Issue #690, opened 2018-10-01
PF4 React Workspace needs a Demo section 


LHinson:
PF4 React Workspace needs a Demo section similar to what is seen on pf-next.com

PR #711, opened 2018-10-04
chore(react-core, react-docs): added ability to add demos to react-docs.

dlabaj:
affects: @patternfly/react-core, @patternfly/react-docs

ISSUES CLOSED: #690

**What**:  Updates to add support to allow demo section for react documentation.

**Additional issues**:  Updated react-core so that the demo section is excluded from the distribution.

### DIFF
--- a/packages/patternfly-4/react-core/.babelrc
+++ b/packages/patternfly-4/react-core/.babelrc
@@ -1,5 +1,6 @@
 {
   "presets": ["../.babelrc.js"],
+  "ignore": ["demos"],
   "env": {
     "production:esm": {
       "plugins": [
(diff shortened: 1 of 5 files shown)


dmiller9911:
I feel like demos should be part of react-docs proper instead of react-core. Since they are not going to be usable components and only for demo/documentation it will reduce the overhead of pulling react-core. I am not super opinionated about this so it might be better to have others weigh in. If it was only in react-docs people would simply add a demo to /pages/demos/{demo}.js

jschuler:
@dmiller9911 Maybe for the same reason that examples got moved into react-core to be easier consumable by the (future) website we should have the demos in react-core? As long as it is not included in the distribution. And we probably want to exclude examples from the distribution as well.

PR #712, opened 2018-10-05
feat(a11y): add basic a11y checks with react-axe

### LINKED ISSUE #603: Enable axe tooling
As a developer, I'd like to have a way to check my work for accessibility violations as I'm developing so that I can resolve issues in real time as they arise. [react-axe](https://github.com/dequelabs/react-axe) looks like a good option and provides a more intimate developer experience as checks are ran on state change as opposed to manual click to run.

Let's integrate with react-axe so that developers can find and address these issues at development time without having to run an entire audit or spot test individual pages manually.

seanforyou23:
fix #603

**What**: Adding automated accessibility checks for PF3 react components

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/603

Is this the best place to install react-axe?

Here are some screenshots that shows what interacting with the tool looks like.

screen shot 2018-10-05 at 9 09 46 am

screen shot 2018-10-05 at 9 10 35 am

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/index.js
+++ b/packages/patternfly-3/patternfly-react/src/index.js
@@ -1,3 +1,11 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+if (process.env.NODE_ENV !== 'production') {
+  const axe = require('react-axe'); // eslint-disable-line global-require
+  axe(React, ReactDOM, 1000);
+}
+
 export { default as controlled } from './common/controlled';
 export * from './common/helpers';
 export * from './components/Alert';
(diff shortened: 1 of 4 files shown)


packages/patternfly-3/patternfly-react/src/index.js
```diff
@@ -1,3 +1,11 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+if (process.env.NODE_ENV !== 'production') {
+  const axe = require('react-axe'); // eslint-disable-line global-require
+  axe(React, ReactDOM, 1000);
+}
+
```
jeff-phillips-18: I'm not sure about adding this here as it could get included by consumers. What about adding it only to storybook in https://github.com/patternfly/patternfly-react/blob/master/storybook/config.js ?

seanforyou23: Ah good call, should have thought of that :)

PR #713, opened 2018-10-05
fix(Alert): Use the correct css class name for screen readers

### LINKED ISSUE #601: PF4: "Danger" prefix in Alert title
The Alert outputs the variant name in the title.

Example:
screen shot 2018-09-05 at 4 53 12 pm

It seems odd to prefix an unlocalizable, "Danger:" string like this. I assume (from accessibleStyles.srOnly) that this is not intended to be seen?

```
const readerTitle = (
    <React.Fragment>
      <span className={css(accessibleStyles.srOnly)}>{capitalize(AlertVariant[variant])}: </span>
      {title}
    </React.Fragment>
  );
```

I don't see srOnly in @patternfly/patternfly-next/utilities/Accessibility/styles.css

boaz1337:
closes #601

react-core is using an old version of react-styles which does not parse class name: "pf-u-something" to "something" but to "uSomething".

To fix this problem updating react-styles to 2.1.x in package.json solves the problem.

### DIFF
--- a/packages/patternfly-4/react-core/package.json
+++ b/packages/patternfly-4/react-core/package.json
@@ -26,7 +26,7 @@
   "homepage": "https://github.com/patternfly/patternfly-react#readme",
   "dependencies": {
     "@patternfly/react-icons": "^2.5.0",
-    "@patternfly/react-styles": "^2.0.0",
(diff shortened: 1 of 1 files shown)


amarie401:
Can you resolve your conflicts @boaz1337 thanks!

Issue #548, opened 2018-08-24
Page Layout 


LHinson:
Introduce PF4 page layout: https://pf-next.com/demos/PageLayout/examples/

In addition to a React review, this issue should be reviewed by the core contributors of this demo: 
* CSS: @andresgalante

jschuler:
@andresgalante Hi Andres, the layout is viewable here:
https://714-pr-patternfly-react-patternfly.surge.sh/patternfly-4/layouts/page

PR #714, opened 2018-10-05
feat(Page layout): Adds the page layout to PF4

jschuler:

WHAT:
Adds the page layout to pf4

ISSUES CLOSED: #548


packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -0,0 +1,113 @@
...
+  isNavOpen,
+  onNavToggle,
+  toolbar,
+  avatar,
+  ...props
+}) => (
+    <React.Fragment>
+      {background}
+      <div {...props} className={css(styles.page, className)}>
+        {/* Added style={{ overflowX: 'unset' }} as a workaround to make dropdowns work */}
```
jschuler: When this issue in pf-next is fixed i can remove it
https://github.com/patternfly/patternfly-next/issues/780

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -0,0 +1,113 @@
...
+                  aria-label="Toggle primary navigation"
+                  variant={ButtonVariant.plain}
+                >
+                  <BarsIcon />
+                </Button>
+              </div>
+            )}
+            <a className={css(styles.pageHeaderBrandLink)}>{logo}</a>
+          </div>
+          {/* Hide for now until we have the context selector component */}
```
jschuler: Should be available for next alpha:
https://github.com/patternfly/patternfly-next/issues/133

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -0,0 +1,113 @@
...
+        </header>
+        {layout === PageLayouts.vertical && (
+          <aside
+            className={css(
+              styles.pageSidebar,
+              isNavOpen && styles.modifiers.expanded,
+              !isNavOpen && styles.modifiers.collapsed
+            )}
+          >
+            {nav}
```
jschuler: Nesting expandable nav inside the `<aside>` has an issue with the chevron icons, tracked here:
https://github.com/patternfly/patternfly-next/issues/779

packages/patternfly-4/react-core/src/layouts/Page/Page.test.js
```diff
@@ -0,0 +1,40 @@
...
+import Page from './Page';
+import PageSection from './PageSection';
+
+const props = {
+  'aria-label': 'Page layout',
+  id: 'PageId',
+  className: 'my-page-class'
+};
+
+test('replace with useful test name', () => {
```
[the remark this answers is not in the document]
jschuler: yep :)

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -0,0 +1,113 @@
...
+  /** Content rendered inside the Page layout */
+  children: PropTypes.node,
+  /** Additional classes added to the Page layout */
+  className: PropTypes.string,
+  /** Page layout type */
+  layout: PropTypes.oneOf(Object.values(PageLayouts)),
+  /** Component to render the background image */
+  background: PropTypes.node,
+  /** Component to render the logo/brand */
+  logo: PropTypes.node,
```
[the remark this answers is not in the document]
jschuler: Could be the Brand component. I am adding a little more descriptive text for these

packages/patternfly-4/react-core/src/layouts/Page/PageSection.d.ts
```diff
@@ -0,0 +1,19 @@
+import { SFC, HTMLProps, ReactNode } from 'react';
+import { OneOf } from '../../typeUtils';
+
+export const PageSectionVariants: {
```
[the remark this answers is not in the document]
jschuler: I am not sure, I was going off of previous definition file examples for this. For example Alert.d.ts. I don't have a typescript workspace set up but what you said makes sense. Maybe we can test these out and update as needed? Or wait until we have typescript support baked in :)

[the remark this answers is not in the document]
amarie401: I would say wait until we have typescript support

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/layouts/Page/page.css';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import { BarsIcon } from '@patternfly/react-icons';
+import { Button, ButtonVariant } from '@patternfly/react-core';
```
dmiller9911: this should be a direct import `'../../components/Button';`

packages/patternfly-4/react-core/src/layouts/Page/Page.js
tlabaj: I wonder if you should add something to the comment so it is clear that a fix is coming and the work around is temporary.

tlabaj: I wonder if we want to add something to the code comment to make it clear that this is a temporary work around.

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -0,0 +1,113 @@
...
+  logo,
+  nav,
+  isNavOpen,
+  onNavToggle,
+  toolbar,
+  avatar,
+  ...props
+}) => (
+    <React.Fragment>
+      {background}
```
tlabaj: I don't think the background belongs in the layout component (that is what is shown in PF-next).  I think when you build the demo, you can add the backgroynd there.

packages/patternfly-4/react-core/src/layouts/Page/Page.test.js
```diff
@@ -0,0 +1,40 @@
...
+const props = {
+  'aria-label': 'Page layout',
+  id: 'PageId',
+  className: 'my-page-class'
+};
+
+test('Check page layout example against snapshot', () => {
+  const PageLogo = <div>Brand</div>;
+  const PageNav = (
+    <ul>
```
andresgalante: you don't really need a ul here, just type "navigation"

tlabaj:
This layout seems more like an actually component rather than a layout.  I would think that it should only layout the header, the aside and the main section.  I would think that the other sub components would be broken into separate components and the consumer would build the page themselves.

dlabaj:
Agree with @tlabaj The layout should only be responsible for laying out the header, aside, and main components.  I believe only main should be required and aside and header should be optional.

jschuler:
@tlabaj @dlabaj All components are optional and they can be anything. Basically just providing placeholders that they can pass "stuff" into

packages/patternfly-4/react-core/src/layouts/Page/Page.js
jschuler: Sounds good!

jschuler:
broke it out a bit, now there is also a PageHeader and PageSidebar component

PR #715, opened 2018-10-05
fix(react-docs deploy): Clean cache

jschuler:
Clear the gatsby cache before build


jschuler:
Looks like issue was that the domain had to be deleted and re-added to another surge account. Closing this PR as it is not necessary :)

PR #717, opened 2018-10-08
feat(checkBox): updated PF4 checkBox component according to PF-next

rvsia:
**What**:

- adds label for checkbox
- adds validation for id, aria-label

1. checkBox must have an id
2. if label is not a string, checkBox must have an aria-label

according to https://pf-next.com/components/Check/examples/

**Link to storybook:**:

https://717-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/checkbox/

**Additional issues**:

How should the component handle different additional attributes for different parts of the component?

e. g. Should we let users to add styles for wrapperDiv, label, input separately?
Right now, a consumer can add props only to input.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.d.ts
@@ -1,4 +1,5 @@
-import { HTMLProps, FormEvent } from 'react';
+import PropTypes from 'prop-types';
+import { HTMLProps, FormEvent, ReactNode } from 'react';
 import { Omit } from '../../typeUtils';
 
 export interface CheckboxProps
@@ -6,6 +7,9 @@ export interface CheckboxProps
   isDisabled?: boolean;
   isValid?: boolean;
   onChange?(checked: boolean, event: FormEvent<HTMLInputElement>): void;
+  id: string;
+  'aria-label': string;
+  label?: ReactNode;
 }
 
 declare const Checkbox: React.SFC<CheckboxProps>;
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
@@ -1,24 +1,31 @@
 import React from 'react';
 import styles from '@patternfly/patternfly-next/components/Check/check.css';
-import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
+import { css, getModifier } from '@patternfly/react-styles';
 
 const propTypes = {
-  /** additional classes added to the Checkbox */
+  /** Additional classes added to the Checkbox. */
   className: PropTypes.string,
   /** Flag to show if the Checkbox selection is valid or invalid. */
   isValid: PropTypes.bool,
   /** Flag to show if the Checkbox is disabled. */
   isDisabled: PropTypes.bool,
   /** A callback for when the Checkbox selection changes. */
-  onChange: PropTypes.func
+  onChange: PropTypes.func,
+  /** Label text of the checkbox. */
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  /** Id of the checkbox. */
+  id: PropTypes.string.isRequired,
+  /** Aria-label of the checkbox. */
+  'aria-label': PropTypes.any.isRequired
 };
 
 const defaultProps = {
   className: '',
   isValid: true,
   isDisabled: false,
-  onChange: () => undefined
+  onChange: () => undefined,
+  label: undefined
 };
 
 class Checkbox extends React.Component {
@@ -27,16 +34,23 @@ class Checkbox extends React.Component {
   };
 
   render() {
-    const { className, onChange, isValid, isDisabled, ...props } = this.props;
+    const { className, onChange, isValid, isDisabled, label, ...props } = this.props;
     return (
-      <input
-        {...props}
-        className={css(styles.check, className)}
-        type="checkbox"
-        onChange={this.handleChange}
-        aria-invalid={!isValid}
-        disabled={isDisabled}
-      />
+      <div className={css(styles.check, className)}>
+        <input
+          {...props}
+          className={css(styles.checkInput)}
+          type="checkbox"
+          onChange={this.handleChange}
+          aria-invalid={!isValid}
+          disabled={isDisabled}
+        />
+        {label && (
+          <label className={css(styles.checkLabel, getModifier(styles, isDisabled && 'disabled'))} htmlFor={props.id}>
+            {label}
+          </label>
+        )}
+      </div>
     );
   }
 }
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.test.js
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.test.js
@@ -8,17 +8,43 @@ const props = {
 };
 
 test('controlled', () => {
-  const view = shallow(<Checkbox checked />);
+  const view = shallow(<Checkbox checked id="check" aria-label="check" />);
   expect(view).toMatchSnapshot();
 });
 
 test('uncontrolled', () => {
-  const view = shallow(<Checkbox />);
+  const view = shallow(<Checkbox id="check" aria-label="check" />);
   expect(view).toMatchSnapshot();
 });
 
 test('isDisabled', () => {
-  const view = shallow(<Checkbox isDisabled />);
+  const view = shallow(<Checkbox id="check" isDisabled aria-label="check" />);
+  expect(view).toMatchSnapshot();
+});
+
+test('label is string', () => {
+  const view = shallow(<Checkbox label="Label" id="check" checked aria-label="check" />);
+  expect(view).toMatchSnapshot();
+});
+
+test('label is function', () => {
+  const functionLabel = () => <h1>Header</h1>;
+  const view = shallow(<Checkbox label={functionLabel()} id="check" checked aria-label="check" />);
+  expect(view).toMatchSnapshot();
+});
+
+test('label is node', () => {
+  const view = shallow(<Checkbox label={<h1>Header</h1>} id="check" checked aria-label="check" />);
+  expect(view).toMatchSnapshot();
+});
+
+test('passing class', () => {
+  const view = shallow(<Checkbox label="label" className="class-123" id="check" checked aria-label="check" />);
+  expect(view).toMatchSnapshot();
+});
+
+test('passing HTML attribute', () => {
+  const view = shallow(<Checkbox label="label" aria-labelledby="labelId" id="check" checked aria-label="check" />);
   expect(view).toMatchSnapshot();
 });
 
@@ -27,7 +53,7 @@ test('checkbox passes value and event to onChange handler', () => {
   const event = {
     currentTarget: { checked: newValue }
   };
-  const view = shallow(<Checkbox {...props} />);
+  const view = shallow(<Checkbox id="check" {...props} aria-label="check" />);
   view.find('input').simulate('change', event);
   expect(props.onChange).toBeCalledWith(newValue, event);
 });
--- a/packages/patternfly-4/react-core/src/components/Checkbox/examples/UncontrolledCheckbox.js
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/examples/UncontrolledCheckbox.js
@@ -4,7 +4,7 @@ import { Checkbox } from '@patternfly/react-core';
 class UncontrolledCheckbox extends React.Component {
   static title = 'Uncontrolled Checkbox';
   render() {
-    return <Checkbox aria-label="uncontrolled checkbox example" />;
+    return <Checkbox label="Uncontrolled CheckBox" aria-label="uncontrolled checkbox example" id="check-2" />;
   }
 }
 
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.docs.js
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.docs.js
@@ -1,12 +1,13 @@
-import { Checkbox } from '@patternfly/react-core';
 import Controlled from './examples/ControlledCheckbox';
 import Uncontrolled from './examples/UncontrolledCheckbox';
 import Disabled from './examples/DisabledCheckbox';
+import Custom from './examples/CustomLabelCheckbox';
+import { Checkbox } from '@patternfly/react-core';
 
 export default {
   title: 'Checkbox',
   components: {
     Checkbox
   },
-  examples: [Controlled, Uncontrolled, Disabled]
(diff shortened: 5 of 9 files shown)


packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -1,24 +1,62 @@
...
   isDisabled: false,
-  onChange: () => undefined
+  onChange: () => undefined,
+  label: undefined,
+  id: null,
+  'aria-label': null
+};
+
+const renderLabel = (label, checkId, isDisabled) => {
+  if (typeof label === 'function') {
```
tlabaj: It looks like we are only applying PF styles when the label is a string.  Can we apply style to this regardless of type? I am concerned that the labels may be unstyled, leading to inconsistencies.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.d.ts
```diff
@@ -1,11 +1,22 @@
...
+};
 
 export interface CheckboxProps
   extends Omit<HTMLProps<HTMLInputElement>, 'type' | 'onChange' | 'disabled'> {
   isDisabled?: boolean;
   isValid?: boolean;
   onChange?(checked: boolean, event: FormEvent<HTMLInputElement>): void;
+  id?: string;
+  'aria-label'?: string;
+  label?: OneOf<typeof TextVariants, keyof typeof TextVariants>;
```
dmiller9911: This is more easily written as `string | ReactNode | Function`. That being said I am curious why function is needed at all. Is there a usecase for it. I understand using render functions, but for a simple label is seems to add complications where they are not needed.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -1,24 +1,62 @@
...
+    if (!props.id) {
+      return new Error('CheckBox must have an id.');
+    }
+    if (typeof props.id !== 'string') {
+      return new Error(`Id must be string. Received ${typeof props.id}.`);
+    }
+    return null;
+  },
+  /** Aria-label of the checkbox. */
+  'aria-label': props => {
```
dmiller9911: If using something like react-intl or react-i18next this will likely always error. I think we should just check for the existence of it since we are allowing nodes in addition to strings.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -1,24 +1,62 @@
...
+  label: undefined,
+  id: null,
+  'aria-label': null
+};
+
+const renderLabel = (label, checkId, isDisabled) => {
+  if (typeof label === 'function') {
+    return label();
+  } else if (typeof label === 'object') {
+    return label;
```
dmiller9911: This would not work if someone actually passed an object here since react cannot render objects. A better check would be `React.isValidElement(label)`, but that would break styles if I was trying to use react-intl. It is a react node that renders a string.

```
<Checkbox 
  label={<FormattedMessage defaultMessage="Some Label" id="someId" />}
/>
```
That will render a string that isn't wrapped by the label. For comparrison this would work correctly in material-ui.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -27,16 +65,19 @@ class Checkbox extends React.Component {
...
+      <div className={css(styles.check, className)}>
+        <input
+          {...props}
+          className={css(styles.checkInput)}
+          type="checkbox"
+          onChange={this.handleChange}
+          aria-invalid={!isValid}
+          disabled={isDisabled}
+        />
+        {label && renderLabel(label, props.id, isDisabled)}
```
dmiller9911: If we get rid of the function (i think we should). Inlining the label down here would be easier to read and follow.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
dmiller9911: Agreed with this. If we really need to support replacing the label component used here (for now just an html label) we should follow something like `labelComponent={CustomLabel}` or `compoents={{ Label: CustomLabel }}` that would then get passed the className, htmlFor, or an other props.

If a basic function is needed the prop should be `renderLabel` and it should be passed the htmlFor and className prop to allow the user access to the styles.

priley86:
*would you mind formatting the commit [semantically](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) ? ... you can use `yarn commit` to walk through the CLI.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
Hyperkid123: > This would not work if someone actually passed an object here since react cannot render objects.

To be absolutely sure that the object is react node i think there should be something like: 
```javascript
typeof label === 'object' && label.$$type === 'React.Node'
``` 
(JS qualifies react nodes ad objects, so if you are not using something like flow types from typescript you need the 'object' compassion)

dmiller9911: `React.isValidElemen`t does this in a more maintainable way: https://github.com/facebook/react/blob/master/packages/react/src/ReactElement.js#L366 This will allow us to not have to make changes based on React internals. the `$$` prefix is React telling consumers that is should be considered private.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
rvsia: I think it (with all of the conditions) would be too complex to put into JSX.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
rvsia: @dmiller9911 

Maybe I don't understand but I can't see any difference between this and your solution except the amount of props. 

If user wants to use a component with props instead of the label he can do something like
`<Checkbox label={<Customcomponent className="..." htmlFor="..." />}`

dmiller9911: I was more indicating that the function support should be removed. It does not add any value. It only complicates things. It should be `node` (covers string, rendered element, number, etc) only. That would support the follow
```jsx
<Checkbox label="<CustomLabel />" />
<Checkbox label="Label" />
const renderLabel =() => 'Some Label';
<Checkbox label={renderLabel()}
```
The only reason we would want to pass a function to label would be to pass the className and any a11y props to it
```jsx
const renderLabel = ({ className, htmlFor }) => (
  <label className={`${className} ${someCustomClass}` } htmlFor={htmlFor}>Label Text</label>
);
<Checkbox label={renderLabel}
```
Otherwise it is no differnet than the last example in the first block.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
dmiller9911: I still do not believe this is resolved. While I mentioned using the check it still does not fix the fact that if I want to render an element which in turn renders only a string. This would render it without being wrapped in our styled `<label>` element.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
amarie401: I agree with getting rid of the function here to simplify this

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
amarie401: I agree with going with `node` here and removing the function support.

amarie401: I agree with going with `node` here and removing the function support.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -1,24 +1,54 @@
...
   isValid: PropTypes.bool,
   /** Flag to show if the Checkbox is disabled. */
   isDisabled: PropTypes.bool,
   /** A callback for when the Checkbox selection changes. */
-  onChange: PropTypes.func
+  onChange: PropTypes.func,
+  /** Label text of the checkbox. */
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.func]),
+  /** Id of the checkbox. */
+  id: props => {
```
jschuler: I think you can just do `id: PropTypes.string.isRequired`

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -1,24 +1,54 @@
...
 };
 
 const defaultProps = {
   className: '',
   isValid: true,
   isDisabled: false,
-  onChange: () => undefined
+  onChange: () => undefined,
+  label: undefined,
+  id: null,
```
jschuler: If this is a string perhaps default to `''`

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
amarie401: Agreed, this would simplify it and give the same error if not specified

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -27,16 +57,19 @@ class Checkbox extends React.Component {
...
-        className={css(styles.check, className)}
-        type="checkbox"
-        onChange={this.handleChange}
-        aria-invalid={!isValid}
-        disabled={isDisabled}
-      />
+      <div className={css(styles.check, className)}>
+        <input
+          {...props}
+          className={css(styles.checkInput)}
```
jschuler: The className from prop is not getting passed through

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.test.js
```diff
@@ -8,17 +8,33 @@ const props = {
...
+  const view = shallow(<Checkbox id="check" isDisabled />);
+  expect(view).toMatchSnapshot();
+});
+
+test('label is string', () => {
+  const view = shallow(<Checkbox label="Label" id="check" checked />);
+  expect(view).toMatchSnapshot();
+});
+
+test('label is function', () => {
```
jschuler: can you add test for making sure a custom class can be passed in, e.g. `class="test-123"` and one for passing a html attribute, e.g. `aria-labelledby="labelId"

packages/patternfly-4/react-core/src/components/Checkbox/examples/UncontrolledCheckbox.js
```diff
@@ -4,7 +4,7 @@ import { Checkbox } from '@patternfly/react-core';
 class UncontrolledCheckbox extends React.Component {
   static title = 'Uncontrolled Checkbox';
   render() {
-    return <Checkbox aria-label="uncontrolled checkbox example" />;
+    return <Checkbox label="Uncontrolled CheckBox" aria-label="uncontrolled checkbox example" id="check-2" />;
```
jschuler: I only see simple strings passed in as labels for the examples. Maybe add another example to pass in custom component

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
rvsia: That's my question from description of this PR. Do we want to pass className to the input or to wrapperDiv? Or should I create classNames props separately for input, label, wrapperdiv?

jschuler: maybe the wrapper div, if user needs to target the input for styling they will be able to, but can also apply style to the wrapper

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -27,16 +34,28 @@ class Checkbox extends React.Component {
...
+          type="checkbox"
+          onChange={this.handleChange}
+          aria-invalid={!isValid}
+          disabled={isDisabled}
+        />
+        {React.isValidElement(label)
+          ? label
```
dmiller9911: this is still an issue in my mind, but maybe @amarie401 or @jschuler can weigh in on if it will cause them any problems. If someone provides
```jsx
import { FormattedMessage } from 'react-intl';

const CustomCheckbox =  (props) => (
  <Checkbox 
    label={<FormattedMessage defaultMessage="Label Value" id="checkbox.label" />}
    {...props}
  />
)
```

The label will go completely unstyled. There is also no way currently to use the label inside this as a component to make sure that I can wrap it any other way. to get around this I would have to do

```jsx
import { injectIntl, defineMessages } from 'react-intl';

const messages = defineMessages({
  label: {
    defaultMessage: 'Label Value',
    id: 'checkbox.label'
  }
})

const CustomCheckbox =  ({ intl, ...props }) => (
  <Checkbox 
    label={intl.formatMessage(messages.label)}
    {...props}
  />
)

const Wrapped = injectIntl(CustomCheckbox);
```

I propose that this return section is just changed to 
```jsx
<label
  className={css(styles.checkLabel, isDisabled && styles.modifiers.disabled)}
  htmlFor={props.id}
>
  {label}
</label>
```
That removes the string and element checks and just passes the node into the label.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.d.ts
```diff
@@ -1,11 +1,15 @@
...
 import { Omit } from '../../typeUtils';
 
 export interface CheckboxProps
   extends Omit<HTMLProps<HTMLInputElement>, 'type' | 'onChange' | 'disabled'> {
   isDisabled?: boolean;
   isValid?: boolean;
   onChange?(checked: boolean, event: FormEvent<HTMLInputElement>): void;
+  id: string;
+  'aria-label': string;
+  label?: string | ReactNode;
```
dmiller9911: ReactNode includes string. This can just be `label?: ReactNode`

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
tlabaj: I agree with Doug here.  I think this leaves room for a label to be unstyled which could lead to an inconsistent user experience.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
jschuler: Yep that's good thanks

jschuler:
Nice job @rvsia, thanks for your contribution!  :tada:

rvsia:
Thanks everyone for helping me! 👍

PR #719, opened 2018-10-08
test(Snapshots): Update snapshots

michaelkro:
**What**:  Update snapshots


jschuler:
Hi @michaelkro, the previous PR that got merged could have only done so if the tests were passing. I think this might be a local workspace issue? I just pulled the latest from master and the tests are passing for me:
Test Suites: 126 passed, 126 total
Tests:       640 passed, 640 total
Snapshots:   490 passed, 490 total
Time:        38.267s, estimated 98s
Ran all test suites.

michaelkro:
@jschuler I'm on latest master as well, and my `yarn.lock` also seems to be not current, weird!  I'll try recloning

amarie401:
Hmm.. yeah I'm thinking that might be the case

michaelkro:
re-cloning fixed it.  I guess just removing `node_modules` would have been a quicker check lol.  Thanks for checking on this for me guys!

PR #725, opened 2018-10-09
fix(CatalogTileView): Fix to not render hidden tiles in catalog tile view categories

jeff-phillips-18:
When a catalog was showing a limited set of tiles, the hidden tiles were still being rendered but not shown due to the height limitation on the container. This allowed the hidden tiles' text to still be considered in the browsers find component.

This PR prevents the category from rendering tiles that are not shown.

@serenamarie125


priley86:
ok - i did not quite follow the View All logic until reviewing on mobile/ipad/desktop. Here is what I'm seeing:

Mobile (iphone): "View All" link shows and items scroll.
iphone

Tablet (ipad): "View All" link does not show and second row is visible:
ipad

Desktop: "View All" link shows and first row is visible:
desktop

PR #727, opened 2018-10-09
fix(Alert): [TS] action type is string while it should've been node

boaz1337:
When adding an action button to Alert, Typescript fails to compile saying action is supposed to be a string.

ts_fix

This commit fixes Alert.d.ts so action accepts ReactNode.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Alert/Alert.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Alert/Alert.d.ts
@@ -1,5 +1,5 @@
 import { SFC, HTMLProps, ReactNode } from 'react';
-import { OneOf } from '../../typeUtils';
+import { OneOf, Omit } from '../../typeUtils';
 
 export const AlertVariant: {
   success: 'success';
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Alert/Alert.d.ts
```diff
@@ -8,9 +8,10 @@ export const AlertVariant: {
   info: 'info';
 };
 
-export interface AlertProps extends HTMLProps<HTMLDivElement> {
+export interface AlertProps extends Omit<HTMLProps<HTMLDivElement>, 'action'> {
   variant: OneOf<typeof AlertVariant, keyof typeof AlertVariant>;
   children?: ReactNode;
+  action?: ReactNode;
```
jschuler: Do you mind adding the other missing props? title, className, 'aria-label' i believe

boaz1337: I didn't add `className`, `title` and `arira-label` because they are coming from `HTMLProps<HTMLDivElement>`.
Should I add them, though?

jschuler: ah that sounds good then, should be fine I think

PR #734, opened 2018-10-10
fix(table-grid): use package import rather than relative path in TableGrid

priley86:
**What**:
I am having issues consuming TableGrid downstream due to the following error:
```
ERROR in ./node_modules/patternfly-react-extensions/dist/esm/components/TableGrid/TableGridHead.js
Module not found: Error: Can't resolve '../../../../patternfly-react/src/components/Tooltip' in '/Users/priley/GitHub/web-ui/frontend/node_modules/patternfly-react-extensions/dist/esm/components/TableGrid'
```

This fixes that by consuming Tooltip from `patternfly-react` instead.

**Additional issues**:


priley86:
you can import directly now @dtaylor113 👍

```
import { OverlayTrigger, Tooltip } from 'patternfly-react';
import { TableGrid } from 'patternfly-react-extensions';
```

Issue #718, opened 2018-10-08
Dropdown component updates


jgiardino:
Several changes were made to the Dropdown component in core (https://github.com/patternfly/patternfly-next/pull/750) that should be reflected here:

- **Dropdown Select** - This is a new variation of the Dropdown that supports the use case of making a selection. The language selector on the login page is a good example of this component being used. Some things to note about this component:
  - The menu items are always `<button>` elements and `aria-pressed` is always present on all of them. The menu item that's currently selected has `aria-pressed="true"`. Otherwise, it should be set to `"false"`
  - An additional `<span>` is included before the `<button>` toggle to provide a visually hidden but screen reader accessible label for the menu. Without seeing this component used in context of different design patterns, I'm not sure if this would be optional or required for accessibility. Maybe we make it optional and present a warning?
  - the menu `<div>` is now a `<ul>` with `<li>` elements that wrap each `<a>` or `<button>`. The `aria-labelledby` that was on the `<div>` is now on the `<ul>` 
  - The `aria-labelledby` attributes are different for this variation. 
    - The button is labelled by the hidden label and itself. 
    - The menu is labelled by the hidden label.

To be implemented in a separate issue/PR:
- **Dropdown Multi-Select** - This is another variation of the Dropdown component that we show in core. There are still some moving parts for this one in core, so we'll add this in a separate issue.

jgiardino:
After further discussion and reviewing wireframe designs from @mcarrano, which are based on the following [example in PF3](https://patternfly-react.surge.sh/patternfly-3/?selectedKind=patternfly-react%2FForms%20and%20Controls%2FFilter&selectedStory=Filter&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs), I wanted to capture some updates to html we need and the expected interaction behavior.

![image](https://user-images.githubusercontent.com/21063328/46823709-e8ba2a00-cd5c-11e8-9716-187d9a1f7609.png)

@mcarrano @tlabaj - let me know if this doesn't match what you were thinking, and if this should be captured in a separate issue.

**Updates needed to the html**

In core, we currently have the following:
```
<div class="pf-c-dropdown pf-m-expanded">
  <span id="dropdown-select-example-expanded-label" hidden>Select a fruit</span>
[...]

PR #736, opened 2018-10-10
feat(PF4-Dropdown): Dropdown keyboard interaction

### LINKED ISSUE #557: Dropdown keyboard interaction
The current implementation of the PF4 Dropdown component needs some enhancements to the keyboard interaction. The following suggestions were noted in the [original PR for Dropdown](https://github.com/patternfly/patternfly-react/pull/506):

- When the menu is visible and the user hits `Esc`
  - Current Behavior: Nothing happens
  - Expected Behavior: menu is hidden and focus is placed on the toggle
- **_updated_** When the menu is visible, the last menu item has focus, and the user hits `Tab`
  - Current Behavior: The next element that can receive focus has focus, and the menu remains visible.
    - This can present usability issues for keyboard-only users who are sighted, since they might not be able to see what has focus if the menu is obscuring it. But we also don't want to automatically close the menu without explicit action, which could also be an issue.
  - ~~Expected Behavior: The next element that can receive focus has focus, and the menu is hidden. Then if the user hits `Shift` + `Tab`, the menu becomes visible again, and the last menu item receives focus.~~
  - Expected Behavior: The first item in the menu that can receive focus receives focus
    - Likewise, when the user is using `Shift` + `Tab` to move up through the menu, when focus leaves the first menu item, it should be placed on the last menu item.
- **_new_** When the toggle has focus and the user hits `Enter`/`Space`:
  - Expected Behavior: The menu displays and the first item in the menu that can receive focus receives focus

**Please note:** There are different variations of a "Dropdown" component. Similar to [Bootstrap](https://getbootstrap.com/docs/4.1/components/dropdowns/), the [PF4 dropdown](https://pf-next.com/components/Dropdown/examples/) is implemented so that anything can be included in the popup section. This means that the elements that receive focus might not be links or buttons, but could be other things like form elements. 

[Additional comments were posted](https://github.com/patternfly/patternfly-react/pull/506#issuecomment-416208197) in response to this suggestion. Any feedback or alternate suggestions are welcome. Our goal is to provide a solution that can accommodate all users, ranging from mouse users, sighted keyboard-only users, and screen reader users.'

michaelkro:
Fixes #557

**What**:
1. Scope `esc` keypress to the dropdown containing the currently focused DOM element
2. Automatically close dropdown if user tabs all the way through the menu (forwards or backwards)
3. Automatically reopen dropdown and focus last menu item under the following conditions.  
    - Keyboard user tabs all the way through the menu (so that it autocloses)
    - Keyboard user uses `shift + tab` to move focus back to the recently closed dropdown

**TODO**:

1.  ~Add type definitions~
2.  ~Add tests~

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
@@ -53,18 +53,35 @@ const defaultProps = {
 
 class Dropdown extends React.Component {
   state = {
-    focusedItemRef: null
+    focusedItemRef: null,
+    focusOnReentry: null
+  };
+
+  componentDidUpdate = (prevProps, prevState) => {
+    if (prevState.focusOnReentry !== this.state.focusOnReentry && !this.state.focusOnReentry) {
+      prevState.focusOnReentry.focus();
+    }
   };
 
   setFocusedItemRef = focusedItemRef => this.setState({ focusedItemRef });
 
-  handleOnFocus = () => this.setState({ focusedItemRef: this.parentRef });
+  handleOnFocus = () => {
+    const { focusOnReentry } = this.state;
+
+    if (focusOnReentry) {
+      this.setFocusedItemRef(focusOnReentry);
+      this.setState({ focusOnReentry: null });
+      return;
+    }
+
+    this.setState({ focusedItemRef: this.parentRef });
+  };
 
   handleOnKeyDown = e => {
     const { focusedItemRef } = this.state;
 
     if (e.key === 'Tab' && !e.shiftKey && focusedItemRef === this.menuRef.lastChild) {
-      this.setState({ focusedItemRef: null });
+      this.setState({ focusedItemRef: null, focusOnReentry: this.menuRef.lastChild });
     }
     if (e.key === 'Tab' && e.shiftKey && focusedItemRef === this.parentRef) {
       this.setState({ focusedItemRef: null });
@@ -94,7 +111,12 @@ class Dropdown extends React.Component {
       >
         {toggle &&
           Children.map(toggle, oneToggle =>
-            cloneElement(oneToggle, { parentRef: this.parentRef, focusedItemRef: this.state.focusedItemRef, isOpen })
+            cloneElement(oneToggle, {
+              parentRef: this.parentRef,
+              focusedItemRef: this.state.focusedItemRef,
+              focusOnReentry: this.state.focusOnReentry,
+              isOpen
+            })
           )}
         <DropdownMenu
           isOpen={isOpen && focusedItemRef}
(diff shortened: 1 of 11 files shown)


jgiardino:
Thanks for working on this! The behavior works as I expect it to. 

What I'm curious about is whether the way we're implementing keyboard interaction here is something that works for other components. 

And also whether we can put a dropdown inside a modal and have the Escape key work as expected, but this might require a separate PR to figure this out.

jeff-phillips-18:
Can we specify clearly what about the dropdown didn't work correctly before? Looking at #557 and the current Dropdown, I see most of this working prior to this PR.

I see a lot of state control being added and wondering how much is really necessary.

michaelkro:
@jgiardino, I can easily remove those attributes, but not sure if it would be best to do that in a separate PR.  I'll let the other reviewers make the call.

@jeff-phillips-18, so this is interesting.  It seems that something about the live examples is affecting interactions with our components.  If you start up your local dev environment with

```sh
GATSBY_LIVE_EXAMPLES=false yarn start:pf4
```

You'll see the behavior #557 mentions, specifically

1.  The dropdown does not automatically collapse when tabbing away
2.  Hitting the esc key will focus the last dropdown example (the kebab example) instead of returning focus to the dropdown that was closed

The "auto-open" on `shift + tab` is a feature that was added to satisfy the requirements of #557

tlabaj:
@jgiardino @michaelkro I ma currently working on  #718

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -63,16 +100,33 @@ class Dropdown extends React.Component {
           isPlain && styles.modifiers.plain,
           direction === DropdownDirection.up && styles.modifiers.top,
           position === DropdownPosition.right && styles.modifiers.alignRight,
-          isOpen && styles.modifiers.expanded,
+          isOpen && focusedItemRef && styles.modifiers.expanded,
           className
         )}
         ref={ref => {
           this.parentRef = ref;
```
dmiller9911: refs should be using the new syntax `React.createRef()` and then they are accessed through `this.parentRef.current` https://reactjs.org/docs/refs-and-the-dom.html#creating-refs we are only supporting the latest react right now so it should be safe to use.

jeff-phillips-18:
@michaelkro So what caused the difference between http://patternfly-react.surge.sh/patternfly-4/components/dropdown and running current locally?

jschuler:
@michaelkro If react-live is causing an issue with your examples, it is also possible to turn it off per example by adding `static live = false;` in the example file. In case this needs to move forward while looking at the react-live interaction issue separately.

dgutride:
Please do not merge this - Jeff and I would like to review the design requirements and implementation with @jgiardino  and @michaelkro before this goes in.

michaelkro:
Closing as we are taking a different approach

PR #743, opened 2018-10-11
feat(Wizard.Pattern): allow hiding cancel and back buttons

suomiy:
Adds a possibility to remove `Cancel` and `Back` buttons from Wizard.Pattern. It is useful for the last final step, if we do not wish to have a  possibility to go back. The last step also has a `Close` button and other buttons would be superfluous (even if disabled).

Additional info:
disscused in https://github.com/kubevirt/web-ui-components/pull/10#issuecomment-428771202

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/Wizard/Patterns/WizardPattern.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Wizard/Patterns/WizardPattern.js
@@ -15,6 +15,8 @@ const WizardPattern = ({
   onBack,
   nextStepDisabled,
   previousStepDisabled,
+  previousStepHidden,
+  cancelButtonHidden,
   title,
   loadingTitle,
   loadingMessage,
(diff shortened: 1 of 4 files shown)


packages/patternfly-3/patternfly-react/src/components/Wizard/Patterns/WizardPattern.js
```diff
@@ -180,6 +199,8 @@ WizardPattern.defaultProps = {
   nextStepDisabled: false,
   previousStepDisabled: false,
   stepButtonsDisabled: false,
+  previousStepHidden: false,
```
suomiy: not sure about the naming: `previousStepHidden` vs `backButtonHidden`

priley86:
from a JS standpoint - no issue from me.

@lizsurette - checking in, are we ok to conditionally hide the back and cancel buttons?

lizsurette:
> @lizsurette - checking in, are we ok to conditionally hide the back and cancel buttons?

@priley86 Yes, I think we would like this for the Results step of the wizard where things have already happened, so there really isn't any going back or cancelling :)

lizsurette:
@suomiy @priley86 - As @serenamarie125 commented above, she followed up with me on the current pattern for the Wizard which just disables these buttons on the optional Results page. Not that I think it's the best design, but I do agre ethat we should stay in line with PatternFly and simply disable them. Sorry for any extra work created here!

suomiy:
Ok, I see. Let's be consistent with the guidelines then.

suomiy:
@lizsurette @serenamarie125

Should I make a PR which disables the cancel button? It is impossible now and is included in the guideline.

jeff-phillips-18:
@suomiy That would be helpful, thanks.

serenamarie125:
Hey @lizsurette yes it does!  Sounds like a great idea, thanks!

PR #744, opened 2018-10-11
fix(background image): fix variable map for background image so when tokens are missing do not assig them

karelhala:

**What**:

There is some nasty bug when core and tokens are out of sync due to wrongly resolved dependencies loading any PF component will throw error and this library can't be used. This fixes such problem in background images.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
+++ b/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
@@ -30,15 +30,15 @@ export const BackgroundImageSrc = {
 };
 
 const variableMap = {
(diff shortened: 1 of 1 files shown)


jschuler:
@dlabrecq looks like the background image examples only work in dev not in prod

karelhala:
@jschuler yeah I noticed that as well, looks like the problem is that gatsby is minifing the name of example file so it tries to access to `/patternfly-4/components/backgroundimage/examples/a` another problem is that on surge base path is `/patternfly-4/` so we end up with `/patternfly-4/patternfly-4/components/backgroundimage/examples/a`

jschuler:
@karelhala i opened #745 so this PR is not blocked as this bug affects other full page URLs as well

Issue #549, opened 2018-08-24
Nav + Page Layout demo 


LHinson:
Introduce nav + page layout demo: https://pf-next.com/demos/PageLayout/examples/

blocked by #547 and #548

Issue #745, opened 2018-10-11
[pf4-react docs] Fix full page urls in react documentation production build


jschuler:
Noticed this issue in #744

Full page URLs are broken in production builds of react documentation. Example:
http://patternfly-react.surge.sh/patternfly-4/components/backgroundimage

The link is `http://patternfly-react.surge.sh/patternfly-4/patternfly-4/components/backgroundimage/examples/a`

Notice that `patterfly-4` appears twice and that the name of the file is `a`.
Presumably that is because gatsby-config for react-docs uses path-prefix `pathPrefix: 'patternfly-4'`
and the gatsby-link component that is used to render the link prefixes it again.

The file name is being picked up as a due to production minification of the name.

jschuler:
This can be reproduced by running `yarn pr-build` in react-docs, then copying the contents of the public folder to `/.public/patternfly-4/. Then uploading the .public folder for example with surge.

jschuler:
Fixed with #749

PR #749, opened 2018-10-12
feat(page layout demos): Add PF4 page layout demos

jschuler:
Closes: #549, #745

**What**:
Adds the page layout demos to PF4
Also rolled in a fix and small improvements to docs


jschuler:
@dlabrecq This PR should fix the full page URLs but I noticed that the background image example cannot find the images since the code is looking for them at the root when it really is under `/patternfly-4/`. So depending on dev/prod build and path-prefixes the assets location can change which the example is not aware of as of yet.

dlabrecq:
Not sure what @dgutride did to support surge, but perhaps we can use that post build process to modify the image paths to include `/patternfly-4?

priley86:
nice work! the expandable navigation demo is particularly appealing to me:
https://749-pr-patternfly-react-patternfly.surge.sh/patternfly-4/demos/pagelayout/examples/page-layout-expandable-nav/

should the expandable nav which is not active have a `ChevronRightIcon`?

jschuler:
Hi @priley86 , yes that is actually a core css bug. I opened an issue for it here
https://github.com/patternfly/patternfly-next/issues/779

PR #754, opened 2018-10-14
feat(Popover): Add PF4 Popover

### LINKED ISSUE #487: Popover component
Popover component from here: https://pf-next.com/components/Popover/examples/

See notes below about leveraging popper.js. This issue also includes identifying the best approach for a wrapper and utilizing core CSS.

Note re: behavior, ensure that popover only closes when the user clicks either the close icon or outside of the popover itself as there are use cases where a user needs to be able to select text from the popover body.

boaz1337:
closes #487


jschuler:
Hi @boaz1337 , thanks for taking a stab at this. Few things, @mcarrano please correct me if I am wrong:
The popover should be able to 'pop' out of the constraints of its parent container. It should basically float on top and not be bounded to or by the layout. So it shouldn't also expand the container it is sitting in to make space for itself when shown.

Also, it's position is dependent on where it is in the DOM. The position prop sets the arrow in the right place, but the popover is not positioned correctly relative to an element (e.g. the button).

Because of this we have suggested incorporating popper.js thidparty library in #487 to help with positioning.

mcarrano:
@jschuler Yes, you are correct. The popover is an element that floats on top of other content and is useful for things like displaying info tips on forms, etc.

boaz1337:
@mcarrano - thanks I will update the props - `header` is no longer required.
@jschuler thanks for the reference, I tried to use `react-popper` but I got a weird error when running the docs. I'm now figuring what is the problem and hopefully update this PR soon.

Thanks!

boaz1337:
I updated the examples + changed the props as suggested.
I am working now on integrating `react-popper` - unfortunately, there documents are not very clear: doing a lot of trial and error.

mcarrano:
Per our discussion, let's remove the Customized Popover variations. We can add these later if a use case comes up that requires them. The Popover without Header requires some styling updates but we will do those in Core first and apply here.

rachael-phillips:
Hi @boaz1337! Are you interesting in continuing your work on this?

boaz1337:
Hi @rachael-phillips 

Sorry that it took me too long - I will be working on this in the next days.

rachael-phillips:
Ok great! Thanks!

boaz1337:
@mcarrano & @christiemolloy  can you PTAL.
Let me know if I'm missing anything.

mcarrano:
@boaz1337 This is looking better, but just a few comments:

* I wasn't sure why the first two examples were different. Can they be combined? It seems like the variables are the position of the popover and whether it has an explicit Close button.
* Besides clicking the close (X) button, a popover should dismiss if the user clicks outside of it.
* The formatting in the third example, headless popover is not quite right. Please refer to the core example here: https://pf-next.netlify.com/components/Popover/examples/

boaz1337:
Updated the Popover examples as you suggested.

However, it seems like the problem with the headless popover comes from `patternfly-next`.
For some reason, `pf-c-popover__body` set `padding-top` to 8 while it should be set to 32.
I will try to figure out how to fix this.

Thanks @mcarrano

tlabaj:
@boaz1337 Core  merged PR  https://github.com/patternfly/patternfly-next/pull [#837](https://github.com/patternfly/patternfly-next/pull/837) that refactored Popover.  Could you update this PR to reflect those changes?

jschuler:
Hi @boaz1337 , every project that has `@patternfly/patternfly-next` in its package.json needs to have the same version number, that includes react-styled-system, react-charts, react-tokens, react-icons, and react-core

jschuler:
Also, in order to have this be ready for consumers we need positioning engine (i.e. popper.js). But this is already great work here! I think we can either a) Merge this but not yet export the components out (in index.js), or b) Include positioning engine.

boaz1337:
Thanks @jschuler I rebased and updated `package.json`.

I am not sure my `popper.js` implementation is good enough. I would rather work on the engine in other PR. Unless, people think differently.

Thanks.

boaz1337:
I updated the examples to match to PF next.
@tlabaj can you please take a look and let me know if I'm missing something.
Thanks a lot.

matthewcarleton:
@christiemolloy yup! we definitely want to consider animation here. We are currently working out how we are going to tackle animation so it likely won't be ready for this PR but we will have this component on this for future updates.

christiemolloy:
@boaz1337 this looks really good thanks for making changes! Will the Close button on the "headless" popover work? 
CSS approved :)

jschuler:
Hey @boaz1337 , I went ahead and merged from master and also removed the popover component from the export list. This way we can merge this code and not expose to the consumer. Once positioning work is ready we can add it back into the index.js files.

jschuler:
@matthewcarleton @tlabaj Are you guys good with the changes?
Please use this URL to view the latest version
http://popover.surge.sh/components/popover

boaz1337:
Thank you so much!

Issue #622, opened 2018-09-14
PF4 Toolbar demo


LHinson:
Add the PF4 Toolbar demo: https://pf-next.com/demos/Toolbar/examples/

Note: there is still a to-do to update the toolbar demo on core with the select component once that has been introduced. See related issues: 
* Add select: https://github.com/patternfly/patternfly-next/issues/708
* Update toolbar with select: https://github.com/patternfly/patternfly-next/issues/709

In addition to a React review, this issue should be reviewed by the core contributors of this demo: 
* CSS: @christiemolloy 
* Design: @mcarrano

amarie401:
The PR for this issue has been merged. See #755

PR #755, opened 2018-10-15
docs(react-core): Updated with toolbar demo

dlabaj:
#622

**What**:

Added toolbar demo for PF4 toolbar layout.

**Additional issues**:


packages/patternfly-4/react-core/src/demos/Toolbar/examples/ComplexToolbarDemo.js
```diff
@@ -0,0 +1,115 @@
...
+          <ToolbarItem>
+            <Button variant="plain" aria-label="Action 1">
+              Action
+            </Button>
+          </ToolbarItem>
+          <ToolbarItem>
+            <Button aria-label="Action 2">Action</Button>
+          </ToolbarItem>
+          <ToolbarItem>
+            <Button variant="plain" aria-label="Kebab">
```
amarie401: Perhaps also add the functionality for the kebab dropdown as well?

packages/patternfly-4/react-core/src/demos/Toolbar/examples/SimpleToolbarDemo.js
```diff
@@ -0,0 +1,109 @@
...
+          <ToolbarItem>
+            <Button variant="plain" aria-label="Action 1">
+              Action
+            </Button>
+          </ToolbarItem>
+          <ToolbarItem>
+            <Button aria-label="Action 2">Action</Button>
+          </ToolbarItem>
+          <ToolbarItem>
+            <Button variant="plain" aria-label="Kebab">
```
amarie401: Same here with perhaps the functionality for the kebab dropdown

packages/patternfly-4/react-core/src/demos/Toolbar/examples/ComplexToolbarDemo.js
dlabaj: Makes sense thanks.  I'll go ahead and add this.

dlabaj:
@jschuler Will add the styles that we discussed today.

packages/patternfly-4/react-core/src/demos/Toolbar/examples/SimpleToolbarDemo.js
dlabaj: Will do.

tlabaj:
I agree with Amiee's comments above. Otherwise looks good with the components that are currently available.  It might be nice to iterate on this as more components become available to make the demo more meaningful.

mcarrano:
@dlabaj This seems like a good start. I agree that we can add more variations of this later when we have additional things to add here. But I just had a couple of questions/comments:

- All icon buttons should have hover states. I don't see these in the core demo either, so maybe that's something that was overlooked? @kybaker or @mceledonia can you take a look and comment?
- This should apply a split layout so that actions are right aligned. I think that's what's being called a Level, but not sure. See the core demo layout.
- The Kabob item is just a type of dropdown to shouldn't I see a menu when I click on that? Was the kabob implemented in React as a dropdown component?

mcarrano:
Thanks for adding the Kebab menu @dlabaj . The only problem I see with this is that the component should not have a border. The toolbar should look like the Core demo here: https://pf-next.netlify.com/demos/Toolbar/examples/  Note that the Kebob is just an icon button that drops down a menu. This looks to be an issue with the way the Kebob menu is implemented in the Dropdown component. There's a border that shouldn't be there.

dlabaj:
@mcarrano If the kebab isn't suppose to have a border as part of the component I can fix it in this PR.  Also I'm going to add the styles for the alignment you brought up now.

mcarrano:
Thanks @dlabaj 

@tlabaj and @LHinson just wanted to make you aware as this relates to my comment on https://github.com/patternfly/patternfly-react/pull/788

jschuler:
I think you can add isPlain prop to the Dropdown to remove the border

dlabaj:
@mcarrano @tlabaj @jschuler @amarie401 There's an isPlain option on the dropdown that changes the dropdown button varriant to "plain".  Missed this the first time I was looking for varriant on the dropdown like we have on the button.  Feel that isPlain should be changed varriant to be consistent with button, even if we don't support all the varriants as button.  Thoughts????

tlabaj:
@mcarrano @jschuler @amarie401 
I just talked to @LHinson.  Since further discussion is needed about the "isPlain" attribute, we can open a separate issue for that.  
@dlabaj can set the "isPlain"attribute on the Dropdown prop in the Toolbar demo to remove the boarder from the kebab for now.

dlabaj:
@mcarrano Updated the styles with the review comments.  Only thing not added is hover state since it's not in core either.  Figured it would be fixed there.

mcarrano:
@dlabaj Thanks for making these changes. Looks great. I agree that the icon hover issue is something to address separately.

PR #761, opened 2018-10-15
feat(component): introduce the pf4 switch component

### LINKED ISSUE #728: PF4 Switch Component
Add PF4 Switch Component: https://pf-next.com/components/Switch/examples/
In addition to a React review, this issue should be reviewed by the core contributors of this demo:
visuals: @maryshak1996

guillaumevincent:
fix #728

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.d.ts
@@ -1,0 +1,16 @@
+import { HTMLProps, FormEvent, ReactNode } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface SwitchProps extends Omit<HTMLProps<HTMLInputElement>, 'type' | 'onChange' | 'disabled' | 'label'> {
+  isDisabled?: boolean;
+  isChecked?: boolean;
+  onChange?(checked: boolean, event: FormEvent<HTMLInputElement>): void;
+  id?: string;
+  'aria-label': string;
+  label?: string;
+  className?: string;
+}
+
+declare const Switch: React.SFC<SwitchProps>;
+
+export default Switch;
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.docs.js
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.docs.js
@@ -1,0 +1,18 @@
+import { Switch } from '@patternfly/react-core';
+import SimpleSwitch from './examples/SimpleSwitch';
+import NoLabelSwitch from './examples/NoLabelSwitch';
+import DisabledSwitch from './examples/DisabledSwitch';
+import UncontrolledSwitch from './examples/UncontrolledSwitch';
+
+export default {
+  title: 'Switch',
+  components: {
+    Switch
+  },
+  examples: [
+    { component: SimpleSwitch, title: 'Simple Switch' },
+    { component: NoLabelSwitch, title: 'Switch with no labels' },
+    { component: DisabledSwitch, title: 'Disabled Switch' },
+    { component: UncontrolledSwitch, title: 'Uncontrolled Switch' }
+  ]
+};
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.js
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Switch/switch.css';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import { getUniqueId } from '../../internal/util';
+
+const propTypes = {
+  /** id for the label. */
+  id: PropTypes.string,
+  /** Additional classes added to the Switch */
+  className: PropTypes.string,
+  /** Text value for the label */
+  label: PropTypes.string,
+  /** Flag to show if the Switch is checked. */
+  isChecked: PropTypes.bool,
+  /** Flag to show if the Switch is disabled. */
+  isDisabled: PropTypes.bool,
+  /** A callback for when the Switch selection changes. (isChecked, event) => {} */
+  onChange: PropTypes.func,
+  /** Adds accessible text to the Switch. */
+  'aria-label': props => {
+    if (!props.id && !props['aria-label']) {
+      return new Error('Switch requires either an id or aria-label to be specified');
+    }
+  }
+};
+
+const defaultProps = {
+  id: '',
+  className: '',
+  label: '',
+  isChecked: true,
+  isDisabled: false,
+  onChange: () => undefined,
+  'aria-label': ''
+};
+
+class Switch extends React.Component {
+  id = this.props.id || getUniqueId();
+
+  render() {
+    const { id, className, label, isChecked, isDisabled, onChange, ...props } = this.props;
+    return (
+      <label className={css(styles.switch, className)} htmlFor={this.id}>
+        <input
+          {...props}
+          id={this.id}
+          className={css(styles.switchInput)}
+          type="checkbox"
+          onChange={event => onChange(event.currentTarget.checked, event)}
+          checked={isChecked}
+          disabled={isDisabled}
+        />
+        <span className={css(styles.switchToggle)} />
+        {label && (
+          <span className={css(styles.switchLabel)} aria-hidden="true">
+            {label}
+          </span>
+        )}
+      </label>
+    );
+  }
+}
+
+Switch.propTypes = propTypes;
+Switch.defaultProps = defaultProps;
+
+export default Switch;
--- a/packages/patternfly-4/react-core/src/components/index.js
+++ b/packages/patternfly-4/react-core/src/components/index.js
@@ -19,6 +19,7 @@ export * from './Nav';
 export * from './Progress';
 export * from './Radio';
 export * from './Select';
+export * from './Switch';
 export * from './TextArea';
 export * from './TextInput';
 export * from './Title';
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.test.js
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.test.js
@@ -1,0 +1,59 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Switch from './Switch';
+
+const props = {
+  onChange: jest.fn(),
+  checked: false
+};
+
+test('switch label for attribute equals input id attribute', () => {
+  const view = shallow(<Switch id="foo" />);
+  expect(view.find('input').prop('id')).toBe('foo');
+  expect(view.find('label').prop('htmlFor')).toBe('foo');
+});
+
+test('switch label id is auto generated', () => {
+  const view = shallow(<Switch aria-label="..." />);
+  expect(view.find('input').prop('id')).toBeDefined();
+});
+
+test('switch is checked', () => {
+  const view = shallow(<Switch id="switch-is-checked" label="On" isChecked />);
+  expect(view).toMatchSnapshot();
+});
+
+test('switch is not checked', () => {
+  const view = shallow(<Switch id="switch-is-not-checked" label="Off" isChecked={false} />);
+  expect(view).toMatchSnapshot();
+});
+
+test('no label switch is checked', () => {
+  const view = shallow(<Switch id="no-label-switch-is-checked" isChecked />);
+  expect(view).toMatchSnapshot();
+});
+
+test('no label switch is not checked', () => {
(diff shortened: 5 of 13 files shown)


guillaumevincent:
I'm waiting for the switch component to be released in patternfly-core
We need to discuss about the API for the switch component (`oneLabel`, `noLabel`, `labelOff`, `labelOn`, etc)

amarie401:
cc: @maryshak1996

maryshak1996:
@guillaumevincent on my end, it isn't showing the element (must not be available in core yet), so I'll check back to review visuals frequently to keep things moving :)

LHinson:
This was a convo on slack but adding here to make sure everyone saw.... 

> the pf-next commit message didn't include a space after fix(topic):asdfdsfads  therefore travis is saying there are no changes that should be included and it didn't kick off a release. 

Once another proper commit is made to core, the switch should be available with that release. @dgutride will be adding stricter rules to help enforce proper commit messages and improve the releases.

guillaumevincent:
I don't know why PatternFly-React preview is not updated, but this PR is ready for a review

packages/patternfly-4/react-core/src/components/Switch/Switch.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Switch/switch.css';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+
+const propTypes = {
+  /** id for the label. */
+  id: PropTypes.string.isRequired,
```
jschuler: since this id is only used to glue the label and the input together for accessibility purposes, i think we can generate it instead and make it optional. Example here:
https://github.com/patternfly/patternfly-react/blob/a331add0ca90f5d3feef9c4752651c957535f219/packages/patternfly-4/react-core/src/components/Nav/NavGroup.js#L27

packages/patternfly-4/react-core/src/components/Switch/Switch.js
```diff
@@ -0,0 +1,63 @@
...
+  render() {
+    const { id, className, labelLeft, labelRight, isChecked, isDisabled, onChange, ...props } = this.props;
+    return (
+      <label className={css(styles.switch, className)} htmlFor={id}>
+        <input
+          {...props}
+          id={id}
+          className={css(styles.switchInput)}
+          type="checkbox"
+          onChange={onChange ? this.handleChange : null}
```
jschuler: Typically, event should be first param sent back. But looking through some other components it appears that we are not consistent with that.
Also could probably remove the extra function and change to:
`onChange={(event) => onChange(event.currentTarget.checked, event)}`

packages/patternfly-4/react-core/src/components/Switch/Switch.js
```diff
@@ -0,0 +1,63 @@
...
+          checked={isChecked}
+          disabled={isDisabled}
+        />
+        {labelLeft ? <span className={css(styles.switchLabel, styles.modifiers.off)}>{labelLeft}</span> : null}
```
jschuler: Instead of ternary operator can to `{labelLeft && <span...>`

packages/patternfly-4/react-core/src/components/Switch/Switch.js
```diff
@@ -0,0 +1,63 @@
...
+        <span className={css(styles.switchToggle)} />
+        {labelRight ? <span className={css(styles.switchLabel, styles.modifiers.on)}>{labelRight}</span> : null}
```
jschuler: same comment about ternary operator

packages/patternfly-4/react-core/src/components/index.js
```diff
@@ -18,6 +18,7 @@ export * from './Nav';
 export * from './Progress';
 export * from './Radio';
 export * from './Select';
+export * from './Switch';
```
jschuler: Also add to src/components/index.d.ts

guillaumevincent:
@jschuler just for your information you can propose change directly in the PR
See https://twitter.com/github/status/1054794897448939520

jschuler:
Looks like you just need to regenerate the snapshots

guillaumevincent:
@christiemolloy I probably miss something but the Switch Disabled is in [pf-next](https://pf-next.com/components/Switch/examples/) and in my [PR](https://761-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/switch)

guillaumevincent:
@jschuler I rebase this patch on top of https://github.com/patternfly/patternfly-react/pull/830
I prefer fixing the getUniqueId consistency issue first, then creating the new Switch component.

christiemolloy:
@guillaumevincent that sounds great, thanks for making that change.

jschuler:
@guillaumevincent Yes we can deal with params order (event first) in separate PR

dmiller9911:
The uncontrolled vs controlled is react specific. I would prefer to keep it. You can read more here. https://reactjs.org/docs/uncontrolled-components.html

guillaumevincent:
I think this one is the good one :smile:

packages/patternfly-4/react-core/src/components/Switch/Switch.d.ts
```diff
@@ -0,0 +1,17 @@
+import { HTMLProps, FormEvent, ReactNode } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface SwitchProps extends Omit<HTMLProps<HTMLInputElement>, 'type' | 'onChange' | 'disabled' | 'label'> {
+  isDisabled?: boolean;
+  isChecked?: boolean;
```
dlabaj: Feel like this should be value to be consistent with PF3, and the other components in PF4.

dlabaj:
@jschuler @tlabaj @christiemolloy Are we no longer supporting tristate?

packages/patternfly-4/react-core/src/components/Switch/Switch.d.ts
dlabaj: Also if we are going to support tri-state we will need to use something other than a boolean.  Maybe constants?

dlabaj: http://patternfly-react.surge.sh/patternfly-3/index.html?knob-Size=normal&knob-On%20Color=primary&knob-Off%20Color=default&knob-animate=true&selectedKind=patternfly-react%2FWidgets%2FSwitch&selectedStory=Switch&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

jgiardino:
Hi @guillaumevincent,

I tested these in JAWS, NVDA and VO. These examples work as expected with one minor change needed.

Where the following elements are included, they result in additional stops/keypresses for the screen reader user (in JAWS or NVDA):
`<span class="pf-c-switch__label pf-m-off">Off</span>`
`<span class="pf-c-switch__label pf-m-on">On</span>`

Having these labels announced doesn't provide any benefit to screen reader users, and is potentially confusing.

We can have these screen readers ignore these elements by added `aria-hidden="true"`:
`<span class="pf-c-switch__label pf-m-off" aria-hidden="true">Off</span>`
`<span class="pf-c-switch__label pf-m-on" aria-hidden="true">On</span>`

This will result in a single keypress to access the control, and when it receives focus, the control is announced with the aria-label given to the checkbox, and the state of the checkbox selection. 

BTW, here's the issue I created to capture this for core: https://github.com/patternfly/patternfly-next/issues/858

mcarrano:
I thought that at some point we decided that we did not actually need two labels on either side of the switch like I'm seeing here. But I don't see that change either here or in the Core version. @mceledonia can you remind me of what we agreed to there? Had you done some visual designs for this?

packages/patternfly-4/react-core/src/components/Switch/Switch.d.ts
tlabaj: I just looked at PF3.  It looks like it uses react-bootstrap-switch for the implementation.  react-bootstrap-switch allows for the consumer to turn triState on.  That allows for a state of null or the indeterminate state.  I think the question there are 2 questions here
1) Do rename the prop to "value" to be consistent with PF3
2) Will we be allowing for the indeterminate state to align with PF3.  If we do support that we need to add another prop to turn it o/ off like react-bootsrap-switch does.

guillaumevincent: patternfly-react should be the react implementation of patternfly-core.
So if triState is not in patternfly-core, it should not be here

for me you should open an issue on pf-core first

guillaumevincent:
@mcarrano @tlabaj @dlabaj this pull request is here to add a Switch React component based on the pf-core implementation.
IMO this is not the right place to discuss about enhancement or update about the Switch component. I think we should open an issue on patternfly-next and update the core Switch component. Then we will update the React component.

packages/patternfly-4/react-core/src/components/Switch/Switch.js
```diff
@@ -0,0 +1,64 @@
...
+  /** Text value for the label on the right */
+  labelRight: PropTypes.string,
+  /** Flag to show if the Switch is checked. */
+  isChecked: PropTypes.bool,
+  /** Flag to show if the Switch is disabled. */
+  isDisabled: PropTypes.bool,
+  /** A callback for when the Switch selection changes. (isChecked, event) => {} */
+  onChange: PropTypes.func,
+  /** Adds accessible text to the input. */
+  'aria-label': PropTypes.string
```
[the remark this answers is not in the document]
guillaumevincent: Nice catch I will do this

guillaumevincent:
@jgiardino I added a PR on pf-core
I will update this PR on monday
Thanks

mcarrano:
That's fair @guillaumevincent .  I will consult with Visuals and open a new issue if we decide that the Switch labeling should be updated in both places. 

I was going to mark this as "UX Approved" but looks like I don't have edit rights on this repo. @LHinson is that something you can enable?

guillaumevincent:
@mcarrano I fix the label in pf-next
I propose to wait for https://github.com/patternfly/patternfly-next/pull/877 to be merged
wdyt?

mcarrano:
Yes, makes sense @guillaumevincent .

rachael-phillips:
@dlabaj @mcarrano @tlabaj patternfly/patternfly-next#877 merged.  Is this one ready for merge?

mcarrano:
@rachael-phillips we had some discussion about the Switch component in the Accessibility meeting today and may be making some changes on the Core side. Opened a new issue here: https://github.com/patternfly/patternfly-next/issues/914  We could merge this and create a new React issue or hold on this until core changes are made.

dgutride:
We should let this in - @guillaumevincent  - please make sure the build is passing and then it can be merged.

rachael-phillips:
Thanks so much, Matt and Dana!

packages/patternfly-4/react-core/src/components/Switch/Switch.docs.js
```diff
@@ -0,0 +1,13 @@
...
+import NoLabelSwitch from './examples/NoLabelSwitch';
+import DisabledSwitch from './examples/DisabledSwitch';
+import UncontrolledSwitch from './examples/UncontrolledSwitch';
+
+export default {
+  title: 'Switch',
+  components: {
+    Switch
+  },
+  examples: [SimpleSwitch, NoLabelSwitch, DisabledSwitch, UncontrolledSwitch]
```
jschuler: This was changed recently, static fields should be moved out of the examples and should now be
```suggestion
  examples: [{ component: SimpleSwitch, title: 'Simple Switch' }, { component: NoLabelSwitch, title: 'Switch with no labels' }, { component: DisabledSwitch, title: 'Disabled Switch' }, { component: UncontrolledSwitch, title: 'Uncontrolled Switch' }]
```

guillaumevincent: Thanks

christiemolloy:
@jgiardino do you know the status of the switch component. Was @mceledonia going to make visual updates? In this react example the text is "on/off" should we keep this wording or switch to "label" like we have in core.

jgiardino:
If this is ready to merge then I'd vote to handle the visual updates (https://github.com/patternfly/patternfly-next/issues/914) in a later PR since they haven't been implemented in core yet.

Let's keep On/Off here. There's also an issue in core to handle toggling the label (https://github.com/patternfly/patternfly-next/issues/915). I suggested we change the label to "On" and "Off" in that issue. So let's keep "On"/"Off" in this one, and core will match when that issue is done.

jschuler:
@jgiardino Sounds good, looks like this is good to merge and there are some followup items in the 2 issues you referenced.

PR #763, opened 2018-10-15
feat(CatalogTile): Use a link for tiles with href or onClick props

jeff-phillips-18:
Allows applications to pass `href` or `onClick` attributes and if given creates the tile as an `<a>` for accessibility.

https://763-pr-patternfly-react-patternfly.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react-extensions%2FCatalog%20Components%2FCatalog%20Tile&selectedStory=Catalog%20Tile

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/less/catalog-tile.less
+++ b/packages/patternfly-3/patternfly-react-extensions/less/catalog-tile.less
@@ -1,6 +1,7 @@
 .catalog-tile-pf {
   background: @card-pf-bg-color;
   border: 1px solid rgba(209, 209, 209, 0.75);
+  color: inherit;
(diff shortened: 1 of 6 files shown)


priley86:
Fine w/ the fix here... did a quick review w/ aXe and it looks like the `<a>` wrapping the tile is respected and tab accessible here. It does appear that the "provided by Red Hat" text is having some contrast issues though. Maybe we can fix those in a future PR?

screen shot 2018-10-15 at 10 17 52 am

jeff-phillips-18:
Yeah, I'd like to have that looked at in a future PR to fix a bunch of these types of warnings. May need to go back to UX/Visual designers

PR #768, opened 2018-10-15
fix(Storybook): Fix horizontal masthead examples using iconImg instead of titleImg

### LINKED ISSUE #767: Storybook examples of Horizontal Masthead are using iconImg where they should use titleImg
After #758 was merged, the brand image and title text began to both render in these examples:

screenshot 2018-10-15 17 00 58

screenshot 2018-10-15 17 28 07

These examples place a wide brand image into the `iconImg` prop, switching it to `titleImg` fixes this. I'll put up a PR shortly.

mturley:
Fix #767

**What**: See #767. Our horizontal nav examples look broken because both a title image and title text are being rendered, due to the use of the wrong prop.

**Additional issues**:


mturley:
@priley86 @jeff-phillips-18

priley86:
good catch @mturley! you are definitely more familiar w/ these components. Thanks for following up!

PR #758, opened 2018-10-15
Show Masthead title

jhellar:
**What**:

This change fixes issue, where title is not shown when provided to `Masthead` component.

**Additional issues**:

https://github.com/patternfly/patternfly-react/issues/757

PR #772, opened 2018-10-16
fix(Masthead): fix Horizontal Menu example in Storybook

jhellar:
**What**:

Fixes storybook example issue introduced with https://github.com/patternfly/patternfly-react/pull/758 - example shows both the image and title. Fixed by changing `iconImg` property to `titleImg` property as suggested in the above PR.


jhellar:
Duplicates https://github.com/patternfly/patternfly-react/pull/768 -> closing.

Issue #782, opened 2018-10-16
PF4: Progress TypeError: has no exported member 'ProgressSize'


dlabrecq:
Attempted to pull the latest from PF4 react-core (v1.23.0), pf-next (1.0.57 and 1.0.58) and encountered the following build error. This prevents koku-ui from rendering.

![screen shot 2018-10-16 at 11 31 19 am](https://user-images.githubusercontent.com/17481322/47028267-0b6e8900-d137-11e8-8067-42e1099e86ee.png)

In addition, it appears that only 'xs', 'md', and 'lg' are known to the component.

Usage:
```
import { Progress, ProgressSize } from '@patternfly/react-core';
...
<Progress value={Number(percent.toFixed(2))} title="Descriptive text here" size={ProgressSize.sm} />
```

karelhala:
Thanks for reporting this, it looks like I made a mistake when exporting these props in typescript, in meantime you can use it without ProgressSize
```TSX
import { Progress } from '@patternfly/react-core';
...
<Progress value={Number(percent.toFixed(2))} title="Descriptive text here" size="sm" />
```

dlabrecq:
My IDE tells me that 'sm' is not valid -- only 'xs', 'md', and 'lg' are available.

dlabrecq:
I took the latest react-core v1.24.0 package and get this error now.

![screen shot 2018-10-16 at 2 57 23 pm](https://user-images.githubusercontent.com/17481322/47040271-e0def900-d153-11e8-8f2c-e9c6917b763e.png)

That said, this latest react-core package is causing a lot more errors. I had to drop back to v1.23.1 just to compile.

![screen shot 2018-10-16 at 2 59 51 pm](https://user-images.githubusercontent.com/17481322/47040531-71b5d480-d154-11e8-89ef-ede49687be49.png)

karelhala:
Yeah, sorry. For the small variant you should use `xs`
```TSX
import { Progress, ProgressSize } from '@patternfly/react-core';
...
<Progress value={Number(percent.toFixed(2))} title="Descriptive text here" size={ProgressSize.xs} />
```

dlabrecq:
The ProgressSize issue has been resolved, but still seeing the build errors above with the latest.

Issue #770, opened 2018-10-15
Typescript definitions out of sync


priley86:
I noted the following while importing PF4 React components downstream today:

```
ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/BackgroundImage/BackgroundImage.d.ts(16,18):
TS2430: Interface 'BackgroundImageProps' incorrectly extends interface 'HTMLProps<HTMLDivElement>'.
  Types of property 'src' are incompatible.
    Type 'string | { lg?: string; md?: string; md2x?: string; sm?: string; sm2x?: string; xl?: string; xs?:...' is not assignable to type 'string'.
      Type '{ lg?: string; md?: string; md2x?: string; sm?: string; sm2x?: string; xl?: string; xs?: string; ...' is not assignable to type 'string'.

ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/Checkbox/Checkbox.d.ts(1,8):
TS1192: Module '"/Users/priley/GitHub/web-ui/frontend/node_modules/@types/prop-types/index"' has no default export.

ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/Checkbox/Checkbox.d.ts(5,18):
TS2430: Interface 'CheckboxProps' incorrectly extends interface 'Pick<HTMLProps<HTMLInputElement>, "default" | "max" | "required" | "media" | "hidden" | "cite" | ...'.
  Types of property 'label' are incompatible.
    Type 'ReactNode' is not assignable to type 'string'.
      Type 'number' is not assignable to type 'string'.

ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/Progress/index.d.ts(1,73):
TS2305: Module '"/Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/Progress/Progress"' has no exported member 'ProgressStatus'.
Version: typescript 2.9.2
Time: 17023ms
   4875 modules
 98% after emitting

```

dgutride:
Closing via #785, #800 and #806

PR #785, opened 2018-10-16
fix(Progress): missing ProgressSize added to exported types

karelhala:
Resolves: #782

**What**:
Wrong attribute was exported as ProgressSize and that made it impossible to use Progress. This PR fixes such issue by exporting correct attribute.

**Additional issues**:


priley86:
this partially fixes #770 as well ;)

Issue #787, opened 2018-10-16
Accessibility updates for Dropdown component


tlabaj:
Several changes were made to the Dropdown component in core (https://github.com/patternfly/patternfly-next/pull/750) that should be reflected here:

- For all instances of this component:
  - remove `aria-haspopup` - After further review of this attribute, it causes issues in JAWS. 
  - Update the `id` attribute to match core. Currently the react component has `id` assigned to `pf-c-dropdown` but it should be assigned to `pf-c-dropdown__toggle`
- **basic Dropdown** - We now have a basic generic Dropdown structure that can accept anything in the menu `.pf-c-dropdown__menu` 
  - remove `aria-labelledby` on the menu `<div>` (for this variation only)
- **Dropdown with links and actions** - This is the version of the Dropdown that's currently implemented. This component is specific to the use case of providing a set of actions/links in a dropdown.
  - remove `aria-haspopup`, as noted above
  - the menu `<div>` is now a `<ul>` with `<li>` elements that wrap each `<a>` or `<button>`. The `aria-labelledby` that was on the `<div>` is now on the `<ul>`

amarie401:
The associated PR for this issue has been merged so I will close this. See https://github.com/patternfly/patternfly-react/pull/788

PR #788, opened 2018-10-16
refactor(patternfly-4/react-core): Updated the Dropdown component to …

tlabaj:
…match accessibility updates in core.

#787

**What**: Updated the Dropdown component to match accessibility updates in core . Also Updated Dropdown to use list for Action Menu and to use a div for the Basic Dropdown.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
@@ -1,16 +1,7 @@
 import { SFC, HTMLProps, ReactType, ReactNode } from 'react';
 import { OneOf } from '../../typeUtils';
 import { DropdownToggleProps } from './DropdownToggle';
-
-export const DropdownPosition: {
-  right: 'right';
-  left: 'left';
-}
-
-export const DropdownDirection: {
-  up: 'up';
-  down: 'down';
-}
+import { DropdownPosition, DropdownDirection } from './dropdownConstants';
 
 export interface DropdownProps extends HTMLProps<HTMLDivElement> {
   children?: ReactNode;
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
@@ -16,8 +16,6 @@ export const DropdownDirection = {
 };
 
 const propTypes = {
-  /** HTML ID of dropdown element */
-  id: PropTypes.string,
   /** Anything which can be rendered as dropdown items */
   children: PropTypes.node,
   /** Classess applied to root element of dropdown */
@@ -37,11 +35,6 @@ const propTypes = {
 };
 
 const defaultProps = {
-  id:
-    new Date().getTime() +
-    Math.random()
-      .toString(36)
-      .slice(2),
   toggle: null,
   children: null,
   className: '',
@@ -54,11 +47,10 @@ const defaultProps = {
 
 class Dropdown extends React.Component {
   render() {
-    const { className, children, isOpen, toggle, direction, onSelect, isPlain, position, id, ...props } = this.props;
+    const { className, children, isOpen, toggle, direction, onSelect, isPlain, position, ...props } = this.props;
     return (
       <div
         {...props}
-        id={id}
         className={css(
           styles.dropdown,
           isPlain && styles.modifiers.plain,
@@ -74,7 +66,11 @@ class Dropdown extends React.Component {
         {toggle && Children.map(toggle, oneToggle => cloneElement(oneToggle, { parentRef: this.parentRef, isOpen }))}
         {isOpen && (
           <FocusTrap>
-            <DropdownMenu isOpen={isOpen} aria-labelledby={id} onClick={event => onSelect && onSelect(event)}>
+            <DropdownMenu
+              isOpen={isOpen}
+              aria-labelledby={toggle ? toggle.props.id : ''}
+              onClick={event => onSelect && onSelect(event)}
+            >
               {children}
             </DropdownMenu>
           </FocusTrap>
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
@@ -2,7 +2,6 @@ import React, { Children, cloneElement } from 'react';
 import styles from '@patternfly/patternfly-next/components/Dropdown/dropdown.css';
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
-// import FocusTrap from 'focus-trap-react';
 import DropdownMenu from './DropdownMenu';
 
 export const DropdownPosition = {
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
@@ -3,16 +3,7 @@ import styles from '@patternfly/patternfly-next/components/Dropdown/dropdown.css
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import DropdownMenu from './DropdownMenu';
-
-export const DropdownPosition = {
-  right: 'right',
-  left: 'left'
-};
-
-export const DropdownDirection = {
-  up: 'up',
-  down: 'down'
-};
+import { DropdownPosition, DropdownDirection } from './dropdownConstants';
 
 // seed for the aria-labelledby ID
 let currentId = 0;
@@ -27,7 +18,7 @@ const propTypes = {
     }
     return null;
   },
-  /** Classess applied to root element of dropdown */
+  /** Classes applied to root element of dropdown */
   className: PropTypes.string,
   /** Array of DropdownItem nodes that will be rendered in the dropdown Menu list */
   dropdownItems: PropTypes.array,
@@ -87,7 +78,6 @@ class Dropdown extends React.Component {
           styles.dropdown,
           isPlain && styles.modifiers.plain,
           direction === DropdownDirection.up && styles.modifiers.top,
-          position === DropdownPosition.right && styles.modifiers.alignRight,
           isOpen && styles.modifiers.expanded,
           className
         )}
@@ -100,6 +90,7 @@ class Dropdown extends React.Component {
           <DropdownMenu
             component={component}
             isOpen={isOpen}
+            position={position}
             aria-labelledby={id}
             onClick={event => onSelect && onSelect(event)}
           >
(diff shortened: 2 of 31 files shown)


mcarrano:
@tlabaj  This is unrelated to the changes you are making, but while reviewing the Toolbar PR, I noticed that the Kabob menu is not correct. Th Kebab should not have a border. @LHinson FYI - do you want me to file this as a separate issue or can we just address here?

LHinson:
@mcarrano after talking to @tlabaj, she identified that she can handle that update within this PR and does not require a separate issue. Thank you both!

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -71,7 +63,11 @@ class Dropdown extends React.Component {
         }}
       >
         {toggle && Children.map(toggle, oneToggle => cloneElement(oneToggle, { parentRef: this.parentRef, isOpen }))}
-        <DropdownMenu isOpen={isOpen} aria-labelledby={id} onClick={event => onSelect && onSelect(event)}>
+        <DropdownMenu
+          isOpen={isOpen}
+          aria-labelledby={toggle ? toggle.props.id : ''}
```
jschuler: i think toggle should be required then you can also remove the conditional here

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -71,7 +63,11 @@ class Dropdown extends React.Component {
...
+          onClick={event => onSelect && onSelect(event)}
```
jschuler:
```suggestion
          onClick={onSelect}
```

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
```diff
@@ -32,6 +34,11 @@ DropdownToggle.propTypes = {
 };
 
 DropdownToggle.defaultProps = {
+  id:
```
jschuler: multiple dropdowns will have the same id this way, generate unique id in constructor instead

packages/patternfly-4/react-core/src/components/Dropdown/KebabToggle.js
```diff
@@ -32,6 +34,11 @@ Kebab.propTypes = {
   isActive: PropTypes.bool
 };
 Kebab.defaultProps = {
+  id:
+    new Date().getTime() +
```
jschuler: Generate in constructor

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -40,13 +40,15 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
     aditionalProps.disabled = isDisabled;
```
amarie401: missing a 'd' for `additionalProps.disabled`

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -40,13 +40,15 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
...
-    <Component
-      {...aditionalProps}
-      className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
-      role={role}
-    >
-      {children}
-    </Component>
+    <li>
+      <Component
+        {...aditionalProps}
```
amarie401: same here with missing a `d`

jgiardino:
Thanks for making these updates!

Here are a couple of things I noticed:

1. In the first example, the following is the parent element for the menu:
    ```
    <ul aria-labelledby="..." class="pf-c-dropdown__menu" role="menu">
    ```
    - Is it possible to change this element to a `<div>`, but keep the `<ul>` for the other examples where actions are included in the menu? Defaulting to a `<div>` would allow the consumer to put whatever they want there.
    - I forgot to mention this in the issue, but can you remove the `role="menu"` attribute from this element?
2. In the second example, the following is the parent element for the menu:
    ```
    <ul aria-labelledby="..." class="pf-c-dropdown__menu" role="menu">
    ````
    - Similar to previous comment, can you remove the `role="menu"` attribute from this element and `role="menuitem"` from the children elements?

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
tlabaj: onSelect could be null.  I think that is why he does the check here.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
tlabaj: Good catch.  Fixed this.

tlabaj:
@jgiardino Do I remove the role from the Separator?  
`role="separator"`

jgiardino:
> It seems like the dropdown arrow sits high relative to the text label and enclosing box. What is it's current alignment? Can it be centered in the rectangle?

This might be a separate issue we should capture. I'm adding @michael-coker here for this one.

When I look at the dropdown arrow `<svg>` in patternfly-next (core), the following classes are included:
```
<svg class="svg-inline--fa fa-caret-down fa-w-10 pf-c-dropdown__toggle-icon" ... >
```

But in patternfly-react, the following classes and additional `height` and `width` classes are added:
```
<svg class="pf-c-dropdown__toggle-icon" fill="currentColor" height="1em" width="1em" ... >
```

@michael-coker - Do you happen to know why there's a difference here? Is this an issue that we should capture?

tlabaj:
@mcarrano, I removed the border from the kebab example.

michael-coker:
@jgiardino I think this is just a workspace thing. Looks like the react workspace is using a different FA family than we use on PF-next.

Looks like this workspace is using
* https://use.fontawesome.com/releases/v5.4.1/css/fontawesome.css
* https://use.fontawesome.com/releases/v5.4.1/css/brands.css
    * `font-family` in this file is "Font Awesome 5 Brands"

PF3 uses these scripts

```        
<script
  defer
  src="https://use.fontawesome.com/releases/v5.0.13/js/solid.js"
  integrity="sha384-tzzSw1/Vo+0N5UhStP3bvwWPq+uvzCMfrN1fEFe+xBmv1C/AtVX5K0uZtmcHitFZ"
  crossOrigin="anonymous"
/>
<script
  defer
  src="https://use.fontawesome.com/releases/v5.0.13/js/fontawesome.js"
  integrity="sha384-6OIrr52G08NpOFSZdxxz1xdNSndlD4vdcf/q2myIUVO0VsqaGHJsB0RaBE01VTOY"
  crossOrigin="anonymous"
/>
```

with [fontawesome.scss](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/assets/fontawesome/fontawesome.scss) and [fa-solid.scss](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/assets/fontawesome/fa-solid.scss) linked from [@patternfly/patternfly-next/assets/fontawesome/](https://github.com/patternfly/patternfly-next/tree/master/src/patternfly/assets/fontawesome) (the `font-family` is "Font Awesome 5 Free")

dmiller9911:
We are using raw svgs not the font icons. I looked into this before, and it has to do with family adding a top attribute to the icons or something similar. I will need to look again. We will need to do this as well in the react-icons package for the font awesome svgs.

tlabaj:
@michael-coker @dmiller9911 @jgiardino Maybe we should ope na separate issue to investigate the FA/svg issues.

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
```diff
@@ -1,16 +1,7 @@
...
-export const DropdownPosition: {
-  right: 'right';
-  left: 'left';
-}
-
-export const DropdownDirection: {
-  up: 'up';
-  down: 'down';
-}
+import { DropdownPosition, DropdownDirection } from './dropdownConstants';
```
jschuler: Can the file name start uppercase?

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -2,26 +2,26 @@ import React, { Children, cloneElement } from 'react';
...
+      return new Error(
+        `Children and dropdownItems props have been provided. Only the dropdownItems prop items will be rendered `
+      );
+    }
+    return null;
+  },
+  /** Classes applied to root element of dropdown */
   className: PropTypes.string,
+  /** Array of DropdownItem nodes that will be rendered in the dropdown Menu list */
+  dropdownItems: PropTypes.array,
```
jschuler: i would 'drop' the prefix dropdown and just call it items

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -54,30 +49,53 @@ const defaultProps = {
...
+      dropdownItems,
+      isOpen,
+      toggle,
+      direction,
+      onSelect,
+      isPlain,
+      position,
+      ...props
+    } = this.props;
+    const id = toggle.props.id || `pf-toggel-id-${currentId++}`;
```
jschuler: this will be updated to use the unique id wrapper? and nit spelling: toggle on the prefix

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.d.ts
```diff
@@ -1,12 +1,16 @@
 import { SFC, HTMLProps, ReactType, ReactNode } from 'react';
 
 export interface DropdownToggleProps extends HTMLProps<HTMLButtonElement> {
+  id: string;
```
jschuler: should be marked optional with '?'

packages/patternfly-4/react-core/src/components/Dropdown/Item.js
```diff
@@ -31,8 +29,7 @@ Item.defaultProps = {
   isHovered: false,
   component: 'a',
   isDisabled: false,
-  href: '#',
-  role: 'menuitem'
+  href: '#'
```
jschuler: Also need to update the typescript definitions for Item, KebabToggle, and Separator?

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
tlabaj: looks like out constants file in the internal directory start with lower case.

packages/patternfly-4/react-core/src/components/Dropdown/Item.js
tlabaj: KebabToggle extends DropdownToggle and Separator Extends DropdownItem.

PR #792, opened 2018-10-16
fix(Form): Do not add any class if action modifier is missing

karelhala:
Fixes currently broken master, because there is no info modifier for FormGroup.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Form/ActionGroup.js
+++ b/packages/patternfly-4/react-core/src/components/Form/ActionGroup.js
@@ -17,7 +17,7 @@ const defaultProps = {
 };
 
 const ActionGroup = ({ className, children, ...props }) => {
-  const customClassName = css(styles.formGroup, getModifier(styles, 'action'), className);
+  const customClassName = css(styles.formGroup, styles.modifiers.action, className);
   const classesHorizontal = css(styles.formHorizontalGroup);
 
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Form/ActionGroup.js
```diff
@@ -17,7 +17,7 @@ const defaultProps = {
 };
 
 const ActionGroup = ({ className, children, ...props }) => {
-  const customClassName = css(styles.formGroup, getModifier(styles, 'action', styles.modifiers.info), className);
+  const customClassName = css(styles.formGroup, getModifier(styles, 'action'), className);
```
dmiller9911: I missed this in the last review. it would probably be better to just use styles.modifiers.action
```suggestion
  const customClassName = css(styles.formGroup, styles.modifiers.action, className);
```

This would allow the babel plugin to throw at build time if this is ever modified. getModifier should only be used if the modifier is coming in via props.

jschuler:
#797 addressed some of these issues as well. We can use this PR to make the change to not use getModifier in the cases where modifier is not coming in via prop

karelhala:
@all sorry for the late response, I like that you merged #797! I've updated this PR not to use `getModifiers` function and just straigh modifier from styles.

Issue #795, opened 2018-10-17
ModelessOverlay is not scrollable


cben:
Modal allows scrolling when content is longer than viewport can fit:
![modal-vscroll](https://user-images.githubusercontent.com/273688/47089185-e5c1ac80-d228-11e8-9e39-2fb69e51f7c6.gif)

ModelessOverlay doesn't, which can make it unusable on small screens:
![modeless-no-vscroll](https://user-images.githubusercontent.com/273688/47089196-ec502400-d228-11e8-9308-638e6bf02d3a.gif)

According to https://www.patternfly.org/pattern-library/forms-and-controls/modeless-overlay/#design (emphasis mine):
> Vertical Navigation: If the content of the panel extends past the initial frame of the window a vertical scrollbar **should be enabled**.
> Horizontal Scrollbar: In the case of the modeless overlay panel, a horizontal scrollbar should not be used.

Sending PR to fix it...

- Didn't test Modal with `.right-side-modal-pf` (#650)

jeff-phillips-18:
Fixed by #796

PR #650, opened 2018-09-24
feat(modal): add class to allow Modal dialog to be placed on the right side

jeff-phillips-18:
affects: patternfly3-react-lerna-root, patternfly-react

**What**:
Add a class `right-side-modal-pf` to allow Modal dialog component to be placed on the right side, similar to the ModelessOverlay. Refactor the ModelessOverlay to use the same class.

**Link to Storybook**:
https://rawgit.com/jeff-phillips-18/patternfly-react/modal-storybook/index.html?selectedKind=patternfly-react%2FForms%20and%20Controls%2FModal%20Overlay&selectedStory=Modal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

PR #796, opened 2018-10-17
ModelessOverlay: allow vertical scrolling

cben:
**What**:

Modal allows scrolling when content is longer than viewport can fit.
ModelessOverlay doesn't (#795), which can make it unusable on small screens:
modeless-no-vscroll

This PR makes the Modal.Body inside scrollable:
modeless-vscroll-body
~~[old recording scrolling whole body](https://user-images.githubusercontent.com/273688/47089570-ce36f380-d229-11e8-89b4-966353ab16d3.gif)~~

According to https://www.patternfly.org/pattern-library/forms-and-controls/modeless-overlay/#design (emphasis mine):
> Vertical Navigation: If the content of the panel extends past the initial frame of the window a vertical scrollbar **should be enabled**.
> Horizontal Scrollbar: In the case of the modeless overlay panel, a horizontal scrollbar should not be used.

---

Modal with `.right-side-modal-pf` (#650) also has a problem:
modal-rightside-no-vscroll

This PR makes it better too (again scrolling just the Modal.Body inside):
modal-rightside-vscroll-body
~~[old recording scrolling whole dialog](https://user-images.githubusercontent.com/273688/47144920-3db3ee00-d2d1-11e8-9e90-c507829d1dfd.gif)~~

**Additional issues**: #795

### DIFF
--- a/packages/patternfly-3/patternfly-react/less/modeless-overlay.less
+++ b/packages/patternfly-3/patternfly-react/less/modeless-overlay.less
@@ -6,6 +6,7 @@
 
     .modal-content {
       height: ~'calc(100vh - @{navbar-pf-height})';
+      overflow: hidden auto;
     }
   }
 
--- a/packages/patternfly-3/patternfly-react/less/modeless-overlay.less
+++ b/packages/patternfly-3/patternfly-react/less/modeless-overlay.less
@@ -6,7 +6,13 @@
 
     .modal-content {
       height: ~'calc(100vh - @{navbar-pf-height})';
-      overflow: hidden auto;
+      display: flex;
+      flex-direction: column;
+
+      .modal-body {
+        flex-grow: 1;
+        overflow: hidden auto;
+      }
     }
   }
 
--- a/packages/patternfly-3/patternfly-react/sass/patternfly-react/_modeless-overlay.scss
+++ b/packages/patternfly-3/patternfly-react/sass/patternfly-react/_modeless-overlay.scss
@@ -6,6 +6,7 @@
 
     .modal-content {
       height: calc(100vh - #{$navbar-pf-height});
+      overflow: hidden auto;
     }
   }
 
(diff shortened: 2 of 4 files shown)


cben:
Note: seems the storybook only uses the `.scss` file, so I didn't test the `.less`.

jeff-phillips-18:
IMO, the body/content should scroll and the header should stay fixed

cben:
>
> IMO, the body/content should scroll and the header should stay fixed
>
I thought about that.  But is it mandatory to use Modal.Header and friends
inside?

>

jeff-phillips-18:
It's not, but the application can add scrolling to the body or where ever they like.

cben:
Hmm.  I'm at a loss how to make only the modal-body scroll.
IIUC for it to scroll we must constrain its `height`.  But what to set it to?  Something in the vein of `calc(100vh - #{$navbar-pf-height} - #{$modal-header-height} - #{$modal-footer-height})` ?  Are header and footer always of same height?  Is a footer always used?
 - BTW, should footer also remain in place, or should body+footer scroll together?

Is there something simpler with flexbox or position: sticky, with sufficient support?
Hmm, https://stackoverflow.com/a/23657083/239657 suggests various alternatives, trying some...

cben:
PTAL, pushed a change to only scroll the `Modal.Body` inside.  Updated screen recordings in PR description.

- [ ] TODO: test IE11, has known flexbox bugs that sound relevant.

jeff-phillips-18:
I'd like to remove the bootstrap grid layout and use flexbox with flex-wrap: wrap.  At large resolutions, the two are too far apart and they break too soon where they could remain on a single line. Using flex-wrap keeps them the same distance and on one line until necessary. I think setting a 20px left margin on manual connection div is about right.

I'd also like to change the form in the manual connection to always remain horizontal (the set width of 350px seems to be wide enough to support this).

Issue #791, opened 2018-10-16
PF4: react-core v1.24.1 is broken


dlabrecq:
I took the latest react-core v1.24.1 package and get this error when building koku-ui. I have to drop back to v1.23.1 in order to compile.

![screen shot 2018-10-16 at 2 59 51 pm](https://user-images.githubusercontent.com/17481322/47040531-71b5d480-d154-11e8-89ef-ede49687be49.png)

dmiller9911:
Looks like the index files are gone along with a few others.....https://unpkg.com/@patternfly/react-core@1.24.1/dist/esm/

dmiller9911:
Not sure why but travis should have bailed since this failed: https://travis-ci.org/patternfly/patternfly-react/builds/442388072#L1003 lerna seemed to ignore the failed exit code.

dlabrecq:
v1.24.3 is working now. Thanks!

PR #797, opened 2018-10-17
fix(react-core 1.24.1 failing build): Failing build

Hyperkid123:
fix 791

**What**:

Distribution package form @patternfly/react-core 1.24.1 did not contain index.js (source of component). As a result build was not usable. More info here: https://github.com/patternfly/patternfly-react/issues/791.

The build was not passing even locally, because of some incorrect style modifier values.

CI did not catch it because the build command was `"build": "yarn build:babel; yarn build:ts"` and it should be `"build": "yarn build:babel && yarn build:ts",` (credit to @dmiller9911 for this)

### DIFF
--- a/packages/patternfly-4/react-core/package.json
+++ b/packages/patternfly-4/react-core/package.json
@@ -35,7 +35,7 @@
     "react-dom": "^15.6.2 || ^16.4.0"
   },
   "scripts": {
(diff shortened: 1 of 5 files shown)


Hyperkid123:
cc @dmiller9911 @dlabrecq

PR #804, opened 2018-10-17
fix(CatalogTile): Fix to not allow the description text to overflow the tile

jeff-phillips-18:
Description text could overflow the tile in Firefox. This is a better CSS constraint.


packages/patternfly-3/patternfly-react-extensions/less/catalog-tile.less
```diff
@@ -75,6 +76,7 @@
   }
 
   .catalog-tile-pf-description {
+    flex: 1;
```
michael-coker: Doesn't look like this is necessary.

PR #756, opened 2018-10-15
feat(radioComponent): updated according to PF-next demo

rvsia:
**What**:

This PR updates radio component according to PF-next demo (https://pf-next.com/components/Check/examples/)

- adds label
- adds more tests

It's basically same component as Checkbox (https://github.com/patternfly/patternfly-react/pull/717), but with `radio` type and required `name` prop.

**Link to docs**:
https://756-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/radio

PR #806, opened 2018-10-18
fix(ts-definitions): fix type definitions for radio

priley86:
**What**:
Re-addressing #770 after adding enhancements in #756.

**Additional issues**:
none


priley86:
thanks @amarie401 - happy to report that I currently have no ts errors w/ the components i am using from `^1.25.2` 👍

PR #607, opened 2018-09-11
feat(DesktopViewer): introduce DesktopViewer component

mareklibra:
**What**:  introduce DesktopViewer component

**Link to Storybook**: [https://607-pr-patternfly-react-patternfly.surge.sh](https://607-pr-patternfly-react-patternfly.surge.sh/patternfly-3/index.html?selectedKind=react-console%20%28%40patternfly%29%2FDesktopViewer&selectedStory=DesktopViewer&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)

**Additional issues**:

**Additional info**:
The components unifies access to virtual machine/server console
connection via desktop (means non-browser) viewer application.

A button for downloading .vv file (of "application/x-virt-viewer" mime-type)
is rendered.
Console details are farther rendered in textual form to enable copy&paste
to a generic desktop viewer application.

This component implements common use case for any virtual machine/server
management application which provides remote console access.
First adopters of this component will be kubevirt/web-ui, ovirt-web-ui or Cockpit.

Examples of use:
```
 - <DesktopViewer spice={spice} vnc={vnc} />
 - <DesktopViewer spice={spice} vnc={vnc}>
     <div>My custom content of more-info section</div>
   </DesktopViewer>
 - <DesktopViewer spice={spice}
                  onDownload={(fileName, content, mimeType) => {/* perform saveAs action */}}
                  onGenerate={({ console, type } => { return ({ content, fileName, mimeType });})} />
```

PR #807, opened 2018-10-18
Rdp console

mareklibra:
**What**: Introduce RDP console access

**Storybook**: [https://807-pr-patternfly-react-patternfly.surge.sh/](https://807-pr-patternfly-react-patternfly.surge.sh/patternfly-3/index.html?selectedKind=react-console%20%28%40patternfly%29%2FDesktopViewer&selectedStory=DesktopViewer%20-%20SPICE%20and%20VNC&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)

**Additional issues**:

**Additional info**:
The `DesktopViewer` component is extended for access via RDP ([Remote Desktop Protocol](https://en.wikipedia.org/wiki/Remote_Desktop_Protocol)).

The access is handled via downloading of a descriptor file (`console.rdp`) of `application/rdp` mime type. 
If user's desktop is properly configured (like runing Windows :-) ), this download action launches external (non-browser based) viewer application.

**Examples of use**:
```
<DesktopViewer rdp={rdp} />
<DesktopViewer rdp={rdp} spice={spice} vnc={vnc} />
<DesktopViewer rdp={rdp}>
  <div />
  <div>
    My custom instructions for RDP viewer installation.
  </div>
</DesktopViewer>
```
---

**Please note:** Depends on #607 (`DesktopViewer` component). The first commit comes from not-yet-merged #607 - will be removed from here once the #607 lands.


mareklibra:
cc @priley86 , @lizsurette

mareklibra:
> Would it make sense to combine the desktop viewer component with this and allow to turn on/off Remote Viewer and RDP?

Thanks, @lizsurette .

The fact is, this is just another data variant for the use-case addressed by the `DesktopViewer` (on top of VNC/SPICE).

The question is, whether the user still needs to access non-RDP protocol in case of Windows VM.
IMO yes since the RDP client works fully just on Windows, non-windows clients (for Linux or Mac) have recently many drawbacks. 

I would not adapt the protocol preference based on the browser's operating system. So it leads us to offer both RDP and SPICE files for download for a single VM. I tried to address that by introducing this new component sharing design with the `DesktopViewer` but recently I agree, having the RDP within the `DesktopViewer` is better approach.

So considering we need to allow download of both types (RDP/SPICE) for a single VM, we can render two download buttons close to each other (in a row??). Any idea how to render the "More Information" section? Both protocols have different instructions and I would not mix them in a single text.

Of course, if the RDP protocol is not available for particular VM, there will be still just a single button/instruction.

For completeness: if the VM supports both VNC and SPICE, the SPICE takes precedence within `DesktopViewer` and the user has no option to download a `.vv` file for the VNC. There are just the copy&paste values on the right side if (s)he needs VNC. But this is fine (SPICE is better :-) ).

lizsurette:
@mareklibra Thanks for taking the time to explain this in detail, it's very helpful to me :)

That makes sense that we might have the use case where we'd want to offer up both RDP and SPICE files for download for a single VM, even if it's not a major use case the design should support it! How about something like this where the user would be given the download buttons first and then could expand sections if they want to read more about the connection details. (I moved the buttons up so that they don't get pushed off of the screen in the case the user has both detail sections and expands them both):
screen shot 2018-10-25 at 5 25 42 pm

Also, it sounds like you are in favor of only showing these Desktop connection options IF the in-browser isn't supported. So there would never be this "toggle between in-browser or desktop". It would just be in-browser if available and desktop if not. Am I understanding correctly? If so, do you think there are any scenarios where a user would just prefer to connect via desktop even if they could connect in-browser? Would they want access to that file anyways to potentially share with someone else? I just want to be sure we aren't missing something by leaving it out :)

mareklibra:
> Also, it sounds like you are in favor of only showing these Desktop connection options IF the in-browser isn't supported. So there would never be this "toggle between in-browser or desktop".

That's misunderstanding. TL/DR; we still need the "switch" functionality.

The user should still have an option to access the console via desktop-application of his/her choice. The in-browser implementations have and probably always will have limitations in usability due to (complicated?) technological stack.

Anyway, the in-broswer console should take priority over desktop-based (by default).
Moreover, the VNC takes precedence over Serial console. 
Unfortunately, there is recently no good-enough HTML based Spice client, it would be the best default.

Maybe out of scope for now, but what do you think about remembering user's choice of console-type (for particular VM? Or generically once for all (his) VMs?) to preselect it next time?

mareklibra:
Reimplemented according to the new design. @lizsurette can you have a look, please?

For code review: the first commit will be removed once #607 lands.

jeff-phillips-18:
@mareklibra  This can be closed correct?

mareklibra:
@jeff-phillips-18 , no. As `DesktopVievwer` component is merged now I'll rebase here. This introduces new feature on top of the DesktopViewer.

mareklibra:
Rebased. 
Ready for review, @lizsurette , @jeff-phillips-18 , wdyt?

mareklibra:
Rebased to resolve conflicts.

dtaylor113:
If this pattern does not yet exist in PatternFly, shouldn't it go into `patternfly-react-extensions`?

PR #808, opened 2018-10-18
fix(CatalogItemHeader): Make subtitle optional

rhamilto:
If a vendor (displayed in the subtitle) is not provided, the component emits an empty subtitle
element, which creates a visual defect

Before:
screen shot 2018-10-18 at 9 13 10 am

After:
screen shot 2018-10-18 at 9 13 58 am

attn: @jeff-phillips-18, @dtaylor113


jeff-phillips-18:
@rhamilto Could you add a subtitle-less CatalogItemHeader to the storybook and also to the tests?

rhamilto:
Updated the Nginx example and test to be vendor-less.

Issue #802, opened 2018-10-17
Backdrop Component not exported


TimoSta:
The issue is that I'm unable to import the `Backdrop` component into my app with `@patternfly/react-core@1.25.1`.

It seems to be implemented, however appears to not be exported in the `components.index.js` & .d.ts files.

LHinson:
@dlabrecq can you help provide some insight here?

dlabrecq:
I recently worked on the BackgroundImage component, but confirmed Backdrop doesn't have an index.js or index.d.ts file. If we add those files, I believe Backdrop would consumable by a typescript app

dlabrecq:
I'll create a PR to add the missing files

PR #809, opened 2018-10-18
PF4: Add missing index files for Backdrop

dlabrecq:
This just adds missing index files to export the Backdrop component.

Fixes https://github.com/patternfly/patternfly-react/issues/802


TimoSta:
I might be wrong here, but wouldn't we need to also add the component to the main components index?

https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/index.js

jschuler:
@TimoSta yep! We need to reopen or create a new PR

tlabaj:
good catch @TimoSta .  I have PR #813  to address this.

PR #810, opened 2018-10-18
fix(CatalogTile): Catalog Tiles can shrink

rhamilto:
Catalog tiles are supposed to have a fixed size.  If the viewport gets small, the tiles shrink.

Before:
screen shot 2018-10-18 at 10 25 52 am

After:
screen shot 2018-10-18 at 10 27 12 am

attn: @jeff-phillips-18, @dtaylor113


jeff-phillips-18:
When in the tile view category, the flex-wrap on the parent takes care of this, but yeah if the parent doesn't have that then, this is necessary

PR #812, opened 2018-10-18
chore(pf4): update core

jschuler:
**What**:

Update PF4 core versions


packages/patternfly-4/react-core/src/components/BackgroundImage/__snapshots__/BackgroundImage.test.js.snap
```diff
@@ -24,17 +13,6 @@ exports[`BackgroundImage 1`] = `
 exports[`BackgroundImage 2`] = `
 .pf-c-background-image.css-jrvmr5 {
   display: block;
-  position: fixed;
```
dmiller9911: These are due to like of pseudo element support for the snapshot utility. The background itself was moved into a `::before` element

PR #814, opened 2018-10-18
chore(pf4): update core for react-icons

jschuler:
**What**:
Update core for react-icons and the compare versions script

### DIFF
--- a/packages/patternfly-4/verifyCoreVersions.js
+++ b/packages/patternfly-4/verifyCoreVersions.js
@@ -8,7 +8,6 @@ const versions = new Map();
 const corePackageName = '@patternfly/patternfly-next';
 
 packages
-  .filter(({ location }) => location.match(/patternfly-4/) || location.match('/react-icons'))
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/verifyCoreVersions.js
```diff
@@ -8,7 +8,7 @@ const versions = new Map();
 const corePackageName = '@patternfly/patternfly-next';
 
 packages
-  .filter(({ location }) => location.match(/patternfly-4/))
+  .filter(({ location }) => location.match(/patternfly-4/) || location.match('/react-icons'))
```
jschuler: Do you think we need this filter at all? shouldn't all @patternfly/patternfly-next versions in this repo be the same? @dmiller9911 @priley86

dmiller9911: Yeah, I think they should all be at the same version. That will keep the dependencies consistent. This could potentially hit the same issue we hit with tokens. Not sure why I filtered in the first place.

jschuler: all right, i will remove the filter then

PR #815, opened 2018-10-19
feat(PF4-Dropdown): Enhance keyboard interaction for a11y

### LINKED ISSUE #557: Dropdown keyboard interaction

michaelkro:
Fixes #557 

# What
- Introduces `focus-trap-react` dependency [(documentation)](https://github.com/davidtheclark/focus-trap-react)
- Fixes `esc` key interaction when there are multiple `Dropdown`s on the same page.  Hitting the `esc` key will close the menu and return focus to that particular `Dropdown`'s toggle.  Previously, focus would get shifted to the last `Dropdown` on the screen

# `focus-trap-react`
- I found `focus-trap-react` while reading the Programmatically Managing Focus section in the [React docs regarding a11y](https://reactjs.org/docs/accessibility.html#programmatically-managing-focus)
- The documentation mentions [react-aria-modal](https://github.com/davidtheclark/react-aria-modal) as as "a great focus management example," the author of which also authored `focus-trap-react`

# Focus Trap Resources
- [CSS-Tricks](https://css-tricks.com/a-css-approach-to-trap-focus-inside-of-an-element/)
- [Deque Article](https://www.deque.com/blog/aria-modal-alert-dialogs-a11y-support-series-part-2/)
- [WAI-ARIA Menu](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html)
- [WAI-ARIA Modal JS](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/dialog-modal/js/dialog.js)

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import styles from '@patternfly/patternfly-next/components/Dropdown/dropdown.css';
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
+import { KEY_CODES } from '../../internal/constants';
 
 const propTypes = {
   /** Anything which can be rendered as dropdown toggle */
@@ -53,7 +54,7 @@ class DropdownToggle extends Component {
   onEscPress = event => {
     const { parentRef } = this.props;
     const keyCode = event.keyCode || event.which;
-    if (keyCode === 27 && parentRef && parentRef.contains(event.target)) {
+    if (keyCode === KEY_CODES.ESCAPE_KEY && parentRef && parentRef.contains(event.target)) {
       this.props.onToggle && this.props.onToggle(false);
       this.toggle.focus();
     }
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
@@ -51,8 +51,9 @@ class DropdownToggle extends Component {
   };
 
   onEscPress = event => {
+    const { parentRef } = this.props;
     const keyCode = event.keyCode || event.which;
-    if (keyCode === 27) {
+    if (keyCode === 27 && parentRef && parentRef.contains(event.target)) {
       this.props.onToggle && this.props.onToggle(false);
       this.toggle.focus();
     }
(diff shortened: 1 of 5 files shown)


jgiardino:
Thanks for sharing the resources you found!

I tested this for the sighted keyboard-only experience, and it works as expected.

There are still some updates pending in #788 that alter the screen reader experience, which is blocking me from reviewing and confirming this works as expected across all screen readers.

packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
```diff
@@ -51,8 +51,9 @@ class DropdownToggle extends Component {
   };
 
   onEscPress = event => {
+    const { parentRef } = this.props;
     const keyCode = event.keyCode || event.which;
-    if (keyCode === 27) {
+    if (keyCode === 27 && parentRef && parentRef.contains(event.target)) {
```
tlabaj: There is a constant for the escape key defined in the internal/constants.js can we use that here?

PR #819, opened 2018-10-19
fix(pf4 NavExpandable): Add animation of expandable nav based on PF4 core with react-transition-group

mturley:
This PR uses `react-transition-group@2.x` to apply classes and attributes at appropriate delays to take advantage of the pre-existing CSS transitions in core.

### DIFF
--- a/packages/patternfly-4/react-core/package.json
+++ b/packages/patternfly-4/react-core/package.json
@@ -28,7 +28,8 @@
     "@patternfly/react-icons": "^2.6.0",
     "@patternfly/react-styles": "^2.3.0",
     "exenv": "^1.2.2",
-    "focus-trap-react": "^4.0.1"
+    "focus-trap-react": "^4.0.1",
+    "react-transition-group": "2.x"
   },
   "peerDependencies": {
     "prop-types": "^15.6.1",
--- a/packages/patternfly-4/react-core/src/components/Nav/NavExpandable.js
+++ b/packages/patternfly-4/react-core/src/components/Nav/NavExpandable.js
@@ -3,6 +3,7 @@ import styles from '@patternfly/patternfly-next/components/Nav/nav.css';
 import a11yStyles from '@patternfly/patternfly-next/utilities/Accessibility/accessibility.css';
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
+import Transition from 'react-transition-group/Transition';
 import NavToggle from './NavToggle';
 import { AngleRightIcon } from '@patternfly/react-icons';
 import { NavContext } from './Nav';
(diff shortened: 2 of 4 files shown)


packages/patternfly-4/react-core/package.json
```diff
@@ -27,7 +27,8 @@
   "dependencies": {
     "@patternfly/react-icons": "^2.5.1",
     "@patternfly/react-styles": "^2.3.0",
-    "exenv": "^1.2.2"
+    "exenv": "^1.2.2",
+    "react-transition-group": "1.x"
```
dmiller9911: Should probably avoid using the 1.x version move to the 2.x since it has been stable for a while.

packages/patternfly-4/react-core/src/components/Nav/NavExpandable.js
```diff
@@ -36,6 +37,52 @@ const defaultProps = {
   id: ''
 };
 
+
+class DelayedHiddenChild extends React.Component {
```
dmiller9911: Delaying the hidden doesn't really cause this to Expand/Collapse it just fades the children out. It might be better to find a way to actually transition the open/close by using height/max-height Here is a pretty good write up on how to do it with vanilla js.

packages/patternfly-4/react-core/package.json
priley86: this is an exciting one to get started for many of us... kudos for that! Curious how you arrived on this lib? I have seen a lot of great animation libraries for React, but my favorite so far is probably PopMotion (extended by Pose for declarative animation styles for React) .

https://github.com/Popmotion/popmotion/tree/master/packages/popmotion-pose

At any rate, here is some nifty comparisons to consider if we intend to extend Core animations with React:
https://logrocket.com/blog/comparing-the-top-js-animation-libraries-for-use-in-react-apps-2/

priley86: I say this having no prior animation design or a11y research on this - this is obviously a bigger topic 😄

mturley:
@priley86 @dmiller9911 I noticed the simpler API of `<Transition>` provided by `react-transition-group@2.x`, and updated this PR with a viable solution! I think this works nicely. Let me know what you guys think.

jschuler:
@mturley Excellent! Looks straightforward to use and works great. I don't know how well this stacks up against the libraries @priley86 suggested, maybe we can do a small comparative analysis and/or try the Nav change out in one of the other ones.

priley86:
Nice work @mturley! I agree w/ @jschuler’s sentiments above. My guess is that we’ll be able to use a common animation library for most animations, so definitely worth the time to compare/contrast a few simple implementations. I believe @dlabreque did something similar with charts...

mturley:
Thanks @priley86! I agree that we should compare/contrast before we choose a library. However, my concern with using something like Popmotion Pose is that it appears to define the style properties being animated directly via JS rather than using CSS transitions. Since we are importing the PF core CSS which already defines the animations as CSS transitions, using Pose would mean a duplicate implementation of the same animations in JS, which could also conflict with the CSS transitions in unpredictable ways. Using something to simply apply CSS classes with certain timings allows the core CSS to provide a single source of truth for animations.

I'm certainly open to trying a few implementations, and Pose does look really cool, but I'm curious what you guys think about this CSS-in-JS versus CSS transitions problem.

priley86:
this is a very good question/distinction...and very noteworthy I think ;-). Here is my $.02.

I _think_ I still take the stance that both CSS in JS (object based /declarative styles/props) should still be supported in conjunction w/ traditional class-based styles in downstream React applications. I.m.h.o. this is more flexible in a functional paradigm like React and most Css-in-JS approaches also support our traditional classes (Pose supports  [classNames](https://popmotion.io/pose/api/posed/#posed-usage-styling) for example). I think one thing many traditionalists may be overlooking is the possibility to test and snapshot the computed styles _after_ an interaction or state is applied on a component. This becomes quite challenging w/ many traditional testing approaches (and difficult to maintain w/o something that is geared for this like Jest snapshots). So w/ PF4 snapshots, we can tell exactly what computed styles [change in the snapshot](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Alert/__snapshots__/Alert.test.js.snap#L4) and review them alongside and/or report issues upstream when noticing unexpected applications of computed styles. This is probably a vastly under-appreciated feature (particularly when considering these unit tests run much faster and scale in a headless browser such as JSDOM). In summary, I still subscribe to the notion that e2e tests are more costly (both for users and build systems) and should focus on some a11y applications or browser gaps (the [test pyramid](https://blog.kentcdodds.com/write-tests-not-too-many-mostly-integration-5e8c7fff591c) sums this up well)... essentially anything that runs faster in a unit test and duplicates an e2e test should probably remain a unit test...

I think an argument could certainly be made that a consumer could incorrectly apply conflicting classes / css variables/ or faulty Css-in-JS styles inline (or out of order), but I am more in the camp that puts this emphasis on the consumer (and prefers the flexibility offered w/ props/object based overrides when desired). If a React component inserts opinion on this, we should ensure it is consistent w/ Core over time (and supports our CSS first approach)...

It is also becoming more apparent that Css-in-Js solutions are becoming equally forward-facing as far as standards go (compared w/ traditional static/pre runtime compilers such as Sass). JSDOM recently [merged support](https://github.com/jsdom/jsdom/pull/2347) for Shadow Dom and [optimized its application](https://github.com/jsdom/jsdom/issues/2343)  for example... 

So that being said, why try to hide or obscure this choice? I think it a bit overbearing...

mturley:
@priley86, sorry I left this hanging for a week. I want to loop back around here and figure out how we should move forward. I'll resolve the merge conflicts on this tomorrow in case we still want to consider this implementation.

Everything you said makes perfect sense! I totally agree. What do you think our next steps should be? I'm a little confused on how we can provide the animations out of the box without hiding/obscuring the choice of implementation.

priley86:
@mturley - to be clear, I am just proposing that we compare a few implementations before solidifying here. Agree that in most cases, a class-based animation library will fit the bill. It's the edge cases I think we should consider here in PF React or in downstream apps. If using react-transition-group will help the most there, then I'm all for it... if not, at least explore all the options available and compare them.

mturley:
Sounds good @priley86. I'll try and make some time shortly to take a look at some of the other options.

mturley:
Things keep coming up in V2V, so this will probably have to wait until next week :(

dgutride:
closing for now - can revisit later

Issue #509, opened 2018-08-03
PF 4 Table Component (Reactabular)


LHinson:
Provide Table component based on design and CSS implementation from this issue: patternfly/patternfly-next#513. It has been decided to use Reactabular for this, similar to PF3.

Related to #402 and https://github.com/patternfly/patternfly-next/issues/513

dmiller9911:
Provide Table component based on design and CSS implementation from this issue: https://github.com/patternfly/patternfly-next/issues/513. It has been decided to use Reactabular for this, similar to PF3.

rachael-phillips:
Are we ready to close this issue, @karelhala ?

karelhala:
@rachael-phillips yes, we are!

PR #821, opened 2018-10-20
feat(Table): introduce new component Table based on pf4 designs

karelhala:
#509

**What**:
Create new component Table based on PF4 designs. Use Reactabular library to allow breaking multiple features into smaller formatters and decorators.

**Additional issues**:
* [x] Create simple table
* [x] Create sortable table
* [x] Create selectable/actions table
* [x] Create expandable table
* [x] Create variants of table
* [x] Create table with width modifiers
* [x] Write examples for all tables
* [x] Write tests for all tables
* [x] Update accessibility

**API**
* HTML:
```JS
<Table rows={rows} cells={cells}>
  <TableHeader />
  <TableBody />
</Table>
```
* `rows`
```
Array(ReactNode || { isOpen: Boolean, cells: Array(ReactNode), parent: Number })
```
* `cells`
```
Array(String || {
  title: ReactNode,
  transforms: Array(Function),
  cellTransforms: Array(Function),
  formatters: Array(Function), 
  cellFormatters: Array(Function) 
})
```
* `cellTransforms`, `cellTransforms`, `formatters`, `cellFormatters`
Functions to change look and feel of body cells (`cell` prefix) or header cells (no prefix)
**Live site**
second-hand-branch.surge.sh


karelhala:
I watned to break this into multiple steps PR and start to add new code as it goes (commits will represent each task... at least Iĺl try to do my best).

@mattnolting since you designed the initial table I have a question for you. There is table with actions in your examples, does it mean that every row will have same actions? If so, we can pass only one array of actions instead of actions per each row. If on the other side we want to allow consumers to have different actions per each row we can't go that way.

karelhala:
Oh yeah and one thing to make reactabular even more confusing. @priley86 first I've used https://www.npmjs.com/package/@reactabular/table but that did not matched code in GH repo so I rolled back to https://www.npmjs.com/package/reactabular-table which is used in PF3 table as well and looks like less bugy as the former.

priley86:
Hey @karelhala 👋- thanks for following up on this. 

I think reactabular is still a solid implementation for many use cases, but I don’t know that this means we can’t continue to explore a completely homemade/customized solution like @jeff-phillips-18 has started with the TableGrid. I think that usecase is attractive when you just want the base features and it fits nicely when you prefer infinite scroll to pagination. It’s also nice that we can completely customize the implementation. Maybe we should concurrently explore a custom PF4 variation like TableGrid along with the Reactabular approach for those needing those features immediately? My guess is that over time we could reach a custom tailored solution to support many use cases   Reactabular has today. 

Thoughts @karelhala @jeff-phillips-18?

priley86:
Ideally we can use the monorepo to separate these into distinct packages for PF4...

karelhala:
@priley86 I like the idea of splitting table into it's own package with multiple mutations and types of table. I'd suggest we finish this effort with simple table with expandable rows since this is most needed component by many apps and after that we can split this into it's own package and add some other variants of table.

packages/patternfly-4/react-core/src/components/Table/Header.js
```diff
@@ -0,0 +1,64 @@
...
+  className: PropTypes.string,
+  headerRows: PropTypes.array
+};
+
+const defaultProps = {
+  className: ''
+};
+
+class ContextHeader extends React.Component {
+  shouldComponentUpdate(nextProps) {
```
karelhala: I'd like to discuss this component more deeply because it requires lodash as dep and that might be unwanted/not necessary. What I wanted to achieve with this is that I don't like passing columns to table provider + reactabular is using outdated `getChildContext` and I'd like to avoid it by using our context. But that requires passing columns from child element to parent element and that might be a bit confusing because we are basicaly changing parent data from child.

dmiller9911: IsEqual is a deep compare. It is likely that this is actually sloeer than letting react run the renders and the vdom diffing. I haven't looked at anymore of this component so maybe it is needed in some way, but the lodash equal will likely need to change

karelhala: That's another reason why I am trying to bring conversation about this thing. The problem is that every time header is changed (new column, d&d, etc.) we have to send updated data to parent trough `Context`. If that happens parrent will try to rerender all of it's children and one of these is actually this component so we end up in never ending cycle or renders.

So there are multiple options... do not use columns on `header` and use them from provider (kinda strange so that's the reason I tried avoiding this).
Keep this deep compare here, honestly I don't think that the header will be changed multiple times (sorting will not effect it) so I don't think that this will be an issue.

dmiller9911: It might be better to start with the desired API for the table. That will likely solve most of this. I do like the idea of providing our own. What is we did something like this:

In React Core provide the styled primitives
```
import { Table, TableCell, TableRow, TableHead, TableBody } from '@patternfly/react-core';
```
Those would be strictly dumb styled elements.

Then we could offer a more opinionated table from a package `@patternfly/react-tabular` or that follows react-tabular's api but uses our components, and adds some other sane overrides. 

This approach would allow a user to either build their own tables or use our opinionated sorting solution. Sorting is not that hard So personally pulling in a lib for this doesn't seem optimal, but I could see if helping people with more complex solutions.

DIY Table Example
```tsx

import { Table, TableCell, TableRow, TableHead, TableBody } from '@patternfly/react-core';

class DataTable extends React.Component {
private handleSortDate = (sortBy: keyof ExampleData) => {....}

public render() {
  return (
      <Table>
        <TableHead>
          <TableRow>
            <TableCell sortDirction={sortBy === 'name' ? sortDir : null} onClick={() => this.sortData('name')}>Name</TableCell>
          </TableRow>
        <TableHead>
        <TableBody>
          {data.map((data) => (
            <TableRow key={data.id}>
              <TableCell label="name">{data.name}</TableCell>
            </TableRow>
           )})}
        </TableBody>
      </Table>
  );
}
```

mattnolting:
> @mattnolting since you designed the initial table I have a question for you. There is table with actions in your examples, does it mean that every row will have same actions? If so, we can pass only one array of actions instead of actions per each row. If on the other side we want to allow consumers to have different actions per each row we can't go that way.

I think it is reasonable to assume that table actions would be the same per row, but I can also see a possible need to define actions specifically for individual types of table entries. Is it possible to take this approach for v1 and extend the component to later handle individual actions per row?

dgutride:
@karelhala  - I'd also like to see this separated out into it's own package before we put it in core.  Let me know if you need some help arranging for a new package in the repo

dlabaj:
Also recommend that we separate this out into it's own package.

karelhala:
@dgutride I've created new package in the repo, but without first release and such it's impossible for me to include it in docs. What would be the best way here? Wrap up this PR with new package without docs or create new package without any component and add docs with everything in this PR? The later one would require new package and such which I'd recommend to be created by someone from PF team. What do you think?

karelhala:
@dgutride the PR is opened https://github.com/patternfly/patternfly-react/pull/899

dgutride:
@karelhala  - should this be closed in favor of a new PR built on top of the new package?  We'd also really like to see an example baked into the react docs website if possible.  Please let me know if we can help with this.

karelhala:
All has been tested and redesigned and such. I have just one small problem with actions table and dropdown with wrong styling for some reason. And I have to test colspan for collapse tables as well.

mcarrano:
@karelhala This is looking good. I compared with the core implementation from a design perspective and saw just a couple of things missing. These include:
* Expandable rows
* Compact Table that has all the same behaviors but compresses the row height.

Also, the responsive behavior does not match the original design intent. When the columns stack, the header and value should be displayed side by side. You can see this in the core implementation.

Let me know if you have questions.

karelhala:
Those travis errors are kinda strange... they are not happening for me locally :cry:

mcarrano:
This is looking good @karelhala . My only comment/question is about the Compact Table. Seems the row height should be much less than what you are showing. Is this consistent with the core CSS? I'm looking here and it looks much more compact. Would it help to see multiple rows?

packages/patternfly-4/react-table/src/components/Table/Header.js
```diff
@@ -0,0 +1,98 @@
...
+  onSort: null,
+  onSelect: null
+};
+
+class ContextHeader extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    const { sortBy: nextSortBy, actions: nextAction } = nextProps;
+    const { sortBy, actions, onCollapse, headerRows } = this.props;
+    const updateCheckers = [
+      () => nextProps.onCollapse && onCollapse && !isEqual(nextProps.onCollapse, onCollapse),
```
dmiller9911: isEqual here seems unneeded. it is a single property, and a normal `===` check should be sufficient.

package.json
```diff
@@ -147,5 +148,8 @@
   },
   "resolutions": {
     "cssom": "0.3.4"
+  },
+  "dependencies": {
+    "lodash": "^4.17.11"
```
dmiller9911: Lodash really shouldn't be needed for anything anymore. I am super opinionated about lodash and how it is overkill for most things though so this is up to others. I expressed my concerns about the deep isEqual checks in other comments

packages/patternfly-4/react-table/src/components/Table/Header.js
```diff
@@ -0,0 +1,98 @@
...
+  sortBy: {},
+  onSort: null,
+  onSelect: null
+};
+
+class ContextHeader extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    const { sortBy: nextSortBy, actions: nextAction } = nextProps;
+    const { sortBy, actions, onCollapse, headerRows } = this.props;
+    const updateCheckers = [
```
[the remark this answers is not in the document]
karelhala: I understand your concern, but this needs to be here so there is not infinite re-render when any kind of data changes. I wanted to keep the API sane, but react-tabular is expecting columns in provider and rows in table body so I had to move columns to table header and pass them up to provider.

But given the controversy over lodash I think that it would be much better to go with different API and pass rows, columns and every config prop to Table provider directly and Header, Body and Footer will consume these props.
```JSX
<Table columns={} rows={} {...props}>
  <Header />
  <Body />
  <Footer />
</Table>
```

packages/patternfly-4/react-table/src/components/Table/BodyCell.js
```diff
@@ -0,0 +1,24 @@
...
+    ...props
+  }
+  return (parentId !== undefined && colSpan === undefined) ?
+    null :
+    <Component {...mappedProps} colSpan={colSpan} />;
+};
+
+BodyCell.propTypes = {
+  'data-label': PropsType.string,
+  component: PropsType.oneOfType([PropsType.string, PropsType.node, PropsType.number])
```
jschuler: node includes string and number

packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -0,0 +1,162 @@
...
+  sortBy: PropTypes.shape({
+    index: PropTypes.number,
+    direction: PropTypes.oneOf(Object.values(SortByDirection))
+  }),
+  /** Function called when user wants to select row. */
+  onSelect: PropTypes.func,
+  /** Function called when user wants to sort table. */
+  onSort: PropTypes.func,
+  /** Additional cell displayed at the end of each row with dropdown of action items. */
+  actions: PropTypes.array,
```
jschuler: Should this have a shape of title and onClick?

packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -0,0 +1,162 @@
...
+      cellFormatters: PropTypes.arrayOf(PropTypes.func),
+      props: PropTypes.shape()
+    })
+  ])).isRequired,
+  /** Aria labeled by this property for select inputs. */
+  selectLabeledBy: PropTypes.string,
+  /** Header to display above table for accessibility reasons. */
+  header: props => {
+    if (!props['aria-label'] && !props.caption && !props.header) {
+      throw new Error('Header is required if no aria-label or caption is supplied!');
```
jschuler: Maybe simpler if each of the 3 warnings say something like
`Specify at least one of: header, caption, aria-label`

packages/patternfly-4/react-table/src/components/Table/examples/CellHeader.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Table, TableHeader, TableBody, headerCol } from '@patternfly/react-table';
+
+class CellHeader extends React.Component {
```
jschuler: Just curious about the application of marking the first cell as header cell?

karelhala:
@jschuler thanks for the review! Fixed all problems and rebased as needed.

jschuler:
@christiemolloy I think this needs UX review, table can be viewed here
https://deploy-preview-821--patternfly-react-gone.netlify.com/components/table

karelhala:
@mcarrano that preview is old, I fixed these issues a month before. I'll publish this to personal netlify so you can check current status.

karelhala:
@mcarrano @jschuler here is the new version https://5c2f8b4aaa55830008fb30df--zen-swanson-2d3350.netlify.com/

jschuler:
@jgiardino can you take a look from a11y perspective?

mcarrano:
Thanks @karelhala This definitely looks better. There were just a couple of styling things.

In the responsive mode, the alignment can be adjusted to ensure that the content aligns in columns. The attached screen shot shows the React and Core implementations side by side where you can see what I mean.

screen shot 2019-01-04 at 12 42 37 pm

Also, somehow the font rendering seems subtly different. @kybaker can you take a quick look at this to verify if it seems right to you?

jgiardino:
Hi @karelhala, Great job on this! I compared your latest preview to what we have in core and there were a few things missing.

- `aria-sort` attribute is missing on the `<th>` element that is sorted. The [wai-aria documentation](https://www.w3.org/TR/wai-aria-1.1/#aria-sort) provides details on using this attribute and the values you can define.
  - Also, the table does not show any column header as the sort value when it first loads. Does this component allow the consumer to set a default sort value?
- The “Details” expand button is missing an `aria-labelledby` attribute. In core, this attribute references both the row header and the Details button, so when the screen reader user is scanning just the buttons, they know what row is getting expanded: “parent-1 Details button, parent-2 Details button, …”. Without this `aria-labelledby` attribute, all the buttons would have the same label "Details", which makes it difficult to navigate and operate the table.
- I do see differences between core and this preview in the kebab button. The core example appears to be incorrect (Issue in pf-next: https://github.com/patternfly/patternfly-next/issues/1183), but in this preview the kebab and menu are left-aligned. Should we default the kebab to the right position, and allow the consumer to change that if they prefer the left position? (@mcarrano did you have any expectations on the alignment of the menu?)

jgiardino:
I just added an issue to core to add the `aria-controls` attribute to the Table component: https://github.com/patternfly/patternfly-next/issues/1184

Let me know if this can also be added in this PR, or if a separate issue is needed.

The updates for this component would be:
- Add an `id` to the element with class `.pf-c-table__expandable-row-content`
- Add the attribute `aria-controls=[id of .pf-c-table__expandable-row-content]` to the button that toggles the expandable rows:
  `<button aria-label="Details" aria-expanded="true" aria-controls=[id] class="pf-c-button pf-m-plain pf-m-expanded" type="button">`

mcarrano:
Good catch @jgiardino .  I would prefer that the kabob actions were right aligned here since the way it is now, the menu will need to overflow the boundary of the table's container.

jgiardino:
Just FYI, I updated my previous comment about how `aria-sort` should be defined for both the sorted and sortable columns.

karelhala:
> In the responsive mode, the alignment can be adjusted to ensure that the content aligns in columns. The attached screen shot shows the React and Core implementations side by side where you can see what I mean.

That sure looks strange... I tried it but looks like I am unable to achieve such problem. Rows and columns are taking styling from pf-next so they should be styled correctly
selection_006
selection_007
selection_008

> Also, somehow the font rendering seems subtly different. 

It looks somewhat different, but computed styles look the same (Font Family, Font size, color, Font weight...) so not sure why they seem to be different.

@jgiardino thanks for the comments, I'll try to implement them and update you about the changes. I particularly like adding option to position where dropdowns should be rendered (left/right - defaults to right).

karelhala:
> aria-sort attribute is missing on the

Really good catch! Fixed it based on your suggestions.

> Also, the table does not show any column header as the sort value when it first loads. Does this component allow the consumer to set a default sort value?

It is to allow non sorted tables and yes, consumers can send sort value if they want to.

> The “Details” expand button is missing an

Updated.

> I do see differences between core and this preview in the kebab button.

Allowed passing both direction and position for dropdowns.

With the aria-controls, that's not gonna be easy. Parents do not know about which children they open (To allow asynchronous loading of collapsed children). And what if there are multiple items collapsed on click `<button aria-label="Details" aria-expanded="true" aria-controls=[id1 id2 id3] class="pf-c-button pf-m-plain pf-m-expanded" type="button">`? However I have added id to expanded row content.

mcarrano:
@karelhala I looked at the responsive behavior again. Looks like the problems I am seeing only show up in Firefox. Can you try that? Looks good in Chrome.

jgiardino:
Hey @karelhala, thanks for working on updates! I'm not seeing changes, but maybe the preview link I'm using is old? Is there an updated preview link I should use?

karelhala:
Yeah, that is a old preview link - this is new one https://5c35a431ca63422985f49310--zen-swanson-2d3350.netlify.com/

Gonna check in FF as well, that sure is strange behavior.

karelhala:
@mcarrano I checked why those tables are not aligned properly and it's actually intended. When grid is so small it couldn't fit all items properly it's gonna loose columns. The reason why you noticed it is that in here the table has much less space to cope with. This thing happens in pf-next as well [1], on other hand when there is enough space for the table it looks better.
[1] selection_016

[2] selection_017

mceledonia:
This should be using Overpass as the font, and it seems like it's just using the default sans serif!

mcarrano:
Thanks for the explanation about the responsive behavior @karelhala . So the only remaining design issue I see is the use as the correct font as noted above by @mceledonia .

jgiardino:
The updates look great! Thanks so much for making them! 

There are just a few minor things I noticed:

- Currently when a sortable column is not sorted, the attribute is defined as `aria-sort="None"`. I've never seen predefined aria attribute values to be anything but lower-case. It might not cause issues, but just to be safe, can we make this `aria-sort="none"`
- When I expand a row, `<tbody>` changes to `<tbody class="pf-m-expanded">` but I don't see that class present in our core example. Was that intentional?

> With the aria-controls, that's not gonna be easy. Parents do not know about which children they open (To allow asynchronous loading of collapsed children). And what if there are multiple items collapsed on click <button aria-label="Details" aria-expanded="true" aria-controls=[id1 id2 id3] class="pf-c-button pf-m-plain pf-m-expanded" type="button">? However I have added id to expanded row content.

I do see an `id` added to the expanded contents—thanks for adding that!  I added an issue (https://github.com/patternfly/patternfly-react/issues/1163) to add the `aria-controls` attribute, and added your comment there as well. Feel free to add more info that might be helpful.

karelhala:
@mceledonia @jgiardino mhh looks like the preview was using older version of dependencies and had older version of font.

> `aria-sort="None"` 

Great catch!

> `<tbody class="pf-m-expanded">`

That was intentional because @mattnolting has PR prepeared to simplify classes and styles

jschuler:
Also, this still has the `Do not merge` label

mcarrano:
Yes, it looks good to me @tlabaj

karelhala:
> Looks good. Only question is did you verify the the d.ts files in a typescript app?

Good point, I have updated TS files based on prop definitions.

priley86:
cloned this locally... getting the following on the Actions table after clicking the kebab:
table-actions-test

karelhala:
@priley86 sadly that is a react core issue and many ppl are experiencing it https://github.com/patternfly/patternfly-react/issues/1191

karelhala:
@priley86 https://github.com/patternfly/patternfly-react/pull/1193 fixes this issue, long story short ref was not assigned in dropdownitem if any child element is valid React element.

tlabaj:
@karelhala  Please resolve conflicts.

karelhala:
And new preview is here https://5c3f5db4b1cda50c4db60305--zen-swanson-2d3350.netlify.com just wait untill it's deployed

Issue #784, opened 2018-10-16
PF4 Progress Bar: Need a unique property for percentage label


dlabrecq:
The Progress component uses the same `value` property for both the progress bar and label shown in the upper right corner. The Koku UI needs to use a custom label as shown below.

![screen shot 2018-10-16 at 12 49 51 pm](https://user-images.githubusercontent.com/17481322/47033469-e4b64f80-d142-11e8-9ee0-1d6d4ada19aa.png)

Ideally, the percent value and percent label should be two properties. Providing another property would allow me to customize the text shown in the Koku dashboard, while providing a more precise percent value (e.g., 82.94345491971737) if needed for the bar.

Note that I attempted to use the valueText property, but that appears to be used for aria-valuetext. IMO, that property name should also be prefixed with "aria", similar to alert's aria-label, to avoid confusion.

tlabaj:
closed by #822

PR #822, opened 2018-10-20
feat(Labeled progress): allow consumers to pass additional label for progressbar

karelhala:
#784

**What**:
To make it possible for consumers to show what indicates current progress we have to allow them pass additional label. This PR addresses such issue by adding new prop `label` to progress component.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Progress/Progress.js
+++ b/packages/patternfly-4/react-core/src/components/Progress/Progress.js
@@ -21,6 +21,8 @@ const propTypes = {
   variant: PropTypes.oneOf(Object.values(ProgressVariant)),
   /** Title above progress. */
   title: PropTypes.string,
+  /** Label to indicate what progress is showing. */
+  label: PropTypes.node,
   /** Actual value of progress. */
   value: PropTypes.number,
   /** DOM id for progress component. */
@@ -59,7 +61,7 @@ class Progress extends Component {
   }
 
   render() {
-    const { className, size, id, value, title, variant, measureLocation, min, max, valueText, ...props } = this.props;
+    const { className, size, id, value, title, label, variant, measureLocation, min, max, valueText, ...props } = this.props;
     const { uniqueId } = this.state;
     const additionalProps = {
       ...props,
@@ -88,6 +90,7 @@ class Progress extends Component {
           parentId={uniqueId}
           value={limitedValue}
           title={title}
+          label={label}
           variant={variant}
           measureLocation={measureLocation}
         />
--- a/packages/patternfly-4/react-core/src/components/Progress/ProgressContainer.js
+++ b/packages/patternfly-4/react-core/src/components/Progress/ProgressContainer.js
@@ -1,6 +1,11 @@
 import React, { Fragment } from 'react';
-import styles from '@patternfly/patternfly-next/components/Progress/progress.css';
-import { css } from '@patternfly/react-styles';
+import {
+  progressDescription,
+  progressMeasure,
+  progressStatusIcon,
+  progressStatus
+} from '@patternfly/patternfly-next/components/Progress/progress.css';
+import { css, StyleSheet } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import { CheckCircleIcon, TimesCircleIcon } from '@patternfly/react-icons';
 import ProgressBar from './ProgressBar';
@@ -23,6 +28,8 @@ const propTypes = {
   parentId: PropTypes.string.isRequired,
   /** Progress title. */
   title: PropTypes.string,
+  /** Label to indicate what progress is showing. */
+  label: PropTypes.node,
   /** Type of progress status. */
   variant: PropTypes.oneOf(Object.values(ProgressVariant)),
   /** Location of progress value. */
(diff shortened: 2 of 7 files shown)


packages/patternfly-4/react-core/src/components/Progress/ProgressContainer.js
```diff
@@ -42,14 +42,16 @@ const variantToIcon = {
...
     <Fragment>
       <div className={css(styles.progressDescription)} id={`${parentId}-description`}>
         {title}
       </div>
       <div className={css(styles.progressStatus)}>
+        {/* TODO: add padding to label */}
+        {label}
         {(measureLocation === ProgressMeasureLocation.top || measureLocation === ProgressMeasureLocation.outside) && (
           <span className={css(styles.progressMeasure)}>{value}%</span>
         )}
```
dlabrecq: I'm looking to wrap the percentage in parentheses like the snapshot below. It looks like I can only add text before it?

screen shot 2018-10-16 at 12 49 51 pm

I'd prefer if the label was wrapped in the span so we get the same text style provided by progressMeasure. Perhaps we can use the `value` by default, if `label` has not been provided?

karelhala: Your example fits this as well, you'll just use `measureLocation={ProgressMeasureLocation.none}` and pass your own label which will be whatever you'll want (string, node, number). If we'd were to force parentheses we'll be targeting only this specific example. But what if someone would want to show `Already donwloaded: 50 %`? We'd have to have another branching here.

But you are correct that I should add styles applied to progress number on top of those custom styles (so label has some padding).

dlabrecq: Not asking to force parentheses. I just want the ability to provide my own label. If measureLocation, will hide it, that's fine. Just need the styling.

packages/patternfly-4/react-core/src/components/Progress/ProgressContainer.js
```diff
@@ -42,20 +56,25 @@ const variantToIcon = {
...
-      <div className={css(styles.progressStatus)}>
+      <div className={css(progressStatus)}>
+        <span className={css(progressMeasure)}>
+          {label}
+        </span>
         {(measureLocation === ProgressMeasureLocation.top || measureLocation === ProgressMeasureLocation.outside) && (
-          <span className={css(styles.progressMeasure)}>{value}%</span>
+          <span className={css(progressMeasure, measureLocation === ProgressMeasureLocation.top && progressLabel)}>
+            {value}%
+          </span>
```
dlabrecq: I feel that the label should override the value / measureLocation. Having to also set ProgressMeasureLocation.none doesn't feel clean, but don't feel strongly about it.

packages/patternfly-4/react-core/src/components/Progress/ProgressContainer.js
karelhala: With this approach you'll be able to pass your own label. I think we don't understand each other on here. Check the [example](https://github.com/patternfly/patternfly-react/blob/f50624a204629638a068e28c441c55990e421e72/packages/patternfly-4/react-core/src/components/Progress/examples/LabeledProgress.js) where you can pass whatever you want
```JSX
<Progress value={33} title="Descriptive text here" label={<div>$106.30 (82.49%)</div>} measureLocation={ProgressMeasureLocation.none} />;
```
Or just plain old string
```JSX
<Progress value={33} title="Descriptive text here" label="$106.30 (82.49%)" measureLocation={ProgressMeasureLocation.none} />;
```

dmiller9911: @karelhala I think he is more speaking to line 56 here since in addition to the label passed the component is also adding in ` <span className={css(styles.progressMeasure)}>{value}%</span>` that would make this:
```jsx
<Progress value={33} title="Descriptive text here" label="$106.30 (82.49%)"  />;
```
Output something like `$106.30 (82.49%) 33%` from the looks of it. 

Is there a way that we can have them just override the actual measure label? like David mentioned, default to the `{value}%`, but user the one provided if there is one. Someone might want to override the label no matter the possition.

packages/patternfly-4/react-core/src/components/Progress/Progress.js
```diff
@@ -21,6 +21,8 @@ const propTypes = {
   variant: PropTypes.oneOf(Object.values(ProgressVariant)),
   /** Title above progress. */
   title: PropTypes.string,
+  /** Label to indicate what progress is shwoing. */
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.number]),
```
jschuler: node includes string and number
```suggestion
  label: PropTypes.node,
```

packages/patternfly-4/react-core/src/components/Progress/ProgressContainer.js
```diff
@@ -23,6 +28,8 @@ const propTypes = {
   parentId: PropTypes.string.isRequired,
   /** Progress title. */
   title: PropTypes.string,
+  /** Label to indicate what progress is shwoing. */
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.number]),
```
jschuler: same comments as above
```suggestion
  label: PropTypes.node,
```

PR #824, opened 2018-10-22
fix(background-image): Add missing fragment identifier to the example…

dlabrecq:
Add missing fragment identifier to the example's filter image. 

The #image_overlay "hash is a fragment identifier and links to the thing in the SVG doc to use".


packages/patternfly-4/react-core/src/components/BackgroundImage/examples/SimpleBackgroundImage.js
```diff
@@ -10,7 +10,7 @@ const images = {
   [BackgroundImageSrc.xl]: '/assets/images/pfbg_2000.jpg',
   [BackgroundImageSrc.xs]: '/assets/images/pfbg_576.jpg',
   [BackgroundImageSrc.xs2x]: '/assets/images/pfbg_576@2x.jpg',
-  [BackgroundImageSrc.filter]: '/assets/images/background-filter.svg'
+  [BackgroundImageSrc.filter]: '/assets/images/background-filter.svg#image_overlay'
```
dmiller9911: Before merging this I would like to see if there is a way to avoid having to specify the `#image_overlay` maybe removing the ID from the SVG? Or to at least document that this is required. This is one of those bugs you don't catch until a customer sees it.

dlabrecq: I'd at least like to get this working on Firefox. Where would we document this (i.e., considering it's specific to the image provided)?

dmiller9911: It is and isn't since most people will be using the filter provided from core their locations might just vary. This will not work for most. I would not know to add it. We just need to note that if they are using the svg provided then they need to add it. This can either be in the JSDOC with the propTypes or the description on the example. The example description might be the better place since it is not specific to the component, but more specific to the urls being used.

dlabrecq: Ok. I've added a comment to the example, where the image is defined.

PR #828, opened 2018-10-24
fix(Slider): Call onSlide on slide event

### LINKED ISSUE #783: PF3: onSlide property for Slider component is not taking any effect
I am using ```Slider``` component like this:

Example:
``` js
<Slider
    label="Memory"
    id="slider-pf"
    tooltip="show"
    showBoundaries
    input
    inputFormat="MiB"
    dropup
    min={0}
    max={16000}
    value={this.state.memory}
    onSlide={this.onMemChange}
/>
```

```onSlide``` handler is never called, although in the docs is mentioned in the input properties for this component.

Looking quickly in the implementation I saw that you overwriting the ```onSlide``` method, here:
https://github.com/patternfly/patternfly-react/blob/67b0cf1d83470c13bc4b56aeb10746996ea519cd/packages/patternfly-3/patternfly-react/src/components/Slider/Slider.js#L72

With the one defined here:
https://github.com/patternfly/patternfly-react/blob/67b0cf1d83470c13bc4b56aeb10746996ea519cd/packages/patternfly-3/patternfly-react/src/components/Slider/Slider.js#L19

Expected results:
```onSlide``` handler passed in component definition should be used.

Let me know if I have misunderstood/misused something.

### PR DESCRIPTION
**Pull request by boaz1337**
closes #783

### CODE REVIEW

**Comment by boaz1337** [comment]
ping @jeff-phillips-18 @priley86

File: packages/patternfly-3/patternfly-react/src/components/Slider/Slider.js [hunk 1]
```diff
@@ -112,7 +112,7 @@ Slider.defaultProps = {
   value: 0,
   step: 1,
   toolTip: false,
-  onSlide: null,
+  onSlide: () => null,
```
jeff-phillips-18: could use helpers.noop

PR #830, opened 2018-10-25
refactor: Use the same mechanism for autogenerated component id

guillaumevincent:
To make the code more consistent, I added a method getUniqueId that is used by all components that
need a unique id. I put this id in the defaultProps object. Add also some tests for util
file.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Progress/Progress.js
+++ b/packages/patternfly-4/react-core/src/components/Progress/Progress.js
@@ -3,6 +3,7 @@ import styles from '@patternfly/patternfly-next/components/Progress/progress.css
 import { css, getModifier } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import ProgressContainer, { ProgressMeasureLocation, ProgressVariant } from './ProgressContainer';
+import { getUniqueId } from '../../internal/util';
 
 export const ProgressSize = {
   sm: 'sm',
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/components/Progress/Progress.js
```diff
@@ -49,20 +50,21 @@ const defaultProps = {
 };
 
 class Progress extends Component {
-  constructor(props) {
```
jschuler: Great PR! But we do need to initialize the unique id in the constructors. Or at least have some other way of assigning unique id to the component. Props get evaluated before the component instances are created so if you have multiple of the same components on the page they will all have the same id otherwise.

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -37,11 +38,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  id:
-    new Date().getTime() +
-    Math.random()
-      .toString(36)
-      .slice(2),
+  id: getUniqueId(),
```
dmiller9911: this does not work. Default props are only instantiated once on when this variable , `defaultProps` is created. The id will need to be setup in a "constructor" or better as a class property
```jsx
class Something extends React.Component {
  this.id = this.props.id || getUniqueId();
}
```

packages/patternfly-4/react-core/src/internal/util.js
```diff
@@ -1,3 +1,12 @@
 export function capitalize(input) {
   return input[0].toUpperCase() + input.substring(1);
 }
+
+export function getUniqueId() {
```
dmiller9911: might be nice to allow a prefix for this so that we can make the id's somewhat useful
```js
getUniqueId(prefix = '') {
  return (
   `${prefix}-` + new Date().getTime() +
    Math.random()
      .toString(36)
      .slice(2)
  );
}
```
this would allow us to do something like `getUniqueId('pf-switch')` we could even add the `pf-` by default and have the preson just pass the component name to it.

guillaumevincent:
@dmiller9911 and @jschuler done
Tell me if this implementation is good

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -37,11 +38,6 @@ const propTypes = {
 };
 
 const defaultProps = {
-  id:
```
jschuler: id in defaultProps is not needed?

PR #832, opened 2018-10-25
chore(patternfly-4/react-core): Added a Wrapper component that will g…

tlabaj:
…enerate a random id for the chi

**What**:

Added a Wrapper component that will generate a random ID for the child component 

**Additional issues**:


jschuler:
Looks like we have 2 competing approaches #830

Issue #753, opened 2018-10-13
Add a styled system component


jschuler:
Expose [PF core css globals](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/_variables.scss) for setting spacers such as padding, margin, font size, font family, line height, colors etc through a `<Box>` component that can be used like this:
`<Box m={2} />`
Looking at utilizing [styled-system](https://github.com/jxnblk/styled-system/blob/master/docs/api.md)

LHinson:
Closing this issue for now since #836 has merged

PR #836, opened 2018-10-25
feat(StyledSystem): Adds styled-system components to map to Core glob…

jschuler:
…al CSS

Closes:
#753

These components give consumers another avenue to style their pages and components so that they are consistent with the PF-Core theme, e.g. for media breakpoints, paddings, margins, colors, fonts, etc.


jschuler:
Added do not merge label since i saw a few things that i want to cleanup first

packages/patternfly-4/react-core/src/components/StyledSystem/StyledBox.d.ts
```diff
@@ -0,0 +1,62 @@
+import { SFC, HTMLProps, ReactNode } from 'react';
+
+export interface StyledBoxProps {
```
dmiller9911: These should be available from styled-system `@types/styled-system` I am still not sure the best way to handle that though. That can either be listed as a peerDependency, and optional, or a straight dependency. @priley86  might have a better idea for that.

packages/patternfly-4/react-core/src/components/StyledSystem/StyledBox.js
```diff
@@ -0,0 +1,285 @@
...
+  left,
+  top,
+  right,
+  bottom
+);
+
+const StyledBox = ({ className, children, component, ...props }) => {
+  const StyledBoxBaseWithComponent = StyledBoxBase.withComponent(component);
+  return (
+    <ThemeProvider theme={StyledTheme}>
```
dmiller9911: Wonder if we should take the time to add a `PatternFlyThemeProvider` that contains this. That would allow people to overwrite the theme anywhere in the tree. Also, it would take some of the overhead out of this.

packages/patternfly-4/react-core/src/components/StyledSystem/StyledTheme.js
```diff
@@ -0,0 +1,368 @@
...
+    xl: global_ZIndex_xl.value,
+    xxl: global_ZIndex_2xl.value
+  },
+  ratio: {
+    golden: global_golden_ratio.value
+  },
+  // use values directly so that the user can concatenate multiple shadows
+  // e.g. boxShadow={`${StyledConstants.sm_right}, ${StyledConstants.sm_bottom}`}
+  shadows: {
+    sm: global_BoxShadow_sm.value,
```
dmiller9911: Does using the `global_BoxShadow_sm.var` work for all of these? That would allow people to choose to overwrite using JS or CSS Custom Properties

priley86:
I am fine w/ iterating on these... these constants will likely change over time, but this gives us a very useful baseline.

packages/patternfly-4/react-core/src/components/StyledSystem/StyledBox.d.ts
jschuler: Okay cool, I can extend my interface with those typings. (just fixing a small issue with them here as well https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30170).
I am adding it as an optional dependency for now, sound good?

priley86:
my vote would be to get this PR in a.s.a.p. and iterate in future issues/PRs. I think it will provide a lot of value to many teams now and would be less refactor later on...thoughts?

priley86:
I think as others have pointed out, this could also be provided as an independent package that could enable using these tokens/constants in _any JS framework_ or even across shadow roots if one really wanted to...

jschuler:
Creating a @patternfly/react-styled-system package today then going to add this component to it

priley86:
this is great news. nice job! I think this interface will potentially help a lot of React components be more versatile and meet PF design specification. I think in the future one could even break out the underlying emotion objects constructed into a base package consumable by other frameworks/systems, but that is definitely less pressing...

packages/patternfly-4/react-core/src/components/StyledSystem/StyledTheme.js
jschuler: Yes, have switched what I could do .var instead. Addressed in #891

jschuler:
Closing in favor of #891

Issue #844, opened 2018-10-26
Checkbox component is missing isChecked


dlabaj:
In PF4 react we are spreading checked down to the input, however a consumer does not know what our implementation is for the checkbox. We should expose checked as isChecked like we do with the Switch component. This will provide consistency between these components and conform to our API standards.

amarie401:
The associated PR has been merged so I will go ahead and close this issue. #848

PR #848, opened 2018-10-28
refactor(Radio/Checkbox): replace checked with isChecked in API

boaz1337:
I think there is some inconsistency in Radio and Checkbox API.
IMHO, it's expected the `isChecked` prop to set the item as checked (the same way `isDisabled` renders the item disabled). However, to check the item the `checked` props is used.

In this PR, I am attempting to fix this behaviour/API.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
@@ -10,6 +10,8 @@ const propTypes = {
   isValid: PropTypes.bool,
   /** Flag to show if the Checkbox is disabled. */
   isDisabled: PropTypes.bool,
+  /** Flag to show if the Checkbox is checked. */
+  isChecked: PropTypes.bool,
   /** A callback for when the Checkbox selection changes. */
   onChange: PropTypes.func,
   /** Label text of the checkbox. */
@@ -24,6 +26,7 @@ const defaultProps = {
   className: '',
   isValid: true,
   isDisabled: false,
+  isChecked: null,
   onChange: () => undefined,
   label: undefined
 };
@@ -34,7 +37,7 @@ class Checkbox extends React.Component {
   };
 
   render() {
-    const { className, onChange, isValid, isDisabled, label, ...props } = this.props;
+    const { className, onChange, isValid, isDisabled, isChecked, label, checked, ...props } = this.props;
     return (
       <div className={css(styles.check, className)}>
         <input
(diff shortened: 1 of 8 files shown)


packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -24,6 +26,7 @@ const defaultProps = {
   className: '',
   isValid: true,
   isDisabled: false,
+  isChecked: false,
```
dmiller9911: I am definitely ok with moving to I checked, but setting this to false will break uncontrolled versions. This needs to not be in default props. Lint will get mad so maybe try undefined or null here. Not sure which will work.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -44,6 +47,7 @@ class Checkbox extends React.Component {
           onChange={this.handleChange}
           aria-invalid={!isValid}
           disabled={isDisabled}
+          checked={isChecked}
```
dmiller9911: This will be breaking so we need to gracefully do this by using checked or isChecked. We can depracate it in next major release.

packages/patternfly-4/react-core/src/components/Radio/Radio.js
```diff
@@ -26,6 +28,7 @@ const defaultProps = {
   className: '',
   isValid: true,
   isDisabled: false,
+  isChecked: false,
```
dmiller9911: Same comments apply to the radio.

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
```diff
@@ -34,7 +37,7 @@ class Checkbox extends React.Component {
   };
 
   render() {
-    const { className, onChange, isValid, isDisabled, label, ...props } = this.props;
+    const { className, onChange, isValid, isDisabled, isChecked, label, checked, ...props } = this.props;
```
dlabaj: Can we just remove checked and only us isChecked?

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
dlabaj: We are still in alpha so I think breaking is fine.  Think it should only be isChecked.

dlabaj:
Fixes issue #844

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.js
dlabaj: Okay with leaving checked for people already using it.

PR #850, opened 2018-10-29
fix(CatalogTile): Fade end of description text rather than truncate

jeff-phillips-18:
By default, fade the text of the description rather than truncating. Truncation can still be
accomplished by setting the maxDescriptionLength parameter.

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/less/catalog-tile.less
+++ b/packages/patternfly-3/patternfly-react-extensions/less/catalog-tile.less
@@ -8,6 +8,7 @@
   height: @catalog-tile-pf-height;
   margin: 0 15px @catalog-tile-pf-margin-bottom 0;
   padding: 15px;
+  position: relative;
   width: @catalog-tile-pf-width;
 
   &.featured {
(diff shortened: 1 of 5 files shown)


jeff-phillips-18:
Direct storybook link:

https://850-pr-patternfly-react-patternfly.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react-extensions%2FCatalog%20Components%2FCatalog%20Tile%20View&selectedStory=CatalogTileView

packages/patternfly-3/patternfly-react-extensions/src/components/CatalogTile/CatalogTile.js
```diff
@@ -23,7 +23,7 @@ const CatalogTile = ({
   const classes = classNames('catalog-tile-pf', { featured }, className);
 
   const defaultTruncateDescription = (text, max) => {
-    if (typeof text !== 'string' || text.length <= max) {
+    if (max === -1 || typeof text !== 'string' || text.length <= max) {
```
priley86: ok... this just allows us to explicitly return text if desired (say you'd prefer to fade the visible characters if they overflow the title). I'm following now 😸

packages/patternfly-3/patternfly-react-extensions/less/catalog-tile.less
```diff
@@ -79,7 +80,18 @@
   .catalog-tile-pf-description {
     flex: 1;
     margin-top: 15px;
-    overflow-y: hidden;
+    overflow: hidden;
+
+    &::after {
+      background: linear-gradient(to right, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 50%);
```
priley86: aXe does not seem to mind this. I am guessing the screen reader will just read the text and ignore it.

Issue #852, opened 2018-10-29
Nav enhancement: click call back functionality


LHinson:
NavItem does currently not allow onClick to be added. Modify to allow callback

PR #854, opened 2018-10-29
Nav item update to handle onClick

jschuler:
Closes: #852 

**What**:

- Allow onClick to be added to NavItem
- Don't stop event propagation from NavItem

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Nav/NavItem.js
+++ b/packages/patternfly-4/react-core/src/components/Nav/NavItem.js
@@ -34,8 +34,8 @@ const defaultProps = {
   onClick: null
 };
 
-const NavItem = ({ className, children, to, isActive, groupId, itemId, preventDefault, onClick, ...props }) => (
-  <li className={css(styles.navItem, className)} {...props}>
+const NavItem = ({ className, children, to, isActive, groupId, itemId, preventDefault, onClick, ...rest }) => {
+  const defaultLink = (
     <NavContext.Consumer>
       {context => (
         <a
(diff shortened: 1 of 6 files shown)


priley86:
thanks @jschuler! I will test this out tomorrow and let you know how it goes.

priley86:
tested this w/ `onClick` handler passed into the React Router `Link` component just now... handler callback was called correctly now. This works for me!

packages/patternfly-4/react-core/src/components/Nav/NavItem.js
```diff
@@ -16,50 +16,40 @@ const propTypes = {
...
-    </NavContext.Consumer>
-  );
-  return (
-    <li className={css(styles.navItem, className)} {...props}>
-      {reactElement ? clonedChild : defaultLink}
-    </li>
-  );
-};
+  </li>
+);
```
priley86: this bit was useful if we want to allow the consumer to still render a `Link` instead of just the `NavItem`, i.e.:

```
      <NavItem isActive={isActive}>
        <Link
          id={id}
          to={this.to}
          target={target}
          onClick={onClick}
          className={classNames({
            'pf-c-nav__link': true,
            'pf-m-current': isActive,
            'pf4-co-external-link': isExternal
          })}
        >
          {name}
```

I find it preferable to use that approach as well (it's a bit more explicit the render the Link tag/anchor separately).

priley86: (in that case, the NavItem just renders the `li`)

jschuler: Thanks, I've added it back in

priley86:
working great @jschuler ! I think this is much more flexible. I can handle clicks on the Link, the NavItem, or the Nav itself. I can also apply custom classes. Very nice!

PR #858, opened 2018-10-30
fix(Progress): import progress style object

boaz1337:
Updating to `1.28.1` shows the following error message.

screenshot-2018-10-30 react app

It looks like `../../@patternfly/patternfly-next/components/Progress/progress.css.js` exports a default object. However, progressBar, progressIndicator, progressMeasure, progressDescription, progressStatusIcon and progressStatus are imported.

I am not sure whether it's a problem on babeljs side or that's an innocent mistake. Nevertheless, it fixed the problem for me.


jschuler:
#861 addressed the issue partially, i've merged in and resolved the conflict from master and we can merge this PR to address the component that was missed

PR #859, opened 2018-10-30
feat(Label): add Label to PF4

boaz1337:


dlabaj:
@boaz1337 Tried to access the PF4 components from the link above but the PF4 link didn't load: https://859-pr-patternfly-react-patternfly.surge.sh/patternfly-4/index.html

packages/patternfly-4/react-core/src/components/Label/Label.d.ts
```diff
@@ -0,0 +1,10 @@
+import { HTMLProps, SFC, ReactType, ReactNode } from 'react';
+
+export interface LabelProps extends HTMLProps<HTMLSpanElement> {
+  children: ReactNode;
+  isActive?: boolean;
```
dmiller9911: should be `isCompact`

packages/patternfly-4/react-core/src/components/Label/Label.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Label/styles.css';
+import { css, getModifier } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+
+const Label = ({ className, children, isCompact }) => (
+  <span className={css(styles.label, className, isCompact && getModifier(styles, isCompact && 'compact'))}>
```
dmiller9911: make this `isCompact && styles.modifiers.compact` It will typecheck. getModifier doesn't type check at all, and it doesn't really help out much here. It was only meant for times were we want the user to pass in a variant and default it if they pass the wrong thing in.

packages/patternfly-4/react-core/src/components/Label/index.d.ts
```diff
@@ -0,0 +1 @@
+export { default as Label } from './Label';
```
dmiller9911: helpful to add the `LabelProps` interface here too.

boaz1337:
Thanks @dlabaj @dmiller9911  can you please take a second look.

Issue #604, opened 2018-09-10
PF4 Login 


LHinson:
Introduce PF4 Login: https://pf-next.com/demos/LoginPage/examples/

Components and Layouts used for this demo include: 
* background image #550 
* Login Box 
* Login Layout 
* Title
* Brand 
* Content 
* List 
* Form

Also note, there is a pending update for the Login demo to update it with a list and form. See issue here: https://github.com/patternfly/patternfly-next/issues/717

In addition to a React review, this issue should be reviewed by the core contributors of this demo: 
* CSS: @andybraren @andresgalante 
* Design: @mceledonia

LHinson:
This is on hold while we wait on related components to merge first: 
* #550

Issue #825, opened 2018-10-23
PF4:Login Form Component


tlabaj:
Introduce PF4 Login Form:https://pf-next.com/demos/LoginPage/examples/
This component is the content of the Login Box Body.

In addition to a React review, this issue should be reviewed by the core contributors of this demo:

CSS: @andybraren @andresgalante
Design: @mceledonia

LHinson:
Closed by #874

PR #863, opened 2018-10-30
feat(LoginPage): introduce the login page component

amarie401:

ISSUES CLOSED: #604  

#### What:
Introduced the Login Page component

#### Note: 
* The login layout and login box are now internal components and share the same directory as the login page component. 
* The login box body will need to include the login form once that is complete and will need to be included in a separate PR. See issue #825 and #866 
* The login footer containing the social icons has been agreed to be completed in a next iteration of this component so it will be missing from this PR but the structure is in place for it. There will need to be a separate PR to include this. See issue #871


Laviro:
Looks great @amarie401 !
Feel free to use [Patternfly 3 - Login Page](https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-3/patternfly-react/src/components/LoginPage)'s logic and inner components that I had implemented. 
I guess you are facing similar challenges as i did :)

amarie401:
Hi @christiemolloy I will look into it to see why the dropdown items aren't aligning. After discussion with @jschuler, @andybraren, @LHinson, @tlabaj, and @dlabaj it was decided to split out the work for the login. The login form is currently in progress by @tlabaj so once that is built, then it will be implemented along with the button and checkbox since that will all go inside of the loginbox body. 

I used placeholder text for now just for the example docs but that will be replaced once we have the form component. I can add the static html/css for the example inside the body but not sure if we should go that route? 

I can check on the background color shade also, I'm not sure why that is. As far as the different options to login, we also decided for this iteration that the login social icons options/footer would not be included due to additional complexities and the components would have to be built for those - so it will be included in a next iteration of the login page. Also the `More...` of the login options has not been implemented in core yet in order to know how the second view screen should be implemented so we are going to hold off on that for now.

christiemolloy:
@amarie401 that sounds great. You're right, if the react component is coming from @tlabaj lets not include the static html/css. You can take out the text so that people don't confuse that for being in PF-Next. And if you want to create a React issue (if there isn't already one) to build out those login/social option links, that would be great. 

@andybraren do you know what the intended implementation is for the "More..." link?

dmiller9911:
What are your thoughts on making this another package? This uses some of the largest CSS files that core distributes (spacing is 76kb). While people using webpack 4 can potentially tree shake this is might be nice to keep react-core the true core components, and then have other packages that build these complex patterns with the core components. This will make it easier for treeshaking for people not using it, and much easier for code splitting for people that are.

 For instance. If I have an application that shows login page if the user is not authenticated, but does not if they are I might not want to bother loading any of the login page JS through dynamic imports. If I do this

**pages/LoginPage.js**
```jsx
export { LoginPage as default } from '@patternfly/react-core';
```
**App.js**
```jsx
import React, {lazy, Suspense} from 'react';
const LoginPage = lazy(() => import('./pages/LoginPage'));
const Dashboard = lazy(() => import('./pages/Dashboard')); 

function App({ isAuthenticated }) {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      {isAuthenticated ? <Dashboard /> : <LoginPage />}
    </Suspense>
  );
}
```
In that example if Dashboard also uses react-core the LoginPage js will end up in the same common bundle as button,etc which means that all login page js/css will load even if it will never get used. If we move it to it's own package and use it like below it will be split into it's own bundle and will not pollute the common bundle.

**pages/LoginPage.js**
```jsx
export { LoginPage as default } from '@patternfly/react-login-page';
```
**App.js**
```jsx
import React, {lazy, Suspense} from 'react';
const LoginPage = lazy(() => import('./pages/LoginPage'));
const Dashboard = lazy(() => import('./pages/Dashboard')); 

function App({ isAuthenticated }) {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      {isAuthenticated ? <Dashboard /> : <LoginPage />}
    </Suspense>
  );
}
```
@priley86 @jschuler  What are your thoughts on this?

andybraren:
@christiemolloy [Here's the issue](https://github.com/patternfly/patternfly-next/issues/801) for the "More..." implementation. I think it's pretty low priority.

I initially thought the background darkness was different between the two when I had both side by side, but looking closer they do seem to be the same (at least in Chrome). Is Core definitely darker for you @christiemolloy?

amarie401:
@christiemolloy I have updated to remove the placeholder text and the dropdown alignment should be fixed as well. I'm not sure about the background image darkness since we seem to be using the same as what's in core.

christiemolloy:
@amarie401 thanks for taking out the text, the dropdown is working now. Great job!

@andybraren I think it depends on the size of your window since the gradient changes, so that was my fault. They're the same.

tlabaj:
@amarie401.  I like @dmiller9911 suggestion above.  
@dgutride should probably weigh-in here.

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
```diff
@@ -0,0 +1,134 @@
...
+  /** Attribute that specifies the URL of the brand image for the LoginPage */
+  brandImgSrc: PropTypes.string,
+  /** Attribute that specifies the alt text of the brand image for the LoginPage. */
+  brandImgAlt: PropTypes.string,
+  /** Attribute that specifies the URL of the background image for the LoginPage */
+  backgroundImgSrc: PropTypes.string,
+  /** Attribute that specifies the alt text of the background image for the LoginPage. */
+  backgroundImgAlt: PropTypes.string,
+  /** Dropdown component for the LoginBox Header of the LoginPage */
+  dropdown: PropTypes.node,
```
tlabaj: I wonder if we should be more specific/opinionated in the comment here about the intended use for the dropdown. I think the name can also be more specific.

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
```diff
@@ -0,0 +1,134 @@
...
+  /** Text for the LoginBox Header of the LoginPage */
+  headerText: PropTypes.string,
+  /** URL for the LoginBox Header of the LoginPage */
+  headerUrl: PropTypes.string,
+  /** Text for the URL of LoginBox Header of the LoginPage */
+  headerUrlText: PropTypes.string,
+  /** Content rendered inside of the LoginBox Footer of the LoginPage */
+  footerContent: PropTypes.node,
+  /** Adds list variant styles */
+  variant: PropTypes.oneOf(Object.values(ListVariant))
```
tlabaj: We use variants for different things (buttons, list etc), so maybe making the name more specific here would be good. Maybe listVariants.  Thoughts?

packages/patternfly-4/react-core/src/components/LoginPage/LoginBoxHeader.js
```diff
@@ -0,0 +1,57 @@
...
+import PropTypes from 'prop-types';
+import { Title } from '../Title';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/patternfly-next/components/LoginBox/login-box.css';
+import spacing from '@patternfly/patternfly-next/utilities/Spacing/spacing.css';
+import displayStyles from '@patternfly/patternfly-next/utilities/Display/display.css';
+import flexAlignment from '@patternfly/patternfly-next/utilities/Flex/flex.css';
+
+const propTypes = {
+  /** Content rendered inside the LoginBox Header */
```
tlabaj: I wonder if there are any components here that are always required  for a loginPage Header.
@christiemolloy do you know?

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
```diff
@@ -0,0 +1,134 @@
...
+import { Text, TextContent } from '../Text';
+import { List } from '../List';
+
+const ListVariant = {
+  grid: 'grid',
+  inline: 'inline'
+};
+
+const propTypes = {
+  /** Anything that can be rendered inside of the LoginPage */
```
tlabaj: I wonder if there are any components here that are always required  for a loginPage.
@christiemolloy do you know?

packages/patternfly-4/react-core/src/components/LoginPage/LoginBoxHeader.js
christiemolloy: looping in @andybraren and @mcarrano to take a look at this

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
christiemolloy: looping in @andybraren and @mcarrano to take a look at this

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
amarie401: @tlabaj what do you recommend?

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
amarie401: listVariants sounds good to me

amarie401:
I would agree with you both @dmiller9911 and @tlabaj and see the benefits of it. Although if we are to go the package route then I would recommend the package conversion being a separate PR/issue due to my time constraints this week as this would broaden the scope of this PR. Thoughts as well @jschuler ?

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
tlabaj: I think the dropdown is specifically meant for language selection.  So maybe languageDropdown?

jschuler:
@dmiller9911 is the other css also large (flex/display..) ? Maybe if it's just spacing, instead of importing the spacing css we can use spacing tokens and create the stylesheet and use that instead.

So instead of:
```
import spacing from '@patternfly/patternfly-next/utilities/Spacing/spacing.css';
...
<Title className={css(spacing.mbSm)} size="3xl">My title</Title>
```
We could do:
```
import { StyleSheet } from '@patternfly/react-styles';
import {
  global_spacer_md
} from '@patternfly/react-tokens';

const tokenStyles = StyleSheet.create({
  title: {
    marginBottom: global_spacer_md.var
  }
});

<Title className={css(tokenStyles.title)} size="3xl">My title</Title>
```

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.d.ts
```diff
@@ -0,0 +1,29 @@
+import { SFC, HTMLProps, ReactNode } from 'react';
+import { OneOf } from '../../typeUtils';
+
+export const ListVariant: {
```
jschuler: Should the name be more unique? Otherwise if there is another ListVariant being exported (e.g. from List package) and the user wants to import { ListVariant } from '@patternfly/react-core' it could spell trouble i think?

amarie401: any suggestions on naming for this? @jschuler

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.d.ts
```diff
@@ -0,0 +1,29 @@
...
+  grid: 'grid';
+  inline: 'inline';
+};
+
+export interface LoginPageProps extends HTMLProps<HTMLElement> {
+  children: ReactNode;
```
jschuler: all of these should be specified as optional `?`

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
jschuler: How about just let the user pass in dropDownItems and a callback if they want to be notified on dropdown selection, and we handle the isOpen state?

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
```diff
@@ -0,0 +1,134 @@
...
+  /** Attribute that specifies the URL of the background image for the LoginPage */
+  backgroundImgSrc: PropTypes.string,
+  /** Attribute that specifies the alt text of the background image for the LoginPage. */
+  backgroundImgAlt: PropTypes.string,
+  /** Dropdown component for the LoginBox Header of the LoginPage */
+  dropdown: PropTypes.node,
+  /** Content rendered inside of the Text Component of the LoginPage */
+  textContent: PropTypes.string,
+  /** Title for the LoginBox Header of the LoginPage */
+  title: PropTypes.string,
```
jschuler: Maybe to make it more obvious where this is used it should be called loginTitle.
In general, maybe we can prefix these, i.e. things that go into the login box with 'login', and things that go into the main? box with 'main'.

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
```diff
@@ -0,0 +1,134 @@
...
+  /** Items rendered inside of the List Component of the LoginPage */
+  listItems: PropTypes.node,
+  /** Text for the LoginBox Header of the LoginPage */
+  headerText: PropTypes.string,
```
jschuler: Instead of 3 props for header, can we have just one `loginSubtitle` perhaps, that accepts a node? Then they can do
```
const subtitle = <p>headerText <a href="headerUrl">headerUrlText</a></p>;
loginSubtitle={subtitle}
```

packages/patternfly-4/react-core/src/components/LoginPage/examples/SimpleLoginPage.js
```diff
@@ -0,0 +1,98 @@
...
+        title="Login to your account"
+        dropdown={myDropdown}
+        headerText="Need an account?"
+        headerUrl="https://www.patternfly.org"
+        headerUrlText="Sign up."
+        brandImgSrc={brandImg}
+        brandImgAlt="pf-logo"
+        backgroundImgSrc={images}
+        listItems={listItem}
+        varient="inline"
```
jschuler: is this a real prop?

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
andybraren: I believe all of these would be required for any generic LoginPage component, except maybe “Brand” (if that’s a PF-brand-specific thing).

packages/patternfly-4/react-core/src/components/LoginPage/LoginBoxHeader.js
andybraren: I believe all of these would be required for any LoginPage Header. Display and Flex certainly, Spacing probably unless the dev wanted to make something super custom.

(Caveat: I’m not at all familiar with how devs use React to build things, I could be wrong.)

amarie401: So everything is required for the Header here?

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.d.ts
amarie401: Looks like everything is required according to @andybraren 's comments

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
jschuler: I think for ease of use we should not require some of these, for example a consumer might not need a language dropdown. I think the only required piece really should be the login form when it gets added later and maybe the loginTitle

packages/patternfly-4/react-core/src/components/index.js
```diff
@@ -13,7 +13,7 @@ export * from './Dropdown';
 export * from './Form';
 export * from './Label';
 export * from './List';
-export * from './LoginBox';
+export * from './LoginPage';
```
jschuler: Same update needs to be done for index.d.ts it looks like

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.js
amarie401: I agree, I think that makes sense @jschuler  also certain components can't be required at the moment for example like loginFooterContent since we don't have the social icons component yet. Might be a bit much to require everything for this.

amarie401: I think the loginTitle and login form should be required when that gets added as well as the brandImgAlt and backgroundImgAlt since we want them to provide alt text.

packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.d.ts
amarie401: As per your comment above, I'll fix this -  as loginTitle and form and the alt text should be the required props and the rest optional

jschuler:
@dmiller9911 yes agree, styled-system would be useful here to reduce the size. going to circle back to styled-system and probably put it in a react-core-extensions package due to request @dgutride

PR #868, opened 2018-10-31
Charts: Added initial area, bar, donut, line, pie & stack components

dlabrecq:
This PR introduces various chart components for a new package, named react-charts. There are ~13 components total, used to create Victory charts such as; area, bar, donut, line, pie, & stack. (Victory doesn't have a donut, but I was able to style one to match.) The other components are required to support custom containers, legend, tooltip, etc. 

Along with some basic unit tests, I also created dark and light themes, based on existing Victory functionality. Each theme has blue, green, & multiple color variations  (6 total) per the mocks below. I realize this is putting the cart before the horse, but the intention is that these components and themes will help @mcoker 's styling efforts in PF-Next.

Mocks:
https://drive.google.com/file/d/1jCxBH0niB78V0PJvzHDBr8Hm5wKmDlgH/view

Note that PF-next only has some of the color variations shown in the mocks -- it's missing quite a few. For example, we only have two greens and three shades of blue I could use for the examples -- no suitable yellow. And there are currently no color variables defined for charts in PF-Next.

Also note there are two charts I did not implement. The bullet chart is a custom HTML/CSS implementation in PF3 -- Victory's Box Plot has similarities, but it's not a best fit. Victory also does not have a heat map; although, the Angular PatternFly implemented its component using D3 directly. Thus, the plan is to port heat map from a-pf later via another PR.

That said, I attempted to add property typings to the chart components, but decided to hold off for now after speaking with @dgutride. This is yet another Typescript issue I'd like to discuss with the Tiger team. Victory provides interfaces that I'd like to use directly when Typescript is enabled.

Before going any further, I'd like to stop here and merge. Obviously there will be revisions, but this is a ton of code. This package will remain alpha for a bit, while the charts can idle until pf-next is ready with styling.

bar-chart

pie-chart

donut-chart

line-chart

stack-chart

area-chart


dlabrecq:
Thanks, Patrick. I intentionally created the examples because the mocks have so many variations. The intention with the extra examples is mainly to help the pf-next styling effort. I expect we will eliminate some examples and themes after the colors and other variables are defined for pf-next.

packages/patternfly-4/react-charts/src/components/ChartArea/ChartArea.js
```diff
@@ -0,0 +1,22 @@
...
+import { default as ChartTheme } from '../ChartTheme/ChartTheme';
+
+export const propTypes = {
+  /**
+   * See TypeScript API docs: https://formidable.com/open-source/victory/docs/victory-area/
+   */
+  '': PropTypes.any
+};
+
+export default class ChartArea extends VictoryArea {
```
dmiller9911: We should not be extending these. It is possible that they could internally remove clasess, etc. A better, and more "React" way, would be to do something like
```jsx
const ChartArea = (props) => <VictoryArea {...props} />

ChartArea.propTypes = propTypes;
ChartArea.defaultProps = {
   ...VictoryArea.defaultProps,
   theme: ChartTheme.default,
}
```

Another option is to do something like
```jsx
const ChartArea = (props) => <VictoryArea them={ChartTheme.default} {...props} />
```
That would pass the default theme down, and allow the user to overwrite it with a theme prop.

dlabrecq: I tried your suggested approach with ChartArea. It renders alone, but won't render as a child of VictoryGroup (i.e., Victory's group container). Looking at the Victory source, there may be other properties being referenced when rendering children (e.g., role, expectedComponents, etc)?

dmiller9911: That is definitely possible. It looks like they do for sure use role quite a bit. Without digging too much I am not sure the best way to handle this, but extending and then calling super.render is not something we should be doing. there is no garuntee that these components are classes. If they ever change (could even be in a minor) it would break. Example
```jsx
function SomeChartFromVictory() {
	return <div/>;
}

class OtherChart extends SomeChartFromVictory {
  render() {
    return super.render();
  }
}
const comp = new OtherChart();
comp.render();
```
That would throw an error.

Victory does have a section for Wrapping components: https://formidable.com/open-source/victory/guides/custom-components/#wrapping-components. That might be useful to look into and could be an option. I have not had time to look into this, and may not anytime soon so I am not positive that is the best option though.

dlabrecq: This doesn't really apply to the ChartArea, tho -- it has no children. That would apply more to a container like ChartGroup. 

To eliminate that wrapping code from the equation, I've been testing ChartArea as a child of VictoryGroup.

jschuler: Since we're wrapping these components perhaps we can extend them for now and then iterate on this later and see if we can compose and pass props through in a way that Victory likes

dlabrecq: Yes, iterating on this would be my preference.

FYI, I was able to get a few components to work using Doug's approach, by adding extra properties. For example, most components required a `role` in order to render, while ChartVoronoiContainer also required `defaultEvents`. And ChartBar required `getDomain`. 

Unfortunately, I've yet to get ChartTooltip to work properly. It renders with ChartVoronoiContainer, but not with ChartPie or ChartDonut. Adding properties seems more fragile than extending because they could change.

I also noticed a new violation message with this approach that is not present when extending.

`[Violation] 'DOMContentLoaded' handler took 422ms`

dlabrecq: I have the components working, using Doug's suggested approach, but had to add some extra properties. For example, most components required a `role` in order to render, while ChartVoronoiContainer also required `defaultEvents`. ChartBar required `getDomain` and ChartStack needed `getChildren`.

I wanted to use Object.assign to add all properties, but that doesn't work in all cases (e.g., I still needed to add `role` & `getChildren` to ChartStack) and some inherited properties were invalid.

This seems more fragile than just extending the class? That is, what's currently a required property could change.

dmiller9911: This might help: https://github.com/mridgway/hoist-non-react-statics That will allow us to not worry about what static properties each of this components need. We can hoist them up to our component.

dlabrecq: Yes, that is exactly what was needed. Thank you!

mcarrano:
I think this is a good start @dlabrecq . I'd also like to invite @mceledonia and @rachael-phillips to comment from a design perspective. The only question I have is what flexibility we have on labeling through React props. I can't really get a sense of that here. I know in using PF3 charts, this was always one of the most problematic things when teams wanted to adjust the labeling. We may need to think about more design guidance here.

dlabrecq:
Victory uses compilation heavily, so there are lots of customization options. For example, if you don't like the tooltip, an React developer could provide their own component. Likewise, if you don't appreciate the labeling on the axis, you can potentially provide your own container and axis components. You could even create your own legend, if desired.

PR #870, opened 2018-11-01
fix(BadgedResource): Use css over inlined style for bg color

jeff-phillips-18:
Also provides defaults for some resource kinds.

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/src/components/BadgedResource/ResourceTypes.js
+++ b/packages/patternfly-3/patternfly-react-extensions/src/components/BadgedResource/ResourceTypes.js
@@ -1,4 +1,6 @@
-import BadgedResource from './BadgedResource';
+import 'patternfly/dist/js/patternfly-settings';
+
+const { patternfly } = window;
 
 const resourceTypes = {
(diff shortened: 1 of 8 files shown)


packages/patternfly-3/patternfly-react-extensions/src/components/BadgedResource/ResourceTypes.js
```diff
@@ -1,4 +1,6 @@
-import BadgedResource from './BadgedResource';
+import 'patternfly/dist/js/patternfly-settings';
+
+const { patternfly } = window;
 
 const resourceTypes = {
```
priley86: so will `resourceTypes` and `kindAbbrs` be referenced globally here downstream? or only when referencing BadgedResource? Just trying to follow this piece...

jeff-phillips-18: It's here for the BadgedResource as helpers in an attempt to keep applications consistent. I doubt downstream would reference it for anything other than the BadgedResource but it would be possible. They do map directly to what is being used in Console atm.

PR #872, opened 2018-11-01
feat(breadcrumb): adds PF4 breadcrumb, breadcrumbItem, breadcrumbTitle components

### LINKED ISSUE #856: PF4 Breadcrumbs Component
Introduce PF4 Breadcrumbs component: https://pf-next.com/components/breadcrumb/examples/
See closed core issue for more details: https://github.com/patternfly/patternfly-next/issues/33

Known changes underway in core can be seen here: https://github.com/patternfly/patternfly-next/issues/832

In addition to a React review, this issue should be reviewed by the core contributors of this demo:
CSS: @michael-coker

rvsia:
fix #856

**What**:

- adds breadcrumbs, breadcrumbItem, breadcrumbHeading components

**Link to docs**:

https://872-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/breadcrumb/

**Additional issues**:

- none

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.d.ts
@@ -1,0 +1,13 @@
+import { SFC, HTMLProps, ReactNode, ReactType } from 'react';
+
+export interface BreadcrumbHeadingProps extends HTMLProps<HTMLLIElement> {
+    children?: ReactNode;
+    className?: string;
+    to?: string;
+    target?: string;
+    component?: ReactType<BreadcrumbHeadingProps>;
+}
+
+declare const BreeadcrumbHeading: SFC<BreadcrumbHeadingProps>;
+
+export default BreeadcrumbHeading;
--- a/packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.js
+++ b/packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.js
@@ -1,0 +1,49 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Breadcrumb/breadcrumb.css';
+import PropTypes from 'prop-types';
+import { css } from '@patternfly/react-styles';
+import { componentShape } from '../../internal/componentShape';
+
+const propTypes = {
+  /** Content rendered inside the breadcrumb title. */
+  children: PropTypes.node,
+  /** Additional classes added to the breadcrumb item. */
+  className: PropTypes.string,
+  /** HREF for breadcrumb link. */
+  to: PropTypes.string,
+  /** Target for breadcrumb link. */
+  target: PropTypes.string,
+  /** Sets the base component to render. Defaults to <a> */
+  component: componentShape
+};
+
+const defaultProps = {
+  children: null,
+  className: '',
+  to: null,
+  target: null,
+  component: 'a'
+};
+
+const BreadcrumbHeading = ({ className, children, to, target, component: Component, ...props }) => (
+  <li {...props} className={css(styles.breadcrumbItem, className)}>
+    <h1 className={css(styles.breadcrumbHeading)}>
+      {to && (
+        <Component
+          href={to}
+          target={target}
+          className={css(styles.breadcrumbLink, styles.modifiers.current)}
+          aria-current="page"
+        >
+          {children}
+        </Component>
+      )}
+      {!to && <React.Fragment>{children}</React.Fragment>}
+    </h1>
+  </li>
+);
+
+BreadcrumbHeading.propTypes = propTypes;
+BreadcrumbHeading.defaultProps = defaultProps;
+
+export default BreadcrumbHeading;
--- a/packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbItem.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbItem.d.ts
@@ -1,0 +1,14 @@
+import { SFC, HTMLProps, ReactNode, ReactType } from 'react';
+
+export interface BreadcrumbItemProps extends HTMLProps<HTMLLIElement> {
+    children?: ReactNode;
+    className?: string;
+    to?: string;
+    isActive?: boolean;
+    target?: string;
+    component?: ReactType<BreadcrumbItemProps>;
+}
+
+declare const BreeadcrumbItem: SFC<BreadcrumbItemProps>;
+
+export default BreeadcrumbItem;
--- a/packages/patternfly-4/react-core/src/components/Breadcrumb/Breadcrumb.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Breadcrumb/Breadcrumb.d.ts
@@ -1,0 +1,11 @@
+import { SFC, HTMLProps, ReactNode } from 'react';
+
+export interface BreadcrumbProps extends HTMLProps<HTMLElement> {
+    children?: ReactNode;
+    className?: string;
+    'aria-label'?: string;
+}
+
+declare const Breadcrumb: SFC<BreadcrumbProps>;
+
+export default Breadcrumb;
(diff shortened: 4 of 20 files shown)


dmiller9911:
For the breadcrumb divider, it looks like core chose to use FontAwesome inside a `::before` element instead of rendering an actual icon. We will need to use an icon @matthewcarleton Is it possible to have core drop the before element and use a real icon element so that React can easily implement this, or is this something React will also need to work around? 

@rvsia Instead of using `isTitle` and having Breadcrumbs do the work what about having a BreadcrumbsTitle component instead
```jsx
<Breadcrumbs>
  <BreadcrumbsItem />
  <BreadCrumbsItem />
  <BreadcrumbsItem />
  <BreadcrumbsTitle />
</Breadcrumbs>
```
This would making opting into the title easier, and make the render logic much simpler.

As a note to make `BreadcrumsItem` support react-router link a `component` prop is going to need to be added that will default to `a`.  Card has an example if this [here](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Card/Card.js#L19).

rvsia:
@dmiller9911 Thanks for the feedback! BreadcrumbTitle component should be fine, I will change it.

If we change `a` for `Link`, will `href` attribute work on `Link` component?

dmiller9911:
@rvsia It will not, but It will be easy for consumers to map it

```jsx
import { BreadCrumbsItem, Breadcrumbs } from '@patternfly/react-core';
import { Link as ReactRouterLink } from 'react-router-dom';

const CustomLink = ({ href, ...props }) => <ReactRouterLink to={href} {...props} />
const CustomBreadcrumsItem = (props) => <BreadcrumbsItem component={CustomLink} />

export {
  Breadcrumbs,
  CustomBreadcrumbsItem as BreadcrumbsItem
}
```

rvsia:
@christiemolloy Thanks for the feedback!

1. So, should I wait for the update in core?
2. The last example is made just for a demonstration that consumers can put custom elements, components, etc. in the breadcrumbItem and it doesn't have to be just a string. And it is their work to make it look good, they have freedom to do it. And I'm using badges as examples, because I really like how they look. :smile: [I updated the last badge] I can imagine that consumers create breadcrumbs with icons or something like that, (but at this moment there is no component/icon which works well in that context). However, I have no problem with removing it.

christiemolloy:
@rvsia 

1. Yeah wait for the update in core
2. I like your idea of extending the design, however all design must come from patternfly-core and the UI/visual designers. So unfortunately we have to remove it from the React example. However, circling in @mcarrano to take a look at the last custom example to see if that is something we would want to create an issue for in the future. But for now we will have to remove it. 

Thank you!!

mcarrano:
I definitely like the idea of being able to include custom elements in a breadcrumb, but as @christiemolloy said. We'd want to implement that in Core first. @rvsia would you mind opening a new issue in patternfly-next requesting this as an enhancement? If you tag me, I will endorse it as something we should consider. Thanks.

matthewcarleton:
@rvsia, @dmiller9911  hey sorry for the delayed response. Ya I think we can switch to a font in core if it makes more sense for React. I'll reach out to the dev working on it and get it updated.

rvsia:
Updated to changes in patternfly-next and removed WIP label

- breadcrumbTitle changed to breadcrumbHeading component
- updated to newest version of PF4 (needed to run it)
- uses Patternfly-react icon as the divider (instead of ::before element)

christiemolloy:
@rvsia great job! thanks for making the changes - CSS approved

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbItem.d.ts
```diff
@@ -0,0 +1,14 @@
+import { SFC, HTMLProps, ReactNode, ReactType } from 'react';
+
+export interface BreadcrumbItemProps extends HTMLProps<HTMLDivElement> {
+    children?: ReactNode;
+    className?: string;
+    to?: string;
+    isActive?: boolean;
```
dlabaj: isActive is used to indicate the current page that the user is on within the breadcrumb.  How does this work with the BreadcrumbHeading.  The expectation is that the BreadcrumbHeading is the same as the current active BreadcrumbItem.

dlabaj: Here's an example from PF3

http://patternfly-react.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react%2FNavigation%2FBreadcrumb&selectedStory=Breadcrumb&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

rvsia: BreadcrumbHeading is always 'active'.

The main difference is that the title heading in PF3 is controlled by CSS on the breadcrumb list => you set title class on the breadcrumb list and the active item will inherit different attributes.

In PF-next you set the title heading class only on `<h>` element in the last child. (And the structure of the heading and a regular item is also different.)

If we want to set the title prop on the list, we will have to add some `breadcrumbContext` to share props and change rendering logic in the breadcrumbItem to something more difficult to render and read. I think it would be too unnecessarily complicated for a list. Better solution would be make changes in CSS to be the same as in PF3.

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.d.ts
```diff
@@ -0,0 +1,13 @@
+import { SFC, HTMLProps, ReactNode, ReactType } from 'react';
+
+export interface BreadcrumbHeadingProps extends HTMLProps<HTMLDivElement> {
```
boaz1337: I think it will be better to extend from `HTMLProps<HTMLLIElement>`.

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.d.ts
```diff
@@ -0,0 +1,13 @@
...
+    children?: ReactNode;
+    className?: string;
+    to?: string;
+    target?: string;
+    component?: ReactType<BreadcrumbItemProps>;
```
boaz1337: I've never heard of `ReactType` maybe you're talking about `ReactElement`?

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbItem.d.ts
```diff
@@ -0,0 +1,14 @@
+import { SFC, HTMLProps, ReactNode, ReactType } from 'react';
+
+export interface BreadcrumbItemProps extends HTMLProps<HTMLDivElement> {
```
boaz1337: I believe this should extend `HTMLProps<HTMLElement>` since the props are passed to `nav` and not to `div`.

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.d.ts
rvsia: We are using `ReacType` in all components where a component prop is used.

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbHeading.js
```diff
@@ -0,0 +1,49 @@
...
+  target: null,
+  component: 'a'
+};
+
+const BreadcrumbHeading = ({ className, children, to, target, component: Component, ...props }) => (
+  <li {...props} className={css(styles.breadcrumbItem, className)}>
+    <h1 className={css(styles.breadcrumbHeading)}>
+      {to && (
+        <Component
+          href={to}
```
jschuler: href and target only make sense for anchor tags, no? So in this case shouldn't the Component always be `a`?

jschuler: Ah, I see @dmiller9911  example above. That's nice then

packages/patternfly-4/react-core/src/components/Breadcrumb/BreadcrumbItem.d.ts
dlabaj: Can we remove isActive then since it does the same thing as the title.

dlabaj: I would remove this then since it's the same the Breadcrumb title.

PR #873, opened 2018-11-01
chore(Release script): Remove rawgit from release script

### LINKED ISSUE #849: FYI: rawgit is shutting down
See the announcement on https://rawgit.com/ :crying_cat_face: 
Actively used links will remain working for a while but should find an alternative...
https://raw.githack.com/ is one alternative.

ibolton336:
**What**:
Remove rawgit from the release scripts. Closes #849 

**Additional issues**:


priley86:
I think we can safely remove rawgit now that surge.sh is in use. thoughts @dgutride ?

priley86:
do we have any PF Core links still referencing PF React Rawgit? Ideally we can forward those all to surge now...

PR #874, opened 2018-11-02
feat(LoginForm): Added the PF4 LoginForm Component   

tlabaj:
**What**:
Added the PF4 LoginForm Component   
   #825

**Additional issues**:


tlabaj:
@christiemolloy Thanks for the review!
I will update the example for point 1 and 3.
Can you please let me know more about item 2 and where I can find the design for this.

christiemolloy:
@andybraren which design did you follow for the form in the login box with the requirement pop over?

christiemolloy:
@andybraren yes, could you point me to the design for the required warning pop-over  "Please fill out this field", that occurs when the user clicks the log-in button and the fields are empty.

andybraren:
@christiemolloy Ah, I see. That popup is Chrome's default behavior for any `input` with a `required` attribute that is empty when the form's submit button is clicked.

Changing `type="button"` of this demo's button to `type="submit"` restores the browser's behavior. I'll create a new issue to add `'type="submit"` to the button in Core's Login demo as well.

Good eye Christie!

christiemolloy:
Great thanks for solving that @andybraren !

christiemolloy:
@tlabaj thanks for making those changes this looks great. Tiny thing i didnt catch before - could you change "Log In" to "Log in" per our new capitalization rules. Thanks! CSS/UX approved!

amarie401:
@tlabaj perhaps rename this to LoginForm, the demo/example isn't showing.

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
```diff
@@ -0,0 +1,177 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@patternfly/react-styles';
+import flexStyles from '@patternfly/patternfly-next/utilities/Flex/flex.css';
+import displayStyles from '@patternfly/patternfly-next/utilities/Display/display.css';
+import alignmentStyles from '@patternfly/patternfly-next/utilities/Alignment/alignment.css';
+import spacingStyles from '@patternfly/patternfly-next/utilities/Spacing/spacing.css';
```
jschuler: I guess there is a similar concern here that this will increase the size of the component too much. Could do something similar as here:
https://github.com/patternfly/patternfly-react/pull/863/files#diff-4e9f1155160426190b5a2ee57a9c0a5fR32
Or use styled-component when the PR is merged :)

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
```diff
@@ -0,0 +1,177 @@
...
+  /** Function that is called when the Login button is clicked */
+  onLoginButtonClick: PropTypes.func,
+  /** Label for the Remember Me Checkbox that indicates the user should be kept logged in */
+  rememberMeLabel: PropTypes.string,
+  /** Flag indicating if the remember me Checkbox is checked. */
+  isRememberMeChecked: PropTypes.bool,
+  /** Function that handles the onChange event for the Remember Me Checkbox */
+  onChangeRememberMe: PropTypes.func,
+  /** Aria Label for the Remember me checkbox */
+  rememberMeAriaLabel: PropTypes.string.isRequired
```
jschuler: should only be required if the rememberMeLabel is also passed in

tlabaj: The Checkbox component requires the Aria Label

jschuler: Yes, but in this form having a checkbox with no label would not make sense, so I think if you require the aria-label you should also require the rememberMeLabel. But i could see consumers maybe not wanting to have this checkbox, so perhaps add the logic here, if rememberMeAriaLabel is passed in, then require the aria-label as well. And render the Checkbox conditionally if rememberMeLabel is used.

packages/patternfly-4/react-core/src/components/LoginPage/examples/SimpleLoginPage.js
```diff
@@ -81,6 +97,26 @@ class SimpleLoginPage extends React.Component {
...
       </React.Fragment>
     );
+
+    const loginForm = (
+      <LoginForm
+        usernameLabel="Username"
+        usernameValue={this.state.usernameValue}
+        onChangeUsername={this.handleUsernameChange}
+        usernameHelperTextInvalid="Unknown Username"
+        isValidUsername
```
jschuler: Can you tie this to a state variable, and maybe in the handleUsernameChange check if username is less than 3 chars mark it as invalid? And something similar for password. So we can play with invalid state in the example

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
```diff
@@ -0,0 +1,185 @@
...
+  passwordLabel: 'Password',
+  passwordValue: '',
+  onChangePassword: () => undefined,
+  passwordHelperText: '',
+  passwordHelperTextInvalid: '',
+  isValidPassword: true,
+  loginButtonLabel: 'Log In',
+  isLoginButtonDisabled: false,
+  onLoginButtonClick: () => undefined,
+  rememberMeLabel: undefined,
```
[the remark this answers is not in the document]
tlabaj: Don had suggested the undefined for the check.  I can switch it back.

packages/patternfly-4/react-core/src/components/LoginPage/examples/SimpleLoginPage.js
tlabaj: Can you open an enhancement issue to add this so we can get the form merged.  the invalid state is really a function of the form.

PR #876, opened 2018-11-02
feat(condensed horizontal nav): updates condensed nav to match Core horizontal & vertical nav

priley86:
**What**:
Changes condensed nav implementation to match Core's new `pf-m-tall` class. The nav is no longer "tall" by default and this class is only applied when using the `useCondensed` feature on desktop. 
See: https://github.com/patternfly/patternfly-next/pull/884

This also introduces the need to consume the `isTall` prop within the `NavList` for Horizontal Navs, thus I have introduced a `PageContext` which will provide this context to any child elements needing this prop. If the `Nav` is wrapped with a `Page`, it will consume the prop from PageContext instead. This same context can also now be used in the `PageHeader` as well. 

**Additional issues**:


packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -33,54 +34,71 @@ const defaultProps = {
...
     const { useCondensed } = this.props;
     if (useCondensed) {
       this.mainRef.current.removeEventListener('scroll', this.handleScroll);
+      window.removeEventListener('resize', this.handleResize);
     }
   }
 
+  // only enable tall header if the user is using the `useCondensed` header feature and we are on desktop
+  isTall = props =>
+    typeof window !== 'undefined' && window.innerWidth >= parseInt(breakpointMd.value, 10) && props.useCondensed;
```
jschuler: Nit: Maybe check props.useCondensed first

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -33,54 +34,71 @@ const defaultProps = {
...
-      const mainPosition = main.scrollTop;
-      if (mainPosition > scrollingDistance && !isCondensed) {
-        this.setState({ isCondensed: true });
-      } else if (mainPosition < scrollingDistance && isCondensed) {
-        this.setState({ isCondensed: false });
-      }
-    });
+    const { scrollingDistance, useCondensed } = this.props;
+    const { isTall } = this.state;
+    if (useCondensed && window.innerWidth >= parseInt(breakpointMd.value, 10)) {
```
jschuler:
```suggestion
    if (this.isTall(this.props)) {
```

packages/patternfly-4/react-core/src/layouts/Page/Page.js
priley86: giving this a spin..

packages/patternfly-4/react-core/src/layouts/Page/Page.js
priley86: 👍

packages/patternfly-4/react-core/src/components/Nav/NavList.js
```diff
@@ -3,24 +3,27 @@ import PropTypes from 'prop-types';
...
 
 const propTypes = {
   /** Indicates the list type. */
   variant: PropTypes.oneOf(Object.values(NavVariants)),
   /** Children nodes */
   children: PropTypes.node,
   /** Additional classes added to the list */
-  className: PropTypes.string
+  className: PropTypes.string,
+  /** nav should have tall modifer */
```
jschuler: Can you make this comment more descriptive? Maybe something like: Horizontal nav only: If true makes the nav taller

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -18,7 +19,7 @@ const propTypes = {
   header: PropTypes.node,
   /** Sidebar component for a side nav (e.g. <PageSidebar />) */
   sidebar: PropTypes.node,
-  /** condensed header on scroll */
+  /** enable condensed header on scroll feature (for desktop only) */
```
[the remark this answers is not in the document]
priley86: added comments

packages/patternfly-4/react-core/src/components/Nav/NavList.js
priley86: added comments

PR #877, opened 2018-11-02
fix(FilterSidePanel/VerticalTabs): Adjust top and bottom spacing of f…

rhamilto:
…ilter-side-panel and vertical-tabs-pf

.filter-panel-pf has top and bottom padding, which causes layout issues when it's the only item
being used.  Removing the top and bottom padding and adding a bottom margin fixes the bug and
provides the proper spacing if another element is used.  As a result of these changes,
.vertical-tabs-pf needs to have a larger bottom margin set on it.

Before:
screen shot 2018-11-02 at 10 09 34 am
screen shot 2018-11-02 at 10 09 44 am

After:
screen shot 2018-11-02 at 10 09 11 am
screen shot 2018-11-02 at 10 09 04 am

attn: @jeff-phillips-18

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/less/filter-side-panel.less
+++ b/packages/patternfly-3/patternfly-react-extensions/less/filter-side-panel.less
@@ -1,5 +1,6 @@
 .filter-panel-pf {
-  padding: 20px 15px;
+  margin: 0 0 30px;
+  padding: 0 15px;
 }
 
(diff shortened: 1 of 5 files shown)


rhamilto:
@dtaylor113 mind reviewing so this can merge?  We've got another fix we want to pull in to console, but don't want to bump the dependency twice given the hassle it currently is.

PR #878, opened 2018-11-02
fix(modal transition): Reduce translate3d (right offset) positioning for right-side-modal-pf to 25%

sg00dwin:
This is a requested modification related to the OpenShift Console Catalog pr
https://github.com/openshift/console/pull/666

**What**:
Reduce offset position for right-side-modal-pf to lessen transition distance and ui movement.

@jeff-phillips-18 

**Additional issues**:
Requested for OpenShift Console Catalog from pr https://github.com/openshift/console/pull/666


sg00dwin:
@jeff-phillips-18

PR #883, opened 2018-11-05
fix(overlay): add anchor tag to filter overlay

dgutride:
**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/demos/PageLayout/examples/PageLayoutDefaultNav.js
```diff
@@ -174,7 +174,7 @@ class PageLayoutDefaultNav extends React.Component {
       [BackgroundImageSrc.xl]: '/assets/images/pfbg_2000.jpg',
       [BackgroundImageSrc.xs]: '/assets/images/pfbg_576.jpg',
       [BackgroundImageSrc.xs2x]: '/assets/images/pfbg_576@2x.jpg',
-      [BackgroundImageSrc.filter]: '/assets/images/background-filter.svg'
+      [BackgroundImageSrc.filter]: '/assets/images/background-filter.svg#image_overlay'
```
dmiller9911: I asked Dan about this earlier. Is there anyway to remove the need for this? We keep watching it in the docs. I would love to see a more permanent solution if there is one available. Doesn't need to block this pr, but it would be nice to avoid the need for the anchor.

PR #884, opened 2018-11-05
Typescript support for react-core

dlabaj:
Initial typescript support with Babel 6.

**What**:

Adds support for typescript to react-core only.  Currently uses tsc for typescript transpiling and babel 6. 

**Additional issues**:

Still need to add tslint.  Should also convert a component to typescript to work out any issues.

Gatsby Demo can be found here: http://chilly-paste.surge.sh/


packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -1,7 +1,9 @@
-import React from 'react';
-import styles from '@patternfly/patternfly-next/components/Avatar/avatar.css';
+import * as React from 'react';
```
dmiller9911: For this you can set the following in tsconfig.json
```
{
  "compilerOptions": {
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    ...
  }
}
```

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -1,7 +1,9 @@
...
+import * as styles from '@patternfly/patternfly-next/components/Avatar/avatar.css';
 import { css } from '@patternfly/react-styles';
-import PropTypes from 'prop-types';
+import * as PropTypes from 'prop-types';
+
+let _styles: any = styles;
```
dmiller9911: should be able to place a d.ts file somewhere under src with the following
```ts
declare module '*.css' {
  const styles: any;
  export default styles;
}
```

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -19,9 +19,28 @@ const defaultProps = {
   src: ''
 };
 
-const Avatar = ({ className, src, alt, ...props }) => (
-  <img {...props} className={css(_styles.avatar, className)} src={src} alt={alt} />
-);
+/**
+ * Column properties.
+ */
+export interface IAvatarProps {
```
dmiller9911: I am not super opinionated about this, but I know most other typescript libs have dropped prefixing interfaces with `I` it was kind of a holdout from C#/Java, and most people in the JS/TS community prefer leaving it off.

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -19,9 +19,28 @@ const defaultProps = {
...
+  /** Additional classes added to the Avatar. */
+  className?: string;
+  /** Attribute that specifies the URL of the image for the Avatar. */
+  src?: number;
+  /** Attribute that specifies the alt text of the image for the Avatar. */
+  alt: string
+}
+
+class Avatar extends React.Component<IAvatarProps, {}> {
```
dmiller9911: This can be `const Avatar: React.SFC<AvatarProps>` instead of using the class.

dmiller9911: Additionally for classes the second generic (state) defaults to `{}`. So no need to specify it if state is not being used.

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,48 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (<img {...this.props.props} className={css(_styles.avatar, this.props.className)} src={this.props.src} alt={this.props.alt} />);
```
dmiller9911: `{...this.props}`

Also `src` and `alt` should spread with the props object.

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dlabaj: @dmiller9911 This worked for style and PropTypes but I get no default error for React so I had to still do * as React.  Thoughts?

dlabaj: Got it working

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dlabaj: Done add a typings folder at the same level as src in react-core

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dlabaj: Changed it to AvatarProps

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dlabaj: Done.

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dlabaj: removed src and alt since they are being spread.

packages/patternfly-4/package.json
```diff
@@ -30,5 +30,8 @@
     "lint:versions": "node ./verifyCoreVersions.js",
     "lint:styles": "yarn stylelint",
     "stylelint": "stylelint '**/sass/**/*.scss' !**/dist/**/*.scss"
+  },
+  "dependencies": {
+    "@types/jest": "^23.3.10"
```
jschuler: Also added in the root package. Can we move the types closer to react-core?

dlabaj: Just tried removing these, but remember that these are needed because jest runs at the root and it can't find the @types causing the tests to fail.

Issue #889, opened 2018-11-06
Dropdowns w/ Icons and custom Buttons


priley86:
I attempted adding an Icon/Button to our Dropdown toggle and it seems the spacing is slightly off. This appears to be due to the fact that `<Button>` will apply a default class of `pf-c-button pf-m-plain` adding the space. If I add `pf-c-dropdown__toggle` to the Button's `className` prop, I will receive `pf-c-button pf-m-plain pf-c-dropdown__toggle` classes (still adding the space). For now, the easy workaround is to just use a jsx `button`.  Maybe we should add a prop specifying type to `Button`, so we can say type "dropdown" only applies `pf-c-dropdown__toggle`? Please advise on the desired effect here for icon dropdown's.

Tried code:
```
      <Toolbar>
        <ToolbarGroup>
          <ToolbarItem className={classNames('pf-u-sr-only', 'pf-u-visible-on-md')}>
            <Dropdown
              isPlain
              direction="top"
              onSelect={this._onSettingsDropdownSelect}
              toggle={
                <Button onClick={this._onSettingsDropdownSelect} id="settings-button" aria-label="Settings actions" className="pf-c-dropdown__toggle">
                  <CogIcon />
                </Button>
              }
              isOpen={isSettingsDropdownOpen}
            >
              <DropdownItem>About</DropdownItem>
            </Dropdown>
          </ToolbarItem>
```

Workaround code:

```
      <Toolbar>
        <ToolbarGroup>
          <ToolbarItem className={classNames('pf-u-sr-only', 'pf-u-visible-on-md')}>
            <Dropdown
              isPlain
              direction="top"
              onSelect={this._onSettingsDropdownSelect}
              toggle={
                <button onClick={this._onSettingsDropdownSelect} id="settings-button" aria-label="Settings actions" className="pf-c-dropdown__toggle">
                  <CogIcon />
                </button>
              }
              isOpen={isSettingsDropdownOpen}
            >
              <DropdownItem>About</DropdownItem>
            </Dropdown>
          </ToolbarItem>
```

jschuler:
Hi @priley86 , I think using DropdownToggle and having the option of hiding the caret solves your needs right? I am closing this issue, but if there's still an open question around this please let me know, thanks!

PR #893, opened 2018-11-07
feat(DropdownToggle - noCaret option): Adds the noCaret option

jschuler:
**What**:

OpenShift needs an option to disable the caret from displaying for the DropdownToggle

Addresses:
https://github.com/patternfly/patternfly-react/issues/889#issuecomment-436458249

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
@@ -5,7 +5,7 @@ import Toggle from './Toggle';
 import styles from '@patternfly/patternfly-next/components/Dropdown/dropdown.css';
 import { css } from '@patternfly/react-styles';
 
-const DropdownToggle = ({ children, IconComponent, ...props }) => (
+const DropdownToggle = ({ children, iconComponent: IconComponent, ...props }) => (
   <Toggle {...props}>
     {children}
     {IconComponent && <IconComponent className={css(styles.dropdownToggleIcon)} />}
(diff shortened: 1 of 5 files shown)


priley86:
looks good!

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
```diff
@@ -5,10 +5,10 @@ import Toggle from './Toggle';
 import styles from '@patternfly/patternfly-next/components/Dropdown/dropdown.css';
 import { css } from '@patternfly/react-styles';
 
-const DropdownToggle = ({ children, ...props }) => (
+const DropdownToggle = ({ children, noCaret, ...props }) => (
   <Toggle {...props}>
     {children}
-    <CaretDownIcon className={css(styles.dropdownToggleIcon)} />
+    {!noCaret && <CaretDownIcon className={css(styles.dropdownToggleIcon)} />}
```
dmiller9911: As an alternative API, what about adding an `IconComponent` prop that can be set to null and defaults to the `CaretDownIcon`? This would allow it to be replaced (maybe with a kebab icon) or removed by setting it to null.

jschuler: nice, I like that idea as well, going to give it a whirl

priley86: hmm interesting. I don't know if PF design will allow different caret icons there but I could see the case for this. This meets the immediate need I am having though (but could also be met by providing a null icon). I think `noCaret` is a bit more explicit right now. I am happy either way but am fine w/ the approach here for now.

jschuler:
@dmiller9911 Do you know if there is a convention for component props to be uppercase? Or should all props be lowercase and then when they are used renamed to its uppercase version to be used as a component? e.g.
```
const DropdownToggle = ({ children, iconComponent: IconComponent, ...props }) => (
  <Toggle {...props}>
    {children}
    {IconComponent && <IconComponent className={css(styles.dropdownToggleIcon)} />}
  </Toggle>
);
```

dmiller9911:
@jschuler that was a mistype on my part. We are using `component` for other areas we expect a component. We should follow that and make it `iconComponent` like you have in that example.

Issue #724, opened 2018-10-08
Update Modal and About Modal components to set 'overflow: hidden' on body when open


tlabaj:
Modal - body should be set to `overflow: hidden` when modal is open 
Depends on patternfly-next issue: [#793](https://github.com/patternfly/patternfly-next/issues/793)

tlabaj:
closed by #895

PR #895, opened 2018-11-07
fix(PF4 AboutModal and Modal): added a class to the body when the Mod…

tlabaj:
…al is open

#724

**What**: added a class to the body when the Modal is open to set overflow to hidden

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalContainer.js
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalContainer.js
@@ -9,6 +9,8 @@ import AboutModalBoxCloseButton from './AboutModalBoxCloseButton';
 import AboutModalBox from './AboutModalBox';
 import Backdrop from '../Backdrop/Backdrop';
 import Bullseye from '../../layouts/Bullseye/Bullseye';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/patternfly-next/components/Backdrop/backdrop.css';
 
 const propTypes = {
   /** content rendered inside the About Modal Box Content. */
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/components/AboutModal/AboutModalContainer.js
```diff
@@ -66,8 +68,10 @@ const ModalContent = ({
   ...props
 }) => {
   if (!isOpen) {
+    document.body.classList.remove(css(styles.backdropOpen));
```
dmiller9911: `document` and `window` do not exist in SSR (server side rendering) scenarios. this will need to be moved to `componentDidMount` and `componentDidUpdate`

dmiller9911: as another side note, it might be helpful to try this in page layout. I do not believe that the body is what scrolls there. I think it is the main section. Setting overflow here would actually not do anything in an app that uses our page layout component.

mattnolting:
@tlabaj Perfect! Thank you!

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalContainer.js
jschuler: @dmiller9911 Correct, in a page layout the section scrolls not the body, so this would not be an issue there

packages/patternfly-4/react-core/src/components/Modal/ModalContent.js
```diff
@@ -7,6 +7,8 @@ import ModalBox from './ModalBox';
 import ModalBoxFooter from './ModalBoxFooter';
 import Backdrop from '../Backdrop/Backdrop';
 import Bullseye from '../../layouts/Bullseye/Bullseye';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/patternfly-next/components/Backdrop/backdrop.css';
```
jschuler: These two imports should be removed

PR #896, opened 2018-11-07
translate strings to English in mockClientPaginationTable.js

tomer:
I see it pointless to have Czech strings in mockClientPaginationTable.js, so I've translated them back to English. I've used Google Translate, so I hope that my translations are correct. :smile:


jeff-phillips-18:
@tomer Looks like you need to update the snapshots. You can use `yarn run test -u` to do so.

Issue #892, opened 2018-11-07
PF4: Dropdown Typescript errors


dlabrecq:
I'm trying to use the PF4 dropdown, but there are a few typescript errors. Looking at the react-core code, it appears that:

1. The Dropdown onSelect is not defined
2. The Dropdown toggle type should probably be ReactNode
3. The DropdownToggle onToggle is not defined

These are only the properties that I attempted to use. There may be others that have also not been tested.

![screen shot 2018-11-07 at 9 43 47 am](https://user-images.githubusercontent.com/17481322/48138641-bbd13800-e272-11e8-8d19-2e70f24d795a.png)

edewit:
There is still an error, when I use this:
```
<DropdownItem
   onClick={(event) => {}}
   key={id}
   component="button"
>{name}</DropdownItem>
```
I get a compile error on `component`:
```
ERROR in [at-loader] ./src/pickers/example-picker/example-picker.tsx:72:25 
    TS2322: Type 'Element' is not assignable to type '"a" | "abbr" | "address" | "article" | "aside" | "b" | "base" | "bdi" | "bdo" | "big" | "blockquote" | "caption" | "cite" | "code" | "data" | "dd" | "del" | "details" | "dfn" | "dt" | ... 40 more ... | undefined'.
  Type 'Element' is not assignable to type 'FunctionComponent<DropdownItemProps>'.
    Type 'Element' provides no match for the signature '(props: DropdownItemProps & { children?: ReactNode; }, context?: any): ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)> | null'.
```
But from the docs a string can be use to define the component property

PR #898, opened 2018-11-07
Dropdown typescript errors

dlabrecq:
This addresses the various typescript errors (i.e., missing props, incorrect types, etc.) encountered while the Dropdown component in the Koku UI

Fixes https://github.com/patternfly/patternfly-react/issues/892

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
@@ -1,15 +1,17 @@
 import { SFC, HTMLProps, ReactType, ReactNode } from 'react';
 import { OneOf } from '../../typeUtils';
-import { DropdownToggleProps } from './DropdownToggle';
+import DropdownToggle from './DropdownToggle';
 import { DropdownPosition, DropdownDirection } from './dropdownConstants';
 
 export interface DropdownProps extends HTMLProps<HTMLDivElement> {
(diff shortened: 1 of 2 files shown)


priley86:
I was not seeing any errors for this before... but now I see this after updating (ts 2.9.2):
```
TS2430: Interface 'DropdownProps' incorrectly extends interface 'HTMLProps<HTMLDivElement>'.
  Types of property 'onSelect' are incompatible.
    Type 'Function' is not assignable to type '(event: SyntheticEvent<HTMLDivElement>) => void'.
      Type 'Function' provides no match for the signature '(event: SyntheticEvent<HTMLDivElement>): void'.
ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/Dropdown/Dropdown.d.ts(14,12):
TS2304: Cannot find name 'DropdownToggle'.
ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/layouts/index.d.ts(5,15):
Version: typescript 2.9.2
```

dmiller9911:
Yeah proper typing is `onSelect(event: React.MouseEvent<HTMLUListElement>): void` Function should usually not be used.

dmiller9911:
Might need to omit on select from the div typings since it is part of the base element handlers.

priley86:
i posted a fix in #900 @dmiller9911

PR #902, opened 2018-11-08
fix(charts): address typescript issues

dlabrecq:
Fixed various typescript issues encountered while using charts in Koku UI. For example, we needed to add a missing title to ChartLegned.

I also cleaned up a couple unused styles from the examples.

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/BarChart/examples/common/getContainerProps.js
+++ b/packages/patternfly-4/react-charts/src/components/BarChart/examples/common/getContainerProps.js
@@ -14,10 +14,6 @@ const styles = StyleSheet.create({
       '.chart-overflow': {
         ':not(foo) svg': {
           overflow: 'visible'
-        },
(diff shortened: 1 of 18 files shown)


packages/patternfly-4/react-charts/src/components/BarChart/examples/common/getContainerProps.js
```diff
@@ -14,10 +14,6 @@ const styles = StyleSheet.create({
       '.chart-overflow': {
         ':not(foo) svg': {
```
dmiller9911: Not needed now, but we can actually replace these with '& svg' the not food was a work around for adding child selectors with Aphrodite. React-styles is now using emotion which supports it out of the box.

dlabrecq: Thanks, I've made the change.

PR #909, opened 2018-11-13
refactor(SerialConsole,VncConsole): Unify use of Toolbar

mareklibra:
**What**: Partially unify code structure of related components.

Both components were created and reviewed independently leading to non-aligned code for the similar
visual effect. This patch unifies use of the Toolbar while having optional integration with
`AccessConsoles` component in mind.

Sass/Less of both components are refactored to use nesting.

No new feature is introduced.


mareklibra:
@jeff-phillips-18 , can you please have look?
It would be great to have this landed soon since bigger portions of code are being moved, so it blocks other work due to merge conflicts.

mareklibra:
Rebased to resolve merge conflicts.

PR #911, opened 2018-11-13
fix(LoginPage): small fixes.

Laviro:
While trying to implement the [LoginPage component in Foreman](https://github.com/theforeman/foreman/pull/6234)
I faced some problems which are fixed in this PR.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardForm.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardForm.js
@@ -16,15 +16,23 @@ const LoginCardForm = ({
   forgotPassword,
   rememberMe,
   submitError,
-  showError
+  showError,
+  attributes
 }) => (
-  <Form onSubmit={onSubmit} noValidate>
+  <Form onSubmit={onSubmit} noValidate {...attributes}>
     <LoginFormError show={showError}>{submitError}</LoginFormError>
     <LoginCardInput {...usernameField} />
     <LoginCardInput {...passwordField} />
     {additionalFields}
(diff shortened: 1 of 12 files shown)


packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardRememberMe.js
```diff
@@ -15,7 +16,7 @@ LoginCardRememberMe.propTypes = {
 };
 
 LoginCardRememberMe.defaultProps = {
-  label: 'Keep me logged in for 30 days',
+  label: '',
```
jeff-phillips-18: not sure about defaulting to text that is not handled by default. There is nothing here that will expire login after 30 days, correct?

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardForm.js
```diff
@@ -16,15 +16,23 @@ const LoginCardForm = ({
   forgotPassword,
   rememberMe,
   submitError,
-  showError
+  showError,
+  attributes
```
jeff-phillips-18: we usually just use `...props` and pass those along. no need to set a propType for this.

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardRememberMe.js
jeff-phillips-18: I think null is better so this doesn't show up inadvertently.

Laviro: done

PR #912, opened 2018-11-13
feat(TextInput): Add Focus TextInput example

boaz0:
I added an example of how to focus on `TextInput`.

Using `innerRef` props that was added in this change and `React.forwardRef` I could get the `input` component and focus on it.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/TextInput/examples/FocusTextInput.js
+++ b/packages/patternfly-4/react-core/src/components/TextInput/examples/FocusTextInput.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import { TextInput, Button } from '@patternfly/react-core';
+
+class FocusTextInput extends React.Component {
+  state = {
+    value: ''
+  };
+
+  inputRef = React.createRef();
+
+  componentDidUpdate() {
+    this.inputRef && this.inputRef.current.focus();
+  }
+
+  handleTextInputChange = value => {
+    this.setState({ value });
+  };
+
+  render() {
+    const { value } = this.state;
+    return (
+      <React.Fragment>
+        <TextInput
+          aria-label={'focus text input example'}
+          ref={this.inputRef}
+          value={value}
+          onChange={this.handleTextInputChange}
+        />
+        <Button
+          onClick={() => {
+            this.inputRef && this.inputRef.current.focus();
+          }}
+        >
+          Focus
+        </Button>
+      </React.Fragment>
+    );
+  }
+}
+
+export default FocusTextInput;
--- a/packages/patternfly-4/react-core/src/components/LoginPage/__snapshots__/LoginForm.test.js.snap
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/__snapshots__/LoginForm.test.js.snap
@@ -19,11 +19,7 @@ exports[`LoginForm with rememberMeLabel and no rememberMeAriaLabel generates con
     label="Username"
   >
     <TextInput
-      aria-label={null}
-      className=""
       id="pf-login-username-id"
-      isDisabled={false}
-      isReadOnly={false}
       isRequired={true}
       isValid={true}
       name="pf-login-username-id"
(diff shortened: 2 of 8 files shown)


packages/patternfly-4/react-core/src/components/TextInput/examples/FocusTextInput.js
```diff
@@ -0,0 +1,52 @@
...
+    this.setState({ value });
+  };
+
+  componentDidUpdate() {
+    this.inputRef && this.inputRef.focus();
+  }
+
+  render() {
+    const { value } = this.state;
+    const FocusTextInput = React.forwardRef(({ value, type, onChange }, ref) => (
```
dmiller9911: This is actually not how this is meant to be used. Forward ref should wrap the actual TextInput component that we export. That would allow for ref passed instead of innerRef

packages/patternfly-4/react-core/src/components/TextInput/examples/FocusTextInput.js
```diff
@@ -0,0 +1,52 @@
...
+        <button
+          onClick={() => {
+            console.log(this.inputRef);
+            this.inputRef && this.inputRef.focus();
+          }}
+        >
+          Focus
+        </button>
+        <FocusTextInput
+          ref={ref => (this.inputRef = ref)}
```
dmiller9911: Refs should use reacts new createRef API.

packages/patternfly-4/react-core/src/components/TextInput/examples/FocusTextInput.js
boaz1337: I am not sure I understand.

If I do 
```js
React.forwardRef((props, ref) => <TextInput ref={ref} {...props} />)
```
the `TextInput` will be referenced instead of the `input` node.

At any rate, I couldn't come up with any way to pass `ref` without using `innerRef`.

tlabaj:
@boaz1337 Can you please resolve conflicts in your branch.

packages/patternfly-4/react-core/src/components/TextInput/examples/FocusTextInput.js
```diff
@@ -0,0 +1,49 @@
...
+    this.setState({ value });
+  };
+
+  componentDidUpdate() {
+    this.inputRef && this.inputRef.current.focus();
+  }
+
+  render() {
+    const { value } = this.state;
+const FocusTextInput = React.forwardRef(({ value, type, onChange }, ref) => (
```
dmiller9911: Example of how forwardRef should be used.
```jsx

// src/components/TextInput/TextInput.js
class TextInput extends React.Component {
  render() {
   const { innerRef } = this.props;
   ...rest
  }
}

export default React.forwardRef((props, ref) => <TextInput innerRef={ref} {...props}  />);

// Example.js
import { TextInput } from '@patternfly/react-core';

class Example extends React.Component {
  inputRef = React.createRef();

  render() {
    return <TextInput ref={this.inputRef} />
  }
}
```

I left out much of the example and component implementation, but that should get the idea acress. React.forwardRef should not be called in render at the very least.

boaz1337: Got it, thanks. I will be updating in the next hours.

boaz1337: It looks OK but now I noticed that `TextInput` props are not shown.

It seems like `React.forwardRef` doesn't return a function like `class Component extends React.Component { ... }` and `const Component = (props) => { ... }` do.

It also looks like this [won't be fixed](https://github.com/facebook/react/issues/12453).
As a workaround I decided to use `react-is.isForwardRef` inside a function and render from the object the component passing `props` passed to the function.

Let me know what you think.

jgiardino:
I'm not able to review this. I get a "page not found" error when I try to navigate to the preview.

boaz1337:
@jgiardino yes me too. I guess because tests failed.

I will try to fix this soon as possible.

Thanks.

tlabaj:
@boaz1337 looks like this one is still failing the build.  Could you please take a look at that.  Thanks!

dgutride:
This PR has become stale - it isn't related to any issues we have open.  If you'd like to continue forward, please respond before March 18, 2019 and we'll keep it open.

dgutride:
Closing due to no response - please reopen once an issue has been created and accepted for this work to begin.

boaz0:
@dgutride - I thought you got a notification about rebasing the PR.
is it possible to open that ticket again :question:

Issue #908, opened 2018-11-12
PF4 Brand alt text


jgiardino:
The alt text in the brand component should be updated to match the example we have in core.

Currently, the react component uses the following alt text:
`alt="Brand Image" `
Whereas the example in core uses:
`alt="Patternfly Logo"`

Whenever text is included in an image, it should be repeated in the `alt` attribute value (this is how screen readers know what the text is). Additionally, alt text can describe the image (e.g. that it's a "logo").  

The Additional Notes section at the end of [this article](https://www.phase2technology.com/blog/no-more-excuses-definitive-guide-alt-text-field) provides a good checklist that's relevant for logos.

PR #917, opened 2018-11-14
fix(a11y): tweak html for better accessibility

seanforyou23:
This PR applies some minor html adjustments for improved a11y experience.

**What**: Update html for better accessibility

**Additional issues**: fix https://github.com/patternfly/patternfly-react/issues/908

@jgiardino I notice that the combination of placeholder text and aria-label gives an interesting experience on VO (for the component search adjustment) - mind seeing how this reacts in JAWS? If you get a same experience, I'm inclined to recommend we dump placeholder text for a text description outside of the input. I know we've discussed this before in a11y reviews, now we finally have a real use case! Let me know what you think.

update alt text to match pf core
correct html attribute name
give component search an accessible name
align alt text for brand images

### DIFF
--- a/packages/patternfly-4/react-docs/src/components/navigation/navigation.js
+++ b/packages/patternfly-4/react-docs/src/components/navigation/navigation.js
@@ -6,6 +6,11 @@ import PropTypes from 'prop-types';
 import logo from '../../assets/logo.png';
 import NavigationItemGroup from './navigationItemGroup';
 import NavigationItem from './navigationItem';
+import {
+  Form,
+  FormGroup,
+  TextInput
+} from '@patternfly/react-core';
 
 const routeShape = PropTypes.shape({
   to: PropTypes.string.isRequired,
@@ -33,8 +38,8 @@ class Navigation extends React.Component {
     searchValue: ''
   };
 
-  handleSearchChange = e => {
-    const searchValue = e.target.value;
+  handleSearchChange = (checked, event) => {
+    const searchValue = event.target.value;
     this.setState(() => ({
       searchValue
     }));
@@ -65,15 +70,20 @@ class Navigation extends React.Component {
               <img src={logo} alt="PatternFly Logo" />
             </Link>
           </div>
-          <div className={css(styles.search)}>
-            <input
-              className={css(styles.input)}
-              placeholder="Search components"
-              type="text"
-              value={searchValue}
-              onChange={this.handleSearchChange}
-            />
-          </div>
+          <Form className={css(styles.search)} onSubmit={event => { event.preventDefault(); return false; }}>
+            <FormGroup
+              label="Search Components"
+              fieldId="primaryComponentSearch">
+              <TextInput
+                type="text"
+                id="primaryComponentSearch"
+                name="primaryComponentSearch"
+                placeholder="For example, &quot;button&quot;"
+                value={searchValue}
+                onChange={this.handleSearchChange}
+              />
+            </FormGroup>
+          </Form>
           <NavigationItemGroup title="Style">
             <NavigationItem to="/styles/tokens" pkg="tokens">Tokens</NavigationItem>
             <NavigationItem to="/styles/icons" pkg="icons">Icons</NavigationItem>
--- a/packages/patternfly-4/react-core/src/components/AboutModal/examples/SimpleAboutModal.js
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/examples/SimpleAboutModal.js
@@ -29,7 +29,7 @@ class SimpleAboutModal extends React.Component {
           productName="Product Name"
           trademark="Trademark and copyright information here"
           brandImageSrc={brandImg}
-          brandImageAlt="Brand Image"
+          brandImageAlt="Patternfly Logo"
           logoImageSrc={logoImg}
           logoImageAlt="AboutModal Logo"
           heroImageSrc={heroImg}
--- a/packages/patternfly-4/react-core/src/components/Brand/examples/SimpleBrand.js
+++ b/packages/patternfly-4/react-core/src/components/Brand/examples/SimpleBrand.js
@@ -4,7 +4,7 @@ import brandImg from './pf_logo.svg';
 
 class SimpleBrand extends React.Component {
   render() {
-    return <Brand src={brandImg} alt="Brand Image" />;
+    return <Brand src={brandImg} alt="Patternfly Logo" />;
   }
 }
 
--- a/packages/patternfly-4/react-core/src/components/Form/examples/HorizontalForm.js
+++ b/packages/patternfly-4/react-core/src/components/Form/examples/HorizontalForm.js
@@ -55,7 +55,7 @@ class HorizontalForm extends React.Component {
     return (
       <Form isHorizontal>
         <FormGroup
-          label="required Name"
+          label="Required Name"
           isRequired
           fieldId="horizontal-form-name"
           helperText="Please provide your full name"
@@ -65,12 +65,12 @@ class HorizontalForm extends React.Component {
             isRequired
             type="text"
             id="horizontal-form-name"
-            describedby="horizontal-form-name-helper"
+            aria-describedby="horizontal-form-name-helper"
             name="horizontal-form-name"
             onChange={this.handleTextInputChange1}
           />
         </FormGroup>
-        <FormGroup label="Email" fieldId="horizontal-form-email">
+        <FormGroup label="Email" isRequired fieldId="horizontal-form-email">
           <TextInput
             value={value2}
             onChange={this.handleTextInputChange2}
--- a/packages/patternfly-4/react-core/src/components/Form/examples/InvalidForm.js
+++ b/packages/patternfly-4/react-core/src/components/Form/examples/InvalidForm.js
@@ -13,7 +13,7 @@ class InvalidForm extends React.Component {
           fieldId="email-label"
           isValid={false}
         >
-          <TextInput isValid={false} value="email||email.com" id="email-label" aria-errormesssage="email-helper" />
+          <TextInput isValid={false} value="email||email.com" id="email-label" aria-describedby="email-label-helper" />
         </FormGroup>
         <FormGroup
           label="Age"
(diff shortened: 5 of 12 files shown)


jgiardino:
Thanks for making these updates!

I need to run out for an appointment, but would like to review in JAWS later today.

I did see a mismatch between the values defined for `aria-errormessage` and `id` of the error message text for the email field.

Regarding placeholder, when I tested in JAWS, there was no indication that placeholder text is placeholder text. When looking at other examples, especially from the a11y thought leaders, I would see examples that defined placeholder text starting with "For example, ..." or "E.g. ...", which seems like a really good idea if we decide to keep placeholder text in our design system. The Forms design documentation is currently up in a PR on core now, and is on my list to review. I'll remember to check that one. 👍

jgiardino:
While looking at this, there are a couple of inputs with the `required` attribute that do not have the required * asterisk in the label:
- Simple Form > Phone number
- Horizontal Form > Email

Are these fixes within the scope of this PR? Or should I capture this in a separate issue?

jgiardino:
Related to the previous comment, I found one more attribute type:
- Horizontal form > required Name field - the attribute `describedby` should be `aria-describedby`	

Your updates look great! Let me know if these typos I captured should be separate issues.

To answer your question about the component search field, the following is what gets announced in JAWS and NVDA, depending on navigation method used. Ultimately I would judge the experience based on the Tab key navigation that's used during forms mode. The experience in NVDA seems nicer in that it identifies that the field is "blank". JAWS doesn't provide that feedback.

- NVDA component search
  - Using forms mode keyboard navigation (i.e. Tab key): 
“Component Search edit Find components, templates blank”
  - Using key command to go to next dom element that is a form element (i.e. F): 
“Component Search edit Find components, templates”
  - Using key command to go to next dom element: 
“edit Find components, templates”


- JAWS component search
  - Using forms mode keyboard navigation (i.e. Tab key): 
“Component Search edit Find components, templates dot dot dot type in text”
  - Using key comment to go to next dom element that is a form element (i.e. F): 
“Component Search edit Find components backslash templates backslash dot dot dot”
  - Using key command to go to next dom element: 
“Component Search”
  - Using key command to go to next dom element:  
“blank edit”

seanforyou23:
Awesome feedback @jgiardino! So, here's what I've done.

- Temporarily replace `aria-errormessage` with `aria-describedby` since it's interaction is more consistent across screen readers (VoiceOver+Safari wasn't reading the error message at all - now it does)
- Fixed those missing required asterisk and other attributes you mentioned
- Added a proper label to the component search and improved the placeholder text to not use ellipsis, it now reads more like a normal sentence

I'm sure that last point will draw some attention, and it probably needs a followup tweak on the styling, but generally it now stands out much better, is more obvious that you can actually interact with it, and solves both the accessible name and color contrast issue that it had before.

Hope this helps, let me know if you see any other issues!

jgiardino:
> Temporarily replace aria-errormessage with aria-describedby since it's interaction is more consistent across screen readers (VoiceOver+Safari wasn't reading the error message at all - now it does)

By testing just in JAWS, the fields with error messages have a much better experience. For example: "Object edit invalid entry [instructions about key commands for announcing descriptive text] Five, Age has to be a number, type in text"


> Fixed those missing required asterisk and other attributes you mentioned

I also noticed a couple of issues that I missed before:

- In the example above, JAWS says "Object" instead of "Age" because `aria-label="Object"` is defined on the input
- In the example Simple Form, the "required Name" input has a mismatch in values between `aria-describedby` and `id`

> Added a proper label to the component search and improved the placeholder text to not use ellipsis, it now reads more like a normal sentence

To use the `<label>` you created, you need to remove the `aria-label` attribute on the input. This will override what's defined in `<label>`

seanforyou23:
Good catches! Will get updates in place for those. I wonder if both of the descriptive attributes can be present on a same element, maybe you've already tried this? I assume errors will arise but I suppose it would be good to confirm.

If it's bad experience in JAWS vs no experience in VO, I'd go with the former. Updates on the way.

packages/patternfly-4/react-docs/src/components/navigation/navigation.js
```diff
@@ -58,10 +58,13 @@ class Navigation extends React.Component {
             </Link>
           </div>
           <div className={css(styles.search)}>
```
mcoker: I dunno if it will be div-itis to do this, but you could change this to

```css
<div class="pf-c-form">
    <div class="pf-c-form__group">
        <label class="pf-c-form__label">
            <span class="pf-c-title pf-m-lg"> /* or use a different modifier for a different size*/
        <input class="pf-c-form-control">
```

That would align with https://pf-next.com/components/form/examples/

packages/patternfly-4/react-docs/src/components/navigation/navigation.js
```diff
@@ -57,15 +57,23 @@ class Navigation extends React.Component {
...
-          <div className={css(styles.search)}>
-            <input
-              className={css(styles.input)}
-              placeholder="Find components, templates,..."
-              type="text"
-              value={searchValue}
-              onChange={this.handleSearchChange}
-            />
-          </div>
+          <form className={[css(styles.search), 'pf-c-form'].join(' ')} onSubmit={event => { event.preventDefault(); return false; }}>
```
dmiller9911: Could the actual form and TextInput components be used here instead? Would be nice to actually use the components. This was put in before they existed, and never converted over.

seanforyou23: Ah good call, I was wondering if there was a reason this was raw markup instead of React components. Learning a ton of history in this issue, thx for pointing this out @dmiller9911 - totally agree.

seanforyou23:
Some questions came up in making these last few tweaks:

1) Why is "checked" prop called "checked" and why is it noted as a boolean type? 
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/TextInput/TextInput.d.ts#L11

When I log it, I see the "checked" param seems to be the actual text value entered and of type `string`. Is it a bug? https://github.com/patternfly/patternfly-react/pull/917/files#diff-152239d6048e71a34f2804c09d7520b0R39

2) Why is part of our css "reset" injected half way through injected style blocks for the workspace? See screenshot below, seems like this should be super early in the cascade, it looks like it was added arbitrarily somewhere in the middle of a bunch of other style references. Surely I'm missing something... I ask because with this being the case, my styling override for the margin around the form isn't applied and the zero value that's used as part of this reset supersedes it. What's the "correct" way to do this?

screen shot 2018-11-26 at 2 31 07 pm

3) Is it just me or is something seriously wrong with the high-level cascade of styles in the workspace? See screenshot below, why are third party libs like fontawesome being injected _after_ core PF styles, which again are injected _after_ the custom component-specific styles. Isn't this like.. exactly backward from what it should be?

screen shot 2018-11-26 at 2 32 47 pm

4) I realize now that new `<style>` elements are only ever added (dynamically) to the workspace, and never removed (dynamically). Is this intentional? Again, considering the above screenshot, I don't see how the order of injected styles leads to any sort of obvious cascade so just unsure if I'm facing a workspace issue or I'm just "doing it wrong" in my approach for overriding the margin around the form.

I'm getting going with styles in PF4 for the first time so please excuse any silly questions as a result of an oversight :)

Maybe @dgutride @jschuler or @dmiller9911 has a clue on some of this?

seanforyou23:
One more question while I'm asking. See the following two screenshots. In the first one I used the following markup to style the label - it worked because I had fine grain control over the html.

`<span className="pf-c-title pf-m-lg">
  Search Components
</span>`

screen shot 2018-11-26 at 2 28 54 pm

screen shot 2018-11-26 at 2 29 22 pm

Now that I'm using the pf react form component, I have slightly less control over the exact markup and am unsure how to apply the styling adjustments to make the label bigger. Using css in js approach, I have no way to target the `<label>` element or anything inside because it's abstracted away as a simple text prop. Is there a way to provide a type of decendant selector using css in js? Or, is there a better and more preferred way of accomplishing this?

seanforyou23:
Apologies for the force pushes, I goofed up a merge earlier today and didn't realize it until now. Had to rebase, should be fixed now. Let me know if I need to open a new PR or anything.

Regarding the label styles, I was able to get some tips over on patternfly.slack.com and have [applied them](https://github.com/patternfly/patternfly-react/pull/917/files#diff-67260aa44fd01223c17742ce1e0362f6R26) - seems to work fine.

As for the ordering of the injected styles, I noticed using string template vs object literal as property values in the styles creation yielded different output positions in the DOM. Not sure if this is intentional or where I can read more about it (probably in the emotion docs) - should be fine to move along with for now anyhow.

mcoker:
Looks awesome @seanforyou23! Some comments:

If your OS shows scroll bars by default, when the scrollbar is visible in the default sidenav view, then disappears when you search for something and the list of things is shorter than the viewport height, the sidebar contents shift. One of the things that makes the shift worse is that the brand image scales to fit the parent width, so when the sidebar becomes wider (when the scrollbar disappears), the brand image becomes wider/taller, shifting the sidebar contents down.

Since the scrollbar is visible on the site by default, I think it would be fine to just show the scrollbar all of the time. You could modify https://github.com/patternfly/patternfly-react/blob/573f7661758a5c69ecb6acb5a7e549813c92c346/packages/patternfly-4/react-docs/src/components/navigation/navigation.styles.js#L17 to be `scroll` instead of `auto`

There are inline styles in `NavigationItemGroup.js` that could be in the `title` CSS already in that component. https://github.com/patternfly/patternfly-react/blob/573f7661758a5c69ecb6acb5a7e549813c92c346/packages/patternfly-4/react-docs/src/components/navigation/navigationItemGroup.js#L25

Would `<Label />` be more appropriate than `<Badge />` here? https://github.com/patternfly/patternfly-react/blob/573f7661758a5c69ecb6acb5a7e549813c92c346/packages/patternfly-4/react-docs/src/components/navigation/navigationItem.js#L28. I don't think it would look as good, but it's probably a better use of the component for what the text in the current badge represents.

seanforyou23:
Thx @mcoker - I applied updates for the issues you pointed out. Regarding Badge vs Label, it was a good point and thank you for bringing this up. I was able to get the best of both worlds by using the Badge as is so we get the intended styling, and then I associated the nav item text with the relevant patternfly package by way of `aria-describedby`, so screen readers will now hear for example

> About Modal... item _x_ of _y_... found in _package name_

This way the badge fulfills the same purpose as what the label would, and we don't have to add any styling overrides. WDYT?

seanforyou23:
Good catch @mcoker ! After discussing further offline, we realized both label and badge didn't really fit the bill semantically, so we didn't use either and simply styled the anchor to look how we wanted (it now looks like a badge). This could potentially be confusing from a UX perspective, but I think that's an issue for another day. Let me know if you see any problems!!

Issue #888, opened 2018-11-06
PF4 progress bar shows 10% as min value


dlabrecq:
The Koku UI currently uses the latest PF4 progress bar and in some cases, shows zero as the percentage value. However, the progress bar itself still shows zero as 10%.

Similar to how 100% shows a full bar, I expect zero to show an empty bar.

Zero percent
![progressbar zero percent](https://user-images.githubusercontent.com/17481322/48086099-034fb980-e1ca-11e8-9037-09b253213ed3.png)

10 percent
![progressbar 10 percent](https://user-images.githubusercontent.com/17481322/48086220-4c077280-e1ca-11e8-8ac3-57e1b9c2a381.png)

ibolton336:
c_progress__indicator_MinWidth is set to "2rem". Not sure where to change this but setting the min width to 0 should fix it.

ibolton336:
This is a PF-next issue. Closing & reopening in patternfly-next.

dlabrecq:
Won't the react-core package need to be updated in order to receive the fix from pf-next? At a min, the pf-next new issue could be added here, before closing, so we can track progress.

ibolton336:
[link to pf-next issue](https://github.com/patternfly/patternfly-next/issues/913)

PR #920, opened 2018-11-14
chore(pf-core): update core for PF4

ibolton336:
**What**: 
Updating the patternfly-next version as some changes were merged in to resolve https://github.com/patternfly/patternfly-react/issues/888

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/layouts/Page/__snapshots__/PageHeader.test.js.snap
+++ b/packages/patternfly-4/react-core/src/layouts/Page/__snapshots__/PageHeader.test.js.snap
@@ -19,6 +19,7 @@ exports[`Check page vertical layout example against snapshot 1`] = `
 .pf-l-page__header.pf-m-tall {
   display: flex;
   grid-area: header;
+  z-index: 300;
   flex-wrap: wrap;
   align-items: flex-end;
   max-width: 100%;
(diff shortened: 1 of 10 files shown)


rvsia:
There is already a newer version of patternfly-next, see https://www.npmjs.com/package/@patternfly/patternfly-next

packages/patternfly-4/react-core/src/layouts/Page/__snapshots__/PageHeader.test.js.snap
```diff
@@ -19,6 +19,7 @@ exports[`Check page vertical layout example against snapshot 1`] = `
 .pf-l-page__header.pf-m-tall {
   display: flex;
   grid-area: header;
+  z-index: 300;
```
dgutride: I'm really interested in why our snapshots need to reference direct css values especially since the react repo doesn't control these.

jschuler: I think it's useful to see, when we upgrade Core, what visually has changed for the components

dmiller9911: The main idea for this before was to catch things core may have missed since there is no unit testing in core. It has caught issues with variables being referenced that didn't actually exist before. This is why variables are replaced with raw values. This was all put in as a stop gap or alternative to visual regression tests.

PR #928, opened 2018-11-16
chore(pf4-docs): show components as secondary items and package

jschuler:


jschuler:
screen shot 2018-11-15 at 11 18 01 pm

jschuler:
Looks like there's an issue with the production build when built through CI. Prod build locally works fine. Continuing to investigate...

packages/patternfly-4/react-docs/package.json
```diff
@@ -37,8 +36,9 @@
     "gatsby"
   ],
   "scripts": {
-    "docbuild": "node build/copyDocs.js && gatsby build",
-    "pr-build": "node build/copyDocs.js && gatsby build --prefix-paths",
+    "patch": "sed -i -e 's/if(c\\.initial) return;/if(!c||c.initial) return;/g' ./node_modules/extract-text-webpack-plugin/index.js",
```
dmiller9911: What is this for?  Is this command available in Linux and windows? If not we might want to make this a node script instead. So that it works cross platform.

jschuler: One of the first travis builds was throwing an error
https://travis-ci.org/patternfly/patternfly-react/builds/455815100#L5627
Even though i didn't see this in a local prod build. On research I found this patch from https://github.com/gatsbyjs/gatsby/issues/1846#issuecomment-324744367. I'll try removing it again since I am not sure if it's still needed as I think it was masking another issue that I fixed afterwards (reserved switch keyword was used in .tmp/Switch.docs.js `    export const switch = docs;` before I changed it to switch_docs).

Unfortunately, the travis build at https://928-pr-patternfly-react-patternfly.surge.sh/patternfly-4/index.html is throwing console errors `webpackJsonp is not defined` and see several issues, for example broken search, gatsby link not properly working etc. Again, local prod build no issues. So was also investigating if switching the travis build environment over to OSX could help...

jschuler:
Docs look good in the "precise" travis environment. Going to revert the travis config just to see if the webpack/travis issue reoccurs or if it was a fluke

jschuler:
Well, looks like it was a fluke before, not sure why it happened. Seems to work now with previous travis config.

Issue #803, opened 2018-10-17
PF4: Radio required properties


dlabrecq:
I installed the latest @patternfly/react-core v1.25.1 package and encountered the following TypeError. It appears that the radio button ID, name, and aria-label are required properties?

![screen shot 2018-10-17 at 3 19 09 pm](https://user-images.githubusercontent.com/17481322/47111073-8a8bbc00-d220-11e8-970c-949668561069.png)

I can understand adding ID/name, but shouldn't aria-label (et al.) be optional?

boaz1337:
It looks like @rvsia patch https://github.com/patternfly/patternfly-react/commit/a62ee271b08b823b3631d38aa1f4579f768894ae#diff-5aebc3b1ceb9e12e2593314778796112R18 changed it to required.

dlabrecq:
Yes, aria-label is now required. Although, I don't believe it should be -- it's not required on all pf4 components. The components should make it easy for apps to implement a11y, but not force it.

Also, ID and name may be required for a radiobutton group, but not for a single radio button.

rvsia:
Hello everyone,

this depends on our vision for the PF4 React. Since the checking of children props has been removed (https://github.com/patternfly/patternfly-react/pull/823 and see discussion here https://github.com/patternfly/patternfly-react/issues/799) we can not alert consumers if children of formGroups do not have labels, ids, etc. So, the simplest solution is to have this props required. Or not to control it and let consumers manage it themselves, but as I said, that depends on what we want to do with a11y.

In terms of a single radio button, I can not imagine use cases for using single radio buttons. There are checkbox or switch components which suit more.

tlabaj:
closed by #930

PR #930, opened 2018-11-16
fix(Radio): Updated Radio aria-label prop so that it is optional

tlabaj:
#803

**What**: Updated Radio aria-label prop so that it is optional.  A warning is displayed in the console at dev time if the aria-label is missing.  I also made some accessibility fixes that were found by AXE .

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Radio/Radio.js
+++ b/packages/patternfly-4/react-core/src/components/Radio/Radio.js
@@ -19,7 +19,12 @@ const propTypes = {
   /** Id of the Radio. */
   id: PropTypes.string.isRequired,
   /** Aria-label of the Radio. */
-  'aria-label': PropTypes.any.isRequired,
+  'aria-label': props => {
+    if (!props['aria-label']) {
+      return new Error('Radio requires an aria-label to be specified');
+    }
+    return null;
+  },
   /** Name for group of Radios */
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/components/Radio/Radio.js
```diff
@@ -19,7 +19,12 @@ const propTypes = {
   /** Id of the Radio. */
   id: PropTypes.string.isRequired,
   /** Aria-label of the Radio. */
-  'aria-label': PropTypes.any.isRequired,
+  'aria-label': props => {
```
dmiller9911: will `PropTypes.string` work here?  It looks like this below is the same as having it required.

tlabaj: The real issue was with the d.ts file. Making the aria-label required there was causing an error.  In order to make the propTypes and the d.ts file match, I removed the isRequired.  We had decided that we want to warn the consumer if they are missing an Accessibility attribute and not cause an error.

jschuler: I think PropTypes.string.isRequired would serve the same purpose. I think we either need to require it for both TS and JS, or optional for both (no message). IMO aria labels should always be optional.

Issue #933, opened 2018-11-17
PF4 Dropdown typescript issues (regression)


dlabrecq:
I installed the latest react-core package and the typescript issues previously fixed have cropped back in again. In particular, with the toggle property.

I believe the change below caused the regression.
https://github.com/patternfly/patternfly-react/pull/900

Typescript errors:
![screen shot 2018-11-16 at 7 46 16 pm](https://user-images.githubusercontent.com/17481322/48654206-eccf0c80-e9d8-11e8-841e-5b87d037e118.png)

My Dropdown usage:
![screen shot 2018-11-16 at 7 47 03 pm](https://user-images.githubusercontent.com/17481322/48654226-11c37f80-e9d9-11e8-818e-ece0db34b79c.png)

dmiller9911:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts#L14 that should be ReactNode. Could make it potentially more strict with ReactElement<DropdownToggleProps>, but I don't see much value in that.

PR #935, opened 2018-11-17
Fix Dropdown typescript issue

dlabrecq:
This PR addresses a typescript regression encountered with the Dropdown. Retested this fix with the Koku UI.

Fixes https://github.com/patternfly/patternfly-react/issues/933


packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.d.ts
```diff
@@ -11,7 +11,7 @@ export interface DropdownProps extends HTMLProps<HTMLDivElement> {
   onSelect(event: React.SyntheticEvent<HTMLDivElement>): void;
```
dmiller9911: Not needed now, but might be nice to make dropdownItems `ReactNode` instead of `ReactNode[]` the array variation is covered in `ReactNode`

PR #936, opened 2018-11-18
fix(LoginPage): Improve inputs validation

Laviro:
As part of the [LoginPage implementation in Foreman](https://github.com/theforeman/foreman/pull/6234) there need to be some fixes in inputs validation.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
@@ -142,7 +142,17 @@ class LoginCardWithValidation extends React.Component {
     !this.isPasswordShort() &&
     this.isUserNameValid();
 
-  isPasswordShort = () => this.state.passwordField.value.length < this.props.passwordField.minLength;
+  isPasswordShort = () => {
+    const {
+      passwordField: { minLength: passwordMinLength }
+    } = this.props;
+    const {
+      passwordField: {
+        value: { length: currentPasswordLength }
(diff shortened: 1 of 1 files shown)


packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
```diff
@@ -122,7 +122,7 @@ class LoginCardWithValidation extends React.Component {
   handleOnInputErrors = () => {
     const { usernameField, passwordField } = this.state;
     if (usernameField.value) {
-      !this.isUserNameValid() && this.handleOnInvalidUsername();
+      !this.isEmailValid() && this.handleOnInvalidUsername();
```
ohadlevy: why? do you always assume that username is an email?

Laviro: Inside this function I validate only it's an email, else return true

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
```diff
@@ -204,7 +214,13 @@ class LoginCardWithValidation extends React.Component {
     });
   };
 
-  isUserNameValid = () => {
+  isEmailValid = () => {
```
jeff-phillips-18: I think the method should remain as `isUserNameValid`, if follow-up PRs add more validation cases there would be no need to re-rename this method.

PR #938, opened 2018-11-19
Fix sub-component references in react-console DesktopViewer

KKoukiou:
Follow up for a1e4d4f.

a1e4d4f fixed subcomponent references for react-console but there are some leftovers.


KKoukiou:
Thanks for quick handling.

PR #942, opened 2018-11-20
fix(LoginPage): improve form's error style and behavior

Laviro:
As part of the [LoginPage implementation in Foreman](https://github.com/theforeman/foreman/pull/6234) 
I found that there was no style to the form-error
and it appeared only after the submit was sent, 
without any control of when to show it.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
@@ -1,14 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Collapse from 'react-collapse';
+import { default as cx } from 'classnames';
+import { excludeKeys } from '../../../../common/helpers';
 
-const LoginFormError = ({ children, show, className, ...props }) => (
-  <div className={`login-form-error ${className}`}>
-    <Collapse {...props} isOpened={show}>
+const LoginFormError = ({ children, show, className, ...props }) =>
+  children && (
+    <Collapse {...excludeKeys(props, ['className'])} isOpened={show} className={cx('login-form-error', className)}>
       {children}
     </Collapse>
-  </div>
-);
+  );
 
 LoginFormError.propTypes = {
   className: PropTypes.string,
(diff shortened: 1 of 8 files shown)


packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
```diff
@@ -4,9 +4,11 @@ import Collapse from 'react-collapse';
...
   <div className={`login-form-error ${className}`}>
-    <Collapse {...props} isOpened={show}>
-      {children}
-    </Collapse>
+    {children ? (
+      <Collapse {...props} isOpened={show}>
+        {children}
+      </Collapse>
+    ) : null}
   </div>
```
jeff-phillips-18: Is it possible to see this in the storybook?

Laviro: Yes, it works there on the `managed login page`, see in the build above :)

jeff-phillips-18: Can we add the `login-form-error` class to `Collapse` and avoid the extra div?

Laviro: yes, great catch @jeff-phillips-18 :)

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
```diff
@@ -1,14 +1,14 @@
...
 import PropTypes from 'prop-types';
 import Collapse from 'react-collapse';
+import { default as cx } from 'classnames';
 
-const LoginFormError = ({ children, show, className, ...props }) => (
-  <div className={`login-form-error ${className}`}>
-    <Collapse {...props} isOpened={show}>
+const LoginFormError = ({ children, show, className, ...props }) =>
+  children && (
+    <Collapse {...props} isOpened={show} className={cx('login-form-error', className)}>
```
jeff-phillips-18: Should probably exclude `className` from the props just so it isn't passed twice if specified.

```suggestion
    <Collapse {...excludeKeys(props, ['className'])} isOpened={show} className={cx('login-form-error', className)}>
```

Issue #945, opened 2018-11-20
Calculate Progress bar width based on max value as well instead of just setting width to value


tlabaj:
By default, set min to 0, max to 100, and then calculate value as a percentage
So for the default of 0 and 100, value IS the percentage
but if max is 50 for example, then the value of 25 would be 50%

kmcfaul:
Closed with #950

PR #950, opened 2018-11-21
feat(Progress): use max parameter to scale value of progress bar

kmcfaul:
**What**: Change max parameter to control scaling of progress bar. Previously, it acted as a limit for the progress value.

Addresses issue: #945

Other notes:
- use floor of scaled value to prevent false 100% reports
- change internal variable name to better represent behavior
- update snapshot of Progress tests to reflect new behavior of max

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Progress/Progress.js
+++ b/packages/patternfly-4/react-core/src/components/Progress/Progress.js
@@ -71,9 +71,14 @@ class Progress extends Component {
       ...props,
       ...(valueText ? { 'aria-valuetext': valueText } : { 'aria-describedby': `${this.id}-description` })
     };
-    let limitedValue;
-    limitedValue = value < min ? min : value;
-    limitedValue = limitedValue > max ? max : limitedValue;
+    const scaledValue =
+      max === 100
+        ? value < min
+          ? min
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Progress/Progress.js
```diff
@@ -71,9 +71,14 @@ class Progress extends Component {
       ...props,
       ...(valueText ? { 'aria-valuetext': valueText } : { 'aria-describedby': `${this.id}-description` })
     };
-    let limitedValue;
-    limitedValue = value < min ? min : value;
-    limitedValue = limitedValue > max ? max : limitedValue;
+    const scaledValue =
+      max === 100
```
jschuler: can you put this in a function and add some test cases?
i tried out with value = 50, min = 50, max = 100 and got result 50 when I expected 0

kmcfaul: The min param still behaves as a minimum value for progress rather than a scaling factor as how max was changed. Should I also be changing how min works?

kmcfaul: Changing min parameter based on slack discussion.

PR #952, opened 2018-11-25
fix(LoginPage): Add submit error callback handling

Laviro:
As part of the [LoginPage implementation in Foreman](https://github.com/theforeman/foreman/pull/6234) 
Adding submit error callback to display the form error.
Added tests and storybook update.


Laviro:
Thanks @jeff-phillips-18, I am about to work on it right now

PR #972, opened 2018-11-27
chore(core): update core to .83

jschuler:
Also removed tall modifier from Page and related props/code as a result of Core changes that removed the modifier


jschuler:
@tlabaj recent core changes removed tall modifier from page. Had to make updates in react for that

PR #974, opened 2018-11-28
[WIP] fix(Page): Remove `pf-m-expanded` and `pf-m-collapsed` modifiers from sidebar on resize

TheRealJon:
These modifiers override the default responsive CSS styles. Changed the behavior so that the modifiers are only added on toggle, and are removed on resize.

### DIFF
--- a/packages/patternfly-4/react-core/src/layouts/Page/Page.js
+++ b/packages/patternfly-4/react-core/src/layouts/Page/Page.js
@@ -1,9 +1,8 @@
 import React from 'react';
 import styles from '@patternfly/patternfly-next/layouts/Page/page.css';
 import { css } from '@patternfly/react-styles';
-import { global_breakpoint_md as breakpointMd } from '@patternfly/react-tokens';
 import PropTypes from 'prop-types';
-import { debounce } from '../../internal/util';
+import { debounce, isDesktop } from '../../internal/util';
 
 export const PageLayouts = {
   vertical: 'vertical',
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -72,16 +73,19 @@ class Page extends React.Component {
...
   }
 
   // only enable tall header if the user is using the `useCondensed` header feature and we are on desktop
-  isTall = props =>
-    props.useCondensed && typeof window !== 'undefined' && window.innerWidth >= parseInt(breakpointMd.value, 10);
+  isTall = props => props.useCondensed && isDesktop();
 
   handleResize = () => {
-    this.setState({ isTall: this.isTall(this.props) });
+    console.log("HERE");
```
TheRealJon: Oops. 😆

packages/patternfly-4/react-core/src/layouts/Page/Page.js
```diff
@@ -72,16 +73,18 @@ class Page extends React.Component {
       } else {
         this.mainRef.current.removeEventListener('scroll', this.handleScroll);
       }
-      window.removeEventListener('resize', this.handleResize);
     }
+    window.removeEventListener('resize', this.handleResize);
```
dmiller9911: You might check out  match media and addListener it will allow using the same query as core and is less exepensive than resize. https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

jschuler:
@TheRealJon I'm closing this PR since we have already merged changes to the nav that removes collapsed/expanded functionality from the nav

PR #977, opened 2018-11-28
fix(CatalogTileView): Support tiles without categories.

jeff-phillips-18:
Currently if tiles are added to the tile view and are not in a category they are displayed in a single tile per row rather than wrapped as they are when they are in a category. 

This PR allows the caller to specify that categories are not used and will wrap the child tiles.

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/src/components/CatalogTileView/__mocks__/mockCatalogTileViewNoCategoriesExample.js
+++ b/packages/patternfly-3/patternfly-react-extensions/src/components/CatalogTileView/__mocks__/mockCatalogTileViewNoCategoriesExample.js
@@ -1,0 +1,101 @@
+import React from 'react';
+import { Icon } from 'patternfly-react';
+import { mockItems } from './mockItems';
+
+import { CatalogTile } from '../../CatalogTile';
+import { CatalogTileView } from '../index';
+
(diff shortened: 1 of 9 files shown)


packages/patternfly-3/patternfly-react-extensions/src/components/CatalogTileView/__mocks__/mockCatalogTileViewNoCategoriesExample.js
```diff
@@ -0,0 +1,101 @@
...
+      badges.push(
+        <CatalogTile.Badge key="certified" id="certified" title="Certified">
+          <Icon type="fa" name="cog" />
+        </CatalogTile.Badge>
+      );
+    }
+
+    if (item.approved) {
+      badges.push(
+        <CatalogTile.Badge key="certified" title="USDA Approved" id="approved">
```
mfrances17: think this should be key="approved"?

mfrances17: nit but i would order these the same as the certified badge - key/id/title
also is 'USDA' intended? would've thought simply Approved would be better

jeff-phillips-18:
Went for the extra-credit. @dtaylor113 @mfrances17 please take another look.

PR #979, opened 2018-11-29
fix(AccessConsoles): Fix child component recognition in production build

mareklibra:
In production build, names/strings are minified causing mismatch of `child.type.name`.
In this fix, the component type matching is fixed by using Ract component's displayName.

**What**: Fix AccessConsoles for child component recognition


mareklibra:
@jeff-phillips-18 , good idea - done.

PR #983, opened 2018-11-29
fix(packageJson) : Updated package.json to fix prod issue when import…

### LINKED ISSUE #982: react-core package.json side-effects: false issue
This issue arises when trying to import the base.css in a production environment with the "import" es6 syntax: 

The mini-css-extract-plugin( which is used with Webpack 4 to bundle a static css asset for production use) will overlook the import when looking for .css files if you arent explicitly using anything from the .css file by hand. See issue here 
https://github.com/webpack-contrib/mini-css-extract-plugin/issues/102#issuecomment-384947142

The side-effects flag in react-core's package.json is set to false so webpack attempts to optimize imports and skips over the import since it thinks you aren't using it. Instead the react-core package.json should use sideEffects: [ "*.css" ].

ibolton336:
…ing base.css (#982)

**What**:

FIxes #982 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/package.json
+++ b/packages/patternfly-4/react-core/package.json
@@ -6,7 +6,7 @@
   "module": "dist/esm/index.js",
   "types": "dist/js/index.d.ts",
   "patternfly:src": "src/",
-  "sideEffects": false,
+  "sideEffects": ["*.css"],
   "publishConfig": {
     "access": "public"
   },


packages/patternfly-4/react-core/package.json
```diff
@@ -6,7 +6,7 @@
   "module": "dist/esm/index.js",
   "types": "dist/js/index.d.ts",
   "patternfly:src": "src/",
-  "sideEffects": false,
+  "sideEffects": ["*.css"],
```
dlabaj: Is this something that should be done in core or something that should be done by the applications webpack?

ibolton336: After a good bit of research over the past few days, I have determined that this should be done at the patternfly package level. Since the react-core package is NOT entirely pure ( the .css files have side effects via adding styles to the dom), the sideEffect property in the core package.json should reflect that. This way when a user just wants to import the base.css from core, they don't have to worry about the file being dropped unintentionally in production mode. 

See https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

ibolton336:
Steps to reproduce: 

1) import patternfly base.css & any patternfly component

`import { Alert, Button } from '@patternfly/react-core';
import '@patternfly/react-core/dist/styles/base.css';`


2) build a production bundle using webpack 4 & the mini-css-extract-plugin 
`
const MiniCssExtractPlugin = require('mini-css-extract-plugin');

module.exports = {
  mode: 'production',
  plugins: [
    new MiniCssExtractPlugin({
      filename: '[name].css',
      chunkFilename: '[id].css'
    })
  ],
  module: {
    rules: [
      {
        test: /\.css$/,
        use: [MiniCssExtractPlugin.loader, 'css-loader']
      }
    ]
  }
};
`
3) Notice import is ignored and base.css styles are not added to generated static css asset.

packages/patternfly-4/react-core/package.json
ibolton336: @dlabaj Looks like Create-react-app handles this on the consumer application side. Here is a snippet from their production config:

            // Don't consider CSS imports dead code even if the
            // containing package claims to have no side effects.
            // Remove this when webpack adds a warning or an error for this.
            // See https://github.com/webpack/webpack/issues/6571
            sideEffects: true,

@seanforyou23 Maybe we should just handle it on the seed app side as a part of the end users webpack configuration?

dmiller9911:
Read that issue linked in the comment from create react app. That is rediculously annoying. Probably fine to have this in there. Will just want to make sure that the components tree shake properly. Sorry for the original push back on this. Seems like a good addition to work around webpacks weirdness.

packages/patternfly-4/react-core/package.json
dlabaj: Thanks @ibolton336 I didn't realize about create-react-app.

Issue #987, opened 2018-11-29
Typos in LoginPage component


mcoker:
* attribute in login form is `mame` instead of `name`
* docs for `passwordHelperText` should refer to `Password`, not `Username`

![image](https://user-images.githubusercontent.com/35148959/49256513-d5d6e400-f3f4-11e8-8707-88d5f678ccd6.png)

![image 1](https://user-images.githubusercontent.com/35148959/49256493-d2435d00-f3f4-11e8-82c6-d247c5a285ad.png)

PR #988, opened 2018-11-29
fix(LoginPage): typos

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/987

**What**:
* attribute in login form is `mame` instead of `name`
* docs for `passwordHelperText` should refer to `Password`, not `Username`


mcoker:
@tlabaj thanks, updated.

PR #989, opened 2018-11-30
Add desktop viewer to access consoles

mareklibra:
**What**: Add desktop viewer to access consoles

With this patch, the DesktopViewer can be a child component of the AccessConsoles.


mareklibra:
Requires #979

mareklibra:
Cc @jeff-phillips-18

PR #991, opened 2018-11-30
fix(DropdownItem): wrap DropdownItem in div when child is a Link node

kmcfaul:
Prevents nesting anchor tags.

**What**: If child element of DropdownItem is a react-router Link, wrapper should be a div.

Still need to add to documentation, potentially an example as well.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -30,10 +30,7 @@ const defaultProps = {
 
 const DropdownItem = ({ className, children, isHovered, component: Component, isDisabled, ...props }) => {
   const additionalProps = props;
-  if (children && children.type && children.type.name === 'Link') {
-    Component = 'div';
-    additionalProps.href = null;
-  } else if (Component === 'a') {
+  if (Component === 'a') {
     additionalProps['aria-disabled'] = isDisabled;
     additionalProps.tabIndex = isDisabled ? -1 : additionalProps.tabIndex;
   } else if (Component === 'button') {
@@ -42,12 +39,16 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
 
   return (
     <li>
-      <Component
-        {...additionalProps}
-        className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
-      >
-        {children}
-      </Component>
+      {React.isValidElement(children) ? (
+        React.Children.map(children, child => React.cloneElement(child, { className }))
+      ) : (
+        <Component
+          {...additionalProps}
+          className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
+        >
+          {children}
+        </Component>
+      )}
     </li>
   );
 };
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -40,7 +40,9 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
   return (
     <li>
       {React.isValidElement(children) ? (
-        children
+        React.Children.map(children, child =>
+          React.cloneElement(child, { className: `${className} ${child.props.className}` })
+        )
       ) : (
         <Component
           {...additionalProps}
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -30,12 +30,16 @@ const defaultProps = {
...
+  } else if (Component === 'a') {
     additionalProps['aria-disabled'] = isDisabled;
     additionalProps.tabIndex = isDisabled ? -1 : additionalProps.tabIndex;
   } else if (Component === 'button') {
     additionalProps.disabled = isDisabled;
   }
+
   return (
     <li>
       <Component
```
jschuler: If there was another routing component with a different name then we'd have to modify this code again.
Also, would you need that extra div for react router link? I think we could change this to something like
```
// Check if we have a component, like react router's Link.
{React.isValidElement(children) ? (
  <li>{children}</li>
) : (
  <li><Component {...props} className=...>{children}</Component></li>
)}
```

kmcfaul: Good call here adding the isValidElement check.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -36,14 +36,19 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
...
   return (
     <li>
-      <Component
-        {...additionalProps}
-        className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
-      >
-        {children}
-      </Component>
+      {React.isValidElement(children) ? (
+        React.Children.map(children, child => React.cloneElement(child, { className }))
```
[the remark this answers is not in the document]
kmcfaul: I added it when I was testing locally to pass the dropdown item styling through to make it consistent with other items, but I agree that it's better to let the user control how the Link looks directly. That way they can style their own lists of custom Links without the default styling stepping on them.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -36,14 +36,19 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
...
   return (
     <li>
-      <Component
-        {...additionalProps}
-        className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
-      >
-        {children}
-      </Component>
+      {React.isValidElement(children) ? (
+        children
```
tlabaj: I think we want to clone the children her an apply the styles.  You can see an example in the NavItem component.

kmcfaul: I had this implemented initially (see Joachim's comment thread). Would it be better to enforce the Dropdown styling or let the user define their own styles because these are custom elements that are embedded into the DropdownItem?

jschuler: Ah, i didn't see any patternfly classes copied on the children, just className. If there are patternfly styles then we can clone and apply them

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -36,14 +36,21 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
...
     <li>
-      <Component
-        {...additionalProps}
-        className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
-      >
-        {children}
-      </Component>
+      {React.isValidElement(children) ? (
+        React.Children.map(children, child =>
+          React.cloneElement(child, { className: `${className} ${child.props.className}` })
```
jschuler: So that was my initial cause for saying that if just className is copied over then we might as well skip the cloning. I guess for this to provide our PF styling here, I would expect to see something like this here:
```
React.cloneElement(child, { 
          className: css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className) 
        }
```

kmcfaul: Ah I forgot about the disabled and hovered styles, I can add those. It's a bit confusing, but className is actually the PF class being applied, child.props.className is whatever was manually declared as a className in the Link itself. Or should I not keep the local class?

jschuler: Yes that sounds good to add the local class as well!

Issue #973, opened 2018-11-28
Update Login Page based on core Updates


tlabaj:
Update Login Page to remove custom styling.  Core PR [#964](https://github.com/patternfly/patternfly-next/pull/964/), created Login Page components from the demos.  React can now pull in the component stylesheet rather than creating their own or pulling in large  utility style sheets.

tlabaj:
closed by #992

Issue #867, opened 2018-10-31
PF4 Data List 


LHinson:
Please note, this is issue is for Data List only (even though documents below also reference Table). 

[Requirements documentation](https://docs.google.com/document/d/1gyD2_Boy3gOHNl5KVo86274HTT_aqZMQz3coviJZIt8/edit)

Also see [Lists and Tables documentation](https://docs.google.com/document/d/1jYKhaFKvMKH1h2VJEDS4dtReMUmepDp99C4x4UNHNZk/edit). 

PF-Core example: https://pf-next.com/components/DataList/examples/

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @mattnolting (CSS) @mcarrano (design)

dlabaj:
Took a look at List View below.  It's made up of a large amount of children components and the api is pretty limited.   Migration shouldn't be an issue for consumers of this component from a list view.

http://patternfly-react.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react%2FContent%20Views%2FList%20View&selectedStory=List%20of%20expandable%20items&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

Issue #986, opened 2018-11-29
Login Page interaction issues


mcarrano:
There were two issues identified while doing an IxD review of the Login Page.

(1) The Language Selector should behave as a select list, i.e. selecting Spanish should change the label to Spanish. **This depends on issue #1214** 

(2) Input fields should not display an error state (red border) prior to validation. Default state should be displayed when this page is loaded and the user has not yet interacted with the page.  **This is fixed by core in PR** [*#1227*](https://github.com/patternfly/patternfly-next/pull/1227)

tlabaj:
@mcarrano   number (1) can not be addressed until issue #718 is resolve.  Dropdown select is not implemented in React yet.
Can you please give more info in number (2).  I was unable to reproduce it..

rachael-phillips:
We'll need the P1 label for this one for Beta

tlabaj:
@rachael-phillips The second point is addressed by core.  I updated description above.  The First point is dependent on the custom select (single select) component.

rachael-phillips:
@tlabaj thanks for the comment!

LHinson:
the language selector on the login page PR was closed without being merged to wait for designs: https://github.com/patternfly/patternfly-next/pull/1610 Therefore I've removed the rc.1 tag

mcarrano:
I think that the Language Selector issue is still pending, but not sure where it is on the roadmap. There is a core issue here: https://github.com/patternfly/patternfly-next/issues/349 @rachael-phillips can you help sort this out? I think this issue remains open unless it is now covered somewhere else.

jenny-s51:
Hi @mcarrano, it looks like the language selector was removed from the login page in this release: https://github.com/patternfly/patternfly-react/compare/@patternfly/react-core@2.1.1...@patternfly/react-core@2.1.2

Can this issue be closed?

mcarrano:
I think this issue can be closed @jenny-s51 . There is an open issue in core here to add the language selector in the background area of the page: https://github.com/patternfly/patternfly-next/issues/1821. It may make sense to close this and open a new React issue when the time is right as this issue was related to the old designs.

@rachael-phillips what do you think?

jenny-s51:
closing per @rachael-phillips

Issue #871, opened 2018-11-01
PF4: Social Icons Component - LoginPage Footer


amarie401:
Introduce the login footer containing the `Or log in with` and social icons component for the LoginPage. This will need to be integrated with the LoginPage component (#604)  and the example will need to be updated. 

**FYI**: There is a `More...` option in core that will show a different view of the social icons.
See: https://github.com/patternfly/patternfly-next/issues/801

rachael-phillips:
@dlabaj @jschuler to investigate for react implementation

tlabaj:
Closed by PR #1319

PR #992, opened 2018-11-30
refactor(LoginPage): Refactored LoginPage to match latest core version

tlabaj:
#973

**What**:Refactored LoginPage to match latest core version.
refactored based on changes in Core PR [#964](https://github.com/patternfly/patternfly-next/pull/964/)
Also updated snapshots to match styling changes Form core PRs [#1019](https://github.com/patternfly/patternfly-next/pull/1019/files) and [#1024](https://github.com/patternfly/patternfly-next/commits/master/src/patternfly/layouts/Page/page.scss) and [#867](https://github.com/patternfly/patternfly-next/pull/867)

I also renamed some of the props to be more intuitive.

**Additional issues**:


mcarrano:
@tlabaj Last week I filed #986 to address issues I saw in doing an IxD review of this. Is it possible to also address these as part of this PR?

tlabaj:
@mcarrano In the interest of unblocking folks who are waiting for the other changes on React, I think it is best to address issue #986  separately.  The first topic in the issue can not be address right away.  we do not have an implementation of dropdown select in react yet.

mcoker:
hey @tlabaj, looks great! Just a few comments.

* in `.pf-c-login__header`, you have `div.pf-c-content > p` - is there any way just to use a `p` there?
* is there support for `.pf-c-login__main-footer` (the "or log in with" links)
* the links in the footer (`.pf-c-login__footer > .pf-c-list > li > a`) have a class `.pf-c-login__footer-link`, can you add that? the links should be white.

tlabaj:
@mcoker Thank you for the review.  I made the changes for the first and last bullet point . The "Login with links" have not been implemented yet.  Issue #871 will cover that work.

packages/patternfly-4/react-core/src/components/LoginPage/LoginFooterItem.js
```diff
@@ -0,0 +1,30 @@
...
+};
+
+const defaultProps = {
+  children: null,
+  className: '',
+  href: '#'
+};
+
+const LoginFooterItem = ({ className, children, ...props }) => (
+  <a className={css(styles.loginFooterLink, className)} {...props}>
```
jschuler: href is not explicitly added here. also might need target="_blank" to open in new tab.
I also think that we should allow user to pass in their own Link component (e.g. react router). Maybe handle similar to here?
https://github.com/patternfly/patternfly-react/pull/991/files#diff-bd11ed69b68db9dcaeb317b19e68e785R42

packages/patternfly-4/react-core/src/components/LoginPage/LoginMainHeader.js
```diff
@@ -0,0 +1,40 @@
...
+  children: null,
+  className: '',
+  dropdown: null,
+  title: '',
+  subtitle: ''
+};
+
+const LoginMainHeader = ({ children, className, dropdown, title, subtitle, ...props }) => (
+  <header className={css(styles.loginMainHeader, className)} {...props}>
+    <Title size="3xl">{title}</Title>
```
jschuler: For optional elements, wonder if we need to surround with conditional, in case it takes up space in the DOM if unused, e.g. `{title && <Title...`

Issue #966, opened 2018-11-27
Badge Component Audit


christiemolloy:
Discrepancies between Core and React:

Badges in React have a margin applied to them. The component shouldn't have margins, its parent should handle its layout, so I'm curious as to what this was used for?
Can we remove this margin from the badge:

`.css-1v6g7pq > span {
    margin: 0.5rem;
}`

kmcfaul:
This span is added to the live preview wrapper and is not a part of the Badge component. Should I still remove this padding to better match the core?

PR #1001, opened 2018-12-03
fix(Badge): remove margin between badges to match core

kmcfaul:
**What**: update badge examples to match core.

References #966

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Badge/Badge.docs.js
+++ b/packages/patternfly-4/react-core/src/components/Badge/Badge.docs.js
@@ -10,13 +10,13 @@ export default {
   },
   examples: [
     {
-      component: Unread,
-      title: 'Unread Badge',
+      component: Read,
+      title: 'Read Badge',
       getContainerProps
     },
     {
-      component: Read,
-      title: 'Read Badge',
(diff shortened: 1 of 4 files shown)


christiemolloy:
@kmcfaul so the spaces in Core are determined by `display: inline-block`, which adds a literal space to either side of the badges. Since React is minifying the HTML this apparently removes those spaces and thats why those badges are now pressed together. Are you able to investigate how to bring those back?

kmcfaul:
Fixed the white space issue but the code in the examples is a bit less readable for it. I also swapped the order of the examples after noticing they were backwards compared to core.

jschuler:
How about instead of inlining the spans, add `{' '}` after the Badges
e.g.
```
<Badge isRead>7</Badge>{' '}
<Badge isRead>24</Badge>{' '}
<Badge isRead>240</Badge>{' '}
<Badge isRead>999+</Badge>
```
This way it is more readable and it will be obvious that we;ve inserted spaces

kmcfaul:
I can do that but when I save the file it auto-reformats into the inline version. Is there a way to disable that on VSCode for the file?

PR #1011, opened 2018-12-04
feat(PF4-Modal): Adds keyboard and screen reader focus trapping

### LINKED ISSUE #561: Modal keyboard interaction
The modal component should follow wai-aria authoring practices recommendations for keyboard interaction, specifically with regard to initial focus and keyboard support. 

**Several resources related to react and modals were included in #815**. 

Guidelines are available in the [wai-aria authoring practices for Dialog](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) as well as in the wai-aria [Modal Dialog Example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog).

The [keyboard interaction](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-6) they describe seems pretty straightforward. 

The guidelines for [initial focus](https://www.w3.org/TR/wai-aria-practices-1.1/#h-note-7) get a little hairy depending on the size and contents of the modal. They go into great detail on this in the [Modal Dialog Example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html), under Accessibility Features, list item 2. Basically, the idea is to avoid having focus placed on an element that is off the screen which would force the first contents to scroll out of view. I’m not sure that there’s a smart way to handle this, so the best alternative might be to let the consumer identify which element will receive focus. 

Any feedback and suggestions for this interaction are appreciated.

michaelkro:
BREAKING CHANGE: The reactRoot prop is required

Fixes #561

**What**:  Adds keyboard and screen reader focus trapping

**Notes/Resources**:

1.  [react accessible modal (github.com/reactjs)](https://github.com/reactjs/react-modal)
2.  [a11y dialog](https://github.com/edenspiekermann/a11y-dialog)
3.  The consumer must pass the react root `HTMLElement` as a prop.
4.  When the modal is opened, `aria-hidden=true` is applied to the react root element, hiding it from screen readers.
5.  When the modal is closed, the `aria-hidden` attribute is removed from the react root element

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.js
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import ModalContent from './ModalContent';
+import safeHTMLElement from '../../internal/safeHTMLElement';
 import { canUseDOM } from 'exenv';
 import { KEY_CODES } from '../../internal/constants';
 import { css } from '@patternfly/react-styles';
@@ -23,7 +24,9 @@ const propTypes = {
   /** A callback for when the close button is clicked */
   onClose: PropTypes.func,
   /** Creates a large version of the Modal */
-  isLarge: PropTypes.bool
+  isLarge: PropTypes.bool,
+  /** React application root element */
+  reactRoot: PropTypes.instanceOf(safeHTMLElement).isRequired
 };
 
 const defaultProps = {
@@ -62,8 +65,10 @@ class Modal extends React.Component {
   componentDidUpdate() {
     if (this.props.isOpen) {
       document.body.classList.add(css(styles.backdropOpen));
+      this.props.reactRoot.setAttribute('aria-hidden', true);
     } else {
       document.body.classList.remove(css(styles.backdropOpen));
+      this.props.reactRoot.removeAttribute('aria-hidden');
     }
   }
 
@@ -73,6 +78,8 @@ class Modal extends React.Component {
   }
 
   render() {
+    const { reactRoot, ...props } = this.props;
+
     if (!canUseDOM) {
       return null;
     }
@@ -81,7 +88,7 @@ class Modal extends React.Component {
       this.container = document.createElement('div');
     }
 
-    return ReactDOM.createPortal(<ModalContent {...this.props} id={this.id} />, this.container);
+    return ReactDOM.createPortal(<ModalContent {...props} id={this.id} />, this.container);
   }
 }
 
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.test.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.test.js
@@ -14,7 +14,8 @@ const props = {
   title: 'Modal',
   onClose: jest.fn(),
   isOpen: false,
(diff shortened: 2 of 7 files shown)


michaelkro:
Having trouble writing tests that check that the `aria-hidden` attribute is applied/removed from the React root.

Using `mount`, it seems that `componentDidUpdate` fires at some unknown time.  Taking a snapshot of `document.body`, I see something like the following

```js
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`the React root is hidden from screen readers when the modal is open 1`] = `
<body
  class="pf-c-backdrop__open"
>
  <div />
  <div />
  <div />
  <div />
  <div />
  <div />
</body>
`;
```
> On a side note, I'm not sure why there are so many divs here

Notice that the `pf-c-backdrop__open` class is being applied.  I can comment out [the line that applies that style](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Modal/Modal.js#L56), and when I rerun the tests (with the mounted Modal being open), the snapshot does not error, i.e., the class is still somehow being applied.

Attempts to do something like

```js
  const view = mount(<Modal {...props} isOpen />);
  const spy = jest.spyOn(document.body.classList, 'add');

  view.update();

  expect(spy).toHaveBeenCalled();
```

also result in failure.  I was hoping to do the something like

```js
test('the React root is hidden from screen readers when the modal is open', () => {
  document.body.appendChild(props.reactRoot);
  const view = mount(<Modal {...props} isOpen />);
  const spy = jest.spyOn(view.prop('reactRoot'), 'setAttribute');

  view.update();

  expect(spy).toHaveBeenCalledWith(stuff);
});
```

Any ideas/suggestions here?

jgiardino:
Thanks for working on this! Everything looks good wrt screen readers. I tested in JAWS, NVDA, and VO, and all trap focus, and contents outside the modal are not accessible while the modal is visible.

packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -23,7 +24,9 @@ const propTypes = {
   /** A callback for when the close button is clicked */
   onClose: PropTypes.func,
   /** Creates a large version of the Modal */
-  isLarge: PropTypes.bool
+  isLarge: PropTypes.bool,
+  /** React application root element */
+  reactRoot: PropTypes.instanceOf(safeHTMLElement).isRequired
```
jschuler: could we default to document.body and make it optional?

jgiardino: I had tested several implementations before to see what worked across the different screen reader / browser combinations we test in. The attributes `role="dialog"` and `aria-modal` don't seem to be supported as described by wai-aria. When I test their solution using JAWS / Chrome, focus is not trapped. It also is not trapped using [focus-trap-react](https://github.com/davidtheclark/focus-trap-react). 

Solutions that did work were:

- https://github.com/edenspiekermann/a11y-dialog 
- https://github.com/gdkraus/accessible-modal-dialog 

I did not test atlaskit, but will do that and let you know what the results were.

jgiardino: I captured part of this discussion in a separate issue: https://github.com/patternfly/patternfly-react/issues/1037

PR #1013, opened 2018-12-05
docs(LoadingState): Add state loading with delay to storybook

boaz0:
Currently there is only one story in LoadingState which demonstrates its props.
However, it's missing an example of setting a delay and waiting for data to be sent by the server.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoadingState/LoadingState.stories.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoadingState/LoadingState.stories.js
@@ -1,35 +1,59 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { withKnobs, boolean, select } from '@storybook/addon-knobs';
+import { withKnobs, boolean, select, number } from '@storybook/addon-knobs';
 import { withInfo } from '@storybook/addon-info';
 import { inlineTemplate } from 'storybook/decorators/storyTemplates';
 import { storybookPackageName, DOCUMENTATION_URL, STORYBOOK_CATEGORY } from 'storybook/constants/siteConstants';
 
-import { LoadingState } from './index';
-
+import LoadableState, { LoadableStateSource } from './Stories/LoadableState';
+import LoadingState from './LoadingState';
 import { name } from '../../../package.json';
 
 const stories = storiesOf(`${storybookPackageName(name)}/${STORYBOOK_CATEGORY.COMMUNICATION}/Loading State`, module);
 
 stories.addDecorator(withKnobs);
 
-stories.add(
-  'Loading State',
-  withInfo('Loading State shows centered spinner when loading')(() => {
-    const loading = boolean('Loading', true);
+stories
+  .add(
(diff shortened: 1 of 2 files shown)


packages/patternfly-3/patternfly-react/src/components/LoadingState/LoadingState.stories.js
```diff
@@ -1,35 +1,84 @@
...
+
+      return inlineTemplate({
+        title: 'Loading State',
+        documentationLink: `${DOCUMENTATION_URL.PATTERNFLY_ORG_COMMUNICATION}loading-state/`,
+        story
+      });
+    })
+  )
+  .add(
+    'Loading State 2',
```
jeff-phillips-18: Let's call this something more like 'Loading State Delay Demo' or the like.

packages/patternfly-3/patternfly-react/src/components/LoadingState/LoadingState.stories.js
```diff
@@ -1,35 +1,84 @@
...
+    withInfo('Loading State shows in delay')(() => {
+      class LoadableState extends React.Component {
+        state = {
+          message: 'State 1',
```
jeff-phillips-18: How about `Pre-Loaded state`? Though I'm not sure showing the message is necessary.

packages/patternfly-3/patternfly-react/src/components/LoadingState/LoadingState.stories.js
```diff
@@ -1,35 +1,84 @@
...
+          const delay = number('spinner delay', 1000);
+          return (
+            <div>
+              <div>
+                <Button
+                  bsStyle="primary"
+                  onClick={() => {
+                    this.setState({ loading: true });
+                    const timer = setTimeout(() => {
+                      this.setState({ loading: false, message: 'State 2' });
```
jeff-phillips-18: And this could be `Loaded State` (if we leave the message)

packages/patternfly-3/patternfly-react/src/components/LoadingState/LoadingState.stories.js
```diff
@@ -1,35 +1,84 @@
...
-      title: 'Loading State',
-      documentationLink: `${DOCUMENTATION_URL.PATTERNFLY_ORG_COMMUNICATION}loading-state/`,
-      story
-    });
-  })
-);
+stories
+  .add(
+    'Loading State',
+    withInfo('Loading State shows centered spinner when loading')(() => {
```
jeff-phillips-18: Add `propTables` to include `LoadingState` component and `propTablesExclude` to exclude `LoadableState`

boaz1337:
//ping @jeff-phillips-18

priley86:
I am OK w/ this if it helps your use case, but would prefer to see us migrate to the React Suspense API (released in 16.6):
https://reactjs.org/blog/2018/11/27/react-16-roadmap.html#react-166-shipped-the-one-with-suspense-for-code-splitting

Issue #984, opened 2018-11-29
Dropdown interaction issues


mcarrano:
I found two issues when reviewing behavior of the Dropdown component:

(1) When the menu is open and the user clicks or touches the down arrow, dropdown toggle, or anywhere outside the menu, then the menu should close.

(2) When clicking a disabled menu item, nothing should happen. Currently the menu closes. This could mislead the user that an action has been taken when it has not.

kmcfaul:
Re: 
1. This looks like focus-trap-react is blocking the toggle, still looking into why and how to get around it.
2. This seems to happen because the onSelect event fires whenever the dropdown menu is clicked and not each dropdown item. Either the select method needs to have code to detect if the target has the disabled status, or I can pipe the onSelect event down to the dropdown item level so if the isDisabled status is set, no onSelect event will fire. Is there any preference on this? @mcarrano

jgiardino:
@mcarrano Regarding (1), would your expectation be that clicking/touching anywhere outside the menu to dismiss the menu?

mcarrano:
Yes, I think so. That's how the PF3 dropdown works (https://www.patternfly.org/pattern-library/widgets/#dropdowns). Do you agree @jgiardino ?

jgiardino:
Yep, sounds good. I updated the description to capture that, and also include touch so that we remember to test both click and touch experiences.

rachael-phillips:
Closing this issue

PR #1014, opened 2018-12-05
Dropdown interaction issues

kmcfaul:
Issue reference: #984 

**What**: Fix focus trap disabling the dropdown toggle function. Pipe onSelect event to DropdownItem, allowing its disabled property to control if the event is fired. In our example docs, this means that clicking a disabled link will not close the dropdown.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
@@ -93,7 +93,7 @@ class Dropdown extends React.Component {
             isOpen={isOpen}
             position={position}
             aria-labelledby={id}
-            onClick={event => onSelect && onSelect(event)}
+            onSelect={event => onSelect && onSelect(event)}
           >
             {renderedContent}
           </DropdownMenu>
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
@@ -3,7 +3,7 @@ import styles from '@patternfly/patternfly-next/components/Dropdown/dropdown.css
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import DropdownMenu from './DropdownMenu';
-import { DropdownPosition, DropdownDirection } from './dropdownConstants';
+import { DropdownPosition, DropdownDirection, DropdownContext } from './dropdownConstants';
 
 // seed for the aria-labelledby ID
 let currentId = 0;
@@ -88,15 +88,11 @@ class Dropdown extends React.Component {
       >
         {Children.map(toggle, oneToggle => cloneElement(oneToggle, { parentRef: this.parentRef, isOpen, id, isPlain }))}
         {isOpen && (
-          <DropdownMenu
-            component={component}
-            isOpen={isOpen}
-            position={position}
-            aria-labelledby={id}
-            onSelect={event => onSelect && onSelect(event)}
-          >
-            {renderedContent}
-          </DropdownMenu>
+          <DropdownContext.Provider value={event => onSelect && onSelect(event)}>
+            <DropdownMenu component={component} isOpen={isOpen} position={position} aria-labelledby={id}>
+              {renderedContent}
+            </DropdownMenu>
+          </DropdownContext.Provider>
         )}
       </div>
     );
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -16,6 +16,7 @@ const propTypes = {
   /** Forces display of the hover state of the element */
   isHovered: PropTypes.bool,
   /** Default hyperlink location */
+<<<<<<< HEAD
   href: PropTypes.string,
 <<<<<<< HEAD
   /** Additional props are spread to the container component */
@@ -24,6 +25,9 @@ const propTypes = {
   /** Function callback called when user selects item */
   onSelect: PropTypes.func
 >>>>>>> feat(Dropdown): disable selection of disabled menu items
+=======
+  href: PropTypes.string
+>>>>>>> feat(Dropdown): improved select event piping, added fix for focus trap toggle
 };
 
 const defaultProps = {
@@ -61,7 +65,7 @@ const DropdownItem = ({ className, children, isHovered, onSelect, component: Com
         <Component
           {...additionalProps}
           className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
-          onClick={!isDisabled ? onSelect : null}
+          onClick={!isDisabled ? onSelect : Function.prototype}
         >
           {children}
         </Component>
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -19,8 +19,10 @@ const propTypes = {
   href: PropTypes.string,
   /** Additional props are spread to the container component */
   '': PropTypes.any,
-  /** Function callback called when user selects item */
-  onSelect: PropTypes.func
+  /** Callback for select event */
+  onSelect: PropTypes.func,
+  /** Callback for click event */
+  onClick: PropTypes.func
 };
 
 const defaultProps = {
@@ -30,10 +32,20 @@ const defaultProps = {
   component: 'a',
   isDisabled: false,
   href: '#',
-  onSelect: Function.prototype
+  onSelect: Function.prototype,
+  onClick: Function.prototype
 };
 
-const DropdownItem = ({ className, children, isHovered, onSelect, component: Component, isDisabled, ...props }) => {
+const DropdownItem = ({
+  className,
+  children,
+  isHovered,
+  onSelect,
+  onClick,
+  component: Component,
+  isDisabled,
+  ...props
+}) => {
   const additionalProps = props;
   if (Component === 'a') {
     additionalProps['aria-disabled'] = isDisabled;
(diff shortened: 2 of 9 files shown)


jgiardino:
Thanks for making these updates! I tested across different devices. 

- Testing this with a mouse, it works as expected.
- Testing this with the keyboard, it works as expected.
- Testing on an iOS device, I can touch the toggle and collapse the menu. But touching anywhere outside the dropdown toggle or menu does not collapse the menu. Whereas with a mouse, I am able to click anywhere outside the toggle and menu to collapse the menu.
  - Since we plan to use the solution implemented for modal in  https://github.com/patternfly/patternfly-react/pull/1011 with the dropdown panel variation, it might be worth checking if this solution provides any similar interaction. Although I'm assuming we would only implement this for the variation we're calling Dropdown Panel (i.e. custom contents in the popup), and this might not help us with the menu solution.
- Testing this with JAWS, NVDA and VO, I am able to place focus on the disabled elements, which is expected (as noted in the [APG for Menu](https://www.w3.org/TR/wai-aria-practices-1.1/#menu)—scroll down to the green info box). When I hit the Enter key while a disabled item has focus, the menu dismisses. Unfortunately, I can't find any suggestion on what should happen if the Enter key is pressed at this time. My suggestion would be to ignore Enter key presses when the menu is open and a disabled item has screen reader focus. I don't think this is a major issue if that's not possible.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -41,6 +41,7 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
       <Component
         {...additionalProps}
         className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
+        onClick={onSelect}
```
michaelkro: There is a bit of an issue with always setting `onClick` to `onSelect` here.  If the dropdown has multiple button items, this will set all of their `onClick` handlers to `onSelect` (so all buttons will fire the same action).

One possible solution would be to add a `handleOnClick` handler to `DropdownItem` and do something like the following

```js
// ...
return (
  <li>
    <Component
      {...additionalProps}
      className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
      onClick={event => handleOnClick && handleOnClick(event, onSelect)}
    >
      {children}
    </Component>
  </li>
);
```

Then the consuming app could do something like

```js
<DropdownItem
  // ...
  component="button"
  handleOnClick={(event, onSelect) => {
    // fire off the particular action for this button
    onSelect();
  }}
>
// ...
</DropdownItem>
```

But this would mean that all `DropdownItem`s (even the "link" ones) will require an `onClick` handler to ensure that `onSelect` is fired.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -28,7 +28,7 @@ const defaultProps = {
   href: '#'
 };
 
-const DropdownItem = ({ className, children, isHovered, component: Component, isDisabled, ...props }) => {
+const DropdownItem = ({ className, children, isHovered, onSelect, component: Component, isDisabled, ...props }) => {
```
michaelkro: whoops!  `onSelect` needs to be added to the `propTypes`

kmcfaul:
Added onSelect to prop types, updated the onClick to also call the button's onClick method in addition to the onSelect, updated snapshots

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -41,6 +44,9 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
       <Component
         {...additionalProps}
         className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
+        onClick={() => {
+          Component === 'button' ? (props.onClick && props.onClick(), onSelect()) : onSelect();
```
michaelkro: Nice, def cleaner than my suggestion!  Just a couple more things :)

1.  Now that we're explicitly setting `onClick` handler using a prop, `onClick` should be added to the propTypes, and can be destructured along with the other props that we exclude from `...props`
1.  We should probably pass the onClick `event` object to the callbacks, just in case the consumers want/need it
2.  Seeing an eslint error here
    > [eslint] Expected an assignment or function call and instead saw an expression.  [no-unused-expressions]

    We can do something like the following to fix the error
    ```js
    onClick={event => {
      if (Component === 'button') {
        onClick && onClick(event);
        onSelect(event);
      } else {
        onSelect(event);
      }
    }}
    ```
    That being said, I'm not sure why Travis isn't running `prettier`, or maybe this particular rule has been removed.  Are we not standardizing around `prettier` and `eslint`?  @jschuler

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
```diff
@@ -91,7 +91,7 @@ class Dropdown extends React.Component {
             isOpen={isOpen}
             position={position}
             aria-labelledby={id}
-            onClick={event => onSelect && onSelect(event)}
+            onSelect={event => onSelect && onSelect(event)}
```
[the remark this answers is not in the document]
michaelkro: Actually, now that I think about it again, I realize that `<Dropdown />` isn't actually a select... it's a menu.  So `onChange` doesn't necessarily make sense either.

This is sort of a weird use case here...  What do you think we should do?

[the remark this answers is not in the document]
kmcfaul: This may be more of a case of bad nomenclature, as this onSelect isn't a property of the DropdownMenu but a prop that gets passed down to the DropdownItem itself where it is set to the onClick function.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -41,6 +56,14 @@ const DropdownItem = ({ className, children, isHovered, component: Component, is
       <Component
         {...additionalProps}
         className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
+        onClick={event => {
+          if (Component === 'button') {
+            onClick && onClick(event);
```
jschuler: Why both onClick and onSelect?

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.js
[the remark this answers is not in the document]
jschuler:
```suggestion
             onSelect={onSelect}
```

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
jschuler: I see, onSelect is added in the Menu. Maybe remove from propTypes so it is only documented that user can add onClick to the item?

jschuler: Actually, nevermind, better would be to use context i think

jschuler: I think it would be best to create a context in Dropdown, and then pass the onSelect through that instead of props. Additionally, when you return the onSelect to the user, apart from the event maybe also return the index

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -25,41 +28,65 @@ const defaultProps = {
...
 
-const DropdownItem = ({ className, children, isHovered, component: Component, isDisabled, ...props }) => {
-  const additionalProps = props;
-  if (Component === 'a') {
-    additionalProps['aria-disabled'] = isDisabled;
-    additionalProps.tabIndex = isDisabled ? -1 : additionalProps.tabIndex;
-  } else if (Component === 'button') {
-    additionalProps.disabled = isDisabled;
-  }
+class DropdownItem extends React.Component {
```
michaelkro: Out of curiosity, what is the reasoning behind switching to a class based component?  In any case, if we decide to keep this as a class, we can switch the context stuff over to the [contextType syntax](https://reactjs.org/docs/context.html#classcontexttype)

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -25,41 +28,65 @@ const defaultProps = {
...
+            ) : (
+              <Component
+                {...additionalProps}
+                className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
+                onClick={event => {
+                  if (!isDisabled) {
+                    if (Component === 'button') onClick && onClick(event);
+                    onSelect && onSelect(event);
+                  } else {
+                    Function.prototype;
```
michaelkro: Getting an eslint error here

> Expected an assignment or function call and instead saw an expression.eslint(no-unused-expressions)

I think, if we want the handler to do nothing in case `isDisabled` is true, we can just omit this `else` block

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -25,41 +28,65 @@ const defaultProps = {
   isHovered: false,
   component: 'a',
   isDisabled: false,
-  href: '#'
+  href: '#',
+  onSelect: Function.prototype,
```
michaelkro: Missing from `propTypes` :)

michaelkro: Oops, I'm wrong, we need to do the opposite.  I think we need to remove `onSelect` from `defaultProps`

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -18,7 +19,9 @@ const propTypes = {
   /** Default hyperlink location */
   href: PropTypes.string,
   /** Additional props are spread to the container component */
-  '': PropTypes.any
+  '': PropTypes.any,
+  /** Callback for click event */
+  onClick: PropTypes.func
```
jschuler: Also need to update the props in the DropdownItem.d.ts file?

kmcfaul: I will add the onClick to the ts file

PR #1032, opened 2018-12-11
feat(core): Popover positioning

jschuler:
Add popover positioning support

Closes: #487


jschuler:
@christiemolloy I have some questions regarding the popover. 
- Is there supposed to be specific animation tied into showing/hiding the popover?
- What should the behavior be if a popover would overflow in the current view? For example, if the popover is opening to the top, but there is not enough space to show the whole popover, should it be flipped to the bottom? This is the functionality I currently have implemented (but can be turned off).

@jgiardino Should the user be able to tab into the popover? What should the flow here be?

jgiardino:
> @jgiardino Should the user be able to tab into the popover? What should the flow here be?

For the popover, I would trap focus like we do for modal (#1011) and use the same keyboard interaction that we use for the modal. The expected keyboard interaction for modal is described in issue https://github.com/patternfly/patternfly-react/issues/561

Let me know if you have questions about this. The documentation regarding where to place initial focus in the modal/popover is quite long.

jschuler:
Thanks @jgiardino , I'll focus trap similar to the modal

christiemolloy:
@jschuler to your second point about overflow of the popover in your current view, I think thats definitely the right approach to flip the positioning, and that would be the case for all 4 positions. 

And to your first point there should be an animation tied to the popover. @mcarrano can you clarify this behavior?

jschuler:
@tlabaj tippy is powered by popper, but also does a lot of things we'd still have to do if we used popper straight up

packages/patternfly-4/react-core/src/components/Popover/Popover.js
```diff
@@ -1,40 +1,300 @@
...
+      <Tippy
+        onCreate={this.storeTippyInstance}
+        size={size}
+        zIndex={zIndex}
+        appendTo={appendTo}
+        content={content}
+        lazy
+        trigger={handleEvents ? 'click' : 'manual'}
+        isVisible={isVisible}
+        hideOnClick={shouldHideOnClick()}
```
tlabaj: do you really want to call the function here?

packages/patternfly-4/react-core/src/components/Popover/PopoverCloseButton.js
```diff
@@ -5,9 +5,9 @@ import { css } from '@patternfly/react-styles';
 import { Button } from '@patternfly/react-core';
 import { TimesIcon } from '@patternfly/react-icons';
 
-const PopoverCloseButton = ({ onClose }) => (
-  <div className={css(styles.popoverClose)}>
-    <Button onClick={onClose} variant="plain" aria-label="Action">
+const PopoverCloseButton = ({ onClose, ...rest }) => (
+  <div className={css(styles.popoverClose)} {...rest}>
+    <Button onClick={onClose} variant="plain" aria-label="Close">
```
tlabaj: Maybe add the aria-label as prop  that can be applied to a button for localization purposes.

packages/patternfly-4/react-core/src/components/Popover/Popover.js
jschuler: Yes, it basically maps the prop hideOnOutsideClick with values of `true` or `false` to `true` and `'toggle'`

packages/patternfly-4/react-core/src/components/Popover/PopoverCloseButton.js
jschuler: Good point, I'll do that

PR #1040, opened 2018-12-12
feat(BulletChart): Allow for custom values (non-percent) and legend text

### LINKED ISSUE #1031: BulletChart assumes values are percentages
The bullet chart assumes that given values given are percentages. Not only with the styled widths, heights, etc., but it also appends a `%` character to the tooltip value.

We need to create a more generic bullet chart, which accepts any value, like so:
screen shot 2018-12-10 at 2 13 27 pm

Here are a couple examples of given values being used as percentages.

BulletChartValue
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChartValue.js#L14

BulletChartAxisTic
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChartAxisTic.js#L10

BulletChartLegend
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChart.js#L125

jeff-phillips-18:
fixes #1031

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChart.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChart.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { patternfly } from '../../../common/patternfly';
 import { Tooltip } from '../../Tooltip/index';
+import { helpers } from '../../../common/helpers';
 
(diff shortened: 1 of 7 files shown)


jeff-phillips-18:
Yes @mturley the thresholds should also adjust if not using percents. Fixed. Thanks!

Issue #1044, opened 2018-12-12
Modal examples needed for testing


jgiardino:
As noted in https://github.com/patternfly/patternfly-next/issues/925, some working examples of the modal are needed to better understand how the aria and role attributes might affect the behavior of a modal in screen readers. 

Can we have a WIP PR created with the following examples? 

- example with `aria-label` (refer to [core](https://pf-next.com/components/ModalBox/examples/) example "ModalBox No Header Example" and use of `aria-label` on `.pf-c-modal-box`)
- example with `aria-labelledby`
- example with `aria-describedby`
- example with contents included in the modal other than the `aria-describedby` contents, header, and buttons (refer to [core](https://pf-next.com/components/ModalBox/examples/) example "ModalBox No Header Example" and use of `<span id="modal-no-header-description">`)
- one example where contents are text and one where contents are a form would be ideal

For each example above, there should be two examples: 
- one with `role="dialog" aria-modal="true"` 
- and one without these attributes

rachael-phillips:
Closing this issue, completed

PR #1052, opened 2018-12-13
WIP PR for aria testing Modal component

kmcfaul:
**What**: Changed aria controls to allow customizable example for screen reader testing. It's worth noting that the current implementation had automatically set aria attributes based on other inputs (aria-label was set by the title attribute, and aria-describedby was set to an id of a container that the component automatically wrapped the body text in, while role and aria-modal were manually declared) and any additional props to Modal were passed to the Modal body and not the Modal box.

@jgiardino You should be able to live-edit the code in the aria example with whatever you need.

**Refer to issue**: #1044

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/ModalBoxHeader.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/ModalBoxHeader.js
@@ -17,7 +17,9 @@ const defaultProps = {
 
 const ModalBoxHeader = ({ children, className, ...props }) => (
   <header {...props} className={css(styles.modalBoxHeader, className)}>
-    <h1 className={css(styles.modalBoxHeaderTitle)}>{children}</h1>
+    <h1 className={css(styles.modalBoxHeaderTitle)} id="modal-header-h1">
+      {children}
+    </h1>
   </header>
 );
 
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.docs.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.docs.js
@@ -1,11 +1,16 @@
 import { Modal } from '@patternfly/react-core';
 import Simple from './examples/SimpleModal';
 import Large from './examples/LargeModal';
+import AriaLabel from './examples/AriaLabelModal';
 
 export default {
   title: 'Modal',
   components: {
     Modal
   },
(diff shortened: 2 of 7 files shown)


jgiardino:
Thanks for creating these examples, @kmcfaul!

I'm able to edit the code to create the variations I need with one exception... 

For the example "Simple Modal", I need to be able to reference the `id` of the `<h1>` element as shown below. I'm not seeing a way to add an attribute to the `<h1>` element. At a minimum, if you just add an id to the `<h1>`, I can add the `aria-labelledby` to the wrapping div.

```
<div aria-labelledby="[some value here]" class="pf-c-modal-box">
  <div class="pf-c-modal-box__close">...</div>
  <header class="pf-c-modal-box__header">
    <h1 class="pf-c-modal-box__header-title" id="[some value here]"> Modal Header </h1>
  </header>
  ...
</div>

**Comment by kmcfaul** [comment]
Added the id "modal-header-h1" to the h1 in the header for you to reference.

**Comment by jgiardino** [comment]
Awesome! Thank you. This should be all I need for testing.

File: packages/patternfly-4/react-core/src/components/Modal/ModalBoxHeader.js [hunk 1]
```diff
@@ -17,7 +17,9 @@ const defaultProps = {
 
 const ModalBoxHeader = ({ children, className, ...props }) => (
   <header {...props} className={css(styles.modalBoxHeader, className)}>
-    <h1 className={css(styles.modalBoxHeaderTitle)}>{children}</h1>
+    <h1 className={css(styles.modalBoxHeaderTitle)} id="modal-header-h1">
```
dmiller9911: Not use if this is just because it is a WIP or not, but this will cause issues of multiple modals are open. Will need to make this generated or passed in before this merges.

jgiardino: There are no plans to merge this one. This is intended to set up examples needed for testing.

jgiardino:
The previews for this PR worked great for testing! Thanks so much @kmcfaul for posting this!

I'm done with testing, and as a result noticed some minor enhancements we should make related to the modal: https://github.com/patternfly/patternfly-react/issues/1383 

I think this PR can be closed now.

Issue #1018, opened 2018-12-05
Document where additional props are added to elements


jschuler:
Something like:
```
/**
   * Additional props are spread to the input element
   */
  '': PropTypes.any
```

dgutride:
Closing via #1054

PR #1054, opened 2018-12-14
docs: add additional props documentation

kmcfaul:
Added documentation where additional props may be passed in to components and spread to containers or inner components. For components that pass props to child components whose props are not listed in documentation, the comment addresses where they are passed ultimately opposed to only stating the props are passed to the child.

**Refer to issue**: #1018


boaz1337:
@kmcfaul - can you rebase?

PR #1063, opened 2018-12-15
PF4: fix(LoginForm): LoginForm.d.ts fix typo passwordValue

### LINKED ISSUE #1062: PF4: Login Password field don't work
Hi there is a typo here https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.d.ts#L12

passwordValue instead of PasswordValue.

aljesusg:
fix #1062

Change a typo error in LoginForm component.

**What**:

Change passwordValue by passwordValue, we can't use the password field when you import this component


aljesusg:
cc @tlabaj I saw that you created this file. Without this we can't use the password field when we import the component. Thanks

PR #1066, opened 2018-12-15
feat(Add placeholder feature to TextInput): Create property placeholder in TextInput

aljesusg:
fix #1064
**What**:

let to the InputText the placehoder property

peek 2018-12-15 16-14

PR #1067, opened 2018-12-15
PF4: feat(LoginForm) let placeholder in username and password input

### LINKED ISSUE #1065: PF4: Can't set placeholder in LoginForm input fields.
Let set placeholder to the user and password fields in LoginForm

Required #1064

aljesusg:
Require: https://github.com/patternfly/patternfly-react/pull/1066
Fix #1065 

This let to the user set a placeholder in the username and password fields inside the LoginForm

peek 2018-12-15 16-20

### DIFF
--- a/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.d.ts
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.d.ts
@@ -5,12 +5,14 @@ export interface LoginFormProps extends HTMLProps<HTMLFormElement> {
   usernameLabel?: string;
   usernameValue?: string;
   onChangeUsername?: Function;
+  usernamePlaceholder?: string;
   usernameHelperText?: string;
   usernameHelperTextInvalid?: string;
   isValidUsername?: boolean;
   passwordLabel?: string;
   PasswordValue?: string;
   onChangePassword?: Function;
+  passwordPlaceholder?: string;
   passwordHelperText?: string;
   passwordHelperTextInvalid?: string;
   isValidPassword?: boolean;
--- a/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
@@ -14,6 +14,8 @@ const propTypes = {
   usernameValue: PropTypes.string,
   /** Function that handles the onChange event for the Username */
   onChangeUsername: PropTypes.func,
+  /** Placeholder for the User Input Field */
+  usernamePlaceholder: PropTypes.string,
   /** Helper Text for the Username Input Field */
   usernameHelperText: PropTypes.string,
   /** Helper Text for the Username Input Field when it is invalid */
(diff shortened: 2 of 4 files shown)


packages/patternfly-4/react-core/src/components/TextInput/TextInput.js
```diff
@@ -50,6 +52,7 @@ const defaultProps = {
   isDisabled: false,
   isReadOnly: false,
   type: 'text',
+  placeholder: '',
```
dmiller9911: Shouldn't need to add this. All input props are spread to the I put element. Placeholder will just work.

aljesusg: Checked, I don't know why not worked for me. Thanks

tlabaj: I think we should take a look and figure out why the spreading of props is not working.

jschuler: When using the LoginForm, the props are only spread to the Form, not the inner TextInput. So looks like we need to have these additional placeholder props on the LoginForm, or add more general userNameProps, passwordProps that are passed along

aljesusg: It's working @tlabaj , I said didn't know why not worked but it's working with this PR see the build. This is outdate

aljesusg:
@tlabaj comments are outdate. I removed the changes of textinput

aljesusg:
Can someone review this? Thanks !!!

maceledo:
We are advising against the use of placeholder text for a couple reasons. [See this article](https://www.nngroup.com/articles/form-design-placeholders/) or [this one](https://www.smashingmagazine.com/2018/06/placeholder-attribute/) for a more comprehensive overview, but I will list some key points below:


- One of Patternfly 4's key aspects is [a commitment to accessibility](https://pf-next.com/accessibility-guide). As such, we want to avoid anything that compromises accessibility, and placeholder text creates a few issues in this regard.
- Placeholder text doesn't meet the minimum contrast ratio which we are requiring all of our content to abide by. See the following - [reference](https://www.w3.org/WAI/WCAG21/quickref/#contrast-minimum) In order for the text color to pass, it would have to look too similar to inputted text, which creates an issue of confusing users as to which fields have been entered and which haven't. This alone is a 
- As placeholder text disappears when the input is active, it is disruptive to users who must navigate with keyboard controls. It also means we can't use placeholder text for crucial information, it would create a burden for the user having to clear the field if they need to re-reference the placeholder information.

With all of that said, we have had discussions around allowing placeholder text for only non-crucial syntax hints. I'm not sure where we landed there so I will page our accessibility expert @jgiardino and interaction expert @mcarrano for more thoughts. 

It's difficult to say we shouldn't support it in react, but we want to advise against the usage of it almost entirely.

tlabaj:
Unfortunately, Placeholders will not be supported due to accessibility concerns.

PR #1070, opened 2018-12-16
feat(Overlay): Export the Overlay component from react-bootstrap

sharvit:
**What**:
The `OverlayTrigger` component is great for most use cases, but as a higher level abstraction it can
lack the flexibility needed to build more nuanced or custom behaviors into your `Overlay`
components. For these cases it can be helpful to forgo the trigger and use the `Overlay` component
directly.

**Storybook**
https://rawgit.com/sharvit/patternfly-react/storybook/feature/overlay/index.html?knob-Placement=top&selectedKind=patternfly-react%2FWidgets%2FOverlay&selectedStory=Overlay%20story&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/Overlay/Overlay.stories.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Overlay/Overlay.stories.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs, select } from '@storybook/addon-knobs';
+import { withInfo } from '@storybook/addon-info';
+import { defaultTemplate } from 'storybook/decorators/storyTemplates';
+import { storybookPackageName, DOCUMENTATION_URL, STORYBOOK_CATEGORY } from 'storybook/constants/siteConstants';
+import { name } from '../../../package.json';
+import { MockOverlayManager, basicExampleSource } from './__mocks__/mockOverlayManager';
+import { Overlay } from './index';
+
+const stories = storiesOf(`${storybookPackageName(name)}/${STORYBOOK_CATEGORY.WIDGETS}/Overlay`, module);
+stories.addDecorator(withKnobs);
+stories.addDecorator(
+  defaultTemplate({
+    title: 'Overlay',
+    reactBootstrapDocumentationLink: `${DOCUMENTATION_URL.REACT_BOOTSTRAP_COMPONENT}overlays`
+  })
+);
+
+stories.add(
+  'Overlay',
+  withInfo({
+    source: false,
+    propTables: [Overlay],
+    propTablesExclude: [MockOverlayManager],
+    text: (
+      <div>
+        <h1>Story Source</h1>
+        <pre>{basicExampleSource}</pre>
+      </div>
+    )
+  })(() => (
+    <MockOverlayManager
+      placement={select('Placement', { top: 'Top', bottom: 'Bottom', right: 'Right', left: 'Left' }, 'top')}
+    />
+  ))
+);
--- a/packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
@@ -1,0 +1,91 @@
+/* eslint-disable react/no-find-dom-node */
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '../../Button';
+import { Popover } from '../../Popover';
+import { Overlay } from '../index';
+
+export class MockOverlayManager extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.buttonRef = React.createRef();
+    this.state = { showOverlay: false };
+  }
+  toggleOverlay = () => {
+    const { showOverlay } = this.state;
+    this.setState({ showOverlay: !showOverlay });
+  };
+  render() {
+    const { placement } = this.props;
+    const { showOverlay } = this.state;
+
+    return (
+      <React.Fragment>
+        <Button bsStyle="primary" bsSize="large" onClick={this.toggleOverlay} ref={this.buttonRef}>
+          Click Here
+        </Button>
+        <Overlay target={this.buttonRef.current} container={this} show={showOverlay} placement={placement}>
+          <Popover id="my-overlay">
+            <strong>Holy guacamole!</strong> Check this info.
+          </Popover>
+        </Overlay>
+      </React.Fragment>
+    );
+  }
+}
+
+MockOverlayManager.propTypes = {
+  placement: PropTypes.oneOf(['top', 'bottom', 'right', 'left'])
+};
+
+MockOverlayManager.defaultProps = {
+  placement: 'top'
+};
+
+export const basicExampleSource = `
+/* eslint-disable react/no-find-dom-node */
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '../../Button';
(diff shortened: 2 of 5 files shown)


packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
```diff
@@ -0,0 +1,99 @@
...
+    const { placement } = this.props;
+    const { showOverlay } = this.state;
+
+    return (
+      <div>
+        <Button bsStyle="primary" bsSize="large" onClick={this.toggleOverlay} ref={this.buttonRef}>
+          Click Here
+        </Button>
+        <Overlay
+          target={() => ReactDOM.findDOMNode(this.buttonRef.current)}
```
Laviro: Is it possible to get the DOM node already from `this.buttonRef.current` instead of using `ReactDOM.findDOMNode` ?

sharvit: I don't think so, see the official doc from react-bootstrap: https://react-bootstrap.github.io/components/overlays/#custom-overlays-overlay

glekner:
would love to see this merged! 👍

packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
sharvit: See: https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components

Laviro: In the link you sent they seem to refer to child nodes,
and they also say about `findDOMNode` that: ` your last option is to use findDOMNode(), but it is discouraged and deprecated in StrictMode.`

packages/patternfly-3/patternfly-react/src/components/Overlay/Overlay.stories.js
```diff
@@ -0,0 +1,37 @@
...
+stories.addDecorator(withKnobs);
+stories.addDecorator(
+  defaultTemplate({
+    title: 'Overlay',
+    reactBootstrapDocumentationLink: `${DOCUMENTATION_URL.REACT_BOOTSTRAP_COMPONENT}overlays`
+  })
+);
+
+stories.add(
+  'Overlay story',
```
Laviro: could you rename to `Overlay` just to be consistent with other components stories labels.

packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
sharvit: I'm confused, isn't it the current case? Do I have another access to the actual dom-node?

Laviro: just changed the overlay target into: `target={this.buttonRef.current}`,
I checked, it works :)

packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+  toggleOverlay = () => {
+    this.setState({ showOverlay: !this.state.showOverlay });
+  };
+  render() {
+    const { placement } = this.props;
+    const { showOverlay } = this.state;
+
+    return (
+      <div>
```
Laviro: optional: you can use `React.Fragment` here instead of `div` so the DOM will be cleaner

packages/patternfly-3/patternfly-react/src/components/Overlay/Overlay.stories.js
```diff
@@ -0,0 +1,37 @@
...
+  withInfo({
+    source: false,
+    propTables: [Overlay],
+    propTablesExclude: [MockOverlayManager],
+    text: (
+      <div>
```
Laviro: ditto

packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable react/no-find-dom-node */
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import { Button } from '../../Button';
+import { Popover } from '../../Popover';
+import { Overlay } from '../index';
+
+export class MockOverlayManager extends React.Component {
+  constructor() {
```
jeff-phillips-18: should take props and pass along props to `super`. I know its just a mock class but good practice 😉

packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
```diff
@@ -0,0 +1,99 @@
...
+    super();
+    this.buttonRef = React.createRef();
+    this.state = { showOverlay: false };
+  }
+  toggleOverlay = () => {
+    this.setState({ showOverlay: !this.state.showOverlay });
```
Laviro: optional: you have been using object destruct already,
can you change `toggleOverlay` into:
```
toggleOverlay = () => {
    const { showOverlay } = this.state;
    this.setState({ showOverlay: !showOverlay });
  };
```

packages/patternfly-3/patternfly-react/src/components/Overlay/Overlay.stories.js
sharvit: This is actually consistent with other stories:
https://github.com/patternfly/patternfly-react/blob/aa839a80a005f5d540faeb0e1a0320b7df1cc982/packages/patternfly-3/patternfly-react/src/components/MessageDialog/MessageDialog.stories.js#L65
https://github.com/patternfly/patternfly-react/blob/aa839a80a005f5d540faeb0e1a0320b7df1cc982/packages/patternfly-3/patternfly-react/src/components/Label/Label.stories.js#L39
https://github.com/patternfly/patternfly-react/blob/aa839a80a005f5d540faeb0e1a0320b7df1cc982/packages/patternfly-3/patternfly-react/src/components/Masthead/Masthead.stories.js#L81
https://github.com/patternfly/patternfly-react/blob/aa839a80a005f5d540faeb0e1a0320b7df1cc982/packages/patternfly-3/patternfly-react/src/components/Wizard/Stories/LoadingWizardExampleStory.js#L22

packages/patternfly-3/patternfly-react/src/components/Overlay/__mocks__/mockOverlayManager.js
sharvit: Awesome :fireworks: 
I'm not sure why the react-bootstrap example is so outdated, was pretty sure there is a reason for it.

sharvit:
Thanks @Laviro @jeff-phillips-18, can you revisit please?

packages/patternfly-3/patternfly-react/src/components/Overlay/Overlay.stories.js
Laviro: oh sorry, I meant using `React.Fragment` here instead of `div`.

sharvit:
Thanks @jeff-phillips-18 :+1:

PR #1071, opened 2018-12-16
fix(DualList): Add item disability functionality.

Laviro:
When an item is disabled,
the user won't be able to move it between the lists:
dec 16 2018 4_22 pm

### DIFF
--- a/packages/patternfly-3/patternfly-react/less/dual-list-selector.less
+++ b/packages/patternfly-3/patternfly-react/less/dual-list-selector.less
@@ -117,12 +117,22 @@
     color: white;
   }
 
+  &.disabled {
+    cursor: not-allowed;
+    background: @color-pf-black-150;
+    color: @color-pf-black-500;
+
+    input[type='checkbox'] {
+      cursor: not-allowed;
+    }
+  }
+
   &.child {
     padding-left: 22px;
   }
 
   &:hover {
-    &:not(.selected) {
+    &:not(.selected):not(.disabled) {
       background-color: @color-pf-blue-100;
       color: inherit;
     }
--- a/packages/patternfly-3/patternfly-react/src/components/DualListSelector/components/DualListItem.js
+++ b/packages/patternfly-3/patternfly-react/src/components/DualListSelector/components/DualListItem.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import DualListItemTooltip from './DualListItemTooltip';
 import { TypeAheadSelect } from '../../../components/TypeAheadSelect';
 import { noop } from '../../../common/helpers';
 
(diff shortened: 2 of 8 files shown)


packages/patternfly-3/patternfly-react/src/components/DualListSelector/components/DualListItem.js
```diff
@@ -70,7 +92,10 @@ DualListItem.defaultProps = {
   filterTerm: null,
   onChange: noop,
   side: null,
-  hidden: false
+  hidden: false,
+  disabled: false,
+  tooltipID: `dual-list-item-tooltip-${UUID()}`,
```
dmiller9911: This only runs once at module import. Not when the component is instantiated so if someone mounts multiple of these they will all share the same id.

Laviro: Wow thanks.. I didn't know that..
maybe set the ID by the item side+position which is unique enough

jeff-phillips-18: What we've done in other components is to set the default to null then check for it and if it is not passed use:
```
const randomId = () => Date.now();
...
<Tooltip id={tooltipID || randomId()}>
```

Laviro: Tell me what do you think, added this:
```
const getTooltipID = () => {
    let uniqueTooltipID = `dual-list-item-tooltip-${side}`;
    if (parentPosition) {
      uniqueTooltipID += `-${parentPosition}`;
    }
    uniqueTooltipID += `-${position}`;
    return uniqueTooltipID;
  };
```
and then using it like this when the default `tooltipID` is `null`.
```
    <DualListItemTooltip text={tooltipText} id={tooltipID || getTooltipID()}>
```

packages/patternfly-3/patternfly-react/src/components/DualListSelector/components/DualListItemTooltip.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import UUID from 'uuid/v1';
```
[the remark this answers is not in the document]
Laviro: oops, forgot to remove this one.. 
what do you think of the implementation above? isn't it unique enough ?
thanks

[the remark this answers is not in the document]
jeff-phillips-18: Yup. Thanks.

packages/patternfly-3/patternfly-react/less/dual-list-selector.less
```diff
@@ -117,12 +117,22 @@
     color: white;
   }
 
+  &.disabled {
+    cursor: not-allowed;
+    background: @color-pf-black-150;
+    color: @color-pf-black-500;
```
dtaylor113: nit. supposed to be in alphabetical order.

jeff-phillips-18: we haven't been keeping up with that in this repo

PR #1074, opened 2018-12-17
feat(BulletChart): Add Thresholds to Legend

jeff-phillips-18:
Provides the ability to add Threshold items to the legend if either text or a function is provided.

### DIFF
--- a/packages/patternfly-3/patternfly-react/less/charts.less
+++ b/packages/patternfly-3/patternfly-react/less/charts.less
@@ -1,0 +1,9 @@
+.bullet-chart-pf-legend-item-box {
+  &.error {
+    background-color: @alert-danger-border;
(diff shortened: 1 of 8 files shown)


dtaylor113:
I'm a little confused.  Should the bar turn orange and/or red when over Warning or Error thresholds...like the arc does in gauge charts?

jeff-phillips-18:
No, the bar remains the same. The threshold lines are used as indicators. See https://www.patternfly.org/pattern-library/data-visualization/bullet-chart/#design

dlabrecq:
Is there a knob to show the threshold in the legend?

jeff-phillips-18:
That would be up to the application to decide. That text is meant for the range bars.

Issue #1002, opened 2018-12-03
Dropdown arrow key interaction


jgiardino:
For Dropdown keyboard interaction to be consistent with the ARIA Authoring Practices Guide and with more complex but similar components like type-ahead select, we should implement arrow key interaction.

Interaction should be based on the documentation provided by ARIA Authoring Practices Guide for the [Menu Button](https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton), which also references documentation provided for [Menus](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) that describes how to navigate within a menu once it's visible. 

  See [this comment](https://github.com/patternfly/patternfly-react/issues/1002#issuecomment-452763351) below with specific details on what these updates are.

Note that the Dropdown currently has two variations, one that displays a menu and one that could display any contents. The arrow key interaction would be implemented for the menu variation only. The other variation would be handled more like a modal or popover in terms of how keyboard interaction is implemented (see issues #839 for trapping focus in the dropdown panel, and #561 for trapping focus in the modal).

~Also note that at this point in time, we will not implement the keyboard interaction labeled as optional for Menu Buttons. See the comments below for more details.~ Refer to [comment](https://github.com/patternfly/patternfly-react/issues/1002#issuecomment-452763351) below for latest updates on this.

jgiardino:
@mcarrano In the documentation referenced above for Menu Button, they state that using an Arrow key to open the menu is optional. Do you have any suggestions on this? I would assume that if we allow the Down arrow key to open the menu for a Dropdown, that it would also open the panel for a Dropdown Panel, since these things look identical when collapsed. But do you think this would be a weird interaction for the Dropdown Panel? 

We could also skip that part for now, since it's labeled as "optional" and revisit it later if we change our mind.

mcarrano:
I agree that we can use the down-arrow key to open the menu or panel as you suggest @jgiardino The down arrow has a good semantic mapping to the action of opening more content below, I think.

jgiardino:
After further discussion with @mcarrano, we are not going to implement the following optional keyboard interaction defined by the ARIA APG for [Menu Buttons](https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton):

> - (Optional) Down Arrow: opens the menu and moves focus to the first menu item.
> - (Optional) Up Arrow: opens the menu and moves focus to the last menu item.

[...]

PR #1075, opened 2018-12-17
Dropdown arrow key support

kmcfaul:
**What**: Dropdown menus should be navigable via up and down arrow keys. 

**Refers to issue**: #1002

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.test.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.test.js
@@ -81,7 +81,7 @@ describe('dropdown', () => {
   test('basic', () => {
     const view = mount(
       <Dropdown isOpen toggle={<DropdownToggle id="Dropdown Toggle">Dropdown</DropdownToggle>}>
-        BASIC
+        <div>BASIC</div>
       </Dropdown>
     );
     expect(view).toMatchSnapshot();
@@ -143,7 +143,7 @@ describe('KebabToggle', () => {
   test('basic', () => {
     const view = mount(
       <Dropdown isOpen toggle={<KebabToggle id="Dropdown Toggle" />}>
-        BASIC
+        <div>BASIC</div>
       </Dropdown>
     );
     expect(view).toMatchSnapshot();
@@ -179,7 +179,7 @@ describe('API', () => {
         isOpen
         toggle={<DropdownToggle id="Dropdown Toggle">Dropdown</DropdownToggle>}
       >
-        Children items
+        <div>Children items</div>
       </Dropdown>
     );
     expect(myMock).toBeCalled();
@@ -203,7 +203,7 @@ describe('API', () => {
     global.console = { error: myMock };
     mount(
       <Dropdown isOpen toggle={<DropdownToggle id="Dropdown Toggle">Dropdown</DropdownToggle>}>
-        Children items
+        <div>Children items</div>
       </Dropdown>
     );
     expect(myMock).not.toBeCalled();
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -57,10 +57,8 @@ class DropdownItem extends React.Component {
     if (event.key === 'Tab') return;
     event.preventDefault();
     if (event.key === 'ArrowUp') {
-      console.log('arrow up pressed');
       this.props.keyHandler(this.props.index, 'up');
     } else if (event.key === 'ArrowDown') {
-      console.log('arrow down pressed');
       this.props.keyHandler(this.props.index, 'down');
     }
   };
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -62,6 +62,8 @@ class DropdownItem extends React.Component {
       this.props.context.keyHandler(this.props.index, 'up');
     } else if (event.key === 'ArrowDown') {
       this.props.context.keyHandler(this.props.index, 'down');
+    } else if (event.key === 'Enter') {
+      this.ref.click();
     }
   };
 
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
@@ -35,23 +35,42 @@ const defaultProps = {
 };
 
 class DropdownItem extends React.Component {
+  constructor(props) {
+    super(props);
+    this.ref = React.createRef();
+    this.onKeyDown = this.onKeyDown.bind(this);
+  }
+
+  onKeyDown = event => {
+    if (event.key === 'Tab') return;
+    event.preventDefault();
+    if (event.key === 'ArrowUp') {
+      console.log('arrow up pressed');
+      this.props.keyHandler(this.ref);
+    } else if (event.key === 'ArrowDown') {
+      console.log('arrow down pressed');
+      this.props.keyHandler(this.ref);
+    }
+  };
+
   render() {
     const {
       className,
       children,
       isHovered,
+      keyHandler,
       onClick,
       component: Component,
       isDisabled,
-      ...additionalProps
+      ...props
     } = this.props;
+    const additionalProps = props;
     if (Component === 'a') {
       additionalProps['aria-disabled'] = isDisabled;
       additionalProps.tabIndex = isDisabled ? -1 : additionalProps.tabIndex;
     } else if (Component === 'button') {
       additionalProps.disabled = isDisabled;
     }
-
     return (
       <DropdownContext.Consumer>
         {onSelect => (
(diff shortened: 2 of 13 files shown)


kmcfaul:
I removed the mapping for the key handler when the children elements are passed in, and not DropdownItems, for the moment, as they were causing invariant errors and React warnings because the keyHandler prop was being read as a DOM element.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -25,28 +28,68 @@ const defaultProps = {
...
-  const additionalProps = props;
-  if (Component === 'a') {
-    additionalProps['aria-disabled'] = isDisabled;
-    additionalProps.tabIndex = isDisabled ? -1 : additionalProps.tabIndex;
-  } else if (Component === 'button') {
-    additionalProps.disabled = isDisabled;
+class DropdownItem extends React.Component {
+  constructor(props) {
+    super(props);
+    this.ref = React.createRef();
```
dmiller9911: No need for this to be in the constructor. The class properties plugin is being used. This will match other components.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -25,28 +28,68 @@ const defaultProps = {
...
+    this.onKeyDown = this.onKeyDown.bind(this);
```
dmiller9911: Same with this. If it is declared as a class property using the arrow notation there is no need to bind here.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -25,28 +28,68 @@ const defaultProps = {
...
+    } else if (Component === 'button') {
+      additionalProps.disabled = isDisabled;
+    }
+    return (
+      <li>
+        <Component
+          {...additionalProps}
+          className={css(isDisabled && styles.modifiers.disabled, isHovered && styles.modifiers.hover, className)}
+          onKeyDown={this.onKeyDown}
+          ref={ref => (this.ref = ref)}
```
dmiller9911: Since you are using createRef you can simply pass this.ref to it. Then you can access the current ref with this.ref.current

packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
```diff
@@ -27,25 +27,85 @@ const defaultProps = {
...
-      </Component>
+class DropdownMenu extends React.Component {
+  constructor(props) {
+    super(props);
+    this.refsCollection = {};
+    this.keyHandler = this.keyHandler.bind(this);
+  }
+
+  keyHandler(index, position, custom = false) {
+    const kids = this.props.children;
```
dmiller9911: This could break if consumers put items in a fragment. It might be better to make use of context to pass things around.

kmcfaul: I'm not sure what you mean by using context? I am seeing that having a child that is a React.Fragment will throw an error but should we just handle that when the children get extended (wrapping them in a div) or is there a better method?

ibolton336: The context API https://reactjs.org/docs/context.html

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
ibolton336: there is a babel plugin that pretty much abstracts away any use of the constructor & adds any definition within the class as a class property instead of having to explicitly do so in the constructor.

jgiardino:
When I test this component using just my keyboard, it works as expected. When I test in JAWS however, I don't see the behavior where hitting the Down arrow key on the last item shifts focus to the first item.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
kmcfaul: Ah I see, cool. I removed the constructor.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
kmcfaul: I can move keyHandler and sendRef to a Context but they only get passed down one layer so I'm not sure it saves much. I'm also confused about the case of putting the items in a Fragment because the top level Dropdown component should be expecting an array.

jgiardino:
Something else I missed in my previous review, when using just the keyboard without screen reader to navigate this component, the user should be able to hit the Tab key while the menu is open to close the menu and shift focus to the next element on the page. 

You can see this behavior in the [wai-aria example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-actions.html)

This behavior is expected for the variations that display a menu, but not for the variation that's labeled "Dropdown Panel" in our preview. In that case, Tab key is now the user would navigate with the contents within the panel.

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.test.js
```diff
@@ -81,7 +81,7 @@ describe('dropdown', () => {
   test('basic', () => {
     const view = mount(
```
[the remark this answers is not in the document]
kmcfaul: Updated the tests.

jgiardino:
Great job on this!!! 🎉 

I am able to test in Chrome, both with VO and normal keyboard interaction, and it works as I'd expect for both cases.

When I try to test in FF or Safari, the Dropdown doesn't even open, but I have been having issues with the pf-react previews (e.g. sometimes they're cached and they don't behave as expected), so I'm not sure if that's an issue with the Dropdown, or with these browsers properly loading these previews. I'm hoping it's the latter :-)

When I test on my phone (Safari or Chrome), I'm able to touch the toggle to close the menu, but touching outside the toggle or menu does nothing. This was something that was fixed in https://github.com/patternfly/patternfly-react/pull/1152. But I also wouldn't be surprised if this is more of a preview issue than a component issue. I'm trying to test the latest version of the workspace, and the Dropdown there doesn't work on my phone at all.

jgiardino:
UPDATE - I was able to get the preview to load in Firefox. It appears to work the same as in Chrome with one exception... When I click the toggle to display the menu, I should be able to hit the Esc key to dismiss the menu. That works in Chrome, but not in FF.

I'll keep trying to get the other previews to work, but wanted to pass that along.

beanh66:
Thanks @kmcfaul, this looks great! Worked well when I tried it out in Chrome.

jgiardino:
I was able to get it working in Safari, too. The issue I noted above is also the case in Safari:

> When I click the toggle to display the menu, I should be able to hit the Esc key to dismiss the menu.

But then I found this, which might explain why:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Clicking_and_focus

Most browsers place focus on the button on click, but Safari and Firefox on a mac don't. 😕  Let me know your thoughts on this one. It _is_ an edge case, and the user can still dismiss the menu with other ways, so I'm not sure how much we should try to handle that interaction.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -31,31 +36,61 @@ const defaultProps = {
...
+  onClick: Function.prototype,
+  index: -1,
+  context: {
+    keyHandler: Function.prototype,
+    sendRef: Function.prototype
+  }
 };
 
 class DropdownItem extends React.Component {
+  componentWillMount() {
```
dmiller9911: ` willMount` is being deprecated. You can write this like below
```
class DropdownItem extends React.Component {
  ref = React.createRef()
  ...
```

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -31,31 +36,61 @@ const defaultProps = {
...
+    // Detected key press on this item, notify the menu parent so that the appropriate
+    // item can be focused
+    if (event.key === 'Tab') return;
+    event.preventDefault();
+    if (event.key === 'ArrowUp') {
+      this.props.context.keyHandler(this.props.index, 'up');
+    } else if (event.key === 'ArrowDown') {
+      this.props.context.keyHandler(this.props.index, 'down');
+    } else if (event.key === 'Enter') {
+      this.ref.click();
```
dmiller9911: This, i beleive should be `this.ref.current.click()`. Also since it is possible to pass a component in you will need to make sure that `click` exists.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
```diff
@@ -29,12 +28,117 @@ const defaultProps = {
   component: 'ul'
 };
 
-const DropdownMenu = ({ className, isOpen, position, children, component: Component, ...props }) => {
-  let menu = null;
-  if (Component === 'div') {
-    menu = (
-      <DropdownContext.Consumer>
-        {onSelect => (
+class DropdownMenu extends React.Component {
```
dmiller9911: to keep consistent with the repo modify this to be
```jsx
class DropdownMenu extends React.Component {
  refsCollection = {};

  keyHandler = (index, position, custom = false) => { ... }
}

Issue #976, opened 2018-11-28
Chip 


rachael-phillips:
Related Core Issue: https://github.com/patternfly/patternfly-next/issues/177 

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @christiemolloy and @mcarrano 

React Interaction: 
- Close button should remove chip. 
- Overflow button should show and hide chips
- Overflow button should count how many chips have overflown

tconn89:
I do not see a Chip component in the code.  Are you referring to a Badge?

If not are you referring to a tag?  If so, would this issue be dependent on PR #351 Introduce Tagging React component? 
https://github.com/patternfly/patternfly-react/pull/351

christiemolloy:
@tconn89 https://pf-next.com/components/Chip/examples/ and https://pf-next.com/components/ChipGroup/examples/

LHinson:
@ibolton336 will you be providing the api review/recommendations only are do you also plan to implement the component?

ibolton336:
@LHinson Ah I thought this issue only involved the API review recommendations.

LHinson:
@ibolton336 the issue covers both. the standard initial API review/recommendations and then moves on to component implementation. I just wasn't sure if you were assigned to complete one or both.

ibolton336:
@LHinson @rachael-phillips I can take a crack at implementing the react component for this.

rachael-phillips:
Ok great! Thank you @ibolton336 !

PR #1076, opened 2018-12-18
feat(Chip): Add the Chip component to PF4

boaz0:
What: [#976](https://github.com/patternfly/patternfly-react/issues/976)

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Chip/Chip.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Chip/Chip.d.ts
@@ -1,0 +1,20 @@
+import { SFC, HTMLProps } from 'react';
+import { Omit, OneOf } from '../../typeUtils';
+import { TooltipPosition } from '../Tooltip';
+
+export const ChipVariant = {
+  overflow: 'overflow',
+  closable: 'closable'
+};
+
+export interface ChipProps extends Omit<HTMLProps<HTMLDivElement>, 'children' > {
+  children: string;
+  isOverflowed?: boolean;
+  variant: OneOf<typeof ChipVariant, keyof typeof ChipVariant>;
+  position: OneOf<typeof TooltipPosition, keyof typeof TooltipPosition>;
+}
+
+declare const Chip: SFC<ChipProps>;
+
+export default Chip;
+
--- a/packages/patternfly-4/react-core/src/components/Chip/Chip.js
+++ b/packages/patternfly-4/react-core/src/components/Chip/Chip.js
@@ -1,0 +1,69 @@
+import React from 'react';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import ChipButton from './ChipButton';
+import { Tooltip, TooltipPosition } from '../Tooltip';
+import { TimesCircleIcon } from '@patternfly/react-icons';
+import styles from '@patternfly/patternfly-next/components/Chip/chip.css';
+import { getUniqueId } from '../../internal/util';
+
+export const ChipVariant = {
+  overflow: 'overflow',
+  closable: 'closable'
+};
+
+const Chip = ({ variant, onClick, children, id, position, className, ...props }) => {
+  const idChip = id || getUniqueId()
+  switch (variant) {
+    case ChipVariant.overflow:
+      return (
+        <div className={css(styles.chip, styles.modifiers.overflow, className)} {...props}>
+          <ChipButton onClick={onClick} aria-label="Expand chip">
+            <span className={css(styles.chipText)}>{children}</span>
+          </ChipButton>
+        </div>
+      );
+    default:
+      const ChipComponent = (
+        <div className={css(styles.chip, className)} {...props}>
+          <span className={css(styles.chipText)} id={idChip}>
+            {children}
+          </span>
+          <ChipButton onClick={onClick} aria-label="Remove" id={`remove_${idChip}`} aria-labelledby={`remove_${idChip} ${idChip}`}>
+            <TimesCircleIcon aria-hidden="true" />
+          </ChipButton>
+        </div>
+      );
+      return children.length < 16 ? (
+        ChipComponent
+      ) : (
+        <Tooltip position={position} content={children}>
+          {ChipComponent}
+        </Tooltip>
+      );
+  }
+};
+
+Chip.propTypes = {
+  /** Content rendered inside the chip text */
+  children: PropTypes.string.isRequired,
+  /** ID of the chip */
+  id: PropTypes.string,
+  /** Additional classes added to the chip item */
+  className: PropTypes.string,
+  /** Builds the chip structure according to variant */
+  variant: PropTypes.oneOf(Object.values(ChipVariant)),
+  /** Position of the tooltip which is displayed if text is longer */
+  position: PropTypes.oneOf(Object.values(TooltipPosition)),
+  /** Function that is called when clicking on the chip button */
+  onClick: PropTypes.func
+};
+
+Chip.defaultProps = {
+  id: undefined,
+  className: '',
+  position: 'top',
+  variant: ChipVariant.closable
+};
+
+export default Chip;
--- a/packages/patternfly-4/react-core/src/components/Chip/examples/SimpleChip.js
+++ b/packages/patternfly-4/react-core/src/components/Chip/examples/SimpleChip.js
@@ -1,0 +1,42 @@
+import React from 'react';
+import { Chip } from '@patternfly/react-core';
+
+class SimpleChip extends React.Component {
+  state = {
+    show: ['chip_one', 'chip_two']
+  };
+  deleteItem = id => {
+    const { show } = this.state;
+    const ix = show.indexOf(id);
+    if (ix < 0) {
+      return;
+    }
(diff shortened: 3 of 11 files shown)


packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,45 @@
...
+import ChipButton from './ChipButton';
+import ChipText from './ChipText';
+import ChipWrapper from './ChipWrapper';
+import { TimesCircleIcon } from '@patternfly/react-icons';
+
+const Chip = ({ isOverflowed, children, id, className, ...props }) => {
+  return !isOverflowed ? (
+    <ChipWrapper className={className} {...props}>
+      <ChipText id={id}>{children}</ChipText>
+      <ChipButton aria-label="Remove" id={`remove_${id}`} aria-labelledby={`remove_${id} ${id}`}>
```
karelhala: You are missing `onClick` for remove button. Clicking on chip anywhere should not call the on click function, this function should be called only for `ChipButton`.

boaz1337: Thanks! I can't believe I missed it. I will fix it soon as possible.

packages/patternfly-4/react-core/src/components/Chip/examples/SimpleChip.js
```diff
@@ -0,0 +1,41 @@
...
+      return;
+    }
+    this.setState(() => ({
+      show: [...show.slice(0, ix), ...show.slice(ix + 1, show.length)]
+    }));
+  };
+  render() {
+    return (
+      <React.Fragment>
+        {this.state.show.includes('chip_one') && (
```
ibolton336: The component looks great - nice work. Just a few things - can we leverage destructuring here for cleaning up the render method of state references?

packages/patternfly-4/react-core/src/components/Chip/ChipWrapper.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
ibolton336: Can we use typescript for this component?

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ChipButton from './ChipButton';
+import ChipText from './ChipText';
+import ChipWrapper from './ChipWrapper';
+import { TimesCircleIcon } from '@patternfly/react-icons';
+
+const Chip = ({ isOverflowed, children, id, className, ...props }) => {
+  return !isOverflowed ? (
```
ibolton336: Can we use a different type of conditional rendering here that would lend itself to more than 2 types of chips? That would help with future proofing the component.

packages/patternfly-4/react-core/src/components/Chip/ChipWrapper.js
boaz1337: I thought `ChipWrapper.d.ts` was good enough. Can you give me an example of doing that? Like rename the file from `ChipWrapper.js` to `ChipWrapper.ts` or something more complicated?

Thanks :smile_cat:

packages/patternfly-4/react-core/src/components/Chip/examples/SimpleChip.js
boaz1337: Yeah sure!

packages/patternfly-4/react-core/src/components/Chip/Chip.js
boaz1337: I am not sure I understand what you mean.

I can do this differently in the following ways:

```js
const chipBody = !isOverflowed ? (
 <ChipText id={id}>{children}</ChipText>
      <ChipButton aria-label="Remove" id={`remove_${id}`} aria-labelledby={`remove_${id} ${id}`}>
        <TimesCircleIcon aria-hidden="true" />
      </ChipButton>
  ) : (
      <ChipButton aria-label={children}>
        <ChipText>{children}</ChipText>
      </ChipButton>
);

return <ChipWrapper isOverflowed={isOverflowed} className={className} {...props}>{chipBody}</ChipWrapper>
```
or maybe:

```js
const { chipType } = this.props;
switch(chipType) {
case "overflow":
  return (
    <ChipWrapper isOverflowed className={className} {...props}>
      <ChipButton aria-label={children}>
        <ChipText>{children}</ChipText>
      </ChipButton>
    </ChipWrapper>
  );
default:
  return (
    <ChipWrapper className={className} {...props}>
      <ChipText id={id}>{children}</ChipText>
      <ChipButton aria-label="Remove" id={`remove_${id}`} aria-labelledby={`remove_${id} ${id}`}>
        <TimesCircleIcon aria-hidden="true" />
      </ChipButton>
    </ChipWrapper>
  );
}
```

Let me know if you have a better or a different idea. Thanks :fish_cake:

ibolton336: I'd take a look at the current patterns used within the library. Maybe something like the button component "variant" prop approach?

packages/patternfly-4/react-core/src/components/Chip/ChipWrapper.js
ibolton336: The typescript stuff isn't fully baked yet for PF4 so I think your are right to just include the d.ts file. Eventually we will want more but for now this should be fine.

boaz1337: Cool :sunglasses: . At any rate, if I can help with that at any point let me know.

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,57 @@
...
+        </ChipWrapper>
+      );
+  }
+};
+
+Chip.propTypes = {
+  /** Content rendered inside the chip text */
+  children: PropTypes.string.isRequired,
+  /** ID of the chip */
+  id: props => {
```
tlabaj: In other components, if the consumer does not provide an id that is required for accessibility,  we  have been generating it internally.  Maybe we should consider doing that here and getting rid of this check.

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,57 @@
...
+export const ChipVariant = {
+  overflow: 'overflow',
+  closable: 'closable'
+};
+
+const Chip = ({ variant, onClick, children, id, className, ...props }) => {
+  switch (variant) {
+    case ChipVariant.overflow:
+      return (
+        <ChipWrapper isOverflowed className={className} {...props}>
```
tlabaj: I don't know if the chip wrapper class is really needed.  Is there a reason we can't have the div here in the Chip Class?  Same Question goes for the ChipText.

boaz0:
@tlabaj thanks. I removed `ChipText` and `ChipWrapper` as you suggested and used `getUniqueId` to set a default id.

@mcarrano yes. I will add `ChipGroup` later. I also did as you suggested: adding tooltip.

@mcoker thanks. I removed `aria-label`. Let me know if I'm missing anything.

Thank you all.

packages/patternfly-4/react-core/src/components/Chip/Chip.d.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Omit, OneOf } from '../../typeUtils';
+import { TooltipPosition } from '../Tooltip';
+
+export const ChipVariant = {
+  overflow: 'overflow',
+  closable: 'closable'
+};
+
+export interface ChipProps extends Omit<HTMLProps<HTMLDivElement>, 'children' > {
+  children: string;
```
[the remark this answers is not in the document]
jschuler: Can you get rid of the Omit 'children' and remove children from the interface? Also don't need id and onClick I believe since they are included in HTMLProps

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,69 @@
...
+        <div className={css(styles.chip, className)} {...props}>
+          <span className={css(styles.chipText)} id={idChip}>
+            {children}
+          </span>
+          <ChipButton onClick={onClick} aria-label="Remove" id={`remove_${idChip}`} aria-labelledby={`remove_${idChip} ${idChip}`}>
+            <TimesCircleIcon aria-hidden="true" />
+          </ChipButton>
+        </div>
+      );
+      return children.length < 16 ? (
```
jschuler: Curious why 16? Should this be customizable?

ibolton336:
@boaz0 I have opened a new PR for this as we need to expedite the process. I was unable to find a way to contact you but perhaps we can meet & pair program if your are in the Raleigh area this week. Thank you for your hard work on this. I have kept all of your commits intact & just cherry picked from your branch.

boaz0:
@ibolton336 no problem.

Although, I am in Raleigh this week, I don't think I have the time to join in.

Thanks for the invitation, though. :grinning:

PR #1078, opened 2018-12-18
fix(DualList): Adding debounce for filter changes

Laviro:
Reduce the events fired on filter change.

Called twice:
selection_110
Instead of 6 times:
selection_111


packages/patternfly-3/patternfly-react/src/components/DualListSelector/DualList.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import cloneDeep from 'lodash/cloneDeep';
+import debounce from 'lodash/debounce';
```
jeff-phillips-18: Please use `debounce` from `helpers` instead.

PR #1083, opened 2018-12-18
fix(pf3): fix usage of noop in BulletChart

jschuler:
Previous travis build was failing because helpers.noop was undefined. I could reproduce this locally. We can import just noop and use that instead.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChart.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Chart/BulletChart/BulletChart.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { patternfly } from '../../../common/patternfly';
 import { Tooltip } from '../../Tooltip/index';
-import { helpers } from '../../../common/helpers';
+import { noop } from '../../../common/helpers';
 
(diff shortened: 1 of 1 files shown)


jeff-phillips-18:
why is it failing? helpers.noop should be defined.

jschuler:
screen shot 2018-12-18 at 12 10 20 pm
@jeff-phillips-18

jschuler:
btw, I am not sure why it thinks the failure is in DualList, i see that the test is mocking common/helpers, wonder if that is related. Regardless, since BulletChart is only using noop it is anyways cleaner to just import that

jeff-phillips-18:
Ah, DualList tests are redefining helpers which causes it to lose the noop function. This is a quick fix but likely not the correct fix as it will probably come back to bite us again.

jeff-phillips-18:
...sigh

jschuler:
@jeff-phillips-18 I agree, at least this should get the build passing again. Let's put a proper fix in for the issue with DualList test redefining common helpers

jschuler:
@jeff-phillips-18 Created an issue here
https://github.com/patternfly/patternfly-react/issues/1085

jeff-phillips-18:
Thanks @jschuler

Issue #993, opened 2018-11-30
PF4 Brand alt text in Login


jgiardino:
The alt text in the brand component should be updated to match the example we have in core.

Currently, the react component uses the following alt text:
`alt="pf-logo" `
Whereas the example in core uses:
`alt="Patternfly Logo"`

Whenever text is included in an image, it should be repeated in the `alt` attribute value (this is how screen readers know what the text is). Additionally, alt text can describe the image (e.g. that it's a "logo").  

The Additional Notes section at the end of [this article](https://www.phase2technology.com/blog/no-more-excuses-definitive-guide-alt-text-field) provides a good checklist that's relevant for logos.

rebeccaalpert:
It looks like this was already fixed in this commit: https://github.com/patternfly/patternfly-react/commit/c1af473039e1ea03dfcab6905db77cbe9ac39315

rebeccaalpert:
Looks like @seanforyou23 fixed this already: https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Brand/examples/SimpleBrand.js

Going to close this up.

jgiardino:
Hey, sorry for not including a link to the example with the issue. I'm still seeing "pf-logo" for the [Login](https://patternfly-react.surge.sh/patternfly-4/components/loginpage/examples/simple-login-page/) component.

rebeccaalpert:
Thanks for the link. I'll take a look and see what's up.

rebeccaalpert:
I have a PR up here: https://github.com/patternfly/patternfly-react/pull/1084

Have you seen the issue anywhere else?

jgiardino:
I think that's it. Thanks for making this update!

PR #1084, opened 2018-12-18
fix: update alt text in LoginPage

rebeccaalpert:
Updated the branding alt text in the LoginPage to "Patternfly Logo." Per @jgiardino, it should match the example in core and repeat the text/describe the image.

Fixes [#993](https://github.com/patternfly/patternfly-react/issues/993).


jgiardino:
Thanks for making this update! I'm just now realizing I had a typo in my original description with "Patternfly logo". Could you update the string to have an upper-case F like "PatternFly logo"?

rebeccaalpert:
I added the capital "F." @tlabaj, could you review this again? Sorry for the trouble.

PR #1095, opened 2018-12-19
fix(DualList): Adding hidden HTML select inputs to reflect the selector items

Laviro:
The DualList doesn't act as an input yet when trying to use it in a form,
providing hidden select inputs will make the data available when the form is submitted.


packages/patternfly-3/patternfly-react/src/components/DualListSelector/components/DualListHeading.js
```diff
@@ -20,14 +20,12 @@ const DualListHeading = ({
   kebabID
 }) => (
   <div className="dual-list-pf-heading">
-    <Form inline>
```
Laviro: can't have a nested form if the DualList will be wrapped in a form.

ohadlevy: +1, imho this component should never introduce a form by it self, the same goes for hidden value, that would be up to developer to decide (out of context of patternfly).

Laviro: I think many consumers would be happy to receive this component as inputs to be able to submit it instead of connecting to it which might be more complex.
we can add a prop for it to let the consumers decide by themselves.

@jeff-phillips-18 @sharvit @glekner what are your thoughts about this?

Laviro:
@ohadlevy added `allowHiddenInputs` props 
to let the consumer determine whether to use this feature.

sharvit:
Looks good and works as expected.

Does it make sense to add another story `DualList in form`?
You can:
1. Use the `allowHiddenInputs`
2. Wrap it with a form
3. Add `onSubmit` with an action logger to show the submitted data

Laviro:
Thanks @jeff-phillips-18  & @sharvit,
added a story for Dual List in a form as you suggested,
see: https://rawgit.com/Laviro/patternfly-react/storybook/feature/dual-list-selector/index.html?selectedKind=patternfly-react%2FForms%20and%20Controls%2FDual%20List%20Selector&selectedStory=Dual%20List%20Selector%20as%20a%20form%20input&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

PR #1096, opened 2018-12-19
feat(KebabToggle): Customizable icon.

martinpovolny:
* Allow passing in alternative content for the drop-down symbol.
 * Remove unused children property.

Note: the `children` property may have been originally intended for this purposo but based on discussion with @karelhala I added a new property.

Needed by: https://github.com/RedHatInsights/insights-chrome/pull/90

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/KebabToggle.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/KebabToggle.js
@@ -5,17 +5,15 @@ import Toggle from './Toggle';
 
 const defaultAriaLabel = 'Actions';
 
-const Kebab = ({ ...props }) => (
+const Kebab = ({icon,  ...props }) => (
   <Toggle {...props}>
-    <EllipsisVIcon />
+    {icon ? icon : <EllipsisVIcon />}
   </Toggle>
 );
 
 Kebab.propTypes = {
   /** HTML ID of dropdown toggle */
   id: PropTypes.string,
-  /** Anything which can be rendered as dropdown toggle */
-  children: PropTypes.node,
   /** Classess applied to root element of dropdown toggle */
   className: PropTypes.string,
   /** Flag to indicate if menu is opened */
@@ -33,11 +31,12 @@ Kebab.propTypes = {
   /** Forces active state */
   isActive: PropTypes.bool,
   /** Display the toggle with no border or background */
-  isPlain: PropTypes.bool
+  isPlain: PropTypes.bool,
+  /** Anything that can be displayed instead of the default EllipsisVIcon as a dropdown toggle **/
+  icon: PropTypes.node
 };
 Kebab.defaultProps = {
   id: '',
-  children: null,
   className: '',
   isOpen: false,
   'aria-label': defaultAriaLabel,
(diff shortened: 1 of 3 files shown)


martinpovolny:
@dlabaj : What you write makes sense.  But it is in conflict with this: https://github.com/patternfly/patternfly-react/pull/1096/files#diff-79a5998575901e0ed514f77a27b535a2L17.

Actually I started by looking at why the `children` property did not work as documented: http://patternfly-react.surge.sh/patternfly-4/components/dropdown#KebabToggle that is passing on the `chindren` but then @karelhala told me to move this to a new property `icon`.

Now I have yet another varian. But right, it makes sense.

So I guess I change this PR to just drop the `children` property to make the documentation match the reality.

Does that make sense to you?

karelhala:
@dlabaj using dropdown toggle with icon as child is great idea.

@martinpovolny I think that updating props of kebab toggle and adding new example with custom icon is good way.

```JSX
import React, { Component } from 'react';
import { Dropdown, DropdownToggle, DropdownItem } from '@patternfly/react-core';
import { AtomIcon } from '@pattenfly/react-icons';

export default class CustomIconDropdown extends Component {
  constructor(props) {
    super(props);
    this.state = {
      isOpen: false
    };
  }

  onToggle = isOpen => {
    this.setState({
      isOpen
    });
  };

  onSelect = event => {
    this.setState({
      isOpen: !this.state.isOpen
    });
  };

  render() {
    const { isOpen } = this.state;
    const dropdownItems = [
      <DropdownItem key="link">Link</DropdownItem>
    ];
    return (
      <Dropdown
        onSelect={this.onSelect}
        toggle={<DropdownToggle onToggle={this.onToggle} iconComponent=""><AtomIcon /></DropdownToggle>}
        isOpen={isOpen}
        isPplain
        dropdownItems={dropdownItems}
      />
    );
  }
}
```

jschuler:
The name of the component `KebabToggle` implies that it is using a Kebab icon. I guess in a way, it is a more specific variant of the DropdownToggle. I agree, if we can open up the DropdownToggle to use any icon that would be preferred.

jgiardino:
I was chatting with designers @mcarrano and @kybaker on slack about their expectations for the toggle. If we use an icon other than kebab, the expectation is that the dropdown arrow would not display, similar to how the kebab doesn't include the dropdown arrow.

I didn't see an example included in this PR that shows how the toggle would look using this update. I'm also not sure if other suggestions mentioned above would result in an icon with or without the dropdown arrow.

Please just keep that expectation in mind as you discuss how to implement this 😄

martinpovolny:
@karelhala : your example is pretty similar what I did: 
https://github.com/RedHatInsights/insights-chrome/pull/90/files

Btw is there a reason for not allowing arrow functions in classes (methods? not sure about the proper javascript terminology) in the Insights Proxy? Or shall I make a PR there to allow it?

dlabaj:
@martinpovolny Added an example in this PR https://github.com/patternfly/patternfly-react/pull/1127

martinpovolny:
@dlabaj : Ok, cool. What about the (wrongly) documented "children" ? Are you going to fix that too?

PR #1098, opened 2018-12-19
fix(modal): rename reactRoot to srHide and add to ts file

jschuler:


packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -25,8 +25,8 @@ const propTypes = {
   onClose: PropTypes.func,
   /** Creates a large version of the Modal */
   isLarge: PropTypes.bool,
-  /** React application root element */
-  reactRoot: PropTypes.instanceOf(safeHTMLElement).isRequired
+  /** The parent element container to hide from screen-reader applications when the modal is open */
+  srHide: PropTypes.instanceOf(safeHTMLElement).isRequired
```
dlabrecq: Can we please remove the isRequired?

packages/patternfly-4/react-core/src/components/Modal/examples/SimpleModal.js
```diff
@@ -14,7 +14,7 @@ class SimpleModal extends React.Component {
 
   render() {
     const { isModalOpen } = this.state;
-    const reactRoot = typeof document !== 'undefined' ? document.querySelector('#___gatsby') : '';
+    const srHide = typeof document !== 'undefined' ? document.querySelector('#___gatsby') : '';
```
dlabrecq: As an app developer, it's not clear what I should use in place of the '#___gatsby' selector. It would be helpful to add a comment in this example describing what I should use instead.

packages/patternfly-4/react-core/src/components/Modal/Modal.d.ts
```diff
@@ -9,6 +9,7 @@ export interface ModalProps extends HTMLProps<HTMLDivElement> {
   isOpen?: boolean;
   onClose?: Function;
   title: string;
+  srHide: Element;
```
dlabrecq: Can we please make srHide optional? IMO, patternfly should make it easy to implement a11y, but not force it.

Not sure why this worked during my test? That is, I didn't provide an srHide property.

jschuler:
@tlabaj from what i can see About Modal does not apply aria-hidden (yet)

PR #1103, opened 2018-12-20
feat(tooltip): add tooltip to PF4

jschuler:
Basically a stripped down version of the Popover which got merged recently


jschuler:
@jgiardino @christiemolloy needs a11y and ux review

jgiardino:
There are a few criteria outlined for [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/quickref/#content-on-hover-or-focus) that we should meet for the tooltip component. I see that the tooltip remains visible during focus/hover and that I can hit the Esc key to dismiss it when visible, which is great! But we don't meet this criteria:
> Hoverable: If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

Would this be something that should be handled in pf-react or pf core?

jschuler:
Thanks @jgiardino, after the travis build finishes the preview should be updated with this feature (hover over tooltip should keep it open)

jgiardino:
Regarding the latest update, I'm seeing this in Firefox and Safari, but not in Chrome. If I move the cursor from the tooltip target toward the tooltip contents, the contents are dismissed.

I also had a chance to review this in screen readers and included my notes below. There are a couple of tweaks that would improve these experiences.
- Do you know if there's some setting in Tippy that is shifting focus to the tooltip, or some way to disable this behavior? This is affecting the experience in VO and NVDA
- Can we add an option for  the consumer to specify whether the tooltip is provided as 
`aria-describedby` text or `aria-labelledby` text? In some cases, the text provided by the tooltip would be more like a label instead of a description. 

Notes from screen reader review:
- In Voiceover, when I place focus on the tooltip toggle, focus is immediately shifted to the tooltip contents. This results in something like this being announced for the first example "I have a Lorem ipsum dolor..." Instead, the full text of the target should be announced before the contents of the tooltip, which should happen if focus stays on the target. 
- In JAWS, there is a difference in behavior depending on the method of navigation I use:
  - Using the Tab key, the tooltip works fine, but would be improved by not shifting focus to the tooltip (this results in an extra keypress)
  - Using the arrow key, the tooltip text isn't announced. `aria-describedby` text is ignored using Arrow key navigation (just because this is how JAWS works) and focus in JAWS doesn't trigger the tooltip to display either.
  - The Tab key would definitely be used when navigating forms. So anything with a tooltip referenced by `aria-describedby` would work. But there are other cases where a tooltip is serving as the label for something, like an icon that isn't obvious. In this case, the better attribute to use would be `aria-labelledby`, which is announced regardless of navigation method. 
- In NVDA, the experience is kind of a mix between VO and JAWS
  - Using the Tab key, the tooltip works fine. 
  - Using the arrow key, the tooltip displays, and the target and contents are announced in sequence. However, similar to VO, focus is placed on the tooltip, which overall is an unconventional experience. First the user's focus is shifted without them realizing it. Then proceeding to move forward, they're at the end of the DOM and have nowhere to go. Since they didn't initiate the change in focus and it happened automatically, it might not be obvious that they can return to their original context by hitting Esc.

jgiardino:
Thanks for opening these issues! (posting my comments from slack here for posterity ;-)

That’s a tough call. I feel like I’m doing a coin flip, but I’d argue that since 1.4.13 was added in WCAG 2.1, but not originally included in 2.0, and since this seems like an enhancement that could benefit some users, but is less likely to be a “broken” experience like the alternative is for screen reader users, then I would opt for not keeping the tooltip open as the user moves the mouse toward it.

PR #1111, opened 2019-01-01
fix(BackgroundImage): remove old tokens

boaz1337:
**What**:

`BackgroundImage` is using `c_background_image_BackgroundImage_<size>` tokens from `react-tokens` to set default values to the `src` props.

One of the tokens is `c_background_image_BackgroundImage_xs` which is no longer available from patternfly-next.

This patch fixes this by removing any token that is no longer available in patternfly-next.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
+++ b/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
@@ -7,13 +7,8 @@ import styles from '@patternfly/patternfly-next/components/BackgroundImage/backg
 /* eslint-disable camelcase */
 import {
   c_background_image_BackgroundImage_lg,
-  c_background_image_BackgroundImage_md,
-  c_background_image_BackgroundImage_md_2x,
   c_background_image_BackgroundImage_sm,
   c_background_image_BackgroundImage_sm_2x,
-  c_background_image_BackgroundImage_xl,
-  c_background_image_BackgroundImage_xs,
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
```diff
@@ -7,31 +7,22 @@ import styles from '@patternfly/patternfly-next/components/BackgroundImage/backg
...
-  xs: 'xs',
-  xs2x: 'xs2x',
   filter: 'filter'
 };
 
 const variableMap = {
   [BackgroundImageSrc.lg]: c_background_image_BackgroundImage_lg && c_background_image_BackgroundImage_lg.name,
   [BackgroundImageSrc.sm]: c_background_image_BackgroundImage_sm && c_background_image_BackgroundImage_sm.name,
   [BackgroundImageSrc.sm2x]: c_background_image_BackgroundImage_sm_2x && c_background_image_BackgroundImage_sm_2x.name,
-  [BackgroundImageSrc.xs]: c_background_image_BackgroundImage_xs && c_background_image_BackgroundImage_xs.name,
```
jschuler: Instead of xs and xs_2x looks like we need to replace with 
`c_background_image_BackgroundImage`
`c_background_image_BackgroundImage_2x`
And since these are smaller images, you probably want to keep the same xs and xs2x constants above

jschuler:
Thanks @boaz1337 , I've pushed this fix via #1124

PR #1114, opened 2019-01-02
fix(ListView): ListViews with One item should render it expanded

glekner:
ListViews with One item should render it expanded.
Since `ListViewItem` manages `expanded` in its state, I changed `ListView` without breaking changes.

**What**:

**Additional issues**:


glekner:
@jeff-phillips-18 @Laviro @sharvit can you review?

jeff-phillips-18:
I would rather see the ability for applications to specify if a ListViewItem is expanded so each particular application can decide if it wants to auto-expand when there is just one item or for whatever reason.

glekner:
- added `initExpanded` prop to `ListViewItem`. Now Items can be rendered expanded if needed. `ListView` reverted

Issue #1048, opened 2018-12-13
Review Components for consistent handling of required aria-labels


tlabaj:
Verify that all components treat required aria labels the same.  Some components generate an error on the JS side but have the prop optional on the TS side.  If we use isRequired on the JS side the prop should also be required in the TS file.

rebeccaalpert:
There are a ton of components that throw an error if there's no aria-label and some other condition applies:

```
if (!props.id && !props['aria-label']) {
    return new Error('TextArea requires either an id or aria-label to be specified');
}
```

Is it okay to leave those optional in the TS files?

tlabaj:
@rebeccaalpert , yes . I would leave both props as optional in this case.
@dlabaj an you weigh in on this please.

jschuler:
Can we implement something similar to what is shown in the examples here?
https://stackoverflow.com/questions/40510611/typescript-interface-require-one-of-two-properties-to-exist
```
type RequireAtLeastOne<T, Keys extends keyof T = keyof T> =
    Pick<T, Exclude<keyof T, Keys>>
    & {
        [K in Keys]-?: Required<Pick<T, K>>
    }[Keys]

interface MenuItem {
  title: string;
  component?: any;
  click?: any;
  icon: string;
}

type ClickOrComponent = RequireAtLeastOne<MenuItem, 'click' | 'component'>
```

PR #1116, opened 2019-01-02
fix(Radio.d.ts): Made aria-label required

rebeccaalpert:
Made aria-label required in Radio.d.ts to match the JS error handling. Per @titani, it should be consistent.

Fixes https://github.com/patternfly/patternfly-react/issues/1048.


jschuler:
Can you also update the prop in Radio.js? Right now it has a custom function but it can be replaced with
`'aria-label': PropTypes.string.isRequired`

tlabaj:
@rebeccaalpert @jschuler I did not see Joachim's comment until after I merged it.  Sorry.  Rebeca do you mind opening a separate PR to address that issue?

jschuler:
It's alright, probably not worth it to open a separate issue for it since it resolves to the same, just syntactic sugar

Issue #1094, opened 2018-12-19
Update typescript in react-core to support tsx in distribution


dlabaj:
The development environment for typescript code in react-core needs to be updated to support tsx files when distributing react-core.

dlabaj:
Fixed by #1117

PR #1117, opened 2019-01-02
Fix 1094: Updated to add support for generating d.ts files for ts, and tsx files.

dlabaj:
**What**:

Adds support for generating d.ts files from ts, and tsx files. (Issue #1094)

**Additional issues**:

N/A


dlabaj:
@jschuler Fixed the issue by renaming the tsconfig.json to tsconfig.dts.json

PR #1127, opened 2019-01-03
Updated drop down with icon only example

dlabaj:
**What**:

Add example to show how to create an icon only drop down.


dlabaj:
Address pf-next issue https://github.com/patternfly/patternfly-next/issues/332

packages/patternfly-4/react-core/src/components/Dropdown/examples/IconDropdown.js
```diff
@@ -0,0 +1,58 @@
+import React, { Component } from 'react';
+import { Dropdown, DropdownItem, DropdownToggle, DropdownSeparator } from '@patternfly/react-core';
+import { ThIcon } from '@patternfly/react-icons';
+
+export default class IconDropdown extends Component {
+  constructor(props) {
```
ibolton336:
```suggestion
state={
isOpen: false
}
``` 
We are using the class properties babel plugin now right? You should be able to get rid of the constructor syntax entirely. https://codeburst.io/use-class-properties-to-clean-up-your-classes-and-react-components-93185879f688

PR #1134, opened 2019-01-06
fix(MessageDialog): conflict between footer and primaryAction props

### LINKED ISSUE #1122: Message Dialog footer vs primaryAction and primaryActionButtonContent props
In `MessageDialog` you can render the Dialog footer by specifying `primaryAction` and `primaryActionButtonContent` or by passing a component directly to `footer`.

The problem is when using the latter `React` type checking prints an error message saying that `primaryAction` and `primaryActionButtonContent` are mandatory although they are redundant when using `footer`.

boaz0:
**What**:

fixes #1122 
`primaryAction` and `primaryActionButtonContent` are required unless `footer` is set.
However, currently an error message will be displayed if `footer` is set and not `primaryAction` or `primaryActionButtonContent`.

This commit fixes this conflict by checking if `footer` is set and if it does these props won't be required.


boaz0:
@michaelkro good catch!
I added unit tests to make sure this time I didn't miss anything.

Thanks a lot!

PR #1139, opened 2019-01-07
fix(AccessConsoles): fix propTypes checking for "null" child

mareklibra:
With this fix, AccessConsole's children of "undefined" or "null" values are handled
correctly by propTypes checking.

Needed for conditionally-rendered children within the higher-level calling code.
**What**: fix propTypes checking for "null" child

**Additional issues**:


mareklibra:
Fixed linter issue

PR #1141, opened 2019-01-07
Toolbar: added 'preventSubmit' prop. to allow prevention of internal form submission

### LINKED ISSUE #1104: PF3 Toolbar uses a `form` element in its toolbar, but doesn't do anything to prevent form submission on Enter key
The Toolbar uses a `<form />` but doesn't do anything to prevent form submission on Enter, which results in a screen refresh which isn't desirable in React/SPA.  We either need to:
- prevent form submission on "Enter" key by default, but wondering if that would break/affect existing code.  Would app devs be expecting form submission on Enter by default?  Currently no way for app devs to access the internal `<form >` Toolbar uses, no way to setup submitHandler or redirect to url.  
- add an `onSubmit` prop/callback which app devs can provide to allow prevention of form submission (`keyEvent.stopPropagation(); keyEvent.preventDefault();`)
- add a `preventSubmit` prop/boolean, default to false but can be set to true by app devs

https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/src/components/Toolbar/Toolbar.js#L22

dtaylor113:
**What**:
Toolbar has an internal `<form .../>` which will be submitted (causing a screen refresh) upon Enter key press.  This PR adds a `preventSubmit` prop. which when set to `true` will prevent form submission.  Default is `false`.

**Additional issues**:

Addresses OpenShift Issue: 
https://bugzilla.redhat.com/show_bug.cgi?id=1661141

Fixes #1104

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/Toolbar/Toolbar.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Toolbar/Toolbar.js
@@ -10,7 +10,7 @@ import ToolbarViewSelector from './ToolbarViewSelector';
 
 import { toolbarContextTypes, getToolbarContext, ToolbarContextProvider } from './ToolbarConstants';
 
-const ContextualToolbar = ({ children, className, ...props }) => {
+const ContextualToolbar = ({ children, className, preventSubmit, ...props }) => {
   const toolbarChildren = filterChildren(children, child => !hasDisplayName(child, ToolbarResults.displayName));
   const resultsChildren = filterChildren(children, child => hasDisplayName(child, ToolbarResults.displayName));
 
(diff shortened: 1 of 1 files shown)


packages/patternfly-3/patternfly-react/src/components/Toolbar/Toolbar.js
```diff
@@ -19,7 +19,16 @@ const ContextualToolbar = ({ children, className, ...props }) => {
       <Grid fluid className={className}>
         <Grid.Row className="toolbar-pf">
           <Grid.Col sm={12}>
-            <form className="toolbar-pf-actions">{toolbarChildren}</form>
+            <form
+              className="toolbar-pf-actions"
+              onSubmit={e => {
```
priley86: maybe just passing `onSubmit` function as a prop is a tad bit more flexible? or spreading `...props` on the  `form`... both alternatives. It looks like we also pass `preventDefault` as a prop on the `Nav` in PF4 so not opposed to this method at all.

jeff-phillips-18: I'm ok either way as well

dtaylor113: I'm leaning towards using `preventDefault` since there is an existing known pattern in `Nav` in PF4 -thanks

dtaylor113:
@priley86 If you are ok with this PR, pls merge, -thanks

PR #1143, opened 2019-01-07
chore(generator): update generator to support typescript files

### LINKED ISSUE #1045: Update PF4 component generator to default to typescript
Now that we support typescript, we should update the contribution docs & the component generators to support using typescript as the default. 

Also, the generator is currently broken as there is no longer an index.js to append the new component definition to. As a result, a new index.js file is being created alongside the existing index.ts file.

ibolton336:
**What**: fix #1045 

**Additional issues**:


scripts/generators/patternfly-4-component/templates/component/component.tsx.hbs
```diff
@@ -0,0 +1,24 @@
...
+  children: null,
+  className: ''
+};
+
+interface {{ componentName }}Props {
+  children?: any;
+  className ?: string;
+}
+
+const {{componentName}}: React.SFC <{{componentName}}Props> = (props) => (
```
dmiller9911: I actually just saw today on another project that `React.SFC` will be replaced by `React.FunctionComponent` in the future. Might be good to switch to that now. Also you might want to remove the space between `SFC` and `<` (`React.FunctionComponent<{{componentName}}Props>`) prettier will likely remove it after the fact, but it would be better to avoid needing the cleanup step.

dgutride: Looks good to me with the space removal - we should add an issue and make sure the FunctionComponent change above is good and make that change separately (to get this in and be consistent).

ibolton336:
Added an issue for the FunctionComponent changes https://github.com/patternfly/patternfly-react/issues/1149

scripts/generators/patternfly-4-component/templates/component/component.tsx.hbs
ibolton336: @dgutride updated the PR to remove the spaces.

priley86:
looks good to me @ibolton336 👍

Issue #926, opened 2018-11-15
PF4 Tabs


LHinson:
Add the PF4 Tabs to PF4-React. PF4-Core can be seen here: https://pf-next.com/components/Tabs/examples/

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @andybraren @mceledonia

tlabaj:
Closed by #1144

tlabaj:
Closed by #1144

Issue #1068, opened 2018-12-16
PF3: SocialLoginPage test warning


Laviro:
![selection_109](https://user-images.githubusercontent.com/26363699/50052287-48043400-012a-11e9-9f18-c42fb6442e09.png)

PR #1144, opened 2019-01-07
PF4: feat(Tabs) Add Tabs for PF4

tlabaj:
**What**:#926

**Additional issues**:


packages/patternfly-4/react-core/src/components/index.ts
```diff
@@ -26,8 +26,12 @@ export * from './Progress';
 export * from './Radio';
 export * from './Select';
 export * from './Switch';
+export * from './Tabs';
 export * from './Text';
 export * from './TextArea';
 export * from './TextInput';
 export * from './Title';
 export * from './Tooltip';
+export * from './TextArea';
```
rvsia: Duplicates here

andybraren:
I've been chatting off-site with Titani, the secondary tab styling fix is in progress (was missing the `.pf-m-tabs-secondary` modifier) and the scrollbar issue that Christie mentioned over in [Core's PR 1068](https://github.com/patternfly/patternfly-next/pull/1068#issuecomment-447378387) will be fixed once React pulls that in. HTML/CSS looks good to me once those are done. 👍

@mcarrano Regarding the presence of scrollbars, based on my note in [Core Issue 85](https://github.com/patternfly/patternfly-next/issues/859#issuecomment-434335331) I think we agreed at the end of our meeting to keep scrollbars and remove the original scrollbar-hiding hack to avoid fighting with browsers (primarily Firefox). Soon after that discussion Firefox made it possible to fully hide scrollbars (covered in [Core Issue 1103](https://github.com/patternfly/patternfly-next/issues/1103)) but doing so will still negatively impact users who explicitly ask for scrollbars to be shown in their browser's settings.

mcarrano:
I do recall that discussion now, @andybraren .  @mceledonia what do you think about the scroll bar?

andybraren:
The CSS is almost there - it just needs the refactoring and tweaks included in [Core's PR #1068](https://github.com/patternfly/patternfly-next/pull/1068) to fix things like the scroll button bug that Christie noted [here](https://github.com/patternfly/patternfly-next/pull/1068#issuecomment-447378387).

tlabaj:
@andybraren Workspace so it pulls in the updated css. Please try the latest preview.  Make sure to clear your cache first.

mcarrano:
Now I see what you are saying about the scroll breaking @andybraren . I have to admit that this is pretty ugly from an IxD perspective, especially when there are two rows of tabs. @mceledonia have you looked at this? Is there possibly another way to handle this?

The original intent was mainly to provide a solution for tabs on a mobile device. Has anyone tested it there? I wonder how well that works. In most desktop scenarios we should not see this unless there are an extreme number of tabs.

andybraren:
The original design linked at the top of [Core PR #868](https://github.com/patternfly/patternfly-next/pull/868) didn't include an example of scrolling secondary tabs. [Here's a video](https://www.youtube.com/watch?v=slJdbzfn5x4) of what they look like on iOS 12; better but not perfect.

If scrollbars on desktop are the main "ugly" part, that circles back to the discussion in [Core Issue #859](https://github.com/patternfly/patternfly-next/issues/859).

Also notice in the video above that two taps are required to "click" a tab - the first one just activates the hover state. This could be solved either via [CSS](https://css-tricks.com/annoying-mobile-double-tap-link-issue/) or [JS](http://cssmenumaker.com/blog/solving-the-double-tap-issue-on-ios-devices/). I'll bring this up within the CSS group to figure out which would be most appropriate.

tlabaj:
@mcarrano @andybraren I updated the code to remove the scroll buttons from the secondary tabs.

mcarrano:
Secondary tabs look good now. But why to tabs show with a gray background now?

screen shot 2019-01-18 at 5 20 59 pm

jgiardino:
I created issue https://github.com/patternfly/patternfly-next/issues/1127 to provide consistent labels for `<nav>` elements. For the `.pf-c-tabs` element, this should be customizable by the consumer to be either `<div class="pf-c-tabs">` or `<nav class="pf-c-tabs" aria-label="Local">`.

`<nav>` with `aria-label` would be used when there is a unique set of tabs on a page, so that screen reader users can find this navigation when browsing by landmark region.

Is this possible with this implementation? Is this something that should be captured in a separate issue?

andybraren:
@jgiardino That makes sense to me - I created [Core #1328](https://github.com/patternfly/patternfly-next/issues/1328) to track that change.

tlabaj:
@mcarrano @andybraren Please take another look.  The PR to updated to the version of core with the background fix went in this morning.

tlabaj:
@andy @jgiardino I added issue #1268 to track the accessibility issue.

packages/patternfly-4/react-core/src/components/Tabs/Tab.d.ts
```diff
@@ -0,0 +1,11 @@
+import { SFC, HTMLProps } from 'react';
+
+export interface TabProps extends HTMLProps<HTMLDivElement> {
+  children: any;
```
ibolton336: We don't need children as that is included in html props.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.d.ts
```diff
@@ -0,0 +1,16 @@
+import { SFC, HTMLProps, FormEvent } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface TabsProps extends Omit<HTMLProps<HTMLDivElement>, 'onSelect'> {
+  children: any;
```
ibolton336: children included in html props

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,221 @@
...
+  isSecondary: false,
+  leftScrollAriaLabel: 'Scroll left',
+  rightScrollAriaLabel: 'Scroll Right'
+};
+
+class Tabs extends React.Component {
+  static propTypes = propTypes;
+  static defaultProps = defaultProps;
+
+  constructor(props) {
```
ibolton336: Since this is a class component & we are using the babel class properties plugin, we don't need a constructor. You can add the createref right on the class like tablist= React.createRef()

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,221 @@
...
+    } = this.props;
+    return (
+      <React.Fragment>
+        <div
+          {...props}
+          className={css(
+            styles.tabs,
+            isFilled && styles.modifiers.fill,
+            isSecondary && styles.modifiers.tabsSecondary,
+            this.state.showLeftScrollButton && styles.modifiers.start,
```
ibolton336: can we destructure these state values like we do for the props?

packages/patternfly-4/react-core/src/components/Tabs/Tab.d.ts
tlabaj: I removed this.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
tlabaj: good idea.  thanks.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.d.ts
dlabaj: You should import this and add this to the typescript integration app since it's more complicated component.

PR #1145, opened 2019-01-07
fix(Page): improve a11y in sidebar and header

### LINKED ISSUE #1136: Page component - remove `<aside>` from nav section
A few updates are needed to improve accessibility of the Page component.

- The `<aside>` element should be changed to a generic `<div>`.
  Both `<aside>` and `<nav>` are [landmark regions](https://www.w3.org/TR/WCAG20-TECHS/ARIA11.html) that serve different purposes. The `<nav>` in the page component should not be wrapped in an `<aside>`, but instead should just be a generic `<div>` element.
- `aria-controls` should be added to `.pf-c-page__header-brand-toggle > .pf-c-button` 
  - This change requires an `id` added to `.pf-c-page__sidebar`, with the `aria-controls` value set to this `id` value
- `aria-expanded="true/false"` should be added to `.pf-c-page__header-brand-toggle > .pf-c-button`
- `aria-label` value for the element `.pf-c-page__header-brand-toggle > .pf-c-button` should be changed to "Toggle global navigation"

These updates with additional explanation are included in issues for core: 
- https://github.com/patternfly/patternfly-next/issues/1185
- https://github.com/patternfly/patternfly-next/issues/1184
- https://github.com/patternfly/patternfly-next/issues/1186
- https://github.com/patternfly/patternfly-next/issues/1127

boaz0:
**What**:

fixes #1136 

//cc @jgiardino

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Page/PageHeader.js
+++ b/packages/patternfly-4/react-core/src/components/Page/PageHeader.js
@@ -20,8 +20,12 @@ const propTypes = {
   topNav: PropTypes.node,
   /** True to show the nav toggle button (toggles side nav) */
   showNavToggle: PropTypes.bool,
+  /** True if the side nav is shown  */
+  isNavOpen: PropTypes.bool,
   /** Callback function to handle the side nav toggle button */
   onNavToggle: PropTypes.func,
+  /** Callback function to handle the side nav toggle button */
+  'aria-label': PropTypes.string,
   /** Additional props are spread to the container <header> */
   '': PropTypes.any
 };
@@ -34,10 +38,12 @@ const defaultProps = {
   avatar: null,
   topNav: null,
   showNavToggle: false,
-  onNavToggle: () => undefined
+  isNavOpen: true,
+  onNavToggle: () => undefined,
+  'aria-label': "Toggle global navigation",
 };
 
-const PageHeader = ({ className, logo, logoProps, toolbar, avatar, topNav, showNavToggle, onNavToggle, ...props }) => (
+const PageHeader = ({ className, logo, logoProps, toolbar, avatar, topNav, isNavOpen, showNavToggle, onNavToggle, 'aria-label': ariaLabel, ...props }) => (
   <header role="banner" className={css(styles.pageHeader, className)} {...props}>
     <div className={css(styles.pageHeaderBrand)}>
       {showNavToggle && (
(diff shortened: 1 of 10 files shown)


jgiardino:
Thanks for working on this, @boaz0! Everything looks good, except one thing. When I click the hamburger button to toggle the display of the navigation panel, the `aria-expanded` attribute doesn't get updated. Currently it only says `aria-expanded="true"` regardless of the display state of the nav panel, but it should be set to `"false"` if the nav panel is hidden.

boaz0:
@jgiardino I fixed it... thanks. :smiley:

jgiardino:
Hmm, I'm still seeing this issue. I'm looking at the runtime html using the preview link above. Could the preview be stale?

boaz0:
@jgiardino looking at [surge.sh](https://1145-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/page) it seems to work:

pf4 page

jgiardino:
Thanks for sharing that clip! I have been experiencing issues with the previews, so it's hard to tell if something is working as expected or not.

I am able to see the Page component work as expected, as shown in your example. 

But when I look at the [Page demo](https://1145-pr-patternfly-react-patternfly.surge.sh/patternfly-4/demos/pagelayout), I'm not seeing a change. But I'm also not sure if that's the correct expectation. I assumed that behavior is built into the Page component, and therefore would be visible wherever the Page component is being used, but I could be wrong about that . 

cc'ing @jschuler on this in case he has input on that expectation, or to review this in case it is working and it's just me having issues with my preview again 😣

boaz0:
Now I see why. I need to update the layout too.

boaz0:
@jgiardino fixed it -- can you take a look please?

jgiardino:
Thanks! The PageLayoutDemo Looks good now!

And I had previously checked the Page component, which looked good, but now I'm noticing the following error for all components. Not sure if that has anything to do with changes in this PR or not.

image

boaz0:
That's really odd. Running it locally I don't see these errors. However, surge.sh  still not being fixed even after doing `commit --amend` and `push -f origin fixes_1136`.

jschuler:
That was an issue with the workspace in master that recently went in, should be fixed now after rebase

packages/patternfly-4/react-core/src/components/Page/PageHeader.js
```diff
@@ -32,18 +34,21 @@ const defaultProps = {
...
+const PageHeader = ({ className, logo, logoProps, toolbar, avatar, topNav, isNavOpen, showNavToggle, onNavToggle, ...props }) => (
   <header role="banner" className={css(styles.pageHeader, className)} {...props}>
     <div className={css(styles.pageHeaderBrand)}>
       {showNavToggle && (
         <div className={css(styles.pageHeaderBrandToggle)}>
           <Button
             id="nav-toggle"
             onClick={onNavToggle}
-            aria-label="Toggle primary navigation"
+            aria-label="Toggle global navigation"
```
tlabaj: @jgiardino should this text be configurable for localization purposes?

boaz0:
@dlabaj I thought about using `React.cloneElement` but since we are passing `header` and `sidebar` component rendered I am not sure it's possible without breaking the current API.

Let me know if I am forgetting something or you believe we can do it in a different way.

Issue #1142, opened 2019-01-07
Need support for UMD and es2015 bundles


ssilvert:
Now that Keycloak's account management console has switched from Angular 2+ to React, we want to use the PatternFly React components.  However, our application must use SystemJs as a module loader instead of a pre-bundling solution like WebPack.  SystemJs requires UMD files, which most major libraries provide.

Dan Labrecque modified the PatternFly Angular build to create these files and it worked beautifully.  The issue is here:
https://github.com/patternfly/patternfly-ng/pull/373

Can you guys do the same for React?

dlabrecq:
Patternfly-NG is using Rollup to build ES and UMD bundles, while react-core uses Babel to generate ES.

That said, I was able to generate UMD by adding the following to the react-core package.json file.

```
"build:babel": "concurrently \"yarn build:babel:cjs\" \"yarn build:babel:esm\" \"yarn build:babel:umd\"",
"build:babel:umd": "cross-env BABEL_ENV=production:umd babel src --out-dir dist/umd --extensions \".css,.js,jsx,.ts,.tsx\"",
```

And in .babelrc.js, I added this:

```
'production:umd': {
      plugins: [
        'babel-plugin-typescript-to-proptypes',
        '@babel/plugin-proposal-export-default-from',
        '@babel/proposal-class-properties',
        '@babel/proposal-object-rest-spread',
        [
          '@patternfly/react-styles/babel',
          {
            srcDir: './src',
            outDir: './dist/umd',
            useModules: true
          }
        ]
      ]
    }
```

The package.json of react-core currently uses these properties. I don't have a SystemJS environment to test, so I cannot confirm if that will be an issue?

"main": "dist/js/index.js",
"module": "dist/esm/index.js",

Note that the react-charts and react-table packages would need to be updated as well.

ssilvert:
A single file would be quick and easy for getting started with development, but it would be wasteful in production.

If it's easy to do, can we publish both?  I see advantages to both approaches.

Would you like me to create a simple SystemJs app for testing?  Alternatively, I could just take what you produce in your branch and verify that it works on my current project.

priley86:
I ran into other issues w/ `rollup-plugin-babel` and importing react-styles. If anyone wants to experiment with this, my branch is [here](https://github.com/priley86/patternfly-react/tree/umd-support). 

[...]

PR #1150, opened 2019-01-09
feat(umd modules): adds babel-transform-es2015-modules-umd 

priley86:
adds babel-transform-es2015-modules-umd for umd module support

fixes: #1142 

**What**:
Docs for [transform-es2015-modules-umd](https://babeljs.io/docs/en/babel-plugin-transform-es2015-modules-umd)

This will convert es2015/esm modules to umd modules so that we can support umd module consumers.

In the future, we may consider rollup-plugin-babel or rollup-plugin-typescript for further support, but this should be the shortest path to support UMD modules today. The hope is to further support teams using RequireJS/CommonJS/SystemJS build systems.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/package.json
+++ b/packages/patternfly-4/react-core/package.json
@@ -39,10 +39,11 @@
     "react-dom": "^15.6.2 || ^16.4.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


ssilvert:
Has anyone been able to get the resulting umd files to work?  

If so, please post a working example.

PR #1152, opened 2019-01-09
fix(DropdownToggle): Listen for touch events on the document to close a Toggle on mobile

### LINKED ISSUE #1131: Dropdown component does not close on touch event outside of the focused area
When touching outside of the dropdown area on a mobile device, the dropdown does not close. The click event needs to support touch events & behave the same way it does on the desktop version.

mturley:
**What**:
Fixes #1131, correcting dropdown behavior on iOS devices.

**Additional issues**:
Also fixes a broken test that was never running assertions due to setTimeout.


dlabaj:
LGTM

PR #1169, opened 2019-01-11
fix(dropdown): add span wrapper for dropdown toggle text

ibolton336:
**What**:

adds this functionality from core: https://github.com/patternfly/patternfly-next/pull/971

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
@@ -7,7 +7,9 @@ import { css } from '@patternfly/react-styles';
 
 const DropdownToggle = ({ children, iconComponent: IconComponent, ...props }) => (
   <Toggle {...props}>
-    {children}
+    <span className={css(styles.dropdownToggleText)} >
+      {children}
+    </span>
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
```diff
@@ -7,7 +7,9 @@ import { css } from '@patternfly/react-styles';
 
 const DropdownToggle = ({ children, iconComponent: IconComponent, ...props }) => (
   <Toggle {...props}>
-    {children}
+    <span className={css(styles.dropdownToggleText)} >
```
rhamilto: It appears this addition broke the Dropdown - Icon Only case at http://patternfly-react.surge.sh/patternfly-4/components/dropdown.  

screen shot 2019-03-05 at 2 41 25 pm

After:
screen shot 2019-03-05 at 2 41 05 pm

`pf-c-dropdown__toggle-text` should only be added if `children` is text and not an icon.

cc: @spadgett, @janwright73, @priley86

priley86: yea i just saw this too...

PR #1178, opened 2019-01-14
fix(DualList): de-Select items after the transition between lists.

Laviro:
After items were moved between the lists,
they should not be selected as specified in the design.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/DualListSelector/helpers.js
+++ b/packages/patternfly-3/patternfly-react/src/components/DualListSelector/helpers.js
@@ -38,14 +38,28 @@ export const makeAllItemsVisible = list =>
 export const sortItems = (items, sortFactor = 'label') =>
   items.sort((a, b) => (a[sortFactor].toLowerCase() > b[sortFactor].toLowerCase() ? 1 : -1));
 
-export const arrangeArray = ({ items, sortBy, isSortAsc = true, isMainChecked = false }) => {
+export const shouldItemBeChecked = (item, isMainChecked, resetAllSelected) => {
+  let checked = item.checked || false;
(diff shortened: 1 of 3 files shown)


packages/patternfly-3/patternfly-react/src/components/DualListSelector/helpers.js
```diff
@@ -38,22 +38,34 @@ export const makeAllItemsVisible = list =>
 export const sortItems = (items, sortFactor = 'label') =>
   items.sort((a, b) => (a[sortFactor].toLowerCase() > b[sortFactor].toLowerCase() ? 1 : -1));
 
-export const arrangeArray = ({ items, sortBy, isSortAsc = true, isMainChecked = false }) => {
+export const shouldItemBeChecked = ({ item, resetAllSelected, isMainChecked }) => {
```
jeff-phillips-18: why take a object with three fields rather than 3 parameters?

Laviro: oops, copy-pasted it from another function signature with a lot of properties,
I don't mind changing it, 
btw from what number of arguments do we usually provide objects instead?

jeff-phillips-18: I'm not sure we ever should in helper functions

Laviro: no problem, opening a PR to refactor that.

Laviro: Added the [refactor PR](https://github.com/patternfly/patternfly-react/pull/1190) now

PR #1180, opened 2019-01-14
Chip component

### LINKED ISSUE #976: Chip

ibolton336:
**What**:
closes #976 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Chip/Chip.js
+++ b/packages/patternfly-4/react-core/src/components/Chip/Chip.js
@@ -1,0 +1,100 @@
+import React from 'react';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import ChipButton from './ChipButton';
+import { Tooltip, TooltipPosition } from '../Tooltip';
+import { TimesCircleIcon } from '@patternfly/react-icons';
+import styles from '@patternfly/patternfly-next/components/Chip/chip.css';
+import GenerateId from '../../internal/GenerateId/GenerateId';
+class Chip extends React.Component {
+
+  renderOverflowChip = () => {
+    const { children, className, onClick } = this.props;
+    return (
+      <div className={css(styles.chip, styles.modifiers.overflow, className)}>
+        <ChipButton onClick={onClick}>
+          <span className={css(styles.chipText)}>{children}</span>
+        </ChipButton>
+      </div>
+    );
+  }
+
+  renderChip = (randomId) => {
+    const {
+      children,
+      closeBtnAriaLabel,
+      tooltipPosition,
+      className,
+      onClick,
+    } = this.props;
+    const isTooltipVisible = children.length > 16;
+    if (isTooltipVisible) {
+      return (
+        <Tooltip position={tooltipPosition} content={children}>
+          <div className={css(styles.chip, className)}>
+            <span className={css(styles.chipText)} id={randomId}>
+              {children}
+            </span>
+            <ChipButton onClick={onClick} ariaLabel={closeBtnAriaLabel} id={`remove_${randomId}`} aria-labelledby={`remove_${randomId} ${randomId}`}>
+              <TimesCircleIcon aria-hidden="true" />
+            </ChipButton>
+          </div>
+        </Tooltip>
+      )
+    } else {
+      return (
+        <div className={css(styles.chip, className)}>
+          <span className={css(styles.chipText)} id={randomId}>
+            {children}
+          </span>
+          <ChipButton onClick={onClick} ariaLabel={closeBtnAriaLabel} id={`remove_${randomId}`} aria-labelledby={`remove_${randomId} ${randomId}`}>
+            <TimesCircleIcon aria-hidden="true" />
+          </ChipButton>
+        </div>
+      )
+    }
+  }
+
+  render() {
+    const {
+      isOverflowChip,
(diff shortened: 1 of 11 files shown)


packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,85 @@
...
+import { TimesCircleIcon } from '@patternfly/react-icons';
+import styles from '@patternfly/patternfly-next/components/Chip/chip.css';
+import GenerateId from '../../internal/GenerateId/GenerateId';
+
+class Chip extends React.Component {
+  renderOverflowChip = () => {
+    const { children, className, onClick } = this.props;
+    return (
+      <div className={css(styles.chip, styles.modifiers.overflow, className)}>
+        <ChipButton onClick={onClick} aria-label="Expand Chip">
```
jschuler: Don't need the aria-label here (from what i can see by looking at the Core docs)

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+  renderChip = (randomId) => {
+    const { children, position, className, onClick, minTooltipChars } = this.props;
+    const ChipComponent = (
+      <div className={css(styles.chip, className)}>
+        <span className={css(styles.chipText)} id={randomId}>
+          {children}
+        </span>
+        <ChipButton onClick={onClick} aria-label="Remove" id={`remove_${randomId}`} aria-labelledby={`remove_${randomId} ${randomId}`}>
```
jschuler: aria label should be customizable

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,85 @@
...
+  onClick: PropTypes.func,
+  /** Min character limit before showing tooltip. */
+  minTooltipChars: PropTypes.number
+};
+
+Chip.defaultProps = {
+  id: undefined,
+  className: '',
+  position: 'top',
+  isOverflow: false,
```
jschuler: nit: Would be clearer if this was called isOverflowChip

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,85 @@
...
+  /** Function that is called when clicking on the chip button */
+  onClick: PropTypes.func,
+  /** Min character limit before showing tooltip. */
+  minTooltipChars: PropTypes.number
+};
+
+Chip.defaultProps = {
+  id: undefined,
+  className: '',
+  position: 'top',
```
jschuler: Should perhaps be tooltipPosition

packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -0,0 +1,85 @@
...
+      <div className={css(styles.chip, className)}>
+        <span className={css(styles.chipText)} id={randomId}>
+          {children}
+        </span>
+        <ChipButton onClick={onClick} aria-label="Remove" id={`remove_${randomId}`} aria-labelledby={`remove_${randomId} ${randomId}`}>
+          <TimesCircleIcon aria-hidden="true" />
+        </ChipButton>
+      </div>
+    );
+    return children.length < minTooltipChars ? (
```
jschuler: Wondering, instead of using a char number, can you check if the span is overflowing (css limits max width)? Example here: https://stackoverflow.com/questions/42012130/how-to-detect-overflow-of-react-component-without-reactdom

packages/patternfly-4/react-core/src/components/Chip/Chip.js
ibolton336: Just had a convo about this with @tlabaj 

"I think what you have actually correct.  I think the example in core should actually have a aria-label.  PF-Next calls for one whenever the "plain" modifier is used for a button."

ibolton336: Should we change the button component to reflect what is in core? As it currently stands, there is a dependency on the aria-label if it is a plain button.

tlabaj: I talked with @jgiardino and @mcoker.  The pf-m-plain documentation will be updated in core.  That modifier can be used for buttons that contain text. If the button contains text (like the chip here), the aria label is not needed.  I have opened an issue #1184  to update the button component.

packages/patternfly-4/react-core/src/components/Chip/Chip.js
ibolton336: After discussing, I have decided to stick with the character approach for now. checking programmatically to see if the span was overflowing did work, but the result of the calculation would have to be used to toggle the display of the tooltip on hover or on some other event. There is no way to render it conditionally based on the actual render of the element itself. Another option @jschuler mentioned was to create a flag to toggle the tooltip visibility for the chip component.

PR #1188, opened 2019-01-15
fix(LoginPage): Submit is disabled when input is empty

MariaAga:
screencast-localhost-6006-2019 01 15-14-38-48

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js
@@ -94,7 +94,6 @@ const createProps = () => {
           href: '#',
           onClick: e => storyAction(e, 'Forgot password was clicked')
         },
-        disableSubmit: false,
         submitText: card.form.submitText,
         onSubmit: (e, onError) => {
           onError(card.form.error);
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
@@ -22,7 +22,7 @@ class LoginCardWithValidation extends React.Component {
     form: {
       showError: this.props.showError,
       submitError: this.props.submitError,
-      disableSubmit: this.props.disableSubmit,
+      disableSubmit: true,
       isSubmitting: this.props.isSubmitting
     }
   };
@@ -35,15 +35,30 @@ class LoginCardWithValidation extends React.Component {
     window.removeEventListener('keyup', this.toggleCapsLock);
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  static getDerivedStateFromProps(props, state) {
+    // disableSubmit prop will only be used in a not validated login card
+    const { validate, disableSubmit } = props;
+    if (validate) return null;
+    if (disableSubmit !== state.form.disableSubmit) {
+      return {
+        form: { disableSubmit }
+      };
+    }
+    return null;
+  }
+
   onInputChange = (e, inputType) => {
     this.props[inputType].onChange && this.props[inputType].onChange(e);
-    this.setState({
-      [inputType]: {
-        ...this.state[inputType],
-        value: e.target.value,
-        showError: false
-      }
-    });
+    const updateState = {
+      usernameField: { value: inputType === 'usernameField' ? e.target.value : this.state.usernameField.value },
+      passwordField: { value: inputType === 'passwordField' ? e.target.value : this.state.passwordField.value }
+    };
+    updateState[inputType].showError = false;
+    updateState.form = {
+      disableSubmit: updateState.usernameField.value.length < 1 || updateState.passwordField.value.length < 1
+    };
+    this.setState(updateState);
   };
 
   onInputFocus = (e, inputType) => {
@@ -273,6 +288,7 @@ LoginCardWithValidation.propTypes = {
   }),
   onSubmit: PropTypes.func,
   submitError: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  // eslint-disable-next-line react/no-unused-prop-types
   disableSubmit: PropTypes.bool,
   isSubmitting: PropTypes.bool,
   showError: PropTypes.bool
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.test.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.test.js
@@ -72,7 +72,6 @@ const createProps = () => {
           href: '#',
           onClick: mockFunction
         },
-        disableSubmit: false,
         submitText: card.form.submitText
       }
     }
@@ -177,6 +176,10 @@ test('Submit while inputs are empty cause specific errors to be shown and onChan
   const component = mount(<LoginPage {...createProps()} />);
   const usernameElement = component.find('input[type="email"]').at(0);
   const passwordElement = component.find('input[type="password"]').at(0);
+  const submit = component.find('button[type="submit"]').at(0);
+
+  expect(submit.props().disabled).toEqual(true);
+
   component
     .find('form')
     .at(0)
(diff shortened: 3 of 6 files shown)


packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
```diff
@@ -44,6 +44,11 @@ class LoginCardWithValidation extends React.Component {
         showError: false
       }
     });
+    this.setState(prevState => ({
+      form: {
+        disableSubmit: prevState.usernameField.value.length < 1 || prevState.passwordField.value.length < 1
+      }
+    }));
```
jeff-phillips-18: I think it should be up to the application to decide if the submit button is disabled or not. We have the correct properties setup but we are not updating state when the properties are updated. We should have a getDerivedStateFromProps method that updates the state if the props have changed. 

That said, the class currently does not use that pattern and updating the props currently has no effect on the component. I guess this change is consistent with that and can be accepted. Changing the entire class to update only if the application updates the props would likely break any existing uses of the component.

packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js
```diff
@@ -94,7 +94,7 @@ const createProps = () => {
           href: '#',
           onClick: e => storyAction(e, 'Forgot password was clicked')
         },
-        disableSubmit: false,
+        disableSubmit: true,
```
Laviro: maybe remove it from the props here,
as it would be true by default on Login Page With Validation ?

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
Laviro: I agree we can add `getDerivedStateFromProps` mechanism here anyway,
@jeff-phillips-18 you think it will really break things for us? 

most of the updates have been called by callbacks until now from components that are wrapped by this container,
and your suggestion adds the ability to update it from its parent, right?

We are still waiting on `patternfly-design` decision over this design change :)

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
```diff
@@ -44,6 +44,11 @@ class LoginCardWithValidation extends React.Component {
         showError: false
       }
     });
+    this.setState(prevState => ({
+      form: {
+        disableSubmit: prevState.usernameField.value.length < 1 || prevState.passwordField.value.length < 1
```
Laviro: could you use destructing here?

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
```diff
@@ -44,6 +44,11 @@ class LoginCardWithValidation extends React.Component {
         showError: false
       }
     });
+    this.setState(prevState => ({
```
Laviro: could you move this into the upper `setState` so it won't update the state twice?

Laviro: I am fine with that,
@jeff-phillips-18 @priley86 what is your opinion on that?

Laviro:
if this UX design will get accepted on patternfly/patternfly-design#671
we might want to remove the `handleOnEmptyInput` [code](https://github.com/patternfly/patternfly-react/blob/0a54b9b2777e82b478c678392812cc89b1c31a86/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js#L211)

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
jeff-phillips-18: If we update to rely on the parent to set the props where we were not before then it would be a breaking change since existing uses wouldn't update the props.

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
jeff-phillips-18: Would prefer not to setState twice, but I understand. Best I can come up with FWIW:

```
     const updateState = {
        usernameField: {value: inputType === 'usernameField' ? e.target.value : this.state.usernameField.value},
        passwordField: {value: inputType === 'passwordField' ? e.target.value : this.state.passwordField.value}
      };
      updateState[inputType].showError = false;
      updateState.form = {
        disableSubmit: updateState.usernameField.value.length < 1 || updateState.passwordField.value.length < 1
      };
      this.setState(updateState);
 ```

**Comment by MariaAga** [comment]
Since submitting on empty input is still technically possible I'm leaving the `handleOnEmptyInput` code in

File: packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js (hunk 4 above)
Comment (MariaAga) [reply packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js]: I've used this suggestion, thanks

File: packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js (hunk 2 above)
Comment (MariaAga) [reply packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js]: Removed and I've added `disableSubmit: true` to default props

**Comment by Laviro** [comment]
seems that in the un-managed LoginPage it blocks the user from submitting.
chrome-capture

File: packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js (hunk 1 above)
Comment (MariaAga) [reply packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js]: I've added `getDerivedStateFromProps`, it will change the state for non validated login card.
For validated login card the disable submit state will only update on input change.

File: packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js [hunk 5]
```diff
@@ -22,7 +22,7 @@ class LoginCardWithValidation extends React.Component {
     form: {
       showError: this.props.showError,
       submitError: this.props.submitError,
-      disableSubmit: this.props.disableSubmit,
+      disableSubmit: this.props.validate,
```
Laviro: Can you set this to `true`? as it would only be accessible if the component is validated

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
```diff
@@ -35,15 +35,29 @@ class LoginCardWithValidation extends React.Component {
     window.removeEventListener('keyup', this.toggleCapsLock);
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  static getDerivedStateFromProps(props, state) {
+    // disableSubmit prop will only be used in a not validated login card
+    if (props.validate) return null;
+    if (props.disableSubmit !== state.form.disableSubmit) {
```
Laviro: could you use destructing here?

I think we can update the state if any of the props is different, not just the `disableSubmit`.
You can use something like this: https://github.com/patternfly/patternfly-react/blob/59ca4e920ba7d6fd06420a02784f0c6bd959083f/packages/patternfly-3/patternfly-react/src/components/DualListSelector/DualListControlled.js#L16

MariaAga: I think updating the state if **any** of the props change is be outside the scope of this PR

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginCardWithValidation.js
MariaAga: yes, thanks

Issue #1191, opened 2019-01-15
PF4:  Toolbar Demo has an error


michaelkro:
Visit https://patternfly-react.surge.sh/patternfly-4/demos/toolbar/ and click on the dropdowns in the Toolbars.  The Toolbars will disappear, and we see an error appear beneath the example

```js
TypeError: this.props.context.sendRef is not a function
```

<img width="659" alt="patternfly_react" src="https://user-images.githubusercontent.com/15141412/51200143-b02f6c80-18c6-11e9-8cf0-7bc86fb83f4c.png">

dlabaj:
Looks like this more than the toolbar. There are errors on multiple pages.  The following error is shown in the console:

bootstrap 585e4cb142013b0929e5:52 Uncaught TypeError: Cannot read property 'call' of undefined
    at t (bootstrap 585e4cb142013b0929e5:52)
    at index.json?5e2e:9
    at loader.js:116
    at loader.js:97
    at index.json?5e2e:9
    at u (patch.js:19)
    at patch.js:40
    at window.webpackJsonp (bootstrap 585e4cb142013b0929e5:20)
    at VM340 path---index-a0e39f21c11f6a62c5ab.js:1

PR #1193, opened 2019-01-16
fix(Dropdown): Pass down ref for dropdown item

karelhala:
**What**:

This PR fixes problems when DropdownItem is React child, without it the ref wouldn't be assigned and was undefined when tried to access it in componentDidmount.

Fixes: #1191


packages/patternfly-4/react-core/src/demos/Toolbar/examples/ComplexToolbarDemo.js
```diff
@@ -69,11 +69,13 @@ class ComplexToolbarDemo extends React.Component {
         position={DropdownPosition.right}
         toggle={<DropdownToggle onToggle={this.onDropDownToggle}>All</DropdownToggle>}
         isOpen={isDropDownOpen}
+        dropdownItems={[
+          <DropdownItem>Item 1</DropdownItem>,
```
michaelkro: Since you're touching these lines already, can you add `key`s to the `DropdownItem`s?  The `Dropdown`s are working properly 🎉 but seeing the "unique key prop" warnings in the console.

karelhala:
@tlabaj yeah, I thought `yarn commit` added this to commit message, turns out it didn't. Updated.

PR #1200, opened 2019-01-16
feat(Alert): Add optional Close button to Alert via onClose prop

### LINKED ISSUE #1036: Alert component requires an optional Close button
Optional Close button is missing from current examples. A corresponding issue has been created to add this in Core. https://github.com/patternfly/patternfly-next/issues/1099

mturley:
**What**:

Fixes #1036.

In order to get the CSS for the close button, this PR upgrades `@patternfly/patternfly-next` from 1.0.120 to 1.0.126. This upgrade causes some inconsistencies with spacing in the Alert component, which I fixed by wrapping the alert description (children of `AlertBody`) in a `<p>` tag, which is consistent with the `alert-description.hbs` template in core (https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Alert/alert-description.hbs) and by changing the Action Button variant from "secondary" to "link", which is consistent with the changes to the core examples (https://github.com/patternfly/patternfly-next/pull/1208/files#diff-80f605f510a73d9a008b8bec47b6b75dR42). These fixes are in a separate commit to isolate them from the actual Close button feature changes.

With the CSS upgrade in place but without these fixes (db197e7716b98ec9a5a7bda7e40b2b96e7bc60de), the Alert component was rendering like this (note the line break before the description link, and the lack of margin to the right of the action button):

screenshot 2019-01-16 10 43 39

With the fixes in place (daecc00c28ea0c55344958c26ff7a05501c266c8), the Alert component renders like this:

screenshot 2019-01-16 14 30 01

The final commit e5c197a2172b9a7aa4e00a62ae8d416c4721be3d adds an optional `onClose` prop to the Alert component which, when defined, will cause a close button to be rendered.

**Additional issues**:

As noted by @andybraren in https://github.com/patternfly/patternfly-next/issues/1172, there is currently a lack of support for a secondary action button and a close button to both be rendered on an Alert component next to each other. Because this issue was closed and this design is being reworked, I did not attempt to support using both the `action` and the `onClose` props of Alert at the same time. If a user tries to do this, it will render like this:

screenshot 2019-01-16 14 24 42

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Alert/Alert.js
+++ b/packages/patternfly-4/react-core/src/components/Alert/Alert.js
@@ -30,6 +30,10 @@ const propTypes = {
   'aria-label': PropTypes.string,
   /** Variant label text for screen readers */
   variantLabel: PropTypes.string,
+  /** A callback for when the close button is clicked (if undefined, no close button is rendered) */
+  onClose: PropTypes.func,
+  /** Allows localization of the accessible label on the close button */
+  closeButtonAriaLabel: PropTypes.string,
   /** Additional props are spread to the container <div>  */
   '': PropTypes.any
 };
@@ -40,7 +44,9 @@ const defaultProps = {
   title: '',
   children: '',
   className: '',
-  variantLabel: null
+  variantLabel: null,
+  onClose: undefined,
+  closeButtonAriaLabel: 'Close'
 };
 
 const getDefaultAriaLabel = variant => `${capitalize(AlertVariant[variant])} Notification`;
@@ -53,6 +59,8 @@ const Alert = ({
   title,
   children,
   className,
+  onClose,
+  closeButtonAriaLabel,
   ...props
 }) => {
   variantLabel = variantLabel || capitalize(AlertVariant[variant]);
@@ -68,7 +76,9 @@ const Alert = ({
   return (
     <div {...props} className={customClassName} aria-label={ariaLabel}>
       <AlertIcon variant={variant} />
-      <AlertBody title={readerTitle}>{children}</AlertBody>
+      <AlertBody title={readerTitle} onClose={onClose} closeButtonAriaLabel={closeButtonAriaLabel}>
+        {children}
+      </AlertBody>
       {action && <AlertAction>{action}</AlertAction>}
     </div>
   );
--- a/packages/patternfly-4/react-core/src/components/Alert/AlertBody.js
+++ b/packages/patternfly-4/react-core/src/components/Alert/AlertBody.js
@@ -2,21 +2,32 @@ import React from 'react';
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import styles from '@patternfly/patternfly-next/components/Alert/alert.css';
+import { TimesIcon } from '@patternfly/react-icons';
+import { Button, ButtonVariant } from '../Button';
 
 const propTypes = {
   title: PropTypes.node,
   children: PropTypes.node,
-  className: PropTypes.string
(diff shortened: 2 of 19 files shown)


packages/patternfly-4/react-core/src/components/Alert/examples/WarningAlert.js
```diff
@@ -5,13 +5,14 @@ class WarningAlert extends React.Component {
   render() {
     return (
       <React.Fragment>
-        <Alert variant="warning" title="Warning notification title">
+        <Alert variant="warning" title="Warning notification title" onClose={() => {}}>
```
tlabaj: I wonder if we should change this example to a class and actually add the logic to close the alert.  Just a suggestion.   Don't know if we really need it though.

packages/patternfly-4/react-core/src/components/Alert/AlertBody.js
```diff
@@ -2,23 +2,32 @@ import React from 'react';
...
-  className: ''
+  className: '',
+  onClose: undefined
 };
 
-const AlertBody = ({ title, className, children, ...props }) => (
+const AlertBody = ({ title, className, children, onClose, ...props }) => (
   <div {...props} className={css(styles.alertBody, className)}>
+    {onClose && (
+      <Button variant={ButtonVariant.plain} onClick={onClose} aria-label="Close">
```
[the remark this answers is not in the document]
mturley: Sure thing.

packages/patternfly-4/react-core/src/components/Alert/examples/WarningAlert.js
mturley: I was considering that, but I didn't want to over-complicate the example... I think you're right though, it would be nice to have this. Do you think it should just be something like `state = { showAlertOne: true, showAlertTwo: true }` with `this.closeAlertOne` and `this.closeAlertTwo` methods? I was thinking of doing something that gives names to each alert and having one method to close by name, but that felt even more convoluted for a basic example. I wanted to have a close button on two of the variants to stay consistent with https://pf-next.com/components/Alert/examples/

packages/patternfly-4/react-core/src/components/Alert/Alert.js
```diff
@@ -30,6 +30,8 @@ const propTypes = {
   'aria-label': PropTypes.string,
   /** Variant label text for screen readers */
   variantLabel: PropTypes.string,
+  /** A callback for when the close button is clicked (if undefined, no close button is rendered) */
+  onClose: PropTypes.func,
```
dlabaj: Do we need to update the d.ts file?

mturley: Whoops, I thought I did, but I was editing the `dist` copy instead of `src`. Will fix.

packages/patternfly-4/react-core/src/components/Alert/AlertBody.js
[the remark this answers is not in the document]
mturley: @tlabaj this would need to be a prop on `Alert` called something like `closeButtonAriaLabel`, and drilled down to `AlertBody`, because the consumer won't render `AlertBody` directly. Does that sound good or would there be a better name for that prop?

mturley:
@tlabaj @dlabaj thanks for your reviews, I need to head out for the day but I'll push updates tomorrow morning.

mturley:
@tlabaj @dlabaj I made the changes you requested. @karelhala my new commits dismissed your approval, do you have time to take another look? Thanks!

Issue #1196, opened 2019-01-16
PF4 [DOCS]:  Dropdown documentation regarding React Router Link needs review


michaelkro:
In the Dropdown documentation, we currently recommend adding `<Link />` components as children directly under `<Dropdown />`.

> A react-router Link may be wrapped by DropdownItem or used directly within Dropdown.

Unfortunately, in this case, keyboard interactivity is lost, and if more than one `<Link />` is added, the consumer will be required to provide custom styles (not sure if this is expected or not)

***Screenshot***
![patternfly_seed](https://user-images.githubusercontent.com/15141412/51261568-7f127300-197e-11e9-8ae7-f59289e66ffa.png)

***Code***
```js
export default class App extends Component {
  state = {
    isOpen: false
  };
  onToggle = () => {
    this.setState(prevState => ({
      isOpen: !prevState.isOpen
    }));
  };
  render() {
    const { isOpen } = this.state;
    return (
      <div>
        <Dropdown
          toggle={
            <DropdownToggle isOpen={isOpen} onToggle={this.onToggle}>
              dude
            </DropdownToggle>
          }
          isOpen={isOpen}
        >
          <Link to="#">First Link</Link>
          <Link to="/">Second Link</Link>
        </Dropdown>
      </div>
    );
  }
}
```
***NOTE***:  Tested by spinning up a starter app with `patternfly-react-seed` and adding `react-router-dom`

***EDIT***:  Removed the wrapping `React.Fragment` after testing.  Results still seem to be the same.

kmcfaul:
Is the behavior the same without wrapping the Links in a React.Fragment? I remember mapping the dropdown item styling and keyboard events to custom child items so it might be that a wrapped block of Links isn't handled right now.

michaelkro:
whoops, good call.  Let me remove the fragment from my sample app and see.

michaelkro:
hmm, it doesn't seem to make a difference.  If you want, I can create a repo with my sample app for you to take a look at

kmcfaul:
I will try testing locally first. How about if you wrap the Link in a DropdownItem?

michaelkro:
If I wrap the Links in DropdownItems, I am seeing some console errors.  I believe that #1193 resolves the second

```
app.bundle.js:65294 Warning: Function components cannot be given refs. Attempts to access this ref will fail.

Check the render method of `Context.Consumer`.
    in Item (created by Context.Consumer)
    in div (created by Context.Consumer)
[...]

PR #1201, opened 2019-01-16
feat(Dropdown): handle keyboard navigation and styling for Link children

kmcfaul:
**What**: Update to custom children to handle keyboard inputs and styling.

**Fixes issue**: #1196


kmcfaul:
The way refs are currently assigned returns the Link component instance instead of the DOM node underneath, so suggestions are welcome if there's a better way to retrieve the node.

kmcfaul:
I was able to test in the docs locally by temporarily adding Link to the imports and editing the examples. I also tested that the Dropdown Panel example with div children still behaved as normal.

PR #1205, opened 2019-01-16
feat(Tooltip): Add optional delay prop to Tooltip

### LINKED ISSUE #1161: PF4: Add delay property to Tooltip
Tooltip component renders on mouse hover immediately.
As a web app user I'm used to tooltips only appearing after some mouse over time (half a second maybe?).
My suggestion is to add a `delay` property to the component that can be defaulted to `0` if you want to keep the current behavior.

PS: Please let me know If I need to add more information to the issue.

rebeccaalpert:
I added an optional delay prop to the Tooltip component to specify the entrance delay. The external tooltip library defaults to an entrance delay of 0 ms.

The null in {[delay, null]} gets the external tooltip library to use its default exit delay (20 ms). Controlling the exit delay wasn't mentioned in the issue, so I didn't add a prop for it. Let me know if you want an extra prop for that.

I didn't make any changes to the overrides; my IDE removed the spaces at the ends of the lines. I can set it back the way it was if you want. Let me know.

Fixes #1161.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.d.ts
@@ -20,6 +20,10 @@ export interface TooltipProps extends Omit<Props, 'content' | 'children'> {
   children: ReactElement<any>;
   /** Tooltip content */
   content: ReactElement<any> | string;
+  /** Delay in ms before a tooltip appears */
+  entryDelay?: number;
+  /** Delay in ms before a tooltip disappears */
+  exitDelay?: number;
   /** The element to append the tooltip to, defaults to body */
   appendTo?: Element | ((ref: Element) => Element);
   /** z-index of the tooltip */
--- a/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.js
+++ b/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.js
@@ -29,6 +29,10 @@ const propTypes = {
   content: PropTypes.node.isRequired,
   /** The reference element to which the tooltip is relatively placed to */
   children: PropTypes.element.isRequired,
+  /** Delay in ms before the tooltip appears */
+  entryDelay: PropTypes.number,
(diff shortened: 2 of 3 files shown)


TimoSta:
Nielson Norman has published [some pretty interesting guidelines on timing of exposing hidden content like tooltips](https://www.nngroup.com/articles/timing-exposing-content/). This research was one of the reasons this discussion originally popped up in our team.

---

They suggest both entry and exit delays in the 300 to 500 ms range:

> For hover interactions, the timing guidelines for speed of visual feedback and exposing hidden elements must be broken down into more steps:
> 1. Mouse cursor enters target area: display visual feedback within 0.1 seconds.
> 2. Wait 0.3–0.5 seconds.
> 3. If cursor remains stopped within target area, display corresponding hidden content within 0.1 seconds.
> 4. Keep displaying the exposed content element until the cursor has left the triggering target area or the exposed content for longer than 0.5 seconds.

---

Their main argument is **detecting intent**: Tooltips should not show up if the user just happens to move the cursor over the tigger element, but only if the user actually intends to activate it -- otherwise, the tooltips will flicker into view for a sub-second, which may be distracting to the user.

On the other hand, tooltips should not disappear instantly instantly when the user moves the cursor off the trigger element, to make the tooltip more resilient against brief exits and re-enters of the cursor, and to make the UX less 'hectic' -- arguably, the exit delay is more relevant e.g. for hover menus, but it seems still valid to me.

---

What I'm taking from this discussion are two points:

1. An **exit delay prop** may be a very useful addition, especially since it's already supported by tippy.js.
2. Maybe a value in the range of **300 to 500ms for both entry and exit delay is a sensible default**, rather than a default of 0ms.

---

I'm curious to hear what your thoughts are on this topic?

rebeccaalpert:
I added entryDelay and exitDelay props that default to 500 ms.

rebeccaalpert:
I rebased. It looks like https://github.com/patternfly/patternfly-react/pull/1244 has been merged. Can you please review?

Issue #1156, opened 2019-01-10
Add jest integration tests for PF4


dgutride:
Ensure after dist is created, we pull into sample app and use jest to validate all components can be loaded

Issue #1153, opened 2019-01-10
Test typescript package distribution as part of CI pipeline


dgutride:
We need to ensure the typescript package is consumable (this is probably a temporary step until we have full typescript definitions auto-generated by the build) - but still a blocking issue for many of our builds if it needs to be done manually.

rachael-phillips:
Are we ready to close this issue?

dlabaj:
Fixed in 1209

PR #1209, opened 2019-01-17
CI Integration tests using Cypress and demo app

dlabaj:
**What**:

Added Cypress for integration testing, and created a simple typescript app to run the integration tests against.

- Fixes: #1156 and #1153 
**Additional issues**:

Right now we only have one sanity test, we may want to add more integration tests as needed.


mturley:
@dlabaj are we sure `demo-app-ts` is importing the `@patternfly/react-core` version from the local dist bundle (so that it would fail if bad changes are made in a PR) and not the version from npm? I made a change that I assume should cause the demo app's build to fail, but the new tests pass.

```diff
diff --git a/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx b/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
index 4fbaa559..239b29d7 100644
--- a/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
+++ b/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
@@ -18,7 +18,7 @@ export interface AvatarProps extends DetailedHTMLProps<ImgHTMLAttributes<HTMLIma
   /** Attribute that specifies the URL of the image for the Avatar. */
   src?: string;
   /** Attribute that specifies the alt text of the image for the Avatar. */
-  alt: string;
+  alt: number;
 }
```

Am I right in assuming that should cause it to fail? Also, should we also add a component to the demo app that is written in JS rather than TS for this initial test, just to make sure it's consuming the type files properly?

mturley:
Oh, I see now that my change above would cause the initial `yarn build` to fail. I'm trying to figure out how we can verify that some change which doesn't cause our existing builds to fail will cause the integration tests to fail.

packages/patternfly-4/react-integration/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "homepage": "https://github.com/patternfly/patternfly-react#readme",
+  "scripts": {
+    "cypress:open": "cypress open",
+    "cypress:run": "cypress run",
+    "integration:setup": "rimraf node_modules && yarn install",
+    "integration:test": "node ./scripts/server.js"
+  },
+  "dependencies": {
+    "@patternfly/react-core": "file:../react-core",
```
priley86: for local e2e runs and development, this makes perfect sense to me. This would essentially test this PR or the latest code in this commit (also good). I would just hope that as a followup eventually, there is some way to run against a released version or tag, i.e. `"@patternfly/react-core" : "next"` to search for regressions we might be targeting before doing a "latest" release of the same code (to leave a consumer unaffected until all components load and pass e2e for the release). Maybe this could be done in a different app/deployment? Nice work! 👍

mturley: I didn't realize you could do this! (depend on `file:...` in package.json). That's pretty cool. @dlabaj you can ignore the first part of my comment above, I hadn't noticed this in my first pass.

dlabaj: @mturley No worries a lot of errors are caught just trying to build the demo app for example:

Example of app using dist unable to build:
*******************************************

$ react-scripts build
Creating an optimized production build...
Failed to compile.

./src/App.tsx
Attempted import error: 'Avatar' is not exported from '@patternfly/react-core'.


info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run build:demo stderr:
error Command failed with exit code 1.

lerna ERR! yarn run build:demo exited 1 in 'demo-app-ts'
error Command failed with exit code 1.

dlabaj: @mturley Others are caught when running the tests for example say a property is named wrong in the d.ts file.

dlabaj: @priley86 That's a good point about testing against other builds, wondering if we can over ride it.  Will look into it further.

mturley: Nice, thanks for the info @dlabaj. This looks great!

packages/patternfly-4/react-integration/README.md
```diff
@@ -0,0 +1 @@
+# react-integration
```
jschuler: Can we update on how to use?

Issue #1195, opened 2019-01-16
PF4 DropdownItem onClick broken with v1.45.2


dlabrecq:
I recently upgraded @patternfly/react-core from v1.44.2 to v1.45.2 and the PF4 DropdownItem onClick event stopped working. I'm no longer able to determine what item was selected.

Note that I'm providing my own onClick for the DropdownItem. However, the dropdown examples do not show how to get the selected drop down item.

Dropdown usage:
```
<Dropdown
  onClick={event => event.preventDefault()}
  onSelect={this.handleGroupBySelect}
  toggle={
    <DropdownToggle onToggle={this.handleGroupByToggle}>
      {t(`group_by.values.${groupById}`)}
   </DropdownToggle>
  }
  isOpen={isGroupByOpen}
  dropdownItems={groupByOptions.map(option => (
    <DropdownItem
      key={option.value}
      onClick={event =>
        this.handleGroupByItemClick(event, option.value)
      }
    >
      {t(`group_by.values.${option.label}`)}
    </DropdownItem>
  ))}
/>
```

Dropdown onClick handlers:
```
public handleGroupByItemClick = (event, groupBy) => {
    const { history, query } = this.props;
    const groupByKey: keyof OcpQuery['group_by'] = groupBy as any;
    const newQuery = {
      ...query,
      group_by: {
        [groupByKey]: '*',
      },
      order_by: { charge: 'desc' },
    };
    history.replace(this.getRouteForQuery(newQuery));
    this.setState({ selectedItems: [] });
  };

  public handleGroupBySelect = event => {
    this.setState({
      isGroupByOpen: !this.state.isGroupByOpen,
    });
  };

  public handleGroupByToggle = isGroupByOpen => {
    this.setState({
      isGroupByOpen,
    });
  };
```

rachael-phillips:
Are we ready to close this issue?

dlabrecq:
At a min, I'd like to confirm that the fix works in this scenario.

I'd also like to see the examples modified to show which item was clicked -- that functionality is missing from the demo.

PR #1211, opened 2019-01-17
fix(Dropdown): add onclick fix to custom items

kmcfaul:
**What**: Adding onclick fix to the updated custom item mapping.

**Part of issue**: #1195


kmcfaul:
Updated the top comment, it's part of #1195 because of the fix #1201 adding onClick to custom children.

PR #1212, opened 2019-01-17
feat(example): Add dark theme toggle button & change background color

### LINKED ISSUE #957: Examples should have Light/Dark Mode
To reflect what is in Core we should have a light/dark mode toggle in React. As of right now, it is only pulling in the light versions and as we progress we will have more examples of components that will exist on light and dark backgrounds.

ibolton336:
**What**:
closes #957 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-docs/src/components/example/liveDemo.js
+++ b/packages/patternfly-4/react-docs/src/components/example/liveDemo.js
@@ -12,6 +12,7 @@ import { LiveProvider, LiveEditor, LiveError, LivePreview, withLive } from 'reac
 import { transform } from 'babel-standalone';
 import Section from '../section';
 import copy from 'clipboard-copy';
+import classNames from 'classnames';
 
 const propTypes = {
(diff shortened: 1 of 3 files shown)


ibolton336:
I removed the background color after discussing with @mcoker . The gray color was throwing off the styles for multiple components. I added a border on the right of the left nav to make sure there still a visual separation between the nav and main. 
Old BG color:

screen shot 2019-01-17 at 9 10 24 am
screen shot 2019-01-17 at 9 11 34 am
screen shot 2019-01-17 at 9 32 30 am

new BG color:
screen shot 2019-01-17 at 1 07 42 pm

matthewcarleton:
@ibolton336 you are the man!

packages/patternfly-4/react-docs/src/layouts/index.css
```diff
@@ -0,0 +1,11 @@
+html,
```
jschuler: This file should be removed

PR #1222, opened 2019-01-21
fix(Dropdown): Do not focus Dropdown if menu is not opened

karelhala:
**What**:
If user clicks anywhere on screen and dropdown is present it will automatically scroll to such dropdown even if it's not opened.

**Additional issues**:
https://github.com/RedHatInsights/insights-frontend-components/issues/222

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Toggle.js
@@ -55,7 +55,7 @@ class DropdownToggle extends Component {
   };
 
   onDocClick = event => {
-    if (this.props.parentRef && !this.props.parentRef.contains(event.target)) {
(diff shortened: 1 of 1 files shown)


karelhala:
Honestly I think that this bug should have been fixed differently, but this works as a hot fix.

Proper fix would require adding domlisteners to DropdownMenu and triggering toggle from there because this menu is destroyed everytime we close it. Plus with this fix we would clean event listener (quite expensive to fire onToggle for multiple dropdowns).

jgiardino:
Thanks for submitting a bug fix for this! I wasn't able to reproduce this issue in the workspace, but also didn't see any issues with the dropdown after these changes.

karelhala:
@jgiardino you can reproduce it by clicking on any dropdown, scrolling a bit so this dropdown is not visible and clicking anywhere on screen.

jgiardino:
Ah, ok. Thanks for the clarification. I was able to see this issue. 
😄👍

PR #1227, opened 2019-01-22
feat(Table): support inline editing in PF4

suomiy:
Hi, I tried to tackle inline editing with reactabular. But it is still not perfect and a bit hacky.

My main goal was to prevent this behavior: https://github.com/kubevirt/web-ui-components/issues/9, but other issues are still present.

- inability to set a cursor to other edited cell (the cursor will firstly appear at the end and then it can be moved)
- no possibility for real time validation (char by char)

This is a due to the reactabular architecture, where we cannot expect our cells to behave in react-like way and have to implement workarounds introduced by this PR. It is inspired by https://github.com/reactabular/react-edit and uses onBlur and autoFocus on input elements

There is one issue with gatsby? where the whole table page is refreshed and the table component is called by costructor. This happens on a dropdown selection in `Editable table With Inline Edit Row`

live site: https://1227-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/inlineedit

TODO:
- [x] finish styles
- [x] move hardcoded styles to pf-next
- [x] move to separate package
- [x] add component extensibility
- [x] add typescript support 
- [x] tests

Current Issues: 
- table borders misbehave in firefox due to https://bugzilla.mozilla.org/show_bug.cgi?id=688556 (https://github.com/patternfly/patternfly-next/issues/1520 )
- gatsby refreshes on dropdown select,  but doesn't happen when the same component is rendered without gatsby - possibly issue with gatsby

 
@priley86 @karelhala


priley86:
Added issue [1228](https://github.com/patternfly/patternfly-react/issues/1228) for this.

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
```diff
@@ -1,23 +1,228 @@
...
+
+export const getEditStyles = (editConfirmationType, { isEditing, isTableEditing, isFirst, isLast }) => {
+  let editStyle;
+
+  if (isTableEditing) {
+    switch (editConfirmationType) {
+      case TableEditConfirmation.ROW:
+        if (isEditing) {
+          editStyle = {
+            background: '#def3ff',
```
priley86: can you replace all color references with actual color tokens from `@patternfly/react-tokens` (ex consumption [here](https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/react-tokens#each-token-as-three-properties) )?

you can also find these and their default mappings here:
http://patternfly-react.surge.sh/patternfly-4/styles/tokens/

If these colors don't exist, maybe we can subtitute an existing [pf-next color](https://github.com/patternfly/patternfly-next/blob/5cf999d5e13eab30303729eabf8c2e26eaedadf7/src/patternfly/sass-utilities/colors.scss)?

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -20,35 +21,69 @@ const defaultProps = {
...
   onRow = (row, props) => {
-    const { onRowClick } = this.props;
+    const { onRowClick, editConfig } = this.props;
     return ({
-      isExpanded: row.isExpanded,
-      isOpen: row.isOpen,
-      onClick: (event) => onRowClick(event, row, props)
+      row,
+      rowProps: props,
+      onMouseDown: (event) => {
```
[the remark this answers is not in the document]
priley86: this seems like kind of a hack, but works. I wonder if proposing an upstream PR to reactabular in parallel to allow hooks into `onChange`\`onBlur` would be a good idea?

dgutride:
This seems like a great idea - we are avoiding adding anything to react-core or the core packages (which include react-table) that do not have backing PatternFly patterns and documentation.  Would it be possible to move this to a new @ patternfly / {name of packaqge} - extension package so we can consider including it into core PatternFly at a later date?

packages/patternfly-4/react-table/internal/util.js
```diff
@@ -0,0 +1,8 @@
+// TODO: better sharing util components between modules
```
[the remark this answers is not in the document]
priley86: added #1232 for this. If it merges you can:

```
import { debounce } from '@patternfly/react-core';
```

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -2,13 +2,14 @@ import React from 'react';
 import { Body } from 'reactabular-table';
 import PropTypes from 'prop-types';
 import { TableContext } from './Table';
+import '../../../internal/closestPolyfill';
```
dmiller9911: This is a `sideEffect` and will break webpack 4 (well not be imported in production)., It is also generally a bad idea to pollute prototypes for the user. This should be an opt-in pollyfill by the consumer. We can still export it and if they want/need it they can use it. Also since by default PF4 is excluding IE support (a workaround is coming for consumers needing it) the closest polyfill is not needed.

If we do for sure want to keep this in the lib it would be better to do something like: 
```js
function closest(element, selector) {
  if (typeof element.closest === 'function') {
    return element.closest(selector);
  }

  if (document.documentElement.contains(el)) {
    while (el && el.nodeType === 1) {
      if (el.matches(selector)) return el;
      el = el.parentElement || el.parentNode;
    }
  }
  return null;
}

File: packages/patternfly-4/react-table/internal/closestPolyfill.js [hunk 5]
```diff
@@ -0,0 +1,23 @@
+export const matches =
+  Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
+
+if (!Element.prototype.matches) {
```
dmiller9911: Not needed at all. This is supported by all major browsers including IE9+

packages/patternfly-4/react-table/src/components/Table/Body.js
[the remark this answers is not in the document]
suomiy: Where would the hooks be? I am not sure something like this would help. 
The idea of this is to get ahead of cell onBlur callbacks. Because of the cell's onBlur callback, the cell will be scrapped and then constructed again. And then we get a chance to fire callbacks (captured here) on our new instance.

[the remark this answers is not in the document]
suomiy: it is a hack nevertheless :)

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
suomiy: yes, I am planning to do the css  cleanup

suomiy:
> This seems like a great idea; however, we are avoiding adding anything to react-core or the core packages (which include react-table) that do not have backing PatternFly patterns and documentation. Would it be possible to move this to a new @ patternfly / {name of packaqge} - extension package so we can consider including it into core PatternFly at a later date?

Sorry for late response because of the PTO.  

I get the idea of putting bigger features to separate packages. 
Nevertheless, I think inline editing is quite integral part of the table and it would be less complicated for the user to not use multiple packages. It is possible, but it would be quite cumbersome and not quite worth it in my opinion.

Btw, do you mean patternfly next patterns or https://www.patternfly.org/pattern-library/forms-and-controls/inline-edit/ ?

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -20,35 +21,69 @@ const defaultProps = {
...
+        }
+
+        // give priority to fire onChange/onBlur callbacks
+
+        setTimeout(() => {
+          if (!row.isEditing) {
+            onRowClick(event, row, props);
+            if (onEditCellChanged) {
+              // edit cell after rerender
+              setTimeout(onEditCellChanged, 0);
```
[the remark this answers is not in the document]
suomiy: In that case we would really have to contribute to reactabular and fix these issues. Yup it is hacky, but there are not many other solutions.

packages/patternfly-4/react-table/src/components/Table/utils/formatters/inlineEditFormatterFactory.js
```diff
@@ -0,0 +1,26 @@
+const inlineEditFormatterFactory = ({ renderEdit, renderValue, resolveValue, isEditable = null } = {}) => (
+  value,
+  additionalData
+) => {
+  const { rowData, columnIndex } = additionalData;
+
+  let result;
```
karelhala: No need to assign to result if we are only returning it. You can add return instead of assign to each if statements and it wil be the same.

karelhala:
@suomiy what @dgutride is proposing is kinda understandable. We want to ship table component as pure as possible with only few enhancers and if consumers want to use inline edit they'd import `@patternfly/react-table-inline-edit` for instance. Inside of this package there would be few functions and components reponsible for edits. The code would still be part of this repository, it would just be moved into special folder and a bit of config is needed. If you want I can help you by moving it there or you can get inspired by @priley86 https://github.com/patternfly/patternfly-react/pull/1250 about virtualized scroll.

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
```diff
@@ -1,24 +1,228 @@
...
+  getConfirmationButtons() {
+    const { row, rowProps, ...props } = this.props;
+    const { isOpen, isExpanded, isLast, isFirst, isEditing, isTableEditing, editConfig } = row;
+
+    if (!editConfig) {
+      return null;
+    }
+    const { editConfirmationType, onEditConfirmed, onEditCanceled } = editConfig;
+
+    let confirmButtons;
```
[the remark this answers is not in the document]
suomiy: I would rather remove it too, but I kept that because of eslint/prettier. I am not sure how do we stand on that?

packages/patternfly-4/react-table/src/components/Table/utils/formatters/inlineEditFormatterFactory.js
suomiy: you are right, thanks

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
```diff
@@ -1,24 +1,228 @@
...
+import { TableEditConfirmation } from './Table';
 
-const RowWrapper = ({ isOpen, isExpanded, ...props }) => (
-  <tr
-    {...props}
-    className={css(isExpanded !== undefined && tableExpandableRow, isExpanded && modifiers.expanded)}
-    hidden={isExpanded !== undefined && !isExpanded}
-  />
-);
+const hasConfirmationButtons = (editConfirmationType, { isEditing, isTableEditing, isFirst, isLast }) => {
```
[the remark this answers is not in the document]
suomiy: you are right, it looks much cleaner now

suomiy:
> @suomiy what @dgutride is proposing is kinda understandable. We want to ship table component as pure as possible with only few enhancers and if consumers want to use inline edit they'd import @patternfly/react-table-inline-edit for instance. Inside of this package there would be few functions and components reponsible for edits. The code would still be part of this repository, it would just be moved into special folder and a bit of config is needed. If you want I can help you by moving it there or you can get inspired by @priley86 #1250 about virtualized scroll.

Ok, I guess it makes sense to separate it. I will look into that. 
Nevertheless I think there will still be some overlap of the code. I will discuss this with @priley86.

karelhala:
@suomiy if you need some special variables to be passed to core implementation that is fine, that code won't be used. But if we split this into another package we have really great way of importing custom packages and promote other teams to contribute and use some other packages (even written outside of PF world)

suomiy:
- Added extensibility:
  - moved inline editing to `react-table-inline-editing` package
  - RowWrapper and Body can be chained and extended using the HOC mechanism

- submitted PR to pf-next for the css: https://github.com/patternfly/patternfly-next/pull/1428

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -19,14 +19,14 @@ const defaultProps = {
...
     const { onRowClick } = this.props;
-    return ({
-      isExpanded: row.isExpanded,
-      isOpen: row.isOpen,
-      onClick: (event) => onRowClick(event, row, props)
-    });
-  }
+    return {
+      row,
+      rowProps,
```
priley86: +1 for passing row props back to consumer

priley86:
this seems like a fairly safe extension to me at this point that we could begin consuming and iterating on downstream. It also helps address #1310. Any concerns on this one @dgutride @janwright73 ?

priley86:
needs a rebase ;)

suomiy:
rebased

priley86:
i think we should rename this package as an extension package - maybe `react-inline-edit-extension`?  Noting the current [CONTRIBUTING.md](https://github.com/patternfly/patternfly-react/blob/master/CONTRIBUTING.md#creating-issues-for-new-components) recommendation on this.

suomiy:
renamed to `react-inline-edit-extension` and added a note to README

suomiy:
The css were moved from pf-next to here. So the preview is now buildable and is up and running at https://1227-pr-patternfly-react-patternfly.surge.sh. 

sadly the css don't work correctly in firefox due to this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=688556

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/index.js
```diff
@@ -0,0 +1,4 @@
+export { TableEditConfirmation } from './constants';
+export { default as editableTableBody } from './BodyHOC';
+export { default as editableRowWrapper } from './RowWrapperHOC';
```
suomiy: I added HOC suffix to the Body and the RowWrapper. 

Errors were thrown by gatsby ( gatsby-transformer-react-docgen -> react-dcogen) when leaving just the RowWrapper. It may be a collision with the other RowWrapper or just that react-docgen doesn't really like HOC. I tried many things (e.g. custom resolver) and nothing worked. 

Just renaming is a simple solution.

other options could be
- rowWrapperHOC
- editableRowWrapper
- EditableRowWrapper

not sure which one is the best

priley86: totally fine w/ "editableRowWrapper" if others are...

priley86:
wonderful job w/ this extension @suomiy 🥇 

I think it would be great to get one more review from @dgutride and @karelhala if they have bandwidth. This PR has grown a lot and it's getting hard to track, but most of it is contained to the inline-edit extension (and you've now encapsulated the CSS there). Are you mostly ready after the CSS issue above is resolved? 

You have done a terrific job kickstarting this feature. We can obviously revisit this w/ the wider PF team later on, but I'm OK to see it merge if others are.

packages/patternfly-4/react-docs/src/components/example/liveDemo.js
```diff
@@ -44,7 +46,7 @@ const scopePlayground = {
 const transformCode = code => {
   try {
     // LiveEditor doesn't work properly with these so need to remove
-    code = code.replace(/^import(.|\s)*?;$/gm, '');
+    code = code.replace(/^import(.|\s)*?;$/gm, '');1
```
priley86: is the "1" just a random typo here? it seems to be ;)

suomiy: thanks :)

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/index.js
suomiy: ok, renamed to "editableRowWrapper"

suomiy:
> You have done a terrific job kickstarting this feature. We can obviously revisit this w/ the wider PF team later on, but I'm OK to see it merge if others are.

thanks :)

> Are you mostly ready after the CSS issue above is resolved?

I do not think it can be resolved easily. This buggy behaviour (makes the borders not rendered correctly in firefox) is already present in `@patternfly/react-table` . The boarders in https://pf4.patternfly.org/components/Table/examples/ are rendered correctly because there are only two rows.

Yes, the implementation part should be ready. I just need to include the ts types and finish the tests.

suomiy:
added typescript

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -19,15 +19,17 @@ const defaultProps = {
...
-      isExpanded: row.isExpanded,
-      isOpen: row.isOpen,
-      onClick: event => {
-        if (event.target.tagName !== 'INPUT' && event.target.tagName !== 'BUTTON') {
-          onRowClick(event, row, props);
-        }
+      row,
+      rowProps,
+      onMouseDown: event => {
+        const computedData = {
```
suomiy: inline edit needs input events. I can send back computedData which are a bit easier to consume and leave the decision logic on consumer. IMO this is the best what can be done. Does somebody have a better idea?

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/editableTableBody.d.ts
```diff
@@ -0,0 +1,36 @@
+import { Component, FunctionComponent, MouseEvent } from 'react';
+import { TableBodyProps } from '@patternfly/react-table';
+
+import { OneOf } from '../../../../react-core/src/typeUtils';
```
suomiy: is it really ok to reference the files in different module? The same thing occurs in `Table.d.ts`

priley86: my preference would be to export these in `react-core` like we did in #1232. Seems they are helpful in other packages too.

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -19,15 +19,17 @@ const defaultProps = {
 };
 
 class ContextBody extends React.Component {
-  onRow = (row, props) => {
+  onRow = (row, rowProps) => {
     const { onRowClick } = this.props;
     return {
-      isExpanded: row.isExpanded,
```
karelhala: Do not remove `isExpanded` and `isOpen` props. They are needed for correct rendering of collapsible tables

suomiy: The RowWrapper interface was changed accordingly and it allows for more rich props like the ones used in `inline-edit` package. E.g. 

```js
const propTypes = {
  ...ReactTableRowWrapper.propTypes,
  trRef: PropTypes.func,
  className: PropTypes.string,
  onScroll: PropTypes.func,
  onResize: PropTypes.func,
  row: PropTypes.shape({
    isOpen: PropTypes.bool,
    isExpanded: PropTypes.bool,
    isEditing: PropTypes.bool,
    isTableEditing: PropTypes.bool,
    isLast: PropTypes.bool,
    isFirst: PropTypes.bool,
    editConfig: PropTypes.object
  }),
  rowProps: PropTypes.object
};
```

The `isOpen` and `isExpanded` are still passed and used. Only `isExpanded` prop is actually used though, because `isOpen` was not used before and is just passed to the RowWrapper. 

`isOpen` is used correctly (taken from rowData) in the CollapseColumn

suomiy: to conclude, I am not aware of any breaking changes

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/examples/EditableTable.js
```diff
@@ -0,0 +1,262 @@
...
+  TableEditConfirmation,
+  TableTextInput
+} from '@patternfly/react-inline-edit-extension';
+import { Dropdown, DropdownToggle, DropdownItem, Checkbox } from '@patternfly/react-core';
+
+class EditableTable extends React.Component {
+  constructor(props) {
+    super(props);
+
+    const textInputFormatter = inlineEditFormatterFactory({
```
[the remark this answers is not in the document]
suomiy: added comments

packages/patternfly-4/react-table/src/components/Table/Body.js
suomiy: ok, you are right, I didn't try to combine them - something breaks it

karelhala: Oh, okay. I tried to set it up and it does allow collapsible with inline edit. Good job!

karelhala: When looking at the collapsed column it does not cover this case. You are unable to edit such row. Honestly I don't think it's really required and can be left as future improvement.

suomiy: yes, I will try to solve it somehow (if I can in reasonable timeframe) so at least that use case is passable somehow.

I will also have to add some property to indicate that the editing includes the children and show the css and buttons accordingly. We are looking at this exact use case which will be needed in the future

suomiy:
> I also experienced an issue when scrolling page the buttons stayed for scroll time at the same place, I think that is caused by fixed and calculating the position of these buttons. Is there any way how not to use fixed position and use relative instead? The calculaction would be slightly easier as well (could be just calculated based on number of rows and their height)

can we leave this to followup as this is a styling issue?

karelhala:
> can we leave this to a followup as this is a styling issue?

Sure no prob on my side, just wanted to point this out.

suomiy:
Added support for:
- collapsible rows
- editing of expanded rows.  I had to change interface of expanded rows a bit in `@patternfly/react-table` to support that (please see usage in `CollapsibleTable` and/or alternative one in `CollapsibleEditableTable`) 
- multiple inputs inside a cell (important for expanded row). To allow that I changed the way editing works. Cell clicks are identified with input ids (together with rowId & columnId). So to support inline editing, all editable fields must provide an id (shows warnings if missing). Please let me know if you have a different idea or I should use data attributes (possibly not feasible in vendor components )?  

Can you please rereview ? I added the changes into a separate commit for better readability. I can squash them before merge if needed.

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/editableTableBody.js
```diff
@@ -0,0 +1,142 @@
...
+        ...rowProps,
+        columnIndex: cellNumber,
+        elementId
+      });
+    };
+  }
+
+  // give priority to fire onChange/onBlur callbacks
+
+  setTimeout(() => {
```
suomiy: I decided to stop checking if the onEdit callback should be fired. Consequently, I renamed `onEditCellChanged` to  `onEditCellClicked`. The checking is up to the consumer and the usage is more permissive.

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/examples/EditableTable.js
```diff
@@ -0,0 +1,308 @@
...
+        activeEditId: null // stop autoFocus
+      };
+    });
+  };
+
+  onEditCellClicked = (event, clickedRow, { rowIndex, columnIndex, elementId }) => {
+    const WORKSPACE_COL = 3;
+    const PRIVATE_REPO_COL = 5;
+    const ACTIONS_COL = 6;
+
```
suomiy: the consumer must supply the logic what to do with the click and can work with the ids

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/utils/utils.js
```diff
@@ -0,0 +1,26 @@
+const requiredElements = ['INPUT', 'BUTTON'];
+let warningCompleted = false;
+
```
suomiy: not sure about this one. I think it might be useful to see a warning when using inline editing without ids

suomiy: nevertheless it is technically possible to use it without ids if there is only one edit component per cell. You just have to pass it to the formatter by yourself.

suomiy: There are problems with that last approach though - you receive a click on td padding as well..

packages/patternfly-4/react-table/src/components/Table/utils/decorators/collapsible.js
```diff
@@ -46,7 +51,7 @@ export const expandedRow = colSpan => {
   ) =>
     rowData.hasOwnProperty('parent') && {
       colSpan,
-      children: <ExpandableRowContent id={contentId + rowIndex}>{value.title || value}</ExpandableRowContent>
```
karelhala: This is not good change :disappointed: We can't change curren API of table and forcing consumers to update it will be bad practice. Do not add edit to children as of yet, just render them and disable editing of them. We can update the whole child render in follwup PR to make it easier.

suomiy: This change is easily fixable by just adding this line to first column

```js
    cellFormatters: [expandable] 
```

This shouldn't be such a big problem for users to adapt to it.

We really need this functionality in cnv and there is no possibility to render custom components in expandable row now.

suomiy: > We can update the whole child render in follwup PR to make it easier.

Can we discuss it in this PR instead?  this functionality is an important part of  `CollapsibleEditableTable` example.

suomiy:
deployed at https://1227-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/inlineedit

suomiy:
> We can update the whole child render in follwup PR to make it easier.

I split the functionality into this PR: https://github.com/patternfly/patternfly-react/pull/1520. We can go through with this one once we are satisfied with the expandable rows implementation.

packages/patternfly-4/react-table/src/components/Table/utils/decorators/collapsible.js
suomiy: You are right, it makes it easier to have the changes separately. Let's resolve https://github.com/patternfly/patternfly-react/pull/1520 first

packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/editableTableBody.d.ts
suomiy: done, can you please take a look at https://github.com/patternfly/patternfly-react/pull/1521?

suomiy:
This PR should be ready for final reviews. I added the tests and they have just passed.

packages/patternfly-4/react-inline-edit-extension/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "reacttabular"
+  ],
+  "author": "Red Hat",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/patternfly/patternfly-react/issues"
+  },
+  "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/",
+  "dependencies": {
+    "@patternfly/patternfly": "1.0.184",
```
jschuler: @redallen a new package coming down the line, not sure if the PR bot automatically picks these up just FYI

jschuler:
I've removed the Do not merge label since it's in its own package now @dgutride 
@priley86 @karelhala please review again :)

suomiy:
Fixed TS and added support for rendering `NO_CONFIRM_ROW` type.

This PR is ready for final reviews (sibling https://github.com/patternfly/patternfly-react/pull/1520 was merged).

priley86:
great job on this @suomiy! I know you are eager to iterate on this extension downstream. I filed my follow-up issue just now in [#1628](https://github.com/patternfly/patternfly-react/issues/1628). I am happy to see this merge when others are.

priley86:
getting these here locally now...can we fix them first?
```
@patternfly/react-docs: error There was a problem parsing component metadata for file: "components/InlineEdit/editableRowWrapper.js"
@patternfly/react-docs: 
@patternfly/react-docs:   TypeError: Property name must be an Identifier or a Literal
@patternfly/react-docs:   
@patternfly/react-docs:   - getPropertyName.js:36 getPropertyName
```

suomiy:
> getting these here locally now...can we fix them first?

fixed, seems like we can't have dynamic object keys (https://github.com/reactjs/react-docgen/issues/317#issue-393678795) 

\+ added a comment about this issue to the code as the constants are now hardcoded

packages/patternfly-4/react-inline-edit-extension/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "reacttabular"
+  ],
+  "author": "Red Hat",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/patternfly/patternfly-react/issues"
+  },
+  "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/",
+  "dependencies": {
+    "@patternfly/patternfly": "1.0.240",
```
priley86: needs updating now (just noticed this by doing `yarn install` and seeing yarn lock diff):
`"@patternfly/patternfly": "1.0.244",`

suomiy:
updated `@patternfly/patternfly` dependency

PR #1237, opened 2019-01-24
fix(Table): Use direct imports instead of spread

karelhala:
**What**:
* When using react table there is tons of warnings due to missing exports in patternfly next styles. That is caused by using spread instead of default import. It is much safer (but a bit less readable) to use direct imports.

* Allow first row to be collapsible by checking undefined instead of casting Boolean.

* Do not show select in collapsed child unless show checkbox is engorced on row.


dmiller9911:
Fyi the reason for this is that the styles plugin only exports a default object. This might be a good candidate for a lint rule since it has happened before.

Issue #1189, opened 2019-01-15
PF4: Popover and Tooltip need an extra CSS import


msglifeIberia:
I followed the documentation and imported the `@patternfly/react-core/dist/styles/base.css` file in my app.
Later I used a `<Popover>` in my app and noticed it didn't behave like it does in [your storybook](https://patternfly-react.surge.sh/patternfly-4/components/popover). 
The main issues were the positioning that was slightly off and the lack of the fading effect when showing the popover.
Inspecting the DOM and the CSS imports I found out I was missing this CSS file import:
`tippy.js/dist/tippy.css`

I wasn't able to find any reference to this import in your documentation.
This import should be done within the `base.css` or the documentation should have a note saying that this import is needed if one wants to use the `<Popover>` or the `<Tooltip>` components (and maybe other components that use `tippy.js` as well?).

PR #1244, opened 2019-01-24
fix(PF4): tooltip/popover - include tippy css

jschuler:
Closes: #1189 

Is there a better way? Guess this css is globally added for consumers


packages/patternfly-4/react-core/src/components/Popover/Popover.js
```diff
@@ -11,6 +11,7 @@ import PopoverBody from './PopoverBody';
 import PopoverHeader from './PopoverHeader';
 import PopoverCloseButton from './PopoverCloseButton';
 import GenerateId from '../../internal/GenerateId/GenerateId';
+import 'tippy.js/dist/tippy.css';
```
dmiller9911: This comes back to requiring users to load CSS and forces them to set up min-css-loader, styled-loader, mini-css-plugin. As for a way around it could do 
1. Use something other than Tippy. Maybe check to see what Material-UI does. They may have written their own though.
2. Modify the babel plugin to support loading external CSS and inject it.
3. Write a script to generate a file that loads the CSS and then uses injectGlobal to inject it.
4. Leave it and require consumers to use webpack, css loaders, etc.

jschuler: I could go with 3, use raw-loader to get the string and then use emotion injectGlobal to inject it

jschuler: Just thought of this, we already recommend they import this to get the base css, wouldn't this fall under the same issues you listed?
`import '@patternfly/react-core/dist/styles/base.css';`

dmiller9911: For the base css that is not the only way to load it. It is also possible to copy it to a public direct, etc. Either way the user has to explicitly import it so it is assumed they know how to set that up. Instead of us assuming they have a good build in place. As for raw loader. It is the same deal. We would assume they have raw loader setup. 

I had in the future planned to make a provider that added the base css for the user, but I admittly have decided to not contribute anymore code. Unless something I wrote needs fixed.

I am ok with any option you go with I just wanted you to be aware that this could potentially be a breaking change and will need documentated well

priley86: I believe I was hitting the same problem in #1250 while trying to include an example CSS file...

I ended up just using `gatsby-plugin-sass` with `iso-morphic-style-loader` for the time being:
https://github.com/priley86/patternfly-react/commit/219c3ac186709a8095e0b86b4bfa77c235899184

I plan to drop this commit unless we determine it useful for others. FWIW, there is a really good list of support CSS options in Gatsby here:
https://www.gatsbyjs.org/tutorial/part-two/#component-css

Would be really nice to see `gatsby-plugin-sass` and `gatsby-plugin-emotion` support in PF-React. To @dmiller9911's point, this may put some impetus on the consumer to use a webpack loader, so it might be worth asking around first...

jschuler: I've just copied the source for now and am injecting it along with my overrides using emotion

priley86: yea i was having trouble w/ the plugins mentioned above. This may be the easiest route for now...

packages/patternfly-4/react-core/src/components/Tooltip/styles.js
```diff
@@ -0,0 +1,71 @@
+import { StyleSheet } from '@patternfly/react-styles';
+
+// inject the base tippy styles ('tippy.js/dist/tippy.css')
+export const tippyStyles = StyleSheet.parse(
```
dmiller9911: This can just use `injectGlobal` it will be faster since parse runs some regexs on the string.

Issue #913, opened 2018-11-13
pf4 PageSidebar does not support the default open on desktop and closed on mobile layout


christianvogt:
On mobile, the `PageSidebar` default should be closed while on desktop it should default to open. However the api for `PageSidebar` only supports `isNavOpen` `true | false`. Therefore one must choose to either be always open or always closed.

Compare to the patternfly next example where on mobile the nav closes:
https://pf-next.com/layouts/page/examples-full/?component=Page%20Layout,%20Nav%20Vertical%20Example

christianvogt:
Further to this I found that even the [demo page](http://patternfly-react.surge.sh/patternfly-4/demos/pagelayout) works around this issue by computing the default open value itself:
```
const isNavOpen = typeof window !== 'undefined' && window.innerWidth >= parseInt(breakpointMd.value, 10);
```

I think this is unreasonable for clients to have to do and should be taken care of from the component.

PR #1249, opened 2019-01-25
PF4: Page component managed sidebar

jschuler:
Closes: #913 

Add new prop to Page component isManagedSidebar which handles the sidebar toggling by maintaining a separate state for "desktop view" and "mobile view".

Use can still manage the sidebar toggle themselves if desired, and can implement this behavior by adding a callback to the new onPageResize prop on the Page component.

@matthewcarleton


jschuler:
Examples here: https://1249-pr-patternfly-react-patternfly.surge.sh/patternfly-4/demos/pagelayout

packages/patternfly-4/react-core/src/components/Page/Page.js
```diff
@@ -25,21 +33,80 @@ const defaultProps = {
...
-        {children}
-      </main>
-    </div>
-  );
-};
+class Page extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
ibolton336: Would be a good opportunity to switch this component to class properties & get rid of the constructor:

https://codeburst.io/use-class-properties-to-clean-up-your-classes-and-react-components-93185879f688

christianvogt:
@jschuler a typical pattern in react for differentiating between controlled and uncontrolled state is that the component is uncontrolled (handles the state internally) if the prop which controls the component is not set. Once that prop is set, the component becomes controlled. Then to provide a default value for the uncontrolled state, you prefix the control prop with `default`.

In this case I would say that `isManagedSidebar` is not necessary. If `isNavOpen` is not set, then the component is uncontrolled and its open state is managed internally. To provide a default value (not that it makes sense for the component though, but as an example), the prop `defaultIsNavOpen` can be provided and still have the component be uncontrolled.

jschuler:
Thanks @christianvogt, that makes a lot of sense. Have opened an issue to follow up on this.

PR #1250, opened 2019-01-25
W.I.P. React virtualized extension

### LINKED ISSUE #1229: PF4: Virtualized Table/Virtualized Extension
This issue builds on to #402 for adding the virtualized feature to PF4 React tables.

Relevant design/core issues:
None as of now (may not be needed)

We would like to release this feature as a "pf4 extensions" package right now. This functionality could be distributed and labeled as `@patternfly/react-virtualized-extension` for the time being so that product teams can consume now and revisit later when design review is completed. The utilities developed in this package could be useful for tables, lists, and other scrolling components.

Example from Reactabular:
https://reactabular.js.org/#/features/virtualization

### LINKED ISSUE #1628: PF4 Bring Reactabular in-house
Reactabular is no longer as active as the project here and I believe thru discussions and recent PRs it would be better to bring this minimal library in-house. Reactabular's current source is a good starting point for writing minimal table extensions, but I believe we can customize and improve it much faster here. The upstream maintainer is no longer able to actively maintain it as noted in [#313](https://github.com/reactabular/reactabular/issues/313) [#338](https://github.com/reactabular/reactabular/issues/338#issuecomment-460069536). 

Noted issues causing a poor developer experience:
- `onChange` \ `onBlur` callbacks for rows are not handled correctly upstream and our current [workarounds](https://github.com/patternfly/patternfly-react/pull/1227/files#r263421663) are not great.
- passing `refs` thru [callback methods](https://github.com/reactabular/reactabular/blob/bd94bbc30e22ba5e0ec5871b2bbf38112391972d/packages/reactabular-table/src/header.js#L35) is no longer a good way to pass React refs, and `React.createRef()` should be used instead.
- `getChildContext` is the old way of using React context. We can update [this](https://github.com/reactabular/reactabular/blob/bd94bbc30e22ba5e0ec5871b2bbf38112391972d/packages/reactabular-table/src/provider.js#L8) to use the new stable [Context API](https://reactjs.org/docs/context.html).

Proposed solution:
- We only currently consume `reactabular-table` (which is luckily only [12 js files](https://github.com/reactabular/reactabular/tree/bd94bbc30e22ba5e0ec5871b2bbf38112391972d/packages/reactabular-table/src) 😎). Bring them in-house and expose as a new package `@patternfly/reactabular-table` and consume them from `@patternfly/react-table`. I think this would be a great first step. Another alternative (which may be better and help us avoid another internal package) would be to bring the source directly into `react-table` and just rename the components to avoid conflict, i.e. `BaseProvider`, `BaseBody`, `BaseBodyRow`, `BaseHeader`, `BaseHeader`, `BaseHeaderRow` or attempt to integrate them inside our current wrappers there. 
- Continue writing our own extensions here in patternfly-react

Additional nice-to-haves:
- convert to TSX
- incrementally fix the issues above
- revisit the API of `Provider`, `Body`, `BodyRow`, `Header`, `HeaderRow` and see if we can improve them.

thoughts?

priley86:
**What**:
A very very rough W.I.P. to start some initial discussions about a "virtualized" PF4 table.

fixes #1229 

Some things I've attempted so far:
* pull in some of the [Reactabular virtualized components](https://github.com/reactabular/reactabular/tree/master/packages/reactabular-virtualized) and wrap our existing table `Body`, `BodyWrapper`, and `RowWrapper` components with virtualized extensions (accepting those as props in our current `Table`) 
* Render a simple virtualized table with percent widths. In order to scroll a virtualized table, the author recommends using flexbox for flexible widths with virtually scrolled rows. Needs more discussion and some CSS review, but I've included some sample CSS to tinker with:
https://reactabular.js.org/#/features/virtualization?a=using-relative-column-widths  
* Render a sortable virtualized table. It's important to keep in mind that virtualized rows are dependent on a scroll position (in order to performantly reuse them). For now, I have made the simple assumption that after virtualized sorting occurs to scrollTop, but we can certainly account for an existing scroll position if desired. The `scrollTo` method should allow the consumer to scroll to any desired row programmatically.
 
Still working on:
- [x] ~~provide an `AutoSizer` (similar to react-virtualized [Autosizer](https://github.com/bvaughn/react-virtualized/blob/master/source/AutoSizer/AutoSizer.js)). These essentially just wrap the table and adjust the height/widths appropriately on browser resize.~~ This should not be needed. PF4 tables already responsively size the row widths and the virtualized table resizes for this.
- [x] provide an option and an example to override the scroll container. This allows the consumer to define the scrollable surface. Similar to [WindowScroller](https://github.com/bvaughn/react-virtualized/tree/master/source/WindowScroller) from react-virtualized or simply using the existing functionality in reactabular-virtualized. After the two options above are finished, we can support a very similar API to the one in use today:
```
<WindowScroller scrollElement={document.getElementById('content-scrollable')}>
        {({height, isScrolling, registerChild, onChildScroll, scrollTop}) =>
            <Table
              caption="Sortable Virtualized Table"
              cells={columns}
              rows={rows}
              bodyWrapper={VirtualizedBodyWrapper}
              rowWrapper={VirtualizedRowWrapper}
              sortBy={sortBy}
              onSort={this.onSort}
            >
              <TableHeader />
              <VirtualizedBody height={height} width={width} onScroll={onScroll} rowKey="id" tableBody={this.tableBody} />
            </Table>}
      </WindowScroller>
```
- [x] ~~test out other Table variations and provide examples w/ virtualized scrolling (selectable table, actions table, collapsible table, etc). These should all work similarly to the way they do today.~~ Selectable table and actions table have been implemented. Collapsible is currently out of scope for this because dynamic row heights in conjunction with virtualized rows introduces problems we should probably avoid for now.
- [x] ~~Continue enhancing the Reactabular API surface with latest React methods and propose those changes upstream in Reactabular. In this PR, I have started these enhancements (using new [Context API](https://reactjs.org/docs/context.html#reactcreatecontext), use `refs` more appropriately, start to upgrade the lifecycles methods to React 16.4 plus.~~ Opened [1628](https://github.com/patternfly/patternfly-react/issues/1628) for this.
- [x] closes #1628  bring [reactabular-table](https://github.com/reactabular/reactabular/tree/master/packages/reactabular-table/src) source in-house. This is necessary to continue improving the base table used and to remove legacy React warnings caused by the [old ref interface](https://github.com/reactabular/reactabular/blob/master/packages/reactabular-table/src/body.js#L40) there.
- [x] rebase PF4 React gatsby updates
- [ ] tests
- [ ] implement [scrollable data grid a11y](https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html) as best as possible (Example 3). 
- [ ] convert to typescript where possible

Running demo:
http://pf4-table-virtualized.surge.sh/virtual-scroll/virtualized/

**Additional issues**:

cc: @karelhala @suomiy


packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -9,69 +9,80 @@ const propTypes = {
...
-            })
-          },
-          { id: oneRowKey }),
-        ...oneRow.parent ? {
-          isExpanded: this.parentsExpanded(oneRow.parent) && rows[oneRow.parent].isOpen
-        } : {},
-      }
-    });
+            }),
+            { id: oneRow[rowKey] || oneRowKey }
```
priley86: we need a specific `row-id` (row key) rendered on the table rows currently viewed to support virtualized scroll (a simple counter index would not support this).

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
```diff
@@ -1,23 +1,27 @@
...
-  />;
-};
+  />
+);
 
 RowWrapper.propTypes = {
-  isOpen: PropsType.bool,
-  isExpanded: PropsType.bool
-}
+  trRef: PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: PropTypes.instanceOf(Element) })]),
```
priley86: refs are needed on the row to provide hooks back to the consumer

packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -147,41 +159,44 @@ class Table extends React.Component {
...
+          headerData,
+          rows
         }}
+      >
+        {header}
+        <Provider
+          {...props}
+          renderers={{
+            body: {
+              wrapper: bodyWrapper ? bodyWrapper(rows) : BodyWrapper(rows),
```
priley86: allows us to inject `BodyWrapper` / `RowWrapper` extensions

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/Body.js
```diff
@@ -0,0 +1,269 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { isEqual } from 'lodash-es';
+import { resolveRowKey } from 'reactabular-table';
+import { TableBody, TableContext } from '@patternfly/react-table';
+import calculateAverageHeight from './utils/calculateAverageHeight';
+import calculateRows from './utils/calculateRows';
+
+const initialContext = {
+  amountOfRowsToRender: 3, // First few rows for initial measurement
```
priley86: these are already recalculated/ readjusted after initial render

packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.js
```diff
@@ -3,27 +3,40 @@ import { css } from '@patternfly/react-styles';
...
+    rowIndex,
+    rowData
+  }
+) => {
   const rowId = rowIndex !== undefined ? rowIndex : -1;
   function selectClick(event) {
-    let selected = rowIndex === undefined ? event.target.checked : rowData && !rowData.selected;
-    onSelect && onSelect(selected, selected, rowId);
+    const selected = rowIndex === undefined ? event.target.checked : rowData && !rowData.selected;
+    onSelect && onSelect(event, selected, rowId, rowData);
```
priley86: rowData is needed to access the `row.id` in cases of a selectable virtualized table

priley86:
Thanks very much for the thorough review @karelhala 🙏 

I agree with pretty much everything you note. Much of this is an artifact of incorporating the upstream here. I am very happy to make these updates early next week after I finish some work on UHC. We can continue making Reactabular a much more compelling table here in PatternFly React. 😊

packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.js
priley86: hey @karelhala - i think this one will be needed sooner than later. I was testing `onSelect` callbacks w/ @TheRealJon today and it looks like we'll definitely need `rowData` returned in order to appropriately select by an OpenShift identifier. I'm happy to open another issue to target that specifically... and resolve in a separate PR if you'd like.

This piece can be targeted for this:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.js#L27

it looks like this change was not included in #1227 either...

thoughts?

suomiy: +1 I think it is better to send rowData anywhere it might be remotely useful

priley86: opened #1472

priley86:
have finally had a chance to revisit this. Started to integrate the react-virtualized implementation of [WindowScroller](https://github.com/bvaughn/react-virtualized/blob/9.21.0/source/WindowScroller/WindowScroller.js) today so we can support the API mentioned above which auto sizes the virtualized table based on an outer scrollable element.

priley86:
* rebased and temporarily disabled tests for now...have some questions. This PR had become quite dated, but it's much closer now.
* Will try to address comments above tomorrow.
* Need to do some performance testing next

packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -226,7 +226,7 @@ class Table extends React.Component {
           {...props}
           renderers={{
             body: {
-              wrapper: bodyWrapper || BodyWrapper,
+              wrapper: bodyWrapper(rows) || BodyWrapper,
```
priley86: @suomiy - if i'd like to filter rows, what is the proper way of doing so w/ this extension model?  Currently I am passing in rows... will likely have to refactor this bit...
https://github.com/patternfly/patternfly-react/pull/1250/files#diff-82199de5c757484a78f5c7f625f15843

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
```diff
@@ -77,7 +77,7 @@ class RowWrapper extends React.Component {
 }
 
 RowWrapper.propTypes = {
-  trRef: PropTypes.func,
+  trRef: PropTypes.object,
```
priley86: not quite sure on this

packages/patternfly-4/react-table/src/components/Table/Table.js
suomiy: We should try to aim at chaining all the components with HoC principles as implemented in `react-inline-edit-extension`. `react-table` should be the most bottom component. We should add necessary basic functionality which  `react-virtualized-extension` might miss to `react-table` and activate it by  each extension. I  will discuss the details with you offline.

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
suomiy: We need the function callbacks (`react-inline-edit-extension` uses that) and then chain the trRef callbacks. Please see `editableRowWrapper` and the usage of `trRef`. Is there something the functional approach is missing?

packages/patternfly-4/react-table/src/components/Table/Table.js
priley86: thanks! just wanted to confirm w/ you first what the API should be. Happy to adjust this piece.

packages/patternfly-4/react-table/src/components/Table/Body.js
priley86: i think this is good to allow consumer to override the `id` column if desired

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -9,69 +9,85 @@ const propTypes = {
...
 };
 
 class ContextBody extends React.Component {
   onRow = (row, props) => {
-    const { onRowClick } = this.props;
-    return ({
+    const { onRowClick, onRow } = this.props;
+
+    // if we have an extended onRow, call it to retrieve the resulting props
+    const extendedRowProps = onRow ? onRow(row, props) : {};
```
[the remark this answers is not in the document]
priley86: done... this is slightly different due to how we pass `rowProps` now, but it is similar to above now:
```
    const extendedRowProps = {
      ...rowProps,
      ...(onRow ? onRow(row, rowProps) : {})
    };
```

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/AutoSizer.js
```diff
@@ -0,0 +1,28 @@
...
+const propTypes = {
+  /** Additional classes for header. */
+  className: PropTypes.string
+};
+
+const defaultProps = {
+  className: ''
+};
+
+class AutoSizer extends React.Component {
```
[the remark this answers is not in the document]
priley86: `AutoSizer` is removed. Should not be needed.

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/RowWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+    // for most purposes.
+    if (this.props.initialMeasurement) {
+      this.updateHeight();
+    }
+  }
+  render() {
+    const { updateHeight, initialMeasurement, ...props } = this.props;
+    return <RowWrapper trRef={this.ref} {...props} />;
+  }
+  updateHeight() {
```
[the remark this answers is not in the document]
priley86: have converted to arrow functions

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/RowWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+VirtualizedRowWrapper.propTypes = {
+  'data-rowkey': PropTypes.string.isRequired,
+  updateHeight: PropTypes.func.isRequired,
+  initialMeasurement: PropTypes.bool.isRequired,
+  trRef: PropTypes.object
+};
+
+VirtualizedRowWrapper.shouldComponentUpdate = function(nextProps) {
```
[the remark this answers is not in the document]
priley86: done

packages/patternfly-4/react-table/src/components/Table/Table.js
priley86: Refactored `BodyWrapper` to use a similar API as in `react-inline-edit-extension`. It is clear to the see difference now. The `react-virtualized-extension` needs to actually override rendering of received `rows` b/c it tracks scroll position with the `start-row`, `end-row` and `extra-row`. These `tr` 's are just used to tracking purposes and are based on the implementation in `reactabular-virtualized`. I have marked them `aria-hidden` so that screen readers ignore them, but they are just used for tracking purposes. They could technically be used for test cases as well as scroll position is adjusted.

priley86: `Body` is also similar, however it needs to calculate the currently rendered rows, thus the `onScroll` is extended to do this and `rowsToRender` are passed down as a prop.

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
priley86: For once, I think I agree w/ this [SO post](https://stackoverflow.com/questions/48007326/what-is-the-correct-proptype-for-a-ref-in-react). It depends on how you pass the ref in. Whether as a function like this:
```
ref={node => this.myRefToBePutInReduxGlobalStore = node}
```
or as a ref from `React.createRef()`:
```
<RowWrapper trRef={this.ref}
```
In this case since RowWrapper is extended, it can technically be either. Are you opposed to this for now:
```
PropTypes.oneOfType([
  PropTypes.func, 
  PropTypes.shape({ current: PropTypes.instanceOf(Element) })
])
```

packages/patternfly-4/react-table/src/components/Table/Table.js
suomiy: Not sure if it is necessary at this testing stage, but again I think the extension should be implemented as HoC in the future to allow chaining of multiple extensions.

packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
suomiy: I would rather forbid of passing the ref object, because the ref might be needed by multiple extensions and the functional approach allows to get the ref for many extensions at once. On the other hand the object approach doesn't seem to support that. 

Example:

https://github.com/patternfly/patternfly-react/blob/96d033c4c5de26a8946885d44dcb99d8b7d90d72/packages/patternfly-4/react-inline-edit-extension/src/components/InlineEdit/editableRowWrapper.js#L219-L225

priley86: This one is a toss up, i think i can refactor to callback refs in the future if desired. React doesn't seem to assume either way...

https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

suomiy: I know, this just seems to be the easier way to use the ref across multiple components. 
Let me know if you can think of an alternative.

priley86: fixed in https://github.com/patternfly/patternfly-react/pull/1250/commits/4a41d57b299535c7df2c1fc4b7a2ed0b062a1b83 and https://github.com/patternfly/patternfly-react/pull/1250/commits/f64e9be45e1799145c6a8f95d365c82e3651b164

priley86:
* WindowScroller example now uses default PF4 table styles and scrolls Window surface like [Reactabular](https://reactabular.js.org/#/features/virtualization?a=scrolling-within-a-container) and [React-Virtualized](https://bvaughn.github.io/react-virtualized/#/components/WindowScroller).

Demo:
http://pf4-table-virtualized.surge.sh/components/virtualized

Note: I have left the other examples w/ fixed headers/scrolling tbody, however they could be used in conjunction w/ `WindowScroller` as well.

cc: @mcoker @jeff-phillips-18

priley86:
* today's trick of the day - added demo of dynamic row heights 🏄 🌊 

Also ensured that window resize is supported.
(see demo link above and new Dynamic Heights Example)

Will try to rebase this, add tests, and get ready to 🚢 soon. 😄

priley86:
* rebased w/ latest pf4-react gatsby and updated demo links above

priley86:
Adding a few more fixes after testing some downstream...

https://github.com/patternfly/patternfly-react/pull/1250/commits/3ee7704451b6a0042fd05e290aa04417ab3aeaed

* `@patternfly/react-table` must be installed as a `peerDependency`. There is currently a problem with transient dependencies and exposing React's Context to consumers. You can expose a new instance of Context if your extension consumes the same context as the root application if placed in `dependencies`. This is really an issue in React that should probably be added to React docs:
https://github.com/facebook/react/issues/13346
https://github.com/reactjs/reactjs.org/pull/1112/files

* fix container offset - using the `offsetHeight` of `tbody` is not guaranteed to provide the correct offset. This should make it more consistent.

* fix possible race condition w/ resize listeners on tbody and container, only listen to resize on container if provided.

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/RowWrapper.js
```diff
@@ -0,0 +1,78 @@
...
+    const margin = parseFloat(styles.marginTop) + parseFloat(styles.marginBottom);
+    return Math.ceil(el.offsetHeight + margin);
+  };
+
+  static shouldComponentUpdate(nextProps) {
+    const { columns, rowData } = this.props;
+    // Update only if a row has not been measured and either
+    // columns or rowData hasn't changed
+    if (nextProps.rowData._measured) {
+      return !(columnsAreEqual(columns, nextProps.columns) && isEqual(rowData, nextProps.rowData));
```
suomiy: 1. `columnsAreEqual` is not enough because user can have arbitrary data tied to each `column`
2. `isEqual` can be slow because the user can have large structures in `rowData`. 
Isn't there an alternative way ? What are the problems of updating the component every time?

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/Body.js
```diff
@@ -0,0 +1,297 @@
...
+        'start index',
+        startIndex,
+        'amount of rows to render',
+        amountOfRowsToRender
+      );
+    }
+
+    const style = { height };
+    if (!container) {
+      // if we do not have a parent container to scroll, set the body to scroll
```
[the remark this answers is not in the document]
priley86: done

packages/patternfly-4/react-table/src/components/Table/Table.js
priley86: i've started this in https://github.com/priley86/patternfly-react/commit/2f73d45129af476faa565c4848e3f623d27b740a but noted some issues, will have to see if this possible or not. At the moment, virtualized is not targeted towards supporting other table features which make further modifications to default row renderings/row heights (i.e. collapse/inline-edit/etc.). This further convolutes row height calculation logic and I believe that making use of server pagination/infinite scroll is better for that use case. Virtual scroll is a simple helper intended to emulate react-virtualized (virtual scroll) allowing the user to simply pass all rows and have a utility calculate visible offsets.

packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -10,13 +10,16 @@ const propTypes = {
   /** Specify key which should be used for labeling each row. */
   rowKey: PropTypes.string,
   /** Function that is fired when user clicks on row.  */
-  onRowClick: PropTypes.func
+  onRowClick: PropTypes.func,
+  /** Virtualized rows (optional provided in place of rows) */
+  rowsToRender: PropTypes.array
```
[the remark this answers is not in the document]
priley86: i can try to internalize this logic within the react-virtualize-extension, will see what i can do...

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/RowWrapper.js
priley86: removed this method - it was legacy logic from reactabular-virtualized. We can revisit this if perf problems become an issue

packages/patternfly-4/react-table/src/components/Table/Body.js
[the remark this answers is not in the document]
priley86: fixed in https://github.com/patternfly/patternfly-react/pull/1250/commits/0d737681bf0473686fe698d8c03ca4b7b3d1c5d9

priley86:
Closing in favor of #2011

PR #1251, opened 2019-01-25
fix(build): Support running npm scripts under Windows

TimoSta:
### What:
With the current setup, Patternfly's npm scripts cannot be executed in a Windows environment.

The changes proposed by this PR should make all npm scripts platform independent and runnable on both Windows and Linux/OS X environments.

### How:
- **Use  `shx` to make Unix specific commands platform independent**: Commands like `rm`, `cp` etc. that are used in PF's npm scripts are not supported by Windows command lines. The [`shx` command line tool](https://www.npmjs.com/package/shx) provides a layer of abstraction that makes basic unix commands work cross-platform with the established Unix syntax.
- **Use `cross-var` to make Unix specific command line variable syntax platform independent**: The PF3 packages use Unix style variables `$variableName` to refer to environment variables in their npm scripts to build SASS style sheets. Under Windows, a different syntax must be used to make those variables work. The [`cross-var` command line tool](https://www.npmjs.com/package/cross-var) provides a layer of abstraction that makes Unix style variables work cross-platform.

### Open Points/Concerns:
- ~**The `coveralls` script in the root package.json**: It uses a pipe operator, which isn't supported on Windows as-is.~ _PowerShell supports Unix style pipe operations ` | `, only cmd does not. Go Powershell!_ 💪 
- ~**Performance**: The additional layer of abstraction for some scripts might cause some additional runtime overhead during the build process. Compare avg. Jenkins build runtimes before and after the change to get an idea of the impact?~ _Comparing the [Travis build time including the changes](https://travis-ci.org/patternfly/patternfly-react/builds/484459161) with similar recent PR builds ([one](https://travis-ci.org/patternfly/patternfly-react/builds/484375459), and [another one](https://travis-ci.org/patternfly/patternfly-react/builds/483547928)) shows no significant preformance degradations. All are around the 25 to 30 minute mark._
- ~**Executing Test suite under Windows environment**: While the build works fine, running the Jest test suite under Windows currently produces a whole bunch of errors that do not show up on the Travis build. Investigating and resolving this should probably be done in a separate PR.~ _Resolved with #1255._
- **Test changes under Unix environment**: I currently only have access to a Windows environment. While it _should_ work just fine under a Unix one, this needs to be verifed.

### DIFF
--- a/packages/react-icons/package.json
+++ b/packages/react-icons/package.json
@@ -25,7 +25,7 @@
   "homepage": "https://github.com/patternfly/patternfly-react#readme",
   "scripts": {
     "prebuild": "node ./build/generateIcons.js",
-    "build": "yarn build:babel; yarn build:ts",
+    "build": "yarn build:babel && yarn build:ts",
     "build:babel": "concurrently \"yarn build:babel:cjs\" \"yarn build:babel:esm\"",
     "build:babel:cjs": "cross-env BABEL_ENV=production:cjs babel src --out-dir dist/js",
     "build:babel:esm": "cross-env BABEL_ENV=production:esm babel src --out-dir dist/esm",
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "react": "^16.3.0",
     "react-dev-utils": "^5.0.0",
     "react-dom": "^16.3.0",
+    "shx": "^0.3.2",
     "travis-deploy-once": "^4.4.1",
     "ts-jest": "^23.10.5"
   },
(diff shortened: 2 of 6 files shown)


packages/react-icons/package.json
```diff
@@ -25,7 +25,7 @@
   "homepage": "https://github.com/patternfly/patternfly-react#readme",
   "scripts": {
     "prebuild": "node ./build/generateIcons.js",
-    "build": "yarn build:babel; yarn build:ts",
+    "build": "yarn build:babel && yarn build:ts",
```
TimoSta: `&&` has slightly different semantics than `;` here: with this change, `yarn build:ts` will only be executed if `yarn build:babel` succeeds, while it used to be always executed, independently of the return code of `yarn build:babel`.

However, `&&` is supported cross-platform, while the `;` syntax is not.

Considering that stopping the command is probably a sensible thing to do if the babel build fails, I believe that is an acceptable change.

priley86: nice catch! thanks for your diligent work on this PR @TimoSta and giving us Windows support 👍

PR #1252, opened 2019-01-25
fix: Add consistent aria labels for <nav> elements

### LINKED ISSUE #1182: Consistent labels for <nav> elements
References this issue: https://github.com/patternfly/patternfly-next/issues/1127

rebeccaalpert:
I updated the aria labels for Nav, Breadcrumbs, and PageHeader as requested by @jgiardino. Tabs and Pagination aren't in PF4 yet, so I wasn't able to address those in this PR. (Tabs are being added by Titani in a separate PR.)

I also updated the examples, tests, and snapshots for Nav and Breadcrumbs to reflect the new aria labels.

Fixes #1182.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Nav/Nav.js
+++ b/packages/patternfly-4/react-core/src/components/Nav/Nav.js
@@ -49,12 +49,6 @@ class Nav extends React.Component {
     });
   }
 
-  // Determine if it's a tertiary nav or not (tertiary needs a different aria-label)
-  determineLabel() {
-    const isTertiary = React.Children.map(this.props.children, child => child.props.variant === 'tertiary');
-    return isTertiary[0] ? 'Local' : 'Global';
-  }
-
   render() {
     const { 'aria-label': ariaLabel, children, className, ...props } = this.props;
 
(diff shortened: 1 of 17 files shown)


packages/patternfly-4/react-core/src/components/Nav/Nav.js
```diff
@@ -49,8 +49,14 @@ class Nav extends React.Component {
     });
   }
 
+  // Determine if it's a tertiary nav or not (tertiary needs a different aria-label)
+  determineLabel() {
+    const isTertiary = React.Children.map(this.props.children, child => child.props.variant === 'tertiary');
```
tlabaj: Why do you need to go through all the children here if you are only looking at the variant from the first child.
Can you just look at the first child and maybe do this inline rather than having a function?

packages/patternfly-4/react-core/src/components/Nav/Nav.js
```diff
@@ -13,7 +13,7 @@ const propTypes = {
   /** Callback for when a list is expanded or collapsed */
   onToggle: PropTypes.func,
   /** Accessibility label */
-  'aria-label': PropTypes.string.isRequired,
+  'aria-label': PropTypes.string,
```
tlabaj: looks like you removed the isRequired label her but did not add a defaultProp.

rebeccaalpert: Pushed a new commit with a default prop.

PR #1255, opened 2019-01-25
fix(build): Support running the test suite under Windows

TimoSta:
Running the linter and test suite under Windows currently fails. The changes proposed by this PR should resolve this.

### What

- **Check out and create new files with Unix style line endings:** By default, files are checked out and created with CRLF line endings on Windows. This is rejected by the `linebreak-style` ESLint rule, which requires Unix style LF line breaks. A `.gitattributes` file was added to prevent automatic conversion to Windows style linebreaks during checkout, and a `.editorconfig` file was added to instruct IDEs to create new files with Unix style linebreaks. (See b0d83fc59d23388121b5fc008f8a126c4abfd34c)
- **Properly load generated files from fs mock when testing react-tokens generation:** The generated files and their contents are compared against a snapshot. Only the last two segments of the file path are used to identify a file, to make the comparison independent of the absolute location in the file system. To correctly extract this piece of the file path, the test script needs to handle back slashes (used on Windows) as well as forward slashes (used on Linux/OS X). (See 024f77e903db152cbd3d5cd7b24826e4c9fa849c)
- **Use OS-specific line endings for expected codemod outputs:** Codemods should produce code which uses OS-specific line endings, which is already the case. However, it used to be compared against an expected value that _always_ used LF line endings. To fix this, the expected values are formatted to use OS-specific line endings at runtime. (See 240f25c983187f051814f42666ccceec32bbd605)
- **Generate proper stylesheet imports during react-core build:** The generated file paths in react-core, which are used to import `.css.js` files created during the build were inserted without slashes between path segments, resulting in failed imports. This is corrected, and the import path segments are now correctly separated by slashes. (See b63f8c0d12fd13e6ff586b345a324b90adfdeece)

image

🎉


TimoSta:
@dlabaj Not quite, there's some other issues that still need some tweaking.

I'm trying to resolve those missing pieces right now, but overall it seems to be fairly straightforward to make it work.

package.json
```diff
@@ -51,7 +51,7 @@
     "minimist": "^1.2.0",
     "npmlog": "^4.1.2",
     "plop": "^2.0.0",
-    "prettier": "^1.14.2",
+    "prettier": "^1.16.1",
```
TimoSta: This upgrade is necessary to enable the `endOfLine` config option. (See https://prettier.io/docs/en/options.html#end-of-line)

packages/react-codemods/transforms/pf3-pf4.button.test.js
```diff
@@ -1,9 +1,29 @@
...
+ * JSCodeShift produces OS-dependent line endings however, LF for Unix based
+ * systems, and CRLF for Windows based ones. This is also likely what most
+ * projects would like to see the codemod prduce.
+ *
+ * To make sure we both adhere to Patternfly's conventions, and compare the
+ * correct OS-specific line endings during the test cases' assertions, we store
+ * expected values with LF line endings, and convert them into the OS-specific
+ * ones at runtime using prettier.
+ */
+const PRETTIER_EOL = SYSTEM_EOL === '\r\n' ? 'crlf' : 'cr';
```
TimoSta: With all the extra effort it takes to work around line ending efforts, maybe it would be a better approach to disable the line ending lint rule, and let Git ensure that they are correctly set?

If Git's `core.autocrlf` option is set correctly, it should make sure that all files have LF line endings after check-in, but devs can work with the OS-specific line endings locally.

There are likely ways to force non-LF endings into the code base with this setup, but that risk may be a worthy tradeoff for the added flexibility?

packages/patternfly-4/react-styles/src/build/util.js
```diff
@@ -40,7 +40,7 @@ export function writeCSSJSFile(rootPath, originalPath, destinationPath, contents
 
 export function getRelativeImportPath(from, to) {
   const parsedTo = path.parse(to);
-  const newImportPath = path.normalize(path.join(relative(from, parsedTo.dir), parsedTo.base).replace(/\\/g, ''));
```
TimoSta: Not quite sure why we've removed backslashes here. From what I can see, it didn't have any effect on Linux/OS X, and completely broke paths on Windows.

Am I missing something?

TimoSta:
@dlabaj The tests are now running fine on Windows for me with the proposed changes. I'd appreciate a review and feedback!

packages/react-codemods/transforms/pf3-pf4.button.test.js
dlabaj: I agree.  @tlabaj @jschuler Thoughts?

dlabaj: @dana Thoughts?

package.json
priley86: if upgrading prettier...please run `yarn prettier` to update all formatting across the project w/ this PR. It seems we also need some updates now in `.prettierignore` now that i try this locally, but i think it can be done in a different PR 😄

packages/react-codemods/transforms/pf3-pf4.test.js
```diff
@@ -1,9 +1,29 @@
+import { EOL as SYSTEM_EOL } from 'os';
 import prettier from 'prettier';
 import { defineInlineTest, runInlineTest } from 'jscodeshift/dist/testUtils';
 import transform from './pf3-pf4';
 
-const prettierConfig = prettier.resolveConfig.sync(process.cwd());
-const pretty = src => prettier.format(src, { parser: 'babylon', ...prettierConfig });
+/**
```
priley86: i am still seeing some tests fail locally in this file on Mac...happy to give you the output if needed.

TimoSta: @priley86 The Travis build is failing too. Looks like that's related to the Prettier upgrade -- some files are missing some formatting it seems. See https://travis-ci.org/patternfly/patternfly-react/builds/486997667#L5198 -- are the issues you are running into the same ones?

I disabled the linting locally, since I get mostly noise due to #1256. Looks like it's the PF3 packages that are causing trouble, I should be able to lint them on my machine without problems.

I'll take a look!

package.json
priley86: with the conversion to typescript/tslint for pf4 upcoming, i don't know how much time we invest on those eslint rules for pf4 - @dgutride can probably weigh this. Regarding prettier though - i think we should just ensure running `yarn prettier` is consistent for now (certainly realize the other issue exists). I think the pf4 team will have to weigh this. Running `yarn prettier` to format our code now is low effort though ;)

TimoSta: f23d5ca44f83db8dcd3d1103ec2a7a885107b48a fixes the formatting for the pf3 stuff. The build is still broken, but less badly now I suppose, trying to figure out what else is wrong with it...

The linting is not failing anymore at least. If you want me to any other formatting, I'm happy to!

packages/patternfly-4/react-styles/src/build/util.js
priley86: i'm honestly not sure... but just ensuring you can run `yarn clean; yarn build; yarn start:pf4` and we verifying we still have styles loading correctly should hopefully verify this...

priley86: (rebuilding styles)

packages/react-codemods/transforms/pf3-pf4.test.js
priley86: yes, i think you've captured it. Seems you just need a few more snapshot updates in react-codemods... `react-codemods` is not yet used in the wild (still a P.O.C.) so definitely not a risk right now. Just need to ensure tests pass. I will try to pull your branch again and build locally on Mac after Travis happy 😸

dgutride:
@redallen  - can you take a look, too?

packages/react-codemods/transforms/pf3-pf4.js
```diff
@@ -149,7 +149,7 @@ module.exports = (file, api, options) => {
 
     return prettier
       ? prettier.format(transformedSource, {
-          parser: 'babylon',
```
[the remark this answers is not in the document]
redallen:
```console.warn node_modules/prettier/index.js:440
      { parser: "babylon" } is deprecated; we now treat it as { parser: "babel" }.```
I also changed it in #1541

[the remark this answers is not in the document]
TimoSta: @dlabaj I had to [upgrade Prettier to 1.16.x to support the `endOfLine` parameter](https://github.com/patternfly/patternfly-react/pull/1255#discussion_r251538863).

The babylon parser [was deprecated with Prettier 1.16.0 and replaced with the bable parser](https://prettier.io/docs/en/options.html#parser). In my understanding, it's the same thing, just a different name.

redallen:
Checked it out locally, tests should pass after updating `yarn.lock`. I like the changes.

TimoSta:
I haven't had the time to look into the build failure yet myself unfortunately, but I'll try to find time to fix it over the next one or two days.

@redallen -- You're saying that merging master into this should help, right? Or is there anything specific around the lockfile that needs to be updated?

redallen:
No. Run `yarn install` in the project root and then commit `yarn.lock`.

redallen:
I'll merge with one more reviewer.

Issue #1238, opened 2019-01-24
BadgedResource tooltip on resource name


lizagilman:
Hi,
Using `BadgedResource` component I would like to have a tooltip on the `resourceName` prop. AFAIU, a tooltip is only shown on the badge itself, but not on the `resourceName`. 

If I pass a DOM fragment (in my case, a `resourceName` with overlay and react router link) instead of an expected `string` it solves my problem, but then I get proptypes validation warning.

Considering [implementation](https://github.com/patternfly/patternfly-react/blob/3e1d440f1900b9a55e7a928492c8c7b9b00536d2/packages/patternfly-3/patternfly-react-extensions/src/components/BadgedResource/BadgedResource.js#L72) , can we not restrict to a `string`? Or provide node as `children` ?

My use case:

Initial goal - have a resource badge with link and tootlip on `resourceName`

code: 
```
    <BadgedResource
          resourceName={clusterName}
          kindStr="Cluster"
          kindAbbr="CL"
          badgeColor="#00b9e4"
      />
```

where `clusterName` is -
```  
    <OverlayTrigger 
        overlay={<Tooltip id={cluster.id}>{``cluster name: ${cluster.name}``}</Tooltip>}
        placement="right"
      >
        <Link to={``/cluster/${cluster.id}``}>{name.trim() !== '' ? name : cluster.name}</Link>
      </OverlayTrigger>
    );
```
Looks like works well, but `resourceName` expected to be a `string`. 

@jeff-phillips-18 @priley86  what do you guys think?

/cc @cben  @Laviro @sharvit

sharvit:
It makes sense to me that `resourceName` should actually be the `children` while the children should be from the type of `node`.

Together with your example, I see it getting used as:
```jsx
<OverlayTrigger 
  overlay={<Tooltip id={cluster.id}>{``cluster name: ${cluster.name}``}</Tooltip>}
  placement="right"
>
  <BadgedResource
    kindStr="Cluster"
    kindAbbr="CL"
    badgeColor="#00b9e4"
  >
    <Link to={``/cluster/${cluster.id}``}>{name.trim() !== '' ? name : cluster.name}</Link>
  </BadgedResource>
</OverlayTrigger>
```

Btw, you might be able to use it with the current implementation as:
```jsx
<OverlayTrigger 
  overlay={<Tooltip id={cluster.id}>{``cluster name: ${cluster.name}``}</Tooltip>}
  placement="right"
>
[...]

PR #1260, opened 2019-01-26
fix(BadgedResource): don't restrict resourceName to a string

lizagilman:
**What**: A fix following discussion on #1238


packages/patternfly-3/patternfly-react-extensions/src/components/BadgedResource/BadgedResource.js
```diff
@@ -99,7 +99,7 @@ BadgedResource.propTypes = {
   /** Full name of the resource kind to display in the badge tooltip and for screen readers (defaulted if valid resourceKind is given) */
   kindStr: PropTypes.string,
   /** Name of the resource */
-  resourceName: PropTypes.string,
+  resourceName: PropTypes.node,
```
jeff-phillips-18: While type node will accept a string, we typically use PropTypes.oneOfType([PropTypes.string, PropTypes.node]) to make it more clear.

lizagilman:
Thanks @jeff-phillips-18 , updated.

PR #1262, opened 2019-01-28
fix(Chip): Replace check for length of chip text with refs

### LINKED ISSUE #1197: Chip tooltip overflow conditional render fix
Currently we are hard coding a check for the length of the chip text to determine if a tooltip is required. This should be replaced with a more elegant solution (e.g. using refs to check the css overflow values). Also, there is a bug with the current implementation causing an error to be thrown if the chip is empty.

rebeccaalpert:
Replaced hard-coded check with refs/check for CSS overflow. Also made children optional to allow for empty chips.

Fixes #1197.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Chip/Chip.js
+++ b/packages/patternfly-4/react-core/src/components/Chip/Chip.js
@@ -7,6 +7,12 @@ import { TimesCircleIcon } from '@patternfly/react-icons';
 import styles from '@patternfly/patternfly-next/components/Chip/chip.css';
 import GenerateId from '../../internal/GenerateId/GenerateId';
 class Chip extends React.Component {
+  span = React.createRef();
+  state = { isTooltipVisible: false };
+
+  componentDidMount() {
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/Chip/Chip.js
```diff
@@ -7,6 +7,17 @@ import { TimesCircleIcon } from '@patternfly/react-icons';
 import styles from '@patternfly/patternfly-next/components/Chip/chip.css';
 import GenerateId from '../../internal/GenerateId/GenerateId';
 class Chip extends React.Component {
+  constructor(props) {
```
ibolton336: can we get rid of this constructor and use class properties? Should be able to since this is a class component. Just do   span = createRef(); and state = {isTooltipVisible: false} right on the class since we are using the babel class properties plugin.

rebeccaalpert:
I just amended the commit to get rid of the constructor and used class properties instead.

PR #1269, opened 2019-01-28
fix(Radio): Modify examples and aria-label/label

### LINKED ISSUE #1007: Radio Component Audit
1. What is the interaction difference between the Controlled and Uncontrolled Radio. I'm assuming that the Uncontrolled Checkbox would be like the Uncontrolled Switch, where the user cannot manipulate its state? Right now I can check it.

2. Custom Label Radio should be removed from the React example as there is no example in Core or in Wireframes. @mcarrano do we want to create an issue for this Custom label radio with the badge?

3. Radio buttons are meant to be mutually exclusive (you can only select one radio button from the group -- see in Core). The Controlled Radio example currently works so that once you select an unchecked radio button, the previous radio button does not uncheck, meaning every item in the group gets selected. Can we fix this functionality so that only one radio button can be selected at a time.

### LINKED ISSUE #1107: PF4 Modal footer occupies space when not in use
I noticed that some of my charts are being cut off at the bottom of the PF4 Modal. Even though there is plenty of room at the bottom of the Modal. Looking at the HTML elements for Modal, I see the `pf-c-modal-box__footer` selector even though I'm not using a footer.

In the screenshot below, I've scrolled to the bottom of the Modal and my charts are still cut off. 
screen shot 2018-12-21 at 4 16 14 pm

Would it be possible to omit the div element when a footer has not been provided? My workaround is to set `display:none` for pf-c-modal-box__footer.

rebeccaalpert:
I removed the Custom Label Radio example, fixed the remaining examples so only one radio button can be selected at a time, and modified aria-labeling so that it's only present if there is no label.

I also updated the relevant snapshot.

Fixes #1007.


tlabaj:
@christiemolloy Similar to the other PR #1270.  I think the examples are fine as is and uncontrolled component should not be changed.

tlabaj:
@christiemolloy @rebeccaalpert I opened an issue against the Switch component so it functions when uncontrolled.
#1275

PR #1270, opened 2019-01-28
fix(Checkbox): Adjustments to examples

### LINKED ISSUE #1006: Checkbox Component Audit
1. What is the interaction difference between the Controlled and Uncontrolled Checkbox. I'm assuming that the Uncontrolled Checkbox would be like the Uncontrolled Switch, where the user cannot manipulate its state? Right now I can check it.
2. Custom Label Checkbox should be removed from the React example as there is no example in Core or in Wireframes. @mcarrano do we want to create an issue for this Custom label checkbox with the badge? 
3. When checkboxes exist in a group, a user can select as many checkboxes as they want. Can we add more checkboxes to the Controlled Checkbox example to demonstrate this functionality?

rebeccaalpert:
I removed the custom label checkbox example and reworked the controlled checkbox example so it could handle more than one checkbox.

I also added another checkbox to the uncontrolled checkbox example for consistency's
sake.

Fixes #1006.


tlabaj:
@christiemolloy is the reason for the user to not control the uncontrolled component?  In React, an uncontrolled component is controlled by the DOM whereas a controlled component is controlled by a component.  So for instance here, the box would be checked based on a prop passed in for the controlled example.  in the uncontrolled example, the DOM handles that.  So in my opinion the example are fine as is.

christiemolloy:
@tlabaj thats fine to leave it like that then. Whats the difference between the uncontrolled switch then?

PR #1271, opened 2019-01-29
fix(LoginPage): remove password's caps lock warning

Laviro:
Continued to the design refactoring [discussion](https://github.com/patternfly/patternfly-design/issues/671#issuecomment-457323229l)
where we decided to remove the inputs warning,

this is the first step - removing the caps lock warning.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/LoginPage.stories.js
@@ -81,7 +81,6 @@ const createProps = () => {
           minLength: 8,
           errors: card.passwordField.errors,
           warnings: card.passwordField.warnings,
-          warning: card.passwordField.warnings.capsLock,
           showWarning: true
         },
         additionalFields: null,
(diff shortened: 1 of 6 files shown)


rebeccaalpert:
Looks good to me.

jeff-phillips-18:
I disagree with making this change at this point. Applications may be using this already and taking it away would break them.

Laviro:
@jeff-phillips-18 this is the results of the discussion mentioned above https://github.com/patternfly/patternfly-design/issues/671#issuecomment-457323229l
where we decided to remove the input errors and use only the top form error,
part of the discussion was to remove the `CAPS-LOCK` warning as it is not secure and maybe the user would like to use it and don't want anyone else to be aware of it,
any other thoughts about this are more than welcome, we can continue on the [discussion](https://github.com/patternfly/patternfly-design/issues/671#issuecomment-457323229l)

jeff-phillips-18:
I don't disagree with the discussion nor the decision. However, applications already using this component may disagree and expect the warning. We could make it an option to show it but the default would have to be to show it so as to not break existing uses. Next major release we can remove the option.

Laviro:
Thanks @jeff-phillips-18 I agree :)

Laviro:
Closing this one as it would be implemented in https://github.com/patternfly/patternfly-react/pull/1278

PR #1274, opened 2019-01-29
build(docbuild): increase gatsby build memory

priley86:
**What**:
Adds the suggested fix [here](https://github.com/gatsbyjs/gatsby/issues/2796#issuecomment-342017625) to increase Gatsby's memory while compiling static assets.

Bumping from [1.5ish gbs](https://github.com/nodejs/help/issues/1377#issuecomment-403801254) to 4gb. I think we can technically increase to 7.5 gb from the [Ubuntu Travis image docs](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system), but this should be enough for now...

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-docs/package.json
+++ b/packages/patternfly-4/react-docs/package.json
@@ -37,8 +37,8 @@
     "gatsby"
   ],
   "scripts": {
-    "docbuild": "node build/copyDocs.js && gatsby build",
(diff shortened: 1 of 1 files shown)


priley86:
validated this previously in #1250... seeing it creep in other builds like this [one](https://github.com/patternfly/patternfly-react/pull/1272) so maybe we merge it sooner than later?

Issue #1214, opened 2019-01-17
PF4: Custom Select Component: Single Select variation


tlabaj:
**This issue covers the Single Select variation of the Custom Select only:**   
(This issue is a child of issue #1206)

Related issue in Core: https://github.com/patternfly/patternfly-next/issues/651  

Info provided on the component: https://docs.google.com/document/d/1rrCdARutABI-Vpa_mdMWGl2xJJTtt2dPQ2bjFx2zwCM/edit?usp=sharing

see: https://pf-next.com/components/Select/examples/

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @srambach (CSS) @mcarrano (interaction design) @christiemolloy (visual design)

rachael-phillips:
Closing this issue because it was completed in PR #1277

PR #1277, opened 2019-01-29
Single select

kmcfaul:
**What**: Added select wrapper and single select variant along with tests and TS declarations. TS was tested locally. 

Tried to structure Select for future use with other variants. Select control/state is outside of the component for now, matching Dropdown, but we may want to move that into Select itself and instead issue a callback when the selected items are updated.

Keyboard navigation is similar to Dropdown.

**Refers to issue**: #1214


packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,16 @@
+import { HTMLProps, FormEvent } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface SelectProps extends Omit<HTMLProps<HTMLOptionElement>> {
+  isExpanded?: boolean;
+  onSelect: Function;
```
dmiller9911: this is part of `HTMLProps<HTMLOptionElement>`

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,16 @@
+import { HTMLProps, FormEvent } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface SelectProps extends Omit<HTMLProps<HTMLOptionElement>> {
```
dmiller9911: `Omit` is only needed if a prop from the html element needs to be overwritten.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,16 @@
...
+  isExpanded?: boolean;
+  onSelect: Function;
+  onToggle: Function;
+  placeholderText?: string;
+  selectOptions?: array;
```
dmiller9911: Array of what? to not type it (which shouldn't be done) you would want to use `any[]` or `Array<any>`

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,16 @@
+import { HTMLProps, FormEvent } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface SelectProps extends Omit<HTMLProps<HTMLOptionElement>> {
+  isExpanded?: boolean;
+  onSelect: Function;
+  onToggle: Function;
```
dmiller9911: correct typing for this is `onToggle(value: boolean): void`

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,16 @@
...
+  placeholderText?: string;
+  selectOptions?: array;
+  selections?: string;
+  variant?: string;
+}
+
+declare const Select: React.SFC<SelectProps>;
```
dmiller9911: `React.FunctionComponent` or `React.FC` those are the new typings. I know people are going through and updating the old `SFC` references now.

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -0,0 +1,102 @@
...
+  isExpanded: false,
+  selectOptions: null,
+  selections: null,
+  placeholderText: 'Please choose one',
+  variant: 'single'
+};
+
+class Select extends React.Component {
+  componentDidUpdate() {
+    if (!this.props.isExpanded) this.openedOnEnter = false;
```
dmiller9911: this should probably live in state.

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -0,0 +1,102 @@
...
+      selectOptions,
+      selections,
+      placeholderText,
+      ...props
+    } = this.props;
+    return (
+      <div
+        className={css(styles.select, isExpanded && styles.modifiers.expanded, className)}
+        ref={ref => {
+          this.parentRef = ref;
```
dmiller9911: refs should use the new `React.createRef` syntax. https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

packages/patternfly-4/react-core/src/components/Select/SelectOption.d.ts
```diff
@@ -0,0 +1,13 @@
+import { HTMLProps, FormEvent } from 'react';
+import { Omit } from '../../typeUtils';
+
+export interface SelectOptionProps extends Omit<HTMLProps<HTMLOptionElement>, 'disabled'> {
```
dmiller9911: sendRef needs added

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import { SelectContext } from './selectConstants';
+
+const propTypes = {
```
dmiller9911: sendRef needs added here too.

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
```diff
@@ -0,0 +1,142 @@
...
+  isFocused: false,
+  isHovered: false,
+  isActive: false,
+  isPlain: false,
+  iconComponent: CaretDownIcon,
+  onToggle: Function.prototype
+};
+
+class SelectToggle extends Component {
+  componentDidMount = () => {
```
dmiller9911: these can be written as `componentDidMount()` and `componentWillMount()` There isn't a bunch of overhead, but making these instance properties will use more memory and have a slower instantiation.

mcarrano:
@kmcfaul The behavior of this looks good with one exception. Is there any way to fix the width of the control (or make that an option)? I wouldn't necessarily want this to resize whenever I select a new value, especially if it's used in a layout with other elements following it.

kmcfaul:
@mcarrano @tlabaj @jschuler 
I agree about the behavior that width should be controllable, and I can add an optional width prop that controls a style attribute on the container div, but I'm not sure if it would be better to just let the user pass the style object into the component instead. Anyone have any thoughts about which method would be better?

mcoker:
I would probably expect the default  width of the select to match the widest item in the select menu. I guess that's a little weird when the initial value (ie, "please choose one") isn't in the menu. In that case, I would probably still expect the menu width to match the width of "please choose one", and not change when you select something else in the menu.

I could also see allowing the menu to be as wide as the parent container, so it's basically 100% width, and that's constrained by the container it's in. And additionally, in this state, you could use a `width` prop and specify a width and overwrite the 100% width.

So maybe 2 modes. 1) The width matches the width of the widest thing in it, so that it doesn't change when you select things, and 2) it's 100% width and fills the width of it's parent element, and you could use this with a `width` prop to overwrite the width with something specific passed by the user.

mcarrano:
This sounds good @mcoker . I think it's important to be able to constrain the width independent of the content. For example, we might use the select to select a server name where the names could be anything and possibly long. Or worse the available values might change based on something else in the page. In this case I would not want the width of the Select to shrink or grow dependent on the values that are loaded. 

On a related note, can I assume that long values will truncate (i.e. if they do not fit)?

mcoker:
@mcarrano 

> On a related note, can I assume that long values will truncate (i.e. if they do not fit)?

Currently the text in the toggle will truncate, but the text in the dropdown options will not. The menu will just grow to match the width of the text inside of it. 

screen shot 2019-02-05 at 8 39 08 am


Looking at Carbon and Bootstrap, Carbon always displays the dropdown menu as wide as the toggle (the entire dropdown is 100% width by default, so it consumes all of the available space it's in), then wraps the text if it exceeds the dropdown menu width.

screen shot 2019-02-05 at 8 41 00 am

Bootstrap behaves the same way we do.

screen shot 2019-02-05 at 8 41 35 am

mcarrano:
In that case @mcoker I like it the way we have it. So long object names can be displayed at full length in the menu without wrapping.

kmcfaul:
> So maybe 2 modes. 1) The width matches the width of the widest thing in it, so that it doesn't change when you select things, and 2) it's 100% width and fills the width of it's parent element, and you could use this with a `width` prop to overwrite the width with something specific passed by the user.

I'm having trouble finding a good way to implement case 1 so far, but case 2 is very straightforward. I'm curious if having the width default to 100% would be the best solution overall?

> Currently the text in the toggle will truncate
 
@mcoker The current build doesn't seem to truncate, if you define a long placeholder text the toggle div expands vertically. Am I missing a css class?

mcoker:
We would likely need to add `width: 100%` to it in core if we're all on board with that change.

> The current build doesn't seem to truncate, if you define a long placeholder text the toggle div expands vertically. Am I missing a css class?

Yep, looks like you're missing `.pf-c-select__toggle-text` in `.pf-c-select__toggle-wrapper`

mcoker:
Agree re: the clear button. I think the default "none" option or whatever the text is in the toggle is also represented as an option in the menu to de-select would work. I also wonder if we could use the `fa-times-circle` button to clear the select, similar to what we have in the typeahead examples. Another example is in this [react select component](https://jedwatson.github.io/react-select/).

kmcfaul:
I added the missing toggle text class, and removed the 'Clear' button I had in the example. I had added the button to test clearing the selection from outside the selection per the integration notes, but it's not necessary for the example.

For allowing the select to control clearing itself, I am working on adjusting the implementation so that instead of separately defining the placeholder text, the user can define their first option for the select to be a placeholder ("Choose...") that will be selected by default.

For width, I was imagining adding a prop where the user could specify either a number (for px width) or string (for % width) that defaulted to 100%. This would be applied in a style attribute on the select container div. This also would solve the resizing the div does based on the select item's length.

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -0,0 +1,102 @@
...
+  /** Classes applied to the root of the Select */
+  className: PropTypes.string,
+  /** Flag to indicate if select is expanded */
+  isExpanded: PropTypes.bool,
+  /** Placeholder text of Select */
+  placeholderText: PropTypes.string,
+  /** Array of SelectOption nodes that will be rendered */
+  selectOptions: PropTypes.array,
+  /** Selected item(s) structure */
+  selections: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
tlabaj: @srambach For the Single selection variation will the selection always be the text string?

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,14 @@
+import { HTMLProps, FormEvent } from 'react';
+
+export interface SelectProps extends HTMLProps<HTMLOptionElement> {
+  isExpanded?: boolean;
+  onToggle(value: boolean): void;
+  placeholderText?: string;
+  selectOptions?: ReactNode[];
+  selections?: string;
```
tlabaj: this says selections is a  string.  the .js file says string or array.

mcarrano:
> For allowing the select to control clearing itself, I am working on adjusting the implementation so that instead of separately defining the placeholder text, the user can define their first option for the select to be a placeholder ("Choose...") that will be selected by default.

Can we make sure this is optional? There are cases where this makes sense and others where it does not. For example, if the select is used in a filter, I may select a value to apply a filter and clear the filter via the Clear Filter(s) action in the toolbar. I wouldn't be sure if resetting the Select would act as a clear filter or if it's just resetting the select list. In this case I'd rather rely on the external clear.

packages/patternfly-4/react-core/src/components/Select/SingleSelect.js
```diff
@@ -0,0 +1,87 @@
...
+    } else {
+      this.refCollection[nextIndex].current.focus();
+    }
+  };
+
+  render() {
+    const { children, className, isExpanded, openedOnEnter, selected, ...props } = this.props;
+    this.renderedChildren = this.extendChildren();
+    return (
+      <ul {...props} className={css(styles.selectMenu, className)} role="listbox">
```
tlabaj: aria-labeledby is missing here.  You need to pass id down from Select.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,99 @@
...
+      selected,
+      sendRef,
+      keyHandler,
+      index,
+      ...props
+    } = this.props;
+    return (
+      <SelectContext.Consumer>
+        {onSelect => (
+          <li>
```
tlabaj: missing aria-pressed here.

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
```diff
@@ -0,0 +1,149 @@
+import React, { Component } from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import { CaretDownIcon } from '@patternfly/react-icons';
+
+const propTypes = {
+  /** HTML ID of dropdown toggle */
+  id: PropTypes.string.isRequired,
```
tlabaj: The id should not come down from the select here.  I think for now we can just generate it here.  I only see it being referenced by the aria-labeledby on this same div.
@jgiardino I know accessibility docs will be updated once this implementation is done,  should we have this match core for now?

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -0,0 +1,102 @@
...
+      ...props
+    } = this.props;
+    const { openedOnEnter } = this.state;
+    return (
+      <div className={css(styles.select, isExpanded && styles.modifiers.expanded, className)} ref={this.parentRef}>
+        <SelectContext.Provider value={onSelect}>
+          {variant === 'single' && (
+            <React.Fragment>
+              <SelectToggle
+                id={`pf-toggle-id-${currentId++}`}
```
tlabaj: this id should be passed to the SingleSelect Component for the aria-labeledby prop.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
+import { css } from '@patternfly/react-styles';
+import PropTypes from 'prop-types';
+import { SelectContext } from './selectConstants';
+
+const propTypes = {
+  /** the value for the option */
+  children: PropTypes.string,
```
tlabaj: I don't think we don't have a need for children here since we are using value.

packages/patternfly-4/react-core/src/components/Select/SingleSelect.js
```diff
@@ -0,0 +1,87 @@
...
+        index
+      })
+    );
+  }
+
+  sendRef = (ref, index) => {
+    this.refCollection[index] = ref;
+  };
+
+  keyHandler = (index, position) => {
```
tlabaj: We may want to move this into a utility class if the other variations of select may use it. Also maybe ad some unit test.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
kmcfaul: What is the usage of aria-pressed? Is it set to true or a value, and is it only when the li is selected?

packages/patternfly-4/react-core/src/components/Select/Select.js
kmcfaul: Is there an aria-labeledby on both SingleSelect and SelectToggle? Your message on SelectToggle makes is sound like the aria-labeledby is only on SelectToggle (so the id should be generated there).

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
kmcfaul: I left it in as an alternate declaration method. So users could still do <SelectOption>value</SelectOption> as well as <SelectOption value={value} />. Should I remove it?

packages/patternfly-4/react-core/src/components/Select/SingleSelect.js
kmcfaul: I'll refactor this as we add more variants if its needed.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
tlabaj: @jgiardino can you give some feedback on this please?

packages/patternfly-4/react-core/src/components/Select/Select.js
tlabaj: Looks like it is on both in core.  I am not sure what the SelectToggle aria-labeleld by is supposed to be linked to.  It is unclear to me in the core example.  @jgiardino do you know?

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
```
dgutride: You'll need to drop the -next in the import here

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
```
dgutride: -next is invalid now

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
```diff
@@ -0,0 +1,146 @@
+import React, { Component } from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
```
dgutride: -next again...

packages/patternfly-4/react-core/src/components/Select/SingleSelect.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import styles from '@patternfly/patternfly-next/components/Select/select.css';
```
dgutride: -next

jgiardino:
Thanks for working on this one!

I tested so far in VO and JAWS. JAWS acts a little strange. I have some ideas on why, but I'm not really able to update the html in the inspector and test if these changes will fix all the weirdness. But I'm happy to test again.

Here are some suggestions on tweaks to make in the HTML:

- Use a `<button>` element for the toggle. When I test other examples of a custom Select component, what gets announced in both JAWS and VO are different than what's announced for this implementation. I think not using a native button is causing that.
  Currently, the toggle is defined like this:
  ```
  <div id="pf-toggle-id-78" class="pf-c-select__toggle" tabindex="0" type="button" aria-expanded="true" aria-haspopup="listbox">
    <div class="pf-c-select__toggle-wrapper">Mr</div>
    <svg ...>...</svg>
  </div>
  ```
  But I think this would yield better info:
  ```
  <button id="pf-toggle-id-78" class="pf-c-select__toggle" aria-expanded="true" aria-haspopup="listbox">
    <div class="pf-c-select__toggle-wrapper">Mr</div>
    <svg ...>...</svg>
  </button>
  ```
  The difference here is the use of `<button>` which allows us to also remove `tabindex` and `role` from this element.
- Use `role="presentation"` on the `<li>`. I'm not sure that this is causing any issues, but since we're changing the role of `<ul>` to `listbox`, and putting the role `option` on the `<button>`, I would remove the `listitem` role from the `<li>` with `role="presentation"`.
- Use `aria-selected="true"` for the option that is currently selected (i.e. the visible text in the toggle, not the option with focus). This is something we're using in the latest [OptionsMenu in core](https://pf4.patternfly.org/components/OptionsMenu/examples/), but it needs to be tested (and if this works as expected, we can update the Select example in core, too)
  - BTW, [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox) talks about using `aria-activedescendent` as an alternative to moving focus among the option elements, which is currently how we have the Dropdown implemented. But none of the examples I've found of custom Select components use anything but `aria-activedescendent`. I'm hoping that we can not use this, and continue managing focus as you have it, and just rely on `aria-selected` to indicate the selected value in the menu.
  - You would use this instead of `aria-pressed`. At the time that the Select component was implemented, the plan was to not manage focus, and aria-pressed was a way to indicate which button was in the selected state. But since we're managing focus, and using `role="listbox"`, `aria-selected` would be the attribute to use.

jgiardino:
For the keyboard interaction, it seems consistent with the Dropdown, which is what I expect. There is one aspect though where I _think_ we should modify the behavior slightly. After a selection is made in the menu (e.g. Enter keypress), can we return focus to the toggle? The idea here is that the user could hit Enter again to change their selection if needed. This would also be more consistent with native `<select>` elements. I'm not sure if this is needed in the Dropdown, although I don't think it would hurt.

Additionally, the [WAI-ARIA guide for listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox) includes these notes about which menu item should receive focus:

> When a single-select listbox receives focus:
>  - If none of the options are selected before the listbox receives focus, the first option receives focus. Optionally, the first option may be automatically selected.
>  - If an option is selected before the listbox receives focus, focus is set on the selected option.


@mcarrano Do you have any thoughts on these keyboard interactions?

kmcfaul:
> Use `aria-selected="true"` for the option

This is actually implemented, but when I renamed the option's value parameter I missed a replacement. Will add this fix in the next commit.

Will add the changes for the toggle focus and which item is focused on keyboard navigation as well.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
jgiardino: I commented on this below, but `aria-pressed` was an earlier implementation based on not managing focus, and was used as a way to indicate current selection.

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
jgiardino: Similar to a form field, a label should accompany this type of input.

The way the labels are handled in [core](https://pf4.patternfly.org/components/Select/examples/) is what I'd expect here, but I would also expect the consumer to decide if they want a visible or hidden label. For this implementation, I'm not sure what should be built into the component versus just demonstrated in an example. Is it possible to update the current example and add another example so that we show how you could implement this with both a visible and hidden label?

packages/patternfly-4/react-core/src/components/Select/Select.js
jgiardino: Assuming that a label is present—like "Title" for the example included in the preview—the aria-labelledby attributes would be defined like this for the following elements:

- toggle: aria-labelledby="[id of Title label][id of toggle]"
- listbox: aria-labelledby="[id of Title label]"

jgiardino:
The updates look great. I don't hear anything for `aria-selected` but it could just be that it isn't supported yet. I'd still leave it.

JAWS behavior is a little weird. It's great if you're already in forms mode (i.e. if this is in the context of a form), but if this is outside the context of a form, it doesn't place focus in the menu on Enter. But it does tell me to hit Space, so I hit Space once, then again, and then it works great! 😄 I think this has more to do with JAWS expecting this to be in the context of a form, and so I think this implementation works they way JAWS expects it to. I also tested it in VO and it works well there.

For a future variation, I'm still inclined that what's implemented for the first example of the [Select component in core](https://pf4.patternfly.org/components/Select/examples/) might be a variation we would want for cases where this component lives outside a form, but that would be a separate conversation/issue from this PR.

packages/patternfly-4/react-core/src/components/Select/Select.js
kmcfaul: There is a hidden span in core that the aria-labelledby of the toggle and listbox reference, but I'm not sure what actually would get filled there. Should this hidden span label change to the currently selected value?

jgiardino: The hidden span is intended to provide a label for the menu, similar to how a native `<select>` would have a label in a form. Since the toggle gets updated with the selected value, the `aria-labelledby` attribute references both the hidden label and the selected value to provide similar information that you'd hear with a native `<select>`. 

That said, I'm not sure what the best approach for implementing this should be. Consumers might want variations where a label is visible and variations where the label is not. And now I'm wondering if that label would be included separate from the component, so that the consumer has more control over how the label is presented, if visible. If that's the case, the consumer would need to specify `aria-labelledby` to reference the id of the toggle and the id of the label, visible or not. Is it possible for the consumer to just pass in the id of the label, and have the component automatically add `aria-labelledby="[id passed by consumer] [id generated for toggle by the component]"?

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
```diff
@@ -0,0 +1,147 @@
...
+          isHovered && styles.modifiers.hover,
+          isActive && styles.modifiers.active,
+          isPlain && styles.modifiers.plain,
+          className
+        )}
+        tabIndex={0}
+        type={type || 'button'}
+        onClick={_event => onToggle && onToggle(!isExpanded)}
+        aria-expanded={isExpanded}
+        aria-haspopup="listbox"
```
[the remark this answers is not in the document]
jgiardino: For this case, `aria-haspopup="listbox"` is correct, since the popup has `role="listbox"`. When the popup has `role="menu"` then this attribute can be either `aria-haspopup="true"` or `aria-haspopup="menu"` (i.e. the default popup role is a menu).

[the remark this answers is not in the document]
jgiardino: sorry for the late reply, I had made this comment last week, but didn't hit the Comment button 😞

jgiardino:
The update to include a label looks as expected. I tested in VO and it works as expected, too. :-)
Thanks!

packages/patternfly-4/react-core/src/components/Select/Select.js
jschuler: Is it an array of strings? Could be more specific, e.g. `PropTypes.arrayOf(PropTypes.string)`

kmcfaul:
@christiemolloy Sorry that was a bug, I hadn't set the width on the toggle, only the container.

@jschuler I'd anticipate an array of strings but I'm unsure of what the other multi-select structures will be so this prop is more confusing than necessary at this point. I will remove it.

packages/patternfly-4/react-core/src/components/Select/SingleSelect.js
tlabaj: Can this use the keyHandler function that Rebecca added to the helpers/util.ts file for AppLauncher and Dropdown to use?

packages/patternfly-4/react-core/src/components/Select/selectConstants.js
```diff
@@ -0,0 +1,17 @@
...
+  onSelect: () => {},
+  onClose: () => {}
+});
+
+export const SelectVariant = {
+  single: 'single'
+};
+
+export const KeyTypes = {
+  Tab: 'Tab',
```
tlabaj: Can you reuse the constants in the helpers/constants file here?

kmcfaul: Sure to both.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
dlabaj: Think we should get rid of children for now.  If this comes in as a separate request we could discuss this option further.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -0,0 +1,15 @@
+import { HTMLProps, FormEvent } from 'react';
+
+export interface SelectProps extends HTMLProps<HTMLOptionElement> {
+  isExpanded?: boolean;
+  onToggle(value: boolean): void;
+  placeholderText?: string;
+  selectOptions?: ReactNode[];
```
dlabaj: We should remove selectOptions and just support children instead of having two ways of doing the same thing.

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -0,0 +1,126 @@
...
+  /** Content rendered inside the Select */
+  children: PropTypes.node,
+  /** Classes applied to the root of the Select */
+  className: PropTypes.string,
+  /** Flag to indicate if select is expanded */
+  isExpanded: PropTypes.bool,
+  /** Placeholder text of Select */
+  placeholderText: PropTypes.string,
+  /** Array of SelectOption nodes that will be rendered */
+  selectOptions: PropTypes.array,
```
dlabaj: Remove selectOptions.

packages/patternfly-4/react-core/src/components/Select/SelectOption.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  value?: string;
+  isValid?: boolean;
+  isDisabled?: boolean;
+  isPlaceholder?: boolean;
+  onClick?: Function;
+  sendRef?: Function;
```
dlabaj: What's this used for?

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,104 @@
...
+  /** the value for the option */
+  value: PropTypes.string,
+  /** flag indicating if the option is disabled */
+  isDisabled: PropTypes.bool,
+  /** flag indicating if the option acts as a placeholder */
+  isPlaceholder: PropTypes.bool,
+  /** Optional on click callback */
+  onClick: PropTypes.func,
+  /** Callback for ref tracking */
+  sendRef: PropTypes.func,
```
dlabaj: Linting error saying it doesn't have a default prop.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,104 @@
...
+  isDisabled: false,
+  isPlaceholder: false,
+  onClick: Function.prototype
+};
+
+class SelectOption extends React.Component {
+  ref = React.createRef();
+
+  componentDidMount() {
+    this.props.sendRef(this.ref.current, this.props.index);
```
dlabaj: Linting error no props validation on index.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  onKeyDown = event => {
+    if (event.key === KeyTypes.Tab) return;
+    event.preventDefault();
+    if (event.key === KeyTypes.ArrowUp) {
+      this.props.keyHandler(this.props.index, 'up');
```
dlabaj: Linting error no props validation on keyHandler.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
```diff
@@ -0,0 +1,104 @@
...
+
+  render() {
+    const {
+      children,
+      className,
+      value,
+      onClick,
+      isDisabled,
+      isPlaceholder,
+      selected,
```
dlabaj: Linting error on selected.

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
```diff
@@ -0,0 +1,156 @@
...
+  /** Anything which can be rendered as dropdown toggle */
+  children: PropTypes.node,
+  /** Classes applied to root element of dropdown toggle */
+  className: PropTypes.string,
+  /** Flag to indicate if select is expanded */
+  isExpanded: PropTypes.bool,
+  /** Callback called when toggle is clicked */
+  onToggle: PropTypes.func,
+  /** Callback for toggle open on keyboard entry */
+  onEnter: PropTypes.func,
```
dlabaj: onEnter, onClose, type all missing default props.  Are they also needed for the d.ts file?

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
```diff
@@ -0,0 +1,156 @@
...
+        }}
+        className={css(
+          styles.selectToggle,
+          isFocused && styles.modifiers.focus,
+          isHovered && styles.modifiers.hover,
+          isActive && styles.modifiers.active,
+          isPlain && styles.modifiers.plain,
+          className
+        )}
+        type={type || 'button'}
```
dlabaj: Can just set 'button' as the default prop.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
kmcfaul: I don't think its bad to have alternative ways of declaring the objects. If we're set on only supporting one way, for SelectOption, should we only support children instead? That would match how DropdownItem behaves.

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
kmcfaul: Running yarn lint isn't throwing me any errors. Am I running it incorrectly? Could you post the specific error?

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
kmcfaul: Will add

packages/patternfly-4/react-core/src/components/Select/SelectToggle.js
kmcfaul: Good catch, will clean up

redallen:
Your Travis build is fine, just merge master to get the latest .travis.yml changes to ignore the [coveralls error](https://github.com/lemurheavy/coveralls-public/issues/632).

packages/patternfly-4/react-core/src/components/Select/SelectOption.js
kmcfaul: I've added the missing props and given them defaults. Let me know if its still throwing errors.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
kmcfaul: Though with select it makes more sense to use the prop. I'll make the changes as stated for the time being.

tlabaj:
@kmcfaul   Looking good.  I still don't understand why you are providing an interface for the SelectToggle.  It is an internal component.

kmcfaul:
Oh yeah, when I was first implementing the component I planned to follow Dropdown in that the user defines the toggle, but I found it wasn't really necessary and changed it to an internal component. I can remove that interface.

PR #1278, opened 2019-01-29
fix(LoginPage): add an option to display errors only on top of the form

Laviro:
Continued to the design refactoring [discussion](https://github.com/patternfly/patternfly-design/issues/671#issuecomment-457323229l)
where we decided to add an ability to pass the inputs errors to the top of the form,

**First glimpse:**
storybook

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
+++ b/packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
@@ -3,12 +3,13 @@ import PropTypes from 'prop-types';
 import Collapse from 'react-collapse';
 import { default as cx } from 'classnames';
 
-const LoginFormError = ({ children, show, className, ...props }) =>
-  children && (
-    <Collapse {...props} isOpened={show} className={cx('login-form-error', className)}>
+const LoginFormError = ({ children, show, topErrorOnly, className, ...props }) => (
+  <div className={cx('login-form-error', className)} style={topErrorOnly ? { minHeight: '20px' } : null}>
+    <Collapse {...props} isOpened={children !== null && show}>
       {children}
     </Collapse>
-  );
+  </div>
+);
 
 LoginFormError.propTypes = {
   /** Additional css classes */
@@ -16,13 +17,16 @@ LoginFormError.propTypes = {
   /** children nodes or text */
   children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   /** Controlls the collapse to open/close */
-  show: PropTypes.bool
+  show: PropTypes.bool,
+  /** if it`s a login page with topErrorOnly, add a fixed space for the form error */
+  topErrorOnly: PropTypes.bool
 };
 
 LoginFormError.defaultProps = {
   className: null,
   children: null,
-  show: false
+  show: false,
+  topErrorOnly: false
 };
 
 export default LoginFormError;
--- a/packages/patternfly-3/patternfly-react/sass/patternfly-react/_login-page.scss
+++ b/packages/patternfly-3/patternfly-react/sass/patternfly-react/_login-page.scss
@@ -4,5 +4,5 @@
 }
 
 .login-pf-input-warning {
-  margin-top: '5px';
+  margin-top: 5px;
 }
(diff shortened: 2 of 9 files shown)


mcarrano:
@Laviro can you give more detail on exactly what you changed here? Based on the discussion from last week with @Rohoover I thought that the plan was to just disable the Log In button until the username and password fields were filled in and then display errors in a fixed space that was allocated above the form. Does this do that?

Laviro:
yes @mcarrano, so the disable button PR is here https://github.com/patternfly/patternfly-react/pull/1188
and on this PR i remove the errors that are raised under the inputs, and move them to be shown only on the top of the form as discussed.

Laviro:
@jeff-phillips-18 could you explain your standpoints on the [discussion page](https://github.com/patternfly/patternfly-design/issues/671#issuecomment-457323229l), thanks

jeff-phillips-18:
I'm not disagreeing with the discussion nor the decision. I disagree with a breaking change in a minor release. Making it optional would get around the breaking change as long as the default is to continue to use the inputs error help block.

mcarrano:
@Laviro To @jeff-phillips-18 's point. I'm wary of making this global change to an existing component that is already being used in other products/projects. Before doing that we would need to review this design with a larger community. Is it possible to just make your customizations at the project level using the "Build your own" version of the login page provided?

mcarrano:
That would also be OK with me @Laviro .

Laviro:
@mcarrano @Rohoover @jeff-phillips-18 Please have a look on the storybooks examples.

mcarrano:
@Laviro so just to make sure I understand what you did... Looks like you added an new variation that moves the error message above the field leaving the existing login page in tact. Is that correct?

My only comment here is doesn't this still have the same problem you were trying to avoid? My thought was that you were going to allocate some fixed space above the form where the error message could live without needing to push context down. Did I misunderstand?

Laviro:
> Looks like you added an new variation that moves the error message above the field leaving the existing login page in tact. Is that correct?

Yes, that's correct.

> My only comment here is doesn't this still have the same problem you were trying to avoid? My thought was that you were going to allocate some fixed space above the form where the error message could live without needing to push context down. Did I misunderstand?

It is more elegant than 3 places where the error jumps.
We can remove the collapse animation..
though if it only happens on the top of the form,
It looks kinda smooth to me :)

Laviro:
@mcarrano added fixed-height to the `top form error` as we discussed offline,
ready for reviews :)

mcarrano:
@Laviro I am not seeing a difference from the last version I looked at.  Is this what I I should be viewing? https://1278-pr-patternfly-react-patternfly.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react%2FApplication%20Framework%2FLogin%20Page&selectedStory=Managed%20Login%20Page%20with%20errors%20display%20on%20top%20of%20the%20form&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

Laviro:
Test errors are non-related.
@mcarrano seems that the storybook deployment is broken 
and I can't also upload a storybook to github pages and send a link at the moment.. 

Can we start with the Javascript review ? :)

Laviro:
@mcarrano, I succeeded to upload the surge storybook manually: http://laviro.surge.sh/patternfly-3/index.html?knob-Social%20List%20Size=12&selectedKind=patternfly-react%2FApplication%20Framework%2FLogin%20Page&selectedStory=Managed%20Login%20Page%20with%20errors%20display%20on%20top%20of%20the%20form&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

packages/patternfly-3/patternfly-react/src/components/LoginPage/components/LoginCardComponents/LoginFormError.js
```diff
@@ -1,26 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Collapse from 'react-collapse';
-import { default as cx } from 'classnames';
 
-const LoginFormError = ({ children, show, className, ...props }) =>
-  children && (
-    <Collapse {...props} isOpened={show} className={cx('login-form-error', className)}>
+const LoginFormError = ({ children, show, topErrorOnly, ...props }) => (
+  <div className="login-form-error" style={topErrorOnly ? { minHeight: '20px' } : null}>
```
jeff-phillips-18: can we add an additional class rather than use inline styling to allow applications to override this setting?

Laviro: Thanks, fixed that

jeff-phillips-18:
LGTM. 

@mcarrano if/when you are satisfied with the UX, please remove `ux review` label and add `ux approved` label.

mcarrano:
This looks good @Laviro @jeff-phillips-18 I think we can merge!

Laviro:
@jeff-phillips-18 could you revisit ? :)

Issue #1243, opened 2019-01-24
Remove language selector from login page


rachael-phillips:
Referencing #986 

After speaking with @mcarrano and @tlabaj , we decided it would be best to remove the language selector from the login page for now and add it back in when the select list is available. Especially since the language selector on the login page is optional anyways.

tlabaj:
Closed by PR #1279

PR #1279, opened 2019-01-29
fix(LoginPage): Remove Language Selector from LoginPage

tlabaj:
#1243

**What**:

**Additional issues**:

Screenshot of Login form without Language selector.
image


mcarrano:
I cannot see the preview. But based on the screen shot above, this looks good to me.

PR #1282, opened 2019-01-30
fix(Table): Allow mixins of object, strings and numbers as row cell

karelhala:
**What**:
Allow different data types for row cells.
cells: [`String | Number | HTML | { title: String | Number | HTML }`]

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
@@ -2,10 +2,6 @@ import { FunctionComponent, HTMLProps, ReactNode } from 'react';
 import { OneOf, Omit } from '../../../../react-core/src/typeUtils';
 import { SortByDirection } from './SortColumn';
 import { DropdownPosition, DropdownDirection } from '@patternfly/react-core';
-export interface ISortBy {
-  index?: Number;
-  direction?: OneOf<typeof SortByDirection, keyof typeof SortByDirection>;
-}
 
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -35,10 +35,16 @@ export interface ICell {
...
+  title: ReactNode;
+  props: Object;
+}
+
 export interface IRow {
-  cells: Array<String>;
+  cells: Array<ReactNode | IRowCell>;
   isOpen: Boolean;
   parent: Number;
+  props: Object;
```
dmiller9911: this should be `object` capital `Object` is advised against. You might try `unknown` though. It is new and is better for these types of things.

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -35,10 +35,16 @@ export interface ICell {
   props: Object;
 }
 
+export interface IRowCell {
```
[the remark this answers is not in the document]
karelhala: Yeah, that is something that kinda stuck with me. I'll redo these interfaces.

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
karelhala: I was kinda thinking about `any` but at the same time trying to avoid it. `unknown` looks much better!

PR #1297, opened 2019-01-31
feat(Pagination): Introduce new PF4 pagination component

karelhala:
**What**:
New component Pagination based on PF-next styles

After brief discussion I chose to use dropdown on entire pagination instead of button to show dorop menu so we don't duplicate one thing on two places (dropdown and this implementation).

Closes: https://github.com/patternfly/patternfly-react/issues/1321

screenshot from 2019-01-31 15-16-58
screenshot from 2019-01-31 15-17-05


karelhala:
I had a discussion with @mcoker earlier about creating custom dropdown functionality for pagination which doesn't sound right. We came to an agreenment that we're going to use dropdown instead of custom logic, however I'd like to discuss this with @jgiardino as well.

tlabaj:
@karelhala Can you please reference issue #1321 in your Description.  Thanks!

mcoker:
hey @karelhala, sorry I may have misunderstood your original question :) Is using the dropdown temporary until the [options menu](https://pf4.patternfly.org/components/OptionsMenu/examples/) is created? Or do you disagree with the use of the Options Menu at all as the menu in pagination? I do think we should use the options menu here, my only concern with it was that currently the text ("1-20 of 20 items") isn't clickable from core, and only the down arrow beside it is. However, we could discuss making that text clickable in the react implementation of the options menu or maybe just in pagination? I'd like @jgiardino's feedback there, too, and she can also speak more to why we chose to make the text and button separate in the options menu toggle area.

karelhala:
@mcoker I didn't know that there's going to be new component as options component. This however brings new question as to why is there such component? I kinda don't like the idea of 2 components for esentially same thing (one with checkboxes the other without them). You see there needs to be a lot of logic behind opening and closing dropdowns, this logic will eventually be converted (copy + paste) to another component. I think that much better approach would be to revisit dropdown component (both from design and react implementation) with option to show checkboxes if such item is currently selected. I know that this is somewhat beyond this PR so let's either meet or discuss this somewhere else.

mcoker:
@karelhala sounds good. Fwiw, the choice to make the options menu a separate component was driven by the react implementation. We try to keep parity between the react and core components, and I asked either @jschuler or @dlabaj and was told this looks like a new component in react, so we made it a new component in core to match.

karelhala:
@/all Any update on this?

jgiardino:
Hey @karelhala! Sorry for the late review on this. 

Currently, the way the dropdown button html is structured in this preview, the button is announced as: "Items per page: Select, popup button" (in VO on Mac).

This means that text that is visible on the screen—"1 - 20 of 20 items"—isn't announced at all, which isn't ideal for screen reader accessibility.

We _could_ combine all of this text into one button label, but I feel like that's really overloading the button label. Yes the information is related, but it's not really specific to describing the action associated with the dropdown. This was why I had recommended to @mcoker that we separate these elements. Having a button with a concise label will help screen reader users find the button. And mouse users suffer a little in that the target size isn't as big, but I feel like that's a fair compromise.

karelhala:
jgiardino, @christiemolloy and @mcarrano  thank you for the review!
> Currently, the way the dropdown button html is structured in this preview, the button is announced as: "Items per page: Select, popup button"

What do you think about changing it to `1 - 20 of 20 items per page: Select, popup button`?

> * In Core I can't see/recall the options menu text having a hover state, did you add this? the numerical text also looks thicker than what we have in Core. Was this style added?

That is brought by dropdown menu, as described earlier I used dropdown because we'd have to duplicate navigation behavior and other things from dropdown component and that didn't felt good. In core it was also stated that options menu will change to be somehwat more aligned to dropdown as well. We can dive deeply into what is best option for us, but I wouldn't want to duplicate this control code because it looks like dropdown will be changed https://github.com/patternfly/patternfly-next/issues/1429

> * Could you change the title from "Page" to "Pagination"

This word is `of Pages` is optional and consumers can change it by passing one attribute. But I will default it to blank space just to be aligned with core. Thanks for poiting it out!

> Remove per page options that are not aplicable

I think we shouldn't limit users in selecting pagesize bigger than there is records.

> * The "items per page" menu is not correct. It should have the text "per page" following the number as shown in the design (see below).

That is interesting, I was using styles from pf4 examples [1] and there was no styling for this, do you have styling for such feature?

> * This seems to take up more horizontal space than the core version.

Yeah it looks sligthly bigger, it's actually 33 pixels wider (8 of them is because of how icons are rendered, sadly can't be changes). Rest is because it uses 16px font instead of 14px because we are using dropdown instead of options component (which will be replaced hopefuly in core design as well).

> * We also don't want the text "of 1 pages." We were trying to keep the width of this as compact as possible for when we include it with other toolbar items.

I understand your concerns, but we have issue in our backlog to include pages in paginator. It can be set by passing one attribute, but I will default it to nothing and if consumers want they can use this feature.

[1] https://pf4.patternfly.org/components/Pagination/examples/

mcarrano:
> > * The "items per page" menu is not correct. It should have the text "per page" following the number as shown in the design (see below).

> That is interesting, I was using styles from pf4 examples [1] and there was no styling for this, do you have styling for such feature?

@mcoker can you comment on this? It may be an oversight in the core CSS as I'm not seeing the correct styling there either. See the design mock I included above.

mcoker:
Sorry, that part was unclear. We introduced this as `.pf-c-pagination__menu-text`, but the way the options menu was introduced, we just included the default options menu in the example, so we aren't using it in the component. I can create an issue in core to update the example with this text in the menu.

mcarrano:
> > * This seems to take up more horizontal space than the core version.

> Yeah it looks sligthly bigger, it's actually 33 pixels wider (8 of them is because of how icons are rendered, sadly can't be changes). Rest is because it uses 16px font instead of 14px because we are using dropdown instead of options component (which will be replaced hopefuly in core design as well).

@jschuler can you comment on the spacing issue mentioned above? Why should the icons render differently in React?

karelhala:
@mcarrano let me elaborate on this. The difference is that in core there are `i` elements with specific font set to show icons. Whereas in here we are using svgs directly and to make the icon larger we just change the width and height of each icon. I think that for easier manipulation it was decided that each icon will take square shape instead of calculating it for each icon specificly.

jgiardino:
>> Currently, the way the dropdown button html is structured in this preview, the button is announced as: "Items per page: Select, popup button"

> What do you think about changing it to `1 - 20 of 20 items per page: Select, popup button`?

My concern with that approach is that the screen reader user has to listen to a lot of text before learning that this element is a button menu with options for changing the page size. Breaking the information up into separate elements allows screen reader users to skip the info they're not interested in to keep looking for the info/controls they're looking for. Whereas grouping a lot of info onto a button could result in burying this functionality and making it harder to find.

On the other hand, if the user is navigating by button type, they will know it's a button, and not just info. If the user is navigating by region or by each element, they will encounter the pagination section and might listen to all the text for each element if they are looking for a page size control. But I still think it makes discoverability for this control harder for screen reader users, and would still recommend breaking this information apart from the page size selection.

mcarrano:
@jgiardino thanks for the elaboration on this. As I recall, I think this is why we broke this apart like we did in the core implementation. Can we just follow the core implementation here? I'd rather not revisit this.

karelhala:
@jgiardino oh yeah, that makes sence! So I will change it that only the arrow will be part of dropdown and on the outside there will be span describing pagination.

karelhala:
@jgiardino @christiemolloy @mcarrano can you please re review once again?

As you can notice the dropdown context menu does not close when clicking on document and it also does not react to enter clicks and such. That's because this behavior is hardwired in toggle component that is not exposed and doc click + esc clicking should be part of context menu. Hopefully once we refactor dropdown this navigation will be part of context menu but as of right now it doesn't hve these features because I don't want to duplicate code..

christiemolloy:
> > * In Core I can't see/recall the options menu text having a hover state, did you add this? the numerical text also looks thicker than what we have in Core. Was this style added?
> 
> That is brought by dropdown menu, as described earlier I used dropdown because we'd have to duplicate navigation behavior and other things from dropdown component and that didn't felt good. In core it was also stated that options menu will change to be somehwat more aligned to dropdown as well. We can dive deeply into what is best option for us, but I wouldn't want to duplicate this control code because it looks like dropdown will be changed [patternfly/patternfly-next#1429](https://github.com/patternfly/patternfly-next/issues/1429)

jgiardino:
> @jgiardino oh yeah, that makes sence! So I will change it that only the arrow will be part of dropdown and on the outside there will be span describing pagination.

I'm not sure if I'm loading a cached version of the preview, but I'm still seeing the text defined as part of the dropdown:
image

Regarding this:
> Hopefully once we refactor dropdown this navigation will be part of context menu but as of right now it doesn't hve these features because I don't want to duplicate code.

I prefer this approach as well. I would like to refactor the dropdown so that we can manage this behavior in one component. But even if we decide to implement separate components, I expect the behavior and styles to be handled by some other component, and not defined here. 

After reviewing in a screen reader, I noticed that there are some `aria-label` attributes defined in core that shouldn't be there. I created https://github.com/patternfly/patternfly-next/issues/1453 to capture these issues. In this preview, it looks like you only have one of these: `<div class="pf-c-pagination__nav-page-select" aria-label="Current page 1 of 4">`. Can you remove this attribute? 

If you were to use this component in your application, do you think each page would have it's own url or not? I'm curious if this component should be flexible enough to allow the consumer to define the previous/next buttons as links and associate a url with those links. Let me know if you think this would be beneficial and even possible. If so, this can be a future enhancement.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
```diff
@@ -0,0 +1,112 @@
...
+          value={page}
+          onChange={(event) => {
+            let inputPage = Number.parseInt(event.target.value);
+            inputPage = isNaN(inputPage) ? page : inputPage;
+            inputPage = inputPage > lastPage ? lastPage : inputPage;
+            inputPage = inputPage < 1 ? 1 : inputPage;
+            onSetPage(event, isNaN(inputPage) ? page : inputPage)
+          }}
+        />
+        <span aria-hidden="true">of {lastPage} {titles.pages}</span>
```
ibolton336: This may be a good use case for the template string method added to helpers/utils.ts. We could pass in the entire string as a prop and use the last page & titles.pages values as embedded expressions.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
```diff
@@ -0,0 +1,112 @@
...
+  onSetPage,
+  onNextClick,
+  onPreviousClick,
+  onFirstClick,
+  onLastClick,
+  className,
+  ...props
+}) => {
+  return (
+    <nav className={css(styles.paginationNav, className)} aria-label="Pagination" {...props}>
```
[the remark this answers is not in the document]
karelhala: I don't think that it's a good idea to add specific aria-label prop, cause it would increase the number of props in prop table. If consumers want to add aria-label they probably know why they want it to be thee.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
```diff
@@ -0,0 +1,112 @@
...
+      <Button
+        variant={ButtonVariant.plain}
+        isDisabled={page === 1}
+        aria-label="Go to first page"
```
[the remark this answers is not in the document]
karelhala: Good catch. I'll update titles object with this as well.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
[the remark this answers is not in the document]
jgiardino: I'm not sure how this should be implemented, but it would be great if we could support the following use cases:

- By default, `aria-label` is set to "Pagination". This is a standard string we're recommending for this component. We also don't expect consumers to realize that this attribute is needed here, which is why we'd like to include something by default.
- If consumers want to provide translated strings for the product, they should be able to override the default value with their translated string. 

In other words, we'd like our components to be accessible by default without consumers having to do much, but we also want the flexibility to allow consumers to replace default text strings.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
```diff
@@ -0,0 +1,112 @@
...
+          min="1"
+          max={lastPage}
+          size="2"
+          value={page}
+          onChange={(event) => {
+            let inputPage = Number.parseInt(event.target.value);
+            inputPage = isNaN(inputPage) ? page : inputPage;
+            inputPage = inputPage > lastPage ? lastPage : inputPage;
+            inputPage = inputPage < 1 ? 1 : inputPage;
+            onSetPage(event, isNaN(inputPage) ? page : inputPage)
```
jschuler:
```suggestion
            onSetPage(event, inputPage)
```

packages/patternfly-4/react-core/src/components/Pagination/OptionsMenu.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+  };
+
+  renderItems = () => {
+    const { perPageOptions, onPerPageSelect, perPage } = this.props;
+    return perPageOptions.map(({ value, title }) =>(
+      <DropdownItem key={value}
+        component="button"
+        data-action={`per-page-${value}`}
+        className={css(styles.optionsMenuMenuItem, perPage === value && 'pf-m-selected')}
```
jschuler: modifier not available from the css object?

packages/patternfly-4/react-core/src/components/Pagination/OptionsMenu.js
```diff
@@ -0,0 +1,99 @@
...
+      itemCount,
+      firstIndex,
+      lastIndex,
+      perPage,
+      onPerPageSelect,
+      ...props
+    } = this.props;
+    return (
+      <div className={css(styles.optionsMenu, className)} {...props}>
+        <span id={`${widgetId}-label`} hidden>Items per page:</span>
```
jschuler: a11y text is a good default, but probably need to be able to pass in as well

packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CaretDownIcon } from '@patternfly/react-icons';
+import styles from '@patternfly/patternfly/components/OptionsMenu/options-menu.css';
+import { css, getModifier } from '@patternfly/react-styles';
+
+const OptionsToggle = ({ items, firstIndex, lastIndex, itemCount, widgetId, onToggle, isOpen }) => (
+  <div className={css(styles.optionsMenuToggle, getModifier(styles, 'plain'), getModifier(styles, 'text'))}>
+    <span className={css(styles.optionsMenuToggleText)}>
+      <b>{firstIndex} - {lastIndex}</b> of <b>{itemCount}</b> {items}
```
jschuler: Should use `strong` tag instead, also need to be able to customize I think. Using the template string builder from utility would be good

packages/patternfly-4/react-core/src/components/Pagination/Pagination.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const PaginationVariant: {
+  top: 'top';
+  bottom: 'bottom';
+  left: 'left';
+  right: 'right';
+};
+
+export interface PaginationProps extends HTMLProps<HTMLDivElement> {
+
```
jschuler: Needs props

packages/patternfly-4/react-core/src/components/Pagination/Pagination.docs.js
```diff
@@ -0,0 +1,17 @@
+import { Pagination, PaginationVariant } from '@patternfly/react-core';
+import PaginationTop from './examples/PaginationTop';
+import PaginationBottom from './examples/PaginationBottom';
+
+export default {
+  title: 'Page',
```
jschuler:
```suggestion
  title: 'Pagination',
```

packages/patternfly-4/react-core/src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,159 @@
...
+  perPageOptions: PropTypes.arrayOf(PropTypes.shape({
+    title: PropTypes.node,
+    value: PropTypes.number
+  })),
+  /** Total number of items. */
+  itemCount: PropTypes.number.isRequired,
+  /** Current page number. */
+  page: PropTypes.number,
+  /** First index of items on current page. */
+  itemsStart: PropTypes.number,
```
jschuler: not required but don't see in defaultProps, same with itemsEnd, amountOfPages, widgetId, dropDirection

packages/patternfly-4/react-docs/package.json
```diff
@@ -39,7 +39,7 @@
   "scripts": {
     "docbuild": "node build/copyDocs.js && env NODE_OPTIONS=--max_old_space_size=4096 gatsby build",
     "pr-build": "node build/copyDocs.js && env NODE_OPTIONS=--max_old_space_size=4096 gatsby build --verbose --prefix-paths",
-    "develop": "gatsby develop"
+    "develop": "gatsby develop -p 9999"
```
jschuler: Not a bad idea in general, just want to make sure this is intended

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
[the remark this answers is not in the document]
karelhala: @jgiardino that's exacly what is implemente here. If consumer didn't pass any `aria-label` to this component it will default to `Pagination` however if they are using some tranaslation strings or want to pass something specific they can choose to add their own attribute and it will be overwritten by their string.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
karelhala: At first I thought that it was a mistake as well. But if you look closely there is page canculation and for instance if somehow lastPage is `NaN` it will go trough propTypes definition and will be passed into `onSetPage`.

packages/patternfly-4/react-core/src/components/Pagination/OptionsMenu.js
karelhala: Sadly options menu does not have `selected` modifier.

packages/patternfly-4/react-docs/package.json
karelhala: Yeah, that was commited by mistake

packages/patternfly-4/react-core/src/components/Pagination/Pagination.js
karelhala: `itemsStart`, `itemsEnd` and `amountOfPages` is deliberate so consumers doesn't have to pass anything and best default value is undefined (in deault props I'd just assign `undefined` to these values).

On other hand `widgetId` and `dropDirection` should be in default props. Realy nice catch!

packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
```diff
@@ -0,0 +1,112 @@
...
+    pages: PropTypes.string
+  }),
+  page: PropTypes.number.isRequired,
+  onSetPage: PropTypes.func.isRequired,
+  onNextclick: PropTypes.func,
+  onPreviousClick: PropTypes.func,
+  onFirstClick: PropTypes.func,
+  onLastClick: PropTypes.func
+};
+const defaultProps = {};
```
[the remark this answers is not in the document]
karelhala: Yeah, I kinda forgot to put them here.

packages/patternfly-4/react-core/src/components/Pagination/Pagination.d.ts
karelhala: Um, yeah, that would be good. Nice catch!

karelhala:
I tried to address review requests as much as possible. The biggest challenge was all those aria labels, can you see if you like the API and such? About the template string where is it? I tried looking for it but couldn't find it. Is there any documentation how to use it?

ibolton336:
> I tried to address review requests as much as possible. The biggest challenge was all those aria labels, can you see if you like the API and such? About the template string where is it? I tried looking for it but couldn't find it. Is there any documentation how to use it?

There is no documentation on how to use it.  We should document it though. In The ChipGroup.js file, you can see on line 46 

`  const collapsedTextResult = fillTemplate(collapsedText, { remaining: React.Children.count(children) - 1 });`

karelhala:
@ibolton336 oh I see, but I don't think that this will be applicable to this use case. You see that function produces only string, but consumers should be able to use html elements as well. If I'd use this function I would have to use [#dangerouslysetinnerhtml](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) and that doesn't feel quite correct. But we could provide another way. Either react node or string. If string then we'd use `fillTemplate` if not then we would use that thing as component and pass attributes to it.

packages/patternfly-4/react-core/src/components/Pagination/OptionsMenu.js
tlabaj: @mcoker can we add this modifier to the Options Menu?  Seems like we would want that.  We can open up another react issue to make the update if core end up adding that.

mcarrano:
Everything I can see here works correctly, however it would be good to rework the example so it represents more than 2 pages. In the current example I can't really test everything. For example, with only two pages the first/last and previous/next buttons have the same effect. Is that possible?

The only small bug that I saw was that there is a space missing after the word "of."

screen shot 2019-03-04 at 12 04 09 pm

jschuler:
This looks pretty good, some of the aria labels and strings could be loosened up a bit more to make it more friendly to customization and other languages but I don't think these are blockers.

karelhala:
I updated the number of records and default pagination label, but looks like it is not updated in surge.

mcarrano:
@karelhala Thanks for including more records. It helps to be able to test how this works. In clicking around, I did uncover one bug however. Seems that if you click to the last page and then increase the number of items per page from the options menu, the math is getting screwed up somehow. Below is a screen shot of the result.

screen shot 2019-03-05 at 4 19 32 pm

There is also a minor deviation of the options in the menu from the original design. Each item should have the text "per page" following it, i.e. "10 per page." I noticed this is also missing from the core implementation. @mcoker is there a reason we did not implement that? See the second screen shot below.

screen shot 2019-03-05 at 4 22 04 pm

mcoker:
@mcarrano I have an issue open to update the pagination example with the "per page" text. You can do it now with the component, it just isn't in the example. https://github.com/patternfly/patternfly-next/issues/1443

karelhala:
> I did uncover one bug however. Seems that if you click to the last page and then increase the number of items per page from the options menu, the math is getting screwed up somehow.

@mcarrano that is caused by the example function. We don't want to hold the state and calculate page based on some complex logic. The only way the consumer interacts with this component is by passing functions for page change and per page change. They calculate current page, per page option and such bythemselves and pass these things to pagination by props. Eventually we don't want to do some complex logic on per page select because eventually this component will be connected to (server) API and this API will dictate current page and per page options.

Some examples I can think of with this approach
1) per page 100, page 15/15. User selects per page 10 - here consumer should recalculate page based on per page. Some might want to go to first page (everytime per page changes) some would want to calculate current page by some complex logic.
2) per page 10, page 150/150. User selects per page 100 - consumer might want to go to first page or calculate current page to 15.

You see we have to let consumers freedom of choice what to do, otherwise we'd have to implement one pagination variant for per page select to go to first page, or to go to some complex calculation. Of course we can do that (it's not that hard) but we **shouldn't**. This is supposed to be library for simple presentational components. So any state calculation should be done wih caution.

mcarrano:
Thanks for the explanation @karelhala . That all makes sense. Is it possible to update the options menu to read "10 per page," 20 per page," etc. per the design? Other than that, I think this is ready to merge.

tlabaj:
@karelhala do want us to open another issue to update the example as @mcarrano mentioned?  We can do that and merge this a is.

tlabaj:
@mcarrano @karelhala I opened issue #1528 To address remaining feedback.

Issue #1267, opened 2019-01-28
Refactor About Modal


tlabaj:
Refactor About Modal to match core once issue [#1110](https://github.com/patternfly/patternfly-next/issues/1110) is resolved.
Changes will include:

- Trademark information should be moved to the bottom of the panel and scroll with the rest of the about content. 
- The logo name should shift up to fill the gap.

tlabaj:
Closed by #1307

PR #1307, opened 2019-02-01
refactor(AboutModal): Upgrade core to 1.0.175 and Refactor AboutModal

tlabaj:
refactored AboutModal and removed heroImg and heroImageSrc and heroImageAlt props

#1267

**What**:

**Additional issues**:


rhamilto:
With the removal of `heroImageSrc`, what's the recommended solution for Webpack?  As it stands, Webpack does not bundle images from patternfly-core.  cc: @jeff-phillips-18

jeff-phillips-18:
@tlabaj what is the intent here? How do application set the heroImageSrc now?

tlabaj:
@jeff-phillips-18 The heroImage is being set by css now.

PR #1313, opened 2019-02-04
fix(Table): Update table styles based on new styling from pf-next v 175

karelhala:
**What**:
Because we updated to newer version of PF-next we have to update collapsible and sort implementation.

#### Before collapsible
screenshot from 2019-02-04 12-20-10
#### After collapsible
screenshot from 2019-02-04 12-19-55

#### Before sortable
screenshot from 2019-02-04 12-22-18
#### After sortable
screenshot from 2019-02-04 12-22-26

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/BodyWrapper.js
+++ b/packages/patternfly-4/react-table/src/components/Table/BodyWrapper.js
@@ -1,10 +1,22 @@
-import React, { Component } from 'react';
+import React, { Component, Children, Fragment } from 'react';
 import styles from '@patternfly/patternfly/components/Table/table.css';
 import { css } from '@patternfly/react-styles';
+import { mapOpenedRows } from './utils/headerUtils';
 
-const BodyWrapper = (rows) => {
+const BodyWrapper = (rows, onCollapse) => {
   class TableBody extends Component {
     render() {
+      if (onCollapse) {
+        return (
+          <Fragment>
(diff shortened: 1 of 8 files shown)


karelhala:
@mcarrano this issue is being resolved in different PR https://github.com/patternfly/patternfly-react/pull/1314/files#diff-ca0af699a3940b14197204af202c2469R28 but I ported some of these changes here as well. Let's hope merging won't break.

karelhala:
@mcarrano because previous PR was merged sorted cells should have correct color.

karelhala:
Thanks @christiemolloy for the review! The collapsible example is indeed similiar to "Expandable Table" in core, however I wanted to show that you can chain parents to achieve treeview table (that would actually need updates core) and that row can expand more than one row. If you want I can remove these examples, but I think we should add issue in core to allow consumers these things as well.

christiemolloy:
@karelhala I believe you should remove these examples and make an issue in Core to address them. Also will you remove the toggle icon on this example as visually it is not placed correctly: 

screen shot 2019-02-08 at 9 29 55 am

mcarrano:
@karelhala Yes, to @christiemolloy 's point, we should remove the tree-view functionality for now as this requires further design consideration and implementation in Core. What we intended here was only simple one-level expansion. Filing a new issue in patternfly-next would be the best way to get this on the feature roadmap.

PR #1314, opened 2019-02-04
fix(Table): Use direct imports for styles

karelhala:
**What**:
Re-adds changes made in this PR https://github.com/patternfly/patternfly-react/pull/1237 because they were removed in https://github.com/patternfly/patternfly-react/pull/1304 

**Additional issues**:


martinpovolny:
I wonder if you are testing the latest patternfy with the insights-chrome?

Each time I upgrade PF I spend several hours figuring what versions work together.

My understanding is that the Insights is the main user of PF4 for now. So it would make sense to me to test the 2 together.

Issue #946, opened 2018-11-20
There should be no space above the Progress bar if description/title is not passed in as prop


tlabaj:
There should be no space above the Progress bar if description/title is not passed in as prop.
Add an example  in the docs to the show how this would display.

This issue will be addressed in core by issue #[961](https://github.com/patternfly/patternfly-next/issues/961) 
React will need to be updated to match cores examples.

**The final implementation should be reviewed by CSS**

PR #1335, opened 2019-02-06
feat(Progress): add single line progress handling and example

kmcfaul:
**What**: If title is not passed in or set to null, the grid gap should not be applied.

**Issue**: #946


packages/patternfly-4/react-core/src/components/Progress/Progress.js
```diff
@@ -82,6 +82,7 @@ class Progress extends Component {
           getModifier(styles, variant, ''),
           getModifier(styles, measureLocation, ''),
           getModifier(styles, measureLocation === ProgressMeasureLocation.inside ? ProgressSize.lg : size, ''),
+          !title && getModifier(styles, 'singleline', ''),
```
jschuler: this should also work when title defaults to empty string

kmcfaul: Will update. I changed the default of title to null, though a user could still pass an empty string.

jschuler: What I meant was I think you can leave the default to empty string, because in js `!'' === !null`. So shouldn't have to add additional null checks

kmcfaul: Oh I see, good catch. I'll update it

packages/patternfly-4/react-core/src/components/Progress/Progress.js
```diff
@@ -93,7 +94,7 @@ class Progress extends Component {
         <ProgressContainer
           parentId={this.id}
           value={scaledValue}
-          title={title}
+          title={title || ''}
```
jschuler: Can also revert this change since title defaults to empty string

PR #1343, opened 2019-02-07
docs(siteConstants.js): PF3: Link to react-bootstrap legacy docs instead of new (incompatible) docs

mturley:
**What**:

react-bootstrap's v1 release targets bootstrap v4, which is incompatible with patternfly.
The new documentation they released breaks all of our deep links into the old docs from Storybook.

This legacy URL comes from the "To find the documentation for the latest Bootstrap 3 compatible
release, go here" link at https://github.com/react-bootstrap/react-bootstrap.

### DIFF
--- a/storybook/constants/siteConstants.js
+++ b/storybook/constants/siteConstants.js
@@ -3,7 +3,7 @@ const PATTERNFLY_ORG = 'http://www.patternfly.org/';
 export const BASE_URL = {
   PATTERNFLY_ORG,
   PATTERNFLY_LIBRARY: `${PATTERNFLY_ORG}pattern-library/`,
-  REACT_BOOTSTRAP: 'https://react-bootstrap.github.io/'
+  REACT_BOOTSTRAP: 'https://5c507d49471426000887a6a7--react-bootstrap.netlify.com/'
 };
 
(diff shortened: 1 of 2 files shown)


storybook/constants/siteConstants.js
```diff
@@ -3,7 +3,7 @@ const PATTERNFLY_ORG = 'http://www.patternfly.org/';
 export const BASE_URL = {
   PATTERNFLY_ORG,
   PATTERNFLY_LIBRARY: `${PATTERNFLY_ORG}pattern-library/`,
-  REACT_BOOTSTRAP: 'https://react-bootstrap.github.io/'
+  REACT_BOOTSTRAP: 'https://5c507d49471426000887a6a7--react-bootstrap.netlify.com/'
```
jeff-phillips-18: Yuck!

mturley: Yeah, I was disappointed to find they didn't have a cleaner URL for their legacy docs. This is what I found in their readme.

mturley:
Force-pushed an eslint fix that was failing CI. That's what I get for assuming I don't need to run tests for a docs change :)

PR #1345, opened 2019-02-07
fix(nav): Allow nav to expand by prop change

### LINKED ISSUE #1016: NavExpandable does not expand or collapse programmatically
If the NavExpandable component prop `isActive` is set to true, it would be great to have an option to expand it automatically. 

Setting the value of `isExpanded` to true or false does not expand or collapse the nav list items. When NavToggle is manually selected, it calls `this.setState({ value: !value });` from within `handleToggle`. This successfully toggles the NavExpandable open and close. We would like to be able to set the value of `isExpanded` programmatically by passing/setting props. Currently, this option isn't available.

nav bug

ibolton336:
**What**: fixes #1016 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
+++ b/packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
@@ -20,7 +20,11 @@ class NavToggle extends React.Component {
   state = {
     value: this.props.defaultValue
   };
-
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.defaultValue !== this.state.value) {
+      this.setState({ value: nextProps.defaultValue });
+    }
+  }
   handleToggle = e => {
     // Item events can bubble up, ignore those
     if (e.target.getAttribute('data-component') !== 'pf-nav-expandable') {


rhamilto:
@ibolton336, this fix appears to have introduced a bug where previously expanded sections that have been collapsed are re-expanded on next navigate.

yikrsqwdlu

cc: @spadgett, @janwright73

packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
```diff
@@ -20,7 +20,11 @@ class NavToggle extends React.Component {
   state = {
     value: this.props.defaultValue
   };
-
+  componentWillReceiveProps(nextProps) {
```
priley86: this lifecycle method is soon to be deprecated. It works with our target version of React, but I recommend looking at `getDerivedStateFromProps` and `componentDidUpdate`.

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops
https://egghead.io/lessons/react-refactor-componentwillreceiveprops-to-getderivedstatefromprops-in-react-16-3

Issue #1163, opened 2019-01-11
Expand button in Table and DataList components


jgiardino:
The expand buttons in the Table and DataList are similar patterns, and should have similar semantics associated with them.

The following aria attributes for the expand button should be used in both of these components:
- **`aria-controls="[id of controlled element]"`** -  This should be used when the controlled contents are not adjacent to the toggle that controls them, as in the Table and DataList. In some screen readers, the user is provided with a key command to shift focus to the controlled element. While the documentation allows multiple id values, there has been some questioning of this, since the benefit is in shifting focus to a specific element. Léonie Watson describes how this attribute should be used in an [article](https://tink.uk/using-the-aria-controls-attribute/), and also commented on the benefits and only using a single id in the second comment to this [article](http://www.heydonworks.com/article/aria-controls-is-poop).
  - This should be applied to both components. **UPDATE** This will be deferred for the Table component
- **`aria-label="Details"`** - This provides an accessible name for the button
  - This is currently defined in the Table component, but needs to be updated in the DataList component
- **`aria-labelledby="[id of row header] [id of expand button]"`** - This combines the row header / list item heading text with the button label to provide an accessible name that is context specific (e.g. "Node 1 Details" versus just "Details")
  - This is currently defined in the Table component, but the order of `id`s specified in the DataList component needs to be updated. 
- **Allow for translated strings** - The consumer should also be able to provide a translated string in this case. 
  - This could be split out as a separate issue if needed.
  - If the consumer provides a translatable string—e.g. `"{itemHeading} Details"`—then this should be defined as the value for `aria-label` and `aria-labelledby` MUST be removed.

jgiardino:
There was some discussion of `aria-controls` in #821. I'm pasting comments from that here:

> With the aria-controls, that's not gonna be easy. Parents do not know about which children they open (To allow asynchronous loading of collapsed children). And what if there are multiple items collapsed on click <button aria-label="Details" aria-expanded="true" aria-controls=[id1 id2 id3] class="pf-c-button pf-m-plain pf-m-expanded" type="button">? However I have added id to expanded row content.

rebeccaalpert:
@jgiardino - It looks like the DataList button lets you put in whatever you want for the aria-label. Should we take that control away from users or do you want "Details" as the default?

jgiardino:
I was wondering if it's possible to have a default implementation that works, but to allow consumers to provide a custom string.

So by default, the button would be this, and announced as "Node 1 Details":
[...]

PR #1347, opened 2019-02-07
fix(DataList): Add aria-label and aria-labelledby improvements

rebeccaalpert:
The DataListToggle aria label now defaults to "Details." If the consumer provides an aria label, there is no aria-labelledby. If the consumer uses the default aria label, aria-labelledby will be in the
following format: aria-labelledby=`${aria-labelledby of DataListItem} ${id}`. Users can also add an id prop to DataListContent and a corresponding aria-controls prop to DataListToggle.

Affects #1163.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
+++ b/packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
@@ -3,13 +3,13 @@ import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import styles from '@patternfly/patternfly/components/DataList/data-list.css';
 import { EllipsisVIcon } from '@patternfly/react-icons';
-import { Button } from '../Button';
+import { Button, ButtonVariant } from '../Button';
 
 class DataListAction extends React.Component {
   render() {
     const {className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, rowid, ...rest} = this.props;
     return <div className={css(styles.dataListAction, className)} {...rest}>
-      <Button variant="plain" id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
+      <Button variant={ButtonVariant.plain} id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
         <EllipsisVIcon />
       </Button>
     </div>;
--- a/packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
+++ b/packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
@@ -5,16 +5,13 @@ import styles from '@patternfly/patternfly/components/DataList/data-list.css';
 import { EllipsisVIcon } from '@patternfly/react-icons';
 import { Button, ButtonVariant } from '../Button';
 
-class DataListAction extends React.Component {
-  render() {
-    const {className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, rowid, ...rest} = this.props;
-    return <div className={css(styles.dataListAction, className)} {...rest}>
-      <Button variant={ButtonVariant.plain} id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
-        <EllipsisVIcon />
-      </Button>
-    </div>;
-  }
-}
+const DataListAction = ({ className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, rowid, ...props }) => (
+  <div className={css(styles.dataListAction, className)} {...props}>
+    <Button variant={ButtonVariant.plain} id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
+      <EllipsisVIcon />
+    </Button>
+  </div>
+);
 
 DataListAction.propTypes = {
   /** Content rendered inside the DataList list */
--- a/packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
+++ b/packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
@@ -5,13 +5,16 @@ import styles from '@patternfly/patternfly/components/DataList/data-list.css';
 import { EllipsisVIcon } from '@patternfly/react-icons';
 import { Button } from '../Button';
 
-const DataListAction = ({ className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, ...props }) => (
-  <div className={css(styles.dataListAction, className)} {...props}>
-    <Button variant="plain" id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
-      <EllipsisVIcon />
-    </Button>
-  </div>
-);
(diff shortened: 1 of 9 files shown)


packages/patternfly-4/react-core/src/components/DataList/DataListToggle.js
```diff
@@ -8,13 +8,15 @@ import { Button } from '../Button';
 const DataListToggle = ({
   className,
   isExpanded,
+  'aria-controls': ariaControls,
   'aria-label': ariaLabel,
   'aria-labelledby': ariaLabelledBy,
+  rowid,
```
jgiardino: I'm seeing `rowid` display as an attribute in almost every child html element of `li.

For example, in this snippet, you can see `rowid` which shouldn't be there. However the `id` with the same value on "Primary content" should be there.
```
<div class="pf-c-data-list__cell" rowid="ex-item1">
  <div id="ex-item1">Primary content</div>
  <a href="#">link</a>
</div>
```

jgiardino:
The updates look great! There was one weird issue I noticed that I commented on inline.

I also tested the following item from the issue, and this works as described. 

* If the consumer provides a translatable string—e.g. `"{itemHeading} Details"`—then this should be defined as the value for `aria-label` and `aria-labelledby` MUST be removed.

But I'm also wondering what the typical expectation is regarding how we handle components when html attributes are defined along with props. Is there a general rule of thumb we're following for this, or is it handled on a component-by-component basis. For example, I tested what would happen if I added both `aria-label` and `aria-labelledby` to `<DataListToggle>`. The `aria-label` appears on the html element but not the `aria-labelledby`. I'm not sure if this is a case we need to support, but I'm also not sure if this behavior would be considered inconsistent with other implementation 🤷‍♀️ @tlabaj Do you have any recommendations on this?

packages/patternfly-4/react-core/src/components/DataList/DataListToggle.js
rebeccaalpert: Yeah, that's because you have to pass the id down from the parent to all the children. Let me see if I can restructure the div to show everything except that prop.

rebeccaalpert:
@jgiardino - I amended the commit with a change to the div. The rowid prop shouldn't be visible to the end user anymore.

rebeccaalpert:
@jgiardino - I amended the commit with some changes to the components involved. The rowid prop shouldn't be visible in the HTML anymore. Let me know if you're still seeing it anywhere.

packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
```diff
@@ -5,13 +5,16 @@ import styles from '@patternfly/patternfly/components/DataList/data-list.css';
...
 import { Button } from '../Button';
 
-const DataListAction = ({ className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, ...props }) => (
-  <div className={css(styles.dataListAction, className)} {...props}>
-    <Button variant="plain" id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
-      <EllipsisVIcon />
-    </Button>
-  </div>
-);
+class DataListAction extends React.Component {
```
tlabaj: why was this converted to a class?

packages/patternfly-4/react-core/src/components/DataList/DataListCell.js
```diff
@@ -3,14 +3,17 @@ import { css, getModifier } from '@patternfly/react-styles';
...
 
-const DataListCell = ({ children, className, width, ...props }) => (
-  <div
-    className={css(styles.dataListCell, width > 1 && getModifier(styles, `flex_${width}`, ''), className)}
-    {...props}
-  >
-    {children}
-  </div>
-);
+class DataListCell extends React.Component {
```
tlabaj: why was this converted to a class?

packages/patternfly-4/react-core/src/components/DataList/DataListContent.js
```diff
@@ -3,22 +3,28 @@ import PropTypes from 'prop-types';
...
-  <section
-    className={css(styles.dataListExpandableContent, className)}
-    hidden={isHidden}
-    aria-label={ariaLabel}
-    {...props}
-  >
-    {children}
-  </section>
-);
+class DataListContent extends React.Component {
```
tlabaj: why was this converted to a class?

packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
rebeccaalpert: I converted these all to classes so I could render everything except the rowid prop in the HTML. The original implementation was showing the rowid prop and @jgiardino flagged it.

packages/patternfly-4/react-core/src/components/DataList/DataListToggle.js
```diff
@@ -8,13 +8,15 @@ import { Button } from '../Button';
...
   id,
   ...props
 }) => (
     <div className={css(styles.dataListToggle, className)} {...props}>
-      <Button id={id} variant="plain" aria-label={ariaLabel} aria-labelledby={ariaLabelledBy} aria-expanded={isExpanded}>
+      <Button id={id} variant="plain" aria-controls={ariaControls !== '' && ariaControls} aria-label={ariaLabel} aria-labelledby={ariaLabel !== "Details" ? null : `${rowid} ${id}`} aria-expanded={isExpanded}>
```
tlabaj: can you import and use the ButtonVariant object from Button and use that instead of using "plain".

packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
```diff
@@ -5,13 +5,16 @@ import styles from '@patternfly/patternfly/components/DataList/data-list.css';
...
+  render() {
+    const {className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, rowid, ...rest} = this.props;
+    return <div className={css(styles.dataListAction, className)} {...rest}>
+      <Button variant="plain" id={id} aria-labelledby={ariaLabelledBy} aria-label={ariaLabel}>
```
tlabaj: can you import and use the ButtonVariant object from Button and use that instead of using "plain".

packages/patternfly-4/react-core/src/components/DataList/DataListToggle.js
rebeccaalpert: I added a new commit with this change.

packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
jschuler: If you destructure the rowid prop out of the ...props it should not show up in the HTML.
```
const DataListAction = ({ className, id, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, **rowid**, ...props })
```

rebeccaalpert: Thanks, I made the change.

PR #1350, opened 2019-02-07
build(sass): Inline the CSS of imported files in patternfly-react.scss

### LINKED ISSUE #673: Distribution css brings along 404 out of box
I've put together a simple demo app that illustrates the problem in a basic use case.

https://github.com/seanforyou23/pf-react-styling

In a nutshell the problem is that, when loading patternfly-react.css stylesheet in standard fashion through a link tag on a webpage, it fails, because it contains an unreliable reference to files that are not guaranteed to exist - Typeahead.css - as its not shipped along with our distribution assets. 

In the demo app I mentioned, you'll see I also included references to several other popular ui styling packages such as patternfly core, bootstrap, foundation and materialize so we can compare the results. Patternfly-react styles load just fine, just that they bring along with them the extra baggage of a 404 on Typeahead.

screen shot 2018-09-27 at 4 28 00 pm

Of course this reference may resolve just fine depending on your environment.. but the point is that this is a very simple yet critical use case that should work without issue - out of the box.

Should we run pre-process import references for third party libs like this before final bundles are produced? Or maybe load them over http? I like the idea of keeping this package self contained, so hoping we can find a way to parse these types of dependencies ahead of time, like we do other assets.

mturley:
Fixes #673.

This PR adds a step to the `build:sass` script in `packages/patternfly-3/patternfly-react` which uses a custom node-sass importer to change the tilde-based path `~react-bootstrap-typeahead/css/Typeahead.css` to a true relative path, and then calls `cleancss` to convert that local import into inlined CSS.

Details on the `--importer` option I'm passing to `node-sass`: https://github.com/sass/node-sass#importer--v200---experimental

Before writing the custom `build/node-sass-tilde-importer.js` file, I tried using https://www.npmjs.com/package/node-sass-tilde-importer, but that converts the import to an absolute path, which cleancss was refusing to resolve. Maybe a different CSS processor could handle that.

I have a feeling there is a better way, but this works. It seems something like this is now necessary because of an upstream change in node-sass: https://github.com/sass/node-sass/issues/2362

@seanforyou23 I'm curious about your thoughts on this solution.

### DIFF
--- a/packages/patternfly-3/patternfly-react/package.json
+++ b/packages/patternfly-3/patternfly-react/package.json
@@ -62,10 +62,11 @@
     "build:babel:cjs": "cross-env BABEL_ENV=production:cjs babel src --out-dir dist/js",
     "build:babel:esm": "cross-env BABEL_ENV=production:esm babel src --out-dir dist/esm",
     "build:less": "shx mkdir -p dist/less && shx cp -r less/* dist/less",
-    "build:sass": "shx mkdir -p dist/sass && shx cp -r sass/patternfly-react/* dist/sass && cross-env node-sass --output-style compressed --include-path sass --include-path $npm_package_sassIncludes_patternfly --include-path $npm_package_sassIncludes_bootstrap --include-path $npm_package_sassIncludes_fontAwesome -o dist/css sass/patternfly-react.scss",
+    "build:sass": "shx mkdir -p dist/sass && shx cp -r sass/patternfly-react/* dist/sass && cross-env node-sass --output-style compressed --include-path sass --include-path $npm_package_sassIncludes_patternfly --include-path $npm_package_sassIncludes_bootstrap --include-path $npm_package_sassIncludes_fontAwesome --importer=build/node-sass-tilde-importer.js -o dist/css sass/patternfly-react.scss && cleancss -o dist/css/patternfly-react.css dist/css/patternfly-react.css",
     "clean": "rimraf dist"
   },
   "devDependencies": {
+    "clean-css-cli": "^4.2.1",
     "react-axe": "^3.0.2",
     "rimraf": "^2.6.2",
     "shx": "^0.3.2"
--- a/packages/patternfly-3/patternfly-react/build/node-sass-tilde-importer.js
+++ b/packages/patternfly-3/patternfly-react/build/node-sass-tilde-importer.js
@@ -1,0 +1,8 @@
+function importer(url) {
+  if (url[0] === '~') {
+    return { file: `../../../../../node_modules/${url.substr(1)}` };
+  }
+  return { file: url };
+}
+
+module.exports = importer;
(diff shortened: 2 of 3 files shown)


mturley:
I'm thinking maybe https://github.com/postcss/postcss-import is a better choice than `cleancss`. But I'd prefer to not have to add any new dependencies at all...

packages/patternfly-3/patternfly-react/package.json
```diff
@@ -63,10 +63,10 @@
     "build:babel:cjs": "cross-env BABEL_ENV=production:cjs babel src --out-dir dist/js",
     "build:babel:esm": "cross-env BABEL_ENV=production:esm babel src --out-dir dist/esm",
     "build:less": "shx mkdir -p dist/less && shx cp -r less/* dist/less",
-    "build:sass": "shx mkdir -p dist/sass && shx cp -r sass/patternfly-react/* dist/sass && cross-env node-sass --output-style compressed --include-path sass --include-path $npm_package_sassIncludes_patternfly --include-path $npm_package_sassIncludes_bootstrap --include-path $npm_package_sassIncludes_fontAwesome -o dist/css sass/patternfly-react.scss"
+    "build:sass": "shx mkdir -p dist/sass && shx cp -r sass/patternfly-react/* dist/sass && cross-env node-sass --output-style compressed --include-path sass --include-path $npm_package_sassIncludes_patternfly --include-path $npm_package_sassIncludes_bootstrap --include-path $npm_package_sassIncludes_fontAwesome --importer=build/node-sass-tilde-importer.js -o dist/css sass/patternfly-react.scss && cleancss -o dist/css/patternfly-react.css dist/css/patternfly-react.css"
```
seanforyou23: It fails for me locally, are there any new steps involved in running a build that I've maybe missed?

screen shot 2019-02-13 at 12 19 47 pm

mturley: Hm, I might have installed `cleancss` globally and referenced that by mistake. Did you run `yarn install`? Maybe we need to use the direct path to cleancss in node_modules. Do you know how the `node-sass` call works here without being a path to node_modules?

seanforyou23: Yea I did a full rebuild with the repo. I haven't had a chance to dig deep into the changes, just got so far as running it and seeing the failure. You can try `git clean -xfd` to blow away all the generated files and directories, and maybe uninstall cleancss globalls, then run a clean install and try again? In case it still works for you and I've goofed this up, let me know and I'll be happy to try again.

mturley: Will do. Thanks for taking a look.

mturley: @seanforyou23 give it a try now. I think I had cleancss in the wrong package.json file. Monorepo problems

dgutride:
@mturley  - ping me about this one?

PR #1357, opened 2019-02-08
fix(docs): bump react docs version

dgutride:
**What**:

**Additional issues**:


packages/patternfly-4/react-docs/package.json
```diff
@@ -1,11 +1,10 @@
 {
   "name": "@patternfly/react-docs",
   "description": "PatternFly React Docs",
-  "version": "3.21.17",
```
redallen: need to add "private": true,

PR #1359, opened 2019-02-08
refactor(ApplicationLauncher): Refactor so it no longer uses Dropdown

### LINKED ISSUE #1241: App Launcher
We now have a separate component for the app launcher so we need to update this to no longer use the dropdown component. 

### App Launcher
https://pf-next.com/components/AppLauncher/examples/
### Core PR
https://github.com/patternfly/patternfly-next/pull/1239

rebeccaalpert:
Refactored ApplicationLauncher into its own component. It still uses DropdownItem and dropdown contexts so as not to break client code.

Fixes #1241.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
+++ b/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
@@ -7,6 +7,7 @@ import Toggle from './Toggle';
 import { ThIcon } from '@patternfly/react-icons';
 import { ApplicationLauncherDirection, ApplicationLauncherPosition } from './applicationLauncherConstants';
 import { DropdownContext } from '../Dropdown/dropdownConstants';
+import GenerateId from '../../helpers/GenerateId/GenerateId';
 
 export const propTypes = {
   /** Additional element css classes */
@@ -26,9 +27,6 @@ export const propTypes = {
   /** Adds accessible text to the button. Required for plain buttons */
   'aria-label': PropTypes.string
 };
-
-// seed for the aria-labelledby ID
-let currentId = 0;
 
 export const defaultProps = {
   className: '',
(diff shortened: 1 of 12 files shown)


packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
```diff
@@ -19,33 +22,71 @@ export const propTypes = {
...
   onToggle: PropTypes.func,
   /** Indicates where menu will be alligned horizontally */
-  position: PropTypes.oneOf(Object.values(DropdownPosition)),
+  position: PropTypes.oneOf(Object.values(ApplicationLauncherPosition)),
   /** Adds accessible text to the button. Required for plain buttons */
   'aria-label': PropTypes.string
 };
 
+// seed for the aria-labelledby ID
+let currentId = 0;
```
jschuler: Is this incremented correctly when you have multiple of these on the same page? I think you'll want to use the generateId util instead

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherMenu.js
```diff
@@ -0,0 +1,130 @@
...
+    } else {
+      custom
+        ? (this.refsCollection[`item-${nextIndex}`].current.focus &&
+            this.refsCollection[`item-${nextIndex}`].current.focus()) ||
+          ReactDOM.findDOMNode(this.refsCollection[`item-${nextIndex}`].current).focus()
+        : this.refsCollection[`item-${nextIndex}`].focus();
+    }
+  };
+
+  onKeyDown = event => {
```
tlabaj: Can we make a constant in the helpers/constants files for the string "up", "ArrowUp" etc.  This keys will be used in multiple places.

Can we also create a helper function in our utils.ts file for the keyHandler.  It is used in multi components.  Make sure that you comment that the custom argument is optional.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
rebeccaalpert: It looks like it's incremented correctly. This is how Dropdown does it. (It's not new code.)

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
```
tlabaj: I think we want to decouple the Dropdown component completely from Dropdown here.  So we want to create a new ApplicationLauncher Item component.  If there is any common functionality that can be reused.  Maybe move it into our helper utilities.

tlabaj:
Follow up issues have been open to 1) 

1. address the refactoring of common code. issue #1374 
2. address refactoring the component to not use dropdownItems #1375

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
```diff
@@ -19,33 +23,69 @@ export const propTypes = {
...
+          isOpen && styles.modifiers.expanded,
+          className
+        )}
+        ref={ref => {
+            this.parentRef = ref;}}>
+        {Children.map(
+          <Toggle id={`pf-toggle-id-${randomId}`} aria-label={ariaLabel} onToggle={onToggle}><ThIcon /></Toggle>, oneToggle =>
+            cloneElement(oneToggle, {
+              parentRef: this.parentRef,
+              id: randomId,
```
tlabaj: should id here match line 63?

rebeccaalpert: I think so. That's how it was originally.

Issue #1362, opened 2019-02-11
PF3: TypeAheadSelect CSS includes `height: fit-content` rule which is only supported by Chrome


mturley:
The CSS we add on top of react-bootstrap-typeahead includes `height: fit-content`.
Seen here in our SCSS file: https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/sass/patternfly-react/_type-ahead-select.scss#L8
and here in our LESS file: https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/less/type-ahead-select.less#L8

According to https://developer.mozilla.org/en-US/docs/Web/CSS/height ,
`fit-content` is only supported in Chrome. My project needs to support IE11, and needs this typeahead component.

I tracked down where this line was originally added in PR https://github.com/patternfly/patternfly-react/pull/409, to fix the issue https://github.com/patternfly/patternfly-react/issues/413. ~~I'm a little confused, because when I remove this `fit-content` rule from the CSS, I still can't reproduce the issue. Maybe we don't need it at all? @Laviro, if you have time maybe you can help me figure out what's going on there.~~ Ok, I can now reproduce the issue in Storybook at https://1363-pr-patternfly-react-patternfly.surge.sh/, but I can't seem to reproduce it in my downstream project... 🤔 I'll try and figure that out separately, but we should find an IE11-friendly way we can fix the bug instead of using `fit-content`.

mturley:
@jeff-phillips-18, @dgutride said I should mention you here to make you aware of the potential IE11 compatibility issue.

jeff-phillips-18:
Thanks for the heads up @mturley

Laviro:
Should this be closed due to the fix in #1369 ?

mturley:
Should we close issues before the fixes are merged? The PR has the magic "fixes" text in it, when it's merged this will close itself.

Laviro:
Oops, I saw it on my phone and thought this was another PR.

mturley:
Haha, no worries! Also, sorry I summoned you in this issue and then realized I was just being dumb.

PR #1363, opened 2019-02-11
fix(TypeAheadSelect): Remove `height: fit-content` rule not supported by IE11

mturley:
Related to issue #1362. It's possible we don't need this rule. PR is just to create a preview link for
testing.


mturley:
Ok, the bug is reproducible in Storybook with this rule removed. We'll need to address #1362 with another solution, closing this one for now.

PR #1368, opened 2019-02-11
PF4 Table:  'Select All' checkbox updates

### LINKED ISSUE #1352: PF4 Table: Select all does not respond properly
When all rows of the table are selected, I expect the select-all checkbox to be selected automatically. Likewise, when the select-all checkbox is selected, and I deselect one or more rows, I expect the select-all check box to be deselected automatically. I don't appear to have a way to access the select-all checkbox directly in order to set it myself?

When the select-all checkbox is selected, the debugger shows all the rows have the property `selected=true`. The select-all checkbox simply fails to respond to the new row selections.

Note that all children are also assigned the `selected=true` property, which is unexpected since they do not have checkbox selections. This forces me to check if each row has a `parent` property when gathering all selected rows. Otherwise, I can mistakenly gather children as selections.

Example:
chrome-capture

dtaylor113:
When all rows of the table are selected, the select-all checkbox is now selected automatically. Likewise, when the select-all checkbox is selected, and I deselect one or more rows, the select-all check box is deselected automatically.

Fixes #1352

pf4-table

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.js
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.js
@@ -141,7 +141,18 @@ class Table extends React.Component {
     this.state = {
       headerData: []
     };
+    this.isSelected = this.isSelected.bind(this);
+    this.areAllRowsSelected = this.areAllRowsSelected.bind(this);
   }
+
+  isSelected(row) {
+    return row.selected === true;
+  }
+
+  areAllRowsSelected(rows) {
+    return rows.every(this.isSelected);
+  }
+
   render() {
     const {
       caption,
@@ -169,6 +180,7 @@ class Table extends React.Component {
       sortBy,
       onSort,
       onSelect,
+      allRowsSelected: onSelect ? this.areAllRowsSelected(rows) : false,
       actions,
       onCollapse,
       rowLabeledBy,
--- a/packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.js
+++ b/packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.js
@@ -7,7 +7,7 @@ export default (
   label,
   {
     column: {
-      extraParams: { onSelect, rowLabeledBy = 'simple-node' }
+      extraParams: { onSelect, allRowsSelected, rowLabeledBy = 'simple-node' }
     },
     rowIndex,
     rowData
@@ -21,9 +21,10 @@ export default (
     };
   }
   const rowId = rowIndex !== undefined ? rowIndex : -1;
+
   function selectClick(event) {
     const selected = rowIndex === undefined ? event.target.checked : rowData && !rowData.selected;
-    onSelect && onSelect(selected, selected, rowId);
+    onSelect && onSelect(event, selected, rowId);
   }
   const customProps = {
     ...(rowId !== -1
@@ -32,6 +33,7 @@ export default (
           'aria-labelledby': rowLabeledBy + rowIndex
         }
       : {
+          checked: allRowsSelected,
           'aria-label': 'Select all rows'
         })
   };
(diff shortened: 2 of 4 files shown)


packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -139,9 +139,29 @@ class Table extends React.Component {
...
+      });
+    }
+  }
+
+  isSelected(row) {
+    return row.selected === true;
+  }
+
+  areAllRowsSelected(rows) {
+    return rows.every(this.isSelected);
```
dlabrecq: Should we skip expanded row children?

It's odd that both parent and child have an selected property, because they could be initialized differently and/or be out of sync. For example, setting the selected property on a child has no affect on its parent.

dtaylor113: Ok, I'll have to look into it, get up to speed on how the expand w/ children is working.

dlabrecq: If I weed them out myself, will rows.every(this.selected) still work?

karelhala: I don't think it's good idea to skip expanded children. There is option to show checkbox for expanded children as well, that might be handy in future once we figure out tree tables.

dlabrecq: Sounds reasonable. Thanks for the explanation

packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -139,9 +139,29 @@ class Table extends React.Component {
...
     this.state = {
-      headerData: []
+      headerData: [],
+      allRowsSelected: props.onSelect ? this.areAllRowsSelected(props.rows) : false
     };
+    this.isSelected = this.isSelected.bind(this);
+    this.areAllRowsSelected = this.areAllRowsSelected.bind(this);
   }
+
+  componentDidUpdate(prevProps) {
```
karelhala: I don't really like `componentDidUpdate` function, it could lead to infinite loop (because of calling set state in it's body). Might be better to use [static-getderivedstatefromprops](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops).

karelhala: After further investigation, I don't think we need to store `allRowsSelected` in state. Rather send
```JS
const headerData = calculateColumns(cells, {
 // ...
 allRowsSelected: this.props.onSelect ? this.areAllRowsSelected(this.props.rows) : false
})
```

packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.js
```diff
@@ -21,9 +21,10 @@ export default (
     };
   }
   const rowId = rowIndex !== undefined ? rowIndex : -1;
+
   function selectClick(event) {
     const selected = rowIndex === undefined ? event.target.checked : rowData && !rowData.selected;
-    onSelect && onSelect(selected, selected, rowId);
+    onSelect && onSelect(event, selected, rowId);
```
karelhala: Good catch!

karelhala:
This PR calls for new UX design as tri-state checkbo https://github.com/patternfly/patternfly-next/issues/1411

packages/patternfly-4/react-table/src/components/Table/Table.js
dtaylor113: > I don't really like `componentDidUpdate` function, it could lead to infinite loop

It does have a conditional check

dtaylor113:
> This PR calls for new UX design as tri-state checkbo [patternfly/patternfly-next#1411](https://github.com/patternfly/patternfly-next/issues/1411)

I think that should be handled in a follow on PR

packages/patternfly-4/react-table/src/components/Table/Table.js
dtaylor113: Fixed

dtaylor113:
@karelhala and @dlabrecq, I believe this PR is ready, pls review when u get a moment -thanks

Issue #1138, opened 2019-01-07
Modal `isLarge` is confussing


boaz0:
Since there is no Wizard in PF next yet, it was decided to use Modal for that.
The problem we have right now is that `isLarge` set the `max-width` attribute of the Modal window while I was expecting it sets the `width` attribute to max.

I need this because in each step in the wizard a form is rendered. Some steps have only two fields and others have description text too. This makes different steps change the size of the Modal window which hurts UX. See [here](https://access.redhat.com/insights/platform/storybook/?selectedKind=Components&selectedStory=Wizard) what I mean

Can we do something about it?

tlabaj:
@boaz0   The link you provided is not working.

mcoker:
Created https://github.com/patternfly/patternfly-next/issues/1202 to allow for this

boaz0:
@tlabaj -- I updated the link. Can you try it again [here](https://access.redhat.com/insights/platform/storybook/?selectedKind=Components&selectedStory=Wizard)

tlabaj:
Closed by PR #1372

PR #1372, opened 2019-02-12
fix(modalbox): add the small modifier to modalbox

### LINKED ISSUE #1291: Update modal examples, add support for width prop
This is a result of https://github.com/patternfly/patternfly-react/issues/1138

The PR: https://github.com/patternfly/patternfly-next/pull/1278

The modal is now full-width by default, and you can specify a `width` now if you want that could add an inline `width` to the element. 

We now have a `pf-m-sm` modifier that sets a small `max-width` on the modal (the same as the old default `max-width` with no modifier), and left the `pf-m-lg` modifier as it was.

boaz0:
**What**:

closes #1291 and #1138 

//cc @mcoker can you take a look to make sure I didn't miss anything. Thanks.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.js
@@ -23,19 +23,25 @@ const propTypes = {
   actions: PropTypes.any,
   /** A callback for when the close button is clicked */
   onClose: PropTypes.func,
+  /** Default width of the Modal. */
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   /** Creates a large version of the Modal */
   isLarge: PropTypes.bool,
+  /** Creates a small version of the Modal */
+  isSmall: PropTypes.bool,
   /** Additional props are passed and spread in the Modal body container <div> */
   '': PropTypes.any
 };
 
 const defaultProps = {
+  width: null,
   className: '',
   isOpen: false,
   hideTitle: false,
   actions: [],
   onClose: () => undefined,
-  isLarge: false
+  isLarge: false,
+  isSmall: false
 };
 
 let currentId = 0;
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.d.ts
@@ -6,6 +6,7 @@ export interface ModalProps extends HTMLProps<HTMLDivElement> {
   className?: string;
   hideTitle?: boolean;
   isLarge?: boolean;
+  isSmall?: boolean;
   isOpen?: boolean;
   onClose?: Function;
   title: string;
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.docs.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.docs.js
@@ -1,11 +1,18 @@
 import { Modal } from '@patternfly/react-core';
 import Simple from './examples/SimpleModal';
+import Width from './examples/WidthModal';
 import Large from './examples/LargeModal';
+import Small from './examples/SmallModal';
 
 export default {
   title: 'Modal',
   components: {
     Modal
   },
-  examples: [{ component: Simple, title: 'Simple Modal' }, { component: Large, title: 'Large Modal' }]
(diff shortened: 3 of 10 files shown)


mcoker:
thanks @boaz0, looks like the preview isn't working so I can't verify visually. though looking at the code, I'm not sure if it should be part of this PR or not, but this change was also made to enable users to pass a `width` prop to the modal component with a custom width that would be used without either `isSmall` (`.pf-m-sm`) or `isLarge` (`.pf-m-lg`) per the issue you opened https://github.com/patternfly/patternfly-react/issues/1138. Per that issue, now if you want a full width modal, you would just omit `isSmall` and `isLarge`, because the modal will now be 100% width by default.

boaz0:
@mcoker fixed that: https://1372-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/modal

mcoker:
Looks like there is an extra div between `div.pf-l-bullseye` and `div.pf-c-modal-box` that is causing an alignment issue. Is it possible to get rid of that or will that always be there with this component? If it will always be there, I'll need to make a small adjustment in core. 

screen shot 2019-02-13 at 10 30 28 am

boaz0:
@mcoker unfortunately I can't remove this. This is one of the "side-effects" of [`FocusTrap`](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Modal/ModalContent.js#L53) to catch clicks outside the modal.

boaz0:
@dlabaj I fixed that. Thanks for your review.

redallen:
Your Travis build is likely fine, just merge master and push to get the latest CI changes.

boaz0:
thanks @redallen - did git rebase.

tlabaj:
@boaz0 Can you please rebase again please.  Some changes were made in Master to fix failing Build. Thanks.

mcoker:
@boaz0 gotcha. Is there any way to add a class or otherwise style the generated element? If not we're going to need to reconsider components we currently center using the `pf-l-bullseye` layout.

boaz0:
@mcoker looking at the documents, `<FocusTrap>` gets three props: `active`, `pasued` and `focursTrapOptions` .
None of those provide a way to set `className` or define the element that is wrapped over the modal.

So I did what you suggested and used `Bullseye` to center the modal.

mcoker:
Gotcha, thanks. I don't think it's ideal to have to use 2 `Bullseye`'s to center something like that. We have a CSS meeting tomorrow, and I'll add this to our agenda of things to discuss.

mcoker:
@boaz0 for now I think what you've done with 2 `Bullseye`s is fine.

is it common for a component like `<FocusTrap>` to allow for a `ClassName`? I'm curious if we can request this be added, so we could add the `.pf-l-bullseye` class to it?

boaz0:
Opened a ticket https://github.com/davidtheclark/focus-trap-react/issues/40 for that thing.

I will update you when I will have something.

boaz0:
Hi @mcoker I have good news. I took a look at the source code and it seems that I was wrong.
You can specify `props` like `id`, `className` or any HTML props and it will be added to the `div` element.

I am updating the code and tests and pushing it for your review.
Hope I didn't make you work too hard on this.

Thanks.

mcoker:
@boaz0 awesome!  Good to know, I'm sure we'll use `FocusTrap` in other places and will need that, too.

packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -23,19 +23,25 @@ const propTypes = {
   actions: PropTypes.any,
   /** A callback for when the close button is clicked */
   onClose: PropTypes.func,
+  /** Default width of the Modal. */
```
jschuler: Should we add clarification, something like `Use either the width or the isLarge/isSmall prop`

boaz0: I will check this out.
Although if I am not wrong `isLarge` and `isSmall` are setting `maxWidth` and `minWidth` to `Modal` respectively.

Thanks.

boaz0:
Awesome! :raised_hands:  Thank you!

PR #1385, opened 2019-02-14
feat(Icons): Add fa regular icons and update solid icons

karelhala:
**What**:
Because new icons were added to solid icons we should pull them and expose them for others as well.

Also add new package with regular icons (outlined solid) with prefix `outlined${IconName}` so we don't have any name clashes.

Overall adds 277 of new icons!


karelhala:
Btw we are still missing [FA-brands-free](https://fontawesome.com/icons?d=gallery&s=brands&m=free) icons. Which is another 414 icons.

PR #1400, opened 2019-02-18
build(fixes): Fix TravisCI PF3 doc deploys

redallen:
**What**: Fix deploying [PF3 docs](http://patternfly-react.surge.sh/patternfly-3/index.html) on publish. Also add root package version to allow `yarn add https://github.com/patternfly/patternfly-react.git`.

**Additional issues**:


lerna.json
```diff
@@ -20,7 +20,7 @@
   },
   "command": {
     "publish": {
-      "message": "chore(release): releasing packages",
+      "message": "chore(release): releasing packages [ci skip]",
```
dgutride: Nice.

PR #1402, opened 2019-02-18
feat(VncConsole): More possibilities to customize component

bond95:
**What**:

Added possibility to extract toolbar to any other component. Added scaling prop to VncConsole, and
enhance `textConnecting` props to use node type as well. And fixed disconnecting.

We need this features for console page in oVirt VM Portal: https://github.com/oVirt/ovirt-web-ui/pull/956

Example from oVirt VM Portal:
image

@mareklibra please, review

### DIFF
--- a/packages/patternfly-3/react-console/src/VncConsole/VncActions.js
+++ b/packages/patternfly-3/react-console/src/VncConsole/VncActions.js
@@ -1,30 +1,59 @@
 import React from 'react';
+import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import { MenuItem, Button, helpers } from 'patternfly-react';
 
 const { Dropdown } = Button;
 const { noop } = helpers;
 
-const VncActions = ({ textSendShortcut, textCtrlAltDel, onCtrlAltDel }) => (
-  <Dropdown bsStyle="default" title={textSendShortcut} id="console-send-shortcut" onClick={noop}>
-    <MenuItem eventKey="1" onClick={onCtrlAltDel}>
-      {textCtrlAltDel}
-    </MenuItem>
-  </Dropdown>
-);
+const VncActions = ({
+  textSendShortcut,
+  textCtrlAltDel,
+  textDisconnect,
+  onCtrlAltDel,
+  onDisconnect,
+  insertToolbarTo
+}) => {
+  const toolbar = (
+    <div>
+      <Dropdown bsStyle="default" title={textSendShortcut} id="console-send-shortcut" onClick={noop}>
+        <MenuItem eventKey="1" onClick={onCtrlAltDel}>
+          {textCtrlAltDel}
+        </MenuItem>
+      </Dropdown>
+      <Button bsStyle="default" onClick={onDisconnect}>
+        {textDisconnect}
+      </Button>
+    </div>
+  );
+
+  if (!insertToolbarTo) {
+    return toolbar;
+  }
+  return (
+    document.getElementById(insertToolbarTo) && ReactDOM.createPortal(toolbar, document.getElementById(insertToolbarTo))
+  );
+};
 
 VncActions.propTypes = {
+  onDisconnect: PropTypes.func.isRequired,
   onCtrlAltDel: PropTypes.func,
 
   textCtrlAltDel: PropTypes.string,
-  textSendShortcut: PropTypes.string
+  textSendShortcut: PropTypes.string,
+  textDisconnect: PropTypes.string,
+
+  insertToolbarTo: PropTypes.string // id of element where VncAction should be inserted
 };
 
 VncActions.defaultProps = {
   onCtrlAltDel: noop,
 
   textCtrlAltDel: 'Ctrl+Alt+Del',
-  textSendShortcut: 'Send Key'
+  textSendShortcut: 'Send Key',
+  textDisconnect: 'Disconnect',
+
+  insertToolbarTo: ''
 };
 
 export default VncActions;
--- a/packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
+++ b/packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
@@ -31,6 +31,7 @@ class VncConsole extends React.Component {
       path,
       encrypt,
       resizeSession,
+      scaleViewport,
       viewOnly,
       shared,
       credentials,
@@ -53,7 +54,7 @@ class VncConsole extends React.Component {
       this.rfb = new RFB(this.novncElem, url, options);
       this.addEventListeners();
       this.rfb.viewOnly = viewOnly;
-      this.rfb.scaleViewport = false; // if the remote session is smaller than HTML container, the view will be centered
+      this.rfb.scaleViewport = scaleViewport;
       this.rfb.resizeSession = resizeSession;
     } catch (e) {
       onInitFailed && onInitFailed(e);
@@ -62,17 +63,17 @@ class VncConsole extends React.Component {
   }
 
   componentWillUnmount() {
-    this.removeEventListeners();
     this.disconnect();
+    this.removeEventListeners();
+    this.rfb = undefined;
   }
 
-  disconnect() {
+  disconnect = () => {
     if (!this.rfb) {
       return;
     }
     this.rfb.disconnect();
-    this.rfb = undefined;
-  }
+  };
 
   onConnected = () => {
     this.setState({ status: CONNECTED });
@@ -95,11 +96,11 @@ class VncConsole extends React.Component {
     this.props.onSecurityFailure(e);
   };
 
-  removeEventListeners() {
+  removeEventListeners = () => {
     this.rfb.removeEventListener('connect', this.onConnected);
     this.rfb.removeEventListener('disconnect', this.onDisconnected);
     this.rfb.removeEventListener('securityfailure', this.onSecurityFailure);
-  }
+  };
 
   setNovncElem = e => {
     this.novncElem = e;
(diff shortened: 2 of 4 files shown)


packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
```diff
@@ -62,17 +63,17 @@ class VncConsole extends React.Component {
   }
 
   componentWillUnmount() {
-    this.removeEventListeners();
     this.disconnect();
+    this.removeEventListeners();
+    this.rfb = undefined;
```
mareklibra: With this change, the `RFB.disconnect()` is skipped. Can you please explain reason for that?

packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
```diff
@@ -62,17 +63,17 @@ class VncConsole extends React.Component {
...
   }
 
-  disconnect() {
+  disconnect = () => {
```
mareklibra: Why this change, please?

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
```diff
@@ -1,30 +1,60 @@
...
-    </MenuItem>
-  </Dropdown>
-);
+const VncActions = ({
+  textSendShortcut,
+  textCtrlAltDel,
+  textDisconnect,
+  onCtrlAltDel,
+  onDisconnect,
+  portalToolbarTo
```
mareklibra: The name is too specific for the `oVirt VM Portal`

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
```diff
@@ -1,30 +1,60 @@
...
+        {textDisconnect}
+      </Button>
+    </div>
+  );
+
+  if (!portalToolbarTo) {
+    return toolbar;
+  }
+  return (
+    document.getElementById(portalToolbarTo) && ReactDOM.createPortal(toolbar, document.getElementById(portalToolbarTo))
```
mareklibra: I don't think we should perform such manipulations. I believe change in the component composition should be preferred over this approach.

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
```diff
@@ -1,30 +1,60 @@
...
-  textSendShortcut: PropTypes.string
+  textSendShortcut: PropTypes.string,
+  textDisconnect: PropTypes.string,
+
+  portalToolbarTo: PropTypes.string
 };
 
 VncActions.defaultProps = {
   onCtrlAltDel: noop,
+  onDisconnect: noop,
```
mareklibra: This is a breaking change.
The consuming application is **silently** required to add implementation. Otherwise the button just appears there without any function.

packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
```diff
@@ -95,11 +96,11 @@ class VncConsole extends React.Component {
     this.props.onSecurityFailure(e);
   };
 
-  removeEventListeners() {
+  removeEventListeners = () => {
```
mareklibra: likewise

packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
```diff
@@ -143,11 +154,21 @@ class VncConsole extends React.Component {
     return (
       <div className={classNames('vnc-console', this.props.topClassName)}>
         {this.props.children}
-        <Toolbar.RightContent>{rightContent}</Toolbar.RightContent>
-        <Toolbar.Results>
-          {status}
-          {this.novncStaticComponent}
-        </Toolbar.Results>
+        {portalToolbarTo ? (
```
mareklibra: as commented above

packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
bond95: But no, it doesn't skipped. Line 66 perform disconnecting. And previously `onDisconnected` event doesn't work, because it removed all listeners before it disconnected.

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
bond95: Well, it calling `portalToolbarTo`, because if it set it do portal toolbar to some element, not because this is from VM Portal. But yeah, it better to rename.

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
bond95: Yes, but in that case appear problem with customization. For example, in VM Portal we need only one toolbar not two and we can't `Send Key` dropdown move to our toolbar, because console is fully under control of `VncConsole` component.

And I think this is most suitable solution.

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
bond95: Ok, I'll do `onDisconnect` props required, but this is was just copy from `onCtlAltDel`. So why in this case `onCtrlAltDel: noop` not required too?

packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
bond95: Because of context.

bond95:
@mareklibra please, review.

jeff-phillips-18:
@mareklibra Could you please take another look?

bond95:
@priley86 could you, please, look at this PR?

packages/patternfly-3/react-console/src/VncConsole/VncConsole.js
priley86: +1

priley86:
like other VncConsole PRs, this is hard to visualize. Can you possibly share an environment this is running w/ me? Or maybe it is easier to record a gif from your machine..

Also, do you plan to add PF4 variation? This is still using PF3 menus/buttons/etc..

bond95:
> like other VncConsole PRs, this is hard to visualize. Can you possibly share an environment this is running w/ me? Or maybe it is easier to record a gif from your machine..

captured (4)

> Also, do you plan to add PF4 variation? This is still using PF3 menus/buttons/etc..

Yes, but in separate PR.

bond95:
@priley86 Could you please take another look?

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
```diff
@@ -1,30 +1,59 @@
...
+        {textDisconnect}
+      </Button>
+    </div>
+  );
+
+  if (!insertToolbarTo) {
+    return toolbar;
+  }
+  return (
+    document.getElementById(insertToolbarTo) && ReactDOM.createPortal(toolbar, document.getElementById(insertToolbarTo))
```
priley86: does this need to be a portal? We don't typically use these for elements that will render in the typical DOM/page content area (they are more often used for appending modals/popovers/etc.). 

I'm not currently using this downstream, so I can't say for sure, but you may be able to just expose it as a simple stateless component...

packages/patternfly-3/react-console/src/VncConsole/__snapshots__/VncActions.test.js.snap
```diff
@@ -1,43 +1,65 @@
...
 exports[`VncActions renders correctly component hierarchy 1`] = `
-<DropdownButton
-  bsStyle="default"
-  id="console-send-shortcut"
-  onClick={[Function]}
-  title="My Send Shortcut description"
->
-  <MenuItem
-    bsClass="dropdown"
+<div>
```
priley86: additional div?

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
```diff
@@ -1,30 +1,59 @@
...
 VncActions.propTypes = {
+  onDisconnect: PropTypes.func.isRequired,
   onCtrlAltDel: PropTypes.func,
 
   textCtrlAltDel: PropTypes.string,
-  textSendShortcut: PropTypes.string
+  textSendShortcut: PropTypes.string,
+  textDisconnect: PropTypes.string,
+
+  insertToolbarTo: PropTypes.string
```
jeff-phillips-18: When would this be used? I don't see it being used here at all. 
These propTypes should all be commented.

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
bond95: > does this need to be a portal?

Yes, this component is depend on VncConsole component and for right work should only be used inside of VncConsole. So from my point of view portals are the best solution.

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
bond95: > When would this be used? I don't see it being used here at all.

It is used in VncActions.

https://github.com/patternfly/patternfly-react/blob/223e582264d215dd6341bd87507a2774b4b0e4b0/packages/patternfly-3/react-console/src/VncConsole/VncActions.js#L30-L35

packages/patternfly-3/react-console/src/VncConsole/__snapshots__/VncActions.test.js.snap
bond95: No, `div` should be there. https://github.com/patternfly/patternfly-react/blob/223e582264d215dd6341bd87507a2774b4b0e4b0/packages/patternfly-3/react-console/src/VncConsole/VncActions.js#L17-L28

packages/patternfly-3/react-console/src/VncConsole/VncActions.js
bond95: > These propTypes should all be commented.

Added comment.

bond95:
@priley86 could you, please, look one more?

priley86:
merging - please be sure to file any follow up issues after testing downstream further.

thanks @bond95 !

Issue #1256, opened 2019-01-25
JS lint npm script broken for patternfly-4 packages


TimoSta:
The npm script `lint:js`, used to lint JavaScript code in the Patternfly 4 packages, and defined in packages/patternfly-4, is currently broken:

```
"lint:js": "eslint --max-warnings 0 **/*.js",
```
[_Source_](https://github.com/patternfly/patternfly-react/blob/62059db60b719fd09164deff327b43f2a9c5eb39/packages/patternfly-4/package.json#L29)

The glob used to select files is picked up correctly in Windows environments, and running the script there results in around 1500 errors and warnings. See the attached lint output:

[**pf4-lint-js.log**](https://github.com/patternfly/patternfly-react/files/2798099/pf4-lint-js.log)

However, in Linux/OS X environments, this glob seems to mach no files, resulting in zero linter errors during the Travis build.

---

Wrapping the glob in quotes resolves this issue, and problems are reported correctly under Linux/OS X as well:

```
"lint:js": "eslint --max-warnings 0 \"**/*.js\"",
```

---

The [ESLint docs](https://eslint.org/docs/user-guide/command-line-interface) give some more details on how to work with globs in ESLint:

> Please note that when passing a glob as a parameter, it will be expanded by your shell. The results of the expansion can vary depending on your shell, and its configuration. If you want to use node glob syntax, you have to quote your parameter (using double quotes if you need it to run in Windows), as follows:
> `eslint "lib/**"`

priley86:
This is a remnant of pf3 unless I am mistaken. I think pf4 typescript/tslint solution should be well defined before attempting this...it does seem reasonable to have consistent eslint configs if we intend to not migrate all at once though.

PR #1408, opened 2019-02-19
build(package.json): use consistent React version 16.4

redallen:
**What**: Use React 16.4.0 consistently everywhere. Towards #1256 

**Additional issues**:


redallen:
Agreed, this is out of scope for now and should look to a bigger upgrade.

Issue #1268, opened 2019-01-28
Update Tabs Component to  match core


tlabaj:
Update the Tab Component to match core once issue [#1328](https://github.com/patternfly/patternfly-next/issues/1328) has been resolved.

> when a set of tabs is totally unique on the page they should be within a <nav> instead so that screen reader users can find them by browsing for navigational landmarks. The HTML would then be:

rachael-phillips:
HI @rebeccaalpert would this be something you could incorporate on what you're working on right now? Let me know what you think about this!

rebeccaalpert:
Yeah, that should be doable. Is it still needed, though? It looks like https://github.com/patternfly/patternfly-next/issues/1328 was closed rather than merged.

Issue #1370, opened 2019-02-11
Add tab content component


mcoker:
This removes the tab sections from the tabs component and moves them to their own component

https://github.com/patternfly/patternfly-next/pull/1371/files

cc @tlabaj

ibolton336:
@rebeccaalpert looks like your PR would close this one too #1413

PR #1413, opened 2019-02-20
feat(Tabs): Allow tabbing of content outside Tabs

### LINKED ISSUE #1364: PF4 Tabs: Lacks support for tabs in header
Per Kyle's high fidelity Insights mocks, we need to place the PF4 tabs inside the `header` tag of Cost Management for accessibility, while the main body is in another tag -- see image below.

At least one Insights app has implemented their own tabs in the header using HTML anchors. This isn't desirable because it doesn't have the same styling and behavior as the PF4 component.

The PatternFly core examples show the tab and tab children in different sections, which could be split into a `header` tag and main body because it's raw HTML. However, the PF4 tabs component won't allow me to do that because a Tab must be a child of the Tabs component.
https://pf4.patternfly.org/components/Tabs/examples/

We could probably style it to look like Kyle's mock, but child elements would also be contained within my `header` tag and/or PF4 header styles, which is not desirable.

That said, I want to do something more like the code below, where children can be defined separately (either by reference or by ID):

```
<section className="pf-l-page-header pf-c-page-header ...">
  <header className="pf-u-display-flex ...">
    <Title size={TitleSize.lg}>{t('overview.title')}</Title>
    <Tabs activeKey={this.state.activeTabKey} onSelect={this.handleTabClick}>
      <Tab eventKey={0} title="Tab item 1" tabChild={tabChild1}/>
      <Tab eventKey={1} title="Tab item 2" tabChild={tabChild2}/>
    </Tabs>
  </header>
</section>
<section className="pf-l-page__main-section pf-c-page__main-section">
  <TabChild id='tabChild1'><div>...</div></Tab>
  <TabChild id='tabChild2'><div>...</div></Tab>
</>
```

Kyle's mock: https://redhat.invisionapp.com/share/ENPJQTBXTAC#/screens/336412758
screen shot 2019-02-11 at 11 13 34 am

rebeccaalpert:
Updated TabContent and modified Tabs so that users can tab through data outside of the Tabs component. The Tab in control of the separated content just needs to be provided with the ref/id of the TabContent.

Also added accessible Tabs option (with nav/aria-label).

Fixes #1364, #1268, and #1597.


packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -49,8 +50,24 @@ class Tabs extends React.Component {
...
-  handleTabClick(event, eventKey) {
+  handleTabClick(event, eventKey, tabChild) {
     this.props.onSelect(event, eventKey);
+    // process any children outside of the component
+    if (tabChild) {
+      React.Children.map(this.props.children, (child, i) => {
+        document.getElementById(child.props.tabChild).hidden = true;
+      });
+      // most recently selected tabChild
+      document.getElementById(tabChild).hidden = false;
```
dlabrecq: We should test if the object returned is not undefined. A dev could potentially provide an incorrect ID.

Probably want to ensure we obtained a `TabChild` type as well. Otherwise, you cannot be certain the returned object has a `hidden` property.

packages/patternfly-4/react-core/src/components/Tabs/examples/TabChildren.js
```diff
@@ -0,0 +1,32 @@
...
+        <Tab eventKey={1} title="Tab item 2" tabChild="tab2Section"></Tab>
+        <Tab eventKey={2} title="Tab item 3" tabChild="tab3Section"></Tab>
+      </Tabs>
+      <div>
+        <TabChild eventKey={0} id="tab1Section">Tab 1 section</TabChild>
+        <TabChild eventKey={1} id="tab2Section">Tab 2 section</TabChild>
+        <TabChild eventKey={2} id="tab3Section">Tab 3 section</TabChild>
+      </div>
+    </React.Fragment>;
+  }
```
dlabrecq: Thinking this should be done by refs so it's more React-idiomatic. 

The link below describes a re-usability issue when using IDs / getElementById.

https://www.javascriptstuff.com/use-refs-not-ids/

"When you use refs, you get the node rendered by `this` component, and you can re-use your component as much as you want."

rebeccaalpert: Yeah, I had it working with refs, but I was getting a warning in the console from whatever builds the docs. As long as the warning's okay, I can add that back in.

dlabrecq: Depends on the warnings. You may need to add a `key` property if you're iterating with a map?

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
rebeccaalpert: Added these checks in new commit; let me know if that's not what you're looking for. I'll add the refs in a separate commit.

packages/patternfly-4/react-core/src/components/Tabs/examples/TabChildren.js
rebeccaalpert: `Warning: Failed prop type: Invalid prop components.TabChild of type object supplied to ComponentDocs, expected function.`

It think it's due to using forwarding refs: https://reactjs.org/docs/forwarding-refs.html

dlabrecq: That doesn't sound related to using refs Vs IDs, but something about how the `Tabs.docs.js` file is processed. I suspect this is a Type issue? Perhaps related to the use of `FunctionComponent` in `TabChild.d.ts`?

dlabrecq: Perhaps something do to with how `TabChildProps` extends `HTMLProps`? looks different than `TabsProps` in `tabs.d.ts`

rebeccaalpert:
I added a third commit that lets you use refs on TabChild components. Caveat: Tabs seems like it can't access the ref prop from its child Tab, so TabChildren can't be hidden in Tabs' ComponentDidMount. Hidden has to be passed in manually on the TabChild if it's needed on page load.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -49,8 +50,34 @@ class Tabs extends React.Component {
   id = getUniqueId();
   tabList = React.createRef();
 
-  handleTabClick(event, eventKey) {
+  handleTabClick(event, eventKey, tabChildId, tabChildRef) {
     this.props.onSelect(event, eventKey);
+    // process any children outside of the component
+    if (tabChildRef) {
+      React.Children.map(this.props.children, (child, i) => {
+        const section = ReactDOM.findDOMNode(child.props.tabChildRef.current);
```
dlabrecq: You may not be passing in the reference properly. You should be able to access the tab child object directly without searching the dom.

Check out this approach https://reactjs.org/docs/forwarding-refs.html

rebeccaalpert: Yup, that's what I was doing earlier and it causes warnings for the docs.

dlabrecq: I believe that's a Type issue we need to resolve. 

If the components property is expecting a certain type, it's possible TabChild.d.ts is not defined properly? Or, our docs generator needs to be more flexible -- I believe @jschuler worked on that?

As a test, you can try something like this below. If it works, then it's probably a Type issue.

```
components: {
    Tabs,
    Tab
    Tab,
    TabChild as any
  },
```

redallen:
Rebase on latest master to fix CircleCI errors.

kmcfaul:
Snapshots need to be updated but otherwise looking good

rebeccaalpert:
Rebased and refactored to work with Ian's fix for https://github.com/patternfly/patternfly-react/issues/1370.

packages/patternfly-4/react-core/src/components/Tabs/Tab.js
```diff
@@ -9,17 +9,40 @@ const propTypes = {
...
+  }
+  return React.forwardRef((props, tabContentRef) => {
+    return <TabContainer {...props} forwardRef={tabContentRef} />;
+  });
+}
+
+class Tab extends React.Component {
+  render() {
+    const { children, eventKey, tabContentId, tabContentRef, forwardRef, ...props } = this.props;
+    return <button {...props} ref={tabContentRef}>{children}</button>;
```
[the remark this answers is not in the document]
rebeccaalpert: eventKey is to eliminate this warning:
`index.js:2177 Warning: React does not recognize the `eventKey` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `eventkey` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

tabContentId is to eliminate this warning:
`index.js:2177 Warning: React does not recognize the `tabContentId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `tabcontentid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

forwardRef is to eliminate this warning:
`index.js:2177 Warning: React does not recognize the `forwardRef` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `forwardref` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

If there's some better way to do it, let me know.

packages/patternfly-4/react-core/src/components/Tabs/Tab.js
```diff
@@ -9,17 +9,40 @@ const propTypes = {
...
+  tabContentId: PropTypes.string,
+  /** child reference for case in which a TabContent section is defined outside of a Tabs component */
+  tabContentRef: PropTypes.object,
 };
 
 const defaultProps = {
   children: null,
-  className: ''
+  className: '',
+  tabContentId: '',
```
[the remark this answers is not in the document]
rebeccaalpert: It's used to generate an aria-controls label in Tabs. I added separate behavior for alternate content sections - it generates an aria-controls label with the id of the tabContent. If you don't want that, I can take it out.

packages/patternfly-4/react-core/src/components/Tabs/examples/SeparateTabContent.js
```diff
@@ -0,0 +1,36 @@
...
+      activeTabKey: tabIndex
+    });
+  };
+
+  render() {
+    return <React.Fragment>
+      <Tabs activeKey={this.state.activeTabKey} onSelect={this.handleTabClick}>
+        <Tab eventKey={0} title="Tab item 1" tabContentId="refTab1Section" tabContentRef={this.contentRef1}></Tab>
+        <Tab eventKey={1} title="Tab item 2" tabContentId="refTab2Section" tabContentRef={this.contentRef2}></Tab>
+        <Tab eventKey={2} title="Tab item 3" tabContentId="refTab3Section" tabContentRef={this.contentRef3}></Tab>
```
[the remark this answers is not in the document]
rebeccaalpert: See https://github.com/patternfly/patternfly-react/pull/1413#discussion_r265803120.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -50,8 +54,16 @@ class Tabs extends React.Component {
...
+  handleTabClick(event, eventKey, tabContentId, tabContentRef) {
     this.props.onSelect(event, eventKey);
+    // process any tab content sections outside of the component
+    if (tabContentRef) {
+      React.Children.map(this.props.children, (child, i) => {
+        child.props.tabContentRef.current.hidden = true;
+      });
+      // most recently selected tabContent
+      tabContentRef.current.hidden = false;
+    }
```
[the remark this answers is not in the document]
rebeccaalpert: Thanks, see new commit.

packages/patternfly-4/react-core/src/components/Tabs/Tab.js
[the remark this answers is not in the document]
dlabrecq: Ok, it appeared that tabContentId was coming from a different child.

Although, Tabs is generating a unique ID. 

`54: id = getUniqueId();`

Considering the dev already provides a unique eventKey, the aria-controls could be simplified to look more like this: 

`aria-controls={`pf-tab-section-${this.id}-${child.props.eventKey}``

packages/patternfly-4/react-core/src/components/Tabs/Tab.js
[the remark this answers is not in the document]
dlabrecq: Ok, it just looks like unused keys. Perhaps you can add a comment above, so no one accidentally removes it?

That said, I do see a couple warnings in the console. For example: 

Screen Shot 2019-03-15 at 10 07 27 AM

Adding a key property to each Tab child should be an easy fix.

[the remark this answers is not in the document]
rebeccaalpert: That warning is actually from Table (it's on master). I can open a separate issue for it. I'll add a comment for the destructuring. Joachim is supposed to look into why the docs generate that error.

rebeccaalpert:
Removed tabContentId and reverted aria-controls behavior to the default.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -149,65 +161,78 @@ class Tabs extends React.Component {
...
-              className={css(styles.tabsScrollButton)}
-              variant="plain"
-              aria-label={rightScrollAriaLabel}
-              onClick={this.scrollRight}
-            >
-              <AngleRightIcon />
-            </button>
-          )}
-        </div>
+        {label !== ''
```
tlabaj: I don't think we want to use the label here to determine if this is <nav> or not.  I think the consumer should decide what they want to use based on their use case. I suggest we have an isNav attribute that defaults to false and switch that way.  If isNav is true we should require the label.  I would also name it aria-label instead.

jgiardino:
Thanks for adding the `label` prop! Could you update the description to be something like this:

> provides an accessible label for the Tabs and defines the Tabs inside a navigation region (labels should be unique for each set of Tabs that are present on a page)

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
jgiardino: This was my suggestion. I was hoping to simplify the process for the consumer to define the tabs as a `<nav>` since an `aria-label` is necessary if they are inside a `<nav>` and also since `aria-label` is useless if they are inside a `<div>`.

rebeccaalpert:
I updated the label prop description and addressed the extra example/aria-label changes from https://github.com/patternfly/patternfly-react/issues/1597.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
tlabaj: I just wonder if the API makes sense this way?  Would it be intuitive for the consumer  to know that if they supply a label, it would change the internal structure of the Tab.  Also this approach does not prevent the consumer from adding an aria-label when it is a <div>.  You would have to have the prop named aria-label rather than label.
@jschuler @dlabaj thought on the API?

redallen: This is why I opened #1529. I think the best option is to make the Tab's property be named `aria-label` and not get propagated to `...props`. In the ES6 object destructuring rename it to `ariaLabel` via `'aria-label': ariaLabel` if some logic needs to take place with it in the rendering.

dlabaj: Agree that we should aim for consistency with the apis.  I would have aria-label as a separate prop instead of keying off the label value.

dlabaj: @tlabaj @rebeccaalpert Can we just move the nav & div element out of tabs and create 2 new tab components?  We could refactor tab.js to be 2 internal components (InternalTabs and TabContainer).

e.g.:

`
const TabsWithNav = ({ ariaLabel, wrapperClass, ...props }) => (<nav><InternalTab{...props}></nav><InternalTabContainer{...props}/>
// 
const Tabs = ({...props}) => (<div><InternalTabs {...props}/></div><InternalTabContainer{...props}/>

// What use to be component in Tabs.js is now InternalTabs, and the map function after that gets moved into a InternalTabContainer component?
`

Then you can keep the implementation props with each component and the users would know what they are using.

tlabaj: Yes I prefer either the composition approach that @dlabaj  has above, or a boolean with the component composing it internally.    I think composition is clearer to the consumer.  Also, with composition it is more like PF# (bootstrap implementation), so consumers you are coming from PF3 might have an easier time upgrading.

dlabaj: Can we go ahead and do this with composition @rebeccaalpert ?

rebeccaalpert: Yeah, no problem.

jgiardino:
Hey @rebeccaalpert, the labels look good! Thanks for adding that!

I was looking at the `id` values defined relative to `aria-controls` and `aria-labelledby` attributes for the **Primary tabs with sections outside of the Tabs component** example. They aren't matching up. I pasted the elements and their attributes below, but here are the mismatches:

1. These values should be equal but are not:
    `button` > `id` > `"pf-tab-2-pf-1553087727722by8m3hp1vts"` 
    ≠
    `section` > `aria-labelledby` > `"pf-tab-2-refTab3Section"`

2. These values should be equal but are not:
    `button` > `aria-controls` > `"pf-tab-section-2-pf-1553087727722by8m3hp1vts"` 
    ≠
    `section` > `id` > `"refTab3Section"`

- For case 1, we could explicitly define the `aria-label` to use the same text string as the tab label, if that's easier to implement.
- For case 2, there really aren't any other options. This is also a lower priority aria attribute, so it would be fine to capture this in a separate issue if it's a huge blocker to moving forward.

```
<li class="pf-c-tabs__item pf-m-current">
  <button 
    title="Tab item 3" 
    class="pf-c-tabs__button" 
    id="pf-tab-2-pf-1553087727722by8m3hp1vts" 
    aria-controls="pf-tab-section-2-pf-1553087727722by8m3hp1vts"
  >
    Tab item 3
  </button>
</li>
```
```
<section ...
  class="pf-c-tab-content" 
  id="refTab3Section" 
  aria-labelledby="pf-tab-2-refTab3Section" 
>
  Tab 3 section
</section>
```

Also, the `<button>` element has a `title` attribute. I'm not seeing that in the current react component or in core. Did you mean for that to be added there? I would suggest removing it unless there's a reason for it being included that I missed.

dlabrecq:
This looks great in it's current form, in regard to #1364.

I don't believe the a11y issues are a result of splitting the tabs and children into different components (i.e., the changes implemented for #1364)? I have to mention this because Cost Management has been waiting for this change for quite a while (early Feburary). 

Can we please address the other issues via another PR?

packages/patternfly-4/react-core/src/components/Tabs/Tab.js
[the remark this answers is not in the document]
rebeccaalpert: Joachim fixed the components.Tab error (fix in latest commit) and the key error is being tracked in https://github.com/patternfly/patternfly-react/issues/1589.

packages/patternfly-4/react-core/src/components/Tabs/TabContent.d.ts
```diff
@@ -0,0 +1,9 @@
+import { FunctionComponent, HTMLProps } from 'react';
+
+export interface TabContentProps extends HTMLProps<HTMLDivElement> {
+  id: string;
```
jschuler: I think `id` of type string might already part of the interface you're extending from
`interface HTMLAttributes` extended by `interface AllHTMLAttributes` extended by `interface HTMLProps` which you're extending
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1432

packages/patternfly-4/react-core/src/components/Tabs/TabContent.js
```diff
@@ -1,36 +1,47 @@
...
-  </section>
-);
+class TabContent extends React.Component {
+  render() {
+    const { id, activeKey, child, children, className, eventKey, innerRef, ...props } = this.props;
+    return <section
+      ref={innerRef}
+      index={eventKey}
+      hidden={children ? null : child.props.eventKey !== activeKey}
+      className={children ? css('pf-c-tab-content', className) : css('pf-c-tab-content', child.props.className)}
```
jschuler: is there no stylesheet you can import for tab content?

packages/patternfly-4/react-core/src/components/Tabs/TabContent.d.ts
rebeccaalpert: Yup, you're right.

packages/patternfly-4/react-core/src/components/Tabs/Tab.js
```diff
@@ -9,17 +9,41 @@ const propTypes = {
   /** Tab title */
   title: PropTypes.string.isRequired,
   /** uniquely identifies the tab */
-  eventKey: PropTypes.number.isRequired
+  eventKey: PropTypes.number.isRequired,
+  /** child id for case in which a TabContent section is defined outside of a Tabs component */
+  tabContentId: PropTypes.string,
```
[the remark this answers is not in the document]
rebeccaalpert: Sounds like yes. Updating the type.

packages/patternfly-4/react-core/src/components/Tabs/TabContent.js
rebeccaalpert: For documentation's sake, there's no stylesheet (similar to the brand component), but it's useful to keep the class around (the brand component is styled by Page this way).

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -14,29 +15,77 @@ const propTypes = {
...
   leftScrollAriaLabel: 'Scroll left',
-  rightScrollAriaLabel: 'Scroll Right'
+  rightScrollAriaLabel: 'Scroll Right',
+  'aria-label': null
 };
 
-class Tabs extends React.Component {
+const Tabs = ({ 'aria-label': ariaLabel, id, ...props }) => {
+  const unique_id = id ? id : getUniqueId();
+  return ariaLabel
```
tlabaj: I still don't think you should be switching with the aria-label.  I think you should expose the API for the TabsWithNav and TabsWithDiv and let the consumer decide what to use and compose it themselves.

rebeccaalpert: My mistake - updated in new commit.

jschuler: :( I think this is a bit too clunky. Would be much better to use a variant prop then if you don't want to key off of the ariaLabel.

packages/patternfly-4/react-integration/demo-app-ts/src/App.tsx
```diff
@@ -69,7 +81,15 @@ class App extends Component {
           <Tab eventKey={2} title="Tab item 3">
             Tab 3 section
           </Tab>
-        </Tabs>
+        </TabsWithNav>
+        <TabsWithDiv>
+          <Tab eventKey={0} title="Tab item 1" tabContentId="refTab1Section" tabContentRef={this.contentRef1}></Tab>
```
jschuler: Use self closing tags if no child content is expected

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
jgiardino: What I like about how it is now is that it's  clear that aria-label would only be used for the `<nav>` element because this prop is only included for the TabsWithNav component.

If we do have a single component, with one prop for specifying element and another for specifying `aria-label` then should we update the `aria-label` description to be more descriptive about this relationship? For example, assuming the prop for specifying html element is named `parentElement`:
> Provides an accessible label for the Tabs. Labels should be unique for each set of Tabs that are present on a page. When `parentElement` is set to `nav`, this prop should be defined to differentiate the tabs from other navigation regions on the page.

@jschuler Were you thinking the consumer should be able to pass whatever they want in for the variant prop (e.g. not just `<div>` vs `<nav>` but whatever)? 

I'm still a little fuzzy on what the expectations are related to documentation between react and html examples, like whether we expect react documentation to stand on it's own, or if we expect react devs to also refer to core documentation to better understand how to define some of these props. I'm ultimately fine with whatever approach you've been taking with other components.

packages/patternfly-4/react-core/src/components/Tabs/TabContent.js
```diff
@@ -1,36 +1,53 @@
...
 
 const propTypes = {
-  /** the index of the active tab */
+  /** the index of the active tab (not needed if used outside Tabs component) */
   activeKey: PropTypes.number,
-  /** content rendered inside the tab content area. */
+  /** content rendered inside the tab content area if used inside Tabs component */
   child: PropTypes.node,
+  /** content rendered inside the tab content area if used outside Tabs component */
+  children: PropTypes.any,
```
jschuler: While I don't think we need separate components for Tabs with div/nav, this one looks like we could split off, maybe a TabOutsideContent

rebeccaalpert:
I refactored this to key off of variant per discussion with @jschuler and @dlabaj, and removed the internal TabContent props from the props table. I updated the label for aria-label to match @jgiardino's suggestion (with slight modifications: I capitalized Tabs and changed parentElement to Tabs).

packages/patternfly-4/react-core/src/components/Tabs/examples/AccessibleSecondaryTabs.js
```diff
@@ -0,0 +1,51 @@
...
+  // Toggle currently active secondary tab
+  handleTabClickSecond = (event, tabIndex) => {
+    this.setState({
+      activeTabKey2: tabIndex
+    });
+  };
+
+  render() {
+    return (
+      <Tabs activeKey={this.state.activeTabKey1} onSelect={this.handleTabClickFirst} aria-label="Local" variant="nav">
```
tlabaj: we should be using the variant constant here.

PR #1409, opened 2019-02-19
fix(lint): fix linting glob and auto-fixable linter errors  (#1256)

redallen:
**What**: Fix linting glob and start linting code again.

**Additional issues**: Even after running `eslint --fix` there are still 912 linter problems. Lots of them are opinionated and will need to be iterated through slowly in order to resolve #1256.

Edit: Breakdown of 902 errors:
```
 318 camelcase
 234 import/no-extraneous-dependencies
 138 react/require-default-props
  80 react/prefer-stateless-function
  49 react/prop-types
  26 no-unused-vars
  26 import/no-unresolved
   8 jest/valid-expect
   5 react/no-find-dom-node
   3 react/default-props-match-prop-types
   3 prefer-destructuring
   3 no-nested-ternary
   2 consistent-return
   1 react/no-unused-state
   1 react/no-did-mount-set-state
   1 no-template-curly-in-string
   1 no-restricted-globals
   1 no-debugger
   1 jsx-a11y/no-redundant-roles
   1 class-methods-use-this
```

PR #1414, opened 2019-02-20
 style(extraneous): fix 234 import/no-extraneous-dependencies errors

redallen:
**What**: Fix no-extraneous-dependencies errors (234 of them). Ref #1409, towards #1256.

**Additional issues**: This is a WIP because it will break something... Haven't tested yet.


packages/patternfly-4/react-charts/src/components/AreaChart/examples/CustomColorsChart.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ChartArea, ChartGroup, ChartLegend, ChartVoronoiContainer } from '@patternfly/react-charts';
+import { ChartArea, ChartGroup, ChartLegend, ChartVoronoiContainer } from '../../index';
```
dmiller9911: If I remember correctly these are required for the docs package since files get copied over.

redallen:
Need to reconsider linting rules for examples.

PR #1416, opened 2019-02-20
build(noerr): suppress error on no package changes

redallen:
**What**:

**Additional issues**:


.circleci/release.sh
```diff
@@ -17,5 +17,5 @@ git checkout master
 
 # helpful for debugging any lerna EUNCOMMIT errors
 git rev-parse HEAD
-yarn run lerna changed
+yarn run lerna changed || true
```
priley86: nice! did not realize this existed.

PR #1422, opened 2019-02-21
build(publish): fix and improve lerna publish

redallen:
**What**: Rewrite `release.sh` to (hopefully) more consistently do what we want.

**Additional issues**: Only way to fully test it is to actually do a release! I'm on standby to fix it if it goes wrong.


.circleci/release.sh
```diff
@@ -2,20 +2,25 @@
...
-git config --global user.name "patternfly-build"
-git config --global push.default simple
-git checkout master
+echo "Doing a release..."
+# Lerna is complicated. Commands: https://github.com/lerna/lerna/tree/master/commands
+git checkout $TRAVIS_BRANCH # Lerna needs to be on a real HEAD
+git rev-parse HEAD # helpful for debugging any lerna EUNCOMMIT errors
+
+# Tag this new commit with each package change. Don't push it yet.
+npx lerna version --conventional-commits --no-commit-hooks --no-git-tag-version --no-push --yes
```
humblej86: ### 
```suggestion
npx lerna version --conventional-commits --no-commit-hooks --no-git-tag-version --no-push --yes
```

.circleci/release.sh
```diff
@@ -2,20 +2,25 @@
...
-git rev-parse HEAD
-yarn run lerna changed || true
-yarn lerna:publish
\ No newline at end of file
+# Check each package.json and determe if any package version is not present in the registry.
+# Any versions not present in the registry will be published.
+if npx lerna publish from-package --no-changelog --no-git-tag-version --no-push --yes ; then
+    # Undo that last commit locally
+    git reset --hard HEAD~1
+    # Now only if it publishes should we also push this commit to Github and do a Github release.
```
priley86: pretty nifty... seems it should work 👍

Issue #1407, opened 2019-02-19
PF4 Table: render row dependent actions 


suomiy:
Currently, actions are rendered with the same content for each row.

Action rendering should be decided by a custom function and take rowData into account. For example row type or when the table is edited (actions would probably not be visible or be disabled).

rachael-phillips:
@tlabaj could you take a look at this one and provide labels?

dgutride:
Fixed in #1441

PR #1441, opened 2019-02-25
 feat(table): add customizable actions for each row

suomiy:
implements  #1407

there are two approaches how to specify actions or kebap disability:
- declarative - SimpleActionsTable example
- with  `actionResolver()` and ` areActionsDisabled()` functions - ActionsTable example

- added passing of isDisabled to DropdownToggle


suomiy:
@karelhala

packages/patternfly-4/react-table/src/components/Table/ActionsColumn.js
```diff
@@ -38,22 +40,22 @@ class ActionsColumn extends React.Component {
   };
 
   onSelect = (event, onClick) => {
-    const { rowId } = this.props;
+    const { rowData, extra } = this.props;
     event.preventDefault();
-    onClick && onClick(event, rowId);
+    onClick && onClick(event, rowData, extra);
```
priley86: +1 for passing `rowData`  back. Maybe rename `extra` to `extraParams` for consistency?

suomiy: fixed

priley86:
Seems like a nice enhancement.

suomiy:
tests added

packages/patternfly-4/react-table/src/components/Table/Table.api.test.js
```diff
@@ -41,6 +41,62 @@ describe('Selectable table', () => {
   });
 });
 
+describe('Action table', () => {
+  test('should call correct functions', () => {
```
suomiy: maybe this test is superfluous?

priley86: do you lose coverage without it? It could be, but i don't see any harm if you are trying to ensure that these props are correctly digested.

suomiy: ok I will leave it in then

suomiy:
tests passing, finally :)

priley86:
hey @suomiy & @karelhala   - so after reviewing this a bit more, I think one thing this API might miss on for some consumers is the ability to _define_ action's JSX alongside rows...for example, given the following (you can see my current alternative):
```
const VMTableRow = (vm) => {
  const { name, namespace } = vm.metadata;
  const migrationResources = getResource(VirtualMachineInstanceMigrationModel, {namespace});
  const resourceMap = {
    pods: {
      resource: getResource(PodModel, {namespace, matchLabels: getLabelMatcher(vm)}),
    },
    importerPods: {
      resource: getResource(PodModel, {namespace, matchLabels: {[CDI_KUBEVIRT_IO]: 'importer'}}),
    },
    migrations: {
      resource: migrationResources,
    },
  };

  return [
    {
      title: <ResourceLink kind={VirtualMachineModel.kind} name={name} namespace={namespace} title={vm.metadata.uid} />,
    },
    {
      title: <ResourceLink kind={NamespaceModel.kind} name={namespace} title={namespace} />,
    },
    {
      title: <WithResources resourceMap={resourceMap}
        resourceToProps={({ pods, importerPods, migrations }) => ({
          launcherPod: findPod(pods, name, VIRT_LAUNCHER_POD_PREFIX),
          importerPods: findImporterPods(importerPods, vm),
          migration: findVMIMigration(migrations, name),
        })}
        loaderComponent={() => DASHES}>
        <VmStatus vm={vm} />
      </WithResources>,
    }, {
      title: <div className="dropdown-kebab-pf">
        <ResourceKebab actions={menuActions}
          kind={VirtualMachineModel.kind}
          resource={vm}
          resources={[
            getResource(VirtualMachineInstanceModel, {name, namespace, isList: false}),
            migrationResources,
          ]} />
      </div>,
      props: { className: 'pf-c-table__action'},
    },
  ];
};
```

The Action's menu items and inner HTML definition depend on the specific row. In that case, our simple actions array defined like this should probably accept row inputs/row props:
```
      actions: [
        {
          title:  (rowProps) => <div>`Some custom ${rowProp.magic} action` </div>,
          onClick: (event, rowId) => console.log('clicked on Some action, on row: ', rowId)
        },
```  
OR... we just allow actions to be alternatively defined alongside rows (this is probably more flexible in the case certain rows do not have certain actions:
```
this.state = {
   columns: [...],
   rows: [
      { cells: ['one', 'two', 'a', 'four', 'five'], actions: [...] }
   ]
}
```

I believe the latter to be most flexible here.

suomiy:
@priley86 

> the ability to define action's JSX alongside rows

you can already define custom JSX (the title property) for each row by writing custom actionResolver (this PR) and render the action depending on the type (data) of the row. 

> OR... we just allow actions to be alternatively defined alongside rows (this is probably more flexible in the case certain rows do not have certain actions:

This is already possible with the actionResolver. You can define certain actions or no actions for each row.

we can also go with the dual approach and have the actions defined by 

1. actionResolver
2. Row Actions
3. Table Actions

in that order.

Should I add the RowActions?

and do you agree with that priority? I think the functional approach (actionResolver) should have the highest priority.

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -58,6 +58,8 @@ export interface TableProps extends Omit<Omit<HTMLProps<HTMLTableElement>, 'onSe
   onSelect?: Function;
   onSort?: Function;
   actions?: Array<IAction | ISeparator>;
+  actionResolver?: (rowData: Object, extraParams: Object) => Array<IAction | ISeparator>;
```
priley86: ok - yes, re-reading this piece, it does resolve my issue. `rowData` is returned in the resolver thus letting the consumer construct custom JSX w/ `rowData` for each action.

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -58,6 +58,8 @@ export interface TableProps extends Omit<Omit<HTMLProps<HTMLTableElement>, 'onSe
...
+  areActionsDisabled?: (rowData: Object, extraParams: Object) => boolean;
```
priley86: yes - this resolves my other issue, allow rows to disable certain actions on a row by row basis. Did not realize `rowIndex` was included in the callback initially.

priley86:
@suomiy - in summary, yes, after further review - i agree that `actionResolver` and `areActionsDisabled` resolves my issue. I will rebase if this PR lands. Not sure how others feel about it yet but will ask.

packages/patternfly-4/react-table/src/components/Table/ActionsColumn.js
```diff
@@ -38,22 +40,22 @@ class ActionsColumn extends React.Component {
   };
 
   onSelect = (event, onClick) => {
-    const { rowId } = this.props;
+    const { rowData, extraParams } = this.props;
     event.preventDefault();
-    onClick && onClick(event, rowId);
+    onClick && onClick(event, rowData, extraParams);
```
priley86: thanks for adding this!

karelhala:
For some reason the actions column is missing in calculated action cell 
screenshot from 2019-03-05 13-07-54

packages/patternfly-4/react-table/src/components/Table/ActionsColumn.js
suomiy: The rowId is already included in extraParams as rowIndex. Does it make a sense to include it separately? We are passing rowData in this fashion in other callbacks as well.

> So we don't break current API.

that is true, but this is already happening (e.g. onRowClick in https://github.com/patternfly/patternfly-react/pull/1227). I mean, aren't tables alpha at the moment?

suomiy:
> For some reason the actions column is missing in calculated action cell

You are right, I will look into that.

packages/patternfly-4/react-table/src/components/Table/ActionsColumn.js
karelhala: Not really they are already used by some projects and we shouldn't break the API for them.

karelhala:
> You are right, I will look into that.

I think that the problem here is that you are not returning null, try returning empty object (or object with `children: ''`) when actions are disabled for row.

suomiy:
fixed

f

suomiy:
fixed
on action click callback is backwards compatible

redallen:

suomiy:
> Rebase on latest master to fix CircleCI errors and get a deploy preview.

thanks, deployed and tests passed

Issue #1378, opened 2019-02-13
PF4 Title: TypeScript error for headingLevel


dlabrecq:
While attempting to set the headingLevel attribute for the Title component, I encountered the TypeScript error below. I can set the size, but cannot make my title an 'h3', for example.

<img width="1218" alt="screen shot 2019-02-12 at 11 38 21 pm" src="https://user-images.githubusercontent.com/17481322/52687368-92821100-2f1f-11e9-91e9-be9b4a592181.png">

dlabrecq:
There is no headingLevel property in the Title.d.ts file.

![screen shot 2019-02-13 at 10 05 04 am](https://user-images.githubusercontent.com/17481322/52721125-07346a00-2f77-11e9-8639-3e1acc3d9206.png)

Issue #1439, opened 2019-02-25
Title component inconsistent prop defaults


seanforyou23:
`size` prop is required without a default
`headingLevel` is not required, but has a default

We should align these two. `headingLevel` should be required as a default of h1 leads to non-semantic markup structures for users who go with defaults. It's not good for accessibility as it stands out of the box.

Further, we'd probably benefit from shortening the name from `headingLevel` to simply `level`.

rachael-phillips:
After discussion, we'd like to move forward with making the headinglevel required but not the name change.

tlabaj:
closed by #3922

PR #1445, opened 2019-02-26
chore(title): title fixes and upgrades

seanforyou23:
**What**: This PR fixes a TS error when trying to set the headingLevel prop for Title. I've gone through the react workspace and updated all instances of `<Title>` and given what I believe is a sensible value for headingLevel, basically corresponding it with the given `size` that was already present. As a result, the demo pages have a true hierarchy that better reflects the structure of the content.

Converts `Title` to a typescript.

Along the way, I noticed `/patternfly-4/react-docs/src/components/docs/` was unused so I deleted it.

With these changes, you'll now get proper IDE support for the level property.

**Additional issues**:
Fixes https://github.com/patternfly/patternfly-react/issues/1378

assign an appropriate headingLevel for each use of <Title> that corresponds to the size prop used
update tests and snapshots
fix a11y issues in workspace
remove unused /patternfly-4/react-docs/src/components/docs/
fix bad path in pagination
convert title to TS


dgutride:
@seanforyou23  - we can't rename props without approval right now - please open an issue tracking that request and we'd like to get any bug fixes taken in separately

seanforyou23:
No problem @dgutride - https://github.com/patternfly/patternfly-react/issues/1439 captures the request.

jschuler:
@seanforyou23  How about you keep the prop name the same as before and also don't make it required so as not to introduce a breaking change, then this can still go in. We can track prop name change and making it required in a separate issue if need be.

seanforyou23:
Working through those updates now, will push these changes soon

packages/patternfly-4/react-docs/src/components/example/example.js
```diff
@@ -66,7 +66,7 @@ const Example = ({
     const path = `/${pathStart}/examples/${exampleName}`;
     return (
       <Section>
-        <Title size="lg" headingLevel="h3">
+        <Title size="lg" level="h3">
```
jschuler: These ones also need to be changed

seanforyou23:
Alrighty @tlabaj @jschuler @dlabaj - thx for the feedback. I believe it's all addressed in the latest changeset. Let me know if you see anything else!

seanforyou23:
If anyone knows where I can see `liveDemo.js` in action, feel free to point me there and I'll make sure it uses a sensible heading level. Right now it's set to h6 which likely isn't accurate.

packages/patternfly-4/react-core/src/components/Title/levels.js
```diff
@@ -0,0 +1,8 @@
+export const HeadingLevels = {
```
jschuler: Looks like you'll want to add this to the index.js file so it can be imported by consumer?

tlabaj: Is this a user facing component? and is There a need for the separate levels.js file?

seanforyou23:
Sorry for the delay @tlabaj, I was taking a bit of time to brush up on some related TS concepts to help me better understand how to best treat this. The purpose for HeadingLevels was only ever to ensure we have good typing support and intellisense for the headingLevel property. I don't really need or see a reason for it to be a separate module in a separate file, or even consumable by end users, but so far that's the only way I've been able to get it to work.

In working through moving HeadingLevels back into the main Title component file, it came up that this would be a good opportunity to go ahead and do the full conversion to TypeScript so we don't have the manually manage the declaration files. That's pretty much where I'm at. This opened a can of worms that I'm still trying to sort through.

If it's ok I'd like to take a little more time to get the best solution in place and make sure I understand what's at stake myself. Is that ok?

seanforyou23:
Ok, I've addressed quite a few issues and with the exception of the prop table content, Title seems to be working well as a TS component. The `headingLevel` and `size` props don't populate the props table as they did before. I'm still trying to figure out how to fix that. I'd love some feedback on the TitleProps interface in case anyone knows a better way to define size and headingLevel. Thanks for all the help!

jschuler:
@seanforyou23 We have implemented basic props table support so far, as we work on more TS components we will have to add more prop types. It is very possible that one of your props needs yet to be handled correctly to show up in the table.

packages/patternfly-4/react-core/src/components/Title/index.d.ts
```diff
@@ -1 +1,2 @@
-export { default as Title, TitleSize, TitleProps } from './Title';
+export { default as Title, TitleProps } from './Title';
```
jschuler: I think you can delete this file

packages/patternfly-4/react-styles/src/utils.d.ts
```diff
@@ -13,7 +14,7 @@ export function getModifier(
   styleObject: any,
   modifier: string,
   defaultModifer?: StyleDeclarationStatic | string
-): StyleDeclarationStatic | string;
+): ClassNameArg | string;
```
jschuler: Revert this change

Issue #1383, opened 2019-02-13
Modal component enhancements


jgiardino:
As a consumer, I should have more flexibility over what gets displayed and what gets announced as part of a modal. 

- Title - Currently, core shows an example of a modal with no visible title, while still including a title for assistive technology in `aria-label`. A similar example should be included on the React preview page.
- Description - Currently, everything inside `.pf-c-modal-box__body` is announced, which is fine when the contents are just text, but this might not make sense if the dialog includes a lot of interactive elements. 
  - The consumer should be able to wrap a subset of contents included in `.pf-c-modal-box__body` with a `<span id="[id]">` and then pass that `[id]` value to `aria-describedby` on the `.pf-c-modal-box`.

An example of both of these use cases can be seen in [core](https://pf4.patternfly.org/components/ModalBox/examples/) in the last example labelled "ModalBox No Header Example".

PR #109, opened 2017-12-13
feat(label): add the pf label component

dabeng:
**What**:
This PR adds the PatternFly label pattern. To be used to close #18 (Label) once done.

Storybook: https://rawgit.com/dabeng/patternfly-react/label-storybook/.out/index.html?knob-Label=Danger%20Will%20Robinson%21&selectedKind=Label&selectedStory=Label&full=0&down=1&left=1&panelRight=0&downPanel=storybooks%2Fstorybook-addon-knobs

**How**:
Providing a label story of patternfly-react storybook based on react-bootstrap label component.

@priley86 , please review.

PR #1453, opened 2019-02-26
fix(Modal): accessibility enhancements and other small fixes

nicolethoen:
Fixes [issue #1383](https://github.com/patternfly/patternfly-react/issues/1383)

Most of the changes are not easy to document in a screen shot, but here are the UI changes:
screen shot 2019-02-26 at 4 02 30 pm
screen shot 2019-02-26 at 4 02 40 pm

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.js
@@ -2,7 +2,6 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import ModalContent from './ModalContent';
-import safeHTMLElement from '../../helpers/safeHTMLElement';
 import { canUseDOM } from 'exenv';
 import { KEY_CODES } from '../../helpers/constants';
 import { css } from '@patternfly/react-styles';
@@ -19,6 +18,8 @@ const propTypes = {
   title: PropTypes.string.isRequired,
   /** Flag to show the title */
   hideTitle: PropTypes.bool,
+  /** id to use for Modal Box description */
+  ariaDescribedById: PropTypes.string,
   /** Action buttons to put in the Modal Footer */
   actions: PropTypes.any,
   /** A callback for when the close button is clicked */
@@ -38,6 +39,7 @@ const defaultProps = {
   className: '',
   isOpen: false,
   hideTitle: false,
+  ariaDescribedById: '',
   actions: [],
   onClose: () => undefined,
   isLarge: false,
@@ -103,7 +105,7 @@ class Modal extends React.Component {
       this.container = document.createElement('div');
     }
 
-    return ReactDOM.createPortal(<ModalContent {...props} id={this.id} />, this.container);
+    return ReactDOM.createPortal(<ModalContent {...props} title={this.props.title} id={this.id} ariaDescribedById={this.props.ariaDescribedById}/>, this.container);
   }
 }
 
--- a/packages/patternfly-4/react-core/src/components/Modal/ModalBox.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Modal/ModalBox.d.ts
@@ -3,10 +3,10 @@ import { FunctionComponent, HTMLProps, ReactNode } from 'react';
 export interface ModalBoxProps extends HTMLProps<HTMLDivElement> {
   children: ReactNode;
   className?: string;
-  id: string;
   isLarge?: boolean;
-  isSmall?: boolean;
+  isSmall?:boolean;
   title: string;
+  id: string;
 }
 
 declare const ModalBox: FunctionComponent<ModalBoxProps>;
--- a/packages/patternfly-4/react-core/src/components/Modal/ModalBox.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/ModalBox.js
@@ -12,7 +12,7 @@ const propTypes = {
   isLarge: PropTypes.bool,
   /** Creates a small version of the ModalBox. */
   isSmall: PropTypes.bool,
-  /* Header title used for Modal Box Header */
+  /** string to use for Modal Box label */
   title: PropTypes.string.isRequired,
   /** id to use for Modal Box description */
   id: PropTypes.string.isRequired,
--- a/packages/patternfly-4/react-core/src/components/Modal/ModalContent.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/ModalContent.js
@@ -33,6 +33,8 @@ const propTypes = {
   /** Creates a small version of the Modal */
   isSmall: PropTypes.bool,
   /** id to use for Modal Box description */
+  ariaDescribedById: PropTypes.string,
+  /** id of the ModalBoxBody */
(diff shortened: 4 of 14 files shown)


jgiardino:
Thanks for making these updates! There are just a couple of things that I noticed...

- `aria-describedby` - The value that is defined here should be the value that's defined for an `id` attribute. In your example, where you have `<span className="no-header-example">`, it should be `<span id="no-header-example">` instead.
- `aria-label` vs `aria-labelledby` - The example we have in core uses `aria-label`, whereas the example you provide uses `aria-labelledby`. I actually don't have an opinion on this, since they both work. I'd opt for the method we show in core because that's less markup, but if the react component is much simpler the way you handled it, then I'd say that's fine.
  - One thing in this implementation that should be updated is the class you're using to hide the title. In this example, you use the class `sr-only`, but in core the class we have is `pf-u-screen-reader`.

packages/patternfly-4/react-core/src/components/Modal/ModalContent.js
```diff
@@ -33,7 +35,9 @@ const propTypes = {
   /** Creates a small version of the Modal */
   isSmall: PropTypes.bool,
   /** id to use for Modal Box description */
-  id: PropTypes.string.isRequired,
```
tlabaj: \Removing id will break the API.  Perhaps keep it and update the description for now.   We can can open an issue about changing the name in resolve that when we are ready for our next major release.

nicolethoen: @tlabaj 
I'm happy to change it back, but (just to make sure I understand) ModalContent is a component that is loaded by Modal.js. I assumed that since the Modal component is what the users would import, that this was more of an internal API. The API on the generated site doesn't change. But this is still a breaking change?

nicolethoen: changing the name of the variable back does not change anything in the code outside of Modal.js [line #109](https://github.com/patternfly/patternfly-react/blob/d6e7c30eaa26d2b9cbb236c58660997cc9b87950/packages/patternfly-4/react-core/src/components/Modal/Modal.js#L109) and the references to the property inside of ModalContent so even though it's a less descriptive name, it's not an inaccurate name, so it may not need to be changed at all if it's really considered a breaking change.

tlabaj: You are right since it is internal, this would not break a consumer API.  And yes the new name is more descriptive so I agree it should change.

redallen:

packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -50,7 +52,8 @@ class Modal extends React.Component {
   static propTypes = propTypes;
   static defaultProps = defaultProps;
 
-  id = `pf-modal-${currentId++}`;
+  bodyId = `pf-modal-${currentId}`;
```
dlabaj: Do you know why we got rid of id?

packages/patternfly-4/react-core/src/components/Modal/ModalBox.d.ts
```diff
@@ -5,8 +5,9 @@ export interface ModalBoxProps extends HTMLProps<HTMLDivElement> {
   className?: string;
   id: string;
   isLarge?: boolean;
-  isSmall?: boolean;
-  title: string;
+  isSmall?:boolean;
+  titleId: string;
```
dlabaj: Why did we change title to titleId?

packages/patternfly-4/react-core/src/components/Modal/Modal.js
tlabaj: I think it is just a rename here.  It is internal so it should be fine

nicolethoen:
@dlabaj 
Generally, the attribute names were changed from simply id to a more descriptive name because of the accessibility requirement in the original issue stated as follows: 
"The consumer should be able to wrap a subset of contents included in .pf-c-modal-box__body with a <span id="[id]"> and then pass that [id] value to aria-describedBy on the .pf-c-modal-box."

aria-describedby on the .pf-c-modal-box was originally using an autogenerated id assigned to the .pf-c-modal-box__body. Now, if a developer passes an ariaDescribedById to the Modal component, the aria-describedBy attribute will use that value. Otherwise, aria-describedBy will be set to the autogenerated id as before.

For this fix I had renamed the autogenerated id prop to bodyId to more accurately reflect which element is being assigned that id.

kmcfaul:
LGTM aside from the id > titleId rename to minimize breaking changes at this time.

nicolethoen:
@kmcfaul @dlabaj 
I was misunderstanding how to use aria-label. The titleId is not necessary and was removed.

packages/patternfly-4/react-core/src/components/Modal/ModalBox.js
```diff
@@ -12,10 +12,10 @@ const propTypes = {
   isLarge: PropTypes.bool,
   /** Creates a small version of the ModalBox. */
   isSmall: PropTypes.bool,
-  /* Header title used for Modal Box Header */
+  /** string to use for Modal Box label */
   title: PropTypes.string.isRequired,
   /** id to use for Modal Box description */
-  id: PropTypes.string.isRequired,
```
tlabaj: we still can't rename this prop since ModalBox is exported.

packages/patternfly-4/react-core/src/components/Modal/ModalBox.js
```diff
@@ -26,12 +26,12 @@ const defaultProps = {
   isSmall: false
 };
 
-const ModalBox = ({ children, className, isLarge, isSmall, title, id, ...props }) => (
+const ModalBox = ({ children, className, isLarge, isSmall, title, ariaDescribedById, ...props }) => (
   <div
     {...props}
```
tlabaj: can you spread the props at the end to allow for overwriting attribute values.

dlabaj:
@nicolethoen Since we are exporting unfortunatly we can not change the api right now as outside consumers might be using it.  If we want to updates these we can look into it when we release a breaking changes release. Can you add back title and id?

packages/patternfly-4/react-core/src/components/Modal/ModalBox.d.ts
```diff
@@ -5,8 +5,9 @@ export interface ModalBoxProps extends HTMLProps<HTMLDivElement> {
   className?: string;
   id: string;
```
dlabaj: Looks like you have id twice now.

kmcfaul:
Snapshots need to be updated.

packages/patternfly-4/react-core/src/components/Modal/ModalBox.js
nicolethoen: @tlabaj just seeing this comment. Are they not already being spread at the end? I'm not sure what you mean

tlabaj: you can ignore that since the id is still coming in explicitly.

Issue #1033, opened 2018-12-11
Gatsby not generating props for typescript


dlabaj:
Currently Gatsby is not generating props for typescript files.  'babel-plugin-react-docgen-typescript' is being used in gatsby-node.js, however we are not currently getting any props generated for tsx files in react-core such as Avatar.tsx

dgutride:
The gatsby team appears to be working on a fix for this on their end - I opened an issue on their repo to try to figure out if there is a timeframe on a fix.  Moving to gatsby-transformer-react-docgen 3.0.5 does start to include some props in the response so that might be an option temporarily.
@jschuler @dlabaj 
https://github.com/gatsbyjs/gatsby/issues/11650

PR #1475, opened 2019-03-01
Generate props for typescript components

jschuler:
Using typedoc to parse ts files which spits out a json file. Read the json file in gatsby and use graphql to find the component. This will likely need to be adjusted when more TS components are added but should give us a good base to build upon.
Closes: #1033

### DIFF
--- a/packages/patternfly-4/react-core/src/tsconfig.dts.json
+++ b/packages/patternfly-4/react-core/src/tsconfig.dts.json
@@ -8,16 +8,28 @@
     "declaration": true,
     "emitDeclarationOnly": true,
     "skipLibCheck": true,
-    "sourceMap": false, // allow sourcemap support
-    "lib": ["es2017", "dom"],
+    "sourceMap": false,
+    "lib": [
+      "es2017",
+      "dom"
+    ],
     "noEmit": false,
-    "module": "commonjs", // specify module code generation
-    "jsx": "react", // use typescript to transpile jsx to js
-    "target": "es5", // specify ECMAScript target version
+    "module": "commonjs",
+    "jsx": "react",
+    "target": "es5",
     "paths": {
-      "*": ["../dist/esm/*"]
+      "*": [
+        "../dist/esm/*"
+      ]
     }
   },
-  "include": ["./**/*"],
-  "exclude": ["./**/*.d.ts", "./**/**.test.tsx", "./**/**.test.ts", "./**/examples/**"]
+  "include": [
+    "./**/*"
+  ],
+  "exclude": [
+    "./**/*.d.ts",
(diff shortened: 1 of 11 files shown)


packages/patternfly-4/react-core/src/tsconfig.dts.json
```diff
@@ -8,16 +8,28 @@
     "declaration": true,
     "emitDeclarationOnly": true,
     "skipLibCheck": true,
-    "sourceMap": false, // allow sourcemap support
-    "lib": ["es2017", "dom"],
+    "sourceMap": false,
```
jschuler: Comments are not allowed in json files

packages/patternfly-4/react-docs/src/components/componentDocs/componentDocs.js
```diff
@@ -57,6 +58,7 @@ class ComponentDocs extends React.PureComponent {
     } = this.props;
     const makeDescription = html => ({ __html: html });
     const getDocGenInfo = name => data.allComponentMetadata.edges.find(edge => edge.node.displayName === name);
+    const getDocGenInfoTs = name => data.allTsDocsJson.edges[0].node.data.find(edge => edge.name === `${name}Props`);
```
jschuler: Right now making the assumption to find the corresponding prop for a component by appending `Props` to the component name, e.g. Avatar => AvatarProps

packages/patternfly-4/react-integration/demo-app-ts/src/App.tsx
```diff
@@ -77,7 +74,7 @@ class App extends Component {
           action={<AlertActionLink>Action Button</AlertActionLink>}
         />
         <Alert variant="success" title="Success notification title" />
-        <Select
+        {/* <Select
```
dgutride: is select broken right now?

jschuler: I must have had outdated react-core in my node_modules folder so it couldn't find those exports. I cleared and reinstalled and now is working. I reverted the change to this file

packages/patternfly-4/react-core/src/tsconfig.dts.json
dlabaj: How did this work before?

jschuler: Gatsby is now reading json files in and parsing them so the newly added plugin doesn't like comments in json

jschuler: https://stackoverflow.com/questions/244777/can-comments-be-used-in-json

PR #1479, opened 2019-03-01
fix(nav): Fix bug where previous was expanding on select

ibolton336:
**What**:
#1016 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
+++ b/packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
@@ -9,7 +9,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  defaultValue: true,
+  defaultValue: false,
   groupId: 0,
   onToggle: () => undefined
 };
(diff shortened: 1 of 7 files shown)


packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
```diff
@@ -18,29 +18,30 @@ class NavToggle extends React.Component {
   static propTypes = propTypes;
   static defaultProps = defaultProps;
   state = {
-    value: this.props.defaultValue
+    toggleValue: this.props.defaultValue
   };
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.defaultValue !== this.state.value) {
-      this.setState({ value: nextProps.defaultValue });
+  componentDidUpdate(prevProps) {
```
priley86: 👍

priley86:
LGTM - 🐛 🔨 

after

packages/patternfly-4/react-core/src/components/Nav/NavToggle.js
```diff
@@ -9,7 +9,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  defaultValue: true,
+  defaultvalue: false,
```
jschuler: Not camel case

rhamilto:
This fix introduces some new bugs:

1.  Upon app load, active sections are not expanded.

yhnpvoa2v6

2.  Previously expanded items that have been collapsed items are (once again) not re-expanded.

ixkroplivy

cc: @spadgett, @janwright73, @priley86

spadgett:
It seems like the underlying problem is that we're tracking the same value using both props and state, which is generally considered an anti-pattern. Would it be simpler just to let the outer component control expanded state and add a callback for when the nav section is activated? Then the expanded prop is the single source of truth.

It puts more work on the outer component, but I don't see a good way around that as long as the expanded state can change based on things that happen outside the nav (such as the user clicking on a link on a details page).

priley86:
this should mostly be resolved now. One step closer here...
https://github.com/openshift/console/pull/1238#issuecomment-469763337

Issue #1217, opened 2019-01-17
PF4: Custom Select Component: Checkbox select (multiple items) variation


tlabaj:
This issue covers the Checkbox select (multiple items) variation of the Custom Select only:
(This issue is a child of issue #1206 )

On hold until the Single select variation is complete (issue #1214 )

Related issue in Core: patternfly/patternfly-next#651

Info provided on the component: https://docs.google.com/document/d/1rrCdARutABI-Vpa_mdMWGl2xJJTtt2dPQ2bjFx2zwCM/edit?usp=sharing

see: https://pf-next.com/components/Select/examples/

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @srambach (CSS) @mcarrano (interaction design) @christiemolloy (visual design)

jgiardino:
There was some discussion in the a11y meeting today related to this component. Here's a recap of some things discussed:

- This variation of the Select component will use the behavior described in this issue: https://github.com/patternfly/patternfly-react/issues/839 (i.e. the panel behaves similar to a popover)
  - #839 is intended to allow consumers to use the Dropdown component and customize what displays in the panel that pops up. 
  - Having the panel behave like a popover allows us to manage the visibility relative to focus. The panel will trap focus, and isn't hidden until the user explicitly exits it. This will be the default behavior of the Dropdown Panel variation of the dropdown.
- Since for this case we know what the contents of the panel are, we can also include additional keyboard interaction behavior so that a sighted keyboard-only user will have similar experiences for components that look the same (i.e. toggle that displays something with a menu). We might not be able to have complete parity, but maybe at a minimum can:
  - enable the arrow keys to shift focus between menu items
  - have Tab on the last menu item exit the menu
- It might make sense to start off with basic functionality first (e.g. basic popover functionality) and do a round of testing, and incrementally add the arrow key and tab key functionality then test again, in case these updates cause the component to not work at all in one of the screen readers.

kmcfaul:
@jgiardino For clarification, should Checkbox Select behave differently than Dropdown for keyboard input? The custom dropdown, Dropdown Panel, currently matches our Dropdown's behavior: it does not use a focus trap, arrow keys are used to navigate items, and tab closes the menu regardless of position. 

The other issue I've run into is that you cannot natively pass focus to a label, so I've added a tabIndex to the labels for the moment to allow focusing. I'm not sure if there's a better way to handle that.

jgiardino:
[...]

PR #1487, opened 2019-03-04
feat(Select): add checkbox select variant

kmcfaul:
**What**: Adds checkbox select variant component.

**Refer to issue**: #1217


mcoker:
Hey @kmcfaul looks good! Compared to the [core example](https://pf4.patternfly.org/components/Select/examples/)...

* `.pf-c-select__menu-group` should wrap the entire group of elements in the group.
* In `.pf-c-select__toggle`, we have `.pf-c-select__toggle-wrapper` + `.pf-c-select__toggle-badge` + `.pf-c-select__toggle-arrow`

Is there any reason you wouldn't match this structure in the react component?

@jgiardino looking at this, do you agree that we should create an issue in core to update `.pf-c-select__menu-group` to be a `section`, and `.pf-c-select__menu-group-title` to be an `h1` by default, like the dropdown component?

jgiardino:
Great work on this, @kmcfaul!

There is currently a PR open in core to change how `<fieldset>` is used: https://github.com/patternfly/patternfly-next/pull/1483

Would it be possible to follow what's done in that PR even though it isn't merged yet? This PR doesn't change the css, just the html structure to use separate fieldsets for each section of checkboxes.

When there are a group of checkboxes, the best way to group them is with a `<fieldset>`, and the label used for the fieldset can provide additional context when the checkbox label is announced (depending on screen reader and nav method).  In this case, instead of using `<legend>`, we're using `aria-labelledby` to reference the visible label, and also are making that node hidden to screen readers to avoid duplication. 

> @jgiardino looking at this, do you agree that we should create an issue in core to update .pf-c-select__menu-group to be a section, and .pf-c-select__menu-group-title to be an h1 by default, like the dropdown component?

I would stick to the updates in that core PR, rather than incorporate section and h1 elements, since they would provide no added benefit. The h1 would be ignored due to `aria-hidden`, and the `fieldset` provides sufficient sectioning, and it could be annoying in some screen readers to wrap it with another section element.

jgiardino:
Is this component also expected to work when there is just one group of checkboxes? I removed one of the groups, and removed the group title, and it rendered like this. This is actually very similar to the example in core, except that there's this empty div that adds extra space:
```
<div class="pf-c-select__menu-group"><div class="pf-c-select__menu-group-title"></div></div>
```

image

In this case, if the consumer has a single group with no label, then can we pull the label from the toggle?

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.js
```diff
@@ -0,0 +1,115 @@
...
+            ref={this.ref}
+            onKeyDown={this.onKeyDown}
+            tabIndex={0}
+          >
+            <input
+              id={value}
+              className={css(checkStyles.checkInput)}
+              type="checkbox"
+              onChange={event => {
+                if (!isDisabled) {
```
jgiardino: There should also be a `disabled` attribute on the `<input>`.

jgiardino:
Looking at the aria-labels, there are a couple of things I notice:

- The example includes `aria-label="Select Input"`, but I don't see this string included in the component
- The Single Select example includes `ariaLabelledBy={titleId}` which includes an additional id in the `aria-labelledby` attribute for the toggle. We should do the same with this component.

kmcfaul:
@mcoker I will address these. I was following the integration document for structure at the outset and may have misread it. For the toggle, I had read it as using the Badge component and didn't want to couple the components, but now I realize I can create a span and reuse the classes instead. 

@jgiardino 
Looks like I just need to create a fieldset under each `pf-c-select__menu-group`, if the select is grouped. I should be able to adjust the child mapping to include that.

For only one group of checkboxes without a label, the implementation would use an array of `CheckboxSelectOption` and leave the `CheckboxSelectGroup` out altogether. It is an optional wrapper and is flagged by the Select `isGrouped` prop. I will add another example to match core, showing a non-grouped checkbox select.

The `aria-label="Select Input` is getting passed down to the fieldset at the moment. With the fieldset changes though, I will move the prop to the outer wrapping div `pf-c-select__menu`. Let me know if the label should be applied to something different instead.

I will add the hidden title as in the SingleSelect example.

kmcfaul:
@mcoker It looks like adjusting the badge has some side effects, seemingly due to the width being set for Select. The toggle contents are justified so the badge floats more or less in the middle opposed to a specific side.

image

If I put the badge div under the toggle wrapper it floats closer to the text, but I'm not sure if it should be close to the text or to the carat.

image

Thoughts?

jgiardino:
> The `aria-label="Select Input` is getting passed down to the fieldset at the moment. With the fieldset changes though, I will move the prop to the outer wrapping div `pf-c-select__menu`. Let me know if the label should be applied to something different instead.

Oh, I see. For this component, I would suggest leaving the prop out, and only providing `aria-label/labelledby` attributes on the `<fieldset>` elements. For the case where there's only one `<fieldset>`, I would have the label come from the toggle.

mcoker:
@kmcfaul Ah, unfortunately that's a bug in core, too.

screen shot 2019-03-06 at 8 47 42 am


My inclination is to match what's in core, even if it's broken, then we can fix it in core and hopefully it's just CSS that will fix it here. I mentioned this to @christiemolloy  in an open PR about select - https://github.com/patternfly/patternfly-next/pull/1527#issuecomment-469831447

Either we'll fix it there, or once that's merged, we'll follow up and fix it then.

kmcfaul:
@mcoker For the fieldset html structure change, it looks like there's now a grid gap between groups from the wrapping form. Is there something I can do to suppress it or is an expected adjustment?

@jgiardino For the aria changes in the fieldsets, I'm using the group label as the aria-labelledby property of each fieldset (and when there are no groups, the previously passed in aria-label is used).

mcoker:
> @mcoker For the fieldset html structure change, it looks like there's now a grid gap between groups from the wrapping form. Is there something I can do to suppress it or is an expected adjustment?

Looks like you're missing a `.pf-c-form__group` as a direct child of `.pf-c-form`

screen shot 2019-03-08 at 1 36 14 pm

jgiardino:
> For the aria changes in the fieldsets, I'm using the group label as the aria-labelledby property of each fieldset (and when there are no groups, the previously passed in aria-label is used).

This approach sounds good. I just have a couple of comments/questions:

-  In the example below, where `id="Status"` is included, if the consumer defines something like "Lorem ipsum" as the group label,  this gets generated as `id="Lorem ipsum"` with a space, which isn't a valid id value. Can we convert the string to a valid value? Are there any standards we follow regarding syntax of id values that we should follow?
- The `id` should be on the `.pf-c-select__menu-group-title` element, not the `.pf-c-select__menu-group`. The `.pf-c-select__menu-group-title` has only the text that we want to use, but `.pf-c-select__menu-group` has the text plus all the checkbox labels.

```
<div id="Status" class="pf-c-select__menu-group">
  <div class="pf-c-select__menu-group-title">Status</div>
  <fieldset aria-labelledby="Status" class="pf-c-form__fieldset">
```

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -113,6 +121,43 @@ class Select extends React.Component {
               )}
             </React.Fragment>
           )}
+          {variant === 'checkbox' && (
+            <React.Fragment>
+              <SelectToggle
```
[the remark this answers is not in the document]
kmcfaul: Yeah I can make it common, I was split between doing that or maintaining one switch type logic for getting the right variant.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -1,16 +1,20 @@
...
+export const SelectVariant: {
+  single: 'single';
+  checkbox: 'checkbox';
 };
 
 export interface SelectProps extends HTMLProps<HTMLOptionElement> {
   isExpanded?: boolean;
+  isGrouped?: boolean;
   onToggle(value: boolean): void;
-  placeholderText?: string;
```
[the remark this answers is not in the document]
kmcfaul: Sounds good, will revert that until we can change the API.

kmcfaul:
@jgiardino I'm converting any non-alpha character from a group label into a hyphen in this latest commit, so that should resolve the spaces in IDs issue.

jgiardino:
Hi @kmcfaul, have you been able to work on the keyboard interaction for this component? I had posted a [comment in the issue](https://github.com/patternfly/patternfly-react/issues/1217#issuecomment-469490146) but wanted to mention that here so that it doesn't get lost. I provided similar feedback on the [context selector](https://github.com/patternfly/patternfly-react/pull/1526#issuecomment-471074929) and am curious if these suggestions are working out.

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -87,32 +93,60 @@ class Select extends React.Component {
...
+            id={selectToggleId}
+            parentRef={this.parentRef.current}
+            isExpanded={isExpanded}
+            onToggle={onToggle}
+            onEnter={this.onEnter}
+            onClose={this.onClose}
+            aria-labelledby={`${ariaLabelledBy} ${selectToggleId}`}
+            style={{ width }}
+          >
+            {variant === 'single' && (
```
tlabaj: nit. can you compare against the variant constant here.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectGroup.js
```diff
@@ -0,0 +1,40 @@
...
+
+const propTypes = {
+  /** Checkboxes within group */
+  children: PropTypes.node,
+  /** Additional classes added to the CheckboxSelectGroup control */
+  className: PropTypes.string,
+  /** Group label */
+  label: PropTypes.string,
+  /** Internal ID for group title */
+  titleId: PropTypes.string,
```
tlabaj: this says it is internal.  If so, we should not be exposing it to the consumer.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.js
```diff
@@ -0,0 +1,116 @@
...
+
+const propTypes = {
+  /** the value for the option */
+  children: PropTypes.string,
+  /** additional classes added to the Select Option */
+  className: PropTypes.string,
+  /** the value for the option */
+  value: PropTypes.string,
+  /** Internal index of the option */
+  index: PropTypes.number,
```
tlabaj: Same comment here about exposing internal index.

mcarrano:
This looks good @kmcfaul except that the badge to indicate the number of items selected should be at a fixed spacing following the dropdown label. See the visual spec below:

Screen Shot 2019-03-18 at 10 50 50 AM

@mcoker can you check this against the CSS implementation? It's hard to tell if this spacing is built in there.

mcoker:
@mcarrano this is currently broken in core, so this PR just reflects what's in core. @christiemolloy has an issue to address this - https://github.com/patternfly/patternfly-next/issues/1539

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectGroup.js
kmcfaul: I can remove it. I think the linter just complains about missing variables in the PropTypes declaration without it however, and I have gotten feedback before to add these kind of variables before. It would be nice to have a way to leave out these internal props when it gets translated into our docs.

kmcfaul:
@jgiardino New commit has updated keyboard interactions. I've added in a focus trap so both tabbing and arrow keys will work, and tab will also open the checkbox select. Additionally, the focus now highlights the input checkbox instead of the whole outer wrapper. **Single select's interaction and focusing remains the same.** 

@tlabaj Updated to using the SelectVariant constant for comparisons. I have left internal props in PropTypes for the moment, do you think they should still be removed despite the linter errors?

jgiardino:
Hey @kmcfaul, thanks for making updates!

> and tab will also open the checkbox select

Can you remove this behavior? We should only open the menu on the Enter key. When this toggle receives focus, the natural behavior of Tab is to move to the next element in the DOM that can receive focus. This will be the user's expectation. But also, the way it is now, it traps keyboard focus, and the user can't ever tab away from the menu.

> Additionally, the focus now highlights the input checkbox instead of the whole outer wrapper.

This is fine. This matches what I would expect and is consistent with how keyboard-only users are used to interacting with checkboxes. We could consider enhancing core to add css that highlights the whole list item when a checkbox has focus (using `:focus-within`), but I'd say that's a refinement we can add later. 

Some other things I noticed:

- Standard keyboard behavior for a checkbox will allow selecting using Space key, but this key doesn't work here.  Can you add this in?
- Can you remove `aria-haspopup="listbox"` from the Checkbox Select examples? The Single Select example has `role="listbox"` on the menu, which is why it also has `aria-haspopup="listbox"` on the toggle, but `role="listbox"` isn't applicable for the Checkbox Select examples. 
- The menu traps focus for the sighted keyboard-only user, which is great 👍, but not for screen readers 😿. This is something that I had noted for ContextSelector in #1526, and captured issue #1615 to generally catch all interaction patterns related to menus including this behavior. I would suggest not trying to add that behavior in this PR, and we can address it as a global pattern as part of #1615.

tlabaj:
@kmcfaul  I think you should leave out the props for now and tell the linter to ignore the line where you are getting the error.

kmcfaul:
> But also, the way it is now, it traps keyboard focus, and the user can't ever tab away from the menu.
This is the behavior of the focus trap. I can change the open/close behavior back though, so tab does not open the menu and if the menu is open tab will close it.

I will add in the Space bar behavior and fix the listbox tags.

Regarding the focus trap, I agree we may want to address this pattern through all components separately. It seems that the focus trap we import may not be the best option for screen readers. We might be able to extend our manual arrow key interactions to include tab and shift+tab however?

jgiardino:
> This is the behavior of the focus trap. I can change the open/close behavior back though, so tab does not open the menu and if the menu is open tab will close it.

Right now, when the menu is open, Tab will shift focus to the next checkbox. Can we keep that behavior? Also, except for when focus is on the first and last checkbox, the Tab key might not need to be handled with JS, since shifting focus to the next focusable element is the default behavior when using the Tab key. The only case we might need to handle is when focus is on the last checkbox, so that we can have Tab shift focus to the first checkbox, and vice versa for the first checkbox.

dlabaj:
@kmcfaul Looks like some of the JEST test failed.

mcarrano:
@rachael-phillips This looks good from a design and mouse interaction perspective except for the CSS issue on badge positioning. But looks like there still may be some keyboard interaction issues @jgiardino ?

jgiardino:
I tested the latest updates. The keyboard interaction works as expected! Thanks!

I did notice while reviewing that there are some weird behaviors when a menu is visible, and I scroll and click another menu. The page will scroll back to the menu that was previously visible.

I'm doing one more review of the aria attributes that are defined, and there are a few that need to be modified:

- `aria-selected` should be removed from `label.pf-c-check pf-c-select__menu-item`
- `aria-haspopup` - This is just a nit, but is it possible to not include this for these components instead of setting it to false? This is not a big deal if it's significant rework.
- `aria-hidden="true"` should be included for `.pf-c-select__menu-group-title`
- `aria-label` on the `<fieldset>` in Checkbox Select Input
  - There was previous discussion on this [here](https://github.com/patternfly/patternfly-react/pull/1487#issuecomment-469776901) and [here](https://github.com/patternfly/patternfly-react/pull/1487#issuecomment-471030433). In looking at this now, I see that the `aria-label` defined on `<Select>` is passed to the `<fieldset>`, but the `<fieldset>` also has `aria-labelledby` associated with the `{titleId}` value provided for `ariaLabelledBy` on `<Select>`. We should only have one of these defined at a time. I'm inclined to say that if a user explicitly defines a text string in `aria-label`, then we should only include that attribute, and not include `aria-labelledby`. Is that possible?
  - When I don't include `ariaLabelledBy` on the component, I still see `aria-labelledby` on the `<fieldset>` with no value. Is it possible to not include this attribute in those cases?
  - Also, is this one of those cases where this is an html attribute defined on the react component and getting passed down to one of the html elements in the component? In cases like this, are we documenting where these attributes get added? (e.g. aside from looking at this example, how would consumers know how this attribute is getting used?)

dlabaj:
@christiemolloy @mcarrano Is this all set for UX / CSS review?

mcoker:
@kmcfaul can you move the badge into the `.pf-c-select__toggle-wrapper` element? We made that change in https://github.com/patternfly/patternfly-next/pull/1595, so the positioning should be fixed now.

dlabaj:
@tlabaj This is good to go the badge offset should be fix when merged into master.

mcarrano:
Looks good @kmcfaul Changed to 'UX Approved'

priley86:
cc: @aljesusg - do you mind giving this a look and seeing if it fits your use case? If not, maybe we can start an issue, but terrific job exploring this w/ the selector pattern Kiali 👍

PR #1490, opened 2019-03-04
fix(navExpand): remove defaultExpand in favor of just using isExpanded

ibolton336:
**What**:
#1016 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Nav/NavExpandable.js
+++ b/packages/patternfly-4/react-core/src/components/Nav/NavExpandable.js
@@ -13,8 +13,6 @@ const propTypes = {
   title: PropTypes.string.isRequired,
   /** If defined, screen readers will read this text instead of the list title */
   srText: PropTypes.string,
-  /** If true will default the list to be expanded */
-  defaultExpanded: PropTypes.bool,
   /** Boolean to programatically expand or collapse section */
   isExpanded: PropTypes.bool,
   /** Anything that can be rendered inside of the expandable list */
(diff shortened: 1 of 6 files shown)


priley86:
this seems to have the desired effect since OpenShift is currently only using the `isExpanded` prop to drive this. I'm not 100% positive of the OpenShift UX rules on this, so please correct me if I'm wrong @rhamilto .

Opening a link with "Home" nav section expanded will correctly now open the "Workloads" section on route change:
pods-link

Opening a link with the "Home" section subsequently closed will correctly maintain its state after route change:
persist-close-state

Does this resolve @rhamilto ?

priley86:
on hard refresh, it seems to correctly expand the current nav for the currently matched route (since a child nav item is active):
hard-refresh

priley86:
we are still resolving one other edge case... the `isExpanded` state is not respected after nav toggle. This is almost there...

packages/patternfly-4/react-core/src/components/Nav/NavExpandable.js
```diff
@@ -44,28 +41,33 @@ const defaultProps = {
...
 class NavExpandable extends React.Component {
   id = this.props.id || getUniqueId();
+  state = {
+    expandedState: false
+  };
+  componentDidMount() {
+    this.setState({ expandedState: this.props.isExpanded });
+  }
+  componentDidUpdate(prevProps) {
+    if (this.props.isExpanded !== prevProps.isExpanded) {
```
priley86: should this be:
```
  componentDidUpdate() {
    if (this.props.isExpanded !== this.state.isExpanded) {
```
?

priley86: sorry:
```
if (this.props.isExpanded !== prevProps.isExpanded || this.props.isExpanded !== this.state.expandedState)
```

priley86: ok, made a few changes to address this. If the consumer is setting expandState externally, we should allow them to hook into this, otherwise they could be become out of sync. My changes here would allow the consumer to override the expand/collapse state at either the top level (Nav and the `onToggle`), or at a more granular level, the `NavExpandable` and the `onExpand` callback). They would then be able to hook into either (or neither if desired). See if this works on your side...
https://github.com/priley86/patternfly-react/commit/ee30e2002615f561c93f5ddf471678fd423f1b9a

priley86: downstream explanation/relevant changes...
https://github.com/openshift/console/pull/1238#issuecomment-469763337

rhamilto:
So close.  Console warning needs fixing (which @priley86 is doing).

```
Warning: Unknown event handler property `onExpand`. It will be ignored.
    in li (created by NavToggle)
    in NavToggle
    in NavExpandable (created by NavSection)
    in NavSection (created by Connect(NavSection))
    in Connect(NavSection) (created by Navigation)
    in ul (created by NavList)
    in NavList (created by Navigation)
    in nav (created by Nav)
    in Nav (created by Navigation)
    in div (created by PageSidebar)
    in PageSidebar (created by Navigation)
    in Navigation (created by App)
    in div (created by Page)
    in Page (created by App)
    in App (created by Route)
    in Route
    in Router
    in Provider
```

priley86:
thanks @rhamilto - https://github.com/patternfly/patternfly-react/pull/1504 is up...  thanks for your patience 👏

PR #1499, opened 2019-03-05
chore(build): suppress travis bugs

redallen:
**What**: Don't double-build bot-created branches, don't fail releases if the Github release doesn't work.

**Additional issues**:


dgutride:
@jschuler  - this can go in after we promote

PR #1503, opened 2019-03-05
fix(Table): Expandable row should not prevent select all checkbox fro…

jharting:
…m being checked

**What**:

If the table combines expandable and selectable rows then select-all checkbox never renders checked.
This happens because areAllRowsSelected() considers all rows even those that do not have checkboxes.
This change modifies areAllRowsSelected() to only consider rows with checkboxes.

**Additional issues**:


jharting:
@karelhala FYI

redallen:

jharting:
@redallen done

PR #1505, opened 2019-03-05
fix(iconOnly): Add conditional for icon only styles

ibolton336:
**What**:
#1169 
Add css conditional for icon only drop down case

**Additional issues**:


priley86:
looks good! thanks @ibolton336 👍

PR #1511, opened 2019-03-06
feat(accordion): Add accordion to pf

### LINKED ISSUE #1387: Introduce the Accordion Component
The Accordion component was just merged in Core - https://github.com/patternfly/patternfly-next/pull/1357.

For any CSS questions talk with @christiemolloy

ibolton336:
**What**: closes #1387 

**Additional issues**:


redallen:

packages/patternfly-4/react-core/src/components/Accordion/AccordionToggle.js
```diff
@@ -0,0 +1,64 @@
...
+        className={css(styles.accordionToggle, isExpanded && styles.modifiers.expanded, className)}
+        {...props}
+        variant="plain"
+        aria-controls={ariaControls !== '' && ariaControls}
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabel !== 'Details' ? null : `${id}`}
+        aria-expanded={isExpanded}
+      >
+        <span className={css(styles.accordionToggleText)}>{children}</span>
+        <span className={css(styles.accordionToggleIcon)}>
```
ibolton336: This class from core is missing the transition that is seen in other similar components. https://github.com/patternfly/patternfly-next/issues/1532 created for this.

packages/patternfly-4/react-core/src/components/Accordion/AccordionItem.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AccordionItem = ({ children, className, 'aria-labelledby': ariaLabelledBy, ...props }) => (
```
kmcfaul: It looks like className and aria-labelledby aren't being applied to the AccordionItem

ibolton336: 👍 just updated

packages/patternfly-4/react-core/src/components/Accordion/AccordionContent.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/patternfly/components/Accordion/accordion.css';
+
+const AccordionContent = ({ className, children, id, isHidden, isFixed, 'aria-label': ariaLabel, ...props }) => (
+  <dd
+    id={id}
+    className={css(styles.accordionExpandedContent, isFixed && styles.modifiers.fixed)}
```
kmcfaul: Missing className from props.

redallen:
CircleCI times out before it shows this, but there is this error that needs to be fixed that happens when building the docs locally @ibolton336 :
```
@patternfly/react-docs: > 1 | export var matches = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
@patternfly/react-docs:     |                      ^
@patternfly/react-docs:   2 | 
@patternfly/react-docs:   3 | if (!Element.prototype.matches) {
@patternfly/react-docs:   4 |   Element.prototype.matches = matches;
@patternfly/react-docs: 
@patternfly/react-docs:   WebpackError: ReferenceError: Element is not defined
@patternfly/react-docs:   
@patternfly/react-docs:   - closestPolyfill.js:1 Module.../../patternfly-3/patternfly-react/dist/esm/com    mon/closestPolyfill.js
@patternfly/react-docs:     lib/Users/zallen/src/patternfly-react-master/packages/patternfly-3/patternfl    y-react/dist/esm/common/closestPolyfill.js:1:22
```

Run `yarn build:docs` locally to reproduce

packages/patternfly-4/react-core/src/components/Accordion/examples/SimpleAccordion.js
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
```
redallen: Try `import  * as React from 'react';`, this has fixed issues with builds in the past (and we'll just have to change it on the Typescript conversion anyways)

mcarrano:
I don't think that the behavior is quite right here. In a classic accordion control, only one panel is ever opened at a time. See the behavior of the PF3 Accordion component, for example: https://www.patternfly.org/pattern-library/widgets/#accordion  Note here that opening a new panel closes the current one. That said, it could be useful to have both variations - one with simple collapsible panels as you have now and another withe the accordion behavior as described above.

The other thing I questioned was the difference between the Simple and the Fixed accordion. My understanding is that the intent of the fixed accordion is that the height is constrained. So if the contents will not fit within the max height, a scrollbar will be exposed within the open panel to scroll the content. Is that also what you intended @mcoker ?

mcoker:
@mcarrano Agreed re: having the option for both types of behavior.

Re: the fixed examples, yep - looking at the [pf3 fixed height accordion](https://www.patternfly.org/pattern-library/widgets/#fixed-height-accordion), this accordion is the same where the max height is defined on the expandable panel for an accordion item. In the PF3 examples, that max-height is 69px. In the PF4 component, looks like we went with 150px.

Also the `.pf-c-accordion__toggle` button shouldn't be a PF4 `<Button>` component (currently `.pf-c-button.pf-m-plain`), it should just use a regular `<button>` html element.

And when an item is expanded, we also need to add `.pf-m-expanded` to the expanded `dd`.

tlabaj:
@mcarrano @mcoker We are going to merge this and make the updates in a new PR.

jgiardino:
Some initial comments after a quick review...

`aria-label` shouldn't be a required prop. Also, it makes sense that it's included on the `<dl>` so I will add an issue to track that.

The section headings that are expand/collapse buttons have `aria-label="Details"` and also `aria-labelledby="..."` defined. These attributes should not be there. This overrides the actual heading text that's specified, preventing that text from being announced.

Core does not include `aria-controls`. This attribute is noted as optional in the [WAI-ARIA practices for disclosure (show/hide)](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure). I would suggest leaving it out, since the controlled contents are immediately after the button.

PR #1515, opened 2019-03-06
chore(LoginPage): a single function to update password and username

### LINKED ISSUE #1462: Login form redundant change handler
Create one change handler to handle all form field updates rather than seperate handlers for username and password.

boaz0:
**What**:

closes #1462 

//cc @ibolton336

### DIFF
--- a/packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.docs.js
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/LoginPage.docs.js
@@ -1,5 +1,6 @@
 import { LoginPage, LoginForm } from '@patternfly/react-core';
 import SimpleLoginPage from './examples/SimpleLoginPage';
+import OneHandlerLoginPage from './examples/OneHandlerLoginPage';
 
 export default {
   title: 'LoginPage',
@@ -8,6 +9,9 @@ export default {
     LoginForm
   },
   variablesRoot: 'pf-c-login',
-  examples: [{ component: SimpleLoginPage, title: 'Simple LoginPage' }],
+  examples: [
+    { component: SimpleLoginPage, title: 'Simple LoginPage' },
+    { component: OneHandlerLoginPage, title: 'One Handler for Username and Password' }
+  ],
   fullPageOnly: true
 };
--- a/packages/patternfly-4/react-core/src/components/LoginPage/examples/OneHandlerLoginPage.js
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/examples/OneHandlerLoginPage.js
@@ -1,0 +1,165 @@
+import React from 'react';
+import brandImg from './pf_logo_color.svg';
+import {
+  LoginFooterItem,
+  LoginForm,
+  LoginMainFooterBandItem,
+  LoginMainFooterLinksItem,
+  LoginPage,
+  BackgroundImageSrc,
+  ListItem
+} from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+
+/**
+ * Note: When using background-filter.svg, you must also include #image_overlay as the fragment identifier
+ */
+
+const images = {
+  [BackgroundImageSrc.lg]: '/assets/images/pfbg_1200.jpg',
+  [BackgroundImageSrc.sm]: '/assets/images/pfbg_768.jpg',
+  [BackgroundImageSrc.sm2x]: '/assets/images/pfbg_768@2x.jpg',
+  [BackgroundImageSrc.xs]: '/assets/images/pfbg_576.jpg',
+  [BackgroundImageSrc.xs2x]: '/assets/images/pfbg_576@2x.jpg',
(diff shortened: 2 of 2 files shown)


packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.test.js
```diff
@@ -74,3 +74,21 @@ test('LoginForm with rememberMeLabel and rememberMeAriaLabel does not generates
...
+});
+
+test('should call onChangeText callback when is set and password is changed', () => {
+  const mockFn = jest.fn();
+  const view = shallow(
+    <LoginForm onChangePassword={mockFn} rememberMeLabel="Login Form" rememberMeAriaLabel="Login Form" />
+  );
+  view.find('#pf-login-password-id').simulate('change');
+  expect(mockFn).toHaveBeenCalled();
+});
```
tlabaj: should probably add a test case for when the user defines both onChangeUsername/onChangePassword and OnChangeText.

boaz0: You're right. Will do

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.test.js
```diff
@@ -74,3 +74,42 @@ test('LoginForm with rememberMeLabel and rememberMeAriaLabel does not generates
...
+test('should call onChangeUsername when is set and username is changed', () => {
+  const mockFn = jest.fn();
+  const view = shallow(
+    <LoginForm onChangeUsername={mockFn} rememberMeLabel="Login Form" rememberMeAriaLabel="Login Form" />
+  );
+  view.find('#pf-login-username-id').simulate('change');
+  expect(mockFn).toHaveBeenCalled();
+});
+
+test('should call onChangeUsername callback when is set and password is changed', () => {
```
tlabaj: Do you mean onChangePassword here?

boaz0: Thanks!

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
```diff
@@ -58,18 +60,19 @@ const defaultProps = {
   helperText: null,
   usernameLabel: 'Username',
   usernameValue: '',
-  onChangeUsername: () => undefined,
+  onChangeUsername: null,
   isValidUsername: true,
   passwordLabel: 'Password',
   passwordValue: '',
-  onChangePassword: () => undefined,
+  onChangePassword: null,
```
dlabaj: Can we keep this as undefined?

boaz0: Sure we can.

The reason I changed it from `() => undefined` to `null` is that I can't compare functions but `null` and `undefined` I can.

jschuler:
It may seem redundant, but it gives user more flexibility to choose to either define separate event handlers for username and password, or the same. I don't think we should add a third option here, that makes the API more confusing
```
onChangeUsername={this.handleChangeText}
onChangePassword={this.handleChangeText}
```

boaz0:
@jschuler thanks for your feedback.
So you're suggesting to remove the `onChangeText` prop, is that right?

Yeah, I don't mind doing so, I just thought that eventually there is going to be only one prop that handles the events of password and username inputs. 

@tlabaj & @ibolton336  what do you think? Should we discard the `onChangeText` prop and keep the old `onChangePassword` and `onChangeUsername`?

jschuler:
@boaz0 I could see us changing the API to have just one prop, but then we need to inform the user which input field the change came from and this would also be a breaking change

dlabaj:
Talked with @ibolton336 and he's good with the solution purposed by @jschuler Do we just want to add an example instead?

boaz0:
@dlabaj yep, just need to change the example. I will update this PR in the next hours.

boaz0:
@tlabaj no problem - created a new example.

jschuler:
@boaz0 basically it's just one prop changed to point to same callback function, no? I don't see too much value in having a whole other example just for this. If you guys are really convinced we need another example for this then ok... if you agree we don't need please 👍 if you disagree 👎 . I'll defer to the outcome of that :)

boaz0:
@jschuler yes, you're right. It is just one prop change. 

I believe there is trade off:
On one hand, I would like to provide all the possible ways to use any component. On the other hand, it does seem like a redundant to have those two examples.

I wish I had a solid opinion :thinking:  but don't have much of experience to decide :man_shrugging:

Thanks. :smile_cat:

jschuler:
Hi @boaz0 , we have many components with various callback props, I believe in this case it would be too much to showcase this variation to bind 2 callback props to the same function, as otherwise the example is nearly identical to the previous one and this could add noise/confusion to the documentation. Thanks!

boaz0:
No problem. Thank you.

PR #1520, opened 2019-03-07
feat(Table): support custom components in expandable rows

suomiy:
by using expandable cellFormatter or ExpandableRowContent component directly

- exposed isRowExpanded function for extensions to use

https://github.com/patternfly/patternfly-react/pull/1227 is dependent on this PR.

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Body.js
+++ b/packages/patternfly-4/react-table/src/components/Table/Body.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Body } from 'reactabular-table';
 import PropTypes from 'prop-types';
 import { TableContext } from './Table';
+import { isRowExpanded } from './utils';
 
 const propTypes = {
   /** Additional classes for table body. */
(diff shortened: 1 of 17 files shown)


packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -32,13 +33,6 @@ class ContextBody extends React.Component {
...
-  parentsExpanded(parentId) {
-    const { rows } = this.props;
-    return rows[parentId].hasOwnProperty('parent')
-      ? this.parentsExpanded(rows[parentId].parent)
-      : rows[parentId].isOpen;
-  }
-
   mapCells = (row, rowKey) => {
     const { headerData } = this.props;
     let shiftKey = Boolean(headerData[0] && headerData[0].extraParams.onSelect);
```
jschuler: It would be good to see more comments, for example it looks like the shiftKey can take values of 0 | 1 | 2 here and then I am not sure how it is used

jschuler:
LGTM but this needs a look by @priley86 @karelhala

priley86:
thanks for reviewing @jschuler. I don't have any issue w/ this and it seems a nice addition, however, @karelhala is much more familiar.

packages/patternfly-4/react-table/src/components/Table/Body.js
suomiy: I added the comments, refactored that part and fixed a small bug in the process

suomiy: I hope it is readable now

suomiy:
\+ enhanced the TS a bit

priley86:
@jschuler @karelhala - any additional thoughts here?

dtaylor113:
> Codewise looks good. I am just a bit concern about breaking current API, can we document that somewhere so consumers will know about that? Also it would be good if we bump the version of table package so everybody knows that this API changed.

Hi, is this a breaking change?

karelhala:
> Hi, is this a breaking change?

Not really if consumers are going to use the old way it's just going to look differently, but the table will still work and expandable actions will work as well.

dtaylor113:
great!  thanks @karelhala

suomiy:
> > > Codewise looks good. I am just a bit concern about breaking current API, can we document that somewhere so consumers will know about that? Also it would be good if we bump the version of table package so everybody knows that this API changed.

> > Hi, is this a breaking change?

> Not really if consumers are going to use the old way it's just going to look differently, but the table will still work and expandable actions will work as well.

Should I update the version? and write something to the `CHANGELOG` ?

suomiy:
I mean, should the version be `1.3.0` ?

jschuler:
No need to manually increment the version or update the changelog, this should be handled correctly by the build

Issue #1525, opened 2019-03-07
PF4 TextInput: TextInput should be modified to allow for aria-labelledby


tlabaj:
TextInput should be modified to allow for aria-labelledby in addition to id or aria-label.  Currently a console error is generated when the TeaxtInput Component is missing an aria-label or an id.  The component should be modified so that it Errors if it is missing and Aria-label, id or an aria-labelledby.

tlabaj:
Closed by #1526

Issue #1293, opened 2019-01-30
Add Context Selector component


mcoker:
New component - https://pf4.patternfly.org/components/ContextSelector/examples/

https://github.com/patternfly/patternfly-next/pull/1288

jgiardino:
There was some discussion in the a11y meeting today related to this component. Here's a recap of some things discussed:

- This component will use the behavior described in this issue: https://github.com/patternfly/patternfly-react/issues/839 (i.e. the panel behaves similar to a popover)
  - #839 is intended to allow consumers to use the Dropdown component and customize what displays in the panel that pops up, when they need something more than just a basic menu. 
  - Having the panel behave like a popover allows us to manage the visibility relative to focus. The panel will trap focus, and isn't hidden until the user explicitly exits it. This will be the default behavior of the Dropdown Panel variation of the dropdown.
- Since for this case we know what the contents of the panel are, we can also include additional keyboard interaction behavior so that a sighted keyboard-only user will have similar experiences for components that look the same (i.e. toggle that displays something with a menu). We might not be able to have complete parity, but maybe at a minimum can:
  - enable the arrow keys to shift focus between menu items
  - have Tab on the last menu item exit the menu
- It might make sense to start off with basic functionality first (e.g. basic popover functionality) and do a round of testing, and incrementally add the arrow key and tab key functionality then test again, in case these updates cause the component to not work at all in one of the screen readers.

dlabaj:
#1526 Fixes this issue.

Issue #1230, opened 2019-01-22
Dropdown cut off if rendered inside element with `overflow: hidden`


TimoSta:
**Problem:**
If a PF4 dropdown is rendered inside a an element that is set to `overflow: hidden`, any part of the dropdown that that is larger than the container is clipped:

![image](https://user-images.githubusercontent.com/5621996/51552845-8256a480-1e3f-11e9-9e7a-4065817c7136.png)

See [this example for a minimal reproduction](https://codesandbox.io/s/j35x1my5).

---

 **Background:**
We are trying to render a dropdown menu on a tab component, to offer options like _duplicate tab_ or _delete tab_.

To make the tabs render properly, we need to set the CSS overflow property of the container to `hidden`. As a result, the dropdown menu on the tab becomes unusable, as it is confined to the space of the tabbar.

---

**Potential Solution:**
A possible solution to this would be to render the dropdown menu not as a child of the toggle in the DOM tree, but rather at the root of the tree, and position it correctly using absolute values. This way, the dropdown is still rendered in the correct place on the UI, but is no longer affected by styles of the parent elements in the tree.

This approach has been taken for example by  Tippy.js, and as a result is in use by PF4 for the tooltip/popover components, and also by Material-UI for [their implementation of dropdown menus](https://material-ui.com/demos/menus/).

---

If you are generally in favor of this approach, I'm more than happy to give the implementation a shot and see how it turns out!

jschuler:
Great suggestion, we should provide users the option to attach the menu to the document body or wherever they choose. We already do this for some other components, like Tooltip and Popover, I think this is another good use-case.

rachael-phillips:
Hi @TimoSta ! Is this something you would be interested in working on?

rachael-phillips:
Hi @TimoSta how is it going on this issue?

rachael-phillips:
Closing because this has been inactive for a long period of time

PR #1526, opened 2019-03-07
feat(ContextSelector): Add PF4 ContextSelector Component

tlabaj:
#1293, #1525

**What**:

**Additional issues**:


jgiardino:
One thing I noticed is that during a search, if I enter a lowercase "a", it returns nothing. Should the search be case-insensitive for this component?

The component isn't keyboard accessible at the moment. For the first implementation, I think the simplest solution is to handle the panel similar to how we handle the popover or modal with regard to trapping and managing focus.

- Given the toggle has focus, when the user hits Enter (or the user clicks the toggle), then:
  - The panel `.pf-c-context-selector__menu` traps focus - This behavior was added to the modal in https://github.com/patternfly/patternfly-react/pull/1011. I'm not sure if the popover uses a similar or different method, and if so, which one would make the most sense for the dropdown. The popover might make the most sense given that it also requires the ability to handle positioning relative to the toggle.
  - The first element in the panel that can receive focus has focus. In this case it would be the search field.
- Given the panel is open, 
  - When the user hits Tab, then the next element that can receive focus gets focus (and Shift-Tab goes to the previous item that can receive focus)
  - And given the last item has focus, when the user hits Tab...?
    - In the modal or popover, we would shift focus to the first element that can receive focus. I think for now, that might be the simplest thing to do here. We can always refine this behavior once we have something that works across all experiences (mouse, keyboard, screen reader), but I would recommend handling this as a separate PR given the amount of effort involved.

tlabaj:
@Jenn, the application implements the search functionality.  I can change the example to be case insensitive if you would like.

jgiardino:
Thanks, @tlabaj!

> @jgiardino the application implements the search functionality. I can change the example to be case insensitive if you would like.

I'm not sure what the general expectation is here. That makes sense that we would be more hands-off and rely on consumers to handle that functionality. But I could also see the case where consumers would want to see an example that shows how we would recommend handling that functionality. 🤷‍♀️ I will totally defer to anyone who has a strong opinion about that.

> Not sure what I changed. I added the focus trap similar to popover like we talked about... I will have to investigate and figure out what is going on.

One thing you might want to check is whether there is anything else in the component that might be managing focus, and remove that. E.g. if you included some of the keyboard navigation for the menu, maybe there's a conflict there.

packages/patternfly-4/react-core/src/components/ContextSelector/ContextSelector.js
```diff
@@ -0,0 +1,138 @@
...
+      screenReaderLabel,
+      toggleText,
+      searchButtonAriaLabel,
+      searchInputValue,
+      onSearchInputChange,
+      searchInputPlaceholder,
+      ...props
+    } = this.props;
+    return (
+      <div className={css(styles.contextSelector, props.className)} ref={this.parentRef} {...props}>
```
kmcfaul: props.className can just be className.

kmcfaul:
Also noticed what Jen mentioned, it looks like the search is case sensitive at the moment. Additionally, if you backspace out some characters (Aw > A > AW to search AWS, or Opens > Open > OpenS to search OpenShift) to correct the case it doesn't look like it re-searches, you have to back out the entire string. 

When the list is filtered by the text search and you click to select the item, it looks like it is selecting based on the current index for that filtered list, so it selects the wrong item based on the whole refCollection.

packages/patternfly-4/react-core/src/components/ContextSelector/ContextSelectorToggle.js
```diff
@@ -0,0 +1,133 @@
...
+        }}
+        className={css(
+          styles.contextSelectorToggle,
+          isFocused && styles.modifiers.focus,
+          isHovered && styles.modifiers.hover,
+          isActive && styles.modifiers.active,
+          className
+        )}
+        type="button"
+        onClick={_event => {
```
jschuler:
```suggestion
        onClick={() => {
```

packages/patternfly-4/react-core/src/components/ContextSelector/ContextSelectorItem.js
```diff
@@ -0,0 +1,72 @@
...
+                styles.contextSelectorMenuListItem,
+                isDisabled && styles.modifiers.disabled,
+                isHovered && styles.modifiers.hover,
+                className
+              )}
+              ref={this.ref}
+              onClick={event => {
+                if (!isDisabled) {
+                  onClick && onClick(event);
+                  onSelect && onSelect(event, children);
```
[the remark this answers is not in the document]
tlabaj: I believe it coms in through the provide/consumer like Select.  I followed that model.

packages/patternfly-4/react-core/src/components/ContextSelector/ContextSelectorToggle.js
tlabaj: I passed the event incase the consumer may want to look at it.

christiemolloy:
@mcarrano should we have an example where you type in the search input, click the search icon and then it searches vs the implementation here where it automatically searches - it just seems as though the button is redundant, and I wanted to get your input on it.

jgiardino:
Great job on this! I just reviewed the updated keyboard and screen reader behavior (in VO). Here's what I noticed:

👍 Behaviors that work as expected:
- Focus lands on the search input when the menu first opens
- For the sighted keyboard-only user, I can only tab through the items in the panel, and focus shifts from the last item to the first item and vice-versa. 
  - Just FYI, I had included a comment in https://github.com/patternfly/patternfly-react/issues/1293 about ways we can refine the keyboard interaction, but my preference would be to defer these as future enhancements
- When the panel is visible, hitting Esc will dismiss the panel

👎 Behaviors that don't work as expected:
- When focus is on the toggle, I should be able to hit Enter to open the panel, but nothing happens.
- When using a screen reader, we should be trapping focus, but currently the screen reader has full access to the DOM. Both the Modal and Popover do this. I think they handle this differently but you might be able to look at one of those components to see how to implement that here. 

And then just a question, not sure who it's for... I noticed that when tabbing through the UI, the Search button can receive focus, but I don't think this button actually does anything. Should we reconsider what type of element this is (for example, is it more of a label for the input than a submit button)? cc'ing @mcarrano on this one. I don't think this is something that needs to be resolved now if we do decide to make changes.

mcarrano:
@christiemolloy @jgiardino I was looking at this earlier today and agree with your comments about the search button being superfluous. Maybe we don't need it. The PF3 design/implementation did not have an explicit search button. @kybaker do you recall the history on this? Is there a reason we included a search button in the design rather than just making this like a type ahead select?

jgiardino:
@mcarrano I just want to add a comment that currently the search icon serves as the visible label for the input, which is required for WCAG AA compliance. It can still be there without being a button.

mcarrano:
Yeah, that makes sense. This was the PF3 design.

Screen Shot 2019-03-18 at 5 01 11 PM

jschuler:
@tlabaj Can you also add the Context Selector to the page demos?

tlabaj:
@mcarrano @jgiardino so will the search button be changing to an icon?

jgiardino:
> Can you also add the Context Selector to the page demos?

👍 I agree, it would be nice to test this in the context of the demo.

> @mcarrano @jgiardino so will the search button be changing to an icon?

I would say that change could be a separate issue, and that we address it in core first. I wouldn't hold this PR up for that since it's not a bug, just a refinement. 

I just reviewed this again. The issue I noted with handling Enter key on the toggle is fixed, which is awesome! 

I tested in JAWS to see if adding `role="dialog" aria-modal="true"` would have any effect on trapping screen reader focus, and it doesn't. 😿 

Considering the differences between popover (which uses tippy) and modal, I wonder if tippy would be our best option for this component. Both require separating the popup element from the `<div>` that contains the main contents, but the modal doesn't require positioning, whereas the popover does. Also, being able to separate the panel and position it on top of the other contents might also let us avoid issues like #1230. I'm fine with capturing this last a11y item in a separate issue if this component is needed and/or this would require a significant amount of work. If we do capture this in a separate issue, then I would remove `aria-modal="true"` since it isn't really modal yet. 

Other than that, there are just a couple of more things I noticed that would be simple updates:

- I just noticed that the order of `id`s on the toggle is reversed. The label id should be first, followed by the button. 
- We can keep `role="dialog"` on the panel, but we should add either `aria-label` or `aria-labelledby` to describe it. 
  - One suggestion is to set `aria-labelledby` to the id of the string defined for `screenReaderLabel`, to avoid adding additional props. But "Selected Project [colon] [dialog]" sounds a little off as the dialog title. Maybe just setting `screenReaderLabel` to "Project" in our example would work for both the toggle and the dialog? This would result in screen readers announcing "Project AWS button collapsed" for the toggle and "Project dialog" for the panel.
  - Another suggestion is to add a prop, similar to `screenReaderLabel` for the toggle, but this prop would be for the panel, and would take a text string and add `aria-label="[text string]"` to the panel. Then in our example, we could define "Select Project" for that prop. This would result in screen readers announcing "Selected Project [colon] AWS button collapsed" for the toggle and "Select Project dialog" for the panel.
  - Let me know if you have a preference on this and I can follow up with design on strings if needed.

tlabaj:
Issue #1615 has been opened and will address remaining accessibility concerns.

dlabaj:
@jgiardino @mcarrano Is enter after typing in search suppose to select the first item in the dropdown from the search results?

jschuler:
IMO button does not make sense if you use typeahead select

dlabaj:
@jschuler @tlabaj I'm going to take this in and if we want to change what happens on enter we can create a separate issue.

PR #1535, opened 2019-03-08
fix(tsconfig): avatar tsx change

priley86:
**What**:
This change disables `allowSyntheticDefaultImports` and `esModuleInterop`. From a library perspective, I believe we will need to ensure that we can support Typescript consumers who have this option disabled OR enabled. This was discovered today while testing OpenShift tests downstream. 

It is generally considered good practice now to enable these flags (and from a Node JS/Common JS perspective) this is the typical path, however some consumers may not elect to enable them.

This should resolve the current issue and help us strategize for future TSX conversion.

cc: @dgutride @rhamilto @spadgett @tlabaj 

For additional detail, with this flags disabled downstream, but enabled upstream, we'll run into these:
```
ERROR in /Users/priley/GitHub/web-ui/frontend/node_modules/@patternfly/react-core/dist/js/components/Avatar/Avatar.d.ts
(1,8): Module '"/Users/priley/GitHub/web-ui/frontend/node_modules/@types/react/index"' has no default export.
```

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
+++ b/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
@@ -1,8 +1,6 @@
-
 import styles from '@patternfly/patternfly/components/Avatar/avatar.css';
 import { css } from '@patternfly/react-styles';
-import React, { DetailedHTMLProps, ImgHTMLAttributes } from 'react';
-
+import * as React from 'react';
 
 const defaultProps = {
   className: '',
@@ -12,7 +10,8 @@ const defaultProps = {
 /**
  * Column properties.
  */
-export interface AvatarProps extends DetailedHTMLProps<ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement>{
+export interface AvatarProps
+  extends React.DetailedHTMLProps<React.ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement> {
   /** Additional classes added to the Avatar. */
   className?: string;
   /** Attribute that specifies the URL of the image for the Avatar. */
@@ -21,7 +20,7 @@ export interface AvatarProps extends DetailedHTMLProps<ImgHTMLAttributes<HTMLIma
   alt: string;
 }
 
-const Avatar:React.FunctionComponent<AvatarProps> = (props) => (
+const Avatar: React.FunctionComponent<AvatarProps> = props => (
   <img {...props} className={css(styles.avatar, props.className)} />
 );
 
--- a/packages/patternfly-4/react-core/src/tsconfig.dts.json
+++ b/packages/patternfly-4/react-core/src/tsconfig.dts.json
@@ -1,35 +1,23 @@
 {
   "compilerOptions": {
     "outDir": "../dist/js",
-    "allowSyntheticDefaultImports": true,
-    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": false,
+    "esModuleInterop": false,
     "baseUrl": ".",
     "moduleResolution": "node",
     "declaration": true,
     "emitDeclarationOnly": true,
     "skipLibCheck": true,
     "sourceMap": false,
-    "lib": [
-      "es2017",
-      "dom"
-    ],
+    "lib": ["es2017", "dom"],
     "noEmit": false,
     "module": "commonjs",
     "jsx": "react",
     "target": "es5",
     "paths": {
-      "*": [
-        "../dist/esm/*"
-      ]
+      "*": ["../dist/esm/*"]
     }
   },
-  "include": [
-    "./**/*"
-  ],
-  "exclude": [
-    "./**/*.d.ts",
-    "./**/**.test.tsx",
-    "./**/**.test.ts",
-    "./**/examples/**"
-  ]
+  "include": ["./**/*"],
+  "exclude": ["./**/*.d.ts", "./**/**.test.tsx", "./**/**.test.ts", "./**/examples/**"]
 }


rhamilto:
Tested and confirmed good.  LGTM.

packages/patternfly-4/react-core/src/tsconfig.dts.json
```diff
@@ -1,35 +1,23 @@
 {
   "compilerOptions": {
     "outDir": "../dist/js",
-    "allowSyntheticDefaultImports": true,
-    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": false,
```
dlabaj: Should we just remove this it will default to esModuleInterop from looking at the typescript documentation I believe.

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -1,8 +1,6 @@
-
 import styles from '@patternfly/patternfly/components/Avatar/avatar.css';
 import { css } from '@patternfly/react-styles';
-import React, { DetailedHTMLProps, ImgHTMLAttributes } from 'react';
```
dgutride: Can we use the import { name, name } from 'react' syntax instead of * as React please?  We might need to add imports to this line instead of switching to the line below.

priley86: if we remove the` import * as React` it would compile to:
```
import { ReactNode } from 'react';

const foo: ReactNode = <div />
```
for example... and in this case:
```
const foo = React.createElement('div') //React is undefined
```

priley86: Seems that React is currently only exported as a default export

priley86: https://unpkg.com/react@16.6.3/index.js

dmiller9911:
@dlabaj Note this will not have any effect on the treeshaking of React. React does not treeshake. All of their optimizations happens during their build process. They export a commonJS module (not ES) since they allow SSR in a node context. adding `import * as React from 'react'` is no different than `import React from 'react'`.

priley86:
For historical purpose with this PR, the proposed downstream change reverts the original proposed downstream change which would support the current configuration here today:
https://github.com/priley86/web-ui/commit/971ecf91f7aac9f4a24e87ca171442b44788c14e#diff-64282092d7b77f064e3f06a569429a99

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dmiller9911: For some more background on this, it all goes back to how babel and TS decided to handle commonjs modules interop with ES Module syntax. React exports a commonJS module (as seen in the link above). They do this instead of an ESModule since they support running React in a Node.js context. Example

Given the following file:
```js
function foo() {}
module.exports = foo;
```
You can import this with babel doing the following
```js
import foo from './foo'
```
With TypeScript Pre 2.7 you had to import the module doing the following
```ts
import * as foo from './foo';
```
Node finally decided that when importing a commonJS module from an ES Module you can import it using the same syntax babel decided on. With TypeScript 2.7+ `esModuleInterop` was added to match this behavior. 

This is great for applications since they can now have standard code on server and client. For libraries though this poses a problem since it requires any consumers of a library to also set `allowSyntheticDefaultImports` for the typings to be correct.

According to the React source and the React typings React does not have a `default` export. They only export a namespace. So if PF emits typing files that have `import React from 'react'` things will fail until a consumer sets `allowSyntheticDefaultImports` to `true`. On the flip side if PF exports a typing with `import * as React from 'react'` then both consumers using esModuleInterop and not using it will work. 

As for the question above for can we use the syntax of `import { FunctionComponent, ImgHTMLAttributes } from 'react'; That is not possible. For React to work `React` (the variable) has to be within the module (unless configured to consume from a CDN and it is a global var). This is the same for any project using babel or typescript. Anything needed off of the React namespace can be referenced from `React.FunctionComponent` instead of importing it. There are ways around this by setting `jsxFactory` in tsconfig of `jsxPragma` in babel config, but it is very much nonstandard and would likely require additional lint or IDE configuration.

If anyone is more curious about this I definitely recommend messing around with the [babel repl](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AoctCAOwGcsII4BeOAHgBNgA3OAegB8pIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.3.4) to view what is actually output when running through a compiler. The best and most up to date documentation is always source code.

priley86: Further reading in this React [issue](https://github.com/facebook/react/issues/11503) would seem to support this.

rhamilto:
Happy Monday.  I am waiting on a resolution to this PR in order to determine a course of action for OpenShift console.  I want to make sure it doesn't get forgotten while @priley86 is out on PTO.  Thank you.

packages/patternfly-4/react-core/src/tsconfig.dts.json
redallen: I'll remove it in my next PR which does some typescript groundwork.

redallen:
@rhamilto Fixed in @patternfly/react-core 2.4.1

PR #1538, opened 2019-03-11
feat(clipboardcopy): add clipboardcopy to PF4

### LINKED ISSUE #1417: Copy to Clipboard
Referencing Core issue: https://github.com/patternfly/patternfly-next/issues/505

boaz0:
**What**:

closes #1417 

Adding copy to clipboard into patternfly 4 react-core.


dgutride:
**https://github.com/patternfly/patternfly-next/issues/505** - here is the relevant patternfly core issue/design links

rachael-phillips:
Related issue: https://github.com/patternfly/patternfly-react/issues/1417

christiemolloy:
@boaz0 I believe that when you click on the "copy to clipboard" button the popover should remain for 2 seconds approximately. @mcarrano are you able to check on this? Otherwise looks really good!

mcarrano:
@boaz0 @christiemolloy Yes, the message that appears after you click Copy should stay for at least 2 seconds (maybe more?). We did not have a spec on this but I would try adding a 2 second timer and see how that feels.

Also, I noticed that in the Expandable version, the expansion area is not editable. It needs to be if a user is to be able to edit the entire contents. Also note that per the design. we should show the truncation ellipses when the text in the field gets truncated. See below.

Screen Shot 2019-03-13 at 5 06 54 PM

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.js
```diff
@@ -0,0 +1,180 @@
...
+import { CopyIcon } from '@patternfly/react-icons';
+import PropTypes from 'prop-types';
+import { TextInput } from '../TextInput';
+import { Tooltip, TooltipPosition } from '../Tooltip';
+import GenerateId from '../../helpers/GenerateId/GenerateId';
+import CopyButton from './CopyButton';
+import ToggleButton from './ToggleButton';
+import ExpandedContent from './ExpandedContent';
+
+export const clipboardCopyFunc = (event, text) => {
```
tlabaj: just wondering if this needs to be exported?  Is there a possibility it will be used outside of this class?

boaz0: Nope, this export is unnecessary. I removed it.
Thanks for noticing. :smile:

jgiardino:
Great job on this, @boaz0! This looks great from an a11y perspective!

> Yes, the message that appears after you click Copy should stay for at least 2 seconds (maybe more?). We did not have a spec on this but I would try adding a 2 second timer and see how that feels.

I agree that the tooltip should remain visible longer. For screen reader users, this pattern requires some updates to the tooltip so that the tooltip can function as an alert. This isn't a variation we support at the moment. I'll capture issues in core and react for this.

> Also, I noticed that in the Expandable version, the expansion area is not editable. It needs to be if a user is to be able to edit the entire contents.

This is handled in core css. When I inspect the element, I don't see the overflow property that handles this in core. Maybe this was recently added to the core css and hasn't been pulled in yet?

christiemolloy:
> @boaz0 I believe that when you click on the "copy to clipboard" button the popover should remain for 2 seconds approximately.

I had left this comment but it hasn't been implemented yet, could you try implementing the success tooltip to remain for 2 seconds?

christiemolloy:
Also, when I type a lot of characters this happens, it might be something that we need on the Core side @mcoker do you know if we should add `white-space: pre`

Screen Shot 2019-03-28 at 4 42 15 PM

boaz0:
@christiemolloy do you mind verifying that it is OK now. Thanks a lot

christiemolloy:
@boaz0 Im still seeing this
Screen Shot 2019-04-01 at 11 26 59 AM

boaz0:
It's going to be replaced by `textarea`. I will update it soon.

Thanks.

boaz0:
@tlabaj yes sure.

christiemolloy:
A few comments:

In React there is no `...` where the text gets cut off in the input group. This is how it should look:
Screen Shot 2019-04-15 at 2 05 25 PM

Looks like the popover isn't directly in the center of the Copy to Clipboard icon
Screen Shot 2019-04-15 at 2 04 45 PM

Also looks like that when the clipboard copy button is selected the popover doesnt remain for 2 seconds.

boaz0:
@christiemolloy 

1 - We need to update the patternfly-next package to have that change.
2 - I am afraid it's a problem with the docs - I see many css attributes disabled. Using the demo-app it works fine
3 - Thanks for noticing, fixed that.

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## Read Only Clipboard Copy
+
+```js
+import React from 'react';
+import { ClipboardCopy } from '@patternfly/react-core';
+class SimpleClipboardCopy extends React.Component {
+  render() {
+    return <ClipboardCopy isReadOnly>This is editable</ClipboardCopy>;
```
mcoker: Should this say "not editable" since it's read only?

mcoker:
What's the purpose of the `<textarea>`? It has inline styles, the font-size is too big, and it scrolls instead of the expandable content container, so the overflow and scrollbar look odd.

Screen Shot 2019-04-18 at 5 46 55 PM

If we're going to keep that in the component, and the `<textarea>` should always wrap the expanded content, can we move that and the styling to core and fix the overflow/scrolling there, too?

If it was added to address the text wrapping issue @christiemolloy mentioned, we should be able to address that behavior in core, too.

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.md
boaz0: Dang! You're right!

boaz0:
@mcoker I added `textarea` because @mcarrano mentioned that the text should also be edited in the expanded block.

Yep, you're right, I should have done in core but I wanted to get feedback as soon as possible. I will take care of it.

Sorry for the mess.

Thanks.

mcoker:
@boaz0 ah, I see! That's definitely a use case for a `<textarea>`. I also wonder if `contenteditable` on the HTMl element would work, which allows us to retain the styling and not need to introduce a new element?

boaz0:
@mcoker 
First time I have heard of `contenteditable` - thanks.
Yes, it seems to work too. :+1:

PR #1540, opened 2019-03-11
chore(TableDoc): Update Documentation for for using 'title' prop in table cells.

### LINKED ISSUE #1473: Update Documentation for for using 'title' prop in table cells.
The documentation is not clear on how to use the title prop in a cell.  Please see issue #1351 for more details.

dtaylor113:
Fixes #1473 - a P1 issue, which is why I chose it, but it's a simple doc change.

Updated table doc to:

image

**Note:** The ability to now use HTML in the Description column.   This is accomplished via:
`<span dangerouslySetInnerHTML={{__html: prop.description.text}}/>`  -wondering if there is any downside/risk in using this method?

### DIFF
--- a/packages/patternfly-4/react-docs/src/components/propsTable/propsTable.js
+++ b/packages/patternfly-4/react-docs/src/components/propsTable/propsTable.js
@@ -54,7 +54,7 @@ export const PropsTable = ({ name, description: preface, props, enumValues }) =>
             </TD>
             <TD align="center">{prop.required && <ExclamationCircleIcon />}</TD>
             <TD>{Boolean(prop.defaultValue) && prop.defaultValue.value}</TD>
-            <TD>{prop.description && prop.description.text}</TD>
+            <TD>{prop.description && <span dangerouslySetInnerHTML={{__html: prop.description.text}}/>}</TD>
           </Row>
         ))}
       </Body>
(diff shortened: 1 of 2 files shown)


dtaylor113:
> Looks good! I don't see any downside to dangerously set HML, it's only on doc site so no biggie.

 I guess any method which has 'dangerously' in its name worries me :-)

packages/patternfly-4/react-docs/src/components/propsTable/propsTable.js
```diff
@@ -54,7 +54,7 @@ export const PropsTable = ({ name, description: preface, props, enumValues }) =>
             </TD>
             <TD align="center">{prop.required && <ExclamationCircleIcon />}</TD>
             <TD>{Boolean(prop.defaultValue) && prop.defaultValue.value}</TD>
-            <TD>{prop.description && prop.description.text}</TD>
+            <TD>{prop.description && <span dangerouslySetInnerHTML={{__html: prop.description.text}}/>}</TD>
```
dlabaj: You're going to have to sanitize this prop if you want to use dangerouslySetInnerHTML.  There's no protection from injecting JavaScript, which can lead to cross site scripting.

dlabaj: Can you use something like this to make sure the prop is okay https://www.npmjs.com/package/dompurify

dtaylor113: Sure @dlabaj I can look into that, however, this is only for our internal api documentation; it is not for public consumable code/components.  Do you think dompurify is necessary in this case?  -thanks

dlabaj: Okay after further discussion since it's only on the doc side it should be okay.  Thanks @dtaylor113 .

Issue #1472, opened 2019-02-28
Table - selection callback requires rowProps


priley86:
Some consumers will need the rowProps passed into the table's `rows` prop passed back in the `onSelect` callback. This would be immediately useful if you need to know a system id or some other object identifier passed in for the row selected (rather than trying to convert simple `rowIndex`). Just opening this to target this issue specifically.

```
  function selectClick(event) {
    const selected = rowIndex === undefined ? event.target.checked : rowData && !rowData.selected;
    onSelect && onSelect(event, selected, rowId); //rowData missing
  }
```

Details: 
https://github.com/patternfly/patternfly-react/pull/1250#pullrequestreview-197863543

rachael-phillips:
@priley86 do you mind telling me more about priorities around this issue? Is this blocking you currently?

priley86:
hey @rachael-phillips - I have started a P.O.C. to showcase table/table selections (but still a W.I.P.). I think this issue is being resolved alongside #1441 currently (that PR seems to resolve it now). I can definitely make use of the table actions there too once it is merged. The minor bugfix in #1503 is also relevant. 

@suomiy is also prototyping inline edit extension table for CNV. #1520, #1521 and #1227 would all support this. #1520 and #1521 should come first though. We could definitely use help reviewing where possible.

thanks for your help!

hope this helps!

priley86:
some further notes on this...after review, #1441 does not seem to handle this and only targets the ActionsColumn `onSelect`. I will need something similar for the Table's `onSelect` and `onSort` callbacks though:

https://github.com/patternfly/patternfly-react/blob/0b637feaf467a860ee69263bb2c477a3d0090c22/packages/patternfly-4/react-table/src/components/Table/examples/SelectableTable.js#L24
```
onSelect(event, isSelected, rowId) {
```
`onSelect` should also return rowProps and extraParams for the row selected:
```
onSelect(event, isSelected, rowId, rowProps, extraParams) {

``` 

On a similar note, I am also seeing that `onSort` does currently not return the column passed in:
https://github.com/patternfly/patternfly-react/blob/0b637feaf467a860ee69263bb2c477a3d0090c22/packages/patternfly-4/react-table/src/components/Table/examples/SortableTable.js#L21
```
onSort(_event, index, direction) {
```
could be:
```
onSort(_event, index, direction, column) {
```
[...]

PR #1542, opened 2019-03-11
feat(Table): pass rowData and extraData to table callbacks

suomiy:
solves https://github.com/patternfly/patternfly-react/issues/1472 issue and passes useful data back to consumer. 

@priley86 @karelhala


packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -12,14 +12,27 @@ export const TableVariant: {
   'compact': 'compact'
 }
 
+export interface IRowData {
+}
```
suomiy: I think it might be useful to name this type, because we might add more typed properties to it in the future + consumers can reference it right now.

PR #1551, opened 2019-03-12
fix: PF4 Fix HelperText definition in LoginForm

### LINKED ISSUE #1550: [PF4][Bug] Login form helperText definition is wrong
In LoginForm.d.ts  helperText is string when in the js is a node.

aljesusg:
HelperText need to be ReactNode

Fix #1550


aljesusg:
please @dlabaj can you review this?

aljesusg:
@serenamarie125 @priley86 there is a typo here

aljesusg:
Thanks @dlabaj @tlabaj , maybe we should think about make some test for this...it's not the first time that there is a problem with the definitions

serenamarie125:
thanks for your contribution here @aljesusg !

Issue #478, opened 2018-07-14
Does this setup work in windows machine ?


Private-SO:
Hi,

I am running into different types of errors when i am installing the setup in my Windows machine.

These are the commands i had used to 

-  yarn add patternfly-react
- yarn install
- yarn build

When i am running yarn build it is throwing different errors like this 

```
C:\Program Files\nodejs\patternfly-react>yarn build
yarn run v1.7.0
$ yarn copy-fonts && yarn copy-img && yarn build-css && yarn build-js
$ ncp node_modules/patternfly/dist/fonts src/fonts
There were errors during the copy.
Error: ENOENT: no such file or directory, mkdir 'C:\Program Files\nodejs\pattern
fly-react\src\fonts'
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this comm
and.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this comm
and.

```

But the setup is working fine in Ubuntu and mac . Why i am getting these errors in my Windows Machine? 

BTW i am running above code in CMD prompt in Admin mode.

Thanks

dlabaj:
Currently OS specific command prevent this from running on Windows.  Converting this to a bug we should fix this by updating package.json to not have OS specific commands.

jschuler:
Close in favor of #1557

PR #1555, opened 2019-03-12
fix(codemod): fix tests failing due to prettier

redallen:
**What**: Remove prettier EOL config option during tests. Tests fail whenever it is specified on *nix-based systems after #1255

**Additional issues**:

### DIFF
--- a/packages/react-codemods/transforms/pf3-pf4.button.test.js
+++ b/packages/react-codemods/transforms/pf3-pf4.button.test.js
@@ -3,25 +3,9 @@ import prettier from 'prettier';
 import { defineInlineTest } from 'jscodeshift/dist/testUtils';
 import transform from './pf3-pf4';
 
-/**
- * Codemod outputs should follow the EOL pattern of the target codebase, not
- * Patternfly's. Patternfly currently enforces LF as line ending, independent of
- * the OS building the codebase.
- *
- * JSCodeShift produces OS-dependent line endings however, LF for Unix based
- * systems, and CRLF for Windows based ones. This is also likely what most
- * projects would like to see the codemod prduce.
- *
- * To make sure we both adhere to Patternfly's conventions, and compare the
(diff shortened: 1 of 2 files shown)


dlabaj:
LGTM.

packages/react-codemods/transforms/pf3-pf4.button.test.js
```diff
@@ -3,25 +3,9 @@ import prettier from 'prettier';
...
- * To make sure we both adhere to Patternfly's conventions, and compare the
- * correct OS-specific line endings during the test cases' assertions, we store
- * expected values with LF line endings, and convert them into the OS-specific
- * ones at runtime using prettier.
- */
-const PRETTIER_EOL = SYSTEM_EOL === '\r\n' ? 'crlf' : 'cr';
 const prettierConfig = {
   ...prettier.resolveConfig.sync(process.cwd()),
-  parser: 'babel',
-  endOfLine: PRETTIER_EOL
```
jschuler: Will this work under Windows? I think this was added to support that
https://github.com/patternfly/patternfly-react/pull/1255/files#r251540310

redallen: I don't think our project even builds on Windows currently due to path resolution problems. I'd love to support it in the future and fix things like #478

redallen: To answer the question, as long as a Windows machine doesn't convert the line-endings to CRLF in JSCodemod, it will still work.

jschuler: Okay, not too long ago there was some effort made to support it but I guess without a Windows environment it is difficult to keep this
https://github.com/patternfly/patternfly-react/pull/1248
https://github.com/patternfly/patternfly-react/pull/1251

jschuler:
Will create an issue to setup a Win environment possibly to test this going forward

Issue #1559, opened 2019-03-12
PF4 Chip:Remove unused prop badgeText from the Chip.d.ts file


tlabaj:
Remove unused prop badgeText from the Chip.d.ts file.  This is an API change so the fix should not be merged until we are ready for our next major release.

boaz0:
@tlabaj I think this issue can be closed.

tlabaj:
Closed by #1728

Issue #1506, opened 2019-03-05
Chip - default attributes are unset


gastaldi:
In order to use the Chip component, I am required to set all the properties below, otherwise I get a compilation error. I am using TypeScript. 

```xml
<Chip key={c.id} closeBtnAriaLabel="none" isOverflowChip={false} isBadge={false} tooltipPosition="top">
```

rachael-phillips:
We can go ahead and close this issue: referencing #1560

PR #1560, opened 2019-03-12
fix(Chip, NavItem): Fixed Typescript definition file errors

tlabaj:
#1559, #1506

**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/components/ChipGroup/Chip.js
```diff
@@ -93,7 +93,8 @@ Chip.defaultProps = {
   closeBtnAriaLabel: 'close',
   className: '',
   isOverflowChip: false,
-  tooltipPosition: 'top'
+  tooltipPosition: 'top',
+  onClick: () => {}
```
jschuler: Don't have to change this, but I've seen it written so many ways, like `null`, `() => undefined`, `Function.prototype` maybe we should have a noop util for consistency at some point

PR #1563, opened 2019-03-13
fix(Modal): remove backdropOpen class from body when Modal is removed

### LINKED ISSUE #1531: pf-c-backdrop__open class stays on <body> even after 'Modal' removal
This prevent user from scrolling after the `<Modal ..>` has been removed from the screen.

image

It seems the class need to be removed: `document.body.classList.remove(css(styles.backdropOpen));`
- when the `isOpen` switch to `false`  
- **and also** when the `componentWillUnmount`

https://github.com/patternfly/patternfly-react/blob/4778a2827ea115b9e1b05683d03a26a26cdb63da/packages/patternfly-4/react-core/src/components/Modal/Modal.js#L92

ia3andy:
1. Add a test that fails because the class is not removed
2. Added the fix (remove the backdropOpen class from the body in `componentWillUnmount`)
3. Test is passing

This is not related to this fix, but having multiple modals in one screen may not be a good idea...

Closes #1531

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.js
@@ -92,6 +92,7 @@ class Modal extends React.Component {
   componentWillUnmount() {
     document.body.removeChild(this.container);
     document.removeEventListener('keydown', this.handleEscKeyClick, false);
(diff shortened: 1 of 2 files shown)


ia3andy:
w00t !!

Thanks for being so reactive :)

PR #1570, opened 2019-03-13
feat(core): clean tsconfigs, add defaultProps to TS-docs

redallen:
**What**: 

- Create 3 different tsconfigs that inherit from a base. Configure .d.ts file generation, typechecking, and jest to use correct tsconfigs.
- Add "@types/react-dom"
- Enable strict typechecking mode and fix miscellaneous errors (mostly in react-core/src/helpers/util.ts)
- Make gatsby use new tsconfig and import props directly from `typedoc` json using new propsHelper.js
- Fix Title to be more like Avatar. Discussion is welcome surrounding default exports.

**Additional issues**: Title is broken in the docs due to ES6 object destructuring renaming when looking for the default for `headingLevel`, I'll fix it Monday.


packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -117,58 +116,58 @@ class Wizard extends React.Component<WizardProps> {
...
     this.toggleSiblingsFromScreenReaders(false);
     document.removeEventListener('keydown', this.handleKeyClicks, false);
   }
 
-  private onNavToggle = isNavOpen => {
-    this.setState({ isNavOpen })
-  }
-
-  private onNext = () => {
+  onNext(): void {
```
jschuler: arrow functions bind `this` automatically , if this is not arrow function i think we need to bind in constructor
`this.onNext = this.onNext.bind(this);`

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -41,7 +41,7 @@ interface ComputedStep extends WizardStep {
 
 export type WizardStepFunctionType = (newStep: { id?: string | number; name: string; }, prevStep: { prevId?: string | number; prevName: string; }) => void;
 
-interface WizardProps {
+export interface WizardProps {
```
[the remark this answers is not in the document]
redallen: In case tests or consumers want to use it.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
redallen: You're right, fixed in latest commit

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -1,6 +1,6 @@
+import * as React from 'react';
 import styles from '@patternfly/patternfly/components/Avatar/avatar.css';
 import { css } from '@patternfly/react-styles';
-import * as React from 'react';
 
 const defaultProps = {
```
jschuler: Delete this?

packages/patternfly-4/react-docs/src/components/componentDocs/componentDocs.js
```diff
@@ -101,20 +101,16 @@ class ComponentDocs extends React.PureComponent {
...
-              return (
-                <PropsTableTs
-                  key={componentName}
-                  name={componentName}
-                  props={componentDocsTs.children}
-                  types={types}
-                  defaultProps={componentFunction.defaultProps}
-                />
-              );
+              console.log('ts', componentName);
```
jschuler: should remove here and below

PR #1574, opened 2019-03-14
fix(VerticalNav): fix attributes for secondary items

glekner:
**What**:
Quick fix to the last PR
pass null instead of empty strings to prevent empty attributes


glekner:
@jeff-phillips-18 @priley86 this can be merged I believe.

Issue #1546, opened 2019-03-12
PF4 Switch adds an unwanted prefix "on|off" to the label


idevat:
When I use the following Switch component:
```javascript
const Test = () => {
  const [isChecked, setIsChecked] = React.useState(false);
  return (
    <Switch
      id="id"
      label="Use custom address"
      isChecked={isChecked}
      onChange={() => setIsChecked(!isChecked)}
      aria-label="aria-label"
    />
  );
};
```
I get the label `OffUse custom address` or `OnUse custom address`.

rachael-phillips:
We are aware of this issue and are working on this on the Core side before moving into React.

rebeccaalpert:
This should be fixed by https://github.com/patternfly/patternfly-react/pull/1591.

PR #1591, opened 2019-03-15
feat(Switch): Add icon and fix labels

### LINKED ISSUE #1296: Switch a11y, css and toggle icon updates
From https://github.com/patternfly/patternfly-next/pull/1300

* changes `aria-label` to `aria-labelledby` on `.pf-c-switch__input`
    * although, @christiemolloy looking at the examples in core, I don't see elements with the `id`'s referenced in `aria-labelledby` in the switch examples. Can you clarify?
* ~CSS now toggles the switch label (on/off)~
* there is now an icon in the switch toggle when the switch has no label

rebeccaalpert:
I added the icon to unlabeled switches, updated HTML and CSS classes to match core/avoid CSS labels, and updated the example labels to match core.

Fixes #1296 and #1546.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.js
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.js
@@ -3,6 +3,7 @@ import styles from '@patternfly/patternfly/components/Switch/switch.css';
 import { css } from '@patternfly/react-styles';
 import PropTypes from 'prop-types';
 import { getUniqueId } from '../../helpers/util';
+import { CheckIcon } from '@patternfly/react-icons';
 
 const propTypes = {
   /** id for the label. */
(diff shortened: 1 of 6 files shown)


jgiardino:
I'm still not seeing a preview link display in this PR. Is there anyway to get one?

redallen:
CircleCI missed this PR (pushing another commit might fix it). In the meantime, here's a preview of 9c06ffcdff5a54762f1573562c22cdd95637e4ac:

http://lovely-pull.surge.sh/

jgiardino:
Thanks, @redallen!

@rebeccaalpert These updates are great! Here are a few comments:

- For the examples that have visible text labels, the `aria-label` prop should be set to the same text string that displays for the `on` state. Specifically, the following examples should all have this prop defined with this value: 
  `aria-label="Message when on"`:
  - Simple Switch
  - Disabled Switch, the first two `<Switch>` component instances
  - Uncontrolled Switch, the first two `<Switch>` component instances
- Are you able to update the description for `aria-label` to include something like: 
  > Adds accessible text to the Switch, and should describe the `isChecked="true"` state. When `label` is defined, `aria-label` should be set to the text string that is visible when `isChecked` is true.

rebeccaalpert:
@jgiardino, I'm going to push the aria-label changes in a PR tied to issue https://github.com/patternfly/patternfly-react/issues/1602 since this merged before I could get those in. I'll have it by the end of today.

PR #1592, opened 2019-03-15
feat(css): change css import methodology

dgutride:
Moving the emotion insertGlobal to a css require statement - many projects have asked that we don't dynamically add style tags to the app.  Without hooks into navigation, the current style injection also doesn't remove styles from the dom as the user moves throughout the app - leading to bloat over time.


packages/patternfly-4/react-styles/src/build/util.js
```diff
@@ -1,20 +1,28 @@
...
 
-export function cssToJS(cssString, useModules = false) {
+export function cssToJS(cssString, cssOutputPath = '', useModules = false) {
+  let cssRequire = '';
+  if (cssOutputPath) {
+    cssRequire = `require('${cssOutputPath}');`;
+  }
   if (useModules) {
     return `import { ${styleSheetToken} } from '${packageName}';
+${cssRequire}
```
redallen: Is this the equivalent of eval() ?

dgutride: nope

PR #1613, opened 2019-03-20
fix(DataList): remove shadow-box

### LINKED ISSUE #1471: Remove box shadow utility classes from DataTable/examples
We removed the box shadow utility class from DataTable in https://github.com/patternfly/patternfly-next/pull/1143

If a DataTable needs to be shown with a box shadow for demo purposes, it can be placed inside of a card (without a header/body/footer) instead.

cc @kybaker

boaz0:
**What**:
closes #1471

Remove the shadow-box class from `DataList`.

//cc @mcoker


boaz0:
Thank you! :smile:

Issue #1421, opened 2019-02-21
Alert: create variation to use for dynamic notifications


jgiardino:
The Alert component needs to be extended to support a variation of the Alert that can be used for toast notifications. 

The html that's currently defined in core does not reflect the html needed to support dynamic alerts. Instead, we'd like to base the core example off of what's implemented for this issue.

A [POC](https://github.com/andresgalante/jenn/pull/25) was created to test what works across the different screen readers we support, and findings are noted in this comment: https://github.com/patternfly/patternfly-next/issues/1253#issuecomment-465355653

Given this testing, we found that:
- A live region must exist on a page 200 ms before the alert message is inserted into the live region
- A separate live region is required for each alert

Some suggestions for how to handle the html structure are noted in this comment: https://github.com/patternfly/patternfly-next/issues/1375#issuecomment-465364138

Ultimately, we'd like the react dev team to decide the best way to address these requirements. Then we will update core to match. 

Please ensure that [all links have discernible text](https://dequeuniversity.com/rules/axe/3.4/link-name?application=AxeChrome) and [all id attribute values are unique](https://dequeuniversity.com/rules/axe/3.4/duplicate-id?application=AxeChrome). These are currently known axe issues with alert group in core that should be considered in this implementation. To check if these conditions are met, please run axe on your solution.

Issue #1529, opened 2019-03-08
Fix accessibility prop name inconsistencies


redallen:
Property names for [aria attributes](https://a11yproject.com/posts/ARIA-properties) are inconsistent. Some examples:
- `AboutModalContainer` takes the following properties: `ariaLabelledbyId`, `ariaDescribedById`. Note inconsistencies in `by` vs `By`.
- `AboutModalBox` takes the raw HTML properties: `aria-labelledby`, `aria-describedby`
- `Alert` takes `variant` and `variantLabel` and sets `aria-label=undefined` as a default property without it being an actual component property that we can typecheck.
- I'm sure there's more...

Codebase names:
- `ariaLabel`: 148
- `aria-label`: 424
- `aria-controls`: 10
- `ariaControls`: 3
- `aria-labelledby`: 74
- `ariaLabelledbyId`: 10
- `aria-describedby`: 18
- `ariaDescribedById`: 12
- `variant`: 694
- `variantLabel`: 13
- There's more... See the [aria property list](https://a11yproject.com/posts/ARIA-properties) and use a global regex tool to spot our naming inconsistencies.

We are already mostly using the HTML properties, so I propose we introduce a breaking change to rename properties like `ariaLabelledbyId` to their HTML counterpart `aria-labelledby`. If we need to do logic with them in the component, we can rename them to camel-case in the property destructuring that already takes place using the `:` syntax:
```javascript
{
  'variant-label': variantLabel,
  'aria-label': ariaLabel,
  'aria-labelledby': ariaLabeledBy,
  'aria-describedby': ariaDescribedBy
}
```

redallen:
Ref #1525

tlabaj:
closed by #3924

tlabaj:
closed by #3924

PR #1622, opened 2019-03-21
feat(alertgroup): add AlertGroup, Alert changes

seanforyou23:
**What**: Introduces an [AlertGroup](https://pf4.patternfly.org/components/AlertGroup/examples/) component.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/1421 and https://github.com/patternfly/patternfly-next/issues/1375

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Alert/Alert.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Alert/Alert.d.ts
@@ -14,6 +14,7 @@ export interface AlertProps extends Omit<HTMLProps<HTMLDivElement>, 'action'> {
   children?: ReactNode;
   action?: ReactNode;
   'aria-label'?: string;
+  isToast?: boolean;
   variantLabel?: string;
 }
 
--- a/packages/patternfly-4/react-core/src/components/Alert/Alert.js
+++ b/packages/patternfly-4/react-core/src/components/Alert/Alert.js
@@ -24,16 +24,19 @@ const propTypes = {
   children: PropTypes.node,
   /** additional classes added to the Alert */
   className: PropTypes.string,
-  /** Adds accessible text to the Alert */
+  /** Adds an accessible name to the Alert */
   'aria-label': PropTypes.string,
+  /** Sets recommended aria attributes and CSS classes */
+  isToast: PropTypes.bool,
   /** Variant label text for screen readers */
   variantLabel: PropTypes.string,
-  /** Additional props are spread to the container <div>  */
+  /** Additional props are spread to the container - div  */
   '': PropTypes.any
 };
 
 const defaultProps = {
   'aria-label': undefined,
+  isToast: false,
   action: null,
   children: '',
   className: '',
@@ -46,6 +49,7 @@ const Alert = ({
   variant,
   variantLabel,
   'aria-label': ariaLabel = getDefaultAriaLabel(variant),
+  isToast,
   action,
   title,
   children,
@@ -60,19 +64,30 @@ const Alert = ({
     </React.Fragment>
   );
 
-  const customClassName = css(styles.alert, getModifier(styles, variant, styles.modifiers.info), className);
-
+  const customClassName = css(
+    styles.alert,
+    getModifier(styles, variant, styles.modifiers.info),
+    className,
+    (isToast && 'pf-m-live')
+  );
   return (
     <div {...props} className={customClassName} aria-label={ariaLabel}>
       <AlertIcon variant={variant} />
-      <h4 className={css(styles.alertTitle)}>{readerTitle}</h4>
-      {children && (
-        <div className={css(styles.alertDescription)}>
-          <p>{children}</p>
-        </div>
-      )}
+      <div
+        aria-live={isToast ? 'polite' : undefined}
+        aria-atomic={isToast ? 'false' : undefined}
+        className="pf-c-alert__content">
+        <h4 className={css(styles.alertTitle)}>{readerTitle}</h4>
+        {children && (
+          <div className={css(styles.alertDescription)}>
+            <p>{children}</p>
+          </div>
+        )}
+      </div>
       {action && (
-        <div className={css(styles.alertAction, className)}>{React.cloneElement(action, { title, variantLabel })}</div>
+        <div className={css(styles.alertAction, className)}>
+          {React.cloneElement(action, { title, variantLabel })}
+        </div>
       )}
     </div>
   );
--- a/packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.md
+++ b/packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.md
@@ -1,0 +1,206 @@
+---
+title: "AlertGroup"
+cssPrefix: "pf-c-alert-group"
+---
+
+import * as React from 'react';
+import { Alert, AlertGroup, AlertVariant, AlertActionCloseButton, InputGroup } from '@patternfly/react-core';
+
+## Static Alert Group
+```js
+import * as React from 'react';
+import { Alert, AlertGroup, AlertVariant } from '@patternfly/react-core';
+
+class StaticAlertGroup extends React.Component {
+  render() {
+    return (
+      <React.Fragment>
+        <AlertGroup>
+          <Alert title="Success Alert" variant={AlertVariant['success']} />
+          <Alert title="Info Alert" variant={AlertVariant['info']} />
+        </AlertGroup>
+      </React.Fragment>
+    );
+  }
+}
+```
+
+## Toast Alert Group
+```js
+import * as React from 'react';
+import { Alert, AlertGroup, AlertActionCloseButton, AlertVariant, InputGroup } from '@patternfly/react-core';
+class ToastAlertGroup extends React.Component {
+  constructor() {
+    this.state = { alerts: [] };
+    this.addAlert = (title, variant, key) => {
+      this.setState({
+        alerts: [ ...this.state.alerts, { title: title, variant: variant, key }]
+      });
+    };
+    this.removeAlert = key => {
+      this.setState({ alerts: [...this.state.alerts.filter(el => el.key !== key)] });
+    };
+  }
+  render() {
+    const btnClasses = ['pf-c-button', 'pf-m-secondary'].join(' ');
+    const getUniqueId = () => (new Date().getTime());
+    const addSuccessAlert = () => { this.addAlert('Toast Success Alert', 'success', getUniqueId()) };
+    const addDangerAlert = () => { this.addAlert('Toast Danger Alert', 'danger', getUniqueId()) };
+    const addInfoAlert = () => { this.addAlert('Toast Info Alert', 'info', getUniqueId()) };
+    return (
+      <React.Fragment>
+        <InputGroup style={{ marginBottom: '16px' }}>
+          <button onClick={addSuccessAlert} type="button" className={btnClasses}>Add Toast Success Alert</button>
+          <button onClick={addDangerAlert} type="button" className={btnClasses}>Add Toast Danger Alert</button>
+          <button onClick={addInfoAlert} type="button" className={btnClasses}>Add Toast Info Alert</button>
+        </InputGroup>
+        <AlertGroup isToast>
+          {this.state.alerts.map(({key, variant, title}) => (
+            <Alert
+              isToast
+              variant={AlertVariant[variant]}
+              title={title}
+              action={
+                <AlertActionCloseButton
+                  onClose={() => this.removeAlert(key)}
+                />
+              }
+              key={key} />
+          ))}
+        </AlertGroup>
+      </React.Fragment>
+    );
+  }
+}
+```
+
+## Singular Additive Alert Group
+```js
+import * as React from 'react';
+import { Alert, AlertGroup, AlertVariant, InputGroup } from '@patternfly/react-core';
+class SingularAdditiveAlertGroup extends React.Component {
+  constructor() {
+    this.state = {
+      alerts: []
+    }
+  }
+  render() {
+    const addAlert = (title, variant, key) => {
+      this.setState({ alerts: [...this.state.alerts, { title: title, variant: variant, key }] });
+    };
+    const btnClasses = ['pf-c-button', 'pf-m-secondary'].join(' ');
+    const getUniqueId = () => (new Date().getTime());
+    const addSuccessAlert = () => { addAlert('Single Success Alert', 'success', getUniqueId()) };
+    const addDangerAlert = () => { addAlert('Single Danger Alert', 'danger', getUniqueId()) };
+    const addInfoAlert = () => { addAlert('Single Info Alert', 'info', getUniqueId()) };
+    return (
+      <React.Fragment>
+        <InputGroup style={{ marginBottom: '16px' }}>
+          <button onClick={addSuccessAlert} type="button" className={btnClasses}>Add Single Success Alert</button>
+          <button onClick={addDangerAlert} type="button" className={btnClasses}>Add Single Danger Alert</button>
+          <button onClick={addInfoAlert} type="button" className={btnClasses}>Add Single Info Alert</button>
+        </InputGroup>
+        <AlertGroup>
+          {this.state.alerts.map(({ title, variant, key }) => (
+            <Alert
+              variant={AlertVariant[variant]}
+              title={title}
+              key={key} />
+          ))}
+        </AlertGroup>
+      </React.Fragment>
+    );
+  }
+}
+```
+
+## Multiple Additive Alert Group
+```js
+import * as React from 'react';
+import { Alert, AlertGroup, AlertVariant, InputGroup } from '@patternfly/react-core';
+class MultipleAdditiveAlertGroup extends React.Component {
+  constructor() {
+    this.state = {
+      alerts: []
+    }
+  }
+  render() {
+    const addAlerts = (incomingAlerts) => {
+      this.setState({ alerts: [...this.state.alerts, ...incomingAlerts] });
+    };
+    const getUniqueId = () => (new Date().getTime());
+    const btnClasses = ['pf-c-button', 'pf-m-secondary'].join(' ');
+    const addAlertCollection = () => {
+      addAlerts([
+        { title: 'First Alert Notification.', variant: 'success', key: getUniqueId() },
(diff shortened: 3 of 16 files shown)


jgiardino:
@seanforyou23 Thanks for pulling this together! I did a quick test in VO, but not in JAWS/NVDA.

Given our previous testing, without testing this in JAWS or NVDA, my guess is that this works in those screen readers. But testing the Async example in VO, the behavior I’m seeing with VO is consistent with our POC, where this method doesn’t work for VO. The issue that happens in VO is that each new alert will interrupt the alert that is currently being announced.

My suggestion for finalizing AlertGroup and Alert is to move the live region off AlertGroup and have each Alert be its own live region. This structure will work as-is for VO, but additionally, we can introduce CSS animation in core to handle the timing needed to make sure alerts are announced in JAWS and NVDA.

Specifically, the updates needed in this PR are:
- Remove these attributes from `.pf-c-alert-group`:
  `aria-atomic="false" aria-live="polite" aria-relevant="additions text"`
- Add this class to the `<li>` elements included in the AlertGroup component:
  `pf-c-alert-group__item`

Other updates needed in react (this can be a separate issue):
- Include a prop to identify an Alert component as an aria-live component (prop name suggestions are welcome). When this prop is true/exists, then add the following `<div>`s that wrap the text message of an Alert:
  ```
  <div aria-atomic="false" aria-live="polite" aria-relevant="additions text">
    <div class="one pf-c-alert__text">
      // the divs above only exist in the aria-live variation of the Alert, and would wrap the following elements
      <h3 class="pf-c-alert__title"></h3>
      <div class="pf-c-alert__description"></div>
  ```

seanforyou23:
Sounds good @jgiardino, I'll make those updates to this PR and can take a stab at making the adjustments to the React Alert component also. Two things;

1) I think we can and should use the `aria-live` input property name directly (any aria-attributes for that matter) as they are natively supported by React. See this [related issue](https://github.com/patternfly/patternfly-react/issues/1529).

2) I wonder if the alerts should be live regions by default? I'll experiment with this and make a recommendation, of course, let me know if you have any idea or preferences around this also.

packages/patternfly-4/react-core/src/components/Alert/Alert.js
```diff
@@ -60,20 +83,32 @@ const Alert = ({
     </React.Fragment>
   );
 
-  const customClassName = css(styles.alert, getModifier(styles, variant, styles.modifiers.info), className);
-
+  const customClassName = css(
+    styles.alert,
+    getModifier(styles, variant, styles.modifiers.info),
+    className,
+    ((toast || ariaLive) && 'pf-m-live')
```
jgiardino: I don't think we want `ariaLive` here. The purpose of `pf-m-live` is that it will enable us to add css to core that will add the delay that some screen readers need in perceiving the contents.

And the purpose of `toast` is to use the Alert component with all the things we recommend to make the alert accessible, following our design patterns. If the consumer doesn't want to use `toast` and instead wants to manually define aria attributes for the live region, then they should also manually add the class `pf-m-live` that handles that delay.

packages/patternfly-4/react-core/src/components/Alert/Alert.js
```diff
@@ -60,20 +83,32 @@ const Alert = ({
...
-        </div>
-      )}
-      {action && (
-        <div className={css(styles.alertAction, className)}>{React.cloneElement(action, { title, variantLabel })}</div>
-      )}
+      <div
+        className="pf-c-alert__content"
+        aria-live={toast ? 'polite' : ariaLive}
+        aria-relevant={toast ? 'additions text' : ariaRelevant}
+        aria-atomic={toast ? 'false' : ariaAtomic}>
```
jgiardino: I just want to throw a suggestion out that would also work, but maybe simplify some of this component...

We could remove these as explicitly defined props and all the logic associated with them, and instead do that thing where the consumer specifies html attributes on the component, and they are passed down to an html element. My personal expectation—which may not be common—is that those cases would add the attributes to the parent html element of the component. If that is the expectation, that would leave us with this:
```
<div class="pf-c-alert pf-m-success pf-m-live" aria-label="Success Alert" aria-live="polite" aria-relevant="additions text" aria-atomic="false">
  <div class="pf-c-alert__icon">...</div>
  <div class="pf-c-alert__content">
    <h4 class="pf-c-alert__title">...</h4>
  </div>
</div>
```

Instead of this:
```
<div class="pf-c-alert pf-m-success pf-m-live" aria-label="Success Alert">
  <div class="pf-c-alert__icon">...</div>
  <div class="pf-c-alert__content" aria-live="polite" aria-relevant="additions text" aria-atomic="false">
    <h4 class="pf-c-alert__title">...</h4>
  </div>
</div>
```

I think this is totally fine. The contents are still announced. It just isn't as clean because ALL the contents are announced, including button labels.

jgiardino:
Thanks for making updates! 

> 1. I think we can and should use the aria-live input property name directly (any aria-attributes for that matter) as they are natively supported by React. See this related issue.

I made a suggestion inline related to this. It might be nice to not have so much built into this component for these aria attributes, and just pass these html attributes if defined on `<Alert>` to the parent `.pf-c-alert` element.

> 2. I wonder if the alerts should be live regions by default? I'll experiment with this and make a recommendation, of course, let me know if you have any idea or preferences around this also.

I definitely think this should not be enabled as a toast by default.

seanforyou23:
Thx for the feedback @jgiardino, I've pushed those updates, here are some thoughts I had while making them.

Not setting the relevant live region attributes automatically when a user sets the `toast` property kinda feels like a missed opportunity. Now, as a consumer, there is no "easy" win if I'm not familiar with the various a11y api parts. It's really nice to be able to use something as simple as `<Alert variant="warning" title="Example" toast={true} />` and just know that I'm getting patternfly recommended live region settings that will work well out of the box.

My fear is that the combination of these attributes being optional and not having a simple flag like this to do the real plumbing will result in people forgetting to set all the necessary props/values or possibly using the wrong values or spending time doing this same research we're going through now to know what values they should use. I think we can save users time and improve the default experience by having the toast prop set these values. It's already a simple component and the extra code required to manage this is nominal so not too worried about that, more just thinking of how we can make this as easy as possible for people who just want a solid factory experience.

I like where it's headed. WDYT?

jgiardino:
Hey, sorry for not being more explicit about this. I agree with you about `toast`. The `toast` prop as you have it looks good, and matches how I expect it to behave. 

My comment was more about how we handle explicitly defined aria attributes. And the updates you made for handling aria attributes manually looks good, too!

seanforyou23:
Thx @jgiardino, just fixed a bug with the nesting of actions. Wanted to get clarification on one more thing before I add tests that will enforce all of what we're discussing. When you mentioned;

>I agree with you about toast. The toast prop as you have it looks good, and matches how I expect it to behave.

I was a little confused because `Alert` currently does not apply aria attributes based off of presence of a toast property, which is what I was recommending. Maybe you were referring to how it was before I pushed that last change? I think we can achieve getting this type of functionality, while still not specifying the aria attributes as explicit props. Do you agree this would be the best of both worlds? Or are you saying the presence of the `toast` prop should only have the single effect of adding the class `pf-m-live`? If the latter is the case, we could probably just omit the toast prop altogether and users can just specify the modifier class as part of the component className prop which would have the same effect.

Let me know when you have time and I'll get this PR all buttoned up. Thanks again.

jgiardino:
> I was a little confused because Alert currently does not apply aria attributes based off of presence of a toast property, which is what I was recommending. Maybe you were referring to how it was before I pushed that last change?

That's probably what happened, because I'm pretty sure I saw an example of the Alert that had `toast` with the three aria attributes applied to `.pf-c-alert__content`, but I don't see that now. I think you and I are expecting the same thing, i.e. this:

> It's really nice to be able to use something as simple as <Alert variant="warning" title="Example" toast={true} /> and just know that I'm getting patternfly recommended live region settings that will work well out of the box.

seanforyou23:
Just realized I forgot about support for `toast`! Working on adding that now, will push updates soon as they are ready.

packages/patternfly-4/react-core/src/components/Alert/Alert.d.ts
```diff
@@ -14,6 +14,7 @@ export interface AlertProps extends Omit<HTMLProps<HTMLDivElement>, 'action'> {
   children?: ReactNode;
   action?: ReactNode;
   'aria-label'?: string;
+  toast?: boolean;
```
jschuler: typically we prefix boolean props, in this case maybe this could be `isToast`

packages/patternfly-4/react-core/src/components/Alert/Alert.js
```diff
@@ -60,19 +64,31 @@ const Alert = ({
     </React.Fragment>
   );
 
-  const customClassName = css(styles.alert, getModifier(styles, variant, styles.modifiers.info), className);
-
+  const customClassName = css(
+    styles.alert,
+    getModifier(styles, variant, styles.modifiers.info),
+    className,
+    (toast && 'pf-m-live')
```
jschuler: is this modifier available through styles.modifiers.live?

packages/patternfly-4/react-core/src/components/Alert/Alert.js
```diff
@@ -60,19 +64,31 @@ const Alert = ({
...
-      {children && (
-        <div className={css(styles.alertDescription)}>
-          <p>{children}</p>
-        </div>
-      )}
+      <div
+        aria-live={toast ? 'polite' : undefined}
+        aria-relevant={toast ? 'additions text' : undefined}
+        aria-atomic={toast ? 'false' : undefined}
+        className="pf-c-alert__content">
```
jschuler: is there `styles.alertContent`?

packages/patternfly-4/react-core/src/components/Alert/Alert.md
```diff
@@ -174,3 +174,133 @@ class DangerAlert extends React.Component {
...
+## Static Live Region Alert
+```js
+import * as React from 'react';
+import { Alert } from '@patternfly/react-core';
+
+class StaticLiveRegionAlert extends React.Component {
+  render() {
+    return (
+      <React.Fragment>
+        <Alert toast={true} variant="info" title="Toast Alert Title"></Alert>
```
jschuler: do not have to specify boolean value `true` for toast. For variants, can we use AlertVariant?

packages/patternfly-4/react-core/src/components/Alert/Alert.md
```diff
@@ -174,3 +174,133 @@ class DangerAlert extends React.Component {
...
+        <Alert
+          aria-live="assertive"
```
jschuler: These aria attributes are spread to the Alert container div, but also the inner div has similar aria attributes?

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.md
```diff
@@ -0,0 +1,194 @@
...
+```js
+import * as React from 'react';
+import { Alert, AlertGroup } from '@patternfly/react-core';
+
+class StaticAlertGroup extends React.Component {
+  render() {
+    return (
+      <React.Fragment>
+        <AlertGroup>
+          <Alert title="Success Alert" variant="success" />
```
jschuler: Can we use AlertVariant for the variants?

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.md
```diff
@@ -0,0 +1,194 @@
...
+    const removeNotification = (index) => {
+      this.setState({ alerts: [...this.state.alerts.filter((el, idx) => !(idx === index))] });
+    };
+    const btnClasses = ['pf-c-button', 'pf-m-secondary'].join(' ');
+    const addSuccessAlert = () => { addAlert('Toast Success Alert', 'success') };
+    const addDangerAlert = () => { addAlert('Toast Danger Alert', 'danger') };
+    const addInfoAlert = () => { addAlert('Toast Info Alert', 'info') };
+    return (
+      <React.Fragment>
+        <div className="pf-c-input-group" style={{ marginBottom: '16px' }}>
```
jschuler: Is there a react component we can use instead?

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { css } from '@patternfly/react-styles';
+import { Omit } from '../../helpers/typeUtils';
+
+export interface AlertGroupProps extends Omit<React.HTMLProps<HTMLUListElement>, 'className'> {
+  /** Additional classes added to the AlertGroup */
+  className?: string;
+  /** Alerts to be rendered in the AlertGroup */
+  children?: React.ReactNode;
+  /** Toast notifications are positioned at the top right cornder of the viewport */
+  toast?: boolean;
```
jschuler: similar point here about possibly prefixing the boolean props

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+  toast,
+  ...props
+}: AlertGroupProps) => (
+    <ul
+      className={css(styles.alertGroup, className, (toast ? 'pf-m-toast' : ''))}
+      {...props}>
+      {React.Children.toArray(children).map(
+        (Alert: React.ReactNode, index: number) => {
+          return (
+            <li className="pf-c-alert-group__item" key={index}>{Alert}</li>
```
jschuler: `styles.alertGroupItem` available?

packages/patternfly-4/react-core/src/components/Button/Button.md
```diff
@@ -39,15 +39,15 @@ import { Button } from '@patternfly/react-core';
 
 LinkButton = () => (
   <React.Fragment>
-    <Button component="a" href="https://pf-next.com/" target="_blank" variant="primary">
+    <Button component="a" href="//pf4.patternfly.org/" target="_blank" variant="primary">
```
jschuler: why the double slash at the beginning?

packages/patternfly-4/react-core/src/components/Alert/Alert.js
seanforyou23: Not at the moment

Screen Shot 2019-04-15 at 10 01 04 AM

packages/patternfly-4/react-core/src/components/Alert/Alert.js
seanforyou23: I don't believe so, similary to `pf-m-live`, I think this was added recently.

Screen Shot 2019-04-15 at 10 00 55 AM

packages/patternfly-4/react-core/src/components/Button/Button.md
seanforyou23: Protocol relative links used to be preferred for portability (syncs http/https paths to that of the protocol on the consuming page) - a quick google tells me this is no longer preferred, I'll update.

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
seanforyou23: Doesn't look like this one is available yet either :/

packages/patternfly-4/react-core/src/components/Alert/Alert.md
seanforyou23: So it's actually an either or kinda thing. I think we want to recommend users just use the `isToast` prop which sets CSS classes as well as aria attributes to what we recommend in general. In this case the aria attributes are planted on the innermost element that contains the notification text.

There may be cases where a user needs more fine grain control over this experience, so in that case a user may omit the `isToast` prop, and set those attributes and classes manually. In this scenario users would be setting those aria attributes directory on the Notification container, which also allows for exposing other bits of content like the icon and actions, should a user want to do that.

It wouldn't be recommended to set both `isToast` and `aria-*` attributes at the same time. Does it make sense to put into place code to prevent this? Seems like it's not worth the added complexity, but I'd like to hear what others think.

In any case I updated the examples to make this more clear as I see how it could be confusing before. Let me know if you think this is enough for now, considering the full docs are still in the works.

mcoker:
@seanforyou23 after adding this to core, we discovered that adding the content section to the alert using it's existing grid layout keeps the title and alert body text from wrapping underneath the action area on the right. I've noted that in the core PR, which is still in progress https://github.com/patternfly/patternfly-next/pull/1726#issuecomment-483357710

christiemolloy:
Also, in this case, when I update the text inside of the first alert in the alert group, and then I click the X, it actually removes the last notification, rather than the Alert that I want to delete.
Screen Shot 2019-05-07 at 11 05 12 AM

mcoker:
@seanforyou23 @jgiardino I looked at this some more in core, and looks like the existing structure won't work. I made notes in the core PR (https://github.com/patternfly/patternfly-next/pull/1726#issuecomment-490226231).

I proposed an option of creating a new layout scoped to the `.pf-m-live` variation of the alert, that should allow us to introduce a new HTML structure without introducing a breaking change.

Another option could be to add the Alert Group as it is, then in our next breaking change release, we introduce the live region changes. What do you think?

mcoker:
@christiemolloy created a core issue to address the warning icon width https://github.com/patternfly/patternfly-next/issues/1804. Also added it to the CSS meeting agenda to discuss a better long-term solution for using consistent icons between core and react.

seanforyou23:
@mcoker I'm fine with getting those changes in core first or going ahead and merging this work - whatever everyone thinks is best. That said, this work already saw the chopping block for last major release so I'm hoping we can get it in and not have to wait for another full cycle to pass. 

I know it's a high visibility component so if there's anything I can do to help get this merged please let me know. I'll keep looking into the issue with not being able to close the toast notifications in the demo in the mean time.

seanforyou23:
Last bit of PR feedback addressed, it has my 👍in case anyone wants to merge it. Thx for the help everyone!

mcoker:
I would like to get @jgiardino's opinion on how to handle the live region. If we should remove that part for now and update it during our next breaking change period, or if we should introduce a new layout scoped to the `.pf-m-live` class for now, then bring that layout back to the default alert in our next breaking change period. If we merge this as it is, the `pf-c-alert__content` element that wraps the live region will keep the alert description from wrapping underneath the action area of the alert like it should.

Issue #1371, opened 2019-02-12
PF4 Table: Cannot read property 'title' of null


dlabrecq:
It appears that the table generates a `Cannot read property 'title' of null` error if any cells are null or undefined.

Appears to be generated from this line of Body.js?
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/Body.js#L51

@karelhala This may interest you? My work around was to ensure `item.label` is not null. Not sure how this is related to the property 'title', but was quite difficult to debug.

```
rows.push({
  cells: [
    item.label ? item.label : '', <<<<
    this.getMonthOverMonthCost(item, index),
    this.getTotalCost(item, index),
  ],
...
```

<img width="1144" alt="screen shot 2019-02-11 at 10 07 06 pm" src="https://user-images.githubusercontent.com/17481322/52609127-34cfc500-2e4a-11e9-9a42-3f056c46bd59.png">

karelhala:
Thanks for bringing this is up yes, it is the same as trying to render null or undefined in React. However we should add warning message or something to notify consumer that they shouldn't send null/undefined to cell.

dlabrecq:
Sounds good. Perhaps we could add a check to the `title: curr.title || curr,` line in Body.js?

rachael-phillips:
Can we go ahead and close this one out?

tlabaj:
Closing this issue.  I have opened issue #1473 to capture updating docs to make this use of the title prop more clear.

dlabrecq:
I believe @karelhala suggested adding a warning message?

PR #1623, opened 2019-03-21
fix(Table): reverts unnecessary fix

suomiy:
https://github.com/patternfly/patternfly-react/pull/1520 already fixed https://github.com/patternfly/patternfly-react/issues/1371 with some other enhacements.

https://github.com/patternfly/patternfly-react/pull/1621 reverted these enhacements, so this PR reverts that PR

It also added 
```
`console.warn(`Cell value at [ ${rowKey}, ${key} ] should not be set to ${curr}`);` 
```

which I do not think is necessary, because there are multiple use cases when the row is represented as structured data and we render that data according to that and do not render cells. In this use case, I think it is not necessary to show warnings (especially one per each cell).

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Body.js
+++ b/packages/patternfly-4/react-table/src/components/Table/Body.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Body } from 'reactabular-table';
 import PropTypes from 'prop-types';
 import { TableContext } from './Table';
+import { isRowExpanded } from './utils';
 
 const propTypes = {
   /** Additional classes for table body. */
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-table/src/components/Table/Body.js
```diff
@@ -32,50 +33,33 @@ class ContextBody extends React.Component {
...
+            };
+
+            // increment the shift index when a cell spans multiple columns
+            if (isCellObject && cell.props && cell.props.colSpan) {
+              additionalColsIndexShift += cell.props.colSpan - 1;
             }
-            console.warn(`Cell value at [ ${rowKey}, ${key} ] should not be set to ${curr}`);
             return {
               ...acc,
-              ...{
```
dlabaj: This is needed since there are times when curr is set to null or undefined causing the application to crash.  The warning is something the PF4 team decided was appropriate for this situation.

suomiy:
> Please add back the handling for where null and undefined are the values for curr.

This should be already handled (I just tested it for null/undefined)

Issue #1451, opened 2019-02-26
PF4 Tabs: scroll button


dlabrecq:
Why is there a scroll button associated with the PF4 tabs? Odly, this does not appear on Chrome, only with Firefox. How do we get rid of it?

![screen shot 2019-02-26 at 11 25 42 am](https://user-images.githubusercontent.com/17481322/53428794-6eabda00-39b9-11e9-85c9-b8b3792321a7.png)

rachael-phillips:
Hi @dlabrecq we don't see this issue on our side. Do you mind telling us how to reproduce it?

dlabrecq:
All I can say is that it shows up on Firefox, not Chrome. And it's still present with the latest prerelease packages.

dlabaj:
@dlabrecq Do you have a link to the file that has this issue?

dlabrecq:
Tested with Firefox v65.0.1 (64-bit) and 66.0 (64-bit).

Note that this is now happening in multiple places:
![Screen Shot 2019-03-19 at 12 59 55 PM](https://user-images.githubusercontent.com/17481322/54626619-e6fd3c80-4a47-11e9-8d60-179328216907.png)

We're using tabs in a few places; for example:

Overview tabs:
https://github.com/project-koku/koku-ui/blob/master/src/pages/overview/overview.tsx#L120

Card tabs:
https://github.com/project-koku/koku-ui/blob/master/src/pages/ocpOnAwsDashboard/ocpOnAwsDashboardWidget.tsx#L295

TheRealJon:
The `Tabs` component uses the `isElementInView` util function to establish whether the first and last `Tab` element in the `TabList` are fully visible. Firefox (and Safari) calculated these `Tab` elements' `clientWidth` to be slightly larger than Chrome did, causing this function to return `false` for the last tab (i.e. the last tab was incorrectly considered partially out of view on the right side). The `showRightScrollButton` state of the `Tabs` component is set to the inverse of that function's return value, thus causing the right scroll button to appear in those browsers.

PR #1625, opened 2019-03-21
fix(util.ts): Improve cross-browser compatibility of `isElementInView` and `sideElementIsOutOfView`

TheRealJon:
Fixes https://github.com/patternfly/patternfly-react/issues/1451

**What**: 
- Use `getBoundingClientRect` and `Math.floor` in `isElementInView` and `sideElementIsOutOfView` util functions to improve consistency of element left and right bound values across browsers. 
- Update `util.test.js` to mock `getBoundingClientRect` function.

### DIFF
--- a/packages/patternfly-4/react-core/src/helpers/util.ts
+++ b/packages/patternfly-4/react-core/src/helpers/util.ts
@@ -22,7 +22,8 @@ export function debounce(this: any, func: (...args: any[]) => any, wait: number)
   };
 }
 
-/** This function returns the side the element is out of view on (right, left or both)
+/** This function returns whether or not an element is within the viewable area of a container. If partial is true,
+ * then this function will return true even if only part of the element is in view.
  * @param {HTMLElement} container  The container to check if the element is in view of.
  * @param {HTMLElement} element    The element to check if it is view
  * @param {boolean} partial   true if partial view is allowed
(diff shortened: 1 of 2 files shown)


TheRealJon:
Since I'm using `Math.floor` here, there is probably an edge case where an `element` overflows on the left or right of `container` by a fraction of a pixel and not be considered partially hidden.

packages/patternfly-4/react-core/src/helpers/util.ts
```diff
@@ -22,6 +22,7 @@ export function debounce(this: any, func: (...args: any[]) => any, wait: number)
   };
 }
 
+// TODO this is not the correct description for this function
```
jschuler: LGTM, can you update the descriptions?

TheRealJon:
@jschuler I've updated the comments for the functions in util.ts

packages/patternfly-4/react-core/src/helpers/util.ts
```diff
@@ -73,7 +76,17 @@ export function sideElementIsOutOfView(container: HTMLElement, element: HTMLElem
   // Return outcome
   return side;
 }
-/** This function returns the side the element is out of view on (right, left or both)
+
+/** Interpolates a parameterized templateString using values from a templateVars object.
```
jschuler: Thanks for fixing this description as well

TheRealJon:
Sorry, noticed we were also missing a unit test for that templateFill function, so I added one.

TheRealJon:
@jschuler @tlabaj Is there a way to re-trigger TravisCI? I think it canceled the last run but didn't start a new one.

tlabaj:
@TheRealJon I just restarted it.

PR #1627, opened 2019-03-22
fix(DataList): set width and isExpanded to optional in d.ts

### LINKED ISSUE #1607: PF4 DataList: width is required?
image
if width is required, we should indicate this in the example

### LINKED ISSUE #1608: PF4 DataList: expanded should not be a required prop for DataListItem
If i am using a regular data list without expandability, I shouldnt have to pass in isExpanded={false} to DataListItem

boaz0:
**What**:

fix #1607 
fix #1608 

//cc @ibolton336


boaz0:
Thank you all :man_dancing:

PR #1638, opened 2019-03-25
fix(dropdown-toggle): toggle props, allow null iconComponent

priley86:
**What**:
Slight tweak to  `DropdownToggleProps` found while testing Kiali today. We currently allow passing `null` to the `iconComponent` to remove the caret icon from the dropdown. If `strictNullChecks` is set true, this will give a TypeScript error to the user. Making the slight adjustment here to the definition to allow `null`.

Screen Shot 2019-03-25 at 10 17 16 AM

cc: @aljesusg 

**Additional issues**:


aljesusg:
Approved from me side !!! :)

Issue #1507, opened 2019-03-05
Chip - allow chip to be read-only


gastaldi:
When setting the Chip as `readOnly`, that would remove the `X` next to it but keeping the same height

rachael-phillips:
@gastaldi do you mind providing us a little more context on what you'd like to achieve? @mcoker making you aware.

gastaldi:
@rachael-phillips sure, for example, in the first Chip Group Toolbar
example in  http://patternfly-react.surge.sh/patternfly-4/components/chipgroup,  I'd like to avoid `Chip 1` to be removed, but others may be removed (as if Chip 1 were a required chip). 

Therefore having the ability to mark a specific chip as `readOnly` would be really helpful

mcoker:
@rachael-phillips we should ask design if there should be any other visual indicators that the chip is read-only/non-removable, or if it's simply removing the button but keeping the existing overall layout, and then add an issue in core to add this variation.

mceledonia:
I think simply removing the icon button should be fine for this. If we start getting a lot of feedback that people aren't understanding why they can't remove the chip, we can reconsider.

rachael-phillips:
Ok great! Thanks everyone!

rachael-phillips:
I added the on hold label so that we can resolve this in our Core repo first (1533 Create read-only variation of chip). Then we will move it onto React. Thanks everyone!

mcoker:
@rachael-phillips this is in core now and ready to be added to react - https://github.com/patternfly/patternfly-next/pull/1555

tlabaj:
Closed by #1639

PR #1639, opened 2019-03-25
feat(Chip): add read-only chip, add single chip examples

kmcfaul:
**What**: Adds read-only flag to chip

**Refer to issue**: #1507


mcoker:
@kmcfaul looks good, except per the [core documentation](https://pf4.patternfly.org/components/Chip/examples/) the chip should be a `<div>` when not used in a chip group.

mcoker:
Not sure if this is in the scope of this PR or not, but looking at the last 2 examples (Chip Group Multi-Select and Badge Chip Group), the HTML looks wrong.

#### Chip group multi-select
* `.pf-c-chip-group` is a `<div>` but the chip inside of it is an `<li>`. I would expect `.pf-c-chip-group` to be a `<ul>`.
* `.pf-c-chip.pf-m-overflow` is nested in `.pf-c-chip-group`. I would expect that to follow the other examples with an overflow chip, where the overflow chip is adjacent to `.pf-c-chip-group`.

#### Badge chip group
* `.pf-c-chip-group` is a `<div>`. I would expect that to be a `<ul>`

LHinson:
Looks good @kmcfaul! Removing the UX review since I've covered the bases there.

PR #1642, opened 2019-03-26
fix(pagination): add "per page" suffix in options

### LINKED ISSUE #1528: PF4 Pagination: Update pagination example to match Core
Per @mcarrano comments on PR #1297, the example should be updated to match core. 
> Thanks for the explanation @karelhala . That all makes sense. Is it possible to update the options menu to read "10 per page," 20 per page," etc. per the design? Other than that, I think this is ready to merge

boaz0:
**What**:

closes #1528 

I added the "per page" suffix into the default props. Let me know if I'm missing something.

//cc @karelhala @mcarrano @tlabaj

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Pagination/OptionsMenu.js
+++ b/packages/patternfly-4/react-core/src/components/Pagination/OptionsMenu.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import styles from '@patternfly/patternfly/components/OptionsMenu/options-menu.css';
+import paginationStyles from '@patternfly/patternfly/components/Pagination/pagination.css';
 import { css } from '@patternfly/react-styles';
 import { Dropdown, DropdownItem, DropdownDirection } from '../Dropdown';
 import { CheckIcon } from '@patternfly/react-icons';
@@ -16,6 +17,7 @@ const propTypes = {
     value: PropTypes.number
   })),
   itemsPerPageTitle: PropTypes.string,
+  perPageSuffix: PropTypes.string,
   itemsTitle: PropTypes.string,
   optionsToggle: PropTypes.string,
   itemCount: PropTypes.number,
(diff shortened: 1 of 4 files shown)


mcarrano:
This looks much better @boaz0 . The only thing I'm seeing is that the text "per page" should have a lighter text style per the design. @mcoker did we ever update this in Core? Do you know what text style should be used here?

mcoker:
@mcarrano it is in core, we don't have an example of it shown. @boaz0 the class to use on that text in the menu is `.pf-c-pagination__menu-text`, so `<span class="pf-c-pagination__menu-text">per page</span>`

boaz0:
@karelhala can you take another look?
I wrapped `title` under `OptionMenu`, is that OK or I should move it to the default per page options?

Thanks.

karelhala:
@boaz0 I think that only the `per page` string should be in different color, am I right @mcarrano? The best way here would probably be to add another string to `titles` object with default `per page` and wrap only this with span.

mcarrano:
Yes, that's correct @karelhala . The different styling should only apply to the text "per page."

mcarrano:
I am also good with this. It does look like the font size for "per page" is a bit large compared with the design, But if that's off, it would be a CSS issue and I can live with it in any case. @mcoker can you double check this?

mcoker:
@mcarrano correct, it's an issue in core. Should the font size be `$pf-global--FontSize--sm` (14px)?

karelhala:
Can you also add this string to [patternfly-4/react-core/src/components/Pagination/Pagination.js#L59](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Pagination/Pagination.js#L59) and pass it to `OptionsMenu`? So we can change this string from outside? Either because of internacionalization, or because we'd want to change it to something different.

mcarrano:
14px sounds right @mcoker but someone on the visual team should verify. @kybaker @mceledonia ? It's the text in the items per page menu within pagniation.

mceledonia:
Yes 14px is correct here. Let's not have this beast take up more room than it already does 🤠

tlabaj:
@boaz0   can you resolve the conflicts please.

boaz0:
@tlabaj I rebased the change. thanks.

boaz0:
Thanks a lot @tlabaj and @dlabaj 
yeay

PR #1645, opened 2019-03-26
fix(InlineEdit): remove flickering of dropdown in an example

suomiy:
Dropdown toggle button focused state was lost on rerender. Sadly the issue with gatsby to select an actual item still persist.

@jschuler @priley86 @karelhala


suomiy:
fixes  https://github.com/patternfly/patternfly-react/issues/1651

PR #1648, opened 2019-03-27
fix(tooltip): trigger tooltip by mouse click or other events

### LINKED ISSUE #1564: PF4 Tooltip - clicking prevents tooltip from opening
If I click a tooltip, the tooltip box does not open, even if I continue to hover the mouse over it. I would expect a click to open the tooltip the same way a mouse-over or keyboard focus does.

boaz0:
**What**:

closes #1564 

Adding the `trigger` props to display tooltip on mouse click.

//cc @keithjgrant

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.js
+++ b/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.js
@@ -21,6 +21,8 @@ export const TooltipPosition = {
 const propTypes = {
   /** Tooltip position */
   position: PropTypes.oneOf(Object.keys(TooltipPosition).map(key => TooltipPosition[key])),
+  /** Tooltip trigger: click, mouseenter, focus */
+  trigger: PropTypes.string,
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.js
```diff
@@ -21,6 +21,8 @@ export const TooltipPosition = {
 const propTypes = {
   /** Tooltip position */
   position: PropTypes.oneOf(Object.keys(TooltipPosition).map(key => TooltipPosition[key])),
+  /** Tooltip trigger: click, mouseenter, focus and etc */
```
jschuler: I would say let's be specific here and limit it to just 3 of the possibly 4 tippy variants (cannot support `manual` at the moment since we don't expose the isVisible prop for controlled mode)
```suggestion
  /** Tooltip trigger: click, mouseenter, focus */
```

boaz0:
Thanks @jschuler, @dlabaj and @tlabaj

PR #1650, opened 2019-03-27
[PF3] Select Component

glekner:
**What**:
WIP for a Select Component with a Search Input.
I took the design from [here](https://www.patternfly.org/pattern-library/widgets/#bootstrap-select) even though there is no search bar.
Screenshot 2019-03-27 17 35 02

Right now the Component accept for options an object e.g. 
```
{
   '1': 'one',
   '2': 'two',
    ...
}
```

**TODO**:
- tests
- css variables
- more features?

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/less/select.less
+++ b/packages/patternfly-3/patternfly-react-extensions/less/select.less
@@ -1,0 +1,140 @@
+@select-max-height: 205px;
+@select-min-width: 220px;
+
+.select-container-pf {
+  display: flex;
+  flex-direction: column;
+  padding-left: 0;
+  padding-right: 0;
+  overflow: hidden;
+  width: 100%;
+  min-width: @select-min-width;
+  border: 1px solid @color-pf-black-400;
+  z-index: 1000;
+
+  &.open {
+    box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
+  }
+
+  .select-dropdown-toggle {
+    display: flex;
+    align-items: center;
+    text-align: left;
+    border: 0;
+
+    .select-toggle-placeholder {
+      font-style: italic;
+      color: @color-pf-black-500;
+    }
+
+    .select-caret {
+      margin-left: auto;
+    }
+  }
+
+  .select-body-container {
+    display: flex;
+    flex-direction: column;
+    padding-left: 0;
+    padding-right: 0;
+    max-height: @select-max-height;
+    overflow: hidden;
+    width: 100%;
+    min-width: @select-min-width;
+    z-index: 1000;
+
+    .select-input-search {
+      position: relative;
+
+      input {
+        padding: 0 25px 0 25px;
+      }
+
+      .fa-search {
+        color: @color-pf-blue-400;
+        position: absolute;
+        top: 7px;
+        left: 8px;
+        z-index: 99;
+      }
+
+      .fa-close {
+        cursor: pointer;
+        position: absolute;
+        top: 7px;
+        right: 10px;
+        z-index: 99;
+      }
+    }
+
+    .select-scrollable-list {
+      overflow: auto;
+
+      .list-group-item {
+        padding: 5px 10px 5px 10px;
+        outline: 0;
+
+        &:hover {
+          background-color: @color-pf-blue-25;
+        }
+
+        &.active {
+          background-color: @color-pf-blue-400;
+        }
+
+        &.cursor:not(.active) {
+          background-color: @color-pf-blue-25;
+        }
+
+        &.no-border {
+          border: 0;
+        }
+
+        &:focus {
+          background-color: @color-pf-blue-400;
+          color: @color-pf-white;
+        }
+
+        &#select-empty {
+          user-select: none;
+          font-style: italic;
+          color: @color-pf-black-500;
+          display: flex;
+
+          #empty-text {
+            margin-left: auto;
+            margin-right: auto;
+          }
+        }
+
+        &#select-loading {
+          &:hover {
+            background-color: @color-pf-white;
+          }
+
+          margin-left: auto;
+          margin-right: auto;
+          display: flex;
+          text-align: center;
+          user-select: none;
+          font-style: italic;
+          color: @color-pf-black-500;
+
+          #select-loading-container {
+            margin-left: auto;
+            margin-right: auto;
+            display: flex;
+
+            #select-spinner {
+              margin-right: 5px;
+            }
+          }
+        }
+      }
+    }
+
+    .list-group {
+      margin-bottom: 0;
+    }
+  }
+}
--- a/packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.js
+++ b/packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.js
@@ -1,0 +1,168 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import EllipsisWithTooltip from 'react-ellipsis-with-tooltip';
+import { Spinner, Button, ListGroup, ListGroupItem, TypeAheadSelect } from '../../../../patternfly-react/src/index';
+import { noop } from '../../common/helpers';
+import { createItemProps } from './SelectHelper';
+import SelectInput from './SelectInput';
+
+const Select = ({
+  disabled,
+  emptyText,
+  focusOnMount,
+  isLoading,
+  isSearchable,
+  loadingText,
+  onItemClick,
+  onKeyDown,
+  onSearchChange,
+  onSearchClear,
+  onToggle,
+  open,
+  options,
+  placeholder,
+  searchValue,
+  selectedItem
+}) => {
+  const classes = classNames('select-container-pf', { open });
+
+  const renderResults = results =>
+    results.length === 0 ? (
+      <ListGroupItem id="select-empty" key="empty" className="select-empty-list">
+        <span id="empty-text">{emptyText}</span>
+      </ListGroupItem>
+    ) : (
+      results.map((opt, i) => (
+        <ListGroupItem {...createItemProps(opt, selectedItem, opt.className || 'no-border', onItemClick)}>
+          <EllipsisWithTooltip>
+            {searchValue && searchValue.length ? (
+              <TypeAheadSelect.Highlighter search={searchValue}>{opt.name}</TypeAheadSelect.Highlighter>
+            ) : (
+              opt.name
+            )}
+          </EllipsisWithTooltip>
+        </ListGroupItem>
+      ))
+    );
+
+  return (
+    <div className={classes}>
+      <Button disabled={disabled} onClick={onToggle} active={open} className="select-dropdown-toggle">
+        {selectedItem.id ? (
+          <EllipsisWithTooltip>{selectedItem.name}</EllipsisWithTooltip>
+        ) : (
+          <EllipsisWithTooltip>
+            <span className="select-toggle-placeholder">{placeholder}</span>
+          </EllipsisWithTooltip>
+        )}
+        <span className="caret select-caret" />
+      </Button>
+      {open && (
+        <div className="select-body-container">
+          {isSearchable && (
+            <SelectInput
+              onClear={onSearchClear}
+              focus={focusOnMount}
+              onChange={onSearchChange}
+              onKeyDown={onKeyDown}
+              placeholder={placeholder}
+              searchValue={searchValue}
+            />
+          )}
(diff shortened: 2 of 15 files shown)


glekner:
- added a disabled prop to an option, so now the options object looks like this
`{ '1': { label: 'one', disabled: false } ... } `

Hyperkid123:
Can we change the format to an Array instead of Object? 

Something like:
```javascript
[{
  label: 'Foo',
  value: 1,
  ...
}, ...]
```

I mean if you want to render anything from Object in React you have to convert it into Array anyway.

glekner:
> Can we change the format to an Array instead of Object?
> 
> Something like:
> 
> ```js
> [{
>   label: 'Foo',
>   value: 1,
>   ...
> }, ...]
> ```
> 
> I mean if you want to render anything from Object in React you have to convert it into Array anyway.
> If we were to use this in existing applications we would have to transform all of our select options into Objects.

@Hyperkid123 
I Agree, but Many apps use JQuery Select that accepts an Object so I thought we might want to support that. I'm fine with switching to array

Hyperkid123:
I don't think we want to model our components on jQuery at all.

glekner:
- Select now accepts an Array
- Fixed css bugs and overflows

@Hyperkid123 @jeff-phillips-18 @serenamarie125 can you please try the component?
we need this component before the freeze. Thanks!

Hyperkid123:
Do you plan adding a multi select version of this?

glekner:
> Do you plan adding a multi select version of this?
In the future if its ok.

> How do you plan to handle clearing selected value? Adding an option with empty value or adding some icon which would clear the value?
Current Select designs doesn't have this feature - clearing the value. Would be able to add it in a future PR.

> Also i think that the select should expand when the selected option label is longer than the value container. It should extend to 100% width of parent element. @terezanovotna can tell you more about the actual PF pattern.
K, working on it.

glekner:
- Changed width to fill 100% of container
- Added `onKeyDown` and `cursor` props, the Storybook now responds to `up` `down` and `enter` keys upon focusing the search input

terezanovotna:
Hey @glekner happy to review ux for this PR, when you have a storybook, feel free to ping me!

Awesome you are adding search into the multi select. 
I suggest having the same design and behavior as [this component](https://www.patternfly.org/pattern-library/widgets/#bootstrap-select) you suggested. 

When more than 4 items selected - we can display `4 selected` instead of naming them all as there is not enough space for it.

glekner:
Thanks @terezanovotna ! the storybook the PR created is a bit outdated but the design is about the same, [try it](https://1650-pr-patternfly-react-patternfly.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react%2FWidgets%2FAsyncSelect&selectedStory=AsyncSelect&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)

@Hyperkid123 anyway to update the PR's storybook?

Hyperkid123:
> @Hyperkid123 anyway to update the PR's storybook?

It should update once all checks are fine (no conflicts tests etc.)

glekner:
> > @Hyperkid123 anyway to update the PR's storybook?
> 
> It should update once all checks are fine (no conflicts tests etc.)

Thanks, any idea why travis isn't running?

Hyperkid123:
Try close/open the PR.

Travis is acting up lately (on every project)

Hyperkid123:
Hmm did not help i will ask someone to reset travis.

glekner:
added `disabled` prop for the Select, and added snapshot file.
travis still doesn't seem to run after rebase..

Hyperkid123:
I have no clue what is going on wit travis. We restated it manually yesterday. Maybe we could try deleting cache on travis.

terezanovotna:
the styling of filter (blue border) should reflect [pf style](https://www.patternfly.org/pattern-library/forms-and-controls/find/#code)

As @Hyperkid123 mentioned it would be great to have multi select.

dtaylor113:
Hi @Hyperkid123, looks great!   Components which do not have PatternFly designs/patterns should go into `patternfly-react-extensions` package.  -thanks

Hyperkid123:
@dtaylor113  i actually did not write any code just trying to help @glekner

glekner:
@dtaylor113 
- Added `isSearchable` prop to control showing the search input
- Moved to `patternfly-react-extensions`

Can we merge? :)

dtaylor113:
> Can we merge? :)

We need another approving review

packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.js
```diff
@@ -0,0 +1,181 @@
...
+
+    if (item.disabled) return { ...itemProps, disabled: true };
+    return { ...itemProps, onClick: e => onChange({ e, id: item.id, name: item.name }) };
+  };
+
+  const renderResults = results =>
+    results.length === 0 ? (
+      <ListGroupItem id="select-empty" key="empty" className="select-empty-list">
+        <span id="empty-text">{emptyText}</span>
+      </ListGroupItem>
```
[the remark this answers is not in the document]
glekner: This is not a separate component, only a function for better readability, pretty common in this library.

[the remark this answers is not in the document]
Laviro: I meant seperating it, for example: 
- EmptyState
- ListItems
And other parts of the code that you could break into smaller components or helpers file.

glekner:
@dtaylor113 @jeff-phillips-18 can you ack? Thanks

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
+@select-max-height: 205px;
+@select-min-width: 220px;
+
+.select-container {
```
jeff-phillips-18: Should have a `-pf` suffix

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
...
+  display: flex;
+  flex-direction: column;
+  padding-left: 0;
+  padding-right: 0;
+  overflow: hidden;
+  width: 100%;
+  min-width: @select-min-width;
+  border: 1px solid #ccc;
```
jeff-phillips-18: Can we use a color from the palette here?

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
...
+    overflow: hidden;
+    width: 100%;
+    min-width: @select-min-width;
+    z-index: 1000;
+
+    .select-input-search {
+      position: relative;
+
+      input {
+        width: 100% !important;
```
jeff-phillips-18: this is a `form-control` which already has width: 100%. Why this and why the `!important`?

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
...
+          background-color: @color-pf-blue-25;
+        }
+
+        &.no-border {
+          border: 0;
+        }
+
+        &:focus {
+          background-color: @color-pf-blue-400;
+          color: #fff;
```
jeff-phillips-18: color-pf-white

packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.js
```diff
@@ -0,0 +1,176 @@
...
+          </ListGroup>
+        </div>
+      )}
+    </div>
+  );
+};
+
+Select.propTypes = {
+  /** cursor position for keyboard navigation */
+  cursor: PropTypes.number,
```
jeff-phillips-18: why are we controlling keyboard navigation via the arrow keys rather than allowing the native control (tab)?

glekner: who said it has to be the arrow keys? its arrow keys just for the storybook example.

jeff-phillips-18: ok, so what is the point in general?

glekner: the arrow keys is a good example but it doesn't scroll if the cursor is going below the visible list. Maybe its best to remove this functionality for now and add it in a future PR, we want this hopefully to be merged today. Thoughts?

jeff-phillips-18: I'd be ok with removing it and adding if/when its deemed necessary

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
...
+
+  .select-dropdown-toggle {
+    display: flex;
+    align-items: center;
+    text-align: left;
+    border: 0;
+
+    .select-toggle-placeholder {
+      font-style: italic;
+      color: #9c9c9c;
```
jeff-phillips-18: palette color

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
...
+
+        &:focus {
+          background-color: @color-pf-blue-400;
+          color: #fff;
+        }
+
+        &#select-empty {
+          user-select: none;
+          font-style: italic;
+          color: #9c9c9c;
```
jeff-phillips-18: palette color

packages/patternfly-3/patternfly-react-extensions/less/select.less
```diff
@@ -0,0 +1,141 @@
...
+            background-color: @color-pf-white;
+          }
+
+          margin-left: auto;
+          margin-right: auto;
+          display: flex;
+          text-align: center;
+          user-select: none;
+          font-style: italic;
+          color: #9c9c9c;
```
jeff-phillips-18: palette color

packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.js
```diff
@@ -0,0 +1,176 @@
...
+  onKeyDown: noop,
+  onSearchChange: noop,
+  onSearchClear: noop,
+  onToggle: noop,
+  placeholder: 'Filter...',
+  searchValue: '',
+  selectedItem: {
+    id: '10',
+    name: 'ten'
+  }
```
jeff-phillips-18: could cause issues for applications not supplying a selectedItem, if they happen to have a matching item it will be inadvertently selected.

glekner:
implemented @jeff-phillips-18 suggestions and removed `cursor` functionality for now.

@dtaylor113 @jeff-phillips-18 ack?

packages/patternfly-3/patternfly-react-extensions/src/components/Select/StatefulWrapperSelect.js
```diff
@@ -0,0 +1,101 @@
+import React, { Component } from 'react';
+import enhanceWithClickOutside from 'react-click-outside';
+import { Select } from './index';
+import { optionsArray } from './Select.fixtures';
+
+class StatefulWrapperSelect extends Component {
+  state = {
+    open: false,
+    cursor: 0,
```
jeff-phillips-18: Should probably remove the cursor code within this class to avoid confusions.

glekner:
@jeff-phillips-18 removed `cursor` from stateful wrapper

jeff-phillips-18:
@glekner Could you please reword your commit?

jeff-phillips-18:
Looks like you need to update your snapshots:

```
Summary of all failing tests
 FAIL  packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.test.js
  ● Select renders properly

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "Select renders properly 1".

    - Snapshot
    + Received

    @@ -1,7 +1,6 @@
      <Select
    -   cursor={-1}
        disabled={false}
        emptyText="No Results"
        focusOnMount={true}
        isLoading={false}
        isSearchable={true}

      16 |   ).getElement();
      17 | 
    > 18 |   expect(component).toMatchSnapshot();
         |                     ^
      19 | });
      20 | 
      21 | test('Select functionality', () => {

      at Object.toMatchSnapshot (packages/patternfly-3/patternfly-react-extensions/src/components/Select/Select.test.js:18:21)
```

glekner:
fixed commit and snapshot @jeff-phillips-18 @dtaylor113

PR #1652, opened 2019-03-28
fix(select): remove the custom override on width

### LINKED ISSUE #1610: Update Select Width
This was just merged in Core - https://github.com/patternfly/patternfly-next/pull/1595
It updates the width for the select to be 100%.

A while back, a custom override was added to the width of the select component-- https://github.com/patternfly/patternfly-react/pull/1277#issuecomment-460663434

But we hadn't followed up with the issue in Core. Now you can remove the custom override for the width of the select.

boaz0:
**What**:
This change removes `width` from `defaultProps` as `width: 100%` is set in the class.
closes #1610 
//cc @christiemolloy @kmcfaul

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.js
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.js
@@ -50,7 +50,7 @@ const defaultProps = {
   selections: null,
   placeholderText: null,
   variant: SelectVariant.single,
-  width: '100%'
+  width: null,
 };
 
 class Select extends React.Component {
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -42,7 +42,6 @@ const defaultProps = {
   selections: null,
   placeholderText: null,
   variant: SelectVariant.single,
-  width: '100%'
```
tlabaj: don't we need a default prop here since the width is not required?

boaz0: If I understood correctly, the `width` should be set by `pf-c-select`.
I didn't remove `width` from props in case someone wants to change that attribute, though.

jschuler: Probably makes more sense to default it to empty string or null, this way your check still works below

boaz0:
Thank you all @kmcfaul, @tlabaj and @jschuler

Issue #1460, opened 2019-02-27
Table - accessible examples


jgiardino:
There are several cases where our examples don't represent best practice with regard to accessibility.

- **aria-labelledby attributes on the checkboxes and expand buttons**
  - **html elements affected:**
    - `<input aria-labelledby="simple-node0" name="checkrow0" type="checkbox">`
    -  `<button aria-labelledby="simple-node1 expandable-toggle1" id="expandable-toggle1" aria-label="Details" aria-expanded="true" class="pf-c-button pf-m-plain pf-m-expanded" type="button">`
  - **issue:** `aria-labelledby` is defined, but the value doesn't exist as an `id` on the first content cell
  - **update:** In previous versions of these examples (at least as far back as https://github.com/patternfly/patternfly-react/pull/1324, maybe sooner), this was handled in the following way:
    ```
    <th data-label="Repositories" data-key="1">
      <div id="simple-node0">one</div>
    </th>
    ```
- **Examples showing the action button should include row headers**
  - One of the accessibility benefits for having a row header is that the row header provides context for other cell contents in that row. This is especially important for examples where the Actions button is included, so that the screen reader user knows which item is being affected by the actions in the kebab. There is a react example that illustrates how to define a row header—“First cell as Header”. This same method should be used in the react examples that show action buttons.

PR #1659, opened 2019-03-29
fix(Table): Fix Table component accessibility issues

TheRealJon:
Fixes https://github.com/patternfly/patternfly-react/issues/1460

**What**: Add headerCol cell transform to the first column of actions, collapsible, selectable, and simple actions table examples. This makes the first cell of each row into a `th` element and wraps the content in a `div` with an id that will match the `aria-labeledby` attribute on selectable row checkboxes and/or expandable row expand buttons.

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.md
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.md
@@ -2,6 +2,7 @@
 title: 'Table'
 cssPrefix: 'pf-c-table'
 ---
+
 Note: Table lives in its own package at [`@patternfly/react-table`](https://www.npmjs.com/package/@patternfly/react-table)!
 
 import {
@@ -17,6 +18,7 @@ import {
 } from '@patternfly/react-table';
 
 ## Simple Table
+
 ```js
 import React from 'react';
 import {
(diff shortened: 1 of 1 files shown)


TheRealJon:
cc @jgiardino

jgiardino:
Sorry for the delay in reviewing. I keep waiting for a preview link to display. Is that something that is easily added?

TheRealJon:
@tlabaj @dlabaj @karelhala I had to rebase, need review again.

TheRealJon:
@jgiardino Once the CI passes, there should be a link. One of the CI tasks was failing, which is probably why we never got the preview link. I just rebased which kicked the CI, so hopefully it will pass this time.

jgiardino:
A preview link is included now, but I'm not able to find the Table component from that link. Any suggestions?

TheRealJon:
@jgiardino The nav looks a little different now, but it's under the 'components' nav item.

jgiardino:
oooooh, thanks! On my screen, it wasn't obvious at all that there were more components listed if I scrolled.

dlabaj:
@TheRealJon Looks like there's a conflict.

PR #1663, opened 2019-03-29
fix(table): Add optional types for table column

### LINKED ISSUE #1463: Table - column configs should have optional typescript types
I noted while reviewing [Facet](https://github.com/metalkube/facet) that column configs should be optionally defined when passing transforms, cellTransforms, formatters, and cellFormatters.

https://github.com/metalkube/facet/blob/master/src/components/HostsTable.tsx#L25

jcaianirh:
fix #1463

**What**: column configs now optionally defined for transforms, cellTransforms, formatters, and cellFormatters

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
@@ -62,9 +62,9 @@ export interface ISeparator {
 export interface ICell {
   title: String;
   transforms: Array<Function>;
-  cellTransforms: Array<Function>;
-  formatters: Array<Function>;
-  cellFormatters: Array<Function>;
+  cellTransforms?: Array<Function>;
+  formatters?: Array<Function>;
+  cellFormatters?: Array<Function>;
   props: any;
 }
 


packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -60,10 +60,10 @@ export interface ISeparator {
 
 export interface ICell {
   title: String;
-  transfroms: Array<Function>;
-  cellTransforms: Array<Function>;
-  formatters: Array<Function>;
-  cellFormatters: Array<Function>;
+  transfroms?: Array<Function>;
```
karangarg45: shouldn't it be `transforms` ?

dgutride: the existing property was named transfroms - we should probably fix the naming as part of this breaking change right now

redallen: Fixed it!

PR #1664, opened 2019-03-30
fix(Table): fix type-o in ICell Interface transform (from transfrom)

### LINKED ISSUE #1662: Table - column configs has type-o in typescript type: transform (transfrom)
In table.d.ts, the column configs has type-o in typescript type: transform (transfrom)

Was going to fix this in pr to fix: https://github.com/patternfly/patternfly-react/issues/1463
but didn't want to introduce a breaking change.

jcaianirh:
Add a fix found by a user to change the type-o of property transfrom to transform

Will only break for typescript users using the type definition files.

fix #1662

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
@@ -60,7 +60,7 @@ export interface ISeparator {
 
 export interface ICell {
   title: String;
-  transfroms: Array<Function>;
+  transforms: Array<Function>;
   cellTransforms: Array<Function>;
   formatters: Array<Function>;
   cellFormatters: Array<Function>;


dgutride:
Please remove BREAKING CHANGE from this commit message and force push to this branch again.  We are not using that semantic messaging right now in order to prevent unnecessary versioning.  This PR should not be merged until we enter a quiet period.

jcaianirh:
@dgutride @dlabaj removed bc text in commit body.

dgutride:
@jcaianirh  - thanks for looking at this, unfortunately, the commit body still contains the text - the commit message will need to be modified and force pushed back to the branch

jcaianirh:
@dgutride, changed commit message, force pushed.

PR #1667, opened 2019-04-01
fix(Dropdown): Send event when clicking on dropdown toggle

karelhala:
**What**:
Consumers might want to choose to override events from dropdown toggle for instance to stop propagation to parent (when dropdown located in clicable element). For this reason we should send event on toggle callback.


redallen:
@tlabaj Rebasing fixed this one, good to merge now.

Issue #1615, opened 2019-03-20
dropdown menu reusable interaction pattern


jgiardino:
Whenever a toggle + menu interaction pattern is implemented in a component, the methods used to handle user interaction should be consistent across all the components that use this pattern. 

This issue is to fully implement this interaction pattern for one of the components.
Then follow-up issues should capture this work for any other components that include this pattern.

Some of the components we have now that include this interaction pattern are:
- Dropdown
- Select
- Pagination
- Context Selector
- Wizard
- AppLauncher

**[Details about interaction pattern requirements](https://docs.google.com/document/d/1YdjO23kgB4MKmc7_rvXbS5ml0a08FobGDGQPYeTPEgo/edit?usp=sharing)**

This work is dependent on having more modular components for menu and toggle available in core - https://github.com/patternfly/patternfly-next/issues/1429

nicolethoen:
This can be considered a duplicate of https://github.com/patternfly/patternfly-react/issues/6014 since all dropdown menus will use the same Menu component under the hood, they will be using the same interaction patterns.

PR #1668, opened 2019-04-01
feat(ApplicationLauncher): Use dropdown implementation instead of custom implementation

karelhala:
**What**:
Application launcher was using custom logic to open/close and select items instead of dropdown implementation. This will eventually cause problems when dropdown changes default behavior and would lead to chasing same bugs in two different components. Plus it removes bunch of duplicate code.

**Additional issues**:


karelhala:
This partially resolves https://github.com/patternfly/patternfly-react/issues/1615

dtaylor113:
Hi @karelhala, just want to verify this isn't a breaking change?  Appears that way as I don't see any changes to the 'examples'  -thanks

karelhala:
@dtaylor113 yes, this should not be breaking change, it is only a changed implementation. Instead of custom logic it is now using shared dropdown logic. We had small discussion with @dgutride about this and he pointed out that we can't have breaking changes in this state, this is not changeing API in any way it is using same tests and same examples as before.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
```diff
@@ -23,81 +17,33 @@ export const propTypes = {
...
-              </DropdownContext.Provider>
-            )}
-          </div>
-        )}
-      </GenerateId>
-    );
+    return <Dropdown
+      { ...this.props }
+      toggle={
+        <DropdownToggle iconComponent={null} onToggle={this.props.onToggle}>
```
jschuler: Can you check if a11y is the same? For example, before aria-label was on the button, now it;s on the outer div. Also the button does not have aria-expanded set to true like before

karelhala: > Also the button does not have aria-expanded set to true like before

I can see that is has aria-expanded, strange.

Before the a11y was broken on menu because it had `aria-labelledby="Actions"` where `"Actions"` was not any ID in DOM, and yeah the aria label should be on the button, not on div, gotta fix that by passing `aria-label` to toggle.

jschuler: Sorry, I can also see aria-expanded now

dlabaj:
@karelhala We use to have this implementation however it was decided to move it to a separate component to handle changes as the application launcher evolves from a drop down.  This was to prevent breaking consumers using the application launcher.  While the logic isn't ideal and there is duplicate code, there is a future change coming that is based off this work being done in core https://github.com/patternfly/patternfly-next/issues/1429  that will address the duplicate code issues.  This change will also diverge the styles that we have from application launcher at the moment.  While the styles are the same now if core is to update the style, or direction changes on the issue above we will be out of sync.  It's best to address down the line once we get resolution on the 1429.  Thanks.

karelhala:
@dlabaj can you explain how this can diverte from core design? If core design were to change apllication launcher from dropdown to something else I could understand that and it would mean that we will have to change this completely evem with current implementation. But code duplicity just because the component can change in future is bad idea. The way how amy dropdown component should work is to have context menu and action component. This improvement will help us to achieve this much easier and in meantime if something changes in dropdown we will have to add it to 4 other components.

PR #1691, opened 2019-04-02
build(lerna): make lerna only do patches [ci skip]

redallen:
**What**: Prevent accidental major version bumps like https://travis-ci.org/patternfly/patternfly-react/builds/514802863

**Additional issues**: No longer get minor releases from semver.


redallen:
We want minor releases, so closing.

PR #1694, opened 2019-04-03
fix(pagination): remove size attr from type="number" input

### LINKED ISSUE #1676: Remove size attr from type="number" input in pagination
https://github.com/patternfly/patternfly-next/pull/1644

boaz0:
**What**:
closes #1676
 
`type="number` inputs don't have the `size` attribute. 
Nevertheless, using `width: auto` (which is the default) the browser determines the right size of the input.


boaz0:
Thanks @mcoker @tlabaj and @jschuler :pray: :1st_place_medal:

PR #1700, opened 2019-04-04
fix(Table): Add ability to center-align header and cell text

### LINKED ISSUE #1482: PF4: Table header and the Table data cell should have prop for center alignment
There is no way to set the Table header or table data cell  alignment to be centered.  Prop should be added that  would indicate the alignment should be centered.

For the Table header,  modifier table-th--modifier="pf-m-center" should be applied to center align
For the Table data cell,  modifier table-td--modifier="pf-m-center" should be applied to center align

TheRealJon:
fix #1482

**What**:

- Add a `textCenter` transform which applies `textCenter=true` prop to header and body cells. This prop conditionally adds the `pf-m-center` css class to header and body cells. This allows the transform to be easily overridden by applying `textCenter=false` prop to an individual cell definition.

- Change the order in which transforms are applied in `components/Table/utils/headerUtils.js:generateCell` function. Previously, props that were defined at the individual cell level were used as the default, then any transforms applied at the column level were overriding those values. This is the reverse of expected behavior. Transforms applied at the column level should be considered the default, then props/transforms should override each other as specificity increases (with the cell level being most specific, overriding all others).

- Add `textCenter` transform and `textCenter` prop override to Simple Table example.

- Fix some lint and prop validation errors in the files that were touched

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/utils/headerUtils.js
+++ b/packages/patternfly-4/react-table/src/components/Table/utils/headerUtils.js
@@ -37,9 +37,9 @@ const generateHeader = ({ transforms: origTransforms, formatters: origFormatters
 const generateCell = ({ cellFormatters, cellTransforms, cell }) => ({
   ...cell,
   transforms: [
-    mapProps,
     ...(cellTransforms || []),
-    ...(cell && cell.hasOwnProperty('transforms') ? cell.transforms : [])
+    ...(cell && cell.hasOwnProperty('transforms') ? cell.transforms : []),
+    mapProps // This transform should be applied last so that props that are manually defined at the cell level will override all other transforms.
   ],
   formatters: [
     defaultTitle,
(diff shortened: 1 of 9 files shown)


TheRealJon:
@karelhala Sorry, had to rebase. Need approval again.

packages/patternfly-4/react-table/src/components/Table/utils/headerUtils.js
```diff
@@ -37,9 +37,9 @@ const generateHeader = ({ transforms: origTransforms, formatters: origFormatters
 const generateCell = ({ cellFormatters, cellTransforms, cell }) => ({
   ...cell,
   transforms: [
-    mapProps,
     ...(cellTransforms || []),
-    ...(cell && cell.hasOwnProperty('transforms') ? cell.transforms : [])
+    ...(cell && cell.hasOwnProperty('transforms') ? cell.transforms : []),
+    mapProps // This transform should be applied last so that props that are manually defined at the cell level will override all other transforms.
```
TheRealJon: @karelhala Is this transform even necessary? I feel like Reactabular takes care of applying body cell props that are defined in the `rows` props.

TheRealJon:
@dlabaj Sorry, had to push a fix for the CI failure. Could you approve again?

Issue #1684, opened 2019-04-01
Add support for .pf-m-grid-xl breakpoint in table


mcoker:
https://github.com/patternfly/patternfly-next/pull/1417

Issue #1683, opened 2019-04-01
Allow for multiple card bodies, pf-m-no-fill modifier


mcoker:
the card should support using multiple body elements, and 
a prop (uses `.pf-m-no-fill`) that keeps a body/bodies from growing to fill the height of the card. https://github.com/patternfly/patternfly-next/pull/1484

tlabaj:
Closed by #1703

Issue #1682, opened 2019-04-01
Add wrapper to accordion, datalist, page (sidebar)


mcoker:
Need to add these elements. They should be a direct child of the parent BEM element

* `.pf-c-data-list__expandable-content-body` (child of `.pf-c-data-list__expandable-content`)
* `.pf-c-page__sidebar-body`(child of `.pf-c-page__sidebar`)
* `.pf-c-accordion__expanded-content-body` (child of `.pf-c-accordion__expanded-content`)

tlabaj:
Accordion has not been merged in react yet. @ibolton336  can you add the accordion wrapper to your PR please?
.pf-c-accordion__expanded-content-body (child of .pf-c-accordion__expanded-content)

tlabaj:
Closed by #1703

Issue #1680, opened 2019-04-01
refactor nav component item separator


mcoker:
The separator is its own element now instead of a modifier on the nav link.

https://github.com/patternfly/patternfly-next/pull/1586

cc @jeff-phillips-18

tlabaj:
Closed by #1703

PR #1703, opened 2019-04-04
fix: Update React with changes from core

tlabaj:
#1680 (API change)
#1706 (API change)
#1683 
#1682 
#1708 (API change)

@rebeccaalpert fixes: 
#1684
#1679 

@jschuler fixes
Remove footerRightAlign prop (patternfly/patternfly-next#1661)
Remove backgroundImageSrc prop (patternfly#1681)

@dlabaj 
#1674  

**What**:

**Additional issues**:


package.json
```diff
@@ -116,7 +116,8 @@
     "test:pf3": "jest packages/patternfly-3",
     "test:pf4": "jest packages/patternfly-4",
     "test:watch": "jest --watchAll",
-    "uninstall": "find . -name node_modules -type d | xargs rm -rf"
+    "uninstall": "find . -name node_modules -type d | xargs rm -rf",
+    "postinstall": "rm -rf node_modules/@patternfly/patternfly && mv node_modules/dg-testrelease/ node_modules/@patternfly/patternfly"
```
jschuler: Make sure to remove this before merging

packages/patternfly-4/react-charts/package.json
```diff
@@ -46,7 +46,7 @@
     "victory-core": "^31.1.0"
   },
   "devDependencies": {
-    "@patternfly/patternfly": "1.0.250",
+    "dg-testrelease": "1.0.10",
```
jschuler: Need to update once core merges to master

packages/patternfly-4/react-core/package.json
```diff
@@ -58,7 +58,8 @@
     "@babel/plugin-transform-object-assign": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
-    "@patternfly/patternfly": "1.0.250",
+    "@babel/preset-typescript": "^7.0.0",
+    "dg-testrelease": "1.0.10",
```
jschuler: ^

packages/patternfly-4/react-core/src/components/Nav/Nav.docs.js
```diff
@@ -0,0 +1,35 @@
+import { Nav, NavGroup, NavList, NavItem, NavItemSeparator, NavExpandable, NavVariants } from '@patternfly/react-core';
```
jschuler: This file should not exist. No more *.docs.js files. Examples are in markdown file

packages/patternfly-4/react-core/src/components/Nav/NavItemSeparator.d..ts
```diff
@@ -0,0 +1,9 @@
+import { FunctionComponent, HTMLProps } from 'react';
```
jschuler: the file name has an extra . `d..ts`

packages/patternfly-4/react-core/src/components/Page/examples/common/getContainerProps.js
```diff
@@ -0,0 +1,19 @@
+import { StyleSheet, css } from '@patternfly/react-styles';
```
jschuler: This puts the borders back?

packages/patternfly-4/react-docs/src/components/css-variables.js
```diff
@@ -0,0 +1,183 @@
+import React from 'react';
```
jschuler: This file should not be merged in

packages/patternfly-4/react-inline-edit-extension/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/",
   "dependencies": {
-    "@patternfly/patternfly": "1.0.250",
+    "dg-testrelease": "1.0.10",
```
jschuler: ^

packages/patternfly-4/react-styled-system/package.json
```diff
@@ -44,7 +44,7 @@
     "react": "^16.4.0"
   },
   "devDependencies": {
-    "@patternfly/patternfly": "1.0.250",
+    "dg-testrelease": "1.0.10",
```
jschuler: ^

packages/patternfly-4/react-core/package.json
```diff
@@ -58,7 +58,8 @@
     "@babel/plugin-transform-object-assign": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
-    "@patternfly/patternfly": "1.0.250",
+    "@babel/preset-typescript": "^7.0.0",
```
jschuler: Why is this added?

package.json
tlabaj: yup. this was just interim so we could start working on changes.

packages/patternfly-4/react-charts/package.json
tlabaj: yup. this was just interim so we could start working on changes.

packages/patternfly-4/react-core/package.json
tlabaj: probably a merge issue.

packages/patternfly-4/react-core/src/components/Nav/NavItemSeparator.js
```diff
@@ -0,0 +1,23 @@
...
+  className: PropTypes.string,
+  /** Additional props are spread to the container <li> */
+  '': PropTypes.any
+};
+
+const defaultProps = {
+  className: ''
+};
+
+const NavItemSeparator = ({ className, ...rest }) => <li className={css(styles.navSeparator, className)} {...rest} />;
```
mcoker: this needs `role="separator"`

mcoker:
@dlabaj in #1678, the empty first cell has `<td class="pf-c-table__toggle">`. That should just be an empty `<td></td>`. Also looks like `.pf-m-no-padding` is applied on the parent `<tr>` and the `div.pf-c-table__expandable-row-content` element. It shouldn't be on those, and should be on the parent of `div.pf-c-table__expandable-row-content` instead. Also, it's an optional class, so there should probably be a prop for it. If you just have text content, you want the padding. If you have a nested table or want to create your own layout in the expanded content, you may want to remove the padding.

However, even with those changes (mocking them up in the browser dev tools), I'm not seeing the padding take effect. I could be doing something wrong, but can we make sure the CSS from core is in?

And FWIW, you can see examples of the use of `.pf-m-no-padding` here - https://pf4.patternfly.org/components/Table/examples/

redallen:
Ah, that's because we need react-docs/package.json to bump @patternfly/patternfly to 2.0.0 as well. I don't see it in this PR.

redallen:
For future reference, see https://github.com/tlabaj/patternfly-react/pulls?q=is%3Apr+is%3Aclosed for the list of breaking changes in here.

mcoker:
@redallen I'm still not seeing some CSS I would expect to see in core 2.0.0. Looks like @tlabaj updated react-docs/package.json. Do you know what the issue might be?

redallen:
Perhaps this should be in the other order in `gatsby-node.js`?
```
import '@patternfly/patternfly/patternfly.css';
import './static/base.css';
```

Not exactly sure what base.css is, I just copied it over from the old docs.

mcoker:
thanks to @dgutride, cleared my cache and looks like I can see all of the changes.

* In dropdown, the separator should be on the `<li>`, not a `<div>` in the `<li>`.
  * Should be `<li class="pf-c-dropdown__separator" role="separator"></li>` instead of `<li role="none"><div href="#" role="separator" class="pf-c-dropdown__separator"></div></li>`
* In the expandable tables, `.pf-m-no-padding` is on the wrong elements.
  * Instead of on `.pf-c-table__expandable-row` and `.pf-c-table__expandable-row-content`, `.pf-m-no-padding` should be on

dlabaj:
@mcoker For the drop down is this a new change for this release?  We didn't see it in the issues for this release for what we have for react, and it doesn't look like that file was changed in this PR.  Just wondering if we missed something.

dlabaj:
@mcoker I removed the no-padding from the table I will do that in a separate PR.  I did add the border fix for #1674, can you verify it.  After that can we just merge this and fix the no padding and the dropdown in a separate PR?

mcoker:
@dlabaj so sorry, I was looking at dropdown and meant to look at nav, and nav looks good. Re: #1674, looks like the functionality is there but the prop might be reversed. In your example you set `borders={false}` to enable the borders.

First example has no `borders` prop. Second has `borders={false}`.

Screen Shot 2019-04-10 at 8 10 19 AM

I'm fine merging at any point and following up in another PR if you want to get this one in.

PR #1716, opened 2019-04-05
chore(title): headingLevel => level and make it required

seanforyou23:
**What**: This PR renames the prop `headingLevel` to simply `level` for the Title component. It also makes this a required prop. It should be required because otherwise, users are inclined to use the default value of h1 which leads to applications with IA that misrepresent the semantics of the content.

Fixes https://github.com/patternfly/patternfly-react/issues/1439

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Title/Title.tsx
+++ b/packages/patternfly-4/react-core/src/components/Title/Title.tsx
@@ -12,14 +12,14 @@ export interface TitleProps extends Omit<React.HTMLProps<HTMLHeadingElement>, 's
   /** Additional classes added to the Title */
   className?: string;
   /** the heading level to use */
-  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  level: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 }
 
 const Title: React.FunctionComponent<TitleProps> = ({
   size,
   className = '',
   children = '',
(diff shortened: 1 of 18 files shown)


packages/patternfly-4/react-core/src/components/Title/Title.tsx
```diff
@@ -12,14 +12,14 @@ export interface TitleProps extends Omit<React.HTMLProps<HTMLHeadingElement>, 's
   /** Additional classes added to the Title */
   className?: string;
   /** the heading level to use */
-  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  level: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
jschuler: Should it be required if we anyways default it to h1?

seanforyou23: It should be required, that's really the primary change for this PR. I wasn't able to remove 'h1' as a default value though, I don't understand why. When I run the `build:docs` target with that in place, it blows up like this;

Screen Shot 2019-04-09 at 3 36 45 PM

I spent a while trying to debug it but didn't wind up with anything useful. In any case, the prop is required so we have the expected experience on the consuming side and the default value seems to have no effect other than silencing the build failure. A code smell, I know.

WDYT? If you have an idea on how to fix this I'd love to hear. The only way I know to get around it is using the `createElement` function but that's been advised against so I'm open to suggestion.

Issue #1485, opened 2019-03-04
Add support for groups in dropdowns


mcoker:
https://github.com/patternfly/patternfly-next/pull/1444

https://pf4.patternfly.org/components/Dropdown/examples/

> .pf-c-dropdown__group | `<section>` | Defines a group of items in a dropdown. Required when there is more than one group in a dropdown

> .pf-c-dropdown__group-title | `<h1>` | Defines the title for a group of items in the dropdown menu

`section.pf-c-dropdown__group` should wrap all groups of items if there is at least 2 groups in the dropdown menu.

A person can change the `<h1>` to another level heading, but it's worth noting it's OK to use an `<h1>` there anywhere in the page since it's in a `<section>`, which allows for it's own heading hierarchy.

rachael-phillips:
Hi @mcoker is this needed for Summit? If not, we may make this lower priority.

mcoker:
@rachael-phillips I don't believe anyone asked for it. It was a result of talking to @mcarrano as to how to add groups of items to a dropdown when other dropdowns collapse into a single dropdown. AFAIK it can wait.

rachael-phillips:
Thank you so much for the update :)

mcarrano:
What I will say about this is that if we don't do it, this leaves us exposed as to how we collapse utility items into the kabob when the viewport shrinks to Mobile. Do you agree @mcoker ?  So I feel like this is less of a new feature request and more like correcting for a design oversight/bug.

mcoker:
Yep, agree with that statement completely.

rachael-phillips:
Hi @kmcfaul can we close this issue?

PR #1725, opened 2019-04-08
feat(Dropdown): add dropdown groups option

kmcfaul:
**What**: Adds dropdown groups option and example

**Refer to issue**: #1485


mcoker:
Do you know where `aria-hidden="true"` on the `h1.pf-c-dropdown__group-title` element came from? We don't have that in core and I'm pretty sure you want the group titles to be read by screen readers.

kmcfaul:
I implemented these groups after the ones in checkbox so it's possible I carried over that property. @jgiardino Should the hidden tag be removed for dropdown group titles?

christiemolloy:
Looks like this should say Group 1 but I cant see it

Screen Shot 2019-05-07 at 3 48 05 PM

kmcfaul:
For the example, I was matching the one in core which shows the Group label can be left out.

Issue #1724, opened 2019-04-08
PF4 docs


jschuler:
- [ ] Fix font size (html should be set to 16px so that rem units can pick it up)
- [ ] Font family should be Overpass everywhere
- [x] Some components are not rendering (LoginPage ((fixed via https://github.com/patternfly/patternfly-react/pull/1696)), Background)
- [x] Layouts should be in their own nav category (#1753)
- [ ] Layouts don't have visual borders anymore
- [ ] Dark theme toggle button is missing
- [x] Fix hot reloading on code changes
- [ ] Fix color scheme.  It is hard to read the grey on grey. (Background should be white)

Additional:
- [x] Fix per-example styles
- [x] Fix empty state centering (#1753)
- [ ] Simple context selector text too large
- [ ] Make react-live code block readable

mcoker:
Wrapping the examples in `<pre>` tags seems to be problematic. `<pre>` applies a lot of font and white-space changes we don't want in the examples. That should also help address some of the font size issues.

jgiardino:
This may or may not be related to this issue. When this issue is fixed, can we check the following, and capture this as an issue if it's unrelated?

I'm seeing a difference in font size for the Alert description:
![image](https://user-images.githubusercontent.com/21063328/55826807-ccfacb00-5ad6-11e9-95f3-476514408c81.png)

I only see this issue in Chrome, and it looks like Chrome is pulling the styles in twice.

dlabaj:
Can this be broken out into separate smaller issues.  This seems like a lot for one PR.  Seems like each of the boxes should be their own issue.

redallen:
My plan is to split up the work over multiple PRs and just reference a few checkboxes on this issue each time.

redallen:
CSS mock for example padding
<img width="1642" alt="Screen Shot 2019-04-12 at 1 10 01 PM" src="https://user-images.githubusercontent.com/47335686/56057654-5e637a80-5d2d-11e9-98d2-d8b4ab800bcd.png">

redallen:
Closing in favor of https://github.com/patternfly/patternfly-org/issues/1208 which will unify docs across repos.

PR #1656, opened 2019-03-28
feat(gatsby): Refactor Gatsby

redallen:
**What**: Rewrite of React-docs. Still using Gatsby so v2 website development can continue.
A few key points for doc writers:

1. Docs now use `*.md` files transformed by `gatsby-transformer-remark`. This encourages writing more documentation and allows for a single source of truth of where doc pages come from (no having to import and link `examples/*.js` files to our custom `*.docs.js` files).
2. Using class fields in examples is no longer supported until [this PR](https://github.com/bublejs/buble/pull/169) is merged. On the bright side, we probably should not assume consumers are using proposal language features like [class properties](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties) and I've already converted all the examples in the meantime.

Key points for contributors:

1. Doc build times take [2 minutes](https://circleci.com/gh/patternfly/patternfly-react/823) for previews (down from ~[25 minutes](https://circleci.com/gh/patternfly/patternfly-react/888))!!! Hopefully this makes them less flaky too.
2. Time to first load for `yarn start:pf4` slightly decreased, and with clearer warning messages as to why resources or property tables might not be loading.

Technical changes:

1. Use custom plugin `react-docgen-typescript` to replace build-time task `gen-ts-docs` to allow hot reloading.
2. Ignore unused source files.
3. Better manage example resource injection (for `examples/*.(js|svg)` files as well as entire modules). This is the hardest part of the docs to understand and is handled in `helpers/dynamicImports.js`. No more components are injected than need to be to keep page sizes smaller.
4. Treat JS and TS the same when it comes to docs.
5. Rewrite all doc components to consume new GraphQL data flow. Data is structured completely differently.
6. Using `@patternfly/react` components in our docs is done safely by using `"@patternfly/react-core": "2.9.2"` in our package.json, but importing doing live code example imports via `require("../../../react-core/dist/js")`. This means if you break a component used in the docs (ie. Title) the whole doc website will not break but the `components/title` page examples will break.

**Additional issues**: ~Still a a day or two of features to tackle:~

~- [ ] Exclude inherited TS props~ Future work.
- [x] Axe
- [x] Nav menu on mobile
- [x] CoreJS warning
- [x] Live edit styling + dropdown
- [x] Background component breaks nav
- [x] Style so it looks decent. I can't CSS for my life.
- [x] Document how to write a *.md file.

PR #1731, opened 2019-04-09
fix(aria): make aria-* component props consistent

### LINKED ISSUE #1529: Fix accessibility prop name inconsistencies

redallen:
**What**: Resolves #1529. Resolves jest console errors related to aria props.

**Additional issues**: Check snapshots to make sure non-custom elements are not changed.


packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.test.js
```diff
@@ -48,11 +48,11 @@ test('modal does not call onClose for esc key if it is not open', () => {
   expect(props.onClose).not.toBeCalled();
 });
 
-test('Each modal is given new ariaDescribedById and ariaLablledbyId', () => {
+test('Each modal is given new aria-describedby and aria-lablledby', () => {
```
jgiardino: tiny typo here: `aria-lablledby` should be `aria-labelledby`

redallen: Good catch. Fixed in latest commit.

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.js
```diff
@@ -20,11 +20,11 @@ const defaultProps = {
   onClose: () => undefined
 };
 
-const AlertActionCloseButton = ({ className, onClose, 'aria-label': ariaLabel, title, variantLabel, ...props }) => (
+const AlertActionCloseButton = ({ className, onClose, variantLabel, 'aria-label': ariaLabel, title, variant, ...props }) => (
```
jgiardino: I see an issue with the aria-label that's coming through. I'm not sure if it's because of this line or somewhere else, but... 

When I compare the html for the Close button,
this is what I see in [the first alert example in the PR preview](https://1731-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/alert/):
```
<button aria-label="Close undefined alert: Info alert title" class="pf-c-button pf-m-plain" type="button">
```
and this is [what is there now](https://patternfly-react.surge.sh/patternfly-4/components/alert):
```
<button aria-label="Close Success alert: Success alert title" class="pf-c-button pf-m-plain" type="button">
```

Notice that the aria-label includes "undefined alert" whereas the current build has "Success alert".

jgiardino:
I'm seeing a difference in font size for the Alert description:
image

I only see this issue in Chrome, and it looks like Chrome is pulling the styles in twice. 

cc'ing @mcoker on this one

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.js
redallen: So is it good now? I (think) I just fixed it in the latest commit that's preview finished a minute ago!

redallen:
Yeah, the styles are generally wrong on the new doc site. Please add that issue (and any others you find) to #1724 !

packages/patternfly-4/react-core/src/components/DataList/DataListContent.js
```diff
@@ -30,10 +18,25 @@ DataListContent.propTypes = {
...
   id: '',
   isHidden: false
 };
 
+const DataListContent = ({ className, children, id, isHidden, 'aria-label': ariaLabel, rowid, ...props }) => (
+  <section
+    id={id}
+    className={css(styles.dataListExpandableContent, className)}
+    hidden={isHidden}
+    aria-label={ariaLabel}
```
jgiardino: This might be out of scope for this PR and belong in a separate issue...

The props documentation states:

Name | Type | Required | Default | Description
-- | -- | -- | -- | --
aria-label | string |   |   | Adds accessible text to the DataList toggle


But the description seems to describe the DataListToggle component. In this context, it's providing an accessible label for the expandable content section.

And when looking at the DataListToggle component the following are documented:

Name | Type | Required | Default | Description
-- | -- | -- | -- | --
aria-labelledby | string |   | '' | Adds accessible text to the DataList toggle
aria-label | string |   | 'Details' | Adds accessible text to the DataList toggle

But only `aria-label` is accepted as a prop. If I provide a value for `aria-labelledby` on the component, that value doesn't display in the html. 

Currently, `aria-labelledby` is defined on the html element only if the consumer doesn't define `aria-label`. If `aria-label` is defined, then the `aria-labelledby` attribute is removed from the html. This is as designed (e.g. either the consumer uses our default label, which we generate using `aria-labelledby`, or they explicitly define the full text string using `aria-label`).

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
```diff
@@ -42,9 +41,9 @@ const propTypes = {
   /** Function that handles the onChange event for the Remember Me Checkbox */
   onChangeRememberMe: PropTypes.func,
   /** Aria Label for the Remember me checkbox */
-  rememberMeAriaLabel: props => {
-    if (props.rememberMeLabel && !props.rememberMeAriaLabel) {
-      return new Error('rememberMeAriaLabel is required with the Remember me checkbox');
+  'aria-label': props => {
```
jgiardino: This one might also be a separate issue, but this attribute should be removed from this checkbox. In this case, the aria-label is overwriting the text defined in the `<label>`.

Let me know if an issue is needed for this.

jgiardino: Also, there's an issue in the preview. This error is displaying: 
> SyntaxError: Invalid or unexpected token

I'm not sure if this is what's causing the issue, but I see that this prop is still being defined on LoginForm even though you changed it to `aria-label`:
`rememberMeAriaLabel="Remember me Checkbox"`

packages/patternfly-4/react-core/src/components/DataList/DataListContent.js
redallen: You are correct, it's ignored. The old logic is:

`aria-labelledby={ariaLabel !== "Details" ? null : `${rowid} ${id}`}`

I'm changing it to
```
if (ariaLabel === "Details") {
    ariaLabelledBy = `${rowid} ${id}`;
}
```
since it probably wasn't intended for the property to do nothing.

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.js
redallen: Why should the attribute be removed? This is starting to get outside the scope of this PR, so a separate issue would be nice.

jgiardino: 👍 I created https://github.com/patternfly/patternfly-react/issues/1737 for my first comment. 

But the second comment might still be relevant to this PR?

redallen: The login page has been broken since #1656 [(see preview).](https://1656-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/loginpage/) Fixing it is a part of #1724 which is next on my list.

packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -106,7 +104,7 @@ class Modal extends React.Component {
       this.container = document.createElement('div');
     }
 
-    return ReactDOM.createPortal(<ModalContent {...props} title={this.props.title} id={this.id} ariaDescribedById={this.props.ariaDescribedById}/>, this.container);
+    return ReactDOM.createPortal(<ModalContent {...this.props} title={this.props.title} id={this.id} />, this.container);
```
jgiardino: There's definitely something weird happening with `aria-describedby`.

I'd suggest looking at how this prop is used in the [current example labeled **No Header Modal**](https://patternfly-react.surge.sh/patternfly-4/components/modal). The idea is that if `aria-describedby` is not defined, it will default to referencing the `id` defined for `.pf-c-modal-box__body`.

But if the consumer needs to control what contents are used for the description, the consumer can specify an `id` for `aria-describedby` and define that `id` on some chunk of contents anywhere inside `pf-c-modal-box__body`. This is what's being illustrated with  **No Header Modal**.

Let me know if you need more details on this one.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelect.js
```diff
@@ -17,6 +17,10 @@ const propTypes = {
   isGrouped: PropTypes.bool,
   /** Currently checked options */
   checked: PropTypes.arrayOf(PropTypes.string),
+  /** Aria label for fieldset */
+  'aria-label': PropTypes.string,
+  /** Aria labelled by for fieldset */
+  'aria-labelledby': PropTypes.string,
```
jgiardino: The old prop `ariaLabelledBy` is still being used in the example.

jgiardino: Oh, and it looks like that's why `arialabelledby` might be getting defined on `fieldset.pf-c-form__fieldset` which shouldn't be there.

redallen: You're killing it! Fixed in latest commit.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectInput.js
```diff
@@ -58,7 +58,7 @@ class CheckboxSelectInput extends React.Component {
           selections={selected}
           isExpanded={isExpanded}
           placeholderText="Filter by status"
-          ariaLabelledBy={titleId}
+          aria-labelledby={titleId}
```
jgiardino: Same issue as previous comment... The old prop ariaLabelledBy is still being used in the example.

packages/patternfly-4/react-core/src/components/Select/GroupedCheckboxSelectInput.js
```diff
@@ -65,7 +65,7 @@ class GroupedCheckboxSelectInput extends React.Component {
           selections={selected}
           isExpanded={isExpanded}
           placeholderText="Filter by status"
-          ariaLabelledBy={titleId}
+          aria-labelledby={titleId}
```
jgiardino: Same issue as previous comment... The old prop ariaLabelledBy is still being used in the example.

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -75,7 +75,7 @@ class Select extends React.Component {
       isExpanded,
       isGrouped,
       selections,
-      ariaLabelledBy,
+      'aria-labelledby': ariaLabelledBy,
```
jgiardino: Same issue as previous comment... The old prop ariaLabelledBy is still being used in the example.

packages/patternfly-4/react-core/src/components/Modal/Modal.js
redallen: This was also a matter of updating the docs. Fixed in latest commit.

redallen:
Let me know if there are still any problem with the Modal or Select, but I think updating them in the docs fixed them! The preview just now redeployed with the changes.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -29,13 +29,11 @@ const propTypes = {
   /** enables Secondary Tab styling */
   isSecondary: PropTypes.bool,
   /** aria-label for the left Scroll Button */
-  leftScrollAriaLabel: PropTypes.string,
+  'aria-label-left-scroll': PropTypes.string,
   /** aria-label for the right Scroll Button */
-  rightScrollAriaLabel: PropTypes.string,
+  'aria-label-right-scroll': PropTypes.string,
```
jgiardino: I like this approach for naming. I'd like to suggest we do this for some of the other components that currently only have one aria-label prop. For example, in the ContextSelector component, we have a search button that gets labeled with the `aria-label` prop. In this case, I'd suggest we use `aria-label-search`. This would let us later update the component to include other things that would use the attribute, and avoid potential confusion by having one prop as `aria-label` and another as `aria-label-[descriptive text]`.

I do think there are cases where having just `aria-label` is fine for the most granular components, like the Button component, where that's the one and only text string we would define. Or where the string defined in `aria-label` is applied to the parent element of a more complex component (like a fancy modal). But other components like Chip or ContextSelector, where it's applied to something buried inside the component, I'd suggest we make the prop name more specific.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -45,22 +43,33 @@ const defaultProps = {
   id: null,
   isFilled: false,
   isSecondary: false,
-  leftScrollAriaLabel: 'Scroll left',
-  rightScrollAriaLabel: 'Scroll Right',
-  'aria-label': null,
+  'aria-label-left-scroll': 'Scroll left',
+  'aria-label-right-scroll': 'Scroll Right',
```
jgiardino: When I define `aria-label-left-scroll` as a prop on the Tabs component, I'm not seeing that string included on the button.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -45,22 +43,33 @@ const defaultProps = {
...
   variant: TabsVariant.div
 };
 
-const Tabs = ({ 'aria-label': ariaLabel, id, variant, ...props }) => {
+const Tabs = ({ id, variant, ...props }) => {
   const unique_id = id ? id : getUniqueId();
   return variant === TabsVariant.nav
-    ? <TabsWithNav ariaLabel={ariaLabel} id={unique_id} {...props} />
+    ? <TabsWithNav id={unique_id} {...props} />
```
jgiardino: Currently there's an example labeled **Accessible primary tabs with aria-label** that shows the `<nav>` variant with `aria-label`. I'm not seeing this example in the preview.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
thesmartwon: I made all of the aria props that aren't just passed to the parent <div> longer names.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
thesmartwon: Yep, fixed in latest commit. TabContent is still a bit whacky.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
thesmartwon: See previous comment.

packages/patternfly-4/react-core/src/components/DataList/DataListContent.js
jgiardino: Thanks! These updates work as expected except for the following...

When I define `aria-label` on the `DataListToggle`, in the PR preview I will see `aria-labelledby` included without a value.

In the current implementation, `aria-labelledby` is not included on the element at all. This is the expected behavior.

As for removing `aria-labelledby` from the documentation for `DataListToggle`, should I create a separate issue for that?

jgiardino:
The issues with Modal and Tabs are resolved now! 🎉

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
jgiardino: Related to the discussion in slack, this should change to `ariaLabelRightScroll` (and same for the left scroll)

packages/patternfly-4/react-core/src/components/Alert/AlertBody.js
```diff
@@ -10,7 +10,7 @@ const propTypes = {
   children: PropTypes.node,
   className: PropTypes.string,
   onClose: PropTypes.func,
-  'aria-label': PropTypes.string
+  'aria-label-button': PropTypes.string
```
jgiardino: Related to discussion in slack, let's use `ariaLabelClose` instead.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
```diff
@@ -25,7 +25,7 @@ export const propTypes = {
   /** Indicates where menu will be alligned horizontally */
   position: PropTypes.oneOf(Object.values(ApplicationLauncherPosition)),
   /** Adds accessible text to the button. Required for plain buttons */
-  'aria-label': PropTypes.string
+  'aria-label-button': PropTypes.string
```
jgiardino: Let's use `ariaLabelToggle` instead.

packages/patternfly-4/react-core/src/components/ChipGroup/Chip.js
```diff
@@ -11,7 +11,7 @@ const propTypes = {
   /** Content rendered inside the chip text */
   children: PropTypes.node,
   /** Aria Label for close button */
-  'aria-label': PropTypes.string,
+  'aria-label-button': PropTypes.string,
```
jgiardino: Let's use `ariaLabelClose` instead.

packages/patternfly-4/react-core/src/components/ContextSelector/ContextSelector.js
```diff
@@ -32,7 +32,7 @@ const propTypes = {
   /** Text that appears in the Context Selector Toggle */
   toggleText: PropTypes.string,
   /** aria-label for the Context Selector Search Button */
-  'aria-label': PropTypes.string,
+  'aria-label-search': PropTypes.string,
```
jgiardino: Let's use `ariaLabelSearch` instead.

packages/patternfly-4/react-core/src/components/Page/PageHeader.js
```diff
@@ -24,8 +24,8 @@ const propTypes = {
   onNavToggle: PropTypes.func,
   /** True if the side nav is shown  */
   isNavOpen: PropTypes.bool,
-  /** Callback function to handle the side nav toggle button */
-  'aria-label': PropTypes.string,
+  /** aria-label for the button */
+  'aria-label-button': PropTypes.string,
```
jgiardino: Let's use `ariaLabelNavToggle` instead.

jgiardino:
Rather than try to find these inline and be wrong, here are a few more prop names to change based on the slack discussion.

- Select Properties:
  - `aria-label` > `ariaLabelPanel`
  - `aria-labelledby` > `ariaLabelledbyToggle`
- CheckboxSelect Properties: 
  - `aria-label` > `ariaLabelPanel`
    - I wasn't sure if this should be `ariaLabelFieldset` instead, but since it's nearly identical in behavior with the first bullet, I'm thinking we can just be consistent with that.

redallen:
With the Typescript conversion, this PR is in a terrible state. The issue still exists and can be resolved in a breaking change release later.

Issue #1675, opened 2019-04-01
Add support for <hr> in content component


mcoker:
https://github.com/patternfly/patternfly-next/pull/1611

tlabaj:
Closing this since no code changes are needed.
PR #1703 updated the version of core that pulls this in.

PR #1743, opened 2019-04-10
Text rule

jschuler:
Add TextRule component

Closes https://github.com/patternfly/patternfly-react/issues/1675


jschuler:
Closing since consumers can do `<TextContent><hr /></TextContent>`. Going forward might even be better to remove the other sub components since they don’t make sense anymore as the css modifications that were made in the TextContent component to target elements for specifically does not work anymore. We can assume that anything placed within the <TextContent> component will receive the content css styling.

Issue #1674, opened 2019-04-01
Add no border rows option to compact table


mcoker:
We added borders to the compact table, and a modifier to disable them

https://github.com/patternfly/patternfly-next/pull/1541

This is also related to https://github.com/patternfly/patternfly-next/pull/1563/ (react issue for that is https://github.com/patternfly/patternfly-react/issues/1679), which says the "no border rows" modifier can't be used with an expandable compact table.

tlabaj:
Closed by #1745

PR #1745, opened 2019-04-10
Updated to border prop for compact table.

dlabaj:
Fixed border prop to show border on true and not show border on false.  #1674

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
@@ -84,6 +84,7 @@ export interface TableProps extends Omit<Omit<HTMLProps<HTMLTableElement>, 'onSe
   children?: ReactNode;
   className?: string;
   variant?: OneOf<typeof TableVariant, keyof typeof TableVariant>;
+  border?: boolean;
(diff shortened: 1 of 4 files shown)


christiemolloy:
Can you switch the titles for the Compact Table so that the one with borders is default, and the one without borders says "Compact Table Borderless Rows"

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -84,6 +84,7 @@ export interface TableProps extends Omit<Omit<HTMLProps<HTMLTableElement>, 'onSe
   children?: ReactNode;
   className?: string;
   variant?: OneOf<typeof TableVariant, keyof typeof TableVariant>;
+  border?: boolean;
```
jschuler: borders

dlabaj: all set.

mcoker:
Noticed one small thing in the prop description from the previous PR. "compacat" is misspelled and it might be better reworded. Maybe something like this?

> Indicates if borders between rows are visible on a compact table. Note that this can not be applied when using an expandable compact table.

Issue #1678, opened 2019-04-01
Add pf-m-no-padding prop to expandable table row


mcoker:
https://github.com/patternfly/patternfly-next/pull/1598

> For no padding add `.pf-m-no-padding` to `.pf-c-table__expandable-row` > `<td>`

cc @christiemolloy

PR #1748, opened 2019-04-10
Updated collapsable table demo for issue #1678

dlabaj:
Added no padding example to collapsable table demo.  Fix for issue #1678


christiemolloy:
@dlabaj "child-1" should be in the first td column - like the "Compact Expandable Table"

PR #1749, opened 2019-04-11
fix(grid): add missing props in d.ts

### LINKED ISSUE #1718: Typescript version of Grid component is missing props
Props (including sm, md, lg, xl) that are defined in the spec and in grid.js are missing from the grid.d.ts file

boaz0:
**What**:

fixes #1718 

Added the span, sm, md, lg and xl props in the d.ts file


boaz0:
Thanks you @tlabaj and @redallen

Issue #1695, opened 2019-04-03
In the Wizard, add the ability to hide the nav.


vbusch:
We have a use case where we want to hide the left side nav for a step.  And when the nav is shown, that step shouldn't appear in it.

Here is the use case:
- Step 1 and Step 2 are entering data into forms.
- The third step is for reviewing and has the finish button.
- The fourth step is an 'in progress... status screen'.  It shouldn't show the navigation items.
- The fifth step is the final status.  .  It shouldn't show the navigation items.

When the left nav is shown (Steps 1-3), the navigation items would only contain:
- Step 1
- Step 2
- Step 3
When on the last 2 steps (step 4 & step 5), the left side nav would be hidden.

amysueg:
@mcarrano For the Wizard design guidelines to take into consideration.

mcarrano:
Yes, this is in fact the way it is intended to work. There is an open issue here for cleaning up the current React implementation: https://github.com/patternfly/patternfly-react/issues/1618. @jschuler please take note. There is an InVision deck here that shows the intended flow: https://redhat.invisionapp.com/share/W3QA5O64XNQ

dlabaj:
@tlabaj Do you think we can get this into the next sprint?

dgutride:
Closing via #1754

Issue #1637, opened 2019-03-25
PF4 Wizzard Export all parts of the component


rvsia:
For our project we need PF4 Wizard parts (header, body, nav, ...) separately. Would it possible to export them as a standalone components?

cc @Hyperkid123

Hyperkid123:
Yep this would be nice. Sometimes we need more control over the steps progression (branching, history, etc.) and this would allow us to completely customize whole wizard.

I do realize it could be done even without all the components but i would like to avoid some hacks.

rachael-phillips:
@jschuler fyi.

jschuler:
All parts of the wizard are now exported

Issue #1630, opened 2019-03-22
PF4 Wizard doesn't allow to control the steps from outside


idevat:
The wizard component encapsulates the workflow of steps. Consider the following (simplified) scenario:

A user wants to add a cluster to cluster management.
1. The user opens a wizard (at step 1).
2. The user enter name of one cluster node and click on button "Next".
3. A check request (whether the back-end is authenticated against chosen node) is sent. (Suppose the answer is not authenticated.)
4. Wizard goes to step 2. It shows form with password (only this field to keep it simple for now).
5. An authentication request is sent. (Suppose password is wrong).
6. At this point, I would expect the wizard to stay at step 2 and the password field is shown again (with an explanation). But I was not able to achieve this behavior.

So, in code it looks like this (in principle, simplified):
```javascript
const Test = () => {
  const [password, setPassword] = React.useState("");
  const [step, setStep] = React.useState(1);
  const [wrongPasswordUsed, setWrongPasswordUsed] = React.useState(false);

  const onNext = () => {
    if (step !== 2 || password === "pw") {
      setStep(step + 1);
      return;
    }
    console.log("I really need to keep step 2. But how?");
    setWrongPasswordUsed(true);
  };

  const steps = [
    {
      name: "Enter node",
      component: "User enter node here...",
    },
    {
      name: "Node authentication",
      component: (
        <React.Fragment>
          Password:
          <input
            value={password}
            onChange={e => setPassword(e.target.value)}
          />
          {wrongPasswordUsed && <p>Wrong password, try again</p>}
        </React.Fragment>
      ),
    },
    {
      name: "Add cluster",
    },
  ];
  return (
    <Wizard
      data-role="add-cluster-wizard"
      isOpen
      startAtStep={step}
      title="Test"
      steps={steps}
      onNext={onNext}
    />
  );
};

```
- The main question here is how to keep step 2 in the case of wrong password.
- I tried to set step with `startAtStep` attribute, but it is really only for start.
- I used my own state `step` to identify my situation. I'm aware that using `onNext` is not sufficient. It is also needed to use at least `onBack` and `onGoToStep` to track current step. It is really uncomfortable.
- If I need to change the button `Next` label (e.g. `Check authentication` in step 1, `Authenticate node` in step 2) I would have to track current step as well. 

Maybe I'm wrong and I misunderstand the patternfly wizard intention. Are there any guidelines for wizard mechanisms  somewhere? I mean something explaining how to implement various scenarios?

[...]

Issue #1618, opened 2019-03-20
Refinements for Wizard Component and Demo


mcarrano:
Just doing a design review of the [wizard component](http://patternfly-react.surge.sh/patternfly-4/components/wizard) as recently merged and have some issues that we would like to see addressed:

* Sidebar styling does not seem right. Background fills for disabled states are too light and difficult to tell visually what steps are disabled. 

* Buttons should always be left aligned per designs. "Validation Wizard" has incorrect button placement.

See core implementation for correct button placement and sidebar styling: https://pf4.patternfly.org/components/Wizard/examples-full/?component=Wizard

For the React demo, it would be good to show how this wizard can be used to progressively disclose steps and have sample screens to indicate progress and completion. This flow is illustrated in the design storyboard here: https://redhat.invisionapp.com/share/SZPQWBQY2KH

mcarrano:
@rachael-phillips I entered a new issue for the React wizard as this got merged before we had an opportunity to do a UX Review. There are some visual issues and also a request for a more robust demo. I think we can prioritize if needed. @kybaker please feel free to add further details about visual issues that you see.

jschuler:
Hi @mcarrano , the first issue needs to be addressed in core @mcoker , the second issue is a simple configuration option that can be passed into the wizard. I can update it to default to left alignment.
I agree it would be great to have a more complex demo. Let's add one this sprint.

mcarrano:
Thanks @jschuler . I will try to pull together a meeting with you, me, @mcoker , and @kybaker to review the sidebar styling and decide what needs to be fixed for release. I think that the demo is less critical and can be added later. We should discuss how to proceed with that, however, and what is needed (if anything) from a CSS perspective or if we only need a demo in React.

mcoker:
> the second issue is a simple configuration option that can be passed into the wizard

@mcarrano per our discussions about button placement, should we remove the option of right-aligning the wizard action buttons? Currently they're left aligned by default, but you can right align them with `.pf-m-align-right`.

mcarrano:
@jschuler @mcoker @kybaker and I met today to review the current implementation. A separate issue has been opened here to track styling changes that are needed in the CSS: https://github.com/patternfly/patternfly-next/issues/1616.

[...]

PR #1754, opened 2019-04-11
PF4 Wizard: Controlled wizard (custom footer)

jschuler:
- Update with latest core changes
- Can create custom footer to use for the wizard which gives full control. Import the `WizardFooter` and pass it into the Wizard as a `footer` prop. Additionally import `WizardContextConsumer` to give the footer additional functionality. Example:
```
const CustomFooter = (
      <WizardFooter>
        <WizardContextConsumer>
          {({ activeStep, goToStepByName, goToStepById, onNext, onBack, onClose }) => {
            return (
              <>
                <Button variant="primary" type="submit" onClick={() => onNext())}>
                  {activeStep.name === 'Review' ? 'Finish' : 'Next'}
                </Button>
                <Button variant="secondary" onClick={() => onBack()} className={activeStep.name === 'Get Started' ? 'pf-m-disabled' : ''}>
                  Back
                </Button>
                <Button variant="link" onClick={onClose}>
                  Cancel
                </Button>
              </>
            )}}
        </WizardContextConsumer>
      </WizardFooter>
    );
```
- In the wizard steps configuration, can use `isFinishedStep` on the step object to hide the nav and footer from the step
- In the wizard steps configuration, can use `canJumpTo` on the step object to enable or disable the ability to directly click on a nav item. Otherwise by default the nav item is clickable.
- Use `nextButtonText` in the step configuration to change the name of the Next button (for example to Close or Finish). Overrides the `nextButtonText` passed into the Wizard
- Use `isCompact` prop for the Wizard to make the nav smaller

Closes: https://github.com/patternfly/patternfly-react/issues/1796, https://github.com/patternfly/patternfly-react/issues/1695, https://github.com/patternfly/patternfly-react/issues/1618, https://github.com/patternfly/patternfly-react/issues/1630, https://github.com/patternfly/patternfly-react/issues/1637


packages/patternfly-4/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -2,33 +2,37 @@ import * as React from 'react';
...
 }: WizardNavItemProps) => (
     <li className={css(styles.wizardNavItem)}>
       <a
-        aria-current={current && !hasChildren ? 'page' : 'false'}
-        onClick={() => onNavItemClick(step)}
-        className={css(styles.wizardNavLink, current && 'pf-m-current', disabled && 'pf-m-disabled')}
-        aria-disabled={disabled ? 'true' : 'false'}
-        tabIndex={disabled ? -1 : undefined}>
-        {label}
+        aria-current={isCurrent && !children ? 'page' : 'false'}
```
dlabaj: 'false' is that suppose to be a string?

redallen: Joachim's on PTO, so I'll respond. They're equivalent actually, React will transform the boolean `false` to the string `"false"`. I think it's slightly cleaner as bools, so I'll change it. Have to push to update snapshots anyway :)

christiemolloy:
In this case it looks like I can't press Next to advance, but i can press the link on the left-side, maybe I missed something but I thought this shouldn't be an option?

Screen Shot 2019-04-24 at 4 25 25 PM

christiemolloy:
For a better user experience I think that on Mobile view, when you click on a new step in the toggle dropdown, the dropdown should disappear so that it reveals the content. Otherwise the content is hidden and the user has to work out how to hide the toggle menu.
Screen Shot 2019-04-24 at 4 27 54 PM

mcoker:
I might change `isCompact` to `isCompactNav`? In the future if we have a compact version of the wizard as a component, `isCompact` should be reserved for that. That's also why I named the modifier `.pf-m-compact-nav` instead of `.pf-m-compact`, since all it does is modify the nav.

christiemolloy:
Looks like if you have too many characters the text isnt wrapping @mcoker 
Screen Shot 2019-04-24 at 4 33 38 PM

mcoker:
@christiemolloy https://github.com/patternfly/patternfly-next/pull/1761

redallen:
To summarize:
- [x] Move alert to top of form
- [ ] Fix `canJumpTo` regression from [initSteps](https://github.com/patternfly/patternfly-react/pull/1754/files#diff-1c20dfabfea0b9f196ef069a34376374L223) for validation wizard
- [ ] isCompact -> isCompactNav
- [ ] Hide dropdown on step click on mobile

I can do these tomorrow if Joachim doesn't tonight.

packages/patternfly-4/react-core/src/components/Wizard/WizardFooter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  onClose,
+  isValid,
+  firstStep,
+  activeStep,
+  nextButtonText,
+  backButtonText,
+  cancelButtonText
+}) => (
+    <footer className={css(styles.wizardFooter)}>
+      <Button variant="primary" type="submit" onClick={onNext} isDisabled={!isValid}>
```
tlabaj: can we use the variant constant here?

packages/patternfly-4/react-core/src/components/Wizard/WizardHeader.tsx
```diff
@@ -5,25 +5,31 @@ import { Button } from '../Button';
...
 import { TimesIcon } from '@patternfly/react-icons';
 
-interface WizardHeaderProps {
+export interface WizardHeaderProps {
+  /** Callback function called when the X (Close) button is clicked */
   onClose?: () => void;
+  /** Title of the wizard */
   title: string;
-  description: string;
-  ariaLabel: string;
```
tlabaj: this would be a breaking change for the consumers already using the wizard.  @dgutride, we are not taking in breaking changes right now correct?

packages/patternfly-4/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -2,33 +2,37 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/patternfly/components/Wizard/wizard.css';
 
-interface WizardNavItemProps {
+export interface WizardNavItemProps {
+  /** Can nest a WizardNav component for substeps */
   children?: React.ReactNode;
-  label?: string;
```
tlabaj: same comment about breaking changes here.

packages/patternfly-4/react-core/src/components/Wizard/WizardHeader.tsx
jschuler: Wizard header was not exported before so not a breaking change

packages/patternfly-4/react-core/src/components/Wizard/WizardNavItem.tsx
jschuler: Same comment, new export

redallen:
I addressed the feedback @mcoker @christiemolloy @tlabaj . [A new preview](https://1754-pr-patternfly-react-patternfly.surge.sh/) just deployed, let me know if there are any new issues.

mcarrano:
@redallen Thanks for moving the alert. In reviewing your changes, my only question is about the behavior of the steps menu for mobile. Did you address the issue raised by @christiemolloy ? In my testing, it looks like the overlay remains open after I select a step.

redallen:
Only in the compact nav do I hide the steps. I can make it any step.

Issue #1767, opened 2019-04-15
PF4 Select: Let set an Icon for Toggle


aljesusg:
Feature to let personalize icon to select component.

rachael-phillips:
Hi @aljesusg do you mind telling us more about your use case and need here? We can't see the example in the PR - would you mind updating that?

aljesusg:
Set another Icon not the caret, for example Angle Down, we want to have the same icons for dropdowns

LHinson:
hey @aljesusg! Trying to understand a bit more here. Is the intent to keep your previous icons an artifact of incremental migration? i.e. you are using PF4 components alongside PF3 and need to address inconsistencies between the two? Thanks in advance for the extra info!

aljesusg:
Yes @LHinson , it is to try to have a correct user experience while making the migration to PF4.
However, if PF4 has this icon as a core-style we can close this issue.

tlabaj:
@aljesusg are you looking to change the default Icon?

dlabaj:
@LHinson Any updates on this?

aljesusg:
@dlabaj I closed this because I didn't know that the PF4 guideline indicates that this is the icon for Select, but I saw that @rachael-phillips added this to the the Sprint If you think that this feature is ok for PF4 I can do it

boaz0:
I will give this +1 as this is what we need for the PF4 filter we are using.
I tried to solve this using `InputGroup` and `InputGroupText` but it doesn't look perfect as having the icon inside the select toggle.

If no one is working on this, can you assign this to me.

related to https://github.com/project-koku/koku-ui/issues/946

//cc @dlabrecq

dlabrecq:
Attn: @LHinson @rachael-phillips 

Cost Management needs the ability to add a filter icon in order to implement the PatternFly filter design shown below.

https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#attribute-value-textbox-filters
![Screen Shot 2019-08-14 at 10 19 59 AM](https://user-images.githubusercontent.com/17481322/63028615-273cdb80-be7d-11e9-9694-4288cd3d5a74.png)

rachael-phillips:
[...]

PR #1768, opened 2019-04-15
feat(Select): Let user personalize icon in Select component

aljesusg:
Issue #1767

## Add feature to Select Component

Let user personalize icon in toggle.
By default is the caret.


dlabaj:
@kmcfaul Can you take a look at this.

redallen:
Select is currently being rewritten in TypeScript, so please open an issue for this and we can address it after it's rewritten.

PR #1776, opened 2019-04-16
[WIP] fix(imports): use a different import pattern

seanforyou23:
**What**: Fix an error I was seeing when importing react-core into a project. I don't see this in every project that uses `@patternfly/react-core`, and I haven't nailed down yet what the difference is in the project I'm seeing this in (https://github.com/syndesisio/syndesis-react), but figured I'd get these changes up so others who may know more about it could chime in.

Screen Shot 2019-04-10 at 2 42 17 PM

These changes fix the build issue I'm seeing, but it breaks the demo's that use the exenv pkg. WIP

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.js
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.js
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import ModalContent from './ModalContent';
-import { canUseDOM } from 'exenv';
+import * as ExecutionEnvironment from 'exenv';
 import { KEY_CODES } from '../../helpers/constants';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/patternfly/components/Backdrop/backdrop.css';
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -98,15 +98,15 @@ class Modal extends React.Component {
   render() {
     const { ...props } = this.props;
 
-    if (!canUseDOM) {
+    if (!ExecutionEnvironment.canUseDOM) {
       return null;
     }
 
     if (!this.container) {
```
redallen: Move the lines that create the container to the onComponentMount() lifecycle method.

packages/patternfly-4/react-core/src/components/Modal/Modal.js
```diff
@@ -98,15 +98,15 @@ class Modal extends React.Component {
   render() {
     const { ...props } = this.props;
 
-    if (!canUseDOM) {
+    if (!ExecutionEnvironment.canUseDOM) {
```
redallen: If you do `document.createElement` in the onComponentMount() lifecycle method, this conditional just becomes `!this.container`

seanforyou23:
I'm not seeing the error I was before, so these changes may not be necessary. Closing this PR for now, can revive later if need be.

PR #1786, opened 2019-04-17
feat(Wizard): support for full width and full height

### LINKED ISSUE #1761: Support for full width and full height wizard
https://github.com/patternfly/patternfly-next/pull/1721

`.pf-c-wizard` now supports `.pf-m-full-width` and `.pf-m-full-height`

boaz0:
**What**:

Add two new props to Wizard:
 - `isFullWidth` - set wizard to full width
 - `isFullHeight` - set wizard to full height

closes #1761 

//cc @mcoker

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
@@ -178,3 +178,57 @@ class ValidationWizard extends React.Component {
   }
 }
 ```
+
+### Full height and width wizard
+
+```js
+import React from 'react';
+import { Button, Wizard } from '@patternfly/react-core';
+import SampleFormOne from './examples/SampleFormOne';
+import SampleFormTwo from './examples/SampleFormTwo';
+
+class SimpleWizard extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: false
+    };
+    this.toggleOpen = () => {
+      this.setState({
+        isOpen: !this.state.isOpen
+      });
+    };
+  }
+
+  render() {
+    const { isOpen } = this.state;
+
+    const steps = [
+      { name: 'Step 1', component: <p>Step 1</p> },
+      { name: 'Step 2', component: <p>Step 2</p> },
+      { name: 'Step 3', component: <p>Step 3</p> },
+      { name: 'Step 4', component: <p>Step 4</p> },
+      { name: 'Final Step', component: <p>Final Step</p>, hideCancelButton: true, nextButtonText: 'Close' }
+    ];
+
+    return (
+      <React.Fragment>
+        <Button variant="primary" onClick={this.toggleOpen}>
+          Show Wizard
+        </Button>
+        {isOpen && (
+          <Wizard
+            isOpen={isOpen}
+            isFullHeight
+            isFullWidth
+            onClose={this.toggleOpen}
+            title="Simple Wizard"
+            description="Simple Wizard Description"
+            steps={steps}
+          />
+        )}
+      </React.Fragment>
+    );
+  }
+}
+```
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
@@ -44,6 +44,14 @@ export type WizardStepFunctionType = (newStep: { id?: string | number; name: str
 export interface WizardProps {
   /** True to show the wizard */
   isOpen?: boolean;
+  /** True to set full height wizard */
+  isFullHeight?: boolean;
+  /** True to set full width wizard */
+  isFullWidth?: boolean;
+  /** Custom width of the wizard */
+  width?: number | string;
+  /** Custom height of the wizard */
+  height?: number | string;
   /** The wizard title */
   title: string;
   /** The wizard description */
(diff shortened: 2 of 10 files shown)


mcoker:
@mcarrano yep, that covers it. Thanks. Sorry @boaz0 I meant to list this in the original description. We just want to add a `width` prop (like we do with modal), and one for `height`, too.

boaz0:
@mcoker & @mcarrano no problem. I will update this PR.
Thanks.

boaz0:
Sorry but I couldn't help myself and added height and width props as required.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -44,6 +44,14 @@ export type WizardStepFunctionType = (newStep: { id?: string | number; name: str
 export interface WizardProps {
   /** True to show the wizard */
   isOpen?: boolean;
+  /** True to set full height wizard */
+  isFullHeight?: boolean;
+  /** True to set full width wizard */
+  isFullWidth?: boolean;
+  /** Default width of the Modal. */
```
tlabaj: Can you update the comments here to say Wizard.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -330,7 +346,7 @@ class Wizard extends React.Component<WizardProps> {
         <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
           <Backdrop>
             <Bullseye>
-              <div {...rest} className={css(styles.wizard, className)} role="dialog" aria-modal="true" aria-labelledby={this.titleId} aria-describedby={description ? this.descriptionId : undefined}>
+              <div {...rest} style={{ width, height }} className={css(styles.wizard, isFullWidth && styles.modifiers.fullWidth, isFullHeight && styles.modifiers.fullHeight,className)} role="dialog" aria-modal="true" aria-labelledby={this.titleId} aria-describedby={description ? this.descriptionId : undefined}>
```
tlabaj: Can you add Logic here to always apply isFullWith when there is a width set and alway set the isFullHeight when the height is set.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
```diff
@@ -178,3 +178,57 @@ class ValidationWizard extends React.Component {
   }
 }
 ```
+
+### Full Height and Width Wizard
```
tlabaj: can you make this title sentence case. 

> Full height and width wizard

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
boaz0: should I remove 'isFullWidth/Height` from props?

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
boaz0: Thanks!

packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
boaz0: yes sure.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
tlabaj: No, they should remain in the props.  When width or height are applied.the take precedence over the isFullWidth/Height.  we need to apply isFullWidth/Height for the scenario when a size that is bigger the max size is applied, the isFullWidth/Height modifiers will force the max to be overridden.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -287,6 +303,8 @@ class Wizard extends React.Component<WizardProps> {
     const computedSteps: ComputedStep[] = this.initSteps(steps, activeStep);
     const firstStep = activeStep === flattenedSteps[0];
     const isValid = activeStep.enableNext !== undefined ? activeStep.enableNext : true;
+    const setFullWidth = isFullWidth || !!(width)
```
tlabaj: Shouldn't this be we want to set these booleans if the width/height exist :
```suggestion
    const setFullWidth = isFullWidth || width
    const setFullHeight = isFullHeight || height
```

boaz0:
I see the same problem on netlify but on my local machine it works fine.

In regards to `null` as the default. This means that we are not going to inject `width` or `height` styling to `wizard` and they will be set by the default `width` and `height` values from patternfly core.

mcarrano:
If that's the case @boaz0 , maybe it makes sense to change the description to "custom width/height of the wizard"? What you say makes sense, but I think the word "default" in the description confused me. The default size is really coming from Core.

redallen:
Wizard docs are indeed broken on master, not sure which commit broke it. It happened recently and I'll try to get a fix out before lunch :)

boaz0:
@mcarrano updated the docs to have "custom" and not "default"

redallen:
So it looks like the fix is in `jschuler:patternfly-react/controlled-wizard`. It has to do with fixing imports/exports. I'll try to cherry pick some commits into here from there.

redallen:
Previews fixed, enjoy! https://1786-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/wizard/

boaz0:
Thank you all @mcoker @mcarrano and @redallen

Issue #1714, opened 2019-04-05
About modal `productName` should be optional


rhamilto:
In the OpenShift console, we use a logo that is the product name for `brandImageSrc`.  As a result, we don't need a separate `productName` that repeats this information as it is redundant and looks a bit silly.  

![Screen Shot 2019-04-05 at 9 53 43 AM](https://user-images.githubusercontent.com/895728/55632912-47e77d00-5789-11e9-867f-4b7ae564cd34.png)

Please make `productName` an optional prop so it is not required.  

Note we can work around the issue by passing an empty string for product name, but the component still outputs markup that renders padding, thus taking up vertical space in the rendered output.

![Screen Shot 2019-04-05 at 9 55 40 AM](https://user-images.githubusercontent.com/895728/55632994-79604880-5789-11e9-9175-41c8002e2c37.png)

cc:  @janwright73, @spadgett

LHinson:
@rhamilto yeah, I think that makes sense. For planning purposes, can you give me some sense of urgency and/or your timeline to get this update?

rhamilto:
> For planning purposes, can you give me some sense of urgency and/or your timeline to get this update?

No rush as the workaround is sufficient for 4.1 release.

LHinson:
Good to know. Thanks @rhamilto

dgutride:
Closing via https://github.com/patternfly/patternfly-react/pull/1797

PR #1797, opened 2019-04-17
feat(aboutmodal): make productName optional

jessiehuff:
#1714

**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
```diff
@@ -17,7 +17,7 @@ const propTypes = {
   /** A callback for when the close button is clicked */
   onClose: PropTypes.func,
   /** Product Name */
-  productName: PropTypes.string.isRequired,
```
tlabaj: You need to make the prop in the AboutModal.d.ts File optional as well.

Issue #1240, opened 2019-01-24
Update background image component.


matthewcarleton:
Background image component
https://pf-next.com/components/BackgroundImage/examples/
Core PR
https://github.com/patternfly/patternfly-next/pull/1285

We now include the svg inline to fix a bug with safari.

priley86:
An additional request here from the Open Shift side would be to conditionally load the background images per the current breakpoint. Today, the `BackgroundImg` loads all images. This is not ideal and we should only load the single image needed for current screen size.

priley86:
Actually, this may have been recently resolved... i'm not seeing this while loading the page here anymore.
http://patternfly-react.surge.sh/patternfly-4/demos/pagelayout/examples/page-layout-default-nav

![screen shot 2019-02-18 at 4 20 25 pm](https://user-images.githubusercontent.com/4237045/52976859-25480300-3399-11e9-8498-8e1cdb657b25.png)

rachael-phillips:
@priley86 just following up. Has this been resolved?

priley86:
@rhamilto ^^ any thoughts? I think last we spoke we wanted to look into the default background size. Would appreciate any further feedback when you get a chance (and let me know if you need anything at all).

rhamilto:
Looks fine.  But note the reason we are not using this component in OpenShift console is because **multiple, resolution dependent, very large** background images were being **bundled** with the console.  And of those large background images, **only a small sliver of each image was ever visible** in the console.  So rather than bundling several megs of images (most of which was never visible in the console), we're opting to only serve **one relevant image** that meets the specific needs of the OpenShift console.  For comparison, [pfbg_1200.jpg](https://github.com/patternfly/patternfly-next/blob/master/static/assets/images/pfbg_1200.jpg) clocks in at 271 KB.  [Our override](https://github.com/openshift/console/blob/master/frontend/public/imgs/pfbg_2000.jpg) is 24.9 KB.  So our override is getting us the same visual effect but with **only 24.9 KB regardless of viewport size**.   The [smallest image from this component](https://github.com/patternfly/patternfly-next/blob/master/static/assets/images/pfbg_576.jpg) still clocks in at 65.9 KB (that's 41 KB more than the image we're using for **all** viewport sizes).

rhamilto:
It may be worth noting the about modal has this same issue (multiple, resolution dependent, very large background images are being used, but only a small portion of each image is ever visible) and we've also implemented an override in OpenShift.

[...]

PR #1811, opened 2019-04-18
feat(BackgroundImage): update background image component

jessiehuff:
#1240

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
+++ b/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
@@ -54,8 +54,8 @@ export const defaultProps = {
 };
 
 const BackgroundImage = ({ className, src, ...props }) => {
-  // Default string value to handle all sizes
-  const variableOverrides =
+    // Default string value to handle all sizes
+    const variableOverrides =
     typeof src === 'string'
       ? Object.keys(BackgroundImageSrc).reduce(
           (prev, size) => ({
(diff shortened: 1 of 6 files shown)


mcoker:
Thanks @jessiehuff! In this case, the SVG is going to need to be inline. This resolves a bug in Safari that requires the filter SVG be inline in the page where the filter is used.

Also I'm seeing this CSS applied in the background-image example

`--pf-c-background-image--Filter: url(/assets/images/background-filter.svg#image_overlay);`

That will need to be removed, too, as it's overwriting this value that comes from core

`--pf-c-background-image--Filter: url(#image_overlay);`

packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
```diff
@@ -66,16 +63,35 @@ const BackgroundImage = ({ className, src, ...props }) => {
         )
       : src;
 
-  const bgStyles = StyleSheet.create({
+    const bgStyles = StyleSheet.create({
```
redallen: Can we fix these tabs to be the way they were before?

redallen: Otherwise, looking good!

packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.js
```diff
@@ -10,26 +10,23 @@ import {
...
+  c_background_image_BackgroundImage_lg
 } from '@patternfly/react-tokens';
 
 export const BackgroundImageSrc = {
   xs: 'xs',
   xs2x: 'xs2x',
   sm: 'sm',
   sm2x: 'sm2x',
-  lg: 'lg',
-  filter: 'filter'
```
tlabaj: @Don Do we need keep the filter here? Removing it would cause a breaking change since it is exported.

packages/patternfly-4/react-core/src/components/BackgroundImage/examples/SimpleBackgroundImage.md
```diff
@@ -0,0 +1,9 @@
+---
+title: 'Background Image'
```
dlabaj: Need to update to sentence case for 'Background image'

Issue #1803, opened 2019-04-17
Add hoverable prop to card


mcoker:
* `.pf-m-hoverable` - applies a box shadow on hover - https://github.com/patternfly/patternfly-next/pull/1734/files

PR #1812, opened 2019-04-19
feat(card): adds card is-hoverable modifier prop

dgutride:
Closes: #1803


mcarrano:
@mceledonia can you review this one to make sure correct drop shadow is applied and tag as 'UX approved' if you are good with it. Thanks!

mceledonia:
I'm not sure I'm seeing the right thing here, is the react preview link above the correct source? I'm just getting a barebones pf4 layout with a couple small cards and no hovers on FF. @mcarrano

mcarrano:
@mceledonia You should be looking at the preview of the Card component here: https://1812-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/card/  If you scroll down the page you will see an example called Card Hover Example.

mceledonia:
Thanks Matt, this looks correct in terms of what was designed.

Issue #1802, opened 2019-04-17
Add padding props to page main section


mcoker:
* `.pf-m-no-padding` - removes padding from main section - https://github.com/patternfly/patternfly-next/pull/1733

* `.pf-m-no-padding-mobile` - removes padding when the page layout breaks to the mobile view - https://github.com/patternfly/patternfly-next/pull/1497

dlabaj:
Are these always added together or you can choose to add one and not the other.

mcoker:
You can choose one or the other.

dgutride:
Closed via https://github.com/patternfly/patternfly-react/pull/1818

PR #1818, opened 2019-04-22
feat(padding): add pagesection no padding props

dgutride:
**What**:

**Additional issues**:


dgutride:
Closes: https://github.com/patternfly/patternfly-react/issues/1802

Issue #1751, opened 2019-04-11
Add selected state to single select


mcoker:
It adds `.pf-m-selected` to `.pf-c-select__menu-item` and adds an icon as the last thing in the menu item.

https://github.com/patternfly/patternfly-next/pull/1717/

tlabaj:
closed by #1827

PR #1827, opened 2019-04-22
refactor(Select): add selected state to single select

tlabaj:
#1751

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/SelectOption.js
+++ b/packages/patternfly-4/react-core/src/components/Select/SelectOption.js
@@ -97,7 +97,7 @@ class SelectOption extends React.Component {
               onKeyDown={this.onKeyDown}
             >
               {value}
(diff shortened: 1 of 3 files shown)


mcoker:
@tlabaj looks good to me. Do you know what the inline `style="vertical-align: -0.125em;"` on the selected icon is from?

tlabaj:
@jschuler please see @mcoker comment above about the vertical align.  I know we add a similar issue with a close button maybe?  Do you remember.  I think it has something to do with the way we pull in Font Awesome if I am remembering correctly.

jschuler:
@tlabaj @mcoker that is coming from react-icons. When we use an icon from there (like `CheckIcon`) it applies the vertical offset. This can be disabled by passing in `noVerticalAlign` prop, e.g. `<CheckIcon noVerticalAlign />`

PR #1828, opened 2019-04-22
feat(Tab): Render Tab as <a> instead of <button> if it contains an href prop

### LINKED ISSUE #1805: Add option to use <a> element on tabs
https://github.com/patternfly/patternfly-next/pull/1722/files

> Note: `<a>` should be used when there is a URL associated with the tab, and should only be used when `.pf-c-tabs` is used with `<nav>`.

mturley:
Closes #1805

**What**: Adds a prop `href` to the `Tab` component that causes it to render `<a>` instead of `<button>`. Adds a note to the props table explaining that this should be done if the consumer is using the `"nav"` variant for their `Tabs` container.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/Tab.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Tabs/Tab.d.ts
@@ -4,6 +4,7 @@ import { Omit } from '../../helpers/typeUtils';
 export interface TabProps extends Omit<HTMLProps<HTMLDivElement>, 'id'> {
   children?: ReactNode;
   className?: string;
+  href?: string;
   eventKey: number;
   id?: string;
   tabContentId?: string | number;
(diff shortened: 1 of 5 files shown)


jgiardino:
These updates look good! 

As part of this PR, can we change the example label from "AccessibleTabs" to "Tabs using the nav element" (like we have in core)? And "Accessible Secondary Tabs" to "Secondary tabs using the nav element".

Using "Accessible" implies that the others are not accessible, which isn't the case.

mturley:
@jgiardino sounds good, I'll change that. Also, I think I'm going to get rid of the `#/item/N` href values and just use `href="#"` in the docs examples, because the URL has changing is messing with gatsby and making it so clicking the tabs doesn't do anything on first click. Also seeing a path in the URL bar like `/components/tabs/#/items/1` is a little confusing since you haven't left the page.

PR #1829, opened 2019-04-22
add bodyWrapper type def

priley86:
**What**:
Discovered these were missed in #1227 today while testing downstream
```
ERROR in /Users/priley/GitHub/web-ui/frontend/public/components/factory/table.tsx(386,12):
TS2322: Type '{ children: Element[]; cells: any[]; rows: any; onSelect: (event: MouseEvent<Element>, isSelected...' is not assignable to type 'IntrinsicAttributes & TableProps & { children?: ReactNode; }'.
  Type '{ children: Element[]; cells: any[]; rows: any; onSelect: (event: MouseEvent<Element>, isSelected...' is not assignable to type 'TableProps'.
    Property 'bodyWrapper' is missing in type '{ children: Element[]; cells: any[]; rows: any; onSelect: (event: MouseEvent<Element>, isSelected...'.
```

do these work OK for now @suomiy  or did you have something else in mind? 

**Additional issues**:
none

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
@@ -104,6 +104,8 @@ export interface TableProps extends Omit<Omit<HTMLProps<HTMLTableElement>, 'onSe
   dropdownDirection?: OneOf<typeof DropdownDirection, keyof typeof DropdownDirection>;
   rows: Array<IRow | Array<String>>;
   cells: Array<ICell | String>;
+  bodyWrapper?: Function;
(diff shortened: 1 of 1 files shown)


suomiy:
This is passed directly to reactabular renderers which invoke React.CreateElement with that as a parameter. So the type should mirror https://reactjs.org/docs/react-api.html#createelement documentation I guess?

Extensions will probably have only the function type if HoC is used.

priley86:
`Function` is safe for now and avoids the ts error above, we can always iterate later though 😉 

thanks all!

PR #1833, opened 2019-04-23
(Fixes #1455) RC.1 updates for Progress Bar.

### LINKED ISSUE #1455: Progress - updates needed based on updates in core
The following issues were opened in core and should be addressed in the react component:

- Progress bar example update - https://github.com/patternfly/patternfly-next/issues/1128 
  - [x] Text strings in examples were updated.
  - [x] With this update, the react examples labeled "Progress (Labeled)" and "Progress (Dynamic Description)" should be merged into a single example, and should have aria attributes similar to the core examples "Progress Finite Step" and "Progress Step Instruction".
  - [x] It's important to note with these changes that the `aria-valuemax` should always be 100 unless a `label` prop is defined (which gets mapped to `aria-valuetext`). Using `aria-valuetext` is the only time we should allow the `aria-valuemax` attribute to be some value other than 100. See a11y documentation in the core example for more explanation about this.
- progress bar - move aria attributes - https://github.com/patternfly/patternfly-next/issues/1456

dlabaj:
Fixes issue #1455 

**What**:

**Additional issues**:


jgiardino:
These updates are great!

The only issue I noticed, which is minor, is in the last example `value` is specified twice.

Other than that, what do you think about updating the prop descriptions to something like this?

`label` - Text description of current progress value to display instead of percentage.
`valueText` - Accessible text description of current progress value, for when value is not a percentage. Use with `label`.

packages/patternfly-4/react-core/src/components/Progress/Progress.md
```diff
@@ -118,16 +111,25 @@ import { Progress, ProgressSize, ProgressMeasureLocation, ProgressVariant } from
...
   value={33}
-  title="Descriptive text here"
+  title="Title"
   measureLocation={ProgressMeasureLocation.none}
   variant={ProgressVariant.danger}
 />
 ```
 
-## Progress (dynamic description)
+## Progress (Finite Step)
```
tlabaj: can you make this sentence case?

packages/patternfly-4/react-core/src/components/Progress/Progress.md
```diff
@@ -118,16 +111,25 @@ import { Progress, ProgressSize, ProgressMeasureLocation, ProgressVariant } from
...
+```js
+import React from 'react';
+import { Progress, ProgressSize, ProgressMeasureLocation, ProgressVariant } from '@patternfly/react-core';
+
+<Progress value={2} min={0} max={5} title="Title" measureLocation={ProgressMeasureLocation.top} label="2 of 5" valueText="2 of 5"/>
+```
+
+## Progress (Step Instruction)
```
tlabaj: sentence case here too

packages/patternfly-4/react-core/src/components/Progress/Progress.md
dlabaj: Done.

packages/patternfly-4/react-core/src/components/Progress/Progress.md
dlabaj: Done.

christiemolloy:
I can't see the Small Progress example (this is in Core)

Screen Shot 2019-04-25 at 10 02 03 AM

christiemolloy:
Lets update the Progress Success example to look like this:

Screen Shot 2019-04-25 at 10 03 17 AM

christiemolloy:
Also update Progress Inside Success to look like this:

Screen Shot 2019-04-25 at 10 04 31 AM

PR #1835, opened 2019-04-23
fix(build): make downstream patternfly-org work upstream

redallen:
**What**: We need a better long-term solution. I believe that solution should be implemented by refactoring patternfly-org to use a [Git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) of this repo and patch the docs however they like.

**Additional issues**: ~Waiting on https://github.com/ChristopherBiscardi/gatsby-mdx/pull/374 to be merged to help out with importing raw examples in our MDX to avoid code duplication for fullscreen component examples.~ Don't have time. Will open issue when merged to fix it.


redallen:
When a build of `gatsby-mdx` comes out that has my merged PR in, we will use that version of the plugins and a raw loader to ditch the duplicated code.

PR #1836, opened 2019-04-23
chore(cleanup): fix dup heading level feature, relax title typedef

seanforyou23:
**What**: Consolidate heading component implementations. Make Title headingLevel prop more flexible.

Currently, there are two ways to generate a heading. The preferred way is using the dedicated Title component, a secondary and legacy way is using the Text component. I don't think we need multiple implementations as it confuses the semantics of the component and makes searching/writing docs harder. This PR removes the heading capability from the `Text` component, deferring to `Title` which has more features and is further tested/documented.

On the note of Title, we had a small but important regression recently. The ability to supply string values for headingLevel in Title I think is important. It's a bit of a tripping hazard. As a developer, I expect to be able to write `<Title size="3xl" headingLevel="h2">...` and not forced to use the enum, for example `<Title size="3xl" headingLevel={TitleLevel.h2}>...`. This PR adds back that capability so developers can use either approach. Makes for smoother developer experience.

Before

Screen Shot 2019-04-23 at 12 17 37 PM

After

Screen Shot 2019-04-23 at 1 59 12 PM

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Title/Title.tsx
+++ b/packages/patternfly-4/react-core/src/components/Title/Title.tsx
@@ -21,7 +21,7 @@ export interface TitleProps extends Omit<React.HTMLProps<HTMLHeadingElement>, 's
   /** Additional classes added to the Title */
   className?: string;
   /** the heading level to use */
-  headingLevel?: TitleLevel;
+  headingLevel?: TitleLevel | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 }
 
 const Title: React.FunctionComponent<TitleProps> = ({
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/components/Title/Title.tsx
```diff
@@ -21,7 +21,7 @@ export interface TitleProps extends Omit<React.HTMLProps<HTMLHeadingElement>, 's
   /** Additional classes added to the Title */
   className?: string;
   /** the heading level to use */
-  headingLevel?: TitleLevel;
+  headingLevel?: TitleLevel | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
redallen: Can't you use something like `TitleLevel | keyof TitleLevel`? A slight bit cleaner.

seanforyou23: I did try that, unfortunately, it does not work. We could create an additional interface as it was before but that felt unnecessary. I think the main thing is that the intellisense works and developers don't need to pull in an entire module just to set a well-known string value.

Issue #1216, opened 2019-01-17
PF4: Custom Select Component: Type-ahead select (multi item) variation


tlabaj:
This issue covers the Type-ahead select (multi item) variation of the Custom Select only:
(This issue is a child of issue #1206 )

On hold until Type-ahead select (single item) complete (issue #1215 )

Related issue in Core: patternfly/patternfly-next#651

Info provided on the component: https://docs.google.com/document/d/1rrCdARutABI-Vpa_mdMWGl2xJJTtt2dPQ2bjFx2zwCM/edit?usp=sharing

see: https://pf-next.com/components/Select/examples/

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @srambach (CSS) @mcarrano (interaction design) @christiemolloy (visual design)

dgutride:
Closed via https://github.com/patternfly/patternfly-react/pull/1839

Issue #1215, opened 2019-01-17
PF4: Custom Select Component: Type-ahead select (single item)


tlabaj:
**This issue covers the Type-ahead select (single item) variation of the Custom Select only:**
(This issue is a child of issue #1206  )

**On hold until the Single select variation is complete (issue #1214 )** 

Related issue in Core: patternfly/patternfly-next#651

Info provided on the component: https://docs.google.com/document/d/1rrCdARutABI-Vpa_mdMWGl2xJJTtt2dPQ2bjFx2zwCM/edit?usp=sharing

see: https://pf-next.com/components/Select/examples/

In addition to a React review, this issue should be reviewed by the core contributors:
CSS: @srambach (CSS) @mcarrano (interaction design) @christiemolloy (visual design)

priley86:
Doing some PF3 typeahead integration work recently in CNV, I noted some of the following things about the React API surface. Would just like to share these for consideration as we begin to review this React component.

* The existing Typeahead supports both synchronous and asynchronous functionality. I would consider synchronous to be most pressing for our needs there first. The API of props here for our existing `Typeahead` and `AsyncTypeahead` would be good to consider as you start to implement the API.
https://github.com/ericgio/react-bootstrap-typeahead/blob/master/docs/Props.md

* Sample usage and some notes:
```
  <TypeAhead
          id={id}
          labelKey={choicesLabelKey}
          defaultValue={defaultValue}
          placeholder={placeholder}
          choices={choices}
          multiple={multiple}
          selected={value}
          onChange={onChange}
          onBlur={onBlur}
          disabled={disabled}
        />
```
This sort of synchronous typeahead would support some default cases that could be used immediately (with multiple variation a nice-to-have for later since this issue pertains to single item). It is important to note that this existing API surface allows the `choices` prop to be an array of strings OR an array of objects. If an array of objects is provided (such as a simple key/value array): 
```
[{ id: 1, name: 'fedora29' }, { id: 1, name: 'fedora28' }]
```
the React component should look at the `labelKey` prop to determine which object property to render for its label. The `onChange` callback would of course return an object instead of a string in this scenario...

[...]

PR #1839, opened 2019-04-24
feat(Select): add typeahead variants (single and multi)

kmcfaul:
…, updated tests

**What**: adding typeahead single and multi variant

**See issue**: #1215, #1216


mcoker:
In the typeahead version of the select dropdown, we wrap the toggle icon on the right in a `button.pf-c-button.pf-c-select__toggle-button` since the parent element is a `<div>`. That's what's creating the extra space there.

jgiardino:
The behavior we have in the [PF3 TypeAhead Select](https://patternfly-react.surge.sh/patternfly-3/?selectedKind=patternfly-react%2FForms%20and%20Controls%2FTypeAhead%20Select&selectedStory=Menu%20Alignment&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) works really well, and matches what I would expect. I suggest reviewing that component and how it behaves. One behavior to check out is how the arrow keys are able to change the selected value while focus remains in the input field. The ability to switch between arrow keys and typing is pretty seamless.

The way the PF3 component handles the menu keyboard interaction is different in that they use `aria-activedescendant` to manage focus instead of `tabindex`, which is what we have with the original Select variants. Both methods are valid, but the `aria-activedescendant` method lets you keep focus on the input, but still use the arrow keys to change the selection, which is more appropriate for this case.  WAI-ARIA describes this method here: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_focus_activedescendant. 

Let me know if you have any questions about this.

kmcfaul:
Added multi typeahead variant
Added missing button wrapper in toggle to typeahead variants

jgiardino:
Per discussion with @dgutride, I captured my previous comment in this issue: https://github.com/patternfly/patternfly-react/issues/1842

kmcfaul:
Regarding the typeahead logic, the code is currently converting the text input to a regex to filter options. Special characters are ignored as they break the regex and would crash the page.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -3,12 +3,15 @@ import { HTMLProps, FormEvent, ReactNode } from 'react';
 export const SelectVariant: {
   single: 'single';
   checkbox: 'checkbox';
+  typeahead: 'typeahead';
+  typeahead_multi: 'typeaheadmulti';
```
tlabaj: Curious why we are not using CamelCase here?

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
```diff
@@ -3,12 +3,15 @@ import { HTMLProps, FormEvent, ReactNode } from 'react';
...
 };
 
 export interface SelectProps extends HTMLProps<HTMLOptionElement> {
   isExpanded?: boolean;
   isGrouped?: boolean;
   onToggle(value: boolean): void;
+  onClear() : void;
```
tlabaj: this should be optional to match JS

packages/patternfly-4/react-core/src/components/Select/Select.js
```diff
@@ -51,18 +57,61 @@ const defaultProps = {
...
+    } else {
+      this.setState({
+        openedOnEnter: false,
+        typeaheadValue: null,
+        filteredChildren: this.props.children
+      });
+    }
+  };
+
+  onChange = e => {
```
tlabaj: can we add some unit test for this.

packages/patternfly-4/react-core/src/components/Select/Select.d.ts
kmcfaul: I can change it

jgiardino:
Great job on pulling the multi-select typeahead together!

Now that keyboard and screen reader stuff is off the table, here are my other comments :-)

- There are some differences between the core and react implementations. In this implementation, I see a lot more in this div that what we have in core:
  `<div aria-labelledby="title-id pf-toggle-id-128" id="pf-toggle-id-128" class="pf-c-select__toggle pf-m-typeahead" aria-expanded="false" aria-haspopup="listbox">`
  This is what we have in core (i.e. just the class names):
  `<div class="pf-c-select__toggle pf-m-typeahead">`

- The following is the input element in the **Typeahead select input** example:
  `<input class="pf-c-form-control pf-c-select__toggle-typeahead" id="select-single-typeahead-typeahead" aria-label="Type to filter" placeholder="Select a state" type="text" value="">`
  Is there a prop available to let the user provide a text string for aria-label? This string works for a default value, but we need to allow consumers to specify their own value.
  - For a prop name, does `ariaLabelTypeAhead` (fyi, as part of https://github.com/patternfly/patternfly-react/pull/1731, we came to some conclusions about how to handle prop names for aria-label attributes, so in this case it's `ariaLabel` + name of component feature, like `Toggle`, `Search`, `Menu`, etc...)

- I get different behaviors for `button.pf-c-select__toggle-clear` between click and Enter keys. These events should always result in the same action (unless there's some edge case I'm not thinking of). The button works as expected on click, but when the button has focus and I hit Enter, it opens the menu instead of clearing the value. 
  - I'm also seeing that this button is included in the react implementation for single select typeahead, but isn't included in the Core example. Is that intentional? Is it missing in core, or should it not be included here? (cc'ing @mcarrano on this one)

- Can you update the text strings that are defined in both examples to use the same string for the hidden span, the aria-label, and the placeholder text, like I show below? I think this is helpful in illustrating what the purpose of these different elements/attributes are (i.e. that they should be communicating the same info). 
  ```
      <div>
        <span id={titleId} hidden>
          Select a state
        </span>
        <Select
          aria-label="Select a state"
          onToggle={this.onToggle}
          onSelect={this.onSelect}
          onClear={this.clearSelection}
          selections={selected}
          isExpanded={isExpanded}
          ariaLabelledBy={titleId}
          placeholderText="Select a state"
        >
  ```

mcoker:
@jgiardino 

> I'm also seeing that this button is included in the react implementation for single select typeahead, but isn't included in the Core example. Is that intentional? Is it missing in core, or should it not be included here?

I see it here, or are you referring to another example? 

Screen Shot 2019-04-24 at 3 41 15 PM

kmcfaul:
@mcarrano 
1. Change input to clear on selection.
2. Toggling on click is now prevented for type ahead variants. Clicking will only open the menu, and clicking outside the menu, selecting an item (for single), or pressing escape will close the menu.
3. As far as I can tell from the code, the overflow of ChipGroup is hardcoded to happen at 2+ chips. It's also possible the large margin of the clear all selections icon is making the div fill up early effectively.

kmcfaul:
@jgiardino 

1. The extra attributes on the toggle label come from the toggle class that is reused by all variants. If there's something specific that shouldn't be on the tag for typeahead let me know.

2. Added the areaLabelTypeAhead parameter that is passed to the aria label here.

3. Removed the toggle's open on enter for type ahead variants. The clear selection button should behave as expected now.

4. Updated the examples with your suggestions.

jgiardino:
Thanks for making updates! I included some more comments below, but if any of these seem like updates for a later PR, just let me know. 

> The extra attributes on the toggle label come from the toggle class that is reused by all variants. If there's something specific that shouldn't be on the tag for typeahead let me know.

The following are the differences in how attributes are defined between these implementations and core for the single select. I'm assuming the multi-select would need similar updates, but let me know if you have questions about that one.

| react | core | diff |
| - | - | - |
| `<div aria-labelledby="title-id pf-toggle-id-12" id="pf-toggle-id-12" class="pf-c-select__toggle pf-m-typeahead" aria-expanded="false" aria-haspopup="listbox">` | `<div class="pf-c-select__toggle pf-m-typeahead">` | `aria-labelledby`, `id`, `aria-expanded`, `aria-haspopup` should **not** be included on this element |
| `<button class="pf-c-button pf-c-select__toggle-button">` | `<button class="pf-c-button pf-c-select__toggle-button" id="select-single-typeahead-expanded-toggle" aria-haspopup="true" aria-expanded="true" aria-labelledby="select-single-typeahead-expanded-label select-single-typeahead-expanded-toggle" aria-label="Select">` | `aria-labelledby`, `id`, `aria-expanded`, `aria-haspopup` should be included on this element |

> Added the areaLabelTypeAhead parameter that is passed to the aria label here.

Awesome! There are a couple of other text strings that we should expose via props if they aren't already. They are on these elements:
- `<button class="pf-c-button pf-m-plain pf-c-select__toggle-clear" aria-label="Clear all">`
  - Does `ariaLabelClear` work?
- `<button class="pf-c-button pf-c-select__toggle-button" id="select-single-typeahead-expanded-toggle" aria-haspopup="true" aria-expanded="true" aria-labelledby="select-single-typeahead-expanded-label select-single-typeahead-expanded-toggle" aria-label="Select">`
  - I'd use `ariaLabelToggle` for this one.
- `<li class="pf-c-chip"> ... <button class="pf-c-button pf-m-plain" aria-label="Remove" ...>` in the multiselect
  - `ariaLabelRemove` probably works
  - Also, in core, this value is set to "Remove", but the default string "close" that's defined for the Chip is currently being used. 

One more thing I notice that is really minor... The new examples have the same `id` values for the hidden span, which results in duplicate ids on the page:
`const titleId = 'title-id';`
But other examples provide something more specific, like: 
`const titleId = 'grouped-checkbox-select-id';`

mcarrano:
This is looking much better @kmcfaul !  Just a couple of additional questions/comments:

> Toggling on click is now prevented for type ahead variants. Clicking will only open the menu, and clicking outside the menu, selecting an item (for single), or pressing escape will close the menu.

Thanks for making this change. the interaction is much better. Seems like the user should also be able to click the toggle (down arrow) itself to close the menu.

> As far as I can tell from the code, the overflow of ChipGroup is hardcoded to happen at 2+ chips. It's also possible the large margin of the clear all selections icon is making the div fill up early effectively.

Where is this hardcoded? In the ChipGroup component? That may be a problem as that's not what I think was intended.

It would be great if we could have a string like "No matches found" display in the dropdown when there is a null query result.

kmcfaul:
@jgiardino 
- Added a control to change where the aria attributes are assigned based on variant. They should remain the same for normal/checkbox select and for typeahead variants the attributes will appear on the toggle button.
- Added the mentioned aria attributes as parameters
- Updated the titleIds on the examples

@mcarrano 
- The carat button should now behave as a toggle, opening and closing the menu
- Added a 'No results found' message when the filtering returns nothing

jgiardino:
Your updates look great! Here are a few more minor things I noticed. Let me know if you have any concerns about getting these addressed in time.

1. The `ariaLabelToggle` should only be used for the type-ahead select variants and is not recommended for other variants because they will have text already which will conflict with the aria-label value. I'm not sure if that's easy to address or not. 
    1. I'd at least update the description from "Label for toggle of select variants" to "Label for toggle of **type ahead** select variants" to discourage consumers from defining that for other variants.

2. In the options menu for similar situations (i.e. a button with an icon instead of text), we use "Options menu" as the default value for `aria-label` unless it is specified by the user. Can we do the same here? I.e. if `ariaLabelToggle` is not defined, automatically set `aria-label` to "Options menu" on the `.pf-c-select__toggle-button` button?
    1. Ideally, we should have a default string, but if not possible in this PR, should we at least add the prop to the example?

3. When `ariaLabelRemove` is not defined in the multi-typeahead example, there isn't a default string defined for aria-label on the chip close buttons (even the default that I see on the Chip component isn't showing up)

4. The `ariaLabelClear` prop should behave similar to what I describe above—when defined, its value is used as the `aria-label` value, but when not defined we should still have `aria-label` with a default value. I'm seeing slightly different behaviors between the two typeahead variants:
    1. This prop seems to have no effect on the multi-select typeahead variant. The text string is always "Clear all" regardless of whether this prop is defined or not. 
    2. If this prop is defined on the single-select typeahead, it works. But if I remove this prop, then there is no default text string that displays.
    3. In both of these examples, I would recommend not defining this prop (or `ariaLabelRemove`), and just have the default text strings used. 

3. This might be completely unrelated to this PR... If I remove the prop `ariaLabelledBy` from the Select component, it will result in this value for the attribute: 
  `aria-labelledby="null pf-toggle-id-60"`
But instead I would expect that attribute to just not be included (i.e. an element is already labeled by itself, so having aria-labelledby reference only itself is unnecessary).

jgiardino:
Great job on making the updates! All the props work as expected! There are just a few minor text string updates needed to match core:

3. For multi-select, in core, the value for aria-label on `li.pf-c-chip`>`button` is set to "Remove", but in the latest updates the default string is set as "Remove selection". This should match core.

4. For both variants, in core, the value for aria-label on `pf-c-select__toggle-clear` is "Clear all", but in the latest updates the default string is set as "Clear all selections". This should match core.
    - Additionally, the prop `ariaLabelClear` is defined in the example as "Clear all", which _is_ the correct string, but in these examples we can leave the prop off and just use the default string.

PR #1843, opened 2019-04-24
feat(DropdownToggle): Add splitButtonItems prop, DropdownToggleCheckbox and split-button dropdown examples

### LINKED ISSUE #1287: Add split-button dropdown
https://github.com/patternfly/patternfly-next/pull/1263/

This is a new variation of a dropdown. The current dropdown toggle element changed from a `button` to a `div`, and you can now put interactive elements in the toggle `div` and those control the menu. Our current example shows a `button.pf-c-dropdown__toggle-button` and `label.pf-c-dropdown__toggle-check` (with an `<input type="checkbox">` in it) in the toggle area, but that could be 3 buttons, 2 buttons and a check, etc. We may add more things like the `pf-c-dropdown__toggle-button/check` later. The aria properties will move from the current `button.pf-c-dropdown__toggle` to the element(s) that control the menu - in this example, that's `.button.pf-c-dropdown__toggle-button`.

mturley:
Closes #1287

**What**: Adds a new prop `splitButtonItems` to the `DropdownToggle` component which causes it to render a wrapper div surrounding those items and the dropdown toggle button. Also adds a new component `DropdownToggleCheckbox` for use as an element in `splitButtonItems`, which copies all of the props of the `Checkbox` component except `label`, and renders the checkbox as a child of the label element instead of rendering a checkbox with an adjacent label.

Note: I didn't write the new component in TypeScript, just because none of the Dropdown components are in TypeScript yet and I figure we should convert them all at once.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.md
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.md
@@ -5,7 +5,7 @@ cssPrefix: 'pf-c-dropdown'
 
 ## Simple dropdown
 
-import { Dropdown, DropdownToggle, DropdownItem, DropdownSeparator, DropdownPosition, DropdownDirection, KebabToggle } from '@patternfly/react-core';
+import { Dropdown, DropdownToggle, DropdownToggleCheckbox, DropdownItem, DropdownSeparator, DropdownPosition, DropdownDirection, KebabToggle } from '@patternfly/react-core';
 import { ThIcon } from '@patternfly/react-icons';
 
 ```js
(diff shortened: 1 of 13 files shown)


mturley:
I found a console warning on the Dropdown docs page that appears to have already existed on master. I don't think it should block this PR, in case anyone notices it. Opened an issue for it: https://github.com/patternfly/patternfly-react/issues/1844

redallen:
There are a lot of such warnings @mturley, and in the Jest tests too :)

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
```diff
@@ -5,19 +5,38 @@ import Toggle from './Toggle';
...
+      )}
+    >
+      {splitButtonItems}
+      <InnerToggle {...props} isSplitButton ariaLabel={props.ariaLabel || 'Select'} />
+    </div>
+  ) : (
+    <InnerToggle {...props} />
+  );
+
+// TODO add splitButtonItems prop
```
redallen: Well, you added it.

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.md
```diff
@@ -349,6 +349,142 @@ class IconDropdown extends React.Component {
...
+      <DropdownItem key="separated action" component="button">
+        Separated Action
+      </DropdownItem>
+    ];
+    return (
+      <Dropdown
+        onSelect={this.onSelect}
+        toggle={(
+          <DropdownToggle
+            splitButtonItems={[
```
redallen: Good choice :)

christiemolloy:
@mcarrano @jgiardino when a user clicks on the checkbox after the dropdown toggle has been clicked once, should the dropdown menu remain open?

redallen:
You also have 20 snapshots to update. Use `yarn test:pf4 -u`.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggle.js
mturley: oops, haha.

mturley:
Thanks @redallen. I also just noticed my handling of `ariaLabel` is a little weird (and caused the aria-label to disappear in a lot of those snapshots), I'm just going to standardize on `aria-label` and fix that.

mcarrano:
@christiemolloy I think you raise a good question. I don't feel strongly either way. It's definitely an edge case and am OK with the way it works now. @jgiardino what do you think?

mturley:
I really don't like how `InnerToggle` is being exposed in the snapshots and making them harder to read. I think I'll refactor slightly to remove the helper component. Feel free to let me know if you guys think I should revert this next commit.

mturley:
There, now the existing snapshots don't need to change at all :)

packages/patternfly-4/react-core/src/components/Dropdown/Dropdown.md
```diff
@@ -349,6 +349,142 @@ class IconDropdown extends React.Component {
...
+              <DropdownToggleCheckbox
```
tlabaj: I like that this is flexible for future use cases of the split button as described in the issue.

jgiardino:
I will not get a chance to review this today to confirm, but any instances of aria-label should be exposed to the consumer so that they can define their own text strings. The naming convention we're using is `aria-label` if the component represents a singular element (like Button), or something more specific if there are multiple elements in a component that can have an aria-label, like `ariaLabelCheckbox`.

mturley:
Yep, I made sure of that @jgiardino 😄 . The `aria-label` prop on `DropdownToggle` is optional and defaults to `"Select"` for split button toggles, or defaults to no aria-label for regular toggles since they have label text as a child. In both cases it can be specified by the consumer. The `aria-label` prop on `DropdownToggleCheckbox` is required (must be specified by the consumer), which matches what we were already doing for the regular `Checkbox`.

Good to know about the naming convention. I happened to get it right this time but I'll keep that in mind in the future.

jgiardino:
I never doubted you, @mturley! And thanks for the explanation! 
LGTM

PR #1851, opened 2019-04-25
feat(table): compound expansion

### LINKED ISSUE #1325: (Table): Compound Expansion
With new PF-next version there were some improvements to tables and one new variation called Compound Expansion. 

screenshot from 2019-02-05 11-13-10

Steps:
* Transformer for action (a with href and text) - consumer loaded
* Transformer for button in cell (tab) - consumer loaded
* Transformer for table as child - auto loaded if some prop defined (it will use same logic as is for rendering table)

priley86:
**What**:
Adds `compoundExpand` transformer and compound expansion example. Closes #1325 

- [x] add `compoundExpand` transformer
- [x] add tests
- [x] add nested table to the example

**Additional issues**:
none


packages/patternfly-4/react-table/src/components/Table/Table.js
```diff
@@ -30,6 +30,8 @@ const propTypes = {
   className: PropTypes.string,
   /** Function called when user wants to collapse row. */
   onCollapse: PropTypes.func,
+  /** Function called when user wants to compound expand row. */
+  onExpand: PropTypes.func,
```
dgutride: Please make sure to add new properties to the d.ts file for table

priley86: adding now...

christiemolloy:
Not sure if its part of this PR but the toggle arrows should face to the right by default.

Screen Shot 2019-04-26 at 10 56 02 AM

christiemolloy:
If its not too much effort, could you add in the other nested tables to these table cells so that we can see the interaction?

Screen Shot 2019-04-26 at 10 59 39 AM

priley86:
> Not sure if its part of this PR but the toggle arrows should face to the right by default.

I think the `vertical-align` on our React icon svg's is defaulting this way now. My vote would be to change this in separate issue (noted the same thing). It looks like we have `noVerticalAlign` prop, but i'm not sure that's intended on these.
https://github.com/patternfly/patternfly-react/pull/971

priley86:
> I'm looking at this and comparing the styling to the core example here: https://www.patternfly.org/v4/documentation/core/components/table#table-with-compound-expansion Looks like the blue highlight is not intended to extend to the row content itself. Just the expansion portion. @kybaker and/or @mceledonia can you also check the visuals on this?

i'm seeing this border/bg class only removes borders on the first `th` (and does not include `td` elements):
```
.pf-c-table tbody.pf-m-expanded>tr>:not(th):first-child:before {
    background-color: var(--pf-c-table__expandable-row--before--BackgroundColor);
}
```
The example [here](https://pf4.patternfly.org/components/Table/examples-full/?component=Table%20with%20Compound%20Expansion) has both `td` and `th` rendered in the row (and I've rendered a `td` instead of a `th` in the transform here):
```
<tbody class="pf-m-expanded">
    <tr>
      <th data-label="Repository Name">
                  <a href="#">siemur/test-space</a>
      </th>
      <td class="pf-c-table__compound-expansion-toggle pf-m-expanded" data-label="Branches">
        <button class="pf-c-button pf-m-link" aria-expanded="true" aria-controls="compound-expansion-table-nested-table-1">
                    <i class="fas fa-code-branch" aria-hidden="true"></i>          10
        </button>
      </td>
      <td class="pf-c-table__compound-expansion-toggle " data-label="Pull Requests">
        <button class="pf-c-button pf-m-link" aria-expanded="false" aria-controls="compound-expansion-table-nested-table-2">
                    <i class="fas fa-code" aria-hidden="true"></i>          4
        </button>
      </td>
``` 
It would be nice if those were consistent (`td` or `th`) or if we extend the class to apply to either I suppose.

christiemolloy:
@mcarrano I see what you're saying regarding visual designs in Sketch, but in every table example and the data list examples the blue border extends to the row that the new row was expanded from. I think if design decides against this we would have the capture this in a Core issue and it would require some updating.

jgiardino:
I'm starting to review this, and noticed something really odd, and am wondering if anyone else has noticed this or can reproduce this behavior?

Open the preview in safari. Scroll down the page to the Compound expansion example. Move the mouse cursor around over the table. This action causes the page to jump back to the top

Other than safari, I've only checked in Chrome, and am not seeing the same behavior there.

jgiardino:
Something else that's weird that's happening is when I hover over the first cell, the text will change position. I only see this behavior when the cell is taller due to wrapping contents in the row.

image

image

priley86:
> Open the preview in safari. Scroll down the page to the Compound expansion example. Move the mouse cursor around over the table. This action causes the page to jump back to the top

I don't know if this the same issue, but I'm seeing some weirdness when hovering the `caption` element in Safari here. This seems like a [bug in safari](https://bugs.webkit.org/show_bug.cgi?id=187903) 😸 - i am not applying any JS that would do this purposefully. Applying `display: block` to the `caption` seems to fix that, but it changes the default order and renders it inside the table then. Shall we remove captions on collapsed tables for now (seeing it in those examples too)?

jgiardino:
Great job on this @priley86!

The only inconsistency with core that I'm seeing so far is in core, we have `aria-controls` on the buttons that show the nested tables. This attribute works in JAWS, and allows the user to easily jump to the contents that are made visible. In past conversations about doing this for the regular expandable table in react, this wasn't possible and wasn't implemented in that case, either. I think the issue was that it wasn't possible to create or get the id of the nested table that we need to set it in `aria-controls`.

This is a minor issue, and not something that I'd consider necessary for this PR.

priley86:
> The only inconsistency with core that I'm seeing so far is in core, we have aria-controls on the buttons that show the nested tables.

there is different ways of achieving this (whether generating them and passing via [React Context](https://reactjs.org/docs/context.html)) or just passing as props down. I have elected to just add them as props to the column for now, and consumed them in the `compoundExpand` transform (that renders the `button`). This allows a user to set them explicitly.

@jgiardino let me know if this works for now!

jgiardino:
@priley86 You are a genius! I was not expecting you to address my comment. Now we need to get you to do this on the regular expandable table 😉 

bitmoji

PR #1858, opened 2019-04-25
feat(Table): Add columnTransforms, classNames decorator, Visibility constants, and hidden/visible breakpoints example

### LINKED ISSUE #1832: Add support for dropping columns in the Table
https://github.com/patternfly/patternfly-next/pull/1748

> | `.pf-m-hidden{-on-[breakpoint]}` | `.pf-c-table tr > *` | Hides a table cell at a given breakpoint, or hides it at all breakpoints with `.pf-m-hidden`. **Note: Needs to apply to all cells in the column you want to hide.** |
> | `.pf-m-visible{-on-[breakpoint]}` | `.pf-c-table tr > *` | Shows a table cell at a given breakpoint. |

This works with any of our global breakpoints. So you can use `.pf-m-hidden, .pf-m-hidden-on-[sm, md, lg, xl]` and the same for `visible`. You can use multiple classes on a single cell, too. For example `.pf-m-hidden` to hide the cell by default, then `.pf-m-visible-on-sm` to show it on `sm`, then `.pf-m-hide-on-md` to hide on `md`, and so on.

I suppose you should be able to add this to any cell in your table, as well as potentially supporting some way to say "hide the 3rd column on md breakpoint" and the react component applies the appropriate class to all the cells in that column?

mturley:
**What**: Closes #1832

This PR adds:
* A new property `columnTransforms` which can be included in a column definition (object in the `cells` array of a `Table`) with an array of reactabular transform functions which will apply to every cell in the column (header and body). This complements the existing properties `transforms` (only applies to header cells) and `cellTransforms` (only applies to body cells).
* A new decorator transform called `classNames` which just applies all of its arguments as classNames to the affected cells.
* A new export `Visibility` which includes all the modifiers for column visibility as properties.
* an example of using these three together to implement hidden/visible column breakpoints in a table.

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.md
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.md
@@ -16,6 +16,8 @@ import {
   expandable,
   cellWidth,
   textCenter,
+  classNames,
+  Visibility
 } from '@patternfly/react-table';
 
 ## Simple table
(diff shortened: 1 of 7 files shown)


priley86:
looking good!

packages/patternfly-4/react-table/src/components/Table/utils/decorators/classNames.js
```diff
@@ -0,0 +1,21 @@
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/patternfly/components/Table/table.css';
+
+const pickProperties = (object, properties) =>
+  properties.reduce((picked, property) => ({ ...picked, [property]: object[property] }), {});
```
redallen: Probably a better way to do (and standardize) this across more components than just table, but this works for now.

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -667,6 +669,66 @@ class WidthTable extends React.Component {
...
+} from '@patternfly/react-table';
+
+class HiddenVisibleBreakpointTable extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      columns: [
+        {
+          title: 'Repositories',
+          columnTransforms: [classNames(Visibility.hidden, Visibility.visibleOnMd, Visibility.hiddenOnLg)]
```
redallen: Love this interface.

packages/patternfly-4/react-table/src/components/Table/utils/decorators/classNames.js
mturley: Yeah I agree, I don't think there is a clean way to get a subset of an object's properties. I figured this was nicer than the repetition of my original version:

```
const {
  hidden,
  hiddenOnSm,
  hiddenOnMd,
  hiddenOnLg,
  hiddenOnXl,
  visibleOnSm,
  visibleOnMd,
  visibleOnLg,
  visibleOnXl
} = styles.modifiers;

export const Visibility = {
  hidden,
  hiddenOnSm,
  hiddenOnMd,
  hiddenOnLg,
  hiddenOnXl,
  visibleOnSm,
  visibleOnMd,
  visibleOnLg,
  visibleOnXl
};
```

packages/patternfly-4/react-table/src/components/Table/Table.md
mturley: Thanks, me too :)

PR #1859, opened 2019-04-25
feat(About modal): Added support for more user control over content

### LINKED ISSUE #1853: Add support for alerts in About Modal
I'm seeing a padding/styling issue when I use Alert inside an AboutModal:
Screen Shot 2019-04-23 at 4 30 30 PM

It's caused by .pf-c-content targeting the h4s. We need to use Alert like this in OpenShift.

I discussed this with patternfly-next in https://github.com/patternfly/patternfly-next/issues/1754. They suggest the easiest way to fix this is let the user specify whether they want their content wrapped in a .pf-c-content component. Right now the React About Modal puts children into a content component automatically.

rebeccaalpert:
Users can specify an optional prop that prevents the component from automatically nesting all content inside a  .pf-c-content div. Users can now use non-text components like Alert in the About modal.

I also fixed a messed up .d.ts file while I was in there (name and prop name were off).

Fixes #1853.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
@@ -33,6 +33,8 @@ const propTypes = {
     }
     return null;
   },
+  /** prevents AboutModal from rendering content inside AboutModalContentBox; allows for more flexible layouts */
+  noAboutModalBoxContentContainer: PropTypes.bool,
   /** Additional props are passed and spread to Modal content container <div> */
   '': PropTypes.any
 };
@@ -44,7 +46,8 @@ const defaultProps = {
   productName: '',
   trademark: '',
   logoImageSrc: '',
-  logoImageAlt: ''
+  logoImageAlt: '',
+  noAboutModalBoxContentContainer: false
 };
 
 let currentId = 0;
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
@@ -8,34 +8,34 @@ import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/patternfly/components/Backdrop/backdrop.css';
 
 const propTypes = {
-  /** content rendered inside the About Modal. */
+  /** Content rendered inside the about modal */
   children: PropTypes.node.isRequired,
-  /** additional classes added to the About Modal */
+  /** Additional classes added to the about modal */
   className: PropTypes.string,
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
```diff
@@ -33,6 +33,8 @@ const propTypes = {
     }
     return null;
   },
+  /** prevents AboutModal from rendering content inside AboutModalContentBox; allows for more flexible layouts */
```
tlabaj: Maybe update the description to sentence case.   I don't know  that AboutModalContentBox is meaningful to the consumer.

rebeccaalpert: Would "Prevents AboutModal from rendering content inside a container" be more meaningful?

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxContent.d.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export interface ModalBoxBodyProps extends HTMLProps<HTMLDivElement> {
+export interface AboutModalBoxContentProps extends HTMLProps<HTMLDivElement> {
   children: ReactNode;
-  tradeMark: string;
+  trademark: string;
   id: string;
 }
 
-declare const ModalBoxBody: FunctionComponent<ModalBoxBodyProps>;
```
dlabaj: Noticed a lot of the other d.ts files are not correct... Where you just ignoring the interfaces before?  How are you using this at the moment in typescript?

rebeccaalpert: We're using AboutModal in a .jsx file in OpenShift. I talked to @dgutride about these weird d.ts files and he said since they're not being exported, they're probably not breaking things yet.

rebeccaalpert:
@tlabaj - I updated the prop description.

christiemolloy:
About Modal doesn't scroll horizontally I think we should add that feature in @mcoker 

Screen Shot 2019-04-26 at 2 59 40 PM

tlabaj:
@christiemolloy will that be a separate issue?

rebeccaalpert:
@mcoker and @christiemolloy I swapped the class.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxContent.d.ts
dlabaj: Thanks!

Issue #1673, opened 2019-04-01
React-charts Axes Hidden


redallen:
The axes are hidden on [a few charts](https://1656-pr-patternfly-react-patternfly.surge.sh/patternfly-4/charts/chartbar/) that need the following styles applied:
```
svg {
  overflow: 'visible'; }
```

We should apply these styles for the consumer instead of making them define their own CSS.

rachael-phillips:
HI @redallen would you consider this a bug or an enhancement?

redallen:
Considering it's also an [upstream issue](https://github.com/FormidableLabs/victory/issues/991) if you structure the <Victory> components the way we do, I'd say so. We should be applying all styling for the consumer so they don't have to worry about this.

dlabrecq:
This has been fixed by adjusting the default padding

PR #1862, opened 2019-04-26
fix(charts): Fix chart themes to prevent axis labels from overflowing parent

### LINKED ISSUE #1819: PF4 chart axis ticks and labels overflow parent
Chart axis tick labels are rendered outside of the parent on the left and bottom. This can be seen in the line and bar chart examples. Created similar vanilla victory charts in OpenShift and the labels were rendered correctly inside the parent element. 

## With PF React Charts
```jsx
<Chart>
  <ChartArea data={data} />
</Chart>
```
image

## With Victory
```jsx
<VictoryChart>
  <VictoryArea data={data} />
</VictoryChart>
```
image

TheRealJon:
Fixes #1819

**What**:

Removed padding prop from the default layout and updated axis stroke color. Victory's default padding allows room for axis tick labels. The color for axis stroke will be included in PF core variables soon, but for now, just hardcoded the hex value.

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/Chart/__snapshots__/Chart.test.js.snap
+++ b/packages/patternfly-4/react-charts/src/components/Chart/__snapshots__/Chart.test.js.snap
@@ -2156,7 +2156,6 @@ exports[`Chart 1`] = `
           "#0066ff",
         ],
         "height": 301,
-        "padding": 8,
         "style": Object {
           "data": Object {
             "fill": "#bee1f4",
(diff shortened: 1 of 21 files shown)


dlabrecq:
FYI, I tried these changes with the Cost management UI and the charts won't render. I'll try again, just to make certain, tho.

Screen Shot 2019-04-26 at 12 27 12 PM

dlabrecq:
Note that the charts would not render while testing in Cost management. I needed to update the hoist-non-react-statics package to the latest version.

Screen Shot 2019-04-26 at 1 26 38 PM

redallen:
Also fixes #1673

TheRealJon:
If we adjust the padding, I think we also need to change the default axis styles to not be hidden.

dlabrecq:
We can adjust the padding, but can't remove all the theme properties for the Chart component, which is why this PR is blocked. I anticipate the padding will be adjusted (per our offline discussions) instead to show axes by default.

tlabaj:
@TheRealJon can you please resolve conflicts.

TheRealJon:
@tlabaj @dlabrecq Rebased on master and resolved conflicts

dlabaj:
@TheRealJon These fixes are in PR #1964 we are going to close this one.

PR #1865, opened 2019-04-26
fix(charts): Fix chart font stack to use PatternFly core sans-serif font stack

### LINKED ISSUE #1864: Chart fonts should use PatternFly core sans serif font stack
Currently, the font stack used for charts is not in line with PF designs. It should be switched to the font stack defined in pf core.

TheRealJon:
Fixes #1864

**What**:
Use `global_FontFamily_sans_serif` variable for chart font stack. Also, correct misspelled 'TYPOGRAPHY' related variable names.


TheRealJon:
cc @kybaker

PR #1867, opened 2019-04-28
fix(Select): fix ListGroupItem import

glekner:
**What**:
Select doesn't work downstream, due to a ListGroupItem import error, this PR tries to fix it.


glekner:
@jeff-phillips-18 thoughts on this? is this the right solution?

dgutride:
@glekner  - please open an issue for this and we'll take it in as soon as we have that

jeff-phillips-18:
This fix is correct given the current build. We should find a better way to build so that we can import with the previous syntax but currently this is how it needs to be.

PR #1420, opened 2019-02-21
[WIP]feat(component): Added Date Time Picker component

MariaAga:
**What**: Add the DateTimePicker component from PF to PF-react.

The PatternFly design is at:https://www.patternfly.org/pattern-library/forms-and-controls/date-and-time/#design

Currently there is only a Date Picker

PR #1875, opened 2019-04-29
feat(component): Add Time Picker component 

MariaAga:
Split the time picker from:
https://github.com/patternfly/patternfly-react/pull/1420


packages/patternfly-3/patternfly-react/sass/patternfly-react/_date-time-picker.scss
```diff
@@ -0,0 +1,74 @@
...
+.bootstrap-datetimepicker-widget.date-time-picker-pf.bottom {
+  margin-top: 10px !important;
+}
+
+.bootstrap-datetimepicker-widget.date-picker-pf,
+.bootstrap-datetimepicker-widget.date-time-picker-pf {
+  max-width: none;
+
+  .arrow {
+    left: 10% !important;
```
Laviro: please try not to use `!important` boith here and also in other places, 
probably we'll need to modify patternfly-core css to make this component more adjustable.

packages/patternfly-3/patternfly-react/sass/patternfly-react/_patternfly-react.scss
```diff
@@ -18,3 +18,4 @@
 @import 'login-page';
 @import 'dual-list-selector';
 @import 'breadcrumbswitcher-popover';
+@import 'date-time-picker';
```
Laviro: nit: add a newline after this one

packages/patternfly-3/patternfly-react/less/patternfly-react.less
```diff
@@ -18,3 +18,4 @@
 @import 'expand-collapse';
 @import 'login-page';
 @import 'dual-list-selector';
+@import 'date-time-picker';
```
Laviro: nit: add a newline

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimeComponents/PickTimeClock.js
```diff
@@ -0,0 +1,105 @@
...
+    ampm: this.props.time.getHours() >= 12 ? 'PM' : 'AM'
+  };
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return {
+      ampm: nextProps.time.getHours() >= 12 ? 'PM' : 'AM'
+    };
+  }
+  setTime = (type, amount) => {
+    const { time } = this.props;
+    if (type === 'H') {
```
Laviro: can you create constants for these? 
you can call them HOUR, MINUTE etc - it would be more readable.

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimeComponents/PickTimeClock.js
```diff
@@ -0,0 +1,105 @@
...
+      time.setHours(time.getHours() + amount);
+    } else if (type === 'M') {
+      time.setMinutes(time.getMinutes() + amount);
+    }
+    this.props.setSelected(time);
```
Laviro: can you add it to the destruction at the top of the method? 
I saw it also in other places over the code.

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
```diff
@@ -0,0 +1,103 @@
...
+    return new Date(`1/1/1 ${time}`);
+  };
+
+  formatDate = time => {
+    const { locale } = this.props;
+    const options = { hour: '2-digit', minute: '2-digit' };
+    if (Date.parse(time)) {
+      const parsedValue = new Date(time);
+      return parsedValue.toLocaleString(locale, options);
+    } else if (Date.parse(`1/1/1 ${time}`)) {
```
Laviro: what does it mean ?

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
```diff
@@ -0,0 +1,103 @@
...
+      const parsedValue = new Date(`1/1/1 ${time}`);
+      return parsedValue.toLocaleString(locale, options);
+    }
+    return time;
+  };
+  state = {
```
Laviro: can you move it to the top of the file?

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
```diff
@@ -0,0 +1,103 @@
...
+            onClick={() => this.setState({ hiddenValue: !hiddenValue })}
+          />
+        </InputGroup.Addon>
+      </InputGroup>
+    );
+  }
+}
+
+TimePicker.propTypes = {
+  value: PropTypes.oneOfType([PropTypes.instanceOf(Date), PropTypes.string]),
```
Laviro: can you add documentation above all of the propTypes so it would be visible in the storybook ?

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
MariaAga: The date object in JavaScript requires a date to create a new object, and this function is used to we can use the Time Picker with time value only like this:
`<TimePicker value="2:22:31 PM" />`
Without this function the component will have to always get a date with the time:
`<TimePicker value="1/1/1 2:22:31 PM" />`

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimeComponents/PickTimeClock.js
MariaAga: Added consts HOUR, MINUTE, AM ,PM. thanks!

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
MariaAga: moved the state, and moved the functions into a helper file

packages/patternfly-3/patternfly-react/sass/patternfly-react/_date-time-picker.scss
MariaAga: The arrow position is set in react-bootstrap's component and can't be changed in a different way

Laviro: maybe you can open an issue or PR to `react-bootstrap` and link it here?

MariaAga: its fixed in the newer versions of react-bootstrap

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
```diff
@@ -0,0 +1,86 @@
...
+    const newTime = getDateFromTime(time);
+    this.setState({ value: newTime, tmpValue: formatTime(newTime, this.props.locale) });
+  };
+  render() {
+    const { value, hiddenValue } = this.state;
+    const { locale, placement, id } = this.props;
+    const popover = (
+      <Popover
+        id={id}
+        className="bootstrap-datetimepicker-widget date-time-picker-pf bootstrap-timepicker-widget  dropdown-menu"
```
jeff-phillips-18:
```suggestion
        className="bootstrap-datetimepicker-widget date-time-picker-pf bootstrap-timepicker-widget"
```
Adding the `dropdown-menu` class is causing the popover to be incorrectly placed at absolute left 0. I don't believe we need that class for any other reason.

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
```diff
@@ -0,0 +1,86 @@
...
+        </ul>
+      </Popover>
+    );
+
+    return (
+      <InputGroup className="input-group date-time-picker-pf time-picker-pf">
+        <FormControl
+          aria-label="time-picker-input"
+          type="text"
+          value={hiddenValue ? '' : this.state.tmpValue}
```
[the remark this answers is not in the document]
MariaAga: Yes, this component is already used in foreman.
When the picker is used inside a form, there needs to be an option for an empty value for submitting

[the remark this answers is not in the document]
jeff-phillips-18: So then it should just be cleared right? no need to prevent re-entry or hiding the actual current value. Why have the button be a toggle? Clear if the X is pressed, disable the X if there is no entry. @mcarrano Can you take a look and make any suggestions from a UX perspective?

At a minimum this functionality should be optional with the default being to not having this functionality. I should be able to clear the text in order to unset the current time.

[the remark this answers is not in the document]
mcarrano: I agree with @jeff-phillips-18 here. The current button that toggles between clear and edit is confusing and unexpected to me. The user should be able to just clear the field to submit a null value. In PatternFly 4, we are specifying a clear button inside the field. See https://github.com/patternfly/patternfly-next/issues/180  If you want to do something equivalent, that would be acceptable @MariaAga .

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
MariaAga: removed. now the popover is above/under the clock icon and the arrow is pointing at the clock icon

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/TimePicker.js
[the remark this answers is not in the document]
MariaAga: Thanks for the feedback!
I've removed the edit/clear button and now the input field can be empty (by deleting the value). Is this a better functionality? @mcarrano

mcarrano:
Looks good @MariaAga . I agree that the use can just delete the text if they want to clear the time value.

packages/patternfly-3/patternfly-react/less/date-time-picker.less
```diff
@@ -0,0 +1,66 @@
...
+      height: initial;
+      line-height: initial;
+    }
+  }
+
+  .timepicker-hours .table-condensed > tbody > tr > td,
+  .timepicker-minutes .table-condensed > tbody > tr > td {
+    padding: 7px 0;
+  }
+}
```
jeff-phillips-18: Looks like this needs to be updated to match the sass version

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/DateTimePicker.stories.js
```diff
@@ -0,0 +1,32 @@
...
+import { select, withKnobs } from '@storybook/addon-knobs';
+import { name } from '../../../package.json';
+import { TimePicker } from './index';
+
+const stories = storiesOf(
+  `${storybookPackageName(name)}/${STORYBOOK_CATEGORY.FORMS_AND_CONTROLS}/Date Time Picker`,
+  module
+);
+stories.addDecorator(withKnobs);
+stories.addWithInfo('DateTimePicker', '', () => {
```
jeff-phillips-18:
```suggestion
stories.addWithInfo('TimePicker', '', () => {
```

packages/patternfly-3/patternfly-react/src/components/DateTimePicker/DateTimePicker.stories.js
```diff
@@ -0,0 +1,32 @@
...
+      <label>Time picker</label>
+      <div className="row">
+        <div className="col-md-3">
+          <TimePicker placement={select('Placement Time', ['top', 'bottom'], 'top')} />
+        </div>
+      </div>
+    </div>
+  );
+  return inlineTemplate({
+    title: 'DateTimePicker',
```
jeff-phillips-18:
```suggestion
    title: 'TimePicker',
```

packages/patternfly-3/patternfly-react/less/date-time-picker.less
MariaAga: Yep, thanks!

Issue #1871, opened 2019-04-29
[PF4] 3.16.0 not work bug in ClipboardCopy component


aljesusg:
node_modules/@patternfly/react-core/dist/js/components/ClipboardCopy/ClipboardCopy.d.ts
(1,63): Module '"../../../../../../@types/react"' has no exported member 'HTMLDivElement'.

- Remove HTMLDivElement from import

```typescript
import { FunctionComponent, HTMLProps, MouseEvent, ReactNode, HTMLDivElement } from 'react';
```
fix
```typescript
import { FunctionComponent, HTMLProps, MouseEvent, ReactNode } from 'react';
```
- Wrong import typeUtils
```typescript
import { OneOf } from '../../typeUtils';
```
fix
```typescript
import { OneOf } from '../../helpers/typeUtils';
```
- 5,37): A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference.
```typescript
export const ClipboardCopyVariant = {
  inline: 'inline',
  expansion: 'expansion'
};
```
fix
```typescript
export enum ClipboardCopyVariant {
  inline = 'inline',
  expansion = 'expansion'
}
```
- onChange is not used  in ClipboardCopy.js
- copyButtonIdPrefix is not used in ClipboardCopy.js
- onCopy method isn't same type

```bash
(10,18): Interface 'ClipboardCopyProps' incorrectly extends interface 'HTMLProps<HTMLDivElement>'.
  Types of property 'onCopy' are incompatible.
    Type '((event: MouseEvent<Element>, text?: string | undefined) => void) | undefined' is not assignable to type '((event: ClipboardEvent<HTMLDivElement>) => void) | undefined'.
      Type '(event: MouseEvent<Element>, text?: string | undefined) => void' is not assignable to type '(event: ClipboardEvent<HTMLDivElement>) => void'.
        Types of parameters 'event' and 'event' are incompatible.
          Type 'ClipboardEvent<HTMLDivElement>' is missing the following properties from type 'MouseEvent<Element>': altKey, button, buttons, clientX, and 10 more.

```

aljesusg:
cc @LHinson

tlabaj:
closed by #1876

PR #1876, opened 2019-04-29
fix(ClipboardCopy): update imports and clean up

nicolethoen:
Fixes issue [#1871](https://github.com/patternfly/patternfly-react/issues/1871)

Note: the issue states that the onChange prop is not used in ClipboardCopy.js but it is used on line 46. It is not used in the render function.

Tested by adding component to index.tsx on patternfly-react-seed (locally)

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.d.ts
+++ b/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.d.ts
@@ -1,16 +1,18 @@
-import { FunctionComponent, HTMLProps, MouseEvent, ReactNode, HTMLDivElement } from 'react';
-import { OneOf } from '../../typeUtils';
+import { ClipboardEvent, FunctionComponent, HTMLProps, ReactNode } from 'react';
+import { OneOf, Omit } from '../../helpers/typeUtils';
 import { PopoverPosition } from '../Popover';
 
-export const ClipboardCopyVariant = {
-  inline: 'inline',
-  expansion: 'expansion'
-};
+export enum ClipboardCopyVariant {
+  inline = 'inline',
+  expansion = 'expansion'
+}
 
-export interface ClipboardCopyProps extends HTMLProps<HTMLDivElement> {
+export interface ClipboardCopyProps extends Omit<HTMLProps<HTMLDivElement>, 'onChange'> {
   className?: string;
   hoverTip?: string;
   clickTip?: string;
+  textAriaLabel?: string;
+  toggleAriaLabel?: string;
   isReadOnly?: boolean;
   variant?: OneOf<typeof ClipboardCopyVariant, keyof typeof ClipboardCopyVariant>;
   position?: OneOf<typeof PopoverPosition, keyof typeof PopoverPosition>;
@@ -18,8 +20,8 @@ export interface ClipboardCopyProps extends HTMLProps<HTMLDivElement> {
   exitDelay?: number;
   entryDelay?: number;
   switchDelay?: number;
-  onCopy?: (event: MouseEvent, text?: string) => void;
-  onChange?: (text: string) => void;
+  onCopy?: (event: ClipboardEvent<HTMLDivElement>, text?: string) => void;
+  onChange?: (text?: string) => void;
   children?: ReactNode;
 }
 
--- a/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.js
+++ b/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.js
@@ -56,18 +56,16 @@ class ClipboardCopy extends React.Component {
       onCopy,
       hoverTip,
       clickTip,
-      'aria-label': ariaLabel,
-      'toggle-aria-label': toggleAriaLabel,
+      textAriaLabel,
+      toggleAriaLabel,
       variant,
       position,
       className,
-      onChange,
       ...props
     } = this.props;
     const textIdPrefix = 'text-input-';
     const toggleIdPrefix = 'toggle-';
     const contentIdPrefix = 'content-';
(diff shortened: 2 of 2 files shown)


packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.d.ts
```diff
@@ -1,16 +1,18 @@
...
-import { OneOf } from '../../typeUtils';
+import { FunctionComponent, HTMLProps, MouseEvent, ReactNode } from 'react';
+import { OneOf } from '../../helpers/typeUtils';
 import { PopoverPosition } from '../Popover';
 
-export const ClipboardCopyVariant = {
-  inline: 'inline',
-  expansion: 'expansion'
-};
+export enum ClipboardCopyVariant {
```
tlabaj: Why are we making this an enum?  We probably want to be consistent with other components that use a const.

nicolethoen: The issue cited this reason: "A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference."

dlabaj: Can you verify that we are seeing this error in our typescript app?  I'm not sure we should make this change.

aljesusg: There is a enum here https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Title/Title.tsx#L14

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.d.ts
```diff
@@ -1,16 +1,18 @@
...
+  inline = 'inline',
+  expansion = 'expansion'
+}
 
 export interface ClipboardCopyProps extends HTMLProps<HTMLDivElement> {
   className?: string;
   hoverTip?: string;
   clickTip?: string;
+  'aria-label'?: string;
+  'toggle-aria-label'?: string;
```
tlabaj: Since we are still prerelease, do you mind changing this prop here and in the JS file to `toggleAriaLabel`.  This is the naming convention we have been using for non HTML aria attributes.

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.d.ts
dlabaj: Trying to figure out exactly why it's throwing this error.  We use constants in other places, so wondering if you will see this error again.

priley86: you can also set this to:
```
export const ClipboardCopyVariant : {
  inline = 'inline',
  expansion = 'expansion'
}
```
to resolve this

priley86: the `=` throws it off for some reason

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.d.ts
```diff
@@ -1,16 +1,18 @@
...
   isReadOnly?: boolean;
   variant?: OneOf<typeof ClipboardCopyVariant, keyof typeof ClipboardCopyVariant>;
```
priley86: seeing a couple more creep in here now in my IDE... i would be up for:

```
import { ClipboardEvent, FormEvent, FunctionComponent, HTMLProps, MouseEvent, ReactNode } from 'react';

export interface ClipboardCopyProps extends HTMLProps<HTMLDivElement> {
   ...
  onCopy?: (event: ClipboardEvent<HTMLDivElement>, text?: string) => void;
  onChange?: (event: FormEvent<HTMLDivElement>) => void;
```

or something else similar...

priley86: or just remove them for now 😆

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.js
```diff
@@ -140,9 +138,9 @@ ClipboardCopy.propTypes = {
   /** Tooltip message to display when clicking the copy button */
   clickTip: PropTypes.string,
   /** Custom flag to show that the input requires an associated id or aria-label. */
-  'aria-label': PropTypes.string,
+  textAriaLabel: PropTypes.string,
```
redallen: Isn't this technically a breaking change? How many people are using `aria-label` or `toggle-aria-label` on ClipboardCopy out in the wild?

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.js
```diff
@@ -56,18 +56,16 @@ class ClipboardCopy extends React.Component {
       onCopy,
       hoverTip,
       clickTip,
-      'aria-label': ariaLabel,
-      'toggle-aria-label': toggleAriaLabel,
+      textAriaLabel,
```
redallen: Pretty!

packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.js
nicolethoen: I believe we are at a point where it is okay to introduce the breaking changes? Is that correct @tlabaj?

tlabaj: This component is new and has not been released yet.  Since we are pre release the change is ok.

PR #1879, opened 2019-04-29
feat(VictoryZoomContainer): Give examples of Victory Zoom component to allow zo…

### LINKED ISSUE #1715: Introduce Victory Charting Brush and Zoom
There is a need for users to be able to zoom in on line charts. I imagine we will leverage Victory charting again for this.... https://formidable.com/open-source/victory/guides/brush-and-zoom/

jcaianirh:
…om capability on charts.

fix #1715

**What**: Add examples on how to use VictoryZoomContainer. Add this component to the highest level Chart component as the containerComponent property to create a zoom-able chart.


packages/patternfly-4/react-charts/src/components/ChartZoomContainer/ChartZoomContainer.js
```diff
@@ -0,0 +1,28 @@
...
+
+  render() {
+    return (
+      <VictoryZoomContainer {...this.props} />);
+  }
+}
+// Note: VictoryZoomContainer.defaultEvents must be hoisted
+hoistNonReactStatics(ChartZoomContainer, VictoryZoomContainer);
+
+export default ChartZoomContainer;
```
dlabrecq: I don't believe we should extend Victory classes.

"We should not be extending these. It is possible that they could internally remove clasess, etc. A better, and more "React" way, would be to do something like" -- @dmiller9911

```jsx
const ChartArea = (props) => <VictoryArea {...props} />

ChartArea.propTypes = propTypes;
ChartArea.defaultProps = {
   ...VictoryArea.defaultProps,
   theme: ChartTheme.default,
}
```

Another option is to do something like
```jsx
const ChartArea = (props) => <VictoryArea them={ChartTheme.default} {...props} />
```
That would pass the default theme down, and allow the user to overwrite it with a theme prop.

_Originally posted by @dmiller9911 in https://github.com/patternfly/patternfly-react/pull/868_

packages/patternfly-4/react-charts/src/components/ChartZoomContainer/ChartZoomContainer.js
```diff
@@ -0,0 +1,28 @@
...
+  static propTypes = {
+    ...VictoryZoomContainer.propTypes
+  };
+
+  renderChildren() {
+    const children = React.Children.toArray(this.props.children);
+    return children.map((child) => {
+      // Some victory components expect to control props like domain for
+      // children, some props should be merged rather than overridden
+      const style = _.merge(child.props.style, this.props.style);
```
redallen: Where is lodash coming from? Shouldn't we explicitly add the dependency?

packages/patternfly-4/react-charts/src/components/ChartZoomContainer/ChartZoomContainer.js
jcaianirh: @dlabrecq i don't think you're seeing the issue with the renderChildren. The passing props is not the problem with wrapping the ZoomContainer, its overriding the renderChildren function that provides provides the missing properties I was telling you about, and its the way that Victory recommends wrapping its components. The other components that do not need to manage child props can be wrapped the way you suggest above. Let me know what you think? See this also: https://formidable.com/open-source/victory/guides/custom-components/#wrapping-components

dlabrecq: The Victory example above shows extending React.Component, but the code is extending VictoryZoomContainer, which is my issue.

packages/patternfly-4/react-charts/src/components/ChartZoomContainer/ChartZoomContainer.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import hoistNonReactStatics from 'hoist-non-react-statics';
+import { VictoryZoomContainer } from 'victory';
+import { merge } from 'lodash';
```
dlabrecq: Where are we picking up the lodash package? I don't see it in the react-core dependencies.

dgutride: @jcaianirh  - I don't believe we currently use lodash in react-charts.  We generally prefer pure es6/typescript methods instead of shelling out to a library for things like this - can you try to use native functionality instead?

jcaianirh:
@dlabrecq got some assistance from victory and found that creating a static role set to "container" as well as hoisting the default events done by hostNonReactStatics, we can achieve a wrapped Zoom component. Added a couple examples and updated the code to reflect the above info. Let me know if this is more or less desirable than just using the Victory component without wrapping it.

jcaianirh:
@dlabrecq removed the ChartZoomController as we discussed and provided examples using VictoryZoomController for Bar, Line, and Stack charts.

jcaianirh:
@dlabrecq I added the addZoom prop to the pf Chart component, and updated the examples to use the 'convenience' allowZoom prop to show how to use it as well as provide the more complex example using the containerComponent.

jcaianirh:
@dlabrecq at first I thought no, but now that I look at victory, they do provide an example of using ChartGroup without Chart, so I'm gonna add it in.

jcaianirh:
@dlabrecq added zoom prop to chart group

jcaianirh:
@dlabrecq updated for tsx conversion. please review

PR #1889, opened 2019-04-30
Add example styles for charts and alert

dgutride:
Also fixes the code example for the docs site
addresses some of:  https://github.com/patternfly/patternfly-react/issues/1724


packages/patternfly-4/react-charts/src/components/ChartArea/examples/chart-area.scss
```diff
@@ -0,0 +1,14 @@
+.ws-preview {
+  & > * {
+    .chart-overflow {
+      & svg {
+        overflow: visible;
+      }
+    }
+
+    .chart-title {
+      margin-bottom: 25px;
```
dlabaj: Should we be using SASS variables for things that are common in overflow, title, container, margin, etc?

dgutride: this was just ported from previous examples and deleted, we can refactor later if you'd like

dlabaj: Okay we can do it later thought this was something new.  Thanks.

PR #1899, opened 2019-04-30
fix(types): Update @types/victory, victory, victory-core, and hoist-non-react-statics packages to latest

### LINKED ISSUE #1898: TypeScript Error for ChartLegend Component
The `@types/victory` package is outdated and is causing typscript errors downstream.

TheRealJon:
fixes #1898

**What**:
Update `@types/victory`, `victory`, `victory-core`, and `hoist-non-react-statics` to latest and remove extra type definition for the `title` prop in `ChartLegend`. The latest victory types include an accurate definition for this prop.

Victory changelog for reference:
https://github.com/FormidableLabs/victory/blob/master/CHANGELOG.md

### DIFF
--- a/packages/patternfly-4/react-charts/package.json
+++ b/packages/patternfly-4/react-charts/package.json
@@ -41,10 +41,10 @@
     "develop": "yarn build:babel:esm --skip-initial-build --watch --verbose"
   },
   "optionalDependencies": {
-    "@types/victory": "^0.9.19",
-    "hoist-non-react-statics": "^3.1.0",
-    "victory": "^30.1.0",
-    "victory-core": "^31.1.0"
+    "@types/victory": "^31.0.15",
+    "hoist-non-react-statics": "^3.3.0",
+    "victory": "^32.2.0",
+    "victory-core": "^32.2.0"
   },
   "devDependencies": {
     "@patternfly/patternfly": "2.6.5",
--- a/packages/patternfly-4/react-charts/src/components/Chart/__snapshots__/Chart.test.js.snap
+++ b/packages/patternfly-4/react-charts/src/components/Chart/__snapshots__/Chart.test.js.snap
@@ -49,7 +49,6 @@ exports[`Chart 1`] = `
             role="presentation"
           />
         }
-        scale="linear"
         standalone={true}
         theme={
           Object {
@@ -76,7 +75,6 @@ exports[`Chart 1`] = `
                   "letterSpacing": "normal",
                   "padding": 10,
                   "stroke": "transparent",
-                  "textAnchor": "middle",
                 },
               },
               "width": 450,
@@ -268,7 +266,6 @@ exports[`Chart 1`] = `
                   "letterSpacing": "normal",
                   "padding": 10,
                   "stroke": "transparent",
-                  "textAnchor": "middle",
                 },
               },
               "width": 450,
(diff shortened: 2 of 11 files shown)


dlabrecq:
The title prop can be removed, but believe we need to update all 3 victory packages to ensure types are is in sync.

packages/patternfly-4/react-charts/package.json
```diff
@@ -41,10 +41,10 @@
     "develop": "yarn build:babel:esm --skip-initial-build --watch --verbose"
   },
   "optionalDependencies": {
-    "@types/victory": "^0.9.19",
+    "@types/victory": "^31.0.15",
     "hoist-non-react-statics": "^3.1.0",
```
dlabrecq: Can you please update hoist-non-react-statics? Cost Management will not run without this also being updated.

Screen Shot 2019-05-01 at 11 02 05 AM

dlabrecq:
Discovered that building react-charts with the latest Victory packages destroyed the bullet charts we created in Cost Management. I'll try to find a solution, but this will be a breaking change for Cost Management with no workaround.

Screen Shot 2019-05-01 at 11 13 38 AM

packages/patternfly-4/react-charts/package.json
TheRealJon: Updated

TheRealJon:
@dlabrecq 
> Discovered that building react-charts with the latest Victory packages destroyed the bullet charts we created in Cost Management. I'll try to find a solution, but this will be a breaking change for Cost Management with no workaround.
> 

See the following in the changelog:

> Breaking Changes
>
> Most Horizontal Charts The change in how props with x and y values are treated (i.e. scale, domain, etc) will be a breaking change for most horizontal charts. In most cases, reversing the x and y values of props you are providing to your horizontal chart will be sufficient to correct the change. For example:
>
```jsx
<VictoryChart horizontal scale={{ x: "log" }} domain={{ y: [4, 9] }}>
  <VictoryBar
    data={[
      { x: 5, y: 0.1 },
      { x: 6, y: 1 },
      { x: 7, y: 10 },
      { x: 8, y: 100 }
    ]}
  />
</VictoryChart>
```
>
> Should be changed to:
>
```jsx
<VictoryChart horizontal scale={{ y: "log" }} domain={{ x: [4, 9] }}>
  <VictoryBar
    data={[
      { x: 5, y: 0.1 },
      { x: 6, y: 1 },
      { x: 7, y: 10 },
      { x: 8, y: 100 }
    ]}
  />
</VictoryChart>
```
>
> Props affected by this change are: scale, domain, maxDomain, minDomain, domainPadding, and categories
>
> Horizontal Charts with Event Containers Dimension props such as brushDimension have changed so that they always refer to the dimension of the target variable (x for the independent variable, and y for the dependent variable). For example, a VictoryBrushContainer component with brushDimension="x" will move and expand only in the independent dimension regardless of whether the chart is horizontal.
>
> Props affected by this change are: brushDimension, cursorDimension, selectionDimension, voronoiDimension, and zoomDimension
>
> Horizontal Charts with Custom dataComponents The position values (i.e. x, y, x0, y0) supplied to custom dataComponents from components like VictoryChart will be scaled for the layout of the horizontal chart. Users who rely on these values may need to flip them or adjust them depending on their use case
>
> Horizontal VictoryBoxPlot Previously VictoryBoxPlot required data to be flipped (x values flipped with y values) in order to plot horizontal charts. This is no longer required, and providing data in this format will no longer work correctly. To correct for this change, it should be sufficient to flip the data used in horizontal charts

PR #1901, opened 2019-05-01
fix(DualList): add callbacks to Controlled Dual List methods

Laviro:
Add callbacks of the updated state for each method in the controlled Dual List


packages/patternfly-3/patternfly-react/src/components/DualListSelector/DualListControlled.js
```diff
@@ -85,4 +97,46 @@ class DualListControlled extends React.Component {
   }
 }
 
+DualListControlled.propTypes = {
+  /**
+   * Function that runs after items have been moved between the lists.
+   * Recieves the updated state as a callback.
```
jeff-phillips-18: nit: typo -> receives
(below as well)

Laviro:
Thanks, fixed :)

Laviro:
can we merge it? :)

PR #1911, opened 2019-05-02
fix(DualList): Distinguish between default hidden to filter hidden

Laviro:
Items that were default set to hidden, shouldn't be modified by the component.


packages/patternfly-3/patternfly-react/src/components/DualListSelector/helpers.js
```diff
@@ -4,32 +4,33 @@ export const filterByHiding = (list, value) => {
...
+        item.children = makeAllHiddenFilterredItemsVisible(item.children);
       }
     }
     return item;
   });
 };
 
-export const makeAllItemsVisible = list =>
+// We don't want to touch default hidden items that the consumer set to hidden.
+export const makeAllHiddenFilterredItemsVisible = list =>
```
jeff-phillips-18:
```suggestion
export const makeAllHiddenFilteredItemsVisible = list =>
```

Laviro:
thanks, fixed that :)

PR #1914, opened 2019-05-02
feat(TextInput): Convert text input to TypeScript

### LINKED ISSUE #1908: Convert TextInput to TypeScript
Like it says on the tin. Leave the examples alone for now.

rebeccaalpert:
Fixes #1908.


nicolethoen:

rebeccaalpert:
Added demo and tests.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+    isValid: true,
+    isDisabled: false,
+    isReadOnly: false,
+    type: 'text'
+  }
+
+  constructor(props: TextInputProps) {
+    super(props);
+    if (!props.id && !props['aria-label'] && !props['aria-labelledby']) {
+      console.error('Text input:', 'Text input requires either an id or aria-label to be specified');
```
redallen: Add `// eslint-disable-next-line no-console` on the line before

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export class TextInput extends React.Component<TextInputProps> {
+  static defaultProps = {
+    className: '',
+    isRequired: false,
+    isValid: true,
+    isDisabled: false,
+    isReadOnly: false,
+    type: 'text'
```
redallen: Add `onChange: () => undefined as any`

redallen: Add `'aria-label': null as string`

rebeccaalpert: @dlabaj said I should drop these in a similar situation another PR, so I'm going to skip these for now unless I hear otherwise.

rebeccaalpert:
Updated enum typing as discussed in meeting.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+  month = 'month',
+  number = 'number',
+  password = 'password',
+  search = 'search',
+  tel = 'tel',
+  time = 'time',
+  url = 'url'
+}
+
+export interface TextInputProps extends Omit<React.HTMLProps<HTMLInputElement>, 'onChange'> {
```
dlabaj: I think you missed | 'disabled'.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** Flag to show if the input is valid or invalid. */
+  isValid?: boolean;
+  /** A callback for when the input value changes. */
+  onChange?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
+  /** Type that the input accepts. */
+  type?: 'text' | 'date' | 'datetime-local' | 'email' | 'month' | 'number' | 'password' | 'search' | 'tel' | 'time' | 'url';
+  /** Value of the input. */
+  value?: string | number;
+  /** Aria-label. The input requires an associated id or aria-label. */
+  'aria-label'?: string;
```
dlabaj: I believe we still want this to default to null for now.  We have a future story to change these to undefined.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
dlabaj: You should keep 'aria-label' undefined for now because of the api.  onChange would be nice if we don't have to do onChange undefined.  @redallen do we need this for a particular reason?

packages/patternfly-4/react-integration/cypress/integration/textinputdemo.spec.ts
```diff
@@ -0,0 +1,23 @@
+describe('Text Input Demo Test', () => {
+  it('Navigate to demo section', () => {
+    cy.visit('http://localhost:3000/');
+    cy.get('#text-input-demo-nav-item-link').click();
+    cy.url().should('eq', 'http://localhost:3000/text-input-demo-nav-link')
+  });
+
+  it('Verify text input', () => {
+    cy.get('#text').type('Hello world');
```
dlabaj: can you add a verification step here to verify it reads Hello world

packages/patternfly-4/react-table/src/components/Table/__snapshots__/Table.test.js.snap
```diff
@@ -3051,7 +3051,6 @@ exports[`Actions table 1`] = `
                                   </Item>,
                                 ]
                               }
-                              isGrouped={false}
```
dlabaj: Wondering why this snapshot changed.

rebeccaalpert: I'm going to look into it. There are also approximately a million linting issues now in files I haven't touched in this PR, so I suspect it might be something from a different PR.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export class TextInput extends React.Component<TextInputProps> {
+  static defaultProps = {
+    'aria-label': null as string,
+    className: '',
+    isRequired: false,
+    isValid: true,
+    isDisabled: false,
+    isReadOnly: false,
+    type: 'text'
```
jeff-phillips-18: onChange and value need default values.

rebeccaalpert: They need to default to undefined, and I was told that leaving them out will do that automatically.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import styles from '@patternfly/patternfly/components/FormControl/form-control.css';
+import { css } from '@patternfly/react-styles';
+import { Omit } from '../../helpers/typeUtils';
+import { FormEvent } from 'react';
+
+export enum inputTypes {
```
jschuler: Would be better to prefix with the component name so as not to cause collisions in the future, e.g. `TextInputTypes`

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+  /** Type that the input accepts. */
+  type?: 'text' | 'date' | 'datetime-local' | 'email' | 'month' | 'number' | 'password' | 'search' | 'tel' | 'time' | 'url';
+  /** Value of the input. */
+  value?: string | number;
+  /** Aria-label. The input requires an associated id or aria-label. */
+  'aria-label'?: string;
+}
+
+export class TextInput extends React.Component<TextInputProps> {
+  static defaultProps = {
```
jschuler: Missing default for onChange. Probably needed since in your handleChange method you call onChange() but what if it is undefined?

rebeccaalpert: They need to default to undefined, and I was told that leaving them out will do that automatically.

packages/patternfly-4/react-core/src/components/TextInput/TextInput.test.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
```
redallen: Change the extension on this file to .tsx

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+  /** Flag to show if the input is read only. */
+  isReadOnly?: boolean;
+  /** Flag to show if the input is required. */
+  isRequired?: boolean;
+  /** Flag to show if the input is valid or invalid. */
+  isValid?: boolean;
+  /** A callback for when the input value changes. */
+  onChange?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
+  /** Type that the input accepts. */
+  type?: 'text' | 'date' | 'datetime-local' | 'email' | 'month' | 'number' | 'password' | 'search' | 'tel' | 'time' | 'url';
```
redallen: To allow use of the enum, I believe you need to add `| TextInputTypes`

packages/patternfly-4/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  handleChange = (event: React.FormEvent<HTMLInputElement>) => {
+    this.props.onChange(event.currentTarget.value, event);
+  }
+
+  render() {
+    const { className, type, value, onChange, isValid, isReadOnly, isRequired, isDisabled, ...props } = this.props;
+    return (
+      <input
+        {...props}
```
redallen: Let's spread the props at the end to be consistent across components.

Issue #1921, opened 2019-05-03
Kebab typedef


priley86:
As mentioned in Slack, I am seeing the following when passing the `component` prop for the Kebab:
```
ERROR in /go/src/github.com/openshift/console/frontend/public/components/utils/kebab.tsx
(139,13): Type '"a" | "button"' is not assignable to type 'ElementType<DropdownItemProps>'.
  Type '"button"' is not assignable to type 'ElementType<DropdownItemProps>'.
```

Usage:
```
<DropdownItem key={option.label} 
            isDisabled={isDisabled} 
            onClick={option.callback}
            href={option.href}
            component={ option.href ? "a" : "button"}>
            {option.label}
          </DropdownItem>
```

Additionally, I tried to remove the `component` prop altogether and just let the `DropdownItem` component handle this for me (with the other props in tact). It rendered properly, however I noted the `onClick` behavior was different after this (and `preventDefault() ` was no longer called on my anchor items).

The suggested fix for now would just be updating the typedef (as the behavior is working fine with the `component` prop passed, it just fails the CI).

https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.d.ts#L4

priley86:
related to #1825  it seems

priley86:
full source of the issue above is here:
https://github.com/priley86/web-ui/blob/react-tables-poc/frontend/public/components/utils/kebab.tsx#L134

tlabaj:
Closed by PR #1924

PR #1924, opened 2019-05-03
fix (dropitems): Updates to drop down item and skip to content api

dlabaj:
**What**:
Fixes for Dropdown item and update to skip to content for issue #1921 

**Additional issues**:


priley86:
thanks @dlabaj !

dgutride:
this commit message is problematic because it doesn't follow the semantic release convention - please update and force push so we can merge

PR #1926, opened 2019-05-03
feat(ChipGroup): Converted ChipGroup to typescript

jessiehuff:


redallen:
Only useful commit is 87864a3, open a new PR with all of ChipGroup converted and none of the checkbox stuff!

PR #1928, opened 2019-05-05
feat(Paginator): Add disableNext,disablePrev Props

glekner:
**What**:
Add disableNext,disablePrev Props to the Paginator component.


glekner:
@jeff-phillips-18 can you approve? thanks

PR #1930, opened 2019-05-05
fix(DualList): The component shouldn't count, toggle or filter hidden items.

Laviro:
The component shouldn't count, toggle or filter hidden items.


Laviro:
rebased, @jeff-phillips-18  could you take a look ? :)

Issue #1919, opened 2019-05-02
No way to set background image for About Modal via JS


seanforyou23:
AboutModal accepts a `logoImageSrc` prop and passes this to AboutModalBoxContainer, but then doesn't do anything with it. I believe it was intended to set the background image on the right side of the modal box. Currently there's no way to override the bg image unless you're using CSS, which isn't always possible on the consuming side.

Since we're not currently using the logoImage{Src,Alt} property for anything, I think we could just rename this to something more appropriate like `bgImgSrc` and use this value to apply the background image when it's provided. Regardless of how it's implemented, we just need a way to override the background image using pure JS, however that's accomplished.

https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalContainer.js#L57

@dlabaj

dgutride:
@seanforyou23  - why would a consumer not be able to use CSS?  There are some properties that will always only be available via css customization and this is one of our guiding principles.  I could see a src image possibly being an exception, but I'd like to understand the foundation of the argument before we move forward.

seanforyou23:
@dgutride sorry wrote that description in a hurry, let me clarify. Not that we can't use CSS at all, just that we can't use CSS to target certain image assets because some assets are generated during the build process with a hash in the name for cache busting purposes, so the exact name isn't really known (for example `/static/media/pf4-downstream-bg.5a6958f1.svg`.

I was hoping we could use a technique kinda like below to allow users to pass a dynamically generated file path or data url, and then plant that on the hero DOM element. Not 100% sure it will work, but seems possible. Thoughts?

```
render(){
  const style = { "--pf-c-about-modal-box__hero--sm--BackgroundImage": myFile } as React.CSSProperties;
  return <div style={style}>...</div>
}
```

rachael-phillips:
Closing as this was completed

PR #1940, opened 2019-05-06
fix(AboutModal): provide a way to set the background using props

seanforyou23:
**What**: This PR removes the `logoImageSrc` and `logoImageAlt` props from about modal as they were never actually used. Beyond removing these dead API parts, I've replaced them with a single prop, `bgImgSrc` which I believe fulfills the original intent - which is making the background graphic on the right side of the modal configurable.

I've added a new example for AboutModal, one that illustrates how to set a custom background image.

The value passed for `backgroundImageSrc` is used to set the value for the css variable "--pf-c-about-modal-box__hero--sm--BackgroundImage". This is different than how we typically set CSS vars, and the need for it is described here https://github.com/patternfly/patternfly-react/issues/1919. In a nutshell, the need is driven by having assets served by a caching service or using generated asset filenames of which you don't know the exact name of a file, which in turn prevents you from setting this in a stylesheet.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/1919

Screen Shot 2019-05-06 at 3 33 55 PM

### DIFF
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.d.ts
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.d.ts
@@ -8,8 +8,7 @@ export interface AboutModalProps extends HTMLProps<HTMLDivElement> {
   trademark?: string;
   brandImageSrc: string;
   brandImageAlt: string;
-  logoImageSrc?: string;
-  logoImageAlt?: string;
+  backgroundImageSrc?: string;
   noAboutModalBoxContentContainer?: boolean;
 }
 
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
@@ -24,20 +24,14 @@ const propTypes = {
   brandImageSrc: PropTypes.string.isRequired,
   /** The alternate text of the brand image */
   brandImageAlt: PropTypes.string.isRequired,
-  /** The URL of the image for the logo */
-  logoImageSrc: PropTypes.string,
-  /** The alternate text of the logo image */
-  logoImageAlt: props => {
-    if (props.logoImageSrc && !props.logoImageAlt) {
-      return new Error('logoImageAlt is required when a logoImageSrc is specified');
-    }
-    return null;
-  },
+  /** The URL of the image for the background */
+  backgroundImageSrc: PropTypes.string,
   /** Prevents the about modal from rendering content inside a container; allows for more flexible layouts */
   noAboutModalBoxContentContainer: PropTypes.bool,
   /** Additional props are passed and spread to the modal content container <div> */
   '': PropTypes.any
 };
+
 
 const defaultProps = {
   className: '',
@@ -45,8 +39,7 @@ const defaultProps = {
   onClose: () => undefined,
   productName: '',
   trademark: '',
-  logoImageSrc: '',
-  logoImageAlt: '',
+  backgroundImageSrc: '',
   noAboutModalBoxContentContainer: false
 };
 
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.d.ts
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.d.ts
@@ -1,5 +1,7 @@
 import { FunctionComponent, HTMLProps } from 'react';
 
-export interface AboutModalBoxHeroProps extends HTMLProps<HTMLImageElement> {}
+export interface AboutModalBoxHeroProps extends HTMLProps<HTMLImageElement> {
+  backgroundImageSrc?: string;
+}
 declare const AboutModalBoxHero: FunctionComponent<AboutModalBoxHeroProps>;
 export default AboutModalBoxHero;
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.js
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.js
@@ -1,11 +1,14 @@
 import React from 'react';
 import { css } from '@patternfly/react-styles';
+import { c_about_modal_box__hero_sm_BackgroundImage } from '@patternfly/react-tokens'
 import PropTypes from 'prop-types';
 import styles from '@patternfly/patternfly/components/AboutModalBox/about-modal-box.css';
 
 const propTypes = {
   /** additional classes added to the About Modal Hero */
   className: PropTypes.string,
+  /** background image data or file path */
+  backgroundImageSrc: PropTypes.string,
   /** Additional props are spread to the container <div> */
   '': PropTypes.any
 };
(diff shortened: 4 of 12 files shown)


packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.d.ts
```diff
@@ -8,8 +8,7 @@ export interface AboutModalProps extends HTMLProps<HTMLDivElement> {
   trademark?: string;
   brandImageSrc: string;
   brandImageAlt: string;
-  logoImageSrc?: string;
```
dlabaj: @tlabaj @dgutride I'm okay with removing these without saying this a breaking change because they never did anything.  If a consumer used them nothing would happen they aren't implemented in the component.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.d.ts
```diff
@@ -8,8 +8,7 @@ export interface AboutModalProps extends HTMLProps<HTMLDivElement> {
...
-  logoImageAlt?: string;
+  bgImgSrc?: string;
```
dlabaj: @seanforyou23 Can we call this backgroundImageSrc instead? Also do we need a backgroundImageAlt?

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.js
```diff
@@ -14,9 +16,14 @@ const defaultProps = {
...
 
-const AboutModalBoxHero = ({ className, ...props }) => (
-  <div {...props} className={css(styles.aboutModalBoxHero, className)} />
-);
+const AboutModalBoxHero = ({ className, bgImgSrc, ...props }) => {
+  const bgStyle = (bgImgSrc !== '') ? {
+    "--pf-c-about-modal-box__hero--sm--BackgroundImage": `url(${bgImgSrc})`
+  } : {};
+  return (
+    <div {...props} style={bgStyle} className={css(styles.aboutModalBoxHero, className)} />
```
dlabaj: What happens if it doesn't find the image?

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
```diff
@@ -22,14 +22,12 @@ const propTypes = {
...
+  bgImgSrc: PropTypes.string,
   /** The alternate text of the brand image */
-  brandImageAlt: PropTypes.string.isRequired,
-  /** The URL of the image for the logo */
-  logoImageSrc: PropTypes.string,
-  /** The alternate text of the logo image */
-  logoImageAlt: props => {
-    if (props.logoImageSrc && !props.logoImageAlt) {
-      return new Error('logoImageAlt is required when a logoImageSrc is specified');
+  brandImageAlt: props => {
```
jschuler: since brandImageSrc is already required, we can also simply set brandImageAlt to required.
Although i'm questioning whether the brand image should be required...

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.js
```diff
@@ -14,9 +16,14 @@ const defaultProps = {
   className: ''
 };
 
-const AboutModalBoxHero = ({ className, ...props }) => (
-  <div {...props} className={css(styles.aboutModalBoxHero, className)} />
-);
+const AboutModalBoxHero = ({ className, bgImgSrc, ...props }) => {
+  const bgStyle = (bgImgSrc !== '') ? {
+    "--pf-c-about-modal-box__hero--sm--BackgroundImage": `url(${bgImgSrc})`
```
dlabaj: Thinking we should make "--pf-c-about-modal-box__hero--sm--BackgroundImage" a constant stored in utils constant.   Maybe make a CSS section where we can add one offs like this that could be affected by a change in core down the road.  This way if we have to change this name with a core update we can find it easier.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalContainer.d.ts
```diff
@@ -10,8 +10,7 @@ export interface AboutModalContainerProps extends HTMLProps<HTMLDivElement> {
   trademark: string;
   brandImageSrc: string;
   brandImageAlt: string;
-  logoImageSrc: string;
-  logoImageAlt: string;
+  bgImgSrc: string;
```
jschuler:
```suggestion
  bgImgSrc?: string;
```

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.d.ts
```diff
@@ -1,5 +1,7 @@
 import { FunctionComponent, HTMLProps } from 'react';
 
-export interface AboutModalBoxHeroProps extends HTMLProps<HTMLImageElement> {}
+export interface AboutModalBoxHeroProps extends HTMLProps<HTMLImageElement> {
+  bgImgSrc?: any;
```
jschuler:
```suggestion
  bgImgSrc?: string;
```

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.d.ts
tlabaj: When initially implemented, the image was not part of css.  So these props did something at one point, but with the current implementation, they are not doing anything.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.js
jschuler: I believe you can
`import { c_about_modal_box__hero_sm_BackgroundImage } from '@patternfly/react-tokens'`
and `c_about_modal_box__hero_sm_BackgroundImage.name` gives you the variable name

dlabaj: If that works then let's go that route.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.d.ts
seanforyou23: Don't need an alt when it's just background styling

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxHero.js
seanforyou23: Good question, so if you're referencing a local module like in the example, and it can't be found, your IDE should notice this and throw an error. If you're providing a value like `https://www.site.com/image.svg` and that asset doesn't properly fetch, you'd get a standard 404 just as if you provide an inaccessible asset in any regular background definition, in which case you'd get a black background.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.js
seanforyou23: For now I left it as required, can change at a later point if you like. The main focus for this PR is just to make the background image configurable.

PR #1945, opened 2019-05-07
feat(Radio): Convert radio to TypeScript

rebeccaalpert:
Converted to TypeScript. I made a small change to the behavior since it was triggering a React warning related to switching from an uncontrolled to a controlled component (and back again) when you clicked the controlled radio and one about defaultChecked and checked being applied at the same time.

Fixes https://github.com/patternfly/patternfly-react/issues/1997.


nicolethoen:

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react';
+import styles from '@patternfly/patternfly/components/Radio/radio.css';
+import { css, getModifier } from '@patternfly/react-styles';
+import { Omit } from '../../helpers/typeUtils';
+
+export interface RadioProps extends Omit<React.HTMLProps<HTMLInputElement>, 'label' | 'onChange'> {
```
dlabaj: Need to Omit type and disable here as well.  We control these internally via isDisable, and we don't want consumers changing the type of input on us.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
...
+  'aria-label'?: string;
+}
+
+export class Radio extends React.Component<RadioProps> {
+  static defaultProps = {
+    className: '',
+    isChecked: null,
+    isDisabled: false,
+    isValid: true,
+    label: undefined,
```
dlabaj: I believe this is set to undefined by default so this line can be removed.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
...
+    onChange: () => undefined
```
dlabaj: Same with onChanage.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** A callback for when the radio selection changes. */
+  onChange?: (checked: boolean, event: React.FormEvent<HTMLInputElement>) => void;
+  /** Aria-label of the radio. */
+  'aria-label'?: string;
+}
+
+export class Radio extends React.Component<RadioProps> {
+  static defaultProps = {
+    className: '',
+    isChecked: null,
```
dlabaj: Shouldn't this be false?

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <input
+          {...props}
+          aria-label={label ? undefined : ariaLabel}
+          className={css(styles.radioInput)}
+          type="radio"
+          onChange={this.handleChange}
+          aria-invalid={!isValid}
+          disabled={isDisabled}
+          checked={isChecked}
+          defaultChecked={undefined}
```
dlabaj: Should be false

packages/patternfly-4/react-core/src/components/Radio/index.ts
```diff
@@ -0,0 +1 @@
+export { Radio } from './Radio';
```
dlabaj:
```suggestion
export * from './Radio';
```

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
rebeccaalpert: It was a hacky workaround for a React warning. I figured out a better way to do it. I'm going to drop  this.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
rebeccaalpert: If you set it to false, it breaks the uncontrolled radio.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
rebeccaalpert: It's to resolve a React warning about having both checked and defaultChecked props. I can declare it as a prop and just not pass it along to the input when there's an isChecked if you'd prefer that.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
...
+  isDisabled?: boolean;
+  /** Flag to show if the radio selection is valid or invalid. */
+  isValid?: boolean;
+  /** Label text of the radio. */
+  label?: React.ReactNode;
+  /** Name for group of radios */
+  name: string;
+  /** A callback for when the radio selection changes. */
+  onChange?: (checked: boolean, event: React.FormEvent<HTMLInputElement>) => void;
+  /** Aria-label of the radio. */
```
tlabaj: Can you update the description here.  Can we say `Aria label` or  `The aria-label`

rebeccaalpert:
Added demo/tests and addressed PR feedback.

packages/patternfly-4/react-integration/cypress/integration/radio.spec.ts
```diff
@@ -0,0 +1,26 @@
+describe('Radio Demo Test', () => {
+  it('Navigate to demo section', () => {
```
tlabaj: Can you add some interactive test for checking and unchecking the controlled and uncontrolled buttons?

rebeccaalpert: It has one for the controlled button, but the uncontrolled buttons don't change at all HTML-wise when they're selected, so I don't know that you can write that kind of test for them.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,80 @@
...
+  name: string;
+  /** A callback for when the radio selection changes. */
+  onChange?: (checked: boolean, event: React.FormEvent<HTMLInputElement>) => void;
+  /** Aria label for the radio. */
+  'aria-label'?: string;
+}
+
+export class Radio extends React.Component<RadioProps> {
+  static defaultProps = {
+    className: '',
```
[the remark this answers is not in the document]
rebeccaalpert: isChecked needs to be undefined so that the uncontrolled buttons work correctly.

PR #1946, opened 2019-05-07
feat(List): Convert list to TypeScript

rebeccaalpert:
Converted to TypeScript.

Fixes https://github.com/patternfly/patternfly-react/issues/1994.


nicolethoen:

packages/patternfly-4/react-core/src/components/List/index.ts
```diff
@@ -0,0 +1,2 @@
+export { List } from './List';
```
dlabaj: Need to convert these to export * with the ts files since Babel only works on a file by file basis.
```suggestion
export * from './List';
export * from './ListItem';
```

rebeccaalpert:
Added demo and test. I also addressed @dlabaj's PR feedback. @seanforyou23, as noted on Slack, git mv doesn't seem to net a different outcome, so I'm leaving that as-is for now.

PR #1957, opened 2019-05-09
fix(build): use native fs-extra commands instead of folderHash

### LINKED ISSUE #1927: Build/devserver breaks with node 10.x
Found a bug where editing any component file prevents running a build or starting the dev server. Seems to be an issue on node10 and not node8.

Steps to reproduce;

0) Ensure you're on Active LTS node version `10.15.3` as directed in our docs for [pf3](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/README.md#node-environment) and [pf4](https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/react-core#node-environment)
1) yarn && yarn bootstrap && yarn build && yarn build:docs
2) edit AboutModal.js (or any component) with a simple line break anywhere
3) yarn build
4) shell crashes, notice the infinite loop of the following error

`(node:21805) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 2 error listeners added. Use emitter.setMaxListeners() to increase limit`

Screen Shot 2019-05-02 at 5 12 41 PM

Following the same steps when running node `8.16.0` and there's no issue. Running similar exercises in another lerna project and there's no issue. Rebooted and recloned the repo and the problem still exists.

Looks like it was introduced with https://github.com/patternfly/patternfly-react/pull/1835

redallen:
**What**: Might resolve some OSX + Node 10 issues. I'm unable to repro, so please test this out @jeff-phillips-18 @rebeccaalpert @nicolethoen @seanforyou23  ! Closes #1927 .

**Additional issues**:


jeff-phillips-18:
This works for me when using Node 10, whereas it failed every time before. Thanks @redallen !

PR #1960, opened 2019-05-09
fix(ts-doc): fix regressions

redallen:
**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
```diff
@@ -5,7 +5,7 @@ import { AngleRightIcon } from '@patternfly/react-icons';
 
 export interface ExpandableProps {
   /** Content rendered inside the Expandable Component */
-  children: React.ReactNode;
+  children: any;
```
dlabaj: Leave this as ReactNode since we haven't released this.

Issue #1925, opened 2019-05-03
Donut chart problems


TheRealJon:
Currently, the donut chart does not meet the PF design criteria. There are a few issues that need to be addressed:

1. The 'donut hole' (innerRadius) is a static value which will cause issues if users pass in width/height values to the ChartDonut component. Need to figure out the best way to calculate this value so that setting width and height don't break the chart. See this example of a donut chart implementation: https://medium.com/@sschannak/creating-a-progress-chart-with-victorypie-in-react-20514c92775

2. Setting width or height on the DonutChart can cause some really strange behavior. For example, setting it to a small value like`10` can cause it to scale up to a huge size within the parent. This might just be an issue with Victory and svg scaling, but I feel like setting a width or height shouldn't cause strange, unwanted behavior. The above implementation with the width and height props abstracted might help with this.

3. Tooltips are not PF design

4. There should be a prop to easily set the label shown in the middle of the donut. As it is now, user has to create their own svg and position the text themselves. This makes the component difficult to use as the design intends.

dgutride:
@dlabrecq  - please take a look at this and let me know the level of effort

janwright73:
@dlabrecq - any chance you can jump on this and help Jon get these issues resolved?

dlabrecq:
Regarding #4, creating a generic label is difficult for the donut chart to do. Victory doesn't have its donut chart, which is why the label is created via compilation. 

Note that this was a first pass at a donut chart. (We haven't made any changes since November and still don't have any pf-core variables for the theme.) There are known issues trying to use the Victory pie chart in this way and don't have a good answer for this.

Regarding #3 et al, there is still some basic architecture that should be addressed first, IMO. For example, theme properties, color variables, etc. That would address tooltips for all charts, for example.

FYI, Cost management overrode the styles for tooltips to show the PF4 white text on black background -- not difficult.

TheRealJon:
@dlabrecq

> Regarding #4, creating a generic label is difficult for the donut chart to do. Victory doesn't have its own donut chart, which is why the label is created via compilation as shown in the example above.

Could we do something like what we are doing in the example, just wrap it into the component? The label is part of the design, so I think we probably need to provide a simple API for it.

> Regarding #3 et al, there is still some basic architecture that should be addressed first, IMO. For example, theme properties, color variables, etc. That would address tooltips for all charts, for example.

[...]

PR #1964, opened 2019-05-10
Update donut chart, examples, & theme

dlabrecq:
This PR updates donut chart features, refactors theme properties, and fixes all the broken examples due to the latest Gatsby refactoring.

Fixes https://github.com/patternfly/patternfly-react/issues/1925

Part #1: New donut chart features.

- Adjust donut innerRadius width
- Styled Victory tooltips to appear similar to PF4
- Added functionality to display a label inside the donut chart
- Added legend responsive property to enable when located inside a chart

Part #2: This not only supports the new donut chart theme, but future themes required for the up coming percentage donut, bullet chart, etc.

- Created a new donut theme
- Refactored for new base and color themes
- Added themeColor and themeVariant props to apply colors, instead of relying on theme property 

Note: Color variants (themes) are merged on top of the base theme. For the donut chart, the extra donut theme properties are merge on top of the base/color themes. Thus, ensuring color scales can be applied to the donut chart (et al.), while user provided themes remain unmodified.

Part #3: Fix the broken examples

- Fixed all broken chart examples by adding missing SCSS
- Fixed incorrectly assigned theme properties, by replacing Object.assign with spread operators

~~Note: Gatsby still has an issue with SCSS in production (e.g., when served by surge.sh). Although, the demo displays fine locally, as demonstrated by the screenshots below.~~

Screen Shot 2019-05-10 at 12 01 46 AM

Screen Shot 2019-05-10 at 12 01 53 AM

Screen Shot 2019-05-10 at 12 01 26 AM

Screen Shot 2019-05-10 at 12 01 11 AM

Screen Shot 2019-05-10 at 12 01 38 AM

Screen Shot 2019-05-10 at 12 02 15 AM


TheRealJon:
Just noticed this bug with the donut chart. For some reason, the center label behaving like a `position: fixed` element:

donut scroll bug

dlabrecq:
The label is positioned absolute. Do you have the inspector open?

TheRealJon:
Yes, but the behavior also occurs if I just add enough content to the page to make it scrollable.

TheRealJon:
Also, the ratio of donut to label size seems to be a bit off from the design, just thinking that all of the defaults should create a donut that looks exactly like the design, then if devs want to make adjustments they can:
image
image

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,164 @@
...
+    const titleStyles = {
+      height: styles.label.height,
+      position: styles.label.position,
+      width: styles.label.width
+    };
+
+    const chartTheme = theme || getDonutTheme(themeColor, themeVariant);
+
+    return (
+      <React.Fragment>
```
TheRealJon: I think we are going to have to use the `standalone={false}` prop on the VictoryPie component in this case. That way we can wrap the whole thing in an svg and position the label accordingly. As it is, we are creating 2 separate svg images, then trying to overlay them using `position: absolute`.

TheRealJon: We can make the api similar to to Victory's by conditionally wrapping the component in an svg based on a `standalone` prop.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,164 @@
...
+      });
+      this.handleResize();
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.handleResize);
+  }
+
+  getSubTitle = () => {
```
TheRealJon: Not sure what the normal pattern is in PF, but could this be a separate `DonutTitle` component instead of a class function?

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,164 @@
...
+    return null;
+  };
+
+  getTitle = () => {
+    const { title } = this.props;
+    const { height, width } = this.state;
+
+    if (title && height && width) {
+      const xVal = width / 2 - title.length / 2;
+      const yVal = height / 2 - 10;
```
TheRealJon: It there is no subtitle (perhaps when a donut is just showing a percentage), then the title will be off center when subtracting 10px from the yVal.

packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md
```diff
@@ -3,58 +3,59 @@ title: 'Donut chart'
...
 import React from 'react';
+import { ChartDonut, ChartLabel, ChartLegend, ChartThemeColor, ChartThemeVariant } from '@patternfly/react-charts';
 
 <div>
-  <div style={{ height: 450, width: 450 }}>
-    <svg style={{ position: 'absolute' }} height={this.size} width={this.size}>
-      <ChartLabel style={{ fontSize: 20 }} text="100" textAnchor="middle" verticalAnchor="middle" x={225} y={200} />
-      <ChartLabel style={{ fill: '#bbb' }} text="Pets" textAnchor="middle" verticalAnchor="middle" x={255} y={250} />
+  <div className="donut-chart-container">
+    <svg className="donut-chart-label">
```
TheRealJon: Should we use the `title` and `subTitle` props in both examples? Might be confusing for users to see two very different examples like this side by side. At the very least, there should be some explanation of why this example is different than the first one.

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
```diff
@@ -0,0 +1,43 @@
...
+const COLOR_LABEL = global_Color_light_100.value;
+const COLOR_SCALE = [
+  global_active_color_100.value,
+  global_active_color_200.value,
+  global_active_color_300.value,
+  '#0066ff'
+];
+const COLOR_STACK_STROKE = global_Color_dark_200.value;
+const COLOR_STROKE = global_active_color_300.value;
+const COLOR_TOOLTIP_FILL = '151515';
```
TheRealJon:
```suggestion
const COLOR_TOOLTIP_FILL = '#151515';
```

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
```diff
@@ -0,0 +1,43 @@
...
+const COLOR_TOOLTIP_STROKE = '151515';
```
TheRealJon:
```suggestion
const COLOR_TOOLTIP_STROKE = '#151515';
```

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
```diff
@@ -0,0 +1,43 @@
...
+const COLOR_TOOLTIP_FLYOUT_FILL = 'ededed';
```
TheRealJon:
```suggestion
const COLOR_TOOLTIP_FLYOUT_FILL = '#EDEDED';
```

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
```diff
@@ -0,0 +1,43 @@
...
+const COLOR_TOOLTIP_FLYOUT_STROKE = 'ededed';
```
TheRealJon:
```suggestion
const COLOR_TOOLTIP_FLYOUT_STROKE = '#EDEDED';
```

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-green.js
```diff
@@ -0,0 +1,43 @@
...
+const COLOR_FILL = global_active_color_200.value;
+const COLOR_LABEL = global_Color_light_100.value;
+const COLOR_SCALE = [
+  global_success_color_100.value,
+  global_success_color_200.value,
+  '#59C768'
+];
+const COLOR_STACK_STROKE = global_Color_dark_200.value;
+const COLOR_STROKE = global_active_color_300.value;
+const COLOR_TOOLTIP_FILL = '151515';
```
TheRealJon: Fix hex values

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-multi.js
```diff
@@ -0,0 +1,42 @@
...
+const COLOR_LABEL = global_Color_light_100.value;
+const COLOR_SCALE = [
+  '#39a5dc',
+  '#9C92F5',
+  '#59C768',
+  '#FFD47D'
+];
+const COLOR_STACK_STROKE = global_Color_dark_200.value;
+const COLOR_STROKE = global_active_color_300.value;
+const COLOR_TOOLTIP_FILL = '151515';
```
TheRealJon: Fix hex values

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/light/theme-color-blue.js
```diff
@@ -0,0 +1,44 @@
...
+const COLOR_LABEL = global_Color_dark_100.value;
+const COLOR_SCALE = [
+  global_active_color_100.value,
+  global_active_color_200.value,
+  global_active_color_300.value,
+  '#0066ff'
+];
+const COLOR_STACK_STROKE = global_Color_light_100.value;
+const COLOR_STROKE = global_active_color_300.value;
+const COLOR_TOOLTIP_FILL = 'ededed';
```
TheRealJon: Fix hex values

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/light/theme-color-green.js
```diff
@@ -0,0 +1,44 @@
...
+const COLOR_FILL = global_active_color_200.value;
+const COLOR_LABEL = global_Color_dark_100.value;
+const COLOR_SCALE = [
+  global_success_color_100.value,
+  global_success_color_200.value,
+  '#59C768'
+];
+const COLOR_STACK_STROKE = global_Color_light_100.value;
+const COLOR_STROKE = global_active_color_300.value;
+const COLOR_TOOLTIP_FILL = 'ededed';
```
TheRealJon: Fix hex values

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/light/theme-color-multi.js
```diff
@@ -0,0 +1,43 @@
...
+const COLOR_LABEL = global_Color_dark_100.value;
+const COLOR_SCALE = [
+  '#39a5dc',
+  '#9C92F5',
+  '#59C768',
+  '#FFD47D'
+];
+const COLOR_STACK_STROKE = global_Color_light_100.value;
+const COLOR_STROKE = global_active_color_300.value;
+const COLOR_TOOLTIP_FILL = 'ededed';
```
TheRealJon: Fix hex values

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-victory.js
```diff
@@ -0,0 +1,235 @@
...
+  padding: 10, // Value must be in pixles
+  stroke: 'transparent'
+};
+const LABEL_CENTERED_PROPS = {
+  ...LABEL_PROPS,
+  textAnchor: 'middle'
+};
+
+// Layout
+const LAYOUT_PROPS = {
```
TheRealJon: Remove?

dlabrecq: I wasn't ready to remove that quite, yet. Wanted to see what variables are created in patternfly core

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/light/theme-color-multi.js
dlabrecq: Yes, will fix

packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md
dlabrecq: Wanted to show the option and let the label bake for a bit?

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
dlabrecq: true

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
dlabrecq: We're already using ChartLabel for this. Perhaps if DonutTitle were exposed via composition, that would be more useful.

packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md
TheRealJon: I feel like that would just create more confusion in the long run. We could just refer the user to the Victory docs for doing custom graphs like this example.

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-victory.js
TheRealJon: Could we just add it back in when we pull in PF variables? It's just dead code that will be merged into master.

dlabrecq: I want to leave it for now because I'm not finished working on with the theme. If we end up with no layout properties (e.g., a default height and width), I will remove it.

packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md
dlabrecq: I will remove the example after we're confident we have a good solution for the label

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
dlabrecq: Fixed

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
dlabrecq: Fixed

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-blue.js
dlabrecq: Fixed

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-multi.js
dlabrecq: Fixed

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/light/theme-color-green.js
dlabrecq: Fixed

packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md
TheRealJon: Okay, are you saying that this PR is a WIP and that you will remove this before merging? If so we should probably prefix the PR title with '[WIP]' and put a `Do Not Merge` label on it.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,164 @@
...
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.handleResize);
+  }
+
+  getSubTitle = () => {
+    const { subTitle } = this.props;
+    const { height, width } = this.state;
+
+    if (subTitle && height && width) {
+      const xVal = width / 2 - subTitle.length / 2;
```
TheRealJon: Isn't 'subTitle.length' the number of chars in the string? Since `width` is pixels, I could see this causing alignment issues. Also, if this were wrapped in an svg with the `VictoryPie` component, they will share the same coordinate system and the svg textAnchor prop should take care of centering the text horizontally.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
TheRealJon: We could actually get rid of these and the resize listener if we use a VictoryContainer to put the ChartLabel and VictoryPie components into a single svg:

```jsx
<VictoryContainer width={400} height={400}>
    <ChartLabel
      style={[styles.label.title, styles.label.subTitle]}
      text={[title, subTitle]}
      textAnchor="middle"
      verticalAnchor="middle"
      x="200"
      y="200"
    />
     <VictoryPie
      standalone={false}
      labelComponent={<ChartTooltip theme={chartTheme} />}
      theme={chartTheme}
      {...rest}
    />
</VictoryContainer>
```

TheRealJon: With defaults and calculated props it could look like this:

```jsx
const ChartDonut = (props) => {
  const { height = 200, padAngle = 1, subTitle, themeColor, themeVariant, title, width = 200, ...rest } = props;

  // Define innerRadius and theme separately since default is calculated from other props
  const { innerRadius = (width - 32) / 2, theme = getTheme(themeColor, themeVariant) } = props;

  return (
    <VictoryContainer width={width} height={height}>
      <ChartLabel
        style={[styles.label.title, styles.label.subTitle]}
        text={[title, subTitle]}
        textAnchor="middle"
        verticalAnchor="middle"
        x={width / 2}
        y={height / 2}
      />
      <VictoryPie
        height={height}
        innerRadius={innerRadius}
        labelComponent={<ChartTooltip theme={theme} />}
        padAngle={padAngle}
        standalone={false}
        theme={theme}
        width={width}
        {...rest}
      />
    </VictoryContainer>
  );
};```

File: packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-multi.js (hunk 10 above)
Comment (TheRealJon) [reply packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-multi.js]: Have you committed the fix for this? Looks like it hasn't changed.

File: packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js (hunk 15 above)
Comment (dlabrecq) [reply packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js]: textAnchor defines how the text is horizontally positioned relative to the given x and y coordinates. It's not perfect, and may iterate on this, but need to get x and y close for now.

File: packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md (hunk 4 above)
Comment (dlabrecq) [reply packages/patternfly-4/react-charts/src/components/ChartDonut/examples/ChartDonut.md]: No, I'm saying there is more work to be done with the theme, but not in this PR.

File: packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js (hunk 2 above)
Comment (dlabrecq) [reply packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js]: This is similar to what I've been experimenting with (i.e., ChartContainer). Although, I was having a difficult time without setting innerRadius dynamically. Your calculation seems to work well, thanks.

My only concern with this approach is that it forces the user to provide a height and width. This particular chart would be used differently than all others. I was using the resize listener so the pie chart would behave similarly and continue to be responsive. 

That said, I'll continue to experiment with the listener. It would be ideal if the charts behave the same.

File: packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-multi.js (hunk 10 above)
Comment (dlabrecq) [reply packages/patternfly-4/react-charts/src/components/ChartTheme/themes/dark/theme-color-multi.js]: Pushed changes

File: packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js (hunk 2 above)
Comment (TheRealJon) [reply packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js]: Yes, `VictoryContainer` is responsive like the other charts. In my example, I provide default width, height, padAngle, innerRadius,  and theme props so that the user doesn't have to define explicit values for those.

File: packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js [hunk 16]
```diff
@@ -1,21 +1,130 @@
...
+   * Because Victory renders responsive containers, the width and height props do not determine the width and
+   * height of the chart in number of pixels, but instead define an aspect ratio for the chart. The exact number of
+   * pixels will depend on the size of the container the chart is rendered into.
+   *
+   * Note: innerRadius may need to be set when using this property.
+   */
+  width: PropTypes.number
 };
 
+class ChartDonut extends React.Component {
```
TheRealJon: I think this should be a stateless component. The `VictoryContainer` component will give us the same responsive behavior as the other charts.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,130 @@
...
+      });
+    }
+  };
+
+  render() {
+    const { subTitle, theme, themeColor, themeVariant, title, ...rest } = this.props;
+    const { height, width } = this.state;
+
+    const chartTheme = theme || getDonutTheme(themeColor, themeVariant);
+    const showLabels = subTitle || title;
```
TheRealJon: I don't think we need this. If no title or subtitle exists, then the ChartLabel component won't render anything.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,130 @@
...
+    const innerRadius = (width - 32) / 2;
+
+    if (showLabels && this.props.standalone !== false) {
+      return (
+        <div ref={this.childeRef}>
```
TheRealJon: We shouldn't need to wrap this in a div. If the standalone prop is `true`, we should wrap the `ChartLabel` and `VictoryPie` in the `ChartContainer`. If standalone is `false`, we should wrap them in a React fragment so that only the inner svg graphics are returned (the user is responsible for providing the svg wrapper).

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
dlabrecq: If there is no title or subtitle, there is no need to wrap with VictoryContainer.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
dlabrecq: Defining a default height and width is not the same behavior. Responsive containers output 100% for height and width.

TheRealJon: Yes, the `VictoryContainer` `width` and `height` props work just like the width and height props for other charts. They define the aspect ratio.

TheRealJon: The actual width and height attributes of the svg element will still be 100%

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
dlabrecq: Wrapping in a div to get the parent container's height and width. If standalone={false} there is no div

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
TheRealJon: Right, but if the user does provide labels, and sets `standalone === false`, this breaks.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
TheRealJon: Something like this:
```jsx
const chart = (
      <React.Fragment>
        <ChartLabel
          style={[styles.label.title, styles.label.subTitle]}
          text={[title, subTitle]}
          textAnchor="middle"
          verticalAnchor="middle"
          x={width / 2}
          y={height / 2}
        />
        <VictoryPie
          innerRadius={innerRadius}
          labelComponent={<ChartTooltip theme={theme} />}
          standalone={false}
          theme={chartTheme}
          {...rest}
        />
      </React.Fragment>
    );
    return standalone ? <ChartContainer>{chart}</ChartContainer> : chart;
```

TheRealJon: This would match the behavior of setting `standalone === false` on a primitive Victory chart.

dlabrecq: I don't believe we should set a default height and width here. This would behave differently than the responsive containers which are 100% height and width.

TheRealJon: From the Victory docs:
> VictoryContainer provides a top-level <svg> element for other Victory components to render within. **By default, VictoryContainer renders responsive SVGs.** VictoryContainer also provides a Portal container that can be accessed via VictoryPortal in order to render specified children above others.

TheRealJon: Victory docs on `height` and `width` prop of container components:
> The height prop determines the height of the containing <svg>. By default VictoryContainer renders responsive containers with the viewBox attribute set to viewBox="0, 0, width, height" and width="100%", height="100%". In responsive containers, the width and height props affect the aspect ratio of the rendered component, while the absolute width and height are determined by the container. To render a static container, set responsive={false}

> The width prop determines the width of the containing <svg>. By default VictoryContainer renders responsive containers with the viewBox attribute set to viewBox="0, 0, width, height" and width="100%", height="auto". In responsive containers, the width and height props affect the aspect ratio of the rendered component, while the absolute width and height are determined by the container. To render a static container, set responsive={false}

dlabrecq: I read those docs, too. :)

My issue is that we're setting a default height and width inside the component. The Victory greyscale theme applies different values for height and width.

Maybe if we can pull those values from the theme itself? That way we don't end up with different aspect ratios -- I'll try that instead.

TheRealJon: Oh, I see, so you mean the height and width props of the `VictoryPie` component itself, not contianer?

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,96 @@
...
+   * pixels will depend on the size of the container the chart is rendered into.
+   *
+   * Note: innerRadius may need to be set when using this property.
+   */
+  width: PropTypes.number
 };
 
+const ChartDonut = ({
+  height = ChartDonutTheme.pie.height,
+  innerRadius = ((height || width) - 38) / 2,
```
TheRealJon: Not sure you can use the value of another destructured variable in a single statement like this.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,96 @@
...
+  standalone = false,
```
TheRealJon: Should default to true to match Victory props

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,96 @@
...
+        innerRadius={innerRadius > 0 ? innerRadius : ChartDonutTheme.pie.innerRadius}
+        labelComponent={<ChartTooltip theme={theme} />}
+        standalone={false}
+        theme={theme}
+        width={width}
+        {...rest}
+      />
+    </React.Fragment>
+  );
+  return standalone ? chart : <ChartContainer width={width} height={height}>{chart}</ChartContainer>;
```
TheRealJon: This logic is reversed from the Victory behavior. `standalone === true` means that the component will be wrapped in an svg, `standalone === false` means it will not.

dlabrecq: Right. Mixed up the standalone default value.

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-donut.js
```diff
@@ -0,0 +1,24 @@
...
+    },
+    width: '100%'
+  }
+};
+
+// Donut properties to extend Victory theme
+export default {
+  pie: {
+    height: 230,
+    innerRadius: 98,
```
TheRealJon: Since we are calculating an innerRadius in the component, this shouldn't be necessary.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
TheRealJon: Disregard! You can do this, that's really cool. Good to know.

packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.js
```diff
@@ -1,21 +1,96 @@
...
+   * The width props specifies the width of the svg viewBox of the chart container. This value should be given as a
+   * number of pixels.
+   *
+   * Because Victory renders responsive containers, the width and height props do not determine the width and
+   * height of the chart in number of pixels, but instead define an aspect ratio for the chart. The exact number of
+   * pixels will depend on the size of the container the chart is rendered into.
+   *
+   * Note: innerRadius may need to be set when using this property.
+   */
+  width: PropTypes.number
```
TheRealJon: Eslint complains about `theme`, `innerRadius`, and `standalone` missing from propTypes.

packages/patternfly-4/react-charts/src/components/ChartContainer/ChartContainer.js
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable react/require-default-props */
 import React from 'react';
 import PropTypes from 'prop-types';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 import { VictoryContainer } from 'victory';
+import { getTheme } from '../ChartTheme/themes/theme-utils';
```
TheRealJon: Unused

packages/patternfly-4/react-charts/src/components/ChartGroup/ChartGroup.js
```diff
@@ -1,22 +1,54 @@
...
+   *
+   * Note: innerRadius may need to be set when using this property.
+   */
+  width: PropTypes.number
 };
 
-// Note: VictoryGroup.role must be hoisted
-const ChartGroup = ({ children, ...props }) => (
-  <VictoryGroup theme={ChartTheme.default} {...props}>
+const ChartGroup = ({ children, theme, themeColor, themeVariant, ...rest }) => (
```
TheRealJon:
```suggestion
const ChartGroup = ({ children, theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => (
```

packages/patternfly-4/react-charts/src/components/ChartGroup/ChartGroup.js
```diff
@@ -1,22 +1,54 @@
...
+  <VictoryGroup theme={theme || getTheme(themeColor, themeVariant)} {...rest}>
```
TheRealJon:
```suggestion
  <VictoryGroup theme={theme} {...rest}>
```

packages/patternfly-4/react-charts/src/components/ChartLegend/ChartLegend.js
```diff
@@ -1,28 +1,68 @@
...
 };
 
-// Note: VictoryLegend.role must be hoisted
-const container = <ChartContainer responsive={false} />;
-const ChartLegend = props => (
+export const defaultProps = {
+  responsive: false
+};
+
+const ChartLegend = ({responsive, theme, themeColor, themeVariant, ...rest}) => (
```
TheRealJon: Same suggestion as ChartGroup, just add the `getTheme` call as a default initial value here in the destructuring statement.

TheRealJon:
```suggestion
const ChartLegend = ({responsive, theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest}) => (
```

packages/patternfly-4/react-charts/src/components/ChartLegend/ChartLegend.js
```diff
@@ -1,28 +1,68 @@
...
   <VictoryLegend
-    containerComponent={container}
+    containerComponent={<ChartContainer responsive={responsive !== undefined ? responsive : defaultProps.responsive} />}
     dataComponent={<ChartPoint />}
-    theme={ChartTheme.default} {...props}
+    theme={theme || getTheme(themeColor, themeVariant)}
```
TheRealJon:
```suggestion
    theme={theme}
```

packages/patternfly-4/react-charts/src/components/ChartLine/ChartLine.js
```diff
@@ -1,18 +1,30 @@
...
+   * Specifies the theme color; blue (default), green, or multi-color. Overridden by the theme property.
+   */
+  themeColor: PropTypes.string,
+  /*
+   * Specifies the theme variant; 'dark' or 'light' (default). Overridden by the theme property.
+   */
+  themeVariant: PropTypes.string
 };
 
+const ChartLine = ({ theme, themeColor, themeVariant, ...rest }) => (
```
TheRealJon:
```suggestion
const ChartLine = ({ theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => (
```

packages/patternfly-4/react-charts/src/components/ChartLine/ChartLine.js
```diff
@@ -1,18 +1,30 @@
...
+  <VictoryLine theme={theme || getTheme(themeColor, themeVariant)} {...rest} />
```
TheRealJon:
```suggestion
  <VictoryLine theme={theme} {...rest} />
```

packages/patternfly-4/react-charts/src/components/ChartPie/ChartPie.js
```diff
@@ -1,19 +1,54 @@
...
+   * Because Victory renders responsive containers, the width and height props do not determine the width and
+   * height of the chart in number of pixels, but instead define an aspect ratio for the chart. The exact number of
+   * pixels will depend on the size of the container the chart is rendered into.
+   *
+   * Note: innerRadius may need to be set when using this property.
+   */
+  width: PropTypes.number
+};
+
+const ChartPie = ({ theme, themeColor, themeVariant, ...rest }) => {
```
TheRealJon:
```suggestion
const ChartPie = ({ theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => {
```

packages/patternfly-4/react-charts/src/components/ChartPie/ChartPie.js
```diff
@@ -1,19 +1,54 @@
...
+  const chartTheme = theme || getTheme(themeColor, themeVariant);
```
TheRealJon:
```suggestion
```

packages/patternfly-4/react-charts/src/components/ChartPie/ChartPie.js
```diff
@@ -1,19 +1,54 @@
...
+  return <VictoryPie labelComponent={<ChartTooltip theme={chartTheme} />} theme={chartTheme} {...rest} />;
```
TheRealJon:
```suggestion
  return <VictoryPie labelComponent={<ChartTooltip theme={theme} />} theme={theme} {...rest} />;
```

packages/patternfly-4/react-charts/src/components/ChartStack/ChartStack.js
```diff
@@ -1,22 +1,32 @@
...
+  /*
+   * Specifies the theme variant; 'dark' or 'light' (default). Overridden by the theme property.
+   */
+  themeVariant: PropTypes.string
 };
 
-// Note: VictoryStack.getChildren & VictoryStack.role must be hoisted
-const ChartStack = ({ children, ...props }) => (
-  <VictoryStack theme={ChartTheme.default} {...props}>
+const ChartStack = ({ children, theme, themeColor, themeVariant, ...props }) => (
```
TheRealJon:
```suggestion
const ChartStack = ({ children, theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...props }) => (
```

packages/patternfly-4/react-charts/src/components/ChartStack/ChartStack.js
```diff
@@ -1,22 +1,32 @@
...
+  <VictoryStack theme={theme || getTheme(themeColor, themeVariant)} {...props}>
```
TheRealJon:
```suggestion
  <VictoryStack theme={theme} {...props}>
```

packages/patternfly-4/react-charts/src/components/ChartBar/ChartBar.js
```diff
@@ -1,18 +1,30 @@
...
+   * Specifies the theme color; blue (default), green, or multi-color. Overridden by the theme property.
+   */
+  themeColor: PropTypes.string,
+  /*
+   * Specifies the theme variant; 'dark' or 'light' (default). Overridden by the theme property.
+   */
+  themeVariant: PropTypes.string
 };
 
+const ChartBar = ({ theme, themeColor, themeVariant, ...rest }) => (
```
TheRealJon:
```suggestion
const ChartBar = ({ theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => (
```

packages/patternfly-4/react-charts/src/components/ChartBar/ChartBar.js
```diff
@@ -1,18 +1,30 @@
...
+  <VictoryBar theme={theme || getTheme(themeColor, themeVariant)} {...rest} />
```
TheRealJon:
```suggestion
  <VictoryBar theme={theme} {...rest} />
```

packages/patternfly-4/react-charts/src/components/ChartAxis/ChartAxis.js
```diff
@@ -1,19 +1,29 @@
...
+  themeColor: PropTypes.string,
+  /*
+   * Specifies the theme variant; 'dark' or 'light' (default). Overridden by the theme property.
+   */
+  themeVariant: PropTypes.string
 };
 
-const ChartAxis = (props) => (
-  <VictoryAxis theme={ChartTheme.default} {...props}/>
+const ChartAxis = ({ theme, themeColor, themeVariant, ...rest }) => (
```
TheRealJon:
```suggestion
const ChartAxis = ({ theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => (
```

packages/patternfly-4/react-charts/src/components/ChartAxis/ChartAxis.js
```diff
@@ -1,19 +1,29 @@
...
+  <VictoryAxis theme={theme || getTheme(themeColor, themeVariant)} {...rest}/>
```
TheRealJon:
```suggestion
  <VictoryAxis theme={theme} {...rest}/>
```

packages/patternfly-4/react-charts/src/components/ChartArea/ChartArea.js
```diff
@@ -1,18 +1,30 @@
...
+   * Specifies the theme color; blue (default), green, or multi-color. Overridden by the theme property.
+   */
+  themeColor: PropTypes.string,
+  /*
+   * Specifies the theme variant; 'dark' or 'light' (default). Overridden by the theme property.
+   */
+  themeVariant: PropTypes.string
 };
 
+const ChartArea = ({ theme, themeColor, themeVariant, ...rest }) => (
```
TheRealJon:
```suggestion
const ChartArea = ({ theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => (
```

packages/patternfly-4/react-charts/src/components/ChartArea/ChartArea.js
```diff
@@ -1,18 +1,30 @@
...
+  <VictoryArea theme={theme || getTheme(themeColor, themeVariant)} {...rest} />
```
TheRealJon:
```suggestion
  <VictoryArea theme={theme} {...rest} />
```

packages/patternfly-4/react-charts/src/components/Chart/Chart.js
```diff
@@ -1,21 +1,49 @@
...
+   * Because Victory renders responsive containers, the width and height props do not determine the width and
+   * height of the chart in number of pixels, but instead define an aspect ratio for the chart. The exact number of
+   * pixels will depend on the size of the container the chart is rendered into.
+   */
+  width: PropTypes.number
 };
 
-const Chart = ({ children, ...props }) => (
-  <VictoryChart theme={ChartTheme.default} {...props}>
+const Chart = ({ children, theme, themeColor, themeVariant, ...rest }) => (
```
TheRealJon:
```suggestion
const Chart = ({ children, theme = getTheme(themeColor, themeVariant), themeColor, themeVariant, ...rest }) => (
```

packages/patternfly-4/react-charts/src/components/Chart/Chart.js
```diff
@@ -1,21 +1,49 @@
...
+  <VictoryChart theme={theme || getTheme(themeColor, themeVariant)} {...rest}>
```
TheRealJon:
```suggestion
  <VictoryChart theme={theme} {...rest}>
```

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-donut.js
dlabrecq: I was keeping this as a default, in case height and width are overridden.

TheRealJon: If height/width are overridden, innerRadius should still be calculated from those values, right? If the calculated innerRadius doesn't meet the need of the user, they can override that too. I don't think there is any need for a static innerRadius value.

dlabrecq: The innerRadius will also be calculated if height and width are overridden. I'm thinking of an edge case where the value could be negative or the given height and width are zero. I just wanted to have a default.

packages/patternfly-4/react-charts/src/components/Chart/Chart.js
dlabrecq: Updated all components to set the theme default in this way, but will only comment once here.

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-donut.js
dlabrecq: I've pushed a change to enure the innerRadius doesn't go negative.

TheRealJon:
@dlabaj @tlabaj @jcaianirh @redallen 

Just want to put this back on your radar. Just needs one more review to merge.

PR #1968, opened 2019-05-10
feat(ts): convert aboutmodal to TS

redallen:
**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/components/AboutModal/index.js
```diff
@@ -1 +0,0 @@
-export { default as AboutModal } from './AboutModal';
```
seanforyou23: Please rename these files with git instead of deleting and re-adding so we can better preserve the history.

AboutModal.js => AboutModal.tsx
AboutModal.test.js => AboutModal.test.tsx
AboutModalBox.js => AboutModalBox.tsx
AboutModalBoxBrand.js => AboutModalBoxBrand.tsx
AboutModalBoxCloseButton.js => AboutModalBoxCloseButton.tsx
AboutModalBoxContent.js => AboutModalBoxContent.tsx
AboutModalBoxHeader.js => AboutModalBoxHeader.tsx
AboutModalBoxHero.js => AboutModalBoxHero.tsx
AboutModalContainer.js => AboutModalContainer.tsx
index.js => index.ts

packages/patternfly-4/react-core/scripts/convertComponentHelpers.js
```diff
@@ -1,98 +1,87 @@
 function getMatches(string, pattern) {
```
seanforyou23: Seems like the changes to convertComponentHelpers.js and convertComponentToTS.js would be better served as part of an independent PR.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+    trademark: '',
+    backgroundImageSrc: '',
+    noAboutModalBoxContentContainer: false
+  };
+
+  constructor(props: AboutModalProps) {
+    super(props);
+    this.state = {
+      container: undefined,
+      ariaLabelledBy: `pf-about-modal-title-${AboutModal.currentId++}`,
```
tlabaj: Can you remove these from the state.  They are not stateful.

packages/patternfly-4/react-core/src/components/AboutModal/helpers/AboutModalBox.tsx
```diff
@@ -0,0 +1,20 @@
...
+  /** additional classes added to the AboutModalBox  */
+  className?: string;
+};
+
+export const AboutModalBox: React.SFC<AboutModalBoxProps> = ({
+  children,
+  className = '',
+  ...props
+}: AboutModalBoxProps) => (
+  <div {...props} role="dialog" aria-modal="true" className={css(styles.aboutModalBox, className)}>
```
tlabaj: It probably does not make a difference since it is internal, but can you spread the props at the end here since you removed the aria attributes from the props so the overwrite any other props.

packages/patternfly-4/react-core/src/components/AboutModal/helpers/AboutModalBoxCloseButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  /** A callback for when the close button is clicked  */
+  onClose?: () => void;
+};
+
+export const AboutModalBoxCloseButton: React.SFC<AboutModalBoxCloseButtonProps> = ({
+  className = '',
+  onClose = () => undefined,
+  ...props
+}: AboutModalBoxCloseButtonProps) => (
+  <div {...props} className={css(styles.aboutModalBoxClose, className)}>
```
tlabaj: can you spread props at the end please.

packages/patternfly-4/react-core/src/components/AboutModal/helpers/AboutModalBoxContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const AboutModalBoxContent: React.SFC<AboutModalBoxContentProps> = ({
+  children,
+  className = '',
+  trademark,
+  id,
+  noAboutModalBoxContentContainer = false,
+  ...props
+}: AboutModalBoxContentProps) => (
+  <div {...props} className={css(styles.aboutModalBoxContent, className)} id={id}>
```
tlabaj: Can you spread props at the end please

packages/patternfly-4/react-core/src/components/AboutModal/helpers/AboutModalContainer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /** additional classes added to the About Modal Box  */
+  className?: string;
+  /** Flag to show the About Modal  */
+  isOpen?: boolean;
+  /** A callback for when the close button is clicked  */
+  onClose?: () => void;
+  /** Product Name  */
+  productName?: string;
+  /** Trademark information  */
+  trademark?: string;
```
tlabaj: Is there a reason this is no longer required?

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
```diff
@@ -0,0 +1,112 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/patternfly/components/Backdrop/backdrop.css';
+import { canUseDOM } from 'exenv';
+import { KEY_CODES } from '../../helpers/constants';
+import { AboutModalContainer } from './helpers/AboutModalContainer';
```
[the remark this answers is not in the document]
redallen: I'll move the files up a directory to be like all the other components.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
redallen: I'll use class attributes instead.

packages/patternfly-4/react-core/src/components/AboutModal/helpers/AboutModalContainer.tsx
redallen: Yes. I removed it to be consistent with `AboutModal` (the component consumers use), which does not require it. Otherwise a type error gets thrown in `AboutModal.tsx` when it constructs an `AboutModalContainer`.

packages/patternfly-4/react-core/scripts/convertComponentHelpers.js
redallen: Reverted.

packages/patternfly-4/react-core/src/components/AboutModal/helpers/AboutModalContainer.tsx
tlabaj: Nice catch.

PR #1970, opened 2019-05-10
Fixes bug with integration app react-core import

dlabaj:
Fixes bug with integration app @patternfly/react-core in package.json

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
+++ b/packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
@@ -12,12 +12,6 @@ export interface AvatarProps
   alt: string;
 }
 
-const defaultProps = {
-  className: '',
-  src: ''
-};
-
-
 export const Avatar: React.SFC<AvatarProps> = ({
   className,
   src,
@@ -26,5 +20,3 @@ export const Avatar: React.SFC<AvatarProps> = ({
 }: AvatarProps) => (
   <img {...props} src={src} alt={alt} className={css(styles.avatar, className)} />
 );
-
-Avatar.defaultProps = defaultProps;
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Label/Label.tsx
```diff
@@ -11,13 +11,22 @@ export interface LabelProps extends React.HTMLProps<HTMLSpanElement> {
...
 export const Label: React.FunctionComponent<LabelProps> = ({
   children,
-  className = '',
-  isCompact = false,
+  className,
+  isCompact,
   ...props
-}) => (
+}) => {
+  console.log(`isCompact = ${isCompact}`)
```
rebeccaalpert: Is there supposed to be a console.log?

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
```diff
@@ -13,8 +13,8 @@ export interface AvatarProps
 }
 
 export const Avatar: React.SFC<AvatarProps> = ({
-  className = '',
```
redallen: These empty string defaults should be added back!

packages/patternfly-4/react-core/src/components/Label/Label.tsx
dlabaj: Nope thanks.

packages/patternfly-4/react-core/src/components/Avatar/Avatar.tsx
dlabaj: Thanks removed the default props like we discussed and added these back.

dlabaj:
Closing pull request the fix will be add here https://github.com/patternfly/patternfly-react/pull/1971

PR #1971, opened 2019-05-10
fix(demo-app): remove demo-app from build

redallen:
**What**: Fixes common build issues with react-demo-app. TypeChecking still happens through `tsc`, so the passed props are still getting checked (like in the Jest tests). If we really want consumers to be able to import our types easier, we should look into exporting a single types file and maybe even hosting it in [@types](https://github.com/DefinitelyTyped/DefinitelyTyped)!

**Additional issues**:


packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TooltipDemo/TooltipDemo.tsx
```diff
@@ -1,15 +1,12 @@
-import { Tooltip, TooltipProps } from '@patternfly/react-core';
```
dlabaj: Missed TooltipProps

PR #1973, opened 2019-05-12
feat(page): add fill and no-fill modifiers to page sections

### LINKED ISSUE #1707: Add fill/no-fill modifiers to page component main sections
https://github.com/patternfly/patternfly-next/pull/1689

boaz0:
**What**:

closes #1707 

Added the `no-fill` and `fill` modifiers to the PageSection component + added an example to `Page.md`.

//cc @mcoker @mcarrano

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Page/Page.md
+++ b/packages/patternfly-4/react-core/src/components/Page/Page.md
@@ -129,3 +129,52 @@ class VerticalPage extends React.Component {
   }
 }
 ```
+## Main section fill/no-fill modifiers
+```js
+import React from 'react';
+import { Page, PageHeader, PageSidebar, PageSection, PageSectionVariants } from '@patternfly/react-core';
+
+class FillPage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isNavOpen: true
+    };
+    this.onNavToggle = () => {
+      this.setState({
+        isNavOpen: !this.state.isNavOpen
+      });
+    };
+  }
+
+  render() {
+    const { isNavOpen } = this.state;
+
+    const logoProps = {
+      href: 'https://patternfly.org',
+      onClick: () => console.log('clicked logo'),
+      target: '_blank'
+    };
+    const Header = (
+      <PageHeader
+        logo="Logo"
+        logoProps={logoProps}
+        toolbar="Toolbar"
+        avatar=" | Avatar"
+        showNavToggle
+        isNavOpen={isNavOpen}
+        onNavToggle={this.onNavToggle}
+      />
+    );
+    const Sidebar = <PageSidebar nav="Navigation" isNavOpen={isNavOpen} />;
+
+    return (
+      <Page header={Header} sidebar={Sidebar}>
+        <PageSection style={{height: '10em'}}>This section is set to the default fill variant</PageSection>
+        <PageSection style={{height: '10em'}} isFilled={true}>This section fills the available space.</PageSection>
+        <PageSection style={{height: '10em'}} isFilled={false}> This section does not fill the available space.</PageSection>
+      </Page>
+    );
+  }
+}
+```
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/components/Page/PageSection.js
```diff
@@ -7,7 +7,9 @@ export const PageSectionVariants = {
   default: 'default',
   light: 'light',
   dark: 'dark',
-  darker: 'darker'
+  darker: 'darker',
+  fill: 'fill',
+  no_fill: 'no_fill',
```
dgutride: I'm not sure if we've defined this anywhere - but the underscore seems unusual in the naming here.  I think we should chat about standardizing on camelCase as we do with other variable names and style names.  @tlabaj - do we have this documented?

tlabaj: Yes our standard is camelCase.  It is not documented, but we should add that.

packages/patternfly-4/react-core/src/components/Page/Page.md
```diff
@@ -129,3 +129,51 @@ class VerticalPage extends React.Component {
   }
 }
 ```
+## Main Section Fill/No-Fill Modifiers
```
tlabaj: Can you make this sentence case.

mcarrano:
I will defer to @mcoker on this.

packages/patternfly-4/react-core/src/components/Page/Page.md
boaz0: Absolutely :+1:

jschuler:
the example at https://pf4.patternfly.org/components/Page/examples/ shows the fill/no-fill modifier a bit better because the workspace's wrapping container has a height of 30em set on it. Maybe we could add `style={{height: '30em'}}` on the `<Page>` component in the example?
Also, i see an issue where you might want to add multiple modifiers, such as fill and dark. Seems like you should be able to combine them @mcoker ?

mcoker:
@jschuler yep, you should be able to use multiple modifiers, like `<div class="pf-c-page__main-section pf-m-light pf-m-no-padding pf-m-fill.....">`

boaz0:
@jschuler updated the height to '10em` because 30 felt too big to me. Let me know what you think.
@mcoker & @tlabaj I can work on multiple modifiers, should I do it here or open a new PR?

Thanks.

jschuler:
@boaz0 I am thinking, instead of adding fill/noFill to the variant list, let's keep the variant prop just for colors (light/dark/darker...) and maintain separate props for fill/noFill (just like we do for padding)

jschuler:
What do you think about making fill a boolean prop?
true: add fill modifier
false: add no-fill modifier
default: undefined (no modifier)

packages/patternfly-4/react-core/src/components/Page/PageSection.js
```diff
@@ -19,6 +19,8 @@ const propTypes = {
   variant: PropTypes.oneOf(Object.values(PageSectionVariants)),
   /** Modifies a main page section to have no padding */
   noPadding: PropTypes.bool,
+  /** Modifies a main page section to have filling/no-filling */
+  fillVariant: PropTypes.bool,
```
tlabaj: Since this is a boolean, I think a better name might be `isFilled` (That is what we have for other components like Card).  And I might update the description to say 'Enables the page section to fill the available vertical space`

packages/patternfly-4/react-core/src/components/Page/PageSection.js
```diff
@@ -19,6 +19,8 @@ const propTypes = {
   variant: PropTypes.oneOf(Object.values(PageSectionVariants)),
   /** Modifies a main page section to have no padding */
   noPadding: PropTypes.bool,
+  /** Enables the page section to fill the available vertical space */
+  isFill: PropTypes.bool,
```
tlabaj: Can you change the name to `isFilled` to match other components.

boaz0:
Thank you!

PR #1975, opened 2019-05-12
fix(Table): fix Table tests snapshots

Laviro:
found that some tests are failing..


Laviro:
or maybe it is only in my stack? waiting to see if someone else gets this test failing before I close..

MariaAga:
@Laviro I think its just your stack, on master the table tests run with no error, on this branch they fail for me

dgutride:
isGrouped is a valid prop on the dropdown - these tests are working on master, moving to closed based on the above comments - you might need to delete your node_modules/reinstall/rebuild

PR #1993, opened 2019-05-13
fix(wizard): wizards props extends div element props

### LINKED ISSUE #1916: Wizard Component Props should extend the correct React HTMLProp
The Wizard Component exported interfaces should extend the correct HTMLProp.

boaz0:
**What**:

closes #1916

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
@@ -42,7 +42,7 @@ export interface WizardStep {
 
 export type WizardStepFunctionType = (newStep: { id?: string | number; name: string; }, prevStep: { prevId?: string | number; prevName: string; }) => void;
 
-export interface WizardProps {
+export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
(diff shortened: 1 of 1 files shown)


boaz0:
Hi @tlabaj 
That's a good question.
From looking at the code examples I see that in each step there are no HTMLProps like `style`, `className` or `id`:

```js
    const steps = [
      { name: 'Step 1', component: <p>Step 1</p> },
      { name: 'Step 2', component: <p>Step 2</p> },
      { name: 'Step 3', component: <p>Step 3</p> },
      { name: 'Step 4', component: <p>Step 4</p> },
      { name: 'Review', component: <p>Review Step</p>, nextButtonText: 'Finish' }
    ];
```

Moreover, looking [at the part in the code](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Wizard/WizardToggle.tsx#L73) where the step component is rendered we are not passing to that component any props from `WizardStepProps`.

So I assume `WizardStepProps` should stay the way it is.

boaz0:
Thank you so much @jschuler @dlabaj @tlabaj

PR #2001, opened 2019-05-14
React-wooden-tree major version update and update the example accordingly

brumik:
**What**: Updated the react-wooden-tree version (from 1.2.* to 2.0.1). Made sure that the example provided is changed accordingly and working with the non-redux part of the tree. The example looks the same with a slightly changed callback function and with one more helper function call.

**Additional issues**: -


jeff-phillips-18:
@redallen Looks like this should be a major release as well, what is the current mechanism for that?

redallen:
Change the commit and squash message to `perf(react-wooden-tree): major version bump` and `lerna version` will deal with it automatically.

jeff-phillips-18:
@redallen Setting the commit message as suggested did not cause a major release bump. Can you fix it?

PR #2011, opened 2019-05-15
React-Virtualized PF4 tables

### LINKED ISSUE #1229: PF4: Virtualized Table/Virtualized Extension

priley86:
**What**:
Refactors #1250 
Fixes #1229 #1628 

Downstream PR:
https://github.com/openshift/console/pull/1465

Q&A 🍿 🍰 

**What's wrong with [reactabular-virtualized](https://reactabular.js.org/#/features/virtualization)?**
As noted in #1250 - dynamic row heights cause a problem when scrolling really long lists. In order to correctly calculate scroll offset, we need to measure all rows. We can't do this when the user drag scrolls an extremely long list. If using fixed heights though, this solution works perfectly fine.

**Why bring in react-virtualized Grid source?**
By default, Grid renders div's instead of `tbody` and `tr`'s [here](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L1048). This negates all PF4 table styling for table body's and table row's.

**Why am I not using @patternfly/react-table TableBody?**
 React-Virtualized uses absolute row positions to resolve dynamic row height issue mentioned above. This breaks the default layout for PF4 table rows. However, we can still emulate PF4 styling by assigning `tr`'s with `display: table; table-layout: fixed`. As such, I am defining a new API surface for table body and table row which works with react-virtualized, but suggests the common use of Table/TableHeader for defining table headers.

**Why am I going through this trouble?**
OpenShift currently has very specific requests about emulating this virtualization behavior as well as retaining the existing mobile behavior. However, integrating @patternfly/react-table now will open the door for using our React tables out-of-the-box when virtualization is not needed. It should also work w/ vanilla PF4 if we were to choose pagination or infinite scroll for some other visualizations (rather than virtual scroll) giving us a more common API surface.

Running Demo:
https://2011-pr-patternfly-react-patternfly.surge.sh/patternfly-4/virtual%20scroll/virtualized/

**A11Y notes:**
The DOM structure allowed by react-virtualized does not fit conventional grid DOM hierarchy b/c it relies on having a nested `rowgroup` element that is scrollable. To resolve this issue, I am using the following fix for fixing incorrect DOM structures with roles/aria. I've tested it with VoiceOver on Mac and it seems to work quite well now. It also passes aXe audits in Chrome.
https://www.html5accessibility.com/tests/aria-table-fix.html

Additional A11y articles on this by Steve Faulkner:
https://developer.paciellogroup.com/blog/2014/10/notes-on-fixing-incorrect-table-structure-using-aria/
https://developer.paciellogroup.com/blog/2018/03/short-note-on-what-css-display-properties-do-to-table-semantics/
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role

remaining todos:
- [ ] tests
- [ ] convert flow js to typescript where necessary
- [x] ensure `aria-label` can be passed as `null` if `role === 'presentation'`. Presentational `table` elements do not get the same aria attributes.
- [x] ~prototype expand/collapse. Expanded rows will be rendered as separate `tr`'s~ Determined this is not needed downstream at the moment. Can be added in a future PR.

cc: @rhamilto @suomiy 

**Additional issues**:


dlabaj:
@karelhala Can you review this one.

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualGrid.js
```diff
@@ -0,0 +1,1477 @@
...
+          ...containerStyle
+        }
+      };
+      innerScrollContainer = React.createElement(
+        innerScrollContainerComponent,
+        innerScrollContainerProps,
+        childrenToDisplay
+      );
+    }
+    return React.createElement(scrollContainerComponent, scrollContainerProps, [
```
priley86: I'll open an upstream PR to [react-virtualized](https://github.com/bvaughn/react-virtualized) for this piece... the consumer should be able to override the default scroll elements rendered.

priley86: this would also add support for the implementation of a virtual scrolling Data List in the future...

jeff-phillips-18:
Set a very long string in one of the columns:

image

Seems the height of the row is not growing correctly and it's over writing the next two rows.

jeff-phillips-18:
The row height issue is not relevant as the height is computed based on the data. I changed the value via the inspector which does not cause a recompute. Sorry for any confusion.

jeff-phillips-18:
When I run locally the column headers align just fine. Using the [demo](http://virtualized-pf-tables.surge.sh/virtual%20scroll/virtualized/) link from above, they don't seem to align correctly.

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualGrid.js
priley86: i've opened [#1377](https://github.com/bvaughn/react-virtualized/pull/1377) upstream in react-virtualized. This would simplify the API consumption there quite a bit if merged.

priley86:
thanks @karelhala - i still have a couple todo's here. Will try to update the PR again soon.

priley86:
thanks @dlabaj and @redallen !

I would like to begin integration testing this downstream and open this as a prerelease for testing to others. I've filed #2179 for future follow up... let me know if anything else is needed before merge.

priley86:
cc: @dgutride

Issue #1842, opened 2019-04-24
Select - TypeAhead variants - keyboard and screen reader accessibility


jgiardino:
The variants of the Select component that include TypeAhead should be updated to support keyboard and screen reader accessibility.

The behavior we have in the [PF3 TypeAhead Select](https://patternfly-react.surge.sh/patternfly-3/?selectedKind=patternfly-react%2FForms%20and%20Controls%2FTypeAhead%20Select&selectedStory=Menu%20Alignment&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) works really well, and matches what I would expect for the PF4 implementation. I suggest reviewing how that component is implemented and how it behaves. 

One behavior to check out is how the arrow keys are able to change the selected value while focus remains in the input field. The ability to switch between arrow keys and typing is pretty seamless.

The way the PF3 component handles the menu keyboard interaction is different in that they use `aria-activedescendant` to manage focus instead of `tabindex`, which is what we have with the original Select variants. Both methods are valid, but the `aria-activedescendant` method lets you keep focus on the input, but still use the arrow keys to change the selection, which is more appropriate for this case.  WAI-ARIA describes this method here: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_focus_activedescendant.

rachael-phillips:
Closing this issue as it has been resolved

PR #2013, opened 2019-05-15
feat(Select): add typeahead keyboard navigation

kmcfaul:
**What**: Adds keyboard interaction to typeahead variants, other variants should remain unchanged.
Enter, with select closed, toggles open the menu.
Focus should remain on input.
Arrow up and down navigates the options with a focus styling.
Enter, with select opened, selects the highlighted item.

**Refer to issue**: #1842


redallen:
You have some merge conflicts now, but I like the interaction on the demo!

PR #2025, opened 2019-05-16
fix(Dropdown): handle on enter focusing when options are disabled

### LINKED ISSUE #2017: PF4 Dropdown throws error if first DropdownItem is disabled with Tab key
Reproduced in StackBlitz: https://react-ax9nl6.stackblitz.io/

If you have something like:
```
<Dropdown>
  <DropdownItem isDisabled>Foo</DropdownItem>
  <DropdownItem>Bar</DropdownItem>
</DropDown>
```

If you try to tab over to the dropdown and hit `Enter` to open it, a console error is thrown: 
`Cannot read property 'focus' of null`

This also removed the Dropdown from the page and it will only appear on reload.

kmcfaul:
**What**: fix for handling when options are disabled. Focus should be on the first non-disabled item.

**Refer to issue**: Fixes #2017

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
@@ -39,10 +39,14 @@ class DropdownMenu extends React.Component {
 
   componentDidMount() {
     if (this.props.openedOnEnter) {
-      if (this.props.component === 'ul') this.refsCollection[0].focus();
+      const focusTarget =
+        this.refsCollection.filter(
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
```diff
@@ -39,10 +39,14 @@ class DropdownMenu extends React.Component {
 
   componentDidMount() {
     if (this.props.openedOnEnter) {
-      if (this.props.component === 'ul') this.refsCollection[0].focus();
+      const focusTarget =
+        this.refsCollection.filter(
```
redallen: In the future, you can use `.find()` instead of `.filter()[0]`

PR #2037, opened 2019-05-18
fix(checkbox): set checked if onChanged is not the default

### LINKED ISSUE #2036: Checkbox regression (uncontrolled version)
I just realized that my change https://github.com/patternfly/patternfly-react/pull/1929 causes a regression in the uncontrolled version.

The `isChecked` prop is set to false which is then making the uncontrolled version being controlled.
Need to check if the user wants to make it controlled or not by verifying if `onChange` was set too. If not, then the `checked` prop in input should not be set.

boaz0:
**What**:

fixes #2036 

This change checks whether `onChange` is set by the user.
If it does, it means the component is controlled by the user. In this case, the input `checked` props should be set by `isChecked/checked` props passed to `Checkbox`. Otherwise, set input `checked` only if `checked` is true or false or `isChecked` is true (because `isChecked` is Boolean TS by default sets it to false).

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
@@ -24,14 +24,16 @@ export interface CheckboxProps
   'aria-label': string;
 }
 
+// tslint:disable-next-line:no-empty
+const defaultOnChange = () => {};
+
 export class Checkbox extends React.Component<CheckboxProps> {
   static defaultProps = {
     className: '',
     isValid: true,
     isDisabled: false,
     isChecked: false,
-    // tslint:disable-next-line:no-empty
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -54,14 +54,18 @@ export class Checkbox extends React.Component<CheckboxProps> {
       isChecked,
       label,
       checked,
+      defaultChecked,
```
redallen: [Default checked](https://www.w3schools.com/jsref/prop_checkbox_defaultchecked.asp) support! :+1:

boaz0:
Thank you @redallen @dlabaj @tlabaj! :smile:

PR #2047, opened 2019-05-21
fix(Select): onSelect prop TypeScript definition

SpyTec:
**What**: TypeScript definition file of Select component. This makes the `onSelect` prop required, and now has the correct parameters. Selections can now also take in null, which is the default value used within the Select component

From:
```js
onSelect(event: ReactEventHandler<HTMLOptionElement, Event>)
//...
selections?: string | Array<string>;
```
to
```js
onSelect(event: ReactEventHandler<HTMLOptionElement, Event>, element: string, isPlaceholder: boolean): void;
//...
selections?: string | Array<string> | null;
```

**Additional issues**: None

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.d.ts
@@ -11,9 +11,10 @@ export interface SelectProps extends HTMLProps<HTMLOptionElement> {
   isExpanded?: boolean;
   isGrouped?: boolean;
   onToggle(value: boolean): void;
+  onSelect(event: ReactEventHandler<HTMLOptionElement, Event>, element: string, isPlaceholder: boolean): void;
   onClear?() : void;
   placeholderText?: string | ReactNode;
-  selections?: string | Array<string>;
+  selections?: string | Array<string> | null;
   variant?: string;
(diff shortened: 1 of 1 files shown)


rhamilto:
I believe this change results in the following error in OpenShift console.  @dlabrecq 
```
ERROR in /Users/rhamilto/go/src/github.com/openshift/console/frontend/node_modules/@patternfly/react-core/dist/js/components/Select/Select.d.ts(10,18):
TS2430: Interface 'SelectProps' incorrectly extends interface 'HTMLProps<HTMLOptionElement>'.
  Types of property 'onSelect' are incompatible.
    Type '(event: any, element: string, isPlaceholder: boolean) => void' is not assignable to type '(event: SyntheticEvent<HTMLOptionElement, Event>) => void'.
ERROR in /Users/rhamilto/go/src/github.com/openshift/console/frontend/node_modules/@patternfly/react-core/dist/js/components/Select/Select.d.ts(14,19):
TS2304: Cannot find name 'ReactEventHandler'.
Version: typescript 3.4.4
```

SpyTec:
Hi @rhamilto, sorry for the error I introduced. I believe this was fixed in #2107, try upgrading to ^3.32.8

Issue #1739, opened 2019-04-09
Tabs - scroll buttons are missing


jgiardino:
The buttons are in the dom, they just don't become visible when expected.

PR #2060, opened 2019-05-22
Tabs: fix scrollable tabs

jschuler:
Closes: #1739 

Makes the scroll buttons visible again when there is tabs overflow

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
+++ b/packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
@@ -247,7 +247,7 @@ class Tabs extends React.Component {
             </button>
           )}
         </Component>
-        {children.map((child, index) =>
+        {React.Children.map(children, (child, index) =>
           !child.props.children ? null : (
             <TabContent key={index} activeKey={activeKey} child={child} id={child.props.id || uniqueId} />
           )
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -264,72 +174,85 @@ class InternalTabs extends React.Component {
...
-          <button
-            className={css(styles.tabsScrollButton)}
-            variant="plain"
-            aria-label={rightScrollAriaLabel}
-            onClick={this.scrollRight}
-          >
-            <AngleRightIcon />
-          </button>
+                <Tab
+                  {...child.props}
```
redallen: Can we stay consistent and propogate at the end?

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
```diff
@@ -264,72 +174,85 @@ class InternalTabs extends React.Component {
...
+              className={css(styles.tabsScrollButton)}
+              variant="plain"
+              aria-label={rightScrollAriaLabel}
+              onClick={this.scrollRight}
+            >
+              <AngleRightIcon />
+            </button>
+          )}
+        </Component>
+        {children.map((child, index) =>
```
redallen: Use `React.Children` instead. `children` could be a string, among other things. Might need to test for that.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.js
jschuler: yes we should thanks

jschuler:
@redallen thanks addressed your comments

PR #2062, opened 2019-05-22
fix(Charts): Fix type definitions for chart themes

### LINKED ISSUE #2061: Type declarations related to chart themes are incorrect.
When building with TypeScript 3.4.4, receive several compile-time errors related to chart theme types. Also, importing ChartThemeColor and ChartThemeVariant in a TypeScript file will import the types instead of instances. The interfaces need to be associated with declared variables in order for module imports to work correctly.

TheRealJon:
Fixes #2061

**What**:
- ChartBaseTheme and ChartDonutTheme type definitions were using `const` instead of `type` or `interface`. Update to interfaces which extend VictoryThemeInterface.
 - None of the types in ChartTheme.d.ts were associated with variable declarations, so module imports did not work as expected. Declare variables to associate types with instances.

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/ChartTheme/ChartTheme.d.ts
+++ b/packages/patternfly-4/react-charts/src/components/ChartTheme/ChartTheme.d.ts
@@ -1,15 +1,19 @@
-export const ChartBaseTheme = any;
-export const ChartDonutTheme = any;
+import { VictoryThemeInterface, VictoryThemeDefinition } from "victory";
 
-export interface ChartThemeColor {
+interface ChartBaseThemeInterface extends VictoryThemeDefinition {}
+interface ChartDonutThemeInterface extends VictoryThemeDefinition {}
+
+interface ChartThemeColorInterface {
   blue: string;
   default: string;
   green: string;
   multi: string;
-};
+}
 
-export interface ChartThemeVariant {
+interface ChartThemeVariantInterface {
   dark: string;
   default: string;
   light: string;
-};
+}
+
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-charts/src/components/ChartTheme/ChartTheme.d.ts
```diff
@@ -1,15 +1,19 @@
...
 
-export interface ChartThemeVariant {
+interface ChartThemeVariantInterface {
   dark: string;
   default: string;
   light: string;
-};
+}
+
+export declare const ChartThemeColor: ChartThemeColorInterface, ChartThemeVariant: ChartThemeVariantInterface, ChartBaseTheme: ChartBaseThemeInterface, ChartDonutTheme: ChartDonutThemeInterface;
```
redallen: Good catch that they were named the same as their components.

packages/patternfly-4/react-charts/src/components/ChartTheme/ChartTheme.d.ts
```diff
@@ -1,15 +1,19 @@
-export const ChartBaseTheme = any;
-export const ChartDonutTheme = any;
+import { VictoryThemeInterface } from "victory";
 
-export interface ChartThemeColor {
+interface ChartBaseThemeInterface extends VictoryThemeInterface {}
+interface ChartDonutThemeInterface extends VictoryThemeInterface {}
```
dlabrecq: Believe this should be VictoryThemeDefinition?

I wasn't using this interface because that is only a subset of the possible charts. Was thinking we needed to create our own? I encountered a type issue while trying to use VictoryThemeDefinition with my own custom theme in Cost Management -- had to use type any for now.

Screen Shot 2019-05-22 at 3 45 41 PM

Everything else looks fine.

TheRealJon: Updated

PR #2072, opened 2019-05-23
fix(docs): use frontmatter for what props to show in docs

### LINKED ISSUE #2003: Some components are missing prop documentation
Table, charts, alert docs are all missing props - need to fix this before next push to org site.

redallen:
**What**: Fixes #2003 

**Additional issues**:


dlabaj:
@redallen Popover.md has a conflict.

PR #2085, opened 2019-05-24
feat(react-styles): move react styles

redallen:
**What**: Change all the React css imports and add types for them. Instead of using a babel plugin to generate the JS at compile time for our CSS-in-JS solution, we statically generate it beforehand in react-styles.

**Additional issues**:


dlabaj:
@redallen Looks like there's a conflict with .babelrc

PR #1109, opened 2018-12-30
feat(BreadcrumbSwitcher): creating a BreadcrumbSwitcher component

glekner:
**What**:
creating a BreadcrumbSwitcher component
screenshot 2018-12-30 15 59 30

**Additional issues**:

PR #2097, opened 2019-05-28
fix(index.js): [PF3] Export BreadcrumbSwitcher from patternfly-react index

mturley:
**What**:

In https://github.com/patternfly/patternfly-react/pull/1109, the new `BreadcrumbSwitcher` component was added, but it was never exported in the main `index.js` file, so it can't be imported by name from the package like `import { BreadcrumbSwitcher } from 'patternfly-react';`, instead it can only be imported from `node_modules/patternfly-react/dist/js/components/BreadcrumbSwitcher`. This PR just adds it to `index.js` so it can be imported normally.

Needed downstream for https://github.com/ManageIQ/manageiq-v2v/issues/865.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/index.js
+++ b/packages/patternfly-3/patternfly-react/src/index.js
@@ -57,3 +57,4 @@ export * from './components/Wizard';
 export { patternfly, layout } from './common/patternfly';
 export * from './components/ExpandCollapse';
 export * from './components/Overlay';
+export * from './components/BreadcrumbSwitcher';


mturley:
Weird... I'm not sure why `Popover` would be undefined when setting up `propTypes` there. We use that pattern elsewhere (e.g., TableGridColumnHeader uses `...Grid.Col.propTypes`). It seems like a lot of our components that consume other components from the main index import them `from '../../index'` like this BreadcrumbSwitcherPopover does, but others import from `'patternfly-react'`. I'll try switching it to the latter and see if that helps. Pretty weird

mturley:
@redallen I played around with a few things and couldn't figure out how to fix it in a sensible way, but I found that moving the new export to the end of `index.js` fixes it... presumably because when it's being exported, it is also importing things from that same file, and they need to have been exported already. Somehow that only fails when running Jest and not the build.

Force-pushed with the export at the end this time... ¯\\\_(ツ)\_/¯

redallen:
Yeah, that seems about right. Will approve once build passes.

PR #2101, opened 2019-05-29
feat(PF4: Application launcher): Use dropdown components in App launcher

karelhala:
**What**:
Application launcher menu is esentially dropdown with predefined values so let's use dropdown components here as well so if dropdown changes we get same changes in app launcher as well.

Notice how there are no changes to example and a lot of code could be removed due to reuse of components.

**Fixes**:
* App launcher can now be disabled
* App menu can now be positioned in same way as dropdown is positioned (**left**, right, up and **down**)
* When App menu is opened correct element is used as `aria-labelledby` (previously it had same value as `aria-label` and no ID reference)
* Tests were completely incorrect in original implementation

**TODO**
* [x] Merge after https://github.com/patternfly/patternfly-react/pull/2227


packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.js
```diff
@@ -1,89 +1,52 @@
...
-    if (!this.props.isOpen) this.openedOnEnter = false;
-  }
-
-  render() {
-    const { 'aria-label': ariaLabel, dropdownItems, className, isOpen, onSelect, onToggle } = this.props;
-    return (
-      <GenerateId>
-        {randomId => (
-          <div
-            className={css(styles.appLauncher, isOpen && styles.modifiers.expanded, className)}
```
redallen: What's the practical difference between `pf-c-app-launcher` and `pf-c-dropdown` @mcoker ? The [current AppLauncher](http://patternfly-react.surge.sh/patternfly-4/components/applicationlauncher/) and [this Dropdown AppLauncher](https://2101-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/applicationlauncher/) look the same.

mcoker: @redallen sorry I missed this. They're effectively the same currently. The app launcher is a dropdown with `.pf-m-plain` on the `.pf-c-dropdown__toggle` element, just using a different icon than we have in our examples.

What the dropdown doesn't allow for is a future variation of the application launcher where we re-arrange the items in the dropdown to be square or whatever (like this - https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/application-launcher.html)

mcoker:
@karelhala this would work for now, but as mentioned [here](https://github.com/patternfly/patternfly-react/pull/2101#discussion_r290895082), there are plans to add additional menu list layouts to the application launcher that will not necessarily be added to the dropdown component. I'm not sure when we would add those. We have the [common dropdown toggle and context menu issue](https://github.com/patternfly/patternfly-next/issues/1429) but it isn't slated in our upcoming milestones, so we have to work with both the dropdown and application launcher components for now.

karelhala:
> @karelhala this would work for now, but as mentioned [here](https://github.com/patternfly/patternfly-react/pull/2101#discussion_r290895082), there are plans to add additional menu list layouts to the application launcher that will not necessarily be added to the dropdown component. I'm not sure when we would add those. We have the [common dropdown toggle and context menu issue](https://github.com/patternfly/patternfly-next/issues/1429) but it isn't slated in our upcoming milestones, so we have to work with both the dropdown and application launcher components for now.

This is esentially another component it jus uses dropdown for sharing logic and the heavy coding, otherwise we'd have to copy all things from dropdown and whenever something changes there it should have be implemented here as well.

Regarding your concerns about different style of menu, with this implementation you are free to update the menu to tiles as well. You see dropdown items can be any HTML element so we can pass icons there if we want to. The only problem I see here is if we'd were to add some classes to dropdown menu, but that is doable by adding new prop to dropdown component and pass additonal classes so they can improve how components are layed out.

mcoker:
@karelhala thanks. Though since this component is using `pf-c-dropdown` classes, wouldn't that imply that if we wanted to support an alternate layout for the application launcher component, we would need to add that functionality to the dropdown component instead, since this component doesn't use the application launcher's CSS?

karelhala:
@mcoker not really. I didn't include app launcher classes in menu, items and toggle because they are the same for now. But if you'd want to include these classes as well I can update dropdown component to allow passing custom classes for each section. That way we could pass classes from app luncher and we would have layout updated from them. So dropdown classes would work as default layout and such, app launcher classes would update these styles.

mcoker:
@karelhala cool, sounds good to me.

karelhala:
@mcoker your thoughts on these updates?

jschuler:
Hi @karelhala , this looks nice, I agree we should use DRY principle where possible and the AppLauncher is a good candidate. But I think to make this really work and future-proof it we'll need to lift some common functionality up a level from Dropdown, for example by creating a common set of components such as BaseDropdown, BaseMenu, BaseToggle components that can be used by both AppLauncher and Dropdown. This way they can run with their own set of classes (pf-c-dropdown / pf-c-app-launcher respectively) and there is less risk of style conflicts. Additionally, the AppLauncher can add its own differentiating functionality without impacting the existing Dropdown components later on.

There are some new feature additions planned for app launcher that should be happening within this sprint as @mcoker mentioned and which I am planning to add. I think it would be good to prioritize the feature additions first (while keeping what we plan to do here in mind), and either as part of the feature additions PR or as a follow-up we can lift common functionality out. Does this sound good to you @karelhala ?

karelhala:
> Hi @karelhala , this looks nice, I agree we should use DRY principle where possible and the AppLauncher is a good candidate. But I think to make this really work and future-proof it we'll need to lift some common functionality up a level from Dropdown, for example by creating a common set of components such as BaseDropdown, BaseMenu, BaseToggle components that can be used by both AppLauncher and Dropdown. This way they can run with their own set of classes (pf-c-dropdown / pf-c-app-launcher respectively) and there is less risk of style conflicts. Additionally, the AppLauncher can add its own differentiating functionality without impacting the existing Dropdown components later on.

I agree that we should update dropdown components to be more generic, however introducing new components for this is not really good idea. For starter we see that we need to update classes so I am going to update this PR to allow passing custom classes to dropdown with React context (already did that but in not so good way).

About the changes, you mean this PR https://github.com/patternfly/patternfly-next/pull/1916? This does not effect app luncher at all honestly, looking at the running styles this will be reallt easy to implement once this PR is merged. But sure, if you feel that these updates will be beter to be merged before this PR I can update it afterwards to match it as well.

karelhala:
Okay I did a bit of a refactor and now it's possible to change classes for base element, toggle, menu and item via React context. I have exported components that are using this context so API stayed the same, hence no need to file a new breaking change release.

@jschuler original tests were not good, they were just checking some random stuff and were completely off so I had to update them and correct them that's the reason why so many files are changed. Plus every component that was using Dropdown needed to update its snapshot (Provider and Consumer was placed there to work correctly with basic dropdown).

@dgutride look at how Dropdown is now using context, this should be prefered way and we should allow consumers this as well so they can override basic bahavior or add some custom logic to their components. Similiar thing is doing Material design and many other libraries.

karelhala:
As promised I have opened new PR to address wrapping dropdown with context so we don't have to deal with different classes on different components. https://github.com/patternfly/patternfly-react/pull/2227

jschuler:
Nice, let's get this rebased and then bring it in @karelhala !

Issue #2106, opened 2019-05-29
TS errors in Select


seanforyou23:
When importing latest `@patternfly/react-core` I'm met with errors for the Select component.

```
ERROR in [at-loader] ../../node_modules/@patternfly/react-core/dist/js/components/Select/Select.d.ts:10:18 
    TS2430: Interface 'SelectProps' incorrectly extends interface 'HTMLProps<HTMLOptionElement>'.
  Types of property 'onSelect' are incompatible.
    Type '(event: any, element: string, isPlaceholder: boolean) => void' is not assignable to type '(event: SyntheticEvent<HTMLOptionElement, Event>) => void'.

ERROR in [at-loader] ../../node_modules/@patternfly/react-core/dist/js/components/Select/Select.d.ts:14:19 
    TS2304: Cannot find name 'ReactEventHandler'.
```

<img width="1138" alt="Screen Shot 2019-05-29 at 1 00 18 PM" src="https://user-images.githubusercontent.com/5942899/58576206-c6085300-8211-11e9-9aef-34591ba2884b.png">

Related to https://github.com/patternfly/patternfly-react/issues/2096

tlabaj:
Closed by #2107

Issue #2099, opened 2019-05-28
Popover Regression


seanforyou23:
A recent [change](https://github.com/patternfly/patternfly-react/commit/be3d1c5e1c057245cb571fd9889ba6378fd35037#diff-d5b83b2e57be9a5646fe7d3bad5c38d9R32) broke the ability to use `position={PopoverPosition.right}` as a value for the position property on Popover. We should be able to use strings or enum values here.

redallen:
Closed via #2113

PR #2107, opened 2019-05-29
fix(Select): update SelectProps interface, clean up demos

seanforyou23:
**What**: This PR updates the SelectProps interface to avoid a type error when importing react-core into a TypeScript application. I also removed some unnecessary code in a couple of the demo's I followed recently. Copying the code as is was creating an error. I think they were safe deletions but worth double checking.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/2106

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Popover/Popover.md
+++ b/packages/patternfly-4/react-core/src/components/Popover/Popover.md
@@ -10,7 +10,7 @@ import { Popover, PopoverPosition, Checkbox, Button } from '@patternfly/react-co
 ## Simple popover
 ```js
 import React from 'react';
-import { Popover, PopoverPosition, Checkbox, Button } from '@patternfly/react-core';
+import { Popover, Button } from '@patternfly/react-core';
 
 SimplePopover = () => (
   <Popover
--- a/packages/patternfly-4/react-core/src/components/Select/Select.d.ts
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.d.ts
@@ -1,4 +1,5 @@
 import { HTMLProps, FormEvent, ReactNode } from 'react';
+import { Omit } from '../../helpers/typeUtils';
 
 export const SelectVariant: {
   single: 'single';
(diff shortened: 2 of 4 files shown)


packages/patternfly-4/react-core/src/components/Popover/Popover.md
```diff
@@ -10,7 +10,7 @@ import { Popover, PopoverPosition, Checkbox, Button } from '@patternfly/react-co
 ## Simple popover
 ```js
 import React from 'react';
-import { Popover, PopoverPosition, Checkbox, Button } from '@patternfly/react-core';
```
jschuler: Can you change `position="right"` to `position={PopoverPosition.right}` instead?

seanforyou23: Not at the moment - https://github.com/patternfly/patternfly-react/issues/2099

jschuler: Oh ok, probably the PopoverPosition const needs to be an enum then? Some of the other examples still use PopoverPosition it seems

seanforyou23: Screen Shot 2019-05-29 at 2 53 15 PM

seanforyou23: Sorry only now does GitHub finally show your comment from an hr ago.. Yep, that's the convention for finite sets of values like this.

I was only removing it from the example where it wasn't used at all, had fixed that locally before the unrelated regression of position property happened. I can remove the use of PopoverPosition object from the other examples but this type is in the process of being fixed and the functionality will be restored soon. I can remove these changes from this PR but hoping we can take this in as-is since the incremental step does improve things and also unblocks the bug we have for SelectProps. Would this be ok for a short term improvement?

jschuler: sounds good thanks @seanforyou23

PR #2112, opened 2019-05-29
fix(Charts): Fix themes for bar chart default colors

### LINKED ISSUE #2111: Chart theme does not apply base color to bar chart
If you don't wrap the `ChartBar` component with `ChartGroup` and a apply a theme, the bars will not have a theme color. Should be able to use a themed simple bar chart without using the `ChartGroup` component. I would expect that the default bar color would be the first color in the color scale.

```jsx
<Chart domainPadding={{ x: [30, 25] }}>
  <ChartBar data={[
    { x: 'Cats', y: 1 }, 
    { x: 'Dogs', y: 2 }, 
    { x: 'Birds', y: 5 }, 
    { x: 'Mice', y: 3 }
  ]} />
</Chart>
```

**Expected:**
image

**Actual:**
image

TheRealJon:
Fixes #2111

**What**:
Add default bar fill color to all themes so that standard (ungrouped) bar charts will have an appropriate fill color.


packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-color.js
```diff
@@ -45,6 +46,9 @@ export default props => {
     bar: {
       colorScale: COLOR_SCALE,
       style: {
+        data: {
+          fill: COLOR_BAR_FILL
+        },
```
dlabrecq: Instead of providing a separate COLOR_BAR_FILL var, my thought was to use the first color defined by the color scale.

dlabrecq: Ultimately, colors will come from pf-core via react-tokens. I'm working to make react-tokens available now.

For example, here is a snippet of colors defined for the bar chart below. 

Screen Shot 2019-05-29 at 4 21 05 PM

dlabrecq: Note that these colors are ordered.

TheRealJon: Yeah, that was my original implementation. I can change back to that if that's what we prefer.

dlabrecq: Yes, please. I'd like the default color to come from the ordered colorScale defined for each chart.

TheRealJon:
I have no idea why snapshots are failing. I've done everything I know to do to make sure the snapshots are up to date, and they should be, but for some reason, unit tests are still failing. @dlabrecq Any ideas?

dlabrecq:
I recall encountering a similar error while running only `yarn test:pf4 -u`, but running `yarn test` seemed to clear it up.

TheRealJon:
Tried that. Tried a clean build, yarn install, and full test run with `-u` flag, got no new snapshot updates.

TheRealJon:
@dlabrecq @tlabaj Could I get approvals again? I had to push updated test snapshots to get CI to pass.

TheRealJon:
@tlabaj Could you approve again? Looks like CI is passing now. Should be good to merge.

PR #2113, opened 2019-05-29
fix(popover): make PopoverPosition an enum

seanforyou23:
**What**: This PR changes PopoverPosition from a const to an enum. This allows users to specify their choice of a string or object property for the `position` property value on Popover component. Also adds a test that exercises this method to ensure it doesn't regress going forward.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/2099


jeff-phillips-18:
One small comment, If you could re-title the PR to make it clear that this is fixing the const (code) not the actual visual position of popovers. I expected something different when I read the title.

PR #2116, opened 2019-05-30
Make Switch OUIA compatible

quarckster:
**What**:

This PR makes Switch component OUIA compatible. What is OUIA? It's a specification from Red Hat QE team for making automation testing easier. Here you can read a discussion about that:
https://forum.patternfly.org/t/open-web-ui-design-specification-for-enabling-automation-ouia/154

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.js
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.js
@@ -45,7 +45,12 @@ class Switch extends React.Component {
   render() {
     const { id, className, label, isChecked, isDisabled, onChange, ...props } = this.props;
     return (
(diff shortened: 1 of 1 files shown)


redallen:
We have a ton of components with`<label>`s. Why just Switch?

quarckster:
Switch is just first shot, of course, we would like to see all components compatible.

redallen:
Have you considered a HOC to wrap all of our components?

This would probably be something we don't want published in our npm package since it would add so many `data-` attributes that are unnecessary for anything but testing. We could create a new `react-core-testing` package that programatically wraps every component in this way for you. What do you think?

redallen:
Closing in favor of #2303, POC PR coming soon.

PR #2117, opened 2019-05-30
feat(react-core): allow dynamic rendering of anchor components

### LINKED ISSUE #2063: PageHeader logo: Control over link element
We are using react-router for our app, which means we need to use their `<Link>` component for links to route correctly. The PageHeader is rendering an `<a>` for the logo, but we need a way to customize this as a `<Link>`.

redallen:
**What**: Allow custom rendering of all `<a>` tags used in PF4 by adding a `linkComponent` prop to the following components:
- LoginMainFooterLinksItem
- NavItem
- PageHeader
- WizardNavItem

 Closes #2063 .

**Additional issues**:


packages/patternfly-4/react-core/src/components/LoginPage/LoginMainFooterLinksItem.tsx
```diff
@@ -11,18 +11,25 @@ export interface LoginMainFooterLinksItemProps extends React.HTMLProps<HTMLLIEle
   target?: string;
   /** Additional classes added to the Footer Link Item  */
   className?: string;
+  /** Component used to render WizardNavItem */
```
jeff-phillips-18: copy/paste error?

packages/patternfly-4/react-core/src/components/Page/PageHeader.js
```diff
@@ -12,6 +12,8 @@ const propTypes = {
   logo: PropTypes.node,
   /** Additional props passed to the logo anchor container */
   logoProps: PropTypes.object,
+  /** Component to use to wrap the passed <logo> */
+  linkComponent: PropTypes.node,
```
jeff-phillips-18: maybe `logoComponent`?

packages/patternfly-4/react-core/src/components/Nav/NavItem.tsx
```diff
@@ -20,35 +20,39 @@ export interface NavItemProps extends Omit<React.HTMLProps<HTMLAnchorElement>,
   /** If true prevents the default anchor link action to occur. Set to true if you want to handle navigation yourself. */
   preventDefault?: boolean;
   /** Callback for item click */
-  onClick?: NavSelectClickHandler
+  onClick?: NavSelectClickHandler;
+  /** Component used to render NavItems */
+  linkComponent?: React.ReactNode;
```
jeff-phillips-18: 'navItemComponent' or simply 'component'?

packages/patternfly-4/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -11,10 +11,12 @@ export interface WizardNavItemProps {
...
   /** Whether the nav item is disabled */
   isDisabled?: boolean;
-  /** Callback for when the nav item is clicked */
-  onNavItemClick?(step: number): any;
   /** The step passed into the onNavItemClick callback */
   step: number;
+  /** Callback for when the nav item is clicked */
+  onNavItemClick?: (step: number) => any;
+  /** Component used to render WizardNavItem */
+  linkComponent?: React.ReactNode;
```
jeff-phillips-18: `navItemComponent`?

jeff-phillips-18:
Please squash commits unless they are broken up to make for an easier review.

jschuler:
@dlabaj @jeff-phillips-18 @redallen I think we should revert this change. Don't think this solves the issue of using react router Link component to render in logo. To make it work, I think you'd have to check if the passed in component is a react component like we do here: https://github.com/patternfly/patternfly-react/blob/edbe2524cf8acdbfaccbe44763728983c7a7f3a9/packages/patternfly-4/react-core/src/components/Nav/NavItem.tsx#L75

The current change to NavItem does not make sense because it needs to render as <a>, if you want to use another component that was already supported and it will render that instead.

Feel free to correct me if I'm wrong, but these props such as navItemComponent have to be valid HTML elements and cannot be a react nodes

Issue #2109, opened 2019-05-29
'RowWrapper' only refers to a type, but is being used as a value here.ts(2693)


Rezney:
While importing 'RowWrapper':

```javascript
import {
  RowWrapper
} from '@patternfly/react-table';
```

and trying to use it:

E.g: 
```javascript
console.log(<RowWrapper></RowWrapper>)
```

I get `'RowWrapper' only refers to a type, but is being used as a value here.ts(2693)` error.

Versions:

```javascript
"@patternfly/react-core": "3.32.5"
"@patternfly/react-table": "2.8.14"
"typescript": "3.4.5"
```

karelhala:
The thing here is that we are using React Tabular[1] and `RowWrapper` is used as cusom rendered[2] so you shouldn't use `RowWrapper` directly, typescript definition is just interface.

By default `RowWrapper` implementation is used directly by the table, if you however want to create custom wrapper for rows (for instance add some special attribute to each `tr`) import this interface and implement it, and then pass this new implementation as `rowWrapper` property.

[1] https://reactabular.js.org/#/
[2] https://reactabular.js.org/#/table/overriding-default-renderers

Rezney:
@karelhala I see. Thanks for checking this. I will close this case then.

shaded-enmity:
@karelhala I think there's a misunderstanding here, see:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/RowWrapper.js
Which is an actual implementation of said interface, essentially we just need to squeeze in an additional class - copy & pasting the aforementioned implementation is a possibility, but it doesn't feel right, we want to compose.
```jsx
<Table rowWrapper={() => <RowWrapper className='our-custom-class' />}>
```

karelhala:
@shaded-enmity oh I see, mhh it's really unfortunate that we did not exported implementation in TS as well. This issue fixes it https://github.com/patternfly/patternfly-react/pull/2118 however I don't know when it's going to be merged because it might cause breaking changes.

PR #2118, opened 2019-05-30
fix(PF4 Table): Export RowWrapper implementation

karelhala:
**What**:
When using react table in typescript project there is no way how to import `RowWrapper` implementation and copy pasting parts of it is bad. So this PR fixes this problem by exporting `RowWrapperProps` as interface and `RowWrapper` as implementation.

**Additional issues**:
There might be problems because we are changing interface to implementation, but this is not that greatly used interface so it should be fine. @suomiy, @priley86 are you using this interface somewhere?

Fixes: https://github.com/patternfly/patternfly-react/issues/2109


redallen:
Fixes #2109

PR #2120, opened 2019-05-30
feat(Modal): Allow custom header and footer

### LINKED ISSUE #1545: Modal - add render prop to allow for custom title content
We would like to more complex content to the title of a `Modal`. This isn't possible now with the existing `title` string prop. Please add support for a render prop to support custom title rendering with more complex content.

jeff-phillips-18:
**What**:
Allows for custom header and/or footer in Modal component

Fixes #1545

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.md
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.md
@@ -315,6 +315,7 @@ class NoHeader extends React.Component {
           isOpen={isModalOpen}
           hideTitle={true}
           title="no header example"
+          showClose={false}
           ariaDescribedById="no-header-example"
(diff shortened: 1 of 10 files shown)


Hyperkid123:
@jeff-phillips-18 can you fix the conflicts please? We could really use this one. Or if you want I cant take this over.

jeff-phillips-18:
Rebased, removed WIP

Hyperkid123:
@jeff-phillips-18 is it possible to include the close button to the modal Header? That would let us to customize the header completely.

Right now the button is the only thing that we cannot do anything about.

jeff-phillips-18:
@Hyperkid123 Added the `showClose` prop, defaulted to true. Set to false to hide the close button.

PR #2122, opened 2019-05-30
feat(Card): Add support for images and actions

### LINKED ISSUE #2038: Card - support images and actions
https://github.com/patternfly/patternfly-next/pull/1745

* adds `.pf-c-card__head` and `.pf-c-card__actions`

Allows you to put either `pf-c-card__header`, `pf-c-card__body`, `pf-c-card__footer`, or a custom element(s) in `pf-c-card__head` (after `pf-c-card__actions`) and that element(s) will display to the left in a flex-row with the actions.

Note that `pf-c-card__actions` is the first element in `pf-c-card__head`, but is displayed after it (with CSS) in the UI.

jessiehuff:
Fixes #2038 

Adds CardAction and CardHead


jessiehuff:
@mcarrano Thanks for taking a look!! I just updated it so let me know if anything still looks off from it. In regards to the image, the consumer is passing their own image so I think I can only set a default size to the example.

mcarrano:
This looks much better @jessiehuff . Just one other question. I noticed that if I open the kabob, the menu can get cropped by the right edge of the window. I don't see that here, for example: https://www.patternfly.org/v4/documentation/react/components/datalist/  Assume that there is a way to position the dropdown so this cropping doesn't happen?

jessiehuff:
Thanks @mcarrano!! I appreciate all of the great feedback. :)

packages/patternfly-4/react-core/src/components/Card/Card.md
```diff
@@ -21,6 +22,260 @@ SimpleCard = () => (
...
+          <CardActions>
+            <Dropdown
+              onSelect={this.onSelect}
+              toggle={<KebabToggle onToggle={this.onToggle} />}
+              isOpen={isOpen}
+              isPlain
+              dropdownItems={dropdownItems}
+              position={'right'}
+            />
+            <Checkbox
```
mcoker: Can we just use an `<input type="check">` here instead of the check component? The check component is designed to work with a check + label pair. For reference, we have plain checkboxes in the "selectable table" example https://2122-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/table/

jessiehuff: I changed it to an input, let me know if that looks ok. :)

PR #2123, opened 2019-05-30
fix(issues): fix stalebot config to include issues

redallen:
**What**: [This bot](https://github.com/probot/stale) should help us close nearly half of our current issues.

**Additional issues**:


dgutride:
@jeff-phillips-18  - that's exactly what stalebot does - it adds a note to the issue warning people that inactive issues will be closed within some period of time (I believe it's 7 days) and then if no activity, it closes them.

jeff-phillips-18:
too bad we don't have community meetings anymore, this would be a good topic to discuss there. I'm ok with it if most others are.

PR #2128, opened 2019-05-31
chore(dependencies): Add "shx" package to root dev dependencies

TheRealJon:
Add "shx" package to root dev dependencies since it is used in the root npm scripts


TheRealJon:
@dlabaj Just want to put this on your radar. This fixes a bug where `shx` package is not always installed and can cause some of our npm scripts to fail.

Issue #1892, opened 2019-04-30
TitleSize: Type not available


dlabrecq:
The TitleSize does not make '2xl' available. I'd like to use the following with Title component:

`<Title size={TitleSize.2xl}>`

Although, I can do this:

`<Title size={TitleSize['2xl']}>`

rachael-phillips:
Hi Dan, thanks for filing this! Is this blocking implementation? Also what priority is this issue to you?

dlabrecq:
Not blocking me. I'm using the workaround described above, but prefer not to do that.

jenny-s51:
@dlabrecq Titani mentioned that there is some discussion about what the best name for the vars would be.

Here are some of the suggestions she listed:

1. Size[‘2xl’] (what we already have)
2. Size.xl2
3. Size.xxl (and also xxxl, xxxxl ?)
4. Size._2xl (folks do not like this due to common use of_ for private variables)
5. Size.$2xl
6. Size.TwoXl

@dlabrecq Please comment here what you think would be most appropriate.

Thank you!

dlabrecq:
I'm just looking to add TitleSize.2xl, not asking to rename all the sizes. 

I don't know why we have TitleSize['2xl'] as a string, but probably not the best choice. That is,  considering we already support TitleSize.sm, TitleSize.xl, etc.

Why don't folks like Size.2xl -- exactly how does that interfere with private vars? I'd rather add TitleSize.2xl than rename all the sizes and introduce a breaking change.

jenny-s51:
@dlabrecq Sorry for not clarifying -- JavaScript prohibits variable names from starting with numbers, which is why 2xl must be accessed as a string: TitleSize['2xl']. 

@tlabaj marked this as a breaking change issue because there isn't a quick fix that would allow for TitleSize.2xl without renaming the sizes.

dlabrecq:
Makes sense now. Thanks for clarifying

jenny-s51:
closed via #2129

PR #2129, opened 2019-05-31
fix(TitleSize): renamed TitleSize variables

jenny-s51:
**What**: renamed BaseSizes (#1892)

- awaiting update from @tlabaj regarding what else to change for this issue

**Additional issues**: see above for issue #

### DIFF
--- a/packages/patternfly-4/react-core/src/styles/sizes.js
+++ b/packages/patternfly-4/react-core/src/styles/sizes.js
@@ -4,9 +4,9 @@ export const BaseSizes = {
   md: 'md',
   lg: 'lg',
   xl: 'xl',
-  '2xl': '2xl',
-  '3xl': '3xl',
-  '4xl': '4xl'
+  xl2: 'xl2',
+  xl3: 'xl3',
+  xl4: 'xl4'
 };
 
 export const DeviceSizes = {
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/styles/sizes.js
```diff
@@ -4,9 +4,9 @@ export const BaseSizes = {
   md: 'md',
   lg: 'lg',
   xl: 'xl',
-  '2xl': '2xl',
-  '3xl': '3xl',
-  '4xl': '4xl'
+  xl2: 'xl2',
+  xl3: 'xl3',
+  xl4: 'xl4'
```
dlabrecq: I was just looking for TitleSize.2xl to be added. Why are we changing all the size strings? 

Note that this would be a breaking change and does not appear to solve my original issue with TitleSize.

jenny-s51: @dlabrecq Sorry for not clarifying -- JavaScript prohibits variable names from starting with numbers, which is why 2xl must be accessed as a string: TitleSize['2xl'].

@tlabaj marked this as a breaking change issue because there isn't a quick fix that would allow for TitleSize.2xl without renaming the sizes

jschuler: @dlabrecq do you prefer this over using bracket notation `Size['2xl']` to access the variable?
other breaking options we could consider:
```
1. Size.xl2 (this PR)
2. Size.xxl
3. Size._2xl
4. Size.$2xl
```

dlabrecq: To be honest, I thought Size['2xl'] was a bug. If that's how it's intended to be used, then we can just add a comment and close the issue? 

Anything would be more readable, tho. Although, Size._2xl looks like the naming convention for a private variable. My vote would be for #1, what's in this PR now.

Whatever is chosen, TitleSize just needs to be in sync.

Issue #1562, opened 2019-03-12
PF4 Pagination per page menu sticks open


dlabrecq:
The menu used to set the number of items per page does not close properly. The menu opens and closes when using the button. However, when clicking elsewhere in the page, the menu remains open unexpectedly. Users can only close the menu by clicking the button, again.

![chrome-capture (1)](https://user-images.githubusercontent.com/17481322/54235595-f03a5680-44e7-11e9-91df-1b0505924817.gif)

rachael-phillips:
Hi @dlabrecq thanks for posting this! We will take this into the current milestone if we have time.

jgiardino:
Capturing some notes from a slack discussion here about updates needed to address this:

- OptionsMenu is updated so that menu panel is closed when clicking outside the menu
- OptionsMenu is updated to include a prop so that the consumer can decide if it stays open or closes when selecting an option. 
  - The prop should default to close the menu when an option is selected. 
  - All examples can use the default except the example labeled **Options menu - multiple options**. For this example, we should set the prop to keep the menu open when selecting an option.
- Pagination component should not need updates, but please check that it works as expected after updating the OptionsMenu. 

Also, here are some comments from @mcarrano on the use case for when a consumer might want the menu to stay open:

> Yes, it was coded this way to all for selecting multiple options without needing to reopen the menu each time. Sort is a good example where I might want to say sort by Name and sort in descending order. For single option selection, there is probably no need to keep the menu open. That would solve the issue raised with the Items per page selection. I also agree that clicking outside of the menu should close it.

PR #2140, opened 2019-06-03
fix(pagination): this PR is incomplete/needs review

jenny-s51:
awaiting feedback from @redallen regarding current changes. 

- pagination optionsToggle now has same toggle behavior and appearance as optionsMenu
- still working to resolve onClick behavior that will hide menu

Screen Shot 2019-06-03 at 2 47 35 PM


jenny-s51:
@tlabaj yes, it's #1562

karelhala:
Thank you for the PR! This issue was introduced by me sadly I have pointed it out in the original implementation and was unable to fix it at that time. However as some component have been refactored recently this is rather easy to fix.

What you should eventually do is use drodpwn Toggle component instead of button component. Pass down required props and use that instead of button component. You are on the right way with rewriting this to class so we can use some other React goodies, but you don't really have to write all those functions to integrate with the toggle component. Here is proposed diff
```diff
diff --git a/packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.js b/packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.js
index b58f345a..5fd7e487 100644
--- a/packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.js
+++ b/packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.js
@@ -1,46 +1,51 @@
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { CaretDownIcon } from '@patternfly/react-icons';
 import styles from '@patternfly/react-styles/css/components/OptionsMenu/options-menu';
 import { css, getModifier } from '@patternfly/react-styles';
 import { fillTemplate } from '../../helpers';
+import { DropdownToggle }   from '../Dropdown';
 
-const OptionsToggle = ({
-  itemsTitle,
-  optionsToggle,
-  firstIndex,
-  lastIndex,
-  itemCount,
-  widgetId,
-  onToggle,
-  isOpen,
-  showToggle,
-  toggleTemplate: ToggleTemplate
-}) => (
-  <div className={css(styles.optionsMenuToggle, getModifier(styles, 'plain'), getModifier(styles, 'text'))}>
-    <span className={css(styles.optionsMenuToggleText)}>
-      {typeof ToggleTemplate === 'string' ? (
-        fillTemplate(ToggleTemplate, { firstIndex, lastIndex, itemCount, itemsTitle })
-      ) : (
-        <ToggleTemplate firstIndex={firstIndex} lastIndex={lastIndex} itemCount={itemCount} itemsTitle={itemsTitle} />
+class OptionsToggle extends Component {
+  render() {
+    const {
+      itemsTitle,
+      optionsToggle,
+      firstIndex,
+      lastIndex,
+      itemCount,
+      widgetId,
+      onToggle,
+      isOpen,
+      showToggle,
+      toggleTemplate: ToggleTemplate
+    } = this.props;
+
+    return <div className={css(styles.optionsMenuToggle, getModifier(styles, 'plain'), getModifier(styles, 'text'))}
+      ref={ref => {
+        this.parentRef = ref;
+      }}
+    >
+      <span className={css(styles.optionsMenuToggleText)}>
+        {typeof ToggleTemplate === 'string' ? (
+          fillTemplate(ToggleTemplate, { firstIndex, lastIndex, itemCount, itemsTitle })
+        ) : (
+            <ToggleTemplate firstIndex={firstIndex} lastIndex={lastIndex} itemCount={itemCount} itemsTitle={itemsTitle} />
+          )}
+      </span>
+      {showToggle && (
+        <DropdownToggle
+          aria-label={optionsToggle}
+          onToggle={onToggle}
+          isOpen={isOpen}
+          id={`${widgetId}-toggle`}
+          isSplitButton
+          className={styles.optionsMenuToggleButton}
+          parentRef={this.parentRef}
+        />
       )}
-    </span>
-    {showToggle && (
-      <button
-        className={css(styles.optionsMenuToggleButton)}
-        id={`${widgetId}-toggle`}
-        aria-haspopup="listbox"
-        aria-labelledby={`${widgetId}-toggle ${widgetId}-label`}
-        aria-label={optionsToggle}
-        aria-expanded={isOpen}
-        onClick={() => onToggle(!isOpen)}
-        type="button"
-      >
-        <CaretDownIcon />
-      </button>
-    )}
-  </div>
-);
+    </div>
+  }
+}
```

jenny-s51:
@karelhala Hello, thank you so much for your helpful feedback and for your proposed diff -- this works!

Cheers 🙂

PR #2142, opened 2019-06-04
feat(build): add source maps

redallen:
**What**: Add source maps for PatternFly 4 packages, BUT exclude them from the published `dist` folders.

**Additional issues**:


dlabaj:
@redallen Does this add it to shipping version?  Do we want to do that ... or do we want a separate debug version.

Issue #2151, opened 2019-06-04
RC2: Tabs: Invalid props 'forwardRef'


dlabrecq:
Seeing the following error in the browser console with Cost Management and the latest packages.

![Screen Shot 2019-06-04 at 1 21 42 PM](https://user-images.githubusercontent.com/17481322/58901711-b6848080-86cf-11e9-8529-42f7cbcee5be.png)

PatternFly rc.2 versions:
    @patternfly/patternfly: 2.8.2
    @patternfly/react-core: 3.34.0
    @patternfly/react-styles: 3.2.3
    @patternfly/react-tokens: 2.5.3
    @patternfly/react-table:2.9.0
    @patternfly/react-charts: 4.0.0
    @patternfly/react-inline-edit-extension: 2.5.0

Issue #2132, opened 2019-06-03
PF4 Tabs Invalid prop `forwardRef` supplied to `Tab`


idevat:
There's an issue with version `3.32.5` (recommended in issue #2100).

Component:
```javascript
import React from "react";
import { Tabs, Tab } from "@patternfly/react-core";

const Test = () => (
  <Tabs activeKey={0}>
    <Tab eventKey={0} title="Tab0">
      Tab0
    </Tab>
    <Tab eventKey={1} title="Tab1">
      Tab1
    </Tab>
  </Tabs>
);

export default Test;
```
produces warning:
![tabs_issue](https://user-images.githubusercontent.com/16188254/58783712-dccbf280-85e1-11e9-8c62-64e6b4b18971.png)

This issue does not occur in a version `3.16.10`

redallen:
This has to do with our incorrect PropTypes. I just opened a PR to fix it!

jschuler:
Thanks @redallen I see your fix covered it as well, I am merged a fix just for tabs that also addresses this

PR #2158, opened 2019-06-04
fix tabs ref proptype

jschuler:
Closes: #2151 #2132


jschuler:
Closing i think there is still an issue left. will reopen when ready

jschuler:
It's good to merge as is

PR #2170, opened 2019-06-05
feat(Topology): Add @patternfly/react-topology package

jeff-phillips-18:
**What**:
Add initial @patternfly/react-topology package containing the TopologyView and TopologyControlBar components.

FYI: @christianvogt


packages/patternfly-4/react-topology/build/snapshot-serializer.js
```diff
@@ -0,0 +1,8 @@
+const fs = require('fs');
```
redallen: Don't need this file since jest uses the one from react-core.

packages/patternfly-4/react-topology/scripts/copyTS.js
```diff
@@ -0,0 +1,15 @@
+const path = require('path');
```
redallen: With no .d.ts files in this project, we don't need this script.

jeff-phillips-18:
Updated per comments, rebased.

packages/patternfly-4/react-docs/gatsby-node.js
```diff
@@ -96,6 +96,7 @@ exports.onCreateWebpackConfig = ({ actions }) => {
         '@patternfly/react-styles': path.resolve(__dirname, '../react-styles'),
         '@patternfly/react-table': path.resolve(__dirname, '../react-table'),
         '@patternfly/react-tokens': path.resolve(__dirname, '../react-tokens'),
+        '@patternfly/react-topology': path.resolve(__dirname, '../react-topology/dist/esm'),
```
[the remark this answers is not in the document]
jeff-phillips-18: I was having a strange build issue w/o the path. That seems to be resolved now so I believe I can remove it.

PR #2190, opened 2019-06-07
build(esm): fix esm output and convert react-virtualized-extension to TS

redallen:
**What**: Fix recent regression from #2085 that made our dist only commons JS for a few packages. Clean babelrc.js and babel build targets to be consistent across packages. Fix jest config afterwards.

**Additional issues**:


packages/patternfly-4/react-core/.babelrc.js
```diff
@@ -1,15 +1,6 @@
-const babelENV = process.env.BABEL_ENV || 'development';
```
dgutride: Can we move the contents of this .babelrc.js file to packages/patternfly-4/ and then refer to it the same way everywhere instead of using the one in react-core in sibling folders?

redallen: Done.

priley86:
thanks @redallen ! will try to test this out this week.

Issue #2096, opened 2019-05-28
Convert Select to Typescript


kmcfaul:

tlabaj:
closed by #2201

PR #2201, opened 2019-06-10
feat(Select): convert to Typescript

kmcfaul:
**What**: Typescript conversion for select. This also contains a couple **breaking changes** for the Checkbox select variant, whose classes have been collapsed into the normal Select classes (now controlled by the variant parameter only), and CheckboxSelectGroup, which has been renamed SelectGroup and is now generic/available to use for other variants. 

**Refer to issue**: #2096, #2223

2223 should be resolved with the change to TS.


kmcfaul:
@redallen Any advice on the failing circleci build? I'm not finding where it's complaining about SelectGroup in the docs.

redallen:
@kmcfaul Double-check that SelectGroup is correctly exported in the `dist`. It should be possible to reproduce locally if you start with `git clean -dfx`.

packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -319,46 +343,48 @@ class Select extends React.Component {
...
         </SelectContext.Provider>
       </div>
     );
   }
 }
 
-Select.propTypes = propTypes;
-Select.defaultProps = defaultProps;
-
 export default Select;
```
redallen: Remove all default exports in favor of named exports.

packages/patternfly-4/react-core/src/components/Select/SelectGroup.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <div {...props} className={css(styles.selectMenuGroup, className)}>
+      <div className={css(styles.selectMenuGroupTitle)} id={titleId} aria-hidden>
+        {label}
+      </div>
+      {children}
+    </div>
+  </React.Fragment>
+);
+
+export default SelectGroup;
```
redallen: Remove all default exports in favor of named exports.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -0,0 +1,154 @@
...
+              </FocusTrap>
+            )}
+          </React.Fragment>
+        )}
+      </SelectConsumer>
+    );
+  }
+}
+
+export default SelectMenu;
```
redallen: Remove all default exports in favor of named exports.

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -0,0 +1,156 @@
...
+              </label>
+            )}
+          </React.Fragment>
+        )}
+      </SelectConsumer>
+    );
+  }
+}
+
+export default SelectOption;
```
redallen: Remove all default exports in favor of named exports.

packages/patternfly-4/react-core/src/components/Select/SelectToggle.tsx
```diff
@@ -0,0 +1,249 @@
...
+              <CaretDownIcon className={css(styles.selectToggleArrow)} />
+            </button>
+          </div>
+        )}
+      </React.Fragment>
+    );
+  }
+}
+
+export default SelectToggle;
```
redallen: Remove all default exports in favor of named exports.

packages/patternfly-4/react-core/src/components/Select/SelectGroup.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const SelectGroup: React.SFC<SelectGroupProps> = ({
+  children = [] as React.ReactNode[],
+  className = '',
+  label = '',
+  titleId = '',
+  ...props
+}) => (
+  <React.Fragment>
```
redallen: No need for `<React.Fragment>` here.

SpyTec:
The failing build in CircleCI seems unrelated as it points to PatternFly 3. But this looks great!

This TypeScript migration doesn't address any existing bugs right? It only migrates existing code and drops some components in favour of one. Is there an ETA for when this is merged?

kmcfaul:
Yeah I'll work on a non-breaking change version.

kmcfaul:
@tlabaj have a patch up that should make this a non-breaking change. The old Checkbox specific classes are back, and users should be able to use them, or use the new way with only Select specific classes, to render the checkbox select. The examples remain updated to show the new method, to further phase out the old method.

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -0,0 +1,154 @@
...
+  className?: string;
+  /** Internal index of the option */
+  index?: number;
+  /** The value for the option */
+  value: string;
+  /** Flag indicating if the option is disabled */
+  isDisabled?: boolean;
+  /** Flag indicating if the option acts as a placeholder */
+  isPlaceholder?: boolean;
+  /** Internal flag indicating if the option is selected */
```
jeff-phillips-18: What does `Internal` flag mean?

packages/patternfly-4/react-core/src/components/Select/SelectToggle.tsx
```diff
@@ -0,0 +1,248 @@
...
+    ariaLabelToggle: '',
+    type: 'button',
+    onToggle: Function.prototype,
+    onEnter: Function.prototype,
+    onClose: Function.prototype
+  };
+
+  constructor(props: SelectToggleProps) {
+    super(props);
+    const { variant } = this.props;
```
jeff-phillips-18:
```suggestion
    const { variant } = props;
```

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
kmcfaul: It means the property is used by the other select classes, and isn't one that has to be passed in by the user. Our docs generate all props so I wanted to specify which aren't expected.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -0,0 +1,152 @@
...
+    keyHandler: Function.prototype
+  };
+
+  extendChildren() {
+    const { children, isGrouped } = this.props;
+    if (isGrouped) {
+      let index = 0;
+      return React.Children.map(children, (group: React.ReactElement) =>
+        React.cloneElement(group, {
+          titleId: group.props.label.replace(/\W/g, '-'),
```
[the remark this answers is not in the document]
kmcfaul: label in the Group has a default value so it shouldn't error out

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectGroup.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Select/select';
+import { css } from '@patternfly/react-styles';
+
+export interface CheckboxSelectGroupProps {
```
nicolethoen: should add `extends React.HTMLProps<HTMLDivElement>` I believe.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
```diff
@@ -0,0 +1,100 @@
...
+  isChecked?: boolean;
+  /** Internal callback for ref tracking */
+  sendRef?: (ref: React.ReactNode, index: number) => void;
+  /** Internal callback for keyboard navigation */
+  keyHandler?: (index: number, position: string) => void;
+  /** Optional callback for click event */
+  onClick?: (event: React.MouseEvent | React.ChangeEvent) => void;
+}
+
+export class CheckboxSelectOption extends React.Component<CheckboxSelectOptionProps> {
```
nicolethoen: Does this need to be a class if there is no state, @redallen? Not sure if there is an alternative to the componentDidMount and componentDidUpdate functions in this case. Should they be modifying a state?

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Select/select';
+import { default as checkStyles } from '@patternfly/react-styles/css/components/Check/check';
+import { css } from '@patternfly/react-styles';
+import { SelectConsumer, KeyTypes } from './selectConstants';
+
+export interface CheckboxSelectOptionProps {
```
nicolethoen: missing `extends React.HTMLProps<HTMLElement>`.

packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -1,89 +1,103 @@
...
 import { SelectContext, SelectVariant } from './selectConstants';
-import { getNextIndex } from '../../helpers/util';
+import { Chip, ChipGroup } from '../ChipGroup';
+import { keyHandler, getNextIndex } from '../../helpers/util';
 
 // seed for the aria-labelledby ID
 let currentId = 0;
 
-const propTypes = {
+export interface SelectProps {
```
nicolethoen: missing `extends React.HTMLProps<HTMLDivElement>`.

packages/patternfly-4/react-core/src/components/Select/SelectGroup.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Select/select';
+import { css } from '@patternfly/react-styles';
+
+export interface SelectGroupProps {
```
nicolethoen: missing `extends React.HTMLProps<HTMLDivElement>`.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -0,0 +1,152 @@
...
+  checked?: string[];
+  /** Internal flag for specifiying how the menu was opened */
+  openedOnEnter?: boolean;
+  /** Internal callback for ref tracking */
+  sendRef?: (ref: React.ReactNode, index: number) => void;
+  /** Internal callback for keyboard navigation */
+  keyHandler?: (index: number, position: string) => void;
+}
+
+export class SelectMenu extends React.Component<SelectMenuProps> {
```
nicolethoen: I think this can also be a React.FunctionComponent since there is no state.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -0,0 +1,152 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Select/select';
+import { default as formStyles } from '@patternfly/react-styles/css/components/Form/form';
+import { css } from '@patternfly/react-styles';
+import { SelectConsumer, SelectVariant } from './selectConstants';
+
+// tslint:disable-next-line
+const FocusTrap: any = require('focus-trap-react');
+
+export interface SelectMenuProps {
```
nicolethoen: missing `extends React.HTMLProps<HTMLElement>`.

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -0,0 +1,154 @@
...
+  isFocused?: boolean;
+  /** Internal callback for ref tracking */
+  sendRef?: (ref: React.ReactNode, index: number) => void;
+  /** Internal callback for keyboard navigation */
+  keyHandler?: (index: number, position: string) => void;
+  /** Optional callback for click event */
+  onClick?: (event: React.MouseEvent | React.ChangeEvent) => void;
+}
+
+export class SelectOption extends React.Component<SelectOptionProps> {
```
nicolethoen: @redallen same question here about stateless components as well

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -0,0 +1,154 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Select/select';
+import { default as checkStyles } from '@patternfly/react-styles/css/components/Check/check';
+import { css } from '@patternfly/react-styles';
+import { CheckIcon } from '@patternfly/react-icons';
+import { SelectConsumer, SelectVariant, KeyTypes } from './selectConstants';
+
+export interface SelectOptionProps {
```
nicolethoen: missing `extends React.HTMLProps<HTMLElement>`.

packages/patternfly-4/react-core/src/components/Select/SelectToggle.tsx
```diff
@@ -0,0 +1,248 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Select/select';
+import buttonStyles from '@patternfly/react-styles/css/components/Button/button';
+import { css } from '@patternfly/react-styles';
+import { CaretDownIcon } from '@patternfly/react-icons';
+import { KeyTypes, SelectVariant } from './selectConstants';
+
+export interface SelectToggleProps {
```
nicolethoen: missing `extends React.HTMLProps<HTMLElement>`.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
kmcfaul: This errors out, but extending `React.HTMLProps<HTMLLabelElement>` works.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
kmcfaul: This errors out due to the `checked` param needing a string[] and not a boolean, and the rendered content can be either a `div` or a `ul`. Is there another extension that would fit better, or do I omit `checked`?

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
nicolethoen: okay that makes sense. I wasn't sure if the consumer would effect which type of element it expects.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
kmcfaul: The several components mentioned all require lifecycle methods to work properly, so I'm unsure if they can be converted to functional components opposed to classes.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
nicolethoen: You can try to omit checked. It should definitely extend some sort of Element, but it may take some research and omitting params to figure out which one is appropriate in this scenario.

kmcfaul: I was able to use the extension with Omit typeUtil helper.

packages/patternfly-4/react-core/src/components/Select/Select.tsx
kmcfaul: Works with Omit for several properties (onSelect, ref, checked, selected). I think it's mostly because I used them differently from what the extension expects, or in the case of ref if can be different types based on which variant is chosen.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
redallen: @kmcfaul is correct.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectGroup.tsx
```diff
@@ -0,0 +1,29 @@
...
+  /** Additional classes added to the CheckboxSelectGroup control */
+  className?: string;
+  /** Group label */
+  label?: string;
+  /** ID for title label */
+  titleId?: string;
+}
+
+export const CheckboxSelectGroup: React.FunctionComponent<CheckboxSelectGroupProps> = ({
+  children = [] as React.ReactNode[],
```
redallen: Then here you can cast to `React.ReactElement[]`, or just say `= null`.

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
nicolethoen: and there are no lifecycle methods here, these methods can be defined in the body of the functionComponent

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
redallen: I opened #2380 for this.

Issue #2145, opened 2019-06-04
Link Button Broken


christiemolloy:
The Link Button variation of the Button is missing the margin between icon and text. Refer to Core for correct spacing. This PR was merged in Core: https://github.com/patternfly/patternfly-next/pull/1890

jschuler:
At the Button is just a container, if it's the Link variant you can place an icon + text or just text or just icon. To add the `pf-c-button__icon` class on the icon for the user which adds the right padding on the icon, we might have to add an additional icon prop for the button that adds it in for the user. otherwise the user could manually insert `<PlusCircleIcon className="pf-c-button__icon" />`

christiemolloy:
@jschuler I was currently fixing this in Core, so that we have a wrapper around the icon with the class that sets the margin. Do you think thats a better way than adding another prop/the user manually setting it? or do you think this can be easily fixed in React?

mcoker:
@jschuler the wrapping element was added to core in https://github.com/patternfly/patternfly-next/pull/1890

kmcfaul:
As @jschuler mentioned, the current React implementation has no knowledge of what the Button contains at the moment, so the icon as well as the text are defined outside. We can either wrap the entirety of the link button's contained content (icon + text), or define an optional icon property that we wrap ourselves.

Is there a preferred way for me to implement this of the options?

christiemolloy:
The idea is that for any button with an icon + text, the icon should have a class wrapped around it and we defined that class in Core. I think the second suggestion about defining an option icon property would work right @mcoker

mcoker:
> define an optional icon property that we wrap ourselves.

Yeah, would you take a prop for the icon to use and the react component generates the icon and wraps it in the new element? FWIW, I was thinking something like this, but maybe a prop would be better?

```
    <Button variant="link">
      <ButtonIcon>
        <PlusCircleIcon />
      </ButtonIcon>
      Link button
    </Button>
```

kmcfaul:
@mcoker 
It would be more along the lines of:
`
<Button variant="link" icon={<PlusCircleIcon />}>
     Link Button
</Button>
`

`<ButtonIcon>` would be another component, but is also doable if we want to add a simple wrapper instead. Users would have to define their link buttons as you have shown in this case.

mcoker:
Your suggestion is much better!

mcoker:
[...]

PR #2205, opened 2019-06-10
feat(Button): add icon property and wrapper css

kmcfaul:
**What**: Adds an optional icon property. When present, wraps the icon in the new margin css class and displays the icon before the text.

**Refer to issue**: #2145

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Button/Button.tsx
+++ b/packages/patternfly-4/react-core/src/components/Button/Button.tsx
@@ -41,6 +41,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   variant?: 'primary' | 'secondary' | 'tertiary' | 'danger' | 'link' | 'plain' ;
   /** Adds accessible text to the button. */
   'aria-label'?: string; 
+  /** Icon for the button */
+  icon?: React.ReactType | null;
 }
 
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Button/Button.tsx
```diff
@@ -41,6 +41,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   variant?: 'primary' | 'secondary' | 'tertiary' | 'danger' | 'link' | 'plain' ;
   /** Adds accessible text to the button. */
   'aria-label'?: string; 
+  /** Icon for the button */
+  icon?: React.ReactType | null;
```
redallen: Change to `React.ReactNode`

packages/patternfly-4/react-core/src/components/Button/Button.tsx
```diff
@@ -41,6 +41,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   variant?: 'primary' | 'secondary' | 'tertiary' | 'danger' | 'link' | 'plain' ;
   /** Adds accessible text to the button. */
   'aria-label'?: string; 
+  /** Icon for the button */
```
redallen: Change to `Icon for the button if variant is link`

Issue #2188, opened 2019-06-07
Page component isManagedSidebar does not work if wrapper components are used for the header and sidebar props


jschuler:
Should use context api to get the isManagedSidebar props value

PR #2208, opened 2019-06-11
PF4 - use context api in page component

jschuler:
Closes: #2188


packages/patternfly-4/react-core/src/components/Page/PageHeader.js
```diff
@@ -56,42 +57,51 @@ const PageHeader = ({
...
+                </div>
+              )}
+              {logo && (
+                <a className={css(styles.pageHeaderBrandLink)} {...logoProps}>
+                  {logo}
+                </a>
+              )}
+            </div>
+          )}
+          {/* Hide for now until we have the context selector component */}
```
tlabaj: We have . context Selector.  Why are we hiding this?

jschuler: This has been here since before the context selector got created. Perhaps we should update the page component now to support it? Should be in a separate PR though. @rachael-phillips

tlabaj: agreed.  We should do tis in a separate PR. @rachael-phillips we can open a separate issue to track that work.

tlabaj:
Closing this.  Newer PR #2273 has been opened to resolve the issue.

Issue #2154, opened 2019-06-04
Alert - provide default value for variant prop


jgiardino:
As suggested in #2030, there should be a default value defined for the `variant` prop. After discussing with other PF contributors, the recommendation is to use `'info'` as the default.

ssilvert:
NOOOOOO!!!!!

**variantProp** should not be displayed as part of the title at all.

As a developer, I should be able to just say:
```javascript
<Alert>My message to the user.</Alert>
```

It should all just work.  The message should be displayed to the user with a default severity level.

tlabaj:
Closed by #2218

PR #2218, opened 2019-06-11
fix(Alert): set the variant prop to info by default

boaz0:
**What**:

Setting `Alert` to info if not specified differently.

//cc @jgiardino @ssilvert
What: #2154


jgiardino:
No comments from me. 😄 I tested the preview by removing the `variant` prop and it behaves as expected.

boaz0:
Thank you @jgiardino @tlabaj and @redallen

PR #2227, opened 2019-06-12
feat(PF4 Dropdown): Wrap dropdown with context

karelhala:
**What**:
Wrap dropdown with context provider so we can change classes for each part of it. Consumers can now change basic classes if they want to. With this approach we will be able to reuse dropdown component as much as possible so we don't have to deal with different classes.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/index.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/index.js
@@ -1,4 +1,4 @@
-export { default as Dropdown, DropdownWithContext as DefaultDropdown } from './Dropdown';
+export { default as Dropdown } from './Dropdown';
 export { DropdownDirection, DropdownPosition, DropdownContext } from './dropdownConstants';
 export { default as DropdownGroup } from './DropdownGroup';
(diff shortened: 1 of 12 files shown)


karelhala:
@jschuler do you know why circle CI fails from time to time?

dgutride:
@karelhala - are you seeing an odd babelrc error that shows up in patternfly-3 packages even when they aren't changed?

jschuler:
@karelhala not sure why it sometimes fails, I have not looked too much into that yet, @redallen would probably know! I restarted the circle ci workflow and has passed now

packages/patternfly-4/react-core/src/components/Dropdown/index.js
```diff
@@ -1,5 +1,5 @@
-export { default as Dropdown } from './Dropdown';
-export { DropdownDirection, DropdownPosition } from './dropdownConstants';
+export { default as Dropdown, DropdownWithContext as DefaultDropdown } from './Dropdown';
```
dlabaj: Can you remove these exports from index.js.  We don't want these exposed to consumers.

Issue #2214, opened 2019-06-11
Default font stack for DonutChartUtilization center label is incorrect


TheRealJon:
The font stack used for the center label on the donut utilization chart is not the PF4 font stack.

dlabrecq:
PatternFly-core is actually making variables for these fonts. They will be updated shortly.

dlabrecq:
dup of https://github.com/patternfly/patternfly-react/issues/2160

Issue #2160, opened 2019-06-04
Gill Sans font in donut charts


stacymcauliffe:
Numbers and secondary text in the center or donut charts is displaying in Gill Sans:

![image](https://user-images.githubusercontent.com/35926653/58911219-62849680-86e5-11e9-8a3d-8c31b32800ee.png)

dlabrecq:
Fixed via  #2232

PR #2232, opened 2019-06-12
fix(charts): labels should use the overpass font

dlabrecq:
The VictoryLabel is not part of the theme, so we must apply a default font via ChartLabel.

Fixes https://github.com/patternfly/patternfly-react/issues/2160
Fixes https://github.com/patternfly/patternfly-react/issues/2214

Screen Shot 2019-06-12 at 11 22 51 AM

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/ChartLabel/ChartLabel.tsx
+++ b/packages/patternfly-4/react-charts/src/components/ChartLabel/ChartLabel.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react';
 import hoistNonReactStatics from 'hoist-non-react-statics';
+import { defaults } from 'lodash';
 import {
   StringOrNumberOrCallback,
-  // TextAnchorType,
-  // VerticalAnchorType,
   VictoryLabel,
   VictoryLabelProps
 } from 'victory';
+import { LabelStyles } from '../ChartTheme/themes/label-theme';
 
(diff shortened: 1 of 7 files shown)


packages/patternfly-4/react-charts/src/components/ChartLabel/ChartLabel.tsx
```diff
@@ -166,8 +166,20 @@ export interface ChartLabelProps extends VictoryLabelProps {
...
+  const newStyles: React.CSSProperties[] = [];
+  if (Array.isArray(style)) {
+    for(let i = 0; i < style.length; i++) {
+      newStyles.push(defaults({}, style[i], LabelStyles));
+    }
+  } else {
+    newStyles.push(style ? defaults({}, style, LabelStyles) : LabelStyles);
+  }
+  return <VictoryLabel style={newStyles as any} {...rest} />;
+}
```
TheRealJon: This could be simplified:
```js
  const applyDefaultStyle = (customStyle) => defaults(customStyle, LabelStyles);
  const newStyle = Array.isArray(style) ? style.map(applyDefaultStyle) : applyDefaultStyle(style);
  return <VictoryLabel style={newStyle as any} {...rest} />; 
```

dlabrecq: Updated, thanks

PR #2247, opened 2019-06-12
chore(topology): add tslint.json

redallen:
**What**: React-topology currently isn't linting.

**Additional issues**:


jeff-phillips-18:
This should be a `chore`

Issue #1950, opened 2019-05-08
Convert React-Table to TypeScript


priley86:
* convert all react-table source to tsx
* bring the reactabular-table source in house and convert it to tsx. This follows along with #1628 which I have already prototyped alongside #1250, but I will pull that change out specifically alongside tsx conversion for now.

priley86:
* noted will need to be able to share tsconfigs/base tslint configs with other packages outside of react-core to accomplish this.

ssilvert:
Hi @priley86, do you have an ETA for this?

We are trying to use react-table with TypeScript and we get errors about declaration files.  Also looks like you need to use the latest version of react-core.

node_modules/@patternfly/react-table/dist/js/components/Table/index.d.ts(20,49): error TS7016: Could not find a declaration file for module './ExpandableRowContent'. 'C:/GitHub/keycloak/themes/src/main/resources/theme/keycloak-preview/account/resources/node_modules/@patternfly/react-table/dist/js/components/Table/ExpandableRowContent.js' implicitly has an 'any' type.
node_modules/@patternfly/react-table/node_modules/@patternfly/react-core/dist/js/components/Select/Select.d.ts(10,18): error TS2430: Interface 'SelectProps' incorrectly extends interface 'HTMLProps<HTMLOptionElement>'.
  Types of property 'onSelect' are incompatible.
    Type '(event: any, element: string, isPlaceholder: boolean) => void' is not assignable to type '(event: SyntheticEvent<HTMLOptionElement, Event>) => void'.
node_modules/@patternfly/react-table/node_modules/@patternfly/react-core/dist/js/components/Select/Select.d.ts(14,19): error TS2304: Cannot find name 'ReactEventHandler'.

priley86:
hi @ssilvert 👋 

i'm guessing we'll be able to introduce this in the next few sprints - currently still trying to resolve a few other things downstream on my side. I'll probably be able to attempt this next week though. As far as I can tell, we'll need `Dropdown` and a few other react-core components converted to TS first. Noted that when starting this [here](https://github.com/priley86/patternfly-react/commit/eaecec3f5783be6843fcf19914547ac71ebd5c99). If you need this sooner and would like to try and contribute it (or you have a short term workaround), happy to review this w/ you.

cc: @LHinson @dgutride

ssilvert:
Hi @priley86,

[...]

PR #2248, opened 2019-06-12
Table type adjustments

seanforyou23:
**What**: This PR updates the type used in react-table to be a little more explicit. Also updates the example code in the readme, following along as-is in a TypeScript application leads to the following error

Screen Shot 2019-06-12 at 10 52 42 AM

I took a stab at using better types for `transforms`, what do people think about this new type? I'd like to use something more specific than `Array<Function>` if possible as this doesn't inform of what the inputs and outputs should or can be. Might be good to wait until this is converted fully to TS to address the others, but I'm curious what others think.

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.d.ts
@@ -48,12 +48,12 @@ export interface IExtra extends IExtraData {
 }
 
 export interface ISortBy {
-  index?: Number;
+  index?: number;
   direction?: OneOf<typeof SortByDirection, keyof typeof SortByDirection>;
 }
 
 export interface IAction {
-  title: String;
+  title: string;
   onClick: (event: MouseEvent, rowIndex: number, rowData: IRowData, extraData: IExtraData) => void;
 }
 
@@ -61,13 +61,18 @@ export interface ISeparator {
   isSeparator: Boolean
 }
 
+export interface IDecorator extends HTMLProps<HTMLElement> {
+  isVisible: boolean;
+  children?: React.ReactNode;
+}
+
 export interface ICell {
-  title: String;
-  transforms?: Array<Function>;
-  cellTransforms?: Array<Function>;
-  columnTransforms?: Array<Function>;
-  formatters?: Array<Function>;
-  cellFormatters?: Array<Function>;
+  title: string;
+  transforms?: ((value: any) => IDecorator)[];
+  cellTransforms?: ((value: any) => IDecorator)[];
+  columnTransforms?: ((value: any) => IDecorator)[];
+  formatters?: ((value: any) => IDecorator)[];
+  cellFormatters?: ((value: any) => IDecorator)[];
   props: any;
 }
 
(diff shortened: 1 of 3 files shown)


priley86:
No issues from me and thanks for taking time to look into this. I think any good improvements on this will be helpful for #1950 later. 👍

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -48,22 +48,27 @@ export interface IExtra extends IExtraData {
...
-  title: String;
+  title: string;
   onClick: (event: MouseEvent, rowIndex: number, rowData: IRowData, extraData: IExtraData) => void;
 }
 
 export interface ISeparator {
   isSeparator: Boolean
 }
 
+export interface ITransform {
```
karelhala: These functions can return object that contains eny HTML property so extendind `HTMLProps` is a good way. And probably renaming it to `IDecorator` so we can reuse this interface in many places?
```TSX
export interface ITransform extends HTMLProps<HTMLElement>{
  isVisible: boolean;
  // children doesn't have to be required
}
```

packages/patternfly-4/react-table/src/components/Table/Table.d.ts
```diff
@@ -48,22 +48,27 @@ export interface IExtra extends IExtraData {
...
+}
+
 export interface ICell {
-  title: String;
-  transforms?: Array<Function>;
+  title: string;
+  transforms?: ((value: any) => ITransform)[];
   cellTransforms?: Array<Function>;
   columnTransforms?: Array<Function>;
   formatters?: Array<Function>;
```
karelhala: Nice way of providing further information o others trough typed array. We should add this to other arrays as well.
```TSX
  transforms?: ((value: any) => IDecorator)[];
  cellTransforms?: ((value: any) => IDecorator)[];
  columnTransforms?: ((value: any) => IDecorator)[];
  formatters?: ((value: any) => IDecorator)[];
```

packages/patternfly-4/react-table/src/components/Table/index.d.ts
```diff
@@ -12,7 +12,8 @@ export {
   IColumn,
   IExtra,
   IExtraData,
-  IExtraColumnData
+  IExtraColumnData,
+  ITransform
```
priley86: sorry, i think i am missing something here. Tested these `.d.ts` downstream... getting 

```
Module '"../../../../../../../../../../Users/priley/GitHub/patternfly-react/node_modules/@patternfly/react-table/src/components/Table/Table"' has no exported member 'ITransform'.ts(2305)
```

priley86: changes in `Table.d.ts` work fine for me though.

karelhala: Yeah it wasn't renamed from `ITransform` to `IDecorator`. Good catch!

seanforyou23: 🤦‍♂️

PR #2251, opened 2019-06-12
feat(datalist): add hidden and visible breakpoints

### LINKED ISSUE #2197: datalist: add support for hidden/visible
datalist: adds breakpoints for actions. Add support for hidden/visible like we did with the table component (the “Table with hidden/visible breakpoint modifiers” example in the react workspace)

### LINKED ISSUE #2221: Bug: Data list checkboxes, actions and additional cells
The primary and secondary buttons on the react datalist example look buggy

Screen Shot 2019-06-11 at 4 49 57 PM

boaz0:
**What**:

closes #2221 and fixes #2197

- [x] update the datalist action component to what is given in pf-core
- [x] add the hidden and visible breakpoint modifiers as props
- [x] update the datalist checkbox and actions example in the docs
- [x] update tests

### DIFF
--- a/packages/patternfly-4/react-core/src/components/DataList/DataList.md
+++ b/packages/patternfly-4/react-core/src/components/DataList/DataList.md
@@ -16,6 +16,7 @@ import {
   DataListCell,
   DataListCheck,
   DataListAction,
+  DataListActionVisibility,
   DataListToggle,
   DataListContent,
   Dropdown,
@@ -24,6 +25,7 @@ import {
   DropdownItem
 } from '@patternfly/react-core';
 import { CodeBranchIcon } from '@patternfly/react-icons';
+import { css } from '@patternfly/react-styles';
 
 ```js
 import React from 'react';
@@ -97,15 +99,35 @@ import {
 class CheckboxActionDataList extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { isOpen: false };
+    this.state = { isOpen1: false, isOpen2: false, isOpen3: false };
 
-    this.onToggle = isOpen => {
-      this.setState({ isOpen });
+    this.onToggle1 = isOpen1 => {
+      this.setState({ isOpen1 });
     };
 
-    this.onSelect = event => {
+    this.onSelect1 = event => {
       this.setState(prevState => ({
-        isOpen: !prevState.isOpen
+        isOpen1: !prevState.isOpen1
+      }));
+    };
+
+    this.onToggle2 = isOpen2 => {
+      this.setState({ isOpen2 });
+    };
+
+    this.onSelect2 = event => {
+      this.setState(prevState => ({
+        isOpen2: !prevState.isOpen2
+      }));
+    };
+
+    this.onToggle3 = isOpen3 => {
+      this.setState({ isOpen3 });
+    };
+
+    this.onSelect3 = event => {
+      this.setState(prevState => ({
+        isOpen3: !prevState.isOpen3
       }));
     };
   }
@@ -144,9 +166,9 @@ class CheckboxActionDataList extends React.Component {
               <Dropdown
                 isPlain
                 position={DropdownPosition.right}
-                isOpen={this.state.isOpen}
-                onSelect={this.onSelect}
-                toggle={<KebabToggle onToggle={this.onToggle} />}
+                isOpen={this.state.isOpen1}
+                onSelect={this.onSelect1}
+                toggle={<KebabToggle onToggle={this.onToggle1} />}
                 dropdownItems={[
                   <DropdownItem key="link">Link</DropdownItem>,
                   <DropdownItem key="action" component="button">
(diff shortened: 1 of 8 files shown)


mcoker:
Thanks @boaz0! Does this only allow you to pass a single breakpoint to `visibleBreakpoint` or `hiddenBreakpoint`? If so, we'll want to extend it so you can, for example, show at xs (0), hide on sm, show on md, hide on lg, and show again on xl. 

And it looks like `hiddenBreakpoint=""` will hide something at all breakpoints? Personally I think that's fine as long as it's documented, but since you're passing breakpoints, seems like you should also be able to pass "xs" (0) there to do the same thing.

And since this is the same exact functionality that we use in the [table component](https://2251-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/table/) with this syntax `classNames(Visibility.hidden, Visibility.visibleOnMd, Visibility.hiddenOnLg)`, and we'll likely add this functionality to other components in the future, I wonder if we should use a similar approach in the classes or names used to show/hide content?

packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
```diff
@@ -28,14 +46,22 @@ class DataListAction extends React.Component {
...
 
     return (
-      <div className={css(styles.dataListItemAction, className)}>
-        <div className={css(styles.dataListAction, className)} {...props}>
-          {children}
-        </div>
+      <div
+        className={css(
+          styles.dataListItemAction,
+          breakPointTable.visible[visibleBreakpoint],
```
redallen: Let's write some utility classes to do what @mcoker suggests here that are reusable for other components. We seem to be lacking it.

boaz0:
@mcoker thanks for your feedback :heart: ! I will update this PR to match the requirements. :smile:

packages/patternfly-4/react-core/src/components/DataList/DataListAction.js
boaz0: absolutely!

boaz0:
@mcoker & @redallen - PR is updated. Feel free to review :smile:  thanks.

packages/patternfly-4/react-core/src/components/DataList/DataList.md
```diff
@@ -96,15 +98,35 @@ import {
...
+    };
+
+    this.onSelect2 = event => {
+      this.setState(prevState => ({
+        isOpen2: !prevState.isOpen2
+      }));
+    };
+
+    this.onToggle3 = isOpen3 => {
+      this.setState({ isOpen3 });
```
tlabaj: I don't see this state defined

packages/patternfly-4/react-core/src/components/DataList/DataList.md
```diff
@@ -16,6 +16,7 @@ import {
   DataListCell,
   DataListCheck,
   DataListAction,
+  DataListVisibility,
```
christiemolloy: I'm not 100% how you name classes in React, but I wonder if this should either be:
1. **DataListActionVisibility** if we intend to only give the user the option to change visibility for Data List actions, or
2. **Visibility** for consistency, as we name it that in the Table.

boaz0: Yes, you're right. Nevertheless, I will go with the former because I am using Data List Action modifiers.
Is that OK?

packages/patternfly-4/react-core/src/components/DataList/DataList.md
boaz0: Good catch! Thanks a lot!

packages/patternfly-4/react-core/src/components/DataList/DataList.md
christiemolloy: Yes that sounds great @boaz0

packages/patternfly-4/react-core/src/components/DataList/DataList.md
```diff
@@ -96,15 +98,35 @@ import {
 class CheckboxActionDataList extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { isOpen: false };
+    this.state = { isOpen1: false, isOpen2: false };
```
tlabaj: is `isOpen3` missing here?

boaz0: That's odd I thought I added it. Thanks for the double check.

packages/patternfly-4/react-core/src/components/DataList/DataListAction.tsx
```diff
@@ -1,8 +1,22 @@
...
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 import { Omit } from '../../helpers/typeUtils';
 
+export const DataListActionVisibility = pickProperties(styles.modifiers, [
+  'hidden',
+  'hiddenOnSm',
+  'hiddenOnMd',
+  'hiddenOnLg',
+  'hiddenOnXl',
+  'hiddenOn2Xl',
```
mcoker: Just want to confirm this should be `[hidden/visible]On2Xl` and not `[hidden/visible]On2xl`? Not sure how numbers play with camel case.

mcoker: Hmm... I'm trying the `[hidden/visible]On2Xl` props and they don't seem to be working. Can someone confirm? @tlabaj or @christiemolloy. I'm probably doing it wrong :)

tlabaj: I actually was wondering the same thing...

redallen: `hiddenOn2Xl` is correct. It comes from our Javascript variable names being camelcased in react-styles based on the css variables.

tlabaj: I also do not see the [hidden/visible]On2Xl being applied.

boaz0: For now I wrote a workaround for that. Please review and let me know what you think.

Thanks.

packages/patternfly-4/react-core/src/components/DataList/DataListAction.tsx
```diff
@@ -1,8 +1,26 @@
...
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 import { Omit } from '../../helpers/typeUtils';
 
+const visibilityModifiers = pickProperties(styles.modifiers, [
+  'hidden',
+  'hiddenOnSm',
+  'hiddenOnMd',
+  'hiddenOnLg',
+  'hiddenOnXl',
+  'hiddenOn_2xl',
```
redallen: Workaround looks good. Thanks as always @boaz0 ! :+1:

mcoker:
@boaz0 the preview URL isn't working for some reason, but I built it locally and looks like the 2xl breakpoints are working, but looks like the base `DataListActionVisibility.hidden` isn't working.

packages/patternfly-4/react-core/src/components/DataList/DataList.md
```diff
@@ -184,6 +225,53 @@ class CheckboxActionDataList extends React.Component {
...
+                dropdownItems={[
+                  <DropdownItem key="pri-action3" component="button">Primary</DropdownItem>,
+                  <DropdownItem key="sec1-action3" component="button">Secondary</DropdownItem>,
+                  <DropdownItem key="sec2-action3" component="button">Secondary</DropdownItem>,
+                  <DropdownItem key="sec3-action3" component="button">Secondary</DropdownItem>,
+                ]}
+              />
+            </DataListAction>
+            <DataListAction
+              className={css(DataListActionVisibility.visibleOnXl, DataListActionVisibility.hidden)}
```
jschuler: The buttons are still visible when i reduce the screen width

boaz0: Thanks. I will be looking into it. :smile:

boaz0: Yep, I found the problem!

boaz0: I am just adding unit tests.

boaz0:
Thanks a lot: @jschuler @redallen @mcoker @tlabaj and @christiemolloy

tlabaj:
@boaz0 @mcoker @redallen 
This PR was merged before it was ready. the 2Xl breakpoints are still not working, I have opened issue #2490 for follow up work.

boaz0:
@tlabaj can you give me a way to reproduce it in https://patternfly-react.surge.sh/patternfly-4/components/datalist/

mcoker:
@boaz0 @tlabaj I also just confirmed it looks OK on master. I was testing previously using the preview build in this PR. Maybe the preview wasn't updating properly. I think we can close #2490

PR #2252, opened 2019-06-13
feat(PF Checkbox): Add 3rd state to checkbox controlled by consumer

karelhala:
**What**:
When using checkbox as hierarchical checkbox (parent checkbox with child checkboxes) and user selects only some of child checkboxes parent one should go to 3rd state to indicate user that only some checkboxes were selected.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.md
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.md
@@ -17,7 +17,8 @@ class ControlledCheckbox extends React.Component {
     super(props);
     this.state = {
       check1: false,
-      check2: false
+      check2: false,
+      check3: null
     };
     this.handleChange = (checked, event) => {
       const target = event.target;
@@ -45,6 +46,14 @@ class ControlledCheckbox extends React.Component {
           aria-label="controlled checkbox example"
           id="check-2"
           name="check2"
+        />
+        <Checkbox
+          label="Controlled CheckBox"
+          isChecked={this.state.check3}
+          onChange={this.handleChange}
+          aria-label="controlled checkbox example"
+          id="check-3"
+          name="check3"
         />
       </React.Fragment>
     );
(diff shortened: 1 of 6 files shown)


rachael-phillips:
@mcarrano @tlabaj have we introduced this in Core? If this is a need we should also get ahead of this in Core.

karelhala:
@rachael-philips this is just a JS update no need to change or anything in core it uses indeterminate API od checkbox [1].

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate

rachael-phillips:
Thanks for the explanation @karelhala !

mcarrano:
I am fine to introduce this directly into React. We may decide to revisit this later based on the discussion here: https://github.com/patternfly/patternfly-next/issues/1411

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.md
```diff
@@ -46,6 +47,14 @@ class ControlledCheckbox extends React.Component {
           id="check-2"
           name="check2"
         />
+        <Checkbox
```
jschuler: would be good to wrap 2 checkboxes inside the checkbox and show the indeterminate state live. And can also get rid of several other duplicate examples

karelhala: Yeah! That's a good point. I like it!

packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -77,6 +79,7 @@ export class Checkbox extends React.Component<CheckboxProps> {
           aria-invalid={!isValid}
           aria-label={ariaLabel}
           disabled={isDisabled}
+          ref={elem => elem && (elem.indeterminate = isChecked === null)}
```
redallen: What's this ref needed for?

karelhala: This prop is can't be set trough react prop for some reason [1]. However plain js api allows to set it trough element, so this ref is to access checkox element. It's probably because indeterminate is a flag and not real prop though.

[1] https://github.com/facebook/react/issues/1798

Issue #2259, opened 2019-06-13
Bug - ChartDonutThreshold should not have tooltip labels by default.


TheRealJon:
If no 'labels` prop is passed to `ChartDonutThreshold`, the default labels just show an index value:

![video-convert](https://user-images.githubusercontent.com/22625502/59447835-f800e400-8dd1-11e9-88fd-c88c8bdcdeaf.gif)

rachael-phillips:
@TheRealJon is this blocking you?

TheRealJon:
@rachael-phillips No, not a blocker, just something I noticed. I am manually overriding the label values, but ideally, the default should be handled upstream. I am working on fixes for this and #2258 now

TheRealJon:
Fixed by https://github.com/patternfly/patternfly-react/pull/2270

PR #2270, opened 2019-06-13
fix(ChartDonutThreshold): Don't show static threshold donut tooltips by default.

### LINKED ISSUE #2258: Bug - ChartDonutThreshold x and y accessors break child ChartDonutUtilization component
If an `x` or `y` accessor props are passed into `ChartDonutThreshold`, they are inadvertently inherited by the `ChartDonutUtilization` child component. This happens on L452 of `ChartDonutThreshold` where the `getData` function is used against the child data.

TheRealJon:
Pie chart tooltips display the x value of each slice as a tooltip label by default. Because we are
doing calculations to mutate the data, this can lead to array indices being displayed as slice labels. Change defaults so that no slice labels are displayed for the static threshold donut. Also, fix a bug where ChartDonutThreshold accessors were being used to format child ChartDonutUtilization data.

Fixes #2258 and #2259

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
+++ b/packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
@@ -392,6 +392,7 @@ export const ChartDonutThreshold: React.FunctionComponent<ChartDonutThresholdPro
   children,
   data = [],
   donutOrientation = 'left',
+  labels = [], // Don't show any tooltip labels by default, let consumer override if needed
   standalone = true,
   themeColor,
   themeVariant,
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
```diff
@@ -405,25 +406,34 @@ export const ChartDonutThreshold: React.FunctionComponent<ChartDonutThresholdPro
...
+        dataPoint._y, // Set the previous value to current y value
+        [
+          ...acc[1],
+          {
+            x: dataPoint._x, // Conditionally add x property only if it is in the original data object
+            y: dataPoint._y - acc[0] // Must be offset by previous value
+          }
+        ]
+      ];
+    }, [0, []]);
```
dlabrecq: Should we apply this to donut utilization as well?

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
```diff
@@ -449,24 +459,25 @@ export const ChartDonutThreshold: React.FunctionComponent<ChartDonutThresholdPro
...
-          height: child.props.height || height,
-          showStatic: child.props.showStatic || false,
+          donutWidth: childProps.donutWidth || donutWidth - (theme.pie.width - dynamicTheme.pie.width),
+          endAngle: childProps.endAngle || 360 * (datum[0]._y ? datum[0]._y / 100 : 100),
+          height: childProps.height || height,
+          showStatic: childProps.showStatic || false,
           standalone: false,
           theme: dynamicTheme,
-          width: child.props.width || width
+          width: childProps.width || width
```
dlabrecq: Could we apply `... childProps` at the end? Thinking it would help make the code more readable.

TheRealJon: Yeah, I think we can do something like this:
```jsx
return React.cloneElement(child, {
  donutDx: getDynamicDonutDx(dynamicTheme, orientation),
  donutDy: getDynamicDonutDy(dynamicTheme),
  donutHeight: donutHeight - (theme.pie.height - dynamicTheme.pie.height),
  donutOrientation: orientation,
  donutWidth: donutWidth - (theme.pie.width - dynamicTheme.pie.width),
  endAngle: 360 * (datum[0]._y ? datum[0]._y / 100 : 100),
  height,
  showStatic: false,
  standalone: false,
  theme: dynamicTheme,
  width: width,
  ...childProps,
});
```

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
TheRealJon: I almost did but thought it was probably outside of the scope of this PR.

TheRealJon: We could probably pull out the utilization/threshold data formatting function into a single util function. The logic is very similar, we just need to handle the difference between a single utilization data point and multiple threshold data points.

dlabrecq: Let's keep them separate. The code is complex and not easy to understand what's happening.

Issue #2213, opened 2019-06-11
DonutUtilizationChart center label is not customizable


TheRealJon:
While using this component for the first time, I've come across the issue that there is no easy way to customize/override the center label text or styling for the utilization donut component. There should be a way to either override the center text label component altogether or at least override the styles, (maybe both).

dlabrecq:
PatternFly-core is actually making variables for these fonts; 24 for title, and 14 for subtitle. (edited) 

Currently, they're a little larger at 26 and 16.

dlabrecq:
After speaking with @gdoyle1, I will add a property to position the subtitle; center, below, right. This will be similar to the PF3 pattern where the subtitle is positioned to the right or bottom of a small chart.

<img width="239" alt="Screen Shot 2019-06-12 at 9 32 21 AM" src="https://user-images.githubusercontent.com/17481322/59355874-d16d7b00-8cf5-11e9-9bb9-82a39e7fdf51.png">

I will also add a labelComponent property to allow users to override the label if necessary.

gdoyle1:
Here's the spec @dlabrecq https://marvelapp.com/i221729

Issue #2183, opened 2019-06-07
Inconsistent chart legends


gdoyle1:
@dlabrecq @mcarrano @rachael-phillips @jgiardino 

After charts were released, I realized that donut utilization with thresholds is the only legend with text including "[variable name] - [data value]" (i.e. storage capacity - 25%). I believe we had to add the data value in order for this chart to be accessible. Wondering if we can add that text to all legends so they are consistent, before our next release? Also if we can change the "-" to a ":". So "Storage capacity: 25%" or "Dogs: 5" etc.

dlabrecq:
In order for the charts to be accessible, we need to add a title and description to the chart. Changing the legend text doesn't make it accessible because it's all an SVG.

IMO, this could just be added to the chart updates described here.
https://github.com/patternfly/patternfly-react/issues/1692

mcarrano:
I'm fine to see these issues combined. @rachael-phillips is this work anywhere on our roadmap?

dlabrecq:
FYI, This is the a11y issue to add title and description.
https://github.com/patternfly/patternfly-react/issues/1498

jgiardino:
Thanks for catching this, @gdoyle1! From a design perspective, I think it's a great idea to include numeric values in the legend. If we do decide to recommend this from a design perspective, I'm not sure how much would be built in to the chart components vs just represented as best practice in our examples.

Re: @dlabrecq's point, I had created https://github.com/patternfly/patternfly-react/issues/1498 to make updates to our chart examples so that they are screen reader accessible. But I see this update as more of a benefit for these users:
- low vision users that might not be using a screen reader - they might have difficulty associating colors in the legend with colors in the chart. The argument could be made that they can get this info on hover, but they might also be keyboard-only users...
- keyboard-only sighted users - the numeric values are available on hover, but not on focus. I think this was one of the main reasons we had decided to add the numbers to the chart legend for donut utilization with thresholds.

rachael-phillips:
These changes still need to be slated on the roadmap. @gdoyle1 @mcarrano do you mind telling me a bit more about what impact you think this is for charting? Also, is this something that needs to be fixed in Core or React?

mcarrano:
I don't believe there is any Core impact here, but @dlabrecq tell me if I'm wrong. User impact concerns only accessibility as @jgiardino has outlined above. I guess there is also now a consistency issue as donut charts have implemented this and other charts have not.

dlabrecq:
[...]

PR #2271, opened 2019-06-13
react-charts: position subtitle center, right, bottom for small charts

dlabrecq:
Refactored the donut, donut utilization, and donut threshold charts in order to position the subtitle to the right or below the chart, similar to the PF3 functionality for small charts (e.g., 150x150px). 

With this change, the label and legend creation has been moved from donut utilization to donut. The donut did not support legends, previously.

In addition, I've refactored the legends so that they are positioned with 16px padding (8px for labels) per the mock below.

Screen Shot 2019-06-13 at 7 24 36 PM

Fixes https://github.com/patternfly/patternfly-react/issues/2213
Fixes https://github.com/patternfly/patternfly-react/issues/2183

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.tsx
+++ b/packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.tsx
@@ -147,11 +147,11 @@ export interface ChartDonutProps extends ChartPieProps {
    */
   donutHeight?: number;
   /**
-   * Defines a horizontal shift from the x coordinate. This should not be set manually.
+   * Defines a horizontal shift from the x coordinate. It should not be set manually.
    */
   donutDx?: number;
   /**
-   * Defines a vertical shift from the y coordinate. This should not be set manually.
+   * Defines a vertical shift from the y coordinate. It should not be set manually.
    */
   donutDy?: number;
   /**
@@ -299,11 +299,11 @@ export interface ChartDonutProps extends ChartPieProps {
    */
   legendData?: any[];
   /**
-   * Defines a horizontal shift from the x coordinate. This should not be set manually.
+   * Defines a horizontal shift from the x coordinate. It should not be set manually.
    */
   legendDx?: number;
   /**
-   * Defines a vertical shift from the y coordinate. This should not be set manually.
+   * Defines a vertical shift from the y coordinate. It should not be set manually.
    */
   legendDy?: number;
   /**
@@ -390,11 +390,11 @@ export interface ChartDonutProps extends ChartPieProps {
    */
   subTitleComponent?: React.ReactElement<any>;
   /**
-   * Defines a horizontal shift from the x coordinate. This should not be set manually.
+   * Defines a horizontal shift from the x coordinate. It should not be set manually.
    */
   subTitleDx?: number;
   /**
-   * Defines a vertical shift from the y coordinate. This should not be set manually.
+   * Defines a vertical shift from the y coordinate. It should not be set manually.
    */
   subTitleDy?: number;
   /**
@@ -506,12 +506,6 @@ export const ChartDonut: React.FunctionComponent<ChartDonutProps> = ({
       return null;
     }
     const props = legendComponent.props ? legendComponent.props : {};
-    const legendDimensions = getLegendDimensions({
-      legendData,
-      legendOrientation,
-      legendProps: props,
-      theme
-    });
     return React.cloneElement(legendComponent, {
       data: legendData,
       orientation: legendOrientation,
@@ -523,7 +517,7 @@ export const ChartDonut: React.FunctionComponent<ChartDonutProps> = ({
         legendData,
         legendOrientation: props.legendOrientation ? props.legendOrientation : legendOrientation,
         legendPosition,
-        legendWidth: legendDimensions.width,
+        legendProps: props,
         theme,
         svgWidth: width
       }),
@@ -532,7 +526,8 @@ export const ChartDonut: React.FunctionComponent<ChartDonutProps> = ({
         chartType: 'pie',
         dy: legendDy,
         legendData: props.data ? props.data : legendData,
-        legendHeight: legendDimensions.height,
+        legendOrientation: props.legendOrientation ? props.legendOrientation : legendOrientation,
+        legendProps: props,
         legendPosition,
         theme
       }),
--- a/packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.tsx
+++ b/packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.tsx
@@ -492,7 +492,7 @@ export const ChartDonut: React.FunctionComponent<ChartDonutProps> = ({
   // destructure last
   theme = getDonutTheme(themeColor, themeVariant),
   legendOrientation = theme.legend.orientation as ChartDonutLegendOrientation,
-  capHeight = title && subTitle ? 1.1 : undefined,
+  capHeight = 1.1,
   height = theme.pie.height,
   width = theme.pie.width,
   donutHeight = Math.min(height, width),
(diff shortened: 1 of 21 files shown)


packages/patternfly-4/react-charts/src/components/ChartDonut/ChartDonut.tsx
```diff
@@ -338,47 +471,167 @@ export interface ChartDonutProps extends ChartPieProps {
...
+
+  // Returns title
+  const getTitle = () => {
+    if (!title) {
+      return null;
+    }
+    const props = titleComponent ? titleComponent.props : {};
+    const showBoth = title && subTitle && subTitlePosition == ChartDonutSubTitlePosition.center;
+    return React.cloneElement(titleComponent, {
+      capHeight: showBoth ? capHeight : undefined,
```
TheRealJon: Just discovered you can do this to conditionally define a prop:
```suggestion
     ...showBoth && { capHeight }
```

If `showBoth` is false, the capHeight prop won't be set at all, instead of getting `{capHeight: undefined}`

dlabrecq: Neat. We only need capHeight when both are present. Don't want it applied when there is no subtitle.

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
```diff
@@ -426,65 +447,100 @@ export const ChartDonutThreshold: React.FunctionComponent<ChartDonutThresholdPro
...
-  // Returns the vertical shift for the dynamic utilization donut cart
-  const getDynamicDonutDy = (dynamicTheme: ChartThemeDefinition) => {
+  // Returns the vertical shift for the donut utilization chart
+  const getDonutDy = (dynamicTheme: ChartThemeDefinition) => {
     const dynamicHeight = donutHeight - (theme.pie.height - dynamicTheme.pie.height);
     return Math.round((donutHeight - dynamicHeight) / 2);
   }
 
+  // Returns the vertical shift for the donut utilization legend and subtitle
+  const getDy = (dynamicTheme: ChartThemeDefinition, position: string) => {
```
TheRealJon:
```suggestion
  const getLegendDy = (dynamicTheme: ChartThemeDefinition, position: string) => {
```

dlabrecq: I'm using this to shift both legend and subtitle, so I had renamed getLegendDy as getDy

dlabrecq: I'll rename it as getLegendAndSubTitleDy to be more clear

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
```diff
@@ -426,65 +447,100 @@ export const ChartDonutThreshold: React.FunctionComponent<ChartDonutThresholdPro
...
-          standalone: false,
-          theme: dynamicTheme,
-          width: child.props.width || width
-        });
+      const props = child.props;
+      if (!props) {
+        return child;
       }
-      return child;
+      const datum = getData([{ ...props.data }]);
```
TheRealJon: No need for spread operator:
```suggestion
      const datum = getData([props.data]);
```

packages/patternfly-4/react-charts/src/components/ChartPie/examples/ChartPie.md
```diff
@@ -40,21 +40,23 @@ import { ChartLegend, ChartPie, ChartThemeColor } from '@patternfly/react-charts
...
   <ChartLegend
-    data={[{ name: 'Cats' }, { name: 'Dogs' }, { name: 'Birds' }]}
-    height={35}
-    orientation={'horizontal'}
-    responsive={false}
-    themeColor={ChartThemeColor.multi}
-    themeVariant={ChartThemeVariant.light}
-    x={8}
-  />
+        data={[{ name: 'Cats: 35' }, { name: 'Dogs: 55' }, { name: 'Birds: 10' }]}
```
TheRealJon: Fix extra whitespace

packages/patternfly-4/react-charts/src/components/ChartUtils/chart-legend.ts
```diff
@@ -27,56 +39,92 @@ export const getChartPadding = (chartType: string = 'chart', theme: ChartThemeDe
...
-    case 'top':
-      return height > lHeight ? height - lHeight + chartDy : 0;
+      return dHeight > lHeight ? Math.round((dHeight - lHeight) / 2) + dy : dy;
     default:
-      return chartDy;
+      return dy;
+  }
+};
+
+const getBorderPaddingWorkAround = (legendData: any[], legendOrientation: string) => {
```
TheRealJon: There is a util function in Victory for estimating text width in pixels that might be useful here:
```js
TextSize.approximateTextSize(string, styleObj);
```

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
TheRealJon: Oh, whoops. Sorry, missed that in the comment

packages/patternfly-4/react-charts/src/components/ChartUtils/chart-legend.ts
dlabrecq: Thanks, I'll look at that.

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
```diff
@@ -426,65 +447,100 @@ export const ChartDonutThreshold: React.FunctionComponent<ChartDonutThresholdPro
...
+      const legendPos = props.legendPosition || legendPosition;
+      const subTitlePos = props.subTitlePosition || subTitlePosition;
+      const dynamicTheme = getDonutThresholdDynamicTheme(
+        props.themeColor || themeColor, props.themeVariant || themeVariant);
+      return React.cloneElement(child, {
+        donutDx: getDonutDx(dynamicTheme, legendPos),
+        donutDy: getDonutDy(dynamicTheme),
+        donutHeight: donutHeight - (theme.pie.height - dynamicTheme.pie.height),
+        donutWidth:donutWidth - (theme.pie.width - dynamicTheme.pie.width),
+        endAngle: 360 * (datum[0]._y ? datum[0]._y / 100 : 100),
```
TheRealJon: Not new, but this could result in an end angle value of `36000`:

```suggestion
        endAngle: 360 * (datum[0]._y || 100 / 100),
```

dlabrecq: That produces some strange results

Screen Shot 2019-06-14 at 11 59 07 AM

dlabrecq: The end angle should be zero with no value, so looking at this:
```
endAngle: 360 * (datum[0]._y ? datum[0]._y / 100 : 0),
```

TheRealJon: That could also be simplified to 
```js
endAngle: 360 * (datum[0]._y || 0 / 100),
```

packages/patternfly-4/react-charts/src/components/ChartUtils/chart-legend.ts
dlabrecq: I used TextSize to help calculate the difference between the actual text size and legend width.

packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutThreshold.tsx
dlabrecq: That also did not work properly because it's either `datum[0]._y` or `0/100`

TheRealJon: Oh, it's just an order of ops problem:

```
endAngle: 360 * ((datum[0]._y || 0) / 100),
```
Would probably work, but it's not really important. Just a nit.

PR #2273, opened 2019-06-14
fix(accordion): adds prop to heading level of parent component

### LINKED ISSUE #1881: Accordion - prop for heading level

jenny-s51:
**What**: closes #1881 , #2273


packages/patternfly-4/react-core/src/components/Accordion/AccordionToggle.js
```diff
@@ -3,21 +3,27 @@ import PropTypes from 'prop-types';
 import { css } from '@patternfly/react-styles';
 import { AngleRightIcon } from '@patternfly/react-icons';
 import styles from '@patternfly/react-styles/css/components/Accordion/accordion';
+import { AccordionContext } from './Accordion';
+import { ApplicationLauncherPosition } from '../ApplicationLauncher/applicationLauncherConstants';
```
jschuler: this should not be here

packages/patternfly-4/react-core/src/components/Accordion/AccordionToggle.js
```diff
@@ -3,21 +3,27 @@ import PropTypes from 'prop-types';
...
-        <AngleRightIcon className={css(styles.accordionToggleIcon)} />
-      </button>
-    </h3>
+    <AccordionContext.Consumer>
+      {({ HeadingLevel }) => (
+        <HeadingLevel>
+          <button
+            id={id}
+            className={css(styles.accordionToggle, isExpanded && styles.modifiers.expanded, className)}
+            {...props}
```
jschuler: would you mind making `{...props}` the last attribute

packages/patternfly-4/react-core/src/components/Accordion/Accordion.js
```diff
@@ -3,17 +3,34 @@ import PropTypes from 'prop-types';
...
 import styles from '@patternfly/react-styles/css/components/Accordion/accordion';
 
-const Accordion = ({ children, className, 'aria-label': ariaLabel, ...props }) => (
-  <dl className={css(styles.accordion, className)} aria-label={ariaLabel} {...props}>
-    {children}
-  </dl>
-);
+export const AccordionContext = React.createContext({ headingLevel: 'h3' });
+
+export const HeadingLevelTypes = {
```
jschuler: Why is this needed? I think we can just pass `headingLevel` through into context

jenny-s51: Do you mean I should just set `AccordionContext = React.createContext();` ?

packages/patternfly-4/react-core/src/components/Accordion/AccordionToggle.js
mturley: Ah, the price of auto-import-on-tab-completion.

packages/patternfly-4/react-core/src/components/Accordion/Accordion.js
mturley: I think this const is good to have for the consumer (it'll be the same as enums we export elsewhere, when this is ported over to the typescript version) but Joachim is saying that on line 18 we don't need `const HeadingLevel = HeadingLevelTypes[headingLevel];`, we can just pass lowercase-h `headingLevel` into the provider. The HeadingLevelTypes object doesn't really do anything for us internally, it's just for the consumer

packages/patternfly-4/react-core/src/components/Accordion/Accordion.tsx
```diff
@@ -2,22 +2,44 @@ import * as React from 'react';
...
+export enum HeadingLevel {
+  h1 = 'h1',
+  h2 = 'h2',
+  h3 = 'h3',
+  h4 = 'h4',
+  h5 = 'h5',
+  h6 = 'h6'
+}
+
+// declare global {
```
mturley: We can fully remove this now.

packages/patternfly-4/react-core/src/components/Accordion/Accordion.tsx
```diff
@@ -2,22 +2,44 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Accordion/accordion';
 
+export const AccordionContext = React.createContext('h3');
+
+export enum HeadingLevel {
```
mturley: Judging by our discussion in Slack, since the user already has `TitleLevel` we don't need this enum after all.

packages/patternfly-4/react-core/src/components/Accordion/Accordion.tsx
```diff
@@ -2,22 +2,44 @@ import * as React from 'react';
...
 export const Accordion: React.FunctionComponent<AccordionProps> = ({
   children = null,
   className = '',
   'aria-label': ariaLabel = '',
+  headingLevel = 'h3',
   ...props
 }: AccordionProps) => (
   <dl className={css(styles.accordion, className)} aria-label={ariaLabel} {...props}>
-    {children}
+    <AccordionContext.Provider value={ headingLevel }>{children}</AccordionContext.Provider>
```
mturley: I guess we still have eslint turned off, but I think when it's eventually turned back on it'll complain about the spacing here, I think you want to remove the spaces around `headingLevel` here (`value={headingLevel}`)

jenny-s51: Thank you! I resolved changes you mentioned above in this commit [https://github.com/patternfly/patternfly-react/pull/2273/commits/9db9ef3ff99503ea299d88585119f0b117745ad7](here)

packages/patternfly-4/react-core/src/components/Accordion/Accordion.tsx
```diff
@@ -1,6 +1,9 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Accordion/accordion';
+import {TitleLevel} from '../Title';
```
mturley: You don't need to import this here, it's not being used in the component itself.

packages/patternfly-4/react-core/src/components/Accordion/AccordionToggle.tsx
```diff
@@ -23,16 +25,20 @@ export const AccordionToggle: React.FunctionComponent<AccordionToggleProps> = ({
...
-        className={css(styles.accordionToggle, isExpanded && styles.modifiers.expanded, className)}
-        {...props}
-        aria-expanded={isExpanded}
-      >
-        <span className={css(styles.accordionToggleText)}>{children}</span>
-        <AngleRightIcon className={css(styles.accordionToggleIcon)} />
-      </button>
-    </h3>
+    <AccordionContext.Consumer>
+      {(TitleLevel: any) => (
```
mturley: Sorry to be confusing here, but I think this is actually better off staying as `HeadingLevel` like it was before, since that's the name of the prop that is passed into Accordion, and it'll be more clear where this came from. `TitleLevel` is just the object a consumer can use to get predefined values for what they should pass into the `headingLevel` prop. I don't think you need any reference to `TitleLevel` at all in these components, you could use it maybe just in a unit test (e.g.`<Accordion headingLevel={TitleLevel.h2}> ...`)

packages/patternfly-4/react-core/src/components/AboutModal/__snapshots__/AboutModalBoxCloseButton.test.tsx.snap
```diff
@@ -11,7 +11,6 @@ exports[`AboutModalBoxCloseButton Test 1`] = `
   >
     <TimesIcon
       color="currentColor"
-      noVerticalAlign={false}
```
mturley: I'm not sure why these unrelated snapshots are updating, @jschuler @redallen do you know what the deal is here?

packages/patternfly-4/react-core/src/components/Accordion/AccordionToggle.tsx
mturley: But even for the unit test, since you'd have to import it from over in the Title folder, I don't think you necessarily need to use `TitleLevel` in this PR at all.

packages/patternfly-4/react-core/src/components/AboutModal/__snapshots__/AboutModalBoxCloseButton.test.tsx.snap
jschuler: react-icons was recently updated to include this prop, pull down latest from master and rerun yarn install

jenny-s51:
final PR is #2273

PR #2276, opened 2019-06-15
CircleCI build: Increment key version to clear cache

dlabrecq:
Added `v2` prefix to the key name. This will clear the cache and get PR builds working, again.

See https://circleci.com/docs/2.0/caching/#clearing-cache

Fixes https://github.com/patternfly/patternfly-react/issues/2277


redallen:
`{{ .Environment.CACHE_VERSION }}` is an environment variable set on Circle's webpage I've been bumping often, so this change won't do much.

PR #2278, opened 2019-06-17
feat(chipGroup): add heading level prop to toolbar chip group label

### LINKED ISSUE #2167: ChipGroup  - prop for heading level
A new prop `headingLevel` should be added to the ChipGroup component to allow the user to specify heading levels. Supported values should be `h1` – `h6`.

This would affect the heading element that is used for `.pf-c-chip-group__label` in the example below:
```
<ul class="pf-c-chip-group pf-m-toolbar">
  <li>
    <h4 class="pf-c-chip-group__label">Category 1</h4>
  ...
```

Similar to my comment in #1881 for Accordion, I think this prop belongs on the parent `<ChipGroup>` component so that the same level is used for all categories. I don't see a use case where these should be different heading levels.

And assuming we land with defaulting to `<h1>` for #1881, then we should be consistent and do that here as well.

boaz0:
**What**:

closes #2167 

- [x] add heading level prop to ChipGroup
- [x] pass heading level to Chip Group Toolbar item through Context API
- [x] update tests
- [x] ~~HeadingLevel type was moved from `Title` to `helpers`~~
- [x] ~~Chip Group Toolbar Item is using HeadingLevel type in its own heading level prop~~

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroup.tsx
+++ b/packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroup.tsx
@@ -4,6 +4,8 @@ import { css } from '@patternfly/react-styles';
 import { Chip } from './Chip';
 import { fillTemplate } from '../../helpers';
 
+export const ChipGroupContext = React.createContext('');
+
 export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside the chip text */
   children?: React.ReactNode;
@@ -15,6 +17,8 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
   collapsedText?: string;
   /** Flag for grouping with a toolbar & category name */
   withToolbar?: boolean;
+  /** Set heading level to the chip item label */
+  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 }
 
 interface ChipGroupState {
@@ -44,7 +48,12 @@ export class ChipGroup extends React.Component<ChipGroupProps, ChipGroupState>{
 
   renderToolbarGroup() {
     const { isOpen } = this.state;
-    return <InnerChipGroup {...this.props} isOpen={isOpen} onToggleCollapse={this.toggleCollapse} />;
+    const { headingLevel = 'h4' } = this.props;
+    return (
+      <ChipGroupContext.Provider value={headingLevel}>
+        <InnerChipGroup {...this.props} isOpen={isOpen} onToggleCollapse={this.toggleCollapse} />
+      </ChipGroupContext.Provider>
+    );
   }
 
   renderChipGroup() {
--- a/packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
+++ b/packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
@@ -1,14 +1,15 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/ChipGroup/chip-group';
+import {ChipGroupContext} from './ChipGroup';
 
 export interface ChipGroupToolbarItemProps extends React.HTMLProps<HTMLUListElement> {
   /**  Category name text */  
   categoryName?: string;
   /** Content rendered inside the chip text */
-  children: React.ReactNode; 
+  children: React.ReactNode;
   /** Additional classes added to the chip item */
-  className?: string;   
+  className?: string;
 }
 
 export const ChipGroupToolbarItem: React.FunctionComponent<ChipGroupToolbarItemProps> = ({
@@ -18,13 +19,19 @@ export const ChipGroupToolbarItem: React.FunctionComponent<ChipGroupToolbarItemP
   ...props
 }: ChipGroupToolbarItemProps) => {
     if (React.Children.count(children)) {
-      return(
-        <ul className={css(styles.chipGroup, styles.modifiers.toolbar)} {...props}>
-          <li>
-            <h4 className={css(styles.chipGroupLabel)}>{categoryName}</h4>
-            <ul className={css(styles.chipGroup)}>{children}</ul>
-          </li>
-        </ul>
+      return (
+        <ChipGroupContext.Consumer>
+          {(HeadingLevel: any) => {
+            return (
+              <ul className={css(styles.chipGroup, styles.modifiers.toolbar)} {...props}>
+                <li>
+                  <HeadingLevel className={css(styles.chipGroupLabel)}>{categoryName}</HeadingLevel>
+                  <ul className={css(styles.chipGroup)}>{children}</ul>
+                </li>
+              </ul>
+            );
+          }}
+        </ChipGroupContext.Consumer>
       );
     }
     return null;
--- a/packages/patternfly-4/react-core/src/components/ChipGroup/__snapshots__/ChipGroup.test.tsx.snap
+++ b/packages/patternfly-4/react-core/src/components/ChipGroup/__snapshots__/ChipGroup.test.tsx.snap
@@ -28,26 +28,30 @@ exports[`ChipGroup chip group default 1`] = `
 `;
 
 exports[`ChipGroup chip group with toolbar 1`] = `
-<InnerChipGroup
-  className=""
-  collapsedText="\${remaining} more"
-  expandedText="Show Less"
-  isOpen={false}
-  onToggleCollapse={[Function]}
-  withToolbar={true}
+<ContextProvider
+  value="h4"
 >
-  <Component>
-    <Chip
-      className=""
-      closeBtnAriaLabel="close"
-      component="div"
-      isOverflowChip={false}
-      isReadOnly={false}
-      onClick={[Function]}
-      tooltipPosition="top"
-    >
-      1.1
-    </Chip>
-  </Component>
-</InnerChipGroup>
+  <InnerChipGroup
+    className=""
+    collapsedText="\${remaining} more"
+    expandedText="Show Less"
+    isOpen={false}
+    onToggleCollapse={[Function]}
+    withToolbar={true}
+  >
+    <Component>
+      <Chip
+        className=""
+        closeBtnAriaLabel="close"
+        component="div"
+        isOverflowChip={false}
+        isReadOnly={false}
+        onClick={[Function]}
+        tooltipPosition="top"
+      >
+        1.1
+      </Chip>
+    </Component>
+  </InnerChipGroup>
+</ContextProvider>
 `;


packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -1,31 +1,35 @@
...
+  className?: string;
+  /** Set heading level to the chip item label */
+  headingLevel?: headingLevelType;
 }
 
 export const ChipGroupToolbarItem: React.FunctionComponent<ChipGroupToolbarItemProps> = ({
   categoryName = '', 
   children = null, 
   className = '', 
+  headingLevel: HeadingLevel = 'h1',
```
redallen: Any particular reason for `h1` as the default instead of `h4` like it used to be?

boaz0: This was one of the requirements made in #2167 by @jgiardino

jgiardino: Ah, right. This is what I had stated in #2167. 

> And assuming we land with defaulting to `<h1>` for #1881, then we should be consistent and do that here as well.

Apologies for not updating that issue after #1881 closed. For that update, we ended up using the same heading level that was originally defined in the component. One concern was that changing the heading level could be considered a breaking change. But also, Accordion ended up presenting a unique case in that we will update it in the future to support a variation that does not use headings (https://github.com/patternfly/patternfly-next/issues/1913). Which means when `headingLevel` is not specified, we would use the non-heading variant.

There was also a lot of conversation around default heading levels when the Accordion was getting updated. The general assumption is that this is a prop that will very likely get overlooked, and so the default value is important to consider.  @seanforyou23 was suggesting that it's probably better to default to lower heading levels for a component that wouldn't be used as the a primary page heading, since lower heading levels are less likely to conflict with the heading outline that exists on the page.  

For this update, I suggest using what we currently have—`<h4>`—and we can open a new issue in the future if we find that this default is resulting in issues with the overall heading outline on a page.

Thanks for working on this, @boaz0!

jgiardino:
The prop works as expected! 🎉 

The only question I have is whether the prop should move to the parent `<ChipGroup>` component?

With the Accordion component, we discussed this and ended up supporting the prop on the parent `<Accordion>` component, because in that component, it would be really odd to have different heading levels for different toggles (because there would then be a disconnect between the visual presentation and the semantic structure).

I would think that same reason is true in this case, but am curious if anyone has some reason for supporting this prop on the `<ChipGroupToolbarItem>` component instead.

I'm pulling in @mturley because he had provided suggestions on how to implement this for Accordion, and might have relevant info that would be helpful to share for this case as well.

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -1,31 +1,35 @@
...
   /**  Category name text */  
   categoryName?: string;
   /** Content rendered inside the chip text */
-  children: React.ReactNode; 
+  children: React.ReactNode;
   /** Additional classes added to the chip item */
-  className?: string;   
+  className?: string;
+  /** Set heading level to the chip item label */
+  headingLevel?: headingLevelType;
```
jessiehuff: Instead of `headingLevelType;` use `headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';`

packages/patternfly-4/react-core/src/components/Title/Title.tsx
```diff
@@ -21,7 +22,7 @@ export interface TitleProps extends Omit<React.HTMLProps<HTMLHeadingElement>, 's
   /** Additional classes added to the Title */
   className?: string;
   /** the heading level to use */
-  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  headingLevel?: headingLevelType;
```
jessiehuff: I think we need `headingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';` for the docs.

packages/patternfly-4/react-core/src/helpers/heading.ts
```diff
@@ -0,0 +1 @@
+export type headingLevelType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
jessiehuff: I don't think we need this file.

mturley:
@boaz0, I agree with @jgiardino that we should probably have the consumer pass one `headingLevel` prop to the parent `<ChipGroup>` instead of having to pass it to each `<ChipGroupToolbarItem>`, since it reduces duplication in the consumer's code and prevents them from using different heading levels in the same chip group.

@jenny-s51 implemented this for Accordion using [React's Context API](https://reactjs.org/docs/context.html) in this PR: https://github.com/patternfly/patternfly-react/pull/2290/files. Basically in the ChipGroup component you'd accept `headingLevel` as a prop and pass it to a `ChipGroupContext.Provider`, and then in ChipGroupToolbarItem you'd render a `ChipGroupContext.Consumer` that takes a function for a child, and this function can pull the `headingLevel` out of its arguments and return the children rendered with that headingLevel.

If you need any help, feel free to let me know!

boaz0:
@mturley no problem I will be working on this.
Thanks for your feedback!

boaz0:
@mturley updated PR feel free to review and give some feedback.

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -1,14 +1,16 @@
...
 export interface ChipGroupToolbarItemProps extends React.HTMLProps<HTMLUListElement> {
   /**  Category name text */  
   categoryName?: string;
   /** Content rendered inside the chip text */
-  children: React.ReactNode; 
+  children: React.ReactNode;
   /** Additional classes added to the chip item */
-  className?: string;   
+  className?: string;
+  /** Set heading level to the chip item label */
```
mturley: Looks like you left a comment behind here :)

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -44,7 +48,8 @@ export class ChipGroup extends React.Component<ChipGroupProps, ChipGroupState>{
 
   renderToolbarGroup() {
     const { isOpen } = this.state;
-    return <InnerChipGroup {...this.props} isOpen={isOpen} onToggleCollapse={this.toggleCollapse} />;
+    const { headingLevel = 'h4' } = this.props;
+    return <ChipGroupContext.Provider value={headingLevel}><InnerChipGroup {...this.props} isOpen={isOpen} onToggleCollapse={this.toggleCollapse} /></ChipGroupContext.Provider>;
```
mturley: We really need to turn our linter back on 😄  I think this would be more readable as multiple lines, like:

```js
return (
  <ChipGroupContext.Provider value={headingLevel}>
    <InnerChipGroup {...this.props} isOpen={isOpen} onToggleCollapse={this.toggleCollapse} />
  </ChipGroupContext.Provider>
);
```

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -18,14 +20,20 @@ export const ChipGroupToolbarItem: React.FunctionComponent<ChipGroupToolbarItemP
...
-          <li>
-            <h4 className={css(styles.chipGroupLabel)}>{categoryName}</h4>
-            <ul className={css(styles.chipGroup)}>{children}</ul>
-          </li>
-        </ul>
+      return (
+        <ChipGroupContext.Consumer>
+            {(HeadingLevel: any) => {
+            return (
+            <ul className={css(styles.chipGroup, styles.modifiers.toolbar)} {...props}>
```
mturley: I might be getting too nit-picky here but without the linter on I get nervous about code style. You've got 4 spaces of indentation on `{(HeadingLevel: any) => {` and no indentation on `return` or `<ul className=...`. Do you mind fixing it so each has 2 spaces of indendation?

```js
        <ChipGroupContext.Consumer>
          {(HeadingLevel: any) => {
            return (
              <ul className={css(styles.chipGroup)}>{children}</ul>
                ...
              </ul>
            )
          }}
        </ChipGroupContext.Consumer>
```

Sorry if this is overkill.

boaz0:
@mturley thanks a lot! Would you mind reviewing this again?

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -18,14 +19,20 @@ export const ChipGroupToolbarItem: React.FunctionComponent<ChipGroupToolbarItemP
...
-      return(
-        <ul className={css(styles.chipGroup, styles.modifiers.toolbar)} {...props}>
-          <li>
-            <h4 className={css(styles.chipGroupLabel)}>{categoryName}</h4>
-            <ul className={css(styles.chipGroup)}>{children}</ul>
-          </li>
-        </ul>
+      return (
+        <ChipGroupContext.Consumer>
+          {(HeadingLevel: any) => {
```
jschuler: it was passed in lowercase, this works? also i looked at the built docs and it doesn't show `headingLevel` prop for ChipGroup, but rather under ChipGroupToolbarItem. Maybe it wasn't rebuilt?

jschuler: Ah just checked, the value is passed directly and then picked up here. So that's good. and the docs looks fine as well when i built it locally, LGTM

boaz0: Hi @jschuler - good questions:

You are correct if chip group context is an object containing the `headingLevel` attribute but since the chip group context is only headingLevel then it's OK that the argument will be `HeadingLevel`.

If you want I can change the context to be object:

```
type ChipGroupCtx = {
  headingLevel: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
}

ChipGroupContext = React.React.createContext<Partial<ChipGroupCtx>>({});
```

And in here I will have to change it to this:

```
{({headingLevel: HeadingLevel}) => (...)
```

Regarding to the props that aren't shown in the docs, I can't look at the CI but I assume it built the docs and uploaded but probably the URL was changed and for some reason it is not shown in here.

Anyway, let me know if I'm missing anything and thanks for your review!

boaz0:
Thank you all
@jgiardino @mturley @jessiehuff @jschuler @tlabaj

PR #2286, opened 2019-06-17
feat( topology-sidebar): Adds TopologySidebar to react-topology

jeff-phillips-18:
**What**:
Adds a TopologySideBar component and the ability to add the side bar to the TopologyVIew


jeff-phillips-18:
@dgutride @mceledonia ping!

Issue #2028, opened 2019-05-17
Radio isChecked changed uncontrolled to controlled


willr3:
Using isChecked={...} results in an error in the Chrome console when changing the selected radio. 
Using checked={...} does not result in the error
Chrome Version 73.0.3683.103 (Official Build) (64-bit)

```javascript
import React from 'react';
import {Radio} from '@patternfly/react-core';
export default class RadioTest extends React.PureComponent{
    constructor(props){
        super(props)
        this.state={ value:"one" }
        this.update=this.update.bind(this)
    }
    update(_,event){
        const { value } = event.currentTarget;
        this.setState({ value })
    }
    render(){
        const {value} = this.state
        return (
            <div>
                <Radio
                    value="one"
                    isChecked={value === "one"}
                    name="test"
                    onChange={this.update}
                    label="One"
                    id="one"
                />
                <Radio
                    value="two"
                    isChecked={value === "two"}
                    name="test"
                    onChange={this.update}
                    label="Two"
                    id="two"
                />
            </div>
        )
    }
}
```
```javascript
{
    "react": "^16.8.6",
    "@patternfly/react-core": "^3.16.10",
}
```
Warning: A component is changing an uncontrolled input of type radio to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (created by Radio)
    in div (created by Radio)
    in Radio (at RadioTest.js:17)
    in div (at RadioTest.js:16)
    in RadioTest (at src/index.js:21)
    ...

rachael-phillips:
@dlabaj making you aware

PR #2292, opened 2019-06-18
fix(Radio): fix warning from undefined starting params

kmcfaul:
**What**: Fixes warning re: controlled/uncontrolled switch. Also fixes warning for missing prop on uncontrolled radio.

**Refer to issue**: #2028

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
+++ b/packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
@@ -3,7 +3,8 @@ import styles from '@patternfly/react-styles/css/components/Radio/radio';
 import { css, getModifier } from '@patternfly/react-styles';
 import { Omit } from '../../helpers/typeUtils';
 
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -60,17 +62,20 @@ export class Radio extends React.Component<RadioProps> {
...
           {...props}
-          aria-label={label ? undefined : ariaLabel}
           className={css(styles.radioInput)}
           type="radio"
           onChange={this.handleChange}
           aria-invalid={!isValid}
           disabled={isDisabled}
           checked={checked || isChecked}
-          defaultChecked={isChecked ? undefined : defaultChecked}
+          {...!isChecked && { defaultChecked }}
```
jschuler: `defaultChecked` is an attribute passed to input? what does it do?

kmcfaul: I think it's required for the uncontrolled version of radio. The undefined was throwing a console error.

Issue #2254, opened 2019-06-13
Support inverted thresholds in ChartDonutUtilization


TheRealJon:
Currently, thresholds only work in a single direction. For representing something like resource utilization (where high data values are undesirable), this works fine. But, for charts that track something like success rates (where high values are desirable), the thresholds should work in the opposite direction. It is possible to implement this functionality downstream using the existing component but requires quite a bit of logic. We should accept an `invert` prop into the `ChartDonutUtilization` and `ChartDonutThreshold` components that will take care of inverting the threshold logic.

**Example**: A chart is used to track API response success rates. If the success rate falls below 85%, the chart should change to a 'warning' state. If the rate falls below 70% it should change to a 'danger' state. The consumer should be able to set thresholds at 15% and 30% and use the `invert` prop to achieve this behavior.

dlabrecq:
Can you please share some snapshots of your implementation?

dlabrecq:
With a couple tweaks to donut utilization chart, I ended up with this:

![chrome-capture](https://user-images.githubusercontent.com/17481322/59708276-6c70c400-91d2-11e9-86ea-874acefcb902.gif)

dlabrecq:
And donut threshold:

![chrome-capture](https://user-images.githubusercontent.com/17481322/59708582-0e90ac00-91d3-11e9-8882-c44e88ccb8df.gif)

gdoyle1:
Mocks for this @dlabrecq @TheRealJon  https://marvelapp.com/c284847/screen/58299421

PR #2294, opened 2019-06-18
donut-charts: add invert prop for reversing thresholds

dlabrecq:
This adds an invert property to the donut utilization and threshold charts, which inverts the threshold color scale used to represent warnings, errors, etc.

Instead of showing a warning at 60% and an error at 90%; for example, this would allow users to show a warning below 60% and an error below 20%.

fixes https://github.com/patternfly/patternfly-react/issues/2254

chrome-capture2

chrome-capture

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
+++ b/packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
@@ -16,6 +16,7 @@ import { ChartDonut, ChartDonutProps } from "../ChartDonut/ChartDonut";
 import { ChartThemeDefinition, ChartDonutUtilizationStaticTheme } from '../ChartTheme/ChartTheme';
 import { getDonutUtilizationTheme } from '../ChartUtils/chart-theme';
 import { DonutUtilizationStyles } from '../ChartTheme/themes/donut-utilization-theme';
+import { cloneDeep } from 'lodash';
 
 export enum ChartDonutUtilizationLabelPosition {
   centroid = 'centroid',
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
```diff
@@ -541,14 +549,20 @@ export const ChartDonutUtilization: React.FunctionComponent<ChartDonutUtilizatio
     if (data) {
       const datum = getData();
       const donutThresholds = getDonutThresholds();
+      const mergeThemeProps = (i: number) => {
+        // Merge just the first color of dynamic (blue, green, etc.) with static (gray) for expected colorScale
+        newTheme.pie.colorScale[0] = donutThresholds[i].color;
+        newTheme.legend.colorScale[0] = donutThresholds[i].color;
+      };
       for (let i = 0; i < donutThresholds.length; i++) {
```
TheRealJon: I think we'll have to enforce sorting on the thresholds array. They need to be sorted in ascending order by the value prop when `inverted === false`, and descending when `inverted === true`.

It's possible to pass a `thresholds` prop in an unsorted or reverse sorted order, which would break this function.

For example:
```js
[
  {
    value: 90, 
    color: 'red'
  }, 
  {
    value: 60, 
    color: 'yellow'
  }
]
```

If a value is greater than 90, we would never get the 'red' color because we would iterate past the `value: 90` threshold, see that the data value is greater than 60 as well, and colorScale[0] would be set to yellow.

TheRealJon: I'll open a separate issue for this and it can be fixed as a follow on since it was an existing issue before this PR.

Issue #2202, opened 2019-06-10
Categories and icons for the app launcher component


LHinson:
The app launcher needs the ability to have categories as well as support icons. 

Related to core issue here: https://github.com/patternfly/patternfly-next/issues/1914

Issue #2084, opened 2019-05-24
Using keyboard navigation to open a Application Launcher with Tooltip'd items fails


jeff-phillips-18:
Console error:

> typeError: Cannot read property 'focus' of undefined
>     at ApplicationLauncherMenu.componentDidMount
> 

Documentation page shows:
![image](https://user-images.githubusercontent.com/11633780/58343408-e35bac80-7e20-11e9-97c5-f1b03e4bc586.png)

tlabaj:
Closed by #2297

PR #2297, opened 2019-06-18
PF4: Application launcher enhancements (groups, icons)

jschuler:
Closes: https://github.com/patternfly/patternfly-react/issues/2202, https://github.com/patternfly/patternfly-react/issues/2084

Update Application launcher to bring in recent core features such as groups and icons.
Made corresponding updates to Dropdown component to enable such. There should be no breaking changes.

One confusing thing about Dropdown component before was that the `Item` component was the one that was exported out as `DropdownItem`, while the `Dropdown` component was actually just an internal component used in the `Item` component. I switched those around and renamed the previous `DropdownItem` to `InternalDropdownItem`.


packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.d.ts
```diff
@@ -1,27 +1,9 @@
...
-  index?: number; 
-  context?: {
-    keyHandler?: (index: number, direction: string) => void,
-    sendRef?: (index: number, ref: any, isDisabled: boolean) => void
-  };
-  /** Callback for click event */
-  onClick?: () => void; 
-}
\ No newline at end of file
+export interface DropdownItemProps extends InternalDropdownItemProps {
```
jschuler: This is confusing, but the contents in this file got replaced with the previous Item.d.ts contents (which was then deleted), and the previous Item.d.ts contents got moved into InternalDropdownItem.d.ts

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js
```diff
@@ -1,169 +1,44 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
-import dropdownStyles from '@patternfly/react-styles/css/components/Dropdown/dropdown';
-import appLauncherStyles from '@patternfly/react-styles/css/components/AppLauncher/app-launcher';
-import { css } from '@patternfly/react-styles';
+import InternalDropdownItem from './InternalDropdownItem';
```
jschuler: This is confusing, but the contents in this file got replaced with the previous Item.js contents (which was then deleted), and the previous Item.js contents got moved into InternalDropdownItem.js

jschuler:
@jgiardino for a11y, and @christiemolloy for css

jschuler:
@tlabaj updated description

packages/patternfly-4/react-core/src/components/ApplicationLauncher/InternalApplicationLauncherText.tsx
```diff
@@ -0,0 +1,19 @@
...
+  /** Additional classes added to the text container */
+  className?: string;
+}
+
+export const ApplicationLauncherText: React.FunctionComponent<ApplicationLauncherTextProps> = ({
+  className = '',
+  children,
+  ...props
+}: ApplicationLauncherTextProps) => (
+  <span className={css('pf-c-app-launcher__menu-item-text', className)} {...props}>
```
karelhala: Needs to use `style` object `css(styles.appLauncherMenuItemText)`

packages/patternfly-4/react-core/src/components/Dropdown/dropdownConstants.d.ts
```diff
@@ -16,7 +16,13 @@ export const DropdownContext: React.Context<{
   menuClass?: string,
   itemClass?: string,
   toggleClass?: string,
-  baseClass?: string
+  baseClass?: string,
+  baseComponent?: string,
+  sectionClass?: string,
+  sectionTitleClass?: string,
```
karelhala: Can you also add `sectionComponent` to this? It usually should be section but some components might use in future something else.

packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.js
```diff
@@ -0,0 +1,173 @@
...
+          }
+          return (
+            <li role="none">
+              {React.isValidElement(children) ? (
+                React.Children.map(children, child => {
+                  const clonedElement = React.cloneElement(child, {
+                    className: css(
+                      isDisabled && disabledClass,
+                      isHovered && hoverClass,
+                      className,
```
jeff-phillips-18: className should be on the `li` not on each of the children - `Classes applied to root element of dropdown item`

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherGroup.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { DropdownGroup, DropdownGroupProps } from '../Dropdown';
+
+export const ApplicationLauncherGroup: React.FunctionComponent<DropdownGroupProps> = ({
```
jeff-phillips-18: DropdownGroup only supports a string for a label. I could see where an icon and string might be desirable in the Application Launcher groups.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/InternalApplicationLauncherText.tsx
jschuler: This class does not actually exist (yet) but it was applied in the core example. @mcoker , should I keep it around or remove it for now?

packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.js
jschuler: I agree with you, but this is a carry over from the current Dropdown code
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.js#L130
technically it could be considered a breaking change to move the className on the parent, although a pretty minor one. @dgutride should this wait for a breaking change release or ok to do here?

packages/patternfly-4/react-core/src/components/ApplicationLauncher/InternalApplicationLauncherText.tsx
mcoker: @jschuler is there a reason we shouldn't keep it? Since it isn't being used, I can always just remove it and add it back when it's needed.

jschuler: i think it's fine to keep it around for now

jgiardino:
Overall, it looks great! The keyboard behavior works as expected. There is some inconsistent behavior I noticed when testing in VO, which I noted below. I can review in JAWS and NVDA later today.

In the first 2 examples, when I use VO, I'm only able to use the Up/Down arrow keys to navigate the menu. If I use the VO command for navigating to the next element, it doesn't shift focus to the next menu item, but instead dismisses the menu and returns me to the toggle. This isn't the behavior with the third example. In that case, if I use the VO command for navigating to the next element, it will traverse the menu items. And at the end/beginning of the menu panel, I'm able to continue shifting focus in this manner, and leave the panel to the adjacent element in the DOM, while the panel remains open. 

I'm not sure that being able to navigate with the VO command for next element is necessarily bad, but I'd expect this behavior to be consistent. And I definitely expect focus to be trapped in the menu for the screen reader user.

Do you see any other differences that could be affecting this? Is there something enabled in the other examples that are managing or trapping focus that aren't included here?

Something I noticed in the HTML is that we now have additional elements than the original examples between the element with `role="menu"` and the element with `role="menuitem"`. I'm not sure that this is affecting the behavior or not, but this would be easy to test again to confirm. The proper role assignments would be to include: 
- `role="group"` on `<section>` elements (this might be redundant, but I've noticed issues when assigning roles for some elements and not all, even when the implicit role of an element is what you want)
- `role="none"` on the `<ul>` elements

**However,** these updates are relative to our current implementation. One of the updates I'd like to see us make with https://github.com/patternfly/patternfly-react/issues/1615 is that we adopt an interaction pattern that is more like a modal and less like the [menu button](https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton) pattern. The menu button pattern could be applicable for some components for specific use cases, as another variant, but not for AppLauncher. The modal pattern is much better for most of the use cases that these types of toggle/menu components need to support.

jschuler:
@christiemolloy which example are you referring to, the last one? Where did you expect to see the tooltip? How did you interact, using the mouse or arrow keys?

jgiardino:
I just tested the examples in NVDA and JAWS. The behavior in those screen readers is consistent across all examples. This makes me wonder if the `role` attribute is what is creating the difference in VO, since I've noticed VO behave differently based on role assignment, or lack of. Let me know when you get the roles updated, and I can test again in VO.

jeff-phillips-18:
Is it possible to iterate on the accessibility requirements or would it be breaking changes to meet them later?

jgiardino:
> Is it possible to iterate on the accessibility requirements or would it be breaking changes to meet them later?

👍 I'd say at a minimum to address the `role` attributes that I mentioned in my first comment. And if nothing else jumps out as obviously different in the implementation, then let's not block this PR with that one issue in VO.

christiemolloy:
@jschuler sorry I should have specified, I was referring to the second example where it says "with tooltip" so I'm wondering where the tooltip is, and also there is a focus state around the text in the dropdown as well as (I think) a focus state around the dropdown item

jgiardino:
The role updates look great. 

Did the items with `role="menuitem"` change from `<a>` to `<div>` or did I just miss this in my first review? I'm assuming these would always be links.

jschuler:
@jgiardino they can be anchor elements but don't have to be. if consumer wants to use a react router item in the nav it should be rendered into a `div` and not an `a`. In the example you'll see both

mcoker:
Looks great overall.

* The divider between groups isn't in a group in core, it's an `<hr>` used between groups. I think either way (as an `hr` between groups or an `li` (or in this case an `li > div`) in a group) creates the same amount of visual space, so it's OK, but is there a use case for wanting to define a separator between groups instead of in a group's menu items? Or should we stick with one type of separator placed in a list of items? 
* The separator here and in the dropdown is `<div href="#"...`. Should we remove `href`?
* Are you adding an inline style of `cursor: pointer` when the item is a `<div>`? Ideally I think we would add that in core, maybe as part of https://github.com/patternfly/patternfly-next/issues/1944

jgiardino:
Thanks for the clarification, @jschuler!

jschuler:
@mcoker 

- I agree, the reason the rule is in the group is because it's the current mechanism for Dropdown (App Launcher is using dropdown under the hood). Changing it to be between groups would break the keyboard interaction at the moment. Perhaps something we can follow up on?

- Yes i can remove the href from the Separator

- I agree it should be added in core which is why I opened the issue you linked :) once it's fixed we can remove it here. I added a comment to the code that mentions this

jschuler:
@mcoker opened #2316 and #2317 for follow-up

packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.js
tlabaj: @dgutride ^^

dgutride: if the api surface is removing anything on an existing component - we can't do it right now

PR #2304, opened 2019-06-19
feat(react-core): add ouia compatibility to Switch

### LINKED ISSUE #2303: Add Custom Data-Attributes to Components for OUIA Compatibility
Custom `data-` attributes on our components can help QE teams with integration tests. One suggestion is [OUIA](https://forum.patternfly.org/t/open-web-ui-design-specification-for-enabling-automation-ouia/154). Investigate how to include such attributes behind a Local Storage flag (for now).

This runtime could be useful for consumers who wish to run their own integration tests that locate PF4 components by `data-` attributes.

On the other hand, this is a violation of KISS in our main `dist` where dead code for these extraneous attributes will exist for a runtime most consumers never use and certainly won't want to ship in production. Webpack will not tree-shake the code because it has no idea what `localStorage` will be at runtime.

We could release a separate `react-core-test` package with these attributes for every major release, or ship a [new module type](https://auth0.com/blog/javascript-module-systems-showdown/) like `amd` and require integration tests use that unpopular module type that doesn't include the `data-` attributes.

Open to discussion.

redallen:
**What**: Fixes #2303 for a Switch by conditionally adding `data-ouia-component-type` and `data-ouia-component-id` for Switch based on `window.localStorage['ouia']`. ~Do not merge until some further discussion takes place on #2303 .~

Edit: Discussion has taken place, we can merge now.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
@@ -4,6 +4,7 @@ import { css } from '@patternfly/react-styles';
 import { CheckIcon } from '@patternfly/react-icons';
 import { getUniqueId } from '../../helpers/util';
 import { Omit } from '../../helpers/typeUtils';
+import { isOUIAEnvironment, getUniqueId as getOIUAUniqueId } from '../../helpers/ouia';
 
 export interface SwitchProps extends Omit<React.HTMLProps<HTMLInputElement>, 'type' | 'onChange' | 'disabled' | 'label'> {
   /** id for the label. */
@@ -24,6 +25,7 @@ export interface SwitchProps extends Omit<React.HTMLProps<HTMLInputElement>, 'ty
 
 export class Switch extends React.Component<SwitchProps> {
   id = '';
+  ouiaId = getOIUAUniqueId();
 
   static defaultProps = {
     id: '',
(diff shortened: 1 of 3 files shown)


quarckster:
thanks @redallen I couldn't test changes of this PR in https://patternfly-react-pr-2304.surge.sh/. I set a localStorage variable `window.localStorage.setItem("ouia", true)`, refreshed the page but nothing changed in the DOM. I couldn't find `data-ouia` attributes. On the other hand, it works if you build PR locally, it's tested by @karelhala

redallen:
@quarckster Gatsby statically renders the site without `window` defined, so after you set the localStorage variable you have to interact with the switch to cause it to re-render, then you'll see the tags. Sorry, that's a caveat I forgot to mention.

redallen:
After discussion with the OUIA team, this PR is good to merge and have many more followups for all our components that would want integration testing.

redallen:
@dgutride @karelhala Fixed merge conflicts, please re-approve!

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -4,6 +4,7 @@ import { css } from '@patternfly/react-styles';
 import { CheckIcon } from '@patternfly/react-icons';
 import { getUniqueId } from '../../helpers/util';
 import { Omit } from '../../helpers/typeUtils';
+import { isOUIAEnvironment, getUniqueId as getOIUAUniqueId } from '../../helpers/ouia';
```
jschuler: nit
```suggestion
import { isOUIAEnvironment, getUniqueId as getOUIAUniqueId } from '../../helpers/ouia';
```

redallen: Fixed.

PR #2305, opened 2019-06-19
fix(Grid): Update grid layout to support new 2xl breakpoint

### LINKED ISSUE #2199: Update grid layout to support new 2xl breakpoint
Update the grid layout to support the 2xl breakpoint added here - https://github.com/patternfly/patternfly-next/pull/1880

jessiehuff:
Fixes #2199

Update the grid layout to support the 2xl breakpoint added here - patternfly/patternfly-next#1880


packages/patternfly-4/react-core/src/layouts/Grid/Grid.js
```diff
@@ -25,6 +25,8 @@ const propTypes = {
   lg: gridItemSpanValueShape,
   /** the number of columns all grid items should span on a xLarge device */
   xl: gridItemSpanValueShape,
+  /** the number of columns all grid items should span on a 2xLarge device */
+  xl2: gridItemSpanValueShape,
```
mcoker: should we call it `2xl` or is there a reason it's `xl2` here?

jessiehuff: I had to name it xl2 to meet JSX attribute name requirements. It won't allow a prop to start with a number. :(

boaz0: Maybe `'2xl': gridItemSpanValueShape` will make JSX happy?

jessiehuff: That works for defining the prop, but it doesn't work when actually creating a `<GridItem 2xl={2}>`

Issue #2019, opened 2019-05-15
PF4: Convert Switch to TypeScript


tlabaj:
Convert the Switch component to Typescript

PR #2311, opened 2019-06-20
chore(Switch): convert Switch to TypeScript

tlabaj:
#2019

**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
+import { HTMLProps, FormEvent } from 'react';
```
redallen: Don't import these and instead use `React.HTMLProps` and `React.FormEvent` to be consistent with other components.

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
...
+  label?: string,
   /** Flag to show if the Switch is checked. */
-  isChecked: PropTypes.bool,
+  isChecked?: boolean,
   /** Flag to show if the Switch is disabled. */
-  isDisabled: PropTypes.bool,
+  isDisabled?: boolean,
   /** A callback for when the Switch selection changes. (isChecked, event) => {} */
-  onChange: PropTypes.func,
+  onChange?(checked: boolean, event: FormEvent<HTMLInputElement>): void;
```
redallen: Use an arrow function so the docs pick this up properly. Like `onChange?: (checked: boolean, event: FormEvent<HTMLInputElement>) => void;

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -77,7 +80,4 @@ class Switch extends React.Component {
   }
 }
 
-Switch.propTypes = propTypes;
-Switch.defaultProps = defaultProps;
-
 export default Switch;
```
redallen: Remove default export.

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
...
-  id: '',
-  className: '',
-  label: '',
-  isChecked: true,
-  isDisabled: false,
-  onChange: () => undefined,
-  'aria-label': ''
-};
+export class Switch extends React.Component<SwitchProps> {
+  id = '';
```
jessiehuff: Remove this class attribute

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
...
+
+  static defaultProps = {
+    id: '',
```
jessiehuff: Change to: `id: getUniqueId()`

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
...
 
-class Switch extends React.Component {
-  id = this.props.id || getUniqueId();
+  constructor(props: SwitchProps) {
+    super(props);
+    if (!props.id && !props['aria-label']) {
+      // tslint:disable-next-line:no-console
+      console.error('Switch: Switch requires either an id or aria-label to be specified');
+    }
+    this.id =props.id || getUniqueId();
```
jessiehuff: This can be removed.

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
...
+  }
 
   render() {
```
jessiehuff: Use this.props.id instead

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -1,49 +1,52 @@
...
+import * as React from 'react';
+import { HTMLProps, FormEvent } from 'react';
 import styles from '@patternfly/react-styles/css/components/Switch/switch';
 import { css } from '@patternfly/react-styles';
 import { CheckIcon } from '@patternfly/react-icons';
 import { getUniqueId } from '../../helpers/util';
+import { Omit } from '../../helpers/typeUtils';
 
-const propTypes = {
+export interface SwitchProps extends Omit<HTMLProps<HTMLInputElement>, 'type' | 'onChange' | 'disabled' | 'label'> {
```
redallen: Also just noticed we don't have to omit `type` or `label`. I like omitting `disabled` in favor of `isDisabled`.

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
dlabaj: Not sure what the purpose of this id is ... is it suppose to be the same as props.id?

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
dlabaj: Seems like the prop.id and this id aren't guaranteed to be the same value ... what's the purpose of the class attribute?

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
tlabaj: it instantiates the id when it is generated. And there is no id prop set.

redallen: Shouldn't it instantiate it as a defaultProp to `getUniqueId()`?

tlabaj: I am not sure.  Does it work differently in TS than JS.  In JS if you set defaultProp to getUniqueId, if you had more  hat one Switch on the page the id's would all be the same?  That is why we had it in the constructor before.

dlabaj: Looks like it will set it for all of the name to be the same based off the time.  Leave it in the constructor.

PR #2313, opened 2019-06-20
fix(virtual-scroll): add columnCount prop to VirtualTableBody

priley86:
**What**:

* adds the `columnCount` prop to `VirtualTableBody`. This will allow fewer renderings when the consumer is virtualizing by row (and using a `rowRenderer` instead of a `cellRenderer`). This is just an 
optional way to implement the `Grid` for the consumer by passing `rows` instead of `columns`.

* fix virtual scroll w/ `WindowScroller` example. This will now use `isScrollingOptOut` for more accurate scrolling. See React-Virtualized issue [638](https://github.com/bvaughn/react-virtualized/issues/638). 

* remove the `pf-l-grid` class. This was actually not needed before (i misinterpreted our CSS and thought that the  `pf-m-[x]-col-on-[breakpoint]` classes were scoped to `pf-l-grid`, however they are completely separate).

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualGrid.ts
+++ b/packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualGrid.ts
@@ -29,7 +29,6 @@ import {
   CellCache,
   StyleCache
 } from './types';
-import { overflow } from 'styled-system';
 /**
  * Specifies the number of milliseconds during which to disable pointer events while a scroll is in progress.
  * This improves performance and makes scrolling smoother.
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualGrid.ts
```diff
@@ -29,7 +29,6 @@ import {
   CellCache,
   StyleCache
 } from './types';
-import { overflow } from 'styled-system';
```
priley86: unused reference

Issue #2275, opened 2019-06-14
Popover close button target area isn't big enough


seanforyou23:
In the react version of Popover, the size of the target area for the close button is too small. While the button seems to be the correct dimensions, the clickable area is much smaller, seems to be roughly the size of the svg inside it. This presents an accessibility issue for those with motor control limitations, and also just seems buggy in general.

Steps to reproduce:

- Launch the examples at http://patternfly-react.surge.sh/patternfly-4/components/popover/
- Hover over the close "x" button
- Notice the cursor flickers when crossing the bounds of svg/button and the area outside the svg isn't clickable

![2019-06-17 09 07 44](https://user-images.githubusercontent.com/5942899/59606599-909fa880-90df-11e9-951d-b90f258878fd.gif)

Expected result: The entire area within the close button is clickable.

The core implementation seems to be ok https://pf4.patternfly.org/components/Popover/examples/

seanforyou23:
Looks like the problem was introduced with https://github.com/patternfly/patternfly-react/pull/1757

Tippy supports an `interactive` property out of the box, which allows for highlighting and selecting text, so I'm not sure why we needed to apply pointer-events in the first place. Will keep investigating, just wanted to share finding so far.

rachael-phillips:
Thanks for filing this @seanforyou23 !

rachael-phillips:
@seanforyou23 are you working on this one currently?

PR #2240, opened 2019-06-12
Converted Page component to Typescript

dtaylor113:
- Converted Page and it's sub-components to Typescript
- Updated unit tests
- Created integration tests

Fixes #2234

PR #2216, opened 2019-06-11
Convert Tooltip component to Typescript

dtaylor113:
Convert Tooltip component to Typescript. 
Update unit and system integration tests.

Addresses #2212

PR #2326, opened 2019-06-21
feat(tippy): upgrade tippy deps and apply api changes

seanforyou23:
**What**: This PR upgrades tippy, which is the backing lib for Tooltip and Popover. It's a follow up to a [recent patch](https://github.com/patternfly/patternfly-react/pull/2284/files#diff-e42107b38e53954fa7c6c9cee3846e7aR9) I sent to allow the Popover close button to capture mouse events properly.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/2275 and https://github.com/patternfly/patternfly-react/issues/2333

Removed the [inline tippy styles](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Tooltip/styles.tsx#L5) as they are now automatically injected by tippy js. Converted `.pf-tippy-theme` class to simply `.pf-tippy` class. Added `tabIndex={-1}` to FocusTrap child element to allow focus trap to continue working amid the new pointer-events that come with tippy2.x.

We now wrap [Tippy children](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx#L284) in a span to ensure refs are forwarded correctly so the auto close features work as expected. An alternate approach would be to require users to manually forwardRefs themselves but this would entail a breaking change, so since the span doesn't introduce any styling complications I think it's the way to go.

I also added some new examples that illustrate using links both inside of popover content area, and also as the base element for a normal tooltip. Ensuring pointer-events are captured correctly in all the various cases proved tricky, so hopefully, these new examples will help in future regression testing.

One change in behavior was for the "Advanced popover" scenario, where you want to have manual control over how the popover can open/close. Previously there was what looks like a bug where clicking outside the popover in the document body would also close the popover. After speaking with some other devs, I don't think this was the desired behavior. So now, if you manually control the popover visibility by setting the `isVisible` prop, you'll not get the auto close behavior as you've taken explicit control over showing the component. Auto close by clicking on the document body is default behavior of Popover so if you want this, just don't explicitly set `isVisible`.

There are new props for Tippy (`aria` and `boundary`, among others) that I'd like to make configurable, but that work may be better done as a dedicated PR so we can update examples/docs for these considerations specifically. For now, I've left them at the default values.

The release notes from https://github.com/atomiks/tippyjs/releases/tag/v4.0.0 and https://github.com/atomiks/tippy.js-react/releases/tag/v2.0.0 are both relevant.


dtaylor113:
LGTM, only issue I see is that top and bottom popover arrows seem to be too close and overlapping the content:
image

Previously:
image

dtaylor113:
Also, I'm not sure if this was an intentional change, but in the new Popover examples, clicking outside of Popover no longer dismisses it.

seanforyou23:
Good catch @dtaylor113 - working on fixing those styles now, will push some updates shortly. Looks like it's just for top/bottom placement that we get the styling issue.

As for clicking outside popover to close, this was an intentional change. You can still get auto-close effect by using the simple popover case where events are handled by the lib. If you want to manually open/close the popover, you'll need to write event handlers to update the state to control the visibility manually (even for the case of clicking outside the popover).

I noticed also there was a bug around this in the previous demo. To see it in action, launch the advanced popover (on master) and then click outside to close it and notice the state property isn't updated, so a subsequent click of the button toggle doesn't open the popover as one would expect. You'd have to click it again to make it work.

Now things are a little more explicit, which I think is what we'd want. Let me know if you see any concern.

dtaylor113:
Hi @seanforyou23, would it be possible to hold off merging this PR until https://github.com/patternfly/patternfly-react/pull/2216 is merged?  I've had to resolve several merge conflicts so far, and I fear this PR would be a big one to have to resolve -thanks

seanforyou23:
I don't mind @dtaylor113 - I've been beatin up by many merge conflicts on my end lately also :) I don't mind helping apply the updates for your TS upgrade PR in case this one gets merged first. Most of the churn was in Popover FWIW, Tooltip was relatively unchanged, a lot of the diff you see for it was me putting the props in alphabetical order so they are easy to compare with Popover props.

dtaylor113:
Thanks @seanforyou23, appreciate it.  I'm trying to get this and https://github.com/patternfly/patternfly-react/pull/2240 merged so I can move onto other OpenShift tasks.  This was supposed to be a quick chore! :-)

PR #2363, opened 2019-06-26
feat(experimental): add POC for experimental button and badge

### LINKED ISSUE #2300: Add support for experimental react components
Once https://github.com/patternfly/patternfly-next/issues/1950 is done - we need to do the same in react.  Experimental components would not be exported from the main index file - but would be in a separate experimental export 

The following should be used for imports
import {experimentalComponent} from '@patternfly/react-core/experimental';  

The documentation should also be updated to create a space for these and a clear way to notate these are experimental as well.

redallen:
**What**: Add `dist/esm/expiremental`, which is not exported from the main `index.js`. Add support for the docs to show them. Towards https://github.com/patternfly/patternfly-org/issues/1247 .

**Additional issues**: Fixes #2300

### DIFF
--- a/packages/patternfly-4/react-core/src/experimental/components/Badge/Badge.md
+++ b/packages/patternfly-4/react-core/src/experimental/components/Badge/Badge.md
@@ -6,11 +6,19 @@ propComponents: ['Badge']
 section: 'experimental'
 ---
 import { Badge as ExperimentalBadge } from '@patternfly/react-core/dist/esm/experimental';
+import { Alert } from '@patternfly/react-core';
+
+<Alert variant="danger" title="Warning">
+  Please don't use this component, it's only an example of what an experimental component could be
+  and likely going away in our next release.
+  Use our <a href="../../components/badge">actual Badge instead.</a>
+</Alert>
+<br />
 
 ## Badge
 ```js
 import React from 'react';
-import { Badge as ExperimentalBadge } from '@patternfly/react-core';
+import { Badge as ExperimentalBadge } from '@patternfly/react-core/dist/esm/experimental';
 
 ReadBadge = () => (
   <React.Fragment>
--- a/packages/patternfly-4/react-core/src/experimental/components/Badge/Badge.md
+++ b/packages/patternfly-4/react-core/src/experimental/components/Badge/Badge.md
@@ -1,0 +1,24 @@
+---
+title: 'Badge'
+cssPrefix: 'pf-c-badge'
+typescript: true 
+propComponents: ['Badge']
+section: 'experimental'
+---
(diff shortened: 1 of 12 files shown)


packages/patternfly-4/react-core/src/expiremental/components/Badge/Badge.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { Badge } from './Badge';
+import React from 'react';
+import { shallow } from 'enzyme';
+
+Object.values([true, false]).forEach(isRead => {
+  test(`${isRead} Badge`, () => {
+    const view = shallow(<Badge isRead={isRead}>{isRead ? 'read' : 'unread'} Badge</Badge>);
+    expect(view).toMatchSnapshot();
```
jeff-phillips-18: I'd really like to see test that check for expected values rather than relying on snapshots. That is, adding a check for the existance of the `pf-m-unread` when isRead is false and for `pf-m-read` when isRead is true.

redallen: These components will likely be removed, they are just to test the site for now. I agree the tests should cover the experimental features.

redallen:
The relevant issue I added the the OP, it's https://github.com/patternfly/patternfly-org/issues/1247 . We'd like for these components to be available as a separate export and be subject to frequent breaking changes or removal.

packages/patternfly-4/react-core/src/experimental/components/Badge/Badge.md
```diff
@@ -0,0 +1,24 @@
...
+typescript: true 
+propComponents: ['Badge']
+section: 'experimental'
+---
+import { Badge as ExperimentalBadge } from '@patternfly/react-core/dist/esm/experimental';
+
+## Badge
+```js
+import React from 'react';
+import { Badge as ExperimentalBadge } from '@patternfly/react-core';
```
jschuler: i know it's just text but make it the same as line 8?

packages/patternfly-4/react-core/src/experimental/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,20 @@
...
+  testThing?: boolean;
+}
+
+export const Badge: React.FunctionComponent<BadgeProps> = ({ 
+  isRead = false, 
+  className = '', 
+  children = '', 
+  ...props }: BadgeProps) => (
+  <span {...props} className={css(styles.badge, (isRead ? styles.modifiers.read : styles.modifiers.unread) as any, className)}>
+    Im expiremental!
```
jschuler: 😄 
```suggestion
    I'm experimental!
```

Issue #2356, opened 2019-06-26
Donut chart main label is misleading due to round up


ares:
Imagine you draw a donut chart for a fireman operator dashboard. It displays how many building in a city are on fire and how many are ok. Let's say there are 1000 buildings in the city. Now first starts burning, so we have 999 ok (green), 1 on fire (red). The donut chart will show only green color, and only after hovering over it, it says "99.9%", which is one problem, there's no visual indication of fire.

The second and the bigger one is, the main label says "100% ok" because it rounds up, I guess thanks to this line https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/src/components/Chart/DonutChart.js#L24.

Poor fireman operator won't notice from the dashboard that one building is on fire, until the fire spreads to few more buildings.

rachael-phillips:
@dlabrecq making you aware of this one

rachael-phillips:
@ares is this blocking you at this time?

dlabrecq:
@rachael-phillips FYI, this is an PF3 issue

ares:
It's not blocking but the bug reported by customer has a big priority. @amirfefer will send a PR with the fix today I believe. And sorry for confusion, yes this is PF3 issue.

ares:
The suggested fix - https://github.com/patternfly/patternfly-react/pull/2375

rachael-phillips:
Thanks @ares and @dlabrecq !

tlabaj:
Closed by #2375

PR #2375, opened 2019-06-27
fix(Donut Chart): Add fixed point notation into percentage donut charts

amirfefer:
donut chart rounds up the title which might lead to a falsy 100%

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/Chart/DonutChart.js
+++ b/packages/patternfly-3/patternfly-react/src/components/Chart/DonutChart.js
@@ -9,20 +9,27 @@ import { getComposer } from './ChartConstants';
 const { pfSetDonutChartTitle } = patternfly;
 const colIndexOfMaxValue = columns => columns.reduce((iMax, x, i, arr) => (x[1] > arr[iMax][1] ? i : iMax), 0);
 
+const truncateNum = (num, precision) => {
+  const pointNotation = num.toString().split('.');
+  if (pointNotation.length === 1) return pointNotation[0];
+  return `${pointNotation[0]}.${pointNotation[1].slice(0, precision)}`;
+};
+
 const setDonutTitle = obj => {
   let primary;
   let secondary;
 
   const { props } = obj;
   const { data, title = {} } = props;
+  const { type, precision = 0 } = title;
   const { columns } = data;
   const sum = columns.reduce((acc, x) => acc + x[1], 0);
   const iMax = colIndexOfMaxValue(columns);
+  const percentage = (100 * columns[iMax][1]) / sum;
 
-  switch (title.type) {
+  switch (type) {
     case 'percent':
-      primary = `${Math.round((100 * columns[iMax][1]) / sum).toString()}%`;
-      [secondary] = columns[iMax];
+      primary = precision ? `${truncateNum(percentage, precision)}%` : `${Math.round(percentage)}%`;
       break;
     case 'max':
       primary = Math.round(columns[iMax][1]).toString();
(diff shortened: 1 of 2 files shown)


packages/patternfly-3/patternfly-react/src/components/Chart/DonutChart.js
```diff
@@ -15,13 +15,14 @@ const setDonutTitle = obj => {
 
   const { props } = obj;
   const { data, title = {} } = props;
+  const { type, percision = 0 } = title;
```
TheRealJon:
```suggestion
  const { type, precision = 0 } = title;
```

packages/patternfly-3/patternfly-react/src/components/Chart/DonutChart.js
```diff
@@ -15,13 +15,14 @@ const setDonutTitle = obj => {
...
   const { columns } = data;
   const sum = columns.reduce((acc, x) => acc + x[1], 0);
   const iMax = colIndexOfMaxValue(columns);
 
-  switch (title.type) {
+  switch (type) {
     case 'percent':
-      primary = `${Math.round((100 * columns[iMax][1]) / sum).toString()}%`;
+      primary = `${((100 * columns[iMax][1]) / sum).toFixed(percision).toString()}%`;
```
TheRealJon: Check out [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat)
```suggestion
      primary = `${Intl.NumberFormat(null, {style: 'percent', maximumFractionDigits: precision}).format(columns[iMax][1]) / sum)}%`;
```

packages/patternfly-3/patternfly-react/src/components/Chart/Stories/DonutChartStory.js
```diff
@@ -67,6 +75,20 @@ const donutChart = stories => {
...
+          <div>
+            <DonutChart
+              id="donunt-chart-3"
+              size={{
+                width: 210,
+                height: 210
+              }}
+              data={donutPercentageData}
+              tooltip={donutConfigTooltip}
+              title={{ type: 'percent', percision: 1 }}
```
TheRealJon:
```suggestion
              title={{ type: 'percent', precision: 1 }}
```

TheRealJon:
> @dlabaj I think using fixed-point notation would lead to decimal places being displayed, regardless of whether the value is fractional. (e.g. If percentage is 100, `precision =1` gives you `100.0%`, `precision = 2` gives you `100.00%`, and so on).

ping @amirfefer

packages/patternfly-3/patternfly-react/src/components/Chart/DonutChart.js
ares: actually this may still be a problem - (100 * 99.99999999 / 100).toFixed(2) # => 100.00
I'm not sure if the other suggestion solves it, maybe it does

dlabrecq: Wouldn't this change the output for existing applications? Perhaps we can leave the existing code as is, unless the  `precision` property is added?

ares:
I'm not sure if this is ever wanted behavior (in fact it may be seen as a bug), @amirfefer, what do you think? I'm fine with configuring through a property.

amirfefer:
@ares, `toFixed` apparently rounds up, so I use the known `slice` function.
@dlabaj - there is an issue for this already - #2356 
@TheRealJon - `Intl.NumberFormat` doesn't work here - getting ` Cannot convert undefined or null to object`, slicing is simpler in this case.
plus, if `precision` prop doesn't exist, the previous functionality remains

ares:
ping, is there something we can do to get this merged?

dlabrecq:
I see that the `precision` prop was added, so I'm good with that. I'd like to give Jon another chance to comment before merging.

ares:
Awesome I see 2 acks now :-)

ares:
Thanks all!

PR #2379, opened 2019-06-27
fix(clipboardcopy): add read only expanded example 

### LINKED ISSUE #2189: fix(clipboardCopy): add read-only example that is expanded
In Core we added a second expandable example for the read-only example. Lets also add this to Core. 

For this example, neither the input nor the expandable section should be editable.

Related to: https://github.com/patternfly/patternfly-next/pull/1896#issuecomment-499932189

jessiehuff:
Fixes #2189


jessiehuff:
@christiemolloy our latest update to tippy created some problems/added a span. I think it's fixed now. :)

PR #2387, opened 2019-06-28
Added OUIA compatibility to Pagination component

quarckster:
Requires #2304 to be merged


redallen:
@quarckster To get the OUIA props rendering in the statically rendered docs, there are two paths we can take:

1. Always include the props when `typeof window === undefined`. I've implemented this in #2487 . This has the effect of if consumers statically render components outside the browser, setting `window.localStorage.ouia` won't matter and OUIA props will be included by default (including in our statically rendered HTML docs). For the majority of downstream sites that are not statically rendered, `window.localStorage.ouia` will still be respected, so I believe this doesn't matter too much.

2. Use a stateful component wrapper. This has more overhead on the library side since it's an additional stateful component, but `window.localStorage.ouia` will be respected properly. @jschuler has implemented this in #2478 , but you'll have to update your PRs!

Pick your poison.

redallen:
We've picked option 2, so closing this PR in favor of ones like #2710 .

PR #2388, opened 2019-06-28
Added OUIA compatibility to Dropdown component

quarckster:
Requires #2304 to be merged


redallen:
Closing in favor of using an OUIA context like #2710

PR #2399, opened 2019-06-28
feat(card): Add compact card variation

### LINKED ISSUE #2392: Compact card variation
Related to https://github.com/patternfly/patternfly-next/issues/1946 and https://github.com/patternfly/patternfly-next/pull/1975

In Core we added a compact variation of the card. In React a prop should be added for the user to specify a card as "compact" so that it gets the compact styling.

jessiehuff:
Fixes #2392

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Card/Card.md
+++ b/packages/patternfly-4/react-core/src/components/Card/Card.md
@@ -362,3 +362,17 @@ HoverableCard = () => (
   </Card>
 );
 ```
+
+## Card compact example
+```js
+import React from 'react';
+import { Card, CardHeader, CardBody, CardFooter } from '@patternfly/react-core';
+
+HoverableCard = () => (
+  <Card isCompact={true}>
+    <CardHeader>Header</CardHeader>
+    <CardBody>Body</CardBody>
+    <CardFooter>Footer</CardFooter>
+  </Card>
+);
+```
--- a/packages/patternfly-4/react-core/src/components/Card/Card.md
+++ b/packages/patternfly-4/react-core/src/components/Card/Card.md
@@ -355,7 +355,7 @@ import React from 'react';
 import { Card, CardHeader, CardBody, CardFooter } from '@patternfly/react-core';
 
 HoverableCard = () => (
-  <Card isHoverable={true}>
+  <Card isHoverable>
     <CardHeader>Header</CardHeader>
     <CardBody>Body</CardBody>
     <CardFooter>Footer</CardFooter>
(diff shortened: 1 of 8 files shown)


christiemolloy:
Looks great @jessiehuff one thing though the headers in the card are set to pf-m-lg when they should be pf-m-md. 16px vs 18px.

Screen Shot 2019-06-28 at 12 26 56 PM
Screen Shot 2019-06-28 at 12 27 04 PM

packages/patternfly-4/react-core/src/components/Card/Card.md
```diff
@@ -362,3 +362,17 @@ HoverableCard = () => (
...
 );
 ```
+
+## Card compact example
+```js
+import React from 'react';
+import { Card, CardHeader, CardBody, CardFooter } from '@patternfly/react-core';
+
+HoverableCard = () => (
+  <Card isCompact={true}>
```
jschuler: usually if the prop is true we leave off the `={true}` part

packages/patternfly-4/react-core/src/components/Card/Card.tsx
```diff
@@ -11,18 +11,21 @@ export interface CardProps extends React.HTMLProps<HTMLDivElement> {
...
   className = '',
   component = 'article',
   isHoverable = false, 
+  isCompact = false,
   ...props
 }: CardProps) => {
   const Component = component as any;
   return (
-  <Component className={css(styles.card, isHoverable && styles.modifiers.hoverable, className)} {...props}>
+  <Component className={css(styles.card, isHoverable && styles.modifiers.hoverable, className, isCompact && styles.modifiers.compact)} {...props}>
```
jschuler: can you make `className` the last one

packages/patternfly-4/react-core/src/components/Card/Card.md
jessiehuff: Ok, cool. I kept it since the previously written test had it, thinking you guys liked it more explicit. I'll take it off of both. :) Thanks!!

packages/patternfly-4/react-core/src/components/Card/Card.tsx
```diff
@@ -11,18 +11,21 @@ export interface CardProps extends React.HTMLProps<HTMLDivElement> {
   component?: React.ReactNode;
   /** Modifies the card to include hover styles on :hover */
   isHoverable?: boolean;
+  isCompact?: boolean;
```
jschuler: oh woops, looks like we also need a descriptive comment here

jessiehuff: Nice catch! :)

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/CardDemo/CardDemo.tsx
```diff
@@ -4,7 +4,7 @@ import { Card, CardHeader, CardBody, CardFooter } from '@patternfly/react-core';
 export class CardDemo extends React.Component {
   render() {
     return (
-      <Card isHoverable={true}>
+      <Card isHoverable>
```
tlabaj: I don't see isCompact in the demo

PR #2406, opened 2019-06-30
fix(Modal): get modal container from state to set/unset aria-hidden

### LINKED ISSUE #2405: Modal should trap screen reader focus
Currently, the method for trapping focus in the Modal component for screen readers is not working as expected. This was previously working, but at some point, the behavior changed. I copied this description of the issue from #1389:

I think the PR where we implemented this was https://github.com/patternfly/patternfly-react/pull/1011. I'm not sure if there were other updates made since then that intended to modify this behavior, or if updates since then were made unintentionally, but the behavior required to trap focus for screen reader users is no longer working.

The method we use for trapping focus is documented in the [accessibility guide](https://pf4.patternfly.org/accessibility-guide#trapping-focus):
> ... apply `aria-hidden="true"` to the parent wrapping element of the page contents. Note that the modal element of the component must not be a descendent of this element with `aria-hidden="true"` and should be included as a sibling to this element.

Currently, if I open a modal, I see that `aria-hidden` is applied to every child of the `<body>` element on the page, including the `<div>` that has the modal. This breaks screen reader accessibility.

boaz0:
**What**:
fixes #2405 

When displaying a modal both the children of the body and the modal container had `aria-hidden=true`. The problem was that the container that was compared with the children is not stored in `this.container` but in the state.

This PR fixes this problem.

//cc @jgiardino

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
@@ -73,7 +73,7 @@ export class Modal extends React.Component<ModalProps, ModalState> {
   toggleSiblingsFromScreenReaders = (hide: boolean) => {
     const bodyChildren = document.body.children;
     for (const child of Array.from(bodyChildren)) {
-      if (child !== this.container) {
+      if (child !== this.state.container) {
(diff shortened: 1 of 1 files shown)


jgiardino:
Thanks for working on this, @boaz0!!

bitmoji

PR #2409, opened 2019-07-01
chore(react): bump react and react-dom versions

### LINKED ISSUE #2408: Bump React to 16.8.6 for Tippy 2
We need to update our React version to support #2326 . Consumers will have to upgrade their version of React to match our peer dependency in order to use Tippy components. Include this in the release notes @rachael-phillips .

redallen:
**What**: Closes #2408

**Additional issues**:


redallen:
Going to revert #2408 , so closing this one for now.

PR #2417, opened 2019-07-01
feat(Pagination): Let users enter numbers into the input field

### LINKED ISSUE #2344: PF4 Table: Pagination input field weird behavior
It's almost impossible to enter page number in the pagination field.
Steps to reproduce using docs:
1. Go to https://www.patternfly.org/v4/documentation/react/components/pagination/
2. Try to enter different numbers in the page field - the first number stays in the field regardless backspace, overall experience is strange.

rebeccaalpert:
Users can now enter numbers into the input field without highlighting the page number in the input field. You have to hit enter in order for the change to go through and affect the dropdown/buttons. Non-integer inputs will default to the previous page number on enter.

Fixes #2344.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
@@ -77,12 +77,12 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
     return inputPage;
   }
 
-  onChange(event: React.ChangeEvent<HTMLInputElement>, lastPage: number): void {
+  private onChange(event: React.ChangeEvent<HTMLInputElement>, lastPage: number): void {
     const inputPage = this.parseInteger(event.target.value, lastPage);
     this.setState({ userInputPage: Number.isNaN(inputPage as number) ? event.target.value : inputPage });
   }
 
-  onKeyDown(event: React.KeyboardEvent<HTMLInputElement>, page: number | string, lastPage: number, onPageInput: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void, onSetPage: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void): void {
+  private onKeyDown(event: React.KeyboardEvent<HTMLInputElement>, page: number | string, lastPage: number, onPageInput: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void, onSetPage: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void): void {
     if (event.keyCode === KEY_CODES.ENTER) {
       const inputPage = this.parseInteger(this.state.userInputPage, lastPage) as number;
       onPageInput(event, Number.isNaN(inputPage) ? page as number : inputPage);
--- a/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
@@ -26,7 +26,7 @@ export interface NavigationProps extends React.HTMLProps<HTMLElement> {
   /** Accessible label for the pagination component */
   paginationTitle?: string;
   /** The number of the current page */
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/components/Pagination/Navigation.js
```diff
@@ -40,99 +41,130 @@ const defaultProps = {
...
-    </Button>
-  </nav>
-);
+class Navigation extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { userInputPage: this.props.page };
+  }
+
+  parseInteger = (input, lastPage) => {
```
redallen: Can we hoist this up outside the class into `const parseInteger = (input, lastPage) => {...}`?

tlabaj: Why? Can it be used elsewhere? If so, we should move it to utilities.

rebeccaalpert: It's not reusable; I just broke it out here since it's been used in this class a couple of times. I made it private.

rebeccaalpert:
Refactored to work with TypeScript conversion.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -25,7 +26,7 @@ export interface NavigationProps extends React.HTMLProps<HTMLElement> {
   /** Accessible label for the pagination component */
   paginationTitle?: string;
   /** The number of the current page */
-  page: number;
+  page: number | React.ReactText;
```
redallen: `type ReactText = string | number;`, so can we just do `string | number` or just `React.ReactText`?

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -40,97 +41,148 @@ export interface NavigationProps extends React.HTMLProps<HTMLElement> {
...
+    currPage: 'Current page',
+    paginationTitle: 'Pagination',
+    onNextClick: () => undefined as any,
+    onPreviousClick: () => undefined as any,
+    onFirstClick: () => undefined as any,
+    onLastClick: () => undefined as any,
+    onPageInput: () => undefined as any,
+  };
+
+  private parseInteger(input: React.ReactText, lastPage: number): number | string {
```
redallen: Make this function `static` since it doesn't depend on `this`. Also, return type should be just `number`, I don't think there's a way this ever returns a `string`.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
jschuler: I wasn't aware of this type, but looks like ReactText is `string | number`, so you can remove number here

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -40,97 +41,148 @@ export interface NavigationProps extends React.HTMLProps<HTMLElement> {
...
-        onLastClick(event, lastPage);
-        onSetPage(event, lastPage);
-      }}
-    >
-      <AngleDoubleRightIcon />
-    </Button>
-  </nav>
-);
+export interface NavigationState {
+  userInputPage?: React.ReactText | number;
```
jschuler: same here about removing number

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
rebeccaalpert: @tlabaj - Are you ok with making it static instead of private?

redallen: I think it can be both! `static private parseInteger`...

rebeccaalpert:
Addressed PR feedback.

Issue #2370, opened 2019-06-26
PF4 Select:  Add ability to use custom component in select children


tlabaj:
> unable to use custom components in select children - we will be loading options from server so we would like to show spinner instead of select options and show options once we load them Critical

Add ability to use custom component in select children

tlabaj:
Closed by #2419

PR #2419, opened 2019-07-01
feat(Select): add optional display via children to option

kmcfaul:
**What**: Allows custom display of options via the children param.

**Additional issues**: #2370

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
@@ -28,7 +28,6 @@ export interface CheckboxSelectOptionProps extends React.HTMLProps<HTMLLabelElem
 export class CheckboxSelectOption extends React.Component<CheckboxSelectOptionProps> {
   private ref = React.createRef<any>();
   static defaultProps = {
-    children: undefined as React.ReactNode,
     className: '',
     value: '',
     index: 0,
@@ -91,7 +90,7 @@ export class CheckboxSelectOption extends React.Component<CheckboxSelectOptionPr
               checked={isChecked || false}
               disabled={isDisabled}
             />
-            <span className={css(checkStyles.checkLabel, isDisabled && styles.modifiers.disabled)}>{children && children || value}</span>
+            <span className={css(checkStyles.checkLabel, isDisabled && styles.modifiers.disabled)}>{children || value}</span>
           </label>
         )}
       </SelectConsumer>
--- a/packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
@@ -5,8 +5,8 @@ import { css } from '@patternfly/react-styles';
 import { SelectConsumer, KeyTypes } from './selectConstants';
 
 export interface CheckboxSelectOptionProps extends React.HTMLProps<HTMLLabelElement> {
-  /** The value for the option */
-  children?: string;
+  /** Optional alternate display for the option */
+  children?: React.ReactNode;
   /** Additional classes added to the Select Option */
   className?: string;
   /** Internal index of the option */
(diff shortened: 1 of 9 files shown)


packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
```diff
@@ -28,7 +28,7 @@ export interface CheckboxSelectOptionProps extends React.HTMLProps<HTMLLabelElem
 export class CheckboxSelectOption extends React.Component<CheckboxSelectOptionProps> {
   private ref = React.createRef<any>();
   static defaultProps = {
-    children: '',
+    children: undefined as React.ReactNode,
```
redallen: You don't ever need to set a variable to `undefined` -- it starts that way.

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
```diff
@@ -91,7 +91,7 @@ export class CheckboxSelectOption extends React.Component<CheckboxSelectOptionPr
               checked={isChecked || false}
               disabled={isDisabled}
             />
-            <span className={css(checkStyles.checkLabel, isDisabled && styles.modifiers.disabled)}>{value}</span>
+            <span className={css(checkStyles.checkLabel, isDisabled && styles.modifiers.disabled)}>{children && children || value}</span>
```
redallen: Isn't this equivalent to `{children}{value}`, which makes more sense?

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -34,6 +36,7 @@ export interface SelectOptionProps extends Omit<React.HTMLProps<HTMLElement>, 't
 export class SelectOption extends React.Component<SelectOptionProps> {
   private ref = React.createRef<any>();
   static defaultProps = {
+    children: undefined as React.ReactNode,
```
redallen: Also don't need to set this to undefined.

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -114,7 +118,7 @@ export class SelectOption extends React.Component<SelectOptionProps> {
                   ref={this.ref}
                   onKeyDown={this.onKeyDown}
                 >
-                  {value}
+                  {children && children || value}
```
redallen: Same thing here -- isn't this `{children}{value}`?

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -144,7 +148,7 @@ export class SelectOption extends React.Component<SelectOptionProps> {
                   checked={isChecked || false}
                   disabled={isDisabled}
                 />
-                <span className={css(checkStyles.checkLabel, isDisabled && styles.modifiers.disabled)}>{value}</span>
+                <span className={css(checkStyles.checkLabel, isDisabled && styles.modifiers.disabled)}>{children && children || value}</span>
```
redallen: Same thing here -- isn't this `{children}{value}`?

packages/patternfly-4/react-core/src/components/Select/CheckboxSelectOption.tsx
tlabaj: @redallen I don't think so.  value is always defined.  The way I am understanding the code is children will rendered if it is defined, otherwise value will be rendered. We never want to render both as `{children}{value}` would do if children is defined. Would updating the prop description help?

redallen: Ah, in that case it should just be `{children || value}`. `{children && children}` is equivalent to `{children}`.

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
redallen: Correction: `{children || value}`

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
redallen: Correction: `{children || value}`

kmcfaul:
@tlabaj I had added some unit tests for the custom display, as I wasn't sure if we wanted to show the custom display in our demos. Which demo should I edit to use the children prop, or should I add another separate demo for it?

tlabaj:
@kmcfaul I was talking about adding a demo to our integration app and adding cypress test.

kmcfaul:
Typeahead filtering still searches based on value, because the format of custom options is not known to the component. The custom filter PR will address this gap, and can be used to properly filter custom options.

PR #2421, opened 2019-07-01
fix(gutters): Remove size from gutter prop

### LINKED ISSUE #1982: "gutter" prop on stack, split, gallery, grid takes an unnecessary argument
In Stack, Split, Gallery, and Grid, the examples use syntax like `<Gallery gutter="md">`. However we don't currently support changing the gutter size (ie, "sm", "md", "lg", etc) in these components - you either add a gutter (`.pf-m-gutter`) or you don't.

jessiehuff:
Fixes #1982


mcoker:
Looks good! But can you also make this change to the level layout and remove the "Options are sm, md or lg" text from the prop docs in each of these layouts?

redallen:
I just tested on the docs site, and passing any of the old strings like `"md"` resolves to `true` and is technically a boolean, so PropTypes doesn't complain. In this case, I don't think it's breaking anything, although consumers should start passing a bool instead of a string.

tlabaj:
That is why it works with the current implementation.
How about folks using `GutterSize.sm`. This will break them.  Also we should not be documenting a type that is not the actual type.  This should be updated in a breaking change release.

packages/patternfly-4/react-core/src/demos/PageLayout/examples/PageLayoutDefaultNav.js
```diff
@@ -203,7 +203,7 @@ class PageLayoutDefaultNav extends React.Component {
             </TextContent>
           </PageSection>
           <PageSection>
-            <Gallery gutter="md">
+            <Gallery gutter>
```
jschuler: boolean props should start with some kind of identifying prefix, e.g. useGutter or hasGutter

seanforyou23: I like `has*`. `use*` could be confused with React hooks api parts. That said, maybe there is a distinction to be made between boolean props that have a presentational impact only vs behavioral impact? What about `usePf*`?

`usePfGutter` is pretty readable and wouldn't likely conflict with other prefixes. Thoughts?

jschuler: not sure about `usePf` but anyways, `hasGutter` would work fine I think

Issue #1389, opened 2019-02-14
AboutModal should trap focus


jgiardino:
Currently the AboutModal does not trap focus like the regular Modal does. The same method used for trapping focus in the Modal should be used for AboutModal.

Details about how to trap focus are documented in the [PF accessibility guide](https://pf4.patternfly.org/accessibility-guide#trapping-focus).

In addition to what's documented there, when a modal is dismissed by the user, focus should be placed on the button that opened the modal.

jgiardino:
I was reviewing the Modal with @jenny-s51 and noticed an issue with how Modal traps focus for screen readers. 

I think the PR where we implemented this was https://github.com/patternfly/patternfly-react/pull/1011. I'm not sure if there were other updates made since then that intended to modify this behavior, or if updates since then were made unintentionally, but the behavior required to trap focus for screen reader users is no longer working.

The method we use for trapping focus is documented in the [accessibility guide](https://pf4.patternfly.org/accessibility-guide#trapping-focus):
> ... apply `aria-hidden="true"` to the parent wrapping element of the page contents. Note that the modal element of the component must not be a descendent of this element with `aria-hidden="true"` and should be included as a sibling to this element.

Currently, if I open a modal, I see that `aria-hidden` is applied to every child of the `<body>` element on the page, including the `<div>` that has the modal. This breaks screen reader accessibility.

This issue with modal should be fixed before updating AboutModal to use a similar implementation.

jenny-s51:
please see #2428

tlabaj:
closed by #2405

PR #2428, opened 2019-07-02
fix(aboutModal+modal): now trap focus in browser

jenny-s51:
**What**: AboutModal and Modal both now trap focus in the browser. However, screenreader accessibility is currently functioning only in modal, which was fixed @boaz0 's PR referenced below -- I added his change to this PR. 

More feedback is necessary as to how to go about making screenreader accessibility work with AboutModal. Any assistance here would be appreciated. 🙂

**Additional issues**:  #2405 , #2406 , #1389


packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
```diff
@@ -29,12 +29,17 @@ export interface AboutModalProps {
...
+interface ModalState {
+  container: HTMLElement;
+}
+
+export class AboutModal extends React.Component<AboutModalProps, AboutModal, ModalState> {
+  private static currentId: number = 0;
   private id = AboutModal.currentId++;
   ariaLabelledBy = `pf-about-modal-title-${this.id}`;
   ariaDescribedBy = `pf-about-modal-content-${this.id}`;
+  container?: HTMLDivElement = undefined;
```
redallen: Remove `= undefined`.

packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
```diff
@@ -73,7 +73,7 @@ export class Modal extends React.Component<ModalProps, ModalState> {
   toggleSiblingsFromScreenReaders = (hide: boolean) => {
     const bodyChildren = document.body.children;
     for (const child of Array.from(bodyChildren)) {
-      if (child !== this.container) {
+      if (child !== this.state.container) {
```
redallen: Accident?

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
```diff
@@ -51,14 +56,26 @@ export class AboutModal extends React.Component<AboutModalProps> {
...
+  handleEscKeyClick = (event: KeyboardEvent) => {
     if (event.keyCode === KEY_CODES.ESCAPE_KEY && this.props.isOpen) {
       this.props.onClose();
     }
   };
 
+  toggleSiblingsFromScreenReaders = (hide: boolean) => {
+    const bodyChildren = document.body.children;
+    for (const child of Array.from(bodyChildren)) {
+      if (child !== this.state.container) {
```
redallen: `this.state.container` will always be undefined?

packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
jenny-s51: This is the change that Boaz added to his PR; I think it needs to be written as `this.state.container` because when it's not, the screenreader can't identify the buttons in the modal.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
jenny-s51: I see what you mean ... My understanding of this function is that as long as a child of `bodyChildren` is not undefined, it is assigned an `ariaHidden` value of either true or false. Is that correct?

jgiardino:
Hey @jenny-s51 - these updates are great!

I tested keyboard accessibility using the following steps, and it worked as expected 🎉 
1. Added a link with `href="#"1  to the contents of the modal. This is to provide another element that can receive focus besides the Close button.
2. Opened the modal. Focus is placed on the Close button, as expected.
3. Hit Tab. Focus is placed on the link, as expected.
4. Hit Tab again. Focus is placed on the Close button, as expected.
5. Repeated steps 3 & 4 with Shift + Tab to check that focus stays in the modal. Works as expected.
6. Closed the modal, and focus is placed on the modal toggle, as expected.

There is an issue in the DOM that would affect screen reader accessibility. The following screencap shows `aria-hidden="true"` is applied to the div that includes the modal. This will hide the contents from assistive technologies, like screen readers. This was the same issue that was fixed for the Modal in https://github.com/patternfly/patternfly-react/pull/2406.

image

tlabaj:
Looking good. In the future, it is better to separate PRs that touch different areas.  Make for smaller code reviews and easier for tracking.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModalBoxCloseButton.tsx
```diff
@@ -13,12 +13,14 @@ export interface AboutModalBoxCloseButtonProps extends React.HTMLProps<HTMLDivEl
...
-  onClose = () => undefined,
+  onClose = () => undefined as any,
   ...props
 }: AboutModalBoxCloseButtonProps) => (
-  <div className={css(styles.aboutModalBoxClose, className)} {...props}>
-    <Button variant={ButtonVariant.plain} onClick={onClose} aria-label="Close Dialog">
-      <TimesIcon />
-    </Button>
-  </div>
+  <React.Fragment>
```
redallen: No need for the extra <React.Fragment> here.

packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
```diff
@@ -29,12 +29,17 @@ export interface AboutModalProps {
...
+interface ModalState {
+  container: HTMLElement;
+}
+
+export class AboutModal extends React.Component<AboutModalProps, ModalState> {
+  private static currentId: number = 0;
   private id = AboutModal.currentId++;
   ariaLabelledBy = `pf-about-modal-title-${this.id}`;
   ariaDescribedBy = `pf-about-modal-content-${this.id}`;
+  container?: HTMLDivElement;
```
redallen: This class attribute can be removed.

packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
```diff
@@ -40,7 +40,7 @@ interface ModalState {
 export class Modal extends React.Component<ModalProps, ModalState> {
   static currentId = 0;
   id = '';
-  container?: HTMLDivElement = undefined;
+  container?: HTMLDivElement;
```
redallen: This variable can be removed too.

packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
redallen: Ah, this makes sense given your refactor of `this.container` -> `this.state.container`. It should have always been in `state` since it affects rendering. Thanks!

Issue #2369, opened 2019-06-26
PF4 Select: Unable to search by callback function 


tlabaj:
> Unable to search by callback function - select options are searchable only on first char, however we would like to query server to retrieve new options Critical

Add a callback function for the search

PR #2434, opened 2019-07-02
feat(Select): add select custom filter callback

kmcfaul:
**What**: Adds a custom callback for the filtering function used in typeahead variants. This replaces the current search entirely, so the function must handle filtering the list of children/options and return the filtered results for the internal state to update.

**Additional issues**: #2369

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
@@ -6,7 +6,6 @@ import { CheckboxSelectOption } from './CheckboxSelectOption';
 import { SelectGroup } from './SelectGroup';
 import { CheckboxSelectGroup } from './CheckboxSelectGroup';
 import { SelectVariant } from './selectConstants';
-import { SelectContext } from '../../../dist/js';
 
 const selectOptions = [
   <SelectOption value="Mr" key="0" />,
--- a/packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
@@ -6,6 +6,7 @@ import { CheckboxSelectOption } from './CheckboxSelectOption';
 import { SelectGroup } from './SelectGroup';
 import { CheckboxSelectGroup } from './CheckboxSelectGroup';
 import { SelectVariant } from './selectConstants';
+import { SelectContext } from '../../../dist/js';
 
 const selectOptions = [
   <SelectOption value="Mr" key="0" />,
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -115,18 +119,24 @@ export class Select extends React.Component<SelectProps, SelectState> {
   };
 
   onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    let input: RegExp;
-    try {
-      input = new RegExp(e.target.value, 'i');
-    } catch (err) {
-      input = new RegExp(e.target.value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'i');
+    const { onFilter } = this.props;
```
tlabaj: Can you add unit test for this.

packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
```diff
@@ -6,6 +6,7 @@ import { CheckboxSelectOption } from './CheckboxSelectOption';
 import { SelectGroup } from './SelectGroup';
 import { CheckboxSelectGroup } from './CheckboxSelectGroup';
 import { SelectVariant } from './selectConstants';
+import { SelectContext } from '../../../dist/js';
```
jschuler: tests shouldn't require the dist to be built, no? can you load it from source?

packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -50,6 +50,8 @@ export interface SelectProps
   onToggle: (isExpanded: boolean) => void;
   /** Callback for typeahead clear button */
   onClear?: (event: React.MouseEvent) => void;
+  /** Optional callback for custom filtering */
+  onFilter?: (e: React.ChangeEvent<HTMLInputElement>) => React.ReactElement[];
```
jschuler: Can you add an example to Select.md?

packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
redallen: Yes, tests don't need the `dist`. Jest builds them itself according to our rules.

jschuler:
LGTM, but noticed two other things:
1. When i type something and filter down the results, and then close and reopen the dropdown (the filter text is still applied) it shows me the whole list again. Is this expected?
2. When i start typing and the dropdown is closed, it does not automatically open. I would expect it to open

kmcfaul:
1. When you close the menu the filter resets. If you select an item, that is filled into the input but it isn't acting as a filter, and maybe that should change?

2. is being addressed in a separate PR

PR #2450, opened 2019-07-03
feat(Level): Convert level to typescript

### LINKED ISSUE #2449: Convert Level to Typescript
Convert level to typescript, add demo, and integration testing

jessiehuff:
Fixes #2449


packages/patternfly-4/react-core/src/layouts/Level/Level.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/layouts/Level/level';
+
+export interface LevelProps extends React.HTMLProps<HTMLDivElement> {
+   /** Adds space between children. Options are sm, md or lg */
+   gutter?: boolean;
```
tlabaj: Same comment as gallery PR.  Why are we changing the type of gutter?

jessiehuff: Like in gallery, I thought we were fixing the gutter prop type in all of these, I'll change them all back.

packages/patternfly-4/react-core/src/layouts/Level/examples/Level.md
```diff
@@ -25,7 +26,7 @@ import { Level, LevelItem } from '@patternfly/react-core';
 import React from 'react';
 import { Level, LevelItem } from '@patternfly/react-core';
 
-<Level gutter="md">
+<Level gutter>
```
tlabaj: this should be reverted

packages/patternfly-4/react-core/src/layouts/Level/Level.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const Level: React.FunctionComponent<LevelProps> = ({
+  gutter = null,
+  className = '',
+  children = null, 
+  ...props
+}: LevelProps) => (
+  <div
+    {...props}
+    className={css(styles.level, gutter && styles.modifiers.gutter, className)}
```
jschuler: needs to use getGutterModifier function

jschuler: Actually looking at this a little more now, it does not seem like there are other gutter sizes we can apply? Just the one 'pf-m-gutter'? Then the gutter prop type does not make sense. Should probably just be a boolean `hasGutter` or something like that

jschuler:
@jessiehuff Does it matter what gets passed into the gutter prop? Seems like the component only has 'pf-m-gutter' and no other variations for sm, md, lg

redallen:
@jschuler It hasn't for quite some time. Gutter sizes don't exist for layouts. That's what she's addressing in #2421 .

jessiehuff:
@jschuler Right now it doesn't matter what gets passed to the gutter prop, and the sizes don't really do anything. When I talked to Coker, he told me that they'd decided to remove the sizes in core so I created this [PR](https://github.com/patternfly/patternfly-react/pull/2421), but it seems there's still discussion about how to handle it.

jschuler:
Ah, gotcha, thanks! At first it seemed to me that the variations accidentally got dropped but now it makes sense :)

PR #2455, opened 2019-07-03
fix(login page): removes ariaLabel from checkbox

### LINKED ISSUE #1737: LoginPage - remove aria-label from checkbox
The checkbox in the LoginPage component includes an aria-label in addition to a traditional `<label>`. This attribute should be removed, since we should be using the string defined inside `<label>` only.

jenny-s51:
closes #1737

### DIFF
--- a/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.tsx
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.tsx
@@ -39,7 +39,9 @@ export interface LoginFormProps extends React.HTMLProps<HTMLFormElement> {
   isRememberMeChecked?: boolean;
   /** Function that handles the onChange event for the Remember Me Checkbox */
   onChangeRememberMe?: (checked: boolean, event: React.FormEvent<HTMLInputElement>) => void;
-  /** Aria Label for the Remember me checkbox, use this to override using the rememberMeLabel */
+  /* THIS PROP IS DEPRECATED AND NO LONGER USED; remove in a future breaking change release
+   * Aria Label for the Remember me checkbox, use this to override using the rememberMeLabel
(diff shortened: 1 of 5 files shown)


jgiardino:
These updates look good! The only question I have is whether this would be a breaking change (and if so, are there any changes we can make now without waiting for a breaking change release)? I'm not able to test what happens if `rememberMeAriaLabel` is defined on `<LoginForm>` (for some reason, the code editor section is read-only).

jenny-s51:
@jgiardino thank you for your feedback! I'm curious as well -- @tlabaj would this be a breaking change?

packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.tsx
```diff
@@ -97,7 +97,6 @@ export const LoginForm: React.FunctionComponent<LoginFormProps> = ({
           label={rememberMeLabel}
           checked={isRememberMeChecked}
           onChange={onChangeRememberMe}
-          aria-label={rememberMeAriaLabel || rememberMeLabel}
```
dgutride: This looks good to me - let's add one more comment up above in rememberMeAriaLabel (to the comment section noting this propType is now deprecated and is no longer used in the code.  Also, please open an issue with a breaking change label indicating we should delete this propType in a future breaking change release.

jenny-s51: Just pushed again with this addition to the comment. Should be all set now -- made a new breaking change issue as well.

tlabaj:
Can you mention new issue here please?  Thnaks.

jenny-s51:
@tlabaj It's #2464 🙂

PR #2457, opened 2019-07-04
fix(BreadCrumbSwitcher): Fix Autocomplete camelCase

glekner:
**What**:
React requires camelCase for DOM attributes, we need to change `autocomplete` to `autoComplete` in PF.

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/BreadcrumbSwitcher/components/BreadcrumbSearchInput.js
+++ b/packages/patternfly-3/patternfly-react/src/components/BreadcrumbSwitcher/components/BreadcrumbSearchInput.js
@@ -25,7 +25,7 @@ class BreadcrumbSearchInput extends React.Component {
           inputRef={input => {
             this.nameInput = input;
           }}
-          autocomplete="off"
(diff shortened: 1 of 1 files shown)


glekner:
@jeff-phillips-18 @priley86 can we merge :) ?

tlabaj:
@glekner can you add the associated issue to the description please.

packages/patternfly-3/patternfly-react/src/components/BreadcrumbSwitcher/components/BreadcrumbSearchInput.js
```diff
@@ -25,7 +25,7 @@ class BreadcrumbSearchInput extends React.Component {
           inputRef={input => {
             this.nameInput = input;
           }}
-          autocomplete="off"
```
redallen: Since this is just a prop passed down to the DOM in https://github.com/nkbt/react-debounce-input/blob/master/src/Component.js, this is good.

glekner:
@jeff-phillips-18 can we merge? thanks!

PR #2463, opened 2019-07-05
fix(charts): remove invalid pf-core var

dlabrecq:
Removes an invalid pf-core import

fixes https://github.com/patternfly/patternfly-react/pull/2461


tlabaj:
Is this follow up from last PR?

PR #2468, opened 2019-07-07
feat(switch): add aria-labelledby to input

### LINKED ISSUE #2435: Switch - add prop for aria-labelledby
Related to this comment: https://github.com/patternfly/patternfly-react/issues/1296#issuecomment-468962918

The switch component can be updated to include `aria-labelledby` instead of `aria-label` so that the runtime html can resemble the examples in core that use `aria-labelledby`.

A couple of notes/questions about this solution:

- Core shows an id on both the on and off message `<span>` elements. Only the `id` on the on message should be used by aria-labelledby.
- One of the benefits for supporting this example is we could allow the user to define their switch like this...
  ```
      <Switch
        id="simple-switch"
        label={isChecked ? 'Message when on' : 'Message when off'}
        isChecked={isChecked}
        onChange={this.handleChange}        
      />
  ```
  ... and automatically associate the aria-labelledby attribute to the id we create for the on message. 
  - This solution would require that we automatically generate an `id` for the on message. Is that possible? Could it be generated from the value provided in the `id` above, so that the switch includes `<input id="simple-switch" ... />` and the on message is `<span class="pf-c-switch__label pf-m-on" id="simple-switch-on" aria-hidden="true">`?
  - If we did implement this behavior, then if the consumer specifies `aria-label`, we would not want to include the `aria-labelledby` attribute on the `<input>` element.
  - This solution would only work if `label` is defined with both on and off text strings. Currently, our disabled examples show just one text string being defined, which would result in the wrong text string getting announced for a disabled switch in the off state. 
- We could also allow the consumer to specify explicit values for `aria-labelledby` and not use defaults. 

@tlabaj please share any suggestions you have on this.

### LINKED ISSUE #2600: PF4 Switch label props breaking change
A follow up to #2468 - to match the implementation to PF4 core the current API should be changed to include the label when the toggle is off and a different label when the toggle is on.

To solve this two ways come to mind:

* update the type of `label` from `string` to `string[2]` where the first item is the label on `on` and the second one is on `off`
* create two props `labelOn` and `labelOff` for the `on` and `off` states respectively.

This is going to be breaking change as it changes the API of the switch component.

//cc @jgiardino

boaz0:
**What**:

closes #2435 
closes #2600 

//cc @jgiardino @tlabaj

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.md
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.md
@@ -28,10 +28,10 @@ class SimpleSwitch extends React.Component {
     return (
       <Switch
         id="simple-switch"
-        label={isChecked ? 'Message when on' : 'Message when off'}
+        label="Message when on"
+        labelOff="Message when off"
         isChecked={isChecked}
         onChange={this.handleChange}
-        aria-label="Message when on"
       />
     );
   }
@@ -70,12 +70,13 @@ import { Switch } from '@patternfly/react-core';
 
 DisabledSwitch = () => (
   <React.Fragment>
-    <Switch id="disabled-switch-on" aria-label="Message when on" label="Message when on" isChecked isDisabled />
+    <Switch id="disabled-switch-on" aria-label="Message when on" label="Message when on" labelOff="Message when off" isChecked isDisabled />
     <br />
     <Switch
       id="disabled-switch-off"
       aria-label="Message when on"
-      label="Message when off"
+      label="Message when on"
+      labelOff="Message when off"
       isChecked={false}
       isDisabled
     />
@@ -94,9 +95,9 @@ import { Switch } from '@patternfly/react-core';
 
 UncontrolledSwitch = () => (
   <React.Fragment>
-    <Switch id="uncontrolled-switch-on" aria-label="Message when on" label="Message when on" isChecked />
+    <Switch id="uncontrolled-switch-on" aria-label="Message when on" label="Message when on" labelOff="Message when off" isChecked />
     <br />
-    <Switch id="uncontrolled-switch-off" aria-label="Message when on" label="Message when off" isChecked={false} />
+    <Switch id="uncontrolled-switch-off" aria-label="Message when on" label="Message when on" labelOff="Message when off" isChecked={false} />
     <br />
     <Switch id="uncontrolled-no-label-switch-on" aria-label="Message when on" isChecked />
     <br />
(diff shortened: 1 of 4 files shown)


boaz0:
fix merge conflict. :smile:

jgiardino:
Thanks for contributing this, @boaz0! 

When I modify the first example, and remove `aria-label` from the component, I see that `aria-labelledby` is added to the `<input>` as expected. I also see that `id` attributes are added to the `<input>` and also the `<span>` elements that provide the visible text labels, as expected. 🎉 

There are just a couple of updates needed:

- When `aria-labelledby` is added to the `<input>`, it should be set to the `id` value that is defined on the `<span>` that is visible during the `on` state. 
  - Currently, the attribute is set to the `id` value assigned to the `<input>` as shown below:
    `<input id="simple-switch" aria-label="" class="pf-c-switch__input" type="checkbox" aria-labelledby="simple-switch" checked="">`
  - Instead, it should render like this:
    `<input id="simple-switch" aria-label="" class="pf-c-switch__input" type="checkbox" aria-labelledby="simple-switch-on" checked="">`
- To align with our core examples, let's remove `aria-label` from the first example, so that the component is defined like this:
  ```
      <Switch
        id="simple-switch"
        label={isChecked ? 'Message when on' : 'Message when off'}
        isChecked={isChecked}
        onChange={this.handleChange}
      />
  ```
  And then when the html is rendered, this will include the `aria-labelledby` attribute like the example in core.
  - I think this would be sufficient for aligning with the core examples, but if we do decide to have complete alignment, then it's important to note that the react component examples showing the disabled switch would need to include both an on and off label (currently the disabled examples only include a string for the label that's visible).

boaz0:
Thanks a lot @jgiardino on your input. 
- I updated this PR, can you please take a second look?
- I opened #2600 to address your comment regarding the mismatch between core and react implementations.

tlabaj:
@boaz0 will you be addressing issue #2600 here?

boaz0:
separate PR :+1:

boaz0:
@tlabaj @jgiardino @redallen I updated the PR to address #2600 too.

Thanks.

boaz0:
Thanks @tlabaj  @kmcfaul @redallen ! :smile:

SpyTec:
This introduced a breaking change in a minor version (from `v3.80.4` to `v3.81.0`). Prior to this the label prop was be shown both when on and off. Not it only shows when on and not when off.

We have to explicitly add both `label` and `labelOff` even when we don't want to change the text. Was this intended?

PR #2473, opened 2019-07-08
fix(Dropdown): Provide option to not autofocus on first item

### LINKED ISSUE #2472: PF4 Dropdown should allow callers to specify the initial focus item
If a dropdown is being used for selection, the currently selected item should get the initial focus. Therefore, the application should be able to specify which dropdown item has initial focus rather than defaulting to the first item.

Also, if using custom children (not an array of dropdown items) the outer container is given focus. In the case of custom children, the DropdownMenu should not try to determine the initial focus.

jeff-phillips-18:
**What**:
Fixes Dropdown to not autofocus when custom children are given. Also provides the option to not autofocus on the first item for the case where the application wants to set autoFocus on one of the items.

**Additional issues**:

Fixes #2472

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
@@ -17,6 +17,9 @@ const propTypes = {
   isOpen: PropTypes.bool,
   /** Flag to indicate if menu should be opened on enter */
   openedOnEnter: PropTypes.bool,
+  /** Flag to indicate if the first dropdown item should gain initial focus, set false when adding
+   * a specific auto-focus item (like a current selection) otherwise leave as true */
+  autoFocus: PropTypes.bool,
   /** Indicates which component will be used as dropdown menu */
   component: componentShape,
   /** Indicates where menu will be alligned horizontally */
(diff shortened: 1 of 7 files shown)


packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
```diff
@@ -41,14 +45,18 @@ class DropdownMenu extends React.Component {
...
   componentDidMount() {
-    const focusTarget =
-      this.refsCollection.filter(
-        ref => ref && ((ref.current && !ref.current.hasAttribute('disabled')) || !ref.hasAttribute('disabled'))
-      )[0] || null;
-    if (this.props.component === 'ul') focusTarget && focusTarget.focus();
-    else {
-      (focusTarget.current.focus && focusTarget.current.focus()) ||
-        (focusTarget && ReactDOM.findDOMNode(focusTarget.current).focus()); // eslint-disable-line react/no-find-dom-node
+    const { autoFocus } = this.props;
```
redallen: Could also destructure `component`.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownMenu.js
```diff
@@ -41,14 +45,18 @@ class DropdownMenu extends React.Component {
...
+
+    if (this.props.component === 'ul' && autoFocus) {
+      const focusTarget =
+        this.refsCollection.filter(
```
redallen: `.find()` is equivalent to `.filter()[0]`.

PR #2489, opened 2019-07-09
fix(unit tests): tests now pass in util.test.js

### LINKED ISSUE #1624: Unit tests for `util.ts` are broken.
The unit tests for `keyHandler` in `util.test.js` are not making any assertions. The `expect` lines should look like this:

```js
    expect(firstDropdownItem).toEqual(document.activeElement);
```

instead of this:

```js
    expect(secondDropdownItem === document.activeElement);
```

As they are, these tests are not actually accomplishing anything and currently fail if the assertions are corrected. These should be revised and updated.

jenny-s51:
**What**: Closes #1624 and fixes keyHandler tests in util.test.js -- tests are now passing.

### DIFF
--- a/packages/patternfly-4/react-core/src/helpers/util.test.js
+++ b/packages/patternfly-4/react-core/src/helpers/util.test.js
@@ -88,18 +88,17 @@ test('sideElementIsOutOfView Returns NONE when in view', () => {
   expect(sideElementIsOutOfView(container, element)).toBe(SIDE.NONE);
 });
 
-// TODO: Fix keyHandler Tests!!!
 // eslint-disable-next-line jest/no-disabled-tests
-xdescribe('keyHandler works on ApplicationLauncher', () => {
+describe('keyHandler works on ApplicationLauncher', () => {
   document.body.innerHTML = '<!doctype html><html><body></body></html>';
   const dropdownItems = [
-    <DropdownItem key="link" id="first">
-      Link
+    <DropdownItem key=" 1" id="first" component="button">
+      Action 1
     </DropdownItem>,
-    <DropdownItem key="action" id="second" component="button">
-      Action
+    <DropdownItem key="2" id="second" component="button">
+      Action 2
     </DropdownItem>,
-    <DropdownItem key="disabled link" id="third" isDisabled>
+    <DropdownItem key="3" id="third" component="button" isDisabled>
       Disabled Link
     </DropdownItem>
   ];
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/helpers/util.test.js
```diff
@@ -88,70 +88,86 @@ test('sideElementIsOutOfView Returns NONE when in view', () => {
...
-    expect(secondDropdownItem === focusElement).toBe(true);
+    const focusedElement = document.activeElement;
+    expect(focusedElement.getAttribute('id')).toEqual('second');
   });
 
   test('keyHandler regresses backward', () => {
-    secondDropdownItem.simulate('keydown', { key: 'ArrowUp', keyCode: KEY_CODES.ARROW_UP, which: KEY_CODES.ARROW_UP });
-    expect(firstDropdownItem === document.activeElement).toBe(true);
+    let focusedElement = document.activeElement;
+    expect(focusedElement.getAttribute('id')).toEqual('second');
```
jeff-phillips-18: I would remove the initial expected focus item check from each of the tests. This makes each test depend on the previous test.

packages/patternfly-4/react-core/src/helpers/util.test.js
```diff
@@ -88,70 +88,86 @@ test('sideElementIsOutOfView Returns NONE when in view', () => {
   expect(sideElementIsOutOfView(container, element)).toBe(SIDE.NONE);
 });
 
-// TODO: Fix keyHandler Tests!!!
 // eslint-disable-next-line jest/no-disabled-tests
```
TheRealJon:
```suggestion
```

packages/patternfly-4/react-core/src/helpers/util.test.js
```diff
@@ -88,70 +88,86 @@ test('sideElementIsOutOfView Returns NONE when in view', () => {
...
+      keyCode: KEY_CODES.ARROW_UP,
+      which: KEY_CODES.ARROW_UP
+    });
+    focusedElement = document.activeElement;
+    expect(focusedElement.getAttribute('id')).toEqual('second');
   });
 });
 
-// TODO: Fix keyHandler Tests!!!
 // eslint-disable-next-line jest/no-disabled-tests
```
TheRealJon:
```suggestion
```

packages/patternfly-4/react-core/src/helpers/util.test.js
TheRealJon: Since there is no set-up or tear-down between each test case, they will depend on the previous test regardless. The following tests will fail if the DOM is not in the correct state. Maybe we should be initializing the DOM state from scratch on each test case?

jenny-s51: @jeff-phillips-18 what do you think?

jeff-phillips-18: I think it's ok to just remove the init checks. The initial state is not handled by the utils and thus need not be tested here.

PR #2492, opened 2019-07-09
fix(ApplicationLauncher): Allow custom icon for application launcher toggle

jeff-phillips-18:
**What**:
Accept a prop (toggleIcon) to allow overriding the application launcher's toggle icon.


jeff-phillips-18:
@redallen Fixed the CI issue.

Issue #1446, opened 2019-02-26
PF4 Tabs: allow eventKey to accept a string


dlabrecq:
As a user, I want to use a string as my tab `eventKey`. Would it be possible to change the `eventKey` type to accept a number or string?

In the Cost Management UI, we iterate over a map of available tabs. We dynamically show/hide tabs, so we identify child components using an ID string; 'aws', 'openshift', etc. 

As a workaround, I've added code to look up IDs based on the `activeTabKey`. However, it would be ideal if we could just provide the tab eventKey with a string instead. This was how our custom tabs component worked.

Below is an example of how we add tabs dynamically.

![screen shot 2019-02-26 at 9 14 09 am](https://user-images.githubusercontent.com/17481322/53418967-e8869800-39a6-11e9-9345-ca57950c28b1.png)

dlabrecq:
It would still be ideal to accept a string or number. I believe this is just a matter of changing the prop type

tlabaj:
closed by #2493

PR #2493, opened 2019-07-10
feat(Tabs): Allow eventKey to accept a string

tlabaj:
#1446

**What**:

**Additional issues**:


packages/patternfly-4/react-core/src/components/Tabs/Tabs.md
```diff
@@ -340,6 +340,7 @@ class SeparateTabContent extends React.Component {
   render() {
     return (
       <React.Fragment>
+      <div id
```
rebeccaalpert: Is this supposed to be here?

rebeccaalpert: Is this supposed to be here? It's a causing a syntax error in the demo.

PR #2495, opened 2019-07-10
fix(chipGroup): Allow variable number of chips to be displayed

### LINKED ISSUE #1855: ChipGroup should allow a variable number of chips to be displayed
The ChipGroup component was designed to group a set of chips to be displayed together and handle potential overflow situations. Looks like as currently implemented, the group is implemented to display a fixed number of chips before hiding additional chips by collapsing the group and exposing a "more" button. This became evident when reviewing the type ahead select (ref https://github.com/patternfly/patternfly-react/pull/1839) and seeing that the list of selections is collapsed even when there is plenty of space to display them.

This should be changed to allow the consumer of this component to either pass in a number of chips to display or choose to make that flexible and calculated based on the width of the group's container. The desired behavior is for the ChipGroup to only collapse the group when there would not be enough space to display all elements without wrapping.

jessiehuff:
Fixes #1855


tlabaj:
Closing this PR.  Issue is on hold awaiting design.

Issue #2496, opened 2019-07-10
Chart tooltip colors are incorrect


dlabrecq:
While adding custom tooltips to the examples, I noticed that the color vars in pf-core are incorrect. The background and label are the same black color, so the text is not visible when overriding labelComponent (i.e., using ChartTooltip).

<img width="521" alt="Screen Shot 2019-07-10 at 4 45 53 PM" src="https://user-images.githubusercontent.com/17481322/61003592-5027f900-a332-11e9-9094-d8f53a10c2ab.png">

We need to adjust a couple pf-core vars and create some new ones. See https://github.com/patternfly/patternfly-next/issues/2037

In addition, we were also applying old, hard coded color properties over the pf-core vars. We need to clean up the theme a bit.

Issue #2482, opened 2019-07-09
PF4: Chart tooltip is unresponsive


karelhala:
When I use bar chart with title and hover over different values the title is not rendering correctly and jumps over different values.

![Peek 2019-07-09 14-08](https://user-images.githubusercontent.com/3439771/60886865-4da2a200-a253-11e9-9b72-ee8efda1a7f5.gif)

karelhala:
@rachael-phillips ping this is quite urgent because our UX person is currently designing report tool that will show bar chart with titles.

dlabrecq:
This tooltip functionality is provided by Victory; however, tooltips can be applied to charts differently. 

In this particular example, we overrode the chart's containerComponent property. However, you can add a tooltip directly to the bar component itself.

See https://formidable.com/open-source/victory/guides/tooltips#simple-tooltips

dlabrecq:
We may change the examples so that bar chart example does not use VoronoiContainer.

dlabrecq:
FYI, the issue appears to be resolved if you set `voronoiDimension`, like so:

`containerComponent={<ChartVoronoiContainer labels={datum => `${datum.name}: ${datum.y}`} voronoiDimension="x"/>}`

dlabrecq:
Added the `voronoiDimension` prop to fix the tooltip hover for our bar chart examples. Then, modified the stack chart example to show how tooltips are applied without a voronoi container.

https://github.com/patternfly/patternfly-react/pull/2497

rachael-phillips:
Hi @karelhala I'm sorry for the delay - I was away and just got back from pto. @dlabrecque did we resolve @karelhala's issue?

dlabrecq:
I believe so. I was able to correct the bar chart example

Issue #2476, opened 2019-07-08
PF4: StackChart - setting the default tooltip background color


priley86:
It seems that our current default theme applied to StackCharts sets the tooltip fill and stroke to #EDEDED making the text unreadable.

![Screen Shot 2019-07-08 at 3 09 18 PM](https://user-images.githubusercontent.com/4237045/60836785-79ab1e00-a194-11e9-965d-2e1a9b061e31.png)

(changing our first StackChart example to reproduce this):
```
    <Chart
      legendData={[{ name: 'Cats' }, { name: 'Birds' }, { name: 'Dogs' }, { name: 'Mice' }]}
      legendOrientation="vertical"
      legendPosition="right"
      height={250}
      padding={{
        bottom: 50,
        left: 50,
        right: 200, // Adjusted to accomodate legend
        top: 50
      }}
      width={600}
    >
      <ChartStack domainPadding={{x: [10, 2]}}>
        <ChartBar data={[{ name: 'Cats', x: '2015', y: 1, label: 'Abcd' }, { name: 'Cats', x: '2016', y: 2 }, { name: 'Cats', x: '2017', y: 5 }, { name: 'Cats', x: '2018', y: 3 }]} labelComponent={<ChartTooltip />} />
        <ChartBar data={[{ name: 'Dogs', x: '2015', y: 2 }, { name: 'Dogs', x: '2016', y: 1 }, { name: 'Dogs', x: '2017', y: 7 }, { name: 'Dogs', x: '2018', y: 4 }]} />
        <ChartBar data={[{ name: 'Birds', x: '2015', y: 4 }, { name: 'Birds', x: '2016', y: 4 }, { name: 'Birds', x: '2017', y: 9 }, { name: 'Birds', x: '2018', y: 7 }]} />
        <ChartBar data={[{ name: 'Mice', x: '2015', y: 3 }, { name: 'Mice', x: '2016', y: 3 }, { name: 'Mice', x: '2017', y: 8 }, { name: 'Mice', x: '2018', y: 5 }]} />
      </ChartStack>
    </Chart>
```

![Screen Shot 2019-07-08 at 2 37 59 PM](https://user-images.githubusercontent.com/4237045/60836882-becf5000-a194-11e9-9487-dd48c6052b44.png)
![Screen Shot 2019-07-08 at 2 38 56 PM](https://user-images.githubusercontent.com/4237045/60836888-c68ef480-a194-11e9-8204-bda2becf985a.png)

[...]

PR #2497, opened 2019-07-10
fix(charts): Adjust pf-core vars & add tooltip examples

dlabrecq:
Adjusted/added pf-core vars to ensure custom tooltip colors are displayed correctly. Previously, the background and label were the same black color, so the text wasn't visible.

We were also applying old, hard coded color properties over the pf-core vars, so the theme had to be cleaned up a bit.

Added the `voronoiDimension` prop to fix the tooltip hover for our bar chart examples. Then, modified the stack chart example to show how similar tooltips are applied without a voronoi container.

Also modified an area chart example, showing how width can be responsive (per @priley86 's request).

Fixes https://github.com/patternfly/patternfly-react/issues/2496
Fixes https://github.com/patternfly/patternfly-react/issues/2482

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/Chart/Chart.tsx
+++ b/packages/patternfly-4/react-charts/src/components/Chart/Chart.tsx
@@ -30,7 +30,7 @@ import { getPaddingForSide } from '../ChartUtils/chart-padding';
  */
 export interface ChartProps extends VictoryChartProps {
   /**
-   * See Victory type docs: https://formidable.com/open-source/victory/docs/victory-area/
+   * See Victory type docs: https://formidable.com/open-source/victory/docs/victory-chart/
    */
   ' '?: any;
   /**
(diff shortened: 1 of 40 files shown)


priley86:
This looks good to me and should resolve my default background issue in #2476 (so I'd be able to use the same theme).

The responsive use case you've added here should help. I'd argue we should try to add those examples for test purposes for each Chart in the future w/ the way these svgs are scaling. I think what's potentially still different about my use case in [35](https://github.com/RedHatInsights/curiosity-frontend/pull/35) is I'm expected to render around 30 x-axis labels responsively (days of the month). When I compress this chart for mobile, I'm seeing some weirdness with the Voronoi tooltip hover, where if bars become close, the first bar hovered is still showing it's tooltip when I move to hover a second bar. I'm not seeing this same buggy behavior happen when I pass the default `ChartTooltip` to the `ChartBar` labelComponent like the example here:
https://formidable.com/open-source/victory/guides/tooltips#simple-tooltips

I will continue to experiment here, but it's just something to test w/ VoronoiContainer in the future. Not sure if you would see the same thing here if trying to render several x-axis labels.

dlabrecq:
Thank you for the reviews!

@karelhala Yes, please use voronoiDimension for your dependent axis. Or, try the approach used by the stack chart example.

@priley86 Although, a voronoi container seems to work with bar charts, it may not be best for the stack chart? (I've also encountered some jumpy hovers.) You could try working with a basic container, instead? I've also modified the stack chart example to show an alternative way to apply tooltips without voronoi container.

Issue #2494, opened 2019-07-10
EmptyStateBody does not allow DOMNesting


rvsia:
EmptyStateBody is wrapper around `<p>` and so we can't use many HTML tags inside the body because of `Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
` etc.

https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateBody.tsx#L17

I propose to change it to `<div>`

tlabaj:
@mcoker thoughts on this one?

mcoker:
@rvsia thanks for the issue! We changed `.pf-c-empty-state__body` from a `<p>` to a `<div>`  in core back in April, and looks like we missed requesting this change to be updated in react. https://github.com/patternfly/patternfly-next/commit/14cc89873991781f26cb2d4f2c916ccbc80d8aae

rvsia:
Thank you for quick response! :+1:

tlabaj:
closed by #2499

PR #2499, opened 2019-07-10
feat(EmptyState): update EmptyStateBody to use <div> instead of <p>

tlabaj:
#2494

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateBody.tsx
+++ b/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateBody.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/EmptyState/empty-state';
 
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateBody.tsx
```diff
@@ -14,7 +14,7 @@ export const EmptyStateBody: React.FunctionComponent<EmptyStateBodyProps> = ({
   className = '',
   ...props
 }: EmptyStateBodyProps) => (
-  <p className={css(styles.emptyStateBody, className)} {...props}>
+  <div className={css(styles.emptyStateBody, className)} {...props}>
```
rebeccaalpert: I think HTMLParagraphElement on line 5 needs to be changed to HTMLDivElement.

tlabaj: nice catch!

Issue #1498, opened 2019-03-05
Charts - provide an accessible title and description


jgiardino:
The charts currently are not accessible to screen reader users. The following css-tricks article describes how charts can be made accessible, by adding a title and description of the contents:
https://css-tricks.com/accessible-svgs/#article-header-id-6

The chart components we provide should include a way for consumers to provide a title and description, and have those strings included in the svg as described in the article.

rachael-phillips:
Hi @dlabrecq, I wanted to make you aware of this issue. Would this be something you would be interested in incorporating in your work on charts?

dlabrecq:
To say that the charts are not accessible is probably not a fair statement. Victory charts support properties for description and title today.

rachael-phillips:
Hi @jgiardino do you think we need to provide the title and description in our examples or is it sufficient that victory charts supports properties for description and title?

jgiardino:
Thanks for the links to the documentation, @dlabrecq! I had looked for similar options for the victory charts but couldn't find them, and assumed that wasn't even supported. So that's awesome that they are!  

I think this issue is still valid in that the argument can be made that our chart examples aren't accessible. I see the following as the two main action items for this issue:
- Verify that `desc` and `title` can be used in our implementation of victory charts. I only did a quick test using the inline editor on our website, but I wasn't able to figure out how to define those in our component.
- Update our chart examples to include `title` and `desc`. Ideally, our examples should represent best practice where possible, since many devs refer to our examples for ideas on how to implement them. These settings are likely to be overlooked if we don't include them. And it also makes it easier to figure out how to set them when they're already defined in our examples.

dlabrecq:
We need to keep this open and add the desc and title properties to the chart examples.

PR #2500, opened 2019-07-11
Charts: provide an accessible title and description

dlabrecq:
Victory provides the ability to add a description and title to SVGs for accessibility. However, users must override `containerComponent` just to set these properties -- it's not clear how to provide accessibility.

In order to make this more intuitive, I've added `ariaDesc` and `ariaTitle` properties to our top level components; `Chart`, `ChartGroup`, etc. The property names we chosen because `ChartDonut` and `ChartDonutUtilization` already have `title` and `subTitle` properties; thus, avoiding a breaking change.

Note that these top-level properties just pass through to the container's `desc` and `title` props. The underlying SVG refers to `desc` and `title` tags via the SVG's aria-labeledby attribute. 

The approach allows users to continue using the `allowZoom` prop of `Chart` and `ChartGroup` without overriding `containerComponent`. If the user must override `containerComponent`, the `allowZoom` property would not be able to provide a default `VictoryZoomContainer` component. 

These top-level properties also help to set `desc` and `title` for `ChartPie`, `ChartDonut` etc. Overriding `containerComponent` would not work in this particular case because the chart is wrapped with a `ChartContainer` component and rendered `standalone=false` -- required to show title, subTitle, and legend within the same SVG. The `desc` and `title` properties must be applied to the outermost container.

Example output:
Screen Shot 2019-07-11 at 2 01 34 AM

Fixes https://github.com/patternfly/patternfly-react/issues/1498


redallen:
Conflicts with #2497 which just went in.

tlabaj:
@dlabrecq can you please link associated issue. Thanks

dlabrecq:
Thanks @jschuler . I've renamed those properties as `ariaDesc` and `ariaTitle`.

packages/patternfly-4/react-charts/src/components/Chart/Chart.tsx
```diff
@@ -349,6 +365,7 @@ export const Chart: React.FunctionComponent<ChartProps> = ({
 
   // destructure last
   theme = getTheme(themeColor, themeVariant),
+  containerComponent = allowZoom ? <VictoryZoomContainer /> : <ChartContainer theme={theme} />,
```
redallen: Can we not pass `desc: ariaDesc, title: ariaTitle,` here and avoid `React.cloneElement(containerComponent as React.ReactElement<any>` later?

dlabrecq: Don't want to modify a given container component directly. I could add the props here, but would be adding it again later in the code

dlabrecq: Didn't want to modify a given containerComponent directly; thus, the clone. Could add props to the defaults while destructuring, but still need to add the same props while cloning the given component. Keeping it simple and following Victory, which also clones the containerComponent, labelComponent, etc.

redallen:
@mattnolting Created relevant issue and followup PR for you.

packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
```diff
@@ -163,6 +167,8 @@ class MultiColorChart extends React.Component {
       <div ref={this.containerRef}>
         <div className="area-chart-legend-bottom-responsive">
           <Chart
+            ariaDesc="Average number of pets"
+            ariaTitle="Donut chart example"
```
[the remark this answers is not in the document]
dlabrecq: Thank you. I'll create a PR to fix the title text.

PR #2505, opened 2019-07-12
chore(deps): remove tippy-react dependency

seanforyou23:
**What**: This PR removes the dependency on @tippy.js/react and brings this functionality in-house as a new helper component called `PopoverBase`. We ran into issues with the latest version of @tippy.js/react (2.x) and decided that best path forward for us is bringing this module in-house. Of course, we're still using tippy.js core so we're basically swapping @tippy.js/react for tippy.js as a dependency.

**Additional issues**: N/A

This is the first of putting back into place work that was abandoned in https://github.com/patternfly/patternfly-react/pull/2326 and https://github.com/patternfly/patternfly-react/pull/2395

There will be follow up PR's for the remaining work - will link them together in the descriptions for easy tracking.


packages/patternfly-4/react-core/src/helpers/PopoverBase/PopoverBase.js
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
```
redallen: Since this is only one file, can we make it Typescript and make a named export `PopoverBase`?

seanforyou23: This is the plan, I'd like to do it in a follow-up PR if that's ok?

seanforyou23: This work has proved tricky for me so I think sending smaller PR's might be good for catching any issues and make the progress a little more steady.

PR #2508, opened 2019-07-12
chore(charts): update chart example aria-titles

redallen:
**What**: Follow-up to #2500

**Additional issues**:


dlabrecq:
This is a dup of https://github.com/patternfly/patternfly-react/pull/2510

Issue #2415, opened 2019-07-01
Unable to click options menu check on the right hand side.


dlabaj:
There's an area between the text and the check that does not respond to click events when a user clicks in that area.  See screenshot below.

<img width="184" alt="Screen Shot 2019-07-01 at 11 18 26 AM" src="https://user-images.githubusercontent.com/6126356/60450229-637eea00-9bf7-11e9-91c7-21d8523547a9.png">
@dlabaj for PR review

boaz0:
IMO it's a problem with core. There are sub elements in the button that are clickable which prevents the event to bubble to the button.

To fix this we need to set something like:

```
.pf-c-button > * {
  pointer-events: none;
}
```

dlabaj:
@mcoker Can you take a look to see if this is a core issue.

mcoker:
@dlabaj weird, I can click anywhere in the button and select something in the options menu. However if I have already selected something in the "single option" example, clicking on the check icon or it's padding toggles the selection. But in the plain, right aligned, top aligned and plain with text examples it's the opposite - clicking on the check icon does nothing, but clicking on the text toggles the selection.

The event should be on the button itself, so clicking anywhere in the button should trigger the selection.

mcoker:
@jenny-s51 we have icons in the [link variation of button](https://patternfly-react.surge.sh/patternfly-4/components/button/) and [App launcher w/ sections and icons](https://patternfly-react.surge.sh/patternfly-4/components/applicationlauncher/). Do you know if those components have the same problem?

jenny-s51:
@mcoker Just checked -- I don't think so! Neither of those seem to have this problem.

mcoker:
@jenny-s51 thanks! Do you know why, and what the difference is as to why the options menu needs it?

jenny-s51:
Closing via #2513

PR #2513, opened 2019-07-12
fix(options menu): update examples to trigger select on the whole item

jenny-s51:
**What**: need feedback on these changes -- this WIP PR is in reference to #2415. Currently, the new click event handler I added seems to work fine on the first three options menu demos, where you can click anywhere on the item to select/deselect it.

However, the last three demos don't seem to work with this new prop,  and when you click directly on the check mark of an item that's been selected, it still doesn't deselect the item. 

Why would this be happening?


jenny-s51:
@jschuler Ah wow, what a magical and simple fix! Now I see the difference between `event.currentTarget`and `event.Target`. Thank you for suggesting this! Will now update this PR to include it.

jenny-s51:
@jschuler Awesome, I appreciate it -- thank you for approving this PR!

PR #2527, opened 2019-07-16
fix(react-table): check empty array in areAllRowsSelected

priley86:
**What**:
If an empty array is passed for `rows` (along with `onSelect` for the "Selectable table" variation), currently the "all rows selection" checkbox in the top left will be set true b/c `rows.every` evaluates true with an empty array.

```
const isSelected = row => row.selected === true; 
let rowsArray = []; 
rowsArray.every(item => isSelected(row));
```
returns `true` when called.

Noted this while testing selection downstream today...

**Additional issues**:


redallen:
There's probably an `isArray()` type deal we can use in case we get passed a weird object with a `length` prop that isn't an array, but this will suffice.

Issue #2528, opened 2019-07-16
Add individual padding vars for donut charts


dlabrecq:
Need to add individual padding vars for each donut chart. (Currently, the donut charts inherit the pie chart's padding property.) That will ensure changes, to the pie chart padding property, do not break donut charts.

Also need to ensure that the donut size is always the minimum of `height` / `width` (default) and the `donutHeight` / `donutWidth` props. This has the potential to cause layout issues depending on the order in which properties are set.

See pf-core vars: https://github.com/patternfly/patternfly-next/issues/2067

PR #2529, opened 2019-07-16
Add individual padding vars for donut charts

dlabrecq:
Added individual padding vars in pf-core for each donut chart. (Currently, the donut charts inherit the pie chart's padding property.) That will ensure changes, to the pie chart padding property, do not inadvertently break donut charts.

Also ensured that the donut size is always the minimum of `height` / `width` (default) and the `donutHeight` / `donutWidth` props. Thus, eliminating the potential to cause layout issues depending on the order in which properties are set.

Minor clean up to docs, etc.

Fixes https://github.com/patternfly/patternfly-react/issues/2528


packages/patternfly-4/react-charts/src/components/ChartUtils/chart-label.ts
```diff
@@ -1,23 +1,24 @@
...
+  center = 'center',
+  right = 'right'
+};
 
 interface ChartLabelPaddingXInterface {
   dx?: number; // Horizontal shift from the x coordinate
   chartWidth: number; // Width of chart (e.g., donut) within SVG
-  labelPosition: string; // Position of label (e.g., bottom, right)
-  legendPosition: string; // Position of legend (e.g., bottom, right)
+  labelPosition?: string; // Position of label (e.g., bottom, right)
```
jschuler: can we type the possible positions out for the positional type props? e.g. 'bottom' | 'right'| etc

redallen: I agree this should be fixed sooner rather than later. We can make a `type Position = 'bottom' | 'right' | 'left' | 'top';`.

dlabrecq: This is just a utility function. The ChartDonut's `legendPosition` and `subTitlePosition` props are already typed with the component.

subTitlePosition?: 'bottom' | 'center' | 'right';
legendPosition?: 'bottom' | 'right';

If I want to continue using the enum below, the utility must use a `string` type.

```
export enum ChartDonutSubTitlePosition {
  bottom = 'bottom',
  center = 'center',
  right = 'right'
}
```

Otherwise, I cannot refer to `ChartLabelPosition.center` in my switch statements. To make that work, I also need to change the code to compare strings (e.g., `'center'`) instead.

That said, the donut component supports multiple legend positions. It's possible that the legendPosition value could be 'bottom', 'bottom-left', 'right', etc. However, there is no where else in `chart-label` utils that we refer to 'bottom-left' or 'right'.

I don't want to maintain unused legend position types here.

The only special legend position for chart-label utils is `bottom`. All other positions use the default position; thus, the `string` type seems appropriate here.

dlabrecq: I've expanded the utility function types as `'bottom' | 'right'` etc., but reverted the switch statements to compare raw strings.

PR #2539, opened 2019-07-18
feat(Tooltip): Added support for aria prop from Tippy library

### LINKED ISSUE #1458: Tooltip - support aria-label
The Tooltip component should support `aria-labelledby` in addition to `aria-describedby`.

For example, a button with an icon might display a tooltip with the label for that icon. This text should be announced as the button label, and not the description.

A fix for tippyjs to support this was included in https://github.com/atomiks/tippyjs/pull/393

rebeccaalpert:
Added aria prop for aria-describedby or aria-labelledby support from the Tippy library. Default is library default.

Fixes #1458.


redallen:
@rebeccaalpert Conflicts :(

Issue #1657, opened 2019-03-28
No exported member 'compoundExpand' for react-table


ibolton336:
![Screen Shot 2019-05-22 at 3 28 22 PM](https://user-images.githubusercontent.com/11218376/58202964-fba7bc00-7ca6-11e9-8a9b-4619301af8ec.png)

ibolton336:
FIxed as of 4-17

ibolton336:
![Screen Shot 2019-05-22 at 3 28 22 PM](https://user-images.githubusercontent.com/11218376/58202908-cdc27780-7ca6-11e9-8af8-295099958698.png)

ibolton336:
My original issue was resolved, but now I am having an issue importing the compountExpand module.

rachael-phillips:
Hi @jcaianirh are we ready to close this issue?

jcaianirh:
yes! thanks this is closed with the mentioned merged pr.

PR #2554, opened 2019-07-19
fix(react-table): Add exports for compoundExpand.

jcaianirh:
**What**: Fix for issue: https://github.com/patternfly/patternfly-react/issues/1657

**Additional issues**:

Add exports for compoundExpand.

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/index.d.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/index.d.ts
@@ -19,5 +19,5 @@ export { default as TableHeader, HeaderProps } from './Header';
 export { default as TableBody, TableBodyProps } from './Body';
 export { default as RowWrapper, RowWrapperRow, RowWrapperProps } from './RowWrapper';
 export { default as ExpandableRowContent } from './ExpandableRowContent';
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-table/src/components/Table/index.d.ts
```diff
@@ -19,5 +19,5 @@ export { default as TableHeader, HeaderProps } from './Header';
 export { default as TableBody, TableBodyProps } from './Body';
 export { default as RowWrapper, RowWrapperRow, RowWrapperProps } from './RowWrapper';
 export { default as ExpandableRowContent } from './ExpandableRowContent';
-export { sortable, headerCol, cellWidth, ISortable, expandable, isRowExpanded } from './utils';
+export { sortable, headerCol, cellWidth, ISortable, expandable, isRowExpanded, compoundExpand } from './utils';
```
redallen: This seems close to `export * from './utils';`

Either way, the TS conversion in #2360 will eventually resolve this too.

priley86: +1... i am planning to circle back and ensure we can do `export *` on all exports, and that they are converted from default exports to named exports... will probably attempt this towards the end to be certain. I've gone ahead and rebased this change for now though!

PR #2558, opened 2019-07-22
fix(PF4: Pagination): When 0 or negative number is passed to Pagintion show 0 pages

### LINKED ISSUE #2557: 0 or negative number of items for pagination
When consumer passes 0 or negative number to pagination user is still able to change page and can see 1 page out of 0.

Screenshot from 2019-07-22 14-05-25

karelhala:
**What**:

When there is 0 items or negative number for pagination show 0th page out of 0 and disable paginating.
Screenshot from 2019-07-26 10-09-40

fix #2557

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
@@ -56,7 +56,7 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
   static defaultProps = {
     className: '',
     lastPage: 0,
-    firstIndex: 0,
+    firstPage: 0,
     pagesTitle: '',
     toLastPage: 'Go to last page',
     toNextPage: 'Go to next page',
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -54,6 +56,7 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
   static defaultProps = {
     className: '',
     lastPage: 0,
+    firstIndex: 0,
```
redallen: Don't think we need this here. Did you mean `firstPage`?

packages/patternfly-4/react-core/src/components/Pagination/Pagination.md
```diff
@@ -36,7 +36,7 @@ class PaginationTop extends React.Component {
   render() {
     return (
       <Pagination
-        itemCount={523}
+        itemCount={0}
```
redallen: Perhaps create another example?

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
karelhala: Yup, at first I passed `firstIndex` and renamed it to `firstPage` good catch!

karelhala:
@mcarrano good point! I disabled all navigation elements including per page (it doesn't make any sense to change per page if server returns 0 items) and fixed that firefox issue by adding correct min value to input.

Issue #2066, opened 2019-05-23
Select w. typeahead breaks search on tab


SpyTec:
If you have a Select component with typeahead and start typing, it displays the options as expected. Clicking tab and then shift-tab to go back however doesn't expand the list

In fact, this issue happens any time you tab to get to the text input rather than clicking

Version: `"@patternfly/react-core": "^3.16.10"`

**Repro steps:**
1. Have a Select component with typeahead variant that's not in focus
1. Tab until you get to text input in the Select component
1. Start typing for a result

**Expected result:**
Options list is expanded

**Actual result:**
Nothing happens

redallen:
Is this still the case in https://2013-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/select/ ?

SpyTec:
@redallen Yes, the examples there have the same bug

tlabaj:
closed by #2572

PR #2560, opened 2019-07-22
fix(select): select menu now opens while typing input

jenny-s51:
**What**: in regards to issue #2066. Select menu now opens immediately as input is entered.


jenny-s51:
This will be closed via #2561

Issue #1936, opened 2019-05-06
Updates for RedHat Typography


dlabaj:
Add updates for new RedHat typography.

- [x] Update README with information on how to pull in the fonts.
- [x] Update examples to use the new font.
- [ ] Update demo app to use the new font.
- [x] Bump core to get these updates once available.

tlabaj:
Closed by #2563

PR #2563, opened 2019-07-22
chore(react-docs): updated react-docs to use Red Hat font

tlabaj:
#1936

**What**:

**Additional issues**:


packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
...
+      <head>
+        <meta charSet="utf-8" />
+        <meta httpEquiv="x-ua-compatible" content="ie=edge" />
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1, shrink-to-fit=no"
+        />
+        {props.headComponents}
+      </head>
+      <body {...props.bodyAttributes} className="pf-m-redhat-font">
```
redallen: It's possible to globally include SASS that adds that styling to `body`, but I like this better to give us more control of the template in the future.

packages/patternfly-4/react-core/README.md
```diff
@@ -49,6 +49,10 @@ export default <Button variant="primary">Button</Button>;
 
 All css related to each component is provided within it. There is no component level CSS to import.
 
+# Applying Red Hat Fonts (Optional) 
+If you would like to add Red Hat font instead of Overpass, simply add the class `.pf-m-redhat-font` to an element that wraps your application (ideally `<html>` or `<body>`) to adopt the CSS changes that introduce the Red Hat font into PatternFly.
```
mcoker: Small change, wdyt?

```suggestion
If you would like to use Red Hat fonts instead of Overpass, simply add the class `.pf-m-redhat-font` to an element that wraps your application (ideally `<html>` or `<body>`) to adopt the CSS changes that introduce the Red Hats font into PatternFly.
```

packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
...
+          id="___gatsby"
+          dangerouslySetInnerHTML={{ __html: props.body }}
+        />
+        {props.postBodyComponents}
+      </body>
+    </html>
+  )
+}
+
+HTML.propTypes = {
```
redallen: The linter is complaining about these, I'd just remove them instead of giving null defaults. Gatsby will always populate them.

packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import PropTypes from "prop-types"
```
redallen: Remove this import.

packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
+import React from "react"
```
redallen: Use single quotes.

packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
...
+        />
+        {props.headComponents}
+      </head>
+      <body {...props.bodyAttributes} className="pf-m-redhat-font">
+        {props.preBodyComponents}
+        <noscript key="noscript" id="gatsby-noscript">
+          This app works best with JavaScript enabled.
+        </noscript>
+        <div
+          key={`body`}
```
redallen: Change to `key="body"`

packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import PropTypes from "prop-types"
+
+export default function HTML(props) {
+  return (
+    <html {...props.htmlAttributes}>
```
redallen: Add `lang="en"`

packages/patternfly-4/react-docs/src/html.js
```diff
@@ -0,0 +1,39 @@
...
+      <head>
+        <meta charSet="utf-8" />
+        <meta httpEquiv="x-ua-compatible" content="ie=edge" />
+        <meta
```
redallen: Make one line.

packages/patternfly-4/react-docs/src/html.js
dgutride: We are using react-helmet in other places to alleviate the need for this - you can add this instead (only the body element needs to be added, other code added for reference so you can find it):

in react-docs/src/templates/sidebarLayout.js 
```
<Helmet>
  <html lang="en-US" />
  <meta charSet="utf-8" />
  <meta name="description" content="PatternFly React Documentation" />
  <meta name="keywords" content="React, PatternFly, Red Hat" />
  <link rel="icon" type="image/x-icon" href="/favicon.ico" />
  <body className='pf-m-redhat-font' />
</Helmet>
```

dgutride: I think for now we don't really need to add a template here, this docs site isn't going to last forever anyway once we can move to just using pf.org.

packages/patternfly-4/react-core/README.md
mcoker: oops. I made a typo 

```suggestion
If you would like to use Red Hat fonts instead of Overpass, simply add the class `.pf-m-redhat-font` to an element that wraps your application (ideally `<html>` or `<body>`) to adopt the CSS changes that introduce the Red Hat fonts into PatternFly.
```

packages/patternfly-4/react-docs/src/html.js
tlabaj: I followed these instructions from the Gatsby documentations.  Are you saying that I do not need to copy the html.js file?
https://www.gatsbyjs.org/docs/custom-html/

dgutride: No - the html.js file isn't needed for just adding a class on the body tag, react-helmet supports that with a single line change to sidebarLayout

PR #2549, opened 2019-07-19
Bump @patternfly/patternfly to 2.23.0, add promotion script, add bumping script

redallen:
**What**: Closes #2336 .

**Additional issues**: Still requires manually running each command for now to build some trust around a sensitive process.

PR #2572, opened 2019-07-23
 fix(select): select menu has consistent behavior on input field 

### LINKED ISSUE #2561: Select menu has incorrect closing behavior
This is an extension to issue #2066 , and PR #2560 

See example below. Currently, the select menu is hidden when the input is initally focused, but backspacing to an empty input leaves the menu open.

ezgif com-video-to-gif

jenny-s51:
**What**: Closes #2561. Select menu now has consistent, open behavior on empty text input field. This is the expected default behavior per @mcarrano's feedback.


jenny-s51:
@mcarrano this is because the react workspace is currently using an outdated version of core. The input field issue was fixed here by @mcoker https://github.com/patternfly/patternfly-next/pull/2075, and is in a newer core release. As soon as @redallen 's PR here is merged, the input field will work as expected. https://github.com/patternfly/patternfly-react/pull/2549

PR #2574, opened 2019-07-23
feat(Modal.tsx): Add targetDocument to Modal to allow showing the Modal in another window

### LINKED ISSUE #2551: Modal PF4: cannot render on specified element
[Potential enhancements]

I'm trying to write a PF4 Modal, then run ReactDOM.render(modal, target), with target being the container that modal would be rendered in (similar to [link](https://github.com/oVirt/ovirt-engine-ui-extensions/blob/83e676c701ffc7696eae1a7d073facbbeec2a6ba/src/integrations/showClusterUpgrade.js#L95)). However, the modal is rendered in the iframe document body, but not in the specified container. 

In PF3, it was possible because Modal accepted a container props. [link](https://github.com/react-bootstrap/react-overlays/blob/7dc63e8e51b2ede669801ead8453645e9311394b/src/Modal.js#L464-L470)

Would it be possible to implement a new props like in PF3 instead of defaulting to `document.body` as [currently implemented](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx#L84)?
@sjd78 @mturley

sjd78:
Current implementation will not allow use of the Modal when the running
code is within an iframe. In order to allow the Modal to escape the
iframe, which is probably invisible, the iframe's parent window's
document must be used by the Modal itself. With the change, the Modal
will be able to create and manage the React portal within another frame.

Fixes #2551

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
@@ -17,6 +17,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   isOpen?: boolean;
   /** Complex header (more than just text), supersedes title for header content */
   header?: React.ReactNode,
+  /** document where the Modal will be created and shown (via React.Portal) */
+  targetDocument?: HTMLDocument;
   /** Simple text content of the Modal Header, also used for aria-label on the body */
   title: string;
   /** Flag to hide the title */
@@ -50,6 +52,7 @@ export class Modal extends React.Component<ModalProps, ModalState> {
   static defaultProps = {
     className: '',
     isOpen: false,
+    targetDocument: window.document,
     hideTitle: false,
     showClose: true,
     ariaDescribedById: '',
@@ -77,7 +80,8 @@ export class Modal extends React.Component<ModalProps, ModalState> {
   };
 
   toggleSiblingsFromScreenReaders = (hide: boolean) => {
-    const bodyChildren = document.body.children;
+    const { targetDocument } = this.props;
+    const bodyChildren = targetDocument.body.children;
     for (const child of Array.from(bodyChildren)) {
       if (child !== this.state.container) {
         hide ? child.setAttribute('aria-hidden', '' + hide) : child.removeAttribute('aria-hidden');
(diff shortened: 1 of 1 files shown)


sjd78:
@mturley - Thanks, this should work well for Modals as long as the code that runs the react app also clones the CSS link and style blocks from the iframe up to the targetDocument's window.  I'm already doing that in ovirt-engine-ui-extensions for PF3's Wizard.

The `build_pf4_docs` error seems to really be because of [GatsbyJS code that runs in node](https://www.gatsbyjs.org/docs/debugging-html-builds/#how-to-check-if-code-classlanguage-textwindowcode-is-defined) without something like jsdoc attached to give it the browser APIs.  I think it is specifically for the gatsby-node phase of the docs creation, but I didn't get much further than that.

So seems like to be able to step around the exact error that's coming up, either do what you suggested and move the "real" default value to a function that doesn't get looked at or run during the gatsby-node phase, or somehow patch jsdom into gatsby for the node phase to make browser API entities usable as defaults.

Or or, setup a React Context to hold the targetWindow/targetDocument (defaulted to standard window / window.document) and pull it from there.  That has the added bonus of being able to be pushed to the other 10ish components that use window or document directly and make them compatible with iframe escaping.  (We all want to escape iframes, just can't always be done :wink:).

mturley:
I like the idea of being able to provide these values in a React Context, just as long as the components behave normally and use the window/window.document defaults when the context is not present (we shouldn't require the majority of users who are not going to change those values to add a Context.Provider to their tree). Which would mean the defaults would still have to be defined at the component level, I think. So we'd have to combine that with one of the other two solutions.

I think patching jsdom into Gatsby sounds like a can of worms, but if you feel like figuring that out, it sounds like it might be nice to have. I wouldn't be surprised if Gatsby has a configuration option somewhere for that.

Since we already call `document.*` only at browser runtime, I think making sure the default value is defined at browser runtime sounds like the easiest solution. I'm happy with all of your ideas though!

packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
```diff
@@ -17,6 +17,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   isOpen?: boolean;
   /** Complex header (more than just text), supersedes title for header content */
   header?: React.ReactNode,
+  /** document where the Modal will be created and shown (via React.Portal) */
+  targetDocument?: HTMLDocument;
```
jschuler: can we change this prop name to `appendTo` since that is a prop we already use in some components?
And then I would make the type more generic
`appendTo?: Element | ((ref: Element) => Element);`

packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
```diff
@@ -50,6 +52,7 @@ export class Modal extends React.Component<ModalProps, ModalState> {
   static defaultProps = {
     className: '',
     isOpen: false,
+    targetDocument: window.document,
```
jschuler: default to something like `document.body` then we don't have to hardcode `body` below and also give user more flexibility to append it anywhere

jschuler: Also probably have to check first that `window` (or `document` if you change it to `document.body` is not undefined, i.e.
`appendTo: typeof document !== 'undefined' && document.body`

jschuler:
@sjd78 Thanks for this PR. I worked with @thaorell to take this idea a little further and make it even more generic in https://github.com/patternfly/patternfly-react/pull/2802. I'll close this PR. Thank you for your contributions!

Issue #2573, opened 2019-07-23
Update charts to use Red Hat font


dlabrecq:
When the `pf-m-redhat-font` selector is added to the page, the Red Hat font is expected to be used instead of the default Overpass font. However, the charts are not picking up this change.

This is due to the fact that the charts are outputting the font value instead of a PatternFly variable.

(Victory typically requires values for colors, pixels, etc. The font size; for example, must be output as a value because Victory concatenates that with 'px'. However, this is one area where we can output a var.)

The issue can be resolved by changing the topology vars in the base theme to output a variable instead of the font value, like so:

![Screen Shot 2019-07-23 at 2 09 58 PM](https://user-images.githubusercontent.com/17481322/61736492-33d58480-ad54-11e9-95e1-730ce84564ae.png)

PR #2584, opened 2019-07-24
fix(charts): use vars to enable Red Hat fonts [ci skip]

dlabrecq:
When the `pf-m-redhat-font` selector is added to the page, the Red Hat font is expected to be used instead of the default Overpass font.

This change ensures that charts and its labels use the Red Hat font as expected.

Fixes https://github.com/patternfly/patternfly-react/issues/2573

Screen Shot 2019-07-24 at 9 50 40 AM


packages/patternfly-4/react-charts/src/components/ChartLabel/__snapshots__/ChartLabel.test.tsx.snap
```diff
@@ -7,7 +7,9 @@ exports[`Chart 1`] = `
   lineHeight={1}
   style={
     Object {
-      "fontFamily": "overpass, overpass, open sans, -apple-system, blinkmacsystemfont, Segoe UI, roboto, Helvetica Neue, arial, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol",
+      "fontFamily": "var(--pf-chart-global--FontFamily)",
+      "fontSize": 14,
```
mcoker: we have vars for the font sizes used in charts - this one is `--pf-chart-global--FontSize--sm`

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/_chart-globals.scss#L55-L61

packages/patternfly-4/react-charts/src/components/ChartLabel/__snapshots__/ChartLabel.test.tsx.snap
```diff
@@ -22,7 +24,9 @@ exports[`Chart 2`] = `
...
```
mcoker: same here

packages/patternfly-4/react-charts/src/components/ChartLabel/__snapshots__/ChartLabel.test.tsx.snap
```diff
@@ -37,7 +41,9 @@ exports[`renders component text 1`] = `
...
```
mcoker: same here

packages/patternfly-4/react-charts/src/components/ChartLabel/__snapshots__/ChartLabel.test.tsx.snap
dlabrecq: Victory typically requires raw values for colors, pixels, etc. The font size; for example, must be output as a value because Victory concatenates that with 'px'. 

The font family and letterSpacing are one of the few places where we can output a variable.

mcoker: Should we remove the unit from these vars so they're usable? We do that with other properties like padding, margin, width, height, etc.

dlabrecq: There is no unit in our pf-core chart vars, it's just a raw number.

Victory is appending the unit automatically. Thus, a CSS variable breaks for properties like `fontSize`, `padding`, etc.

I've created an issue with Victory, asking they change that.
`https://github.com/FormidableLabs/victory/issues/1357`

dlabrecq: There is no unit in our pf-core chart vars for this very reason, it's just a raw number.

Victory is appending the unit automatically. Thus, a CSS variable breaks for properties like `fontSize`, `padding`, etc.

I've created an issue with Victory, asking they change that. Although, I suspect that won't be possible because they use the raw values to calculate heights & widths of labels, legends, etc.
`https://github.com/FormidableLabs/victory/issues/1357`

Issue #2315, opened 2019-06-20
Disabled variation of pagination component


mcoker:
Once we complete the work needed in core and add an example in https://github.com/patternfly/patternfly-next/issues/1963, let's offer a variation of the pagination in a disabled state.

This is related to @boaz0's issue https://github.com/patternfly/patternfly-next/issues/1785

rachael-phillips:
@mcoker are we ready to get started on this on the React side?

mcoker:
@rachael-phillips yep!

PR #2586, opened 2019-07-24
feat(Pagination): add disabled flag for whole component

kmcfaul:
**What**: Adds 'isDisabled' flag to disable the whole pagination component.

**See issue**: #2315


kmcfaul:
> The options menu toggle isn't aligning correctly

Am I missing a class?

packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.tsx
```diff
@@ -41,16 +43,17 @@ export const OptionsToggle: React.FunctionComponent<OptionsToggleProps> = ({
   showToggle = true,
   onToggle = (_isOpen: boolean) => undefined as any,
   isOpen = false,
+  isDisabled = false,
   parentRef = null,
   toggleTemplate: ToggleTemplate = '',
 }:OptionsToggleProps ) => {
   return (
-    <div className={css(styles.optionsMenuToggle, getModifier(styles, 'plain'), getModifier(styles, 'text'))} >
+    <div className={css(styles.optionsMenuToggle, isDisabled && getModifier(styles, 'disabled'), getModifier(styles, 'plain'), getModifier(styles, 'text'))} >
```
jschuler: nit: it generally makes more sense to use getModifier if a variable prop can be matched to the style object. In this case i would change these to `styles.disabled`, `styles.plain`, etc..

packages/patternfly-4/react-core/src/components/Pagination/Pagination.md
```diff
@@ -87,6 +87,48 @@ class PaginationBottom extends React.Component {
 }
 ```
 
+## Pagination disabled
+```js
+import React from 'react';
+import { Pagination, PaginationVariant } from '@patternfly/react-core';
+
+class PaginationTop extends React.Component {
```
jschuler: nit: class name should be `PaginationDisabled`. Doesn't affect the rendering but shows up in the sample code

packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.tsx
kmcfaul: Ah I thought that was just how we retrieved modifier styles. I can change it to `styles.modifiers.disabled` and so on.

Issue #1658, opened 2019-03-29
Allow managed sidebar to default to closed


srambach:
Possibly related to #1311. I would like to use a managed sidebar, but just want to be able to specify that it is always closed initially.

rachael-phillips:
Should be completed in unison with #1311 based on conversation with @tlabaj

srambach:
Still would like to have this.

tlabaj:
closed by #2608

PR #2608, opened 2019-07-29
feat(Page): Allow the nav to default to closed for managed sidebar

tlabaj:
#1658

**What**:  Allow the nav to default to closed for managed sidebar

**Additional issues**:

Also fixed Page integration test to be more specific about sidebar that is being selected.


packages/patternfly-4/react-core/src/components/Page/Page.tsx
```diff
@@ -31,6 +31,10 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
    * the sidebar component or add a callback onNavToggle function into the PageHeader component
    */
   isManagedSidebar?: boolean;
+  /**
+   * If true, the managed the sidebar is initially open for desktop view
+   */
+  isManagedSidebarOpen?: boolean;
```
jschuler: It might be clearer that this sets the default behavior by naming the prop `defaultManagedSidebarOpen` instead. Otherwise it would seem that you're able to change the prop programmatically after the page is already created.

packages/patternfly-4/react-core/src/components/Page/Page.tsx
```diff
@@ -31,6 +31,10 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
    * the sidebar component or add a callback onNavToggle function into the PageHeader component
    */
   isManagedSidebar?: boolean;
+  /**
+   * If true, the managed the sidebar is initially open for desktop view
```
jschuler:
```suggestion
   * If true, the managed sidebar is initially open for desktop view
```

packages/patternfly-4/react-core/src/components/Page/Page.tsx
```diff
@@ -61,10 +60,23 @@ export class Page extends React.Component<PageProps, PageState> {
...
+    isManagedSidebarOpen: true,
     onPageResize: ():void => null,
     mainContainerId: null as string
   };
 
+  constructor(props: PageProps) {
+    super(props);
+
+    const {isManagedSidebar, isManagedSidebarOpen} = props;
+    const managedSidebarOpen = !isManagedSidebar ? true : isManagedSidebar && isManagedSidebarOpen;
```
jschuler:
```suggestion
    const managedSidebarOpen = !isManagedSidebar ? true : isManagedSidebarOpen;
```

PR #2612, opened 2019-07-30
feat(Select): Adds the ability to pass in a user defined object

### LINKED ISSUE #2045: pf4 Select and Multi select component API
First of all thank you for adding Multi select variant of select component.

I am interested who is in charge of design of API for PF4 components. With whom can I get in touch? @rachael-phillips 

There seems to be a gap between the designs and actual use cases. In order to use these in Insights apps we would probably need some modifications as the current solution does not work too well.

I have a few suggestions about about the component API. 
- options format
  - what if i want the option label to be different from option value? Sometimes we also need to have additional information in the options
  - currently we would have to iterate over all options and search the options object based on 
  - i would propose something like: 
  - The on change would return whole object with all values.
  - currently its also impossible to use this with i18n. We would have to translate the strings back to its original value in order to make it work with back end.
```
{ label: 'New York', value: 'NY', something: 'else' }
```

- options component
  - what if i want to use custom component for options?
  - For instance add icon to option. I know you have checkbox variant so i think its possible to allow users to add their own components. We could have a wrapper which would handle all the events and renders its children.
```JSX
<SelectOptionWrapper>
  {(options) => <div>{options.label} <SomeIcon /></div>}
</SelectOptionWrapper>
```

I think we can use [react-select](https://react-select.com/home) for inspiration. We could even just create custom styling that would restyle react select to be PF4 compatible.
 
link to docs: http://patternfly-react.surge.sh/patternfly-4/components/select/

dlabaj:
This enhancement allows a user to now pass in a user defined object to store additional data besides just the string value to a select option.  The object must have a toString function that is
responsible for returning the the localized string.

fix #2045

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.md
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.md
@@ -518,12 +518,12 @@ class MultiTypeaheadSelectInputCustomObjects extends React.Component {
     }
   }
     this.options = [
-      { value: this.createState('Alabama', 'AL', 'Montgomery', 1846), disabled: false },
-      { value: this.createState('Florida', 'FL', 'Tailahassee', 1845), disabled: false },
-      { value: this.createState('New Jersey', 'NJ', 'Trenton', 1787), disabled: false },
-      { value: this.createState('New Mexico', 'NM', 'Santa Fe', 1912), disabled: false },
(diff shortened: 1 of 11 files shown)


kmcfaul:
Echoing adding the tests, otherwise looks great

PR #2615, opened 2019-07-31
feat(Table): Add the wrap transform

### LINKED ISSUE #2578: Add wrap prop for <td> and <th> cells in a <thead>
I don't see support for the `.pf-m-wrap` class you can add to cells of the `<thead>` element to allow the content in the cells to wrap. If we don't support that, we should add it, and also add an example. We recently added an example to the core workspace here - https://github.com/patternfly/patternfly-next/pull/2065

boaz0:
**What**:

closes #2578 

Add the wrap modifier to react-table through transforms.

//cc @karelhala @mcoker


boaz0:
Thank you @redallen & @karelhala :smiley:

PR #2628, opened 2019-08-01
feat(ExpandCollapse): Add onToggle callback 

### LINKED ISSUE #2610: PF3 ExpandCollapse missing onToggle callback
The `Expandable` component in PF4 provides `onToggle` callback, however the PF3 `ExpandCollapse` component has no such callback, which makes it impossible to use when the expanded part triggers a re-render of a parent component. 

I can submit a PR for this, let me know if such change is welcomed.

shaded-enmity:
So that state of the component can be controlled externally

This is to bring feature parity with PF4. Also this makes it possible to keep the UI consistent if
the children of ExpandCollapse trigger a rerender of the component containing the ExpandCollapse.

Fix #2610

**What**: Add onToggle callback for ExpandCollapse so that the expanded state can be controlled externally.

**Additional issues**: #2610

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/ExpandCollapse/ExpandCollapse.js
+++ b/packages/patternfly-3/patternfly-react/src/components/ExpandCollapse/ExpandCollapse.js
@@ -24,7 +24,11 @@ class ExpandCollapse extends React.Component {
   }
 
   onClick = () => {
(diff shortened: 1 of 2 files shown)


kmcfaul:
@shaded-enmity If you rebase that should resolve the test error I think.

Looks good code wise.

shaded-enmity:
@kmcfaul Thanks, it's all green now :tada:

PR #2634, opened 2019-08-02
fix(docs): Fix a11y issues in Documentation

jessiehuff:
Fix accessibility issues in Accordion, ClipboardCopy, and DataList

Fixes https://github.com/patternfly/patternfly-org/issues/1311 
Fixes https://github.com/patternfly/patternfly-org/issues/1310 
Fixes https://github.com/patternfly/patternfly-org/issues/1305


jessiehuff:
Hey Jenny, thanks for taking a look! So from what I can tell, I believe in org we have AutoLinkHeaders that set the id. So in the mdxPF4Template.js and index.js there is an AutoLinkHeader that generates the size, className, id, content, etc. It does this from the AutoLinkHeader.js file. For the other PR I had on unique org ids, I had to pass a suffix to the AutoLinkHeader to add that to the id and  differentiate them. With this issue, it looked like it was with what was being pulled from React so I changed the aria-controls in ClipboardCopyToggle.tsx. That's been my thinking up until this point anyway. :)

jenny-s51:
@jessiehuff That makes sense. Thank you for clarifying! Just out of curiosity (and for future reference), is `aria-controls` somehow modifying the value of the `id` there?

packages/patternfly-4/react-core/src/components/DataList/DataList.md
```diff
@@ -476,7 +476,7 @@ class ExpandableDataList extends React.Component {
               dataListCells={[
                 <DataListCell isIcon key="icon">
                   <CodeBranchIcon />
-                </DataListCell>,
```
jschuler: looks like the comma should be here

jessiehuff:
@jenny-s51 `aria-controls` provide a relation between a parent element and a child element (the toggle's button in this case). The reason I think it needed to be changed here is because otherwise we get the issue: "ARIA attributes must conform to valid values" basically pointing out that we need to pass the correct information to screen readers. This [article](https://accessibilityresources.org/aria-controls) seems to give a good summary of `aria-controls.` :)

PR #2638, opened 2019-08-04
docs(Dropdown): fix a11y issues discovered by axe

### LINKED ISSUE #2581: a11y issue in documentation/react/dropdown
only need to fix the "buttons must have discernible text" issue. 

https://github.com/patternfly/patternfly-org/issues/1312

To find these errors / test if your solution works:

1. Right click on the page and choose inspect
2. In developer tools, choose "axe" from the top menu 
3. Run the test

boaz0:
**What**:

closes #2581 

//cc @lareplog @christiemolloy

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ApplicationLauncher/__snapshots__/ApplicationLauncher.test.js.snap
+++ b/packages/patternfly-4/react-core/src/components/ApplicationLauncher/__snapshots__/ApplicationLauncher.test.js.snap
@@ -140,7 +140,9 @@ exports[`ApplicationLauncher custom icon 1`] = `
                 id="pf-toggle-id-5"
                 type="button"
               >
-                <span>
+                <span
+                  aria-label="Application launcher"
+                >
                   <svg
                     aria-hidden="true"
                     fill="currentColor"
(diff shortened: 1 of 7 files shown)


jgiardino:
Thanks for working on this, @boaz0! 

Similar updates were made in https://github.com/patternfly/patternfly-react/pull/2517 and when I review the preview link for that PR, the dropdown component and examples look as I expect them to.  But when I look at the current example (i.e. the one we have now, before this PR), the `aria-label` attribute is missing on the App launcher example. And then the updates in this example aren't quite what we want, either. 

Maybe some updates were made to the Dropdown component since #2517 that caused a regression? 

Should we try to figure out what changes were made after #2517 that caused the regression, instead of moving forward with this PR? Or should we continue moving forward with updates in this PR? I'm not sure which is the easier path forward, but please let me know if further clarification is needed.

boaz0:
@jgiardino I actually think you're right. I will take another look at it. I didn't know it is a regression.

jgiardino:
@boaz0 Apologies for the unnecessary work here. :-(  We realized a regression happened when reviewing this PR.

These are some other PRs related to this work:
- Originally the ability to specify an aria-label on a dropdown toggle was included in https://github.com/patternfly/patternfly-react/pull/2517
- Then there was a regression that happened during a Typescript conversion. I don't know that PR number, but this PR fixes that: https://github.com/patternfly/patternfly-react/pull/2690

tlabaj:
Closing this PR since regression is fixed by PR #2690

PR #2639, opened 2019-08-05
fix(Pagination): Fix pagination to update page count

### LINKED ISSUE #2629: PF4 Pagination doesn't update page count
On the PF4 `Pagination` component, updating the `page` prop doesn't cause the page number text field to update.

When I update the page number from an external event such as a button press, this is what happens:

ScreenRecording2019-08-01at1

Code used for this:

```
class Test extends React.Component<{}, { page: number }> {
    constructor(props) {
        super(props);
        this.state = { page: 1 };
    }
    render() {
        return (
            <div className='body' style={{ display: 'flex' }}>
                <Pagination
                    itemCount={10000}
                    perPage={10}
                    page={this.state.page}
                />
                <button
                    onClick={() => this.setState({ page: this.state.page + 1 })}
                >
                    Increase Page
                </button>
            </div>
        );
    }
}
```

jessiehuff:
Fixes #2629

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
@@ -93,6 +93,13 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
     }
   }
 
+  componentDidUpdate() {
+    //Typical usage, don't forget to compare the props
+    if (this.state.userInputPage !== this.props.page) {
+      this.setState({ userInputPage: this.props.page });
+    }
+   }
+
   render () {
     const {
       page,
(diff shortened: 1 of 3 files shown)


AllenBW:
The only issue with this is... now yah cant change the date via input :(

jessiehuff:
@AllenBW I was trying to look into this, but I'm not positive I know what you mean. Can you explain further or show me what's happening for you? I'll definitely work on whatever is going on! :)

AllenBW:
@jessiehuff  hi!! I believe its the same issue we see in the test failure, if i drop into the PatternFly-React preview and attempt to change the page via the input (rather than the flanking page buttons) there is no change in page value 😬

AllenBW:
https://github.com/patternfly/patternfly-react/pull/2417 Is definitely the work that resulted in this issue, another fe (in addition to insights) that demonstrates this issue is `https://cloud.redhat.com/beta/rhel/vulnerability/cves`  so you'll update the top pagination and the bottom one will not reflect the changes

jessiehuff:
Hey @AllenBW! I think I got it figured out. :) Sorry it took me so long, I had issues with my development environment that I had to solve first haha. Thanks for pointing out the issue!

AllenBW:
WOOHOO!! thanks for effort! lookin forward to incorporating this fix 😋 💃🏾🌺

PR #2640, opened 2019-08-05
feat(react-table): Move from lodash-es to lodash

aljesusg:
Fix #2616 

Remove lodash-es by lodash

PR openshift migration : https://github.com/openshift/console/pull/1724

Impact: Kiali https://github.com/kiali/kiali-ui/pull/1319

cc @priley86

PR #2641, opened 2019-08-05
chore(table): replace lodash-es with lodash

### LINKED ISSUE #2616: [PF4] lodash-es problems in React-table
Hi

After add the react-table for the VirtualScroll in kiali we have a problem with lodash-es library imported in react-table.

Why react-table use lodash-es instead lodash like react-charts?
```bash
  This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/aljesusg/Projects/kiali/kiali-ui/node_modules/@patternfly/react-table/node_modules/lodash-es/lodash.js:10
    export { default as add } from './add.js';
    ^^^^^^

    SyntaxError: Unexpected token export

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:471:17)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:513:25)
      at Object.<anonymous> (node_modules/@patternfly/react-table/dist/js/components/Table/base/evaluate-transforms.js:8:17)
```

Thanks

### LINKED ISSUE #2617: Replace lodash-es with lodash
We have no reason to depend twice on the same library. We should replace *`lodash-es` with `lodash`.

redallen:
**What**: Fixes #2617 Fixes #2616

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-virtualized-extension/package.json
+++ b/packages/patternfly-4/react-virtualized-extension/package.json
@@ -40,7 +40,6 @@
   },
   "peerDependencies": {
     "@patternfly/react-table": "^2.10.0",
-    "lodash-es": "4.x",
     "prop-types": "^15.6.1",
     "react": "^16.4.0",
     "react-dom": "^15.6.2 || ^16.4.0"
--- a/jest.config.js
+++ b/jest.config.js
@@ -33,9 +33,9 @@ module.exports = {
   testPathIgnorePatterns: [
     '<rootDir>/scripts/generators/',
     '<rootDir>/packages/patternfly-4/react-integration/',
-    '<rootDir>/node_modules/(?!lodash-es/.*)'
+    '<rootDir>/node_modules/(?!lodash/.*)'
   ],
-  transformIgnorePatterns: ['node_modules/(?!@patternfly|@novnc|tippy.js|lodash-es)'],
+  transformIgnorePatterns: ['node_modules/(?!@patternfly|@novnc|tippy.js|lodash)'],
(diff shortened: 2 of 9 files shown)


packages/patternfly-4/react-virtualized-extension/package.json
```diff
@@ -40,7 +40,6 @@
   },
   "peerDependencies": {
     "@patternfly/react-table": "^2.10.0",
-    "lodash-es": "4.x",
```
priley86: should `lodash` be added as a `peerDependency` now?

redallen: `react-virtualized-extension` doesn't use `lodash-es` anywhere.

priley86:
this is nearly identical to #2640  now... let's pick one 😄

aljesusg:
This Fix #2616 too.

aljesusg:
Hooray !!! Thanks

PR #2645, opened 2019-08-05
fix(formselect): adds isRequired prop

### LINKED ISSUE #1397: Form - required attribute missing for some of the components
All of the form inputs that can have the `required` attribute should be enabled to have it with the `isRequired` prop.

This attribute is supported on `<select>`, `<textarea>`, and several types of `<input>` ([Note on mdn about what `<input>` types don't support `required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-required))

jenny-s51:
Closes #1397

### DIFF
--- a/packages/patternfly-4/react-core/src/components/FormSelect/FormSelect.tsx
+++ b/packages/patternfly-4/react-core/src/components/FormSelect/FormSelect.tsx
@@ -1,7 +1,5 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/FormControl/form-control';
-import { ASTERISK } from '../../helpers/htmlConstants';
-import { FormContext } from '../Form/FormContext';
 import { css } from '@patternfly/react-styles';
 import { Omit } from '../../helpers/typeUtils';
 
--- a/packages/patternfly-4/react-core/src/components/FormSelect/FormSelect.tsx
+++ b/packages/patternfly-4/react-core/src/components/FormSelect/FormSelect.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/FormControl/form-control';
+import { ASTERISK } from '../../helpers/htmlConstants';
+import { FormContext } from '../Form/FormContext';
 import { css } from '@patternfly/react-styles';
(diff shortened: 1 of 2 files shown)


tlabaj:
@jenny-s51 Why is this labeled breaking change?

jenny-s51:
per @tlabaj , this is not considered a breaking change as it adds an optional (but not required) prop

packages/patternfly-4/react-core/src/components/FormSelect/FormSelect.tsx
```diff
@@ -48,14 +51,15 @@ export class FormSelect extends React.Component<FormSelectProps> {
   };
 
   render() {
-    const { children, className, value, isValid, isDisabled, ...props } = this.props;
+    const { children, className, value, isValid, isDisabled, isRequired, ...props } = this.props;
     return (
       <select
         {...props}
```
redallen: We should move `{...props}` to the end to not break consumers using the `required` property instead.

jgiardino:
I can't say I understand what @redallen is suggesting 😉 so his suggestion might be really good and something worth considering, but when I review the updates in this PR, the form elements now work as I expect. I.e. I can add `isRequired` to `<FormSelect>` and see a `required` attribute on the `<select>` element.

As future work, it would be nice to have examples that clearly illustrate how to create required elements in a form (especially for the select element, because creating the options for a required select can be tricky), and also refine the strings so that they more accurately describe exactly what effect the `isRequired` prop will have. But I'm also curious what @redallen is suggesting and if that might affect what I would suggest in an issue.

redallen:
I'm suggesting that if we have the following JSX:
```jsx
<FormGroup isRequired>
  <FormSelect />
</FormGroup>
```
that we automatically make `<FormSelect>` have `isRequired` passed to it since it's in a `FormGroup` that has `isRequired` set. I thought wrong, though!

jgiardino:
> ...that we automatically make <FormSelect> have isRequired passed to it since it's in a FormGroup that has isRequired set. I thought wrong, though!

Ah, thanks for clarifying! I'm not opposed to this suggestion at all. I would suggest capturing this as a separate issue, though, assuming it's something others agree would be good to implement.

PR #2647, opened 2019-08-05
fix(select): disable focus trap on checkbox select with no children

### LINKED ISSUE #2618: PF4 Select checkbox variant cannot expand with empty children
If you have no children with `SelectVariant.checkbox` and try to expand you will be met with a "You can't have a focus-trap without at least one focusable element" error

## Expended results
1. Click to expand Select
1. Met with empty window

## Actual results
1. Click to expand select
1. Error crashes the application

Reproducible sandbox:
https://codesandbox.io/s/amazing-dawn-ueed9?fontsize=14

boaz0:
**What**:

It seems like Focus Trap throws exception if no children to ref. In order to avoid this error, render the component with no Focus Trap when no children is passed.

closes #2618 

//cc @SpyTec

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
@@ -136,7 +136,7 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
                 {this.extendChildren()}
               </ul>
             )}
-            {variant === SelectVariant.checkbox && (
+            {variant === SelectVariant.checkbox && React.Children.count(children) > 0 && (
               <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
                 <div className={css(styles.selectMenu, className)}>
                   <form noValidate className={css(formStyles.form)}>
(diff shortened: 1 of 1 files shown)


jenny-s51:
Hi @boaz0, your changes look great! When I set the select dropdown to have no children (delete the `options` object from `CheckboxSelectInput`, the focus trap error no longer appears. However, if I try to expand it looks like this. 
ezgif com-video-to-gif(1)

I'm wondering what should be the default behavior of the select dropdown if there are no children. Though it's also possible that I'm not testing it like you did... how did you test your changes?

SpyTec:
Hi @jenny-s51, I would say that issue is out of scope for this PR. The behavior exists on all select components so it's probably better suited for patternfly-next

jenny-s51:
@SpyTec Makes sense to me -- thank you for clarifying Eric!

boaz0:
@jenny-s51 sorry for the late response. I actually thought an empty div is an expected behavior. Though, I agree with @SpyTec when he said we should ask the pf4 core team what they think about it.

//cc @mcoker @mcarrano - can you take a look at this. What should the expanded select item look like when there are no children/options?

mcarrano:
Can you give me an example use case of how/where this would show up?

jenny-s51:
Yes @mcarrano please see https://github.com/patternfly/patternfly-next/issues/2132#issuecomment-518867007 for @SpyTec 's response to your question

mceledonia:
Chiming in here (I think a typo led me here) but I'd think just a basic text-only empty state, maybe secondary text color body style "No options available" or some similar language would work! Same spacing as a single entry in the list... nothing too fancy needed.

jschuler:
This fixes an bug that is already present (throws an exception if no children present), merging this and in the meantime we can explore a loading pattern/skeleton for when there are no children in https://github.com/patternfly/patternfly-next/issues/2132

Issue #2596, opened 2019-07-26
Add default alert 


LHinson:
Add an "default" alert. See the core PR here: https://github.com/patternfly/patternfly-next/pull/2107

rachael-phillips:
@tlabaj are you ready for me to close this issue?

tlabaj:
Closed by #2648

PR #2648, opened 2019-08-05
 feat(Alert): Add the default alert variant 

tlabaj:
**What**: Add default variant for Alert.

**Additional issues**:
#2596


packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
```diff
@@ -10,13 +10,14 @@ export enum AlertVariant {
   success = 'success',
   danger = 'danger',
   warning = 'warning',
-  info = 'info'
+  info = 'info',
```
dlabaj: feel like these should be in alphabetical order ...

PR #2649, opened 2019-08-06
feat(Expandable): Support uncontrolled expandable

### LINKED ISSUE #2350: support uncontrolled Expandable
Often times the expandable section doesn't need to be controlled and for developer ease it would be good to have the Expandable component support an uncontrolled variant if the `isExpanded` prop isn't supplied.

This would also require support for the expanded vs unexpanded text.

jessiehuff:
Fixes #2350


jessiehuff:
Sure thing! I'm on it. :)

PR #2651, opened 2019-08-06
 fix(chip group): add prop to allow chip group to be initially expanded

### LINKED ISSUE #1697: PF4: ChipGroup - allow chip group to be expanded via passing prop isOpen
Defaulting isOpen = false is ok, but allow an override to allow the chip group to be initially expanded.

jenny-s51:
**What**: Closes #1697.


packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -11,6 +11,8 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   /** Additional classes added to the chip item */
   className?: string; 
+    /** Flag for having the chip group default to expanded */
+  isOpenOnInit?: boolean;
```
tlabaj: Hi.  I was just talking to @jschuler  about the name of this prop.  We want to be consistent with the way we name our props across components.  The Page has a similar prop called 'defaultManagedSidebarIsOpen' so we think the chip Group should have a similar name.  How about 'defaultIsOpen`.  I will also rename the Page prop to defaultManagedSidebarIsOpen for consistency.

jenny-s51: Sounds good, thank you Titani -- I'll rename it accordingly

jenny-s51:
@tlabaj absolutely!

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -11,6 +11,8 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   /** Additional classes added to the chip item */
   className?: string; 
+    /** Flag for having the chip group default to expanded */
```
tlabaj: looks like an extra space here

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/ChipGroupDemo/ChipGroupDemo.tsx
```diff
@@ -46,7 +46,7 @@ export class BadgeChip extends Component<{}, BadgeChipState> {
   render() {
     const { badgeChipArray } = this.state;
     return (
-      <ChipGroup>
+      <ChipGroup defaultIsOpen={false}>
```
tlabaj: Can you add a demo for the defaultIsOpen being true and a test for that as well.

packages/patternfly-4/react-integration/cypress/integration/chipgroupdefaultisopen.spec.ts
```diff
@@ -0,0 +1,36 @@
+describe('Chip Group Demo Test', () => {
+    it('Navigate to demo section', () => {
+      cy.visit('http://localhost:3000/');
+      cy.get('#chipgroup-default-is-open-demo-nav-item-link').click(); 
+      cy.url().should('eq', 'http://localhost:3000/chipgroup-default-is-open-demo-nav-link');
+    });
+  
+    it('Verify chip default text', () => {
```
tlabaj: shouldn't we be verifying the chip group is open?

jenny-s51: Yes, nice catch! fixed it 👍

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -29,15 +31,16 @@ export class ChipGroup extends React.Component<ChipGroupProps, ChipGroupState>{
   constructor(props: ChipGroupProps) {
     super(props); 
     this.state = {
-      isOpen: false
+      isOpen: this.props.defaultIsOpen
```
redallen: Don't need `this.`, but `super(props)` does set `this.props` so you're fine :)

PR #2659, opened 2019-08-07
fix(loginForm): adds autoFocus to username input

### LINKED ISSUE #1954: Expose Refs to Input Fields to Allow Focusing
Related Issue: https://github.com/patternfly/patternfly-react/issues/816

I would like to autofocus the username input on the Login form, but currently this is not a supported feature. Without access to refs, the only solution around this is to manipulate the actual DOM.

jenny-s51:
Closes #1954.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.tsx
+++ b/packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.tsx
@@ -5,6 +5,8 @@ import { Button } from '../Button';
 import { Checkbox } from '../Checkbox';
 
 export interface LoginFormProps extends React.HTMLProps<HTMLFormElement> {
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/LoginPage/LoginForm.tsx
```diff
@@ -71,7 +71,7 @@ export const LoginForm: React.FunctionComponent<LoginFormProps> = ({
       {helperText}
     </FormHelperText>
     <FormGroup label={usernameLabel} isRequired isValid={isValidUsername} fieldId="pf-login-username-id">
-      <TextInput
+      <TextInput autoFocus
```
redallen: Can we add a `noFocus` prop in case a consumer doesn't want this new behaviour? Otherwise, nice find!!!

Issue #1294, opened 2019-01-30
html, body, and .pf-c-page changed to "height: 100%;"


mcoker:
As result of some iOS scrolling issues, we changed the height of `.pf-c-page` from `height: 100vh;` to `height: 100%;` requiring the ancestors of `.pf-c-page` to also have `height: 100%`. This was only a CSS change, but we just need to ensure it works in react. It may require adding `height: 100%` to any ancestors of `.pf-c-page`. 

https://github.com/patternfly/patternfly-next/pull/1290/

@tlabaj for PR review

jschuler:
Should probably add a description on the page component that this may be required by the consumer to ensure the ancestors have height 100%

PR #2670, opened 2019-08-08
fullscreen examples should have 100% height

jschuler:
Particularly noticeable for the page demos

Closes: https://github.com/patternfly/patternfly-react/issues/1294


jschuler:
Not sure, Core doesn't mention it either, perhaps we should? @mcoker

mcoker:
@jschuler yeah, I think we should. I don't think having it in the demo is enough for a user to figure out it's required to make the page component full height.

packages/patternfly-4/react-docs/src/templates/mdxTemplate.js
```diff
@@ -3,15 +3,27 @@ import PropTypes from 'prop-types';
...
+import { Title, PageSection, Text, TextList, TextListItem } from '@patternfly-safe/react-core';
 import { MDXProvider } from '@mdx-js/react';
 import { MDXRenderer } from '../components/mdx-renderer';
 
 const components = {
+  wrapper: componentProps => <div className="theme-pf-c-content" {...componentProps} />,
   code: LiveEdit,
-  pre: React.Fragment
+  pre: React.Fragment,
+  p: componentProps => <Text component="p" {...componentProps} />,
```
redallen: We (eventually) need to make sure every HTML tag is covered here that doesn't look right inside `<div className="theme-pf-c-content">`.

packages/patternfly-4/react-docs/src/styles/workspace.css
```diff
@@ -0,0 +1,7 @@
+#___gatsby {
```
redallen: This works, but could also be solved via `wrapper: componentProps => <div className="theme-pf-c-content" style={{height: "100%"}} {...componentProps} />` in `mdxFullscreenTemplate.js` to be consistent with `mdxTemplate.js`.

jschuler: that would only set 100% height on the wrapper, but the ancestors #__gatsby and a child div still need 100% as well

redallen: Ah -- you're right!

Issue #2358, opened 2019-06-26
PF4 Please provide a lightweight composable component set for tables


idevat:
Patternfly-table provides heavyweight component `Table`. This component tries to provide complex functionality to ease a development. But my experience with this component is a bit painful and I would welcome a set of simple composable components beside this super "smart" component. Let me elaborate it a bit.

As can be seen for example in issue #1399, component `Table` does not allow the developer to control some details. It tries to make things universal. In the case of issue #1399 it is about a control over the keys. It can fit to some use cases but not to all.

Sometimes, a particular use case could be achieved much more easily with set of simple tools. Actually, I haven't yet encountered a case where the current `Table` component saved me the hard work that a set of simple components couldn't handle. On the contrary. But it is true that my experience with this component is not extensive.

Most of the examples in https://www.patternfly.org/v4/documentation/react/components/table/ could be implemented with a set of simple components, I guess. Just a note, I haven't found the difference between `First cell as header table` and `Simple table`. When I looked into the code I found `{ title: 'Header cell', cellTransforms: [headerCol('selectable')] }` but  I have no idea what it should do (without more detailed study).

Take for example the issue #2357. If a user had components similar to these:

```javascript
const PFTable = ({ children }) => (
  <table className="pf-c-table pf-m-grid-md">
    {children}
  </table>
);

const PFCompoundExpansionToggle = ({ expanded, onClick, children }) => (
  <td
    className={
      `pf-c-table__compound-expansion-toggle ${expanded ? "pf-m-expanded" : ""}`
    }
  >
    <button
      className="pf-c-button pf-m-link"
      type="button"
      onClick={onClick}
    >
      {children}
    </button>
  </td>
);

const PFExpandableRow = ({ children }) => (
  <tr className="pf-c-table__expandable-row pf-m-expanded">
    {children}
  </tr>
);

```

the implementation of task from issue #2357 might look like this:

```javascript
const ClusterRow = ({ cluster }) => {
  const [expanded, setExpanded] = React.useState(false);
  return (
    <React.Fragment>
      <tr>
        <td>{cluster.name}</td>
        <PFCompoundExpansionToggle
          expanded={expanded}
          onClick={() => setExpanded(!expanded)}
        >
          {cluster.nodeList.length}
        </PFCompoundExpansionToggle>
      </tr>
      {expanded && (
        <PFExpandableRow>
          <td colSpan={2}>Nodes</td>
        </PFExpandableRow>
      )}
    </React.Fragment>
[...]

PR #2672, opened 2019-08-08
First pass at updating the examples for the patternfly/react-table library

riccardo-forina:
**What**: 

I have personally struggled with patternfly/react-table's examples for a number of reasons:
* examples are provided as class-based components. This is ok, but some of us moved to function components (with hooks. Given that that seems to be the direction the React community is following, I think it makes sense to update the examples to the new style.
* heavy and confusing use of `setState`. I think `setState` in an example should be used only if the example requires it. But more importantly, for a developer that's still learning React it teaches a bad lesson. The React way should be to derive the state required for the rendering as much as possible, and avoid storing multiple copies of the same data using the state.
* arrays and objects are mutated in place. This is a bad practice in general, but with hooks it's also a pitfall since all hooks perform a shallow comparison between objects, which doesn't catch object mutations.

So I took a stab at updating the examples (well, some of them for the moment) to address the above points.

Finally, when updating the selectable rows example I got bit really hard by something that looked like a bug in the library but ended up being the right behavior. Using an immutable data structure and keeping track of the selected rows by storing the id in some state, this is what was happening: 

1. initial render

| Select  | Name | Notes |
| ------------- | ------------- | ------------- |
| [ ]  | Foo  |  |
| [ ]  | Bar  |  |
| [ ]  | Baz  |  |

2. we select the Bar line

| Select  | Name | Notes |
| ------------- | ------------- | ------------- |
| [ ]  | Foo  | this row will not be re-rendered |
| [x]  | Bar  | this row will be re-rendered because it's been selected |
| [ ]  | Baz  | this row will not be re-rendered |

3. we select the Baz line. We expect Bar to stay select, but it gets deselected.

| Select  | Name | Notes |
| ------------- | ------------- | ------------- |
| [ ]  | Foo  | this row will not be re-rendered |
| [ ]  | Bar  | this row will be re-rendered but with an unchecked state, because that was its state when Baz was rendered the first time |
| [x]  | Baz  | this row will be re-rendered because it's been selected |

This was happening because the Foo's and Baz's `onSelect` callback was not being updated because the `BodyRow:shouldComponentUpdate` function _correctly_ detected that nothing changed in the data, and skipped the render.
The solution was simple, the `onSelect` callback had to make use of `useRef` to track the latest snapshot of the selected rows. 

I then realized that the logic for the selectable table wasn't trivial, especially for a developer new in the React ways, and also we were exposing _a lot_ of logic to the end-user that I think we could abstract away. So I moved all the logic to a reusable hook called `useSelectableRows` that does all the above for the user, asking only for the rows and, optionally, a function to retrieve the row's key.

The example is now super short and I think we can save other devs some good amount of time because of this little hook.

I plan to do apply the same treatment to the remaining examples, but I'll probably not be able to do so for a few weeks. In the meantime, I'd like to collect your feedback about these changes. Are you all ok with it? Something that can be improved or changed?

**Additional issues**: kinda related #2358

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.md
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.md
@@ -13,6 +13,7 @@ import {
   TableHeader,
   TableBody,
   sortable,
+  SortHelpers,
   SortByDirection,
   headerCol,
   TableVariant,
@@ -22,13 +23,16 @@ import {
   textCenter,
   wrappable,
   classNames,
-  Visibility
+  Visibility,
+  useSortableRows,
+  useSelectableRows
 } from '@patternfly/react-table';
-
 import {
   CodeBranchIcon,
   CodeIcon,
-  CubeIcon
+  CubeIcon,
+  CheckIcon,
+  TimesIcon
 } from '@patternfly/react-icons';
 
 import DemoSortableTable from './demo/DemoSortableTable';
@@ -241,11 +245,11 @@ function SortableTable () {
   }; 
 
   const cells = [
-    { title: 'Repositories', transforms: [sortable(SortHelpers.strings)] },
-    { title: 'Branches', transforms: [sortable(SortHelpers.numbers)] },
-    { title: 'Pull requests', transforms: [sortable(SortHelpers.numbers)] },
-    { title: 'Workspaces', transforms: [sortable(SortHelpers.numbers)] },
-    { title: 'Last Commit', transforms: [sortable(sortLastCommit)] }
+    { title: 'Repositories', transforms: [sortable] },
+    { title: 'Branches', transforms: [sortable.numbers] },
+    { title: 'Pull requests', transforms: [sortable.numbers] },
+    { title: 'Workspaces', transforms: [sortable.numbers] },
+    { title: 'Last Commit', transforms: [sortable.custom(sortLastCommit)] }
   ];
 
   const [sortedRows, onSort, sortBy] = useSortableRows(rows);
@@ -268,7 +272,6 @@ import {
   TableHeader,
   TableBody,
   headerCol,
-  sortable,
   useSelectableRows
 } from '@patternfly/react-table';
 
@@ -307,6 +310,7 @@ import {
   TableHeader,
   TableBody,
   headerCol,
+  sortable,
   useSelectableRows
 } from '@patternfly/react-table';
 
@@ -318,11 +322,11 @@ function SortableAndSelectableTable () {
   ];
   
   const cells = [
-    { title: 'Repositories', transforms: [sortable(SortHelpers.strings)], cellTransforms: [headerCol()] },
-    { title: 'Branches', transforms: [sortable(SortHelpers.numbers)] },
-    { title: 'Pull requests', transforms: [sortable(SortHelpers.numbers)] },
-    { title: 'Workspaces', transforms: [sortable(SortHelpers.numbers)] },
-    { title: 'Last Commit', transforms: [sortable(SortHelpers.numbers)] }
+    { title: 'Repositories', transforms: [sortable], cellTransforms: [headerCol()] },
+    { title: 'Branches', transforms: [sortable.numbers] },
+    { title: 'Pull requests', transforms: [sortable.numbers] },
+    { title: 'Workspaces', transforms: [sortable.numbers] },
+    { title: 'Last Commit', transforms: [sortable.numbers] }
   ];
   
   const [sortedRows, onSort, sortBy] = useSortableRows(rows, cells);
@@ -404,7 +408,7 @@ class SimpleActionsTable extends React.Component {
   render() {
     const { columns, rows, actions } = this.state;
     return (
-      <Table caption="Actions Table" actions={actions} cells={columns} rows={rows}>
+      <Table caption="Simple Actions Table" actions={actions} cells={columns} rows={rows}>
         <TableHeader />
         <TableBody />
       </Table>
--- a/packages/patternfly-4/react-table/src/components/Table/Table.md
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.md
@@ -41,70 +41,175 @@ import {
   Table,
   TableHeader,
   TableBody,
-  sortable,
-  SortByDirection,
-  headerCol,
-  TableVariant,
-  expandable,
+  textCenter,
+} from '@patternfly/react-table';
+
+function SimpleTable() {
+  const cells = [
+    'Repositories',
+    'Branches',
+    'Pull requests',
+    'Workspaces',
+    'Last Commit'
+  ];
+  const rows = [
+    ['Foo', 2, 0, 6, 1533635470],
+    ['Bar', 1, 11, 2, 1564566670],
+    ['Baz', 6, 4, 99, 1565167870],    
+  ];
+
+  return (
+    <Table caption="Simple Table" cells={cells} rows={rows}>
+      <TableHeader />
+      <TableBody />
+    </Table>
+  );
+}
+```
+
+## Cell/row options
+
+```js
+import React from 'react';
+import {
+  Table,
+  TableHeader,
+  TableBody,
   cellWidth,
   textCenter,
 } from '@patternfly/react-table';
 
-class SimpleTable extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      columns: [
-        { title: 'Repositories' },
-        'Branches',
-        { title: 'Pull requests' },
-        'Workspaces',
-        {
-          title: 'Last Commit',
-          transforms: [textCenter],
-          cellTransforms: [textCenter]
-        }
-      ],
-      rows: [
-        {
-          cells: ['one', 'two', 'three', 'four', 'five']
-        },
-        {
-          cells: [
-            {
-              title: <div>one - 2</div>,
-              props: { title: 'hover title', colSpan: 3 }
-            },
-            'four - 2',
-            'five - 2'
-          ]
-        },
-        {
-          cells: [
-            'one - 3',
-            'two - 3',
-            'three - 3',
-            'four - 3',
-            {
-              title: 'five - 3 (not centered)',
-              props: { textCenter: false }
-            }
-          ]
-        }
-      ]
+function SimpleCustomTable() {
+  const cells = [
+    // Equivalent to just passing the string
+    { title: 'Repositories' }, 
+    'Branches',
+    { title: 'Pull requests' },
+    'Workspaces',
+    // Will run the `transform` functions on the header, and the 
+    // `cellTransforms` on the cells. Transformers are used to inject
+    // extra props to the the cell component. In this case we use the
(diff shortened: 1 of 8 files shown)


priley86:
Hey @riccardo-forina !

Really awesome points, and love this feedback! ❤️ 

A couple things come to mind for me personally:

- Completely agree we should start targeting the [React Hooks API](https://reactjs.org/docs/hooks-reference.html) in future PF React components and examples. I believe this is the first such addition so a major kudus on this! ⭐️  I think this may spark a larger discussion about when to move away from class based components here... cc: @dgutride @dlabaj @jcaianirh 

- Overall, I can see a lot of improvements that can be made to `@patternfly/react-table` if adopting Hooks. Using `useContext` Hooks for providing the current `TableContext` would be much cleaner as well down the road...

- Agree that using `setState` is non desirable and `useState` will be preferable once we adopt Hooks. Much of this is legacy and the way Reactabular showed that `rows` and `cols` were ultimately stateful (and dynamic) but it is better to think of them as props from a React perspective. Some consumers had these tied to Redux for providing table data and attributes, but I've noted a really good [blog](https://kentcdodds.com/blog/application-state-management-with-react) on how to move away from Redux and the old React state models like you propose here. I see a ton of value in moving towards this kind of API for React based consumption.

- Regarding `BodyRow:shouldComponentUpdate` and `useRef` - very good catch w/ this. I think `useSelectableRows` is a nice addition and exactly the kind of API feedback we are looking for now that the Reactabular source is "in-house". Happy to see these kind of improvements coming!

Overall, I don't have any issue moving the React Table docs here towards Hooks if others are OK w/ this effort. I think it would be a really nice addition to test alongside #2360 after that PR goes and set the stage nicely for future...

Will add a few inline comments and await any feedback from others!

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -40,125 +40,179 @@ import {
...
+  ];
+  const rows = [
+    ['Foo', 2, 0, 6, 1533635470],
+    [
+      // Cell content can also be defined as a JSX element.
+      // Extra props can be passed to the cell (the `td` element).
+      {
+        title: <div>Bar 👾</div>,
+        props: { title: 'hover title', colSpan: 3 }
+      },
```
priley86: nice! I think these comments certainly help understanding. That is the goal!

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -40,125 +40,179 @@ import {
...
+  // the main component function in a real world application; it's
+  // done this way in this example as a work-around against the 
+  // limitation of one component per example of the documentation
+  // system in use.
+  const CIStatusIcon = ({ passing }) => {
+    const styles = {
+      color: passing 
+        ? 'var(--pf-global--success-color--200)' 
+        : 'var(--pf-global--danger-color--100)'
     };
```
priley86: I am all for showing examples of custom CSS applications if they are useful and match PF standards. I don't know if we want a single "style customized" example or many, but I like this. "Cell Formatters" example is a very welcome addition. We could potentially have other examples of this coming in the future too (i.e. a table that has "favorited rows") cc: @patternfly/patternfly-core-ux

riccardo-forina:
I'm not sure why the doc fails building. This is the error:
`@patternfly/react-docs:   WebpackError: ReferenceError: CheckIcon is not defined`
But it's working fine in dev mode, and I believe the icon is there?

riccardo-forina:
Today I focused my efforts making sure that sortable and selectable tables are easier to pull off. There is now a `useSortableRows` hook that allows a user to sort the rows automatically, providing the right sort function to the cell. There are a few common sorting functions already implemented and usable, but a custom one can be provided too (check the example with the cell with the emoji).

I also updated the Typescript typings a bit. Not sure how friction-free with the PR to migrate the lib to TS will be, but it helped in writing the hooks.

I doubt I'll be able to push any more updates to this for a few weeks, and CI doesn't pass for reasons I'm not sure I understand. So if you feel like closing this and have me reopen it again in the future I'll understand :D
Else if anyone wants to jump in and push PRs against my branch, that works too.

packages/patternfly-4/react-table/src/components/Table/Table.md
mturley: I agree that these examples might be useful, maybe a section of the docs page on "customizing styles" with a few examples would be good in the long term, but we could start with just one.

As far as using the `--pf-global` variable directly as a string here, @priley86 do you happen to know if there is a way to import these color variables as constants from `@patternfly/react-styles` somehow?

mturley:
+1, this is a really cool direction for us to move in! I love the introduction of Hooks, that will be a great way to clean up some of the more convoluted patterns in the original reactabular source. Nice work @riccardo-forina !

I do think we should rebase/restart this after https://github.com/patternfly/patternfly-react/pull/2360 is merged, because there will be conflicts and because converting the whole of react-table to TypeScript before making changes like this to the API/behavior will make the history easier to track.

riccardo-forina:
Ok I finally managed to sync this back with upstream and find a way around the `sortable` API change I did introduce.

Are you ok with reviewing this now? I plan to follow up with other PRs for the remaining examples to avoid making this too big.

packages/patternfly-4/react-table/src/components/Table/Table.md
seanforyou23: Love the "favorited rows" idea as an example.

@mturley yep, we could do something like `import { global_breakpoint_lg } from '@patternfly/react-tokens';` which basically results to 

```
{
    "name": "--pf-global--breakpoint--lg";
    "value": "992px";
    "var": "var(--pf-global--breakpoint--lg)";
}
```

seanforyou23:
One more thing, I started with the code from the CellFormatters example as a base, then, I attempted to incorporate some behavior from the "compact expandable table" example but hung and realized that now some examples use a basic array instead of array of objects representing the rows. I wonder if it would be too much trouble to align this one thing, that way the demos go together easier like legos?

I then think about this sort of related comment `Please do not use rowKey as row index for more complex tables. Rather use some kind of identifier like ID passed with each row.` and wonder if we could use this as an opportunity to make exactly that change (since we're here in the spirit of DX) in the demo code instead of suggesting it to each user to do on their own? Especially for the simple case. The more we can set users up out of the box the better and these changes might help put people onto the fast track. Thoughts?

Screen Shot 2019-09-05 at 7 51 25 PM

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -41,127 +45,221 @@ import {
...
+    { title: 'Branches', transforms: [sortable.numbers] },
+    { title: 'Pull requests', transforms: [sortable.numbers] },
+    { title: 'Workspaces', transforms: [sortable.numbers] },
+    { title: 'Last Commit', transforms: [sortable.custom(sortLastCommit)] }
+  ];
+
+  const [sortedRows, onSort, sortBy] = useSortableRows(rows);
+
+  return (
+    <Table caption="Sortable Table" sortBy={sortBy} onSort={onSort} cells={cells} rows={sortedRows}>
```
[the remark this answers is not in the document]
riccardo-forina: I see a `Table.d.ts` referenced there, but there is no such thing anymore. Wondering, did you run a full build?

packages/patternfly-4/react-table/src/components/Table/Table.md
riccardo-forina: Sure!

packages/patternfly-4/react-table/src/components/Table/Table.md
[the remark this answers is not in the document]
seanforyou23: Good call, not sure why I'm still seeing the old files, totally did a full rebuild, must be the caches. Give me a bit to re-test this and I'll update my comments - for now please ignore them!

mturley:
@riccardo-forina @seanforyou23 this got marked stale, do we think it's worth rebasing and trying to get it moving again or should we close it?

riccardo-forina:
I know that the table package is being actively developed, unfortunately I don't have much time to update this. I'm ok for closing it unless @seanforyou23 is of a different view.

seanforyou23:
Thanks to everyone for being patient while this one sat on the back burner. We've finally worked through several critical type issues that we needed to have in place before we could really think about making moves on things mentioned in this PR. Here's where I think we're at..

We want to embrace the idea of keeping the table and example implementations current, and also balance that with continuing the support the existing implementations which may be favorable to some. We have learned from this PR and implemented a hook based approach in our [beginner katacoda tutorial](https://www.katacoda.com/patternfly/courses/react-components/table-intro), it was easier to use this new pattern here because there was no pre-existing code to combat with. I think it's been well received and hope that we can continue to add more of this style tutorial.

>arrays and objects are mutated in place

This ^ is a critical point in my opinion. When I follow the examples from our docs and convert them into function components, some of the things simply do not work as expected. For example, these lines do not work for me in a function component based approach, and I think it's for the reason you mention, that we are mutating objects in place.

```js
const { rows } = this.state;
rows[rowKey].isOpen = isOpen;
// then set state with `rows`
```

The fix was simple, just create a new array based off of the `rows` state property, and modify that instead, then set state with that. I think at the very least we should make this type of change to our class-based component examples. This goes to show how function components can illuminate problems in the code we don't realize even exist.

There have been some enhancements made to the selectable rows doc, and I think further enhancements would be welcome. `useSelectableRows` looks really awesome and I love the idea of offsetting complex logic like this so each user doesn't have to reinvent it. The truth is, though, that we're still building our competency as a team on the hooks-based approach, and we need to make sure we can support these types of design patterns as a team before we go all-in on them in the docs. For the sake of incremental progress, I say let's make the action item to fix these cases of mutating objects in place so that at the very least, those who wish to use hooks/function components can more easily copy code from our examples without immediately running into problems. I don't think we can have too many examples/demos, so instead of converting the existing examples to hooks, can we instead add _new_ examples/tutorials that illustrate these concepts so that all people have bases covered, regardless of their preferred pattern?

That's the best **_and_** solution I can think of at the moment. In that case, we can close this PR and resurrect it as time allows as an addition to the docs instead of a refactor. Does that cover all our bases? I will send a PR to fix the in-place mutations to soften the problem in the mean time.

Let me know what you think, sorry for the wall, just wanna make sure all the context was captured for anyone interested to see. @riccardo-forina @mturley lemme know what you think.

riccardo-forina:
Thanks for this! I’m :+1: on everything, examples can (maybe should) be showing how to do things both with classes and function components. It makes sense to find a way to do it.

mturley:
Totally agree on all points. Especially the in-place mutations.. I don't know if that's a practice we inherited from Reactabular, but it for sure needs to go. To be honest, ever since we decided to pull Reactabular's source in house and convert it to TypeScript, I started asking myself why that became necessary and why the original Reactabular devs stopped maintaining it. It's possible we will need to move farther away from their original implementation ideas to make sure ours can keep up.

I think it makes sense to modernize what we can when we can, while coming to speed with things like hooks, and honoring the original implementation for those who already use it. Thanks for looking at this @seanforyou23 !

Issue #2436, opened 2019-07-02
Add 2xl breakpoint to [hidden/visible]On[breakpoint] table cell visibility classes


mcoker:
we have a new breakpoint `$pf-global--breakpoint--2xl` that should be supported with both `visibleOn` and `hiddenOn` with the `classNames(Visibility.hidden, Visibility.visibleOnMd, Visibility.hiddenOnLg, ...)` classes.

tlabaj:
@priley86 for review of PR

priley86:
sounds good... should be a simple add here hopefully:

https://github.com/patternfly/patternfly-react/blob/fdbde6d0fa3e6151300f1bbb8322a3308fb606c0/packages/patternfly-4/react-table/src/components/Table/utils/decorators/classNames.js

kmcfaul:
Closed in #2676

PR #2676, opened 2019-08-09
feat(Table): add new visibleOn and hiddenOn breakpoint

kmcfaul:
**What**: Adds the new 2xl breakpoint to visibleOn and hiddenOn classes.

**Refer to issue**: #2436

@priley86 Is there a good way to add tests for this? I'm unfamiliar with table and unsure how it uses the Visibility object.


priley86:
this looks good to me @kmcfaul. I believe it would be great to add unit tests in `Table.tests.tsx` for this after #2360 lands. We can add some new tests for `columnTransforms` there I'm guessing.

Issue #2366, opened 2019-06-26
PF4 Select: Add isDisabled prop to the component


tlabaj:
The consumer should be able to disable the Select component similar to the Dropdown.

This issue is dependent on the core issue [#1985](https://github.com/patternfly/patternfly-next/issues/1985)

@jschuler will be in charge on the PR review for this one.

mcoker:
@tlabaj @kmcfaul the work for this is done in core, merged in https://github.com/patternfly/patternfly-next/pull/2028

tlabaj:
Closed by #2678

PR #2678, opened 2019-08-09
feat(Select): add disabled flag for component

kmcfaul:
**What**: Adds isDisabled flag to select.

**Refer to issue**: #2366


kmcfaul:
Requires core to be updated to v2.26 to get styling

packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -28,6 +28,8 @@ export interface SelectProps
   isGrouped?: boolean;
   /** Display the toggle with no border or background */
   isPlain?: boolean;
+  /** Flad to inficate if select is disabled */
```
tlabaj: typo

packages/patternfly-4/react-core/src/components/Select/Select.md
```diff
@@ -325,6 +407,84 @@ class TypeaheadSelectInput extends React.Component {
...
+
+```js
+import React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+class TypeaheadSelectInput extends React.Component {
+  constructor(props) {
+    super(props);
+    this.options = [
+      { value: 'Alabama', disabled: false },
```
jschuler: I'd remove the disabled keys from the options array in the two added examples for clarity purposes

tlabaj: I was thinking they same thing.

tlabaj:
@jschuler ,  I think the examples are fine as is.  They match core.

PR #2683, opened 2019-08-12
fix(button): set the disable modifier when component is not button

### LINKED ISSUE #2666: disabled button should only apply the disabled attribute
Currently if you use

```
<Button isDisabled variant="primary">Primary</Button>
```

That generates

```
<button class="pf-c-button pf-m-primary pf-m-disabled" disabled="" type="button">
  Primary
</button>
```

When disabling a `<button>` html element, all you need is the `disabled` attribute, not the class `.pf-m-disabled`. The class doesn't break anything, but it's unnecessary. The class `.pf-m-disabled` should only apply to elements that don't support the `disabled` attribute, like `<a>` elements, for example.

In the case above, the rendered HTML should be:

```
<button class="pf-c-button pf-m-primary" disabled type="button">
  Primary
</button>
```

boaz0:
**What**:

closes #2666 

If component is not `button` and `isDisabled` is true then add `pf-m-disabled` to class name.

//cc @mcoker

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Button/Button.tsx
+++ b/packages/patternfly-4/react-core/src/components/Button/Button.tsx
@@ -76,7 +76,7 @@ const Button: React.FunctionComponent<ButtonProps & InjectedOuiaProps> = ({
       styles.button,
       getModifier(styles.modifiers, variant),
       isBlock && styles.modifiers.block,
-      isDisabled && styles.modifiers.disabled,
+      isDisabled && !isButtonElement && styles.modifiers.disabled,
       isActive && styles.modifiers.active,
       isFocus && styles.modifiers.focus,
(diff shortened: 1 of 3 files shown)


boaz0:
fixed merge conflicts

tlabaj:
@mcoker I can open up a follow up issue.  Just to clarify,  For the Select,  the  <div> containing the <button> should not have the pf-m-disabled modifier applied correct?

mcoker:
@tlabaj Yeah, basically the toggle can either be a `<div>` or a `<button>`. When the toggle is a `<button>`, we should use the `disabled` attribute to disable it. If the toggle is a `<div>`, we should use `.pf-m-disabled` to disable it.

PR #2684, opened 2019-08-12
fix(radio): group radio inputs in docs (axe-core)

### LINKED ISSUE #2568: fix(a11y): Radio component
This issue was found in org but is coming from React: https://github.com/patternfly/patternfly-org/issues/1318

The issue is: "Radio inputs with the same name attribute value must be part of a group"

boaz0:
**What**:

closes #2568 

Radio inputs that have the same name should be grouped. One way to group them is inside a div with `role="group"` and `aria-label` to describe this group.

//cc @christiemolloy

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Radio/Radio.md
+++ b/packages/patternfly-4/react-core/src/components/Radio/Radio.md
@@ -8,6 +8,7 @@ propComponents: ['Radio']
 import { Radio } from '@patternfly/react-core';
 
 ## Controlled radio
+
 ```js
 import React from 'react';
 import { Radio } from '@patternfly/react-core';
@@ -16,12 +17,12 @@ class ControlledRadio extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: '4'
+      check1: false,
     };
 
     this.handleChange = (_, event) => {
       const { value } = event.currentTarget;
-      this.setState({ value });
+      this.setState({ [value]: true });
     };
   }
 
@@ -29,20 +30,11 @@ class ControlledRadio extends React.Component {
     return (
       <React.Fragment>
         <Radio
-          value="3"
-          isChecked={this.state.value === '3'}
-          name="pf-version"
+          isChecked={this.state.check1}
+          name="radio-1"
           onChange={this.handleChange}
-          label="Controlled radio 1"
-          id="radio-1"
-        />{' '}
-        <Radio
-          value="4"
-          isChecked={this.state.value === '4'}
-          name="pf-version"
-          onChange={this.handleChange}
-          label="Controlled radio 2"
-          id="radio-2"
+          label="Controlled radio"
+          id="radio-controlled"
         />
       </React.Fragment>
     );
@@ -51,27 +43,54 @@ class ControlledRadio extends React.Component {
 ```
 
 ## Uncontrolled radio
+
 ```js
 import React from 'react';
 import { Radio } from '@patternfly/react-core';
 
 UncontrolledRadio = () => (
   <React.Fragment>
-    <Radio label="Uncontrolled radio example" id="radio-4" name="radio-4" />
-    <Radio label="Uncontrolled radio example" id="radio-5" name="radio-4" />
+    <Radio label="Uncontrolled radio example" id="radio-uncontrolled" name="radio-2" />
+  </React.Fragment>
+);
+```
+
+## Reversed radio
+
+```js
+import React from 'react';
+import { Radio } from '@patternfly/react-core';
+
+ReversedRadio = () => (
+  <React.Fragment>
+    <Radio isLabelBeforeButton label="Reversed radio example" id="radio-reversed" name="radio-3" />
+  </React.Fragment>
+);
+```
+
+## Label wraps input radio
+
+```js
+import React from 'react';
+import { Radio } from '@patternfly/react-core';
+
+LabelWrapsInputRadio = () => (
+  <React.Fragment>
+    <Radio isLabelWrapped label="Label wraps input example" id="radio-label-wraps-input" name="radio-4" />
   </React.Fragment>
 );
 ```
 
 ## Disabled radio
+
 ```js
 import React from 'react';
 import { Radio } from '@patternfly/react-core';
 
 DisabledRadio = () => (
   <React.Fragment>
-    <Radio label="Disabled checked radio example" defaultChecked isDisabled name="group-1" id="radio-disabled" />{' '}
-    <Radio id="radio-disabled-2" label="Disabled radio example" isDisabled name="group-2" />
+    <Radio id="radio-disabled" label="Disabled radio example" isDisabled name="radio-5" />
+    <Radio id="radio-disabled-checked" defaultChecked label="Disabled and checked radio example" isDisabled name="radio-6" />
   </React.Fragment>
 );
 ```
--- a/packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
+++ b/packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
@@ -9,6 +9,10 @@ export interface RadioProps
   className?: string;
   /** Id of the radio. */
   id: string;
+  /** Flag to show if the radio label is wrapped on small screen. */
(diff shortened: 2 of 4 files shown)


jgiardino:
Thanks for working on this, @boaz0! First of all, your solution is valid in terms of accessibility. However, I question if we should be fixing this issue at all. 

If I use the Radio component in the context of a form, I would expect to provide a visible label that names the group of radio buttons. 

There is an issue related to this in core that's still open and provides a suggestion on how to handle that: https://github.com/patternfly/patternfly-next/issues/953

That issue is in the context of the Form component, which would consume the Radio component. If we want to handle how radio buttons are grouped, I _think_ that would be addressed in the Form component. 

If we were to update the Radio component with this PR, how would that update affect the ability to define the group and group label in the context of the Form component? 

If we rely on this grouping and naming to happen in the Form component, is it ok to have this component fail the a11y audit? 

My concern with trying to make things accessible out of context is that the consumer might not realize that the additional markup was only included because the component was out of context. My suggestion would be to not address a11y failures in situations like this, and instead refer the consumer to a component or demo that illustrates how to use the component in an accessible way.

I'm curious what others think about this.

boaz0:
@jgiardino :+1: 

How about aligning the example with what we have in the pf4 core [here](https://codepen.io/pen/?&editable=true) - it will fix the a11y message till https://github.com/patternfly/patternfly-next/issues/953 will be solved.

What do you think?

Thanks.

jgiardino:
👍 Works for me. Thanks for updating this, @boaz0!!

jessiehuff:
@boaz0 I was just about to say the same thing as Zack above. I noticed the radio functionality seems to be broken. :(

jgiardino:
I'm not sure if this is still the goal or not, but my understanding with creating examples is that they should be similar between core and react. The example from core also doesn't wrap the radio buttons in the same group (i.e. creating the behavior you described).

 @jessiehuff @redallen 
If there is an issue with the example created for this PR, I would suggest raising that question with both core and react contributors to figure out what the example should be before asking @boaz to continue making changes. 

I also question if that's something that should block this PR or not. I thought the scope of this PR was to address the accessibility issue found by aXe, and since the behavior between core and react examples seem to match, I think this PR could move forward. 🤷‍♀

christiemolloy:
Currently I can't click out of the radio checkboxes, are you getting the same thing?
Screen Shot 2019-09-04 at 11 22 33 AM

jgiardino:
Here is the example from html. The ability to select a radio button in a group is native to html, and not functionality that's provided by javascript. 

cc'ing @mcoker on this to weigh in on whether we should be wrapping the Radio component with additional things that are not part of the component in our component examples (e.g. like a `<fieldset>`), or if that type of example belongs somewhere else, like a demo?

image

mcoker:
@jgiardino this PR looks good to me. My expectation is that the examples should match, too, whenever possible.

That said, for this example, I think the examples should just ideally just represent the component, and not include anything that isn't necessary to demonstrate the component's functionality. Here I think labeling the radio button examples as checked/unchecked would suffice to serve as an example for the radio component, then we illustrate how radio buttons should be named, grouped, etc in the context of a form in a demo.

mcoker:
We chatted about it as a team and decided to keep any additional form elements needed to create a group of radio buttons in a form out of the component examples, and since it's a big confusing now that you can't toggle between radio buttons (like you would in a form), just to update the examples so we only have a single radio in the example, instead of a group. I created an issue to do that in the core workspace here - https://github.com/patternfly/patternfly-next/issues/2221. Can we do that for the examples in the react workspace, too?

boaz0:
@mcoker yes - I will update according to https://github.com/patternfly/patternfly-next/issues/2221 
no problem. :smile_cat:

boaz0:
fixed merge conflicts

jgiardino:
Thanks for making these updates! I only see a couple of updates needed for the example **Label wraps input radio**:

- The element that has the class `pf-c-radio` should be a `<label>` and not a `<div>` in this case.
- The `for` attribute that is on the `<span>` should be moved to the `label.pf-c-radio` element.

packages/patternfly-4/react-core/src/components/Radio/Radio.md
```diff
@@ -16,33 +17,25 @@ class ControlledRadio extends React.Component {
...
 
   render() {
     return (
       <React.Fragment>
         <Radio
-          value="3"
-          isChecked={this.state.value === '3'}
-          name="pf-version"
+          value="check1"
+          isChecked={this.state.check3}
```
tlabaj: should this be `this.state.check1`

boaz0: That's right

boaz0:
@jgiardino thank you for the review.
A question: is it possible that both the radio will be reversed and the label wraps input?

mcoker:
>  is it possible that both the radio is reversed and the label wraps input?

@boaz0 yep!

boaz0:
@jgiardino @mcoker done - I added `isReversed` and `isWrapped` props and discarded the variant prop idea.

Let me know what you think.

I have a question though when I am in "label wraps input" making the window slower doesn't put the label in one row and the radio button in the second row. Am I missing something?

Thanks.

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
```diff
@@ -1,14 +1,21 @@
...
+  extends Omit<
+    React.HTMLProps<HTMLInputElement>,
+    "disabled" | "label" | "onChange" | "type"
+  > {
   /** Additional classes added to the radio. */
   className?: string;
   /** Id of the radio. */
   id: string;
+  /** Flag to show if the radio label is wrapped on small screen. */
+  isWrapped?: boolean;
```
tlabaj: Maybe we can be little more descriptive with prop names here? e.g. isLabelWrapped and is isLabelBeforeButton.

packages/patternfly-4/react-core/src/components/Radio/Radio.md
```diff
@@ -51,27 +43,54 @@ class ControlledRadio extends React.Component {
...
+
+## Reversed radio
+
+```js
+import React from 'react';
+import { Radio } from '@patternfly/react-core';
+
+ReversedRadio = () => (
+  <React.Fragment>
+    <Radio variant="reversed" label="Reversed radio example" id="radio-reversed" name="radio-3" />
```
tlabaj: This should be using the boolean correct?

packages/patternfly-4/react-core/src/components/Radio/Radio.md
```diff
@@ -51,27 +43,54 @@ class ControlledRadio extends React.Component {
...
+
+## Label wraps input radio
+
+```js
+import React from 'react';
+import { Radio } from '@patternfly/react-core';
+
+LabelWrapsInputRadio = () => (
+  <React.Fragment>
+    <Radio variant="label-wraps" label="Label wraps input example" id="radio-label-wraps-input" name="radio-4" />
```
tlabaj: Same comment as above

mcoker:
> I have a question though when I am in "label wraps input" making the window slower doesn't put the label in one row and the radio button in the second row. Am I missing something?

Ah, the difference between the normal and "label wraps input" variations doesn't have anything to do with content wrapping onto multiple lines. It's that the normal markup is like:

```
div.pf-c-radio
   input.pf-c-radio__input
   label.pf-c-radio__label
```

And when the label wraps, it's:

```
label.pf-c-radio
    input.pf-c-radio__input
    span.pf-c-radio__label
```

packages/patternfly-4/react-core/src/components/Radio/Radio.tsx
boaz0: I like that idea

boaz0:
@tlabaj thank you so much for the review - that was very helpful.
@mcoker thanks a lot for the explanation. I also appreciate your time testing and reviewing again and again.

boaz0:
@tlabaj conflict was fixed.

boaz0:
Thank you all!! :tada:

Issue #1433, opened 2019-02-22
PF4 Table: Provide empty state support


dlabrecq:
The Cost Management UI needs to provide an empty state in a table, per Kyle's high fidelity mock below. For example, if the applied filter results in no data. However, I cannot place the PF4 empty state component inside the Table, nor the TableBody.

If I place the empty state components within the Table, the empty state component does not span columns. The column headers shift right, while the empty state is aligned left.

<img width="1255" alt="screen shot 2019-02-22 at 3 45 30 pm" src="https://user-images.githubusercontent.com/17481322/53270170-f98a8d00-36b8-11e9-9cac-6805c9837f52.png">

I can trick the table by placing the empty state components within a row cell. However, I still get the expanding row checkbox, which is not desired.

<img width="346" alt="screen shot 2019-02-22 at 3 47 39 pm" src="https://user-images.githubusercontent.com/17481322/53270256-3ce4fb80-36b9-11e9-9b73-2bd06adbc075.png">

<img width="1256" alt="screen shot 2019-02-22 at 3 37 28 pm" src="https://user-images.githubusercontent.com/17481322/53270268-42424600-36b9-11e9-83ad-efab010e0111.png">

A possible work around is to add my own tbody, tr, and td tags like this:

<img width="674" alt="screen shot 2019-02-22 at 4 00 23 pm" src="https://user-images.githubusercontent.com/17481322/53270919-fee8d700-36ba-11e9-8924-2f32169495d5.png">

That said, my workaround is to place the empty state outside the table in a div and apply my own styling. The problem with that is everyone will likely implement that differently.

Ideally, TableBody could be modified to include the empty state as a child? Or, create another component (e.g., TableEmptyState) just for this purpose that spans all columns? 

That way, default padding, height, etc. could be applied uniformly.

Kyle's Mock:
<img width="1501" alt="screen shot 2019-02-22 at 2 58 34 pm" src="https://user-images.githubusercontent.com/17481322/53267880-867e1800-36b2-11e9-9ede-49b254a61358.png">

rachael-phillips:
@mcoker making you aware of this issue :) What do you think next steps should be here?

mcoker:
[...]

PR #2696, opened 2019-08-13
feat(Table): add empty state demo

kmcfaul:
**What**: adds an example showing a table using the empty state component for cases with no rows.

**Refer to issue**: #1433

### DIFF
--- a/packages/patternfly-4/react-core/src/layouts/Bullseye/examples/Bullseye.md
+++ b/packages/patternfly-4/react-core/src/layouts/Bullseye/examples/Bullseye.md
@@ -14,7 +14,7 @@ import './bullseye.scss';
 import React from 'react';
 import { Bullseye } from '@patternfly/react-core';
 
-<Bullseye>
+<Bullseye className="layout-example">
   <div>Bullseye ◎ layout</div>
 </Bullseye>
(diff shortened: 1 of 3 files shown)


redallen:
Something doesn't seem right in Firefox 68:
image

kmcfaul:
Huh that's pretty odd. It seems like taking it out of the Bullseye sorts it out, but I'm not sure why. Any thoughts @mcoker?

jschuler:
It's also getting the border from workspace styles that should not be applied here. Should only be visible on the layouts page to visualize it.

kmcfaul:
Is there a way to remove that? I'm not seeing where that styling gets applied.

jschuler:
@kmcfaul i think it's coming from here `packages/patternfly-4/react-core/src/layouts/Bullseye/examples/bullseye.scss`

kmcfaul:
Design undergoing review, PR will likely be changing so do not merge tag added.

mcarrano:
I have opened a design/core issue here: https://github.com/patternfly/patternfly-next/issues/2171

tlabaj:
Closing this PR.  We will re-open it once design is ready.

PR #2698, opened 2019-08-13
feat(TextArea): allow textarea to be uncontrolled

seanforyou23:
**What**: Allows for setting `defaultValue` instead of `value` in the case where we want to use it as an uncontrolled component.

Helps us get past this error, which we get when using TextArea in our form builder wrapper component.

Screen Shot 2019-08-12 at 5 04 11 PM


tlabaj:
do you mind opening any issue to track this.  Thanks!

PR #2701, opened 2019-08-14
feat(Icons - config): Publish configs as well with icons

karelhala:
**What**:
Some usecases require the original SVG path to be used instead of React component (for instance when using icons in svg). So this PR will introduce iconConfig (for each icon) alongSide the exported icon.

**Additional issues**:


packages/react-icons/build/templates/iconFile.hbs
```diff
@@ -1,14 +1,14 @@
...
   width: {{width}},
   svgPath: '{{svgPath}}',
   yOffset: '{{yOffset}}',
   xOffset: '{{xOffset}}',
   transform: '{{transform}}'
-});
+};
 
-export default {{name}};
+export default createIcon({{name}}Config);
```
redallen: I like this approach.

Issue #2425, opened 2019-07-02
PF4: make common components OUIA compatible


quarckster:
This issue tracks implementation of OUIA compatibility for common components:

- [ ] Alert
- [ ] Breadcrumb
- [ ] Button
- [ ] Nav
- [ ] Pagination #2387
- [ ] Switch #2304 
- [ ] Tabs

PR #2704, opened 2019-08-14
feat(Tabs): Added OUIA compatibility to Tabs component

abalakh:
**What**: Added OUIA props to Tabs as per #2425

**Additional issues**: 2 Tabs tests are failing after this change. Any help with figuring out what's causing the issue is appreciated.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/__snapshots__/Tabs.test.tsx.snap
+++ b/packages/patternfly-4/react-core/src/components/Tabs/__snapshots__/Tabs.test.tsx.snap
@@ -1,79 +1,105 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`should call handleScrollButtons tabs with scrolls 1`] = `
-<Tabs
-  activeKey={0}
-  className=""
+<Component
   id="handleScrollButtons"
   isFilled={true}
-  isSecondary={false}
-  leftScrollAriaLabel="Scroll left"
-  onSelect={[Function]}
-  rightScrollAriaLabel="Scroll right"
-  variant="div"
 >
-  <div
-    className="pf-c-tabs pf-m-fill"
-    onSelect={[Function]}
-  >
-    <button
-      aria-label="Scroll left"
-      className="pf-c-tabs__scroll-button"
-      onClick={[Function]}
-    >
-      <AngleLeftIcon
-        color="currentColor"
-        noVerticalAlign={false}
-        size="sm"
-        title={null}
-      >
-        <svg
-          aria-hidden={true}
-          aria-labelledby={null}
-          fill="currentColor"
-          height="1em"
-          role="img"
-          style={
-            Object {
-              "verticalAlign": "-0.125em",
-            }
-          }
-          viewBox="0 0 256 512"
-          width="1em"
-        >
-          <path
-            d="M31.7 239l136-136c9.4-9.4 24.6-9.4 33.9 0l22.6 22.6c9.4 9.4 9.4 24.6 0 33.9L127.9 256l96.4 96.4c9.4 9.4 9.4 24.6 0 33.9L201.7 409c-9.4 9.4-24.6 9.4-33.9 0l-136-136c-9.5-9.4-9.5-24.6-.1-34z"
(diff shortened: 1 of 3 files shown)


quarckster:
@redallen @karelhala @jschuler

packages/patternfly-4/react-core/src/components/Tabs/__snapshots__/Tabs.test.tsx.snap
```diff
@@ -1,1665 +1,1162 @@
...
-  <ForwardRef
-    activeKey={0}
-    child={
-      <ForwardRef
-        eventKey="three"
-        id="tab3"
-        title="Tab item 3"
-      >
-        Tab 3 section
-
```
abalakh: @redallen first off, thanks for help. But are you sure this snap will actually cover test scenario? It's basically empty, and it doesn't look like it will ensure "should render tabs with eventKey Strings".
Appologies if it's a dumb question, i'm pretty new to jest testing, trying to understand to not introduce same mistakes in future PRs.

abalakh: The diff is too big so GH points to wrong line. My question was re this snap:
```
exports[`should render tabs with eventKey Strings 1`] = `
<ContextConsumer>
  <Component />
</ContextConsumer>
`;
```

redallen: Yep, it makes sense to make the snapshot tests that use OUIA use `mount` instead of `shallow` since there's now a `<ContextConsumer>`.

abalakh:
@redallen so i returned branch to previous state with `mount` instead of `shallow`. The thing is, 2 tests are failing with `mount` (snaps for them aren't updated as they can't be generated yet), see ci/circleci: test_jest_pf4 result. Any ideas what's causing failures?

redallen:
Yes, this is the same issue I fixed yesterday. You have to fix the selectors Jest uses for those tests since now there's a `<ContextConsumer><Component /></ContextConsumer>` in the VDOM.

abalakh:
@redallen not sure if i follow. Those 2 failing tests are comparing mount result with snap, no interactions and no css selectors are used there.

karelhala:
Thanks for adding this! I think that the best way would be to replace `shallow` test renders with `render` from enzyme. We don't really test ouia so no need to check if that is present in snapshots and mount just adds a lot of code to snaps that we really don't care right? Plus render is a bit faster than mount (doesn't have to call componentDidMount etc.).

redallen:
Good call @karelhala , I've updated the PR with that, and I wouldn't have thought of that :+1: 

I am rather confused why `mount` was having problems only for these two tests:
```jsx
test('should render secondary tabs', () => {
  const view = render(
    <Tabs id="primarieTabs">
      <Tab eventKey={0} title="Tab item 1">
        <Tabs id="secondaryTabs">
          <Tab id="secondary tab1" eventKey={10} title="Secondary Tab 1">
            Secondary Tab 1 section
          </Tab>
          <Tab id="secondary tab2" eventKey={11} title="Secondary Tab 2">
            Secondary Tab 2 section
          </Tab>
          <Tab id="secondary tab3" eventKey={12} title="Secondary Tab 3">
            Secondary Tab 3 section
          </Tab>
        </Tabs>
      </Tab>
      <Tab id="tab2" eventKey={1} title="Tab item 2">
        Tab 2 section
      </Tab>
      <Tab id="tab3" eventKey={2} title="Tab item 3">
        Tab 3 section
      </Tab>
    </Tabs>
  );
});
```
```jsx
test('should render tabs with eventKey Strings', () => {
  const view = render(
    <Tabs id="eventKeyTabs">
      <Tab id="tab1" eventKey={'one'} title="Tab item 1">
        Tab 1 section
      </Tab>
      <Tab id="tab2" eventKey={'two'} title="Tab item 2">
        Tab 2 section
      </Tab>
      <Tab id="tab3" eventKey={'three'} title="Tab item 3">
        Tab 3 section
      </Tab>
    </Tabs>
  );
});
```

Issue #1600, opened 2019-03-18
Bullet Chart


rachael-phillips:
This issue was created in reference to https://github.com/patternfly/patternfly-next/issues/871

Cc @dlabrecq

dlabrecq:
I believe we still want to port the bullet chart. Spent most of this time addressing basic architecture issues in react-charts, first.

PR #2708, opened 2019-08-15
feat(charts): add bullet chart

dlabrecq:
This PR adds the bullet chart component.

Note that new theme variables will be added to PatternFly core via a separate PR.

Worked from @mceledonia 's color mock, using the latest PF color scales:
https://docs.google.com/document/d/1cw10pJFXWruB1SA8TQwituxn5Ss6KpxYPCOYGrH8qAY/edit

And included features from the existing design on pf.org:
See https://www.patternfly.org/v3/pattern-library/data-visualization/bullet-chart/#design

Fixes https://github.com/patternfly/patternfly-react/issues/1600

There are various examples available via the demo, but here are a few snapshots:

Screen Shot 2019-08-15 at 1 56 07 PM

Screen Shot 2019-08-15 at 2 01 56 PM

Screen Shot 2019-08-15 at 1 56 29 PM

Screen Shot 2019-08-15 at 1 56 41 PM

Screen Shot 2019-08-15 at 1 57 38 PM

Screen Shot 2019-08-15 at 1 57 50 PM


packages/patternfly-4/react-charts/src/components/ChartUtils/chart-domain.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (yHighSearch) {
+      if (dataPoint._y > yHigh) {
+        yHigh = dataPoint._y;
+      }
+    }
+  });
+  return {x: [xLow, xHigh], y: [yLow, yHigh]};
+};
+
+export const getDomains = ({
```
redallen: Where is `getDomains` used? If it's a helper function for the consumer, we can include it in the docs.

dlabrecq: It's used by the ChartBullet component to figure out the min/max domain based on the given data points. The utility / helper functions are intended for use by chart components -- don't really want/need to expose that in the top level doc.

redallen: Ah -- I missed that it's used in the component when I used Ctrl+F on Github. Thanks.

Issue #2700, opened 2019-08-14
Tooltip arrows can't point at the wrong node or nothing at all


rhamilto:
See https://github.com/openshift/console/pull/2345 for details.

cc:  @janwright73, @dgutride, @spadgett

PR #2711, opened 2019-08-15
Improve tooltip and popover positioning

jschuler:
- Added `auto` option to `position` prop so that it can place the tooltip/popover where there's space
- Use tippy's arrow and style it with PF. This way we get the auto-positioning of the arrow for free. For example if the tooltip/popover content is getting pushed by its boundary, the arrow will continue to point at the wrapped node.
- Expose tippy's boundary prop. Default it to `window`.

Closes: #2700


redallen:
Snapshots needs updating.

redallen:
More snapshots :(

rhamilto:
Thanks, @jschuler.  I think the distance from the arrow to the node it points at can be problematic in some cases (see attached screenshots).

Screen Shot 2019-08-15 at 3 39 31 PM
Screen Shot 2019-08-15 at 3 39 37 PM
Screen Shot 2019-08-15 at 3 39 44 PM

PR #2733, opened 2019-08-20
feat(PF4 - Drawer): Update drawer to expose all parts

karelhala:
**What**:

Thank you @dlabaj for experimental Drawer component! Usage is nice and straight forward, what do you think about exposing DrawerContent as well? This way users can add the `DrawerPanelContent` when they want to, for instance when data in it should be loaded from server so they want to show it only when user clicks on some element to show it. Also special classes and attributes can be aplied to Content so for instance if app has some special attributes to track where user is located they can do that easilly with this.

It's essentially the same, except that content is not part of Drawer itself, but special component and Panel is not sent trough prop but as child.

### DIFF
--- a/packages/patternfly-4/react-core/src/experimental/components/Drawer/Drawer.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/Drawer/Drawer.tsx
@@ -11,8 +11,6 @@ export interface DrawerProps extends React.HTMLProps<HTMLDivElement> {
   isExpanded?: boolean;
   /** Indicates if the content element and panel element are displayed side by side. */
   isInline?: boolean;
-  /** Content rendered in the drawer */
-  panelContent?: React.ReactNode;
 }
 
 export const Drawer: React.SFC<DrawerProps> = ({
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/experimental/components/Drawer/Drawer.tsx
```diff
@@ -11,24 +11,19 @@ export interface DrawerProps extends React.HTMLProps<HTMLDivElement> {
   isExpanded?: boolean;
   /** Indicates if the content element and panel element are displayed side by side. */
   isInline?: boolean;
-  /** Content rendered in the drawer */
-  panelContent?: React.ReactNode;
```
redallen: Experimental :tada:

karelhala: Exactly! Good way we added it in experimental so we can improve the API before it's full fledged component.

packages/patternfly-4/react-core/src/experimental/components/Drawer/__snapshots__/Drawer.test.tsx.snap
```diff
@@ -4,13 +4,9 @@ exports[`Drawer isExpanded = false and isInline = false 1`] = `
 <div
   className="pf-c-drawer"
 >
-  <div
-    className="pf-c-drawer__content"
-  >
-     
+  <Component>
```
jschuler: Should change test to use mount instead of shallow so we can get more info here

karelhala: Sure thing!

PR #2735, opened 2019-08-20
feat(Ansible icon): Add ansible tower icon to custom icons definition

karelhala:
**What**:
Since Ansible Tower is RH product, let's add it's icon to our icon drawer as well so app devs can include it in their apps.

Screenshot from 2019-08-20 10-55-21

**Additional issues**:


AllenBW:
This is SO long over due. 🥰🥰🥰🥰

PR #2749, opened 2019-08-21
fix(switch): use defaultChecked instead of checked

### LINKED ISSUE #2669: Turning switch on/off does not change anything in DOM
I'm looking at examples in docs (both patternfly-react.surge.sh and patternfly.org). Switch's input is loaded with or without "checked" attribute (`<input class="pf-c-switch__input" ... checked>`), but when changing switch state that attr is not added/removed. No classes or anything else are updated as well.
Moreover, both labels (for "on" and "off" states) are present in DOM at the same time, so from test automation perspective i have no idea what state switch is currently in and which label is actually shown.
Making "checked" attr reflect actual switch state (or intoducing some other attr/class/etc) would help a lot.

redallen:
**What**: Fixes #2669 and actually changes the `<input>`'s `checked` prop in the DOM for our Switch. This is a somewhat tricky React bug, but the solution is to use `defaultChecked` instead of `checked`. Setting `checked={true}` or `checked={false}` has no effect in the DOM outside of the first render, which is semantically incorrect.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/DataList/DataListCheck.tsx
+++ b/packages/patternfly-4/react-core/src/components/DataList/DataListCheck.tsx
@@ -37,7 +37,7 @@ export const DataListCheck: React.FunctionComponent<DataListCheckProps> = ({
         onChange={(event) => onChange(event.currentTarget.checked, event)}
         aria-invalid={!isValid}
         disabled={isDisabled}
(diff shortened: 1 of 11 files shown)


boaz0:
@redallen seems like you need to resolve a conflict or two.

redallen:
Rebased.

PR #2754, opened 2019-08-21
feat(virtual-table): debounce resize event

priley86:
**What**:
After further review in OpenShift, we need to debounce/throttle the resize event which handles resizing the virtualized grid. Failure to do so can lead to inconsistent grid sizing.

See fix in [console](https://github.com/openshift/console/pull/2419). 

**Additional issues**:


tlabaj:
HI @priley86 is there an issue associated with this one?

priley86:
opened [2798](https://github.com/patternfly/patternfly-react/issues/2798) for this just now @tlabaj

PR #2755, opened 2019-08-21
Docs: target the container with more specificity

jschuler:
Previous selector was also targeting the div appended by the [2nd Popover example](https://patternfly-react.surge.sh/patternfly-4/components/popover/)


jschuler:
Not exactly sure what you mean @redallen , but gatsby adds 2 outer divs which we both want to set to 100% height
Screen Shot 2019-08-21 at 2 30 37 PM

Issue #2328, opened 2019-06-21
PF4: React Virtualized tests


priley86:
Add unit/e2e tests for `@patternfly/react-virtualized-extension`.

jenny-s51:
https://github.com/priley86/patternfly-react/pull/10

tlabaj:
closed by #2761

PR #2761, opened 2019-08-22
PF4: (virtualized-extension) jest tests for VirtualizedTable

jenny-s51:
WIP for #2328 -- currently working on resolving these type errors shown after running `yarn run jest virtualizedTable`. 

Screen Shot 2019-08-22 at 3 42 47 PM


jenny-s51:
Resolving TS errors with `any` type for now. Can change the types to what's more appropriate, but need some feedback on what would be best here. In addition, `/Virtualized`  and`/Virtualized/utils` are now showing up in the coverage report, but coverage still really needs to be improved (13.23%  and 19.29%, respectively). Will be working on this

packages/patternfly-4/react-core/src/declarations.d.ts
```diff
@@ -6,3 +6,24 @@ declare module 'victory-core' {
   export const Path: any;
   export const TextSize: any;
 }
+
+declare module 'linear-layout-vector' {
```
priley86: hey @jenny-s51 - my only ounce of feedback so far here would be - can we move this declaration into patternfly-4/react-virtualized-extension some place? I believe `linear-layout-vector` is only a dependency of that package (so probably unnecessary in react-core).

Also, i think you can safely remove package-lock.json as well. We don't ship lock files to consumers (just reference the dependences in the package.json for each package) and all of the dev dependencies for PF React end up getting locked in the root `yarn.lock` file (which should update when you install new packages or run `yarn install` in the root).

Great job w/ this, it's a huge help!!! ⭐️

priley86:
hey @jenny-s51 - i've pushed a commit w/ some proposed changes to this PR today:
https://github.com/priley86/patternfly-react/commits/iss2328

Let me know what you think. I'm mostly happy here aside from one warning in the jest log still...
```
 PASS  packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualizedTable.test.tsx
  ● Console

    console.error node_modules/prop-types/checkPropTypes.js:20
      Warning: Failed prop type: Invalid prop `children` supplied to `Table`, expected a ReactNode.
          in Table
    console.error packages/patternfly-4/react-table/dist/js/components/Table/Table.js:237
      Table: Specify at least one of: header, caption, aria-label
    console.error node_modules/prop-types/checkPropTypes.js:20
      Warning: Failed prop type: Invalid prop `children` supplied to `Provider`, expected a ReactNode.
          in Provider (created by Table)
          in Table (created by WrapperComponent)
          in WrapperComponent
    console.error node_modules/react-dom/cjs/react-dom.development.js:180
      Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Componen
t /> from render. Or maybe you meant to call this function rather than return it.
          in table (created by Provider)
          in Provider (created by Table)
          in Table (created by WrapperComponent)
          in WrapperComponent
    console.error packages/patternfly-4/react-table/dist/js/components/Table/Table.js:237
      Table: Specify at least one of: header, caption, aria-label
```

Did not get a chance to address it. I will try testing these changes downstream tomorrow and report back though!

packages/patternfly-4/react-table/src/components/Table/Table.tsx
```diff
@@ -180,7 +180,7 @@ export interface TableProps {
   contentId?: string;
   dropdownPosition?: 'right' | 'left';
   dropdownDirection?: 'up' | 'down';
-  rows: (IRow | string[])[];
+  rows?: (IRow | string[])[];
```
priley86: 👍

package.json
```diff
@@ -117,5 +117,8 @@
     "packages": [
       "packages/**"
     ]
+  },
+  "dependencies": {
+    "@types/react-virtualized": "^9.21.4"
```
priley86: i don't think this is needed in the root anymore since we've added it in react-virtualized-extension's package.json

packages/patternfly-4/react-table/src/components/Table/Table.test.tsx
```diff
@@ -179,6 +179,16 @@ test('Collapsible nested table', () => {
   expect(view).toMatchSnapshot();
 });
 
+test('Header only table', () => {
```
priley86: 👍

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/utils/CellSizeAndPositionManager.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable */
 
-import LinearLayoutVector from 'linear-layout-vector';
+const LinearLayoutVector = require("linear-layout-vector");
```
priley86: is this still necessary? (changing import to require)

priley86:
hey @jenny-s51 - i'm mostly OK w/ this change. I would just remove any package-lock.json files (since we havent been checking those in)

mturley:
@jenny-s51 +1 about `package-lock.json`.. and actually, if that file is being created you're probably using the `npm` CLI, I'd make sure you're running `yarn install` instead of `npm install` because we do want any updates to `yarn.lock` 🙂

jenny-s51:
@priley86 do you mean I can just remove `package-lock.json` from this PR entirely?

redallen:
@jenny-s51 We use `yarn` with `yarn.lock` to install packages, so yes please remove `package-lock.json`!

jenny-s51:
@redallen gotcha, should be all set now -- thank you!

package.json
jenny-s51: how would you suggest I address this? would it just be `yarn remove @types/react-virtualized`?

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/__snapshots__/VirtualizedTable.test.tsx.snap
```diff
@@ -0,0 +1,7975 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Actions virtualized table 1`] = `
```
mturley: These huge snapshots are pretty unwieldy, it's hard to tell exactly what they are trying to assert. These tests are basically just "it renders correctly" tests, which is fine, but any little code change will cause these to fail and need updated snapshots, so it's up to the person updating the snapshots to look through them and determine if the change was bad or not.

I think that can be addressed as a future enhancement, these are better than no tests. In the future it might be nice to add some tests that use shallow rendering and assert some specific behavior of the code in reaction to props or state changing, or simulated user interaction. Those kinds of smaller, more targeted tests will be more likely to only fail if something is really wrong.

package.json
mturley: I think so, yes, or you could remove the line from the package.json and re-run `yarn install`. If the build still passes without this dependency in here, yeah, I think remove it. But it's probably ok either way.

packages/patternfly-4/react-virtualized-extension/src/components/Virtualized/VirtualizedTable.test.tsx
```diff
@@ -0,0 +1,203 @@
+import * as React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
mturley: I think you can remove this `eslint-disable`, you use `clsx` below so the linter is probably not calling it an extraneous dependency anymore

package.json
redallen: You can delete the line. Likely no need to re-run `yarn install`, but better safe than sorry.

mturley:
It looks like you have a merge conflict in `yarn.lock`. A good way to fix this is to first reset your `yarn.lock` to the version from master (`git checkout master yarn.lock`) and then run `yarn install` which will update that lockfile based on the changes you made to `package.json`.

jenny-s51:
@mturley thanks for all your help on this!

package.json
```diff
@@ -116,5 +116,7 @@
     "packages": [
       "packages/**"
     ]
+  },
+  "dependencies": {
```
redallen: Remove this empty `dependencies` section.

packages/patternfly-4/react-core/src/declarations.d.ts
```diff
@@ -1,8 +0,0 @@
-// This is for charts for jest.
```
redallen: Thanks for getting rid of this!

packages/patternfly-4/react-virtualized-extension/package-lock.json
```diff
@@ -0,0 +1,5779 @@
+{
```
mturley: @jenny-s51 sorry if I missed this before, I'm not sure when it snuck back in, but you've still got a `package-lock.json` file in here you should delete. Make sure you're not using the `npm` CLI commands in a yarn project, this file comes from using `npm install` instead of `yarn install`. `yarn.lock` is the only lockfile we do want to keep.

jenny-s51: Sorry about that -- I'm not sure how this snuck back in either! I've been using `yarn add`.

mturley: Might have happened in a rebase. Old changes have a way of sneaking back in during a complex rebase :)

Issue #2536, opened 2019-07-18
Add ability to use icons with Tabs


SpyTec:
Right now in PF4, you cannot add PF icons to Tab components. This issue is a feature request to add that functionality

My first thought is having a new prop for icon where the React component is sent in

rachael-phillips:
Hi @SpyTec thanks for filing this! @mcarrano is this something we ever talked about in the design process or would recommend?

mcarrano:
@rachael-phillips we had not considered an option to add icons to tabs. @SpyTec can you give me a bit more background about your usage and why you need this?

SpyTec:
As an example of what I mean:

![image](https://user-images.githubusercontent.com/6598829/61464280-4d528700-a976-11e9-9708-10b681065750.png)

I could just see this as better UX as you can at a glance see what each tab is, such as you would on a navigation. On smaller devices or screens one could even reduce the overall size by only showing icon

![image](https://user-images.githubusercontent.com/6598829/61464534-b1754b00-a976-11e9-9283-81aa98c36aa6.png)

kybaker:
@mcarrano My only concern with this proposed solution is that it maintains accessibility when only showing an icon in the smallest view port size.

mcarrano:
@SpyTec @rachael-phillips We currently have some work planning to rework our tab designs here https://github.com/patternfly/patternfly-next/issues/1855  We can factor this in.

karelhala:
No need to add new prop to tab component. You should be able to pass any component to title of tab
`<Tab eventKey={0} title={<div><UserFriendsIcon/>Users</div>}>` however the prop is wrongly defined as it states that you can pass sring only and it will throw error in console because of this.

karelhala:
> Speaking of different resolutions, I would do something want something like this as well

This doesn't really have to be part of PF core styles, it is nice to have but not something critical. I would assume that if you want such thing you'd have to write styles for this. Because you'd also have to provide accessibility attributes and such.

SpyTec:
> No need to add new prop to tab component. You should be able to pass any component to title of tab
> `<Tab eventKey={0} title={<div><UserFriendsIcon/>Users</div>}>` however the prop is wrongly defined as it states that you can pass sring only and it will throw error in console because of this.

We can't simply update the type definitions however since TabProps extends React.HTMLProps. We'd then get this error:
```
[...]

PR #2681, opened 2019-08-11
feat(tab): support JSX elements in Tab title

boaz0:
**What**:

closes #2623 

In general Tab supported JSX elements the problem is that the `title` props type was set to string. In addition, `Tab` didn't use the `title` prop at all, which was confusing, because it was passed as a child in the `Tabs` component. So what I did was removing `title` from props. 

//cc @xeviknal

PR #2766, opened 2019-08-23
fix(Tab icon): Change tab title prop to allow custom components

karelhala:
Fixes: #2536

**What**:
Allow consumers to pass whatever component they choose in tab title to allow icons and such be part of tab.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/Tab.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Tabs/Tab.test.tsx
@@ -11,6 +11,15 @@ test('should render tab', () => {
   expect(view).toMatchSnapshot();
 });
 
+test('should render tab with node', () => {
+  const view = shallow(
+    <Tab eventKey={0} title={<span>Tab item 1</span>}>
+      Tab 1 section
+    </Tab>
+  );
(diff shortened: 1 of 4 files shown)


karelhala:
@mcarrano Additional spacing shouldn't be required here (or shouldn't be styled by core team), if someone wants to add icon in he tab title they can also use some layout - level or so to add some spacing. Or adding custom styles for themselves.

karelhala:
@mcoker does it really have to be specific styles applied to icon only? What if someone would like to add two spans or would like to flip the order of icon and text? I feel like it would be much better to style children of tab component (does't matter if it's span, svg, div, w/e) instead of force direct flow and add specific prop just to add icon in there.

karelhala:
Nevertheless this PR is not about icon, but to allow any HTML element in tab component. And I feel we can merge this one and continue the discussion in PF core issue addressing styling of childre of certain components. I feel like for instance button component should allow flipping the order of text and icon as well, and I think that it can be achieved by styling children elements instead of adding specific styles to one specific element.

karelhala:
Closing in favor of https://github.com/patternfly/patternfly-react/pull/2681 we can continue discussion of spacing in https://github.com/patternfly/patternfly-next/issues/2195

PR #2769, opened 2019-08-23
fix(tooltip): allow manul trigger, access to tippy.js props

jeff-phillips-18:
**What**:
Allows passing properties through to PopoverBase and hence tippy.js
Adds the manual trigger option with the isVisible prop.


packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -53,23 +53,28 @@ export interface TooltipProps {
...
   position?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
-  /** Tooltip trigger: click, mouseenter, focus */
+  /** Tooltip trigger: click, mouseenter, focus, manual  */
   trigger?: string;
+  /** value for visibility when trigger is 'manual' */
+  isVisible?: boolean;
   /** z-index of the tooltip */
   zIndex?: number;
+  /** additional Props to pass through to tippy.js */
+  tippyProps?: TippyProps;
```
jschuler: i think we should just add what we need in terms of props instead of a passthrough? otherwise could be more maintenance if some tippy props don't behave as expected in our component

jeff-phillips-18: I think it's better to allow the full set to be passed thru. How can we know up front which props consumers want to customize? Limiting the options will cause limits to its adoptability. If there are any issues with using a particular prop, then an issue can be raised.

jschuler: what about the fact that there'll be duplicate props, for example we expose `enableFlip` which would be the same as `tippyProps.flip`

jeff-phillips-18: whatever props we accept/set will override what is in the tippyProps

jschuler: okay let's go with this, can you also add the passthrough to the Popover component as well then? Thanks

jeff-phillips-18:
@jschuler @dtaylor113 Added passthru props to the Popover component as well.

PR #2771, opened 2019-08-23
feat(catalog-tile): custom icon jsx for react-extensions CatalogTile

### LINKED ISSUE #2770: Catalog Tile - custom icon
As a consumer of PF4 React Extensions, I'd like to be able to fully customize the icon provided to the Catalog Tile. This would allow me to pass custom JSX (including PF React Icons).

Todo:
* add an `icon` prop that can replace the current JSX rendered for the `iconImg` or the `iconClass`.

https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react-extensions/src/components/CatalogTile/CatalogTile.js#L120

priley86:
**What**:
Closes #2770 

cc: @jeff-phillips-18 @christianvogt

**Additional issues**:


christianvogt:
@priley86 the other icon options apply `catalog-tile-pf-icon`. The size is no longer controlled if supplying a component as an icon. Is this intended? Do you expect the dev to explicitly supply an `catalog-tile-pf-icon` or equivalent?

jeff-phillips-18:
@priley86 @christianvogt I think wrapping the provided icon component in a span with the `catalog-tile-pf-icon` class would work to keep the icon the correct size.

priley86:
updated and tried to simplify the logic using new default for `iconClass`. Let me know if this works. 

@jeff-phillips-18 @christianvogt

PR #2774, opened 2019-08-26
fix(Modal): FocusTrap works with showClose being false

### LINKED ISSUE #2757: [PF4] FocusTrap throws exception on Modal when showClose prop is false
When creating a modal with `showClose` to false it will throw an error with FocusTrap

```js
The above error occurred in the <FocusTrap> component:
    in FocusTrap (created by ModalContent)
    in div (created by Backdrop)
    in Backdrop (created by ModalContent)
    in ModalContent (created by Modal)
    in Modal (at App.tsx:140)
    in div (created by Grid)
    in Grid (at App.tsx:139)
    in App (created by inject-App-with-serverStore-configStore)
    in inject-App-with-serverStore-configStore (at src/index.tsx:23)
    in Provider (at src/index.tsx:22)

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries. index.js:1437
Error: You can't have a focus-trap without at least one focusable element
```

## Reproducible steps
1. Create a Modal that is open and has `showClose` set to false

## Reproducible sandbox
https://codesandbox.io/embed/fast-river-zlm5n?fontsize=14

SpyTec:
**What**: Fixes FocusTrap breaking on Modal if `showClose` prop is false.

**Additional issues**: Fixes #2757

I have not been able to test this out as I cannot build PatternFly at the moment.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
@@ -37,7 +37,7 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   isLarge?: boolean;
   /** Creates a small version of the Modal */
   isSmall?: boolean;
-  /** Flag to disable focus trap if no focusable elements are available */
+  /** Flag to disable focus trap */
   disableFocusTrap?: boolean;
 }
 
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.tsx
@@ -37,6 +37,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   isLarge?: boolean;
   /** Creates a small version of the Modal */
   isSmall?: boolean;
+  /** Flag to disable focus trap if no focusable elements are available */
+  disableFocusTrap?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


SpyTec:
Just noticed that this breaks the [Modal (no header)](https://patternfly-react-pr-2774.surge.sh/patternfly-4/components/modal/)

SpyTec:
@redallen I've updated based on your feedback, thanks :)

I also took the liberty of using Bullseye instead of css when FocusTrap isn't used.

SpyTec:
I didn't even consider testing those props, I'll give it a try to see if we can just use that instead :)

I noticed in one of the other PF4 components that we went with the solution that's currently in this PR

SpyTec:
Sounds like a good idea to me, though is there a way to hint this when an error is thrown if a user stumbles upon the issue? Such as catching the error and adding a message to it about the prop.

SpyTec:
Added `disableFocusTrap`. FocusTrap should now be active at all times unless teh prop is true.

I also fixed ModalContent documentation to have description, which it didn't before

jeff-phillips-18:
@SpyTec Don't see the doc change to `Modal` though the commit message indicates you meant to include it.

SpyTec:
Sorry, glanced over that part. Fixed Modal prop description :)

Issue #94, opened 2017-12-04
Ensure storybook-deployer can support npm run scripts 


priley86:
We noted in #93 that storybook-deployer does not support custom npm run script for `build-storybook`. I will follow up w/ an upstream PR to ensure this support and then circle back to pf-react.

priley86:
Upstream PR is up:
https://github.com/storybooks/storybook-deployer/pull/28

PR #2796, opened 2019-08-29
area chart - horizontal indicator example

priley86:
**What**:
Horizontal Indicator example for Area Charts. Currently reviewing this from a UX/design perspective.

Opened PF issue [2797](https://github.com/patternfly/patternfly-react/issues/2797) for this documentation.

Demo (Multi-color chart with threshold indicators and responsive container example):
https://patternfly-react-pr-2796.surge.sh/patternfly-4/charts/chartarea/

Current Questions:
~~* What color themes do we think appropriate for indicators?~~
~~* What stroke width do we find legible for indicators?~~
~~* Should indicators be connected lines or independent line segments?~~
~~* How should we handle mobile legends gracefully?~~
~~* Do we need custom threshold tooltip behavior?~~

Desktop:
chart-desktop

Tablet:
chart-tablet

Mobile:
mobile

### DIFF
--- a/packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
+++ b/packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
@@ -2,10 +2,19 @@
 title: 'Area'
 section: 'charts'
 typescript: true
-propComponents: ['Chart', 'ChartAreaProps', 'ChartAxis', 'ChartGroup', 'ChartLegend', 'ChartVoronoiContainer']
+propComponents:
+  [
+    'Chart',
+    'ChartAreaProps',
+    'ChartAxis',
+    'ChartGroup',
+    'ChartLegend',
+    'ChartThreshold',
+    'ChartVoronoiContainer',
+  ]
 ---
 
-import { Chart, ChartArea, ChartAxis, ChartGroup, ChartVoronoiContainer } from '@patternfly/react-charts';
+import { Chart, ChartArea, ChartAxis, ChartGroup, ChartLegend, ChartThreshold, ChartThemeColor, ChartThemeVariant, ChartVoronoiContainer } from '@patternfly/react-charts';
 import '@patternfly/patternfly/patternfly-charts.css';
 import './chart-area.scss';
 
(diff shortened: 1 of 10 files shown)


packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
```diff
@@ -224,6 +224,122 @@ class MultiColorChart extends React.Component {
...
+        <div className="area-chart-legend-bottom-responsive">
+          <Chart
+            ariaDesc="Average number of pets"
+            ariaTitle="Area chart example"
+            containerComponent={<ChartVoronoiContainer labels={datum => `${datum.name}: ${datum.y}`} />}
+            legendData={[
+              { name: 'Cats' },
+              { name: 'Birds' },
+              { name: 'Cats Threshold', symbol: { fill: ChartThemeColor.blue, type: 'dash' }},
+              { name: 'Birds Threshold', symbol: { fill: ChartThemeColor.orange, type: 'dash' }}
```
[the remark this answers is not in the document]
dlabrecq: I believe we should change those values to something like `blue: "blue-theme-color"`, so they are not used incorrectly. I'll create an issue to do that.

priley86:
linking [#94](https://github.com/RedHatInsights/curiosity-frontend/pull/94) related (and started today in Insights Subscriptions)

jeperry:
Not sure where to note this, but i think overlapping area charts with transparency makes this trickier. I'd like to see this applied to a stacked area chart as well. Speaking of which.. we don't seem to have one?

dlabrecq:
A threshold line can be applied on top of other areas / lines, grids, etc. to be more visible. It's simply a matter of ordering.

Regarding stack chart, you can see that here:
https://patternfly-react.surge.sh/patternfly-4/charts/chartstack/

mceledonia:
Hey, working on getting a stacked chart example for you here, and I'll try to answer those questions:

---

What color themes do we think appropriate for indicators?
- I almost feel like they should be treated the same as the values, just the next colors in the color order. So if you have three values and 2 threshold indicators, you'd have something like this:

Value 1 = Color 1
Value 2 = Color 2
Value 3 = Color 3
Threshold 1 = Color 4
Threshold 2 = Color 5

Is something like the above possible?

---

What stroke width do we find legible for indicators?
- It looks like you have 2px which should be a good choice here. What variables can we control for the dotted line with victory? i.e. do we have control over dash length + gap size?

---

Should indicators be connected lines or independent line segments?
- I'm not sure I'm following this one, but I like the dashed line style which differentiates from the solid line style we use for values in line charts.

---

How should we handle mobile legends gracefully?
- I'd be curious if victory offers any solutions or has any examples of solutions for this. It seems like a problem universal to charts in general, but there's definitely a few solutions we could explore. One is a simple wrapping list under the chart, similar to what you have above in the mobile example but "birds threshold" would wrap to a second line.

---

Do we need custom threshold tooltip behavior?
- I don't think it would hurt for thresholds to have the same tooltip behavior as the values do, but if that's a technical challenge I don't think it's necessary by any means.

dlabrecq:
Considering the threshold is just another line, the ChartLine component would simply apply the next color in the color scale / family.

We can customize the dash length and gap size. However, if a threshold requires different properties, we may need to consider creating a new component?

Regarding the dashed lines, Cost Management uses both solid and dashed lines to show the difference between the current and previous month's data. Can the dash symbol be optional depending on your usage? Perhaps we can suggest possible alternatives?

Tooltips are a definitely a technical challenge. What did you have in mind?

jeperry:
This looks really great, and the proposed rules @mceledonia has here would work for the use case I am aware of for Subscriptions UI (one threshold line not specifically related to any of the 2-3 area values).

mceledonia:
@dlabrecq I don't think the tooltips are necessary for the thresholds (chime in if anyone disagrees here) but figured if they're easy to toss in, they wouldn't hurt. I'm thinking the juice isn't worth the squeeze there.

For the dashed lines, since we have control over gap length and size, it's possible we recommend a dash style for values and a dash style for thresholds. They can be distinguishable enough to work I think, here's a quick example:

image

priley86:
i've started a `ChartThreshold` component just now alongside this PR. No visual changes yet for the dash style, but it would be defaulted inside the `ChartThreshold` component and overridable by the consumer.

@dlabrecq @cdcabrera - please weigh in and give a bit more specific feedback here once ready. I am happy to move forward however best!

dlabrecq:
IMO, we need a different dash style for thresholds. That's the main reason for creating a new ChartThreshold component (i.e., so users won't need to override styling).

//cc @mceledonia

dlabrecq:
Took the liberty of creating a POC, showing how to create a new symbol for thresholds. You can find the code here: https://github.com/patternfly/patternfly-react/pull/2884

New threshold sybmol
Screen Shot 2019-09-08 at 10 57 07 PM

Existing dash symbol
Screen Shot 2019-09-08 at 11 04 14 PM

priley86:
ok, after chatting w/ @dlabrecq - the proposal is to use the new `threshold` legend symbol created in #2884 instead of `dash` symbol. Also, we can move `ChartThreshold` towards a theme for adding the default `strokeDashArray` and add a `getThresholdTheme` function (similar to ChartDonut's `getDonutTheme` function [here](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartUtils/chart-theme.ts#L91))

packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
```diff
@@ -242,6 +251,139 @@ class MultiColorChart extends React.Component {
...
+                    name: 'Birds Threshold',
+                    symbol: { fill: ChartThemeColor.orange, type: 'threshold' }
+                  }
+                ]}
+                itemsPerRow={2}
+              />
+            }
+            height={250}
+            padding={{
+              bottom: 100, // Adjusted to accomodate legend
```
priley86: adding some additional height in new container class `area-chart-threshold-bottom-responsive` and inside the legend bottom padding to accommodate multi row legend

packages/patternfly-4/react-charts/src/components/ChartThreshold/ChartThreshold.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { getThresholdTheme } from '../ChartUtils/chart-theme';
+
+export const ChartThreshold: React.FunctionComponent<ChartLineProps> = ({
+  themeColor,
+  themeVariant,
+  style, // do not let style defaults override threshold theme
+  ...rest
+}: ChartLineProps) => {
+  const theme = getThresholdTheme(themeColor, themeVariant);
+  return (
```
priley86: when extending ChartLineProps this way, it seems `style` prop will override `theme`. This method works for allowing theme colors, but would prevent the consumer from override `style` and `theme` directly. Any suggestions @dlabrecq ? I think this is close to what we discussed...

packages/patternfly-4/react-charts/src/components/ChartTheme/themes/threshold-theme.ts
```diff
@@ -0,0 +1,10 @@
+// Threshold theme
+export const ThresholdTheme = {
+  line: {
+    style: {
+      data: {
+        strokeDasharray: '6,6'
```
priley86: updated to `6,6`

packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
```diff
@@ -224,6 +224,122 @@ class MultiColorChart extends React.Component {
...
+              <ChartLine
+                data={[
+                  { name: 'Cats Threshold', x: 0, y: 4},
+                  { name: 'Cats Threshold', x: 3, y: 4},
+                  { name: 'Cats Threshold', x: 3, y: 6},
+                  { name: 'Cats Threshold', x: 5, y: 6},
+                ]}
+                style={{
+                  data: {
+                    stroke: ChartThemeColor.blue,
```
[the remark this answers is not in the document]
priley86: now using new `ChartThresholdTheme` instead

priley86:
thanks for the extensive feedback, and finally circling back here....here's my updates on this:

* added new `threshold` symbol for legend
* made legend multi-line using `itemsPerRow`
* added new `ChartThresholdTheme` and made default `strokeDasharray` 6,6 instead of 3,3
* updated screenshots above

Noted that this should still be safe to consume later in Subscriptions (current usage here):
https://github.com/RedHatInsights/curiosity-frontend/blob/ci/src/components/chartArea/chartArea.js#L216

packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
```diff
@@ -242,6 +251,140 @@ class MultiColorChart extends React.Component {
...
+              />
+              <ChartThreshold
+                data={[
+                  { name: 'Birds Threshold', x: 0, y: 2 },
+                  { name: 'Birds Threshold', x: 2, y: 2 },
+                  { name: 'Birds Threshold', x: 2, y: 3 },
+                  { name: 'Birds Threshold', x: 5, y: 3 }
+                ]}
+                themeColor={ChartThemeColor.orange}
+                themeVariant={ChartThemeVariant.light}
```
dlabrecq: You can remove the `themeVariant`, it defaults to `ChartThemeVariant.light`

packages/patternfly-4/react-charts/src/components/ChartThreshold/ChartThreshold.tsx
dlabrecq: I believe this is expected behavior. For example, when overriding the `padding` prop, users must provide all variables; top, bottom, left, right.

mceledonia:
LGTM!

PR #2802, opened 2019-08-29
appendTo prop

jschuler:
Closes: https://github.com/patternfly/patternfly-react/issues/2551

Adds `appendTo` prop to the Modal which makes it more flexible to where the DOM element can get appended to

Examples:
1. 
```
<Modal
          appendTo={document.getElementById('my-target')} />
```
2.
```
<Modal
          appendTo={() => {
            const innerFrame = document.getElementById('my_iframe').contentDocument;
            return innerFrame.getElementsByTagName('body')[0];
          }} />
```


sjd78:
@jschuler - This looks pretty good, and I know it's merged, but I do have a question.

Doesn't the `appendTo` still really need to be a `document.body` for the backdrop styles to be applied correctly and for the ESC key events to be handled properly?  In other words, will your example 1 actually work?

In the case of my PR #2574, and Issue #2551, the situation is the react app running in an iframe and needing to render the component up in the frame's parent...
```jsx
<Modal appendTo={window.top.document.body} />
```
...so I don't see an immediate issue for us.

Issue #2362, opened 2019-06-26
Code style/formatting rules for TypeScript are not enforced by TSLint


mturley:
In patternfly-4/react-core, in our tslint.json we extend `tslint-config-prettier`, but we're not running Prettier when we lint our TypeScript.

From https://www.npmjs.com/package/tslint-config-prettier:

> Do you want to use TSLint and Prettier without conflicts?  `tslint-config-prettier` disables all conflicting rules that may cause such problems. Prettier takes care of the formatting whereas tslint takes care of all the other things.

With this configuration, TSLint is no longer enforcing any code formatting / style rules. For code formatting we need to run Prettier, either by using its CLI directly, or by adding `tslint-plugin-prettier` to our tslint configuration.

I noticed these rules were not being enforced when I reviewed this commit: https://github.com/patternfly/patternfly-react/commit/6eb1ffd5fe651e0f99eeb17174fcbf5d67d133fc#diff-d2816c60d6fa971b0f62310ff6bef46eR26
Note the lack of indentation in that block.
Here's the CI task that ran tslint on that commit and passed: https://circleci.com/gh/patternfly/patternfly-react/3602

@redallen suggested that maybe we should remove `eslint-plugin-prettier` from our eslint configuration and use the `prettier` CLI directly for both eslint and tslint, to get a consistent configuration. This needs more investigation.

redallen:
Prettier CLI docs: https://prettier.io/docs/en/cli.html

mturley:
Copying my relevant Slack comments here; we may want to get rid of TSLint and use `typescript-eslint` instead of fixing the TSLint config. Then we could use `eslint-plugin-prettier` for both languages.

> I wonder if we should consider moving away from tslint entirely in favor of `typescript-eslint`? Looks like it's going to have better support moving forward, Palantir announced they will be deprecating TSLint, and the core TypeScript team is switching over:
> https://eslint.org/blog/2019/01/future-typescript-eslint
> https://github.com/typescript-eslint/typescript-eslint#what-about-tslint
> It might be nice to publish one recommended linter config for consumers to use no matter whether they are writing JS or TS.
> I guess it would need to be a slightly different config for TS to enable the TS parser. But that could extend the JS rules and it would still be simpler to use one linter.
> Prettier supports both JS+TS out of the box too so we could include eslint-plugin-prettier without any worry about config differences

redallen:
[...]

PR #2787, opened 2019-08-27
Improve linting

redallen:
**What**:
- Tweak tslint.json to be nicer
- Fix autofixable errors using new tslint.json

Closes #2362

cc @mturley 

**Additional issues**:

PR #2810, opened 2019-08-30
Add tslint-plugin-prettier and apply code formatting

mturley:
**What**:

Because we realized that [some important code-formatting-related TSLint rules are broken](https://github.com/patternfly/patternfly-react/pull/2787#issuecomment-526628618), @redallen and I decided that it would be appropriate to add [Prettier](https://prettier.io/) to our linter, and run it on existing code.

This PR:
* Adds `tslint-plugin-prettier` which runs Prettier as part of TSLint (currently these errors do not fail the CI checks, but eventually when other TSLint errors are resolved they probably should)
* Re-adds `tslint-config-prettier` which was removed in #2787, this disables TSLint rules which would conflict with Prettier
* Re-adds the `yarn prettier` script which contributors can run to automatically format their code
* Includes code changes from an initial run of `yarn prettier` and `yarn lint:ts --fix`. This cleans up all of our indentation issues and other small formatting gripes.

Hopefully this will help to make our code a little more readable and consistent going forward!

cc @redallen 

Closes https://github.com/patternfly/patternfly-react/issues/2362 (for real this time)


mturley:
Hmm, `yarn lint:ts --fix` still makes some conflicting changes. Hang on for one more commit.

mturley:
Ok, this should be all set now. I ran `yarn lint:ts --fix` and it had some more changes to make, then ran `yarn prettier` again and it made no changes, so they don't conflict.

mturley:
Rebased on master

redallen:
Let's try to do get this in post code-freeze next week after one final rebase. Things are a little hectic now.

mturley:
Fair enough 🙂

mturley:
I rebased this on the latest master, and squashed it into two commits to make it easier to rebase in case we need to do so again:

* https://github.com/patternfly/patternfly-react/pull/2810/commits/a56c6b88578b047b9bab16e2f4ed37a00aa328a3 makes the actual configuration changes, this commit should be included in a rebase
* https://github.com/patternfly/patternfly-react/pull/2810/commits/51a85c05483fa40044710c934de9124742ed6d1e runs the linter and formatter on existing code, this commit should be dropped and replaced in a rebase

mturley:
Hmm.. @redallen I'm having trouble parsing the errors in the CircleCI log, do you know why this would be failing CI? I didn't make any different configuration changes compared to before the rebase.

redallen:
@mturley During `yarn build` it's something to do with Chart types:
```
src/components/ChartVoronoiContainer/ChartVoronoiContainer.tsx(153,6): error TS2322: Type '{ activateData?: boolean; activateLabels?: boolean; disable?: boolean; labels?: (point: any, index: number, points: any[]) => string; mouseFollowTooltips?: boolean; onActivated?: (points: any[], props: VictoryVoronoiContainerProps) => void; ... 14 more ...; theme: ChartThemeDefinitionInterface; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<VictoryVoronoiContainer> & Readonly<VictoryVoronoiContainerProps> & Readonly<{ children?: ReactNode; }>'.
  Property 'className' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<VictoryVoronoiContainer> & Readonly<VictoryVoronoiContainerProps> & Readonly<{ children?: ReactNode; }>'
```

Charts has been touched a lot lately, so I'd recheck the merge conflicts for that file. Worst case, checkout master's version and then relint the file.

mturley:
> Worst case, checkout master's version and then relint the file
Hmm, that's what I did. I'll take a closer look at that file, thanks!

packages/patternfly-4/react-charts/src/components/ChartVoronoiContainer/ChartVoronoiContainer.tsx
```diff
@@ -143,16 +140,18 @@ export const ChartVoronoiContainer: React.FunctionComponent<ChartVoronoiContaine
...
     constrainToVisibleArea,
     theme,
-    ...labelComponent.props,
+    ...labelComponent.props
   });
 
   // Note: theme is required by voronoiContainerMixin, but @types/victory is missing a prop type
   // @ts-ignore
-  return <VictoryVoronoiContainer className={chartClassName} labelComponent={chartLabelComponent} theme={theme} {...rest} />;
+  return (
```
mturley: Ahh, @redallen it looks like it was this `@ts-ignore` not being applied anymore because the offending code was moved down a line by prettier. I'll add a `prettier-ignore` on the same line, and we might have to do that in the future where we need `@ts-ignore` on a long line.

It's weird that we can't just `ts-ignore` multiple lines... looks like an open issue: https://github.com/Microsoft/TypeScript/issues/19573

mturley:
I rebased this on the latest master, and squashed it into fewer commits to make it easier to rebase in case we need to do so again:

* https://github.com/patternfly/patternfly-react/pull/2810/commits/61fe744dd9cb868e3526ab7f078e53b427ba4a4b makes the actual configuration changes, this commit should be included in a rebase
* https://github.com/patternfly/patternfly-react/pull/2810/commits/de1e7e5ec6ee32be25c6328cdb0847d4f1c9f806 stops prettier from making the build fail by removing ts-ignore from a line, this commit should be included in a rebase
* https://github.com/patternfly/patternfly-react/pull/2810/commits/e771fff6067e40866d37ce9b0400b63cb74ee8af runs the linter and formatter on existing code, this commit should be dropped and replaced in a rebase

(edited 2x)

mturley:
@dgutride do you think it would make sense to try and merge this (after a final rebase) immediately after this week's code freeze ends, to minimize conflicts with other PRs? It doesn't change our dist, but it touches a lot of files.

mturley:
Rebased and resolved conflicts with https://github.com/patternfly/patternfly-react/pull/2895

mturley:
@dlabrecq what formatting opinions of Prettier do you dislike? We can reconfigure it: https://prettier.io/docs/en/options.html. I'm not really concerned about how the code is formatted as long as it's formatted consistently. Messy indentation was my main gripe.

mturley:
I do agree somewhat that Prettier tries too much to fit the most code into every line, but I think maybe that could be helped by lowering the print-width option. They actually recommend that you don't exceed 80 characters on that option (see the yellow box at the top of the [options docs](https://prettier.io/docs/en/options.html), and we have it set to 120.

I would be open to shortening this to 80 or 100 (and then obviously there would be many lines which would end up longer than that).

dlabrecq:
@mturley I often use separate lines when I destructure multiple items or sometimes when importing long names, but that's just my preference to make things more readable. It's not a big deal, just not a fan. I am a big fan of running `yarn lint:ts --fix`, tho

mturley:
I agree actually, those sound like good reasons to disagree with Prettier. We could lower the print width to make that less of an issue, and in extreme cases we could use `// prettier-ignore` if it's being really fussy. This PR doesn't automatically run it going forward though, so we can always run it and then change things back if we disagree with it.

mturley:
@dlabrecq I'll bug you just one last time (I know you're not thinking of this as a big deal), but just for comparison's sake I made a few separate branches and ran the same process with the print-width set to 100 characters (https://github.com/mturley/patternfly-react/commit/c5fbfe4a62f6fcb0f1a5a7517da20d5b7c782325) and set to 80 characters (https://github.com/mturley/patternfly-react/commit/fc02c3ba249931e780a874bfd3fdaa4cfe14282d).

The lower the print-width, the more readable a lot of code gets, but files end up larger in terms of lines of code. Do you have an opinion on whether either of these would be better than the 120-character width in this PR?

packages/patternfly-4/react-charts/src/components/ChartBullet/utils/chart-bullet-size.ts
```diff
@@ -121,41 +126,34 @@ export const getQualitativeRangeBarWidth = ({
...
+    value: ChartBulletStyles.qualitativeRangeWidth,
+    width
+  });
+
+const scale = ({ defaultSize, height, horizontal = true, scale = 1, value, width }: ChartBulletScaleInterface) =>
+  horizontal
+    ? height > defaultSize
+      ? value + (height - defaultSize) * scale
+      : value - (defaultSize - height) * scale
+    : width > defaultSize
```
dlabrecq: This indentation looks much better compared to the others I noted

packages/patternfly-4/react-charts/src/components/ChartPoint/ChartPoint.tsx
```diff
@@ -75,8 +75,18 @@ export interface ChartPointProps {
...
+    | 'diamond'
+    | 'plus'
+    | 'minus'
+    | 'square'
+    | 'star'
+    | 'triangleDown'
+    | 'triangleUp'
+    | 'dash'
+    | 'threshold'
+    | Function;
```
dlabrecq: Better...

packages/patternfly-4/react-charts/src/components/ChartBullet/ChartBullet.tsx
```diff
@@ -720,12 +715,16 @@ export const ChartBullet: React.FunctionComponent<ChartBulletProps> = ({
...
+        ? -defaultPadding.bottom + 60
+        : -defaultPadding.bottom;
     } else if (legendPosition === ChartLegendPosition.bottomLeft) {
       dy = horizontal
-        ? defaultPadding.top * .5 + (defaultPadding.bottom * .5 - (defaultPadding.bottom)) - 25
-        : title ? -defaultPadding.bottom + 60 : -defaultPadding.bottom;
+        ? defaultPadding.top * 0.5 + (defaultPadding.bottom * 0.5 - defaultPadding.bottom) - 25
+        : title
+        ? -defaultPadding.bottom + 60
+        : -defaultPadding.bottom;
```
dlabrecq: This indentation looks a little off?

mturley: It looks like Prettier will indent a nested ternary if it is nested within the true-branch (`foo ? bar ? 1 : 2 : 3`) but not if it is trailing / nested in the false-branch (`foo ? 1 : bar ? 2 : 3`). This is a bit weird but seems reasonable to me... to be honest though, maybe we should avoid nested ternary expressions unless we use parentheses to make them more clear. Maybe that would help here.

mturley: Looks like this is a subject of debate in Prettier: https://github.com/prettier/prettier/issues/5814

mturley: Looks like this is a subject of debate in Prettier: prettier/prettier#5814

mturley: Huh. Adding parens around the second expression there, Prettier removes them. Maybe that's a case where `// prettier-ignore` would be useful if we need an exception to make that code extra clear.

dlabrecq: I didn't realize `// prettier-ignore` was available. That could be quite useful for those couple of weird issues

mturley: Agreed. You can even configure a `.prettierignore` file to ignore entire files/directories if necessary

mturley:
Rebased, if you want to re-approve @jschuler @dlabrecq

mturley:
🎶 _...one more time..._ 🎶 

giphy

PR #2816, opened 2019-09-02
fix(switch): backward compatibility in case labelOff is not set

### LINKED ISSUE #2814: [PF4] Switch label prop does not show when Switch is off
In https://github.com/patternfly/patternfly-react/pull/2468 the `label` prop was changed to only show when the Switch is on. Prior to this it was always showing `label` prop with no regard for if it was on or off.

`label` prop should always show unless `labelOff` is provided, otherwise this is a breaking change and should be rolled back and published in a major version

boaz0:
**What**:

Add a check if `label` is set but not `labelOff` - in that case display `label` on both labels.
if both `label` and `labelOff` are set display in the new format
Otherwise - display switch without labels.

Added tests to verify that if `labelOff` not given the label is shown.

fixes #2814 

//cc @SpyTec

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
@@ -70,7 +70,25 @@ class Switch extends React.Component<SwitchProps & InjectedOuiaProps> {
           aria-labelledby={isAriaLabelledBy ? `${this.id}-on`: null}
           {...props}
         />
-        {label !== '' || labelOff !== '' ? (
+        {label !== '' ? (
+          <React.Fragment>
+            <span className={css(styles.switchToggle)} />
+            <span
+              className={css(styles.switchLabel, styles.modifiers.on)}
+              id={isAriaLabelledBy ? `${this.id}-on` : null}
+              aria-hidden="true"
+            >
+              {label}
+            </span>
+            <span
+              className={css(styles.switchLabel, styles.modifiers.off)}
+              id={isAriaLabelledBy ? `${this.id}-off` : null}
+              aria-hidden="true"
+            >
+              {labelOff || label}
+            </span>
+          </React.Fragment>
+        ) : label !== '' && labelOff !== '' ? (
           <React.Fragment>
             <span className={css(styles.switchToggle)} />
             <span
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.test.tsx
@@ -28,6 +28,18 @@ test('switch is not checked', () => {
   expect(view).toMatchSnapshot();
 });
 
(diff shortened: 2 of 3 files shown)


packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -70,7 +70,25 @@ class Switch extends React.Component<SwitchProps & InjectedOuiaProps> {
...
+              aria-hidden="true"
+            >
+              {label}
+            </span>
+            <span
+              className={css(styles.switchLabel, styles.modifiers.off)}
+              id={isAriaLabelledBy ? `${this.id}-off` : null}
+              aria-hidden="true"
+            >
+              {label}
```
jschuler: should remove labelOff from the conditional above and simplify this. 
```suggestion
              {labelOff || label}
```

boaz0: If I'm not wrong, in that case labelOff is equal to `''` (empty string) which will evaluate `label` value.

boaz0:
fixed merge conflicts

jschuler:
Hi @boaz0 , what i meant is that we don't need to code duplication, we can simply fix the conditionals. 
Before:
```
        {label !== '' || labelOff !== '' ? (
          <React.Fragment>
            <span className={css(styles.switchToggle)} />
            <span
              className={css(styles.switchLabel, styles.modifiers.on)}
              id={isAriaLabelledBy ? `${this.id}-on` : null}
              aria-hidden="true"
            >
              {label}
            </span>
            <span
              className={css(styles.switchLabel, styles.modifiers.off)}
              id={isAriaLabelledBy ? `${this.id}-off` : null}
              aria-hidden="true"
            >
              {labelOff}
            </span>
          </React.Fragment>
        ) : (
          <span className={css(styles.switchToggle)}>
            <div className={css(styles.switchToggleIcon)} aria-hidden="true">
              <CheckIcon noVerticalAlign />
            </div>
          </span>
        )}
```
After:
```
        {label !== '' ? (
          <React.Fragment>
            <span className={css(styles.switchToggle)} />
            <span
              className={css(styles.switchLabel, styles.modifiers.on)}
              id={isAriaLabelledBy ? `${this.id}-on` : null}
              aria-hidden="true"
            >
              {label}
            </span>
            <span
              className={css(styles.switchLabel, styles.modifiers.off)}
              id={isAriaLabelledBy ? `${this.id}-off` : null}
              aria-hidden="true"
            >
              {labelOff || label}
            </span>
          </React.Fragment>
        ) : (
          <span className={css(styles.switchToggle)}>
            <div className={css(styles.switchToggleIcon)} aria-hidden="true">
              <CheckIcon noVerticalAlign />
            </div>
          </span>
        )}
```

boaz0:
@jschuler I got it.

Issue #2665, opened 2019-08-07
Creatable and New options support for select component


dlabaj:
PF4 select component is missing the ability to add creatable items as seen here: https://react-select.com/creatable .  This feature is needed by product teams.  Product teams also need the ability add new options by typing them in input (these options should stay in pool of options).

Please see issue #2531 for further details.

PR #2820, opened 2019-09-03
feat(Select): add creatable and new features

kmcfaul:
**What**: Adds the creatable and new use cases to select via `isCreatable` and `onCreateOption` properties. Specifically applies to typeahead variants.

**Refer to issue**: #2665


kmcfaul:
@mcarrano 
When a new item is entered, the addition occurs on the user-side. In the example, I just appended the option to the list. Select does not sort options.

Both new and creatable should work for multi-type ahead select as well. I will add the checkbox toggles to that example!

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/SelectDemo/SelectDemo.tsx
```diff
@@ -277,6 +294,12 @@ export class SelectDemo extends Component<SelectDemoState> {
     }
   }
 
+  typeaheadCreateNew = (newValue: string) => {
```
dlabaj: Should also allow SelectOptionObject.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/SelectDemo/SelectDemo.tsx
```diff
@@ -47,7 +50,15 @@ export class SelectDemo extends Component<SelectDemoState> {
...
     customTypeaheadMultiIsExpanded: false,
     customTypeaheadMultiSelected: [],
-    direction: SelectDirection.down
+    direction: SelectDirection.down,
+    typeaheadOptions: [
+      { value: 'Alabama', disabled: false},
+      { value: 'Florida', disabled: false },
+      { value: 'New Jersey', disabled: false }
+    ],
+    typeaheadCreateNew: undefined,
```
dlabaj: Why does this need to be set to undefined?

kmcfaul: It doesn't, I will remove

jschuler:
I'm getting the browser autofill popup, should be turned off?
Screen Shot 2019-09-05 at 4 04 14 PM

christiemolloy:
How is `isCreatable` different from `onCreateOption`?

When I have onCreateOption selected I can't create an option:
Screen Shot 2019-09-05 at 4 05 03 PM

christiemolloy:
For the multiple typeahead select input when I have Alabama filtered and then I search for Alabama again and add it, both of them delete and it wipes Alabama clear? I don't know if this is the expected behavior?

Screen Shot 2019-09-05 at 4 06 57 PM

christiemolloy:
When the menu expands up, should the arrow rotate @mcarrano 

Screen Shot 2019-09-05 at 4 09 32 PM

kmcfaul:
@jschuler autocomplete is set to off for all selects, so I'm unsure why it would be showing up. Which browser are you using?

@christiemolloy `isCreatable` handles whether the user can add inputs, but will only do so temporarily without `onCreateOption` defined. `onCreateOption` is a callback for the user to store the new option in the overall options list, and is used together with isCreateable. 

Re: multi-typeahead, it looks like you are selecting Alabama twice, first which selects it and second which unselects it and removes it form the chip group.

jschuler:
@kmcfaul strange, Chrome 76.0.3809.132. If no one else sees this then it might be something on my end

kmcfaul:
That's my version of Chrome and I'm not seeing the autocomplete on the surge. Very odd. Anyone else seeing it?

christiemolloy:
I see autocomplete in Chrome @kmcfaul @jschuler

jschuler:
@kmcfaul do you have autofill enabled? `chrome://settings/addresses`

kmcfaul:
I do have it enabled.

jschuler:
Might have to wrap in form tags as this comment suggests https://gist.github.com/niksumeiko/360164708c3b326bd1c8#gistcomment-3004386

kmcfaul:
That seems less than ideal, html structure wise. Will try it out though, hopefully shouldn't change anything visually.

jschuler:
@kmcfaul ok, i don't recall seeing autocomplete before, i wonder why it's showing now

mcarrano:
Regarding your question @christiemolloy:

> When the menu expands up, should the arrow rotate @mcarrano

I don't feel strongly about this, but I think we should just be consistent with what we do for other dropdown or select instances.

kmcfaul:
I think it makes sense to have the arrow icon change to an upward direction on direction toggle. This is more of a quick bugfix than related to this particular PR though. I can put up a different PR for it.

jschuler:
@kmcfaul wrapping inputs with form elements fixed the auto complete issue for me

PR #2821, opened 2019-09-03
docs(ApplicationLauncher): add examples for disabled, align right, an…

### LINKED ISSUE #2576: AppLauncher: Add support for right and top alignment
This PR was merged in Core: https://github.com/patternfly/patternfly-next/pull/2081

evwilkin:
…d top

**What**:

Fixes #2576 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
+++ b/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
@@ -62,7 +62,169 @@ class SimpleApplicationLauncher extends React.Component {
 }
 ```
 
-## Application launcher with tooltip
+## App launcher disabled
+```js
+import React from 'react';
+import { ApplicationLauncher, ApplicationLauncherItem } from '@patternfly/react-core';
+
+class SimpleApplicationLauncher extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: false
+    };
+    this.onToggle = isOpen => {
+      this.setState({
+        isOpen
+      });
+    };
+    this.onSelect = event => {
+      this.setState({
+        isOpen: !this.state.isOpen
+      });
+    };
+  }
+
+  render() {
+    const { isOpen } = this.state;
+    const appLauncherItems = [
+      <ApplicationLauncherItem key="application_1a" href="#">
+        Application 1 (anchor link)
+      </ApplicationLauncherItem>,
+      <ApplicationLauncherItem key="application_2a" onClick={() => alert('Clicked item 2')}>
+        Application 2 (div with onClick)
+      </ApplicationLauncherItem>,
+      <ApplicationLauncherItem key="application_3a" onClick={() => alert('Clicked item 3')}>
+        Application 3 (div with onClick)
+      </ApplicationLauncherItem>,
+      <ApplicationLauncherItem key="disabled_application_4a" isDisabled>
+        Unavailable Application
+      </ApplicationLauncherItem>
+    ];
+    return (
+      <ApplicationLauncher
+        onSelect={this.onSelect}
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
```diff
@@ -62,7 +62,169 @@ class SimpleApplicationLauncher extends React.Component {
...
+    this.state = {
+      isOpen: false
+    };
+    this.onToggle = isOpen => {
+      this.setState({
+        isOpen
+      });
+    };
+    this.onSelect = event => {
+      this.setState({
```
priley86: I think a better practice is to use the function form of `setState` for simple toggles like this:

```
this.setState( prevState => ({ isOpen: !prevState.isOpen }) );
```

https://medium.com/@wisecobbler/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1

redallen: Quote from the article:

> I am relying on the value of this.state.showForm to modify the next state of my form. In this simple example, we probably would not run into any issues by relying on this value, but you might imagine that as an application gets more complex and there are multiple setState calls happening and queuing up data to be rendered to the DOM, there is a possibility that the actual value of this.state.showForm might not be what you think.

I think it's fine to use `setState({})` for simple examples.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
```diff
@@ -62,7 +62,169 @@ class SimpleApplicationLauncher extends React.Component {
...
+        Unavailable Application
+      </ApplicationLauncherItem>
+    ];
+    return (
+      <ApplicationLauncher
+        onSelect={this.onSelect}
+        onToggle={this.onToggle}
+        isOpen={isOpen}
+        items={appLauncherItems}
+        isDisabled="true"
```
jschuler:
```suggestion
        isDisabled
```

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
```diff
@@ -62,7 +62,169 @@ class SimpleApplicationLauncher extends React.Component {
...
+        Application 2 (div with onClick)
+      </ApplicationLauncherItem>,
+      <ApplicationLauncherItem key="application_3a" onClick={() => alert('Clicked item 3')}>
+        Application 3 (div with onClick)
+      </ApplicationLauncherItem>,
+      <ApplicationLauncherItem key="disabled_application_4a" isDisabled>
+        Unavailable Application
+      </ApplicationLauncherItem>
+    ];
+    // const style = { marginTop: '100px' };
```
jschuler: Should be cleaned up here and line 220

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
evwilkin: Leaving as-is for now to match existing examples, we may want to refactor all examples if this change does end up being necessary.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
evwilkin: @jschuler updated

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.md
evwilkin: Removed unnecessary code

PR #2825, opened 2019-09-04
fix(HorizontalNavMenuItem): Allow HorizontalNavMenuItem's title to be a node

martinpovolny:
Allow HorizontalNavMenuItem's title to be a node rather than just a string.

Needed by: https://github.com/ManageIQ/react-ui-components/pull/115 and https://github.com/ManageIQ/manageiq-ui-classic/pull/5997

Ping @karelhala

### DIFF
--- a/packages/patternfly-3/patternfly-react/src/components/HorizontalNav/HorizontalNavMenuItem.js
+++ b/packages/patternfly-3/patternfly-react/src/components/HorizontalNav/HorizontalNavMenuItem.js
@@ -35,7 +35,7 @@ const HorizontalNavMenuItem = props => {
 HorizontalNavMenuItem.propTypes = {
   children: PropTypes.node,
   onItemClick: PropTypes.object,
-  title: PropTypes.string,
+  title: PropTypes.node,
   active: PropTypes.bool,
   dropdown: PropTypes.bool,
   submenu: PropTypes.bool,


packages/patternfly-3/patternfly-react/src/components/HorizontalNav/HorizontalNavMenuItem.js
```diff
@@ -35,7 +35,7 @@ const HorizontalNavMenuItem = props => {
 HorizontalNavMenuItem.propTypes = {
   children: PropTypes.node,
   onItemClick: PropTypes.object,
-  title: PropTypes.string,
+  title: PropTypes.node,
```
jeff-phillips-18: While node works for both, we typically have used:
```suggestion
  title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
to be more explicit

karelhala: Not sure that it's required node can be anything that can be rendered https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

jeff-phillips-18: Understood, just looking for consistency and explicitly noting in the docs.

martinpovolny: Guys, I do not care ;-)

I am seeing both variants in places. Tell me if I need to change it to get the PR merged or not.

Yes - change / No - leave it.

That's what I care about :-D Because w/o that I cannot finish the PRs above.

martinpovolny: Thx!

jeff-phillips-18:
@martinpovolny Could you rebase your branch to fix the build error?

martinpovolny:
Rebased.

PR #2830, opened 2019-09-04
feat(Wizard): add support for in page rendering

### LINKED ISSUE #2751: Provide In-page wizard demo
The related core issue is: https://github.com/patternfly/patternfly-next/issues/2073

suomiy:
fixes #2751

inpage

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
@@ -820,3 +820,32 @@ class SimpleWizard extends React.Component {
   }
 }
 ```
+
+### Wizard - in page
+
+```js
+import React from 'react';
+import { Wizard } from '@patternfly/react-core';
+
+class InPageWizard extends React.Component {
+  render() {
+    const steps = [
+      { name: 'Step 1', component: <p>Step 1</p> },
+      { name: 'Step 2', component: <p>Step 2</p> },
+      { name: 'Step 3', component: <p>Step 3</p> },
+      { name: 'Step 4', component: <p>Step 4</p> },
+      { name: 'Final Step', component: <p>Final Step</p>, hideCancelButton: true, nextButtonText: 'Close' }
+    ];
+
+    return (
+      <div style={{height: '400px'}}>
+        <Wizard
+          isInPage
+          onClose={() => console.log('closed')}
+          steps={steps}
+        />
+      </div>
+    );
+  }
+}
+```
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
@@ -43,8 +43,10 @@ export interface WizardStep {
 export type WizardStepFunctionType = (newStep: { id?: string | number; name: string; }, prevStep: { prevId?: string | number; prevName: string; }) => void;
 
 export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
-  /** True to show the wizard */
+  /** True to show the wizard (not applicable for isInPage)*/
   isOpen?: boolean;
+  /** True to show the wizard without the modal */
+  isInPage?: boolean;
   /** If true makes the navigation more compact */
   isCompactNav?: boolean;
   /** True to set full height wizard */
@@ -55,8 +57,8 @@ export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
   width?: number | string;
   /** Custom height of the wizard */
   height?: number | string;
-  /** The wizard title */
-  title: string;
+  /** The wizard title (required unless isInPage is used) */
+  title?: string;
   /** The wizard description */
   description?: string;
   /** Callback function to close the wizard */
@@ -100,9 +102,11 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
   private static currentId = 0;
   static defaultProps = {
     isOpen: false,
+    isInPage: false,
     isCompactNav: false,
     isFullHeight: false,
     isFullWidth: false,
+    title: '',
     description: '',
     className: '',
     startAtStep: 1,
@@ -122,12 +126,20 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
   private container: HTMLDivElement;
   private titleId: string;
   private descriptionId: string;
+  private isModal: boolean;
 
   constructor(props: WizardProps) {
     super(props);
     const newId = Wizard.currentId++;
-    this.titleId = `pf-wizard-title-${newId}`;
-    this.descriptionId = `pf-wizard-description-${newId}`;
+    this.isModal = !props.isInPage;
+    if (this.isModal) {
+      if (!props.title) {
+        console.warn('Title is required for modals!');
+      }
+      this.titleId = `pf-wizard-title-${newId}`;
+      this.descriptionId = `pf-wizard-description-${newId}`;
+    }
+
     this.state = {
       currentStep: this.props.startAtStep && Number.isInteger(this.props.startAtStep) ? this.props.startAtStep : 1,
       isNavOpen: false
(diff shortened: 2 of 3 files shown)


suomiy:
The css require at least `"@patternfly/patternfly": "2.28.0"`  dependency but only has `"@patternfly/patternfly": "2.27.3"`. Should it be resolved in this PR?

packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
```diff
@@ -820,3 +820,33 @@ class SimpleWizard extends React.Component {
...
+      { name: 'Step 3', component: <p>Step 3</p> },
+      { name: 'Step 4', component: <p>Step 4</p> },
+      { name: 'Final Step', component: <p>Final Step</p>, hideCancelButton: true, nextButtonText: 'Close' }
+    ];
+
+    return (
+      <div style={{height: '400px'}}>
+        <Wizard
+          isInPage
+          isOpen
```
suomiy: I kept the behaviour of `isOpen`. Meaning one has to always specify it when using in-page wizard. Otherwise nothing is rendered.

Should this behaviour be changed and render the in-page wizard always?

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -56,11 +58,11 @@ export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
   /** Custom height of the wizard */
   height?: number | string;
   /** The wizard title */
-  title: string;
+  title?: string;
```
suomiy: title is now optional, because the header is not rendered in in-page wizard

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -56,11 +58,11 @@ export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
...
   /** The wizard description */
   description?: string;
   /** Callback function to close the wizard */
-  onClose?: () => void;
+  onClose: () => void;
```
suomiy: I changed onClose to be required because otherwise the wizard crashes.

We can either:
- keep it optional. In that case multiple null checks have to be added to the wizard.
- change it to required which would be a breaking change

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -122,12 +125,17 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
   private titleId: string;
   private descriptionId: string;
+  private isModal: boolean;
 
   constructor(props: WizardProps) {
     super(props);
     const newId = Wizard.currentId++;
-    this.titleId = `pf-wizard-title-${newId}`;
-    this.descriptionId = `pf-wizard-description-${newId}`;
+    this.isModal = !props.isInPage;
```
suomiy: user cannot should not have an ability to change the inpage wizard to modal one after it is constructed

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -122,12 +125,17 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
+    if (this.isModal) {
+      this.titleId = `pf-wizard-title-${newId}`;
+      this.descriptionId = `pf-wizard-description-${newId}`;
+    }
```
suomiy: not sure if this is needed, but the titleId/descriptionId are not used when in-page

suomiy:
I guess we need to update `@patternfly/patternfly"` dependency for the tests to pass (fails on missing `inPage` modifier ).

should I do it in this PR?

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
suomiy: I removed the fix from this PR (as it is for a discussion) and created an issue instead: https://github.com/patternfly/patternfly-react/issues/2831

kmcfaul:
There's a PR up to bump the version to 2.28 so when that merges in it should resolve your issues.

kmcfaul:
@suomiy Version bump PR has been merged, try updating and that should solve the build error!

suomiy:
updated & tests passed

matthewcarleton:
@suomiy this looks awesome, thanks for doing this so quickly!

packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
jschuler: I think it makes sense to make isOpen the default in the code if isInPage is true. Just like any other in page components they should show by default, if user needs to toggle visibility they can do so on their end

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -414,61 +429,77 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
+          <div {...rest}
+               className={css(
+                 styles.wizard,
+                 !this.isModal && styles.modifiers.inPage,
+                 isCompactNav && 'pf-m-compact-nav',
+                 activeStep.isFinishedStep && 'pf-m-finished',
+                 setFullWidth && styles.modifiers.fullWidth,
+                 setFullHeight && styles.modifiers.fullHeight,
+                 className)}
+               role="dialog"
```
jschuler: i think this is only applicable to modal

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
mcoker: The default wizard requires a title currently, and per conversations with @mcarrano here (https://github.com/patternfly/patternfly-next/issues/2191) I believe we want to keep it that way. Making it optional creates these 2 scenarios:

Title/description is undefined renders an empty header, and the header is required at a minimum to position the close button.

Screen Shot 2019-09-05 at 9 18 11 AM

Title is undefined but a description exists, which seems to work OK, but we haven't really accounted for that in core and I'm not sure we want to allow that per the design:

Screen Shot 2019-09-05 at 9 21 06 AM

Can we make this optional if `isInPage` is `true`? And technically it isn't optional - the entire header shouldn't render if using an in-page wizard, so `title` and `description` shouldn't be allowed.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.md
suomiy: done

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
suomiy: fixed

suomiy:
> Can we make this optional if isInPage is true? And technically it isn't optional - the entire header shouldn't render if using an in-page wizard, so title and description shouldn't be allowed.

@mcoker 

These two are mutually exclusive from the type perspective. I.e. having title optional in page and required in modal.

Our options are:

Quick Patch
- leave it required. Users will be force to specify the title and this will have no effects
- leave it optional. Users will have mentioned positioning problems if they do not specify it

Fix Patch
- pass WizardHeader component to the wizard (breaking changes)
- fix css

mcoker:
> leave it required. Users will be force to specify the title and this will have no effects
> leave it optional. Users will have mentioned positioning problems if they do not specify it

I'm not sure between the two, what do you think @tlabaj?

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
tlabaj: We can change the title to optional here since it technically is not always required.  However, can you generate a warning if the title is not provided AND isModal is true.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -100,6 +102,7 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
   private static currentId = 0;
   static defaultProps = {
     isOpen: false,
+    isInPage: false,
```
jschuler: also want to add `title: ''` to defaultProps

jschuler:
Alternative workaround to optional title, in `WizardHeader` render empty non-breaking space if not provided
```
{title ? (
        <Title
          size="3xl"
          className={css(styles.wizardTitle)}
          aria-label={title}
          id={titleId}
        >
          {title}
        </Title>
      ) : (
        <Title
          size="3xl"
          className={css(styles.wizardTitle)}
          aria-label={title}
          id={titleId}
        >
          &nbsp;
        </Title>
      )}
```

suomiy:
I decided to go the `&nbsp;` route. Should I still add the warning?

suomiy:
Also, should I create an issue for that (to fix it properly and remove the non breaking space)?

jschuler:
@suomiy that would be great thanks. can you just also update the prop documentation for the `title` prop to say that it is required unless `isInPage` is set to true?

suomiy:
added the warning and the documentation

suomiy:
nice! :tada:

Issue #2744, opened 2019-08-20
Left-align buttons on modal 


rachael-phillips:
The issue opened in Core is: https://github.com/patternfly/patternfly-next/issues/1599

This issue is on hold until completed in Core.

tlabaj:
Closed by #2835

PR #2835, opened 2019-09-04
feat(Modal): add modifier to left align Modal footer

tlabaj:
#2744

**What**:

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/Modal.md
+++ b/packages/patternfly-4/react-core/src/components/Modal/Modal.md
@@ -309,6 +309,11 @@ class NoHeader extends React.Component {
 
   render() {
     const { isModalOpen } = this.state;
+    const footer = (
+      <React.Fragment>
+        Modal Footer
+      </React.Fragment>
+    );
 
     return (
       <React.Fragment>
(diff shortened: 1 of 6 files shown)


mcoker:
@mcarrano hmmm. We do have this example in core, but we show how to use the existing modal close button without a header - https://pf4.patternfly.org/components/ModalBox/examples/

Screen Shot 2019-09-05 at 12 07 14 PM

Personally I think we should update the react example to match, assuming that's the preferred design approach for a modal without a header (to use the existing close button versus a custom one), since we don't have an example for that in react currently.

Then I think we could discuss if we need to add another example that removes the modal's close button and demonstrates how to use a custom button to close the modal. If we do provide that example, I guess how we present that would depend on whether we want to either officially support or recommend a design for that use case. If we support it, we could add a variant to the footer that positions the custom close button. If we just recommend a design, maybe we use the flex layout as a demo of how we suggest doing it. Or do we just leave it as is, demonstrating how to add the button, but not necessarily recommending a design for the button's placement.

mcarrano:
In the context of this PR, I like your suggestion @mcoker . @tlabaj can we just update the 'Modal with no header' example to match core?

I think we could take up the question of whether we want an explicit Close button as a separate design issue.

Issue #2758, opened 2019-08-22
Add text to Split Button 


rachael-phillips:
Related Core issue: https://github.com/patternfly/patternfly-next/issues/2161

This issue is on hold until 2161 is completed in Core.

PR #2840, opened 2019-09-05
feat(Split button): Allow texts in split button

karelhala:
Fixes: #2758

**What**:
Allows passing children to split button in order to allow passing number of selected to this component.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
@@ -16,6 +16,8 @@ export interface DropdownToggleCheckboxProps extends Omit<React.HTMLProps<HTMLIn
   checked?: boolean | null;
   /** A callback for when the checkbox selection changes */
   onChange?(checked: boolean, event: React.FormEvent<HTMLInputElement>): void;
+  /** */
+  children?: React.ReactNode;
   /** Id of the checkbox */
   id: string;
   /** Aria-label of the checkbox */
@@ -38,7 +40,14 @@ export class DropdownToggleCheckbox extends React.Component<DropdownToggleCheckb
   }
 
   render() {
-    const { className, onChange, isValid, isDisabled, isChecked, checked, ...props } = this.props;
+    const { className, onChange, isValid, isDisabled, isChecked, checked, children, ...props } = this.props;
+    const text = children && <span
+      className={css(styles.dropdownToggleText, className)}
+      aria-hidden="true"
+      id={`${props.id}-text`}
+    >
+      {children}
+    </span>;
     return (
       <label className={css(styles.dropdownToggleCheck, className)} htmlFor={props.id}>
         <input
@@ -49,6 +58,7 @@ export class DropdownToggleCheckbox extends React.Component<DropdownToggleCheckb
           disabled={isDisabled}
           defaultChecked={isChecked || checked}
         />
+        {text}
       </label>
     );
   }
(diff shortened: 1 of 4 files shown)


karelhala:
I see that there is also PR https://github.com/patternfly/patternfly-react/pull/2838 for version bump. I'll wait for it to merge and rebase.

mcoker:
@mcarrano good catch, that's because there is a space character in the HTML example and no space in the react example. I don't think we want to rely on a space being there. I can submit a PR to address this in core and we can pull that in for this PR.

mcoker:
@karelhala if you can update to >= 2.31.3, the spacing should be fixed via https://github.com/patternfly/patternfly-next/issues/2223

Also looks like there are conflicts now.

karelhala:
@mcarrano @mcoker good catch! And thank you for the new styles update. I updated PF core to newest version, can you check if it looks good now?

mcarrano:
@karelhala @mcoker When I look at this in Firefox, looks like the alignment of the checkbox and text are a bit off.

Chrome:

Screen Shot 2019-09-06 at 9 13 33 AM

Firefox:

Screen Shot 2019-09-06 at 9 13 24 AM

mcoker:
@mcarrano checkboxes can be hard to style consistently cross-browser. We ran into this with the check component before, and I see the same thing now on that component

Screen Shot 2019-09-06 at 8 42 09 AM

Feel free to open an issue if you'd like us to look at it again. Using custom checkboxes should solve this.

packages/patternfly-4/react-inline-edit-extension/package.json
```diff
@@ -27,9 +27,15 @@
   },
   "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/",
   "dependencies": {
+<<<<<<< HEAD
```
redallen: You committed merge conflicts.

karelhala: Ah, thanks for noticing!

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
```diff
@@ -16,6 +16,8 @@ export interface DropdownToggleCheckboxProps extends Omit<React.HTMLProps<HTMLIn
   checked?: boolean | null;
   /** A callback for when the checkbox selection changes */
   onChange?(checked: boolean, event: React.FormEvent<HTMLInputElement>): void;
+  /** */
```
redallen: Add comment: `Element to be rendered inside the <span>`

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
```diff
@@ -49,6 +58,7 @@ export class DropdownToggleCheckbox extends React.Component<DropdownToggleCheckb
           disabled={isDisabled}
           defaultChecked={isChecked || checked}
         />
+        {text}
```
redallen: You can just put the whole statement here:

```
children && (
<span
  className={css(styles.dropdownToggleText, className)}
  aria-hidden="true"
  id={`${props.id}-text`}
>
  {children}
</span>
)
```

karelhala: Yeah, I did that and the TS lint complained about `jsx-no-multiline-js` https://github.com/palantir/tslint-react

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
karelhala: Yeah, good point. Missed that.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
redallen: I think you need the `(` and `)` to comply with that linting rule in the future. I'm fine with it either way.

Issue #2841, opened 2019-09-05
Dropdown Split button 3rd state input


karelhala:
Consumers should be able to add 3rd state in split button dropdown in order to show that current page is partially selected.

rachael-phillips:
Hi @karelhala thanks for filing. I'll follow up with members on our team to see what next steps look like here.

PR #2842, opened 2019-09-05
fix(Dropdown split button): Add 3rd state to split button

karelhala:
Fixes: #2841

**What**:
In order to allow 3rd state in checkbox of split button we have to use `Checkbox` component.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { DropdownToggleCheckbox } from './DropdownToggleCheckbox';
 
 const props = {
@@ -44,4 +44,15 @@ test('passing HTML attribute', () => {
     <DropdownToggleCheckbox label="label" aria-labelledby="labelId" id="check" isChecked aria-label="check" />
   );
   expect(view).toMatchSnapshot();
+});
+
+test('checkbox passes value and event to onChange handler', () => {
+  const newValue = true;
+  const event = {
+    target: { checked: newValue }
+  };
+  const view = mount(<DropdownToggleCheckbox id="check" {...props} aria-label="check" />);
+  view.find('input').simulate('change', event);
+  expect(props.onChange.mock.calls[0][0]).toBe(newValue);
+  expect(props.onChange.mock.calls[0][1]).toMatchObject(event);
 });
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
@@ -27,6 +27,11 @@ test('isDisabled', () => {
   expect(view).toMatchSnapshot();
 });
 
+test('3rd state', () => {
+  const view = shallow(<DropdownToggleCheckbox id="check" isChecked={null} aria-label="check" />);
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
```diff
@@ -28,26 +30,29 @@ export class DropdownToggleCheckbox extends React.Component<DropdownToggleCheckb
...
+    const { className, onChange, isValid, isDisabled, isChecked, ref, checked, ...props } = this.props;
     return (
       <label className={css(styles.dropdownToggleCheck, className)} htmlFor={props.id}>
-        <input
+        <Checkbox
           {...props}
-          type="checkbox"
-          onChange={this.handleChange}
+          {...(this.calculateChecked() !== undefined) && { onChange: this.handleChange }}
+          ref={ref as any}
```
karelhala: Anybody knows something better to use here?

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
```diff
@@ -35,13 +40,3 @@ test('passing HTML attribute', () => {
   );
   expect(view).toMatchSnapshot();
 });
-
-test('checkbox passes value and event to onChange handler', () => {
```
karelhala: I had to remove this test because jest is not playing nicely with mount and `currentTarget` https://github.com/airbnb/enzyme/issues/218.

I had to use mount because checkbox is now in `Checkbox` component.

redallen: I've run into this same issue (I think also with Checkbox). My solution was to change `handleChange` in `Checkbox.tsx` from:
```ts
private handleChange = (event: React.FormEvent<HTMLInputElement>): void => {
  this.props.onChange(event.currentTarget.checked, event);
}
```
to:
```ts
private handleChange = (event: React.FormEvent<HTMLInputElement>): void => {
  this.props.onChange(event.target.checked, event);
}
```

If Checkbox doesn't break, feel free to try that.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
redallen: Why do you have to pass the ref here? It will be passed automatically with `{...props}` with the correct type.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
redallen: I think we should open a high-priority issue to add this test back if we aren't willing to address it in the scope of this PR.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.tsx
karelhala: It's because TS does not recognized the ref correctly in here. I tried casting it to expected and it failed. This was the only solution that didn't thre any error.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownToggleCheckbox.test.tsx
karelhala: Yeah, that came up in head as well, I didn't wanted to change the original Checkbox though because it's not part of this PR. But I'll probably change it to use `target` instead of `currentTarget`.

I could also change `handleChange` of this component to use `target` instead of `currentTarget` which is probably a bit safer.

PR #2848, opened 2019-09-05
fix(Select): display selections data properly

### LINKED ISSUE #2828: [PF4] Changing selections value outside Select does not update correctly in Select component
Changing the `selections` prop outside the Select component does not update the component correctly. In typeahead, it shows the clear icon but does not show the text for the selected item. Clicking dropdown and then closing it makes it show up.

## Expected Results
Selected item is shown

## Actual results
Selected item is not shown

## Reproducible sandbox
This sandbox has a timeout of 2 seconds after mounted where it changes the selection. This is reflected in the UI with the clear icon showing.
https://codesandbox.io/s/modern-sky-2zhnh

SpyTec:
**What**: This fixes `selections` prop being properly displayed if updated externally.

**Additional issues**: Fixes #2828

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.tsx
@@ -120,6 +120,12 @@ export class Select extends React.Component<SelectProps, SelectState> {
         typeaheadFilteredChildren: React.Children.toArray(this.props.children)
       });
     }
+
+    if (prevProps.selections !== this.props.selections) {
+      this.setState({
+        typeaheadInputValue: (this.state.typeaheadActiveChild && this.state.typeaheadActiveChild.innerText)
+      });
+    }
   }
 
   onEnter = () => {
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -120,6 +120,12 @@ export class Select extends React.Component<SelectProps, SelectState> {
         typeaheadFilteredChildren: React.Children.toArray(this.props.children)
       });
     }
+
+    if (prevProps.selections !== this.props.selections) {
```
redallen: Change to `if (prevProps.selections !== this.props.selections && this.state.typeaheadActiveChild) {`

packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -120,6 +120,12 @@ export class Select extends React.Component<SelectProps, SelectState> {
...
+      this.setState({
+        typeaheadInputValue: (this.state.typeaheadActiveChild && this.state.typeaheadActiveChild.innerText)
```
redallen: Change to `typeaheadInputValue: this.state.typeaheadActiveChild.innerText`.

SpyTec:
Thanks @redallen, moved check in the latest commit :)

PR #2849, opened 2019-09-05
fix(ClipboardCopy): Update if children change

### LINKED ISSUE #2846: ClipboardCopy doesn't update when children are updated
Screen Recording 2019-09-05 at 9 18 04 AM mov

The `ClipboardCopy` component should be updating along with the counter in that video, but it doesn't.

Here's the code to generate that example:

```
class Test extends React.Component<{}, { counter: number }> {
    constructor(props) {
        super(props);
        this.state = { counter: 0 };
    }

    componentDidMount() {
        setInterval(
            () => this.setState({ counter: this.state.counter + 1 }),
            1000,
        );
    }

    render() {
        return (
            <React.Fragment>
                {this.state.counter.toString()}
                <ClipboardCopy>{this.state.counter.toString()}</ClipboardCopy>
            </React.Fragment>
        );
    }
}
```

SpyTec:
**What**: This updates ClipboardCopy to update component if children changes

**Additional issues**: fixes #2846

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
+++ b/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
@@ -91,6 +91,12 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
     toggleAriaLabel: 'Show content'
   };
 
+  componentDidUpdate = (prevProps: ClipboardCopyProps, prevState: ClipboardCopyState) => {
+    if (prevProps.children !== this.props.children) {
+      this.updateText(this.props.children as string | number);
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
```diff
@@ -91,6 +91,12 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
     toggleAriaLabel: 'Show content'
   };
 
+  componentDidUpdate = (prevProps: ClipboardCopyProps, prevState: ClipboardCopyState) => {
+    if (prevProps.children !== this.props.children) {
```
redallen: This check will work because we technically only support `children` being a string (despite it's type being ReactNode).

Issue #2782, opened 2019-08-26
Dark nav: Opt-in


rachael-phillips:
Relevant blog post: https://blog.patternfly.org/patternfly/how-do-different-patternfly-visual-themes-affect-navigation-performance-perception-and-preference/

Core issue: https://github.com/patternfly/patternfly-next/issues/2192

PR #2856, opened 2019-09-05
feat(Page, Nav): add dark theme modifier flag

kmcfaul:
**What**: adds dark theme opt-in option to page sidebar and nav components. Example added to PageLayout demos

**Refer to issue**: #2782


christiemolloy:
@kmcfaul In Core @mcoker updated all of the page demos to use the dark nav, the nav component to use the dark nav (except for one example with the white), and the page component to use the dark nav. If this issue is pressing I think it would be ok to merge it and then follow up with an issue to update all of those components.

jschuler:
If you change the Nav examples to use theme="dark" in the Nav.md file can you also add a frontmatter prop for `optIn` with a description?
e.g.
---
title: 'Nav'
cssPrefix: 'pf-c-nav'
typescript: true
propComponents: ['Nav', 'NavList', 'NavGroup', 'NavItem', 'NavItemSeparator', 'NavExpandable']
optIn: 'My description here'
---
@rachael-phillips is there a specific description we should use?

rachael-phillips:
> If you change the Nav examples to use theme="dark" in the Nav.md file can you also add a frontmatter prop for `optIn` with a description?
> e.g.
> 
> ```
> ---
> title: 'Nav'
> cssPrefix: 'pf-c-nav'
> typescript: true
> propComponents: ['Nav', 'NavList', 'NavGroup', 'NavItem', 'NavItemSeparator', 'NavExpandable']
> optIn: 'My description here'
> ---
> ```
> 
> @rachael-phillips is there a specific description we should use?

@christie @joachim does this description make sense to you "Opt-in and update to the latest design by using this prop"

kmcfaul:
Working on updating the demos and adding the legacy/light demo, will update asap

jschuler:
@rachael-phillips almost, have to also call out what the user needs to set I think. "In a future breaking-change release, the nav will default to the dark theme. You can opt-in and update to use the nav dark theme now by setting the theme prop to 'dark'"

rachael-phillips:
> @rachael-phillips almost, have to also call out what the user needs to set I think.
> 
> > In a future breaking-change release, the nav will default to the dark theme. You can opt-in and update to use the nav dark theme now by setting the theme prop to 'dark'.

@jschuler that looks fantastic to me. You're a genius - let's go with that. @kmcfaul any concerns or additions to that approach?

kmcfaul:
Looks good to me

kmcfaul:
Added in legacy/light example and swapped other page and nav examples to dark theme. Added in the `optIn` frontmatter prop to Nav (should this go on Page somewhere too?). Fixed the missing type from PageSidebar's theme prop.

kmcfaul:
@redallen Any thoughts on the docs build error? I tried nuking my cache and rerunning just that step locally but it runs fine

packages/patternfly-4/react-core/src/components/Nav/examples/nav.scss
```diff
@@ -0,0 +1,5 @@
+.ws-preview {
+  & .pf-m-dark {
+    background-color: var(--pf-global--BackgroundColor--dark-300);
```
jschuler: can you set this on .ws-preview directly? would look better :)

kmcfaul: Setting it there directly changes it for all examples haha xD
I'm not sure if there's a better alternative to get the same effect. Tried to apply it on a wrapper div but it looks the same as the current setup.

jschuler: ah, then it's fine

lboehling:
This looks great! Only one thing to note, the divider line in the expanded section should be pf-black-850 (#212427) and the divider line outside of expanded section on main nav bkg should be pf-black-700 (#4F5255).

kmcfaul:
> the divider line in the expanded section should be pf-black-850 (#212427) and the divider line outside of expanded section on main nav bkg should be pf-black-700 (#4F5255).

@mcoker Are these overrides I need put in the example css as well?

mcoker:
@kmcfaul nope, that's a bug I need to fix in core. We're using the `<Divider>` component in the core workspace, and the react workspace uses `<NavItemSeparator>`, which is still valid. I'll push a fix ASAP.

Issue #2009, opened 2019-05-14
Refactored Toolbar


rachael-phillips:
The related core issue is https://github.com/patternfly/patternfly-next/issues/1528

mattnolting:
I spoke to @mcarrano and @mceledonia about toolbar details. Here are the takeaways.

- Elements need to be moved in the DOM, based on content. 
Example 1: when multiple filters are present, these filters need to be moved to the `expandable-content` element and controlled via filter button toggle.

![Screen Shot 2019-05-14 at 3 38 49 PM](https://user-images.githubusercontent.com/5385435/57727050-86ad0480-765e-11e9-802c-615d1499fc1d.png)

![Screen Shot 2019-05-14 at 3 39 23 PM](https://user-images.githubusercontent.com/5385435/57727033-7d239c80-765e-11e9-983c-248b9a5b80fe.png)


Example 2: `primary` and `secondary` action buttons should move to the actions menu dropdown at specific breakpoint (this may be controlled by JS as well, we'll have to test and work w/design).

![Screen Shot 2019-05-14 at 3 38 13 PM](https://user-images.githubusercontent.com/5385435/57726969-54030c00-765e-11e9-9fb1-eb72273d3853.png)

- Toolbar may include `bulk-select`, `filter/multiple filters`, `overflow menu action button w/primary and secondary buttons`, `chip-group + chips` and `pagination`.

karelhala:
@nicolethoen any update on this? We want to change how our toolbars behave on smaller screens and would like to use these new components.

nicolethoen:
@karelhala it's all I'm doing :) hoping to have a PR up this week.

rachael-phillips:
Closing this issue

PR #2861, opened 2019-09-06
feat(DataToolbar): Add Data Toolbar component to experimental

nicolethoen:
[#2009 ](https://github.com/patternfly/patternfly-react/issues/2009)


packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarUtils.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface DataToolbarContextProps {
+  isExpanded: boolean;
+  toggleIsExpanded: () => void;
+  expandableContentRef: RefObject<HTMLDivElement>;
+  expandableContentId: string;
+}
+
+export const DataToolbarContext = React.createContext<Partial<DataToolbarContextProps>>({});
+
+export type DataToolbarBreakpointMod = {
```
karelhala: Please change the name to `DataToolbarBreakpointVariant`

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarUtils.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /** The attribute to modify  */
+  modifier: 'hidden' | 'visible' | 'align-right' | 'align-left';
+  /** The breakpoint at which to apply the modifier */
+  breakpoint: 'md' | 'lg' | 'xl' | '2xl';
+};
+
+export const formatBreakpointMods = (breakpointMods: DataToolbarBreakpointMod[]) => {
+
+  let formattedBreakpointMods = '';
```
karelhala: Let, assign and return can usually be rewritten to `reduce`
```JS
return breakpointMods.reduce((acc, curr) => (
  `${acc} ${getModifier(styles, `${curr.modifier}-on-${curr.breakpoint}`)}`
), '')
```

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarUtils.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /** The size of the spacer */
+  spacerSize: 'none' | 'sm' | 'md' | 'lg';
+  /** The breakpoint at which to apply the spacer */
+  breakpoint?: 'md' | 'lg' | 'xl';
+};
+
+export const formatItemSpacers = (spacers: DataToolbarSpacer[]) => {
+
+  let formattedSpacers = '';
+  for (const spacer of spacers) {
```
karelhala: Same as above, this can be rewritten to reduce
```JS
return spacers.reduce((acc, curr) => (
  `${acc} pf-m-spacer-${curr.spacerSize}${curr.breakpoint ? `-on-${curr.breakpoint}` : ''}`
), '')
```

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarUtils.tsx
```diff
@@ -0,0 +1,64 @@
...
+    if (spacer.breakpoint) {
+      formattedSpacers += '-on-' + spacer.breakpoint + ' ';
+    } else {
+      formattedSpacers += ' ';
+    }
+  }
+  return formattedSpacers;
+};
+
+export const formatGroupSpacers = (spacers: DataToolbarSpacer[]) => {
```
karelhala: This function can actually reuse the function above with one additional attribute `type` that is by default `pf-m-spacer-`

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  closeExpandableContent = () => {
+    this.setState(() => ({
+      isExpanded: false
+    }));
+  }
+
+  componentDidMount() {
+    window.addEventListener('resize', this.closeExpandableContent);
```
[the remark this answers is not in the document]
dlabaj: @karelhala There's an issue where if the toggle drawer is open and the window is resized when the breakpoint is hit the content will not layout correctly.  We're looking for a solution to make it behave the same way that dropdown does.  Will look at that code to see how that one works to see if there's a better solution.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
```diff
@@ -0,0 +1,399 @@
+---
+title: 'DataToolbar'
+cssPrefix: 'pf-c-data-toolbar'
+typescript: true
+propComponents: ['DataToolbar', 'DataToolbarContent', 'DataToolbarItem', 'DataToolbarGroup', 'DataToolbarToggleGroup', 'DataToolbarBreakpointMod', 'DataToolbarSpacer']
```
mattnolting: Looks like we need `DataToolbarSpaceItems` on order to apply `.pf-m-space-items{-on-[breakpoint]}` to groups

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { css, getModifier } from '@patternfly/react-styles';
+
+import {
+  DataToolbarBreakpointMod,
+  DataToolbarSpacer,
+  formatBreakpointMods,
+  formatItemSpacers
+} from './DataToolbarUtils';
+
+export enum DataToolbarItemMod {
```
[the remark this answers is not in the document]
nicolethoen: I added clear, but the shared-item modifier and chip-group are going to be used in the expandable content - and (probably) only used internally, applied programmatically. When the chip group is implemented soon it should be clear.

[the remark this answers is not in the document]
nicolethoen: and actually i just got clarification - clear is only intended to be used alongside multiple chip groups in the expandable content so it will not be provided by the consumer. so i think the existing variants are appropriate.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+  'overflow-menu' = 'overflow-menu',
+  pagination = 'pagination',
+  'search-filter' = 'search-filter',
+}
+
+export interface DataToolbarItemProps extends React.HTMLProps<HTMLDivElement> {
+  /** Classes applied to root element of the Data toolbar item */
+  className?: string;
+  /** A type modifier which modifies spacing specifically depending on the type of item */
+  mod?: DataToolbarItemMod | 'separator' | 'bulk-select' | 'overflow-menu' | 'pagination' | 'search-filter';
```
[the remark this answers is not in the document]
nicolethoen: clear is only intended to be used alongside multiple chip groups in the expandable content so it will not be provided by the consumer - it will only ever be added (in the datatoolbar) programmatically.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarUtils.tsx
mattnolting: - `__group`s should accept both `.pf-m-space-items` and `.pf-m-spacer` modifiers, or `DataToolbarSpaceItems` and `DataToolbarSpacer`
- `__item`s should accept only a `.pf-m-spacer` modifier, or `DataToolbarSpacer`

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
```diff
@@ -0,0 +1,399 @@
...
+               isDisabled={option.disabled}
+               key={index}
+               value={option.value}
+             />
+           ))}
+        </Select>
+      </DataToolbarItem>
+    </React.Fragment>;
+    
+    const items =  <DataToolbarToggleGroup toggleIcon={<FilterIcon />} breakpoint='xl'>{toggleGroupItems}</DataToolbarToggleGroup>;
```
mattnolting: 👍 💯

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
```diff
@@ -0,0 +1,399 @@
...
+    const toggleGroupItems = <React.Fragment>
+      <DataToolbarItem>
+        <InputGroup>
+          <TextInput name="textInput2" id="textInput2" type="search" aria-label="search input example" onChange={this.onInputChange} value={inputValue}/>
+          <Button variant={ButtonVariant.tertiary} aria-label="search button for search input">
+            <SearchIcon />
+          </Button>
+        </InputGroup>
+      </DataToolbarItem>
+      <DataToolbarItem spacers={[{spacerSize: 'none'}]}>
```
mattnolting:
```suggestion
      <DataToolbarItem>
```

mattnolting: In this case, the filters should be a nested `DataToolbarGroup`, as `__group`s are nestable.

```
<DataToolbarToggleGroup>
  <DataToolbarItem>
    <InputGroup />
  </DataToolbarItem>

  <DataToolbarGroup variant="filter-group">
    <DataToolbarItem>
      <Select />
    </DataToolbarItem>
    <DataToolbarItem>
      <Select />
    </DataToolbarItem>
  </DataToolbarGroup>

  <DataToolbarItem>
    <Stuff />
  </DataToolbarItem>

 </DataToolbarToggleGroup>
```

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItem.tsx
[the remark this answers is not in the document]
mattnolting: `clear`, `shared-item` and `chip-group` are dynamically added to toolbar. 

- `clear`: when multiple filters have been selected
- `shared-item`: this will probably go as React doesn't need it
- `chip-group`: automatically added when a filter is selected

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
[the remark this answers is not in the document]
karelhala: This is really unaesthetic because imagine that you are looking at table with this on mobile, click on the expand button, realise that landscape view might be better for this and once you flip the phone this expanded menu is lost and user has to click on it once again.

If the layout is not rendered correctly on window resize I think that is more of a issue in PF core and we shouldn't try to fix it here.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
nicolethoen: Both spacing of Items and Groups can use the DataToolbarSpacer, it's built to be smart enough to build both -space-items and -spacer, but you're right I was missing the -spacer on Groups, so I'll add a property to groups to specify if you are adding the spacer to the group or to the items in the group.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
nicolethoen: this is needed if the filters in this example will look like they are in a filter group in the unexpanded toolbar, while still taking up 100% width in the expandable content... in the core example, the html is structured differently in the content vs. the expandable content. Unfortunately, if the expandable content is going to be generated dynamically, then they cannot be different... (or i need an explicit rule about how they are different every time)

nicolethoen: they are nestable, but then they will not take up 100% width in the expandable content like they do in the core example... they will still be grouped like a filter-group in the expandable content...

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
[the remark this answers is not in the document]
dlabaj: @mcarrano Any thoughts on the behavior with the toggle?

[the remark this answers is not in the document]
mcarrano: This is a tough one to answer because it's not clear what the best approach is. I agree that just collapsing the drawer on resize it not great. Maybe if the drawer is open and a breakpoint is crossed, the content should not reconfigure itself (i.e. suppress the behavior in this case). Is that even possible? @mattnolting any thoughts here?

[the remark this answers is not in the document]
nicolethoen: @mcarrano - matt nolting is on PTO until Tuesday.

regarding that being possible, yes I believe it is, but we'd have to use the breakpoint passed in by the user and look up the global size of that breakpoint and make it reconfigure itself based on the global size (hopefully no one will have over written it). I may not have time to figure that out today - but I can try. Would that be an acceptable approach @dlabaj?

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarToggleGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+  static defaultProps = {
+    breakpointMods: [] as DataToolbarBreakpointMod[],
+    spacers: [] as DataToolbarSpacer[],
+  };
+
+  render() {
+    const { toggleIcon, breakpoint, variant, breakpointMods, spacers, className, children, ...props } = this.props;
+
+    return (
+      <DataToolbarContext.Consumer>
```
mturley: I forgot to mention this before, and you can take this or leave it, but there is a less verbose way to consume Context if you only need type of Context in your component.

Instead of rendering `<DataToolbarContext.Consumer>` with a render prop child, you could change this to a regular render method that pulls these values out from `this.context` like `const { isExpanded, toggleIsExpanded, expandableContentRef, expandableContentId } = this.context`, as long as you add `static contextType = DataToolbarContext;` up there next to your defaultProps.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarToggleGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+                  variant="plain"
+                  onClick={toggleIsExpanded}
+                  {...isExpanded && { 'aria-expanded': true }}
+                  // TODO aria-haspopup when isExpanded = true && viewport is smaller than lg global breakpoint
+                  aria-controls={expandableContentId}
+                >
+                  {toggleIcon}
+                </Button>
+              </div>
+              { isExpanded ? ReactDOM.createPortal(children, expandableContentRef.current): children }
```
mturley: I'm so glad this worked! For a while there I thought there might not be a declarative way to achieve this... but context, refs and portals = magic ✨

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
```diff
@@ -0,0 +1,793 @@
...
+            <SelectOption
+              isDisabled={option.disabled}
+              key={index}
+              value={option.value}
+            />
+          ))}
+        </Select>
+     </DataToolbarItem>
+     <DataToolbarItem>
+        <Select
```
mturley: There's some inconsistent indentation/spacing here and elsewhere in this file, but I don't think we need to hold up the PR for it. I'll address cases like this when I rebase https://github.com/patternfly/patternfly-react/pull/2810 after the release, but I'm not sure if Prettier will run on the JSX in our Markdown files, so that's something else I want to look into.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarToggleGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+                getModifier(styles, `reveal-on-${breakpoint}`),
+                className)}
+              {...props}
+            >
+              <div className={css(styles.dataToolbarToggle)}>
+                <Button
+                  variant="plain"
+                  onClick={toggleIsExpanded}
+                  {...isExpanded && { 'aria-expanded': true }}
+                  // TODO aria-haspopup when isExpanded = true && viewport is smaller than lg global breakpoint
```
mturley: Is this something we need to implement with this PR, and if not can we open an issue to track it?

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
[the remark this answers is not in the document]
mattnolting: @mcarrano I think, when using a toggle group, the content should reconfigure at the specified breakpoint if expanded content is expanded.

nicolethoen:
> @nicolethoen Again, great work! Can we add an example of `.pf-m-space-items` on a group to the **Data toolbar spacers** example? Perhaps the last two action buttons?
> 
> @nicolethoen can we add a horizontal divider to the **Data toolbar group stacked** example?
> 

Yes and Yes!

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarToggleGroup.tsx
nicolethoen: good thought

mcarrano:
@nicolethoen This is looking great. Agree with adding the horizontal divider to the stacked example. The only other thing I found is that for the Consumer managed toggle group example, I can get into a situation where the drawer remains open but the toggle disappears by opening the drawer and then stretching the window. See below.

Screen Shot 2019-09-10 at 11 24 30 AM

If this is not easy to fix. I'd be OK to approve this and come back to address this as a bug in the next release.

karelhala:
@mcarrano that's exactly what is discussed in https://github.com/patternfly/patternfly-react/pull/2861#discussion_r321627623

Consumers should be rensponsible when the pane is hidden and shown in this example. So if you open this toggle change the screen size and want to close the pane consumer is responsible for that. It's actually expected bahavior.

nicolethoen:
> @nicolethoen This is looking great. Agree with adding the horizontal divider to the stacked example. The only other thing I found is that for the Consumer managed toggle group example, I can get into a situation where the drawer remains open but the toggle disappears by opening the drawer and then stretching the window. See below.
> 
> Screen Shot 2019-09-10 at 11 24 30 AM
> 
> If this is not easy to fix. I'd be OK to approve this and come back to address this as a bug in the next release.

I noted this issue in the documentation above the example.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarToggleGroup.tsx
dlabaj: Let's open an issue for this.

nicolethoen: Issue created

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
```diff
@@ -0,0 +1,800 @@
...
+    const items = <React.Fragment>
+          <DataToolbarItem spacers={firstSpacers}><Button variant="secondary">Action</Button></DataToolbarItem>
+          <DataToolbarItem spacers={secondSpacers}><Button variant="secondary">Action</Button></DataToolbarItem>
+          <DataToolbarItem spacers={thirdSpacers}><Button variant="secondary">Action</Button></DataToolbarItem>
+          <DataToolbarItem spacers={fourthSpacers}><Button variant="secondary">Action</Button></DataToolbarItem>
+          <DataToolbarItem variant="separator"></DataToolbarItem>
+          <DataToolbarItem spacers={fifthSpacers}><Button variant="secondary">Action</Button></DataToolbarItem>
+          <DataToolbarItem><Button variant="primary">Action</Button></DataToolbarItem>
+          <DataToolbarItem variant="separator"></DataToolbarItem>
+          <DataToolbarGroup itemSpacers={[{spacerSize: 'lg'}]} spacers={[{spacerSize: 'lg'}]}>
```
mattnolting: Can you remove ` spacers={[{spacerSize: 'lg'}]}`

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md
```diff
@@ -0,0 +1,800 @@
...
+            <DataToolbarItem><Button variant="secondary">Action</Button></DataToolbarItem>
+ `          <DataToolbarItem><Button variant="secondary">Action</Button></DataToolbarItem>
```
mattnolting: looks like you have a floating tick mark ^^

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { css } from '@patternfly/react-styles';
+import { DataToolbarExpandableContent } from './DataToolbarExpandableContent';
+import { DataToolbarContext } from './DataToolbarUtils';
+
+export interface DataToolbarProps extends React.HTMLProps<HTMLDivElement> {
+  /** Classes applied to root element of the Data toolbar */
+  className?: string;
+  /** Content to be rendered as rows in the Data toolbar */
+  children?: React.ReactNode;
+  /** Flag indicating if a Data toolbar toggle group's expandable content is expanded */
```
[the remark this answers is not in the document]
mturley: If the consumer is managing the expansion, they would already know, right? If the consumer is not managing the expansion, why do they need to be able to find out if it's currently expanded?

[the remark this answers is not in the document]
nicolethoen: There is some documentation just before the toggle group examples that i believe explains this? should I add more info to that documentation?

PR #2862, opened 2019-09-06
refactor(Modal): Simplify disabling of FocusTrap

SpyTec:
**What**: This removes complexity and uses FocusTrapReact's `active` prop instead of switching to Bullseye component

**Additional issues**: #2966

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Modal/ModalContent.tsx
+++ b/packages/patternfly-4/react-core/src/components/Modal/ModalContent.tsx
@@ -15,7 +15,6 @@ import { ModalBoxHeader } from './ModalBoxHeader';
 import { ModalBoxCloseButton } from './ModalBoxCloseButton';
 import { ModalBox } from './ModalBox';
 import { ModalBoxFooter } from './ModalBoxFooter';
-import { Bullseye } from '../../layouts/Bullseye';
 
 export interface ModalContentProps {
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -102,16 +101,9 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
...
-        <Bullseye>
-          {modalBox}
-        </Bullseye>
-      ) 
-      : (
-        <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }} className={css(styles.bullseye)}>
-          {modalBox}
-        </FocusTrap>
-      )}
+      <FocusTrap active={!disableFocusTrap} focusTrapOptions={{ clickOutsideDeactivates: true }} className={css(styles.bullseye)}>
```
redallen: Neat [`active` property!](https://github.com/davidtheclark/focus-trap-react#active)

SpyTec:
@tlabaj there's no corresponding issue. I was looking at a different component that uses FocusTrap and just wanted to clean the code up for Modal

tlabaj:
Looks great @SpyTec!  We do require a corresponding issue to be linked before we merge a PR.  I would be happy to open one for you if you would like.

PR #2863, opened 2019-09-06
fix(Wizard): adds default function to onClose prop

### LINKED ISSUE #2831: PF4: Wizard - missing onClose crashes the wizard
The `onClose` prop is optional but the wizard crashes when it is missing.

We can either:

- keep it optional. In that case multiple null checks have to be added to the wizard.
- change it to required which would be a breaking change

SpyTec:
**What**: This adds a default function to `onClose` so that it no longer crashes the component if left out

**Additional issues**: fixes #2831

cc @suomiy

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
@@ -122,7 +122,7 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     width: null as string,
     height: null as string,
     footer: null as React.ReactNode,
-    onClose: () => undefined as () => void
+    onClose: () => undefined as any
   };
   private container: HTMLDivElement;
   private titleId: string;
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
@@ -122,7 +122,7 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     width: null as string,
     height: null as string,
     footer: null as React.ReactNode,
-    onClose: () => undefined
+    onClose: () => undefined as () => void
   };
   private container: HTMLDivElement;
   private titleId: string;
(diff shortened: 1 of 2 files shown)


suomiy:
+1, but snapshots need to be updated

SpyTec:
Right, thank you! - Snapshots updated

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -121,7 +121,8 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     onGoToStep: null as WizardStepFunctionType,
     width: null as string,
     height: null as string,
-    footer: null as React.ReactNode
+    footer: null as React.ReactNode,
+    onClose: () => undefined as () => void
```
redallen: Nitpick: `() => undefined as any` or `(): any => undefined` to be consistent with the rest of our code.

packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -121,7 +121,8 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     onGoToStep: null as WizardStepFunctionType,
     width: null as string,
     height: null as string,
-    footer: null as React.ReactNode
+    footer: null as React.ReactNode,
+    onClose: () => undefined as any
```
tlabaj: Can we do this instead for consistency?
```suggestion
    onClose: (): any  => undefined 
```

SpyTec: Right now there's 11 results for `): any => undefined` and 22 for `) => undefined as any`. Are we preferring the former?

dlabaj: We should fix these consistency issues and decide on which format we would like to use.  Could add some tslint rules for things when possible.  I would recommend leaving this for now and fixing some of these consistency issues in a separate issue.

PR #2899, opened 2019-09-10
Fix live edit caret color so it is visible

TheRealJon:
**What**:

Live editor caret was invisible since it was the same color as the background. Changed to white so that it is visible against the black background.

screencast-localhost-8000-2019 09 10-12-01-42


TheRealJon:
@redallen @jschuler Who wants to press the big green button?

redallen:
The repo is frozen now, but we'll click it real soon :)

Issue #1448, opened 2019-02-26
PF4 Tabs: lazy loading support


dlabrecq:
As a user, I want tabs to be lazily loaded.

In the Cost Management UI, we have multiple tabs. Each tab child generates multiple requests (3 each) required to populate cards containing charts, progress bars etc.

While implementing the PF4 tabs component, I noticed that all tabs are rendered at the same time. This means all 9 requests are generated immediately, when the page is loaded, regardless if other tabs are active or not.

Unfortunately, none of my children components are updated when switching tabs in this scenario. We can step thru the tab code, but our child's componentDidUpdate method is never called because nothing changed. 

As a side effect, the charts and progress bars in other tabs, do not render as expected when the tab is clicked. Only when a change is made; for example, the window resizes, the child components update and cards (i.e., charts, progress bars, etc.) render properly.

As a workaround, I've implemented code to return null instead of providing a tab child component. This forces the tabs to lazy load when active, which is what I want. Instead of generating 9 request, we only see 3 for the current tab. Requests are cached, so this happens only once, as needed.

Ideally, it would be great if the tabs component did this for us -- we could omit our custom code to handle this. For example, the tabs component might omit rendering children instead of just hiding them? We could even use a flag, like isLazyLoad, to enable this as a feature.

Below is an example of how we return null for lazy loading.

![screen shot 2019-02-26 at 9 14 09 am](https://user-images.githubusercontent.com/17481322/53420277-811e1780-39a9-11e9-9cce-ffc94121f32b.png)

dlabrecq:
Loading all tabs at once is still an issue.

rachael-phillips:
We're ready to close this one

PR #2902, opened 2019-09-10
feat(Tab): Added two new props: mountOnEnter and unmountOnExit

### LINKED ISSUE #2609: Feature: Tab component misses mountOnEnter, unmountOnExit properties
In PF3 we had something like the following:

```
            <TabContent>
              <TabPane eventKey="info" mountOnEnter={true} unmountOnExit={true}>
                <AppInfo
                  app={this.state.app}
                  namespace={this.props.match.params.namespace}
                  onRefresh={this.doRefresh}
                  activeTab={this.activeTab}
                  onSelectTab={this.tabSelectHandler}
                  health={this.state.health}
                />
              </TabPane>
              <TabPane eventKey="traffic" mountOnEnter={true} unmountOnExit={true}>
              </TabPane>
            <TabContent>
```

In patternfly 4, we want to implement Tabs also using the `mountOnEnter` and `unmountOnExit`

```
      <Tab title="Inbound metrics" eventKey={2}>
        <IstioMetricsContainer
          namespace={this.props.match.params.namespace}
          object={this.props.match.params.app}
          objectType={MetricsObjectTypes.APP}
          direction={'inbound'}
        />
      </Tab>
```

Is there the possibility to implement those properties for PF4 for Tab component?

rebeccaalpert:
When mountOnEnter is provided as a prop, TabContents will mount on the DOM as you click the corresponding tab. When unmountOnExit is provided, TabContents will unmount from the DOM when no longer visible.

Fixes #2609 and #1448.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/patternfly-4/react-core/src/components/Tabs/Tabs.tsx
@@ -6,8 +6,8 @@ import { Omit } from '../../helpers/typeUtils';
 import { AngleLeftIcon, AngleRightIcon } from '@patternfly/react-icons';
 import { getUniqueId, isElementInView, sideElementIsOutOfView } from '../../helpers/util';
 import { SIDE } from '../../helpers/constants';
-import { TabContent } from './TabContent';
 import { Tab } from './Tab';
+import { TabContent } from './TabContent';
 import { InjectedOuiaProps, withOuiaContext } from '../withOuia';
 
 export enum TabsVariant {
@@ -38,6 +38,10 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   variant?: 'div' | 'nav';
   /** provides an accessible label for the Tabs. Labels should be unique for each set of Tabs that are present on a page. When variant is set to nav, this prop should be defined to differentiate the Tabs from other navigation regions on the page. */
   'aria-label'?: string;
+  /** waits until the first "enter" transition to mount tab children (add them to the DOM) */
+  mountOnEnter?: boolean;
+  /** unmounts tab children (removes them from the DOM) when they are no longer visible */
+  unmountOnExit?: boolean;
 }
 
 export interface TabsState {
@@ -45,6 +49,7 @@ export interface TabsState {
   showRightScrollButton: boolean;
   highlightLeftScrollButton: boolean;
   highlightRightScrollButton: boolean;
+  shownKeys: (string|number)[];
 }
 
 class Tabs extends React.Component<TabsProps & InjectedOuiaProps, TabsState> {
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -257,10 +275,13 @@ class Tabs extends React.Component<TabsProps & InjectedOuiaProps, TabsState> {
             <AngleRightIcon />
           </button>
         </Component>
-        {React.Children.map(children, (child: any, index) =>
-          !child.props.children ? null : (
-            <TabContent key={index} activeKey={activeKey} child={child} id={child.props.id || uniqueId} />
-          )
+        {React.Children.map(children, (child: any, index) => {
+          if (!child.props.children || (unmountOnExit && child.props.eventKey !== activeKey) || (mountOnEnter && shownKeys.indexOf(child.props.eventKey) === -1)) {
+            return null;
```
dlabaj: Should this be undefined instead?

rebeccaalpert:
Added integration test and changed from null to undefined. Also fixed minor bug in Tabs where you couldn't actually apply an id to the main Tabs div.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -257,10 +276,13 @@ class Tabs extends React.Component<TabsProps & InjectedOuiaProps, TabsState> {
             <AngleRightIcon />
           </button>
         </Component>
-        {React.Children.map(children, (child: any, index) =>
-          !child.props.children ? null : (
-            <TabContent key={index} activeKey={activeKey} child={child} id={child.props.id || uniqueId} />
-          )
+        {React.Children.map(children, (child: any, index) => {
+          if (!child.props.children || (unmountOnExit && child.props.eventKey !== activeKey) || (mountOnEnter && shownKeys.indexOf(child.props.eventKey) === -1)) {
+            return undefined;
```
dlabrecq: When you don't want to render a component, you typically return null. In this case; however, perhaps you can return only when there is tab content? That is, skip returning null or undefined.

rebeccaalpert: Yeah, @dlabaj asked me to do this. Do you have a preference, Don? Thanks.

dlabrecq: I don't think you need to return anything, but the items you intend to render here. Easy enough to try and see how it works...

dlabaj: @rebeccaalpert Since react docs say to return null we should do that instead.  In other frameworks I was use to it being undefined.

rebeccaalpert: All set.

Issue #1882, opened 2019-04-29
Pagination - add prop for page size menu toggle label


jgiardino:
The consumer should be able to specify the text string for the toggle that displays the page size menu.

We should also include a default string that clearly describes the button. 

NOTE: these updates are expected to be completed after or as part of the update to refactor the Pagination component to use the OptionsMenu component. As part of this update, `button.pf-c-options-menu__toggle` would not be expected to have the attribute `aria-labelledby`.

**Updates for this issue:**
- Include a prop named `ariaLabelPageSize` that, when specified, provides the text string for a value for `aria-label` on `button.pf-c-options-menu__toggle-button`
- ~Define a default string "Items per page" to use when this prop is not specified by the user.~ This was completed in https://github.com/patternfly/patternfly-react/pull/2904

jgiardino:
@mcarrano @stacymcauliffe 

For the page size toggle, does "Page size" work? Or "Items per page"? 

When announced, the whole button gets announced as "[label] button collapsed".

I also need to create an issue for this in core, btw.

mcarrano:
"Items per page" IMO. Page size is a bit ambiguous.

jgiardino:
Thanks! That sounds good to me. I updated the comment description to use that as a text string but kept the prop name as "ariaLabelPageSize" for brevity.

dgutride:
can this one be closed @jgiardino ?

jgiardino:
I don't see these updates in the component, can we keep it open? 

Currently, we include an `aria-label` attribute on the toggle with a static string of "Select". The consumer should be able to provide their own localized string for this. 

Also, the default value for this text string should be "Items per page", which is a more specific label.

jessiehuff:
This seems like a fairly straightforward issue, just giving flexibility to change the `aria-label` of the pagination toggle (which would likely help with localization as well).

From what I understand, we just need to add a prop named ariaLabelPageSize that, when specified, provides the text string for a value for aria-label on button.pf-c-options-menu__toggle-button

tlabaj:
The prop is there it is not documented.

PR #2904, opened 2019-09-11
fix(Pagination): Styling discrepancy with HTML version

### LINKED ISSUE #2903: Styling of Pagination component is incorrect
The styling of the "items per page" select menu is not correct. This should not use standard Select list styling as shown below.

Core implementation is correct. See

SpyTec:
**What**: Fix differences with upstream. This makes Pagination be identical to HTML variant

**Additional issues**: Fixes #2903 

~~Note, work in progress. Trying to figure out the best approach for fixing the font size and padding of the dropdown button as well as the font color.~~ Fixed

image

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/index.ts
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/index.ts
@@ -1,5 +1,6 @@
 export * from './Dropdown';
 export * from './DropdownMenu';
+export * from './DropdownWithContext';
 export * from './dropdownConstants';
 export * from './DropdownGroup';
 export * from './DropdownItem';
--- a/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
@@ -174,7 +174,7 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
               onChange={(event) => this.onChange(event, lastPage)}
             />
             <span aria-hidden="true">
-              of {pluralize(lastPage, pagesTitle)}
+              of {pagesTitle ? pluralize(lastPage, pagesTitle) : lastPage}
             </span>
           </div>
         )}
--- a/packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/OptionsToggle.tsx
@@ -11,6 +11,8 @@ export interface OptionsToggleProps extends React.HTMLProps<HTMLDivElement> {
   itemsTitle?: string;
   /** The text to be displayed on the Options Toggle */
   optionsToggle?: string;
+  /** The Title of the Pagination Options Menu */
+  itemsPerPageTitle?: string;
   /** The first index of the items being paginated */
   firstIndex?: number;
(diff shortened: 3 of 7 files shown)


SpyTec:
Trying to change DropdownToggle in OptionsToggle to OptionsMenuToggleWithText doesn't help much. HTML variant has `button.pf-c-options-menu__toggle-button`, but I'm unable to see how we can get this without the `.pf-c-dropdown__toggle` being added

SpyTec:
I've managed to sync everything with HTML version. One note to keep in mind, to get around applying `.pf-c-dropdown__toggle` class to the dropdown I added a `DropdownContext.Provider` with toggleClass value of a space which is shown in the outputted HTML. 

image

mcoker:
Looks like `.pf-c-options-menu` is nested in another `.pf-c-options-menu`, and in the top level `.pf-c-options-menu`, there is a `#pagination-options-menu-top-label` element with "Items per 
 page:" that we don't have in core. 

Screen Shot 2019-09-17 at 1 58 29 PM

"Items per page" is an `aria-label` on the toggle button in core, and the `.pf-c-options-menu__menu` element's `aria-labelledby` is set to the ID of the toggle button.

Also looks like `.pf-c-options-menu__menu-item` is listed twice on the menu items

Screen Shot 2019-09-17 at 2 00 44 PM

And for the `.pf-c-options-menu__menu-item-icon`, can we just apply that class directly to the `<svg>` and remove the `<i>`?

Screen Shot 2019-09-17 at 4 11 23 PM

SpyTec:
Fixed all but the `.pf-c-options-menu__menu-item-icon`. Putting on the SVG itself didn't work. Personally I think it would be better to keep it to an `<i>` as it is in core so that the elements don't change

mcoker:
@SpyTec thanks! I opened https://github.com/patternfly/patternfly-next/issues/2297 re: the SVG/icon class.

This looks good to me, although I'm still seeing this element in the react component that isn't in the core examples:

Screen Shot 2019-09-23 at 4 20 36 PM

And in the react component, `.pf-c-options-menu__toggle-button` has `aria-label="Select"` and in core, we have `aria-label="Items per page"`.

@jessiehuff or @jgiardino can you confirm if we need to make updates there?

jgiardino:
Replying to @mcoker's [comment above](https://github.com/patternfly/patternfly-react/pull/2904#issuecomment-534291671), there is a related open issue to the aria label applied to the toggle: https://github.com/patternfly/patternfly-react/issues/1882

However, when I compare the html in core to the html in react, they are different, and the solution we want is dependent on which html we're using. So that issue might need to be updated depending on that. 

The main difference I'm seeing is that in core, the text that displays in the toggle "x-y of z items" is not part of the button. But in react, it is.

jgiardino:
Doh! I was looking at the current version of the react component, and not the fixes included in this PR. 😜 

This hidden span was an artifact of the old way we were labelling the toggle. It can be removed.
`<span id="pagination-options-menu-top-label" hidden="">Items per page:</span>`

That issue I mention above is still relevant, but not a blocker for this PR.

SpyTec:
Fixed the last bit with `Select` aria-label so it's now `Items per page`. Also removed old labelling `<span id="pagination-options-menu-top-label" hidden="">Items per page:</span>` as suggested

Issue #2876, opened 2019-09-06
Hide default bulk selection from table for bulk select 


LHinson:
When we introduce the bulk select, we will want to ensure there is the ability to hide the default bulk select seen in the table header today. This should be configurable dependent on whether or not you want to use you bulk select component or not. 

related issue #2364

PR #2926, opened 2019-09-13
feat(Table): add flag to hide select all

kmcfaul:
**What**:

**Additional issues**: #2876


packages/patternfly-4/react-table/src/components/Table/utils/headerUtils.tsx
```diff
@@ -93,12 +93,12 @@ const mapHeader = (column: ICell, extra: any, key: number, ...props: any) => {
...
  */
-const selectableTransforms = ({ onSelect }: { onSelect: OnSelect}) => [
+const selectableTransforms = ({ onSelect, canSelectAll }: { onSelect: OnSelect, canSelectAll: boolean}) => [
   ...(onSelect
     ? [
         {
           title: '',
-          transforms: [selectable],
+          transforms: (canSelectAll && [selectable]) || null,
           cellTransforms: [selectable]
```
priley86: this seems correct - if we have a consumer that passes `onSelect` callback with `canSelectAll` set `false`, they'd receive the selectable cell transform (in the rows), but they'd not receive the "select all" header transform. 

I feel like another example showing that use case in Table.md might be warranted if you have time, but this appears correct to me!

kmcfaul: Yeah I'll throw in a checkbox to toggle it on and off in the UI of the docs, it will make it easier to see without needing to edit the code section.

PR #2932, opened 2019-09-16
feat(tables): support height auto modifier in tr

### LINKED ISSUE #2930: Height: auto; prop for table component <tr> elements
We need to add a prop so that a `<tr>` element can have class `.pf-m-height-auto`, which assigns an auto height. This is to address a bug in the table where the child(ren) of a cell that are set to `height: 100%` will end up only being `1px` tall because we have CSS that defines `<tr>` elements as `1px`.

Core PR - https://github.com/patternfly/patternfly-next/pull/2254/

boaz0:
**What**:

closes #2930 

//cc @mcoker @dlabrecq


mcoker:
How do you use `cellHeightAuto`? And can we document that in the table component?

priley86:
yea, PF upgrades are typically done in separate PRs... but this looks fine to me 👍 

cc'ing: @redallen

boaz0:
Thank you @tlabaj & @dlabrecq

Issue #2483, opened 2019-07-09
Select has no max height


SpyTec:
Select has no `max-height`, meaning that for long lists it will expand every entry. A height limit should be imposed so it can become scrollable

tlabaj:
@mcoker can you please take a look at this one too

mcoker:
@mcarrano can you weigh in on this one?

mcarrano:
I don't feel strongly about this one @mcoker but I do see @SpyTec 's point. I can see arguments both ways. Maybe we add the ability to set max-height as an option? What do you think?

SpyTec:
Though if max-height was an optional thing, would #2485 be possible? Given hundreds of entries, would there then be a different component more suitable?

rachael-phillips:
@mcoker and @mcarrano what are your thoughts to @SpyTec 's comment above?

mcoker:
I agree setting a `max-height` as an optional variation would be useful, though I don't understand the technical implications of a virtual scroller, so not sure why it wouldn't (maybe?) be possible if the choice to limit the height was a prop?

I'm also not sure if another component would be more suitable. I suppose it depends on the use case - maybe a selectable table or datalist with the option for filtering and/or pagination?

SpyTec:
> I agree setting a `max-height` as an optional variation would be useful, though I don't understand the technical implications of a virtual scroller, so not sure why it wouldn't (maybe?) be possible if the choice to limit the height was a prop?

I've never implemented a virtual scroller myself. Though from my POV it would seem like a lot of work to implement if it will only be used if max-height is active.

> I'm also not sure if another component would be more suitable. I suppose it depends on the use case - maybe a selectable table or datalist with the option for filtering and/or pagination?

All of these require you to have visible data that isn't hidden behind a dropdown, and it would need to occupy space and not be floating like dropdowns do.

tlabaj:
cc @priley86  Can you weigh in here please.

mceledonia:
Chiming in here, I don't think it would hurt to have as an option, but I'd concur that if you had a dropdown with enough items for this to be an issue another design approach would be worth considering. Maybe that's something we suggest in design documentation. How does it currently behave when the dropdown list extends beyond the browser height in say a mobile size viewport?

mcoker:
@mceledonia it overflows the viewport and you have to scroll the window

[...]

PR #2946, opened 2019-09-17
feat(Select): add max height property, update tests

kmcfaul:
**What**: adds a maximum height property to select menu

**Refer to issue**: #2483

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
@@ -137,7 +137,7 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
         {({ variant }) => (
           <React.Fragment>
             {variant !== SelectVariant.checkbox && (
-              <ul {...props} className={css(styles.selectMenu, className)} role="listbox" {...maxHeight && {style:({ maxHeight, overflow: 'auto' })}}>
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -133,21 +137,21 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
         {({ variant }) => (
           <React.Fragment>
             {variant !== SelectVariant.checkbox && (
-              <ul {...props} className={css(styles.selectMenu, className)} role="listbox">
+              <ul {...props} className={css(styles.selectMenu, className)} role="listbox" {...maxHeight && {style:({ maxHeight, overflow: 'auto' })}}>
```
redallen: Not a giant fan of using a style tag as our implementation. Can we add a custom React class so the consumer can override it?

redallen: Ah, I realize now there's not much to override other than the `overflow` type. Let's just move the `{...props}` to the end in case a `style` tag is passed in `props`.

kmcfaul: I'm not sure how we could assign a class and still have a variable the user could customize.

Issue #2908, opened 2019-09-11
Table spike: Column reordering


LHinson:
Can we make [Column re-ordering on table](https://github.com/patternfly/patternfly-react/issues/2875) available? We need to ensure design understand the default behavior to make sure they approve 

Clarifications: 
- This is only for PatternFly table (not extensions)
- This is an investigation 
- Follow up with Dana, Jan, Leslie, Titani to clarify effort and agree upon next steps

seanforyou23:
@LHinson here's the takeaways from the investigation

Column reordering is totally possible, we'll be looking at rolling our own implementation of the drag and drop functionality as well as the column reordering, as [the one from reactabular](https://reactabular.js.org/#/drag-and-drop) currently has issues and is a bit overengineered it seems.

Design states needed

- Styling for the source column when a user has dragged it (in the middle of a drag operation - "item marked for drag")
- Styling for target insertion point (when a user has grabbed a column and is hovering over a potential drop point)
- Styling for a grab handle (consider how this works when a column is both sortable and reorderable)
- Styling for target insertion point in disallowed drag state (when a user has grabbed a column and is hovering over a drop point that is not allowed)
- Styling for the source node in disallowed drag state (when a user has grabbed a column and is hovering over a drop point that is not allowed)

Open design questions

- If using this feature from a touch device or driving from a keyboard, are the visual indicators the same?

Open accessibility questions

- If design calls for dedicated "grabbable" elements, how should these be labelled?
- Aria attributes for drag and drop are deprecated in 1.1 ([aria-dropeffect](https://www.w3.org/TR/wai-aria-1.1/#aria-dropeffect) and [aria-grabbed](https://www.w3.org/TR/wai-aria-1.1/#aria-grabbed)), is there anything special we need to do in this area to ensure the functions are made accessible to SRU?
- How to determine keyboard support? We could strive to match the keypress combinations that facilitate drag and drop in a native screen reader (for example `ctrl+opt+<` to mark item as drag and `ctrl+opt+>` to drop item). Is there another strategy to consider?

PR #2949, opened 2019-09-17
[WIP] Table column reorder

seanforyou23:
**What**: This PR is only here to give a very early look at what reorderable table columns could look/function like. It is far from complete both implementation and UX-wise, so please hold off on giving any comprehensive review. This PR is mainly just a way to share the demo with a few folks who will be gathering requirements for how this feature should _actually_ function and look.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/2908

You can currently drag not only headings, but also row cells. Trying to drag anything other than a header onto another header will result in error - this is known, please ignore.

A limitation we should discuss is that currently, we allow users to specify `colSpan` for row data. This will cause the drag/drop logic to become much more elaborate than it is currently. We could make this work, but it may be better to simply require users to supply an empty object (or something of the sort) to represent empty row data.

Another talking point from UX standpoint is regarding the various drag states. How to best indicate where a column will be dropped - currently you drop a column ontop of another column, but now that we've gotten this far it doesn't really make sense. Might be better to have the visual indicator point to in-between columns instead of on top. This would align nicely with drop and drop in other components (treeview for example), where dropping on top of another node produces a child instead of a drag before/after operation.

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.md
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.md
@@ -33,84 +33,6 @@ import {
 
 import DemoSortableTable from './demo/DemoSortableTable';
 
-## Reorderable Column Table
-
-```js
-import React from 'react';
-import {
-  Table,
-  TableHeader,
-  TableBody,
-  sortable,
-  SortByDirection,
-  headerCol,
-  TableVariant,
-  expandable,
-  cellWidth,
-  textCenter,
-} from '@patternfly/react-table';
-
-
-class ColReorderTable extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      columns: [
-        { title: 'Repositories' },
-        { title: 'Branches' },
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -108,6 +108,110 @@ class SimpleTable extends React.Component {
...
+  render() {
+    const { columns, rows } = this.state;
+
+    const swapArrPosition = (from, to, arr) => {
+      let newArr = arr.slice();
+      [newArr[from], newArr[to]] = [newArr[to], newArr[from]];
+      return newArr;
+    }
+
+    const applyColumnReorder = (from, to) => {
```
redallen: I know it's a pain to maintain, but I'd like us to take care of this problem on the PatternFly side, even if a wonky default implementation that can be overridden. We'll have to make a copy of the data or keep some state about the order of columns and iterate over them according to that.

Issue #2737, opened 2019-08-20
a11y issues in documentation/react/table 


lareplog:
for https://github.com/patternfly/patternfly-org/issues/1319 one type of a11y issue needs to be fixed:

1. scope attribute should be used correctly

goes along with this issue https://github.com/patternfly/patternfly-next/issues/2165

rachael-phillips:
Hi @lareplog what was the rationale for tagging this as P1?

lareplog:
@rachael-phillips the original issue was tagged as p1, so i assumed it and the next issue i created would also be considered p1

PR #2953, opened 2019-09-18
fix(table): remove scope=col from td elements

mattnolting:
fixes: #2737


jgiardino:
Thanks for sharing the deque article, @jessiehuff! That was a nice find.

> Were they meant to be a `<th>` in these instances? 

Yes, this example was included to illustrate how to define a row header in the Table component. Row headers can be as beneficial as column headers when navigating a table using a screen reader.

> I'm fine with accepting these changes for now and opening another issue, but it seems like that is something that needs further investigation/discussion.

I'm also good with accepting this PR as-is. I assume adding `scope="row"` for these cases would be out of scope for this PR. It would be worth adding an issue to track it, though.

PR #2957, opened 2019-09-19
feat(emptyState): Support primary and icon as div elements

### LINKED ISSUE #2929: Empty state icon and primary button updates
`.pf-c-empty-state__icon` can now be a `div` that allows for an icon, in case that icon is a component, like the `Spinner` component.

Also added a new element `.pf-c-empty-state__primary` that can be used in lieu of `<Button variant="primary">` as the primary button in the empty state component.

Core PR - https://github.com/patternfly/patternfly-next/pull/2254/

boaz0:
**What**:

closes #2929 

I wanted to add the demos (loading state table and empty state table) but I couldn't because:
1. I don't have the spinner component in react-core
2. They're relayed on the height-auto modifier which isn't in master yet: https://github.com/patternfly/patternfly-react/pull/2932
3. I am waiting on patternfly core to be upgraded to `2.31.9` (please ignore the second commit).

### DIFF
--- a/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateIcon.tsx
+++ b/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateIcon.tsx
@@ -19,14 +19,25 @@ export interface IconProps extends Omit<React.HTMLProps<SVGElement>, 'size'> {
 export interface EmptyStateIconProps extends IconProps {
   /** Additional classes added to the EmptyState */
   className?: string;
-  /** Icon component to be rendered inside the EmptyState */
-  icon: string | React.FunctionComponent<IconProps>;
+  /** Icon component to be rendered inside the EmptyState on icon variant */
+  icon?: string | React.FunctionComponent<IconProps>
+  /** Component to be rendered inside the EmptyState on container variant */
+  component?:  React.FunctionComponent<any>;
+  /** Adds empty state icon variant styles  */
+  variant?: 'icon' | 'container';
 }
 
 export const EmptyStateIcon: React.FunctionComponent<EmptyStateIconProps> = ({
   className = '',
-  icon: IconComponent,
+  icon: IconComponent = null,
+  component: AnyComponent = null,
+  variant = 'icon',
   ...props
-}: EmptyStateIconProps) => (
-  <IconComponent className={css(styles.emptyStateIcon, className)} {...props} aria-hidden="true" />
-);
+}: EmptyStateIconProps) => {
+  const classNames = css(styles.emptyStateIcon, className);
+  return variant === 'icon' ? (
+    <IconComponent className={classNames} {...props} aria-hidden="true" />
+  ) : (
+    <div className={classNames}><AnyComponent /></div>
+  );
+};
--- a/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStatePrimary.tsx
+++ b/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStatePrimary.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/EmptyState/empty-state';
+
+export interface EmptyStatePrimaryProps
+  extends React.HTMLProps<HTMLDivElement> {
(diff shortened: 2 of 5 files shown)


tlabaj:
@boaz0 the patternfly core package has been updated.

boaz0:
@karelhala :scream: for some reason my vim-prettier got crazy.

packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateIcon.tsx
```diff
@@ -21,12 +21,21 @@ export interface EmptyStateIconProps extends IconProps {
...
   ...props
-}: EmptyStateIconProps) => (
-  <IconComponent className={css(styles.emptyStateIcon, className)} {...props} aria-hidden="true" />
-);
+}: EmptyStateIconProps) => {
+  const classNames = css(styles.emptyStateIcon, className);
+  return isWrapped ? (
+    <div className={classNames}>
+      <IconComponent {...props} aria-hidden="true" />
+    </div>
```
mcoker: Will this allow you to put a the `<Spinner>` component in the empty state as the "icon"?

packages/patternfly-4/react-core/src/components/EmptyState/EmptyStatePrimary.tsx
```diff
@@ -0,0 +1,21 @@
...
+  className?: string;
+  /** Content rendered inside the EmptyStatePrimary */
+  children: React.ReactNode;
+}
+
+export const EmptyStatePrimary: React.FunctionComponent<
+  EmptyStatePrimaryProps
+> = ({ children, className = '', ...props }: EmptyStatePrimaryProps) => {
+  return (
+    <div className={css(styles.emptyStatePrimary, className)} {...props}>
```
mcoker: Hmm, when I put `<EmptyStatePrimary className="custom-empty-state-prim-cls" id="empty-state-prim-id">
<Button variant="link">Link</Button></EmptyStatePrimary>` in one of the empty state component examples, I get `ReferenceError: EmptyStatePrimary is not defined`. Do you know what I'm doing wrong?

packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateIcon.tsx
boaz0: I just realized that it's a little complicated. Is `<Spinner>` imported from `react-icons` or is a separate component?

packages/patternfly-4/react-core/src/components/EmptyState/EmptyStatePrimary.tsx
boaz0: No, I think this is my mistake. I forgot to export it. I will fix that. Thank you!

packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateIcon.tsx
mcoker: It’s a separate component.

mcoker: It’s a separate component.

boaz0:
@mcoker ok I updated the PR, I also temporarily (just for the sake of this to review) added the loading state and the no-match state to the examples.

Also, I am assuming that the component that will be wrapped inside the `div` will get the same props: ('size', 'color' and 'title') as the icon component. Can I assume that or we're not sure how it's going to look like.

At the worst case, we may create a new component that will support both wrapping general components whether it's a general React component or an Patternfly Icon component

boaz0:
Thanks @karelhala I guess vim-prettier doesn't love me :sob: - I am going to turn it off. May it rest in peace :pray:

mcoker:
@boaz0 thanks! My expectation is that if `pf-c-empty-state__icon` is used as a container, you can put whatever you want in it and `color`, `size`, `title`, etc no longer apply since it's now a `<div>` and not an `<svg>`. I would expect it to work similarly to `pf-c-empty-state__primary`.

boaz0:
@mcoker thanks.
to make this backward compatible I set a new prop that is only used in the container variant.

We should figure out a better API (probably will introduce a breaking change) to support one simple prop for both icon and component under the empty state icon.

Issue #2888, opened 2019-09-09
PF4 DataToolbar:  Add chip groups to toolbar


tlabaj:
This is a follow up issue for the initial implementation of the experimental DataToolbar component.  This issue will t rack the work need to add the Chipgroup to the toolbar.

rachael-phillips:
Closing because this has been complete

Issue #1454, opened 2019-02-27
ChipGroup - accessibility refinements


jgiardino:
The following accessibility issues are present in the ChipGroup component:

- **Issue:** In the HTML that gets rendered, an `aria-label="Close"` attribute is defined on the "# More" and "Show Less" buttons. This label doesn't start with the text string that's visible, and it doesn't describe the action that's happening. In this case, since the button has visible text, `aria-label` shouldn't be used.
  - **Update:** remove `aria-label` attribute from the "# More" and "Show Less" buttons
- **Issue:** When the chip groups are expanded, there should only be one `<ul class=“pf-c-chip-group pf-m-toolbar”>` element that wraps all the first level list items (`<li><h*><ul></li>`), so that the number of chip groups are communicated to screen reader users. Currently, each chip group is a separate list, with one list item. 
  - **Update:**
    - Replace the following structure:
      ```
      <ul class="pf-c-chip-group pf-m-toolbar"
        <li>
          <h4> class="pf-c-chip-group__label">Category 1</h4>
          <ul class="pf-c-chip-group">...</ul>
        </li>
      </ul>
      <ul class="pf-c-chip-group pf-m-toolbar"
        <li>...</li>
      </ul>
      <ul class="pf-c-chip-group pf-m-toolbar"
        <li>...</li>
      </ul>
      ```
    - With the following structure:
      ```
      <ul class="pf-c-chip-group pf-m-toolbar"
        <li>
          <h4> class="pf-c-chip-group__label">Category 1</h4>
          <ul class="pf-c-chip-group">...</ul>
        </li>
        <li>...</li>
        <li>...</li>
      </ul>
      ```
- **Issue:** Heading level is set to `<h4>`, but heading level should be defined by the consumer relative to context in which this component is used. 
  - **Update:** Allow the consumer to set the heading level in the chip group.

jgiardino:
Closing this issue. Some updates requested are no longer relevant due to updates to core (e.g. specifying a heading level), and other updates have been addressed.

PR #2969, opened 2019-09-19
feat(DataToolbar): add Chip groups to DataToolbar

nicolethoen:
Addresses [#2888](https://github.com/patternfly/patternfly-react/issues/2888)

### DIFF
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
@@ -16,6 +16,8 @@ export interface DataToolbarProps extends React.HTMLProps<HTMLDivElement> {
   toggleIsExpanded?: () => void;
   /** optional callback for clearing all filters in the toolbar */
   clearAllFilters?: () => void;
+  /** Flag indicating that the Clear all filters button should be visible */
+  showClearFiltersButton?: boolean;
   /** Id of the Data toolbar */
   id: string;
 }
@@ -25,8 +27,6 @@ export interface DataToolbarState {
   isConsumerManagedToggleGroup: boolean;
   /** Flag indicating if the component managed state has expanded content or not */
   componentManagedIsExpanded: boolean;
-  /** Flag indicating that the Clear all filters button should be visible */
-  showClearFiltersButton: boolean;
 }
 
 export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarState> {
@@ -34,7 +34,8 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
   private chipGroupContentRef = React.createRef<HTMLDivElement>();
 
   static defaultProps = {
-    isExpanded: false
+    isExpanded: false,
+    showClearFiltersButton: false,
   };
 
   constructor(props: DataToolbarProps) {
@@ -43,7 +44,6 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
     this.state = {
       isConsumerManagedToggleGroup: props.isExpanded || !!props.toggleIsExpanded,
       componentManagedIsExpanded: false,
-      showClearFiltersButton: false,
     };
   }
 
@@ -57,12 +57,6 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
     this.setState(() => ({
       componentManagedIsExpanded: false
     }));
-  }
-
-  updateShowClearFiltersButton = (showClearFiltersButton: boolean) => {
-    if (showClearFiltersButton !== this.state.showClearFiltersButton) {
-      this.setState(() => ({ showClearFiltersButton }));
-    }
   }
 
   componentDidMount() {
@@ -82,8 +76,17 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
 
   render() {
 
-    const { className, children, isExpanded, toggleIsExpanded, id, clearAllFilters, ...props} = this.props;
-    const { isConsumerManagedToggleGroup, componentManagedIsExpanded, showClearFiltersButton } = this.state;
+    const {
+      className,
+      children,
+      isExpanded,
+      toggleIsExpanded,
+      id,
+      clearAllFilters,
+      showClearFiltersButton,
+      ...props
+    } = this.props;
+    const { isConsumerManagedToggleGroup, componentManagedIsExpanded } = this.state;
 
     const expandableContentId = `${id}-expandable-content`;
 
@@ -97,7 +100,6 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
               expandableContentRef: this.expandableContentRef,
               expandableContentId,
               chipGroupContentRef: this.chipGroupContentRef,
-              updateShowClearFiltersButton: this.updateShowClearFiltersButton,
             }
           }
         >
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
@@ -3,6 +3,8 @@ import styles from '@patternfly/react-styles/css/components/DataToolbar/data-too
 import { css } from '@patternfly/react-styles';
 import { DataToolbarExpandableContent } from './DataToolbarExpandableContent';
 import { DataToolbarContext } from './DataToolbarUtils';
+import { DataToolbarChipGroupContent } from './DataToolbarChipGroupContent';
+
 
 export interface DataToolbarProps extends React.HTMLProps<HTMLDivElement> {
   /** Classes applied to root element of the Data toolbar */
@@ -13,6 +15,8 @@ export interface DataToolbarProps extends React.HTMLProps<HTMLDivElement> {
   isExpanded?: boolean;
   /** A callback for setting the isExpanded flag */
   toggleIsExpanded?: () => void;
+  /** optional callback for clearing all filters in the toolbar */
+  clearAllFilters?: () => void;
   /** Id of the Data toolbar */
   id: string;
 }
@@ -26,6 +30,7 @@ export interface DataToolbarState {
 
 export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarState> {
   private expandableContentRef = React.createRef<HTMLDivElement>();
+  private chipGroupContentRef = React.createRef<HTMLDivElement>();
 
   static defaultProps = {
     isExpanded: false
@@ -36,7 +41,7 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
 
     this.state = {
       isConsumerManagedToggleGroup: props.isExpanded || !!props.toggleIsExpanded,
-      componentManagedIsExpanded: false
+      componentManagedIsExpanded: false,
     };
   }
 
@@ -69,7 +74,7 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
 
   render() {
 
-    const { className, children, isExpanded, toggleIsExpanded, id, ...props} = this.props;
+    const { className, children, isExpanded, toggleIsExpanded, id, clearAllFilters, ...props} = this.props;
     const { isConsumerManagedToggleGroup, componentManagedIsExpanded } = this.state;
 
     const expandableContentId = `${id}-expandable-content`;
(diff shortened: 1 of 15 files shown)


packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItemWithChipGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  componentDidMount() {
+    /* eslint-disable-next-line */
+    this.setState({ isMounted: true });
+  }
+
+  render() {
+    const { children, chips, deleteChip, label, ...props } = this.props;
+
+    return (
+      <DataToolbarContext.Consumer>
```
redallen: Why can't we make this a part of `DataToolbarItem`?

nicolethoen:
@redallen I suppose they could be combined, there were a couple reasons I chose to separate them.

1. An item is a very simple component that primarily manipulates spacing. 
2. There are multiple props that should be considered if a consumer wants to use an item with a chip group - such as a list of chips, and I didnt want to bog down a very simple item with those props.
3. The item with chip group uses the item component multiple times in the render method, so it seemed natural that it was a different sort of concept. 

are those valid enough reasons? I started trying to combine them and it seems like such over kill for the more commonly used and simple item to have all the logic and props from the chip group item, but it's certainly possible.

jgiardino:
Hey @nicolethoen, this is great! I played around with some of the controls, and noticed a couple of things that are a little odd. I'm copying @mcoker on this, because I'm not sure if they are core issues or not.

1. When I clear all filters so that the chip groups don't display, there's still a grid gap that appears to be visible. Is that expected?
2. When I clear all filters, then add one back so that there's only one chip, I see the following layout:
    image


I also wanted to mention that the html structure of the chip group is off, and that there's an issue I had created a while back for the original chip group component: https://github.com/patternfly/patternfly-react/issues/1454

But then I saw that the core example for data toolbar also has the same issue:
image

To simplify the example above, we basically have this html structure:
```
<ul><li></li></ul>
<ul><li></li></ul>
<ul><li></li></ul>
```
Instead of this:
```
<ul>
  <li></li>
  <li></li>
  <li></li>
</ul>
```

I'm not sure what the best path forward on this is. Should the issue in core be captured as a separate issue, and revisited in react when it's fixed?

nicolethoen:
@jgiardino thanks! i'll play around with these comments and makes sure these issues are related to the chip group component and not just my use of it. I'll report back

nicolethoen:
@jgiardino the extra grid gap is something I can fix - so I'm looking into that. I have been unable to reproduce the wrapping one chip. not sure how that's happening. does anyone else see that?

karelhala:
Thanks for opening this PR! Can we stop for a moment and discuss the API? I think that passing selected items to toolbar items is unlucky, we can't be certain what consumers include in these toolbar items so it doesn't seem to have any reason to bind together toolbar item and chips. I would expect that chips would be specific prop on root toolbar component at most in a shape of `PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node])` and one callback to remove some chips `onChipRemove` this function would be called for either global remove and one chip remove.

Imagine a usecase where user clicks on a tile and is brought to a table with filters, one filter is automatically aplied to show for instance `OS=Mac`, this filter cannot be removed or changed it is not represented in any filter in toolbar. Where should these chips go? Empty `toolbarItemWithChips`?

Plus the API and implementation will be greatly improved, only one place where to stick your filter values and only one place to control removing such chips.

**Long story short** - toolbar items doesn't need to know about selected filters and vice versa.

nicolethoen:
@karelhala 

I just spoke to @mcarrano and @mattnolting.
Here are a few things we have talked about.
- if a consumer wants to add a chip group without a filter, they can totally do that. The toolbar will accept any component as an item if they want. They could even put the chip group in its own second row in the toolbar.
- if they want an orphan chip group to behave like a chip group that is part of a filter, that will probably have to be a feature enhancement for a future release, because that is not how this feature is designed to work
- as far as the concerns about the API, that is something that can be investigated, and iterated on in future releases as well, since that is the purpose of an experimental component. but my instinct is still that filters with their chip groups should remained coupled. the toolbar should be agnostic

karelhala:
@nicolethoen yeah, I looked at the example on smaller screen and those chips needs to be moved to the specific toolbarItem. The problem here is that you have to write callback function for each of these chip groups/filters. So it would probably be better to have one callback function so bulk remove and single remove would be easier.

What do you think about passing these filters to data toolbar and keeping `DataToolbarItemWithChips`, but instead passing object as chips passing key in filters. The filters object would look like (given the example)
```JS
{
  risk: ['Low'],
  status: ['New', 'Pending']
}
```

And callback function for removing chips would have singature `onDelete(event, newFilters, group, removed)` where `newFilters` would be calculated filters after remove and `group` would be (for our example) either `risk` or `status` and removed would be filters that user wants to remove. This way it is much easier for consumers to notify users what they want to remove and also call server in more suitable manner.

Any filters that are not assigned to any item would be left off at the bottom above `Remove chips`. This way you can have filters that are not directly assigned to any filter and interacting with them is much easier (no need to have 3 callbacks for one action).

What do you think about this approach? I know that I could use chipgroup in tooblarContent, but these filter chip groups in this PR are easier to set up and ineract with them.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItemWithChipGroup.tsx
```diff
@@ -0,0 +1,86 @@
...
+import * as ReactDOM from 'react-dom';
+import { DataToolbarItem, DataToolbarItemProps } from './DataToolbarItem';
+import { ChipGroup, Chip, ChipGroupToolbarItem } from '../../../components/ChipGroup';
+import { DataToolbarContext } from './DataToolbarUtils';
+import styles from '@patternfly/react-styles/css/components/DataToolbar/data-toolbar';
+import { getModifier } from '@patternfly/react-styles';
+
+export interface DataToolbarItemWithChipGroupProps extends DataToolbarItemProps {
+  /** An array of strings to be displayed as chips in the expandable content */
+  chips?: string[];
```
karelhala: This should be array of `React.ReactNode` in order to allow translations and custom components.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItemWithChipGroup.tsx
```diff
@@ -0,0 +1,86 @@
...
+  /** Callback passed by consumer used to delete a chip from the chips[] */
+  deleteChip?: (chip: string) => void;
+  /** Content to be rendered inside the Data toolbar item associated with the chip group */
+  children: React.ReactNode;
+  /** Unique label for the chip group */
+  label: string;
```
karelhala: This should also be `React.ReactNode`.

karelhala:
Codewise looking good! I noticed one small bug thought, if you remove all filters trough their `x` buttons the `Clear all filters` is still present.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
```diff
@@ -52,6 +59,12 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
     }));
   }
 
+  updateShowClearFiltersButton = (showClearFiltersButton: boolean) => {
+    if (showClearFiltersButton !== this.state.showClearFiltersButton) {
```
mturley: You don't actually need to check to see whether a setState is necessary, if you call `this.setState` with the exact same (`===`) value that is already in state, React will optimize that for you and prevent a re-render IIRC. I think this kind of check is only necessary if the value is mutable and React wouldn't be able to tell that a different instance of it represents the same state. But for a boolean, it shouldn't be necessary (I think?)

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItemWithChipGroup.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  render() {
+    const { children, chips, deleteChip, label, ...props } = this.props;
+
+    return (
+      <DataToolbarContext.Consumer>
+        {({ isExpanded, chipGroupContentRef, updateShowClearFiltersButton }) => {
+
+          if (chips.length > 0) {
+            updateShowClearFiltersButton(true);
```
mturley: Updating state at render time is an antipattern we should avoid (it can lead to render loops and incorrect state, since React doesn't guarantee a render on every update). If you wanted to implement what this code is trying to do (call `updateShowClearFiltersButton` when the value of `chips` changes), that should be done in `componentDidUpdate` instead of `render`.

However, we're duplicating sources of truth here; ideally, `showClearFiltersButton` shouldn't be its own value in state at all, it should be derived from whether or not there are any filters applied across all chip groups.
Duplicating that state is problematic, I think it's causing the bug @karelhala noticed:
> if you remove all filters trough their x buttons the Clear all filters is still present.

This call (`if (chips.length > 0) { updateShowClearFiltersButton(true); }`) makes sure that when any chips appear the Clear all filters button is shown, but there is nothing making sure that when the last chip disappears the Clear all filters button is hidden. And since the DataToolbarItemWithChipGroup components only know about their own chips, there's no way for them to know that a chip being deleted is the last chip in the whole toolbar, so they can't confidently call that method with `false`.

I think the real way to fix all these problems is just to have `showClearFiltersButton` be a prop of `DataToolbar`, instead of having it in state, and to get rid of the whole `updateShowClearFiltersButton` method entirely. The example's root component is the only one who knows whether there are no chips left in any of the DataToolbarItems, so that prop in the example would just be defined as `showClearFiltersButton={statusSelection.length > 0 && riskSelection.length > 0}`.

The only true way to avoid that prop and make this truly "automatic" would be for the parent DataToolbar to "know about" the `chips` props of its children, which would mean they call some method to "register and unregister" their chips when they update... but that sounds like a huge mess, and it violates the React principle of one-way data flow. I think the right solution is to just make it a prop.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarItemWithChipGroup.tsx
nicolethoen: so, I renamed this prop to categoryName, because that is what it is - and categoryName is a prop on a chip group that must be a string.

nicolethoen:
I believe my changes address the remaining code review comments.

The only outstanding issue for this PR is the bug @karelhala noticed that clearing the last chip using its x button, the 'clear all filters' link does not disappear. That is not a super straight forward fix. I will see if I can get it working by tomorrow at a reasonable time. Otherwise I will open an issue for the next milestone. Anyone who wants to try to solve the puzzle before lunch tomorrow with me is more than welcome! :)

nicolethoen:
Alright! This last change should have addressed all outstanding comments and concerns. Rather than trying to combine DataToolbarItem with DataToolbarItemWithChipGroup, I renamed it to DataToolbarFilter. I think that is more intuitive for the consumer and I think it's more appropriate that the DataToolbarFilter uses DataToolbarItems rather than trying to get the consumer to think about items that include chip groups. If there are further iterations that need to be made on this aspect of the API, I suggest we create issues for that and move them to the next milestone at this point.

And the bug regarding the 'Clear all filters' button is gone. I don't love that the consumer manages that, but that can also be iterated upon in the future if that seems to be an issue for consumers.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarUtils.tsx
```diff
@@ -6,27 +6,16 @@ interface DataToolbarContextProps {
   toggleIsExpanded: () => void;
   expandableContentRef: RefObject<HTMLDivElement>;
   expandableContentId: string;
+  chipGroupContentRef: RefObject<HTMLDivElement>;
+  updateShowClearFiltersButton: (showClearFiltersButton: boolean) => void;
```
mturley: @nicolethoen looks like you missed removing a `updateShowClearFiltersButton` here

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
mturley: Turns out this might not be true in all cases, I may have misunderstood the API here (just commenting for posterity in case anyone reads this later)

mturley:
@nicolethoen, your last commit is causing tests to fail, I think we should address the problem with OptionsMenu in a different PR. Instead, we can avoid the bug you were running into by taking @mcarrano 's advice and using checkbox Select instead of OptionsMenu for the DataToolbarFilter examples.

In the interest of time I went ahead and implemented that for you, if you want to take my commit:
https://github.com/mturley/patternfly-react/commit/e290fbeac4a4e42a527545d4fa6835db5f09adb4

You should drop your last commit (https://github.com/patternfly/patternfly-react/pull/2969/commits/820d09b3318f713e4518d8ae9d89718cf54d599c) and then you can:

```
git remote add mturley https://github.com/mturley/patternfly-react.git
git fetch mturley
git cherry-pick e290fbeac4a4e42a527545d4fa6835db5f09adb4
```

PR #2981, opened 2019-09-20
fix(copytoclipboard): wrap content in pre tag

### LINKED ISSUE #2924: add prop to clipboardcopy expandable content to render code
Once https://github.com/patternfly/patternfly-next/pull/2260 is merged, this issue should be started. Add a prop like `isCode` to `pf-c-clipboard-copy__expanded-content` to add the html element `<pre>` which will render the text inside of `pf-c-clipboard-copy__expanded-content`  to display as code.

jenny-s51:
Closes #2924. Any feedback/suggestions would be very much appreciated regarding how to clean up this PR. Thanks!

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopyExpanded.tsx
+++ b/packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopyExpanded.tsx
@@ -42,8 +42,6 @@ export class ClipboardCopyExpanded extends React.Component<ClipboardCopyExpanded
       contentEditable={!isReadOnly}
       {...props}
     />;
-    // tslint:disable-next-line: no-console
-    console.log( typeof clipboardCopyExpandableComponent);
 
     return (isCode ?
     <pre><React.Fragment>{clipboardCopyExpandableComponent}</React.Fragment></pre>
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/ClipboardCopy/ClipboardCopyExpanded.tsx
```diff
@@ -30,16 +32,19 @@ export class ClipboardCopyExpanded extends React.Component<ClipboardCopyExpanded
...
+    <div
+      suppressContentEditableWarning
+      ref={this.contentRef}
+      className={css(styles.clipboardCopyExpandableContent, className)}
+      onInput={(e: any) => onChange(e.target.innerText, e)}
+      contentEditable={!isReadOnly}
+      {...props}
+    />;
+
+    return (isCode ?
```
jschuler: could probably work a little cleaner like this:
```
return isCode ? <pre>{clipboardCopyExpandableComponent}</pre> : clipboardCopyExpandableComponent
```

jenny-s51: Fixed -- thank you Joachim!

christiemolloy:
@jenny-s51 I can't see an example of code inside of the clipboard expanded content, i think mimicking the example in core will work here to demonstrate the `<pre>` tag

jenny-s51:
I just added the same prop to the parent component so now when you add the `isCode` prop to any of the demos, replace the sample text with code, and enclose it to look like `{` code `}` it should render as code normally. Is that what you meant?

christiemolloy:
Thanks for updating, some more comments. When I add `isCode` it looks like this:
Screen Shot 2019-09-23 at 4 56 45 PM

And what I should have specified is that you should add an example to the React workspace with the isCode prop so that there is consistency with Core

Screen Shot 2019-09-23 at 4 57 11 PM

jenny-s51:
@christiemolloy absolutely -- thanks for your feedback here, will plan to add a separate example for this

PR #2985, opened 2019-09-23
feat(Flex): Added support for flex layout, tests, etc.

### LINKED ISSUE #2860: Add `.pf-l-flex__item` wrapper to examples
Here is the related Core issue: https://github.com/patternfly/patternfly-next/issues/1984

The layout needs to be built with all the options.

rebeccaalpert:
Added React support for Flex and FlexItem. Moved Nicole's DataToolbar breakpoint utility to main utils and added tests so I could reuse it for Flex.

Fixes #2860.

### DIFF
--- a/packages/patternfly-4/react-core/src/layouts/Flex/examples/flex.scss
+++ b/packages/patternfly-4/react-core/src/layouts/Flex/examples/flex.scss
@@ -1,0 +1,23 @@
+.ws-preview {
+  & .pf-l-flex {
+    padding: 1rem;
+  }
+
+  & .pf-l-flex > pf-l-flex {
+    border-style: dashed;
+    border-width: 2px;
+    padding: 1rem;
+  }
+
+  & .pf-l-flex > div {
+    border-style: dashed;
+    border-width: 2px;
+    padding: 1rem;
+  }
+
+  .example-border {
+    border-style: dashed;
+    border-width: 2px;
+    padding: 1rem;
+  }
+}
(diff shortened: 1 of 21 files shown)


rebeccaalpert:
Added demo and integration test.

packages/patternfly-4/react-core/src/layouts/Flex/examples/flex.scss
```diff
@@ -0,0 +1,23 @@
+.ws-preview {
```
tlabaj: Why do we need this?

rebeccaalpert: I used it because other layouts seemed to be using it. I removed it to see what would happen and it throws off some of the existing styling.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/FlexDemo/FlexDemo.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Flex, FlexItem } from '@patternfly/react-core';
+
+export class FlexDemo extends React.Component {
+  render() {
+    return (
+      <Flex>
+        <FlexItem>Flex item</FlexItem>
```
tlabaj: can we add an example using the breakpointMods prop?

rebeccaalpert: Pretty much any of the lower variations use it. (Inline, flex-1, etc.)

PR #2988, opened 2019-09-23
fix(charts): replace empty tags with React.Fragment

dlabrecq:
In ChartBullet, we're use empty <>...</> tags instead of React.Fragment. For consistency, we want to use React.Fragment. It's possible that gatsby / empty tags generate a "Minified React error" on pf.org?

Fixes https://github.com/patternfly/patternfly-org/issues/1477


dlabrecq:
This should fix https://github.com/patternfly/patternfly-org/issues/1475

PR #2995, opened 2019-09-24
feat(Chip): Added OUIA compatibility to Chip component 

quarckster:
What: Added OUIA props to Tabs as per #2425

~Additional issues: tests failed with these errors and I cannot figure out what I should do next:~


redallen:
Same deal as with Select in #2996 .

quarckster:
@tlabaj fixed

Issue #2874, opened 2019-09-06
Compact pagination 


LHinson:
Introduce compact pagination as a follow up to core issue, https://github.com/patternfly/patternfly-next/issues/2110

mcoker:
Compact pagination was merged - https://github.com/patternfly/patternfly-next/pull/2275

tlabaj:
closed by #2998

PR #2998, opened 2019-09-24
feat(Pagination): add compact variation flag

kmcfaul:
**What**: Adds a flag to set whether the pagination component should be compact.

**Refer to issue**: #2874

 PR #2904 updates the html structure and styling of pagination's toggle to match core


packages/patternfly-4/react-core/src/components/Pagination/Pagination.test.tsx
```diff
@@ -12,7 +12,12 @@ describe('component render', () => {
     const wrapper = mount(<Pagination itemCount={20} variant={PaginationVariant.bottom} />);
     expect(wrapper).toMatchSnapshot();
   });
-
+  
+  test('should render correctly bottom', () => {
```
rebeccaalpert: I think this should be "should render correctly compact."

Issue #3015, opened 2019-09-26
Charts: Theme examples


dlabrecq:
A custom theme is typically applied to multiple charts, while `style` properties are applied to individual charts. As a user, I would like to view examples of the different approaches. 

An example showing how to create a custom theme that overrides default padding and colors. And as an alternative, an example showing how to override those same properties via the chart's `style` attribute.

These examples should be in its own "Theme" area, similar to the Sparline examples.

PR #3023, opened 2019-09-27
Charts: Add custom theme example

dlabrecq:
Added an example showing how to create a custom theme that overrides default padding and colors. And as an alternative, an example showing how to override those same properties via the chart's style attribute.

Fixes https://github.com/patternfly/patternfly-react/issues/3015

Screen Shot 2019-10-03 at 9 38 30 AM

Screen Shot 2019-10-03 at 9 51 05 AM


packages/patternfly-4/react-charts/src/components/ChartArea/examples/ChartArea.md
```diff
@@ -247,10 +245,148 @@ class MultiColorChart extends React.Component {
 }
 ```
 
+<<<<<<< HEAD
```
redallen: Merge conflict inside of here.

dlabrecq: Thanks, Zack! Fixed

PR #3030, opened 2019-09-27
SerialConsole: Use xterm v4

### LINKED ISSUE #2912: Update to xterm v4 in console
[Currently](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/react-console/package.json#L41) version 3.3.0 is used, which is a year old release and since then there has been some interesting improvements and fixes.

Now is also available version 4.0.0 which changes quiet a few things around.
One of them is the fact, that addons are now separate packages, so [fit](https://www.npmjs.com/package/xterm-addon-fit) package cannot be loaded as [is now](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/react-console/src/SerialConsole/XTerm.js#L5).

This block [cockpit](https://github.com/cockpit-project/cockpit) team from upgrading to xterm-4.0.0 since patternfly react console needs xterm < 3.14

marusak:
**What**: Use xterm v4 in SerialConsole. Fixes #2912

### DIFF
--- a/packages/patternfly-3/patternfly-react-extensions/package.json
+++ b/packages/patternfly-3/patternfly-react-extensions/package.json
@@ -60,7 +60,8 @@
     "react-bootstrap-typeahead": "^3.4.1",
     "rimraf": "^2.6.2",
     "shx": "^0.3.2",
-    "xterm": "^3.3.0"
+    "xterm": "^4.0.0",
+    "xterm-addon-fit": "^0.2.1"
   },
   "peerDependencies": {
     "prop-types": "^15.6.1",
--- a/packages/patternfly-3/react-console/README.md
+++ b/packages/patternfly-3/react-console/README.md
@@ -29,7 +29,7 @@ Example with LESS:
 @import "~bootstrap/less/variables";
 @import "~patternfly/dist/less/variables";
 @import "~patternfly-react/dist/less/patternfly-react.less";
-@import "~xterm/dist/xterm.css";
+@import "~xterm/css/xterm.css";
 @import "~@patternfly/react-console/dist/less/console.less";
 ```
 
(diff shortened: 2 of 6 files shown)


marusak:
Can I get a review on this? (not really sure who to ping, maybe @mareklibra as he has some commits in serial console)

jeff-phillips-18:
code changes look ok to me but I don't really know this package all that well. If @mareklibra is OK with it then I am.

marusak:
yes, users needs to move xterm to 4.0.0. But otherwise there should be no other visible change nor they do not need to adjust their code in any way.

dlabrecq:
As long as it's not considered a breaking change, I believe it would be ok.

martinpitt:
Sorry for nagging, but is there anything else to be done on our (@marusak 's) side? This seems approved and all tests passed. Thanks!

jeff-phillips-18:
Should this be a breaking change? Regardless, the commit message needs to be updated for semantic release. If breaking, then please use `major` otherwise `feat` would do.

marusak:
> Should this be a breaking change?

I answered this in a previous comment

> Regardless, the commit message needs to be updated for semantic release. If breaking, then please use major otherwise feat would do.

Can this be done on squashing? If it can, then I don't need to push force and then tests don't need to re-run. Otherwise I can edit the commit message (`feat(SerialConsole): Use xterm v4` would work, I guess?)

jeff-phillips-18:
Sorry, I guess my question is, if adopters are forced to move to xterm 4.0.0 then wouldn't this be a breaking change?

marusak:
> Sorry, I guess my question is, if adopters are forced to move to xterm 4.0.0 then wouldn't this be a breaking change?

I don't know what is defined as breaking change. It requires users to update xterm to `4.0.0`, but otherwise they should not notice any difference and they do not need to update their code (unless they use xterm on other places as well).

jeff-phillips-18:
I'm going to flag this for a major release as it would break consumers if they pull this change and do nothing else.

PR #3034, opened 2019-09-28
docs(button): add an example of a link button with anchor

### LINKED ISSUE #3020: Button: Add an example of a link button with anchor component
In the "Link button" example, add a link button that has an anchor component .  It is not clear from the examples that you when you specific the link variant you still need to specify component to be 'a' for the button to function as a hyperlink. 

e.g. of what to add
` <Button component="a" href="https://pf-next.com/" target="_blank" variant="link">
      Secondary Link to Core Docs
    </Button>`

boaz0:
**What**:

closes #3020

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Button/Button.md
+++ b/packages/patternfly-4/react-core/src/components/Button/Button.md
@@ -59,6 +59,9 @@ LinkButton = () => (
     <Button component="a" isDisabled href="https://pf-next.com/" target="_blank" variant="tertiary">
       Tertiary Link to Core Docs
     </Button>
+    <Button component="a" href="https://pf4.patternfly.org/contribution/#modifiers" variant="link">
+     Jump to modifiers in contribution guidelines
+    </Button>
   </React.Fragment>
 );
 ```


packages/patternfly-4/react-core/src/components/Button/Button.md
```diff
@@ -62,3 +62,23 @@ LinkButton = () => (
   </React.Fragment>
 );
 ```
+
```
tlabaj: Can  you please add the example to the  "Link Button" section  a Button with `variant="link"` and ` component=“a”` and and 'href prop.  See the example in the issue.

packages/patternfly-4/react-core/src/components/Button/Button.md
```diff
@@ -59,6 +59,15 @@ LinkButton = () => (
     <Button component="a" isDisabled href="https://pf-next.com/" target="_blank" variant="tertiary">
       Tertiary Link to Core Docs
     </Button>
+    <h2 id="anchor1">Anchor</h2>
```
tlabaj: Can you either modify this to have the href be a url or add another example (like the one in the issue).  The issue was opened because a consumer did not understand they needed to have `variant=link` and `component="a"` to get the link to work.

PR #3049, opened 2019-10-01
feat(Popover): Ability to hide close button

### LINKED ISSUE #3043: For Popover, allow to hide close box
In migrating the from PF3 to PF4 the new [popover](https://patternfly-react.surge.sh/patternfly-4/components/popover/) component always displays the close box and there is no way to hide the close box.

Here is the PF4 Popover with close box:
hide-closebox

This is our [Kiali](https://www.kiali.io) current PF3 screen:
old-popover
That is what I want to replicate with PF4 version of Popover.

SpyTec:
**What**: Adds ability to show/hide close button. To do this two props was added, `showClose` and `disableFocusTrap`. The latter is required when no focusable elements are in the popover.

An example was added to the bottom of Popover documentation page

**Additional issues**: fixes #3043

cc @mtho11

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Popover/Popover.md
+++ b/packages/patternfly-4/react-core/src/components/Popover/Popover.md
@@ -149,3 +149,23 @@ HeadlessPopover = () => (
   </Popover>
 );
 ```
+
+## Popover without close button
+```js
+import React from 'react';
+import { Popover, PopoverPosition, Button } from '@patternfly/react-core';
+
+SimplePopover = () => (
+  <Popover
+    showClose={false}
+    disableFocusTrap={true}
+    headerContent={<div>Popover Header</div>}
+    bodyContent={
+      <div>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla turpis.</div>
+    }
+    footerContent="Popover Footer"
+  >
+    <Button>Toggle Popover</Button>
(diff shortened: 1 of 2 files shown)


mtho11:
@mcarrano there is also a more complicated use case for namespace popover:
namnespace

SpyTec:
Now that I think about it, this can all be solved with Select really. It gives a dropdown and you can put in custom nodes as well, doesn't need to be text

Only thing you'd loose is the triangle pointing to it

kybaker:
I agree with @mcarrano here. It does seem like this is same desired effect can be achieved with a select which we already have in the library.

mceledonia:
@mcarrano Do you remember why we decided to do that? It might have been accessibility but I can't quite remember.

Either way I agree with @SpyTec , select should work well here!

jessiehuff:
I talked to Jenn to see if she could remember if there was accessibility reasoning behind this, but she didn't quite remember what the original driving reason was. She was saying her goal has been to make this component interaction (i.e. specifically the Dropdown with customizable panel) behave similar to popover so there wouldn't really be any accessibility issues with not having a close button (at least until proven otherwise via usability research). I think it's really a question of design instead of accessibility. And for the use case mentioned, I think the suggestion from design to use the other component makes sense.

tlabaj:
Closing this issue.  It looks like another component  will be used for the use case.  Please open another issue if have any more trouble.

PR #3053, opened 2019-10-01
feat(tooltip): allow tooltip text to be left-aligned

### LINKED ISSUE #3013: Allow tooltip text to be left-aligned
The related Core PR is here: https://github.com/patternfly/patternfly-next/pull/2284?

In this issue we should allow for the tooltip text to be left-aligned. This issue will be populated with more description.

jenny-s51:
will close #3013

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
+++ b/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
@@ -2,10 +2,10 @@
 title: 'Tooltip'
 cssPrefix: 'pf-c-tooltip'
 typescript: true
-propComponents: ['Tooltip']
+propComponents: ['Tooltip', 'TooltipContent']
 ---
 
-import { Button, Tooltip, TooltipPosition, Checkbox } from '@patternfly/react-core';
+import { Button, Tooltip, TooltipPosition, TooltipContent, Checkbox } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 
 ## Simple tooltip
@@ -14,6 +14,21 @@ import React from 'react';
 import { Tooltip, TooltipPosition } from '@patternfly/react-core';
 
 <Tooltip
+  content={
+    <div>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla turpis.</div>
+  }
+>
+  <span>I have a tooltip!</span>
+</Tooltip>
+```
+
+## Simple tooltip with left-aligned text
+```js
+import React from 'react';
+import { Tooltip, TooltipPosition } from '@patternfly/react-core';
+
+<Tooltip
+  isTextLeftAligned
   content={
     <div>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla turpis.</div>
   }
--- a/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
+++ b/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
@@ -28,7 +28,7 @@ import React from 'react';
 import { Tooltip, TooltipPosition } from '@patternfly/react-core';
 
 <Tooltip
-  isTextLeftAligned
+  isContentLeftAligned
   content={
     <div>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla turpis.</div>
   }
--- a/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.tsx
@@ -77,6 +77,7 @@ export class Tooltip extends React.Component<TooltipProps> {
     position: 'top',
     trigger: 'mouseenter focus',
     isVisible: false,
+    isTextLeftAligned: false,
     enableFlip: true,
     className: '',
     entryDelay: 500,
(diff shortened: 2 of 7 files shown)


redallen:
Very close:
```
Snapshot Summary
 › 2 snapshots failed from 1 test suite. Inspect your code changes or run `yarn run test:pf4 -u` to update them.

Test Suites: 1 failed, 157 passed, 158 total
Tests:       2 failed, 849 passed, 851 total
Snapshots:   2 failed, 621 passed, 623 total
```

packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -61,6 +61,8 @@ export interface TooltipProps {
   position?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Tooltip trigger: click, mouseenter, focus, manual  */
   trigger?: string;
+  /** Flag to indicate that the text content is left aligned */
+  isContentLeftAligned?: boolean;
```
dlabrecq: Why only support a left aligned option? What if designers want to support right-aligned text, next?

Even if we don't support right-aligned text for now, something more like this would scale better in the future.

```
textPosition?: 'bottom' | 'center' | 'right';
```

Then we could create an enum

```
export enum TooltipTextPosition {
  bottom = 'bottom',
  center = 'center',
  right = 'right'
}
```

Looks like we already take this approach for the `position` property.

jenny-s51: Sure, I'm happy to add a right-aligned option. I was wondering why this issue was opened for left-aligned text only. @dlabrecq @tlabaj would you prefer I open a new issue to add support for right-aligned content? Or add an `enum` to this PR instead, like you have above?

packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
```diff
@@ -2,10 +2,10 @@
 title: 'Tooltip'
 cssPrefix: 'pf-c-tooltip'
 typescript: true
-propComponents: ['Tooltip']
+propComponents: ['Tooltip', 'TooltipContent']
```
redallen: Good catch, we export `TooltipContent` in `Tooltip/index.ts` but don't show its props in the docs.

Edit: We should add a followup issue to provide examples on how/why to use `TooltipContent`.

packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
```diff
@@ -2,10 +2,10 @@
...
 ---
 
-import { Button, Tooltip, TooltipPosition, Checkbox } from '@patternfly/react-core';
+import { Button, Tooltip, TooltipPosition, TooltipContent, Checkbox } from '@patternfly/react-core';
```
redallen: Unused import.

packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.md
jenny-s51: All set! #3088

packages/patternfly-4/react-core/src/components/Tooltip/Tooltip.tsx
jenny-s51: @dlabrecq @tlabaj I've opened a separate issue for this at #3089. If agreeable with you, I'm happy to assign myself to it and add support right-aligned content as well.

dlabrecq: I wasn't necessarily asking for right-aligned tooltips, just that we use a property that is more scalable. Once we introduce the `isContentLeftAligned` prop, it will be difficult to change without a breaking change. That's why I suggested something like the existing `position` property.

dlabrecq: I wasn't necessarily asking for right-aligned tooltips, just that we use a property that is more scalable. Once we introduce the isContentLeftAligned prop, it will be difficult to change without a breaking change. That's why I suggested we use something more like the existing position property.

I just want to consider a use case where we end up with both isContentLeftAligned and isContentRightAligned props together? Or, would it be better to have something more like contentPosition?: 'bottom' | 'center' | 'right' where we can easily add future support.

tlabaj: Dan makes good point here.  I don't think there are plans of introducing the right aligned variant and the naming here seems to be consistent with other components.  That being said those components would be hard to change if we added another variant.

jenny-s51: @tlabaj Thank you for your feedback! Does that mean I should change the prop to look like `contentPosition?: 'left' | 'center'`?

dlabrecq: If you don't feel we'll ever have another variant, I'm fine leaving it as is. Just wanted to raise the question.

dlabaj: @jenny-s51 I think we should use `contentPosition?: 'left' | 'center'`.  It's also consistent with the position enum we are already using in this component.  Can we change it to that?

jenny-s51: @dlabaj Absolutely -- I'm on it

tlabaj: @jenny-s51 I like what @dlabrecq suggested.  But, if we are not going to have a right alignment variation I don't see the need to introduce the inconstancy.   We have at least one other component that uses isLeftAligned (or similar).
@mcarrano can you verify that there is no intent on introducing a right aligned variant.

mcarrano: I see no reason to support right-aligned tooltips.

PR #3068, opened 2019-10-03
feat(Expandable): Allow for dynamic toggle text in uncontrolled version

### LINKED ISSUE #2914: add support for expanded vs collapsed text in Expandable uncontrolled variant
Support for the expanded vs collapsed text for Expandable uncontrolled variant.

This issue, https://github.com/patternfly/patternfly-react/issues/2350, requested an uncontrolled variant for the Expandable component. Which was implemented. However it also requested support to supply two pieces of text. One for expanded, and another for collapsed. Without this feature, we're left with having to add an `onToggle` listener and track the expanded state still which defeats the purpose of using the uncontrolled variant.

Even the demos exhibit this missing feature:
expand-collapse

Ideally the Expandable component could be used as follows:
```
  <Expandable
    textExpanded="Show Less"
    textCollapsed="Show More"
  >...</Expandable>
```

rebeccaalpert:
Added two new props that allow user to specify collapsed and expanded toggle text for uncontrolled expandable.

Fixes #2914.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
+++ b/packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
@@ -10,8 +10,12 @@ export interface ExpandableProps {
   className?: string;
   /** Flag to indicate if the content is expanded */
   isExpanded?: boolean;
-  /** Text that appears in the  toggle */
+  /** Text that appears in the toggle */
   toggleText?: string;
+  /** Text that appears in the toggle when expanded (will override toggleText if both are specified; used for uncontrolled expandable with dynamic toggle text) */
+  toggleTextExpanded?: string;
+  /** Text that appears in the toggle when collapsed (will override toggleText if both are specified; used for uncontrolled expandable with dynamic toggle text) */
+  toggleTextCollapsed?: string;
   /** Callback function to toggle the expandable content */
   onToggle?: () => void;
   /** Forces focus state */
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
```diff
@@ -38,12 +42,24 @@ export class Expandable extends React.Component<ExpandableProps, ExpandableState
...
+    }
+    if (!propOrStateIsExpanded && toggleTextCollapsed !== '') {
+      return toggleTextCollapsed;
+    }
+    return toggleText;
+  }
+
   render() {
     const {
       onToggle: onToggleProp,
```
dlabrecq: On line 77, onToggle is defined, again

`let onToggle = onToggleProp;`

packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
```diff
@@ -68,6 +86,8 @@ export class Expandable extends React.Component<ExpandableProps, ExpandableState
       };
     }
 
+    let computedToggleText = this.calculateToggleText(toggleText, toggleTextExpanded, toggleTextCollapsed, propOrStateIsExpanded);
```
dlabrecq: This should be a const, since it never changes

packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
rebeccaalpert: It looks like this is being done intentionally by whoever wrote that section.

rebeccaalpert:
Adjusted variable definition in amended commit.

packages/patternfly-4/react-core/src/components/Expandable/Expandable.tsx
dlabrecq: A comment would help here, but I won't hold up the PR

PR #3077, opened 2019-10-06
fix(button): component type should accept react-router Link

### LINKED ISSUE #2980: Button shows error in console on reload
When I go to a page that has a button the page loads fine without any error in the console.  However, when I refresh the page I get a console error. `Warning: Failed prop type: Invalid prop component supplied to `Button`, expected a ReactNode.`

Button Bug

boaz0:
**What**:

closes #2980 

ReactNode is any of the following types: `ReactChild`, `ReactFragment`, `ReactPortal`, `boolean`, `null`, `undefined`.[1]

While `Link` is a component type: `ComponentClass` or `FunctionComponent`.[2]

This PR adds `ComponentType` to the type of `component`. I did set it to `any` to allow other router components in the future or a custom router link. But if we want to support only react-router `Link` I can add a specific type to the component type there.

[1]https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L191
[2]https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L71

//cc @AlexSCorey

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Button/Button.tsx
+++ b/packages/patternfly-4/react-core/src/components/Button/Button.tsx
@@ -25,7 +25,7 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   /** Additional classes added to the button */
   className?: string;
   /** Sets the base component to render. defaults to button */
-  component?: React.ReactNode;
+  component?: React.ElementType<any>;
   /** Adds active styling to button. */
   isActive?: boolean;
   /** Adds block styling to button */


packages/patternfly-4/react-core/src/components/Button/Button.tsx
```diff
@@ -25,7 +25,7 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   /** Additional classes added to the button */
   className?: string;
   /** Sets the base component to render. defaults to button */
-  component?: React.ReactNode;
+  component?: React.ReactNode | React.ComponentType<any>;
```
redallen: Does just `component?: React.Element` work? I _think_ it should and that it's the better type to use.

boaz0: I assume you meant `React.ElementType`.

redallen: Yessir!

Issue #2910, opened 2019-09-11
Add bulk select to table 


LHinson:
Design file: https://marvelapp.com/i50gj84
related to #2876

mcarrano:
Updated design link to reflect what I believe to be the approved/latest designs.

mcarrano:
Also note that this may be more like a demo rather than an updated component as it requires interaction between the selector in the Toolbar and the Table. It's not really adding this capability to the table directly. There is an issue for this demo open in Core here: https://github.com/patternfly/patternfly-next/issues/1878

PR #3082, opened 2019-10-07
feat(BulkSelect demo): add bulk select demo with table and toolbar

kmcfaul:
**What**: adds an interactive demo between table and toolbar for bulk select functionality.

**Refer to issue**: #2910

### DIFF
--- a/packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
+++ b/packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
@@ -161,10 +161,12 @@ class BulkSelectTableDemo extends React.Component {
               <DropdownToggleCheckbox
                 id="example-checkbox-2"
                 key="split-checkbox"
-                aria-label="Select all"
+                aria-label={anySelected ? 'Deselect all' : 'Select all'}
                 isChecked={isChecked}
-                onClick={() => this.onDropDownToggle(!isDropDownOpen)}
-                onChange={anySelected ? () => this.handleSelectClick('none') : () => this.handleSelectClick('all')}
+                onClick={() => {
+                  this.onDropDownToggle(!isDropDownOpen);
+                  anySelected ? this.handleSelectClick('none') : this.handleSelectClick('all');
+                }}
               >
                 {numSelected !== 0 && <React.Fragment>{numSelected} selected</React.Fragment>}
               </DropdownToggleCheckbox>
@@ -184,6 +186,7 @@ class BulkSelectTableDemo extends React.Component {
         <ToolbarGroup>
           <ToolbarItem className="pf-u-mr-md">{this.buildSelectDropdown()}</ToolbarItem>
         </ToolbarGroup>
+        {this.renderPagination()}
       </Toolbar>
     );
   }
--- a/packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
+++ b/packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
@@ -136,7 +136,21 @@ class BulkSelectTableDemo extends React.Component {
     const { isDropDownOpen, selectedItems, total } = this.state;
     const numSelected = selectedItems.length;
     const allSelected = numSelected === total;
+    const anySelected = numSelected > 0;
     const isChecked = allSelected ? true : numSelected > 0 ? null : false;
+
+    const items = [
+      <DropdownItem key="item-1" onClick={() => this.handleSelectClick('none')}>
+        Select none (0 items)
+      </DropdownItem>,
+      <DropdownItem key="item-2" onClick={() => this.handleSelectClick('page')}>
+        Select page ({this.state.perPage} items)
+      </DropdownItem>,
+      <DropdownItem key="item-3" onClick={() => this.handleSelectClick('all')}>
+        Select all ({this.state.total} items)
+      </DropdownItem>
+    ];
+
     return (
       <Dropdown
         onSelect={this.onDropDownSelect}
@@ -150,7 +164,7 @@ class BulkSelectTableDemo extends React.Component {
                 aria-label="Select all"
                 isChecked={isChecked}
                 onClick={() => this.onDropDownToggle(!isDropDownOpen)}
-                onChange={allSelected ? () => this.handleSelectClick('none') : () => this.handleSelectClick('all')}
+                onChange={anySelected ? () => this.handleSelectClick('none') : () => this.handleSelectClick('all')}
               >
                 {numSelected !== 0 && <React.Fragment>{numSelected} selected</React.Fragment>}
               </DropdownToggleCheckbox>
@@ -159,17 +173,7 @@ class BulkSelectTableDemo extends React.Component {
           ></DropdownToggle>
         }
         isOpen={isDropDownOpen}
-        dropdownItems={[
-          <DropdownItem key="item-1" onClick={() => this.handleSelectClick('none')}>
-            Select none (0 items)
-          </DropdownItem>,
-          <DropdownItem key="item-2" onClick={() => this.handleSelectClick('page')}>
-            Select page ({this.state.perPage} items)
-          </DropdownItem>,
-          <DropdownItem key="item-3" onClick={() => this.handleSelectClick('all')}>
-            Select all ({this.state.total} items)
-          </DropdownItem>
-        ]}
+        dropdownItems={items}
       />
     );
   }
--- a/packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
+++ b/packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
@@ -5,12 +5,12 @@ section: 'demos'
 
 ## Bulk select table demo
 
-import { Pagination, PaginationVariant, Title, Checkbox } from '@patternfly/react-core';
+import { Pagination, Title, Checkbox } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody} from '@patternfly/react-table';
 
 ```js
 import React from 'react';
-import { Pagination, PaginationVariant, Title, Checkbox } from '@patternfly/react-core';
+import { Pagination, Title, Checkbox } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
 
 class BulkSelectTableDemo extends React.Component {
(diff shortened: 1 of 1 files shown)


mcoker:
Besides the interaction @mcarrano mentioned, there seems to be a bug where if you select all on the page, you get the indeterminate state checkbox. However if you select all, then select the page, you're left with the checked checkbox. I would expect the checkbox to be the same in those 2 scenarios.

kmcfaul:
@mcarrano Should the 'select page' button in the dropdown behave as a toggle for the page (deselecting, if page is selected)? Per dlabreq's feedback.

kmcfaul:
@mcarrano I will update the checkbox to act as a 'select all' toggle

@mcoker If you select all, and select the page, the demo currently checks for unchecked rows and checks them, adding them to the selected count/state. The total number of items doesn't change, all are still selected, so the check mark should be correct. Unless the behavior is meant to be 'when the page is all selected, unselect the whole page'.

mcoker:
@kmcfaul ah gotcha! I misunderstood, I thought if you went from "select all" to "select page", it would deselect anything that wasn't on the current page.

kmcfaul:
That's a good point though @mcoker, it could work like that and make sense. How do we want this 'select page' interaction @mcarrano?

mcarrano:
@kmcfaul a couple of comments...

1- The checkbox now works as expected. Thanks for fixing that.
2- I don't think I'd expect clicking 'Select page' when all are selected would deselect items on other pages, but it's a reasonable question. Let's leave this alone for now.
3- @dlabrecq makes an interesting point about having 'Select page' toggle to 'Deselect page' when on a page that is already selected. The only think that's tricky is what happens if I just deselect one item on that page, does it stay deselect? Toggle back? It would require tracking the selection state of a page. @katierik what do you think.
4- I seem to have com across another bug. If I select 2 pages, i.e. 40 items are selected, and then move to the next page, suddenly all 100 items become selected. @kmcfaul can you try to reproduce?

katierik:
In the POC for this we found that when anything was selected, it made the most sense to *deselect* is the user clicked on the bulk select box again (vs the original design/this implementation that clicking the partially selected selects more). 
I suspect that change would account for part of the "select page" confusion.

kmcfaul:
@mcarrano I've been unable to reproduce the selection bug. Selecting the two pages manually or with two sets of 'select page' and then clicking on the next page button doesn't select all for me. Is it the next page button that is triggering the selection for you or is it another button?

kmcfaul:
@katierik So the interaction for the select page should be: when there are no selections, select the whole page; when there are some selections / when the whole page is selected, deselect the whole page?

katierik:
@kmcfaul yes! 
This gif may help (it doesn't show the clicks though :/) 
http://www.giphy.com/gifs/cn2fWChDzdmzZVnKKS
1 - Click the toolbar select all - whole list is selected
2 - Click the first two items, then click the toolbar select all about - whole list deselected

kmcfaul:
This commit changes the top level toggle to clear selections when anything is selected (previously, selected everything first and only would deselect if everything was selected). That should match the gif.

Should the 'select page' button switch between select and deselect for just the page though? I'm still confused if that should go in or not. @katierik

mcarrano:
> @mcarrano I've been unable to reproduce the selection bug. Selecting the two pages manually or with two sets of 'select page' and then clicking on the next page button doesn't select all for me. Is it the next page button that is triggering the selection for you or is it another button?

@kmcfaul I did some more detective work on this and here's what seems to be happening. If I'm careful to click on the toggle in the split button it works fine, but if I accidentally click on the text, it's the same as clicking the checkbox. Can we make it so the text is not active?

Also, I think it would help the demo if we can drop a pagination component in the top toolbar as well as the footer. It would just be easier to know what page I'm on. Thanks!

kmcfaul:
@mcarrano The text fires off the click/change events natively as it is a label for the checkbox. I would have to restructure the DropdownToggleCheckbox component a bit to disable the text firing off these events. Should I go ahead and include that in this PR, or would this be a separate issue?

I will add the extra pagination location.

mcarrano:
@kmcfaul Yes, I think we will want to do that restructuring. Normally we do want to have a checkbox label toggle the state of a checkbox, but in this case the text does not label the checkbox exactly, so it becomes a usability problem. I would be OK if clicking on the text either did nothing or if it opened the dropdown menu. Will this also require a core change? There may also be an accessibility consideration here. @jessiehuff thoughts?

As for whether to combine with this PR or open a new issue, I will defer to @tlabaj . If you need me to open a separate issue for this, I can do that.

kmcfaul:
@mcarrano I found an alternative, taking the text out of the checkbox component, the spacing is a little wider but it should behave as intended. I also took out the behavior where clicking the checkbox also opens the dropdown, but if that should be preset I can add it back in.

mcarrano:
Excellent @kmcfaul !  I actually like the wider spacing as it makes it clear that the text and checkbox are not functionally connected. I think the only pending question is whether the 'select page' should toggle to 'deselect page'. Let me connect with @katierik today and we'll give you an answer.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+    this.onDropDownSelect = event => {
+      this.setState({
+        isDropDownOpen: !this.state.isDropDownOpen
+      });
+    };
+  }
+
+  fetch(page, perPage) {
+    this.setState({ loading: true });
+    fetch(`https://jsonplaceholder.typicode.com/posts?_page=${page}&_limit=${perPage}`)
```
redallen: Not a huge fan of being dependent on jsonplaceholder.typicode.com , but understandable for now.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+      cells: [post.title, post.body],
+      selected: post.selected
+    }));
+
+    return (
+      <React.Fragment>
+        {this.renderToolbar()}
+        {!loading && (
+          <Table
+            header={<div></div>}
```
seanforyou23: Can we supply an `aria-label` prop instead of this empty header div? Something like `aria-label="Bulk Select Table Demo"` would make the table show up nicely in the rotor menu.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+            rows={rows}
+            onSelect={this.onSelect}
+            canSelectAll={false}
+          >
+            <TableHeader />
+            <TableBody />
+          </Table>
+        )}
+        {loading && (
+          <center>
```
seanforyou23: Can we position this in a patternfly way, like maybe `<div className="pf-l-bullseye">` or something like that?

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+      />
+    );
+  }
+
+  buildSelectDropdown() {
+    const { isDropDownOpen, selectedItems, total } = this.state;
+    const numSelected = selectedItems.length;
+    const allSelected = numSelected === total;
+    const anySelected = numSelected > 0;
+    const isChecked = allSelected ? true : numSelected > 0 ? null : false;
```
seanforyou23: Can we avoid using nested ternary? Breaking the statements out into separate vars would help the readability, at the least we could include parenthesis around the expressions to make it easier to follow with the eyeballs.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+
+    this.onDropDownToggle = isOpen => {
+      this.setState({
+        isDropDownOpen: isOpen
+      });
+    };
+
+    this.onDropDownSelect = event => {
+      this.setState({
+        isDropDownOpen: !this.state.isDropDownOpen
```
seanforyou23: Can we use the [function form of setState](https://medium.com/@wisecobbler/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1) instead? Might help prevent some bugs down the road.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+import React from 'react';
+import { Pagination, PaginationVariant, Title, Checkbox } from '@patternfly/react-core';
+import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+
+class BulkSelectTableDemo extends React.Component {
+  constructor(props) {
+    this.state = {
+      res: [],
+      perPage: 20,
+      total: 100,
```
seanforyou23: Does `total` need to be a state property? Might be good to only capture properties that need to change in the component state.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,225 @@
...
+      .then(resp => this.setState({ res: resp, perPage, page, loading: false }))
+      .then(() => this.updateSelected())
+      .catch(err => this.setState({ error: err, loading: false }));
+  }
+
+  componentDidMount() {
+    this.fetch(this.state.page, this.state.perPage);
+  }
+
+  renderPagination(variant = 'top') {
```
seanforyou23: Since we don't customize this `variant` property anywhere in the demo, it might help readability to simply use the string value 'top' directly when setting this property like `variant={'top'}`, then we could remove this param/default here.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
kmcfaul: The empty div here is to resolve a console error with table (as it expects a header/caption/etc). I will add some text here and the aria label to make it clearer.

packages/patternfly-4/react-core/src/demos/BulkSelectTable.md
```diff
@@ -0,0 +1,228 @@
...
+---
+
+## Bulk select table demo
+
+import { Pagination, PaginationVariant, Title, Checkbox } from '@patternfly/react-core';
+import { Table, TableHeader, TableBody} from '@patternfly/react-table';
+
+```js
+import React from 'react';
+import { Pagination, PaginationVariant, Title, Checkbox } from '@patternfly/react-core';
```
seanforyou23: Can probably remove `PaginationVariant` from the imports

kmcfaul: You are right, good catch!

PR #3083, opened 2019-10-07
feat(eslint): add eslint-plugin-react-hooks

jeremylenz:
* Add [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) to `eslint-plugin-patternfly-react` so that Foreman and plugin devs will benefit from its guidance when using Hooks.

This plugin checks if you're following the [Rules of Hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) and is useful for anyone developing components that use React Hooks.

In Foreman we've started using Hooks, but don't yet have this rule.  See [this discussion](https://github.com/theforeman/foreman/pull/6981#issuecomment-535225407) with @sharvit

### DIFF
--- a/packages/eslint-plugin-patternfly-react/lib/config/recommended.js
+++ b/packages/eslint-plugin-patternfly-react/lib/config/recommended.js
@@ -47,7 +47,9 @@ module.exports = {
     'react/jsx-filename-extension': 'off',
     'react/jsx-uses-vars': 'error',
     'react/no-danger': 'off',
-    'react/sort-comp': 'off'
+    'react/sort-comp': 'off',
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn'
   },
   extends: [
     'standard',
@@ -64,6 +66,6 @@ module.exports = {
     node: true,
     jest: true
   },
-  plugins: ['prettier', 'jest', 'react', 'patternfly-react'],
+  plugins: ['prettier', 'jest', 'react', 'react-hooks', 'patternfly-react'],
   parser: 'babel-eslint'
 };
--- a/packages/eslint-plugin-patternfly-react/package.json
+++ b/packages/eslint-plugin-patternfly-react/package.json
@@ -35,6 +35,7 @@
     "eslint-plugin-prettier": "^2.6.0",
     "eslint-plugin-promise": "^3.7.0",
     "eslint-plugin-react": "^7.7.0",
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin-patternfly-react/lib/config/recommended.js
```diff
@@ -47,7 +47,9 @@ module.exports = {
     'react/jsx-filename-extension': 'off',
     'react/jsx-uses-vars': 'error',
     'react/no-danger': 'off',
-    'react/sort-comp': 'off'
+    'react/sort-comp': 'off',
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn'
```
dlabrecq: I don't believe we should be adding rules that we don't use ourselves in patternfly-react, even if it's good practice for others. How would folks disable the rule if they have trouble with it down the road?

jeremylenz: @dlabrecq I think the idea was that since `eslint-plugin-patternfly-react` is used in Foreman and most of its plugins, it would be better to just add the rule here in one place rather than every individual repo.

Like any eslint rule, these can be disabled with an `eslint-disable` or `eslint-disable-next-line` magic comment, or I think you can probably override rules in your own `.eslintrc` file.

dlabrecq: Sounds reasonable, but a little odd considering PatternFly does not use hooks. Can you please create an issue?

dlabrecq: //cc @dgutride

jeremylenz: @dlabrecq Ok, created #3092

sharvit: @dlabrecq patternfly-react are not going to use hooks in the future?

dlabrecq: Anything is possible, but we don't yet have any plans to introduce it. 

A benefit of using hooks is you don't need to write a class. Considering the majority of PF4 components don't maintain state, we only have a few component classes (e.g., tabs, dropdown, etc.). I don't know if hooks would be helpful enough to refactor those few components? 

I believe the PF team wants to discuss this a bit more...

Issue #2607, opened 2019-07-29
Tabs as nav example doesn't match html equivalent


jgiardino:
When I look at the HTML examples for Tabs, the [Tabs using the nav element](https://www.patternfly.org/v4/documentation/core/components/tabs#tabs-using-the-nav-element) shows the tabs as `<a>` instead of `<button>` elements. 

When I look at a similar [example for react](https://www.patternfly.org/v4/documentation/react/components/tabs/#tabs-using-the-nav-element), I don't see `<a>` elements being used for the tabs. And it isn't obvious to me from this example how to use `<a>` elements instead of `<button>` elements.

If I add `href="#someurl"` to a `<Tab>` and inspect the html, I do see that the tab converts to a `<a href="#someurl" ...>` instead of a `<button>`. But I don't see this prop/attribute noted in the props documentation. 

For this issue:
- Should an item be added for `href` in the `<Tab>` props documentation? 
  - actually, are there other props that should be included for `<Tab>`? I don't see a section listing any props for `<Tab>`
- Can we update the react example to have parity with the html example, and use `<a>` instead of `<button>` in the examples that use the `<nav>` element?

cmwylie19:
Hi Guys, just submitted the fix for this. I made the react example identical to the html version.

rachael-phillips:
Hi @cmwylie19 do you mind linking the issue here? @tlabaj I'm going to assign @cmwylie19

cmwylie19:
sure https://github.com/patternfly/patternfly-react/issues/2607 @rachael-phillips thanks! Submitting a PR now

tlabaj:
Closed by #3527

PR #3084, opened 2019-10-07
Update Tabs nav example to match core

cmwylie19:
https://github.com/patternfly/patternfly-react/issues/2607

**What**: updated the Tabs.md file to make the react example match the html example.

**Additional issues**:

Was a pleasure!

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/Tabs.md
+++ b/packages/patternfly-4/react-core/src/components/Tabs/Tabs.md
@@ -449,22 +449,63 @@ class TabsNavVariant extends React.Component {
 
   render() {
     return (
-      <Tabs
-        activeKey={this.state.activeTabKey}
-        onSelect={this.handleTabClick}
-        aria-label="Local"
-        variant={TabsVariant.nav}
-      >
-        <Tab eventKey={0} title="Tab item 1">
-          Tab 1 section
+       <React.Fragment>
+        <Tabs activeKey={this.state.activeTabKey} onSelect={this.handleTabClick} leftScrollAriaLabel="scrollLeft" rightScrollAriaLabel="right">
+          <Tab eventKey={0} title="Tab item 1">
+            Tab 1 section
         </Tab>
-        <Tab eventKey={1} title="Tab item 2">
-          Tab 2 section
+          <Tab eventKey={1} title="Tab item 2">
+            Tab 2 section
         </Tab>
-        <Tab eventKey={2} title="Tab item 3">
-          Tab 3 section
+          <Tab eventKey={2} title="Tab item 3">
+            Tab 3 section
         </Tab>
-      </Tabs>
+          <Tab eventKey={3} title="Tab item 4">
+            Tab 4 section
+        </Tab>
+          <Tab eventKey={4} title="Tab item 5">
+            Tab 5 section
+        </Tab>
+          <Tab eventKey={5} title="Tab item 6">
+            Tab 6 section
+        </Tab>
+          <Tab eventKey={6} title="Tab item 7">
+            Tab 7 section
+        </Tab>
+          <Tab eventKey={7} title="Tab item 8">
+            Tab 8 section
+        </Tab>
+          <Tab eventKey={8} title="Tab item 9">
+            Tab 9 section
+        </Tab>
+          <Tab eventKey={9} title="Tab item 10">
+            Tab 10 section
+        </Tab>
+          <Tab eventKey={10} title="Tab item 11">
+            Tab 11 section
+        </Tab>
+          <Tab eventKey={11} title="Tab item 12">
+            Tab 12 section
+        </Tab>
+        </Tabs>
+        <br />
+        <Tabs
+          activeKey={this.state.activeTabKey}
+          onSelect={this.handleTabClick}
+          aria-label="Local"
+          variant={TabsVariant.nav}
+        >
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/react-core/src/components/Tabs/Tabs.md
```diff
@@ -449,22 +449,63 @@ class TabsNavVariant extends React.Component {
...
-        aria-label="Local"
-        variant={TabsVariant.nav}
-      >
-        <Tab eventKey={0} title="Tab item 1">
-          Tab 1 section
+       <React.Fragment>
+        <Tabs activeKey={this.state.activeTabKey} onSelect={this.handleTabClick} leftScrollAriaLabel="scrollLeft" rightScrollAriaLabel="right">
+          <Tab eventKey={0} title="Tab item 1">
+            Tab 1 section
         </Tab>
```
dlabrecq: The `</Tab>` tags are not indented properly.

jgiardino:
Thanks for working this issue, @cmwylie19! 

One of our goals with the React component examples is that they align with the core HTML examples. Part of this issue is to bring the React examples [[1]](https://www.patternfly.org/v4/documentation/react/components/tabs/#tabs-using-the-nav-element) [[2]](https://www.patternfly.org/v4/documentation/react/components/tabs/#secondary-tabs-using-the-nav-element) into closer alignment with the [HTML example](https://www.patternfly.org/v4/documentation/core/components/tabs#tabs-using-the-nav-element) which shows how to use `<a>` html elements with the component.

As part of this issue, the react examples **Secondary tabs using the nav element** [[1]](https://www.patternfly.org/v4/documentation/react/components/tabs/#tabs-using-the-nav-element)  and **Tabs using the nav element** [[2]](https://www.patternfly.org/v4/documentation/react/components/tabs/#secondary-tabs-using-the-nav-element) should demonstrate using `href="#"` on the `<Tab>` components to result in using the `<a>` html element instead of `<button>` for the tabs (i.e. `.pf-c-tabs__button`).

Additionally, I see that the **Tabs using the nav element** includes an additional set of tabs that is not using `variant={TabsVariant.nav}`. I'm not sure these extra tabs are needed to demonstrate this behavior, but I also don't have a strong preference. I'm fine with either removing them, or updating them to be consistent with the other examples that use `<nav>` as the `.pf-c-tabs` element.

packages/patternfly-4/react-core/src/components/Tabs/Tabs.md
cmwylie19: Thanks for the feedback, will work to address your two concerns.

cmwylie19:
Hi @jgiardino, it is a pleasure.

> As part of this issue, the react examples Secondary tabs using the nav element [1] and Tabs using the nav element [2] should demonstrate using href="#" on the <Tab> components to result in using the <a> html element instead of <button> for the tabs (i.e. .pf-c-tabs__button).

Regarding the statement above, I did gathered that was your intention from the comment on the issue but i did not address it directly because I did not know you would be reading it, so sorry about that! First, let me give you some clarity around your question.

So Tabs component does have two buttons, but just to scroll left and to scroll right depending on the VP, the child of the scroll buttons are an unordered list. The unordered list contains function components: These function components decide to render an anchor or button element _solely based on the href prop that they receive_. If they are passed the href prop, they will render an anchor, if not it will be a button.

`const Component = (props.href ? 'a' : 'button') as any;`

So to address your question on whether or not you can declarative use a an anchor element instead of the button element the answer is no, not declaratively , but you can make the Tab component render the anchor tab if you feed is the href prop.

Now, i build a component that utilized href as a prop, and added a hash to the end of the url and it worked flawlessly, I wanted to add that also to the demo, exactly as it was in the html example, but the Markdown file that we are using show the examples did not play nice with the href prop. It ruined the demo in the page by freezing component state as soon as the hash was added to the end of the URL.

So, all in all, you can totally make Tab render as an anchor if you just add the prop href.

`
<Tab eventKey={2} title="Tab item 3" href="#forexample">
          Tab 3 section
</Tab>`

jgiardino:
Thanks for the explanation, @cmwylie19 😄 You actually articulate a really great issue, that I'm really stumped on how to move forward. 

Ideally, our examples would illustrate for consumers how to use components. When I look at the react examples for Tabs, there's no example or information on how to render the tabs as links. And I question if consumers would realize that putting `href` on a `<Tab>` would work without seeing an example or having some documentation to explain that? 

I'm open to suggestions on how to move forward on this.

dlabrecq:
FYI, we spoke during today's PatternFly standup and the suggestion was to try changing the href in the example to point at patternfly.org instead of using a hash tag -- hope that works for everyone

dlabrecq:
Check out the text and tab examples. 

The text example uses a hash tag:
```
<Text component={TextVariants.a} href="#">
  non ante
</Text>
```

While the tabs example uses:
```
      <Tab id="tab3" eventKey={2} title="Tab item 3" href="#/items/3">
        Tab 3 section
      </Tab>
```

There are more examples with breadcrumbs, link button, alert, etc., which also use hash tags.

mcoker:
The examples still don’t match core - first example uses `<button>`, not `<a>`, and the second example uses`<a>`, but the tabs link to google (instead of opening tab sections), and there is no secondary tabs in the second example like there is in core.

PR #3102, opened 2019-10-09
feat(aboutmodal,wizard): append component to any element in DOM

### LINKED ISSUE #2917: cannot render Wizard and AboutModal to specified element
Similar to #2551, an appendTo() props could be added to Wizard and AboutModal like what @jschuler did in #2802 so that these components could be rendered in specified element on the DOM tree.

boaz0:
**What**:

closes #2917 

I added `appendTo` to both Wizard and AboutModal.

//ping @jschuler can you have a look?

### DIFF
--- a/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
@@ -27,6 +27,8 @@ export interface AboutModalProps {
   backgroundImageSrc?: string;
   /** Prevents the about modal from rendering content inside a container; allows for more flexible layouts  */
   noAboutModalBoxContentContainer?: boolean;
+  /** The parent container to append the modal to. Defaults to document.body */
+  appendTo?: HTMLElement | (() => HTMLElement);
 }
 
 interface ModalState {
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/AboutModal/AboutModal.tsx
```diff
@@ -76,34 +79,45 @@ export class AboutModal extends React.Component<AboutModalProps, ModalState> {
...
+  };
+
   componentDidMount() {
     const container = document.createElement('div');
+    const target: HTMLElement = this.getElement(this.props.appendTo);
     this.setState({ container });
-    document.body.appendChild(container);
-    document.addEventListener('keydown', this.handleEscKeyClick, false);
+    target.appendChild(container);
+    target.addEventListener('keydown', this.handleEscKeyClick, false);
```
dlabrecq: Considering the default value for `appendTo` is null, it appears that `target.appendChild` could generate a null exception.

boaz0: I can add a check before calling these making sure it's not null.
But when target is `null` it means `window.document` is not defined which is odd.

dlabrecq: I saw the check for `document !== 'undefined'`, so assumed there's a condition where it could be null. Maybe just use `appendTo: document.body`?

boaz0:
@dlabrecq - I see the pf4 docs build fails:

> WebpackError: ReferenceError: document is not defined

This is because in [Gatsby document will be undefined on build](https://www.gatsbyjs.org/docs/debugging-html-builds/)

To fix this, I set the `appendTo` default value to `null` and in `getElement` I will return `document.body` if it's null. Is that OK?

PR #3107, opened 2019-10-10
fix(Dropdown/Application Launcher/Options Menu): Fix unit tests and options menu

### LINKED ISSUE #3029: Disappearing-selection bug in OptionsMenu caused by event handler change in InternalDropdownItem
In the OptionsMenu examples at https://patternfly-react.surge.sh/patternfly-4/components/optionsmenu/, changing the selection in the "single option" menu fails and causes the selection to disappear entirely.

yuqLFtEQDl

This is a bug that @nicolethoen noticed when working on her PR for chip groups in DataToolbar: https://github.com/patternfly/patternfly-react/pull/2969. It was causing errors in her usage of OptionsMenu for that chip groups example, and she initially fixed it with the help of @jschuler in this commit: https://github.com/patternfly/patternfly-react/pull/2969/commits/5544f2822bd7697e32bd6dbbc36b9fe3b7d1bde2

That fix caused unit test failures, and it turned out that according to @mcarrano we should have used Select instead of OptionsMenu for that PR's examples, so the fix was reverted and the issue was worked around by instead removing OptionsMenu from that PR. As a result, this bug still exists in master.

@jschuler found that it was caused by changes to the event handlers in the InternalDropdownItem component made in his recent PR https://github.com/patternfly/patternfly-react/pull/3011. I verified this by testing [the surge deployment from the PR merged immediately before that one](https://patternfly-react-pr-2981.surge.sh/patternfly-4/components/optionsmenu/) (https://github.com/patternfly/patternfly-react/pull/2981), where the bug is not present, and [the surge deployment from 3011](https://patternfly-react-pr-3011.surge.sh/patternfly-4/components/optionsmenu/), where it is.

To fix this we can probably cherry-pick [Nicole's commit](https://github.com/patternfly/patternfly-react/pull/2969/commits/5544f2822bd7697e32bd6dbbc36b9fe3b7d1bde2) and then work out the unit test failures.

rebeccaalpert:
Applied Nicole's fix for the Options menu.

Verified that keyhandler is working as expected. Added `componentID` prop in order to fix failing unit tests on keyhandler due to the id being passed to the `<li>`.

Fixes #3029.

### DIFF
--- a/packages/patternfly-4/react-core/src/helpers/util.test.js
+++ b/packages/patternfly-4/react-core/src/helpers/util.test.js
@@ -93,13 +93,13 @@ test('sideElementIsOutOfView Returns NONE when in view', () => {
 describe('keyHandler works on ApplicationLauncher', () => {
   document.body.innerHTML = '<!doctype html><html><body></body></html>';
   const dropdownItems = [
-    <DropdownItem key=" 1" id="first" component="button">
+    <DropdownItem key=" 1" id="first" componentID="first-button" component="button">
       Action 1
     </DropdownItem>,
-    <DropdownItem key="2" id="second" component="button">
+    <DropdownItem key="2" id="second" componentID="second-button" component="button">
       Action 2
     </DropdownItem>,
-    <DropdownItem key="3" id="third" component="button" isDisabled>
+    <DropdownItem key="3" id="third" componentID="third-button" component="button" isDisabled>
       Disabled Link
     </DropdownItem>
   ];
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/helpers/util.test.js
```diff
@@ -93,13 +93,13 @@ test('sideElementIsOutOfView Returns NONE when in view', () => {
...
-    <DropdownItem key=" 1" id="first" component="button">
+    <DropdownItem key=" 1" id="first" componentID="first-button" component="button">
       Action 1
     </DropdownItem>,
-    <DropdownItem key="2" id="second" component="button">
+    <DropdownItem key="2" id="second" componentID="second-button" component="button">
       Action 2
     </DropdownItem>,
-    <DropdownItem key="3" id="third" component="button" isDisabled>
+    <DropdownItem key="3" id="third" componentID="third-button" component="button" isDisabled>
```
dlabrecq: Can the dropdown provide a default ID here? 

It seems odd that users must provide an 'optional' property to have the dropdown function properly.

rebeccaalpert: This is just for the unit test.

Issue #3117, opened 2019-10-11
Do not expect `onFilter` to return filtered values


karelhala:
We are trying to implement filterable select that filters data on server. We saw that we can use `onFilter` function for filtering. However it kept throwing errors so checked the codebase and saw that PF expects from this function to return actually filtered stuff. That is not possible to do, we don't know when the request on server will finish and we can't simply send async function. Plus this is not good way of consuming consumer defined callback functions this kinda looks like 2 way data binding and you are bypassing React's lifecycle.

PR #3123, opened 2019-10-11
fix(Select): use props children if custom on filter used

karelhala:
Fixes: #3117

**What**:
Do not expect filtered values to be returned from `onFilter` function.

Fix errors when opened Select menu by omiting them from spreaded props.

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
@@ -22,10 +22,10 @@ class User implements SelectOptionObject {
 }
 
 const selectOptions = [
-  <SelectOption value="Mr" key="0" />,
-  <SelectOption value="Mrs" key="1" />,
-  <SelectOption value="Ms" key="2" />,
-  <SelectOption value="Other" key="3" />
+  <SelectOption value="Mr" key="00" />,
+  <SelectOption value="Mrs" key="01" />,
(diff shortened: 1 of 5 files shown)


karelhala:
This is rather blocking issue for us, can we look at it with high priority? Thank you.

packages/patternfly-4/react-core/src/components/Select/Select.test.tsx
```diff
@@ -22,10 +22,10 @@ class User implements SelectOptionObject {
...
 
 const selectOptions = [
-  <SelectOption value="Mr" key="0" />,
-  <SelectOption value="Mrs" key="1" />,
-  <SelectOption value="Ms" key="2" />,
-  <SelectOption value="Other" key="3" />
+  <SelectOption value="Mr" key="00" />,
+  <SelectOption value="Mrs" key="01" />,
+  <SelectOption value="Ms" key="02" />,
+  <SelectOption value="Other" key="03" />
```
dlabrecq: Curious about the key values, is the change necessary?

karelhala: There was problem with tests complaining about same keys.

dlabrecq: ok, thanks for the explanation

PR #3126, opened 2019-10-11
WIP(table with empty state support)

### LINKED ISSUE #1433: PF4 Table: Provide empty state support

jenny-s51:
Will resolve #1433

### DIFF
--- a/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEmptyStateDemo.tsx
+++ b/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEmptyStateDemo.tsx
@@ -1,0 +1,82 @@
+import * as React from 'react';
+import {
+  Table,
+  TableHeader,
+  TableBody,
+  TableProps,
+  sortable,
+  SortByDirection,
+  headerCol,
+  TableVariant,
+  expandable,
+  cellWidth,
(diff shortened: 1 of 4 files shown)


jenny-s51:
Closing per @boaz0 's PR at #3133

Issue #923, opened 2018-11-15
Attribute value (text box) Filter Demo


LHinson:
This is a follow up to https://github.com/patternfly/patternfly-next/issues/2218

For intended design behaviors see the Marvel prototpye: https://marvelapp.com/1i69c6ig

mcarrano:
@LHinson @tlabaj Not sure if we still need this component. Do we have what we need to build filters today on React? I think there is still work on Filters pending, correct? I think we could make a TextBox filter from an input group.

Maybe we should just close this and open new issues if needed. Thoughts?

tlabaj:
closed by #3127

PR #3127, opened 2019-10-11
docs(Table Filter Demo)

kmcfaul:
**What**: Adds a demo of a filterable table. Adds a property to DataToolbarFilter (to allow filter chips to be displayed without the toolbar dropdown being displayed), and fixes a console warning in Select (create&no data text props were being passed to the DOM)

**Refer to issue**: #923


kmcfaul:
Added do not merge to wait for other PRs to go in first

mcarrano:
@kmcfaul I took a quick peek at this realizing that it still may be WIP. I think it's looking really good. Here are just a few small things I noticed:

* Styling should be consistent with core styling here https://patternfly-next-pr-2287.surge.sh/documentation/core/demos/datatoolbar Note difference in the search field specifically.

* For the search field, typing and pressing enter should have the same effect as clicking the search button.

* When the filter leads to 0 results, the empty state pattern should be used as seen in this demo: https://www.patternfly.org/v4/documentation/core/demos/table-full?component=Table%20-%20empty%20state%20demo

* Is it possible to fix the width of the input group so that you don't get an accordion affect as you select different items? I'm concerned to in the context of a toolbar, items following the filter will move around which is not a great experience. It can be up to the designer to specify a width that will work for the likely content.

All other behavior is working exactly as I expected. Great work!

kmcfaul:
@mcarrano Added updates based on your feedback, except the empty state. I cannot get to the full page demo via the link - it returns a 404 page. Is there another place I can look for this? I don't think we have an empty state demo for the table in react yet.

mcarrano:
Looks like there is a website issue preventing you from seeing the empty state demo @kmcfaul .

mcarrano:
Actually, here is a screen shot of what the empty state should look like @kmcfaul:

Screen Shot 2019-10-22 at 4 57 05 PM

You should be able to access this under core demos/table  Look for the Table-empty state demo

nicolethoen:
Is this demo supposed to be in an experimental category?

kmcfaul:
@mcarrano Added the empty state, thanks for the screenshot & location!

@nicolethoen @rachael-phillips We don't have a demos section in Experimental, but maybe it should be added?

nicolethoen:
I think we probably don't want to encourage people to implement the toolbar as seen in the demo if breaking changes could still be in the component's future. So the demo should be specifically categorized or consumers sufficiently warned... I thought that it was agreed that the demo would be in an experimental section or held off until the component is not longer experimental

When the other PR gets merged, you'll want to add a 'collapseListedFiltersBreakpoint' to the DataToolbar component to make it more responsive, just a heads up.

kmcfaul:
Yeah I agree, I'll try creating a new folder for demos in experimental and see if that gets picked up!

Thanks for the heads up about the PR changes as well.

packages/patternfly-4/react-core/src/demos/FilterTableDemo.md
```diff
@@ -0,0 +1,388 @@
...
+      });
+      this.onFilterSelect();
+    };
+  }
+
+  buildCategoryDropdown() {
+    const { isCategoryDropdownOpen, currentCategory } = this.state;
+
+    return (
+      <Dropdown
```
nicolethoen: This dropdown should also be in a DataToolbarItem or some other DataToolbar wrapper component

packages/patternfly-4/react-core/src/demos/FilterTableDemo.md
```diff
@@ -0,0 +1,388 @@
...
+    return (
+      <DataToolbar
+        id="data-toolbar-with-chip-groups"
+        clearAllFilters={this.onDelete}
+        showClearFiltersButton={
+          filters.location.length !== 0 || filters.name.length !== 0 || filters.status.length !== 0
+        }
+      >
+        <DataToolbarContent>
+          <DataToolbarGroup style={{ width: '50%' }}>
```
nicolethoen: If we need to make this toolbar responsive, you should only need to use a DataToolbarToggleGroup rather than a DataToolbarGroup here. You should at least try it (whether or not it ends up in the demo) to make sure that your multi-drop-down style filter still works in a responsive toggle group.

packages/patternfly-4/react-core/src/demos/FilterTableDemo.md
nicolethoen: It probably should be in the same DataToolbarGroup variant='filter-group' as the other dropdown.

packages/patternfly-4/react-core/src/demos/FilterTableDemo.md
nicolethoen: feel free to reach out to me if it doesn't. I think the changes you made to the DataToolbarFilter component are good. So it'll just come down to how you've implemented it in the demo to get it to work.

kmcfaul:
@mcarrano @nicolethoen 
Adding in the toggle creates some issues.

- The biggest issue is that the chip groups aren't located properly in the collapsed state's expanded menu for this use case. They are located after their respective dropdown, and contain spacing even when no chips are present. With 2 of the 3 dropdowns hidden, there are varying white space gaps above and below each filter dropdown. I would expect all chip groups to be inline at the end of all dropdowns.
- The inline styling to prevent the dropdowns from resizing also limit them in the collapsed state menu when its expanded. Should I remove these limits?
- It always appears collapsed despite setting a breakpoint. Might be user error on my part, I'm fiddling with it still.

kmcfaul:
Talked with Nicole and rebased with the data toolbar changes. The responsive collapse and expansion is working now. 

I removed the width styling for now, and scheduled a meeting to go over the chip group locations in the collapsed menu.

kmcfaul:
As this demo uses an experimental component, we are moving forward with this PR and will increment on it in the next release re: the chip group stacking in the collapsed menu.

nicolethoen:
And people are okay with an experimental component under Demos?

kmcfaul:
I believe that was what was decided after a recent standup. @tlabaj Can you confirm?

christiemolloy:
I wonder if this demo should be under Toolbar demo because in Core it is, and the reason for the demo was to show how the new toolbar works when filtering the table. If it sits under Table demo, I don't know what functionality of the table its demonstrating? Seems like its just demonstrating the filtering. Would be great to hear what @mcarrano thinks

mcarrano:
This is tricky @christiemolloy . I discussed this with @kmcfaul and @nicolethoen in a meeting earlier today. Naming of demos is something that needs more thought in general, IMO. Should they be based on the thing being demonstrated (i.e. Toolbar) or the use case being demonstrated (i.e. Filtering). I don't know, but the naming will effect how easy it is for developers to find these. Not sure how the website search indexes these (or does it). Let's leave alone for now, but @rachael-phillips @LHinson I think this is a topic that should be on the radar for more discussion.

PR #3133, opened 2019-10-13
fix(table): set height auto modifier on table row

### LINKED ISSUE #3132: [PF4] [react-table] auto modifier is set on `td` instead of `tr`
@jenny-s51 messaged me about a mistake on my side regarding where the  `pf-m-height-auto` class is set.

Currently we are passing `cellHeightAuto` to `transforms` of a cell. This adds to classname the `pf-m-height-auto` class. The problem is: this is not where this modifier should be set and that's why she couldn't get it work.

To fix that, the modifier should be set in the row, `tr` element.

//cc @mcoker @karelhala

boaz0:
**What**:

close #3132

I mistakenly added the `height auto` modifier to `td` element instead to `tr`.

This PR fixes this by adding the `HeightAuto` attribute to row item.
In addition to that, added a deprecation warning message to whom is using `cellHeightAuto`.

More tests and an example were added.

//cc @jenny-s51 @mcoker @karelhala @tlabaj @redallen

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.md
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.md
@@ -26,10 +26,16 @@ import {
 } from '@patternfly/react-table';
 
 import {
-    Checkbox
+    Checkbox,
+    Button,
+    EmptyState,
+    EmptyStateBody,
+    EmptyStatePrimary,
+    Bullseye,
 } from '@patternfly/react-core';
 
 import {
+  SearchIcon,
   CodeBranchIcon,
   CodeIcon,
   CubeIcon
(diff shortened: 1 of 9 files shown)


jenny-s51:
will also close #1433

packages/patternfly-4/react-table/src/components/Table/utils/decorators/cellHeightAuto.ts
```diff
@@ -1,6 +1,9 @@
 import { css, getModifier } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Table/table';
 
-export const cellHeightAuto = () => ({
-  className: css(getModifier(styles, 'heightAuto'))
-});
+export const cellHeightAuto = () => {
+  console.error('cellHeightAuto:', 'is deprecated. Use heightAuto instead. Read more here: https://github.com/patternfly/patternfly-react/issues/3132')
```
karelhala: This should be `warn` instead of `error`.

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -1110,3 +1116,43 @@ class WrappableHeadersTable extends React.Component {
...
+
+```js
+import React from 'react';
+import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import { Button, EmptyState, EmptyStateBody, EmptyStatePrimary, Bullseye } from '@patternfly/react-core';
+
+EmptyStateTable = () => {
+  const columns = ['Repositories', 'Branches', 'Pull request', 'Workspaces', 'LastCommit']
+  const rows = [{
+    heightAuto: true,
```
karelhala: @priley86 what do you think about introducing row decorators? Similiar to how cell formatters work. This can be potentially problem in future when we'll update API and we will have to keep it intact.

Issue #3111, opened 2019-10-10
Console error when using radio buttons in build 3.113.4


dlabaj:
I'm seeing the following error when using radio buttons with version 3.113.4 of react-core.

<img width="1690" alt="Screen Shot 2019-10-10 at 2 41 17 PM" src="https://user-images.githubusercontent.com/6126356/66600785-52d5fc80-eb74-11e9-81a7-1ac97b36bea1.png">

Here' is the code in integr8tly causing the issue:

<React.Fragment>
              <Form>
                <FormGroup controlId="radio" disabled={false} bsSize="small" label="Check your work">
                <FormGroup fieldId="radio" disabled={false} label="Check your work">
                  <Radio
                    id={`${blockId}verificationYes`}
                    name={`${blockId}Yes`}
                    checked={isYesChecked}
                    className="integr8ly-task-radio-btn"
                    isChecked={isYesChecked}
                    onChange={e => {
                      this.handleVerificationInput(e, blockId, true);
                    }}
                    label="Yes"
                  >
                    Yes
                  </Radio>
                  />
                  <Radio
                    id={`${blockId}verificationNo`}
                    name={`${blockId}No`}
                    checked={isNoChecked}
                    className="integr8ly-task-radio-btn"
                    isChecked={isNoChecked}
                    onChange={e => {
                      this.handleVerificationInput(e, blockId, false);
                    }}
                    label="No"
                  >
                    No
                  </Radio>
                  />

...

tlabaj:
Closed by #3138

PR #3138, opened 2019-10-14
fix(Radio): Fix console error

tlabaj:
#3111

**What**:

**Additional issues**:


dlabaj:
@tlabaj Verified in integr8tly and the error is gone.  Thanks.

Issue #3097, opened 2019-10-09
DataToolbar: implement structural and interaction changes to match PF-Core


nicolethoen:
- .pf-c-data-toolbar__expandable-content mov into the .pf-c-data-toolbar__content to which it relates

- add .pf-m-toggle-group-summarymodifier to address mcarrano’s design update

tlabaj:
closed by #3144

Issue #3046, opened 2019-09-30
Add support for expanded state on data toolbar's toggle button


mcoker:
core issue - https://github.com/patternfly/patternfly-next/issues/2303

tlabaj:
closed by #3144

PR #2342, opened 2019-06-24
feat(Notification Badge): Adds the notification badge component

dlabaj:
**What**:
Adds the notification badge component to react core.  

**Additional issues**:
#2021

PR #3144, opened 2019-10-15
fix(DataToolbar): implement changes to match PF-Core

nicolethoen:
Addresses [issue #3097](https://github.com/patternfly/patternfly-react/issues/3097)
Addresses https://github.com/patternfly/patternfly-react/issues/3046

Implements changes made to core in this PR [#2342](https://github.com/patternfly/patternfly-next/pull/2342)

### DIFF
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
@@ -82,7 +82,7 @@ export class DataToolbarChipGroupContent extends React.Component<DataToolbarChip
         {
           showClearFiltersButton && !isExpanded &&
           <DataToolbarItem className={css(getModifier(styles, 'clear'))}>
-            <Button variant="link" onClick={clearChipGroups}>
+            <Button variant="link" onClick={clearChipGroups} isInline>
               {clearFiltersButtonText}
             </Button>
           </DataToolbarItem>
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
@@ -11,6 +11,8 @@ import { globalBreakpoints } from './DataToolbarUtils';
 export interface DataToolbarChipGroupContentProps extends React.HTMLProps<HTMLDivElement> {
   /** Classes applied to root element of the data toolbar content row */
   className?: string;
+  /** Flag indicating if a data toolbar toggle group's expandable content is expanded */
+  isExpanded?: boolean;
   /** Chip group content reference for passing to data toolbar children */
   chipGroupContentRef: RefObject<HTMLDivElement>;
   /** optional callback for clearing all filters in the toolbar */
@@ -34,6 +36,7 @@ export class DataToolbarChipGroupContent extends React.Component<DataToolbarChip
   render() {
     const {
       className,
+      isExpanded,
       chipGroupContentRef,
       clearAllFilters,
       showClearFiltersButton,
@@ -70,14 +73,14 @@ export class DataToolbarChipGroupContent extends React.Component<DataToolbarChip
           {...(collapseListedFilters && { 'aria-hidden': true })}
         />
         {
-          collapseListedFilters && numberOfFilters > 0 && (
+          collapseListedFilters && numberOfFilters > 0 && !isExpanded && (
             <DataToolbarGroup className={css(getModifier(styles, 'toggle-group-summary'))}>
               <DataToolbarItem>{numberOfFilters} filters applied</DataToolbarItem>
             </DataToolbarGroup>
           )
         }
         {
-          showClearFiltersButton &&
+          showClearFiltersButton && !isExpanded &&
           <DataToolbarItem className={css(getModifier(styles, 'clear'))}>
             <Button variant="link" onClick={clearChipGroups}>
               {clearFiltersButtonText}
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
@@ -59,10 +59,10 @@ export class DataToolbarChipGroupContent extends React.Component<DataToolbarChip
         className={css(
           styles.dataToolbarContent,
           getModifier(styles, 'chip-container'),
-          getModifier(styles, 'hidden'),
+          numberOfFilters === 0 &&  getModifier(styles, 'hidden'),
           className
         )}
-        hidden
+        {...(numberOfFilters === 0 && {hidden: true})}
         ref={chipGroupContentRef}
         {...props}
       >
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
@@ -9,15 +9,15 @@ import { DataToolbarGroup } from './DataToolbarGroup';
 import { globalBreakpoints } from './DataToolbarUtils';
 
 export interface DataToolbarChipGroupContentProps extends React.HTMLProps<HTMLDivElement> {
-  /** Classes applied to root element of the Data toolbar content row */
+  /** Classes applied to root element of the data toolbar content row */
   className?: string;
-  /** Chip group content reference for passing to Data toolbar children */
+  /** Chip group content reference for passing to data toolbar children */
   chipGroupContentRef: RefObject<HTMLDivElement>;
   /** optional callback for clearing all filters in the toolbar */
   clearAllFilters?: () => void;
-  /** Flag indicating that the Clear all filters button should be visible */
+  /** Flag indicating that the clear all filters button should be visible */
   showClearFiltersButton: boolean;
-  /** Text to display in the Clear all filters button */
+  /** Text to display in the clear all filters button */
   clearFiltersButtonText?: string;
   /** Total number of filters currently being applied across all DataToolbarFilter components */
   numberOfFilters: number;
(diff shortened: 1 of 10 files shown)


packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
```diff
@@ -12,17 +13,107 @@ export interface DataToolbarContentProps extends React.HTMLProps<HTMLDivElement>
...
-    <div className={css(styles.dataToolbarContent, formatBreakpointMods(breakpointMods, styles), className)} {...props}>
-      {children}
-    </div>
-  );
-};
+export interface DataToolbarContentState {
+  /** Flag indicating the if the expandable content's expanded state is consumer managed or not */
+  isConsumerManagedToggleGroup: boolean;
+  /** Flag indicating if the component managed state has expanded content or not */
+  componentManagedIsExpanded: boolean;
```
dlabrecq: The `isConsumerManagedToggleGroup` and `componentManagedIsExpanded` names are a little confusing to me. 

Perhaps something more like `isToggleManaged` would suffice?

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
```diff
@@ -12,17 +13,107 @@ export interface DataToolbarContentProps extends React.HTMLProps<HTMLDivElement>
...
+      <div
+        className={css(styles.dataToolbarContent, formatBreakpointMods(breakpointMods, styles), className)}
+        {...props}
+      >
+        <DataToolbarContentContext.Provider
+          value={{
+            isExpanded: isConsumerManagedToggleGroup ? isExpanded : componentManagedIsExpanded,
+            toggleIsExpanded: isConsumerManagedToggleGroup ? toggleIsExpanded : this.toggleIsExpanded,
+            expandableContentRef: this.expandableContentRef,
+            expandableContentId
```
dlabrecq: It's difficult to understand the difference between isExpanded, toggleIsExpanded, expandableContent*, etc. It would be good to clarify these names.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
```diff
@@ -12,17 +13,107 @@ export interface DataToolbarContentProps extends React.HTMLProps<HTMLDivElement>
   breakpointMods?: DataToolbarBreakpointMod[];
   /** Content to be rendered as children of the content row */
   children?: React.ReactNode;
+  /** Flag indicating if a Data toolbar toggle group's expandable content is expanded */
+  isExpanded?: boolean;
```
dlabrecq: Is this for the toggle group or content? The description, "toggle group's expandable content is expanded" sounds confusing to me.

Considering there is a `toggleIsExpanded` prop below, perhaps something more like `isContentExpanded` would help to clarify the use case?

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
dlabrecq: I'm struggling with the `componentManagedIsExpanded` name. Couldn't a user use `isExpanded` directly, when `isConsumerManagedToggleGroup` is in use? 

Could this be simplified at all?

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
nicolethoen: so the toggle group has an associated expandable content that is what gets expanded or collapsed when the toggle icon in the toggle group is clicked. It is confusing. I feel passing an `isExpanded` flag as a prop to the `DataToolbarToggleGroup` is the clearest way to state it. Do you have another idea?

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
nicolethoen: I will change the names of the `isConsumerManagedToggleGroup and componentManagedIsExpanded` to try and clear it up some. 

Basically, I allow the consumer to control the expanded state of the DataToolbarContentGroup by passing `isExpanded` and `toggleIsExpanded` as props. If they are not provided by the consumer, then both are managed internally by the component. That was what I was trying to communicate. I use the state to track whether or not the user has provided those props - thus wants to manage the expanded state of the toggle group themself. 

In the react docs I do try to provide a write up explaining some of that. Maybe that clears it up some? I could add more of that into the comments of the component code for the sake of the developers in the future.

nicolethoen:
@mcarrano 
> * It still looks like we are not applying the latest styling from core. I'm specifically looking at the search fields which should pick up the revised input group styling.

Good catch, I had to make a code change to apply the new style for inputGroup. I guess that was a change made after I first used the InputGroups in the examples.

> * The way the toolbar responds as I shrink the viewport in the examples is not consistent with the design intent. See the section on Responsive toolbar here: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/toolbar. Maybe this is OK for these examples and we should build something more like a Toolbar demo that responds correctly. Not sure. Perhaps we didn't address this fully in core. @mattnolting what are your thoughts? Should we be using the Overflow component in the place of just groups of buttons?

If you are referring to the groups of buttons in the first few examples. Those examples are not demonstrating the responsive behavior of the toolbar, they are demonstrating the spacing between items and groups of items and how that can be manipulated. 

As far as I understand, the responsive solution in the toolbar is heavily tied to the Toggle Group, so that is not used or demonstrated until later examples. Is there something else that is missing?

> * Without more documentation, I'm not sure that the examples here have enough context. Do we need to explain the spacing system? I tried to do that in the design documentation (linked above) so maybe we just need to reference that?

I'm happy to add more text - when doing my own poking around, it didn't seem that the react examples tended to include a grand deal of explanation. So I assumed that we expected consumers to look elsewhere for design decision explanation and such - but I can definitely see how more text is useful to a developer. If no one sees a problem with adding more information. I'll do that.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
nicolethoen: `isConsumerManagedToggleGroup` is a state field calculated based on whether or not the consumer provided `isExpanded` and the `toggleIsExpanded` handler, not a passed prop. But I will rename it to `isToggleManaged` and make the comments clearer?

dlabrecq: gotcha, ok

mcarrano:
> If you are referring to the groups of buttons in the first few examples. Those examples are not demonstrating the responsive behavior of the toolbar, they are demonstrating the spacing between items and groups of items and how that can be manipulated.

Upon further thinking, I guess this is OK. Maybe this would be solved if there was some documentation that describes the purpose of those examples. I just didn't want someone to look at this and think it was the responsive behavior that we want (with everything wrapping to multi-lines).

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
```diff
@@ -16,40 +17,73 @@ export interface DataToolbarChipGroupContentProps extends React.HTMLProps<HTMLDi
   clearAllFilters?: () => void;
   /** Flag indicating that the Clear all filters button should be visible */
   showClearFiltersButton: boolean;
-  /** Flag indicating if a Data toolbar toggle group's expandable content is expanded */
-  expandableContentIsExpanded: boolean;
+  /** Text to display in the Clear all filters button */
```
tlabaj: can you make this sentence case.  That goes for all prop descriptions.

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarChipGroupContent.tsx
```diff
@@ -16,40 +17,73 @@ export interface DataToolbarChipGroupContentProps extends React.HTMLProps<HTMLDi
   clearAllFilters?: () => void;
   /** Flag indicating that the Clear all filters button should be visible */
```
tlabaj: can you make this sentence case.

nicolethoen:
@mattnolting 
does this incorporate changes necessary to consider this issue addressed?
https://github.com/patternfly/patternfly-react/issues/3046

mattnolting:
@nicolethoen yes, `.pf-m-expanded` is applied to `pf-c-data-toolbar__toggle` when expanded

nicolethoen:
> Toolbar height shrinks when toggle is opened.
> 
> Screen Shot 2019-10-21 at 4 22 09 PM
> 
> Screen Shot 2019-10-21 at 4 22 15 PM
> 
> Can you add `pf-m-toggle-group-container` to `pf-c-data-toolbar__content`s that contain `__toggle-group`s?

It shrinks because the chips and clear filters button are hidden when the toggle is open. 
I've tried adding that class to the content that contain a toggle group but its not super straight forward.... I can try again...

packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbarContent.tsx
mturley: Storing `isToggleManaged` in state seems like an antipattern to me, since it is something derived from props. If those props were to change, the state wouldn't update. Since the component will probably never go from managed to unmanaged or vice versa while it is mounted (I don't see why a consumer would do that) it's probably fine here, but I could see this pattern causing bugs if used elsewhere. Instead maybe putting that logic in a method and calling `this.isToggleManaged()` wherever you need to check that would be better.

nicolethoen: oh I see

mattnolting:
> It shrinks because the chips and clear filters button are hidden when the toggle is open.
I've tried adding that class to the content that contain a toggle group but its not super straight forward.... I can try again...

@nicolethoen yeah, idk what to do about that. We'll see how it behaves in the context of a `table`/`data-list`. Just observing..

nicolethoen:
Gah ... still an issue in the mobile view... wait for one more update

PR #3149, opened 2019-10-16
fix(OUIA): Use consumer context if specified

karelhala:
**What**:
When rendering Table the ouia context is not correctly set up. This can be fixed by using context passed from props because table components copies these context values to different context and pass them along with props.


karelhala:
@quarckster ping

PR #3158, opened 2019-10-17
fix(Select): set onFilter to null to run the default filter

### LINKED ISSUE #3167: Select: Typeahead (multi/single) doesn't filter options
The problem is the follwoing:
When typing in the input box the options that are relative to what I'm entering are not shown.

For example, go to https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/select and scroll down to the Typeahead example. Then click on the input box, you will see option menu. Type "A" but the options "New Jersey", "New Mexico", "New York" and etc are still shown even though they don't have A in them.

//cc @dlabaj @dlabrecq

boaz0:
**What**:

fixes #3167 

Typeahead doesn't filter options correctly. For example, in the `Typeahead select input` example, clicking "Ala" will display all the results even though they don't contain this substring.

The problem is that in the `defaultProps` object the `onFilter` attribute is set to a noop function which makes the `if(onFilter)` block to execute and eventually sets options to the children.

This PR fixes it by setting `onFilter` to `null` which makes the default filter function in `onChange` to be executed.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.tsx
@@ -104,7 +104,7 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
     isPlain: false,
     isDisabled: false,
     isCreatable: false,
-    "aria-label": '',
+    'aria-label': '',
     ariaLabelledBy: '',
     ariaLabelTypeAhead: '',
     ariaLabelClear: 'Clear all',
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -183,6 +183,7 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
             )
           : React.Children.toArray(this.props.children);
     }
+    console.log('ffffffffffffffff', typeaheadFilteredChildren)
```
dlabrecq: Please remove

boaz0: Thanks!

boaz0:
@dlabrecq good catch. I think I fixed it here too - I just passed `this.onClick` to the input box and that worked

tlabaj:
@boaz . are you still having an issue?  If so, can you open an issue for this.  Thanks.

boaz0:
Thank you all :cat:

Issue #3121, opened 2019-10-11
Table - apply custom classes on Rows


rachael-phillips:
I could not find anywhere documented how to apply custom classes on Rows (in our use case we want to mark rows with a color that signals a warning if the resource on that row has some issue)

tlabaj:
closed by #3160

Issue #3118, opened 2019-10-11
Improvement from table documentation


rachael-phillips:
What is RowWrapper and BodyWrapper? the props description mentions only that it's a Function.

Issue #3018, opened 2019-09-26
PF4 Table: TableBody documentation not accurate


priley86:
I noted in conversation w/ @KKoukiou today that some props currently showing in the `TableBody` definition are calculated (and not typically passed by a consumer). We should hide those props from the documentation:

<img width="969" alt="Screen Shot 2019-09-26 at 12 53 32 PM" src="https://user-images.githubusercontent.com/4237045/65708523-c4825680-e05c-11e9-8f2b-406326ed0365.png">

Alongside this issue, it may be nice to add an example to our docs with a "custom row wrapper". Here is one such example. This could also be modified to have `CustomRowWrapper` just render `RowWrapper` passing className prop... 
https://github.com/priley86/patternfly-react/commit/a020e837ff7e1e2ce0e40b65847c2e3ef0e148c7

redallen:
Yes, I agree we should add a frontmatter or some special syntax in the interface definition (maybe `@hide`?) to hide props from the docs. I prefer the latter, but it'll take tweaking our `gatsby-transformer-react-docgen-typescript`.

PR #3160, opened 2019-10-17
Custom row wrapper example for table

seanforyou23:
**What**: This PR adds an example for how to customize a row wapper in table. Also registers the example with react-integration.

Screen Shot 2019-10-17 at 7 24 04 PM

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/3121 and https://github.com/patternfly/patternfly-react/issues/3018 and https://github.com/patternfly/patternfly-react/issues/3118

I built off the top of @priley86 's initial example of how custom classes per row can be achieved.

It also clears a few console errors for table examples that were introduced recently.

Screen Shot 2019-10-17 at 5 31 08 PM

Screen Shot 2019-10-17 at 5 31 17 PM


packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -118,6 +114,55 @@ class SimpleTable extends React.Component {
...
+
+```js
+import React from 'react';
+import {
+  Table,
+  TableHeader,
+  TableBody,
+} from '@patternfly/react-table';
+
+import { CustomRowWrapper } from './demo/CustomRowWrapper';
```
dlabrecq: As a user, I want to see the CustomRowWrapper code. Can we please include it in the SimpleTable class?

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableRowWrapperDemo.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
```
dlabrecq: I see the table example in Table.md, but don't see anything new under demos. Is this intended to be under demos, the table example, or both?

Demo navigation links:
Screen Shot 2019-10-17 at 11 12 58 PM

packages/patternfly-4/react-table/src/components/Table/Table.md
seanforyou23: Totally agree, but it doesn't seem like our docs allow more than one class to be present in a single example area

Screen Shot 2019-10-18 at 9 12 16 AM

I added a comment that will point to where the implementation lives (once it's merged) so users are not forced to hunt around for it. Think this is an ok stop-gap until we get support for multiple classes in a single example snippet?

Screen Shot 2019-10-18 at 9 17 00 AM

We could possibly introduce custom decorator support to help here, something like `@noparse`. TBH though I'd much prefer we go back to using proper javascript or typescript for these examples instead because these markdown files continue to require more custom code to make them act like normal script files again. WDYT? Can we go with the simple comment for now?

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableRowWrapperDemo.tsx
seanforyou23: Ah, I just realized I should have mentioned "example" instead of "demo" in the PR, wasn't intending for anything to show up here. I'll update the PR to reflect this

packages/patternfly-4/react-table/src/components/Table/Table.md
seanforyou23: Btw [the url in that comment](https://github.com/patternfly/patternfly-react/pull/3160/files#diff-f6c0274c6d946d8f7472804d7595cf78R129) will only work once this PR is merged. In the mean time you can view what it would contain on [my working branch](https://github.com/seanforyou23/patternfly-react/blob/row-wrapper-demo/packages/patternfly-4/react-table/src/components/Table/demo/CustomRowWrapper.js).

dlabrecq: Considering CustomRowWrapper.js does not contain any state, do you need a class? Could you add something similar (below) in your constructor?

```
this.customRowWrapper = ({
  trRef,
  className,
  rowProps,
  row: { isExpanded, isHeightAuto }
}) => {
  const isEvenOddRow = (rowProps.rowIndex + 1) % 2;
  const customStyle = {
    borderLeft: '3px solid var(--pf-global--primary-color--100)'
  };
  return (
    ...
  );
}
```

dlabrecq: I ran into the same situation (not external variables) with the chart legend examples. I use the same approach (above) to create my custom legend and label components.

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -924,7 +923,7 @@ class CompoundExpandableTable extends React.Component {
           parent: 0,
           compoundParent: 1,
           cells: [
-            { 
+            {
               title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-1', 'three', 'four','five']} id="compoound-expansion-table-1" />,
```
jenny-s51:
```suggestion
              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-1', 'three', 'four','five']} id="compound-expansion-table-1" />,
```

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -934,8 +933,8 @@ class CompoundExpandableTable extends React.Component {
           parent: 0,
           compoundParent: 2,
           cells: [
-            { 
-              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-2', 'three', 'four','five']} id="compoound-expansion-table-2" />, 
+            {
+              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-2', 'three', 'four','five']} id="compoound-expansion-table-2" />,
```
jenny-s51:
```suggestion
              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-2', 'three', 'four','five']} id="compound-expansion-table-2" />,
```

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -944,8 +943,8 @@ class CompoundExpandableTable extends React.Component {
           parent: 0,
           compoundParent: 3,
           cells: [
-            { 
-              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-3', 'three', 'four','five']} id="compoound-expansion-table-3" />, 
+            {
+              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-3', 'three', 'four','five']} id="compoound-expansion-table-3" />,
```
jenny-s51:
```suggestion
              title: <DemoSortableTable firstColumnRows={['parent-0', 'compound-3', 'three', 'four','five']} id="compound-expansion-table-3" />,
```

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -986,8 +985,8 @@ class CompoundExpandableTable extends React.Component {
           parent: 4,
           compoundParent: 1,
           cells: [
-            { 
-              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-1', 'three', 'four','five']} id="compoound-expansion-table-4" />, 
+            {
+              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-1', 'three', 'four','five']} id="compoound-expansion-table-4" />,
```
jenny-s51:
```suggestion
              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-1', 'three', 'four','five']} id="compound-expansion-table-4" />,
```

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -996,8 +995,8 @@ class CompoundExpandableTable extends React.Component {
           parent: 4,
           compoundParent: 2,
           cells: [
-            { 
-              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-2', 'three', 'four','five']} id="compoound-expansion-table-5"/>, 
+            {
+              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-2', 'three', 'four','five']} id="compoound-expansion-table-5"/>,
```
jenny-s51:
```suggestion
              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-2', 'three', 'four','five']} id="compound-expansion-table-5"/>,
```

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -1006,8 +1005,8 @@ class CompoundExpandableTable extends React.Component {
           parent: 4,
           compoundParent: 3,
           cells: [
-            { 
-              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-3', 'three', 'four','five']} id="compoound-expansion-table-6"/>, 
+            {
+              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-3', 'three', 'four','five']} id="compoound-expansion-table-6"/>,
```
jenny-s51:
```suggestion
              title: <DemoSortableTable firstColumnRows={['parent-4', 'compound-3', 'three', 'four','five']} id="compound-expansion-table-6"/>,
```

packages/patternfly-4/react-table/src/components/Table/demo/CustomRowWrapper.js
```diff
@@ -0,0 +1,37 @@
...
+  render() {
+    const {
+      trRef,
+      className,
+      rowProps,
+      row: { isExpanded, isHeightAuto },
+      ...props
+    } = this.props;
+
+    const isEvenOddRow = (rowProps.rowIndex + 1) % 2;
```
jenny-s51: Do you think this prop could possibly be renamed to `isOddRow`? I was confused by the name of it since in this example you're applying the styles to only the odd rows

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/RowWrapperForTableRowWrapperDemo.tsx
```diff
@@ -0,0 +1,38 @@
...
+  render() {
+    const {
+      trRef,
+      className,
+      rowProps,
+      row,
+      ...props
+    } = this.props;
+
+    const isEvenOddRow = rowProps && (rowProps.rowIndex) % 2;
```
jenny-s51: Looks like you're checking if it's odd https://github.com/patternfly/patternfly-react/pull/3160/files#diff-7b522b96301a408012fc2cc357c0f91fR27. Do you think `isEvenOddRow` could possibly be renamed to `isOddRow`? I was confused by the name of it since in this example you're applying the styles to only the odd rows.

jenny-s51: Also, if you're trying to check for odd rows, you do `(rowProps.rowIndex + 1) % 2` in https://github.com/patternfly/patternfly-react/pull/3160/files#diff-140e5f0b0acc2241453a7c0859691b15R15 so I'm guessing you want to do the same thing here too?

jenny-s51: and just out of curiosity, what does `rowProps && ...` do to the value of isEvenOddRow? It looks like `rowProps` is declared as an object in `RowWrapper.tsx` so I'm really curious about how the bool expression is evaluated here

seanforyou23: > and just out of curiosity, what does `rowProps && ...` do to the value of isEvenOddRow? It looks like `rowProps` is declared as an object in `RowWrapper.tsx` so I'm really curious about how the bool expression is evaluated here

I used logical AND here because [rowProps can be undefined](https://github.com/patternfly/patternfly-react/pull/3160/files#diff-89b3fdfa90b01f903ab08494efcf7b4dR19). Without this, copying the example code into a TypeScript based app will immediately throw errors. This happens because we're accessing the rowIndex prop, and in the case where rowProps are undefined, attempting to access `.rowIndex` would throw an error. So it is basically being used as a sort of type guard.

Does that make sense? Probably a lot we can do to improve this, let me know if you see a better way.

seanforyou23:
Thanks for the feedback! I've rebased with master and just waiting on the docs to finish building. I introduced one more [change](https://github.com/patternfly/patternfly-react/pull/3160/files#diff-9bda7132cc01104382345447864888d5R223) which I think will be helpful but needs to be tested further. Let me know what you think!

dlabrecq:
@jenny-s51 I see you requested changes. Do you want to review, again?

PR #3168, opened 2019-10-20
feat(textinput): support passing a ref

### LINKED ISSUE #3153: TextInput: expose ref of input element with forwardRef
The ref of TextInput's input element should be settable to allow for react hooks.

See: #2913

boaz0:
**What**:

closes #3153 

- I had to update the tests because of enzyme snapshot (TextInput -> ForwardRef)
- I added an example for react-hook-form. I am not sure we should have it.
- I added `withInnerRef` to the helpers directory. This will make it easier to add support to `ref` in other component.

//cc @kmcfaul @ia3andy

### DIFF
--- a/packages/patternfly-4/react-core/src/components/TextInput/examples/TextInput.md
+++ b/packages/patternfly-4/react-core/src/components/TextInput/examples/TextInput.md
@@ -6,7 +6,7 @@ propComponents: ['TextInput']
 typescript: true
 ---
 
-import { TextInput } from '@patternfly/react-core';
+import { TextInput, Button } from '@patternfly/react-core';
 
 ## Examples
 ```js title=Basic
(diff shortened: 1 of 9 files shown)


packages/patternfly-4/react-core/src/components/TextInput/examples/TextInput.md
```diff
@@ -79,3 +79,19 @@ class InvalidTextInput extends React.Component {
...
+import React from 'react';
+import { TextInput, Button } from '@patternfly/react-core';
+
+TextInputFocus = () => {
+  const [value, setValue] = React.useState('');
+  const ref = React.useRef(null);
+  return (
+    <>
+      <TextInput ref={ref} value={value} onChange={value => setValue(value)} aria-label="focused" />
+      <Button onClick={() => ref && ref.current && ref.current.focus()}>Focus</Button>
```
dlabrecq: Instead of a primary button, thinking we should follow the select example and use a checkbox?

https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/select

boaz0: Checkbox for what? ref and refless?

dlabrecq: The button is used to set focus, but I don't know that we should use primary buttons in our examples like this. 

We need to be careful about the use cases we demo; otherwise, folks may use this incorrectly. I don't believe that using a primary button here is a valid pattern.

I didn't go thru all the examples, but the "select" example uses checkboxes to show features. I'm suggesting we use that as precedence.

boaz0: Hi @dlabrecq - I updated the example to something much simpler. This one doesn't require checkbox/radio/button to demonstrate "ref".

boaz0:
Thanks a lot :smile: :confetti_ball:

PR #3169, opened 2019-10-20
fix(copytoclipboard): wrap pre inside the expand element

### LINKED ISSUE #3163: Bug in CopyToClipboard
I'm having issues implementing the `isCode` variant on the CopyToClipboard component in the MDC Product. It looks as though there are discrepancies between whats in Core and whats in React in terms of where the `<pre>` tag is wrapped. In Core it is wrapped inside of the expanded content container, and in React its wrapped on the outside.

Its also not picking up the styles that were set in Core to make sure code isn't bleeding out of its parent container:

Screen Shot 2019-10-18 at 2 40 51 PM

This is what it looks like in MDC when I add the `isCode` variant

Screen Shot 2019-10-18 at 2 45 30 PM

@tlabaj could we prioritize this bug as it is crucial for the MDC

boaz0:
**What**:

closes #3163 

//cc @christiemolloy @jessiehuff @jenny-s51


boaz0:
Thank you all!

PR #3182, opened 2019-10-22
feat(Pagination): automate pagination and overflow logic

### LINKED ISSUE #3136: Pagination and pagination overflow OOB logic
This is a feature request for our existing [Pagination](https://www.patternfly.org/v4/documentation/react/components/pagination/) component. Currently, Pagination seems to expose only some fundamental aspects of pagination such as tracking the current page, number of items per page, how many items are in the dataset, etc.

While this certainly helps prop a consumer up for implementing pagination on a collection, it doesn't offer two primary sets of logic that most users will need;

1) The logic for splitting and paginating the collection itself
2) Logic to handle the case of pagination overflow (when the user sets the number of items to display per page to an amount that exceeds the remaining amount of data)

For the second case here, I think a good experience from a user standpoint would be if the component walked the pages in reverse until a page index is found that will display the amount of data the user set, or page 1 if there are not enough items - and render the collection based on that page index.

I imagine there are other ways people may wish to paginate a dataset. I'm not requesting a generic feature that attempts to solve pagination for all the ways people may need to accomplish it, but rather just a basic "out of the box" method that would work for common scenarios with lists, tables, etc.

evwilkin:
**What**:
fixes #3136 

This feature allows the component to automatically paginate the users's data for them, and provides a "defaultToFullPage" flag that will automatically move the user back to the last full page of results when they change the number of rows to view per page.

**Additional issues**:


packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -57,7 +57,135 @@ class ComplexPaginationTableDemo extends React.Component {
...
+      { cells: [
+        { title: "Row 12 column 1" },
+        { title: "Row 12 column 2" },
+        { title: "Row 12 column 3" }
+      ]}
+    ];
+    this.state = {
+      perPage: 10,
+      page: 1,
+      rows: this.defaultRows.slice(0, 10)
```
seanforyou23: The `10` here corresponds to the value supplied for perPage, is there something we can do to make this more clear? Maybe add a default constant that's used for both of these places? If they are different, the default row count doesn't match the default specified perPage until the user interacts with one of the controls.

packages/patternfly-4/react-core/src/components/Pagination/PaginationOptionsMenu.tsx
```diff
@@ -89,16 +96,31 @@ export class PaginationOptionsMenu extends React.Component<PaginationOptionsMenu
     });
   };
 
+  handleNewPerPage = (_evt: React.MouseEvent | React.KeyboardEvent | MouseEvent, newPerPage: number) => {
+    const { page, onPerPageSelect, itemCount, defaultToFullPage } = this.props;
+    let newPage = page;
+    if (defaultToFullPage) {
+      if (itemCount / newPerPage !== newPage) {
+        while (newPage && (itemCount - (newPerPage * newPage)) < 0) {
```
seanforyou23: We might want to tweak this logic to handle the case where the user selects a number to display per page that's greater than the entire dataset. For example, I have 6 items, but specify 50 per page. In this case, I'd think to just display all the available rows we do have.



Otherwise, this seems to work great, nice work!

packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -57,7 +57,135 @@ class ComplexPaginationTableDemo extends React.Component {
...
+
+  renderPagination(variant = 'top') {
+    const { page, perPage, total } = this.state;
+    return (
+      <Pagination
+        itemCount={this.defaultRows.length}
+        page={page}
+        perPage={perPage}
+        defaultToFullPage
+        onSetPage={(_evt, newPage, perPage, startIdx, endIdx) => this.setState(
```
seanforyou23: While nesting these function calls is nice a terse it kind of makes the example a little difficult to follow. Even though `onSetPage` and `onPerPageSelect` follow a similar path in the code, they represent very different user interactions and we could maybe benefit from being more explicit and using dedicated callbacks for each. Just a thought!

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -33,8 +33,10 @@ export interface NavigationProps extends React.HTMLProps<HTMLElement> {
   paginationTitle?: string;
   /** The number of the current page */
   page: React.ReactText;
+  /** Number of items per page. */
+  perPage?: number;
   /** Function called when page is changed */
-  onSetPage: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void;
+  onSetPage: (_evt: React.MouseEvent | React.KeyboardEvent | MouseEvent, newPage: number, perPage?: number, startIdx?: number, endIdx?: number) => void;
```
seanforyou23: Can we define and share a type alias for this, since it's used in multiple places?

packages/patternfly-4/react-core/src/components/Pagination/Pagination.tsx
```diff
@@ -98,7 +100,7 @@ export interface PaginationProps extends React.HTMLProps<HTMLDivElement> {
   /** Function called when user inputs page number. */
   onPageInput?: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void;
   /** Function called when user selects number of items per page. */
-  onPerPageSelect?: (event: React.MouseEvent | React.KeyboardEvent | MouseEvent, perPage: number) => void;
+  onPerPageSelect?: (_evt: React.MouseEvent | React.KeyboardEvent | MouseEvent, newPerPage: number, newPage?: number, startIdx?: number, endIdx?: number) => void;
```
seanforyou23: Can we define and share a type alias for this also, since it's used in multiple places?

packages/patternfly-4/react-core/src/components/Pagination/Pagination.tsx
```diff
@@ -52,6 +52,22 @@ export interface PaginationTitles {
...
+  newPage: number,
+  perPage?: number,
+  startIdx?: number,
+  endIdx?: number
+) => void;
+
+export type OnPerPageSelect = (
+  _evt: React.MouseEvent | React.KeyboardEvent | MouseEvent,
+  newPerPage: number,
+  newPage?: number,
```
seanforyou23: I was wondering why `newPage` needs to be optional? It's required in `OnSetPage` - is it possible to also make it required here?

packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -66,3 +68,140 @@ class ComplexPaginationTableDemo extends React.Component {
...
+        itemCount={this.defaultRows.length}
+        page={page}
+        perPage={perPage}
+        defaultToFullPage
+        onSetPage={(_evt, newPage, perPage, startIdx, endIdx) => {
+          this.handleSetPage(newPage, startIdx, endIdx);
+        }}
+        onPerPageSelect={(_evt, newPerPage, newPage, startIdx, endIdx) => {
+          this.handlePerPageSelect(newPage, newPerPage, startIdx, endIdx);
+        }}
```
seanforyou23: I think we can simplify this demo by skipping the intermediate anonymous function here and just call `handlePerPageSelect` handler directly. Since handlePerPageSelect isn't used elsewhere there isn't as much value in breaking it out.

packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -66,3 +68,140 @@ class ComplexPaginationTableDemo extends React.Component {
...
+
+  renderPagination(variant = 'top') {
+    const { page, perPage, total } = this.state;
+    return (
+      <Pagination
+        itemCount={this.defaultRows.length}
+        page={page}
+        perPage={perPage}
+        defaultToFullPage
+        onSetPage={(_evt, newPage, perPage, startIdx, endIdx) => {
```
seanforyou23: Similar request here, can we simplify this by removing the arrow function and call `handleSetPage` directly? This makes it one less step a user needs to do to follow along.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -96,15 +99,21 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
     page: number | string,
     lastPage: number,
     onPageInput: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void,
-    onSetPage: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void
   ): void {
     if (event.keyCode === KEY_CODES.ENTER) {
       const inputPage = Navigation.parseInteger(this.state.userInputPage, lastPage) as number;
       onPageInput(event, Number.isNaN(inputPage) ? (page as number) : inputPage);
-      onSetPage(event, Number.isNaN(inputPage) ? (page as number) : inputPage);
+      this.handleNewPage(event, Number.isNaN(inputPage) ? (page as number) : inputPage);
```
seanforyou23: I wonder if there's a better way of handling that `page` here could be a string or a number. Could we instead cast it to a number in the place where it returns a string so we can avoid having to use `as number` type assertion in so many places? No worries if not, just curious in case it's an easy improvement.

packages/patternfly-4/react-core/src/components/Pagination/Pagination.tsx
evwilkin: The original thought was that changing the **perPage** option would set a new **newPerPage** value but may not change the page - that said, the **newPage** argument is always provided by this function, updating it to required per your suggestion.

packages/patternfly-4/react-core/src/components/Pagination/Pagination.tsx
```diff
@@ -69,6 +85,8 @@ export interface PaginationProps extends React.HTMLProps<HTMLDivElement> {
   perPage?: number;
   /** Select from options to number of items per page. */
   perPageOptions?: PerPageOptions[];
+  /** Indicate whether to show last full page of results when user selects perPage value greater than remaining rows */
+  defaultToFullPage?: boolean;
```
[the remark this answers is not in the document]
evwilkin: Thanks for calling this out @dlabrecq , doing this now

packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -168,6 +168,9 @@ class ComplexPaginationTableDemo extends React.Component {
 }
 ```
 
+The below example illustrates the `defaultToFullPage` prop, which makes the following changes when the user sets the number of items to display per page to an amount that exceeds the remaining amount of data:
+- The component automatically changes the page back to the last full page of results, rather than defaulting to the final page of results.
```
seanforyou23: >defaulting to the final page of results

I wasn't quite sure what "final page" means or what its impact on the overall interaction would be. What's the default experience when the `defaultToFullPage` prop is omitted?

evwilkin: @seanforyou23 in the default experience the last page will be an incomplete set of results, but I don't know that "incomplete" makes sense until you already understand how it will look.  I thought about typing out an example, but am not sure that putting more words towards it will reduce the confusion - perhaps adding steps for the user to follow in the demo and then pointing out the difference would be a clearer option.

evwilkin: @seanforyou23 I've updated the demo docs based on my feedback above, please let me know if this makes it any better from your perspective.

PR #3185, opened 2019-10-23
feat(popover): allow to set popover trigger

### LINKED ISSUE #3116: Allow to trigger Popover on mouseenter
Hi,

In Kiali we have a regression with PF4, as it seems popovers can't be triggered on mouse-enter anymore (or I didn't find how).
Is it done on purpose? It would be nice to get it back.

Before you ask, I've also considered using `Tooltip` instead of `Popover` but it doesn't really look well because it's not short content (as supposed to be in tooltips) and also tooltip text is centered which is not what I want.

Related PR in Kiali: https://github.com/kiali/kiali-ui/pull/1475

boaz0:
**What**:

closes #3116 

I added the `triggers` prop to modify the trigger option passed to `TippsyJS`.

**Please notice** that if you set `isVisible` this props is irrelevant since you are the one who manages when to display the popover and thus you will have to implement those triggers or catch those events and mutate the state that determines whether to display or hide the popover.

//cc @jotak

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Popover/Popover.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Popover/Popover.test.tsx
@@ -62,3 +62,20 @@ test('popover passes along values to tippy.js', () => {
   );
   expect(view).toMatchSnapshot();
 });
+
+test('popover with custom triggers', () => {
+  const view = shallow(
+    <Popover
+      triggers={['mouseenter', 'focus']}
+      headerContent={<div>Popover Header</div>}
+      bodyContent={
+        <div>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla turpis.
+        </div>
+      }
+    >
+      <div>Tippy Props Test</div>
+    </Popover>
+  );
+  expect(view.prop('trigger')).toBe('mouseenter focus');
+});
--- a/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
+++ b/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
@@ -27,6 +27,15 @@ export enum PopoverPosition {
   right = 'right'
 }
 
+export enum PopoverTriggers {
+  manual = 'manual',
+  focus = 'focus',
+  mouseenter = 'mouseenter',
+  click = 'click',
+}
+
+type Triggers = 'manual' | 'focus' | 'mouseenter' | 'click';
+
 export interface PopoverProps {
   /** Accessible label, required when header is not present */
   'aria-label'?: string;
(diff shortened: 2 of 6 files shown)


packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
```diff
@@ -69,6 +78,8 @@ export interface PopoverProps {
    * Instead, the consumer is responsible for closing the popover themselves by adding a callback listener for the shouldClose prop.
    */
   isVisible?: boolean;
+  /** The events which cause the popover to show. **/
+  triggers?: Triggers[];
```
tlabaj: Can you please note that is isVisible is set, this prop is ignored.

packages/patternfly-4/react-core/src/components/Popover/Popover.test.tsx
```diff
@@ -62,3 +62,20 @@ test('popover passes along values to tippy.js', () => {
   );
   expect(view).toMatchSnapshot();
 });
+
+test('popover with custom triggers', () => {
```
tlabaj: this is great.  Can you also add a demo to the demo app and an integration test for this.

packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
boaz0: Good idea!

packages/patternfly-4/react-core/src/components/Popover/Popover.test.tsx
boaz0: Yeah sure! I added a test into its spec to make sure that setting a focus trigger will display the popover. :man_technologist:

jotak:
Sorry I was on PTO . Thanks @boaz0  for taking this. I'll check asap.

jotak:
It looks good to me!

jotak:
@mcarrano it happens that we actually moved to `Tooltip` just yesterday but for a different reason, more as a workaround to another issue we had with Popover: https://github.com/kiali/kiali-ui/pull/1521

jgiardino:
> I do understand that our tooltip is very constrained and may not be giving you the flexibility you need for formatting the content. If we are to make a change, I would rather update the tooltip component to accept formatted content. @mcoker can you take a look at this issue kiali/kiali-ui#1475 What would it take to add that content to a tooltip? If that seems like a reasonable approach, I can open an issue in core to create that variant.

Regarding this comment from @mcarrano, I posted a [comment in the original issue](https://github.com/patternfly/patternfly-react/issues/3116#issuecomment-550487833) with an example of what the Tooltip currently supports.

mcoker:
@jgiardino @mcarrano technically you can put anything you want in the tooltip content as far as content goes, though as @jgiardino discovered, there may be font color issues as we set that explicitly. Instead, we could assign dark theme variable substitutions inside of the tooltip content, which would turn anything that uses our dark text color white inside of the tooltip content area. That change would happen in core and would go into react with no updates other than pulling in the core CSS changes.

And to mimic that behavior currently, you could use the class `.pf-t-dark` on an element that wraps the toolbar content, and that will do the same variable substitutions, resulting in anything that uses our default text color changing white in the tooltip content area.

mcarrano:
@boaz0 @jotak Given all of this, can we close this PR? If we want to follow up on improvements to the Tooltip component, a new issue should be opened for that.

boaz0:
Sure, let me open a follow up ticket for that.

Issue #3172, opened 2019-10-16
react-table 2.22.19 typing issue


israel-hdez:
In version 2.22.19 of react-table, which seems to be the latest at the time of writing, the sample found at this link is generating errors when building: https://www.patternfly.org/v4/documentation/react/components/table/#table-with-width-modifiers

The error is:

```
error TS2322: Type '(() => { className: string; })[]' is not assignable to type '((label?: (string & { title?: ReactNode; props?: any; }) | (object & { title?: ReactNode; props?: any; }) | (ComponentClass<any, any> & { title?: ReactNode; props?: any; }) | (FunctionComponent<any> & { ...; }) | undefined, rowData?: IRowData | undefined, columnIndex?: number | undefined, column?: IColumn | undefine...'.

168         transforms: [cellWidth(70)],
            ~~~~~~~~~~
```

Indeed, apparently, the `transforms` attribute of `ICell` is typed as `ITransforms` as seen in this code: https://github.com/patternfly/patternfly-react/blob/73c61699720ba29c5281183b3411dccfdde5aa05/packages/patternfly-4/react-table/src/components/Table/Table.tsx#L152-L164

But the `cellWidth` transform does not comply with that interface, so, typescript transpiler fails with error.

I think types/interfaces should be fixed.

seanforyou23:
@israel-hdez this should be fixed in latest master, gonna close this one. Feel free to reopen if it still isn't fixed for you. Thx for reporting it!

Issue #3148, opened 2019-10-15
Specifying expandable cellFormatter breaks columns type for table


seanforyou23:
When I specify `cellFormatters: [expandable]` as part of a column definition that's typed `(ICell | string)[]` (as directed in the docs) I get [type errors](https://gist.github.com/seanforyou23/60484029a78c463c79a129c885177dd5).

I created a [branch](https://github.com/seanforyou23/patternfly-react-seed/tree/table-type-issue) using pf-seed where you can easily see the problem. The implementation seems to be fine (live demo of the working branch [here](patternfly-react-seed-demo-url.surge.sh)) but the types complain so I'm forced to supply a type of `any[]` in the mean time.

Am I just doing it wrong or is it a regression?

<img width="267" alt="Screen Shot 2019-10-16 at 4 32 13 PM" src="https://user-images.githubusercontent.com/5942899/66956542-8e6c3d00-f032-11e9-8ad0-6f68dd5c0c25.png">

PR #3203, opened 2019-10-24
chore(table): fix types for table transforms

seanforyou23:
**What**: This PR updates the types used around the existing table decorators (transforms & formatters) so that consumers can specify the standard `(ICell | string)[]` type for table column definitions in TypeScript based apps. It also makes use of the existing types throughout a lot of the table related code, and modifies a few types slightly for consistency.

It should also help clear the `strictFunctionTypes` errors for Table.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/3148 and https://github.com/patternfly/patternfly-react/issues/3172 and https://github.com/patternfly/patternfly-react/issues/3224

### DIFF
--- a/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableActionsDemo.tsx
+++ b/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableActionsDemo.tsx
@@ -5,14 +5,20 @@ import {
   TableBody,
   TableProps,
   headerCol,
+  ICell,
   IRow,
   IRowData,
   IExtra,
-  IAction,
+  IActions,
   ISeparator
 } from '@patternfly/react-table';
 
-export class TableActionsDemo extends React.Component<TableProps, { columns: any; rows: IRow[] }> {
+interface TableState {
+  columns: (ICell | string)[];
+  rows: IRow[];
+}
+
+export class TableActionsDemo extends React.Component<TableProps, TableState> {
   constructor(props: TableProps) {
     super(props);
     this.state = {
@@ -47,7 +53,7 @@ export class TableActionsDemo extends React.Component<TableProps, { columns: any
       return null;
     }
 
-    const thirdAction =
+    const thirdAction: IActions =
       rowData.type === 'blue'
         ? [
             {
@@ -59,7 +65,7 @@ export class TableActionsDemo extends React.Component<TableProps, { columns: any
               onClick: (event, rowId, rowData, extra) =>
                 // tslint:disable-next-line:no-console
                 console.log(`clicked on Third action, on row ${rowId} of type ${rowData.type}`)
-            } as IAction
+            }
           ]
         : [];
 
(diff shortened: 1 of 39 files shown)


seanforyou23:
Just realized we probably don't actually want to swap IExtra for IRowData in the places I've done currently in this PR, but I do think that's at least the right place to be making the adjustment. Flipping the strictFunctionTypes switch really brings the issues to light. Will keep hacking away!

seanforyou23:
I've pushed some updates that improve the approach. It seems to be working well and I've only had to modify the tests slightly.  I'm not sure why the [`children` property type isn't inferred as something acceptable](https://github.com/patternfly/patternfly-react/pull/3203/files#diff-aae0f66d24fcaff183ecfbbaafe2830bR65), will look into this more.

Beyond that, the main thing I'd like to get some opinion on is [breaking out a shared return type for transformers and formatters](https://github.com/patternfly/patternfly-react/pull/3203/files#diff-9bda7132cc01104382345447864888d5R132-R145). It seems like this may have been the original intention, but I'm partly guessing. Are transforms and formatters considered "subclass" of decorators? The system will work in a variety of ways, just trying to match the correct terminology used to how the public api parts are named.

It seems to run fine in OpenShift on a quick test, and solves the type errors mentioned in the issues attached. I need to give it all another once over tomorrow and see if I can wrap these last bits up, would love any feedback in the meantime.

packages/patternfly-4/react-table/src/components/Table/utils/decorators/sortable.tsx
```diff
@@ -2,13 +2,15 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Table/table';
 import buttonStyles from '@patternfly/react-styles/css/components/Button/button';
-import { SortByDirection, IExtra, IFormatterValueType } from '../../Table';
+import { SortByDirection, IExtra, IFormatterValueType, ITransform } from '../../Table';
+import { ExtraParamsType } from '../../base/types';
```
karelhala: Looks like this type is not used anywhere.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableActionsDemo.tsx
```diff
@@ -5,14 +5,20 @@ import {
   TableBody,
   TableProps,
   headerCol,
+  ICell,
   IRow,
   IRowData,
   IExtra,
-  IAction,
+  IActions,
```
dlabrecq: Would this be considered a breaking change? That is, the `IAction` interface has been renamed `IActions`

seanforyou23: I don't think so, it should work both ways. There are quite a few places in the table code where the type assertions are currently necessary so just trying to make the code more readable by favoring the "normal" approach of applying types where we can. If everything is read `as SomethingElse` it becomes a little hard to read. Does it make sense?

dlabrecq: Just want to make sure that someone isn't relying on the `IActions` interface and we break them by renaming it as `IAction`

Issue #3073, opened 2019-10-04
Charts: interactive legend


jotak:
Is there any plan to have a built-in interactive legend as there was in pf3 / c3 charts ?
Previously, clicking on legend icons would show/hide the related series in charts.

With victory chart I believe there's no such mechanism built-in, but we can implement it through events. I'm doing something like that in Kiali: https://github.com/kiali/kiali-ui/pull/1457/commits/63d400d58c2fdcbbb4a77c087c2b76511be6801e#diff-82e26dca83e13db6b8969687fd9268cbR71-R156 . Maybe that can be part of the patternfly value-added & standard look n feel?

Also note that, currently I cannot do it using the `ChartLegend` component from PF4, as it seems events are not fired when targeting the legend (I believe it's due to the wrapping component) so I need to use `VictoryLegend` directly instead of `ChartLegend`.

jotak:
PS: this is something I can contribute for, if it gets some chance to be accepted

dlabrecq:
I have implemented this in Cost Management using the ChartLegend. This implementation allows the user to click on either the legend symbol or label.

This is the custom Legend component with event handlers for onClick:
 
<img width="644" alt="Screen Shot 2019-10-04 at 9 36 38 AM" src="https://user-images.githubusercontent.com/17481322/66211797-b9e24580-e68a-11e9-8663-78596b5614fc.png">

And this is how the custom Legend component is added to the chart:

<img width="659" alt="Screen Shot 2019-10-04 at 9 38 07 AM" src="https://user-images.githubusercontent.com/17481322/66211810-bea6f980-e68a-11e9-9bc9-59ec46557361.png">

The chart is composed of multiple components, but the legend doesn't know anything about the charts data. That's what we're wiring up above.

We can certainly add an example of this -- this was on my next to do. However, it may be possible for the `Chart` component to wire this up, if you also allow the chart to create the legend. It would also need to be implemented in other components that don't use `Chart`.

jotak:
I see, it's another approach of dealing with events, it's interesting.

> A chart is composed of multiple components, so the legend doesn't really know anything about the chart's data. That is, unless it's assumed you're providing the same data object? That's what we're wiring up above.

[...]

PR #3253, opened 2019-11-01
Charts: Interactive legend example

dlabrecq:
Adds functionality for interactive legends.

chrome-capture

Fixes https://github.com/patternfly/patternfly-react/issues/3073
Related: https://github.com/patternfly/patternfly-next/issues/2352

**Example** https://patternfly-react-pr-3253.surge.sh/patternfly-4/documentation/react/charts/chartlegend


packages/patternfly-4/react-charts/src/components/ChartLegend/examples/ChartLegend.md
```diff
@@ -475,7 +407,7 @@ class InteractiveLegendChart extends React.Component {
   }
 
   // Tips:
-  // 1. Don't omit components unless using custom colors -- that will reassign color scale
```
mturley: Should we move these Tips from the example source code to the Tips section of the markdown documentation below, so they appear without having to expand the code block? I'm not sure, since we do still want people reading this example code to make sure they scroll down to see the tips. If we do move them, maybe we can leave a comment here like `// See the Tips section below for important notes about color scale, axis labels and tooltips`.

dlabrecq: Possibly. The tips at the bottom apply to all examples, while this applies only to this particular example

mturley: Ah, I didn't realize it was specific to this example. I could go either way then.

mcarrano:
@mceledonia can you review this to make sure the implementation is consistent with design intent?

PR #3254, opened 2019-11-01
PF4: fix(README.md): provides URL location of workspace 

### LINKED ISSUE #3154: Regression in local workspace development workflow/doc
Some time ago, after starting the PF4 workspace with `yarn start:pf4`, we included output that confirmed the development server had started successfully and also a note about what url/port the server could be reached at.

At some point, months ago, this stopped working. Aside from it being a general regression of behavior in our development workflow, it also made me realize that we haven't documented this information anywhere else where users can find. Currently, the below snippet shows the last thing users are made aware of after running yarn start:pf4

`@patternfly/react-docs: atternfly/react-docs`

Notice it doesn't tell me the dev server has started (or not), it doesn't tell me what URL I can access the workspace from, and the `@p` was mangled in a previous linebreak.

The real problem is that if any ol' community member decides to pull the project locally and try to start it, they'll get to this point and likely not know what to do next. This means they need to reach out to someone (and have an exchange) before they can get up and running for the very first time.

Why did we remove these helpful and critical bits of information? Can we add them back?

jenny-s51:
Temporary fix for https://github.com/patternfly/patternfly-react/issues/3154#issuecomment-545600553

### DIFF
--- a/packages/patternfly-4/react-core/README.md
+++ b/packages/patternfly-4/react-core/README.md
@@ -64,7 +64,7 @@ Note: All commands below assume you are on the root directory in this repository
 
 ### Install
 
-Run to install all the dependencies
+Run to install all the dependencies.
 
 ```sh
 yarn install && yarn build
@@ -72,15 +72,17 @@ yarn install && yarn build
 
 ### Running
 
-To start the site locally.
+To start the site locally and view the site:
 
 ```sh
 yarn start:pf4
 ```
 
+Go to [localhost:8000](http://localhost:8000/).
+
 ### Building
 
-To build the site.
+To build the site:
 
 ```sh
 yarn build:docs
--- a/packages/patternfly-4/react-core/README.md
+++ b/packages/patternfly-4/react-core/README.md
@@ -72,13 +72,13 @@ yarn install && yarn build
 
 ### Running
 
-To start the site locally and view the site:
+To launch the development server and view the workspace:
 
 ```sh
 yarn start:pf4
 ```
 
-Go to [localhost:8000](http://localhost:8000/).
+Go to localhost:8000.
 
 ### Building
 


packages/patternfly-4/react-core/README.md
```diff
@@ -64,23 +64,25 @@ Note: All commands below assume you are on the root directory in this repository
...
+Run to install all the dependencies.
 
 ```sh
 yarn install && yarn build
 ```
 
 ### Running
 
-To start the site locally.
+To start the site locally and view the site:
```
seanforyou23: Maybe consider referring to these like "start the _development server_" and/or "launch the _workspace_" would be more intuitive? I wonder if people will assume the right thing when they read "site".

packages/patternfly-4/react-core/README.md
```diff
@@ -64,23 +64,25 @@ Note: All commands below assume you are on the root directory in this repository
...
 
 ```sh
 yarn start:pf4
 ```
 
+Go to [localhost:8000](http://localhost:8000/).
```
seanforyou23: Using a static link here could pose a couple of issues. The first is that this will show as a broken link for any crawler-bots since `http://localhost:8000/` doesn't point to anything public. Another potential trip up could be that users could be already running another service on port 8000 locally, in which case following the link would take them somewhere else. Maybe we could use a plaintext version of the address instead?

jenny-s51:
@seanforyou23 Thank you for your helpful feedback Michael! I've updated the README according to your suggestions.

packages/patternfly-4/react-core/README.md
```diff
@@ -64,23 +64,25 @@ Note: All commands below assume you are on the root directory in this repository
...
 
-To start the site locally.
+To launch the development server and view the workspace:
 
 ```sh
 yarn start:pf4
 ```
 
+Go to localhost:8000.
+
```
dlabrecq: Although adding more info to the README is a good thing, this does not appear to address the original issue. That is, the `yarn start:dev` command no longer confirms exactly where the app is running.

This also does not account for the example app running on a different port, should port 8000 be occupied already.

jenny-s51: @dlabrecq Thank you for your feedback Dan! This has to do with something Zack addressed here: https://github.com/patternfly/patternfly-react/issues/3154#issuecomment-545600553 and I added that line based on what was agreed upon in this comment.

Good point about the different port -- do you happen to know what port the app runs on if localhost:8000 is already occupied?

seanforyou23: @jenny-s51 if the port is already in use it prints to the console `@patternfly/react-docs: Something is already running at port 8000`. I think it used to be the case that an interactive console would take over and allow you to specify a different port, then proceed to launch the dev server using that new config. It seems to have regressed or been removed at some point. 

I'm personally ok with pulling these changes in, but I agree it doesn't really fix the root cause so as long as we can leave the related issue open for future fixes we should be ok.

dlabrecq: Ok, can leave that issue opened

PR #3258, opened 2019-11-03
fix(flex): in breakpointMods set breakpoint as optional and add enums

### LINKED ISSUE #3244: PF4 Layout Flex: Typing improvements for breakpointMods
There is an inconsistency between documentation and typings for `breakpointMods`.
Typing for breakpointMods suggests `justify-content-flex-space-between` for `modifier` which is invalid and undefined. Correct value is `justify-content-space-between`.
Using `justify-content-space-between` leads to TypeError: 
```
TS2322: Type 'string' is not assignable to type '"column" | "flex" | "inline-flex" | "column-reverse" | "row" | "row-reverse" | "nowrap" | "wrap" | "wrap-reverse" | "full-width" | "spacer-none" | "spacer-xs" | "spacer-sm" | ... 44 more ... | "align-content-space-around"'.
```
`breakpoint` should be optional because docs shows `breakpoint` not being used with `justify-content-space-between`.
Documentation for PF4 show the correct implementation for this: https://patternfly-react.surge.sh/patternfly-4/documentation/react/layouts/flex#justify-content-space-between

`<Flex breakpointMods={[{ modifier: 'justify-content-space-between' }]} />`

### LINKED ISSUE #3248: Flex component TypeScript error
Attempting to use this simple Flex component example

```
<Flex className="example-border">
  <FlexItem breakpointMods={[{modifier: "align-right"}]}>Flex item</FlexItem>
  <FlexItem>Flex item</FlexItem>
</Flex>
```

And encountered the TypeScript error below:

Screen Shot 2019-10-30 at 5 45 45 PM

Also tried importing FlexBreakpointMod, but it is not exported properly.

Screen Shot 2019-10-30 at 5 45 15 PM

boaz0:
**What**:

closes #3244 
closes #3248

- Set breakpoint as optional in breakpointMd
- Add FlexModifiers and FlexItemModifiers
- Update examples

### DIFF
--- a/packages/patternfly-4/react-core/src/layouts/Flex/FlexUtils.tsx
+++ b/packages/patternfly-4/react-core/src/layouts/Flex/FlexUtils.tsx
@@ -2,98 +2,112 @@ import * as React from 'react';
 import { getModifier } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/layouts/Flex/flex';
 
+export enum FlexModifiers {
+  flex = 'flex',
+  'inline-flex' = 'inline-flex',
+  'spacer-none' = 'spacer-none',
+  'spacer-xs' = 'spacer-xs',
+  'spacer-sm' = 'spacer-sm',
+  'spacer-md' = 'spacer-md',
+  'spacer-lg' = 'spacer-lg',
+  'spacer-xl' = 'spacer-xl',
+  'spacer-2xl' = 'spacer-2xl',
+  'spacer-3xl' = 'spacer-3xl',
+  'space-items-none' = 'space-items-none',
+  'space-items-xs' = 'space-items-xs',
+  'space-items-sm' = 'space-items-sm',
+  'space-items-md' = 'space-items-md',
+  'space-items-lg' = 'space-items-lg',
+  'space-items-xl' = 'space-items-xl',
+  'space-items-2xl' = 'space-items-2xl',
+  'space-items-3xl' = 'space-items-3xl',
+  grow = 'grow',
+  shrink = 'shrink',
+  'flex-1' = 'flex-1',
+  'flex-2' = 'flex-2',
+  'flex-3' = 'flex-3',
+  'flex-4' = 'flex-4',
+  'flex-default' = 'flex-default',
+  'flex-none' = 'flex-none',
+  column = 'column',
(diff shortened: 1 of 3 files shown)


rebeccaalpert:
@tlabaj wasn't there a discussion about how we wouldn't use enums in TypeScript components? Has that changed?

boaz0:
@rebeccaalpert I didn't know that you're not going to use enums.
What should I use instead? Because if I am using literal types, passing a string will show a typescript error unless I will use the `as` keyword.

rebeccaalpert:
@boaz0 - It sounds like enums are ok. (Wanted to check with more active PatternFly devs.) Back when we were converting all the components to TypeScript there had been some back-and-forth over declaring everything as strings vs. using enums.

packages/patternfly-4/react-core/src/layouts/Flex/FlexUtils.tsx
```diff
@@ -2,98 +2,112 @@ import * as React from 'react';
...
+  'align-right' = 'align-right',
+  'align-left' = 'align-left',
+  'align-self-flex-start' = 'align-self-flex-start',
+  'align-self-flex-end' = 'align-self-flex-end',
+  'align-self-flex-center' = 'align-self-flex-center',
+  'align-self-flex-baseline' = 'align-self-flex-baseline',
+  'align-self-flex-stretch' = 'align-self-flex-stretch',
+  'justify-content-flex-end' = 'justify-content-flex-end',
+  'justify-content-center' = 'justify-content-center',
+  'justify-content-flex-space-between' = 'justify-content-flex-space-between',
```
sahil143: @tlabaj Not sure if this value is correct. Docs example says to use `justify-content-space-between` https://github.com/patternfly/patternfly-react/pull/3258/files#diff-fee6334c94230e770989b535c35f87f7R428
Also, `justify-content-flex-space-between` doesn't work when this is used an undefined class is applied to the element.

boaz0: :fearful: ho snap! I see this now. It looks like the Modifier type was in correct to begin with. It didn't have `justify-content-space-between` and the flex.css doesn't have `justify-content-space-between` but `justify-content-space-between`.

My mistake for not paying attention. I will suggest a fix

Issue #3263, opened 2019-11-05
Needs example/doc for onRowClick in Table component


seanforyou23:
Currently, there is no example of how to use the `onRowClick` prop in Table. Further, the type documentation doesn't give too much information around what params this handler can receive.

<img width="902" alt="Screen Shot 2019-11-05 at 9 17 48 AM" src="https://user-images.githubusercontent.com/5942899/68215470-4abf8000-ffad-11e9-8863-c2adf8f7dc8b.png">

Let's add some more doc/examples for this.

PR #3265, opened 2019-11-05
chore(table): add types, examples, and demo for onRowClick

seanforyou23:
**What**: This PR adds an example of how to use the onRowClick prop, and specifies more specific types for this handler function. It also adds a demo in react-integration and tests. I also noticed that the [default onRowClick function](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/Body.tsx#L119) would [fire on ever row click](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/Body.tsx#L40), even if it wasn't defined by the user, so I added a check so that it only fires if a handler is provided by the user.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/3263

Screen Shot 2019-11-05 at 9 19 40 AM

@KKoukiou FYI

add IComputedData type
provide type for onRowClick handler
add onRowClick example to docs
add onRowClick test for table
add on row click demo in react-integration
clean up customRowWrapper demo/example
only fire onRowClick when it is provided by user


packages/patternfly-4/react-table/src/components/Table/Body.tsx
```diff
@@ -116,7 +124,7 @@ export const TableBody = ({
   className = '' as string,
   children = null as React.ReactNode,
   rowKey = 'id' as string,
-  onRowClick = (...args: any) => undefined as any,
+  onRowClick = undefined,
```
karelhala: Is there any reason why you removed the empty function with undefined and `if`?

karelhala:
> I also noticed that the default onRowClick function would fire on ever row click, even if it wasn't defined by the user

I don't think that it's a bad behavior, the function that is called is just an empty function. If we change this to `undefined` we have to check for it to be defined and that adds to a complexity of this already really complex component.

packages/patternfly-4/react-table/src/components/Table/Body.tsx
seanforyou23: Yes, I suppose it was two-fold. I replaced the default function because it executed every time a row was clicked, even if the table doesn't specify an `onRowClick` prop, which kinda looked like an error. Having two instances of `any` along with the type assertion `undefined as any` in the same function signature made me think it was just temporary implementation details that could stand to be cleaned up.

seanforyou23:
Thx @karelhala - I added the default function back and removed the check. I still feel kinda weird about firing stub function handlers like this as it could lead to performance issues if the pattern is repeated across many features of many components. I did take the time to add a type for the handler, though, so at least now it's a little more type-safe. Lemme know if you find anything else!

packages/patternfly-4/react-table/src/components/Table/Body.tsx
```diff
@@ -37,6 +49,7 @@ class ContextBody extends React.Component<TableBodyProps, {}> {
           isInput: (event.target as HTMLElement).tagName !== 'INPUT',
           isButton: (event.target as HTMLElement).tagName !== 'BUTTON'
         };
+
         onRowClick(event, row, rowProps, computedData);
```
dlabrecq: Should we test if `onRowClick` exists first? That way you don't need to fire a stub function?

karelhala: That's what was in the previous version and it's not really that good. Firing empty function is nothig compared to checking if function exists. The undefined instead of empty function might hurt us in future when we forgot to check if this function is defined and blow UI.

Overall it boils down to be consistent and since almost every other component is using either `() => undefined`  or `Function()` instead of checking before call I think empty function is better.

dlabrecq: Yes, we use this pattern a lot with stateless components. Don't agree it would "blow UI", but don't feel strongly about it.

PR #3291, opened 2019-11-08
fix(CatalogTile): Edit target selector

rebeccaalpert:
Adjusted selector for gradient effect.

### DIFF
--- a/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
+++ b/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
@@ -60,13 +60,7 @@
 .catalog-tile-pf-description {
   margin-top: 0;
 
-  > span {
-    position: initial;
-  }
-}
-
-.catalog-tile-pf-body {
-  .catalog-tile-pf-description span.truncated {
+  .truncated {
     overflow: hidden;
     position: relative;
 


packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
```diff
@@ -59,26 +59,20 @@
...
-    }
+  & span.truncated::after {
+    background: linear-gradient(to right, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 75%);
+    bottom: 0;
+    color: transparent;
+    content: ".";
+    position: absolute;
+    right: 0;
+    width: 50%;
+    text-align: right;
```
jeff-phillips-18: Better to have:
```
 .truncated {
    overflow: hidden;
    position: relative;

    &::after {
      background: linear-gradient(to right, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 75%);
      bottom: 0;
      color: transparent;
      content: ".";
      position: absolute;
      right: 0;
      width: 50%;
      text-align: right;
    }
  }
}
```

However, I don't believe we want this only applied when `truncated` is set. The idea is for this to auto truncate when necessary and the caller need not provide truncated text.

rebeccaalpert: The class is applied whenever the truncation function is called. The user doesn't have to provide truncated text.

rebeccaalpert: Updated the commit with the change.

jeff-phillips-18: Why the check for `isTruncated` at all?

rebeccaalpert: If all the descriptions have these styles, even really short descriptions that don't need truncation get faded out. This is a due to a difference between the old card and the new card. The old card had a fixed height applied based on the amount of space the text took up, so it wasn't an issue. I didn't think it was a great idea to use the fixed height thing with PF4 card styles.

jeff-phillips-18: OK, this PR is fine by me. The height issue might be a problem if applications try to show tiles with differing header text lengths (causing wrapping). In PF3, that was taken into consideration and the height allowed for the descriptive text auto adjusted for it. Something to watch out for.

PR #3232, opened 2019-10-29
docs(table): add empty state and spinner to pagination table demo

boaz0:
**What**:

closes #3184

PR #3294, opened 2019-11-08
fix(PaginationTableDemo): add Spinner and empty state to demo

### LINKED ISSUE #3184: Docs: update PaginationToolbar demo to include spinner and error handling
Add to the pagination toolbar demo spinner instead of just displaying the "Loading" text.
In addition, add error handling to display an error state if there is no data fetched.

nicolethoen:
Fixes #3184

### DIFF
--- a/packages/patternfly-4/react-core/src/demos/PaginationTable.md
+++ b/packages/patternfly-4/react-core/src/demos/PaginationTable.md
@@ -48,8 +48,14 @@ class ComplexPaginationTableDemo extends React.Component {
       total: 0,
       page: 0,
       error: null,
-      loading: true
+      loading: true,
+      forceLoadingState: false,
     };
+    
+    this.handleCheckboxChange = (checked) => {
+      console.log(checked);
+      this.setState({ forceLoadingState: checked });
+    }
   }
 
   fetch(page, perPage) {
@@ -62,6 +68,19 @@ class ComplexPaginationTableDemo extends React.Component {
 
   componentDidMount() {
     this.fetch(this.state.page || 1, this.state.perPage || 20);
+  }
+  
+  renderLoadingStateCheckbox() {
+    return (
+      <Checkbox
+        label="View loading state"
+        isChecked={this.state.forceLoadingState}
+        onChange={this.handleCheckboxChange}
+        aria-label="view loading state checkbox"
+        id="check"
+        name="check"
+      />
+    )
   }
 
   renderPagination(variant = 'top') {
@@ -79,7 +98,7 @@ class ComplexPaginationTableDemo extends React.Component {
   }
 
   render() {
-    const { loading, res, error } = this.state;
+    const { loading, res, error, forceLoadingState } = this.state;
     if (error) {
       const noResultsRows = [{
         heightAuto: true,
@@ -129,14 +148,15 @@ class ComplexPaginationTableDemo extends React.Component {
     
     return (
       <React.Fragment>
+        {this.renderLoadingStateCheckbox()}
         {this.renderPagination()}
-        {!loading && (
+        {!(loading || forceLoadingState) && (
           <Table cells={['Title', 'Body']} rows={res.map(post => [post.title, post.body])} aria-label="Pagination Table Demo">
             <TableHeader />
             <TableBody />
           </Table>
         )}
-        {loading && (
+        {(loading || forceLoadingState) && (
           <Table cells={['Title', 'Body']} rows={loadingRows} aria-label="Pagination Table Demo">
             <TableHeader />
             <TableBody />
--- a/packages/patternfly-4/react-core/src/demos/PaginationTable.md
+++ b/packages/patternfly-4/react-core/src/demos/PaginationTable.md
@@ -26,7 +26,8 @@ import {
   EmptyStateIcon,
   EmptyStateBody,
   EmptyStateSecondaryActions,
-  Bullseye 
+  Bullseye,
+  Radio 
 } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody} from '@patternfly/react-table';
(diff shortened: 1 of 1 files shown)


dlabrecq:
Looks like we have two PRs for the same issue?
https://github.com/patternfly/patternfly-react/pull/3232

packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -50,17 +70,50 @@ class ComplexPaginationTableDemo extends React.Component {
...
+                </Title>
+                <EmptyStateBody>
+                  No results match the filter criteria. Remove all filters or clear all filters to show results.
+                </EmptyStateBody>
+              </EmptyState>
+            </Bullseye>
+            )
+          },
+        ]
+      }]
```
dlabrecq: How do users see the empty state? Do we need to edit the code to remove rows?

packages/patternfly-4/react-core/src/demos/PaginationTable.md
```diff
@@ -50,17 +70,50 @@ class ComplexPaginationTableDemo extends React.Component {
...
         {this.renderPagination()}
         {!loading && (
-          <Table cells={['Title', 'Body']} rows={this.state.res.map(post => [post.title, post.body])}>
+          <Table cells={['Title', 'Body']} rows={res.map(post => [post.title, post.body])} aria-label="Pagination Table Demo">
             <TableHeader />
             <TableBody />
           </Table>
         )}
-        {loading && <center><Title size="3xl">Please wait while loading data</Title></center>}
+        {loading && <center><Spinner size="xl"/></center>}
```
dlabrecq: I'm not seeing a spinner. I briefly see the empty state prior to loading the table rows. Is there a way to better see this?

packages/patternfly-4/react-core/src/demos/PaginationTable.md
boaz0: we should probably add a filter to the table and if a filter is set show empty state.

mcarrano:
@nicolethoen for some reason I don't seem to be able to access the preview link. It says "project not found."

nicolethoen:
@dlabrecq @boaz0 I can work to add a way to force view these states.

nicolethoen:
@mcarrano those changes should be no problem! There are no filters on the table though, so the 'Clear All Filters' action would be a dead action, so do we still want it? I removed it intentionally for just that reason.

mcarrano:
That's an interesting question @nicolethoen . I think it's important to have the Clear filters action since this message would only really appear as the result of a filter. But you are right that this would not really do anything in the current demo. Would it take much to add a simple filter to this, like just a Select list with some values?

nicolethoen:
@mcarrano we have a filterable table demo... would it need to also use the experimental data toolbar to be implemented correctly? seems out of scope for this demo. The only legitimate reason there would be no data to display is if no data is returned from the API. It'd be more of an error.

mcarrano:
Yes, that's true @nicolethoen . But this is a slightly different empty state pattern in that case. I will try to set up some time to sync with you on what to put here.

nicolethoen:
Note: radio buttons were removed - it was decided they were muddying context for the user a little bit. And 'Empty State' was replaced with more of an 'Error state' that is not demonstrated intentionally as part of the demo - but if there is an ever some unknown error retrieving data, it will fail gracefully. Any anyone curious about error handling in this case can look at how it's handled in the code. It was decided this demo is not going to highlight the error state, because it is not an Error State demo, it is a Pagination Table demo.

nicolethoen:
@dlabrecq I dont think that the Pagination Demo is the place to demonstrate the empty state or error state of a table. It is simply for demonstrating pagination. The other demos don't explicitly demonstrate empty states or error states either. 

Before the change, as data was being loaded and paginated, you saw a message that quickly flashed `Please wait while loading data` before it disappeared. Now you see the spinner instead. 

And we've added an error state if no data is retrieved, which should never happen unless there is a connectivity issue - so I added the error state so that any issues that arise will fail gracefully using a Patternfly pattern. So it's not explicitly demonstrable, but is more there as it should if anyone wants to implement Pagination in their table correctly. I think that is following the demo precedent we've established.

These changes were made after talking to @mcarrano and Content strategist Margot

nicolethoen:
I'll also add that the Filterable Table demo does demonstrate an empty state when all options have been filtered out. That seems to be a more relevant place to demonstrate this.

dlabrecq:
I'm just going by what issue #3184 is asking for. If the functionality is not appropriate for the pagination demo, then perhaps we should not modify this particular demo? Or, leave the issue open?

Regardless, my issue is that I don't even see the spinner.

nicolethoen:
Your machine must be processing the data too fast for you to see it. @mcarrano would it be appropriate to add a 'View Loading State" checkbox to force the table to stay in a loading state?

dlabrecq:
How about a timeout that loads the table after a second or two? Then, there is no need for a checkbox

mcarrano:
@nicolethoen @dlabrecq I'm good with either of those approaches. It was the inclusion of the empty state here that was causing me the most confusion. I'm wondering if we should have a demo that shows the various types of empty states in context as there are lots of different reasons to get an empty state. Do you think that would be useful?

nicolethoen:
@dlabrecq I updated the attached issue to reflect the changes made in this PR. And I opened a new  issue to document the missing demos. #3330

nicolethoen:
Any other thoughts @dlabrecq or @mcarrano? Can we get this PR in soon :)

mcarrano:
@nicolethoen It makes sense to me to create these in a separate but related demo. I've left a comment to see if we can get that om the roadmap.

Issue #3199, opened 2019-10-24
Add coverage reporting for cypress tests


dgutride:

jenny-s51:
In progress

evwilkin:
Resolved in #3295

PR #3295, opened 2019-11-08
Coverage for Cypress tests

jenny-s51:
PR for #3199


.circleci/config.yml
```diff
@@ -4,6 +4,7 @@ version: 2
 # I'd love to find docs on this syntax.
 references:
   js_deps_paths: &js_deps_paths
+  - /home/circleci/cache/Cypress
```
redallen: Don't need this now

.circleci/config.yml
```diff
@@ -26,7 +27,7 @@ references:
   - packages/react-codemods/node_modules/
   - packages/react-icons/node_modules/
   build_cache_paths: &build_cache_paths
-  - .cache/
+  - /home/circleci/cache/Cypress
```
redallen: Don't need this now

.circleci/config.yml
```diff
@@ -141,8 +142,11 @@ jobs:
         paths: *build_cache_paths
         key: *build_cache_key
     - persist_to_workspace:
-        root: ~/project
-        paths: *build_cache_paths
```
redallen: Leave this in

.circleci/config.yml
```diff
@@ -141,8 +142,11 @@ jobs:
...
+    #for integration tests
+        root: ~/
```
redallen: Add new `- persist_to_workspace:`

yarn.lock
```diff
@@ -1,4 +1,5 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+
```
redallen: You can revert this now

.circleci/config.yml
```diff
@@ -182,6 +186,17 @@ jobs:
...
+        name: Run Cypress Integration Tests
+        command: yarn test:integration && yarn run codecov -f packages/patternfly-4/react-integration/results/merged-test-results.xml -F patternfly4integration
+        no_output_timeout: 180m
+    - run: 
+        name: Create HTML file of Cypress coverage
+        command: yarn build:integration:report
+    - persist_to_workspace:
+        root: ~/project
+        paths:
+        - packages/patternfly-4/react-integration/results
```
redallen: Please make https://github.com/patternfly/patternfly-react/blob/master/.circleci/copy-docs.sh copy this HTML file to the PF4 docs directory! Then it'll get uploaded along with a doc build.

redallen: Optionally, also store the folder using
```yml
- store_artifacts:
          path: /tmp/artifacts
```
so we can download it from CircleCI if we ever need to! Docs: https://circleci.com/docs/2.0/artifacts/

.circleci/config.yml
```diff
@@ -26,7 +26,6 @@ references:
   - packages/react-codemods/node_modules/
   - packages/react-icons/node_modules/
   build_cache_paths: &build_cache_paths
-  - .cache/
```
redallen: You can put this back, it's related to Gatsby, not Cypress.

.circleci/config.yml
```diff
@@ -182,6 +187,19 @@ jobs:
     - run:
```
redallen: After `- *attach_workspace` add:
```
- attach_workspace:
      at: ~/
```
and I think that will restore the workspace you cached!

evwilkin:
Resolves #3199

Issue #3279, opened 2019-11-07
Selectable transform issues


seanforyou23:
We have a `selectable` transform for react-table, but it isn't clear how it should be used from the [selectable table example](https://www.patternfly.org/v4/documentation/react/components/table#selectable) in our docs.

It's also [unused](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSelectableDemo.tsx#L10) in the TableSelectableDemo in react-integration.

Seems like the `canSelectAll` api part is integral to the selectable functionality but the relationship between this prop and the transform isn't clear.

There's also a [type mismatch](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/utils/decorators/selectable.tsx#L31) mentioned that we'd like to go ahead and address.

Let's fix any issues with the transform, put it to use, and ensure it's well tested/documented.

Issue #3231, opened 2019-10-29
Some component demos don't meet "noImplicitAny" requirements


seanforyou23:
I noticed that when a consuming typescript app uses the noImplicitAny compiler option, errors are thrown during build time. I'm seeing this issue in the following component demos in react-integration:

- [x] AboutModal/AboutModalDemo.tsx
- [x] AccordionDemo/AccordionDemo.tsx
- [x] AlertDemo/AlertDemo.tsx
- [x] ApplicationLauncherDemo/ApplicationLauncherDemo.tsx
- [x] CheckboxDemo/CheckboxDemo.tsx
- [x] ChipGroupDemo/ChipGroupDefaultIsOpenDemo.tsx
- [x] ChipGroupDemo/ChipGroupDemo.tsx
- [x] ContextSelectorDemo/ContextSelectorDemo.tsx
- [x] DataToolbarDemo/DataToolbarDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationSmallRightDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationSmallStaticRightDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationStaticInvertedRightDemo.tsx
- [x] DonutUtilizationChartDemo/DonutUtilizationStaticRightDemo.tsx
- [x] FormDemo/FormDemo.tsx
- [x] FormSelectDemo/FormSelectDemo.tsx
- [x] InputGroupDemo/InputGroupDemo.tsx
- [x] NavDemo/NavDemo.tsx
- [x] NotificationBadgeDemo/NotificationBadgeDemo.tsx
- [x] OptionsMenuDemo/OptionsMenuDemo.tsx
- [x] OverflowMenuDemo/OverflowMenuDemo.tsx
- [x] PaginationDemo/PaginationDemo.tsx
- [x] RadioDemo/RadioDemo.tsx
- [x] SelectDemo/SelectDemo.tsx
- [x] SwitchDemo/SwitchDemo.tsx
[...]

Issue #2673, opened 2019-08-08
PF4 React-Table - TypeScript followup


priley86:
In #2360, I noted a couple of additional nice-to-haves:

- [x] It would be better to expose all formatter/transformer props in `Table.tsx` with our explicitly defined types. Doing so would also require ensuring all formatters and transformers are tested via that interface (as explicit props to `Table`). https://github.com/patternfly/patternfly-react/pull/2360#issuecomment-517371474

- [ ] Ensure that Prettier formatter is applied to React-Table after it is applied in React-Core: #2362 

- [ ] Determine if we'd like to introduce breaking API changes for the `tslint` settings different from React-Core: https://github.com/patternfly/patternfly-react/pull/2360#issuecomment-513822602

- [x] Add Cypress/e2e tests

- [ ] Consider using an interface for static type analysis of aria props/captions/headers. See [comment here](https://github.com/patternfly/patternfly-react/pull/2360#discussion_r312525317).

- [ ] Consider rewriting this logic in `collapsible.tsx` for better type analysis: https://github.com/patternfly/patternfly-react/pull/2360#discussion_r312541336

- [ ] Update event types in `selectable.tsx`: see [comment](https://github.com/patternfly/patternfly-react/pull/2360#discussion_r312542774)

mturley:
Just commenting here so this issue ends up in my "participating" filter. This all sounds like good stuff!

mturley:
We haven't addressed all these things yet, so I'm bumping stalebot -- @seanforyou23 do you want to keep this issue open or track these kinds of things in individual issues?

seanforyou23:
Thanks @mturley - I'm just now coming across this issue and find it very helpful! I'm working on fixing the `selectable` transform today - and I've had my eye on some of the other stuff mentioned in this issue so let's keep it open and I'll work on seeing through the remaining TODO's. Will tag you for review once I get a PR up.

mturley:
Bumping stalebot because https://github.com/patternfly/patternfly-react/pull/3296 is still open

Issue #3223, opened 2019-10-28
react-integration compiler options masks type issues


seanforyou23:
Currently, our react-integration pkg serves the purpose of ensuring our documentation examples will build and compile as expected. It works great, but I recently realized [strict type checking is turned off](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-integration/demo-app-ts/tsconfig.json#L20-L21). Because of this, issues go undetected until they are happened upon by community members. Setting `strict` to `true` is standard practice and it doesn't make sense to suggest consumers lessen their compiler strictness just to be able to consume our components.

I've increased the strictness of the checks we can without making further changes in https://github.com/patternfly/patternfly-react/pull/3222.

This leaves us needing to work through changes required to enable the following compiler options:

- [ ] strictPropertyInitialization
- [x] [noImplicitAny](https://github.com/patternfly/patternfly-react/issues/3231)
- [ ] [strictNullChecks](https://github.com/patternfly/patternfly-react/issues/3230)
- [x] [strictFunctionTypes](https://github.com/patternfly/patternfly-react/issues/3224)

seanforyou23:
This is a critical issue, commenting so stale bot doesn't close it.

dgutride:
@seanforyou23  - since the strict changes were made in https://github.com/patternfly/patternfly-react/pull/3222 is there some work this issue is still left open to capture?

seanforyou23:
Hi @dgutride - yes we still have [several strict modes ](https://github.com/patternfly/patternfly-react/pull/3222/files#diff-2b96ca5fa7e86654dc36bfcf59237f20R24-R27)left turned off. I was kinda thinking of this issue as an epic and linked out to them from this one so it's easier to follow this work as a sort of "theme" since it requires breaking changes. Does this help or answer your question? Let me know if I should change anything about how they are captured.

dgutride:
that makes sense - so we still need to enable them one by one and fix as we go - thx for the clarification!

PR #3296, opened 2019-11-11
chore(table): update types in table

seanforyou23:
**What**: This PR aligns the onSelect event type from

`React.MouseEvent` & `React.ChangeEvent<HTMLInputElement>`

to

`React.FormEvent<HTMLInputElement>`

The `target` property of the event object is what the user actually clicked on, and it can vary as the underlying implementation changes in structure. The `currentTarget` property is the element the event was bound to, and will never change.

By using FormEvent, we now correctly get assistance from the IDE. 

Before:
Screen Shot 2019-11-11 at 10 00 33 AM

After:
Screen Shot 2019-11-11 at 10 16 02 AM

This PR also clears all noImplicitAny errors for Table in react-integration and simplifies the Selectable Table example in the docs.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/3279 https://github.com/patternfly/patternfly-react/issues/2673 https://github.com/patternfly/patternfly-react/issues/3231 

update/align onSelect event type from React.MouseEvent and React.ChangeEvent<HTMLInputElement> to React.FormEvent<HTMLInputElement>
clear todo and tslint error in selectable transform
simplify selectable table example
export types for ICustomCell and ISelectTransform
clear noImplicitAny errors for Table

## Breaking changes
- Change type of `onSelect` in Table from `(event: React.ChangeEvent<HTMLInputElement>) => void` to `(event: React.FormEvent<HTMLInputElement>) => void`

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSelectableDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSelectableDemo.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Table, TableHeader, TableBody, TableProps, headerCol, ICell, IRow } from '@patternfly/react-table';
 import { Checkbox } from '@patternfly/react-core';
+import '@patternfly/patternfly/utilities/Spacing/spacing.css';
 
 interface TableState {
   columns: (ICell | string)[];
@@ -36,9 +37,10 @@ export class TableSelectableDemo extends React.Component<TableProps, TableState>
       canSelectAll: true
     };
     this.onSelect = this.onSelect.bind(this);
+    this.toggleSelect = this.toggleSelect.bind(this);
   }
 
-  onSelect(event: React.MouseEvent, isSelected: boolean, rowId: number) {
+  onSelect(event: React.FormEvent, isSelected: boolean, rowId: number) {
     let rows: IRow[];
     if (rowId === -1) {
       rows = this.state.rows.map(oneRow => {
@@ -65,28 +67,29 @@ export class TableSelectableDemo extends React.Component<TableProps, TableState>
   }
 
   render() {
-    const { columns, rows } = this.state;
+    const { columns, rows, canSelectAll } = this.state;
 
     return (
       <div>
-        <Table
-          caption="Selectable Table"
-          onSelect={this.onSelect}
-          cells={columns}
-          rows={rows}
-          canSelectAll={this.state.canSelectAll}
-        >
-          <TableHeader />
-          <TableBody />
-        </Table>
         <Checkbox
-          label="canSelectAll"
-          isChecked={this.state.canSelectAll}
+          label="Can select all"
+          className="pf-u-mb-lg"
+          isChecked={canSelectAll}
           onChange={this.toggleSelect}
           aria-label="toggle select all checkbox"
           id="toggle-select-all"
           name="toggle-select-all"
         />
+        <Table
+          onSelect={this.onSelect}
+          canSelectAll={canSelectAll}
+          caption="Selectable Table"
+          cells={columns}
+          rows={rows}
+        >
+          <TableHeader />
+          <TableBody />
+        </Table>
       </div>
     );
   }
--- a/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSimpleDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSimpleDemo.tsx
@@ -1,7 +1,10 @@
 import * as React from 'react';
 import { Table, TableHeader, TableBody, TableProps, textCenter, ICell, IRow } from '@patternfly/react-table';
 
-export class TableSimpleDemo extends React.Component<TableProps, { columns: (ICell | string)[]; rows: IRow[] }> {
+export class TableSimpleDemo extends React.Component<
+  TableProps,
+  { columns: (ICell | string)[]; rows: (IRow | string[])[] }
+> {
   constructor(props: TableProps) {
     super(props);
     this.state = {
@@ -17,25 +20,31 @@ export class TableSimpleDemo extends React.Component<TableProps, { columns: (ICe
         }
       ],
       rows: [
-        ['one', 'two', 'three', 'four', 'five'],
-        [
-          {
-            title: <div>one - 2</div>,
-            props: { title: 'hover title', colSpan: 3 }
-          },
-          'four - 2',
-          'five - 2'
-        ],
-        [
-          'one - 3',
-          'two - 3',
-          'three - 3',
-          'four - 3',
-          {
-            title: 'five - 3 (not centered)',
-            props: { textCenter: false }
-          }
-        ]
+        {
+          cells: ['one', 'two', 'three', 'four', 'five']
+        },
+        {
+          cells: [
+            {
+              title: <div>one - 2</div>,
+              props: { title: 'hover title', colSpan: 3 }
+            },
+            'four - 2',
+            'five - 2'
+          ]
+        },
+        {
+          cells: [
+            'one - 3',
+            'two - 3',
+            'three - 3',
+            'four - 3',
+            {
+              title: 'five - 3 (not centered)',
+              props: { textCenter: false }
+            }
+          ]
+        }
       ]
     };
   }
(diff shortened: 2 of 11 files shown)


packages/patternfly-4/react-table/src/components/Table/SelectColumn.tsx
```diff
@@ -4,13 +4,13 @@ export interface SelectColumnProps {
   name?: string;
   children?: React.ReactNode;
   className?: string;
-  onSelect?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onSelect?: (event: React.FormEvent<HTMLInputElement>) => void;
```
redallen: This changes our exported types.

packages/patternfly-4/react-table/src/components/Table/SelectColumn.tsx
```diff
@@ -4,13 +4,13 @@ export interface SelectColumnProps {
...
 }
 
 export const SelectColumn: React.FunctionComponent<SelectColumnProps> = ({
   children = null as React.ReactNode,
   className = '',
-  onSelect = null as (event: React.ChangeEvent<HTMLInputElement>) => void,
+  onSelect = null as (event: React.FormEvent<HTMLInputElement>) => void,
```
redallen: This changes our exported types.

packages/patternfly-4/react-table/src/components/Table/Table.tsx
```diff
@@ -46,7 +46,7 @@ export type OnExpand = (
   extraData: IExtraData
 ) => void;
 export type OnSelect = (
-  event: React.MouseEvent,
+  event: React.FormEvent<HTMLInputElement>,
```
redallen: This changes our exported types.

seanforyou23:
Thanks for the feedback @redallen 

> Typescript consumers who rely on these types ... will have their Typescript build fail if we go ahead with this change.

This 👆 is why I consider these changes a breaking change. 

> (are there many? who are they?)

As far as who and how many, I'm not aware of any upstream apps using the selectable transform currently, which is why I went ahead and posted the change. I'm going off of [patternfly-analytics](https://github.com/redallen/patternfly-analytics/tree/master/stats/2019-11-26) stats 

> How we should introduce this to consumers without "breaking type changes" I'm unsure.

> It's also not good to break Typescript consumers from one minor version to the next.

I'm not sure why we would do either of these, neither is my recommendation. What I can say is that this is one of [several](https://github.com/patternfly/patternfly-react/issues/3223) type changes that need to be made to get the components to a point where they work seamlessly in TypeScript apps that use strict mode. So, it might make sense to bundle all these types of changes into one big breaking release that's specifically for addressing the known type mismatches. I don't have as much an opinion on when and how they are released, just that they are fixed at all. Suggestions on how to proceed are welcome!

packages/patternfly-4/react-table/src/components/Table/Table.tsx
seanforyou23: This is intentional

packages/patternfly-4/react-table/src/components/Table/SelectColumn.tsx
seanforyou23: Yes, this was intentional. Do you happen to know why OnSelect used `React.MouseEvent` while SelectColumn onSelect method was using `React.ChangeEvent<HTMLInputElement>`? It seems like general misalignment but figured it's worth asking about.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSelectableDemo.tsx
```diff
@@ -47,7 +45,7 @@ export class TableSelectableDemo extends React.Component<TableProps, TableState>
     this.onSelect = this.onSelect.bind(this);
   }
 
-  onSelect(event: React.MouseEvent, isSelected: boolean, rowId: number) {
+  onSelect(event: React.FormEvent<HTMLInputElement>, isSelected: boolean, rowId: number) {
```
dlabrecq: We use `onSelect` in Cost Management. Technically, this would be a breaking change for the build.

If you look at the event in the debugger, it is `MouseEvent`.

seanforyou23: Yes I see what you mean, it was generally the purpose of this PR to update the type to FormEvent so that we get correct type checking when attempting to use the `checked` property from the event target. Seems like I'm missing something?

I'm not really sure what's being suggested, can you help me understand? Happy to do whatever is needed to fix the problem.

dlabrecq: When I look at the `onSelect` function, the `event` being passed in is of type `MouseEvent` (i.e., when I select any checkbox). In Cost Management, our handleOnSelect callback also defines this as `event: MouseEvent`. I'm simply suggesting that using `FormEvent` would be a breaking change.

seanforyou23: Thanks for clarifying!

packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSelectableDemo.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Table, TableHeader, TableBody, TableProps, headerCol, ICell, IRow } from '@patternfly/react-table';
 import { Checkbox } from '@patternfly/react-core';
+import '@patternfly/patternfly/utilities/Spacing/spacing.css';
```
redallen: What classes are we using that we need this import for?

seanforyou23: [`pf-u-mb-lg`](https://github.com/patternfly/patternfly-react/pull/3296/files#diff-1719150855fcc83a974c78f98e5b276cR76)

packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableSimpleDemo.tsx
```diff
@@ -17,25 +20,31 @@ export class TableSimpleDemo extends React.Component<TableProps, { columns: (ICe
...
-        [
-          {
-            title: <div>one - 2</div>,
-            props: { title: 'hover title', colSpan: 3 }
-          },
-          'four - 2',
-          'five - 2'
-        ],
-        [
-          'one - 3',
```
redallen: I never liked this API. Thanks for removing examples with it.

packages/react-table/src/components/Table/utils/decorators/headerCol.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { IExtra, IFormatterValueType, ITransform } from '../../Table';
 
 export const headerCol = (id = 'simple-node') => {
+  // tslint:disable-next-line:no-shadowed-variable
```
redallen: I don't think this is needed since `headerCol` and `headerColObj` are two different variable names.

packages/react-table/src/components/Table/utils/decorators/selectable.tsx
```diff
@@ -27,13 +27,12 @@ export const selectable: ITransform = (
...
+   * @param {React.FormEvent} event - React form event
    */
-  function selectClick(event: React.ChangeEvent<HTMLInputElement>) {
-    const selected = rowIndex === undefined ? event.target.checked : rowData && !rowData.selected;
-    // todo: change event type to React.FormEvent<HTMLInputElement> in the future, breaking change a.t.m.
+  function selectClick(event: React.FormEvent<HTMLInputElement>) {
+    const selected = rowIndex === undefined ? event.currentTarget.checked : rowData && !rowData.selected;
     // tslint:disable-next-line:no-unused-expression
-    onSelect && onSelect((event as unknown) as React.MouseEvent, selected, rowId, rowData, extraData);
+    onSelect && onSelect(event, selected, rowId, rowData, extraData);
```
redallen: Yay!!!

PR #3297, opened 2019-11-11
fix(Pagination): fixed previous page navigation issues with 1 row perpage

### LINKED ISSUE #3087: Pagination prev button always disabled  with perPage of 1
I notice that when setting `perPage` to 1 in the Pagination component, it starts to have issues. Seems to be two separate issues. One is that when paging one page at a time, the "previous" button and "jump to beginning" button remain disabled, so the only way you can go backward is to use the numeric input, which isn't an option for the compact version.

The second issue also happens when using a perPage value of 1, it's when you use the "jump to end of set" button and the entire control lands in a disabled state where the only way I can tell to revive the component is to refresh the page.

evwilkin:
**What**:
fixes #3087 

**Additional issues**:


evwilkin:
> What about these lines?
> 
> ```
> <ToggleTemplate
>             firstIndex={firstIndex}
>             lastIndex={lastIndex}
>             itemCount={itemCount}
>             itemsTitle={titles.items}
> />
> ```

@redallen thanks for taking a look - those lines shouldn't affect this navigation issue, unless we wanted to change the text displayed on the page.

PR #3318, opened 2019-11-19
Add displayName to exports wrapped in higher-order components

### LINKED ISSUE #3317: <Button> rendered as <Component> in jest snapshots after updating to the latest PatternFly, breaking our tests
After updating to `@patternfly/react-core@3.120.5` (we previously used `@patternfly/react-core@3.86.1`), every `<Button>` in our snapshot is now `<Component>`, and some other components are affected too.

If this was just an issue with the snapshot itself, it wouldn't be a big deal, but it also affects more functional tests that we have, for example finding a button in the shallow rendered component, simulating a click, and checking the result. Now that the snapshot no longer has `<Button>`, we can't use `.find('Button')` anymore, as it doesn't find anything.

This means that we can't upgrade to the latest PatternFly, and this blocks us from implementing important features and improvements.

mturley:
**What**: Fixes #3317

### DIFF
--- a/packages/patternfly-4/react-core/src/components/AboutModal/__snapshots__/AboutModalBoxCloseButton.test.tsx.snap
+++ b/packages/patternfly-4/react-core/src/components/AboutModal/__snapshots__/AboutModalBoxCloseButton.test.tsx.snap
@@ -4,7 +4,7 @@ exports[`AboutModalBoxCloseButton Test 1`] = `
 <div
   className="pf-c-about-modal-box__close"
 >
-  <Component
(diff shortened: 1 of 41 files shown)


mturley:
For anyone who finds this: the test failures here (wrapper.find() suddenly finding two of a component instead of one, in several places) indicate that this quick-fix for snapshots will break lots of other tests. We're currently discussing other options.

mturley:
Closing this since we'll need to work around the other HOC-related enzyme issues.

PR #3333, opened 2019-11-21
feat(Select): Add the ability to have custom content in the Select Menu

### LINKED ISSUE #3193: Select - add "basic" example
https://github.com/patternfly/patternfly-next/pull/2337

We added a "panel" example in core - https://pf4.patternfly.org/documentation/core/components/select#panel

This will mimic the "basic" dropdown react example - https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/dropdown#basic

tlabaj:
**What**: Closes #3193

**Additional issues**:


packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -119,7 +121,8 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
     onClear: (_e: React.MouseEvent) => undefined as void,
     onCreateOption: (_newOptionValue: string) => undefined as void,
     toggleIcon: null as React.ReactElement,
-    onFilter: null
+    onFilter: null,
+    customContent: null as React.ReactNode
```
dlabaj: This needs to be defined as ReactNode?

packages/patternfly-4/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -45,21 +47,23 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
...
     maxHeight: '',
     sendRef: Function.prototype,
-    keyHandler: Function.prototype
+    keyHandler: Function.prototype,
+    isCustomContent: false
   };
 
   extendChildren() {
     const { children, isGrouped } = this.props;
+    const childrenArray = children as React.ReactElement[];
```
dlabaj: Can you do this instead const childrenArray: React.ReactElement[] = children as React.ReactElement[];    ... so it's not implicit? Or will this not work with here?

packages/patternfly-4/react-core/src/components/Select/Select.tsx
tlabaj: I was just following what was done for toggleIcon.  I can change it.

dlabaj: I think if we are going to change it would should have it as a coding standard.  It's fine to leave it ... however we might want to spend some time looking into our recommended coding standards so we are coding consistently through out the project.  Maybe create an issue (chore) in our backlog to update coding standards for typescript and fix the code so we are consistent.

Issue #3338, opened 2019-11-21
Options Menu - support for groups and titles


mcoker:
This is currently an open PR but will be merged after the code freeze.

https://github.com/patternfly/patternfly-next/pull/2403

Need to add support for and an example of groups & titles

https://patternfly-next-pr-2403.surge.sh/documentation/core/components/optionsmenu#with-groups

tlabaj:
closed by #3360

Issue #2403, opened 2019-06-28
GridItem warnings


dlabrecq:
I updated to the latest pre-release packages and encountering a bunch of new warnings with GridItem.

Note that I've also updated React v16.3.9 and React.dom v16.0.5 to use v16.8.0 -- required for PatternFly's use of Tippy.

<img width="267" alt="Screen Shot 2019-06-28 at 3 30 25 PM" src="https://user-images.githubusercontent.com/17481322/60366498-b0757d00-99b9-11e9-8007-913303ce595f.png">

<img width="1186" alt="Screen Shot 2019-06-28 at 3 36 37 PM" src="https://user-images.githubusercontent.com/17481322/60366923-f0892f80-99ba-11e9-96ec-3eda6fb15c6e.png">
<img width="1182" alt="Screen Shot 2019-06-28 at 3 36 50 PM" src="https://user-images.githubusercontent.com/17481322/60366924-f0892f80-99ba-11e9-89cd-9d50ca4f62d1.png">
<img width="1184" alt="Screen Shot 2019-06-28 at 3 37 02 PM" src="https://user-images.githubusercontent.com/17481322/60366925-f0892f80-99ba-11e9-8f35-f49160b8cc04.png">
<img width="1184" alt="Screen Shot 2019-06-28 at 3 37 13 PM" src="https://user-images.githubusercontent.com/17481322/60366926-f0892f80-99ba-11e9-88bf-18b03df19b91.png">
<img width="1177" alt="Screen Shot 2019-06-28 at 3 37 41 PM" src="https://user-images.githubusercontent.com/17481322/60366927-f0892f80-99ba-11e9-8ed5-571148fa3168.png">

PR #3360, opened 2019-11-27
fix(Options menu): added support for groups and titles

nicolethoen:
Addresses #3338 

Follow up to core PR [#2403](https://github.com/patternfly/patternfly-next/pull/2403)


packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export interface OptionsMenuGroupProps extends React.HTMLProps<HTMLElement> {
+  /** Content to be rendered in the Options menu items component */
+  children?: React.ReactNode;
+  /** Classes applied to root element of the Options menu items group */
+  className?: string;
+  /** Provides an accessible name for the Options menu items group */
+  ariaLabel?: string;
+  /** TODO */
+  groupTitle?: string | React.ReactNode;
```
mcoker: should we change this to `label` to match the app launcher and dropdown components?

packages/patternfly-4/react-core/src/components/OptionsMenu/examples/OptionsMenu.md
```diff
@@ -457,3 +459,63 @@ class PlainWithText extends React.Component {
...
+    const toggle = <OptionsMenuToggle onToggle={this.onToggle} toggleTemplate={toggleTemplateText} />
+    
+    return (
+      <OptionsMenu 
+        id="options-menu-align-right-example" 
+        position={OptionsMenuPosition.right} 
+        menuItems={menuGroups} 
+        toggle={toggle} 
+        isOpen={isOpen} 
+        isGrouped />
```
mcoker: if `isGrouped`, the `.pf-c-options-menu__menu` should be a `<div>` instead of a `<ul>`.  The dropdown and app launcher work that way if it helps to use as a reference.

packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
nicolethoen: i think that would be a breaking change at this point...

packages/patternfly-4/react-core/src/components/OptionsMenu/examples/OptionsMenu.md
nicolethoen: hm... the menu is built using the dropdown pieces so i'm surprised that didn't behavior come for free. Nice catch. I'll look into it

packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
tlabaj: Isn't this an internal Class?  Changing the name should be ok if that that is so.  We already changed the name of the Class itself correct?

packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/OptionsMenu/options-menu';
+import { css } from '@patternfly/react-styles';
+
+export interface OptionsMenuGroupProps extends React.HTMLProps<HTMLElement> {
+  /** Content to be rendered in the Options menu items component */
```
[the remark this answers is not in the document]
nicolethoen: okay. I guess in general we don't consider component names to be proper nouns?

packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
nicolethoen: It's not an internal class, so I actually might need to change the name back.

packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children?: React.ReactNode;
+  /** Classes applied to root element of the Options menu items group */
+  className?: string;
+  /** Provides an accessible name for the Options menu items group */
+  ariaLabel?: string;
+  /** TODO */
```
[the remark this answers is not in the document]
nicolethoen: oops

packages/patternfly-4/react-core/src/components/OptionsMenu/OptionsMenuGroup.tsx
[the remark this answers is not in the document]
tlabaj: I think think so?  We can double check with Abi.  I know I have been asked to change it past...

PR #3363, opened 2019-12-01
feat(list): add the ordered variant

### LINKED ISSUE #3356: PF4 - List - no option for ordered list.
PF4 List component has no option to display ordered lists.

List component has no option to choose `<ol>` list:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/List/List.tsx#L24

Design include ordered list:
https://pf4.patternfly.org/documentation/core/components/list#ordered

cc:// @matthewcarleton

boaz0:
**What**:

Closes #3356

### DIFF
--- a/packages/patternfly-4/react-core/src/components/List/List.tsx
+++ b/packages/patternfly-4/react-core/src/components/List/List.tsx
@@ -1,27 +1,53 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/List/list';
 import { css, getModifier } from '@patternfly/react-styles';
+import { Omit } from '../../helpers/typeUtils'
 
-export enum ListVariant {
-  inline = 'inline'
+export enum OrderType {
+  number = '1',
+  lowercaseLetter = 'a',
+  uppercaseLetter = 'A',
+  lowercaseRomanNumber = 'i',
+  uppercaseRomanNumber = 'I',
 }
 
-export interface ListProps extends React.HTMLProps<HTMLUListElement> {
+export enum ListVariant {
+  inline = 'inline',
+}
+
+export enum ListComponent {
+  ol = 'ol',
+  ul = 'ul',
+}
+
+export interface ListProps extends Omit<React.HTMLProps<HTMLUListElement | HTMLOListElement>, 'type'> {
   /** Anything that can be rendered inside of the list */
   children?: React.ReactNode;
-  /** Additional classes added to the list. */
+  /** Additional classes added to the list */
   className?: string;
   /** Adds list variant styles */
-  variant?: 'inline';
+  variant?: ListVariant.inline;
+  /** Sets the way items are numbered if variant is set to ordered */
+  type?: OrderType;
+  component?: 'ol' | 'ul';
 }
 
 export const List: React.FunctionComponent<ListProps> = ({
   className = '',
   children = null,
   variant = null,
+  type = OrderType.number,
+  ref = null,
+  component = ListComponent.ul,
   ...props
-}) => (
-  <ul {...props} className={css(styles.list, variant && getModifier(styles.modifiers, variant), className)}>
-    {children}
-  </ul>
-);
+}) => {
+  return component === ListComponent.ol ? (
+    <ol ref={ref as React.LegacyRef<HTMLOListElement>} type={type} {...props} className={css(styles.list, variant && getModifier(styles.modifiers, variant), className)}>
+      {children}
+    </ol>
+  ) : (
+      <ul ref={ref as React.LegacyRef<HTMLUListElement>} {...props} className={css(styles.list, variant && getModifier(styles.modifiers, variant), className)}>
+        {children}
+      </ul>
+    );
+}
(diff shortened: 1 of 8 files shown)


packages/patternfly-4/react-core/src/components/List/List.tsx
```diff
@@ -2,26 +2,40 @@ import * as React from 'react';
...
+  variant = ListVariant.unordered,
+  type = null,
   ...props
-}) => (
-  <ul {...props} className={css(styles.list, variant && getModifier(styles.modifiers, variant), className)}>
-    {children}
-  </ul>
-);
+}) => {
+  if (variant === ListVariant.ordered) {
```
yaacov: maybe use ternary operator here ?

packages/patternfly-4/react-core/src/components/List/List.tsx
```diff
@@ -2,26 +2,40 @@ import * as React from 'react';
...
+
 export enum ListVariant {
-  inline = 'inline'
+  inline = 'inline',
+  ordered = 'ordered',
+  unordered = 'unordered',
 }
 
-export interface ListProps extends React.HTMLProps<HTMLUListElement> {
+export interface ListProps extends React.HTMLProps<HTMLOListElement | HTMLOListElement> {
```
yaacov: `<HTMLOListElement | HTMLOListElement>` typo ?
should be:
`<HTMLUListElement | HTMLOListElement>`

packages/patternfly-4/react-core/src/components/List/List.tsx
```diff
@@ -2,26 +2,40 @@ import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/List/list';
 import { css, getModifier } from '@patternfly/react-styles';
 
+export type OrderTypes = '1' | 'a' | 'A' | 'i' | 'I';
+
 export enum ListVariant {
-  inline = 'inline'
+  inline = 'inline',
+  ordered = 'ordered',
+  unordered = 'unordered',
```
yaacov: do we need this variant ?

mcoker:
Can `variant` have more than 1 `ListVariant`? For example, something like `variant={ListVariant.inline, ListVariant.ordered}`?

boaz0:
@mcoker - I believe we can do that but I wonder why `inline` should be a variant in that case. Why not using `isInline` flag to add that modifier?

mcoker:
@boaz0 I'm not sure why we didn't use `isInline`, but ultimately it should be valid to create either an `<ol>` or `<ul>` with `.pf-m-inline`.

Not sure if it helps, but here are some other components that are similar where we allow the user to define the type of element(s) to be used. There are probably more I'm not thinking of.

* `<Button>`
  * uses `<Button component="a">` to render as an `<a>` instead of a `<button>`
* <Accordion>`
  * uses `<Accordion asDefinitionList>` to render as a `<dl>` with `<dt>`'s as titles and `<dd>`'s as the body, instead of `<div>`'s with heading elements as the titles. 
* `<Tabs>`
  * uses `variant={TabsVariant.nav}` to render as a `<nav>` instead of a `<div>`

packages/patternfly-4/react-core/src/components/List/List.tsx
```diff
@@ -1,27 +1,50 @@
...
+  component?: ListVariant.ordered | ListVariant.unordered;
 }
 
 export const List: React.FunctionComponent<ListProps> = ({
   className = '',
   children = null,
   variant = null,
+  type = OrderType.number,
+  ref = null,
+  component = ListVariant.unordered,
```
yaacov: using `ListVariant` for the `component` prop type ?

yaacov: sorry, didn't notice this is done in Text component too, ok by me.

boaz0: I can change that to `ListComponent` if you like, unless you have a better name.
I am happy to hear any idea. :D

yaacov: I like the `isSomething` pattern:
for example: leave the variant like it was in the original component, and add a `isOrdered` prop ...

But since I saw that `Text` is using the `TextVariant` for it's `component` view, I'm fine with continuing the same pattern.

packages/patternfly-4/react-core/src/components/List/List.tsx
```diff
@@ -1,27 +1,50 @@
...
+  lowercaseLetter = 'a',
+  uppercaseLetter = 'A',
+  lowercaseRomanNumber = 'i',
+  uppercaseRomanNumber = 'I',
+}
 
 export enum ListVariant {
-  inline = 'inline'
+  inline = 'inline',
+  ordered = 'ol',
```
tlabaj: You want to remove the `ul` and `ol` variants here and but them in a separate enum.

packages/patternfly-4/react-core/src/components/List/List.tsx
```diff
@@ -1,27 +1,53 @@
...
+}
 
 export enum ListVariant {
-  inline = 'inline'
+  inline = 'inline',
 }
 
-export interface ListProps extends React.HTMLProps<HTMLUListElement> {
+export enum ListComponent {
+  ordered = 'ol',
```
tlabaj: Looks good.  My preference would be to name them by the component name.
`export enum ListComponent {
    ordered = 'ol',
    unordered = 'ul',
}`

Issue #3252, opened 2019-11-01
Popover `hideOnOutsideClick` does not appear to work


rhamilto:
![ekTvTcf0vU](https://user-images.githubusercontent.com/895728/68031484-1e94be00-fc92-11e9-8a31-0ab4db47377a.gif)

rachael-phillips:
Hi @rhamilto I've seen this, and I will be discussing this with the team tomorrow.

PR #3370, opened 2019-12-03
Update tippy.js dependency

nicolethoen:
Addresses: [#3252 ](https://github.com/patternfly/patternfly-react/issues/3252)

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
+++ b/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
@@ -95,7 +95,7 @@ export interface PopoverProps {
   /** z-index of the popover */
   zIndex?: number;
   /** additional Props to pass through to tippy.js */
-  tippyProps?: TippyProps;
+  tippyProps?: object;
 }
 
 export interface PopoverState {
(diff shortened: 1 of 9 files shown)


packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
```diff
@@ -95,7 +95,7 @@ export interface PopoverProps {
   /** z-index of the popover */
   zIndex?: number;
   /** additional Props to pass through to tippy.js */
-  tippyProps?: TippyProps;
+  tippyProps?: object;
```
redallen: Are Tippy's types no longer working? Avoiding `object` would be nice for consumers.

nicolethoen: there are like 43 required TippyProps. So in the tests, for example, we tried only passing two tippy props, we get a type error.

redallen: Does `Partial<TippyProps>` work?

redallen:
Thanks Nicole!

PR #3371, opened 2019-12-03
Application launcher enhancements

### LINKED ISSUE #3227: Application launcher enhancements
This is a React implementation follow-up to Core issue https://github.com/patternfly/patternfly-next/issues/2365. This React issue is on hold until https://github.com/patternfly/patternfly-next/issues/2365 has been closed.

kmcfaul:
**What**: Closes #3227

### DIFF
--- a/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
+++ b/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
+import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/AppLauncher/app-launcher';
+import formStyles from '@patternfly/react-styles/css/components/FormControl/form-control';
 import { ThIcon } from '@patternfly/react-icons';
 import { DropdownDirection, DropdownPosition, DropdownToggle, DropdownContext } from '../Dropdown';
 import { DropdownWithContext } from '../Dropdown/DropdownWithContext';
@@ -40,6 +42,12 @@ export interface ApplicationLauncherProps extends React.HTMLProps<HTMLDivElement
   favorites?: string[];
   /** Enables favorites. Callback called when an ApplicationLauncherItem's favorite button is clicked */
   onFavorite?(itemId: string, isFavorite: boolean): void;
+  /** Enables search. Callback called when text input is entered into search box */
+  onSearch?(textInput: string): void;
+  /** Placeholder text for search input */
+  searchPlaceholderText?: string;
+  /** Additional properties for search input */
+  searchProps?: any;
 }
 
 export const ApplicationLauncherContext = React.createContext({ onFavorite: Function.prototype });
@@ -58,7 +66,24 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
     onToggle: (_value: boolean): any => undefined,
     'aria-label': 'Application launcher',
     isGrouped: false,
-    toggleIcon: <ThIcon />
+    toggleIcon: <ThIcon />,
+    searchPlaceholderText: 'Filter by name...'
+  };
+
+  createSearchBox = () => {
+    const { onSearch, searchPlaceholderText, searchProps } = this.props;
+
+    return (
+      <div className={css(styles.appLauncherMenuSearch)}>
+        <input
+          type="search"
+          className={css(formStyles.formControl)}
+          placeholder={searchPlaceholderText}
+          onChange={e => onSearch(e.target.value)}
+          {...searchProps}
+        ></input>
+      </div>
+    );
   };
 
   createRenderableFavorites = () => {
@@ -113,6 +138,9 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
       dropdownItems,
       favorites,
       onFavorite,
+      onSearch,
+      searchPlaceholderText,
+      searchProps,
       items,
       ref,
       ...props
@@ -131,6 +159,10 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
       renderableItems = favoritesGroup.concat(this.extendItemsWithFavorite());
     } else {
       renderableItems = items;
+    }
+
+    if (onSearch) {
+      renderableItems = [this.createSearchBox(), ...renderableItems];
     }
 
     return (
--- a/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
+++ b/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
@@ -83,14 +83,18 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
     const { onSearch, searchPlaceholderText, searchProps } = this.props;
 
     return (
-      <div className={css(styles.appLauncherMenuSearch)}>
-        <input
-          type="search"
-          className={css(formStyles.formControl)}
-          placeholder={searchPlaceholderText}
-          onChange={e => onSearch(e.target.value)}
-          {...searchProps}
-        ></input>
+      <div key="search" className={css(styles.appLauncherMenuSearch)}>
+        <ApplicationLauncherItem
+          customChild={
+            <input
+              type="search"
+              className={css(formStyles.formControl)}
+              placeholder={searchPlaceholderText}
+              onChange={e => onSearch(e.target.value)}
+              {...searchProps}
+            ></input>
+          }
+        ></ApplicationLauncherItem>
       </div>
     );
   };
@@ -100,17 +104,15 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
     if (isGrouped) {
       let favoriteItems: React.ReactNode[] = [];
       (items as React.ReactElement[]).forEach(group =>
-        favoriteItems.push(
-          (group.props.children as React.ReactElement[])
-            .filter(item => favorites.includes(item.props.id))
-            .map(item => React.cloneElement(item, { isFavorite: true }))
-        )
+        (group.props.children as React.ReactElement[])
+          .filter(item => favorites.includes(item.props.id))
+          .map(item => favoriteItems.push(React.cloneElement(item, { isFavorite: true, enterTriggersArrowDown: true })))
       );
       return favoriteItems;
     }
     return (items as React.ReactElement[])
       .filter(item => favorites.includes(item.props.id))
-      .map(item => React.cloneElement(item, { isFavorite: true }));
+      .map(item => React.cloneElement(item, { isFavorite: true, enterTriggersArrowDown: true }));
   };
 
   extendItemsWithFavorite = () => {
@@ -161,14 +163,18 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
 
     if (onFavorite) {
       let favoritesGroup: React.ReactNode[] = [];
-      if (favorites.length > 0)
+      let renderableFavorites: React.ReactNode[] = [];
+      if (favorites.length > 0) {
+        renderableFavorites = this.createRenderableFavorites();
         favoritesGroup = [
           <ApplicationLauncherGroup key="favorites" label={favoritesLabel}>
-            {this.createRenderableFavorites()}
+            {renderableFavorites}
             <ApplicationLauncherSeparator key="separator" />
           </ApplicationLauncherGroup>
         ];
-      renderableItems = favoritesGroup.concat(this.extendItemsWithFavorite());
+      }
+      if (renderableFavorites.length > 0) renderableItems = favoritesGroup.concat(this.extendItemsWithFavorite());
+      else renderableItems = this.extendItemsWithFavorite();
     } else {
       renderableItems = items;
     }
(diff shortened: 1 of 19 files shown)


kmcfaul:
@mcarrano Regarding the search/filtering enhancement, what is the expected behavior when no results are found? What is the expected behavior when no items in a group are found? Should filtering apply to the favorites list?

Currently, groups are still displayed even when no results are found within them, and the favorites list is also filtered as a side effect of passing in the filtered results.

mcarrano:
> Regarding the search/filtering enhancement, what is the expected behavior when no results are found? What is the expected behavior when no items in a group are found? Should filtering apply to the favorites list? Currently, groups are still displayed even when no results are found within them, and the favorites list is also filtered as a side effect of passing in the filtered results.

@kmcfaul Thanks for raising that question. It was definitely missing from the design. I want to loop in @maryshak1996 as she was the primary designer. I'm thinking that maybe we just insert a string "No items found" under each category header with no hits, but am open to ideas. The other thing that I wondered about was the behavior of refreshing the menu when things are added and removed from the Favorites set. While this is implemented as designed, it does feel disorienting to me to have the menu refresh (and items move) when the user toggles a favorites tag. I'm wondering if the list of Favorites should only include favorites as loaded when the menu was opened and not update dynamically. Mary, Interested in your thoughts.

maryshak1996:
@kmcfaul @mcarrano those are really good points to bring up
In regard to having newly-favorited items getting added to the top 'Favorites' section, I don't personally find it disorienting, but I also don't have very strong feelings around it -- @mcarrano I'm open to hearing any alternatives!
And, to the question about the empty state, I would agree with Matt's thought for sure. @kmcfaul I mocked some references here: 
**Search finds no application names**
image
**Search finds applications in the same group**
image
**Search finds applications in multiple groups**
image
**Note: I saw in your implementation that the search retains the groupings even when there's nothing in the group -- is this something that can change (screenshot of what I'm referring to below), so that there isn't so many unneeded lines?
Screen Shot 2019-12-03 at 3 47 42 PM

mcarrano:
Thanks @maryshak1996 . I don't feel that strongly about the favoriting behavior either. @chart @bmignano can you also take a look at the behavior for toggling favorites off and on and let us know if this is what you expected? Agree that clarifying the filtering behavior is more critical.

bmignano:
This favoriting behavior makes sense to me – I don't think it's too disorienting since the movement is a direct result of an action the user takes. If we waited to load the favorites until the menu was closed and reopened I worry it'd be more confusing as the feedback would be delayed upon clicking a star.
As for filtering, I'm not sure we need to maintain headers when no items in specific groups are found. I think Mary's screenshots above are a good approach to filtering - only show the groups when an item matches the filter, and if no items match, show "No applications found".

jessiehuff:
This looks awesome! It's a cool addition to our app launcher. :) I know you're still working out specifics, but I just wanted to put a few things on your radar. 

It looks like the input field isn't accessible through keyboard interaction and the favorites option isn't accessible by keyboard or screen reader. I'll try to research suggestions on the best way to handle it, but I imagine it would work similar to other dropdown components. 

Currently I believe our typeahead select puts focus on the input field first instead of the first menu item, and when your focus is in the menu, shift + tab takes you to the input field. (However, I know that we've been in discussion about the best way to handle typeahead select so I'm interested in others' opinions here.)

In terms of the favorites option, I would think it would be treated similarly to other menu items in using the arrow keys, but I'd have to do some research to see if there is a recommended way for this use case. Interested to hear your thoughts on that interaction @mcarrano. 

For the screen reader, I'll have to play with it a bit to see what would allow the favorites icon to be announced. I imagine it needs some kind of discernible text like an aria-label so the screen reader can see it, but I'd have to test to verify that theory.

Thanks for working on this, Katie!!

mcoker:
> For the screen reader, I'll have to play with it a bit to see what would allow the favorites icon to be announced. I imagine it needs some kind of discernible text like an aria-label so the screen reader can see it, but I'd have to test to verify that theory.

FWIW, I looked at google and they change the `aria-label` between "starred" and "not starred". Not sure if that's a good way to announce, or to announce it more as the action you'll perform, like "favorite" and "unfavorite"?

kmcfaul:
I'll add in the aria-label swapping and review some of the keyboard behavior. As ApplicationLauncher wraps Dropdown it should be there, but I think the search input might need to be wrapped in an ApplicationLauncherItem to get properly picked up by Dropdown's keyboard interactions.

I'm also working on refining the empty state behavior, removing the last group's separator and removing groups that contain no filtered items. @mcoker @maryshak1996 What is the structure of the "No results found" visual? I currently have a placeholder in an unstyled <div>.

kmcfaul:
@jessiehuff After experimenting with keyboard behavior, the desired interactions may be more complicated to implement. Since ApplicationLauncher wraps Dropdown, it has Dropdown's behavior which doesn't quite work for this use case as we are passing in custom elements and expanding DropdownItem beyond it's original structure.

The search input can be targeted using arrow keys if it is wrapped in an ApplicationLauncherItem, which gets the context used to build out Dropdown's ref collection, but unfortunately loses its functionality as an input field. I haven't found a clean way to insert a custom element into Dropdown's ref collection yet, but am trying a couple different things.

The favorites button exists within the DropdownItem `li` - where the ref sits for the keyboard to target. However there isn't a way yet to tab into or select the secondary element, and it defaults to targeting the first element, which is the button/anchor that contains the text. This issue also extends to the external link feature, which sits inside the primary element but may point elsewhere. How do we want to handle this? It might be possible to build out a ref collection of each targetable element instead of just the item itself, and then arrow keys would take the user through each item.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx
```diff
@@ -25,29 +27,57 @@ export interface ApplicationLauncherItemProps {
...
-    >
-      {children && <ApplicationLauncherContent>{children}</ApplicationLauncherContent>}
-    </DropdownItem>
+    <ApplicationLauncherContext.Consumer>
+      {({ onFavorite }) => (
+        <DropdownItem
+          id={id}
+          component={component}
+          href={href || null}
+          className={css(isExternal && styles.modifiers.link, className)}
```
mcoker: Oh sorry this probably wasn't clear from the implementation. Prior to this PR, the app launcher item (`.pf-c-app-launcher__menu-item`) is either a button or a link (some sort of interactive element you click on and it does the thing). In this PR, we introduce the ability to have multiple interactive things (items) inside of a single main row/item in the menu. To enable this (for example, when an item in the menu is favoritable, there are 2 individual interactive items in the main menu item), we allow this by wrapping the individual items with `.pf-c-app-launcher__menu-wrapper`, then placing the individual items in the wrapper. Those individual items will need classes to handle the spacing. The main item (the button/link) will use the modifier `.pf-c-app-launcher__menu-item.pf-m-link`, and any actions to the right of it will use `.pf-c-app-launcher__menu-item.pf-m-action`.

Which is a long way of saying that all of the buttons/links in a favoritable item need `styles.modifiers.link`, not just external links.

mcoker:
> What is the structure of the "No results found" visual? I currently have a placeholder in an unstyled div.

@maryshak1996 should we just use an empty state there?

kmcfaul:
Updated with current 'no results' div and logic to remove groups/separators when a middle group happens to be the last group when filtered.

Added a small fix for keyboard to prevent it from failing when there is an undefined ref in the ref collection (as is the case when the search input div is present).

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
```diff
@@ -62,47 +137,80 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
...
       ref,
       ...props
     } = this.props;
+    let renderableItems: React.ReactNode[] = [];
+
+    if (onFavorite) {
+      let favoritesGroup: React.ReactNode[] = [];
+      if (favorites.length > 0)
+        favoritesGroup = [
+          <ApplicationLauncherGroup key="favorites" label="Favorites">
```
tlabaj: This label should be configurable.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
```diff
@@ -62,47 +137,80 @@ export class ApplicationLauncher extends React.Component<ApplicationLauncherProp
...
+          </ApplicationLauncherGroup>
+        ];
+      renderableItems = favoritesGroup.concat(this.extendItemsWithFavorite());
+    } else {
+      renderableItems = items;
+    }
+    if (items.length === 0 && dropdownItems.length === 0) {
+      renderableItems = [
+        <ApplicationLauncherGroup key="no-results-group">
+          <ApplicationLauncherItem key="no-results">No results found</ApplicationLauncherItem>
```
tlabaj: This text should also be configurable.

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx
```diff
@@ -25,29 +27,58 @@ export interface ApplicationLauncherItemProps {
...
+                className={css(styles.appLauncherMenuItem, styles.modifiers.action)}
+                onClick={() => {
+                  onFavorite(id, isFavorite);
+                }}
+              >
+                <StarIcon />
+              </button>
+            )
+          })}
+          {...(isFavorite !== null && { 'aria-label': isFavorite ? 'starred' : 'not starred' })}
```
tlabaj: should the consumer be able to change the aria-lable?  @jessiehuff

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -23,6 +23,8 @@ export interface DropdownItemProps extends InternalDropdownItemProps {
   tooltip?: React.ReactNode;
   /** Additional tooltip props forwarded to the Tooltip component */
   tooltipProps?: any;
+  /** Additional node to include alongside item within the <li> */
+  additionalChildren?: React.ReactNode;
```
tlabaj: Do we really need this prop?  I am not understanding why AplicationLauncher can't create a single child for the dropdownItem.  Is it because it would be 2 fragments and not a single node?  If so, can we wrap it in `<React.Fragment>`?

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx
tlabaj: I saw @mcoker's comment above.  We still need to be able to configure these strings for localization purposes.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.tsx
kmcfaul: This is needed to match the html structure in core, and prevents nesting buttons inside anchors/buttons (if it was a child of the dropdown).

kmcfaul:
Updated with pr feedback and added cypress/TS demo

kmcfaul:
Updated with keyboard navigation changes to support left/right on favorites menu. Should not be breaking because keyboard handling is internal. 

I've tested Dropdown, Select, OptionsMenu as well to ensure that keyboard interaction still works.

kmcfaul:
@maryshak1996 @mcoker Any update on the empty state visual?

maryshak1996:
It looks great to me, @kmcfaul :)

packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx
```diff
@@ -25,29 +27,71 @@ export interface ApplicationLauncherItemProps {
    * } />
    */
   component?: React.ReactNode;
+  /** Flag indicating if the item is favorited */
+  isFavorite?: boolean;
+  /** Aria label text for favoritable button when favorited */
+  ariaIsFavorite?: string;
```
tlabaj: Maybe append `Label` to the end of these so it is more obvious that it is a label and not a boolean.

mcoker:
Overall this looks great to me @kmcfaul. I noticed there is a bit of a structure difference in the HTML, mainly in the `__menu-item` class. In react, there is a `__menu-item-text` in some of the app launcher examples. I don't see any style in core for the `__menu-item-text` element. Do you know what that is for? 

And `__menu-item-external-icon` is a `<span>` instead of a class on the icon itself in core.

I don't think either of these are breaking but they don't quite match. 

Also looks like the external icon in the [sections and icons](https://patternfly-react-pr-3371.surge.sh/patternfly-4/documentation/react/components/applicationlauncher#with-sections-and-icons) example is off. Looks like this is because the `.pf-m-link` class is being applied to those links. For reference, the `.pf-m-link` class should only apply to the main link/action in a `__menu-wrapper` element as a means to space the link/action from whatever may be placed beside it.

Screen Shot 2019-12-12 at 9 45 03 AM

I would also like to fix this spacing between the search divider and the first element, and have opened a core issue https://github.com/patternfly/patternfly-next/issues/2514

Screen Shot 2019-12-12 at 9 31 49 AM
Screen Shot 2019-12-12 at 9 41 13 AM
Screen Shot 2019-12-12 at 9 41 19 AM

kmcfaul:
@mcoker 

I'm not sure what `__menu-item-text` is used for. I can remove it if it doesn't do anything.

`__menu-item-external-icon` only works when it's applied to the outer span from what I tested, not when its applied to the icon itself.

Fixed the application of the `.pf-m-link` and `.pf-m-external`.

The lack of space could be because the item is still wrapped in a group, but the group doesn't have a title.

mcoker:
@kmcfaul I'm not seeing the external icon in the menus now. It has to do with the difference between the SVG icons used in react and the font awesome icons in core. We can make an update in core so that the icon works with your existing code, but it will require a fix after our code freeze. 

Personally I think applying the class to the `<span>` like you had it is a better approach, as it allows for more flexible spacing and works with whatever kind of icon or image we want to use in it. Then we can update the examples in core to match. The other icons in the menu are all placed inside of an element that handles the spacing, too.

PR #3378, opened 2019-12-05
improvement(CatalogTile): Removed truncation fade and fixed length

rebeccaalpert:
Removed truncation fade per OpenShift request and added CSS/SCSS to deal with the description length (limit to 3 lines maximum).

### DIFF
--- a/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
+++ b/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
@@ -68,7 +68,13 @@
   margin-top: 0;
 
   .truncated {
+    display: -webkit-box;
     overflow: hidden;
-    position: relative;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 3;
+  }
+
+  .has-footer {
+    -webkit-line-clamp: 2;
   }
 }
--- a/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
+++ b/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
@@ -1,10 +1,3 @@
-@mixin co-line-clamp($lines: 3) {
-  display: -webkit-box;
-  overflow: hidden;
-  -webkit-box-orient: vertical;
-  -webkit-line-clamp: $lines;
-}
-
 .catalog-tile-pf {
   &.featured {
     border-top: 2px solid $color-pf-blue-300;
@@ -66,15 +59,7 @@
 
 .catalog-tile-pf-description {
   margin-top: 0;
-
-  .truncated {
-    display: -webkit-box;
-    overflow: hidden;
-    -webkit-box-orient: vertical;
-    -webkit-line-clamp: 3;
-  }
-
-  .has-footer {
-    -webkit-line-clamp: 2;
-  }
+  display: -webkit-box;
+  overflow: hidden;
+  -webkit-box-orient: vertical;
 }
--- a/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
+++ b/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
@@ -1,3 +1,10 @@
+@mixin co-line-clamp($lines: 3) {
+  display: -webkit-box;
+  overflow: hidden;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: $lines;
+}
+
 .catalog-tile-pf {
   &.featured {
     border-top: 2px solid $color-pf-blue-300;
@@ -63,16 +70,5 @@
   .truncated {
     overflow: hidden;
     position: relative;
-
-    &::after {
-      background: linear-gradient(to right, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 75%);
-      bottom: 0;
-      color: transparent;
-      content: ".";
-      position: absolute;
-      right: 0;
-      width: 50%;
-      text-align: right;
-    }
   }
 }
--- a/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
+++ b/packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
@@ -59,7 +59,15 @@
 
 .catalog-tile-pf-description {
   margin-top: 0;
-  display: -webkit-box;
-  overflow: hidden;
-  -webkit-box-orient: vertical;
+
+  span {
+    display: -webkit-box;
+    overflow: hidden;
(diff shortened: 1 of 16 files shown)


packages/patternfly-4/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
```diff
@@ -1,3 +1,10 @@
+@mixin co-line-clamp($lines: 3) {
```
jeff-phillips-18: co?

rebeccaalpert:
So the reason we have so much empty vertical space in OpenShift right now is to accommodate the absolute longest tile. This new change will keep everything a more consistent length so we don't need as much space (we will potentially need some vertical space given that titles and providers vary so much in length -- that's something we could potentially do this with as well if Sam is game for it).

jeff-phillips-18:
This is patternfly-react, not specific to Console, right?

rebeccaalpert:
Yes, this is PatternFly-React rather than a console-specific change. We've been using OpenShift as a staging area to see how this component behaves with more variable data.

packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
```diff
@@ -143,7 +143,7 @@ export class CatalogTile extends React.Component<CatalogTileProps> {
         </CardHeader>
         <CardBody className="catalog-tile-pf-body">
           <div className="catalog-tile-pf-description">
-            <span className={classNames({'truncated': isTruncated})}>
+            <span className={classNames('co-line-clamp', {'truncated': isTruncated})}>
```
jeff-phillips-18: If we are auto-truncating via the `co-line-clamp` then we should not be truncating the string at all and no need for the `truncated` class. 

We should remove the truncation properties (and update the text in the example tiles)

rebeccaalpert:
The original catalog tiles had a truncation function and allowed you to pass one in. Removing it would be a breaking change; is that what you want to do?

I worked with Joe on a refs-based solution. We have to do more complicated stuff here since it's not a fixed width and height container, but it seems to work. I know you're on vacation, but let us know your thoughts when you get back.

rebeccaalpert:
Thanks Joe.

packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
```diff
@@ -59,11 +62,43 @@ export class CatalogTile extends React.Component<CatalogTileProps> {
...
+  state = {
+    heightStyle: {}
+  };
+
+  descFullHeight = null as any;
+  descLineHeight = null as any;
+  footerFullHeight = null as any;
+  cardFullHeight = null as any;
+  headerFullHeight = null as any;
+  headFullHeight = null as any;
```
jeff-phillips-18: Shouldn't the all be numbers?

packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
```diff
@@ -59,11 +62,43 @@ export class CatalogTile extends React.Component<CatalogTileProps> {
...
+
+  componentDidMount() {
+    this.computeDescHeight();
+  }
+
+  computeDescHeight() {
+    let heightStyle = { maxHeight: null as string, WebkitLineClamp: null as string };
+    if (this.cardFullHeight && this.headFullHeight && this.headerFullHeight && this.descFullHeight) {
```
jeff-phillips-18: Could the heights be `0`? If so the computation should still occur.

rebeccaalpert: Fixed.

packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
rebeccaalpert: All set.

packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/examples/CatalogTile.md
```diff
@@ -126,7 +100,7 @@ Link = () => (
       ]}
       href="http://patternfly.org/v4"
       title="Patternfly-React"
-      vendor={<React.Fragment>provided by <a href="http://redhat.com">Red Hat</a></React.Fragment>}
+      vendor="provided by Red Hat"
       description={
         'This is a very long description that should be truncated after 112 characters. ' +
         '112 is the default but can be overridden if need be. You can also provide a custom truncation function ' +
```
jeff-phillips-18: I would suggest setting `maxDescriptionLength` to -1 and updating the text that references the 112 default in all the examples. IMO, we should never truncate the text since we are using the visual truncation anyhow to avoid the truncated last line not using the full width of the tile.

We should deprecate the text truncation and remove it all together in the next release.

packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
```diff
@@ -59,11 +62,45 @@ export class CatalogTile extends React.Component<CatalogTileProps> {
...
+  cardFullHeight = null as number;
+  headerFullHeight = null as number;
+  headFullHeight = null as number;
+
+  componentDidMount() {
+    this.computeDescHeight();
+  }
+
+  computeDescHeight() {
+    let heightStyle = { maxHeight: null as string, WebkitLineClamp: null as string };
```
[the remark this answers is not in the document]
rebeccaalpert: I think we have to keep this as a let because it gets modified further down. I addressed the other comments; will add a new commit.

jeff-phillips-18:
I still maintain that a hard coded line truncation is a bad idea.

rebeccaalpert:
Hey - I wound up speaking to Dana about this yesterday because I was noticing some performance issues with the refs on the PF4 developer site. He and Michael Coker suggested we use a CSS solution, so I'm in the process of rolling back to the original approach. Coker thought line clamp was a decent approach for this.

Serena just pointed out the issue we've been trying to fix with this PR to a larger audience and Joe and I would like to fast-track this fix so we can get it in under the wire into this upcoming version of OpenShift.

jeff-phillips-18:
Why is this less performant than it was in the previous version?

rebeccaalpert:
In addition to the issues with the line height that we discussed over Slack, there seemed to be an issue where we were intermittently receiving the incorrect clientHeight. It happened intermittently on refresh on the dev site. I asked Dana about it and we tried running a production build to see if it was an issue with the dev site. However, on the production build, you would see a delay, where there was an incorrect clientHeight and then the correct values seemed to kick in, resizing the content. This is not acceptable behavior for OpenShift.

rebeccaalpert:
They mostly have text badges. We're going to shrink them down in OpenShift to widths and heights that make sense for those layouts.

rebeccaalpert:
Rebased.

PR #3398, opened 2019-12-11
fix(DataToolbar): hide Clear all filters button if no event handler provided

nicolethoen:
Fixes issue #3400 found while testing Katacoda tutorial.

### DIFF
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.tsx
@@ -106,7 +106,7 @@ export class DataToolbar extends React.Component<DataToolbarProps, DataToolbarSt
 
     const isToggleManaged = this.isToggleManaged();
     const numberOfFilters = this.getNumberOfFilters();
-    const showClearFiltersButton = numberOfFilters > 0;
+    const showClearFiltersButton = clearAllFilters && numberOfFilters > 0;
 
     return (
       <div className={css(styles.dataToolbar, className)} id={id} {...props}>
--- a/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/__snapshots__/DataToolbar.test.tsx.snap
+++ b/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/__snapshots__/DataToolbar.test.tsx.snap
@@ -2018,7 +2018,6 @@ exports[`data toolbar DataToolbarItemsAndGroups 1`] = `
       collapseListedFiltersBreakpoint="lg"
       isExpanded={false}
       numberOfFilters={0}
-      showClearFiltersButton={false}
     >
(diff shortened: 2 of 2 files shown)


tlabaj:
Can you please add the corresponding issue number

nicolethoen:
> @mcarrano is this what we want to do here? I ask because we don't always do this elsewhere if a button is not provided with a handler function. Eg. the close button on a Modal.

@tlabaj without this fix, the Clear All Filters button is a blank button. There is no default behavior for this button because all the filters are managed by the consumer.

tlabaj:
@nicolethoen by blank button you mean it does nothing correct?  If so, it is not the only button in the library that behaves that way.  I am just making sure we are being consistent with how we handle cases like this.  Maybe we should be hiding buttons elsewhere or maybe this is right for this use case in particular.  That is why I am asking.

nicolethoen:
@tlabaj I see. it was causing confusion in the katacoda tutorial. If it's better to have a button that doesn't do anything without an optional event handler prop provided, then I'll just have to explain that in the tutorial.

mcarrano:
@tlabaj @nicolethoen My 2 cents is that we should be consistent with whatever we do. If I read this correctly, sound like you are saying that it's the consumer's responsibility to clear the filters and restore the original data, is that correct? Hiding the button does not seem like the right thing because the feature is not really optional. The end-user must be able to clear any filters that were set. Not providing a handler would constitute a bug on the consumer side IMO.

nicolethoen:
@mcarrano @tlabaj I can close this PR. I just want to make sure we are all on the same page before I do.

A delete event handler is required to be passed to the individual chips so that a user is always able to delete a chip[applied filter] if it appears in the toolbar. 

This issue is that right now, a ClearAllFilters function is not required (since many toolbars don't have filters at all).

Currently the ONLY way to make that button disappear is to remove all applied filters. If there are no applied filters, then the clear all filters button rightfully does not display.  

If the user does not want a "Clear All Filters" button alongside their filters, they don't have that option. if they have not provided the optional ClearAllFilters prop, the "Clear All Filters" button/link still appears, but it does not do anything. There is no way to make the dead button disappear. 

This PR just adds the one piece of logic so that if the optional ClearAllFilters property is not provided, the Clear All Filters button will be hidden even if filters are applied. Remember, they can still clear chips individually in such a case. 

If the Clear All Filters button needs to exist in EVERY toolbar with a filter, then the existence of a dead button that does not do anything does force them to implement the button. I was under the assumption that it was not a requirement for this component that it contain a Clear All Filters button anytime there are filters. So, if we are all agreed that I was mistaken, I'll close the PR.

Phew, sorry for the essay.

mcarrano:
@nicolethoen I am in favor of closing this PR. But we could revisit this if there is an actual use case that requires the button to be hidden. Right now our design position is that Clear all should be possible whenever multiple filters are applied.

PR #3402, opened 2019-12-12
fix(Table): Fix the ActionColumn to allow a link in the actions menu

### LINKED ISSUE #3392: Table ActionsColumn: Action can't be a link due to preventDefault
We have a table on our app (OCM) with an actions column. One of the actions is an external link. When we implemented this with the usual Dropdown kebab toggle (before we heard of the actions column), this worked well - a Dropdown item is an `<a>` by default, so we can just specify an `href`, and it works well.

Now we want to use the actions column feature of the Table component.

The ActionColumn allows me to specify a list of actions. The [IAction interface](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/Table.tsx#L110-L115) is defined as a superset of the [DropdownItem](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.tsx#L5-L26) props. 

This means that it should, in theory, accept an `href`, and it does indeed render this correctly. However, [the onSelect event handler in the ActionsColumn](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/ActionsColumn.tsx#L49-L62) calls `event.preventDefault();`, which **disable's the browser default link behavior**, making the link non functional.

Since the default `onSelect` event handler is [always called](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/ActionsColumn.tsx#L79)  we have no way to work around this behavior, making this **a blocking bug** for us.

tlabaj:
Closes  #3392

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/ActionsColumn.tsx
+++ b/packages/patternfly-4/react-table/src/components/Table/ActionsColumn.tsx
@@ -53,9 +53,12 @@ export class ActionsColumn extends React.Component<ActionsColumnProps, ActionsCo
       | undefined
   ): void => {
     const { rowData, extraData } = this.props;
-    event.preventDefault();
-    // tslint:disable-next-line:no-unused-expression
-    onClick && onClick(event as React.MouseEvent, extraData && extraData.rowIndex, rowData, extraData);
+    // Only prevent default if onClick is provided.  This allows href support.
+    if (onclick) {
+      event.preventDefault();
+      // tslint:disable-next-line:no-unused-expression
+      onClick(event as React.MouseEvent, extraData && extraData.rowIndex, rowData, extraData);
+    }
     this.setState(prevState => ({
       isOpen: !prevState.isOpen
     }));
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-table/src/components/Table/ActionsColumn.tsx
```diff
@@ -53,9 +53,12 @@ export class ActionsColumn extends React.Component<ActionsColumnProps, ActionsCo
       | undefined
   ): void => {
     const { rowData, extraData } = this.props;
-    event.preventDefault();
-    // tslint:disable-next-line:no-unused-expression
-    onClick && onClick(event as React.MouseEvent, extraData && extraData.rowIndex, rowData, extraData);
+    // Only prevent default if onClick is provided.  This allows href support.
+    if (onclick) {
```
jschuler: should the c be capitalized?
```suggestion
    if (onClick) {
```

Issue #3233, opened 2019-10-29
Hover for list view 


LHinson:
Hover for list view as a follow up to https://github.com/patternfly/patternfly-next/issues/2370

rachael-phillips:
The only functionality you’ll need from core is the ability to designate a data list as selectable or not (a prop that applies a CSS modifier class), which is what will differentiate the row hover and expanded left blue border color differences from non-selectable data lists.
Then when you select a row in a selectable data list, that applies a CSS modifier class to the row

mcoker:
This is ready to be worked on. From https://github.com/patternfly/patternfly-next/pull/2491

* Selectable rows get the class `.pf-m-selectable` and attribute `tabindex="0"`
* Selected rows get the class `.pf-m-selected`

@jessiehuff how do you think we should announce that a row is selected? Each row (`.pf-c-data-list__item`) is `aria-labelledby` the `id` of the primary cell in the row. Could we incorporate "selected" text in that label somehow, or maybe use `aria-selected`?

PR #3404, opened 2019-12-12
feat(DataList): add selectable variant

nicolethoen:
addresses #3233 

I have not figured out a clean way to prevent clicking the kebab (or interacting with components which have their own event handlers) from also selecting a DataListItem row. Events keep propagating up to parent components. I'd love any help people are willing to give... (@evwilkin @jschuler).

Is it possible the interaction as is would be acceptable for merging before code freeze monday? I could open an issue to figure out the last part? @mcarrano @mcoker

### DIFF
--- a/packages/patternfly-4/react-core/src/components/DataList/DataListItem.tsx
+++ b/packages/patternfly-4/react-core/src/components/DataList/DataListItem.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 import { Omit } from '../../helpers/typeUtils';
+import { DataListContext } from './DataList';
+import { KeyTypes } from '../Select';
 
 export interface DataListItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'children'> {
   /** Flag to show if the expanded content of the DataList item is visible */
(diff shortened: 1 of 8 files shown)


mcarrano:
@nicolethoen without actually seeing this, my opinion is that it would be a significant usability problem if opening a kabob, for example also selected the row. If we were to release this, I would only do so with the restriction that the row should not include active elements to avoid this behavior. @LHinson @rachael-phillips do you recall who was asking for this feature and what their intended usage was?

dlabaj:
@nicolethoen Can you do a prevent default when you click the kebab?

packages/patternfly-4/react-core/src/components/DataList/DataListItem.tsx
```diff
@@ -12,32 +14,77 @@ export interface DataListItemProps extends Omit<React.HTMLProps<HTMLLIElement>,
...
-        React.isValidElement(child) &&
-        React.cloneElement(child as React.ReactElement<any>, {
-          rowid: ariaLabelledBy
-        })
-    )}
-  </li>
-);
+export class DataListItem extends React.Component<DataListItemProps> {
+
+  // @ts-ignore
```
tlabaj: Why do we need this?

nicolethoen: typescript has some issues with pulling in context this way. For a handful of versions of React, pulling in a static context was a deprecated practice. It is now supported again, but the ts linter doesn't know that? All instances of a static context in the project need the ts-ignore flag. And its the only way to reference the context outside of a context.consumer in a render function.

nicolethoen:
I rebased with the latest changes, but now selecting the rows doesn't appear to be working. let me investigate quickly

nicolethoen:
> I rebased with the latest changes, but now selecting the rows doesn't appear to be working. let me investigate quickly

I take it back! It looks like the example is working in the surge preview :) @mcarrano @mcoker @tlabaj

Issue #3268, opened 2019-11-05
react-icons build optimization


Hyperkid123:
Hello. We are building our docs page and we have serious issues with the react-icons bundle size.

We are using server side rendering and there is 8 chunks that all have additional 1.8mb bundle size caused just in `@patternfly/react-icons`. In our project we are using only 5 distinct icons, but the build includes all of them. We have tree shaking enabled but with the way how the icons are exposed it will never mark them as unreachable and it will not exclude them from the build.

We could theoretically mark each icon file in node modules as a side effect but that would mean marking each and every single icon file separately which is not really. It is much simpler to take these 5 svgs and store them in our code base and drop the dependency all together (that is what will do for now).

Here you have a screenshot of bundle sizes of every page created for SSR chunks:
![Screenshot from 2019-11-05 16-35-07](https://user-images.githubusercontent.com/22619452/68222368-326e5600-ffeb-11e9-99e8-5e9fe051bba6.png)

Here is comparision of chunk sizes before (with react-icons) and after (without, moved svgs to code base)

### Before
![Screenshot from 2019-11-05 17-52-14](https://user-images.githubusercontent.com/22619452/68229063-d27dac80-fff6-11e9-941e-3313cb47923b.png)

### After
![Screenshot from 2019-11-05 17-51-24](https://user-images.githubusercontent.com/22619452/68229056-d0b3e900-fff6-11e9-8ed5-8e36ab170d8e.png)

You can see that `react-icons/dist/ems` is in all of the biggest chunks. Each having 1.8 MB. That means that 8 of our SSR pages has over 2.2 MB of size which is not acceptable for SRR and completely defeats the purpose of this technology.

Can you please look at this issue? We are not able to use PF4 efficiently and we have to use different means (like duplicating components and having them in our code base).

martinpovolny:
@dlabaj, @tlabaj : any updates?

Whom should we ping regarding this?

karelhala:
This was first opened in https://github.com/patternfly/patternfly-react/issues/2533 without any resolution. 

@rachael-phillips this has been observed from many teams, we see similiar issue in cloud services as well and our bundle is getting bigger with each icons bump. We were told that treeshaking will take care of unused icons. Turns out not to be the case and all icons are included multiple times if you create multiple bundles.

I would mark this as high priority, because it bloats our resources and either example how to properly bundle icons or some work that enables tree shaking.

evwilkin:
[...]

PR #3410, opened 2019-12-16
fix(docs): documented CJS tree shaking solution for icons and react-core

evwilkin:
**What**: Addresses #3268 

**Additional issues**:

### DIFF
--- a/packages/patternfly-4/react-core/README.md
+++ b/packages/patternfly-4/react-core/README.md
@@ -112,3 +112,54 @@ yarn test packages/patternfly-4/react-core
 [docs]: https://patternfly-react.surge.sh/patternfly-4
 
 
+## Tree Shaking
+
+Ensure optimization.sideEffects is set to true within your Webpack config:
+```JS
+optimization: {
+  sideEffects: true
+}
+```
+
+Use ESM module imports to enable tree shaking with no additional setup required.
+```JS
+import { TimesIcon } from '@patternfly/react-icons';
+```
+
+To enable tree shaking with named imports for CJS modules, utilize [babel-plugin-transform-imports](https://www.npmjs.com/package/babel-plugin-transform-imports) and update a babel.config.js file to utilize the plugin:
+```JS
+require.extensions['.css'] = () => undefined;
+const components = require('@patternfly/react-core/dist/js/components');
+const experimental = require('@patternfly/react-core/dist/js/experimental');
+const layouts = require('@patternfly/react-core/dist/js/layouts');
+
+module.exports = {
+  presets: ["@babel/preset-react"],
+  plugins: [
+    [
+      "transform-imports",
+      {
+        "@patternfly/react-core": {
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/README.md
```diff
@@ -112,3 +112,52 @@ yarn test packages/patternfly-4/react-core
...
+
+Use ESM module imports to enable tree shaking with no additional setup required.
+```JS
+import { TimesIcon } from '@patternfly/react-icons';
+```
+
+To enable tree shaking with named imports for CJS modules, utilize [babel-plugin-transform-imports](https://www.npmjs.com/package/babel-plugin-transform-imports) and update a babel.config.js file to utilize the plugin:
+```JS
+require.extensions['.css'] = () => undefined;
+const components = require('@patternfly/react-core/dist/js/components');
```
jschuler: our package.json files have `main` and `module` entries, when doing require it should find the cjs entry point as specified in `main` and when doing imports it should find the esm entry point in `module`. So I don't think you need the whole path?

jschuler: anyways, might be different for babel js resolution. looks good :)

Hyperkid123:
@evwilkin @dlabaj @redallen Hello, I did some testing and there are some issues with this particular config. Adding link to a issue with more detail: https://github.com/patternfly/patternfly-react/issues/3268#issuecomment-566442100

martinpovolny:
@evwilkin : First on all I'd like to thank you for keeping working on this. I really appreciate that despite what I am going to write next.

So I have to say that this issue makes be wonder about Patternfly development processes.

You have an issue, that is complex, hard to fix, hard to understand, hard to discuss.

You discuss it for weeks. Have ideas, some wrong, some right, obviously it's not a simple thing.

The issue was already once prematurely closed and reopened.

Then you make a PR and merge it within 1 hour. Merge it within one hour since it was opened w/o giving the people who where previously discussing the issue a chance to check it/test it/review it.

Well and no. It did not resolve the issue.

Wow. Just wow.

evwilkin:
@martinpovolny please note that we've kept the original issue open (#3268 ) until we're able to validate together that we have a successful fix, and are happy to continue working together to resolve this.

Hyperkid123:
@evwilkin I tried to apply the transform plugin to the react-core repo to use full imports on react icons and the results are great:

```

Compiled successfully.

Automatically optimizing pages  

Page                    Size
┌ ○ /                   27 kB
└   /_app               1.11 kB
+ shared by all         71.8 kB
  ├ chunks/commons.js   65.7 kB
  ├ runtime/main.js     5.26 kB
  └ runtime/webpack.js  812 B

λ  (Server)  server-side renders at runtime (uses getInitialProps or getServerProps)
○  (Static)  automatically rendered as static HTML (uses no initial props)
●  (SSG)     automatically generated as static HTML + JSON (uses getStaticProps)

Done in 4.31s.
```

The bundle now contains only the components that it should. I will create PR with the changes to the react-core repo.

PR #3448, opened 2020-01-02
Use direct paths to react-icons in prod build

Hyperkid123:
**What**: Addresses #3268

Added transform-imports for react-core to use direct paths to react-icons dependencies. This change marginally improves build sizes of libraries which depends on a library that is also using react-core package icons and then is consumed by some app.

Bundle size of an app using just PF4 Modal from a third-party library was reduced from 513kB to 27Kb 

Right now, if you use any component with an icon from react-icons (Modal for example), it sometimes bundles all react-icons even if not used.

A reproducible example can be found here: https://github.com/Hyperkid123/pf-core-icons-deps

Detailed info can be found in #3268

### DIFF
--- a/packages/patternfly-4/react-core/.babelrc
+++ b/packages/patternfly-4/react-core/.babelrc
@@ -1,3 +1,0 @@
-{
-  "extends": "../.babelrc"
-}
--- a/packages/patternfly-4/react-core/babel.config.js
+++ b/packages/patternfly-4/react-core/babel.config.js
@@ -1,0 +1,31 @@
+const envDistMapping = {
+  cjs: 'js',
+  esm: 'esm',
+  umd: 'umd'
+}
+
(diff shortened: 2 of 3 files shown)


evwilkin:
@Hyperkid123 thanks for putting in this PR, it looks good to me but I am going to leave open for @redallen to review when he's back in office as well before merging.

evwilkin:
addresses #3268

Hyperkid123:
@evwilkin I am doing more testing and this will have to be applied to all PF4 extensions using icons. I will do that now. I also have some updates regarding the react-core transform imports. I will post some notes and possible solutions to the issues thread.

redallen:
This change is good for CJS bundles. If we are to ever switch to building react-core with Typescript we'll have to change all icon import paths manually to import directly from react-icons' dist.

PR #3474, opened 2020-01-09
Lint md code blocks 3469

ddonahue007:
linting code blocks in any markdown file. We are only producing warnings so not to break the build. (This will change) Initial rules could change as well. This is a good starting point to cleanup linting errors.

fixed all hard errors in this commit so not to break the build.

linting is run for pf4 only. It is run as part of the 'build:pf4' target. It also can be run by itself calling 'lint:md'


packages/patternfly-4/react-styles/README.md
```diff
@@ -105,7 +105,7 @@ const styles = StyleSheet.parse(`
 `);
 
 const Button = ({
-  variant // primary | secondary,
+  primary, // variant primary | secondary,
```
dgutride: should primary actually be "variant" here?

ddonahue007: fixed...

packages/patternfly-4/react-core/src/components/Dropdown/examples/Dropdown.md
```diff
@@ -1111,6 +1111,7 @@ A react-router Link may be wrapped by DropdownItem or used directly within Dropd
 
 ```js noLive
 /** Wrapped Link for DropdownItem list **/
+<div>
```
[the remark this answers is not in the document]
ddonahue007: done...

packages/patternfly-4/react-core/src/components/Dropdown/examples/Dropdown.md
```diff
@@ -1111,6 +1111,7 @@ A react-router Link may be wrapped by DropdownItem or used directly within Dropd
 
 ```js noLive
 /** Wrapped Link for DropdownItem list **/
+<>
```
redallen: @ddonahue007 Is it possible to make the eslint check our JSX to ensure we use `<React.Fragment>` instead of `<>`? That would fix some of our fullscreen examples.

Issue #3155, opened 2019-10-16
Tree shaking issue


johnparkTopQuad:
We noticed that even though we're only using a couple of icons and a single react component, we're getting ove 2MB of icons included in our production bundles. Is the tree shaking working as intended? Is there something we need to do configuration-wise on our end to take advantage of tree-shaking on the consuming side? We're using a custom webpack configuration that is similar in setup to the seed project in terms of how the loaders are managed.

dgutride:
@johnparkTopQuad  - sorry I haven't been able to repro this locally - treeshaking is working when I test in webpack, it's possible this is a babel issue.  If you can post any additional info you get from your investigation, that would be helpful.

riccardo-forina:
Syndesis suffer of the same problem. Or, in a smaller scale, the [example app in use-patternfly](https://github.com/riccardo-forina/use-patternfly/tree/master/example) shows the same behavior. Checked with [source-map-explorer](https://www.npmjs.com/package/source-map-explorer)

johnparkTopQuad:
Hey I wanted to give an update so you don't think this ticket is dead. I've been poking around with Webpack and Patternfly and measuring bundle size differences and I've noted the following:

Patternfly Seed Project -
TSX branch :
    Before adding Wizard:
        All (768.66 KB)
        app.bundle.js (757.71 KB)
        support.bundle.js (10.94 KB)

    After adding Wizard:
        All (850.78 KB)
        app.bundle.js (839.84 KB)
        support.bundle.js (10.94 KB)

I'm not sure if this is the expected growth of importing a single Wizard component. 


My project that consumes Patternfly -
Current bundle size  -   7.49 MiB
Current bundle size without Patternfly -  4.06 MiB

Since we are really only importing a single Wizard component I think were getting a lot more than we need. I tried changing out the babel-loaders for ts loaders and replicating what I could from seed but my results were the same. 

At this point ill see if i can maybe get seed project tree shaking to break / keep messing with Webpack.

seanforyou23:
Related to https://github.com/patternfly/patternfly-react/issues/3268

evwilkin:
@johnparkTopQuad @riccardo-forina FYI - for react-icons you can use the `babel-plugin-transform-imports` plugin and `babel.config.js` setup laid out here:

https://github.com/patternfly/patternfly-react/pull/3448

Keeping this issue open as we work on react-core components, which will require a different solution due to the directory structure & export names.

[...]

PR #3476, opened 2020-01-09
Use transform imports plugin on react-icons for all pf4 builds.

Hyperkid123:
**What**: Adress #3448  and https://github.com/patternfly/patternfly-react/issues/3155

Changes build paths to icons to use direct paths instead of using icons from `index.js`


Hyperkid123:
cc @karelhala

packages/patternfly-4/.babelrc
```diff
@@ -1,11 +0,0 @@
...
-  "presets": ["@babel/react"],
-  "ignore": ["**/__snapshots__", "**/*.d.ts", "**/*.test.*"],
-  "plugins": [
-    "@babel/plugin-transform-typescript",
-    "babel-plugin-typescript-to-proptypes",
-    "@babel/plugin-proposal-export-default-from",
-    "@babel/proposal-class-properties",
-    "@babel/proposal-object-rest-spread",
-  ]
-}
```
redallen: Why couldn't you have just renamed this to `babel.config.js` and added your plugin here? That would apply it to all the pf4 builds. I worked hard to unify our babel config and get away from `BABEL_ENV=cjs` or whatnot so that Windows works without having to use yet another tool called cross-env.

Soon I'm planning on removing babel in pf4 in favor of `tsc`, which we're already using for typechecking, so I'm also not really a fan of adding this plugin to begin with. Could we just change all the instances of `import { SomeIcon } from '@patternfly/react-icons';` to `import { SomeIcon } from '@patternfly/react-icons/dist/js/SomeIcon` in all `patternfly-4/*.{tsx,js}` files? There's only about 50 or so. We could easily add a linting step that gets upset if we try to use ESM imports like `import { SomeIcon } from '@patternfly/react-icons';`.

Hyperkid123: If you are planing the direct imports that would be perfect and we can close this. I would prefer that anyway so you don't have to rely on some plugin.

I did not want to do any changes to the component code.

I would like to point out that the same thing should be done with react-core components in extensions like react-table.

redallen:
Does #3517 work for you @Hyperkid123 ?

Hyperkid123:
@redallen works. But I've noticed that you have marked #3155 as resolved. I think it could be closed after all extensions use direct imports not only for `react-icons` but for `react-core` as well.

If you have your hands full I can start working on that.

redallen:
I updated #3517 to just say "towards #3155".

I do currently have my hands full, so working on that would be appreciated and help reduce CJS bundle sizes further.

Hyperkid123:
Ok, I will start the changes ASAP. I think this can be closed in favor of your PR.

Issue #3401, opened 2019-12-12
Select - custom option data comparison


kmcfaul:
**Describe the issue. What is the expected and unexpected behavior?** Custom option data should be compared via a custom comparison function (added to the SelectOptionObjec interface) opposed to the current direct equality comparison.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?** There is a work around using strings for comparison.

PR #3491, opened 2020-01-10
feat(Select): allow custom option data comparison

nicolethoen:
Addresses: #3401

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
@@ -7,7 +7,9 @@ import { SelectConsumer, SelectVariant, KeyTypes } from './selectConstants';
 import { Omit } from '../../helpers/typeUtils';
 
 export interface SelectOptionObject {
+  /** Function returns a string to represent the select option object */
   toString(): string;
+  /** Function returns a true if the passed in select option is equal to this select option object, false otherwise */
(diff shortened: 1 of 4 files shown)


packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -8,6 +8,7 @@ import { Omit } from '../../helpers/typeUtils';
 
 export interface SelectOptionObject {
   toString(): string;
+  compareTo?(selectOption: any): boolean;
```
dlabaj: We should add comments to this interface for to toString and compareTo

packages/patternfly-4/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -18,7 +19,7 @@ export interface SelectOptionProps extends Omit<React.HTMLProps<HTMLElement>, 't
   index?: number;
   /** Indicates which component will be used as select item */
   component?: React.ReactNode;
-  /** The value for the option, if passing an object you most provide a toString function */
+  /** The value for the option, if passing an object you most provide a toString function and optional compareTo function */
```
dlabaj: If we add the comments to the interface we could just tell them see the SelectOptionObject.  E.g. The value for the option, can be a string or a SelectOptionObject.

PR #3493, opened 2020-01-12
fix(linting): eslint for *.md file fixes

ddonahue007:
for react-catalog-view-extension

initial cleanup of linting errors on *.md files.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1761,6 +1761,15 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
+"@babel/types@^7.4":
+  version "7.8.0"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -1761,6 +1761,15 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
+"@babel/types@^7.4":
```
jschuler: are these new dependencies? Don't see anything in this PR that should require it

ddonahue007: These are dependencies for other packages. They could have bumped due to version change as well.

jschuler: okay 👍

PR #3497, opened 2020-01-13
chore(packages): downgrade all packages to react-^16.4.0

### LINKED ISSUE #3435: Downgrade React  to 16.4.0
We should downgrade React across this repo so that we don't accidentally use new React features in our library while claiming to support ^16.4.0.

boaz0:
**What**: Closes #3435


packages/patternfly-3/patternfly-react/package.json
```diff
@@ -50,8 +50,8 @@
   },
   "peerDependencies": {
     "prop-types": "^15.6.0",
-    "react": "^16.3.2",
-    "react-dom": "^15.6.2 || ^16.2.0"
```
dlabrecq: I'm don't believe we should change the dependencies for PatternFly 3. There may have been a reason for supporting these specific versions?

packages/patternfly-4/react-core/package.json
```diff
@@ -73,6 +73,6 @@
   "peerDependencies": {
     "prop-types": "^15.6.1",
     "react": "^16.4.0",
-    "react-dom": "^15.6.2 || ^16.4.0"
+    "react-dom": "^16.4.0"
```
dlabrecq: This looks like we're forcing apps to upgrade from 15.6.2, which may be a breaking change for some. I don't believe it would hurt leaving 15.6.2 there, unless there is specific functionality required by 16.4.0?

packages/patternfly-3/patternfly-react-extensions/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "peerDependencies": {
     "prop-types": "^15.6.1",
-    "react": "^16.3.2",
-    "react-dom": "^16.3.2"
+    "react": "^16.4.0",
+    "react-dom": "^16.4.0"
```
dlabrecq: Is the intention to be consistent or do we really want to downgrade / lock at 16.4.0? 

Specifying "^16.4.0" will potentially install the latest features of 16.12.0.

packages/patternfly-3/patternfly-react-extensions/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "peerDependencies": {
     "prop-types": "^15.6.1",
-    "react": "^16.3.2",
-    "react-dom": "^16.3.2"
+    "react": "16.4.0",
+    "react-dom": "16.4.0"
```
dlabrecq: @redallen we don't want to change peer dependencies, correct?

redallen: Correct, let's leave these alone.

Nice new icon btw @boaz0 !

redallen:
Closing in favor of #3875 in an upcoming breaking change release. Thanks for this PR anyway @boaz0 , without it we wouldn't have known #3875 is the best course of action.

Issue #2580, opened 2019-07-23
Configurable roles for Page and PageHeader


seanforyou23:
Currently, these components default to `role="main"` and `role="banner"` respectively. These are sensible defaults, but shouldn't be hardcoded. We shouldn't assume that Patternfly layouts are always used at the root of an application. It's a feature that the page markup/styling is nestable, so we need to make sure users can customize these settings in these cases.

An immediate example is within our integration test application.

<img width="1110" alt="Screen Shot 2019-07-23 at 4 31 13 PM" src="https://user-images.githubusercontent.com/5942899/61745378-6f2d7e80-ad67-11e9-90d8-47abd4c0732d.png">

Generally, the banner role should only be used for the outermost header of an application. Similarly, there should only be a single main role and it should be the outermost container for the app's dynamic content.

jenny-s51:
After corresponding with @seanforyou23 regarding this issue, he added some helpful feedback on what he meant by "configurable":

"We do want the `role` to be controllable by the user in case some advanced scenario comes up where people need more fine grain control over them"

I.e. the attributes can be removed, but add a "prop there as an option in case someone wants/needs to set it explicitly". This prop would set the `role` on either the `head` or the `main` element. 

I'd be happy to take on this issue once it's prioritized and I have some extra bandwidth 🙂

PR #3499, opened 2020-01-13
fix(PageHeader): removing redundant role="banner" attribute

### LINKED ISSUE #3498: PageHeader: role=banner attribute on <header>

This issue is related to [#2580](https://github.com/patternfly/patternfly-react/issues/2580). 

The error I get from AXE is “Banner landmark must not be contained in another landmark”, and the element looks like this currently: `<header role="banner" class="pf-c-page__header">` .

According to the ARIA docs, the `role="banner"` doesn’t need to be there at all, which is where the error is coming from.  

"Assistive technologies can identify the main header element of a page as the banner if is a descendant of the body element". In this case, it _is_ a descendant of the body element, and therefore the `role="banner"` is redundant. (See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_Role).

It looks like the folks at Jekyll have addressed a similar issue:
https://github.com/jekyll/minima/issues/75

jenny-s51:
Closes #3498.

**Additional issues**:  Addresses #2580.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Page/PageHeader.tsx
+++ b/packages/patternfly-4/react-core/src/components/Page/PageHeader.tsx
@@ -57,7 +57,7 @@ export const PageHeader = ({
         const navOpen = isManagedSidebar ? managedIsNavOpen : isNavOpen;
 
         return (
-          <header role="banner" className={css(styles.pageHeader, className)} {...props}>
+          <header className={css(styles.pageHeader, className)} {...props}>
             {(showNavToggle || logo) && (
               <div className={css(styles.pageHeaderBrand)}>
                 {showNavToggle && (
--- a/packages/patternfly-4/react-core/src/components/Page/__snapshots__/Page.test.tsx.snap
+++ b/packages/patternfly-4/react-core/src/components/Page/__snapshots__/Page.test.tsx.snap
@@ -40,7 +40,6 @@ exports[`Check dark page against snapshot 1`] = `
     >
       <header
         className="pf-c-page__header"
-        role="banner"
       >
         <div
           className="pf-c-page__header-brand"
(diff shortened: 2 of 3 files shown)


tlabaj:
@jenny-s51 This will not close #2580 correct?

jenny-s51:
@tlabaj yes that's correct -- that issue involves the `main` attribute of `Page` as well. I'm happy to remove that attribute so that issue can also be closed! It was out of scope of this PR, but after looking into it, seems like we can just remove that attribute without running into any problems. Would you like me to do that? Should be a very quick fix 🙂

tlabaj:
@jenny-s51 Yes, can you please do that.  Thanks!

jenny-s51:
@tlabaj yes! checking with @seanforyou23 now to see if doing that would resolve the issue in the way he hoped

packages/patternfly-4/react-core/src/components/Page/PageHeader.tsx
```diff
@@ -57,7 +57,7 @@ export const PageHeader = ({
         const navOpen = isManagedSidebar ? managedIsNavOpen : isNavOpen;
 
         return (
-          <header role="banner" className={css(styles.pageHeader, className)} {...props}>
+          <header className={css(styles.pageHeader, className)} {...props}>
```
seanforyou23: Can we allow this attribute to be omitted by default, but set if the user provides a value explicitly?

jenny-s51: Absolutely! Will open up a separate issue for adding a new prop that will do this.

seanforyou23: Ty!

Issue #3460, opened 2020-01-07
[Select] Missing clear button


karelhala:
**Describe the issue. What is the expected and unexpected behavior?**
When using filterable Select there is clear button missing. It is designed https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#type-ahead-filters and implemented in CSS core repository https://www.patternfly.org/v4/documentation/core/components/select#single-with-typeahead-expanded.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

* Use Select component
* Pass options to it
* Use custom filter function to filter out items

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
It is blocking us from using this component since users has no way how to clear filter string.

**What is your product and what release version are you targeting?**
Cloudservices

tlabaj:
closed by #3502

PR #3502, opened 2020-01-14
fix(Select): clear button appears with any typeahead input

nicolethoen:
Addresses #3460

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Select/Select.tsx
+++ b/packages/patternfly-4/react-core/src/components/Select/Select.tsx
@@ -463,7 +463,7 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
                     disabled={isDisabled}
                   />
                 </div>
-                {selections && (
+                {(selections || typeaheadInputValue) && (
                   <button
                     className={css(buttonStyles.button, buttonStyles.modifiers.plain, styles.selectToggleClear)}
                     onClick={e => {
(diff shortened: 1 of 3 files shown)


karelhala:
Codewise it looks really good and simple! Can we also call `onSelect` with the emtpy input so if consumers are using custom filtering they get notified and can filter based on their decision. This is important so when user clears out the text for instance server can be called and queried for all the data instead of filtered ones.

nicolethoen:
> Codewise it looks really good and simple! Can we also call `onSelect` with the emtpy input so if consumers are using custom filtering they get notified and can filter based on their decision. This is important so when user clears out the text for instance server can be called and queried for all the data instead of filtered ones.

I could do that. Might it be a breaking change if consumers have written onSelect functions without designing them to handle empty inputs? We do provide an onClear property that is only called when the user clicks the clear button. Can that not serve the same purpose?

kmcfaul:
`onClear` will be called already when the clear button is clicked, and this will clear the filtering by default. If you want to prompt an `onSelect` call, you would be able to do so from the application-side of the code by calling the function passed into `onSelect` inside the function passed into `onClear`.

@nicolethoen It does look like there's a small issue with the clear button not resetting the filtering when cleared in this way. When the menu is reopened it retains the previous filtered state until new input is entered. I think the internal `onChange` function has to be called with an empty input when the clear button is clicked to get it to update.

Issue #3483, opened 2020-01-09
Promote Spinner from Experimental


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Promote the Spinner component from Experimental.

tlabaj:
closed by #3516

Issue #3482, opened 2020-01-09
Promote OverflowMenu from Experimental


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Promote the OverflowMenu component from Experimental.

tlabaj:
closed by #3516

Issue #3481, opened 2020-01-09
PF4: Promote Divider from Experimental


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Promote the Divider component from Experimental.

tlabaj:
closed by #3516

Issue #3480, opened 2020-01-09
PF4: Promote DataToolbar from Experimental


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Promote DataToolbar component from Experimental.

tlabaj:
closed by #3516

tlabaj:
Re-opening since we reverted the promotion.

PR #3516, opened 2020-01-15
chore(Spinner, DataToolbar, Divider, OverflowMenu): Promote Spinner, DataToolbar, Divider and OverflowMenu from experimental

tlabaj:
#3483 #3480 #3481 #3482

**What**: Closes #

**Additional issues**:


evwilkin:
@tlabaj the file changes look good to me but I don't see "Spinner" listed under the Component in the preview link - is this expected when promoting components?  It does still show in Experimental, as expected.

packages/patternfly-4/react-core/src/experimental/components/index.ts
```diff
@@ -1,5 +1,5 @@
 export * from './DataToolbar';
-export * from './Divider';
+export * from '../../components/Divider';
 export * from './Drawer';
-export * from './Spinner';
```
dlabrecq: Would it be possible to deprecate the spinner (e.g., for at least one milestone) from the experimental package Vs removing it immediately?

We're using the spinner in Cost Management and this change would break us. It would give folks a little warning that apps must update.

dlabrecq: Never mind, I see that spinner is in the index file.

PR #3519, opened 2020-01-16
Accessibility updates to alert for toast alerts

### LINKED ISSUE #1421: Alert: create variation to use for dynamic notifications

jessiehuff:
**What**: Closes #1421


packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
```diff
@@ -57,6 +60,7 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
   const customClassName = css(
     styles.alert,
     isInline && styles.modifiers.inline,
+    isToast && 'pf-m-live',
```
tlabaj: Why are you using the class name here instead of react-styles?

PR #3523, opened 2020-01-16
fix(demos): Added comments above svg import statements to instruct us…

ddonahue007:
For now we have decided to add comments that will help if the user wants to upload the svg images. This needs to be done manually through codesandbox.


redallen:
Eh, we technically could upload them instead of giving up and leaving a comment. SVGs are also imported in 43 other places across 13 files:

image


Want to spend an hour or so peer programming this one?

ddonahue007:
> Eh, we technically could upload them instead of giving up and leaving a comment. SVGs are also imported in 43 other places across 13 files:
> 
> image
> 
> Want to spend an hour or so peer programming this one?



> Eh, we technically could upload them instead of giving up and leaving a comment. SVGs are also imported in 43 other places across 13 files:
> 
> image
> 
> Want to spend an hour or so peer programming this one?

Sure I would think if we could fix it easily it should be. Rather than putting these comments all over the place. To me the comments just look sloppy.

ddonahue007:
Closing this PR as we are addressing this issue in the patternfly-org code base.
PR: https://github.com/patternfly/patternfly-org/pull/1632

PR #3525, opened 2020-01-17
feat(pf4): Use absolute import paths for react core in pf4 packages.

Hyperkid123:
**What**: towards #3155

### DIFF
--- a/packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
+++ b/packages/patternfly-4/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
@@ -1,5 +1,10 @@
 import * as React from 'react';
-import { Card, CardActions, CardHead, CardHeader, CardBody, CardFooter } from '@patternfly/react-core';
+import { Card } from '@patternfly/react-core/dist/js/components/Card/Card';
+import { CardActions } from '@patternfly/react-core/dist/js/components/Card/CardActions';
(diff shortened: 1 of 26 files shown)


Hyperkid123:
@redallen everything looks good except with build of react-topology/TopologyControlBar and react-topology/TopologySideBar components.

Typescript props checking is getting crazy and throwing this error:
```
error TS2741: Property 'css' is missing in type '{ children: (string | number | boolean | {} | ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)> | ReactNodeArray | ReactPortal | Element)[]; ... 5 more ...; variant: "tertiary"; }' but required in type 'Pick<ButtonProps & InjectedOuiaProps, "className" | "children" | "accept" | "acceptCharset" | "action" | "allowFullScreen" | "allowTransparency" | "alt" | "as" | "async" | ... 360 more ... | "icon"> 
```

Adding the `css` prop to the Button component fixes that, but that is obviously not the solution to the issue. Apparently this is caused by emotion and how it applies the styles to components.

Both components have css imports in the file but removing them has no effect.

I was looking for some solutions online but without luck. They mostly recommend updating emotion `^10.x`, but that is a very invasive change.

Have you ever encounter this issue?

Hyperkid123:
@redallen @evwilkin I've reverted the changes in components that caused the TS issues. This only affects the topology package. I went through the dist files and there are no imports from `react-core` with the exception of the topology component.

Issue #3454, opened 2020-01-06
Prop descriptions are missing from the docs


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Prop descriptions are missing from the docs

ddonahue007:
It appears this could be caused by build infra changes. These descriptions used to be in place. I have attached a screen shot that shows the missing descriptions on the accordion component.

![Screen Shot 2020-01-16 at 3.23.02 PM.png](https://images.zenhubusercontent.com/5dc5dcfe586229000122b6e7/7cf47c9e-b66c-497e-831c-678eff93f416)

ddonahue007:
This was fixed by the following PR in the patternfly-org repo
https://github.com/patternfly/patternfly-org/pull/1636

ddonahue007:
Need to up the version of the following package to take in the fixes.

packages/patternfly-4/react-docs/package.json
             "gatsby-theme-patternfly-org": "^1.2.9",

PR #3527, opened 2020-01-17
fix(Tabs): update tabs with nav examples + add Tab component to props docs

nicolethoen:
Addresses issue #2607 

Issue (#3454) opened to address the fact that props descriptions are missing (from all docs it seems).

Note: Tab component cannot return null, or the gatsby parser does not recognize that Tab.tsx returns a component to build the props documentation for Tab. Possible alternative solution could be updating our gatsby-transformer-react-docgen-typescript plugin.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/patternfly-4/react-core/src/components/Tabs/Tab.tsx
@@ -18,6 +18,4 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   tabContentRef?: React.RefObject<any>;
 }
 
-export const Tab: React.FunctionComponent<TabProps> = ({ className = '' }: TabProps) => {
-  return null;
-};
+export const Tab: React.FunctionComponent<TabProps> = ({ className = '' }: TabProps) => null;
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -22,6 +22,6 @@ export const Tab: React.FunctionComponent<TabProps> = ({
   className = '',
 }: TabProps) => {
   return (
-    null
+    <React.Fragment/>
```
dlabrecq: Best practice is to return null when there is nothing to render.

"You can return an empty value from a React render method. However return null is the most preferred as it signifies that nothing is returned"

https://stackoverflow.com/questions/42083181/is-it-possible-to-return-empty-in-react-render-function

dlabaj: I agree we should return null.

kmcfaul: Adding the reversion as Nicole is out. 

@redallen Is it possible to update our docgen to detect the Tab properties while maintaining the null return?

PR #3531, opened 2020-01-20
feat(AlertGroup): Add alert group component

### LINKED ISSUE #3528: PF4: Alert Group
Add the Alert Group feature to PF4
https://pf4.patternfly.org/documentation/core/components/alertgroup

jessiehuff:
**What**: Closes #3528

**Additional issues**: #3519 should go in first to allow for accessible toast alerts


packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (!canUseDOM) {
+      return null;
+    }
+    if (!this.container) {
+      this.container = document.createElement('div');
+    }
+    const { className, children, isToast, ...rest } = this.props;
+    const alertGroup = (
+      <ul
+        className={css(styles.alertGroup, className, (isToast ? 'pf-m-toast' : ''))}
```
tlabaj: Why are you using the hard coded class name here rather than the react-styles css function?

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class AlertGroup extends React.Component<AlertGroupProps> {
+  container?: HTMLDivElement = undefined;
+  constructor(props: AlertGroupProps) {
+    super(props);
+  }
+  componentDidMount() {
+    if (this.container) {
+      document.body.appendChild(this.container);
```
jschuler: I think we should add a prop so user can define where this should be appended to, like in AboutModal something like:`appendTo?: HTMLElement | (() => HTMLElement);`

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
jschuler: should be able to use the imported styles object for the class name strings here and below

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
jessiehuff: Sounds good to me! :)

packages/patternfly-4/react-core/src/components/AlertGroup/examples/alert-group.css
```diff
@@ -0,0 +1,3 @@
+.ws-react-c-alertgroup .pf-c-alert-group__item {
+  margin-bottom: 21px;
```
mcoker: Why do we need this? It isn't an accurate representation of alert group spacing.

jessiehuff: I saw it in the alert component and thought we'd want to keep the same spacing. Happy to take it out! :)

mcarrano:
@jessiehuff This is definitely looking better. Just a couple of remaining issues:

* For the "Singular dynamic alert group", can we change the styling of those to inline alerts?
* In the Async alert group, doesn't seem like the Stop button is working. Could not stop the alerts.

packages/patternfly-4/react-core/src/components/Alert/Alert.test.tsx
```diff
@@ -120,7 +120,7 @@ Object.values(['success' as AlertVariant]).forEach(variant => {
     test('Toast alert live regions are not atomic', () => {
       const wrapper = mount(
         <Alert
-          isToast={true}
+          isLiveRegion={true}
```
tlabaj: I like this prop name change!

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface AlertGroupProps extends Omit<React.HTMLProps<HTMLUListElement>, 'className'> {
+  /** Additional classes added to the AlertGroup */
+  className?: string;
+  /** Alerts to be rendered in the AlertGroup */
+  children?: React.ReactNode;
+  /** Toast notifications are positioned at the top right corner of the viewport */
+  isToast?: boolean;
+  /** Determine where the alert is appended to */
+  appendTo?: HTMLElement | (() => HTMLElement);
```
jschuler: looks like appendTo is only invoked as a function below, so this should only take an HTMLElement or the code is updated below with something like:
```
getModalRoot = (appendTo: HTMLElement | (() => HTMLElement)) => {
    if (typeof appendTo === 'function') {
      return appendTo();
    }
    return appendTo || document.body;
  };
```

Also appendTo should allow the user to append the modal to any element not just document.body, so i think this needs to be changed a bit. To put it all together, something like:

```
constructor(props) {
    super(props);
    this.el = document.createElement('div');
  }

componentDidMount() {
  getModalRoot().appendChild(this.el);
}

componentWillUnmount() {
  getModalRoot(). removeChild(this.el);
}

getModalRoot = (appendTo: HTMLElement | (() => HTMLElement)) => {
    if (typeof appendTo === 'function') {
      return appendTo();
    }
    return appendTo || document.body;
  };

render() {
  const alertGroup = ...
    return ReactDOM.createPortal(
      alertGroup,
      this.el,
    );
  }
```

jschuler: Might be good to create an example / test for it as well

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -0,0 +1,63 @@
...
+  getTargetElement() {
+    const appendTo = this.props.appendTo;
+    if (typeof appendTo === 'function') {
+      return appendTo();
+    }
+    return appendTo || document.body;
+  };
+
+  render() {
+    const alertGroup = <AlertGroupInline {...this.props} />;
```
jschuler: Great changes! One suggestion:
Could you spread out the props that you need and pass them explicitly?
e.g. something like
```
const { className, children, isToast } = this.props;
const alertGroup = <AlertGroupInline className={className} isToast={isToast}>{children}</AlertGroupInline>
```
Then you also don't need the appendTo prop in AlertGroupInline

packages/patternfly-4/react-core/src/components/AlertGroup/AlertGroupInline.tsx
```diff
@@ -0,0 +1,20 @@
...
+  children,
+  isToast,
+  appendTo, // Don't spread
+  ...rest
+}: AlertGroupProps) => (
+  <ul
+    className={css(styles.alertGroup, className, (isToast ? styles.modifiers.toast : ''))}
+    {...rest}>
+    {React.Children.toArray(children).map(
+      (Alert: React.ReactNode, index: number) => <li className="pf-c-alert-group__item" key={index}>{Alert}</li>
```
jschuler: Is there `styles.alertGroupItem` we can use here?

PR #3534, opened 2020-01-21
fix(pagination): calculate navigation input according to last page

### LINKED ISSUE #3530: The input box in the Pagination component is too small
The input box is very narrow and also occupied by spinners. This makes the pagination bar unable to display triple digit page numbers correctly. 
In my opinion, its width should be calculated while it's rendered, based upon the total page count, rather than setting it to a fixed width.
image

boaz0:
**What**: Closes #3530 

//cc @LeshemNoa

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Pagination/Pagination.tsx
+++ b/packages/patternfly-4/react-core/src/components/Pagination/Pagination.tsx
@@ -6,6 +6,8 @@ import { css } from '@patternfly/react-styles';
 import { Navigation } from './Navigation';
 import { PaginationOptionsMenu } from './PaginationOptionsMenu';
 import { InjectedOuiaProps, withOuiaContext } from '../withOuia';
+import { c_pagination__nav_page_select_c_form_control_width_chars as widthChars } from '@patternfly/react-tokens';
+import { PickOptional } from '../../helpers';
 
 export enum PaginationVariant {
   top = 'top',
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -54,6 +56,18 @@ export interface NavigationState {
...
+// Github issue: 3530
+const calculateInputWidth = (lastPage: number) => {
+  const inputControllerWidth = 38;
+  const digitWidth = 8;
+  const digitNo = String(lastPage).length;
+  if (digitNo < 3) {
+    // the current input width fits 2 digits number
+    return null;
+  }
+  return `${inputControllerWidth + digitWidth * digitNo}px`;
```
dlabrecq: I feel like there should probably be a CSS solution for this in pf-core, instead of hard coding inputControllerWidth (38px) and digitWidth (8px) here.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -14,6 +14,8 @@ export interface NavigationProps extends React.HTMLProps<HTMLElement> {
   isDisabled?: boolean;
   /** Flag indicating if the pagination is compact */
   isCompact?: boolean;
+  /** Pagination input width. */
+  inputWidth?: string;
```
dlabrecq: Is there an example showing how the new `inputWidth` prop would be used?

boaz0:
DO NOT MERGE till `@patternfl/patternfly` is updated to `2.58.0` or later.

boaz0:
ci failure unrelated:

```
@patternfly/react-docs: error Generating JavaScript bundles failed
@patternfly/react-docs: Can't resolve '@patternfly/react-core/dist/esm/beta' in '/home/circleci/project/packages/patternfly-4/react-docs/.cache/caches/gatsby-plugin-mdx/mdx-scopes-dir'
```

yarn.lock
```diff
@@ -3821,6 +3821,11 @@
   resolved "https://registry.yarnpkg.com/@patternfly/patternfly/-/patternfly-2.60.0.tgz#fb319c92514706220007437d7c6e11e9383343f8"
   integrity sha512-oGivL/sudNzQ8ERIfghT1MApQumnixdii/pkHh69fRjyjgo50Np+Ne+j3WAO4NtIqBB5nuSvlI5UnBZegYBB0A==
 
+"@patternfly/patternfly@2.58.0":
+  version "2.58.0"
+  resolved "https://registry.yarnpkg.com/@patternfly/patternfly/-/patternfly-2.58.0.tgz#699f811e12a2372af4b930800826b16b81a280f6"
+  integrity sha512-evVE9/5BJWznY8yXYk0mxq9Cws84GFNLPF9OIaHuOSDIcy8HTQ30bTsnLqrTJLrJ5+o+duYttaNO/buskGJNDw==
+
```
dlabrecq: Everything else looks ok, but we don't want to update the lock file. That is, we rely on the build automation to bump the patternfly version for all packages.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -125,6 +127,15 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
       this.state.userInputPage !== this.props.page
     ) {
       this.setState({ userInputPage: this.props.page });
+      if (this.inputRef) {
+        if (this.props.lastPage >= 100) {
+          const charLength = String(this.props.lastPage).length;
+          this.inputRef.current.style.setProperty(
+            '--pf-c-pagination__nav-page-select--c-form-control--width-chars',
```
redallen: I think there is a react-style for this now.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -125,6 +127,15 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
...
+            String(charLength)
```
redallen: What happens without casting to a string? I think `setProperty` will still work.

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -125,6 +127,15 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
       this.state.userInputPage !== this.props.page
     ) {
       this.setState({ userInputPage: this.props.page });
+      if (this.inputRef) {
+        if (this.props.lastPage >= 100) {
```
jschuler: would be good to explain in a comment why this is checked for

packages/patternfly-4/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -111,13 +117,27 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
     }
   }
 
+  handleInputWidth = () => {
+    const { lastPage } = this.props;
+    const len = String(lastPage).length;
+    if (len >= 3) {
+      if (this.inputRef.current) {
+        this.inputRef.current.style.setProperty(inputTotalWidth.name, `calc(${inputWidthBase.value} + ${len} * 1ch)`);
```
redallen: Is there not a way to solve this with normal CSS upstream in https://github.com/patternfly/patternfly-next? If not, resorting to CSS-in-JS is okay with me.

dlabrecq: This looks like a style that belongs in pf-core

mcoker: To set the input width, from the solution in this comment https://github.com/patternfly/patternfly-react/issues/3530#issuecomment-576937340, you want to use `setProperty` to define `--pf-c-pagination__nav-page-select--c-form-control--width-chars` with the number of characters you want the input to display. The `calc()` and `1ch` conversion happens in the CSS here:

https://github.com/patternfly/patternfly-next/blob/72790c83d8cc1704f831e42ba06bf828c67d029f/src/patternfly/components/Pagination/pagination.scss#L25

boaz0: I tried to update `--pf-c-pagination__nav-page-select--c-form-control--width-chars` but it didn't work maybe you can find why.

mcoker: What element are you calling `setProperty()` on? Currently it will need to be added to the element with `.pf-c-pagination` (the component's outer most container)

boaz0: @mcoker thanks - I did as you said and it works now!

PR #3555, opened 2020-01-23
fix(Dropdown): Set a random id for dropdown item if id is not given

### LINKED ISSUE #3549: DropdownItems have bad id, cause console warnings
Using DropdownItems creates items with duplicate invalid id settings causing warnings in the console:

```
t-s:1 [DOM] Found 7 elements with non-unique id #: (More info: https://goo.gl/9p2vKq) 
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…​
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…
<button tabindex=​"-1" type=​"button" class=​"pf-c-dropdown__menu-item" id>​…
```

boaz0:
**What**: Closes #3549 

//cc @jeff-phillips-18

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.tsx
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.tsx
@@ -66,8 +66,8 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
       keyHandler: () => {},
       sendRef: () => {}
     },
-    id: '',
-    componentID: '',
+    id: undefined,
+    componentID: undefined,
     enterTriggersArrowDown: false
   };
 
--- a/packages/patternfly-4/react-core/src/components/ApplicationLauncher/__snapshots__/ApplicationLauncher.test.js.snap
+++ b/packages/patternfly-4/react-core/src/components/ApplicationLauncher/__snapshots__/ApplicationLauncher.test.js.snap
@@ -294,25 +294,21 @@ exports[`ApplicationLauncher custom icon 1`] = `
                     role="menu"
                   >
                     <li
-                      id=""
                       role="menuitem"
                     >
                       <a
                         aria-disabled="false"
                         class="pf-c-app-launcher__menu-item"
-                        id=""
                         tabindex="-1"
                       >
                         Link
                       </a>
                     </li>
                     <li
-                      id=""
                       role="menuitem"
                     >
                       <button
                         class="pf-c-app-launcher__menu-item"
-                        id=""
                         tabindex="-1"
                         type="button"
                       >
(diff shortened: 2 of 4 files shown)


jeff-phillips-18:
Could we simply not set the id if not given? So either remove the default props for `id` and `componentId` or, if linter doesn't allow it, then do:

`id={id || undefined}`

boaz0:
@jeff-phillips-18 - yes, that makes more sense. Thanks for the feedback.

packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -4,6 +4,9 @@ import { DropdownContext } from './dropdownConstants';
 import { KEY_CODES, KEYHANDLER_DIRECTION } from '../../helpers/constants';
 import { Tooltip } from '../Tooltip';
 import styles from '@patternfly/react-styles/css/components/Dropdown/dropdown';
+import GenerateId from '../../helpers/GenerateId/GenerateId';
+
+let currentId = 0;
```
jeff-phillips-18: This change is no longer needed.

packages/patternfly-4/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -66,8 +66,8 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
       keyHandler: () => {},
       sendRef: () => {}
     },
-    id: '',
-    componentID: '',
+    id: null,
+    componentID: null,
```
dlabrecq: Instead of setting this to null, couldn't we just not set a default? That is, remove this from `defaultProps`.

We shouldn't need to test `componentID || undefined` because it would already be undefined.

jeff-phillips-18: that was my thought as well, but if i recall the lint settings require optional props to have defaults, though `undefined` may be a valid default 🤷‍♂

dlabaj: yeah if you remove the default props here then it will just default to undefined and be fine.

PR #3556, opened 2020-01-23
Bump victory 34.0.x

jotak:
I'd like to bump victory charts to 34.0.x, to get a patch I did there: https://github.com/FormidableLabs/victory/pull/1474

This will enable having better / more customized tooltips. Like here in kiali: 

Capture d’écran de 2020-01-20 19-36-11

This version upgrade (33.x to 34.x) is supposedly a breaking change but I'm not sure if it can have any impact on patternfly. There's actually just a single breaking change: https://github.com/FormidableLabs/victory/blob/master/CHANGELOG.md#3400-2019-12-20
It's about using react's context API. I've built and did some smoke tests locally without seeing anything broken.

### DIFF
--- a/packages/patternfly-4/react-charts/package.json
+++ b/packages/patternfly-4/react-charts/package.json
@@ -34,9 +34,8 @@
     "@patternfly/react-tokens": "^2.7.20",
     "hoist-non-react-statics": "^3.3.0",
     "lodash": "^4.17.15",
-    "victory": "^33.0.5",
-    "victory-core": "^33.0.1",
-    "victory-legend": "^33.0.1"
+    "victory": "^34.0.1",
+    "victory-core": "^34.0.0"
   },
   "peerDependencies": {
     "prop-types": "^15.6.1",
@@ -62,7 +61,7 @@
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
     "@types/lodash": "^4.14.138",
-    "@types/victory": "^31.0.22",
+    "@types/victory": "^33.1.4",
     "babel-plugin-transform-es2015-modules-umd": "^6.24.1",
     "babel-plugin-typescript-to-proptypes": "^0.17.1",
     "css": "^2.2.3",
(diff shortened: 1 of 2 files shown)


jotak:
> I've build Cost Management with 34.0, which seems to work fine.
> 
> My concern is that this will force app's to update their own dependency? For example, I'd likely need to update Cost Management's Victory packages.
> 
> Perhaps we can change the version to "^33.0 | ^34.0"?
> 

Not sure to understand, I don't know how Cost Management dependencies are managed, isn't it a simple dependency to PF that transitively pulls victory? So victory packages are automatically updated when you update patternfly?

> That said, the `activePoints` prop would not be available in ChartTooltip. We would need to add the new prop there to make it available to users. In that case, we would need to update to 34.0.

This is not actually the Tooltip component that gets this new props, but the internal Flyout and Label components... which can be customized. So it doesn't necessitate changes in pf's `ChartTooltip` to make use of.
In Kiali we define a CustomFlyout and a CustomLabel passed to ChartTooltip:

```
      <ChartTooltip
        flyoutComponent={<CustomFlyout />}
        labelComponent={<CustomLabel colors={colors} />}
        constrainToVisibleArea={true}
      />
```

So it's in CustomFlyout and CustomLabel that we can get `activePoints` and make the magic happen.

dlabrecq:
It appears that VictoryTooltip has a new `activePoints` prop. Are you suggesting we don't want to sync this with ChartTooltip props?

https://github.com/jotak/victory/blob/6993f657670d554c433cc17d9313bde87d07eed4/packages/victory-tooltip/src/victory-tooltip.js#L27

jotak:
@dlabrecq sorry, yes there's this new props what I wanted to say is that it's not *required* to have it formalized in `ChartTooltip` to have the whole thing working, because this props is actually injected by the `VictoryVoronoiContainer` and not by the user directly. This is more like an "internal prop", as several similar prop exist in Victory, that doesn't have to be exposed to the user in ChartTooltip (actually I don't see how the user would provide it, unless by writing its own container).

redallen:
Breaking release change time is here! Rebase on the [`v4` branch](https://github.com/patternfly/patternfly-react/tree/v4) and open a new PR against the [`v4` branch](https://github.com/patternfly/patternfly-react/tree/v4) and we can take this in.

PR #3567, opened 2020-01-23
fix(Table): Fixes issue 3559, for table we now default to a div i…

### LINKED ISSUE #3559: DropdownItem defaults to anchor component
`DropdownItem` `component` prop defaults to anchor link. This makes the DOM invalid if we were to provide an anchor to a DropdownItem without changing component prop.

Table component uses the default anchor link component instead of changing it to button, making it impossible to have an anchor link in a dropdown without getting errors

Example of anchor link in anchor link: https://codesandbox.io/s/determined-nobel-1eqjh?fontsize=14&hidenavigation=1&theme=dark

1. Click ellipsis
1. see error in console

Bug

dlabaj:
This fixes the double anchor tag that was showing up in the table.

fix #3559


SpyTec:
By just changing to div, we loose the cursor pointer on normal actions. Do we need to change styling that would apply to all `DropdownItems`?

PR #3572, opened 2020-01-24
fix(wizard): add typeto wizard context

### LINKED ISSUE #3545: WizardContext props are `{}` when using typescript

When trying to use the WizardContextProvider, the props are actually `{}` (see: [here](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Wizard/WizardContext.ts#L3)) and we can only guess the values (or inspect [here](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Wizard/Wizard.tsx#L451-L458))

Add a custom footer using typescript and try to use any attribute of the WizardContextProvider
```tsx
<WizardContextConsumer>
                { (props) => { // props type is `{}`
                }}
</WizardContextConsumer>
```

Bug, the workaround would be to use the following

```tsx
interface WizardContextProps {
    goToStepById: (stepId: number | string) => void;
    goToStepByName: (stepName: string) => void;
    onNext: () => void;
    onBack: () => void;
    onClose: () => void;
    activeStep: WizardStep;
}

<WizardContextConsumer>
                { (_props) => { // props type is `{}`
                    const props = _props as WizardContextProps;
                }}
</WizardContextConsumer>

```

Custom Policies https://github.com/RedHatInsights/custom-policies-ui-frontend

I believe that [this](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Wizard/WizardContext.ts#L3) should have something similar to:
```tsx
const WizardContext = React.createContext<WizardContextProps>({});
```

boaz0:
**What**: Closes #3545

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Wizard/WizardContext.ts
+++ b/packages/patternfly-4/react-core/src/components/Wizard/WizardContext.ts
@@ -1,5 +1,25 @@
 import * as React from 'react';
+import { WizardStep } from './Wizard'
 
-const WizardContext = React.createContext({});
+
+export interface WizardContextType {
+    goToStepById: (stepId: number | string) => void;
+    goToStepByName: (stepName: string) => void;
+    onNext: () => void;
+    onBack: () => void;
+    onClose: () => void;
+    activeStep: WizardStep;
+}
+
+const WizardContext = React.createContext<WizardContextType>({
+    goToStepById: () => null,
+    goToStepByName: () => null,
+    onNext: () => null,
+    onBack: () => null,
+    onClose: () => null,
+    activeStep: {name: null}
+
+});
+
 export const WizardContextProvider = WizardContext.Provider;
 export const WizardContextConsumer = WizardContext.Consumer;
--- a/packages/patternfly-4/react-core/src/components/Wizard/Wizard.test.tsx
+++ b/packages/patternfly-4/react-core/src/components/Wizard/Wizard.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
-import { Wizard, WizardStep, WizardStepFunctionType } from './Wizard';
+import { Wizard, WizardStepFunctionType, WizardStep  } from './Wizard';
 
 test('Wizard should match snapshot', () => {
   const steps: WizardStep[] = [


packages/patternfly-4/react-core/src/components/Wizard/WizardContext.ts
```diff
@@ -1,5 +1,46 @@
...
+  steps?: WizardStep[];
+  /** (Unused if footer is controlled) Can change the Next button text. If nextButtonText is also set for the Wizard, this step specific one overrides it. */
+  nextButtonText?: string;
+  /** (Unused if footer is controlled) The condition needed to enable the Next button */
+  enableNext?: boolean;
+  /** (Unused if footer is controlled) True to hide the Cancel button */
+  hideCancelButton?: boolean;
+  /** (Unused if footer is controlled) True to hide the Back button */
+  hideBackButton?: boolean;
+}
```
dlabrecq: I don't know that we can move the WizardStep interface without introducing a breaking change. Won't this change where WizardStep is imported from?

josejulio: Users of this lib import it from the top-level [index.ts](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/index.ts). 
e.g.
```ts
import {
    WizardContextConsumer,
    WizardStep
```

Since that's the document way to import stuff from react-core, i think is not a breaking change :). - Also, after re-reading the docs, I think this `WizardStep` is not even documented (I didn't find it [here](https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/wizard), only references that it exists in some types) - Is there a `typescript` version of these docs?

dlabrecq: Lack of doc wouldn't negate a breaking change. However, as long as it doesn't change the import path, I'm fine with it.

jschuler: this interface is not context specific, can we leave this in Wizard.tsx? Doesn't have to be moved

packages/patternfly-4/react-core/src/components/Wizard/WizardContext.ts
```diff
@@ -1,5 +1,46 @@
...
+export interface WizardContextType {
+    goToStepById: (stepId: number | string) => void;
+    goToStepByName: (stepName: string) => void;
+    onNext: () => void;
+    onBack: () => void;
+    onClose: () => void;
+    activeStep: WizardStep;
+}
+
+const WizardContext = React.createContext<WizardContextType>({
```
jschuler: 👍 good addition of adding the type here thanks

Issue #3357, opened 2019-11-27
Example usage of table inline editing


dalaicam:
Cheers,

Was trying to integrate the table inline editing in the project I'm currently in, but having a bit of difficulty to understand how to actually include and use it properly.
The project is being developed using Typescript and React.

Was following the examples as base to have the component integrated, though can't advance much since getting the error:
'EditableTableBody' only refers to a type, but is being used as a value here.ts(2693)

eg.
```
import React from 'react'
import { Table, TableHeader} from '@patternfly/react-table';
import {
  EditableTableBody
} from '@patternfly/react-inline-edit-extension';

export default class InlineEditTableTest extends React.Component {

  render() {
    const editConfig = {
    };

    return (<Table
        aria-label="Editable Table"
        cells={[]}
        rows={[]}
      >
        <TableHeader />
       <EditableTableBody editConfig={editConfig} />
      </Table>
    )
  }
}
```

This error happen to all other components related with the table inline editing.

Not sure if the way we are using is not the correct or if an issue in the configuration.
If not issue, would it be possible to provide a small example of how it is expected to be used with Typescript?

Thanks for the help

rachael-phillips:
Hi @dalaicam  our team has seen this, and we will be taking a look at it this week.

dalaicam:
Were you able to check the situation?

Thanks for the help

rachael-phillips:
@dalaicam thank you for following up! Is this currently blocking you? We are going to be introducing an inline edit feature, and that could be worth waiting on since we aren't offering support for the extension at this time.

dalaicam:
@rachael-phillips thank you for the quick response.
Currently this feature is not blocking the development, but the sooner we have it implemented/available the better as we will be able to test and validate if it's necessary to make adjustments to the application (user feedback, etc...).

When are you predicting to have this feature included and available (at least as experimental) in the API?

Once again, thank you for the help and attention

seanforyou23:
Hi @dalaicam - it has taken a while but we just merged the new inline edit feature. It's supported in core and react, and now supersedes the previous inline-edit-extension. Since it's a critical new feature, we've marked it as `beta`, and for a while, we'll probably leave the inline edit extension around in case people are still using it and want to test out the new version before migrating. I'd love any feedback you may have, feel free to give it a try when you have time.

[...]

Issue #3058, opened 2019-10-02
inline edit for single row (with no expand) 


LHinson:
As a follow up to the issue https://github.com/patternfly/patternfly-next/issues/2315

rachael-phillips:
Can I close this?

seanforyou23:
Yep, I was just about to!

Issue #2679, opened 2019-08-09
PF4 - React Table - Inline Edit Feature


priley86:
In this issue, I will try to capture additional needs for the upcoming Inline Edit feature 😸 . Please feel free to append...

This issue will include the following:

- Provide an implementation of the new Inline Editable table coming to PF Core. Link to [PF Core issue](https://github.com/patternfly/patternfly-next/issues/404).
- Refactor inline editable row and inline editable table to meet latest design spec and Core HTML/CSS (following). This will involve revisiting and refactoring the existing source in `@patternfly/react-inline-edit-extension`. Special consideration should be taken to ensure this is now consistent w/ our current React-Table implementation.
- Use TypeScript and available React-Table types/interfaces exposed in #2360 where possible. [Link to a relevant comment](https://github.com/patternfly/patternfly-react/pull/2360#discussion_r312588396) and [another comment](https://github.com/patternfly/patternfly-react/pull/2360#discussion_r312494646).
- Expose the new InlineEditable source from `@patternfly/react-table` instead of `@patternfly/react-inline-edit-extension`. We'd like to make this part of the Core distribution of React Table going forward.

This may also include additions to the Toolbar in React-Core for the new design as well as demos integrated w/ the Inline Edit Toolbar options in the Inline Edit examples.

mturley:
Just commenting here so this issue ends up in my "participating" filter. I will definitely want to get involved in this, I wrestled with the PF3 inline edit feature quite a bit.

johnsonm325:
Attaching new issue for future implementation on this issue:

https://github.com/patternfly/patternfly-react/issues/2808

mturley:
Noticed that this issue was marked as stale but there is ongoing work for this in core (https://github.com/patternfly/patternfly-next/issues/2315, https://github.com/patternfly/patternfly-next/issues/2319). @mattnolting @tlabaj @rachael-phillips @LHinson should we keep this issue open or close it until the current core work is ready for a React implementation?

rachael-phillips:
@mturley, yes we should leave it open.

PR #3581, opened 2020-01-24
Inline edit for Table

### LINKED ISSUE #3367: Type mismatch in RowWrapper
I've noticed this type error, spent a bit of time trying to fix it but no luck so far. Using visual studio code, it flags this in standard fashion.

Screen Shot 2019-12-02 at 8 18 38 PM

When I remove the [ouia context](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/RowWrapper.tsx#L106-L109) bits the error goes away. It's not affecting anything directly, but I'm looking to extend the row wrapper code for inline edit and want to make sure the testing related types are correctly attached so the new features can be tested.

seanforyou23:
**What**: This PR adds row-level inline edit feature and supporting tests to react-table and react-integration packages. It also fixes an issue where rows used non-unique react keys by default and a type issue in RowWrapper, along with some other type enhancements that (hopefully) don't introduce any breaking changes. We've also added a row-level validation hook which can be used to [apply arbitrary rulesets](https://github.com/patternfly/patternfly-react/pull/3581/files#diff-f6c0274c6d946d8f7472804d7595cf78R1449-R1470) for any given row as defined by the user. This validation hook only runs on submission of the row edit, not onChange of the individual inputs, there are plans to add validation onBlur in a future effort.

Should help close https://github.com/patternfly/patternfly-react/issues/3367 https://github.com/patternfly/patternfly-react/issues/2679 https://github.com/patternfly/patternfly-react/issues/3357 and https://github.com/patternfly/patternfly-react/issues/3058


mcarrano (changes requested):
I did not test extensively, but from what I can see, the behavior looks good @seanforyou23 There was just one visual thing I'm questioning. Why doesn't the edit (pencil) icon align vertically with the kabob in the normal state? 

Screen Shot 2020-01-24 at 3 42 24 PM

Also, I think the example would make more sense if all of the rows were initialized to non-edit state since that's normally how the table would be presented. A user could then click the edit icon to place any or all rows into edit mode.

packages/patternfly-4/react-table/src/components/Table/EditColumn.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
```
tlabaj: I wonder if this actually belongs in the experimental package?

mattnolting (dismissed):
LGTM! Great work @seanforyou23!

packages/patternfly-4/react-table/src/components/Table/utils/utils.tsx
```diff
@@ -33,3 +33,73 @@ export const isRowExpanded = (row: IRow, rows: IRow[]) => {
...
+    const hasEditableValue = cell.props.editableValue !== undefined && cell.props.editableValue !== null;
+
+    if (cell.props && hasValue && hasEditableValue) {
+      if (type === 'save') {
+        const errorMsg = Object.keys(validationErrors)
+          .filter((validatorName) => validationErrors[validatorName].includes(cell.props.name))
+          .map((validatorName) => getErrorTextByValidator(validatorName, row.rowEditValidationRules));
+
+        if (errorMsg.length) {
+          cell.props.errorText = cell.props.name ? errorMsg.join(', ') : 'Validation requires unique name property for row cells';
```
tlabaj: It looks like the function is only being called by the demo, but might a consumer call it?  If so, is there any way to localize these strings?

packages/patternfly-4/react-table/src/components/Table/utils/utils.tsx
```diff
@@ -33,3 +33,73 @@ export const isRowExpanded = (row: IRow, rows: IRow[]) => {
...
+    cell.props.isValid = true;
+  });
+
+  row.isEditable = !row.isEditable;
+  row.isValid = true;
+  return row;
+};
+
+export const validateCellEdits = (row: IRow, type: RowEditType, validationErrors: RowErrors) => {
+  row.isValid = !!Object.keys(validationErrors).length ? false : true;
```
kmcfaul: Does the `!!` do something specific here? It seems like it can be removed.

packages/patternfly-4/react-table/src/components/Table/EditColumn.tsx
tlabaj: I am not well versed on the Table.  Can this not be an experimental component?

packages/patternfly-4/react-table/src/components/Table/utils/utils.tsx
boaz0: Maybe we can use `!Boolean(Object.keys(...).length)`

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -1202,3 +1206,420 @@ EmptyStateTable = () => {
...
+              name: 'required',
+              validator: value => value.trim() !== '',
+              errorText: 'This field is required'
+            }
+          ],
+          cells: [
+            {
+              title: (value, rowIndex, cellIndex, props) => (
+                <React.Fragment>
+                  <div className="pf-c-inline-edit__value">{value}</div>
```
tlabaj: Why are we doing this inline like this?  Shouldn't it be part of an inline edit component?

tlabaj (commented):
So looking at this

packages/patternfly-4/react-table/src/components/Table/Table.md
kmcfaul: It's because Table does not edit or peer into its cells at all, the way it is currently. And the functional rendering allows the user to control what the input/value fields look like and how they behave.

I do believe we can wrap the whole `React.Fragment` into a simple component, to serve as a 'default' behavior/look.

packages/patternfly-4/react-table/src/components/Table/EditColumn.tsx
kmcfaul: It should be experimental. But I am confused on what that means for the overall `Table`. Does the whole thing get marked as experimental suddenly, or do we duplicate the whole folder into experimental and have two copies of Table while inline-edit is 'experimental'?

packages/patternfly-4/react-table/src/components/Table/Table.md
seanforyou23: Right, what @kmcfaul said :D - I created a default component that people can get up and running with, can be replaced with a custom component as users see fit. Definitely makes things cleaner.. let me know if you see any issues!

packages/patternfly-4/react-table/src/components/Table/EditColumn.tsx
jschuler: How about prefixing relevant props with `experimental`? e.g. in the Table component
```
export type ExperimentalOnRowEdit
experimentalOnRowEdit?: ExperimentalOnRowEdit
```

seanforyou23: We're looking into flagging the feature as experimental since the Table component itself is already available for production use.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEditableDemo.tsx
```diff
@@ -0,0 +1,304 @@
...
+      ],
+      rows: [
+        {
+          isEditable: true,
+          rowEditValidationRules: rowLevelValidationRules,
+          cells: [
+            {
+              title: (value: string, rowIndex: number, cellIndex: number, props: any) => (
+                <React.Fragment>
+                  <div className="pf-c-inline-edit__value">{value}</div>
```
dlabaj: Why are you adding css into the integration app?

packages/patternfly-4/react-table/src/components/Table/Table.md
tlabaj: I still am not understanding.  We want the consumers to add the pf classnames on there own?

dlabaj: Having the user specify the className like this doesn't seem like something we want to promote.  How have we handled this before in the table?  I don't recall us having to specify classNames like this with the table for components that are released as part of the PF4 react library ... should this be an InlineEditValue component... or have another component wrapped around this where we set this styling by passing in a prop?

packages/patternfly-4/react-table/src/components/Table/EditColumn.tsx
tlabaj: Can we also at least mark the associated props related to the feature (eg isEditable) as experimental?

dlabaj (dismissed):
Took a look what was done in core... this doesn't seem like it maps easily to a react component.  Might want to go back to core to reevaluate the implementation.  Seems more like a modifier then a component... or some combination of both.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEditableDemo.tsx
```diff
@@ -0,0 +1,304 @@
...
+                    <TextInput
+                      value={props.editableValue !== undefined ? props.editableValue : value}
+                      isValid={props.isValid}
+                      type="text"
+                      onChange={(value, event) => {
+                        this.handleTextInputChange(value, event, rowIndex, cellIndex);
+                      }}
+                      aria-label="Row 1 cell 1 content" />
+                    <div className="pf-c-form__helper-text pf-m-error" aria-live="polite">{props.errorText}</div>
+                  </div>
```
dlabrecq: +1. I'd like to see another TextInupt variation, specifically for inline edit, rather than manually wrap component with pf-c-inline-edit__input

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEditableDemo.tsx
```diff
@@ -56,20 +57,13 @@ export class TableEditableDemo extends React.Component<TableProps, { columns: (I
...
-                    <div className="pf-c-form__helper-text pf-m-error" aria-live="polite">{props.errorText}</div>
-                  </div>
-                </React.Fragment>
+                <EditableTextCell
+                  value={value}
+                  rowIndex={rowIndex}
+                  cellIndex={cellIndex}
+                  props={props}
+                  handleTextInputChange={this.handleTextInputChange}
+                  inputAriaLabel="Row 1 cell 1 content" />
```
dlabrecq: Can an editable cell be invalid, do we need an isValid prop?

seanforyou23: Yep! `isValid` is actually part of that props object, EditableTextCell component looks for this and updates the TextInput accordingly.

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEditableDemo.tsx
seanforyou23: I asked the core folks about creating a dedicated variation of TextInput for inline edit, but it didn't seem desirable. When you initially left this comment, I hadn't gotten a chance to make use of that `EditableTextCell` wrapper component I created yet - just added it, thinking it will help ease the concern here. Let me know if you see any issues with this approach? Thx!

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEditableDemo.tsx
seanforyou23: The markup used here in react-integration was out of date, have since replaced with a wrapper component `EditableTextCell` to make it easier to work with and less error-prone for users. Is it ok?

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/TableDemo/TableEditableDemo.tsx
dlabrecq: Yes, I like this better

mcarrano (changes requested):
I am still seeing the Edit icon bottom justified in the row. Should be top-justified like the Kabob icon.

packages/patternfly-4/react-table/src/components/Table/EditColumn.tsx
seanforyou23: Looks like `onRowEdit` is the only one of the new beta/experimental props that's listed in the props table for Table component. I've added the new badges for this prop and the inline edit example - lemme know what you think!

dlabrecq (commented):
The edit icon position still appears to be off a bit

Screen Shot 2020-02-12 at 1 29 26 PM

mcarrano (dismissed):
Looks great @seanforyou23 !

packages/patternfly-4/react-table/src/components/Table/Table.md
```diff
@@ -1269,3 +1274,312 @@ EmptyStateTable = () => {
   );
 }
 ```
+
+```js title=Editable-Rows isBeta
```
tlabaj: Can you make this title sentence case please.
I noticed that some of the other Table example titles are not sentence case either.  Can you fix those while you are in that file?

kmcfaul (dismissed):

mcarrano (dismissed):
Still looks good!

kmcfaul (approved):

dlabrecq (approved):

mcarrano (approved):
One more time...

tlabaj (approved):
LGTM

PR #3593, opened 2020-01-27
fix(a11y): fix a11y errors and enable a11y in CI

### LINKED ISSUE #3591: Fix a11y errors & enforce for PR's
Enforce a11y tests to pass for new PR's submitted, and fix any existing errors as part of this process.

redallen:
**What**: Closes #3591

## Components
- Add `role="listbox"` to outer `ul` on **Datalist** so that `aria-checked` attribute on an `li` is valid
  - Use `aria-checked` instead of `aria-selected` on the `li` per @jessiehuff 's suggestion since current **Datalist** logic only allows a single selected item
- Add `linkComponentProps` prop to **LoginPage** to allow adding an `aria-label` to `<LinkComponent>`
- Make **Paginiation** and its OptionsToggle have unique IDs using some global counters
- Fix misc examples that have duplicate ids
- Fix misc examples that lack aria labels

## CI
- Enable a11y tests failing build
- Remove [`start-server-and-test`](https://www.npmjs.com/package/start-server-and-test) in favor of [`background: true`](https://circleci.com/docs/2.0/configuration-reference/#background-commands) in CircleCI
- Use `resource_class: xlarge` for a few build steps now that we have a paid CircleCI plan
  - Remove machine instance used for doc builds
- Cache Gatsby folders for faster doc build
- Remove `upload_docs` job in favor of uploading after any step that has a report (a11y or pf3/pf4)
  - Add `upload-preview.js` to use Surge, can use S3 in the future. Has logic to leave Github comments as artifacts are uploaded like this:
> PF3 preview: https://patternfly-react-pr-3593-pf3.surge.sh
> PF4 preview: https://patternfly-react-pr-3593.surge.sh
> A11y report: https://patternfly-react-pr-3593-coverage.surge.sh

PR workflow now looks like this (~9m total):
image


packages/patternfly-4/react-core/src/experimental/components/DataToolbar/examples/DataToolbar.md
```diff
@@ -342,7 +344,7 @@ class DataToolbarComponentMangedToggleGroup extends React.Component {
     };
     
     this.clearRiskSelection = () => {
-      this.setState({
+      this.setState({Edit
```
jessiehuff: I think this is a typo? :)

PR #3595, opened 2020-01-27
fix(Page & PageHeader): adds optional "role" prop

### LINKED ISSUE #2580: Configurable roles for Page and PageHeader

jenny-s51:
**What**: Closes #2580. 

**Additional issues**: Following a discussion with @seanforyou23, it was decided that instead of removing the `role` attribute entirely (as was done in #3499), we should instead allow the user to optionally configure it using a prop.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Page/Page.tsx
+++ b/packages/patternfly-4/react-core/src/components/Page/Page.tsx
@@ -24,8 +24,6 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   sidebar?: React.ReactNode;
   /** Skip to content component for the page */
   skipToContent?: React.ReactElement;
-  /** A flag that specifies whether the value for role is set on the <main> element */
-  isRoleSet?: boolean;
   /** Sets the value for role on the <main> element */
   role?: string;
   /** an id to use for the [role="main"] element */
@@ -68,8 +66,7 @@ export class Page extends React.Component<PageProps, PageState> {
     defaultManagedSidebarIsOpen: true,
     onPageResize: (): void => null,
     mainContainerId: null as string,
-    isRoleSet: false,
-    role: null as string
+    role: undefined as string
   };
 
   constructor(props: PageProps) {
@@ -133,7 +130,6 @@ export class Page extends React.Component<PageProps, PageState> {
       sidebar,
       skipToContent,
       role,
-      isRoleSet,
       mainContainerId,
       isManagedSidebar,
       defaultManagedSidebarIsOpen,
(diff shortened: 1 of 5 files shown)


packages/patternfly-4/react-core/src/components/Page/Page.tsx
```diff
@@ -147,7 +155,13 @@ export class Page extends React.Component<PageProps, PageState> {
           {skipToContent}
           {header}
           {sidebar}
-          <main role="main" id={mainContainerId} className={css(styles.pageMain)} tabIndex={-1} aria-label={mainAriaLabel}>
+          <main
+            role={isRoleSet ? role : undefined}
```
dlabrecq: I'm confused as to why isRoleSet is necessary as a prop? Considering role is undefined by default, there wouldn't need to test it here. I don't see an example how this is used, unless I'm missing something?

jenny-s51: Yes @dlabrecq that makes sense, you're right. We'd probably never need the prop so I removed it -- thank you for clarifying

packages/patternfly-4/react-core/src/components/Page/Page.tsx
```diff
@@ -63,7 +65,8 @@ export class Page extends React.Component<PageProps, PageState> {
     isManagedSidebar: false,
     defaultManagedSidebarIsOpen: true,
     onPageResize: (): void => null,
-    mainContainerId: null as string
+    mainContainerId: null as string,
+    role: undefined as string
```
tlabaj: Should this default to `main` rather that undefined?

jenny-s51: AXE sometimes throws an error when it's set to `main`, so @seanforyou23 and I decided it was better to remove the default and have the user configure it to whatever they prefer, rather than having it be hard-coded.

tlabaj: @jessiehuff the core doc say this attribute is required.  Is that not the case,  do the docs need to be updated? https://pf4.patternfly.org/documentation/core/components/page#accessibility

jessiehuff: @tlabaj I think with this change, the docs would need updated to reflect it. :)

tlabaj: @jenny-s51  can you open issue in core please

jenny-s51: @tlabaj yes!

PR #3601, opened 2020-01-28
feat(Popover): add min-width override property

### LINKED ISSUE #3192: min-width on popover is buggy
The first example here is what Core's popover looks like with no content pushing the width out.
Screen Shot 2019-10-23 at 12 50 03 PM
This is what the React version looks like. It seems like this is being influenced by the Tippy tooltip css that is applied additionally via the react implementation. It would good to have a min-width for this work as I've seen it look pretty odd on the product side when there is a loading state for a Popover's content and it's really narrow.
Screen Shot 2019-10-23 at 12 49 44 PM

kmcfaul:
**What**: Closes #3192 . Adds in an optional `midWidth` property that will override the default and allow a custom minimum width on the popover.

Update: the default min-width property will be adjusted to match max-width, making both examples match. This PR could still be added in if we want to allow users to customize the width of a popover.

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
+++ b/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
@@ -70,6 +70,8 @@ export interface PopoverProps {
    * Instead, the consumer is responsible for closing the popover themselves by adding a callback listener for the shouldClose prop.
    */
   isVisible?: boolean;
+  /** Minimum width of the popover (default 6.25rem) */
+  minWidth?: string;
   /** Maximum width of the popover (default 18.75rem) */
   maxWidth?: string;
   /** Lifecycle function invoked when the popover has fully transitioned out. */
@@ -167,6 +169,7 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
   }
 
   storeTippyInstance = (tip: TippyInstance) => {
+    if(this.props.minWidth) tip.popperChildren.tooltip.style.minWidth = this.props.minWidth;
     tip.popperChildren.tooltip.classList.add(styles.popover);
     this.tip = tip;
   };
@@ -236,6 +239,7 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
       onShown,
       onMount,
       zIndex,
+      minWidth,
       maxWidth,
       closeBtnAriaLabel,
       distance,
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-core/src/components/Popover/Popover.tsx
```diff
@@ -167,6 +169,7 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
   }
 
   storeTippyInstance = (tip: TippyInstance) => {
+    if(this.props.minWidth) tip.popperChildren.tooltip.style.minWidth = this.props.minWidth;
```
dlabrecq: Shouldn't we use `c_popover_MinWidth` as the default from react-tokens to set this, similar to the default `maxWidth`?

dlabrecq: @mcoker Will pf-core apply a default `minWidth`? Note that this component sets a default `maxWidth`.

mcoker: @dlabrecq are you referring to this? https://github.com/patternfly/patternfly-next/blob/7649c9c0a40c2e6e4f6f1e4d1b9cdc6e50aeac66/src/patternfly/components/Popover/popover.scss#L37-L38

dlabrecq: @mcoker Yes, but we pull it from `@patternfly/react-tokens`. 

The Popover component currently sets a default `maxWdith` using `c_popover_MaxWidth` below. I'm just asking if we should also set a default for `minWidth`?

See https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Popover/Popover.tsx#L126

mcoker: @dlabrecq hmm, do you know why we set that? Doesn't seem like either `minWidth` or `maxWidth` would need to be set by default as they're in the CSS. I would expect that we wouldn't apply either unless it's explicitly defined as a prop, then we just apply whatever the prop value is.

dlabrecq: I don't know the history behind that. If we don't need it, I'm fine with that. 

We should ensure `maxWidth` is removed in the PR, then.

kmcfaul: `maxWidth` is passed to tippy and set on the instance, I think it has to have that default value.

PR #3607, opened 2020-01-28
react-core: Generate snapshot tests for all components 

jschuler:
and move tests and snapshots into the `__tests__` sub directory
merge existing tests into the same files as the generated test file


jschuler:
I'll do separate PRs in the following order:
1. Move existing tests and snapshots into the `__tests__` sub directory
2. Fix some more TS issues that I found through the test generation
3. Generate new tests and add them to `__tests__`

PR #3612, opened 2020-01-29
Column functions are not always equal

### LINKED ISSUE #3620: Using new functions don't trigger a render on the Table rows

I'm providing a new functions to `onCollapse` (The same should apply for other params that are functions) due to some changes in my state while using React hooks, but the new function is not applied to every row.

Check the reproducer:
https://codesandbox.io/s/agitated-cannon-gd2ny

Peek 2020-01-29 23-06

Step 1- Close the first - It closes correctly
Step 2- Close the second - It closes but opens the first. 

Reason of "Step 2" outcome, is because is using an old `onCollapse` function that has an old `state`. The row is not being rendered again with the new `onCollapse` funcition.

Bug, the issue is blocking me, no know workaround when using hooks.

Custom Policies https://github.com/RedHatInsights/custom-policies-ui-frontend

josejulio:
- This is important when using [React hooks](https://reactjs.org/docs/hooks-state.html)

Fixes #3620

In my case, I'm using the Table and passing a `onCollapse` method within a React component that uses hooks.

I need the columns that are a function, but a different one, to be updated, I'm using as example `onCollapse`, but this applies to any other column.

Consider below code:

```ts

interface RowState {
    isOpen: boolean;
    // more stuff
}

export const PolicyTable: React.FunctionComponent<PolicyTableProps> = (props) => {

    const [ rowsState, setRowsState ] = React.useState<Record<string, RowState>>({});

    const onCollapse = (_event, idx: number, isOpen: boolean) => {
        const rowState = { ...rowsState[idx], isOpen };
        const newRowsState = { ...rowsState };
        newRowsState[idx] = rowState;
        setRowsState(setRowsState);
    }

    // Other props not show for simplicity
    return <Table
        onCollapse={ onCollapse }
    />
}

```

With this in place, on first render, all the rows have my initial `rowState` value.
If i click the first cell, only that cell is going to have the updated `rowState` (where it is open) and the others will have the old `rowState` (nothing open).
When clicking the second row, it will discard my modified `rowState` (because this cell doesn't have access to it) and only open what I clicked.

See below gif without the contents of this PR:

without-pr-content

and with this PR content:

with-pr-content

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/base/columns-are-equal.ts
+++ b/packages/patternfly-4/react-table/src/components/Table/base/columns-are-equal.ts
@@ -10,7 +10,7 @@ import { ColumnsType } from './types';
 export function columnsAreEqual(oldColumns: ColumnsType, newColumns: ColumnsType) {
   return isEqualWith(oldColumns, newColumns, (a, b) => {
     if (isFunction(a) && isFunction(b)) {
-      return true;
+      return a === b;
     }
 
     return undefined;


josejulio:
Requesting review from @priley86 as per github's suggestion. Please let me know if I should do something else. Thanks!

PR #3621, opened 2020-01-30
feat(radio/check): add support for optional description

### LINKED ISSUE #3611: Radio/check: support for optional description
https://github.com/patternfly/patternfly-next/pull/2579/files

Adds `div.pf-c-[check/radio]__description` that follows the check/radio and label.

boaz0:
**What**: Closes #3611

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
@@ -22,6 +22,8 @@ export interface CheckboxProps
   id: string;
   /** Aria-label of the checkbox. */
   'aria-label'?: string;
+  /** Description text of the checkbox. */
+  description?: React.ReactNode;
 }
 
 // tslint:disable-next-line:no-empty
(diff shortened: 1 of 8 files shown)


packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -22,6 +22,8 @@ export interface CheckboxProps
   id: string;
   /** Aria-label of the checkbox. */
   'aria-label'?: string;
+  /** Description text of the checkbox. */
+  description?: string;
```
mcoker: is `string` limiting in any way? I tried putting an HTML tag in `description` and it didn't work. I think it's fair for that section to contain links or maybe elements like `<strong>`, `<span>`, etc.

boaz0: Yes, you're right - `string` is limit. I can change that to `React.Node` which allows `<strong>`, `<i>` and etc.

PR #3622, opened 2020-01-30
[WIP] Added PF4 namespace to data-ouia-component-type attributes

### LINKED ISSUE #3626: Add namespacing into data-ouia-component-type attributes

There are cases when a web application can use more than one component library and there could be used similar components such as Dropdown. In order to distinguish components with OUIA in QA suite  `data-ouia-component-type` should contain namespace e.g. `data-ouia-component-type=PF4/Dropdown`

Here spec excerpt https://ouia.readthedocs.io/en/latest/README.html#ouia-component:

> A root level HTML element with a data-ouia-component-type attribute describing a unique name identifying ALL HTML components that can be controlled with the same code or interactions. These identifiers MAY and SHOULD be namespaced when used within a framework, particularly when the name is generic. The delimiter between namespace and type name should be a single / character.

> * e.g. A page that has a special dropdown could choose to name that dropdown as FrameworkA/CustomDropdown. All instances of this FrameworkA/CustomDropdown component MUST be expected to be able to be controlled via the same automation.

quarckster:
In order to distinguish similar components from different libraries  `data-ouia-component-type` attribute should have namespaces.
 Fixes #3626

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
+++ b/packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
@@ -70,7 +70,8 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
       className={customClassName}
       aria-label={ariaLabel}
       {...(ouiaContext.isOuia && {
-        'data-ouia-component-type': 'PF4/Alert',
+        'data-ouia-component-type': 'Alert',
+        'data-ouia-component-ns': 'PF4',
         'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
       })}
       {...(isLiveRegion && {
--- a/packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
+++ b/packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
@@ -70,7 +70,7 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
       className={customClassName}
       aria-label={ariaLabel}
       {...(ouiaContext.isOuia && {
-        'data-ouia-component-type': 'Alert',
+        'data-ouia-component-type': 'PF4/Alert',
         'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
       })}
       {...(isLiveRegion && {
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownWithContext.tsx
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownWithContext.tsx
@@ -24,7 +24,7 @@ class DropdownWithContext extends React.Component<DropdownProps & InjectedOuiaPr
     direction: DropdownDirection.down,
     onSelect: (): void => undefined,
     autoFocus: true,
-    ouiaComponentType: 'Dropdown'
+    ouiaComponentType: 'PF4/Dropdown'
   };
 
   constructor(props: DropdownProps & InjectedOuiaProps) {
--- a/packages/patternfly-4/react-core/src/components/Dropdown/DropdownWithContext.tsx
+++ b/packages/patternfly-4/react-core/src/components/Dropdown/DropdownWithContext.tsx
@@ -24,7 +24,7 @@ class DropdownWithContext extends React.Component<DropdownProps & InjectedOuiaPr
     direction: DropdownDirection.down,
     onSelect: (): void => undefined,
     autoFocus: true,
-    ouiaComponentType: 'PF4/Dropdown'
+    ouiaComponentType: 'Dropdown'
   };
 
   constructor(props: DropdownProps & InjectedOuiaProps) {
@@ -94,6 +94,7 @@ class DropdownWithContext extends React.Component<DropdownProps & InjectedOuiaPr
               ref={this.baseComponentRef}
               {...ouiaContext.isOuia && {
                 'data-ouia-component-type': ouiaComponentType,
+                'data-ouia-component-ns': 'PF4',
                 'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
               }}
             >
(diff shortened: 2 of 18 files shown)


packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
```diff
@@ -70,7 +70,7 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
       className={customClassName}
       aria-label={ariaLabel}
       {...(ouiaContext.isOuia && {
-        'data-ouia-component-type': 'Alert',
+        'data-ouia-component-type': 'PF4/Alert',
```
dlabrecq: Probably want an issue for this to discuss the format. For example, we may want to omit the version and use something more like `pf-alert`?

redallen: I agree with @dlabrecq . Is there a formal spec for this?

quarckster: @dlabrecq What if a someday new version of Patternflywill be released let's say PF5 and some application will have to use both of them?

quarckster: @redallen There is a spec https://ouia.readthedocs.io/en/latest/README.html#ouia-component:

```
A root level HTML element with a data-ouia-component-type attribute describing a unique name identifying ALL HTML components that can be controlled with the same code or interactions. These identifiers MAY and SHOULD be namespaced when used within a framework, particularly when the name is generic. The delimiter between namespace and type name should be a single / character.

    e.g. A page that has a special dropdown could choose to name that dropdown as FrameworkA/CustomDropdown. All instances of this FrameworkA/CustomDropdown component MUST be expected to be able to be controlled via the same automation.
```

dlabrecq: We don't have this for our CSS selector names, so thought to ask. I'm fine with whatever the team decides here. Just thought the team should discuss...

quarckster:
Updated the PR after discussion with @karelhala and @psav. Namespace should be defined in a separate attribute and be always enabled.

packages/patternfly-4/react-core/src/components/Dropdown/DropdownWithContext.tsx
```diff
@@ -24,7 +24,7 @@ class DropdownWithContext extends React.Component<DropdownProps & InjectedOuiaPr
     direction: DropdownDirection.down,
     onSelect: (): void => undefined,
     autoFocus: true,
-    ouiaComponentType: 'Dropdown'
+    ouiaComponentType: 'PF4/Dropdown'
```
karelhala: I don't think this is no longer a thing, is it?

quarckster: Right, I missed that.

quarckster:
@karelhala @redallen, please review

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
```diff
@@ -71,6 +71,7 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
       aria-label={ariaLabel}
       {...(ouiaContext.isOuia && {
         'data-ouia-component-type': 'Alert',
+        'data-ouia-component-ns': 'PF4',
```
dlabrecq: Could this be defined in one place, instead of within each component?

quarckster: @dlabrecq it's a good point but my knowledge of typescript is limited, I believe it can be done somewhere here https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/withOuia/withOuia.tsx

dlabrecq: Thinking we just need to import a common file. Perhaps `ouia.ts` or `withOuia.tsx` would be a good location to store the namespace?

.../packages/patternfly-4/react-core/src/components/withOuia/ouia.ts

redallen:
I think the need to add a namespace has revealed our OUIA implementation has too much boilerplate. We should not be hardcoding the string 'PF4' for each OUIA component. We'll discuss the best way to implement this, which might be doing away with the `withOUIA` HOC.

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
seanforyou23: Any chance we can use a qualified URI as the namespace, instead of a basic string? Something like `https://www.patternfly.org/v4` as the namespace is more semantic and readable.

quarckster: @psav @karelhala ^

redallen:
Hey @karelhala , @quarckster , and @psav , our infrastructure team discussed our OUIA implementation today and we would like to ignore `localstorage` and always include OUIA attributes using a helper function like:
```js
export function getOUIAProps(componentName, id) {
  return {
        'data-ouia-component-ns': 'PF4',
        'data-ouia-component-type': componentName,
        'data-ouia-id': id
  };
}
```
and then in components:
```js
import { getOUIAProps } from '../../helpers/getOUIAProps';
export class Alert extends React.Component {
  render() {
    return <div {...getOUIAProps(Alert.displayName, props.ouiaID)} />
  }
}
```

However, we would like to time this with our breaking change release since it changes the way components are rendered in snapshots. Is this a pressing issue, or can we wait until our breaking change release? Who do we expect to modify the current outstanding PRs after this change?

cc @dgutride @evwilkin

quarckster:
@redallen Let's wait for the breaking change release. After that this PR should be closed and new issue and PR should be created. I will take over it.

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
seanforyou23: Here's a good article in case it is helpful

https://www.w3.org/2001/tag/doc/URNsAndRegistries-50-2006-08-17.html#args_http

redallen:
Closing in favor of #3927.

PR #3623, opened 2020-01-30
Changed ouia variable name in localStorage in order to comply with spec

### LINKED ISSUE #3625: OUIA localStorage attribute name doesn't comply OUIA spec
According to OUIA spec:
```
If OUIA attributes are not enabled by default there SHOULD be an HTML local storage variable ouia:enabled, to enable the usage of these attributes. Such action MAY incur a page restart.
```
Currently PF4 checks `window.localStorage.ouia.enabled` but it should be `window.localStorage.getItem("ouia:enabled")`

quarckster:
According to OUIA spec:

> If OUIA attributes are not enabled by default there SHOULD be an HTML local storage variable ouia:enabled, to enable the usage of these attributes. Such action MAY incur a page restart.

This PR makes localStorage ouia related item to comply with the spec.

Fixes  #3625

### DIFF
--- a/packages/patternfly-4/react-core/src/components/withOuia/ouia.ts
+++ b/packages/patternfly-4/react-core/src/components/withOuia/ouia.ts
@@ -1,6 +1,6 @@
 export const isOUIAEnvironment = (): boolean => {
   try {
-    return (typeof window !== 'undefined' && window.localStorage && window.localStorage.ouia && window.localStorage.ouia.toLowerCase() === 'true') ||
+    return (typeof window !== 'undefined' && window.localStorage && window.localStorage.getItem("ouia:enabled") && window.localStorage["ouia:enabled"].toLowerCase() === 'true') ||
     false;
   } catch (exception) {
     return false;


packages/patternfly-4/react-core/src/components/withOuia/ouia.ts
```diff
@@ -1,6 +1,6 @@
 export const isOUIAEnvironment = (): boolean => {
   try {
-    return (typeof window !== 'undefined' && window.localStorage && window.localStorage.ouia && window.localStorage.ouia.toLowerCase() === 'true') ||
+    return (typeof window !== 'undefined' && window.localStorage && window.localStorage.getItem("ouia:enabled") && window.localStorage["ouia:enabled"].toLowerCase() === 'true') ||
```
dlabrecq: Looks ok to me, but believe we need an issue to go along with the PR?

quarckster: Ah sorry, let me create it.

quarckster: @dlabaj #3625

quarckster: @dlabrecq #3625

Issue #3592, opened 2020-01-27
Enforce TS linting on new PRs


evwilkin:
Fix any existing TS linting errors & enforce new PRs to pass TS linting tests

evwilkin:
Remaining post-code freeze:
- React styles
- React tokens
- React virtualized extension

evwilkin:
Closed by #3632, #3692, #3693, #3713, #3714, #3715, #3717, #3718, #3719, #3721, #3723, #3725, and #3726 .

PR #3632, opened 2020-01-31
Fix/tslint errors 3605

evwilkin:
**What**: Towards #3592 - fixes eslint errors in react-core


package.json
```diff
@@ -94,9 +98,9 @@
     "clean:build": "shx rm -rf .cache",
     "commit": "git-cz",
     "generate": "yarn plop",
-    "lint:md": "yarn eslint --ext md ./packages/patternfly-4 --config ./.eslintrc_md.json",
+    "lint:md": "yarn eslintpackages/patternfly-4 --ext md --config .eslintrc-md.json",
```
jschuler: space missing after `eslint`?

redallen: Yeah, let's make a followup PR to fix that and turn it on in CI.

PR #3642, opened 2020-02-03
fix(card): add wrapper to image to resize it properly

### LINKED ISSUE #3539: Card bug fix
In Core an issue was filed for a bug with the Card component. When an image was added to `__head` it wouldn't resize properly. To fix this bug we added a new wrapper element called `pf-c-card__head-main` to wrap the image. We should update React accordingly.

Core issue: https://github.com/patternfly/patternfly-next/pull/2578/files

boaz0:
**What**: Closes #3539

### DIFF
--- a/packages/react-core/src/components/Card/CardHeadMain.test.tsx
+++ b/packages/react-core/src/components/Card/CardHeadMain.test.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import { CardHeadMain } from './CardHeadMain';
+import { shallow } from 'enzyme';
+
+test('renders with PatternFly Core styles', () => {
+  const view = shallow(<CardHeadMain>text</CardHeadMain>);
+  expect(view).toMatchSnapshot();
+});
+
+test('className is added to the root element', () => {
+  const view = shallow(<CardHeadMain className="extra-class">text</CardHeadMain>);
+  expect(view.prop('className')).toMatchSnapshot();
+});
+
+test('extra props are spread to the root element', () => {
+  const testId = 'card-head-main';
+  const view = shallow(<CardHeadMain data-testid={testId} />);
+  expect(view.prop('data-testid')).toBe(testId);
+});
--- a/packages/react-core/src/components/Card/CardHeadMain.tsx
+++ b/packages/react-core/src/components/Card/CardHeadMain.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+
+export interface CardHeadMainProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the Card Head Main */
+  children?: React.ReactNode;
+  /** Additional classes added to the Card Head Main */
+  className?: string;
+}
+
+export const CardHeadMain: React.FunctionComponent<CardHeadMainProps> = ({
+  children = null,
+  className = '',
+  ...props
+}: CardHeadMainProps) => (
+  <div className={css('pf-c-card__head-main', className)} {...props}>
+    {children}
+  </div>
+);
--- a/packages/react-core/src/components/Card/__snapshots__/CardHeadMain.test.tsx.snap
+++ b/packages/react-core/src/components/Card/__snapshots__/CardHeadMain.test.tsx.snap
@@ -1,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`className is added to the root element 1`] = `"pf-c-card__head-main extra-class"`;
+
+exports[`renders with PatternFly Core styles 1`] = `
+<div
+  className="pf-c-card__head-main"
+>
+  text
+</div>
+`;
--- a/packages/react-core/src/components/Card/examples/Card.md
+++ b/packages/react-core/src/components/Card/examples/Card.md
@@ -3,10 +3,10 @@ title: 'Card'
 section: components
 cssPrefix: 'pf-c-card'
 typescript: true
-propComponents: ['Card', 'CardHeader', 'CardBody', 'CardFooter']
+propComponents: ['Card', 'CardHeadMain', 'CardHeader', 'CardBody', 'CardFooter']
 ---
 
-import { Card, CardActions, CardHead, CardHeader, CardBody, CardFooter, Checkbox, DropdownActions } from '@patternfly/react-core';
+import { Brand, Card, CardActions, CardHead, CardHeadMain, CardHeader, CardBody, CardFooter, Checkbox, DropdownActions } from '@patternfly/react-core';
 import pfLogo from './pfLogo.svg'; 
 
 ## Examples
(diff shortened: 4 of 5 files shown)


packages/patternfly-4/react-core/src/components/Card/examples/Card.md
```diff
@@ -75,7 +75,9 @@ class KebabDropdown extends React.Component {
     return (
       <Card>
         <CardHead>
-          <img src={pfLogo} alt="PatternFly logo" style={{ height: '50px' }}/>
+          <CardHeadMain>
+            <img src={pfLogo} alt="PatternFly logo" style={{ height: '50px' }}/>
+          </CardHeadMain>
```
dlabrecq: I know `img` already exist here, but shouldn't we use the Brand component for the PatternFly logo? That way, we don't need to specify the image path.

<Brand src={pfLogo} alt="Patternfly Logo" />

packages/patternfly-4/react-core/src/components/Card/examples/Card.md
```diff
@@ -6,7 +6,7 @@ typescript: true
 propComponents: ['Card', 'CardHeader', 'CardBody', 'CardFooter']
```
tlabaj: You need to add CardHeadMain here.

packages/patternfly-4/react-core/src/components/Card/CardHeadMain.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Card/card';
+
+export interface CardHeadMainProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the Card Head Main */
+  children?: React.ReactNode;
```
tlabaj: To me it is not intuitive when this wrapper should be used.  Maybe we should add the children should be text, image or logo to make it more clear. Even if we add that a consumer could easily not include the wrapper and the styling would be off @christiemolloy @mcoker.

mcoker: Agree. I wonder if we should always have the `<CardHeadMain>` contain the contents of whatever you put in the `<CardHead>` that isn't actions? That way it's consistent. wdyt @christiemolloy?

```
<CardHead>
  <CardActions></CardActions> // optional
  <CardHeadMain>
    // images, text, <CardHeader>, etc
  </CardHeadMain>
</CardHead>
```

christiemolloy: Yeah it would be good if we could generate that wrapper automatically so users dont forget it. Good point!

packages/patternfly-4/react-core/src/components/Card/CardHead.tsx
```diff
@@ -13,8 +19,37 @@ export const CardHead: React.FunctionComponent<CardHeadProps> = ({
...
+}: CardHeadProps) => {
+
+  const actionChildren = React.Children.toArray(children).filter(child => isReactElement(child) && child.type === CardActions)
+  const mainChildren = React.Children.toArray(children).filter(child => !isReactElement(child) || child.type !== CardActions)
+  return (
+    <div className={css(styles.cardHead, className)} {...props}>
+      <CardHeadMain>
+        {mainChildren}
+      </CardHeadMain>
+      {actionChildren}
```
dlabrecq: There is an implied order here. What if I've defined `<CardActions>` first? 

I believe the output would be quite unexpected if users provided a layout like this:
```

<CardHead>
  <img .../>
  <CardActions>
  <img .../>
</CardHead>
```

@mcoker I don't know that we want to do this with React composition? I believe users should be explicit when wrapping components with `<CardAction/>` and `<CardHeadMain/>`

boaz0: That's what I thought but I wasn't sure. However, every items that's not `CardActions` will be wrapped individually. If that's OK, then I have something to work on. Otherwise, we need to think more.

dlabrecq: It seems problematic to me

dlabrecq: It seems problematic to me (e.g., if someone wants a custom layout)

boaz0: @tlabaj @christiemolloy @mcoker - your thoughts.
I am trying automatically to wrap items under `CardHead` into `CardHeadMain` as you thought but I need more information.

Thanks.

tlabaj: I agree with Dan here about the implied order.  This is too restrictive and can break current implementations.   I think in this case, that it is better to update the docs and guide the consumer to use the wrapper rather than restricting the order of the children.

packages/patternfly-4/react-core/src/components/Card/CardHeadMain.tsx
```diff
@@ -0,0 +1,19 @@
...
+  /** Additional classes added to the Card Head Main */
+  className?: string;
+}
+
+export const CardHeadMain: React.FunctionComponent<CardHeadMainProps> = ({
+  children = null,
+  className = '',
+  ...props
+}: CardHeadMainProps) => (
+  <div className={css('pf-c-card__head-main', className)} {...props}>
```
tlabaj: Why are you suing the class name directly here?  Is it not in the stylesheet?

boaz0: No, it doesn't exist there. Read more [here](https://github.com/patternfly/patternfly-react/issues/3539#issuecomment-578488823)

packages/patternfly-4/react-core/src/components/Card/CardHead.tsx
```diff
@@ -15,6 +25,11 @@ export const CardHead: React.FunctionComponent<CardHeadProps> = ({
   ...props
 }: CardHeadProps) => (
   <div className={css(styles.cardHead, className)} {...props}>
-    {children}
+    {React.Children.map(children, child => {
+      if (isReactElement(child) && (child.type === CardActions || child.type === CardHeader)) {
```
tlabaj: This still does not look right.  `CardHeader` should be wrapped by `CardHeadMain`.  You are also still forcing the actions to come before 'CardHeadMain` and I don't think we want to add that restriction. 
Since 'CardHeadMain` is not required,  I still think your original implementation with leaving it up to the consumer to add the wrappper is the best solution here.  We just need to make sure it is documented appropriately.

tlabaj:
@boaz this one needs a rebase.  The package structure changed in the repo.  Also see the last comment on the PR.  Thanks!

tlabaj:
@Boaz.  We can take this one in once snapshots are updated.

boaz0:
@tlabaj no problems.
I will update the snapshot now...
Thanks.

PR #3645, opened 2020-02-03
feat(SimpleList): add simple list

### LINKED ISSUE #3610: Add experimental "Simple list" component
https://github.com/patternfly/patternfly-next/pull/2573/files

https://pf4.patternfly.org/documentation/core/experimental/simplelist

kmcfaul:
**What**: Closes #3610

@mcoker It appears that `pf-c-simple-list__list` and `pf-c-simple-list__list-item` do not exist/are not in use despite appearing on the HTML structure in core. Let me know if I should add them manually, but currently I have left them off.


mcoker:
* I wonder if we should add an `id` to the `__title`, then `aria-labelledby` to the `__section` to match the nav component? Also curious if we should make that update to the dropdown, app launcher, options menu, etc?
* We probably want to add `type="button"` to the buttons so that they aren't `submit` buttons.
* This is a bug in core, but we need to add `text-decoration: none` to the `<a>` items so that they don't get an underline on hover.
* Add an `href` attribute to the `<a>` `__item-link` elements so that they're in the tabindex. Or maybe add `tabindex="0"` to elements that aren't a button, or just all for consistency?

cc @jessiehuff 

> It appears that pf-c-simple-list__list and pf-c-simple-list__list-item do not exist/are not in use despite appearing on the HTML structure in core. Let me know if I should add them manually, but currently I have left them off.

I'm curious what @tlabaj thinks about this. I'm not sure that we have a consistent way of doing this, and it often comes up in the PR when we give an element a class that doesn't have a corresponding style.

The main reason we do it is that it helps build the BEM structure (creates a namespace for elements), and allows us to style those elements at some later point without requiring users update their markup to add a new class. That mostly applies to core users, since we can always add a missing class to a react component after the fact.

kmcfaul:
@mcoker Added your bullets (minus the core one). I went with adding the `tabindex="0"` only to the anchor elements.

@jessiehuff Let me know if I'm missing any other aria labels or tags that should be explicitly defined. I think in most cases using the `...props` to pass these in should work.

packages/patternfly-4/react-core/src/experimental/components/SimpleList/SimpleList.tsx
```diff
@@ -0,0 +1,75 @@
...
+import styles from '@patternfly/react-styles/css/components/SimpleList/simple-list';
+import { SimpleListGroup } from './SimpleListGroup';
+
+export interface SimpleListProps {
+  /** Content rendered inside the SimpleList */
+  children?: React.ReactNode;
+  /** Additional classes added to the SimpleList <ul> */
+  className?: string;
+  /** Callback when an item is selected */
+  onSelect?: (ref: React.RefObject<HTMLButtonElement> | React.RefObject<HTMLAnchorElement>) => void;
```
dlabrecq: Instead of providing the button associated with clicking on each menu option, I'd prefer to provide my own ID. I can get the label from the button, but it's difficult to work with considering the label may be localized.

Screen Shot 2020-02-05 at 10 59 26 AM
...
Screen Shot 2020-02-05 at 10 59 46 AM

@dlabaj I believe we may be using `SelectOption` to help with this?

mcoker:
@mcarrano that's a bug from core. Just put up a PR to fix https://github.com/patternfly/patternfly-next/pull/2679****

packages/patternfly-4/react-core/src/experimental/components/SimpleList/SimpleList.test.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
```
jschuler: can you add this under the `__tests__` directory?

mcoker:
@kmcfaul the underline is fixed https://github.com/patternfly/patternfly-next/pull/2679 (v2.59.1)

kmcfaul:
@dlabrecq 
I added another property to the callback function that will pass back the item's props. This will work for `id` and the like without any special declaration. 

However, I could instead add another property to SimpleListItem that would be used to pass back data, instead of props. It would be clearer what gets passed back using this approach but would require extra user input to define the structure.

Any thoughts about which approach would be better?

packages/patternfly-4/react-core/src/components/SimpleList/SimpleList.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/SimpleList/simple-list';
+import { SimpleListGroup } from './SimpleListGroup';
+import { SimpleListItemProps } from './SimpleListItem';
+
+export interface SimpleListProps {
```
jschuler: looks like it should extend `React.HTMLProps<HTMLDivElement>`

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListGroup.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/SimpleList/simple-list';
+
+export interface SimpleListGroupProps {
```
jschuler: should extend HTMLProps for section

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+            <li className={css(listItemClassName)}>
+              <Component
+                className={css(styles.simpleListItemLink, isCurrentItem && styles.modifiers.current, className)}
+                onClick={(evt: React.MouseEvent) => {
+                  onClick(evt);
+                  updateCurrentRef(this.ref, this.props);
+                }}
+                ref={this.ref}
+                {...componentProps}
+                {...props}
```
jschuler: can `...props` be moved to the parent element `li`?

packages/patternfly-4/react-core/src/components/SimpleList/SimpleList.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    return (
+      <SimpleListContext.Provider
+        value={{
+          currentRef: this.state.currentRef,
+          updateCurrentRef: this.handleCurrentUpdate
+        }}
+      >
+        <div className={css(styles.simpleList, className)} id={id} {...props}>
```
jschuler: why is the 'id' needed?

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListGroup.tsx
```diff
@@ -0,0 +1,32 @@
...
+  /** Content rendered inside the SimpleList group */
+  children?: React.ReactNode;
+  /** Additional classes added to the SimpleList <ul> */
+  className?: string;
+  /** Additional classes added to the SimpleList group title */
+  titleClassName?: string;
+  /** Title of the SimpleList group */
+  title?: string | React.ReactNode;
```
jschuler: I believe ReactNode can be string already so not really needed to specify string explicitly

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    return (
+      <SimpleListContext.Consumer>
+        {({ currentRef, updateCurrentRef }) => {
+          const isButton = Component === 'button';
+          const isCurrentItem = this.ref && currentRef ? currentRef.current === this.ref.current : isCurrent;
+
+          const componentProps = isButton
+            ? {
+                type
```
jschuler: just making sure this is intended, but type can be `submit` or `reset` as well if that's what's passed in as prop

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+              }
+            : {
+                tabIndex: 0,
+                href
+              };
+
+          return (
+            <li className={css(listItemClassName)}>
+              <Component
+                className={css(styles.simpleListItemLink, isCurrentItem && styles.modifiers.current, className)}
```
jschuler: `className` should be on the root element

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListItem.tsx
kmcfaul: we want to make sure the default type is "button", this line lets the user change that if they wish. otherwise adding a `type` property manually would be caught by the component and not actually set.

packages/patternfly-4/react-core/src/components/SimpleList/SimpleList.tsx
kmcfaul: I think it should be able to be removed, `id` can be set via the `...props`

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListItem.tsx
kmcfaul: I'm not sure either way if we want to allow this

@mcarrano

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListGroup.tsx
kmcfaul: I could only find `React.HTMLProps<HTMLTableSectionElement>`, would this be correct? It seems to work with the console logging warnings

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListItem.tsx
kmcfaul: Moved these (className, props) to root element and added properties to customize the inner component.

packages/patternfly-4/react-core/src/components/SimpleList/SimpleListGroup.tsx
jschuler: hm, how about passing `<any>` in this case since we just want the default HTML attributes

mattnolting:
I would also really like to see the component selectors manually added to the `<li>`s, but that's a larger discussion and not a blocker 😄

mattnolting:
Really great work @kmcfaul ! Only one more thing for me: 

If `.pf-c-simple-list` doesn't contain child `.pf-c-simple-list`s, you don't need `role="list"`. 

Screen Shot 2020-02-13 at 1 18 06 PM

Issue #3643, opened 2020-02-03
react-tokens build optimizations


Hyperkid123:
cc @evwilkin @redallen @karelhala 

React tokens are currently built into a single file. This causes issues in non-esm environments. Since there were great optimizations for pf4 components packages, which greatly improved bundle sizes of consumer apps (so far it was around 50% - 80%, depends on use-case), I think we can do the same with tokens

Here is an example of how this can affect the build of app using pf4:
![Screenshot from 2020-02-03 15-00-43](https://user-images.githubusercontent.com/22619452/73662069-77ae2c80-469b-11ea-9643-93fdca99679b.png)

This is one of the cloud services app. As you can see, react-tokens has almost the same **parsed** size as react-core.

basically the solution to this would be the same as with the rest of pf4 packages. Build each token into one file (while keeping the index with all tokens), and use absolute import path on pf4 react packages.

The code change would be rather simple. Here are some examples.

Here is something that can be added to the template files
```diff
diff --git a/packages/patternfly-4/react-tokens/src/templates/cjs.js b/packages/patternfly-4/react-tokens/src/templates/cjs.js
index 9cc1b1313..683e4f2ff 100644
--- a/packages/patternfly-4/react-tokens/src/templates/cjs.js
+++ b/packages/patternfly-4/react-tokens/src/templates/cjs.js
@@ -3,5 +3,7 @@ const { join } = require('path');
 module.exports = {
   getOutputPath: ({ outDir }) => join(outDir, 'js/index.js'),
   getContent: ({ tokens }) =>
-    Object.keys(tokens).reduce((acc, key) => `${acc}module.exports.${key} = ${JSON.stringify(tokens[key])}\n`, '')
+    Object.keys(tokens).reduce((acc, key) => `${acc}module.exports.${key} = ${JSON.stringify(tokens[key])}\n`, ''),
+  getSingleOutputPath: ({ outDir, tokenName }) => join(outDir, `js/${tokenName}.js`),
+  getSingleContent: ({ tokenValue }) => `module.exports = ${JSON.stringify(tokenValue)}\n`
 };

```

And in the generating function:

```diff
diff --git a/packages/patternfly-4/react-tokens/src/generateTokens.js b/packages/patternfly-4/react-tokens/src/generateTokens.js
index 83d4d3942..1177e8118 100644
--- a/packages/patternfly-4/react-tokens/src/generateTokens.js
+++ b/packages/patternfly-4/react-tokens/src/generateTokens.js
[...]

PR #3647, opened 2020-02-04
Generate single file for each react token

Hyperkid123:
re #3643

Adds additional generators for react-tokens. Each token has its own file. This should improve build sizes for libraries/apps using react-tokens.

### DIFF
--- a/packages/patternfly-4/react-tokens/src/templates/d.ts.js
+++ b/packages/patternfly-4/react-tokens/src/templates/d.ts.js
@@ -9,5 +9,9 @@ module.exports = {
         .map(tokenKey => `${JSON.stringify(tokenKey)}: ${JSON.stringify(token[tokenKey])};`)
         .join(' ');
       return `${acc}export const ${key}: { ${tokenTypeValue} }\n`;
-    }, '')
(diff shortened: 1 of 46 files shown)


packages/patternfly-4/react-tokens/src/templates/d.ts.js
```diff
@@ -9,5 +9,9 @@ module.exports = {
         .map(tokenKey => `${JSON.stringify(tokenKey)}: ${JSON.stringify(token[tokenKey])};`)
         .join(' ');
       return `${acc}export const ${key}: { ${tokenTypeValue} }\n`;
-    }, '')
+    }, ''),
+  getSingleOutputPath: ({ outDir, tokenName }) => join(outDir, `js/${tokenName}.d.ts`),
+  getSingleContent: ({ tokenName, tokenValue }) =>
+    `const ${tokenName}: {${Object.entries(tokenValue).map(([key, value]) =>
```
jschuler: can you add `declare` before the const? This is so TS won't complain and throw this message:
`TS1046: 'declare' modifier required for top level element`

Issue #3624, opened 2020-01-30
Change experimental components to beta components


dgutride:
- [x] Remove the experimental category from the documentation site
- [x] The experimental folder on disk should be renamed to “beta” and a placeholder for the old experimental/index.ts barrel file should remain but be updated to point to the moved beta components.
- [x] All beta components should be exported from the root index.ts file now
- [x] Beta components should be tagged as “beta” and a banner or some descriptive information should be included in the documentation indicating these could potentially have minor adjustments as we receive feedback on them.
- [x] Allow for any new properties added to existing components to be potentially marked as “beta” using metadata in the property description.  These properties should be clearly called out in the documentation.
- [x] When a beta component is mounted, a console warning/info (?)  statement should be logged that indicates a beta component is being used that could potentially have a changing api.
- [x] When a beta property is used on an existing component, a console warning should be logged that indicates a beta property is used that could potentially be susceptible to changes.
  - Note - we probably don’t need to warn on all beta properties that are on beta components, a single higher level warning is sufficient.

evwilkin:
Closed by #3663, https://github.com/patternfly/patternfly-org/pull/1682, https://github.com/patternfly/patternfly-next/pull/2680, #3755 , #3764 , #3856 , #3813 , and #3775 and #3764

PR #3663, opened 2020-02-05
chore(nomenclature): phrase non production components as beta

seanforyou23:
**What**: Steps toward closing https://github.com/patternfly/patternfly-react/issues/3624

This PR renames "experimental" things to "beta" so that the intent of the in-progress status is more clearly signaled to users.


packages/patternfly-4/react-core/src/beta/components/Drawer/examples/Drawer.md
```diff
@@ -49,7 +49,7 @@ class SimpleDrawer extends React.Component {
 
 ```js title=Basic-inline
 import React, { ReactFragment } from 'react';
-import { Drawer, DrawerPanelContent, DrawerContent } from '@patternfly/react-core/dist/esm/experimental';
+import { Drawer, DrawerPanelContent, DrawerContent } from '@patternfly/react-core/dist/esm/beta';
```
dgutride: Can these be changed to @patternfly/react-core directly now?

packages/patternfly-4/react-docs/gatsby-browser.js
```diff
@@ -16,9 +16,9 @@ import '@patternfly/patternfly/patternfly.css';
 // Utilities
 import '@patternfly/patternfly/patternfly-addons.css';
 
-// Experimental components
+// Beta components
```
dgutride: Are these just temporarily necessary until a core change is made to include css in the main package?

packages/patternfly-4/react-docs/gatsby-config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
             { section: 'charts' },
             { section: 'components' },
             { section: 'demos' },
-            { section: 'experimental' },
+            { section: 'beta' },
```
dgutride: This should go away - all beta components will be listed with the others - just noted differently in the documentation.  If you'd rather do that in a separate PR (or have somebody else make that change, it's fine with me).

packages/patternfly-4/react-styled-system/README.md
```diff
@@ -1,6 +1,6 @@
 # @patternfly/react-styled-system
 
-## Warning: the styled system components are experimental. Use at your own risk!
+## Warning: the styled system components are beta. Use at your own risk!
```
dgutride: These actually truly are experimental (and I think we should deprecate and remove) - can we leave the wording as-is?

packages/react-codemods/README.md
```diff
@@ -1,6 +1,6 @@
 # @patternfly/react-codemods
 
-_PatternFly React codemods are currently experimental only._
+_PatternFly React codemods are currently beta only._
```
dgutride: These are also truly experimental - maybe we should leave it this way for now?

packages/patternfly-4/react-core/src/beta/components/Drawer/examples/Drawer.md
tlabaj: I agree for new components. I actually think we may want to still export from experimental (and from react-core) for existing components and remove that with the breaking change release.

dgutride: I agree - for this, I'm just talking about the import statements we are using - I'd rather that we just treat these components like other components to indicate the commitment to them.    That does remind me that I need to look to see if the experimental/index.js still exists...

seanforyou23: Ok think I see what you mean, pushed some changes - let me know if I missed anything!

seanforyou23:
I've pushed some updates. We now display previously "experimental" components alongside all production components and visually mark them as "Beta" with an inline alert. This alert message needs to be reviewed.

Screen Shot 2020-02-07 at 12 35 17 PM

In addition to now having the notion of beta components, we also have support for beta properties/features on production components. No beta properties exist yet, but below is a screenshot of how they are currently wired up to display. Notice the property name is `exampleBetaProperty` and the visual treatment is to simply prefix the property name with `Beta: `, just like we do for deprecated properties. We could style it some special way, or instead just generate a new "beta" column, but I figured a first pass would be good to follow the existing convention. #3581, which is in-progress, will put this new feature to use for the first time, I'm happy to address any changes now or later to this part of the experience.

Screen Shot 2020-02-07 at 12 26 20 PM

packages/patternfly-4/react-docs/gatsby-browser.js
seanforyou23: Just waiting on these last few [changes from core](https://github.com/patternfly/patternfly-next/commit/f24376e58d941420c477e0b0aa29f985af5111ae) to come in so we can delete the rest, until they come through we still need to manually import a couple of these

seanforyou23:
Thanks @mcarrano - checking on the styles, think we just need to pull in a recent change from core. Good catch.

As far as using a badge to flag beta properties, it shouldn't be a problem. I will take a look!

seanforyou23:
Thx for the feedback @mcarrano - I pulled in the latest changes from core and that seems to have fixed the styles on the Data toolbar/Drawer beta component pages. Also, made some changes in an open PR in [org](https://github.com/patternfly/patternfly-org/pull/1682) that should better reflect the design @rachael-phillips provided for beta styles (sry, didn't see that before). Here's an example of what those changes across org/core produce:

Marking examples as beta:
Screen Shot 2020-02-10 at 1 14 29 PM

And then marking props as beta in the props table at the bottom: 
Screen Shot 2020-02-10 at 1 28 10 PM

Let me know if you see any issues!

packages/patternfly-4/react-core/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './beta';
```
dgutride: I'm still a bit confused about why we have a beta location - the goal was to move the components to their final place (with the other components) and just use the documentation to indicate they are beta so we wouldn't need to keep moving them.

seanforyou23: I see, sorry I missed this! Getting that updated now

PR #3693, opened 2020-02-06
chore(lint): eslint react-table

evwilkin:
**What**: Towards #3592 - linting for react-table


packages/patternfly-4/react-table/src/components/Table/BodyCell.tsx
```diff
@@ -22,8 +22,10 @@ export const BodyCell: React.FunctionComponent<BodyCellProps> = ({
   isVisible,
   parentId,
   textCenter = false,
+  /* eslint-disable @typescript-eslint/no-unused-vars */
   isOpen,
```
dlabrecq: Is there a reason to keep `isOpen` if it's not used?

packages/patternfly-4/react-table/src/components/Table/SelectColumn.tsx
```diff
@@ -9,14 +9,13 @@ export interface SelectColumnProps {
 
 export const SelectColumn: React.FunctionComponent<SelectColumnProps> = ({
   children = null as React.ReactNode,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   className = '',
```
dlabrecq: Is there a reason we need to keep `className` if it's unused?

packages/patternfly-4/react-table/src/components/Table/Body.tsx
```diff
@@ -125,11 +127,14 @@ class ContextBody extends React.Component<TableBodyProps, {}> {
 }
 
 export const TableBody = ({
-  onRow = (...args: any) => Object,
   className = '' as string,
   children = null as React.ReactNode,
   rowKey = 'id' as string,
-  onRowClick = (event: React.MouseEvent, row: IRow, rowProps: IExtraRowData, computedData: IComputedData) => undefined as OnRowClick,
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  onRow = (...args: any) => Object,
```
dlabrecq: `onRow` and `onRowClick` appear to be used?

evwilkin: Both onRow and onRowClick were throwing TS errors - it looks like the unused variable errors were for the arguments to both functions (...args for onRow, and event/row/rowProps/computedData for onRowClick)

packages/patternfly-4/react-table/src/components/Table/BodyCell.tsx
evwilkin: @dlabrecq thanks for calling this out - I've seen this pattern where we pull certain variables out of props (even if we don't use them) so they're not passed in with the remaining `...props` - I'm not as familiar with tables so didn't know if that was the case here, so simply avoided the existing error by disabling it for these variables.  Do you think the unused variables should be removed instead?  Same issue for your next comment below on SelectColumn.

packages/patternfly-4/react-table/src/components/Table/Body.tsx
redallen: Can we track down how `onRow` is used and give it a proper type?

Also, for the unused variables, just put underscores before the variable names. Like `_event: React.MouseEvent`. Then you can remove the eslint-disable.

packages/patternfly-4/react-table/src/components/Table/Body.tsx
```diff
@@ -139,6 +144,7 @@ export const TableBody = ({
         rows={rows as IRow[]}
         onRow={onRow}
         className={className}
+        // eslint-disable-next-line react/no-children-prop
```
redallen: Can we move children to actually be children?

packages/patternfly-4/react-table/src/components/Table/SelectColumn.tsx
redallen: I guess it's because we don't want to allow a `className` to be passed to the `<input>`.

packages/patternfly-4/react-table/src/components/Table/base/columns-are-equal.ts
```diff
@@ -7,6 +7,7 @@
 import { isFunction, isEqualWith } from 'lodash';
 import { ColumnsType } from './types';
 
+// eslint-disable-next-line jsdoc/require-jsdoc
```
redallen: Why don't we add JSDoc?

packages/patternfly-4/react-table/src/components/Table/base/evaluate-formatters.ts
```diff
@@ -6,6 +6,7 @@
  */
 import { formattersType, formatterValueType, ExtraParamsType } from './types';
 
+// eslint-disable-next-line jsdoc/require-jsdoc
```
redallen: Why don't we add JSDoc?

packages/patternfly-4/react-table/src/components/Table/base/body.tsx
```diff
@@ -23,6 +23,7 @@ export interface BodyProps {
 
 class BaseBody extends React.Component<BodyProps, {}> {
   static defaultProps = {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     onRow: (...args: any) => Object
```
redallen: A proper type for onRow would be wonderful.

packages/patternfly-4/react-table/src/components/Table/base/evaluate-transforms.ts
```diff
@@ -8,6 +8,7 @@ import { isFunction } from 'lodash';
 import { mergeProps } from './merge-props';
 import { transformsType, ExtraParamsType } from './types';
 
+// eslint-disable-next-line jsdoc/require-jsdoc
```
redallen: Why don't we add JSDoc?

packages/patternfly-4/react-table/src/components/Table/base/resolve-row-key.ts
```diff
@@ -8,6 +8,9 @@
 import { isArray } from 'lodash';
 import { RowType, RowKeyType } from './types';
 
+/**
+ *
+ */
```
redallen: You can add some simple JSDoc instead of leaving it blank.

packages/patternfly-4/react-table/src/components/Table/utils/decorators/headerCol.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import { IExtra, IFormatterValueType, ITransform } from '../../Table';
 
-export const headerCol = (id: string = 'simple-node') => {
-  // tslint:disable-next-line:no-shadowed-variable
+export const headerCol = (id = 'simple-node') => {
+  // eslint-disable-next-line no-shadow
   const headerCol: ITransform = (value: IFormatterValueType, { rowIndex }: IExtra = {}) => {
```
redallen: Renaming this variable is probably wise.

packages/patternfly-4/react-table/src/components/Table/BodyCell.tsx
seanforyou23: I think due to the way we're spreading props, if we don't destructure these properties in this way, they end up getting passed down to the DOM element which throws warnings in the console like so

`Warning: React does not recognize the `isOpen` prop on a DOM element.`

packages/patternfly-4/react-table/src/components/Table/Body.tsx
evwilkin: Done

packages/patternfly-4/react-table/src/components/Table/base/columns-are-equal.ts
evwilkin: Done

packages/patternfly-4/react-table/src/components/Table/base/evaluate-formatters.ts
evwilkin: Done

packages/patternfly-4/react-table/src/components/Table/base/evaluate-transforms.ts
evwilkin: Done

packages/patternfly-4/react-table/src/components/Table/base/resolve-row-key.ts
evwilkin: Done

packages/patternfly-4/react-table/src/components/Table/utils/decorators/headerCol.tsx
evwilkin: Done

packages/patternfly-4/react-table/src/components/Table/SelectColumn.tsx
dlabrecq: If there is a specific reason to pull it out, that's fine. It's just odd to set a default value, so a comment would help.

packages/patternfly-4/react-table/src/components/Table/Table.tsx
```diff
@@ -61,12 +64,14 @@ export enum SortByDirection {
   desc = 'desc'
 }
 
-export interface IHeaderRow extends ColumnType { }
+export type IHeaderRow = ColumnType;
```
dlabrecq: Should we have changed this from `interface` to `type`?

PR #3706, opened 2020-02-10
fix(switch): avoid switch id override by props

### LINKED ISSUE #3662: Changing Switch component id causes it to not be clickable
When you change the id attribute on the switch the for attribute on the label that wraps around the switch input does not get changed. I would expect that when I change the id attribute, both the for attribute on the label and the id attribute on the input should be the same whenever the id is changed. The problem is that on [line #49](https://github.com/patternfly/patternfly-react/blob/6290a542f17baae3e84779dcc18bfe7dbbab40a1/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx#L49) the id is set and never reset if the switch id gets changed, and then on [line #72](https://github.com/patternfly/patternfly-react/blob/6290a542f17baae3e84779dcc18bfe7dbbab40a1/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx#L72) all the props are being deconstructed on the input which overwrites the the old id with the new, which is what causes the mismatch between the label's for attribute and the input's id attribute.

https://codesandbox.io/s/jovial-montalcini-3dfdu

This is a bug and the only workaround is that if you make the id static. For example, I just made the id's for the list of switches their index of the array they were in.

boaz0:
**What**: Closes #3662 

//cc @jeudy100

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
+++ b/packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
@@ -50,7 +50,20 @@ class Switch extends React.Component<SwitchProps & InjectedOuiaProps> {
   }
 
   render() {
-    const { className, label, labelOff, isChecked, isDisabled, onChange, ouiaContext, ouiaId, ...props } = this.props;
+    const {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      id,
+      className,
+      label,
+      labelOff,
+      isChecked,
+      isDisabled,
+      onChange,
+      ouiaContext,
+      ouiaId,
+      ...props
+    } = this.props;
+
     const isAriaLabelledBy = props['aria-label'] === '';
     return (
       <label
--- a/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/SwitchDemo/SwitchDemo.tsx
+++ b/packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/SwitchDemo/SwitchDemo.tsx
@@ -1,26 +1,55 @@
 import React from 'react';
-import { Switch } from '@patternfly/react-core';
+import { Switch, Button } from '@patternfly/react-core';
+import { Table, TableBody, TableHeader } from '@patternfly/react-table';
 
 interface SwitchState {
-  isChecked: boolean;
+  simple: {
+    isChecked: boolean;
+  };
+  table: {
+    data: { id: string; isChecked: boolean }[];
+    all: boolean;
+  };
 }
+
 export class SwitchDemo extends React.Component<{}, SwitchState> {
   state = {
-    isChecked: true
+    simple: {
+      isChecked: true
+    },
+    table: {
+      data: [{ id: 'id1', isChecked: false }, { id: 'id2', isChecked: false }, { id: 'id3', isChecked: false }],
+      all: true
+    }
   };
 
-  handleChange = isChecked => {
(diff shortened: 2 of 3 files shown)


packages/patternfly-4/react-core/src/components/Switch/examples/Switch.md
```diff
@@ -6,9 +6,63 @@ propComponents: ['Switch']
 typescript: true
 ---
 
-import { Switch } from '@patternfly/react-core';
+import { Switch, Table, TableHeader, TableBody, Button } from '@patternfly/react-core';
 
 ## Examples
+```js title=Multi switch table
```
tlabaj: I don't think we should add this example of the switch here.  It seems to be a specific use case.  If you want to verify it works the demo-app and integration test would be a better place for this.

boaz0: good idea

packages/patternfly-4/react-core/src/components/Switch/Switch.tsx
```diff
@@ -62,14 +62,14 @@ class Switch extends React.Component<SwitchProps & InjectedOuiaProps> {
...
         <input
+          {...props}
           id={this.id}
           className={css(styles.switchInput)}
           type="checkbox"
           onChange={event => onChange(event.target.checked, event)}
           checked={isChecked}
           disabled={isDisabled}
           aria-labelledby={isAriaLabelledBy ? `${this.id}-on` : null}
-          {...props}
```
dlabrecq: I don't know that we want all props added to the input? For example, there are props (i.e., className, ouiaId, etc.) that are only applied to the label.

Probably should be using `rest` instead of all `props`? Then, we can destructure the `id` so that it does not override anything else.

This:
`const { className, label, labelOff, isChecked, isDisabled, onChange, ouiaContext, ouiaId, ...props } = this.props;`

becomes:
`const { className, label, labelOff, id, isChecked, isDisabled, onChange, ouiaContext, ouiaId, ...rest } = this.props;`

Finally:       
```
<input
  id={this.id}
  className={css(styles.switchInput)}
  type="checkbox"
  onChange={event => onChange(event.target.checked, event)}
  checked={isChecked}
  disabled={isDisabled}
  aria-labelledby={isAriaLabelledBy ? `${this.id}-on` : null}
  {...rest} // <<< This should be last
/>
```

packages/patternfly-4/react-core/src/components/Switch/examples/Switch.md
redallen: Use dashes between words as well, i.e. `title=Multi-switch-table`

packages/patternfly-4/react-integration/demo-app-ts/src/components/demos/SwitchDemo/SwitchDemo.tsx
```diff
@@ -35,4 +64,44 @@ export class SwitchDemo extends React.Component<{}, SwitchState> {
...
+  }
+
+  render() {
+    return (
+      <React.Fragment>
+        {this.renderSimple()}
+        {this.renderTable()}
+      </React.Fragment>
+    );
+  }
```
dlabrecq: Trying to run the demo app locally, but getting this error.

>Attempted import error: 'EditableTextCell' is not exported from '@patternfly/react-table'.

There are a lot of changes to the demo, considering we're only changing the how the ID is destructured.

PR #3714, opened 2020-02-10
chore(react-catalog-view-extension): fix linting errors

evwilkin:
**What**: Towards #3592 - resolves existing linting issues for react-catalog-view-extension


packages/patternfly-4/react-catalog-view-extension/scripts/copyTS.js
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
redallen: These aren't var requires? Can this rule be configured for JS files to not complain?

evwilkin: @redallen thanks for the callout - these do meet the definition of incorrect code based on this rule: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-var-requires.md

It doesn't look like this rule can be configured - this is an error you'll see a lot of across upcoming PRs

packages/patternfly-4/react-catalog-view-extension/node-sass-patternfly-importer.js
```diff
@@ -1,5 +1,9 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
ddonahue007: Just curious on this. instead of disable the check why not switch it to proper ES6 typescript format?
import path = require('path');

This is in multiple places

evwilkin: Good thought - I tried that initially but then it causes prettier to throw an error, and at that point it seemed like multiple fixes to get around a linter issue:
Screen Shot 2020-02-11 at 5 00 36 PM

PR #3715, opened 2020-02-10
chore(react-docs): resolved existing linting errors

evwilkin:
**What**: Towards #3592 - resolves linting errors for react-docs


packages/patternfly-4/react-docs/gatsby-config.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
 const path = require('path');
```
dlabrecq: This is a build script. Not sure we need to apply linter here? Perhaps a different config for js?

evwilkin: Thanks - updated this!

PR #3719, opened 2020-02-11
chore(react-styled-system): fixed linter errors

evwilkin:
**What**: Towards #3592 - resolves existing linter issues for react-styled-system

### DIFF
--- a/packages/patternfly-4/react-styled-system/scripts/copyTS.js
+++ b/packages/patternfly-4/react-styled-system/scripts/copyTS.js
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 const path = require('path');
 const glob = require('glob');
 const fse = require('fs-extra');
+/* eslint-enable @typescript-eslint/no-var-requires */
 
 const srcDir = path.join('./src');
 const distDir = path.join('./dist/js');
--- a/packages/patternfly-4/react-styled-system/scripts/copyTS.js
+++ b/packages/patternfly-4/react-styled-system/scripts/copyTS.js
@@ -1,8 +1,6 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
 const path = require('path');
 const glob = require('glob');
 const fse = require('fs-extra');
-/* eslint-enable @typescript-eslint/no-var-requires */
 
 const srcDir = path.join('./src');
 const distDir = path.join('./dist/js');
--- a/packages/patternfly-4/react-styled-system/src/components/StyledSystem/StyledTheme.d.ts
+++ b/packages/patternfly-4/react-styled-system/src/components/StyledSystem/StyledTheme.d.ts
@@ -1,5 +1,6 @@
 import { FunctionComponent, HTMLProps, ReactNode } from 'react';
 
+/* eslint-disable camelcase */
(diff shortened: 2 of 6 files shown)


packages/patternfly-4/react-styled-system/scripts/copyTS.js
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 const path = require('path');
 const glob = require('glob');
 const fse = require('fs-extra');
+/* eslint-enable @typescript-eslint/no-var-requires */
```
dlabrecq: This is a build script. Not sure we need to apply linter here? Perhaps a different config for js?

packages/patternfly-4/react-styled-system/src/components/StyledSystem/StyledTheme.d.ts
```diff
@@ -161,7 +162,7 @@ export interface StyledConstants {
 }
 
 export interface PatternFlyThemeProviderProps extends HTMLProps<HTMLDivElement> {
-  theme?: Object;
+  theme?: Record<string, any>;
```
dlabrecq: Do we want to change the type or wait for a breaking change release?

packages/patternfly-4/react-styled-system/src/components/StyledSystem/StyledTheme.d.ts
```diff
@@ -161,6 +162,7 @@ export interface StyledConstants {
 }
 
 export interface PatternFlyThemeProviderProps extends HTMLProps<HTMLDivElement> {
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
seanforyou23: Do you know why we need the `Object` type here? If we can safely remove the general type in favor of the primitive `object` I think that's the way to go, then we don't need the ignore eslint statement

packages/patternfly-4/react-styled-system/src/components/StyledSystem/StyledTheme.d.ts
```diff
@@ -161,7 +162,7 @@ export interface StyledConstants {
 }
 
 export interface PatternFlyThemeProviderProps extends HTMLProps<HTMLDivElement> {
-  theme?: Object;
+  theme?: object;
```
dlabrecq: Shouldn't this be `Object` not lowecase

evwilkin: @seanforyou23 pointed out [above](https://github.com/patternfly/patternfly-react/pull/3719#discussion_r379619119) that typically [Object is used incorrectly and object is preferred](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types)

packages/patternfly-4/react-styled-system/src/components/StyledSystem/StyledTheme.d.ts
evwilkin: Updated based on [https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types](feedback)

PR #3723, opened 2020-02-11
Chore/lint react styles

evwilkin:
**What**: Towards #3592 - resolves existing linter errors for react-styles

### DIFF
--- a/packages/patternfly-4/react-styles/scripts/copyStyles.js
+++ b/packages/patternfly-4/react-styles/scripts/copyStyles.js
@@ -1,6 +1,7 @@
-/* eslint-disable no-case-declarations */
+/* eslint-disable @typescript-eslint/no-var-requires */
 const { copySync } = require('fs-extra');
 const { resolve, dirname, join } = require('path');
+/* eslint-enable @typescript-eslint/no-var-requires */
 
 const toDir = resolve(__dirname, '../css');
 const fromDir = dirname(require.resolve('@patternfly/patternfly/package.json'));
--- a/packages/patternfly-4/react-styles/scripts/copyStyles.js
+++ b/packages/patternfly-4/react-styles/scripts/copyStyles.js
@@ -1,7 +1,5 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
 const { copySync } = require('fs-extra');
 const { resolve, dirname, join } = require('path');
-/* eslint-enable @typescript-eslint/no-var-requires */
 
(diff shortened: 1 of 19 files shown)


packages/patternfly-4/react-styles/scripts/copyStyles.js
```diff
@@ -1,6 +1,7 @@
-/* eslint-disable no-case-declarations */
+/* eslint-disable @typescript-eslint/no-var-requires */
 const { copySync } = require('fs-extra');
 const { resolve, dirname, join } = require('path');
+/* eslint-enable @typescript-eslint/no-var-requires */
```
dlabrecq: This is more of a build script. Should we apply linter rules here? 

Perhaps we should create a different config for js?

packages/patternfly-4/react-styles/src/build/jest/snapshot-serializer/cssUtils.js
```diff
@@ -39,16 +58,19 @@ function getComputedStyles(classNames, cssStr) {
   );
   const cs = dom.window.getComputedStyle(dom.window.document.body.children[0]);
   const values = {};
-  for (let i = 0; i < cs.length; i++) {
-    const key = cs[i];
+  for (const key of cs.length) {
     const value = cs
       .getPropertyValue(key)
-      .replace(/rgb\(([\d|,|\s]+)\)/g, (full, match) => rgbToHex(...match.split(',').map(n => parseInt(n, 10))));
+      .replace(/rgb\(([\d|,|\s]+)\)/g, (full, match) => rgbToHex(...match.split(',').map(n => parseInt(n))));
```
ddonahue007: We should be assigning the Radix '10' to the parseInt function. Not doing this can give us unreliable results from the function call.

packages/patternfly-4/react-styles/scripts/copyStyles.js
```diff
@@ -1,6 +1,7 @@
-/* eslint-disable no-case-declarations */
+/* eslint-disable @typescript-eslint/no-var-requires */
```
dlabrecq: revert the no-var-requires rule

packages/patternfly-4/react-styles/src/build/jest/snapshot-serializer/cssUtils.js
evwilkin: Added radix 10 back in

PR #3725, opened 2020-02-11
Chore/lint react tokens

evwilkin:
**What**: Towards #3592 - resolves existing linter errors for react-tokens


packages/patternfly-4/react-tokens/src/generateTokens.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable global-require,import/no-dynamic-require,no-restricted-globals */
+/* eslint-disable @typescript-eslint/no-var-requires */
```
dlabrecq: Should we remove the no-var-requires rule? Or perhaps have a different linter config for js?

dlabrecq: Isn't this more of a build script. Should we apply the linter here?

dlabrecq: Revert the no-var-requires rule

evwilkin: Remove no-var-requires rule

PR #3745, opened 2020-02-13
fix(docs): updated table property descriptions

ddonahue007:
Updated table property descriptions

### DIFF
--- a/packages/patternfly-4/react-table/src/components/Table/Table.tsx
+++ b/packages/patternfly-4/react-table/src/components/Table/Table.tsx
@@ -203,13 +203,17 @@ export interface IRow extends RowType {
 }
 
 export interface TableProps {
+  /** Adds accessible text to the Table */
   'aria-label'?: string;
   /** Content rendered inside the Table */
   children?: React.ReactNode;
   /** Additional classes added to the Table  */
   className?: string;
+  /** Style variant for the Table  */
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/react-table/src/components/Table/Table.tsx
```diff
@@ -219,25 +223,37 @@ export interface TableProps {
   onExpand?: OnExpand;
   /** Function triggered when a row's checkbox is selected. When this is used, one checkbox will be positioned in the first or second cell of a non-expandable row */
   onSelect?: OnSelect;
+  /** Enables or Disables the ability to select all  */
   canSelectAll?: boolean;
   /** Function triggered when sort icon is clicked */
   onSort?: OnSort;
+  /** Action to add to the Table */
```
seanforyou23: Can we make `Action` plural here?

packages/patternfly-4/react-table/src/components/Table/Table.tsx
```diff
@@ -203,13 +203,17 @@ export interface IRow extends RowType {
 }
 
 export interface TableProps {
+  /** Adds accessible text to the Table */
```
seanforyou23: Might consider rewording this to say that it adds "_an accessible name for_" the Table

Issue #3250, opened 2019-10-31
PF4: BackgroundImage produces console error


keithjgrant:
The BackgroundImage component generates a console error due to the empty `width` attribute on a `<filter>`. Probably caused by https://github.com/patternfly/patternfly-react/pull/2742

Chrome:
![Screen Shot 2019-10-31 at 11 19 10 AM](https://user-images.githubusercontent.com/410794/67975076-b6ca6e80-fbd0-11e9-8153-1848b93ff0b1.png)

Firefox:
![Screen Shot 2019-10-31 at 11 19 19 AM](https://user-images.githubusercontent.com/410794/67975091-bcc04f80-fbd0-11e9-9106-51c3ff4e8885.png)

This is in react-core v3.120.2. Seems like a regression of https://github.com/patternfly/patternfly-react/issues/2100

nicolethoen:
Just in case I am not in Monday - here are my thoughts so far:

I wonder if a valid solution would be to remove the width=“” property all together from the <filter> tag in the background image component. 
I don’t see a difference and it makes the errors go away. 

Also the example here: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/filter does not have a width in the filter tag. I’m not sure what the intent is for having it there.

nicolethoen:
I do see in the documentation here: https://www.w3.org/TR/SVG11/filters.html#FilterEffectsRegion

"Negative values for ‘width’ or ‘height’ are an error (see Error processing). Zero values disable rendering of the element which referenced the filter."

so I imagine that is why it was there with a "0" value originally?

seanforyou23:
@keithjgrant I took a stab at fixing this in https://github.com/patternfly/patternfly-react/pull/3758, would you mind test driving that PR to see if the error is gone? Thx!

keithjgrant:
@seanforyou23 looks like that fixes it for us. thanks!

Issue #2693, opened 2019-08-13
PF4 Login Page background image doesn't show up in Firefox


akshygupt:
System Details:
Operating System: Fedora 29
Browser: Firefox Quantum 68
@patternfly/react-core:

If you visit the login page [example](https://www.patternfly.org/v4/documentation/react/components/loginpage/) provide in the documentation, the background image doesn't show up. 

This seems to be an issue with the filter property, if i were to remove this, the image starts to show up

.pf-c-background-image{
...
~~--pf-c-background-image--Filter: url(#image_overlay);~~
}

Could someone please take a look into this?

jschuler:
@mcoker can you see from the DOM why the filter is not working? (I am not sure how the filter works :) )

mcoker:
@jschuler in https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/BackgroundImage/BackgroundImage.tsx#L82 we should remove the `0` from `width` so that it is just an empty value (`width=""`)

PR #3758, opened 2020-02-17
Remove width attr from BackgroundImage filter element

seanforyou23:
This PR removes the `width` attribute for the filter element in BackgroundImage. Looks like it could also be set to the default value of 120%, but omitting it also seems to work so taking that route. Should help clear console errors when using this component.

**What**: Closes https://github.com/patternfly/patternfly-react/issues/3250


redallen:
I tried to remove this attribute a few months ago. I settled for setting it to `""` instead of `"0"` because of a Firefox bug I think? (see #2693) However, this PR appears to work in Firefox.

If @mcoker approves, this is good by me and we should change [the HTML in Core too.](https://pf4.patternfly.org/documentation/core/components/backgroundimage)

PR #3761, opened 2020-02-17
Version fix 3759

ddonahue007:
I missed a few versions

### DIFF
--- a/packages/patternfly-4/react-inline-edit-extension/package.json
+++ b/packages/patternfly-4/react-inline-edit-extension/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@patternfly/patternfly": "2.65.0",
     "@patternfly/react-core": "^3.140.7",
-    "@patternfly/react-icons": "^3.15.1",
+    "@patternfly/react-icons": "^3.15.0",
     "@patternfly/react-styles": "^3.7.2",
     "@patternfly/react-table": "^2.27.7",
     "classnames": "^2.2.5",
--- a/packages/patternfly-4/react-table/package.json
+++ b/packages/patternfly-4/react-table/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@patternfly/patternfly": "2.65.0",
     "@patternfly/react-core": "^3.140.7",
-    "@patternfly/react-icons": "^3.15.1",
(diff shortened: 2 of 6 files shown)


packages/patternfly-4/react-topology/package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "@patternfly/patternfly": "2.65.0",
     "@patternfly/react-core": "^3.140.6",
```
jschuler: this one also?

jschuler: actually i do see a prerelease with 3.140.6 in npm

packages/patternfly-4/react-inline-edit-extension/package.json
```diff
@@ -28,8 +28,8 @@
   "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/",
   "dependencies": {
     "@patternfly/patternfly": "2.65.0",
-    "@patternfly/react-core": "^3.140.6",
```
evwilkin: The latest prerelease for react-core is 3.140.6, so I believe this needs to stay as-is?

packages/patternfly-4/react-table/package.json
```diff
@@ -28,8 +28,8 @@
   "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/react-table#readme",
   "dependencies": {
     "@patternfly/patternfly": "2.65.0",
-    "@patternfly/react-core": "^3.140.6",
```
evwilkin: The latest prerelease for react-core is 3.140.6, so I believe this needs to stay as-is?

packages/patternfly-4/react-topology/package.json
ddonahue007: Is this al set?  This is not a version i changed, it appears to be the latest pre-release though

evwilkin: Yes I think this looks correct, the latest prerelease is 3.140.6 - the other places where this changes (react-inline-edit-extension/package.json and react-table/package.json) should be reverted back to match this 3.140.6 version number

Issue #3730, opened 2020-02-12
PF4 Alert: React does not recognize the `variantLabel` prop on a DOM element.


Hyperkid123:
Using the Alert component injects `variantLabel` prop to the alert action. This prop is only accepted by the `AlertActionCloseButton` and other AlerAction components.

There are two possible solutions to this

1. Stop enforcing `AlertActionCloseButton` (and others) to be used as the action component by stopping cloning the action and injection of additional props. Let the consumers take care of accessibility. This would be a preferred option IMO because we can actually use component composition.
2. Instead of using type node for the actions, make the action a render function and pass these additional props as a `innerProps` or something. Then the dev could take these accessibility props and apply them correctly.
3. Remove the `action` prop from the alert and replace it by `onClose` (and others for other alert action components) prop. The only accepted components are `AlertAction*`. Customization of the action is pointless and we only need the actions.

Line of code responsible: https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Alert/Alert.tsx#L85

karelhala:
@tlabaj can you look at this issue and prioritize it? It is kinda pressing issue, because whenever we show toast notification we receive error in dev tools that the `variantLabel` prop is not recognised.

seanforyou23:
This bug was captured as part of https://github.com/patternfly/patternfly-react/issues/2847

tlabaj:
@seanforyou23 I converted t#2847 to an epic and added this issue to the epic.  I will create issues for the other components as well.

tlabaj:
@Hyperkid123 @karelhala This issue has been investigated by @seanforyou23 and he has a PR up.  The fix would be considered a breaking change and we have it scheduled for our 2020.04 milestone.

Issue #2847, opened 2019-09-05
Console errors in test suite


seanforyou23:
Below is a recent list of components that are throwing errors in the jest test suite.

- [ ] VirtualizedTable.test.tsx
- [ ] Alert.test.tsx
- [ ] OptionsMenu.test.tsx
- [ ] SelectToggle.test.tsx
- [ ] DropdownToggle.test.tsx
- [ ] ContextSelectorItem.test.tsx
- [ ] EmptyStateIcon.test.tsx

seanforyou23:
FWIW I see that running the test suite multiple times in a row may generate a slightly different set of results, but the reoccurring theme is that there are several warnings and sometimes errors that could stand to be addressed.

seanforyou23:
I'm able to reproduce the errors consistently now. Updated the description to reflect latest master

PR #3771, opened 2020-02-18
fix(alert): fix console error

seanforyou23:
**What**: This PR fixes the `React does not recognize the "variantLabel" prop on a DOM element.` bug in Alert.

Closes https://github.com/patternfly/patternfly-react/issues/3730

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/2847

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -35,6 +35,13 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   isLiveRegion?: boolean;
 }
 
+interface AlertContext {
+  title: React.ReactNode;
+  variantLabel?: string;
+}
+
+export const AlertContext = React.createContext<AlertContext>(null);
+
 const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
   variant = AlertVariant.info,
   isInline = false,
@@ -49,7 +56,7 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
   ouiaId = null,
   ...props
 }: AlertProps & InjectedOuiaProps) => {
-  const readerTitle = (
+  const getHeadingContent = (
     <React.Fragment>
       <span className={css(accessibleStyles.screenReader)}>{variantLabel}</span>
       {title}
@@ -78,11 +85,13 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
       })}
     >
       <AlertIcon variant={variant} />
-      <h4 className={css(styles.alertTitle)}>{readerTitle}</h4>
+      <h4 className={css(styles.alertTitle)}>{getHeadingContent}</h4>
       {children && <div className={css(styles.alertDescription)}>{children}</div>}
-      {action && (
-        <div className={css(styles.alertAction)}>{React.cloneElement(action as any, { title, variantLabel })}</div>
-      )}
+      <AlertContext.Provider value={{ title, variantLabel }}>
+        {action && (typeof action === 'object' || typeof action === 'string') && (
+          <div className={css(styles.alertAction)}>{action}</div>
+        )}
+      </AlertContext.Provider>
     </div>
   );
 };
--- a/packages/react-core/src/components/Alert/AlertActionCloseButton.tsx
+++ b/packages/react-core/src/components/Alert/AlertActionCloseButton.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Button, ButtonVariant, ButtonProps } from '../Button';
 import TimesIcon from '@patternfly/react-icons/dist/js/icons/times-icon';
+import { AlertContext } from '..';
 
 interface AlertActionCloseButtonProps extends ButtonProps {
   /** Additional classes added to the AlertActionCloseButton */
@@ -18,16 +19,19 @@ export const AlertActionCloseButton = ({
   className = '',
   onClose = () => undefined as any,
   'aria-label': ariaLabel = '',
-  title,
-  variantLabel = '',
+  variantLabel,
   ...props
 }: AlertActionCloseButtonProps) => (
-  <Button
-    variant={ButtonVariant.plain}
-    onClick={onClose}
-    aria-label={ariaLabel === '' ? `Close ${variantLabel} alert: ${title}` : ariaLabel}
-    {...props}
-  >
-    <TimesIcon />
-  </Button>
+  <AlertContext.Consumer>
+    {({ title, variantLabel: alertVariantLabel }) => (
+      <Button
+        variant={ButtonVariant.plain}
+        onClick={onClose}
+        aria-label={ariaLabel === '' ? `Close ${variantLabel || alertVariantLabel} alert: ${title}` : ariaLabel}
+        {...props}
+      >
+        <TimesIcon />
+      </Button>
+    )}
+  </AlertContext.Consumer>
 );
--- a/packages/react-core/src/components/Alert/AlertActionLink.tsx
+++ b/packages/react-core/src/components/Alert/AlertActionLink.tsx
@@ -10,7 +10,7 @@ export interface AlertActionLinkProps extends ButtonProps {
 
 export const AlertActionLink: React.FunctionComponent<AlertActionLinkProps> = ({
   className = '',
-  children = '',
+  children,
   ...props
 }: AlertActionLinkProps) => (
   <Button variant={ButtonVariant.link} className={className} {...props}>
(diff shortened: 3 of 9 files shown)


packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
```diff
@@ -79,10 +98,11 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
...
     >
       <AlertIcon variant={variant} />
-      <h4 className={css(styles.alertTitle)}>{readerTitle}</h4>
+      <h4 className={css(styles.alertTitle)}>{getHeadingContent}</h4>
       {children && <div className={css(styles.alertDescription)}>{children}</div>}
-      {action && (
-        <div className={css(styles.alertAction)}>{React.cloneElement(action as any, { title, variantLabel })}</div>
+      {action && typeof action === 'object' && <div className={css(styles.alertAction)}>{action}</div>}
+      {action && typeof action === 'function' && (
+        <div className={css(styles.alertAction)}>{action(title, variantLabel)}</div>
```
karelhala: Can you pass the `title` and `variantLabel` as object? Also what are your thoughts on ignoring altogether the `div` and pass the alertAction className to action callback?
```JS
{action && typeof action === 'function' && action({ title, variantLabel, className: css(styles.alertAction) })}
```

This can be helpfull for consumers to write whatever they want as action, but it has drawbacks that they have to use the className from the action callback props.

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
```diff
@@ -36,11 +38,28 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   isLiveRegion?: boolean;
 }
 
+export const closeBtnAriaTxt = (ariaLabel: string, variantLabel?: string, title?: string) => {
```
jschuler: I think we should use context to pass title and variant from the Alert to the action components.
The only prop that the AlertActionCloseButton should have related to the label should be an aria-label if they want to override the default we provide. I think it's overcomplicating it a bit to have 3 different props for setting the aria-label

seanforyou23:
Thx for the feedback @karelhala @jschuler - gonna work on making some updates now. I will ping you both when it is ready for another round of review!

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
seanforyou23: @karelhala split this out into another PR so we can get the bug fixed without breaking change https://github.com/patternfly/patternfly-react/pull/3822

You can achieve something similar to what you describe without the div by using a fragment, but I see what you mean. We can definitely explore those changes over in #3822 if that works for you

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
seanforyou23: Updated, take a look and lemme know what you think!

packages/patternfly-4/react-core/src/components/Alert/Alert.tsx
karelhala: I think that this is great first step! Let's iterate on it, I think it is good to keep it this way and if any need for custom component as a wrapper arise we can address it later on.

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.tsx
```diff
@@ -9,25 +10,25 @@ interface AlertActionCloseButtonProps extends ButtonProps {
...
-  variantLabel?: string;
 }
 
 export const AlertActionCloseButton = ({
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   className = '',
   onClose = () => undefined as any,
   'aria-label': ariaLabel = '',
-  title,
-  variantLabel = '',
```
dlabrecq: This appears to be a breaking change

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.tsx
```diff
@@ -9,25 +10,25 @@ interface AlertActionCloseButtonProps extends ButtonProps {
   onClose?: () => void;
   /** Aria Label for the Close button */
   'aria-label'?: string;
-  /** Variant Label for the Close button */
-  variantLabel?: string;
```
dlabrecq: This appears to be a breaking change

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.tsx
```diff
@@ -9,25 +10,25 @@ interface AlertActionCloseButtonProps extends ButtonProps {
...
-    {...props}
-  >
-    <TimesIcon />
-  </Button>
+  <AlertContext.Consumer>
+    {({ title, variantLabel }) => (
+      <Button
+        variant={ButtonVariant.plain}
+        onClick={onClose}
+        aria-label={ariaLabel === '' ? `Close ${variantLabel} alert: ${title}` : ariaLabel}
```
jschuler: Right now if variantLabel is not set in `Alert` it will use the default value which is set to 
variantLabel = `${capitalize(variant)} alert:`, 
so it comes out like this: `Close Default alert: alert: Default alert title 1`

I think we should not have any hard-coded words here, for example if this should be in another language etc.. it will cause issues.

I think this would be better as `aria-label={ariaLabel || variantLabel}`

And you could put additional wording into the default for variantLabel in the Alert component instead?

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.tsx
seanforyou23: Oops, will revert!

packages/patternfly-4/react-core/src/components/Alert/AlertActionCloseButton.tsx
seanforyou23: The hard coded text is only the default and can be overridden by setting a custom aria-label.

The problem with defaulting to variantLabel for the aria-label if ariaLabel is not set by the user is that the variant label alone doesn't distinguish the close button from others when multiple alerts are present. This is critical, otherwise screen reader users don't have an easy way of knowing which close button they are pressing without further navigating the page for surrounding context clues.

I think any changes to the experience should probably go into a separate PR/work-item, and let's just focus on fixing the bug for this one. Does this make sense?

packages/react-core/src/components/Alert/AlertActionLink.tsx
```diff
@@ -3,14 +3,14 @@ import { Button, ButtonVariant, ButtonProps } from '../Button';
 
 export interface AlertActionLinkProps extends ButtonProps {
   /** Content rendered inside the AlertLinkAction  */
-  children?: string;
+  children: string;
```
kmcfaul: Would this change be considered breaking? 

Everything else lgtm

seanforyou23: Yea I think I changed this because otherwise there's no way to guarantee an accessible name for the component and I had run into it during some of this work (this was part of why I had previously added the accessibility/breaking-change labels)

Probably safe to revert it (although not the best for a11y), checking now.

tlabaj: We are starting breaking changes next milestone.  We can pull this in to the breaking change branch?

seanforyou23: Sure, I'll include this enhancement as part of https://github.com/patternfly/patternfly-react/pull/3822 and revert it from this PR so we can get the bug fixed.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -79,11 +87,11 @@ const Alert: React.FunctionComponent<AlertProps & InjectedOuiaProps> = ({
...
     >
       <AlertIcon variant={variant} />
-      <h4 className={css(styles.alertTitle)}>{readerTitle}</h4>
+      <h4 className={css(styles.alertTitle)}>{getHeadingContent}</h4>
       {children && <div className={css(styles.alertDescription)}>{children}</div>}
-      {action && (
-        <div className={css(styles.alertAction)}>{React.cloneElement(action as any, { title, variantLabel })}</div>
-      )}
+      <AlertContext.Provider value={{ title, variantLabel }}>
+        {action && typeof action === 'object' && <div className={css(styles.alertAction)}>{action}</div>}
```
jschuler: type for action is ReactNode so it could also be a string

packages/react-core/src/components/Alert/AlertActionCloseButton.tsx
```diff
@@ -18,16 +19,19 @@ export const AlertActionCloseButton = ({
   className = '',
   onClose = () => undefined as any,
   'aria-label': ariaLabel = '',
-  title,
```
jschuler: this is also technically a breaking change

redallen: True, but I don't think anyone was using the `title` prop as an `aria-label` on the `<Button>`.

PR #3775, opened 2020-02-18
fix(experimental): Added experimental exports 

### LINKED ISSUE #3774: Experimental should export Divider, OverflowMenu, and Spinner
Experimental directory should export Divider, OverflowMenu, and Spinner (in addition to Drawer and DataToolber which are currently exported)

evwilkin:
**What**: Closes #3774


packages/patternfly-4/react-core/src/experimental/components/Divider/index.ts
```diff
@@ -0,0 +1 @@
+export * from '../../../components/Divider';
```
tlabaj: Question.  Do we need this?  can't we export directly in packages/patternfly-4/react-core/src/experimental/components/index.ts ?

evwilkin: @tlabaj technically we don't need this, as we are exporting everything from the `/experimental/components/index.ts` file you mentioned - but I've added these in for any users who are importing directly from the component file, which was part of the original feedback we got on this:

> I just got this build error...Due to specific import we have from `@patternfly/react-core/dist/js/experimental/components/Drawer;`

tlabaj: Makes sense.  thanks!

evwilkin:
Note: had to remove Divider, OverflowMenu, and Spinner from `/react-core/src/components/index.ts` after adding their exports back into Experimental in order to avoid duplicate exports.

packages/patternfly-4/react-core/src/components/index.ts
```diff
@@ -17,7 +17,6 @@ export * from './ChipGroup';
 export * from './ClipboardCopy';
 export * from './ContextSelector';
 export * from './DataList';
-export * from './Divider';
```
tlabaj: Why did we remove these?  Does that they can only be exported form experimental

jschuler: You will still be able to do
`import { Divider } from '@patternfly/react-core` since the react-core/index barrel file exports out the experimental components which themselves point back into the components folder. For that same reason, Drawer and DataToolbar and not re-exported again in this file.

evwilkin: We had to remove Divider, OverflowMenu, and Spinner from /react-core/src/components/index.ts after adding their exports back into Experimental in order to avoid duplicate exports - when we had them in previously it was throwing errors for each and failing the build.  We did this to mirror how Drawer and DataToolbar are currently set up.

tlabaj:
@evwilkin We had duplicate exports before.  I am not understanding why we need to remove them now.  How about the consumers who might be exporting from react-core?

packages/patternfly-4/react-core/src/components/index.ts
tlabaj: Ok. I see it now.  thanks.

PR #3798, opened 2020-02-19
feat(Button): add link icon position for link buttons

### LINKED ISSUE #3283: Link Button
I'm using the Link button with an icon in a Patternfly-org page. I want the icon to sit after the text like below, however the default is for the icon to sit before the link. Could we explore the option of having another variant that lets you determine where the icon sits whether its before or after the icon.
I can use a regular link button and then drop in the icon, however I don't get the correct spacing as seen below:

Screen Shot 2019-11-07 at 11 27 20 AM

kmcfaul:
**What**: Closes #3283

### DIFF
--- a/packages/patternfly-4/react-core/src/components/Button/Button.tsx
+++ b/packages/patternfly-4/react-core/src/components/Button/Button.tsx
@@ -42,6 +42,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   type?: 'button' | 'submit' | 'reset';
   /** Adds button variant styles */
   variant?: 'primary' | 'secondary' | 'tertiary' | 'danger' | 'link' | 'plain' | 'control';
+  /** Sets position of the link icon */
+  linkPosition?: 'left' | 'right';
   /** Adds accessible text to the button. */
   'aria-label'?: string;
   /** Icon for the button if variant is a link */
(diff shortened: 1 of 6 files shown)


packages/patternfly-4/react-core/src/components/Button/Button.tsx
```diff
@@ -95,8 +98,14 @@ const Button: React.FunctionComponent<ButtonProps & InjectedOuiaProps> = ({
         'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
       })}
     >
-      {icon && variant === ButtonVariant.link && <span className="pf-c-button__icon">{icon}</span>}
-      {children}
+      {icon && variant === ButtonVariant.link && linkPosition === 'left' && (
```
mcoker: Not sure if this is beyond the scope of this PR since this was the previous behavior, but do we need to restrict use of `.pf-c-button__icon` to `ButtonVariant.link`? This is probably a question for @mcarrano or @mceledonia. Seems to me like it could be supported in any button, except maybe the plain variation.

Also should we call it `iconPosition` instead of `linkPosition`?

mcarrano: I don't see any reason to restrict it, what do you think @mceledonia ?

Issue #1894, opened 2019-04-30
Tooltips for disabled elements


jgiardino:
For a tooltip to be accessible for keyboard-only users, the element with the tooltip needs to be able to receive keyboard focus. Elements like `<button>` that support the `disabled` attribute cannot receive keyboard focus. 

To continue to render these elements as disabled but enable keyboard focus, the following changes are needed:
- Remove `disabled`
- Add `aria-disabled="true"`
- Make sure the `pf-m-disabled` class or similar is included on the element to apply the disabled styling.
- If the element is a link, remove `tabindex="-1"`

We should update all interactive elements to include a prop that allows the consumer to make the element disabled but focusable so the element can be used with the Tooltip component. This prop should apply the appropriate attributes and classes for the given element.

Unless someone has a better suggestion, the prop could be named `isDisabledFocusable` (totally open to suggestions on this one)

Components that should be updated are:
- [ ] Application launcher
- [x] Button (see note above about links, which are supported in the Button)
- [ ] Checkbox
- [ ] Context selector
- [ ] Dropdown (DropdownToggle, KebabToggle)
- [ ] FormSelect
- [ ] Options menu
- [ ] Radio
- [ ] Select
- [ ] Switch
- [ ] Text area
- [ ] Text input

We could simplify this list if there are any components here that we think are less likely to be disabled and with a tooltip.

rachael-phillips:
Hi @jgiardino what priority would you consider this issue?

jgiardino:
Given the possibility that consumers would choose a workaround that is not accessible, and given this solution might not be a priority to fix later when this use case is better supported, I'd make this a higher priority.

jgiardino:
We can also split this out with a subset of components that are most likely going to be affected, if that helps. I think components like Button, Dropdown, and all the form components would be a good start.

This should also be a relatively easy fix. It's a new prop that changes what attributes are defined.

LeshemNoa:
I really need a tooltip for disabled radio buttons and it's kind of a blocker for me. Currently, wrapping a radio button with a tooltip doesn't look great, and unfortunately I couldn't find a CSS workaround that's good enough.

seanforyou23:
@LeshemNoa we've just pulled this into our current milestone and will bump radio button up in the list so it's worked on soon. Will ping you for review once it's ready!

seanforyou23:
[...]

PR #3801, opened 2020-02-19
feat(button): add isAriaDisabled prop

seanforyou23:
**What**: This PR introduces a new prop `isAriaDisabled`. When provided as a prop, buttons are set into a partially abled state, where the primary behavior and events are suppressed - but the button can still be focused. This way, users can attach tooltips when the buttons are not yet enabled. Click and keypress events are suppressed by default and can be overridden by specifying an event list to `inoperableEvents`. It's a more flexible version of `isDisabled`.

**Why**: People want to be able to decorate "disabled" element with tooltips to describe why they are disabled.

Tested in VO and the tooltip content is announced as one would expect, following any accessible text for the button itself.

Closes https://github.com/patternfly/patternfly-react/issues/3786


seanforyou23:
Thx @jgiardino  - good call on the hover action. We are only suppressing [click and keypress](https://github.com/patternfly/patternfly-react/pull/3801/files#diff-e12cc92cf54b18eb2db75d97d17c772fR77-R84) events currently. I think the hover/mouseenter effect is more easily (and better) controllable via the `pointer-events` CSS property so I'd say we could simply set that property to achieve this effect. I've pushed a change to set this property inline, we probably want to make it a utility class and pull from core but you get the idea.  WDYT?

jgiardino:
Thanks for the clarification, @seanforyou23! 

I tested and now see the tooltip display on hover. I also see other hover effects happening (border color changes), which might not be intended for the button design. But up until now, these hover effects were never happening, because `.pf-m-disabled` applies `pointer-events: none;`.

@mcoker Do you have any recommendations on how to best support enabling `pointer-events` for this use case.

mcoker:
@jgiardino, @seanforyou23 and I met and agreed we can support variations of the component in core that manage `pointer-events` and any other custom styling we may need to offer for creating disabled-looking, focusable elements on a per component basis.

packages/react-core/src/components/Button/examples/Button.md
```diff
@@ -52,18 +52,46 @@ import { TimesIcon, PlusCircleIcon } from '@patternfly/react-icons';
...
+    <Button component="a" isDisabled href="https://pf4.patternfly.org/" target="_blank" variant="tertiary">
+      Tertiary link to core docs
     </Button>
     <Button component="a" href="https://pf4.patternfly.org/contribution/#modifiers" variant="link">
       Jump to modifiers in contribution guidelines
     </Button>
   </React.Fragment>
 );
 ```
+
```
mturley: It might we worth adding a note in the docs here about when a consumer would want/need to use the inoperable state over the disabled state.

mturley: I guess the example having a tooltip demonstrates that, but looking at that in the docs doesn't immediately make it clear that a tooltip won't work on a disabled button.

mturley:
Any chance we can get this merged before today's code freeze? I'm talking to a consumer who is asking for this feature.

seanforyou23:
@mturley I doubt it - the way we move forward on this sort of sets a precedent for how we manage this type of thing for a lot of components so I anticipate a few more rounds of discussion and tweaks to the implementation/documentation/recommendations on how to best craft a good experience for this case.

Is there any chance your consumer friend would be willing to review this PR or maybe the [Focusable](https://github.com/patternfly/patternfly-react/pull/3845) one to see if it adequately solves their use case? Any feedback would be really helpful!

mturley:
Sure! I'll send them along. Thanks @seanforyou23

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -69,29 +72,46 @@ const Button: React.FunctionComponent<ButtonProps & InjectedOuiaProps> = ({
...
   ouiaContext = null,
   ouiaId = null,
-  tabIndex = null as number,
+  tabIndex = null,
   ...props
 }: ButtonProps & InjectedOuiaProps) => {
   const Component = component as any;
   const isButtonElement = Component === 'button';
+  const preventDefaultActions = {
+    onKeyPress: (event: Event) => {
```
jschuler: could we see this being modifiable from props? lots of events exist https://reactjs.org/docs/events.html.
if so it could take an array list of strings and then we can dynamically build out this object

mturley: I like that idea, and we could just give it a sensible default. I wonder if the default should include anything more than `['onKeyPress', 'onClick']`, I expect most people won't bother overriding it.

seanforyou23: Good call! I [took a stab at it](https://github.com/patternfly/patternfly-react/pull/3801/commits/ad1c64b05402fff98f7598dc17eeb7bbda5b5760#diff-51ac3b180a71f3ee94153db042f8acd9R85-R86). It seems to be working as one would expect. Please let me know if you find any issues with this! I added a simple cypress test to verify this new prop `inoperableEvents` works - let me know if you can think of any other scenarios that would be good to verify.

packages/react-core/src/components/Button/examples/Button.md
seanforyou23: Good point - I made this more clear using the [tooltip text](https://github.com/patternfly/patternfly-react/pull/3801/files#diff-5ef4ad66c44ff42efa5f8359d5ccce44R77)

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -69,29 +75,68 @@ const Button: React.FunctionComponent<ButtonProps & InjectedOuiaProps> = ({
...
+
+  if (isAriaDisabled && process.env.NODE_ENV !== 'production' && !hasWarnBeta) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'You are using a beta component feature (isAriaDisabled). These api parts are subject to change in the future.'
+    );
+    hasWarnBeta = true;
+  }
+
+  const preventedEvents = Object.assign({}, linkBtnEvents(), ...getDefaultPreventedEvents());
```
mturley: Take it or leave it, but you could simplify this by replacing `getDefaultPreventedEvents` and `linkBtnEvents` and this `Object.assign` with a single object-building array reduce:

```js
const preventDefault = (event: React.SyntheticEvent<HTMLButtonElement>) => {
  event.preventDefault();
};
const preventedEvents = inoperableEvents.reduce((handlers, eventToPrevent) => ({
  ...handlers,
  [eventToPrevent]: preventDefault,
  ...(!isButtonElement && eventToPrevent === 'onKeyPress' && { onClick: preventDefault })
}), {});
```

mturley: Or if the logic in the reduce is too awkward, this works too:

```js
const eventsToPrevent =
  !isButtonElement && inoperableEvents.includes('onKeyPress')
    ? [...inoperableEvents, 'onClick']
    : inoperableEvents;
const preventedEvents = eventsToPrevent.reduce(
  (handlers, eventToPrevent) => ({
    ...handlers,
    [eventToPrevent]: (event: React.SyntheticEvent<HTMLButtonElement>) => {
      event.preventDefault();
    },
  }),
  {}
);
```

seanforyou23: Thanks! Yea I didn't feel quite settled on this, I'll give it another round of refactoring. Thx for the snippet :)

jgiardino:
This looks great! I have no suggestions in the context of react. I think the remaining changes left would be handled by core.

Testing this now, I see that the tooltip displays on focus, but not on hover:
image

I agree with @mcarrano's feedback above, but I think this is a change we need in core. For example, the variant classes (e.g. secondary, tertiary) applied to the button are providing the style changes on hover and focus. Currently, "pf-m-disabled" prevents these styles from displaying on hover by using `pointer-events: none`. Can the "pf-m-disabled" styles be defined in a way so that the variant styles are not included on hover or focus, without using `pointer-events` as it does today? This change would enable the ability to see the tooltip on both hover and focus, and also not see the unintended style changes on hover or focus as @mcarrano requested.

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -59,6 +63,8 @@ const Button: React.FunctionComponent<ButtonProps & InjectedOuiaProps> = ({
   isActive = false,
   isBlock = false,
   isDisabled = false,
+  isAriaDisabled = null,
```
dlabrecq, line 66: This should this initialized as false

packages/react-core/src/components/Button/examples/Button.md
```diff
@@ -52,18 +52,46 @@ import { TimesIcon, PlusCircleIcon } from '@patternfly/react-icons';
...
+```js title=Aria-disabled-button-with-tooltip
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+import { TimesIcon, PlusCircleIcon, Tooltip } from '@patternfly/react-icons';
+
+AriaDisabledBtnWithTooltip = () => (
+  <Tooltip content="Aria-disabled buttons are like disabled buttons, but focusable. Allows for tooltip support.">
+    <Button isAriaDisabled variant="secondary" onKeyPress={() => {alert('KeyPress')}} onClick={() => {alert('Click')}}>
```
dlabrecq, line 78: Why have `onKeypress` and `onClick` here if the button `isAriaDisabled`, is this leftover for testing? These events should not function, correct?

seanforyou23: Yes exactly, it was there to show the events are correctly being suppressed. I can move it over into react-integration and just make sure it's covered from a test perspective. Good catch!

mturley:
This is looking good to me. Have @jgiardino's comments been addressed / is it ready for a final review?

seanforyou23:
@mturley I think the react changes are ready for review, but no changes have been addressed in core https://github.com/patternfly/patternfly/issues/2946

seanforyou23:
I just realized I had misread part of what @mcarrano was saying, we do also want the tooltip on hover, just not the cursor styling. I'll work on updating this PR, sorry bout that!

mcarrano:
@seanforyou23 Looks like you fixed part of the problem. Yes, we do want the tooltip to display on hover without changing the cursor. But we also do not want to see the hover styles applied to the button itself. Right now the border color is changing on hover which is not expected. The button appearance should be static except for displaying the tooltip (i.e. continues to look like a disabled button).

jgiardino:
@mcarrano The updates you're describing would need to be handled in core with https://github.com/patternfly/patternfly/issues/2946.

I also think that core issue should handle the inline styles that @seanforyou23 recently added to enable the hover effects: `style="pointer-events: auto; cursor: default;"`. It could be debated whether those styles should even be added to react, if we expect they ultimately would be handled in core. 

The team needs to decide if this PR is blocked by that issue in core. And if not, should the inline styles be included now, and removed later when the core issue is addressed?

mcarrano:
I feel like we should hold on introducing this PR until the core issue can also be addressed. @mcoker is that something we can get into the current sprint?

seanforyou23:
Right, getting those styling adjustments implemented on core side is all we've been waiting on. I've left the inline styles in the meantime so that the behavior could be reviewed. Can definitely update this PR once core styles are in place to use instead.

mcoker:
@seanforyou23 @mcarrano sorry for the delay, there is a core PR up to enable a `.pf-m-aria-disabled` state at https://github.com/patternfly/patternfly/pull/3166

jgiardino:
Thanks for sharing the link, @mcoker! One thing I noticed in reviewing the core examples is that the react examples for Button don't include examples showing `isDisabled`. Consumers could potentially get really confused by seeing only examples for `isAriaDisabled` and none for `isDisabled`. As part of this PR, it would be great if the react examples could be updated to be consistent with the examples in core.

This also made me realize that the prop descriptions for `isDisabled` and `isAriaDisabled` could use some refinement to further clarify the differences and similarities. Here are a couple of suggestions:
- Display these two props adjacent in the list of props, if possible.
- Rewrite the props descriptions, since these are the only form of explanation on the React page, to better convey the differences and similarities.
  - Current text:
    - **isDisabled**: Disables the button and adds disabled styling
    - **isAriaDisabled**: Sets button as aria-disabled
  - Suggested text:
    - **isDisabled**: Adds disabled styling and disables the button using the `disabled` html attribute.
    - **isAriaDisabled**: Adds disabled styling and communicates that the button is disabled using the `aria-disabled` html attribute.

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -55,34 +59,66 @@ export const Button: React.FunctionComponent<ButtonProps & OUIAProps> = ({
...
   icon = null,
   ouiaId = null,
-  tabIndex = null as number,
+  tabIndex = null,
   ...props
 }: ButtonProps & OUIAProps) => {
   const Component = component as any;
   const isButtonElement = Component === 'button';
+
+  let hasWarnBeta = false;
```
redallen, line 77: Can you remove this var? We can just delete lines 79-85 once it's no longer beta.

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -55,34 +59,66 @@ export const Button: React.FunctionComponent<ButtonProps & OUIAProps> = ({
...
+    }),
+    {}
+  );
+
+  const getDefaultTabIdx = () => {
+    if (isDisabled) {
+      return isButtonElement ? null : -1;
+    } else if (isAriaDisabled) {
+      return null;
+    }
```
redallen, line 102: How about just mutating `tabIndex`?
 ```
if (tabIndex !== null) {
  tabIndex = (isAriaDisabled || (isDisabled && isButtonElement) ? null : -1);
}`

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -55,34 +59,66 @@ export const Button: React.FunctionComponent<ButtonProps & OUIAProps> = ({
...
-        isDisabled && !isButtonElement && styles.modifiers.disabled,
+        isDisabled && styles.modifiers.disabled,
+        isAriaDisabled && styles.modifiers.ariaDisabled,
         isActive && styles.modifiers.active,
         isInline && variant === ButtonVariant.link && styles.modifiers.inline,
         className
       )}
       disabled={isButtonElement ? isDisabled : null}
-      tabIndex={isDisabled && !isButtonElement ? -1 : tabIndex}
+      tabIndex={tabIndex !== null ? tabIndex : getDefaultTabIdx()}
```
redallen, line 121: That way, you can just do `tabIndex={tabIndex}`

packages/react-core/src/components/Button/Button.tsx
seanforyou23: We generally try to avoid mutations unless necessary, what's the goal of the suggestion though? It looks like something you'd get from a truth table and isn't very readable. This optimization may save a few characters but makes it difficult to understand the code's purpose in the process. Also, this prevents the user from being able to set the tabindex attribute explicitly, which would break a couple of the tests attached.

redallen: I think a single if statement with mutation is easier to follow than a ternary plus an arrow function with 2 control flows and another ternary.

I don't see any reason to "try to avoid mutations unless necessary." Setting default props is a mutation. Since we want to set a default prop based on other props, I think it's fine to do it as a mutation.

As for the logic, you're correct. The code should read
```js
if (tabIndex === null) {
 tabIndex = (isAriaDisabled || (isDisabled && isButtonElement) ? null : -1);
}
```

You could even do `tabIndex = getDefaultTabIndex(props)` in the prop destructuring and move `getDefaultTabIndex` up in scope. I think that reads nicely in the docs and is explicit.

seanforyou23: With this code, wouldn't all buttons that don't explicitly specify a tabIndex end up with `tabIndex={-1}`?

tlabaj: Thanks for discussion around this topic.  Let's just not make these updates at the moment.  The logic works in its current form, and each person has a different style of coding.  We can discuss a coding style document for future updates if we want to have a best practices guideline.

Issue #3791, opened 2020-02-19
Tooltips for disabled radio


jessiehuff:

seanforyou23:
I was able to come up with a few implementations between #3817 and #3845 that each facilitates tooltips for disabled radios, but in discussing the direction of the options we determined there is a design solution that's better suited to address the problem for radios. Not sure when/where that will manifest, but below are the sample implementations in case it is helpful.

Wrap the entire disabled Radio component in a tooltip
https://patternfly-react-pr-3817.surge.sh/documentation/react/components/radio#inoperable-with-wrapping-tooltip

Wrap only the disabled radio input element in a tooltip
https://patternfly-react-pr-3817.surge.sh/documentation/react/components/radio#inoperable-with-nested-tooltip

Wrap _anything_ in a tooltip using a "Focusable" component
https://patternfly-react-pr-3845.surge.sh/documentation/react/components/focusable#wrapping-disabled-radio-with-tooltip

thatblindgeye:
For now will remove the Stale label and keep a pulse check on this issue.

FWIW this may also require:

- Core update, to add a `pf-m-aria-disabled` class -- right now the component only suppots `pf-m-disabled`, at least on the label text.
- a prop for passing a Tooltip to the component so that it wraps the correct element -- right now in a quick test, wrapping the Radio component in a Tooltip causes the tooltip to render possibly strangely if the Radio takes up its entire horizontal space. Additionally, a tthat point focusing the radio button doesn't trigger the tooltip since the trigger is technically the wrapper that contains the input and label.

PR #3817, opened 2020-02-21
feat(radio): allow radios to be "inoperable"

seanforyou23:
**What**: This PR introduces a new prop `isInoperable`. When provided as a prop, radios are set into a partially enabled state. This way, users can still attach tooltips when the radios are not fully available for whatever reason. Click, change, and keypress events are suppressed when this prop is present so effectively the only thing you can do is focus the element and view the tooltip.

Closes # https://github.com/patternfly/patternfly-react/issues/3791

### DIFF
--- a/packages/react-core/src/components/Card/examples/Card.md
+++ b/packages/react-core/src/components/Card/examples/Card.md
@@ -7,7 +7,7 @@ propComponents: ['Card', 'CardHeadMain', 'CardHeader', 'CardBody', 'CardFooter']
 ---
 
 import { Brand, Card, CardActions, CardHead, CardHeadMain, CardHeader, CardBody, CardFooter, Checkbox, DropdownActions } from '@patternfly/react-core';
-import pfLogo from './pfLogo.svg'; 
+import pfLogo from './pfLogo.svg';
 
 ## Examples
 ```js title=Basic
@@ -24,9 +24,9 @@ SimpleCard = () => (
 ```
 
 ```js title=With-image-and-actions
-import React from 'react'; 
-import { Brand, Card, CardHead, CardHeadMain, CardActions, CardHeader, CardBody, CardFooter, Dropdown, DropdownToggle, DropdownItem, DropdownSeparator, DropdownPosition, DropdownDirection, KebabToggle, } from '@patternfly/react-core'; 
-import pfLogo from './pfLogo.svg'; 
+import React from 'react';
+import { Brand, Card, CardHead, CardHeadMain, CardActions, CardHeader, CardBody, CardFooter, Dropdown, DropdownToggle, DropdownItem, DropdownSeparator, DropdownPosition, DropdownDirection, KebabToggle, } from '@patternfly/react-core';
+import pfLogo from './pfLogo.svg';
 
 class KebabDropdown extends React.Component {
   constructor(props) {
@@ -46,11 +46,11 @@ class KebabDropdown extends React.Component {
       });
     };
     this.onClick = (checked, event) => {
-      const target = event.target; 
-      const value = target.type === 'checkbox' ? target.checked : target.value; 
-      const name = target.name; 
-      this.setState({ [name]: value }); 
-    }; 
+      const target = event.target;
+      const value = target.type === 'checkbox' ? target.checked : target.value;
+      const name = target.name;
+      this.setState({ [name]: value });
+    };
   }
 
   render() {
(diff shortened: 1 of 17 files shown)


seanforyou23:
@dlabrecq you're right at the heart of what's at stake - thanks for calling this out!

I don't think this is going against any standard. We are removing `disabled` because we are exposing a tiny bit of functionality, even if it's only focusability and hoverability with tabindex and pointer-events - that counts :) 

When this is the case, it's replaced by `[aria-disabled]` and `.pf-m-disabled`, which together achieve a partially disabled state both visually and with AT. Screen reader users are made aware that the content is (dimmed) which is the case for disabled elements, and sighted users see the control with standard disabled styling.

Agree it's expected that disabled elements behave a specific way, but also we can acknowledge that this is bleeding edge of what's even possible from an experience standpoint so we're reaching hard for gains even if it requires managing state/perception in unique ways. I think as long as it [passes standard html validation](https://validator.w3.org/nu/#textarea) then it's fair game if the community is ok with it from the UX side.

All of this is in an effort to facilitate tooltips for disabled elements. It we're ok with limiting the tooltip to being attached to the container element of the Component then we might be able to achieve the same thing without having to remove the disabled attribute on these natively disablable elements. In other words, would we be ok with supporting [only this way of using tooltips with disabled elements](https://patternfly-react-pr-3817.surge.sh/documentation/react/components/radio#disabled-focusable-with-tooltip), and not [this one](https://patternfly-react-pr-3817.surge.sh/documentation/react/components/radio#disabled-focusable-with-custom-tooltip)? Notice the difference is that in one case, we simply nest the target element within a tooltip as usual, and in the other we send the tooltip down as props. I think the latter is the potentially dangerous api we might benefit from steering clear of.

I need to explore this a bit more to ensure it is even possible and works in all the cases of how it can be used/consumed.

IMO we should be thinking of this as setting the Component (capitol C) in a disabled/focusable state, not the native controls that may be associated with the control. I think with this distinction we can reason about it better and have more options to consider. Does that help?

I did notice that the `.pf-c-radio__description` element fails html validation with `Element div not allowed as child of element label in this context.`. Using a `span` element instead seems to fix it.

seanforyou23:
@dlabrecq  I just realized I may not have actually answered your question after all of that 😆 

Yes, I think that would work also. At the very least, let's limit to managing the events/properties on the container of the React Component we're wrapping, and at best introduce a wrapper component to carry the related properties. It would be great if we don't have to do a lot of surgery on all these components that need this feature, the latter probably does this best. Does that help?

dlabrecq:
We should ensure the community is really ok with it, perhaps push for a change in the standard, before introducing something unexpected? An HTML element is expected to be disabled, not just look that way.

seanforyou23:
If we can use the approach where we introduce a container to carry this functionality, it seems like it would eliminate the problem, right? If so, that also has my vote.

dlabrecq:
Yes, that would be my preference. As long as were not circumventing what disabled means for an HTML element.

seanforyou23:
Does not having the disabled property at the DOM level cause an issue with interaction or behavior of the component? I'm trying to better understand any limitations to the approach we've explored here. In addition to making it look visually disabled, we've also blocked the common/relevant events that people associate with an "enabled" element. The only (intentional) interactions we've exposed are focusability and hoverabilty. I just want to make sure I understand what problem this is presenting. Thanks!

seanforyou23:
Does not having the disabled property at the DOM level cause an issue with interaction or behavior of the component? I'm trying to better understand any limitations to the approach we've explored here. In addition to making it look visually disabled, we've also blocked the common/relevant events that people associate with an "enabled" element. The only (intentional) interactions we've exposed are focusability and hoverabilty. I just want to make sure I understand what problem this is presenting from the experience standpoint. Thanks!

seanforyou23:
Hmm I see. I suppose one way to identify this case would be to check for the presence of `aria-disabled` and the absence of `disabled` at the same time. I think the container/wrapper based approach would also use this markup strategy. What do you think of using this as the primary method of verification for testing?

Do you think the name `isDisabledFocusable` is confusing? I wonder if `isLimitedAvailability` (or something like that) would be a better reflection of the end resulting effect. I can see how using "disabled" as part of the prop name would invite people to expect to find a disabled property in there.

I can start exploring a container-based approach some we have more options to compare. Also, I think some of these pointer-event adjustments are no longer necessary now that we're preventing default actions for click, keypress, and change events. I think I added the pointer-event bits before I was properly preventing default actions - will send updates for this soon.

seanforyou23:
Here's an alternate approach we could consider - https://github.com/patternfly/patternfly-react/pull/3845

PR #3821, opened 2020-02-24
feat(Modal): add description property

### LINKED ISSUE #3777: PF4 Modal: Modal enhancements for table column management
related core issue https://github.com/patternfly/patternfly-next/issues/2636

kmcfaul:
+ move ID to description when it exists.

**What**: Closes #3777

### DIFF
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -44,6 +44,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   appendTo?: HTMLElement | (() => HTMLElement);
   /** Flag to disable focus trap */
   disableFocusTrap?: boolean;
+  /** Description of the modal */
+  description?: React.ReactNode;
 }
 
 interface ModalState {
--- a/packages/react-core/src/components/Modal/ModalContent.tsx
+++ b/packages/react-core/src/components/Modal/ModalContent.tsx
@@ -7,6 +7,7 @@ const FocusTrap: any = require('focus-trap-react');
 
 import styles from '@patternfly/react-styles/css/layouts/Bullseye/bullseye';
 import titleStyles from '@patternfly/react-styles/css/components/Title/title';
+import modalStyles from '@patternfly/react-styles/css/components/ModalBox/modal-box';
 import { css } from '@patternfly/react-styles';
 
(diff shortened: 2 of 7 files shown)


packages/patternfly-4/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -29,6 +30,8 @@ export interface ModalContentProps {
   isOpen?: boolean;
   /** Complex header (more than just text), supersedes title for header content */
   header?: React.ReactNode;
+  /** Description */
+  description?: React.ReactNode;
```
tlabaj: SHouldn't we also add this prop to `Modal` so the consumer knows that is where they pass it?

mcoker: Does this mean the description can contain markup, too, and doesn't have to just be a string?

tlabaj: @mcoker yes.

packages/patternfly-4/react-core/src/components/Modal/examples/Modal.md
```diff
@@ -62,6 +64,58 @@ class SimpleModal extends React.Component {
...
+    return (
+      <React.Fragment>
+        <Button variant="primary" onClick={this.handleModalToggle}>
+          Show Modal
+        </Button>
+        <Modal
+          title="Modal Header"
+          isOpen={isModalOpen}
+          onClose={this.handleModalToggle}
+          description="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
```
redallen: Can you change the description to match core so it's clear which text is the description? The text from core is:
> A description is used when you want to provide more info about the modal than the title is able to describe. The content in the description is static and will not scroll with the rest of the modal body.

packages/patternfly-4/react-core/src/components/Modal/examples/Modal.md
```diff
@@ -62,6 +64,58 @@ class SimpleModal extends React.Component {
 }
 ```
 
+```js title=Basic-with-description
```
redallen: Change to just "With description" to be consistent with https://pf4.patternfly.org/documentation/core/components/modalbox#with-description

PR #3822, opened 2020-02-24
Flexible alert actions

seanforyou23:
**What**: Splitting out some work from #3771 - allowing actions to be any custom content, not just our action/close buttons. 

**Additional issues**: #3771 

Screen Shot 2020-02-24 at 6 50 16 PM

## Breaking changes
1. **Alert**: Alert `action` prop uses new type of [`AlertActionRenderer | React.ReactNode`](https://github.com/patternfly/patternfly-react/pull/3822/files#diff-3c7439478e25fad3c720e081f6dfc4bfR33). Previous valid values remain valid, but the type applied may need to be updated accordingly.

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -14,6 +14,14 @@ export enum AlertVariant {
   default = 'default'
 }
 
+export type AlertActionRenderer = ({
+  title,
+  variantLabel
+}: {
+  title: React.ReactNode;
+  variantLabel: string;
+}) => React.ReactNode;
+
 export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'action' | 'title'> {
   /** Adds Alert variant styles  */
   variant?: 'success' | 'danger' | 'warning' | 'info' | 'default';
@@ -21,8 +29,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   isInline?: boolean;
   /** Title of the Alert  */
   title: React.ReactNode;
-  /** Action button to put in the Alert. Should be <AlertActionLink> or <AlertActionCloseButton> */
-  action?: React.ReactNode;
+  /** Action button to put in the Alert. Often <AlertActionLink> or <AlertActionCloseButton> */
+  action?: AlertActionRenderer | React.ReactNode;
   /** Content rendered inside the Alert */
   children?: React.ReactNode;
   /** Additional classes added to the Alert  */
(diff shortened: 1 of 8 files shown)


jschuler:
needs to be rebased

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -80,6 +88,9 @@ export const Alert: React.FunctionComponent<AlertProps & OUIAProps> = ({
         {action && (typeof action === 'object' || typeof action === 'string') && (
```
redallen, line 88: Perhaps a better API would be to leave `action`'s type as `React.ReactNode` and use `React.cloneElement` to add the `title` and `variantLabel` props? This way class components could be passed as well, and we don't have to amend the API.

```tsx
{React.isValidElement(action) && (
  <div className={css(styles.alertAction)}>
    {React.cloneElement(action as React.ReactElement, {
      title,
      variantLabel
    })}
  </div>
)}
```

Consumers would then be able to use added `title` and `variantLabel` props like this:

```tsx
<Alert action={({title, variantLabel}) => <div>{title}{variantLabel}</div>} />
```

OR like this:

```tsx
class MyAlertAction extends React.Component {
  render() {
    return <div>{this.props.title}{this.props.variantLabel}</div>;
  }
}

<Alert action={<MyAction />} />
```

packages/react-core/src/components/Alert/AlertActionLink.tsx
```diff
@@ -3,7 +3,7 @@ import { Button, ButtonVariant, ButtonProps } from '../Button';
 
 export interface AlertActionLinkProps extends ButtonProps {
   /** Content rendered inside the AlertLinkAction  */
-  children?: string;
+  children: string;
```
jschuler, line 6: Should be ReactNode so it could work with i18n libraries for example

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -80,6 +88,9 @@ export const Alert: React.FunctionComponent<AlertProps & OUIAProps> = ({
...
           <div className={css(styles.alertAction)}>{action}</div>
         )}
+        {action && typeof action === 'function' && (
+          <div className={css(styles.alertAction)}>{action({ title, variantLabel })}</div>
```
jschuler, line 92: I don't understand, couldn't this be
`<div className={css(styles.alertAction)}>{action}</div>` and since consumer is passing in `title` and `variantLabel` as props they already have knowledge of the props so aren't we just giving it back to them?

PR #3826, opened 2020-02-25
feat(ChipGroup): Added a tooltip to the Chipgroup label

### LINKED ISSUE #3490: Chip group: add tooltip to chip group label when text overflows
When a chip group's label text overflows, we should show the full text on hover via a tooltip, similar to the behavior on a single chip when the chip text overflows.

Screen Shot 2020-01-10 at 10 27 04 AM

cc @mcarrano

related core issue: https://github.com/patternfly/patternfly-next/issues/2524

tlabaj:
**What**: Closes #3490

### DIFF
--- a/packages/react-core/src/components/ChipGroup/Chip.tsx
+++ b/packages/react-core/src/components/ChipGroup/Chip.tsx
@@ -116,6 +116,7 @@ class Chip extends React.Component<ChipProps & InjectedOuiaProps, ChipState> {
     return (
       <Component
         className={css(styles.chip, isReadOnly && styles.modifiers.readOnly, className)}
+        tabIndex="0"
         {...(ouiaContext.isOuia && {
           'data-ouia-component-type': 'Chip',
           'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
--- a/packages/react-core/src/components/ChipGroup/Chip.tsx
+++ b/packages/react-core/src/components/ChipGroup/Chip.tsx
@@ -91,6 +91,7 @@ class Chip extends React.Component<ChipProps & InjectedOuiaProps, ChipState> {
         <Tooltip position={tooltipPosition} content={children}>
           <Component
             className={css(styles.chip, isReadOnly && styles.modifiers.readOnly, className)}
+            tabIndex="0"
             {...(ouiaContext.isOuia && {
               'data-ouia-component-type': 'Chip',
               'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
@@ -116,7 +117,6 @@ class Chip extends React.Component<ChipProps & InjectedOuiaProps, ChipState> {
     return (
       <Component
         className={css(styles.chip, isReadOnly && styles.modifiers.readOnly, className)}
-        tabIndex="0"
         {...(ouiaContext.isOuia && {
           'data-ouia-component-type': 'Chip',
           'data-ouia-component-id': ouiaId || ouiaContext.ouiaId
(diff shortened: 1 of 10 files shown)


packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -19,47 +20,113 @@ export interface ChipGroupToolbarItemProps extends React.HTMLProps<HTMLUListElem
...
-      <ChipGroupContext.Consumer>
-        {(HeadingLevel: any) => <GenerateId>{randomId => renderChipGroup(randomId, HeadingLevel)}</GenerateId>}
-      </ChipGroupContext.Consumer>
-    );
+  componentDidMount() {
+    this.setState({
+      isTooltipVisible: Boolean(
+        this.heading.current && this.heading.current.offsetWidth < this.heading.current.scrollWidth
+      )
+    });
```
dlabrecq: Thank you for implementing this tooltip feature. We can definitely use this in Cost Management.

Instead of setting `isTooltipVisible` state, could you test this in the `render`? I'm wondering if `offsetWidth` or `scrollWidth` can change?

It appears that the `render` may not step into the `this.state.isTooltipVisible` block the first time through?

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
```diff
@@ -19,47 +20,113 @@ export interface ChipGroupToolbarItemProps extends React.HTMLProps<HTMLUListElem
...
+      const renderChipGroup = (id: string, HeadingLevel: any) => {
+        if (this.state.isTooltipVisible) {
+          return (
+            <ul className={css(styles.chipGroup, styles.modifiers.toolbar, className)} {...rest}>
+              <li>
+                <Tooltip position={tooltipPosition} content={categoryName}>
+                  <HeadingLevel tabIndex="0" ref={this.heading} className={css(styles.chipGroupLabel)} id={id}>
+                    {categoryName}
+                  </HeadingLevel>
+                </Tooltip>
```
dlabrecq: Could make `HeadingLevel` a const and wrap this code with a Boolean? Other than adding the Tooltip, the following code block looks the same.

```
const headingLevel = (
  <HeadingLevel tabIndex="0" ref={this.heading} className={css(styles.chipGroupLabel)} id={id}>
    {categoryName}
  </HeadingLevel>
);
...
<li>
   {Boolean(isTooltipVisible) ? (
     <Tooltip position={tooltipPosition} content={categoryName}>
       headingLevel
     </HeadingLevel>
    ): (
      headingLevel
    )}
<ul className={css(styles.chipGroup)}>{children}</ul>
...
```

packages/patternfly-4/react-core/src/components/ChipGroup/ChipGroupToolbarItem.tsx
tlabaj: Thanks for the feedback Dan.  For this I tried to follow what was already implemented for the Chip.  Do you think the Chip needs to be updated too?

dlabrecq: Yes, I suspect the `Chip` should be updated as well. It seems problematic to set `isTooltipVisible` once.

dlabrecq: It seems odd that we need to render twice, but the `ref` needs to be assigned to `HeadingLevel` first. Setting state in `componentDidMount` triggers another call to `render`, which is why it works the second time thru. Let's leave it as is.

PR #3830, opened 2020-02-25
fix(CatalogTile): Removed truncation and maxLength props

### LINKED ISSUE #3471: PF4 Catalog Tile: Drop support for truncation function
We want to drop support for Catalog Tile's truncation function in patternfly-react-extensions after we migrate OpenShift off of using it. It's not needed once the wip improvements go through since the tile will automatically truncate text itself. This is a placeholder issue so I remember to take care of it.

This is not blocking.

OpenShift, tbd.

rebeccaalpert:
Descriptions are now truncated via CSS, so the truncation function and maxLength props are unnecessary. The only folks using this are migrating off of it in a current PR.

Fixes #3471.


rebeccaalpert:
@jschuler - Only OpenShift is using this, and we have a PR up to transition them off the truncation/maxLength props. If this needs to wait for the breaking changes release, though, that is fine.

redallen:
I have approval from @tlabaj , in it goes!

PR #3838, opened 2020-02-27
fix(select): fix group select options in single variant

### LINKED ISSUE #3780: PF4 Select test suite console errors
Fix console errors in Select.test.tsx

boaz0:
**What**: Closes #3780 

When running the tests in `Select.test.tsx` we get an error message saying children element doesn't have `sendRef` and `isSelected` in its props.

I tracked what tests cause this error message which led me to this particular test:
https://github.com/patternfly/patternfly-react/blob/692ea8492d70c7d231535f4663c565f1519da6cf/packages/patternfly-4/react-core/src/components/Select/__tests__/Select.test.tsx#L125-L130

After reading the `Select.tsx` code, I noticed that `isGrouped` is not passed to `SelectMenu` when variant is `single` -- which I fixed.
In addition, I had to add a case when `isGrouped` is true when displaying the selected value in the select box because the selected option is a grand child of the Select component.

Unfortunately, this change causes the DOM to look differently which forced me to update the snapshots.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -292,8 +292,15 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
       return;
     }
 
-    const { children } = this.props;
-    const item = children.filter(child => child.props.value.toString() === value.toString())[0];
+    const { children, isGrouped } = this.props;
+    let item = children.filter(
+      child => child.props.value !== undefined && child.props.value.toString() === value.toString()
+    )[0];
+    if (isGrouped) {
+      item = children
+        .reduce((acc, curr) => [...acc, ...React.Children.toArray(curr.props.children)], [])
+        .filter(child => child.props.value.toString() === value.toString())[0];
+    }
     if (item) {
       if (item && item.props.children) {
         if (type === 'node') {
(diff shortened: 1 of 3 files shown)


seanforyou23:
Thanks for this @boaz0! Would you mind adding a description of the changes required to fix the issue? Just the high level is all I'm talking about, so it's easy for people to follow along without having to read the code to understand.

boaz0:
Yes, no problem! @seanforyou23

packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -292,8 +292,15 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
       return;
     }
 
-    const { children } = this.props;
-    const item = children.filter(child => child.props.value.toString() === value.toString())[0];
+    const { children, isGrouped } = this.props;
+    let item = null;
+    if (!isGrouped) {
+      item = children.filter(child => child.props.value.toString() === value.toString())[0];
```
dlabrecq: Could you test for `isGrouped`, and make the default case be the original `item = children.filter`... code?

tlabaj:
@boaz.  This one also needs a rebase.

PR #3843, opened 2020-02-28
feat(Select): add inline filtering to checkbox select

### LINKED ISSUE #3778: PF4 Select :Add filtering to checkbox select menu
Add filtering to the checkbox Select
related core issue https://github.com/patternfly/patternfly-next/issues/2565

kmcfaul:
**What**: Closes #3778 

Requires new `hasInlineFilter` flag, and the `onFilter` callback (because grouped filtering is not a part of the default filtering function). The `onClear` callback is required for the clear button, but may be left out to opt out of the whole button. 

Theoretically can be extended to be used by other variants, but only tested and implemented in checkbox select (both grouped and ungrouped) currently.


packages/patternfly-4/react-core/src/components/Select/Select.tsx
```diff
@@ -440,6 +484,22 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
...
+                        this.clearSelection(e);
+                        onClear(e);
+                      }}
+                      aria-label={ariaLabelClear}
+                      type="button"
+                      disabled={isDisabled}
+                    >
+                      <TimesCircleIcon aria-hidden />
+                    </button>
+                  )}
```
dlabrecq: Currently, the close button is only added when using the `typeahead` feature. Adding it for `hasInlineFilter` seems inconsistent with the remaining examples?

IMO, I feel that the clear button shouldn't be limited to `typeahead` and `hasInlineFilter`. Seems like this feature would be useful, regardless? 

Can we show the clear button if `onClear` is present?

kmcfaul:
@dlabrecq @tlabaj Removed typeahead restriction on the clear button. Should now be applicable to all variants.

If `onClear` is defined and anything is selected, the toggle will change to a div similar to typeahead toggles (from a button). Is this html structure fine for select variants other than typeahead? @mcoker

mcoker:
I wonder if we should use a "Search" placeholder for the text input, similar to context selector? We also use "Filter by name" as the placeholder in the app launcher text input. In the core implementation for this text input, we use `aria-label="Search"`. Seems like maybe we could be consistent with this, context menu, and app launcher - wdyt @jessiehuff?

packages/react-core/src/components/Select/examples/Select.md
```diff
@@ -300,6 +300,111 @@ class GroupedCheckboxSelectInput extends React.Component {
...
+    const { isExpanded, selected, filteredOptions } = this.state;
+    const titleId = 'checkbox-select-id';
+    return (
+      <div>
+        <span id={titleId} hidden>
+          Checkbox Title
+        </span>
+        <Select
+          variant={SelectVariant.checkbox}
+          aria-label="Select Input"
```
mcoker: Just checking if this is on purpose? Looks like this labels the outer fieldset. In the previous checkbox select example, the `titleId` labels the outer fieldset.

kmcfaul: I think this was just a side-affect of copying another example to base it on. It has `aria-labelledby` already so this is unneeded. I will remove it.

PR #3844, opened 2020-02-28
feat(EmptyState): Added support for extra-large empty state

### LINKED ISSUE #3540: Introduce Empty State X-Large Variant
Core issue: https://github.com/patternfly/patternfly-next/issues/2472 is now merged

cc: @tlabaj

rebeccaalpert:
Added support for new size and updated examples to include extra large empty state.

Fixes #3540.

### DIFF
--- a/packages/react-core/src/components/EmptyState/EmptyState.tsx
+++ b/packages/react-core/src/components/EmptyState/EmptyState.tsx
@@ -3,14 +3,14 @@ import { css, getModifier } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/EmptyState/empty-state';
 
 export enum EmptyStateVariant {
-  extraLarge = 'extraLarge',
+  'xl' = 'xl',
   large = 'large',
   small = 'small',
   full = 'full'
 }
 
 const maxWidthModifiers: { [variant in keyof typeof EmptyStateVariant]: string } = {
-  extraLarge: 'xl',
+  xl: 'xl',
   large: 'lg',
   small: 'sm',
   full: ''
@@ -22,7 +22,7 @@ export interface EmptyStateProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside the EmptyState */
   children: React.ReactNode;
   /** Modifies EmptyState max-width */
-  variant?: 'small' | 'large' | 'full' | 'extraLarge';
+  variant?: 'small' | 'large' | 'full' | 'xl';
 }
 
 export const EmptyState: React.FunctionComponent<EmptyStateProps> = ({
--- a/packages/react-core/src/components/EmptyState/EmptyState.tsx
+++ b/packages/react-core/src/components/EmptyState/EmptyState.tsx
@@ -3,12 +3,14 @@ import { css, getModifier } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/EmptyState/empty-state';
 
 export enum EmptyStateVariant {
+  extraLarge = 'extraLarge',
   large = 'large',
   small = 'small',
   full = 'full'
 }
 
 const maxWidthModifiers: { [variant in keyof typeof EmptyStateVariant]: string } = {
+  extraLarge: 'xl',
   large: 'lg',
   small: 'sm',
   full: ''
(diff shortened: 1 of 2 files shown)


packages/patternfly-4/react-core/src/components/EmptyState/EmptyState.tsx
```diff
@@ -3,12 +3,14 @@ import { css, getModifier } from '@patternfly/react-styles';
...
 
 export enum EmptyStateVariant {
+  extraLarge = 'extraLarge',
   large = 'large',
   small = 'small',
   full = 'full'
 }
 
 const maxWidthModifiers: { [variant in keyof typeof EmptyStateVariant]: string } = {
+  extraLarge: 'xl',
```
dlabrecq: I believe we use xLarge in the grid component. We should follow that convention

yarn.lock
```diff
@@ -3819,6 +3819,7 @@
 "@patternfly/patternfly@2.65.3":
```
dlabrecq: Why is the yarn.lock being updated?

packages/patternfly-4/react-core/src/components/EmptyState/EmptyState.tsx
rebeccaalpert: It looks like Grid uses the same sm, md, lg, xl convention other components are using.

yarn.lock
rebeccaalpert: Yarn install. On Slack Zack said it was ok to commit these.

packages/patternfly-4/react-core/src/components/EmptyState/EmptyState.tsx
dlabrecq: I'm referring to the `extraLarge` prop name. Grid uses prop names like lg, xl, etc.

yarn.lock
dlabrecq: I don't see any harm, but there's really no need for this file to change, unless we're introducing new dependencies.

packages/react-core/src/components/EmptyState/EmptyState.tsx
```diff
@@ -3,12 +3,14 @@ import { css, getModifier } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/EmptyState/empty-state';
 
 export enum EmptyStateVariant {
+  extraLarge = 'extraLarge',
```
jschuler: nitpick: most components use the short-hand names, sm, lg, xl, etc. Example: EmptyStateIcon
we should try to be uniform i think. Unfortunately we already have large and small spelled out :) What if there is a xxl variant next? Maybe `xl = 'xl'` would be better?

rebeccaalpert: Would this resolve your concern @dlabrecq? I'll open a bug for this inconsistency; maybe it can get fixed in an upcoming breaking changes release.

dlabrecq: Is the suggestion to use 'xl'?

Yes, I'd rather introduce the name we want now (i.e., use 'xl') and change the remaining props during the breaking change release.

tlabaj: @rebeccaalpert can you open an issue to update the remaining prop names during breaking change.  You can put the `breaking change` label on it.

@dlabrecq @jschuler what would `full` be in shorthand?  I don't think we have that one in other components.

rebeccaalpert: @tlabaj - I opened this on Wednesday: https://github.com/patternfly/patternfly-react/issues/3873 (just added the label, thanks).

tlabaj: Thanks Rebecca!

packages/react-core/src/components/EmptyState/examples/EmptyState.md
```diff
@@ -126,6 +126,42 @@ SimpleEmptyState = () => (
...
+  EmptyStateIcon,
+  EmptyStateBody,
+  EmptyStateSecondaryActions
+} from '@patternfly/react-core';
+import { CubesIcon } from '@patternfly/react-icons';
+
+SimpleEmptyState = () => (
+  <EmptyState variant={EmptyStateVariant.xl}>
+    <EmptyStateIcon icon={CubesIcon} />
+    <Title headingLevel="h5" size="xl">
```
mcoker:
```suggestion
    <Title headingLevel="h5" size="4xl">
```

rebeccaalpert:
@mcoker - Thanks, just pushed the change.

packages/react-core/src/components/EmptyState/EmptyState.tsx
dlabrecq: Looking at the 'full' variant and it appears be an empty string. It could be 'xxl', or possibly omitted?

PR #3845, opened 2020-02-28
feat(focusable): add focusable component

seanforyou23:
**What**: This PR adds a new `Focusable` component that can be used to wrap other elements/components to make them focusable. I haven't had a chance to add cypress tests yet, but wanted to go ahead and get the work up for review in case people have time to take a look and provide feedback.

It also serves as an alternative approach to https://github.com/patternfly/patternfly-react/pull/3817 

Screen Shot 2020-02-28 at 3 34 49 PM

### DIFF
--- a/packages/react-core/src/components/Focusable/Focusable.tsx
+++ b/packages/react-core/src/components/Focusable/Focusable.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+
+export interface FocusableProps {
+  children?: string | React.ReactElement;
+  /** Any additional classes to apply to the focusable element */
+  className?: string;
+  /** An HTML element to use as the container */
+  component?: React.ElementType;
+  /** Sets the tabindex value */
+  tabIndex?: number;
+  /** Whether an intermediary DOM node container is needed */
+  withContainer?: boolean;
+  /** A prop for dynamically controlling focusability */
+  isFocusable?: boolean;
+}
+
+const renderElement = ({ children, tabIndex, className, component: Component, ...props }: FocusableProps) =>
+  typeof children === 'object'
+    ? React.cloneElement(children as React.ReactElement, {
+        tabIndex,
+        className: css(className),
+        ...props
+      })
+    : React.createElement(Component, { tabIndex, className: css(className), ...props }, children);
+
+const Focusable: React.FunctionComponent<FocusableProps> = ({
+  children,
+  className,
+  component: Component = 'div',
+  tabIndex = 0,
+  isFocusable = true,
+  withContainer = false,
+  ...props
+}) => {
+  const focusableElement = children as React.ReactElement;
+
+  const getDefaultTabIndex = () =>
+    !isFocusable || focusableElement.props.isDisabled || focusableElement.props.disabled ? tabIndex : null;
+
+  return withContainer ? (
+    <Component className={css(className)} tabIndex={getDefaultTabIndex()} {...props}>
+      {children}
+    </Component>
+  ) : (
+    renderElement({ children, tabIndex, className, component: Component, ...props })
+  );
+};
+
+export { Focusable };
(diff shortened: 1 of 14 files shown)


packages/react-core/src/components/Focusable/Focusable.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+
+export interface FocusableProps {
+  children?: string | React.ReactElement;
+  /** any additional classes to apply to the focusable element */
```
tlabaj: not.  This should be sentence case.

jgiardino:
When I test the examples of Focusable with tooltips in VO, the tooltip contents are not announced. 

Additionally for the disabled buttons, there are slight differences with how VO behaves depending on navigation method. 

Using the tab key, VO announces the element as a group instead of a button. It does add "you are currently on a button," but the semantics in this case are slightly different. For example, the disabled button is announced as:
> Disabled button text group main. [this is where the tooltip text would be announced if working as expected] You are currently on a button. This item is dimmed. 

Using the next element navigation or navigating by form control, the element is announced as a button and the button is announced as disabled. 

> Disabled button text dimmed button

Assuming the issue with the tooltip getting announced is fixed, there are still a couple of concerns I have with this method:
1. If different methods of navigation are focusing on different elements (i.e. the outer div vs the inner button), then the tooltip should be applied to all focusable elements to ensure that it's announced regardless of which element receives focus.
2. How does this behave in other screen readers? VO does eventually announce that it's a button, but is that the case with all screen readers? It's really important that the semantics of elements are preserved. If an element is a button, it should always be announced as a button. Those semantics should not change simply because it's disabled.  user is expecting a button element and comes across that element at a time when it's disabled, it , and the element should still be announced as a button even when disabled. 

I still favor the original implementation of adding the prop directly on the element to state that it's disabled but focusable. It avoids the issues mentioned above, since we're not adding addltional wrapping elements that alter the semantics of the normal state of the element. It also seemed like a much cleaner solution, both in implementation, but also when using it (as a consumer, I should be able to just add a prop to say that something is disabled).

mcoker:
> Could the default div tag use pf-u-display-inline-block

I'm not sure if you would need to import the entire display utility CSS (which is 3.7k) but an alternative could be to use a `<span>`.

----

This looks pretty cool! My main concern is supporting layouts that depend on a `.parent` > `.child` relationship, where the child is disabled. Adding an extra element around `.child` will break those layouts. In some cases we may be able to refactor the CSS to support an extra element being added between the parent and child, but not all. For example, something like this https://codepen.io/mcoker/pen/ZEGvyWw

I included in there an example of `display: contents`, which is great at resolving the layout issue where an element is added between the `.parent` > `.child`, but that also removes the `tabindex`. 

Side note, this may be something that is resolved in later implementations of `display: contents` as the current implementation by browsers has [a11y issues](https://github.com/w3c/csswg-drafts/issues/3040) due to incorrectly implementing [the spec](https://drafts.csswg.org/css-display/#valdef-display-contents). Namely removing semantics and including the ability to tab to an element.

seanforyou23:
Thanks for the feedback! Wanted to address a couple of things while we collect more comments/feedback.

>When I test the examples of Focusable with tooltips in VO, the tooltip contents are not announced.

I noticed they are not announced when focusing with VO cursor (reader focus), however if you tab onto the element with the standard tab key, the tooltip content does get announced - however, it's announced later after the reader has finished reading other contextual aspects of the current view. So (for this implementation strategy) I don't think it's as much a matter of getting the announcements "fixed" as it is just realizing when we should and shouldn't expect the announcements to take place. Seems like descriptions/tooltips are generally only announced on TAB focus, not screen reader focus.

I think exploring the impact/differences of screen reader focus vs keybaord focus is in play here, and adjusting our expectations around how these things should function can make the path forward a little more clear and flexible.

I'm starting to see how both implementations we're comparing here can be useful as tools in the toolset to build accessible experiences, and maybe there's less value in forcing to pick one as the "correct" path forward. IMO the need here is becoming more about offering tools to enable developers to build good AX rather than doing it for them or constructing tools in such a way that "prevents" them from composing solutions in a way that may be inaccessible.

>>Could the default div tag use pf-u-display-inline-block

>I'm not sure if you would need to import the entire display utility CSS (which is 3.7k) but an alternative could be to use a `<span>`.

Yes would need to import the display utilities, the size difference I think is nominal though. Using a `span` isn't a good default because certain types of markup isn't valid as children

Screen Shot 2020-03-10 at 11 08 56 AM

PR #3856, opened 2020-03-02
added console warning to Drawer beta component

evwilkin:
**What**: Towards #3624 - this PR adds a console warning for the Drawer beta component, similar to existing implementations in the DataToolbar and SimpleList components.


SergioEscobales:
Great !!! this add it's awesome!! thanks for this...

PR #3865, opened 2020-03-03
feat(FileUpload) Add new beta File Upload component

### LINKED ISSUE #3735: PF4: Add  File Upload Component
Introduce File Upload component according to the UI design here: https://marvelapp.com/de0ggah

This component will be introduced for uploading text files only. The user can upload a file by browsing to it, pasting its contents, or dragging and dropping it.

This is related to core issue https://github.com/patternfly/patternfly-next/issues/2624

mturley:
**What**: Closes #3735 

Preview: https://patternfly-react-pr-3865.surge.sh/documentation/react/components/fileupload

**Additional issues**: Based on core implementation from https://github.com/patternfly/patternfly-next/pull/2681, https://github.com/patternfly/patternfly-next/pull/2731 and https://github.com/patternfly/patternfly-next/pull/2776

### DIFF
--- a/packages/react-core/src/components/FileUpload/FileUpload.tsx
+++ b/packages/react-core/src/components/FileUpload/FileUpload.tsx
@@ -1,0 +1,150 @@
+import * as React from 'react';
+import Dropzone, { DropzoneProps, DropFileEventHandler } from 'react-dropzone';
+import { Omit } from '../../helpers';
+import { FileUploadField, FileUploadFieldProps } from './FileUploadField';
+import { readFile, fileReaderType } from '../../helpers/fileUtils';
+
+export interface FileUploadProps
+  extends Omit<
+    FileUploadFieldProps,
+    'children' | 'onBrowseButtonClick' | 'onClearButtonClick' | 'isDragActive' | 'containerRef'
+  > {
+  /** Unique id for the TextArea, also used to generate ids for accessible labels. */
+  id: string;
+  /** What type of file. Determines what is is passed to `onChange` and expected by `value`
+   * (a string for 'text' and 'dataURL', or a File object otherwise. */
+  type?: 'text' | 'dataURL';
+  /** Value of the file's contents
+   * (string if text file, File object otherwise) */
+  value?: string | File;
+  /** Value to be shown in the read-only filename field. */
+  filename?: string;
+  /** A callback for when the file contents change. */
+  onChange?: (
+    value: string | File,
+    filename: string,
+    event:
+      | React.DragEvent<HTMLElement> // User dragged/dropped a file
+      | React.ChangeEvent<HTMLTextAreaElement> // User typed in the TextArea
+      | React.MouseEvent<HTMLButtonElement, MouseEvent> // User clicked Clear button
+  ) => void;
+  /** Additional classes added to the FileUpload container element. */
+  className?: string;
+  /** Flag to show if the field is disabled. */
+  isDisabled?: boolean;
+  /** Flag to show if the field is read only. */
+  isReadOnly?: boolean;
+  /** Flag to show if a file is being loaded. */
+  isLoading?: boolean;
+  /** Aria-valuetext for the loading spinner */
+  spinnerAriaValueText?: string;
+  /** Flag to show if the field is required. */
+  isRequired?: boolean;
+  /* Value to indicate if the field is modified to show that validation state.
+   * If set to success, field will be modified to indicate valid state.
+   * If set to error,  field will be modified to indicate error state.
+   */
+  validated?: 'success' | 'error' | 'default';
+  /** Aria-label for the TextArea. */
+  'aria-label'?: string;
+  /** Placeholder string to display in the empty filename field */
+  filenamePlaceholder?: string;
+  /** Aria-label for the read-only filename field */
+  filenameAriaLabel?: string;
+  /** Text for the Browse button */
+  browseButtonText?: string;
+  /** Text for the Clear button */
+  clearButtonText?: string;
+  /** Flag to hide the built-in preview of the file (where available).
+   * If true, you can use children to render an alternate preview. */
+  hideDefaultPreview?: boolean;
+  /** Flag to allow editing of a text file's contents after it is selected from disk */
+  allowEditingUploadedText?: boolean;
+  /** Additional children to render after (or instead of) the file preview. */
+  children?: React.ReactNode;
+
+  // Props available in FileUpload but not FileUploadField:
+
+  /** A callback for when a selected file starts loading */
+  onReadStarted?: (fileHandle: File) => void;
+  /** A callback for when a selected file finishes loading */
+  onReadFinished?: (fileHandle: File) => void;
+  /** A callback for when the FileReader API fails */
+  onReadFailed?: (error: DOMException, fileHandle: File) => void;
+  /** Optional extra props to customize react-dropzone. */
+  dropzoneProps?: DropzoneProps;
+}
+
+export const FileUpload: React.FunctionComponent<FileUploadProps> = ({
+  id,
+  type,
+  value = type === fileReaderType.text || type === fileReaderType.dataURL ? '' : null,
+  filename = '',
+  children = null,
+  onChange = () => {},
+  onReadStarted = () => {},
+  onReadFinished = () => {},
+  onReadFailed = () => {},
+  dropzoneProps = {},
+  ...props
+}: FileUploadProps) => {
+  const onDropAccepted: DropFileEventHandler = (acceptedFiles: File[], event: React.DragEvent<HTMLElement>) => {
+    if (acceptedFiles.length > 0) {
+      const fileHandle = acceptedFiles[0];
+      if (type === fileReaderType.text || type === fileReaderType.dataURL) {
+        onChange('', fileHandle.name, event); // Show the filename while reading
+        onReadStarted(fileHandle);
+        readFile(fileHandle, type as fileReaderType)
+          .then(data => {
+            onReadFinished(fileHandle);
+            onChange(data as string, fileHandle.name, event);
+          })
+          .catch((error: DOMException) => {
+            onReadFailed(error, fileHandle);
+            onReadFinished(fileHandle);
+            onChange('', '', event); // Clear the filename field on a failure
+          });
+      } else {
+        onChange(fileHandle, fileHandle.name, event);
+      }
+    }
+    dropzoneProps.onDropAccepted && dropzoneProps.onDropAccepted(acceptedFiles, event);
+  };
+
+  const onDropRejected: DropFileEventHandler = (rejectedFiles: File[], event: React.DragEvent<HTMLElement>) => {
+    if (rejectedFiles.length > 0) {
+      onChange('', rejectedFiles[0].name, event);
+    }
+    dropzoneProps.onDropRejected && dropzoneProps.onDropRejected(rejectedFiles, event);
+  };
+
+  const onClearButtonClick = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    onChange('', '', event);
+  };
+
+  return (
+    <Dropzone multiple={false} {...dropzoneProps} onDropAccepted={onDropAccepted} onDropRejected={onDropRejected}>
+      {({ getRootProps, getInputProps, isDragActive, open }) => (
+        <FileUploadField
+          {...getRootProps({
+            ...props,
+            refKey: 'containerRef',
+            onClick: event => event.preventDefault() // Prevents clicking TextArea from opening file dialog
+          })}
+          tabIndex={null} // Omit the unwanted tabIndex from react-dropzone's getRootProps
+          id={id}
+          type={type}
+          filename={filename}
+          value={value}
+          onChange={onChange}
+          isDragActive={isDragActive}
+          onBrowseButtonClick={open}
+          onClearButtonClick={onClearButtonClick}
+        >
+          <input {...getInputProps()} /* hidden, necessary for react-dropzone */ />
+          {children}
+        </FileUploadField>
+      )}
+    </Dropzone>
+  );
+};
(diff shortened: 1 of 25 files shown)


mturley:
@tlabaj I still have a few small TODO items here, but it people are free I could use some early reviews since we want this merged by Friday. All that's left to do is add the loading spinner case and tests, the component and docs are mostly done.

evwilkin:
@mturley this looks good to me so far, pending your remaining updates - thanks for your work on this!

packages/react-core/src/components/FileUpload/FileUploadField.tsx
```diff
@@ -0,0 +1,163 @@
...
+  /** Additional children to render after (or instead of) the TextArea. */
+  children?: React.ReactNode;
+  /** A callback for when the Browse button is clicked. */
+  onBrowseButtonClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+  /** A callback for when the Clear button is clicked. */
+  onClearButtonClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+  /** Flag to show if a file is being dragged over the field */
+  isDragActive?: boolean;
+  /** A reference object to attach to the FileUploadField container element. */
+  containerRef?: React.Ref<any>;
```
seanforyou23: Are we able to type this? `HTMLDivElement`?

packages/react-core/src/components/FileUpload/FileUploadField.tsx
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export const FileUploadField: React.FunctionComponent<FileUploadFieldProps> = ({
+  id,
+  value = '',
+  filename = '',
+  onChange = (): any => undefined,
```
seanforyou23: Are these default functions necessary? If not can we remove them, and if they are - can we type them?

packages/react-core/src/components/FileUpload/FileUploadField.tsx
```diff
@@ -0,0 +1,163 @@
...
+  isRequired = false,
+  isDragActive = false,
+  validated = 'default' as 'success' | 'error' | 'default',
+  'aria-label': ariaLabel = 'File upload',
+  filenamePlaceholder = 'Drag a file here or browse to upload',
+  filenameAriaLabel = filename ? 'Read only filename' : filenamePlaceholder,
+  browseButtonText = 'Browse...',
+  clearButtonText = 'Clear',
+  clearButtonDisabled = !filename && !value,
+  containerRef = null as React.Ref<any>,
```
seanforyou23: Are we able to type this? `HTMLDivElement`?

seanforyou23:
Would using a [`FileReader`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) help with the blocking-ui behavior?

mturley:
Thanks for the review comments @seanforyou23. I'll fix up the types as you suggested.

As far as blocking file types that are not supported, I was thinking I would leave that up to the consumer (via dropzoneProps, see my second example), but maybe there's a sensible default I can set there. I didn't want to make an assumption here since the react-dropzone docs point out that MIME type determination is not reliable across platforms, and I don't see an easy way to reliably identify if a file is binary or not via FileReader or the File object.

https://react-dropzone.js.org/#!/Dropzone

> accept -- Set accepted file types. See https://github.com/okonet/attr-accept for more information. Keep in mind that mime type determination is not reliable across platforms. CSV files, for example, are reported as text/plain under macOS but as application/vnd.ms-excel under Windows. In some cases there might not be a mime type set at all. See: https://github.com/react-dropzone/react-dropzone/issues/276

I'm still not sure how to solve the blocking-ui behavior, I'm already using a FileReader. I'll keep looking into that for now, but if I don't find a solution today I wonder if it would be acceptable to merge the initial version of this component with this issue and address it in a follow-up. For most small text files, the load happens instantly.

packages/react-core/src/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,151 @@
...
+beta: true
+---
+
+import { FileUpload, Form, FormGroup, FileUploadField, Checkbox } from '@patternfly/react-core';
+
+## Examples
+
+The basic `FileUpload` component can handle simple text files via browse or drag-and-drop, loading them into memory and passing their contents as a string to an `onChange` prop.
+
+```js title=Simple-text-file isBeta
```
seanforyou23: I think the `isBeta` flag on the individual examples is only needed if the main component/feature is not beta. Since the entire FileUpload is currently beta, we can probably omit these in favor of the main beta flag at the top.

packages/react-core/src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,128 @@
...
+  dropzoneProps?: DropzoneProps;
+}
+
+// TODO handle an optional message for errors without using FieldGroup
+
+export const FileUpload: React.FunctionComponent<FileUploadProps> = ({
+  id,
+  value = '',
+  filename = '',
+  onChange = (): any => undefined,
```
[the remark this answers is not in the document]
mturley: As discussed, just noops for optional callbacks so I don't have to check that they are defined on each call. I don't think the typing is necessary on them at all, just replaced them all with `() => {}`. Let me know if you think we should use something better there.

mturley:
Ok @seanforyou23 I'm happy with this as it is now, we just need some tests and some more reviews! Feel free to jump in if you start before I do tomorrow. I'll be offline between about 1:30-3:30 and then I'll be around for the last hour or two before code freeze.

mturley:
One thing I do want to note is that the "file type and size summary" I added to support non-text files is something I made up in response to feedback, and we will probably need to go back through the design and core implementation process to support a better-looking version of that.

christiemolloy:
I think we should show an editable and non-editable version of this, and label them

Screen Shot 2020-03-06 at 2 32 06 PM

Do you this focus should just go to the file upload component and not the text? Would probably be a change in Core, but wondering what people think

Screen Shot 2020-03-06 at 2 34 19 PM

What is the purpose of adding children? When I check the children box , the border is cut off by the child
Screen Shot 2020-03-06 at 2 35 53 PM

is there a way to restrict this so that if someone chooses loading they have to select showPreview?

Screen Shot 2020-03-06 at 2 37 24 PM

seanforyou23:
The recommended a11y practice is to send focus back to whatever element had it before the file selector process started. I think this would mean the button in most cases, but it does make me think about the case of dragging onto the dropzone area. Even though the component container real estate is the droppable area, it doesn't receive focus (nor does it need to) so I'd still think to send focus to the browse button after the drop operation also. It seems to be working as intended when you use the browser button, but after dropping currently focus seems to be lost to the page.

@christiemolloy I think `children` was intended for consumers to be able to have a custom preview area if they wanted some special formatting. Not sure about the border thingie.

mturley:
@christiemolloy I think maybe I should remove that last example if it's causing confusion. Also I changed the `showPreview` prop to a `hideDefaultPreview` which defaults to false in my last commit, so that might help with the confusion there a little.

children is intended to be used only if hideDefaultPreview is also set, since they are for a custom preview. That border will not be cut off if it's used this way. So maybe if children are passed, the preview should be automatically hidden. (but it's still useful to be able to hide the preview entirely).

mturley:
Per discussion with Titani, I removed the default preview for a non-text file since it needs design and core consideration. Here's what it looked like for posterity, after this is merged I'll open another WIP PR to play around with it:

Screenshot 2020-03-06 13 52 13

mturley:
Thanks to @seanforyou23 for helping me out here with tests and helping me with some decisions!

mcoker:
@mturley this looks great!

I'm curious why we need the `tabindex` on the outer `.pf-c-file-upload` element?

Also I think you're on an older version of core? I'm not seeing the blue border on drag/hover.

I wonder if we should limit the drag/hover styles to the form elements, and not apply them over any messaging underneath the form elements. wdty @mcarrano? 

Screen Shot 2020-03-06 at 5 30 10 PM

FWIW, I used a 5mb jpg image to upload to the "simple text file" example, and my browser froze and was unresponsive for 3 minutes before updating the file preview. Not sure if there's anything we could do about that.

mturley:
@christiemolloy:

> I think we should show an editable and non-editable version of this, and label them

@mcarrano:

> The only thing I was not sure about was whether there was a way to optionally edit the text uploaded into the text area. Looks like that is currently read-only in the example, but if I recall our conversation, we wanted to make editing optional.

I can add a prop like `allowEditingUploadedFile` and show an example with that, would that meet both your needs? Sorry I overlooked that.

@mcoker:

> I'm curious why we need the `tabindex` on the outer `.pf-c-file-upload` element?

I was just duplicating what I saw in core, I'm not sure exactly 😆 but I can do whatever we need there.

> Also I think you're on an older version of core? I'm not seeing the blue border on drag/hover.

You're right! I need to bump that version, sorry.

> FWIW, I used a 5mb jpg image to upload to the "simple text file" example, and my browser froze and was unresponsive for 3 minutes before updating the file preview. Not sure if there's anything we could do about that.

Yeah, that's unfortunately what happens when you take that much data and shove it into a JS variable in React state. We could maybe set a sensible default for the maximum size allowed for that basic text file case, and advise the consumer not to use `type="text"` if they need larger files. If you don't use `type="text"`, it doesn't try to read the file at all, it just passes a handle to the consumer so they can do what they need to to read the file themselves. See my "custom file preview" example, you can drop a large file there and it'll just show the size without doing anything else.

mcoker:
> I was just duplicating what I saw in core, I'm not sure exactly 😆 but I can do whatever we need there.

@mturley hmm, I'm looking on [master](https://pf4.patternfly.org/documentation/core/components/fileupload) and don't see `tabindex` defined on any of the file upload component elements. Maybe it was there at some point and removed?

mturley:
@mcoker that's probably the case. I'll bump the core version and look over it all to make sure I'm up to date.

seanforyou23:
I think generally if `tabindex` is present we need to also ensure there's an accessible name for that element. Otherwise, as it stands currently the extra ways we can interact with the container don't get exposed to consumers using AT, and we have a mismatch between visual, non-visual, and keyboard experience.

That said, having two focusable elements to facilitate the same operation (for **each** file upload component on the page) is redundant and exposing this in an intuitive way to SRU will be unnecessarily involved without any gain, but definitely more clutter. The names need to be unique, and there are already many elements to wade through on any given web page these days so if we can have less nodes that are better described it will be better than having overwhelming number of options with names close enough to be confusing. Not worth the extra cognitive load it brings. 

We could either: Add an aria-label to the container that has tabindex or, we could remove tabindex. My vote is to go with the latter and only use the container for drag/drop purposes - then nail making the component browse/clear btn names unique among many of a same page. That would make for the best accessible experience in my opinion. I hope it helps, lemme know what you think!

mturley:
@mcoker I'm not actually seeing a `tabindex` in my source, where are you seeing that?

mturley:
Ah, I think the tabindex is being applied automatically by the library I'm using for drag-and-drop (react-dropzone). I can override that and prevent it

mcarrano:
> I wonder if we should limit the drag/hover styles to the form elements, and not apply them over any messaging underneath the form elements. wdty @mcarrano?

Good catch @mcoker . I don't think this is a big deal, but if we can restrict the styles to form elements, that would be preferrable.

mturley:
@mcarrano @mcoker In the core examples, the `.pf-c-file-upload__message` element is inside the `.pf-c-file-upload` container which gets the `.pf-m-drag-hover` class applied to it. For the message to be outside the drag/hover box, we'll need to change that in core.

The other option is, we can remove the styles for error/info messaging from FileUpload entirely, since the user can get the same result by wrapping the whole thing in a FormGroup component with its own `helperText`/`helperTextInvalid` props. Maybe that's what the examples should use instead?

mcoker:
> remove the styles for error/info messaging from FileUpload entirely, since the user can get the same result by wrapping the whole thing in a FormGroup component with its own helperText/helperTextInvalid props.

That sounds great to me if you think that makes sense from the developer's perspective.

mturley:
@seanforyou23 :

> We could either: Add an aria-label to the container that has tabindex or, we could remove tabindex. My vote is to go with the latter and only use the container for drag/drop purposes - then nail making the component browse/clear btn names unique among many of a same page.

I agree. I'm omitting the `tabindex` from react-dropzone's `getRootProps()` there. 
For making the browse/clear button names unique, what specific attributes do you think we need there? Based on the core implementation, I gave the filename field a unique id and aria-label, and set its aria-describedby to the Browse button which has a unique id (all generated based on the `id` prop passed by the consumer for the whole component). But the Clear button doesn't have its own unique id. Do we need to change that (both here and in core) or is it enough to have it be adjacent to the uniquely-identified filename stuff?

mturley:
@mcoker I think it does, it'll match how the developer needs to provide validation messages on other types of form fields. I'll make that change in my examples and remove the `message` prop, and we'll have to loop back and remove the error message HTML/CSS from the core examples and maybe use formgroup stuff there too. (cc @christiemolloy )

mturley:
@mcoker here's what it looks like using a FormGroup for the message instead. The dropzone area is now restricted to the form fields:

Screenshot 2020-03-09 11 32 30

It is worth noting that if the consumer passes `hideDefaultPreview` and `children` to render their own custom preview (e.g. an image thumbnail if they are handling images), that will still be within the dropzone (so they could drag/drop another image on top of the previous image). If they don't have any preview at all (`hideDefaultPreview` and no `children`) then the dropzone will be only the filename/browse/clear bar.

mturley:
Ok, I just rebased and resolved merge conflicts caused by my version bump to `@patternfly/patternfly@2.68.3`. I think I've addressed all the feedback so far, this should be ready to go!

mturley:
Yeah, 51 commits, I guess I should squash 😆 if anyone wants to back and look at the original history I'll leave it preserved here: https://github.com/patternfly/patternfly-react/compare/2aac12f9683d44cfa603236f68adaee8b082fcbf...mturley:3735-file-upload-component-presquash

mturley:
Left the pf version bump unsquashed in case we need to revert/replace it, since it was the thing causing conflicts.

@seanforyou23 @mcoker would you mind re-approving?

PR #3877, opened 2020-03-05
fix(aboutmodal): set a customize aria-label to the close button

### LINKED ISSUE #3855: AboutModal:  aria-label on the close button can not be overridden
The aria-label in the About Modal is hard coded.  The consumer should be able to override it.

boaz0:
**What**: Closes #3855

### DIFF
--- a/packages/react-core/src/components/AboutModal/AboutModalBoxCloseButton.tsx
+++ b/packages/react-core/src/components/AboutModal/AboutModalBoxCloseButton.tsx
@@ -9,15 +9,18 @@ export interface AboutModalBoxCloseButtonProps extends React.HTMLProps<HTMLDivEl
   className?: string;
   /** A callback for when the close button is clicked  */
   onClose?: () => void;
+  /** Set close button aria label */
+  'aria-label'?: string;
 }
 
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/AboutModal/AboutModalBoxCloseButton.tsx
```diff
@@ -9,15 +9,18 @@ export interface AboutModalBoxCloseButtonProps extends React.HTMLProps<HTMLDivEl
   className?: string;
   /** A callback for when the close button is clicked  */
   onClose?: () => void;
+  /** set aria label on the close button */
```
tlabaj: Can you update this to be sentence case please.

packages/react-core/src/components/AboutModal/AboutModalContainer.tsx
```diff
@@ -37,6 +37,8 @@ export interface AboutModalContainerProps extends React.HTMLProps<HTMLDivElement
   ariaLabelledbyId: string;
   /** id to use for About Modal Box aria described by  */
   ariaDescribedById: string;
+  /** set aria label to the close button */
```
tlabaj: Sentence case here too please.

PR #3879, opened 2020-03-05
fix react-docs version

redallen:
**What**: Publishing on master is broken because we published react-docs from our `v4` branch. This should fix it!

**Additional issues**:


evwilkin:
Build failing

PR #3880, opened 2020-03-05
fix (datatoolbar):  Added support for key with categoryName.

### LINKED ISSUE #3552: DataToolbarFilter: categoryName label required to delete chips
The `categoryName` prop of `DataToolbarFilter` is used to display the category name in the chip. In the example below, the "Project" category (i.e., chip) is a localizable string.

Screen Shot 2020-01-22 at 8 13 49 PM

When the chip is deleted, the same string is provided as the `type` variable to callback.

Screen Shot 2020-01-22 at 8 15 31 PM

In the filterable table demo below, the `type` is converted using `toLowerCase()`. However, considering that  this is a localizable label, we cannot use that approach to store filter values an array.

Screen Shot 2020-01-22 at 8 26 54 PM

I don't want the localizable "Project" label here, but some ID (e.g., 'project'). (All my filter and menu options are based on IDs.) Perhaps a `categoryID` prop would be helpful here?

dlabaj:
Updated code to allow for a key to be used with categoryName.  

TODO:
* Working on updating the demo wanted to get some eyes on the fix though.

**What**: Closes #3552

### DIFF
--- a/packages/react-core/src/components/DataToolbar/DataToolbarFilter.tsx
+++ b/packages/react-core/src/components/DataToolbar/DataToolbarFilter.tsx
@@ -51,12 +51,14 @@ export class DataToolbarFilter extends React.Component<DataToolbarFilterProps, D
   }
 
   componentDidMount() {
-    this.context.updateNumberFilters(this.props.categoryName, this.props.chips.length);
+    const { categoryName, chips } = this.props;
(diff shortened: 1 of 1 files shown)


kmcfaul:
Fix looks good, just need to update the examples/demo 👍

dlabaj:
Created an issue #3890 to add the demo.

Issue #3720, opened 2020-02-11
Change recommended React Router usage in Dropdowns


SpyTec:
Right now, in Dropdown docs we recommend using
```jsx
<DropdownItem key="link">
  <Link to={'/'}>Link</Link>
</DropdownItem>
```

Since this by default will create nested `<a>` we will get an error. On top of that there will also be applied global link styling.

One way we could fix this is by using hooks.
```jsx
const history = useHistory();
//...

<DropdownItem key="link" onClick={() => history.push("/")}>
  Link
</DropdownItem>
```

And for class-based usage we can recommend usage from this [StackOverflow answer](https://stackoverflow.com/a/42679052). That way we can use it the same way with `onClick` as shown above.

kmcfaul:
Adding the property `component="div"` will also prevent the nesting anchor issue. Our example will have to be updated to include that, at the very least.

SpyTec:
Then we still have the issue described on #3641, where the link will have link styling applied as well as only registering when you click the actual text instead of somewhere in the dropdown item

SpyTec:
Fixed in #3889

PR #3889, opened 2020-03-06
feat(Dropdown): Add new prop so Links are treated better

### LINKED ISSUE #3641: DropdownItem variation for use with interactive child element
https://www.patternfly.org/v4/documentation/react/components/dropdown#react-router-link-usage

```
<DropdownItem key="link" component="div">
  <Link to={'/'}>Link</Link>
</DropdownItem>
```

This results in markup that looks like

```
<div class="pf-c-dropdown__menu-item">
  <a href="#">Link</a>
</div>
```

and looks like the "Link action" link in this screenshot

Untitled

Our dropdown component isn't setup to support this structure and would require some enhancements to be able to support

* The touch area of the link filling the entire dropdown item
  * The link is currently just the text, and does not include any of the padding around it
* The proper hover/focus interaction
  * The background color should be an indication that you're hovering or focused on the interactive element (link/button). Currently the background changes when your mouse enters the padding around the link, and clicking the padding does nothing. The pointer should change to a cursor along with the background color, indicating that you're hovering the interactive element.
* Ensuring the link color is correct

After talking with @tlabaj and @kmcfaul about it today, we considered allowing for a prop to be used with `<DropdownItem>` that would be an indication that the interactive element is a child of `<DropdownItem>`, and the prop would effectively remove the `container` element and apply the dropdown item class `.pf-c-dropdown__menu-item` to the child instead.

rebeccaalpert:
When prop is used, class is applied to child so that the link is full-width and even clicking on the padding gets you where you need to go. Link color also remains black.

@kmcfaul, @tlabaj, and @mcoker - Do you want to use a different prop? How do you want to handle the documentation on how to use React router links? Should I incorporate those into the example here?

Fixes #3641 and #3720.


mcoker:
Hmm... I wonder if the prop being called "interactive" indicates that the `<DropdownItem>` is interactive, which isn't true in this case - the child is interactive. Would it be too specific to call this a `variant="routerLink"`?

Also this is currently rendering nested `<a>` elements, which is invalid. The outer `<a>` should just disappear. And the `<a>` inside should render like any `<DropdownItem key="link">Link</DropdownItem>` would.

Screen Shot 2020-03-09 at 10 12 13 AM

So the structure would render like:

```
<ul class="pf-c-dropdown__menu" role="menu">
  <li role="menuitem">
    <a tabindex="-1" aria-disabled="false" class="pf-c-dropdown__menu-item">Link</a></li><li role="menuitem"><button tabindex="-1" type="button" class="pf-c-dropdown__menu-item">Action</button></li><li role="menuitem"><a tabindex="-1" aria-disabled="true" class="pf-m-disabled pf-c-dropdown__menu-item">Disabled Link</a></li><li role="menuitem"><button tabindex="-1" disabled="" type="button" class="pf-m-disabled pf-c-dropdown__menu-item">Disabled Action</button></li><li role="separator"><div class="pf-c-dropdown__separator"></div></li><li role="menuitem"><a tabindex="-1" aria-disabled="false" class="pf-c-dropdown__menu-item">Separated Link</a></li><li role="menuitem"><button tabindex="-1" type="button" class="pf-c-dropdown__menu-item">Separated Action</button></li></ul>
```

rebeccaalpert:
@mcoker  - Thanks, fixing this now.

kmcfaul:
This should be able to close out #3720 as well, with the new example. Can you remove the old, out of date react-router examples?

rebeccaalpert:
@mcoker and @kmcfaul - Can you take another look? I removed the wrapper when the prop is provided (prop name is now "routerLink"), as well as the original examples. Let me know if you have other thoughts on this.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import classNames from 'classnames';
```
redallen: Can you not use the import below this line (`import { css } from '@patternfly/react-styles';`)? `classnames` technically isn't a dependency of react-core and this would be the first time we depend on it in react-core.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -125,6 +126,23 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
...
+  extendChildClass(
+    children: React.ReactNode,
+    itemClass: string,
+    ref: React.ReactNode,
+    id: string,
+    additionalProps: any
+  ) {
+    return React.Children.map(children as React.ReactElement<any>, child =>
+      React.cloneElement(child, {
+        className: classNames(child.props.className, itemClass),
```
redallen: Change to `css(child.props.className, itemClass)`

jschuler:
We need a general way to be more router-friendly, i.e. a router-friendly component that can be used instead of the default `<a href="">` type. We will have the same issue in breadcrumbitem, navitem, etc

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
jschuler: maybe it would make sense if this method is a helper method that could be utilized by other components?

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -15,7 +16,7 @@ export interface InternalDropdownItemProps extends React.HTMLProps<HTMLAnchorEle
   /** Indicates which component will be used as dropdown item */
   component?: React.ReactNode | string;
   /** Variant of the item. The 'icon' variant should use DropdownItemIcon to wrap contained icons or images. */
-  variant?: 'item' | 'icon';
+  variant?: 'item' | 'icon' | 'routerLink';
```
jschuler: could we make this a boolean instead? e.g. `isRouterLink: boolean` then we can adopt this pattern across components

rebeccaalpert: Are you okay with this @kmcfaul, @tlabaj, and @mcoker?

mcoker: Sounds good to me

rebeccaalpert: All set!

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
rebeccaalpert: All set! I think I subconsciously thought I was working on OpenShift.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
rebeccaalpert: Fixed. I think I subconsciously thought I was doing product work.

rebeccaalpert:
Addressed PR feedback by adjusting prop trigger and removed unneeded library.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
tlabaj: @jschuler @ralpert.  I think this is good now.  Do we need to open follow up issues for other components?

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -45,6 +45,8 @@ export interface InternalDropdownItemProps extends React.HTMLProps<HTMLAnchorEle
   customChild?: React.ReactNode;
   /** Flag indicating if hitting enter on an item also triggers an arrow down key press */
   enterTriggersArrowDown?: boolean;
+  /** Flag indicating if the item is a router link */
+  isRouterLink?: boolean;
```
dlabrecq: Could this prop just be isLink? Don't see "router" as adding value, considering you could add an HTML anchor instead.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
jschuler: @tlabaj yes I think we should track the other components as well so we can align them

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -207,6 +227,8 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
                     ...additionalProps,
                     className: css(classes, itemClass, variant === 'icon' && styles.modifiers.icon)
                   })
+                ) : isRouterLink ? (
```
jschuler: we are already checking if the component is a `isComponentReactElement`. Can we instead update that code path to work for router? Then we might not even need an extra prop

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
tlabaj: @jshuler or @rebeccaalpert can you please open an issue for that and listed the impacted components.

PR #3897, opened 2020-03-10
feat(Modal): move description to separate component

kmcfaul:


packages/react-core/src/components/Modal/ModalBoxDescription.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/ModalBox/modal-box';
+
+export interface ModalBoxDescriptionProps {
+  /** Content rendered inside the Footer */
```
tlabaj: Ca you update the description to say `description` instead of footer

PR #3903, opened 2020-03-11
feat(Alert): change default alert variant to default

### LINKED ISSUE #2646: PF4 Alert: Alert should default to "default" instead of "info"
1. The Alert should default to "default" instead of "info" alert
2. The default variant should be removed from the variant union since it has no matching modifier in core.  Instead if the consumer does not provide variant type, the default (meaning no modifier) alert will be used.

kmcfaul:
**What**: Closes #2646

@tlabaj `default` cannot be removed from the `variant` property union because it is now set to be the default value. Should the `variant` property accept a string instead of the specific "default" string, or should it remain accepting "default"?

## Breaking changes
1. Changes the default of the `variant` property to `default` instead of `info`. To maintain current default behavior, set the `variant` property to `info`.

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -36,7 +36,7 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
 }
 
 export const Alert: React.FunctionComponent<AlertProps & OUIAProps> = ({
-  variant = AlertVariant.info,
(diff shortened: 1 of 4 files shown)


tlabaj:
@kmcfaul it is fine to leave it in the Union.  I did something similar with the EmptyState.

tlabaj:
Looks good. maybe add a little bit more to the description like "in order to maintain current default behaviour, set the `variant` to `info`.

PR #3909, opened 2020-03-13
improvement(Table): implement indeterminate checkbox state

### LINKED ISSUE #3648: Add indeterminate checkbox to table header
When clicking a checkbox for a row in a `<Table>`, the header selectAll checkbox should display as indeterminate. Right now it's always unchecked until all rows have been marked.

1. Click checkbox on row
1. Expect header selectAll checkbox to change to indeterminate

See https://www.patternfly.org/v4/documentation/react/components/table#selectable

Enhancement

michpetrov:
**What**: Closes #3648

Added indeterminate state; this cannot be achieved simply through HTML attributes, hence the use of hooks. Left in the `allRowsSelected` prop in case it's used somewhere but it's made redundant with this change.

(Kinda new to React, please correct me if I went against conventions)

### DIFF
--- a/packages/react-table/src/components/Table/Table.tsx
+++ b/packages/react-table/src/components/Table/Table.tsx
@@ -31,6 +31,12 @@ export enum TableVariant {
   compact = 'compact'
 }
 
+export enum SelectedRowsAmount {
+  all = 'all',
+  some = 'some',
+  none = 'none'
+}
+
 export type RowEditType = 'save' | 'cancel' | 'edit';
 
 export interface RowErrors {
@@ -103,7 +109,7 @@ export interface IColumn {
     contentId?: string;
     dropdownPosition?: DropdownPosition;
     dropdownDirection?: DropdownDirection;
-    allRowsSelected?: boolean;
+    selectedRowsAmount?: SelectedRowsAmount;
   };
 }
 
@@ -333,11 +339,17 @@ export class Table extends React.Component<TableProps & OUIAProps, {}> {
 
   isSelected = (row: IRow) => row.selected === true;
 
-  areAllRowsSelected = (rows: IRow[]) => {
+  howManyRowsSelected = (rows: IRow[]) => {
     if (rows === undefined || rows.length === 0) {
-      return false;
+      return SelectedRowsAmount.none;
     }
-    return rows.every(row => this.isSelected(row) || (row.hasOwnProperty('parent') && !row.showSelect));
+    const selectableRows = rows.filter(row => !(row.hasOwnProperty('parent') && !row.showSelect));
+    const count = selectableRows.filter(row => this.isSelected(row)).length;
+    return count === 0
+      ? SelectedRowsAmount.none
+      : count === selectableRows.length
+      ? SelectedRowsAmount.all
+      : SelectedRowsAmount.some;
   };
 
   componentDidMount() {
@@ -394,7 +406,7 @@ export class Table extends React.Component<TableProps & OUIAProps, {}> {
       onSort,
       onSelect,
       canSelectAll,
-      allRowsSelected: onSelect ? this.areAllRowsSelected(rows as IRow[]) : false,
+      selectedRowsAmount: onSelect ? this.howManyRowsSelected(rows as IRow[]) : SelectedRowsAmount.none,
       actions,
       actionResolver,
       areActionsDisabled,
--- a/packages/react-table/src/components/Table/utils/decorators/selectable.tsx
+++ b/packages/react-table/src/components/Table/utils/decorators/selectable.tsx
@@ -10,7 +10,7 @@ export const selectable: ITransform = (
   { rowIndex, columnIndex, rowData, column, property }: IExtra
 ) => {
   const {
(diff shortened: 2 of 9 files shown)


michpetrov:
Okay, I think I fixed the tests; one question regarding `Selectable table with selected expandable row`, is the purpose of the test to simply check the presence of a prop, or to check that the `check-all` checkbox is indeed checked? Since those two are somewhat independent.

Should I add test for "some rows are selected"?

michpetrov:
@seanforyou23 I see `useRef` being used in [TextInput](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/TextInput/examples/TextInput.md) but I can refactor the component.

seanforyou23:
Thx for pointing this out! Checking into what we can do, keep you posted!

tlabaj:
@michpetrov  `useRef` is only used int the TextInput example to show how a PF consumer might be able to use it.  We do not yet utilize hooks in our PF library as @seanforyou23 mentioned above.

michpetrov:
Refactored the component, rebased the branch. There is still the one test failing, I don't know what it is testing so I don't know what to fix.

seanforyou23:
@michpetrov I'll take a look later today and see if I can help find out what gives with the test. Thanks for the refactoring!

seanforyou23:
Hi there @mpetrov - just a heads up I sent a [PR to your working branch](https://github.com/michpetrov/patternfly-react/pull/1) - seems to be related to this commit https://github.com/patternfly/patternfly-react/commit/bd3f49f526367c60de4cb8bbe05cb697c1ffba05


Looking into the issue I think I found some problems that while unrelated to this work item - is impacted by it. I'm going to see if I can isolate the unrelated problem ([described in the cypress test](https://github.com/michpetrov/patternfly-react/pull/1/files#diff-248d329a9763dc838fb7a393baf4ac6dR24-R39)) without these changes and maybe that will fix the problem without having to do anything further here. It's just a guess at this point though, so feel free to proceed investigating on your end with the demo I provided in the mean time. Hope it helps, keep you posted!

packages/react-table/src/components/Table/Table.tsx
```diff
@@ -30,6 +30,12 @@ export enum TableVariant {
   compact = 'compact'
 }
 
+export enum SelectedRowsAmount {
```
seanforyou23, line 33: Do you think we could assign strings to represent these enum values? I think using the enum indices is really elegant but makes it a little difficult to debug when comparing against the resulting markup. In other words I thought `data-rowsamount="1"` meant there was a single row selected instead of it referring to "some".

packages/react-table/src/components/Table/utils/decorators/selectable.tsx
```diff
@@ -42,8 +42,8 @@ export const selectable: ITransform = (
           'aria-labelledby': rowLabeledBy + rowIndex
         }
       : {
-          checked: allRowsSelected,
-          'aria-label': 'Select all rows'
+          'aria-label': 'Select all rows',
+          'data-rowsamount': selectedRowsAmount
```
seanforyou23, line 46: Could we use `data-selectedrowsamount` (or something similar) instead so the markup is more representative of what's going on behind the scenes?

packages/react-table/src/components/Table/utils/decorators/selectable.tsx
```diff
@@ -9,7 +9,7 @@ export const selectable: ITransform = (
   { rowIndex, columnIndex, rowData, column, property }: IExtra
 ) => {
   const {
-    extraParams: { onSelect, allRowsSelected, rowLabeledBy = 'simple-node' }
+    extraParams: { onSelect, allRowsSelected, selectedRowsAmount, rowLabeledBy = 'simple-node' }
```
seanforyou23, line 12: If we no longer need `allRowsSelected`, can we remove it?

packages/react-table/src/components/Table/utils/decorators/selectable.tsx
```diff
@@ -42,8 +42,8 @@ export const selectable: ITransform = (
           'aria-labelledby': rowLabeledBy + rowIndex
         }
       : {
-          checked: allRowsSelected,
```
seanforyou23, line 45: Seems like this may be what's causing that test to fail.

packages/react-table/src/components/Table/utils/decorators/selectable.tsx
michpetrov: I'll remove it, I kept it in in case something depended on it.

michpetrov:
@seanforyou23 I've changed the failing test and added your cypress tests

michpetrov:
@dlabrecq rebased

mcarrano:
I think that @dlabrecq makes some good points. This show work in the same way as the bulk selection from the toolbar. The only reason to still include this was for cases where you might want to use a table without a toolbar.

SpyTec:
I can see how the bulk selection could work if you have a toolbar, but at the same time it will increase the size of the toolbar, which might already be full enough. Unless the select all functionality in the Table will be deprecated I believe having a select-all checkbox that can be indeterminate makes sense

PR #3927, opened 2020-03-18
perf(ouia): make ouia functional, add namespace

### LINKED ISSUE #3317: <Button> rendered as <Component> in jest snapshots after updating to the latest PatternFly, breaking our tests

### LINKED ISSUE #3626: Add namespacing into data-ouia-component-type attributes

redallen:
**What**: Remove `withOUIA` HOC as discussed in #3622. In this PR it is replaced in favor of a functional:
```ts
function getOUIAProps(componentName: string, id: OuiaId | undefined, ouiaSafe: Boolean = true) {
  return {
    'data-ouia-component-type': `PF4/${componentName}`,
    'data-ouia-safe': ouiaSafe,
    'data-ouia-id': id === undefined ? uid++ : id
  };
}
```

Many snapshots now correctly identify the component since our old [withOUIAContext HOC](https://github.com/patternfly/patternfly-react/blob/83924aa5c081796f3e72667f23d90d27d96c5672/packages/react-core/src/components/withOuia/withOuia.tsx#L19) with a [displayName of `Component`](https://reactjs.org/docs/react-component.html#displayname) is gone.

Implementation in components is much cleaner:
```tsx
import { getOUIAProps, OUIAProps } from '../../helpers';

export const Alert: React.FunctionComponent<AlertProps & OUIAProps> = () => (
  <div {...getOUIAProps('Alert', ouiaId)} />
);
```

Spec: https://ouia.readthedocs.io/en/latest/README.html#ouia-component

**Additional issues**: Closes #3626 , Closes #3317


mturley:
@redallen does this need to be in v4 or can it go into master too? What's the breaking change?

mturley:
This issue was closed by `stale-bot`, but I think this PR resolves it (which is why it might be nice to backport to master): https://github.com/patternfly/patternfly-react/issues/3317

redallen:
There is no breaking change, but because of the +178,936/−235,682 of snapshot changes we considered this "breaking" enough to put on this branch since most consumer snapshots will change.

 See https://github.com/patternfly/patternfly-react/pull/3622#issuecomment-582557302

mturley:
Fair enough. I'll talk to folks downstream about upgrading to v4.

Issue #3734, opened 2020-02-12
Select: reorganize selections property


kmcfaul:
**Describe the issue. What is the expected and unexpected behavior?**
Divide `selections` property into two properties: `selection` (`string | SelectOptionObject`) for single select variants and `selections` (`(string | SelectOptionObject)[]`) for multiple select variants. This will clarify the expected data structure type for the different variants.

PR #3932, opened 2020-03-19
feat(Select): Convert selections prop to type

### LINKED ISSUE #3931: Feat: Select selections prop could be replaced with a type definition
`Select`'s [`selections`](https://github.com/patternfly/patternfly-react/blob/cf3559ee61fff0c1b679743a17e4521011fc7651/packages/react-core/src/components/Select/Select.tsx#L46) is able to take in multiple different prop types, namely `string | SelectOptionObject | (string | SelectOptionObject)[]`

If we were to change this to 

```ts
export type SelectionsType = string | SelectOptionObject | (string | SelectOptionObject)[];
export interface SelectProps {
  // ...
  /** Selected item for single select variant.  Array of selected items for multi select variants. */
  selections?: SelectionsType;
}
```
I would be able to simply define my selections variable in my code as follows which makes it much easier to read.
```ts
const selections: SelectionsType = "some selection"
<Select
  selections={selections}
/>
```

SpyTec:
**What**: Closes #3931

**Additional issues**:


tlabaj:
@kmcfaul I think this conflicts with the changes you are making for issue #3734

kmcfaul:
The `selections` prop will be broken into `selections` and a new `selection` prop, dividing out the single selection usage. So this prop will be simplified to two props each with two types (sting and SelectOptionObject, and their arrays).

SpyTec:
Gotcha, might be easier to handle then instead of current implementation. Thanks :)

PR #3943, opened 2020-03-20
fix(Empty State): renames example to 'Basic'

### LINKED ISSUE #3538: Empty state - move single example to end, rename to "basic"
This request came from design - move the "single" example to the end of the size variations for the empty state, so that we list in order: small -> large -> single. And can we update the example name to match what we have in core? Rename from "single" to "basic." 

Here are the core examples for reference - https://www.patternfly.org/v4/documentation/core/components/emptystate

Here's the PR where we made the change: https://github.com/patternfly/patternfly-next/pull/2499

And the issue with the request from design: https://github.com/patternfly/patternfly-next/issues/2469

jenny-s51:
**What**: Closes #3538. Changing to remain consistent with the core docs: [https://www.patternfly.org/v4/documentation/core/components/emptystate](https://www.patternfly.org/v4/documentation/core/components/emptystate)

per @mcoker


tlabaj:
@mcoker the issue you opened says to change the order.  When I look at this link, the "Basic' Example is still at the top.
https://www.patternfly.org/v4/documentation/core/components/emptystate

mcoker:
@tlabaj good catch, I was just talking to the core devs about that. It was changed recently for another reason, so I'm going to submit a PR shortly and revert it so the order is correct. So we can go ahead and make that update so that the order is small -> large -> x-large -> basic.

jenny-s51:
@mcoker @tlabaj will do -- thanks for the feedback on this!

tlabaj:
@jenny-s51 can you mov this to the v4 branch please.

PR #3945, opened 2020-03-20
feat(Select): change isExpanded to isOpen, split selections prop, rem…

### LINKED ISSUE #2327: Select Update Breaking Changes
These are the breaking changes we see in #2211 

- isExpanded - all components have isOpen however select has isExpanded is there any reason for it?

cc @karelhala

### LINKED ISSUE #3734: Select: reorganize selections property

kmcfaul:
**What**: Closes #3734, Closes #2327
Additionally, removes deprecated CheckboxSelect classes and cleans up code a bit.

## Breaking changes
1. Renames `isExpanded` property to `isOpen`.
2. Splits `selections` property into `selection` and `selections` for use in single versus multiple select variants respectively.
3. Removes deprecated `CheckboxSelect` and `CheckboxSelectOption` components. Please use the `variant = "checkbox"`.


kmcfaul:
@jenny-s51 Was able to update the demo to get the proper behavior. It wasn't set up to return the updated filtered items so the internal logic to add the 'no results option' wasn't executing.

jschuler:
On the custom filtering demo, after i select something and then click the clear button i get an error. maybe another case of the example being setup wrong? 

Screen Shot 2020-03-26 at 12 18 48 PM

kmcfaul:
@jschuler Fixed. I had added in a toString but it isn't needed after the prop split, and it was throwing the error on the null input.

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -446,11 +451,54 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
...
+    let variantChildren: any;
+    if (customContent) {
+      variantProps = {
+        selected: selections,
+        openedOnEnter,
+        isCustomContent: true
+      };
+      variantChildren = customContent;
+    } else {
+      switch (variant) {
```
redallen: This is so much more readable! Thanks! :star:

PR #3971, opened 2020-03-25
feat(rollup): create unified react-core dist

### LINKED ISSUE #3836: Create unified react dist file

redallen:
**What**:

- Add rollup and plugins
- Remove 3 circular dependencies created from shared contexts by extracting context to separate file
- Create `umd.html` as an example

Closes #3836

**Additional issues**: Still need to publish this to NPM to get the `/umd.html` link to work. Then the following tag will work:`<script src="https://unpkg.com/@patternfly/react-core@3/dist/umd/react-core.umd.js"></script>`


packages/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx
```diff
@@ -3,11 +3,10 @@ import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/AppLauncher/app-launcher';
 import { DropdownItem, DropdownItemProps } from '../Dropdown';
 import { ApplicationLauncherContent } from './ApplicationLauncherContent';
-import { ApplicationLauncherContext } from './ApplicationLauncher';
+import { ApplicationLauncherContext } from './ApplicationLauncherContext';
+import { ApplicationLauncherItemContext } from './ApplicationLauncherItemContext';
 import StarIcon from '@patternfly/react-icons/dist/js/icons/star-icon';
 
-export const ApplicationLauncherItemContext = React.createContext({ isExternal: false, icon: null });
```
dlabrecq: Wouldn't removing an export be a breaking change? 

I believe it's fine to add ApplicationLauncherItemContext.ts , but perhaps remove this in v4?

packages/react-core/rollup.config.js
```diff
@@ -0,0 +1,31 @@
...
+    file: `dist/umd/react-core${isProduction ? '.min' : ''}.js`,
+    format: 'umd',
+    name: 'PatternFlyReact',
+    globals: {
+      react: 'React',
+      'react-dom': 'ReactDOM',
+      'prop-types': 'PropTypes'
+    }
+  },
+  external: ['react', 'react-dom', 'prop-types'],
```
jschuler: Didn't you remove prop-types references in another PR?

redallen: Yes, but `prop-types` are still generated from the `typescript-to-proptypes` plugin. It seemed good to include these in our UMD bundle because there's no type support when using the `PatternflyReact` object.

packages/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx
redallen: Good catch! I've reexported it in the `index.ts` file so this isn't breaking.

PR #3978, opened 2020-03-26
fix(icons): fix types, upgrade typescript

### LINKED ISSUE #3977: Icons have incorrect types and are still JS
Currently react-icons is one of the few packages that hasn't undergone TS conversion. As such, some of its types are off for icon props. It should be refactored to be used with just `tsc` after source code generation.

redallen:
**What**: 
- Upgrade Typescript to same version as on master
- Fix 2 chart Typescript errors
- Refactor icons
  - Change FontAwesome/PatternFly/Custom icon files to have exports match aggregate exports
  - Change createIcon HOC to SVGIcon class component
  - Generate TSX source files with _named_ exports (no need for displayName)
  - Generate `index.ts` file in `icons` folder for easier gitignoring

Closes #3977

### @patternfly/react-icons
- Removed `OutlinedFontAwesomeLogoFullIcon`.


packages/react-charts/src/components/Chart/Chart.tsx
```diff
@@ -460,7 +459,7 @@ export const Chart: React.FunctionComponent<ChartProps> = ({
     >
       {children}
       {getLegend()}
-    </VictoryChart>
+    </VictoryChartWithContainerComponent>
```
dlabrecq: Is this a required change (i.e., do we need to cast to `any`)?

There are more components that use containerComponent prop, other than Chart and ChartAxis, so curious if this is blocking?

redallen: These are the only 2 that threw errors. I agree it'd be better to redefine `victory` types or update them upstream. We can open an issue for this.

PR #3989, opened 2020-03-30
chore(packages): bump @patternfly/patternfly to 2.71.0

redallen:
For release. Depends on https://github.com/patternfly/patternfly/pull/2900 first.

Also updates release notes.


redallen:
Ah, good catch @evwilkin . I did the same things in Core's release notes... I'll fix it here and I suppose we can live with it there for now.

PR #3991, opened 2020-03-30
chore(release): merge master for 2020.04

redallen:
Merge master for the past 2-3 weeks into v4


packages/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx
```diff
@@ -7,7 +7,6 @@ import { CardBody } from '@patternfly/react-core/dist/js/components/Card/CardBod
 import { CardFooter } from '@patternfly/react-core/dist/js/components/Card/CardFooter';
 import classNames from 'classnames';
 
-import { Omit } from '../../helpers/typeUtils';
 export interface CatalogTileProps extends Omit<React.HTMLProps<HTMLElement>, 'title'> {
```
rebeccaalpert: Do we really not need Omit anymore? It looks like we're still using it.

redallen: Typescript 3.5 introduced the [new Omit helper type.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type)

PR #4004, opened 2020-04-01
fix: PF4 Switch: label does not accept React.node

### LINKED ISSUE #3850: PF4 Switch: label does not accept React.node
If react node is passed as a label to Switch component, you will get an invalid prop type error, even though the label gets rendered:
```jsx
/**
* outputs error
*/
<Switch label={<span>Hello<span>} ... />


/**
* no error
*/
<Switch label="Hello" ... />
```

Is it possible to allow Reat.node as a valid prop type?

wojta:
**What**: Closes #3850

### DIFF
--- a/packages/react-core/src/components/Switch/__tests__/Switch.test.tsx
+++ b/packages/react-core/src/components/Switch/__tests__/Switch.test.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
 import { Switch } from '../Switch';
+import { assertPropTypes } from 'check-prop-types';
 
 const props = {
   onChange: jest.fn(),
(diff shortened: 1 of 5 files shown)


dgutride:
@tlabaj - a new dependency was added globally - do you have any thoughts about this?

packages/react-core/src/components/Switch/__tests__/Switch.test.tsx
```diff
@@ -67,3 +68,11 @@ test('switch passes value and event to onChange handler', () => {
   input.simulate('change', { target: { checked: true } });
   expect(props.onChange.mock.calls[0][0]).toBe(true);
 });
+
+test('switch accepts React.Node', () => {
+const view = mount(<Switch id="switch-react-node" label={<span>on</span>} labelOff={<span>off</span>}></Switch>);
+  expect(view.find('span#switch-react-node-on').find('span').first().text()).toBe('on');
+  expect(view.find('span#switch-react-node-off').find('span').first().text()).toBe('off');
+  assertPropTypes(Switch.propTypes, { label: <span>on</span> }, 'prop', Switch.name);
```
redallen, line 76: The reason this even works is because we are using a babel plugin to generate prop types. So `Switch.propTypes` does exist in `dist/esm/components/Switch/Switch.js`.

wojta: ok, removed and added demo

PR #4010, opened 2020-04-01
perf(packages): Remove babel in favor of tsc

### LINKED ISSUE #3972: Remove `babel` in favor of `tsc`
## What
When running `yarn start` contributors do not see `tsc` errors until submitting a PR or manually running `yarn build`, which is confusing. All Typescript files are compiled twice (once to JS and once to .d.ts) so compilation takes twice as long as necessary.

Our current [incremental build script](https://github.com/patternfly/patternfly-react/blob/9860938ac7641a5cc4278d91ceda71cc6a742159/scripts/incrementalBuild.js) is slow since it has to regenerate the source of `react-icons` each build. Also its terminal output confusing: when running `yarn start` we miss a crucial piece of information: that the dev server runs at http://localhost:8000.

### History
We took on technical debt when we started adopting `tsc` in our repo to emit types from Typescript (`.ts` or `.tsx`) files in #884. Each Typescript file in our repo is compiled twice, first with `babel`, and then a second time with `tsc` to create `.d.ts` files. Now that the majority of our code is Typescript, compilation takes twice as long as necessary.

Since we have a monorepo, we also require support for incremental builds. This way touching `react-core` doesn't require building all packages. #1817 originally implemented our custom incremental build script.

## Why
The benefits of this change would be:
- To save patternfly-react developers time on initial `yarn start`/`yarn build`
  - A full build with `tsc --build` takes just 30s on cold start with 2 threads
- To move building all packages at once to a standard system of `tsc --build`
  - `tsc --build` is better optimized to only rebuild when necessary and allows for cross-package typechecking
  - A rebuild with `tsc --build` takes just 6s when touching a single file in `react-core`
- Emit cleaner code in our `dist` that makes Webpack's life easier and performs better in the browser
  - `babel` currently is emitting code older than ES2015 bloated with helpers at the top
- Type-checking on save
- Ctrl+click on cross-project types with `.d.ts` sourcemap support
- Open up possibilities of writing tests and examples in TypeScript, improves support for `ts-jest` to type check our tests in the future

## How
I propose we move off of `lerna run build`, which runs `babel` 3 times in each of our packages to `tsc --build` to run a `tsc` incremental build once across all of our packages. The only thing we lose in this conversion is support for [`babel-plugin-typescript-to-proptypes`](https://www.npmjs.com/package/babel-plugin-typescript-to-proptypes). Per discussion on Slack in #patternfly-react, now that Typescript conversion is complete, our types are a good substitute for `prop-types` for now. Adopting something like [ts-proptypes-transformer](https://github.com/joelday/ts-proptypes-transformer) may be good future work.

### Epic
Issue | PRs | Completed | Description
--------|------|-----------|----------------
#4042 | #4048 | :heavy_check_mark: | Improve `eslint` speed on `.tsx?` files to something more reasonable than the over an hour it currently takes
#4043 | #4073 | :heavy_check_mark: | Convert react-inline-edit-extension to Typescript and drop `prop-types` from it
#4044 | #4058 (tokens) #4066 (styles) #4069 (icons) | :heavy_check_mark: | Make JS/TS generated in `react-tokens`, `react-styles`, and `react-icons` consistent with `tsc` (closes #4023 as well)
#4045 | #4050  | :heavy_check_mark: | Correctly type all function components
| | (If needed for #4046) Upgrade Jest
#4049 | #4051 | :heavy_check_mark: | Make imports compatible with tsc
#4046 | #4076 | :heavy_check_mark: | Move to `tsc --build` (closes #3763 as well)
#4047 | | | Add list of supported ES2015 browsers to README.md

Depends on #3977

redallen:
**What**: Closes #3972

- Use `tsc` instead of `babel` to build react-* packages.
  - Add `tslib` dependency and use it via `importHelpers: true` flag to avoid hundreds of spread/import helpers at the top of most files.
  - Remove `propTypes` as a `peerDependency` since the babel plugin [`typescript-to-proptypes`](https://www.npmjs.com/package/typescript-to-proptypes) can no longer be used.
  - Add newly generated `dist/**/*.map` and `dist/*.tsbuildinfo` files to each project's `.npmignore`.
- Remove [custom incremental build script](https://github.com/patternfly/patternfly-react/blob/9860938ac7641a5cc4278d91ceda71cc6a742159/scripts/incrementalBuild.js) in favor of `tsc`'s [project references](https://github.com/RyanCavanaugh/project-references-demo) which are well suited for monorepos. Cold build times for ESM and CJS are ~60s. Hot build times when touching react-core are ~6s.
- Drop per-package UMD output in favor of #3971 since there's lots of overhead and it's inadvisable to use UMD imports that make 50+ web requests. Bundles should be used instead.
- Change `generate` script to generate _entire_ `dist` folder for react-tokens, react-icons and react-styles without using babel. Refactor code generation to match `tsc`'s generated code.
  - Rename old `yarn generate` command (which ran `plop`) to `yarn template` or `yarn plop`.
    - Update plop's templates.
- Convert last JS project `react-inline-edit-extension` to TS so it can compile without enabling `tsc`'s `allowJS` flag.

**Additional issues**: 
- About a dozen function components (like TooltipArrow) were not defined as function components. I added types for these.
- `react-inline-edit-extension` had > 200 type errors when setting `allowJS: true`.  I refactored it to TS and remove `propTypes` from it to match the rest of our projects.
- The linter in CI took >10m because I added/changed too many *.tsx files. I've removed the eslint typescript parser's [`"project"`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser#parseroptionsproject) option for the time being which brings down the cold run time to ~50s and hot run time to 1.5s.

**Further work**:
- Replace `import * as React from 'react';` with `import React from 'react';`
- Use `ts-jest` instead of `babel-jest` to type check test files while compiling them. Combined with @jschuler 's auto generated snapshot tests this eliminates the need for `react-integration` to type check.
- Convert JS examples to TS and compile to JS using `tsc`.
- Add list of supported ES2015 browsers to README.md

### @patternfly/react-virtualized-extension
- Many types have been added and changed while converting this project to TS.

### ALL PACKAGES
- We no longer support UMD builds for individual packages. Use our new [react-core.umd.js bundle](https://unpkg.com/@patternfly/react-core@3/dist/umd/react-core.umd.js) instead.
- We no longer define `propTypes` for our components. Consider using our Typescript types under `dist/js` instead.


redallen:
@seanforyou23 It'd take a couple hours to split out the `inline-edit-extension` work and I'd have to get it to work with `babel` rather than `tsc`... but I'm happy to spend time doing that if getting it working with `babel` is valuable and then rebasing this PR on that.

There's a couple of merge commits in the history, so I can't easily squash them. Github will do that hard work upon merging, sorry for having to scroll to see the comments at the bottom here.

mturley:
Do we know if switching from `babel` to `tsc` will have any implications on the behavior of the compiled code? I have heard about subtle caveats, but I think they are mostly things `tsc` supports that `babel` doesn't. I can't find it now but someone told me about an issue where async/await code wasn't compiled exactly the same between them.

redallen:
@mturley Yes our compiled code is different. This PR targets ES2015 and use Typescript helpers and transforms to compile our code rather than Babel's. This allows for the use of arrow functions, although I'm not sure about async/await (but we don't have any usages of async/await in our exported codebase). Overall, I think our generated code quality has increased. [Here's a diff of `esm/components/Brand/Brand.js`](https://www.diffchecker.com/xE3yOqHg), and [another of `js/components/Brand/Brand.js`.](https://www.diffchecker.com/NGwMcYXt) I encourage testing out all the components in https://patternfly-react-pr-4010.surge.sh to make sure there aren't any obvious regressions!

Also now that we clearly target ES2015 we can list browsers + versions that we support. I'll add that to the future work section of the OP.

mturley:
Nice! Yeah, definitely an improvement.

seanforyou23:
> Drop per-package UMD output in favor of #3971 

This is the first time I've heard of this, have we checked with relevant teams who depend on this feature?

> It'd take a couple hours to split out the inline-edit-extension work and I'd have to get it to work with babel rather than tsc... but I'm happy to spend time doing that if getting it working with babel is valuable

I think you miss the point, it's not about getting the extension to work with babel but rather not sending a PR that's so large in volume that it becomes difficult to read and follow.

mturley:
@redallen do you mean you'd have to get the TS version of inline-edit-extension to work with babel in order to merge that before this PR? Could we instead leave its JS as-is for now but compiled by tsc (merge this PR with the `allowJS` flag set), and then have another PR to convert inline-edit-extension to TS and unset `allowJS`?

mturley:
Oh, I'm seeing this now:

> react-inline-edit-extension had > 200 type errors when setting allowJS: true

Maybe we can just have TS ignore that package entirely for this PR.

mturley:
@redallen when you were using `allowJs`, did you also have `checkJs` on? According to https://github.com/microsoft/TypeScript/issues/19573#issuecomment-401664604 we should be able to allow JS without emitting type errors within it.

Even if we can't do it that way, maybe we can use `// @ts-nocheck` in each of those files for this PR. https://github.com/microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files

packages/tsconfig.base.json
```diff
@@ -2,15 +2,24 @@
...
-    "module": "commonjs",
+    "target": "es2015",
+    "module": "es2015",
     "moduleResolution": "node",
+    "esModuleInterop": true,
     "skipLibCheck": true,
+    "importHelpers": true,
+    "composite": true,
+    "sourceMap": true,
+    "declarationMap": true,
```
seanforyou23, line 16: Do we need to also add `"declaration": true`? Or does the declarationDir setting in the individual package configs maybe take care of generating these?

redallen: By default `tsc` emits declarations. [This is the flag we were using before:](https://www.typescriptlang.org/docs/handbook/compiler-options.html)

Option | Type | Default | Description
-- | -- | -- | --
--noEmit | boolean | false | Do not emit outputs.

redallen:
I'm turn #3972 into an epic and break up this PR into smaller pieces that are easier to understand and review.

PR #4019, opened 2020-04-02
Fix/3223 integration ts

### LINKED ISSUE #3223: react-integration compiler options masks type issues

### LINKED ISSUE #3224: Some component demos don't meet "strictFunctionTypes" requirements
I noticed that when a consuming typescript app uses the `strictFunctionTypes` compiler option, following certain component examples causes errors to be thrown. Below is an example of what the error looks like, a typical type mismatch from what I can tell..

Screen Shot 2019-10-28 at 3 00 24 PM

These are the component demos in the react-integration package I see that are affected.

- [x] ContextSelectorDemo/ContextSelectorDemo.tsx
- [x] SelectDemo/SelectDemo.tsx
- [x] TableDemo/TableActionsDemo.tsx
- [x] TableDemo/TableCompactExpandableDemo.tsx
- [x] TableDemo/TableSortableForCompoundExpandableDemo.tsx
- [x] TabsDemo/TabsDemo.tsx
- [x] TabsDemo/TabsStringEventKeyDemo.tsx

### LINKED ISSUE #3230: Some component demos don't meet "strictNullChecks" requirements
I noticed that when a consuming typescript app uses the `strictNullChecks` compiler option, errors are thrown during build time. I'm seeing this issue in the following component demos in react-integration:

- [ ] AccordionDemo/AccordionDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSmallRightDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSmallStaticBottomDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSmallStaticDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationSmallStaticRightDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationStaticBottomDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationStaticDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationStaticInvertedRightDemo.tsx
- [ ] DonutUtilizationChartDemo/DonutUtilizationStaticRightDemo.tsx
- [ ] SelectDemo/SelectDemo.tsx
- [ ] TableDemo/TableActionsDemo.tsx
- [ ] TableDemo/TableSortableForCompoundExpandableDemo.tsx

It seems like a big list, but I think the problem for charts is the same one repeated in all of the chart demos. That said, I don't guess this will be a trivial fix.

Below is an example of how the error can manifest:

Screen Shot 2019-10-28 at 8 19 20 PM

### LINKED ISSUE #3231: Some component demos don't meet "noImplicitAny" requirements

evwilkin:
**What**:
- Towards #3223 
- Closes #3231 
- Towards #3230 
- Towards #3224 

This PR resolves typescript errors within `react-integration/demo-app-ts` related to `strictFunctionTypes` and `noImplicitAny` and enables both of these options in the `tsconfig`. 

This PR also makes progress towards errors related to `strictPropertyInitialization` and `strictNullChecks`, though both had to remain disabled pending further decisions around `null` conflicts with Victory charts types.

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
@@ -42,7 +42,7 @@ export class DonutUtilizationGreenStaticRightDemo extends React.Component<{}, { 
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: used }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
               legendData={[
                 { name: `Storage capacity: ${used}%` },
                 { name: 'Warning threshold at 60%' },
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
@@ -7,9 +7,9 @@ import {
 } from '@patternfly/react-charts';
 
 export class DonutUtilizationGreenStaticRightDemo extends React.Component<{}, { used: number }> {
-  interval: number;
+  interval: number = 0;
   constructor(props: {}) {
-    super(props as null);
+    super(props);
     this.state = {
       used: 0
     };
@@ -42,7 +42,7 @@ export class DonutUtilizationGreenStaticRightDemo extends React.Component<{}, { 
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: used }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
               legendData={[
                 { name: `Storage capacity: ${used}%` },
                 { name: 'Warning threshold at 60%' },
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
@@ -4,7 +4,7 @@ import { ChartDonutUtilization } from '@patternfly/react-charts';
 export class DonutUtilizationInvertedRightDemo extends React.Component<{}, { used: number; spacer: string }> {
   interval: any;
   constructor(props: {}) {
-    super(props as null);
+    super(props);
     this.state = {
       spacer: '',
       used: 100
@@ -36,7 +36,7 @@ export class DonutUtilizationInvertedRightDemo extends React.Component<{}, { use
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: used }}
             invert
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
             legendData={[{ name: `Storage capacity: ${spacer}${used}%` }, { name: 'Unused' }]}
             legendOrientation="vertical"
             subTitle="of 100 GBps"
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
@@ -36,7 +36,7 @@ export class DonutUtilizationInvertedRightDemo extends React.Component<{}, { use
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: used }}
             invert
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
             legendData={[{ name: `Storage capacity: ${spacer}${used}%` }, { name: 'Unused' }]}
             legendOrientation="vertical"
             subTitle="of 100 GBps"
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
@@ -16,7 +16,7 @@ export class DonutUtilizationSimpleBottomDemo extends React.Component {
             constrainToVisibleArea
             data={{ x: 'Storage capacity', y: 45 }}
             height={275}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
             legendData={[{ name: `Storage capacity: 45%` }, { name: 'Unused' }]}
             legendPosition="bottom"
             padding={{
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
@@ -16,7 +16,7 @@ export class DonutUtilizationSimpleBottomDemo extends React.Component {
             constrainToVisibleArea
             data={{ x: 'Storage capacity', y: 45 }}
             height={275}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
             legendData={[{ name: `Storage capacity: 45%` }, { name: 'Unused' }]}
             legendPosition="bottom"
             padding={{
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
@@ -12,7 +12,7 @@ export class DonutUtilizationSimpleDemo extends React.Component {
         <div style={{ backgroundColor: 'white', height: '230px', width: '500px' }}>
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: 75 }}
-            labels={({ datum }) => (datum.x ? `${datum.x} - ${datum.y}%` : '')}
+            labels={({ datum }) => (datum.x ? `${datum.x} - ${datum.y}%` : null)}
             subTitle="of 100 GBps"
             title="75%"
             width={500}
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
@@ -12,7 +12,7 @@ export class DonutUtilizationSimpleDemo extends React.Component {
         <div style={{ backgroundColor: 'white', height: '230px', width: '500px' }}>
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: 75 }}
-            labels={({ datum }) => (datum.x ? `${datum.x} - ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x} - ${datum.y}%` : '')}
             subTitle="of 100 GBps"
             title="75%"
             width={500}
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
@@ -2,10 +2,10 @@ import React from 'react';
 import { ChartDonutUtilization } from '@patternfly/react-charts';
 
 export class DonutUtilizationSimpleRightDemo extends React.Component<{}, { used: number; spacer: string }> {
-  interval: number;
+  interval: number = 0;
 
   constructor(props: {}) {
-    super(props as null);
+    super(props);
     this.state = {
       spacer: '',
       used: 0
@@ -36,7 +36,7 @@ export class DonutUtilizationSimpleRightDemo extends React.Component<{}, { used:
         <div style={{ backgroundColor: 'white', height: '230px', width: '435px' }}>
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: used }}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
             legendData={[{ name: `Storage capacity: ${spacer}${used}%` }, { name: 'Unused' }]}
             legendOrientation="vertical"
             subTitle="of 100 GBps"
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
@@ -36,7 +36,7 @@ export class DonutUtilizationSimpleRightDemo extends React.Component<{}, { used:
         <div style={{ backgroundColor: 'white', height: '230px', width: '435px' }}>
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: used }}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
             legendData={[{ name: `Storage capacity: ${spacer}${used}%` }, { name: 'Unused' }]}
             legendOrientation="vertical"
             subTitle="of 100 GBps"
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
@@ -2,9 +2,9 @@ import React from 'react';
 import { ChartDonutUtilization, ChartThemeColor, ChartThemeVariant } from '@patternfly/react-charts';
 
 export class DonutUtilizationSimpleRightGreenDemo extends React.Component<{}, { used: number; spacer: string }> {
-  interval: number;
+  interval: number = 0;
   constructor(props: {}) {
-    super(props as null);
+    super(props);
     this.state = {
       spacer: '',
       used: 0
@@ -39,7 +39,7 @@ export class DonutUtilizationSimpleRightGreenDemo extends React.Component<{}, { 
             constrainToVisibleArea
             data={{ x: 'Storage capacity', y: used }}
             height={300}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
             legendData={[{ name: `Storage capacity: ${spacer}${used}%` }, { name: 'Unused' }]}
             legendOrientation="vertical"
             legendPosition="bottom"
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
@@ -39,7 +39,7 @@ export class DonutUtilizationSimpleRightGreenDemo extends React.Component<{}, { 
             constrainToVisibleArea
             data={{ x: 'Storage capacity', y: used }}
             height={300}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
             legendData={[{ name: `Storage capacity: ${spacer}${used}%` }, { name: 'Unused' }]}
             legendOrientation="vertical"
             legendPosition="bottom"
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
@@ -12,7 +12,7 @@ export class DonutUtilizationSmallDemo extends React.Component {
           <ChartDonutUtilization
             data={{ x: 'Storage capacity', y: 75 }}
             height={150}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
             subTitle="of 100 GBps"
             title="75%"
             width={150}
--- a/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
@@ -12,7 +12,7 @@ export class DonutUtilizationSmallDemo extends React.Component {
           <ChartDonutUtilization
             data={{ x: 'Storage capacity', y: 75 }}
             height={150}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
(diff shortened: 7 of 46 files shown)


packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
```diff
@@ -42,7 +42,7 @@ export class DonutUtilizationGreenStaticRightDemo extends React.Component<{}, {
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: used }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 45: This is suppose to be null to prevent the tooltip from displaying in certain scenarios. 

It's possible that it could be undefined, but an empty string is not what we want here. An empty string will cause the tooltip to display an empty box.

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
```diff
@@ -36,7 +36,7 @@ export class DonutUtilizationInvertedRightDemo extends React.Component<{}, { use
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: used }}
             invert
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 39: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
```diff
@@ -16,7 +16,7 @@ export class DonutUtilizationSimpleBottomDemo extends React.Component {
             constrainToVisibleArea
             data={{ x: 'Storage capacity', y: 45 }}
             height={275}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 19: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
```diff
@@ -12,7 +12,7 @@ export class DonutUtilizationSimpleDemo extends React.Component {
         <div style={{ backgroundColor: 'white', height: '230px', width: '500px' }}>
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: 75 }}
-            labels={({ datum }) => (datum.x ? `${datum.x} - ${datum.y}%` : null)}
```
dlabrecq, line 15: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
```diff
@@ -36,7 +36,7 @@ export class DonutUtilizationSimpleRightDemo extends React.Component<{}, { used:
         <div style={{ backgroundColor: 'white', height: '230px', width: '435px' }}>
           <ChartDonutUtilization
             data={{ x: 'GBps capacity', y: used }}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 39: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
```diff
@@ -39,7 +39,7 @@ export class DonutUtilizationSimpleRightGreenDemo extends React.Component<{}, {
             constrainToVisibleArea
             data={{ x: 'Storage capacity', y: used }}
             height={300}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 42: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
```diff
@@ -12,7 +12,7 @@ export class DonutUtilizationSmallDemo extends React.Component {
           <ChartDonutUtilization
             data={{ x: 'Storage capacity', y: 75 }}
             height={150}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 15: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallRightDemo.tsx
```diff
@@ -36,7 +36,7 @@ export class DonutUtilizationSmallRightDemo extends React.Component<{}, { used:
           <ChartDonutUtilization
             data={{ x: 'Storage capacity', y: used }}
             height={150}
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 39: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallStaticDemo.tsx
```diff
@@ -11,14 +11,17 @@ export class DonutUtilizationSmallStaticDemo extends React.Component {
...
+              { x: 'Danger at 90%', y: 90 }
+            ]}
             height={175}
             labels={({ datum }) => (datum.x ? datum.x : null)}
             width={175}
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: 45 }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 24: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallStaticBottomDemo.tsx
```diff
@@ -28,7 +28,7 @@ export class DonutUtilizationSmallStaticBottomDemo extends React.Component {
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: 45 }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 31: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallStaticRightDemo.tsx
```diff
@@ -29,14 +29,17 @@ export class DonutUtilizationSmallStaticRightDemo extends React.Component<{}, {
...
+              { x: 'Danger at 90%', y: 90 }
+            ]}
             height={175}
             labels={({ datum }) => (datum.x ? datum.x : null)}
             width={425}
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: used }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 42: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationStaticDemo.tsx
```diff
@@ -11,12 +11,15 @@ export class DonutUtilizationStaticDemo extends React.Component {
...
+            data={[
+              { x: 'Warning at 60%', y: 60 },
+              { x: 'Danger at 90%', y: 90 }
+            ]}
             labels={({ datum }) => (datum.x ? datum.x : null)}
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: 45 }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 22: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationStaticInvertedRightDemo.tsx
```diff
@@ -41,7 +41,7 @@ export class DonutUtilizationStaticInvertedRightDemo extends React.Component<{},
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: used }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 44: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationStaticRightDemo.tsx
```diff
@@ -35,7 +35,7 @@ export class DonutUtilizationStaticRightDemo extends React.Component<{}, { used:
           >
             <ChartDonutUtilization
               data={{ x: 'Storage capacity', y: used }}
-              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+              labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : '')}
```
dlabrecq, line 38: This is suppose to be null

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
evwilkin: Got it - thanks for the feedback here, and in the related issues!  I'll update the type to allow for null and revert these back

evwilkin:
@dlabrecq it looks like each change requested is around the `labels` prop of the `ChartDonutUtilization` component, which (with the added `strictNullChecks`) allows returning only a string or an array of strings:

`labels?: string[] | ((data: any) => string);`

Because of this, an error is thrown if I revert the code back to return null in the examples in this PR:

```
labels={({ datum }) => (datum.x ? '${datum.x}: ${datum.y}%' : null)}
// type 'null' is not assignable to type 'string'
```

I wanted to get your feedback based on your experience with react-charts on whether I should go ahead and update this `labels?` definition in the `ChartDonutUtilization` component to allow the function to return `null` as follows:

`labels?: string[] | ((data: any) => string | null);`

...or use a workaround for these issues for now as follows (assuming this still registers `null` as needed), and log a separate issue towards react-charts to update the type outside of this PR:

`labels={({ datum }) => (datum.x ? '${datum.x}: ${datum.y}%' : ((null as unknown) as string))}`

Thanks for taking a look!

dlabrecq:
`null` is a valid for strings, it should not need to be cast. 

Does every prop need to accept null explicitly? I didn't realize strictNullChecks was added, can we omit it?

Like I said, `undefined` might also work, but `null` is valid

dlabrecq:
Like I said, undefined might also work, but null is valid. It should not need to be cast.

Does every prop need to define null and undefined explicitly? I didn't realize strictNullChecks was added, can we omit it?

evwilkin:
@seanforyou23 wanted to get your feedback on @dlabrecq's feedback above on `strictNullChecks` as it's directly related to the work in this PR towards #3230 .

From my understanding, it seems better to get it up and working across the board and explicitly define it as an option for the one component where it's breaking (`ChartDonutUtilization`, which should implicitly accept `null` as a return value) or type cast if needed, but I'm just looking at the scope of this PR and thought you may have a better understanding of implications on a bigger picture - thanks!

dlabrecq:
Is this just an issue with the demo? That is, it's the demo that's requiring the `strictNullChecks`, correct? In that case, go ahead an cast `null` or use `undefined`.

FYI, it may be more of an issue if we turn this on for react-charts itself. There are many places where null is valid. That is, these types come directly from @types/victory. Typically, if the types are not exact, it won't compile.

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
```diff
@@ -12,7 +12,7 @@ export class DonutUtilizationSimpleDemo extends React.Component {
...
+            labels={({ datum }) => (datum.x ? `${datum.x} - ${datum.y}%` : '')}
```
seanforyou23, line 15: Poking around and found that if we swap the type for [labels here](https://github.com/patternfly/patternfly-react/blob/v4/packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx#L326) to be `labels?: string[] | ((data: any) => string | null);` then users can specify `null` as a value without throwing errors and without showing the empty tooltip box.

WDYT? @dlabrecq would this cause any problems you're aware of?

dlabrecq: My concern is that it's not the only place where `null` and `undefined` are valid. Yes, we could change this one prop, but it doesn't address any other (pass-thru) Victory props. The demo is only touching a single point here -- we're not testing anything else.

Does this mean all strings cannot be null, unless we explicitly add null to the type? This would be contrary to the types defined by @types/victory. 

Curious what happens when we use VictoryPie directly, instead of the ChartPie component?

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/victory/index.d.ts#L1550

dlabrecq: My concern is that it's not the only place where `null` and `undefined` are valid. Yes, we could change this one prop, but it doesn't address any other (pass-thru) Victory props. The demo is only touching a single point here -- we're not testing anything else.

Does this mean all strings cannot be null, unless we explicitly add null to the type? This would be contrary to the types defined by @types/victory. 

Curious what we would do if using VictoryPie directly, instead of the underlying ChartPie component? It appears that VictoryPie would have the same issue with the demo's strict rules.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/victory/index.d.ts#L1550

seanforyou23: > Does this mean all strings cannot be null, unless we explicitly add null to the type?

This is a great question, and it's exactly the understanding that I was _starting_ to formulate, but wanted to verify. That does seem like what this setting is geared toward enforcing.

> This would be contrary to the types defined by @types/victory

Yea, and while we do this in other places (create our own abstraction on third party lib props), I guess we've treated Victory/charts stuff differently to this point to keep things simple/pass-through in terms of api and documentation.

> it doesn't address any other (pass-thru) Victory props

What's an example of this? If we need to support this, then we could add to the integration app an example that exercises this code path. Is it something that's easily illustrated?

dlabrecq: There are many Victory-based props that can accept null or undefined. Unfortunately, I do not know all the use-cases, but anything related to data. My point is that enabling strict mode may cause some breakage we haven't anticipated.

seanforyou23: I see, thanks for explaining. I knew it would be a potentially challenging thing to achieve. It does raise good discussion though. If the incoming Victory types don't work out of box in strict mode, seems like we have the options of either continuing to expose these props as-is and for PF to not support this specific strict mode for chart components, or adjust our types as necessary and possibly reduce our support of the Victory props to the subset that we feel is worth supporting and relevant to PF needs.

Victory has a large API surface area, so just brainstorming to see how we can best align it with what we're supporting on our end.

Is there another path forward I'm not thinking of? Otherwise, let me know what you think :)

dlabrecq: Omitting Victory props probably isn't a good option. Perhaps we should discuss offline, along with the infrastructure team?

evwilkin:
@dlabrecq @seanforyou23 reverted Victory conflicts and disabled `strictNullChecks`, which threw an error for me to also disable `strictPropertyInitialization` as it requires strictNullChecks.  I've updated the PR description to reflect these updates.

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationGreenStaticRightDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleBottomDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationInvertedRightDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSimpleRightGreenDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallRightDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallStaticDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallStaticBottomDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationSmallStaticRightDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationStaticDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationStaticInvertedRightDemo.tsx
evwilkin: Reverted this

packages/react-integration/demo-app-ts/src/components/demos/DonutUtilizationChartDemo/DonutUtilizationStaticRightDemo.tsx
evwilkin: Reverted this

Issue #4005, opened 2020-04-01
Cursor in table actions is no correct


karelhala:
**Describe the issue. What is the expected and unexpected behavior?**
When hovering over action item in table cursor is not correct, it shows `text` variant instead of `pointer`
**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
Any version, PF react demo page.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

* bug - kinda blocking, it's not user friendly to show this type of cursor

**What is your product and what release version are you targeting?**

* cloud services - the sooner the better

seanforyou23:
Hi @karelhala - having a little trouble reproducing this. Can you help me understand the environment a little better? What browser/platform combo are you seeing this on? I don't see it on the [docs site](https://www.patternfly.org/v4/documentation/react/components/table#simple-actions) or when running latest master locally. Thanks!

karelhala:
If you click on the 3 dots representing the action and hover over the `Some action` text you'll see differet cursor than what you can see in regular dropdown.

seanforyou23:
Ah, I see what you mean now. Thanks - taking a look!

PR #4021, opened 2020-04-03
(fix): Table actions cursor

seanforyou23:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4005

This PR updates the table actions dropdown to use `button` elements by default as we do in core. I noticed this was adjusted recently, would be good to verify this doesn't introduce any problems.

**Additional issues**: https://github.com/patternfly/patternfly-react/issues/3559

@SpyTec ping

### DIFF
--- a/packages/react-table/src/components/Table/ActionsColumn.tsx
+++ b/packages/react-table/src/components/Table/ActionsColumn.tsx
@@ -79,7 +79,7 @@ export class ActionsColumn extends React.Component<ActionsColumnProps, ActionsCo
               <DropdownSeparator {...props} key={itemKey || key} data-key={itemKey || key} />
             ) : (
               <DropdownItem
-                component="div"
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/DataList/__tests__/Generated/DataList.test.tsx
```diff
@@ -15,6 +15,7 @@ it('DataList should match snapshot (auto-generated)', () => {
       aria-label={'string'}
       onSelectDataListItem={(id: string) => undefined as void}
       selectedDataListItemId={"''"}
+      isCompact={false}
```
jschuler, line 18: Looks like this is updating tests outside of Table as well, would be good if this PR only focused on the Table, perhaps spin these other ones out into a separate PR?

PR #4025, opened 2020-04-05
feat(form): support removing top spacer form from labels

### LINKED ISSUE #3848: Support removing top spacer from form labels
This adds a `.pf-m-no-padding-top` modifier to a `.pf-c-form__label` element. Details in https://github.com/patternfly/patternfly-next/pull/2200

boaz0:
**What**: Closes #3848 

//cc @mcoker @tlabaj

### DIFF
--- a/packages/react-core/src/components/Form/FormGroup.tsx
+++ b/packages/react-core/src/components/Form/FormGroup.tsx
@@ -20,6 +20,8 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   validated?: 'success' | 'error' | 'default';
   /** Sets the FormGroup isInline. */
   isInline?: boolean;
+  /** Removes top spacer from label. */
+  hasNoPaddingTop?: boolean;
   /** Helper text after the field. It can be a simple text or an object. */
   helperText?: React.ReactNode;
   /** Helper text after the field when the field is invalid. It can be a simple text or an object. */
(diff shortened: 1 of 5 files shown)


mcoker:
@boaz0 thanks! The prop looks good, however we wouldn't use it when the label is adjacent to a regular form control element because the padding aligns the label text with the form control content.

Can you modify the example so that this prop is used on a label next to a checkbox or group of radio elements? You can see how it's used in the core workspace here - https://www.patternfly.org/v4/documentation/core/components/form#horizontally-aligned-labels

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -20,6 +20,8 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   validated?: 'success' | 'error' | 'default';
   /** Sets the FormGroup isInline. */
   isInline?: boolean;
+  /** Removes top spacer from label. */
+  isNoPaddingTop?: boolean;
```
tlabaj, line 24: I would change either change this name to `hasNoPaddingTop` or I would call is `hasPaddingTop` and and set the default to `True`.  The second option would be my preference.

@jschuler do you have preference here?

mcoker: I left a comment about this in a drawer PR - https://github.com/patternfly/patternfly-react/pull/4036, and mentioned how the `<PageSection>` supports `noPadding`, and `noPaddingMobile`.

tlabaj:
@boaz0,  This one just needs e rebase then it can go in.

Issue #3986, opened 2020-03-27
Popover - popover header changes from xl title to md


mcoker:
https://github.com/patternfly/patternfly/pull/2762/files

PR #4030, opened 2020-04-06
fix(popover): update default popover header size

seanforyou23:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/3986

This PR updates the default size of the popover header to be medium instead of extra-large.

## Breaking changes
1. **Popover**: PopoverHeader title now has a default size of medium. Extra large is no longer the default size.


tlabaj:
Looks good.  Just needs the breaking change notes.

Issue #3674, opened 2020-02-06
Button updates to allow text/icons to be placed on either side of one another


mcoker:
placeholder for https://github.com/patternfly/patternfly-next/issues/2574, will likely result in adding an additional element to the button to be able to better control the CSS

mcoker:
The work for this is is done in core - https://github.com/patternfly/patternfly/pull/2828/

* Removes `.pf-c-button__text`
* Adds `.pf-m-start` and `.pf-m-end` variations to `.pf-c-button__icon` for when the icon is either positioned at the start or end of the button.

PR #4032, opened 2020-04-06
feat(Button): allow icon to be used in any variant, add icon css modifiers

### LINKED ISSUE #3993: Button - allow any button variant to support icon prop

Related to https://github.com/patternfly/patternfly-react/issues/3674

Looks like the button's `icon` prop only works if `variant === ButtonVariant.link`. I confirmed with design that any button should support the `icon` prop.

I'm not sure about the `plain` variant though. Technically the icon for the plain variant will work with either the `icon` prop or the current behavior where the icon is passed as the children. However a plain button should only be used with an icon, so we don't want to encourage use with text. However for the sake of consistency, I would be in favor of supporting either `<Button variant="plain" aria-label="Action"><TimesIcon /></Button>` or `<Button variant="plain" aria-label="Action" icon={<TimesIcon />}></Button>`, and we just make it clear in the documentation that you shouldn't pass regular text to the plain button. WDYT @tlabaj? Also not sure if this is a breaking change?

kmcfaul:
**What**: Closes #3993, #3674

### DIFF
--- a/packages/react-core/src/components/Button/Button.tsx
+++ b/packages/react-core/src/components/Button/Button.tsx
@@ -46,7 +46,7 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   iconPosition?: 'left' | 'right';
   /** Adds accessible text to the button. */
   'aria-label'?: string;
-  /** Icon for the button if variant is a link */
+  /** Icon for the button. Usable by all variants except for plain. */
   icon?: React.ReactNode | null;
   /** Set button tab index unless component is not a button and is disabled */
   tabIndex?: number;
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/Button/Button.tsx
```diff
@@ -94,14 +94,11 @@ export const Button: React.FunctionComponent<ButtonProps & OUIAProps> = ({
...
-      )}
-      {variant === ButtonVariant.link && <span>{children}</span>}
-      {variant !== ButtonVariant.link && children}
-      {icon && variant === ButtonVariant.link && iconPosition === 'right' && (
-        <span className={styles.buttonIcon}>{icon}</span>
+      {icon && iconPosition === 'left' && (
+        <span className={css(styles.buttonIcon, styles.modifiers.start)}>{icon}</span>
       )}
+      {children}
+      {icon && iconPosition === 'right' && <span className={css(styles.buttonIcon, styles.modifiers.end)}>{icon}</span>}
```
mcoker, line 101: just curious why this conditional block is formatted differently than the other?

kmcfaul:
I think that's a good solution. I will prevent the icon prop from being used with the plain button variant and make a note of it in the description!

packages/react-core/src/components/Button/Button.tsx
kmcfaul: Not sure, I think the autoformatter just preferred a new line for the other one. The new update has them formatted the same.

PR #4034, opened 2020-04-06
fix(drawer): remove hidden when drawer is expanded

### LINKED ISSUE #3984: Drawer panel updates
- https://www.patternfly.org/v4/documentation/react/components/drawer#static-drawer has `hidden="" aria-hidden="true" aria-expanded="false"` set by default

- `pf-c-drawer__panel` only needs to toggle `hidden` attribute. `aria-hidden="true" aria-expanded="false"` can be removed

boaz0:
**What**: Closes #3984 

//cc @mattnolting

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -31,7 +31,7 @@ export const DrawerPanelContent: React.SFC<DrawerPanelContentProps> = ({
   ...props
 }: DrawerPanelContentProps) => (
   <DrawerContext.Consumer>
-    {({ isExpanded }) => (
+    {({ isExpanded, isStatic }) => (
       <div
         className={css(
           styles.drawerPanel,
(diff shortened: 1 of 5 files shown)


boaz0:
Thanks @jenny-s51 - I updated the changes.

mattnolting:
@boaz0 Thanks for taking this task! I don't see `hidden` showing up on `__panel` right now. Looks like an issue in your ternary. Also, there's some discussion as to whether the css handles the timeout or JS does that we should clarify. @mcoker wdyt?

boaz0:
Thanks @mattnolting I updated my changes again :D

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
```diff
@@ -43,8 +43,6 @@ export const DrawerPanelContent: React.SFC<DrawerPanelContentProps> = ({
           className
         )}
         hidden={!isExpanded}
```
mcoker, line 45: Hey @boaz0 looks good, the only issue I see is that `hidden` is applied to the panel by default unless the drawer `isExpanded`. Which means it's also being applied when the drawer is `isStatic`. With a static drawer, the panel is always expanded, so you end up with this:

Screen Shot 2020-04-24 at 12 27 44 PM

boaz0: Got it, I will fix that. :)

PR #4036, opened 2020-04-07
feat(Drawer): remove flag to add border, adjust flag for no border

### LINKED ISSUE #3988: Drawer - remove ability to add a border to panel
In https://github.com/patternfly/patternfly/pull/2887/ we removed the `.pf-m-border` variation. All non-inline/static drawers where the panel overlays the content should have a drop shadow. All inline/static drawer panels should have a border. Either the shadow or border can be removed with `.pf-m-no-border`, but there is already an issue to handle that - https://github.com/patternfly/patternfly-react/issues/3983

kmcfaul:
**What**: Closes #3988

Adjusts `hasBorder` to `hasNoBorder`, as the default of hasBorder was preventing any border or drop shadow from displaying. `hasNoBorder` is an opt-in to get the `.pf-m-no-border` class.

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -9,7 +9,7 @@ export interface DrawerPanelContentProps extends React.HTMLProps<HTMLDivElement>
   /** Content to be rendered in the drawer panel. */
   children?: React.ReactNode;
   /* Flag indicating that the drawer panel should not have a border. */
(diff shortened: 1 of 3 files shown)


kmcfaul:
@mcoker I mostly just followed the previous prop name, but I think that `noBorder` would work with more consistency. I will update.

redallen:
According to #3988 shouldn't we merge this into master and then merge it into v4? Also, @tlabaj , what's the naming convention we want to move towards for boolean props that add classes?

PR #4056, opened 2020-04-09
fix(DataToolbar): fix filter context from updating unnecessarily

### LINKED ISSUE #3770: DataToolbarFilter: Maximum update depth exceeded
In Cost Management, we discovered a "Maximum update depth exceeded" issue with the data toolbar.

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops

Screen Shot 2020-02-18 at 11 17 17 AM

This appears to be a problem with how the PatternFly requires a unique DataToolbarFilter component to represent the chip values for each tag below. If there are 100 tags, we end up with 100 DataToolbarFilter components.

Screen Shot 2020-02-18 at 10 25 02 AM

kmcfaul:
**What**: Closes #3770

DataToolbar's state was getting updated by each DataToolbarFilter even when nothing was changed, so I added a check to only kick off the update when a filter's chips got updated.

Moving DataToolbar's internal `filterInfo` object out of its state, to a private variable, would be another way to resolve the bug.

### DIFF
--- a/packages/react-core/src/components/DataToolbar/DataToolbarFilter.tsx
+++ b/packages/react-core/src/components/DataToolbar/DataToolbarFilter.tsx
@@ -56,15 +56,19 @@ export class DataToolbarFilter extends React.Component<DataToolbarFilterProps, D
     this.setState({ isMounted: true });
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: DataToolbarFilterProps) {
     const { categoryName, chips } = this.props;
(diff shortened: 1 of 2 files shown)


kmcfaul:
@redallen Any ideas about the difference in snapshots? I can't seem to get them to update to what the build is wanting, they are passing on my machine. Maybe I have to clean a cache somewhere?

Issue #4044, opened 2020-04-08
Make code generation consistent with `tsc`


redallen:
Currently, `react-tokens`, `react-styles`, and `react-icons` all generate their JS/TS differently from one another. Additionally, their output does not match what `tsc` would generate from similar source files.

`react-icons` has the additional issue that there are a few static files related to icon creation that must be compiled in addition to dynamically generating the code for each icon.

Bonus points for unifying the templating system that's currently a combination of interpolation, plop, and a custom API like this:
```
module.exports = {
  getOutputPath:
  getContent:
  getSingleOutputPath:
  getSingleContent:
};
```

Towards #3972

PR #4066, opened 2020-04-13
feat(react-styles): standardize css-in-js generation

redallen:
**What**: Generate CSS-in-JS files that are compatible with what `tsc` would generate. Use generation scripts like react-tokens does (one file to generate an object, another to write it to files). Stop generating `.ts` files that are compiled with `tsc` and later removed with a cleanup script, and instead just generate the `.js` and `.d.ts` files separately. This also reduces build time to 0.22s, down from 2.96s (since we had to run `tsc` before).

[Here's a diff for `css/components/AboutModalBox/about-modal-box.js`](https://www.diffchecker.com/aIouneRS) and [another diff for `css/components/AboutModalBox/about-modal-box.d.ts`.](https://www.diffchecker.com/prtU0Mon) The JS is effectively unchanged, but the `.d.ts` now has the actual class name strings instead of generic `string` types so VS Code's Intellisense can do it's work!

Before:
image

After:
image

**Additional issues**: Towards #4044


packages/react-styles/scripts/generateClassMaps.js
```diff
@@ -0,0 +1,74 @@
+const path = require('path');
+const fs = require('fs-extra');
+const glob = require('glob');
+const camelcase = require('camel-case');
+
+/**
+ * @param {string} cssString - CSS string
+ */
+function getCSSClasses(cssString) {
+  return cssString.match(/(\.)(?!\d)([^\s.,{[>+~#:)]*)(?![^{]*})/g);
```
mturley, line 10: Is there any way we can use some CSS parser library to pull out these class names, instead of our own regex? Maybe that would be less performant. I'm hesitant to take on maintaining stuff like this in case it's missing some corner case or could be broken by future changes to the CSS spec (however unlikely).

Maybe we could use something like one of these, and then walk the resulting AST:
* https://www.npmjs.com/package/css-selector-parser (https://stackoverflow.com/a/38422908)
* https://www.npmjs.com/package/css
* https://www.npmjs.com/package/string-extract-class-names

redallen: I just copied this over from our existing build script. I think it would be fantastic to leverage https://www.npmjs.com/package/css which we already use.

> Also, can we convert these scripts to TS for the sake of maintainability, or does that complicate the build too much?

We could, but this would also require us using `ts-node` instead of `node`. We have maybe 20 or so such script files in the repo that it'd be good to convert all of them.

Issue #3980, opened 2020-03-26
Set goals for integration suite


redallen:
Right now [we require contributors write demos in TSX in our integration suite](https://github.com/patternfly/patternfly-react/blob/master/packages/react-integration/README.md) for typechecking purposes. These demos are often copied from MD files, types added haphazardly, and many are out of date from the MD files.

We should set goals for what we want this integration suite to accomplish. Some possible goals are:
- Typechecking published examples outside of what `ts-jest` could
- Interaction testing complex interactions outside of what Jest can
- Only requiring a single set of Typescript demos rather than the 2 sets we have now
  - Contributors only have to write their demo _once._
  - Examples are documented in Typescript for our increasing number of Typescript consumers.
- Verifying react-inline-edit-extension, react-topology, and react-virtualized-extension per https://github.com/patternfly/patternfly-react/pull/4073#pullrequestreview-393141638

seanforyou23:
Type checking the exported components 👍
Interaction testing complex interactions 👍

About using react-integration src as a single source of truth, I'd like to open up a discussion about it. I've always seen a lot of value in having two workspaces that each serve different purposes. We need the main one like we have today, which shows the main examples in their best form, those examples are curated and optimized by the whole team with special care and they only show a subset of the functionality. These represent the "blessed" interaction patterns as described in the usage and behavior guidelines (or something along these lines).

That's when this second workspace comes into play. We need a place where we can verify things like complex customizations of the components, case-specific usages, all possible variations of a prop, etc. Those are often things we don't necessarily want to show from the main docs and it affords us more flexibility around the types of demos we can create without having to run each one through a whole design process.

One way this could fit in incrementally is that the TypeScript examples could then be exposed in the docs in some way where the JavaScript examples are. Maybe we could formalize how the TS demos are formatted/added and simply link to them from the main docs (leaving JS as the primary example snippets). I can see how publishing the react-integration build would be helpful for a lot of reasons. Thoughts?

PR #4073, opened 2020-04-14
feat(react-inline-edit-extension): convert to TS

### LINKED ISSUE #4043: Convert react-inline-edit-extension to typescript
Other than `eslint-plugin-patternfly-react`, `react-inline-edit-extension` is our only remaining JS package. It should be converted to Typescript for proper typings and readability. Part of #3972 so that it can be compiled with `tsc` without having to:

1. Enable extra flags
2. Fix errors when using `import { ... } from '@patternfly/react-*';`

redallen:
**What**: Closes #4043

This conversion was mostly converting propTypes to Typescript and adding/amending helper types that were incorrect.

A LegacyRef was also replaced with a modern `ref.current` in `editableRowWrapper.tsx`.

This PR changes the directory structure by moving `Body`, `CancelButton`, `ConfirmButton`, `utils` from under the `src/components/InlineEdit` directory to their own `component` or `util` directories that matches react-core and other package's layouts.

### @patternfly/react-virtualized-extension
- Many types have been added and amended while converting this project to TS.


redallen:
@seanforyou23 I've added that point to #3980 and we can discuss there.

@seanforyou23 I am seeing the "mv" history preserved except with `CancelButton.js` and `ConfirmButton.js` which are both very small files that _are_ completely changed. Are you okay with losing the commit history on those, or should I open a new PR?

seanforyou23:
Sounds good @redallen thx. And the renames I noticed were under utils like inlineEditFormatterFactory. I don't want to hold up progress so feel free to proceed using your discretion (it's a deprecated component anyway).

PR #4076, opened 2020-04-15
perf(packages): Use tsc instead of babel for build

### LINKED ISSUE #3763: PF4 component source maps
Source maps are not generated per component basis, only for the `index.js`. If you use absolute import paths, the console in the browser will be swarmed with warnings:

Screenshot from 2020-02-18 10-25-29

Solution: generate source maps for all components.

### LINKED ISSUE #4046: Move to `tsc --build`
Replace the babel config with tsconfig per-package. Fix the dependencies in `package.json` files, and update the `yarn plop` template to boot.

redallen:
**What**: Replace running `babel` per-package with running [`tsc --build`](https://www.typescriptlang.org/docs/handbook/project-references.html) at the project root. [`tsc --build`](https://www.typescriptlang.org/docs/handbook/project-references.html) offers a number of advantages outlined in #3972.

**How:**

- Replace per-package `.babelrc` files with `tsconfig.json` and `tsconfig.cjs.json` files
- Replace incremental build script with `tsc --build`
  - Use `"importHelpers": true` option to avoid duplicated spread and destructuring helpers at top of every file which helps with dist size and bundling
    - Add tslib as a dependency to non-generated packages
    - Disable `importHelpers` option in react-icons to match other generated packages. Remove single usage of `import *` from it
    - Add `dist/*.tsbuildinfo` files to `.npmignore`s
  - Always run styles, icons, tokens, and SASS generation with new `build:generate` target
- Remove per-package UMD builds
  - Run `yarn build:umd` which generates UMD bundles only in CI (not needed for dev env)
- Remove prop-types as a peer dependency since `tsc` can no longer use the babel plugin [typescript-to-proptypes](https://www.npmjs.com/package/typescript-to-proptypes)
  - Remove prop-types as an external dependency from our rollup UMD build
- Fix root `babel.config.js` and add its deps to root `package.json` for Jest's `babel-jest` (possible future work to use `ts-jest`)
  - Update snapshots which now use `dist/js` builds that render cross-package imports as `<Component>` (possible future work could be to define a `displayName` or give a function name to these components)
- Fix plop package generation and add option for Typescript

**Results:**

Overall, I think our compiled code quality has increased. [Here's a diff of `esm/components/Brand/Brand.js`](https://www.diffchecker.com/xE3yOqHg), and [another of `js/components/Brand/Brand.js`](https://www.diffchecker.com/NGwMcYXt) (Babel is on the left, and `tsc` + tslib is on the right).

I don't believe that any component broke as `yarn start` and all the components I've checked on the [doc preview](https://patternfly-react-pr-4076.surge.sh) work the [same as before.](http://patternfly-react-v4.surge.sh)

**Additional issues**: Closes #4046. Closes #3763 (by publishing source maps). Towards #3972.

## Breaking changes
**ALL packages:**

- We no longer support UMD builds for individual packages. Consider using our [react-core.umd.js bundle](https://unpkg.com/@patternfly/react-core@3/dist/umd/react-core.umd.js) instead.
- We no longer define `propTypes` for our components. Consider using our Typescript types under each packages' `dist/js` folder instead.


packages/react-icons/src/createIcon.tsx
```diff
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React from 'react';
```
seanforyou23, line 1: What changes here?

redallen: We don't need the [importStar helper](https://github.com/microsoft/tslib/blob/master/tslib.js#L227) in our generated code anymore.

babel.config.js
```diff
@@ -1,4 +1,11 @@
 module.exports = {
-  extends: './packages/.babelrc',
-  presets: [['@babel/env', {'modules': 'commonjs'}], '@babel/react'],
-};
+  presets: [
```
seanforyou23, line 2: What are these babel configs used for?

redallen: Jest still uses `babel-jest`. Future work is to possibly use `ts-jest` to typecheck tests.

packages/react-icons/scripts/writeIcons.js
```diff
@@ -55,7 +55,7 @@ export default ${jsName};
 
 const writeDTSExport = (fname, jsName, icon) => {
   outputFileSync(
-    join(outDir, 'js/icons', `${fname}.ts`),
+    join(outDir, 'js/icons', `${fname}.d.ts`),
```
seanforyou23, line 58: 👍

Issue #4079, opened 2020-04-15
Title enhancement for size/headingLevel


seanforyou23:
We'd like to improve the developer experience of the Title component a bit, to help with new usage as well as upgrade paths.

The enhancement is to provide a default `size` property value based on the provided `headingLevel` property value.

The user should still be able to override the size if they want to explicitly set it, otherwise, the default size value should be derived by the mapping table below. So 4-2xl = h1, xl = h2, etc.

`4xl | 3xl | 2xl | xl  | lg  | md | md | md`
`h1  | h1  | h1  | h2  | h3 | h4 | h5 | h6`

This would change [size](https://github.com/patternfly/patternfly-react/blob/v4/packages/react-core/src/components/Title/Title.tsx#L18) from being required to optional.

PR #3922, opened 2020-03-17
fix(title): remove problematic default headingLevel

seanforyou23:
**What**: This PR makes the `headingLevel` prop required, replace TitleSize with TitleSizes, and also removes TitleLevel enum as these values are not expected to change.

Closes # https://github.com/patternfly/patternfly-react/issues/1439

## Breaking changes
1. **Title**: Removes enum `TitleSize`. The enum `TitleSizes` should be used instead.
2. **Title**: Removes valid values `xs` and `sm` for `size` prop. Use size `md` instead.
3. **Title**: Removes enum `TitleLevel`. Use one of the string values 'h1', 'h2', 'h3', 'h4', 'h5', or 'h6' instead.
4. **Title**: `headingLevel` is now a required prop. Update any existing usage of `Title` to supply a string value for headingLevel of h1-h6.
5. **Title**: Removes default value `h1` for `headingLevel` prop. Use one of the string values 'h1', 'h2', 'h3', 'h4', 'h5', or 'h6' explicitly.

PR #4081, opened 2020-04-15
feat(title): supply defaults for title size property

seanforyou23:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4079

This PR makes `size` optional and generates a default value for it (when not supplied by user) that corresponds to to closest likely size given the heading level. This should not only simplify Title usage, but also ease upgrade pains in some cases.

**Additional issues**: https://github.com/patternfly/patternfly-react/pull/3922

### DIFF
--- a/packages/react-core/src/components/Title/Title.tsx
+++ b/packages/react-core/src/components/Title/Title.tsx
@@ -11,11 +11,20 @@ export enum TitleSizes {
   '4xl' = '4xl'
 }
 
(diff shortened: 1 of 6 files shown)


seanforyou23:
Not sure we want to display those default sizes in the doc examples, should I only test these defaults in react-integration?

seanforyou23:
Thx @jschuler - updated the description.

PR #4090, opened 2020-04-17
fix(page): use xl breakpoint instead of md

redallen:
**What**: Core has updated when to hide the Page sidenav in their media queries. Towards #1801. New breakpoint is *1200px, demo can be seen here: https://patternfly-react-pr-4090.surge.sh/documentation/react/demos/pagelayout/default-nav

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -100,15 +100,11 @@ export class Page extends React.Component<PageProps, PageState> {
   handleResize = () => {
     const { onPageResize } = this.props;
     const windowSize = window.innerWidth;
-    // eslint-disable-next-line radix
-    const mobileView = windowSize < Number.parseInt(globalBreakpointXl.value, 10);
+    const mobileView = windowSize < Number.parseInt(globalBreakpointXl.value);
     if (onPageResize) {
       onPageResize({ mobileView, windowSize });
     }
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    this.setState(prevState => ({
-      mobileView
-    }));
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/Page/Page.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/Page/page';
 import { css } from '@patternfly/react-styles';
-import globalBreakpointMd from '@patternfly/react-tokens/dist/js/global_breakpoint_md';
+import globalBreakpointLg from '@patternfly/react-tokens/dist/js/global_breakpoint_lg';
```
mcoker, line 4:
```suggestion
import globalBreakpointXl from '@patternfly/react-tokens/dist/js/global_breakpoint_xl';
```

packages/react-core/src/components/Page/Page.tsx
```diff
@@ -101,7 +101,7 @@ export class Page extends React.Component<PageProps, PageState> {
     const { onPageResize } = this.props;
     const windowSize = window.innerWidth;
     // eslint-disable-next-line radix
-    const mobileView = windowSize < Number.parseInt(globalBreakpointMd.value, 10);
+    const mobileView = windowSize < Number.parseInt(globalBreakpointLg.value, 10);
```
mcoker, line 104:
```suggestion
    const mobileView = windowSize < Number.parseInt(globalBreakpointXl.value, 10);
```

redallen:
Need to investigate why ALL integration tests are failing. That's very odd.

redallen:
Fixed in #4116

PR #4092, opened 2020-04-17
chore(packages): Bump version of react-scripts used by demo-app-ts to prevent node-gyp errors on yarn install

mturley:
For some reason, on my machine when running `yarn install` after https://github.com/patternfly/patternfly-react/pull/4076 was merged I was getting failures related to node-gyp trying to compile an old version of `fsevents` (1.2.4), which does not have prebuilt binaries available for macOS. `yarn why fsevents` found that the only reason we depend on `fsevents@1.2.4` is because we're on an old version of `react-scripts` in `packages/react-integration/demo-app-ts`.

This PR bumps that version of react-scripts (which should have no real impact as long as integration tests still run), and `yarn install` no longer needs to compile `fsevents` on my machine. No more errors, and a faster install time :)

Thanks @redallen for the help troubleshooting this.

### DIFF
--- a/packages/react-integration/demo-app-ts/package.json
+++ b/packages/react-integration/demo-app-ts/package.json
@@ -17,7 +17,7 @@
     "@types/react": "^16.8.0",
     "@types/react-dom": "^16.8.0",
     "@types/react-router-dom": "^4.3.1",
-    "react-scripts": "2.1.3",
+    "react-scripts": "^3.4.1",
     "typescript": "^3.8.3"
   },
   "browserslist": [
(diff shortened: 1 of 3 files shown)


mturley:
Just rebased and pushed a commit that should fix the `build_demo_app` CI failure, but I'm not sure why the `test_jest_pf4` one was failing. Let's see what happens this time.

mturley:
Hm.

mturley:
@redallen bizarrely enough, I'm not able to locally reproduce the build failures CircleCI is showing here, and I am now able to install/build on the existing v4 branch perfectly fine (no longer seeing the errors that led me to open this PR in the first place). Maybe we should just close this.. but I am thoroughly confused, would love to figure out wtf is happening.

mturley:
I'm just going to close this, things seem to be working fine now.

PR #4127, opened 2020-04-24
chore(EmptyStateIcon): update docs to reflect suggested icons

### LINKED ISSUE #3810: EmptyStateIcon should provide options to pick the right icon
The EmptyStateIcon icon prop accepts a string, but that doesn't really do anything `icon?: string | React.FunctionComponent<IconProps>`.

Looking at the design guidelines, we encourage a specific set of icons to be used https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state
We should provide an enum/a list of string literals to make it easy to pick one of these.

So perhaps instead of `string` it should accept an enum of supported react-icons, like `SearchIcon`

nicolethoen:
**What**: Closes #3810

### DIFF
--- a/packages/react-core/src/components/EmptyState/EmptyStateIcon.tsx
+++ b/packages/react-core/src/components/EmptyState/EmptyStateIcon.tsx
@@ -5,7 +5,8 @@ import styles from '@patternfly/react-styles/css/components/EmptyState/empty-sta
 export interface EmptyStateIconProps {
   /** Additional classes added to the EmptyState */
   className?: string;
-  /** Icon component to be rendered inside the EmptyState on icon variant */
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/EmptyState/EmptyStateIcon.tsx
```diff
@@ -5,7 +5,8 @@ import styles from '@patternfly/react-styles/css/components/EmptyState/empty-sta
 export interface EmptyStateIconProps {
   /** Additional classes added to the EmptyState */
   className?: string;
-  /** Icon component to be rendered inside the EmptyState on icon variant */
+  /** Usually a CheckCircleIcon, ExclamationCircleIcon, LockIcon, PlusCircleIcon, RocketIcon
```
tlabaj, line 8: I think you should keep the description of the prop here as well.

PR #4129, opened 2020-04-24
fix(nav): fix scrolling

### LINKED ISSUE #3200: NavTertiaryList scroll bugs
In OpenShift console, we have tabbed navigation that can include more tabs than can be displayed at a given viewport.  `NavTertiaryList` seems like a good way for us to solve for this problem.  However, I've noticed a few bugs in attempting to implement `NavTertiaryList` in OpenShift console (see https://github.com/openshift/console/pull/3096). 

* In my OpenShift console prototype, the scroll buttons don't appear on initial load.  Resizing the viewport fixes the bug.  Initialization bug?
* Sometimes clicking the scroll buttons doesn't result in scrolling in the entire nav bar (stops short and remaining tabs are never revealed no matter how much clicking occurs). 
* Rapidly clicking the scroll button can result in the first or last tab being selected, which can be confusing (see below).  In my OpenShift console prototype, I accounted for this by not overlaying the scroll buttons on top of the tabs.
rYKy0zMiSx
* Scrolling the nav with the mouse pointer instead of the buttons does not show or hide the buttons when reaching the beginning or end of the scrollable area.
RCGvmphUB1
* Font weight increase for active item causes a "shimmy" of the nav items on select, and the last nav item can end up clipped. 
nknYg1HqRK

cc: @spadgett

### LINKED ISSUE #4054: Nav - updates to support new design
A new design was introduced in https://github.com/patternfly/patternfly/pull/2884. There are a few small changes to support the design:

* Remove `<NavItemSeparator />` - the `<Divider>` component should be used instead. 
* Grouped navigation lists should use `.pf-c-nav__simple-list` instead of `.pf-c-nav__list`

redallen:
**What**: Fix the scrolling buttons for the horizontal and tertiary navigation variants.

Closes #3200
Closes #4054 

**Additional issues**: The scrollable tertiary nav must now be in a PageSection, but @mcoker is working upstream so that it can work outside of one. Once that work is complete the following breaking change note can be removed.

## Breaking changes
-  **Nav**: The tertiary variant of NavList must now be inside a `<PageSection type="nav">` to scroll properly.

### DIFF
--- a/packages/react-core/src/components/Nav/NavList.tsx
+++ b/packages/react-core/src/components/Nav/NavList.tsx
@@ -21,6 +21,13 @@ export interface NavListProps
   ariaRightScroll?: string;
 }
 
+const variantStyle = {
+  [NavVariants.default]: '', // TODO: Core wants a BEM element here.
+  [NavVariants.simple]: styles.navSimpleList,
+  [NavVariants.horizontal]: styles.navHorizontalList,
+  [NavVariants.tertiary]: styles.navTertiaryList
+};
+
 export class NavList extends React.Component<NavListProps> {
   static contextType = NavContext;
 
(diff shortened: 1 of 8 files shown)


packages/react-core/src/components/Nav/NavList.tsx
```diff
@@ -92,45 +101,41 @@ export class NavList extends React.Component<NavListProps> {
...
+    this.handleScrollButtons();
   }
 
   componentWillUnmount() {
-    const { variant } = this.props;
-    const isHorizontal = variant === NavVariants.horizontal || variant === NavVariants.tertiary;
-    if (isHorizontal) {
-      document.removeEventListener('resize', this.handleScrollButtons, false);
-    }
+    document.removeEventListener('resize', this.handleScrollButtons, false);
```
jschuler, line 109: looks like previous bug, but should be on window as well

mcoker:
If the nav has overflow and is in the scrollable state, if you scroll with your mouse, the buttons don't update like you would expect. Is that an easy thing to support?

redallen:
@mcoker We didn't originally have this functionality, but I just added it with one line. Let me know how it works for you once the preview updates.

Issue #4040, opened 2020-04-08
Drawer:  update padding prop names for consistency


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
We currently have inconsistent prop naming for adding/removing padding.
We should update props from names from `noPadding` to `hasNoPadding` in DrawerPanelBody and DrawerHead

redallen:
What about just `hasPadding`?

redallen:
Closed in #4133

Issue #4039, opened 2020-04-08
Consistent Prop names for adding/removing padding


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
We currently have inconsistent prop naming for adding/removing padding.
We should update props from names frmo `noPadding` to `hasNoPadding`.  The following Components need updates

- DataListContent
- Page Section

redallen:
What about just `hasPadding`?

redallen:
Closed in #4133

PR #4133, opened 2020-04-27
feat(Drawer, DataList, Page): rename noPadding to hasNoPadding

### LINKED ISSUE #4013: Page - updates to noPadding*
As part of https://github.com/patternfly/patternfly/pull/2816, we dropped the `pf-m-no-padding-mobile` class and replaced it with the use of `pf-m-padding[-on{breakpoint}]` and `pf-m-no-padding[-on{breakpoint}]`.

We will need to update page sections to support those classes and their ability to be applied to a single element like `pf-m-no-padding`, `pf-m-padding-on-md`, `pf-m-no-padding-on-lg`, `pf-m-padding-on-xl`, which would remove default padding, add it back on md viewports, remove it on lg, added it back on xl. 

Also the existing `noPaddingMobile` prop can be updated from using `.pf-m-no-padding-mobile` (no longer exists) to use `.pf-m-no-padding.pf-m-padding-on-md`.
* @tlabaj curious what you think of that, creating `noPaddingMobile` from a combination of classes in core. Does that seem OK, or for something specific like that, would you prefer core manage that combination and continue to offer a class like `.pf-m-no-padding-on-mobile`?

cc @christiemolloy

kmcfaul:
And add Page padding breakpoint logic

**What**: Closes #4013, #4039, #4040

## Breaking changes
1. Renamed `noPadding` to `hasNoPadding` for Drawer, DataList and Page components.
2. Added `hasPaddingOn` and `hasNoPaddingOn` properties to PageSection, accounting for page size breakpoints. Breakpoints are defined in the `PageSectionBreakpoints` enum.

### DIFF
--- a/packages/react-core/src/components/DataList/DataListContent.tsx
+++ b/packages/react-core/src/components/DataList/DataListContent.tsx
@@ -14,7 +14,7 @@ export interface DataListContentProps extends React.HTMLProps<HTMLElement> {
   /** Flag to show if the expanded content of the DataList item is visible */
   isHidden?: boolean;
   /** Flag to remove padding from the expandable content */
-  noPadding?: boolean;
+  hasNoPadding?: boolean;
   /** Adds accessible text to the DataList toggle */
   'aria-label': string;
 }
@@ -25,7 +25,7 @@ export const DataListContent: React.FunctionComponent<DataListContentProps> = ({
   id = '',
   isHidden = false,
   'aria-label': ariaLabel,
-  noPadding = false,
+  hasNoPadding = false,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   rowid = '',
   ...props
@@ -37,6 +37,8 @@ export const DataListContent: React.FunctionComponent<DataListContentProps> = ({
     aria-label={ariaLabel}
     {...props}
   >
-    <div className={css(styles.dataListExpandableContentBody, noPadding && styles.modifiers.noPadding)}>{children}</div>
+    <div className={css(styles.dataListExpandableContentBody, hasNoPadding && styles.modifiers.noPadding)}>
+      {children}
+    </div>
   </section>
 );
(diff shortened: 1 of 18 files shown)


packages/react-core/src/components/DataList/DataListContent.tsx
```diff
@@ -14,7 +14,7 @@ export interface DataListContentProps extends React.HTMLProps<HTMLElement> {
   /** Flag to show if the expanded content of the DataList item is visible */
   isHidden?: boolean;
   /** Flag to remove padding from the expandable content */
-  noPadding?: boolean;
+  hasNoPadding?: boolean;
```
redallen, line 17: Can we change this to `hasPadding` and default it to `true`? Sounds a little better and codemods fix it all the same.

packages/react-core/src/components/Page/PageSection.tsx
```diff
@@ -41,21 +53,46 @@ const variantStyle = {
...
 }: PageSectionProps) => (
-  // TODO: Implement https://github.com/patternfly/patternfly/pull/2816
   <section
     {...props}
     className={css(
       variantType[type],
-      noPadding && styles.modifiers.noPadding,
+      hasNoPadding && styles.modifiers.noPadding,
+      hasPaddingOn &&
+        (hasPaddingOn.constructor !== Array
```
redallen, line 89: We have helpers for this! Checkout https://github.com/patternfly/patternfly-react/blob/v4/packages/react-core/src/helpers/util.ts#L243 and follow what Flex or Toolbar do for example.

packages/react-core/src/components/DataList/DataListContent.tsx
kmcfaul: @tlabaj Thoughts?

I don't mind either way.

packages/react-core/src/components/Page/PageSection.tsx
kmcfaul: I had seen them but didn't use them as PageSection only needed breakpoints on padding and no other modifiers, but its probably better this way for any future expansion on breakpoint specific modifiers. I'll update.

packages/react-core/src/components/DataList/DataListContent.tsx
kmcfaul: Keeping as `hasNoPadding` per standup discussion for now

PR #4140, opened 2020-04-28
chore(Wizard): make wizard inPage

### LINKED ISSUE #3668: Wizard: The wizard is another implementation of modal
This issue is the follow up work needed in React once core issue [#2046](https://github.com/patternfly/patternfly-next/issues/2046) is resolved.  The Wizard will default to an in-page variation so the consumer can place it in a modal.

(This will address the first bullet point in issue #2114 )

nicolethoen:
### To Modal component
- Uses pf-c-modal-box__title in the modal header if title is provided.
- Make title prop optional.
- Add aria-label prop for use if title is not provided.
- Add noPadding prop to place Modal children in the ModalContent rather than a ModalBody
- Add 'Modal with wizard' and 'No title' example to docs
### To Wizard component
- Only display wizard header if a title is provided.
- Added hideClose prop to allow user to hide the close button on wizard header.
- remove wizard's inPage prop.
- Update wizard examples. Add in page example showing wizard header with no close button. Add - Wizard in modal example.
- implemented Wizard's height prop.

Above changes address Core updates:
https://github.com/patternfly/patternfly/pull/2969
https://github.com/patternfly/patternfly/pull/2964

**What**: Closes #3668 

## Breaking changes
- Modal: Remove hideTitle prop. To hide the Modal header, do not pass a title prop, a description prop, or a header prop. If there is no title or header passed, please provide an aria-label prop to the Modal component to make it accessible.
- Wizard: Remove inPage prop. By default the Wizard will be displayed in page, filling its parent container. If the consumer passes a managed isOpen flag, then the Wizard will be displayed in a modal.


packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -44,10 +38,6 @@ export type WizardStepFunctionType = (
 ) => void;
 
 export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
-  /** True to show the wizard (not applicable for isInPage)*/
-  isOpen?: boolean;
-  /** True to show the wizard without the modal */
-  isInPage?: boolean;
```
dlabrecq, line 50: I don't understand why isOpen was removed? Are all wizards now being placed within a modal?

dlabrecq (dismissed):
You can see the adverse effect of the wizard changes in our examples

Screen Shot 2020-04-28 at 10 21 23 AM

nicolethoen:
@dlabrecq 
Yeah, this is a breaking change PR for sure. 
The wizard is just a regular component now, filling its parent element. So the steps scroll in the examples for Wizard. 

To see the Wizard component in a Modal, there is now an example in the Modal component examples. I could conceivably add it to the wizard examples as well.

The issue I addressed wanted to decouple the Wizard with the Modal component so that it wasn't a second modal implementation.

nicolethoen:
@tlabaj 
I added a variant to the Modal so that it can now house a Wizard - you can see it in the examples.
Do we want a variant to the Wizard that puts it in a modal? Do we want variants going both directions? 
Easy enough to add.

dlabrecq:
As an app developer, I wouldn't think to look at the modal component for wizard examples. The previous wizard examples had shown how to launch the wizard.

Note that this change will effect a lot of application code, in addition to the exponentially increasing number of breaking changes.

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -123,18 +111,9 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
   constructor(props: WizardProps) {
     super(props);
-    const newId = Wizard.currentId++;
-    this.isModal = !props.isInPage;
-    if (this.isModal) {
-      this.titleId = `pf-wizard-title-${newId}`;
-      this.descriptionId = `pf-wizard-description-${newId}`;
-    }
 
     this.state = {
```
redallen, line 118: If all you're doing is setting state, you don't need a constructor and can just set the class member `state`.

redallen (changes requested):
I'm liking where this is going~! You can create a `wizard.css` file with a rule like
```css
.ws-react-c-wizard {
  min-height: 400px;
}
```

and just import it in `Wizard.md` to fix the height of the examples in the preview :)

redallen:
As for the API, I'm fine with whatever we choose since there aren't that many products using a Wizard:

image

My 2 cents are that we should create a wrapper `WizardModal` component and export that. I could more easily write codemods for that.

mcoker:
This looks great @nicolethoen! A couple of things. In the example where we show the wizard in a modal
* the large modal size variation is preferred for a wizard in a modal. Can we update the example to use `size={ModalVariant.large}`?
* In a modal, the wizard header should be present. That will also introduce a close button.
* Is `disableFocusTrap={true}` required with `isBasic` if there isn't a close button? If I don't add that to the examples, I get an error _"You can't have a focus-trap without at least one focusable element"_. I wonder if we need to call any of that out in the docs?

You can see examples of the size variation and wizard header in the core docs under demos -> wizard - http://patternfly-v4.surge.sh/documentation/core/demos/wizard

nicolethoen:
> * In a modal, the wizard header should be present. That will also introduce a close button.
> * Is `disableFocusTrap={true}` required with `isBasic` if there isn't a close button? If I don't add that to the examples, I get an error _"You can't have a focus-trap without at least one focusable element"_. I wonder if we need to call any of that out in the docs?

These are the sorts of things that would actually be solved much easier with a WizardModal wrapper. Because the Wizard needs to know it's in a modal to have these features work correctly. It cannot remain parent agnostic. I will begin implementing a WizardModal component as Zac recommended.

nicolethoen:
> > * In a modal, the wizard header should be present. That will also introduce a close button.
> > * Is `disableFocusTrap={true}` required with `isBasic` if there isn't a close button? If I don't add that to the examples, I get an error _"You can't have a focus-trap without at least one focusable element"_. I wonder if we need to call any of that out in the docs?
> 
> These are the sorts of things that would actually be solved much easier with a WizardModal wrapper. Because the Wizard needs to know it's in a modal to have these features work correctly. It cannot remain parent agnostic. I will begin implementing a WizardModal component as Zac recommended.

So, I think I was able to get it to work. If the consumer provides a Title, the WizardHeader will appear. So a Wizard with a Title placed in a Modal with the isBasic flag works like an old Wizard :)

karelhala:
> As an app developer, I wouldn't think to look at the modal component for wizard examples. The previous wizard examples had shown how to launch the wizard.

The wizard component was from the start fundamentally wrong, it's not specific usecase only to modals. It should be a component that can be placed wherever you want and be free with it. Why not placing the wizard component in dropdown? It might sound strange, but the useage of it can be actually pretty nice. As for the demo, the wizard page demo could have an example in modal since that's one usecase, but I wouldn't limit the wizard component to modals only.

I think this is great direction with the wizard compoent, thank you @nicolethoen for taking care of it!

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -90,13 +93,16 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
...
       id={modalBoxAriaDescribedById || id}
     >
-      {showClose && <ModalBoxCloseButton onClose={onClose} />}
-      {modalBoxHeader}
-      {description && <ModalBoxDescription id={id}>{description}</ModalBoxDescription>}
-      <ModalBoxBody {...props} {...(!description && { id })}>
-        {children}
-      </ModalBoxBody>
-      {modalBoxFooter}
+      {!isBasic && showClose && <ModalBoxCloseButton onClose={onClose} />}
```
karelhala, line 96: Can we remove this `isBasic` prop and unify all these elements into one component?
```JSX
const ModalBody = () => {
  onst modalBoxHeader = header ? (
    <div className={css(titleStyles.title)}>{header}</div>
  ) : (
    <ModalBoxHeader hideTitle={hideTitle}> {title} </ModalBoxHeader>
  );

  const modalBoxFooter = footer ? (
    <ModalBoxFooter>{footer}</ModalBoxFooter>
  ) : (
    actions.length > 0 && <ModalBoxFooter>{actions}</ModalBoxFooter>
  );
  return <Fragment>
    {showClose && <ModalBoxCloseButton onClose={onClose} />}
    {modalBoxHeader}
    {description && <ModalBoxDescription id={id}>{description}</ModalBoxDescription>}
    <ModalBoxBody {...props} {...(!description && { id })}>
          {children}
        </ModalBoxBody>
    {modalBoxFooter}
</Fragment>
}
```

This way we won't have a ton of ifs in this component and it can just have children as with any other component.

If we don't want to break backwards compatibility that much we could add prop called bodyWrapper to Modal where you'd by default use `ModalBody` component, if you'd don't want to use you could pass in `null` or w/e. The bodyWrapper prop could be either null, string, number, function or node.

karelhala:
I've had discussion with a few folks about this PR, I really want it to be part of PF, because it can be great example of how to write and export componets as building blocks. However this can eventually be hard to migrate to as we usually have a lot of other things to do instead just fixing breaking changes from component libray.

This PR fixes the linked issue in elegant way, but it will introduce a breaking change and would require from app developers to fix this. There can be another approach to this, it will bloat the PR, but I think it'll help us in the long run:
* Add new component exported as Wizard that will be just a wrapper that passes down all props except `isModal`. Based on this prop it'll use Wizard in modal (current way) or modal in page without modal at all (new and proposed way)
* It would be also beneficial to export a few new components as proposed in the original issue. App devs can then combine these components toger as they choose and as they want.
   * WizardHeader
   * WizardStepper
   * WizardBody
   * WizardInPage - combination of WizardHeader, WizardStepper and WizardBody. This component can by default have these 3 components used trough render funcion so consumers can override these part with their own. Meaning if I don't want to use WizardStepper and I want to implement my own I can do that easilly by adding `wizardStepper={(...props) => <div>My awesome stepper</div>}`.

@nicolethoen the work you've done here is awesome, it actually led me to these ideas and proposals, without this change we'd probably won't know what would be the ideal outcome of original issue. But I feel like this is the right path. I'd be more than happy to implement it as well, because this might be interesting addition. That being said if you'd need any more information or didn't understood anything, don't hesitate to ask. I'll monitor this PR and try help you as much as I could so we can have this change un in the wilderness soon.

packages/react-core/src/components/Modal/Modal.tsx
```diff
@@ -15,9 +15,9 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   /** Complex header (more than just text), supersedes title for header content */
   header?: React.ReactNode;
   /** Simple text content of the Modal Header, also used for aria-label on the body */
-  title: string;
-  /** Flag to hide the title */
-  hideTitle?: boolean;
```
jschuler, line 20: This looks great! Can you please add all the breaking changes to the bottom of your PR? in this form:
```
## Breaking changes
- Your list of changes.
- You can use MD
```

packages/react-core/src/components/Modal/examples/Modal.md
```diff
@@ -383,3 +382,107 @@ class NoHeader extends React.Component {
   }
 }
 ```
+
+```js title=No-title
```
jschuler, line 386: the no header and no title examples seem identical?

packages/react-core/src/components/Modal/examples/Modal.md
```diff
@@ -383,3 +382,107 @@ class NoHeader extends React.Component {
...
+          aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
+          occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+        </Modal>
+      </React.Fragment>
+    );
+  }
+}
+```
+
+```js title=With-wizard
```
jschuler, line 437: The Escape button does not seem to be working to close the wizard.

jschuler (changes requested):

packages/react-core/src/components/Modal/examples/Modal.md
nicolethoen: Lol, yep. That's evident of me experiencing tunnel vision.

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -73,29 +76,37 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
   const modalBoxHeader = header ? (
     <div className={css(titleStyles.title)}>{header}</div>
   ) : (
-    <ModalBoxHeader hideTitle={hideTitle}> {title} </ModalBoxHeader>
+    title && <div className={css(styles.modalBoxTitle)}>{title}</div>
```
jschuler, line 79: I think it might be good to keep the ModalBoxHeader and just update it. This way, existing consumers won't be too broken

jschuler (approved):
LGTM!

redallen (approved):
This API is a crowd pleaser!

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -293,42 +273,22 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
 
   componentDidMount() {
-    const { appendTo } = this.props;
-    const target: HTMLElement = this.getElement(appendTo);
-    if (this.isModal) {
-      if (this.container) {
-        target.appendChild(this.container);
-      }
-      this.toggleSiblingsFromScreenReaders(true);
+    const target = (typeof document !== 'undefined' && document.body) || null;
```
redallen, line 276: Creative way to write a ternary.

packages/react-core/src/components/Modal/Modal.tsx
```diff
@@ -38,6 +38,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   description?: React.ReactNode;
   /** Variant of the modal */
   variant?: 'small' | 'large' | 'default';
+  /** Flag indicating if modal content should have no padding*/
+  noPadding?: boolean;
```
mcoker, line 42: I wonder if this prop name is too generic and might overlap or be confusing with other props we could introduce into the modal, like one that removes all padding from the modal, or just the padding from the `<ModalBoxBody>` element. I have seen potential use cases for the former.

Technically this prop removes the `<ModalBoxBody>` element all together. 

Here is what it looks like now with regular modal content and `noPadding`

Screen Shot 2020-05-04 at 3 46 26 PM

mcoker:
Should we remove the `description` prop from wizard examples that don't also have a `title`, since `description` will only show up if you have a `title`?

packages/react-core/src/components/Modal/Modal.tsx
nicolethoen: hm... would it makes sense to have it be called noModalBoxBody? Would that make sense to a consumer? Could something like removeBody make more sense?

packages/react-core/src/components/Wizard/Wizard.tsx
nicolethoen: lol I guess I too quickly borrowed the logic from a different part of the code :P

packages/react-core/src/components/Modal/Modal.tsx
mcoker: Yeah, `noModalBoxBody` definitely isn't as easy of a name, though it's a more accurate description of what it does. Or maybe `noModalBoxBodyWrapper` since you can still pass content and it puts it where the body would go. `noModalBoxBody` and `removeBody` could be confused with the modal having no body content. At any rate, I think as long as we document it, any of those names seem OK to me. I'm interested in others' thoughts.

mcoker (approved):
Looks good to me! Thanks for all of your hard work on this @nicolethoen!

karelhala (approved):
Looking good! I am wondering if we could also somehow export helper functions `this.goToStepById`, `this.goToStepByName`, `this.onNext` and `this.onBack`. No need to be part of this PR at all! But it would make the further usage of smaller building blocks much easier.

redallen (approved):
I love seeing more lines removed than added :)

Thanks for the refactor @nicolethoen !

packages/react-core/src/components/Modal/Modal.tsx
nicolethoen: I could go with noModalBoxBodyWrapper... it's a mouthful.
@tlabaj @jschuler @dlabrecq any thoughts?

PR #4141, opened 2020-04-28
fix(docs): bump theme and improve speed

redallen:
Speeds up Gatsby for `yarn start` from 5-10 minutes to 1m33s on first run and 58s on second run.

Accomplished by tweaking webpack config in the theme and NOT importing the entirety of react-icons and react-tokens in both ESM and CJS formats.

Bonus: change top nav from always showing scroll buttons to conditionally showing them.

Before:
image
After:
image


redallen:
CI build_docs_pf4 speed is down to 5m from 10m! That's 5m off each of our build :tada:

PR #4144, opened 2020-04-28
chore(packages): bump react peer dep to 16.8.0

redallen:
**What**: Upgrade our peer dependency from `react@^16.4.0` (2 years old) to `react@^16.8.0` (1 year old) as a part of our breaking changes. This change would allow our library code to use React hooks moving forward. Previous discussion is in the #patternfly-react channel on Slack.

Closes patternfly/patternfly-org#1712

**Additional issues**: There are some concerns with products having to upgrade React, which in turn would break other 3rd party dependencies like Redux. These products are on a timeline where upgrading is feasible in the future, and the price to pay of not upgrading to `@patternfly/react-core@^4.0.0` is only that our components with hooks will not work (which currently, there are none).

## Breaking changes
- Our packages now can possibly use hooks, which requires `react@^16.8.0` instead of `react@^16.4.0`. We recommend upgrading your version of React if it is below 16.8.0.

### DIFF
--- a/packages/react-catalog-view-extension/package.json
+++ b/packages/react-catalog-view-extension/package.json
@@ -50,7 +50,7 @@
     "typescript": "^3.8.3"
   },
   "peerDependencies": {
-    "react": "^16.4.0",
-    "react-dom": "^15.6.2 || ^16.4.0"
+    "react": "^16.8.0",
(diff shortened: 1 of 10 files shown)


mturley:
Nice!! Excited to use some hooks

PR #4146, opened 2020-04-28
Tabs updates

### LINKED ISSUE #3951: Tabs - updates to support new design
https://github.com/patternfly/patternfly/pull/2757/

* adds vertical variation
* adds support for insets that align with our spacer system and can be applied at breakpoints
* adds box variation
* adds elements for tab button text and icon

Changes to the scroll behavior
* removes `pf-m-[start/end/start-current/end-current]`
* adds `pf-m-scrollable` variation applied to tabs to show and hide the scroll buttons. `pf-m-scrollable` should be applied when there is overflow in either direction
  * when `pf-m-scrollable` is applied and you're already scrolled as far as you can in one direction, that button should be disabled with the `disabled` attribute.
  * when `pf-m-scrollable` is not applied, buttons should be `disabled` and receive `aria-hidden="true"`

tlabaj:
**What**: Closes #3951

## Breaking changes
**Tabs**:
1. Changed prop `variant` for consistency.  You will need to update all instances of `variant` prop to `component` 
2. Changed `TabVariant` enum Name to `TabComponent` for consistent naming.  You will now need to update all instances of `tabVariant` to `TabComponent`. 

**Tab**: 
1. The title should be wrapped with `<TabTitleText>` for proper styling.  If you would like to place an Icon in the Tab, it should be wrapped with `<TabTitleIcon>` for proper styling.

### DIFF
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -7,7 +7,7 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   className?: string;
   /** URL associated with the Tab. A Tab with an href will render as an <a> instead of a <button>. A Tab inside a <Tabs variant="nav"> should have an href. */
   href?: string;
-  /** Content rendered in the tab title */
+  /** Content rendered in the tab title. Should be <TabTitleText> and/or <TabTitleIcon> for proper styling. */
   title: React.ReactNode;
   /** uniquely identifies the tab */
   eventKey: number | string;
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -5,7 +5,7 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   children?: React.ReactNode;
   /** additional classes added to the Tab */
   className?: string;
-  /** URL associated with the Tab. A Tab with an href will render as an <a> instead of a <button>. A Tab inside a <Tabs variant="nav"> should have an href. */
+  /** URL associated with the Tab. A Tab with an href will render as an <a> instead of a <button>. A Tab inside a <Tabs component="nav"> should have an href. */
   href?: string;
   /** Content rendered in the tab title. Should be <TabTitleText> and/or <TabTitleIcon> for proper styling. */
   title: React.ReactNode;
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -7,7 +7,7 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   className?: string;
   /** URL associated with the Tab. A Tab with an href will render as an <a> instead of a <button>. A Tab inside a <Tabs variant="nav"> should have an href. */
   href?: string;
-  /** Tab title */
+  /** Content rendered in the tab title */
   title: React.ReactNode;
   /** uniquely identifies the tab */
   eventKey: number | string;
--- a/packages/react-core/src/components/Tabs/TabTitleIcon.tsx
+++ b/packages/react-core/src/components/Tabs/TabTitleIcon.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
+
+export interface TabTitleIconProps extends React.HTMLProps<HTMLSpanElement> {
+  /** Icon to be rendered inside the tab button title. */
+  children?: React.ReactNode;
+  /** additional classes added to the tab title icon */
+  className?: string;
+}
+
+export const TabTitleIcon: React.FunctionComponent<TabTitleIconProps> = ({
+  children,
+  className = '',
+  ...props
+}: TabTitleIconProps) => (
+  <span className={css(styles.tabsItemIcon, className)} {...props}>
+    {children}
+  </span>
+);
--- a/packages/react-core/src/components/Tabs/TabTitleIcon.tsx
+++ b/packages/react-core/src/components/Tabs/TabTitleIcon.tsx
@@ -4,7 +4,7 @@ import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
 
 export interface TabTitleIconProps extends React.HTMLProps<HTMLSpanElement> {
   /** Icon to be rendered inside the tab button title. */
-  children?: React.ReactNode;
+  children: React.ReactNode;
   /** additional classes added to the tab title icon */
   className?: string;
 }
--- a/packages/react-core/src/components/Tabs/TabTitleText.tsx
+++ b/packages/react-core/src/components/Tabs/TabTitleText.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
+
+export interface TabTitleTextProps extends React.HTMLProps<HTMLSpanElement> {
+  /** Text to be rendered inside the tab button title. */
+  children?: React.ReactNode;
+  /** additional classes added to the tab title text */
+  className?: string;
+}
+
+export const TabTitleText: React.FunctionComponent<TabTitleTextProps> = ({
+  children,
+  className = '',
+  ...props
+}: TabTitleTextProps) => (
+  <span className={css(styles.tabsItemText, className)} {...props}>
+    {children}
+  </span>
+);
--- a/packages/react-core/src/components/Tabs/TabTitleText.tsx
+++ b/packages/react-core/src/components/Tabs/TabTitleText.tsx
@@ -4,7 +4,7 @@ import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
 
 export interface TabTitleTextProps extends React.HTMLProps<HTMLSpanElement> {
   /** Text to be rendered inside the tab button title. */
-  children?: React.ReactNode;
+  children: React.ReactNode;
   /** additional classes added to the tab title text */
   className?: string;
 }
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -186,7 +186,8 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
     }
   };
 
-  insetString = (s: string) => (s === '2xl' ? '_2xl' : capitalizeFirstLetter(s));
+  // Format the inset prop string by capitalizing the first letter.  If the string prop is '2xl' append '_' to beginning of the string so we can key the correct modifier.
+  FormatInsetString = (s: string) => (s === '2xl' ? '_2xl' : capitalizeFirstLetter(s));
 
   componentDidMount() {
     window.addEventListener('resize', this.handleScrollButtons, false);
@@ -238,12 +239,12 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
             isVertical && styles.modifiers.vertical,
             isBox && styles.modifiers.box,
             showScrollButtons && !isVertical && styles.modifiers.scrollable,
-            inset && styles.modifiers[`inset${this.insetString(inset)}` as keyof typeof styles.modifiers],
-            insetOnMd && styles.modifiers[`inset${this.insetString(insetOnMd)}OnMd` as keyof typeof styles.modifiers],
-            insetOnLg && styles.modifiers[`inset${this.insetString(insetOnLg)}OnLg` as keyof typeof styles.modifiers],
-            insetOnXl && styles.modifiers[`inset${this.insetString(insetOnXl)}OnXl` as keyof typeof styles.modifiers],
+            inset && styles.modifiers[`inset${this.FormatInsetString(inset)}` as keyof typeof styles.modifiers],
+            insetOnMd && styles.modifiers[`inset${this.FormatInsetString(insetOnMd)}OnMd` as keyof typeof styles.modifiers],
+            insetOnLg && styles.modifiers[`inset${this.FormatInsetString(insetOnLg)}OnLg` as keyof typeof styles.modifiers],
+            insetOnXl && styles.modifiers[`inset${this.FormatInsetString(insetOnXl)}OnXl` as keyof typeof styles.modifiers],
             insetOn2Xl &&
-              styles.modifiers[`inset${this.insetString(insetOn2Xl)}On_2xl` as keyof typeof styles.modifiers],
+              styles.modifiers[`inset${this.FormatInsetString(insetOn2Xl)}On_2xl` as keyof typeof styles.modifiers],
             className
           )}
           {...getOUIAProps('Tabs', ouiaId)}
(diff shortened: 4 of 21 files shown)


packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -63,11 +83,18 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
...
     rightScrollAriaLabel: 'Scroll right',
     variant: TabsVariant.div,
     mountOnEnter: false,
-    unmountOnExit: false
+    unmountOnExit: false,
+    inset: undefined,
+    insetOnMd: undefined,
+    insetOnLg: undefined,
+    insetOnXl: undefined,
+    insetOn2Xl: undefined
```
dlabrecq, line 97: Is there a particular reason these props are defined as `undefined`?

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -87,13 +114,39 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
       // most recently selected tabContent
       tabContentRef.current.hidden = false;
     }
+    // Update scroll button state and which button to highlight
+    setTimeout(() => {
+      this.handleScrollButtons();
+    }, 1);
```
dlabrecq, line 120: Using `setTimeout` is kind of a red flag. Is there something we can do via `componentDidUpdate` instead, when state has changed?

Better yet, perhaps use the callback provided by `setState`?

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -133,6 +186,18 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
     }
   };
 
+  insetString = (s: string) => (s === '2xl' ? '_2xl' : capitalizeFirstLetter(s));
```
dlabrecq, line 189: It's not clear what format is expected here. Some comments (i.e., examples) would help clarify how `capitalizeFirstLetter` is used to build modifiers.

packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -7,7 +7,7 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   className?: string;
   /** URL associated with the Tab. A Tab with an href will render as an <a> instead of a <button>. A Tab inside a <Tabs variant="nav"> should have an href. */
   href?: string;
-  /** Tab title */
+  /** Content rendered in the tab title */
   title: React.ReactNode;
```
dlabrecq, line 11: It appears that `TabTitleText` is required for proper styling. Shouldn't the type reflect that?

packages/react-core/src/components/Tabs/Tabs.tsx
tlabaj: This is an internal function used only by this component  would a comment be enough or should I add examples to?

packages/react-core/src/components/Tabs/Tabs.tsx
tlabaj: Do you recommend setting it to null?    We only want to apply the modifier if one is supplied?

tlabaj:
@mcarrano to consolidate the examples, I added a checkbox at the bottom, "isBox" to switch modifier on and off.  If it is too confusing, I can separate the examples (that is how I i had it initially).

mcarrano:
I like this approach @tlabaj . It reduces the number of examples which is a good thing :) My only concern is it could be easily missed. Would it be possible to add some text at the top of the examples section to explain this? I can give you some text if you want.

tlabaj:
@mcarrano yes,  please provide some text that you think would work.

mcarrano:
@tlabaj how about, "Most tab variations are available as open (default) or box style tabs. Select the 'isBox' checkbox to preview an example with box styled tabs."

packages/react-core/src/components/Tabs/TabTitleIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+export interface TabTitleIconProps extends React.HTMLProps<HTMLSpanElement> {
+  /** Icon to be rendered inside the tab button title. */
+  children?: React.ReactNode;
+  /** additional classes added to the tab title icon */
+  className?: string;
+}
+
+export const TabTitleIcon: React.FunctionComponent<TabTitleIconProps> = ({
+  children,
+  className = '',
```
redallen, line 14: Nit: we don't have to set these to a default of empty string anymore on `v4`.

packages/react-core/src/components/Tabs/Tabs.tsx
redallen: Nit: no reason to set defaults to undefined. Just don't specify a default and they will be undefined.

packages/react-core/src/components/Tabs/Tabs.tsx
redallen: Why is this needed?

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -16,35 +16,52 @@ export enum TabsVariant {
...
+  /** Provides an accessible label for the tabs. Labels should be unique for each set of tabs that are present on a page. When variant is set to nav, this prop should be defined to differentiate the tabs from other navigation regions on the page. */
   'aria-label'?: string;
-  /** waits until the first "enter" transition to mount tab children (add them to the DOM) */
+  /** Waits until the first "enter" transition to mount tab children (add them to the DOM) */
   mountOnEnter?: boolean;
-  /** unmounts tab children (removes them from the DOM) when they are no longer visible */
+  /** Unmounts tab children (removes them from the DOM) when they are no longer visible */
   unmountOnExit?: boolean;
+  /* Modifies the tabs component padding/inset to visually match padding of other adjacent components.*/
+  inset?: 'none' | 'sm' | 'md' | 'lg' | 'xl' | '2xl';
```
redallen, line 50: Maybe something like our breakpoint mods helpers would be a better API here? The structure is looking similar.

packages/react-core/src/helpers/util.ts
```diff
@@ -266,3 +266,9 @@ export const toCamel = (s: string) => s.replace(/([-_][a-z])/gi, camelize);
  * Copied from exenv
  */
 export const canUseDOM = !!(typeof window !== 'undefined' && window.document && window.document.createElement);
+
+/**
+ *
+ * @param {string} s string to make first letter a capital
+ */
+export const capitalizeFirstLetter = (s: string) => s.charAt(0).toUpperCase() + s.slice(1);
```
redallen, line 274: There's a `capitalize` helper on line 9 already that does the exact same thing.

packages/react-core/src/components/Tabs/TabTitleText.tsx
```diff
@@ -0,0 +1,20 @@
...
+export interface TabTitleTextProps extends React.HTMLProps<HTMLSpanElement> {
+  /** Text to be rendered inside the tab button title. */
+  children?: React.ReactNode;
+  /** additional classes added to the tab title text */
+  className?: string;
+}
+
+export const TabTitleText: React.FunctionComponent<TabTitleTextProps> = ({
+  children,
+  className = '',
```
redallen, line 14: Nit: we don't have to set these to a default of empty string anymore on `v4`.

packages/react-core/src/components/Tabs/Tabs.tsx
tlabaj: This was preexisting.  Not sure why it is here.  I am looking into it.

packages/react-core/src/components/Tabs/TabTitleIcon.tsx
tlabaj: I followed the API we have in the drawer component for width.

packages/react-core/src/components/Tabs/Tabs.tsx
tlabaj: @jschuler this is what I was asking about.  I do not understand why we need this.

jschuler: good question, i did that a year ago :D let's remove it

packages/react-core/src/helpers/util.ts
tlabaj: I see that now.  I just read the description and thought it capitalized the whole string.  Thanks!  I will update description of original function.

packages/react-core/src/components/Tabs/Tab.tsx
jschuler: Can you also add that to your PR description? Looks like the components to use are not mentioned

packages/react-core/src/components/Tabs/TabTitleIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
+
+export interface TabTitleIconProps extends React.HTMLProps<HTMLSpanElement> {
+  /** Icon to be rendered inside the tab button title. */
+  children?: React.ReactNode;
```
jschuler, line 7: i'd make this required

packages/react-core/src/components/Tabs/TabTitleText.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
+
+export interface TabTitleTextProps extends React.HTMLProps<HTMLSpanElement> {
+  /** Text to be rendered inside the tab button title. */
+  children?: React.ReactNode;
```
jschuler, line 7: this should be required

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -133,6 +176,19 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
...
+  formatInsetString = (s: string) => (s === '2xl' ? '_2xl' : capitalize(s));
+
+  componentDidMount() {
+    window.addEventListener('resize', this.handleScrollButtons, false);
+    // call the handle resize function to check if scroll buttons should be shown
+    this.handleScrollButtons();
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('resize', this.handleScrollButtons, false);
```
jschuler, line 189: in componentDidMount you added the listener to window. They both should be the same

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -133,6 +176,19 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
     }
   };
 
+  // Format the inset prop string by capitalizing the first letter.  If the string prop is '2xl' append '_' to beginning of the string so we can key the correct modifier.
+  formatInsetString = (s: string) => (s === '2xl' ? '_2xl' : capitalize(s));
+
+  componentDidMount() {
+    window.addEventListener('resize', this.handleScrollButtons, false);
```
jschuler, line 183: Should we only listen to resize if `!isVertical`? Looks like it needs that condition to add the modifier

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -173,10 +248,12 @@ export class Tabs extends React.Component<TabsProps & OUIAProps, TabsState> {
             className={css(styles.tabsScrollButton, isSecondary && buttonStyles.modifiers.secondary)}
             aria-label={leftScrollAriaLabel}
             onClick={this.scrollLeft}
+            disabled={disableLeftScrollButton}
+            aria-hidden={disableLeftScrollButton}
           >
             <AngleLeftIcon />
           </button>
-          <ul className={css(styles.tabsList)} ref={this.tabList}>
+          <ul className={css(styles.tabsList)} ref={this.tabList} onScroll={this.handleScrollButtons}>
```
jschuler, line 256: similar question here, should we only call the function if `! isVertical`?

packages/react-core/src/components/Tabs/TabTitleIcon.tsx
tlabaj: We do not generally make children for similar components required. Is it because the title is required?

packages/react-core/src/components/Tabs/Tabs.tsx
tlabaj: Interesting this is how it was before.  I did not update here.  Should we have it on window on both?

redallen: Both should be on `window`.

packages/react-core/src/components/Tabs/TabTitleIcon.tsx
jschuler: without children this component does nothing right? so that's why i think it's not optional. when else should something be required? :)

mattnolting:
Screen Shot 2020-05-07 at 3 27 06 PM

Looks like `.pf-c-tabs__link` is being added twice

mattnolting:
Nav link example scrolls top on click `.pf-c-tab__link` click https://patternfly-react-pr-4146.surge.sh/documentation/react/components/tabs#using-the-nav-element

packages/react-core/src/components/Tabs/Tabs.tsx
redallen, line 50: I experimented with our breakpoint modifier API and I think it'd be a good fit and extendable to future modifiers:
```ts
export interface TabBreakpointMod {
  /** The attribute to modify  */
  modifier:
    | 'insetNone'
    | 'insetSm'
    | 'insetMd'
    | 'insetLg'
    | 'insetXl'
    | 'inset-2xl';
  /** The breakpoint at which to apply the modifier */
  breakpoint?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
}
```
...
```tsx
  /** Array of objects representing the various modifiers to apply to tabs at various breakpoints */
  breakpointMods?: TabBreakpointMod[];
```
...
```
formatBreakpointMods(breakpointMods, styles),
```
...
```tsx
<Tabs
  activeKey={activeTabKey}
  onSelect={this.handleTabClick}
  breakpointMods={[
    { breakpoint: 'md', modifier: 'insetSm' },
    { breakpoint: 'lg', modifier: 'insetLg' },
    { breakpoint: 'xl', modifier: 'inset_2xl' }
  ]}
  isBox={isBox}>
```

Let me know what you think!
```

File: packages/react-core/src/components/Tabs/Tabs.tsx (hunk 3 above)
Comment (dlabrecq) [reply packages/react-core/src/components/Tabs/Tabs.tsx]: Yes, just looking for code comments here

File: packages/react-core/src/components/Tabs/Tabs.tsx (hunk 6 above)
Comment (tlabaj) [reply packages/react-core/src/components/Tabs/Tabs.tsx:50]: Fair enough.  I see how this allows for a more flexible interface.  But, I do we think if we make this change we should probably go back and update Drawer for consistency.

Comment (redallen) [reply packages/react-core/src/components/Tabs/Tabs.tsx:50]: I agree. I think Drawer should use the same API and we should document it in the breaking change release notes.

Comment (tlabaj) [reply packages/react-core/src/components/Tabs/Tabs.tsx:50]: Drawer is still beta.  Do you mind opening the drawer issue.  I will update the Tabs API.

Comment (redallen) [reply packages/react-core/src/components/Tabs/Tabs.tsx:50]: I opened #4238

File: packages/react-core/src/components/Tabs/Tabs.tsx [hunk 14]
```diff
@@ -16,35 +16,52 @@ export enum TabsVariant {
...
+  /** Enables vertical tab styling */
+  isVertical?: boolean;
+  /** Aria-label for the left scroll button */
   leftScrollAriaLabel?: string;
-  /** Aria-label for the right Scroll Button */
+  /** Aria-label for the right scroll button */
   rightScrollAriaLabel?: string;
-  /** determines what tag is used around the Tabs. Use "nav" to define the Tabs inside a navigation region */
+  /** Determines what tag is used around the tabs. Use "nav" to define the tabs inside a navigation region */
   variant?: 'div' | 'nav';
```
mcoker, line 42: do we normally use `component` where you specify the type of element to be used?

PR #4147, opened 2020-04-28
Chore/4112 dropdownitem icon

### LINKED ISSUE #4112: Update DropdownItem to take `icon` prop
#4038 updated the DropdownToggle to accept an `icon` prop to display an icon within the Dropdown which is displayed before any component children (separate from the toggleIndicator icon which appears after the component children).

The DropdownItem should follow this same pattern, rather than wrapping an icon component with a `DropdownItemIcon` component.  See [example here](https://github.com/patternfly/patternfly-react/pull/4038#issuecomment-614833537).

evwilkin:
**What**: Closes #4112 

This PR updates the DropdownItem component to take an `icon` prop to mirror the pattern used in #4038.  This change removes the need for DropdownItemIcon component, that component is removed completely.


redallen:
Can you add this to your PR body's text?
```
## Breaking changes
- **Dropdown:** Remove DropdownItemIcon in favor of <Dropdown icon={<Icon />}
```

Issue #3679, opened 2020-02-06
Divider: Update components that have their own divider to use the divider component instead


mcoker:
We'll remove the component-specific dividers and update the examples to use the divider component

https://github.com/patternfly/patternfly-next/issues/2621

mcoker:
The core work is complete - https://github.com/patternfly/patternfly/pull/2944

evwilkin:
May be partially addressed in #4116

nicolethoen:
@tlabaj 
I imagine we would want to introduce as few breaking changes as possible? And just change the definitions of DropdownSeparator and other components with their own separators so they render a Divider. Rather than removing the whole DropdownSeparator component and requiring consumers to replace all instances with a Divider component.

nicolethoen:
@mcoker @tlabaj @mattnolting 

There is no implementation of a vertical divider variant. I don't need to make one in order to make the vertical dividers in the DataToolbar work, since the vertical separator is a variant of a DataToolbarItem. But I figure if people see them in the DataToolbar, they may want the variant to be accessible to them? 

I understand that the vertical Divider variant would always need to be placed in a Flex layout in order to work. That'd have to be documented in the example if implemented as part of addressing this issue. Unless there is still more work to be done or decisions to be made surrounding this on the core side?

mcoker:
Hey @nicolethoen there is an issue to add support for the vertical variant here that hasn't been completed - https://github.com/patternfly/patternfly-react/issues/3736

> I understand that the vertical Divider variant would always need to be placed in a Flex layout in order to work.

That's one way. The divider either needs to have some sort of height established, or be the child of an element with a height. Here are a couple of examples - https://codepen.io/mcoker/pen/wvKqpzy 

 A flex layout assigns an intrinsic height by default so being the child of a flex layout works.

PR #4154, opened 2020-04-30
chore(Divider): add vertical dividers and use Divider in DataToolbar

### LINKED ISSUE #3736: Add support for Vertical Divider
#### SUMMARY
Divider Component: https://www.patternfly.org/v4/documentation/react/components/divider

Currently the Divider component only supports horizontal line breaks. We use vertical dividers to visually break up toolbar actions and headers, it would be awesome if you could provide a vertical variation of the Divider. 

##### Header
Screen Shot 2020-02-12 at 11 07 15 AM

##### Toolbar
Screen Shot 2020-02-12 at 11 09 07 AM

##### DataList
Screen Shot 2020-02-12 at 11 10 11 AM

nicolethoen:
**What**: Closes #3736 and #3679

**Additional issues**:
I imagine we would want to introduce as few breaking changes as possible. And just change the implementations of any components with their own separators so they render a Divider. Much of that work has already been done.

So I don't plan to make more breaking changes unless we'd prefer I remove the whole DropdownSeparator component and other component specific separators (even if they all ultimately use the Divider component). This would require consumers to replace all instances with a Divider component.

### DIFF
--- a/packages/react-core/src/components/Toolbar/ToolbarItem.tsx
+++ b/packages/react-core/src/components/Toolbar/ToolbarItem.tsx
@@ -19,6 +19,7 @@ export enum ToolbarItemVariant {
 export interface ToolbarItemProps extends React.HTMLProps<HTMLDivElement> {
   /** Classes applied to root element of the data toolbar item */
   className?: string;
+  /** A type modifier which modifies spacing specifically depending on the type of item */
   variant?:
     | ToolbarItemVariant
     | 'bulk-select'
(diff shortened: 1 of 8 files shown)


nicolethoen:
> LGTM, but should new components be experimental?

It is a variant of the Divider component, not a new component.

packages/react-core/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -18,15 +19,15 @@ export enum ToolbarItemVariant {
 export interface ToolbarItemProps extends React.HTMLProps<HTMLDivElement> {
   /** Classes applied to root element of the data toolbar item */
   className?: string;
-  /** TODO: Support 'separator' element as a <Divider component="div" variant="horizontal" />. A type modifier which modifies spacing specifically depending on the type of item */
```
redallen, line 21: Add back comment for docs: `/** A type modifier which modifies spacing specifically depending on the type of item */`

PR #4155, opened 2020-04-30
fix(fillTemplate): ie11 compatibility

### LINKED ISSUE #4148: [Bug] ChipGroup is not working in IE11 browser

If I try to use chip group in an IE11 browser. A single chip without a chip group works fine though. The whole app breaks with the exception `Invalid character`
I tried digging into the code and this is what i think could be the root cause of the issue

Chip group uses `fillTemplate` function in utils
https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/helpers/util.ts#L115

This util function creates a function on the fly which is using template literals to parse a string. My guess is babel doesn't transpile this code using a polyfill and this is why it breaks in IE11.

I also tried using this function in isolation and i could reproduce the issue

This is a bug. I would say this is critical for us and not a blocker. The workaround is not to use the filter chips on IE11 by detecting the browser.

Latest release of the patternfly and patternfly-react.

redallen:
**What**: Closes #4148

**Additional issues**:

### DIFF
--- a/packages/react-core/src/helpers/util.ts
+++ b/packages/react-core/src/helpers/util.ts
@@ -113,7 +113,7 @@ export function sideElementIsOutOfView(container: HTMLElement, element: HTMLElem
  * @returns {string} The template string literal result
  */
 export function fillTemplate(templateString: string, templateVars: any) {
-  return templateString.replace(/\${(.*)}/g, (_, match) => templateVars[match] || '');
+  return templateString.replace(/\${(.*?)}/g, (_, match) => templateVars[match] || '');
 }
 
 /**
--- a/packages/react-core/src/helpers/util.ts
+++ b/packages/react-core/src/helpers/util.ts
@@ -107,13 +107,13 @@ export function sideElementIsOutOfView(container: HTMLElement, element: HTMLElem
  *    const result = fillTemplate(templateString, templateVars);
  *    // "My name is Jon Dough"
  *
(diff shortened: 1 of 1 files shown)


packages/react-core/src/helpers/util.ts
```diff
@@ -107,14 +107,13 @@ export function sideElementIsOutOfView(container: HTMLElement, element: HTMLElem
...
- * @param {object} templateString  The string passed by the consumer
+ * @param {string} templateString  The string passed by the consumer
  * @param {object} templateVars The variables passed to the string
  *
  * @returns {string} The template string literal result
  */
 export function fillTemplate(templateString: string, templateVars: any) {
-  const func = new Function(...Object.keys(templateVars), `return \`${templateString}\`;`);
-  return func(...Object.values(templateVars));
+  return templateString.replace(/\${(.*?)}/g, (_, match) => templateVars[match] || '');
```
jschuler, line 116: would be good to have some tests for this, and verify before and after is the same

redallen: We have this test: https://github.com/patternfly/patternfly-react/blob/v4/packages/react-core/src/helpers/__tests__/util.test.js#L217

Also we have snapshots for our 3 components that use this helper that remain unchanged. What other tests would you like to see?

jschuler: ah good, all set then

PR #4165, opened 2020-05-01
feat(Label): refactor for updated design and features

### LINKED ISSUE #3688: Label: updates to support new design
Placeholder for https://github.com/patternfly/patternfly-next/issues/2656. Will likely result in changes to the current label component elements.

kmcfaul:
**What**: Closes #3688

The `onClose` prop, if present, will add the close button to the label, and the `href` prop, if present, will change the <span> to an <a> element.

## Breaking changes
1. Default color changed to grey. Design also adjusted.

### DIFF
--- a/packages/react-core/src/components/Label/Label.tsx
+++ b/packages/react-core/src/components/Label/Label.tsx
@@ -17,12 +17,10 @@ export interface LabelProps extends React.HTMLProps<HTMLSpanElement> {
   icon?: React.ReactNode;
   /** Close click callback for removable labels. If present, label will have a close button. */
   onClose?: (event: React.MouseEvent) => void;
-  /** Aria label of the close button. */
-  closeBtnAriaLabel?: string;
-  /** Id of the close button. */
-  closeBtnId?: string;
-  /** Text id of the text within the label for the close button aria-labelledby field. */
-  closeBtnTextId?: string;
+  /** Node for custom close button. */
+  closeBtn?: React.ReactNode;
+  /** Additional properties for the default close button. */
+  closeBtnProps?: any;
   /** Href for a label that is a link. If present, the label will change to an anchor element. */
   href?: string;
 }
@@ -44,22 +42,16 @@ export const Label: React.FunctionComponent<LabelProps> = ({
   variant = 'filled',
   icon,
   onClose,
-  closeBtnAriaLabel,
-  closeBtnId,
-  closeBtnTextId,
+  closeBtn,
+  closeBtnProps,
   href,
   ...props
 }: LabelProps) => {
   const Component = href ? 'a' : 'span';
-  const closeBtn = (
-    <Button
-      type="button"
-      variant="plain"
-      onClick={onClose}
-      aria-label={closeBtnAriaLabel}
-      id={closeBtnId}
-      aria-labelledby={`${closeBtnId} ${closeBtnTextId}`}
-    >
+  const button = closeBtn ? (
+    closeBtn
+  ) : (
+    <Button type="button" variant="plain" onClick={onClose} {...closeBtnProps}>
       <TimesIcon />
     </Button>
   );
@@ -73,7 +65,7 @@ export const Label: React.FunctionComponent<LabelProps> = ({
         {icon && <span className={css(styles.labelIcon)}>{icon}</span>}
         {children}
       </Component>
-      {onClose && closeBtn}
+      {onClose && button}
     </span>
   );
 };
--- a/packages/react-core/src/components/Label/Label.tsx
+++ b/packages/react-core/src/components/Label/Label.tsx
@@ -9,13 +9,21 @@ export interface LabelProps extends React.HTMLProps<HTMLSpanElement> {
   children: React.ReactNode;
   /** Additional classes added to the label. */
   className?: string;
+  /** Color of the label. */
   color?: 'blue' | 'cyan' | 'green' | 'orange' | 'purple' | 'red' | 'grey';
+  /** Variant of the label. */
   variant?: 'outline' | 'filled';
+  /** Icon added to the left of the label text. */
   icon?: React.ReactNode;
+  /** Close click callback for removable labels. If present, label will have a close button. */
   onClose?: (event: React.MouseEvent) => void;
+  /** Aria label of the close button. */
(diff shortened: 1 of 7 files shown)


packages/react-core/src/components/Label/examples/Label.md
```diff
@@ -7,15 +7,251 @@ propComponents: ['Label']
...
+);
+```
+
+```js title=Outline
 import React from 'react';
 import { Label } from '@patternfly/react-core';
+import { InfoCircleIcon } from '@patternfly/react-icons';
 
-SimpleLabel = () => (
+FilledLabels = () => (
```
jenny-s51, line 132: Rename to be `OutlineLabels`

packages/react-core/src/components/Label/examples/Label.md
```diff
@@ -7,15 +7,251 @@ propComponents: ['Label']
...
+    <br />
+    <Label variant="outline" color="blue">
+      Blue
+    </Label> <Label variant="outline" color="blue" icon={<InfoCircleIcon />}>
+      Blue icon
+    </Label> <Label variant="outline" color="blue" onClose={Function.prototype}>
+      Blue removeable
+    </Label> <Label variant="outline" color="blue" icon={<InfoCircleIcon />} onClose={Function.prototype}>
+      Blue icon removeable
+    </Label> <Label variant="outline" color="blue" href="#">
```
jenny-s51, line 160: Would it make sense to set the `href`s to `#filled` or `#outline` for the link labels in both examples? At the moment when I click an outlined label with a link, it takes me back up to the top of the page.

packages/react-core/src/components/Label/Label.tsx
```diff
@@ -1,16 +1,79 @@
...
+  /** Variant of the label. */
+  variant?: 'outline' | 'filled';
+  /** Icon added to the left of the label text. */
+  icon?: React.ReactNode;
+  /** Close click callback for removable labels. If present, label will have a close button. */
+  onClose?: (event: React.MouseEvent) => void;
+  /** Aria label of the close button. */
+  closeBtnAriaLabel?: string;
+  /** Id of the close button. */
+  closeBtnId?: string;
```
redallen, line 23: I already know someone is going to want to customize the button they put in here... Can we just make the close button a `React.ReactNode` that the consumer can pass in, but has a default of:
```jsx
    <Button
      type="button"
      variant="plain"
      onClick={onClose}
      aria-label={'some-default'}
      id={'some-default'}
      aria-labelledby={`'some-default'`}
    >
      <TimesIcon />
    </Button>
```
if they pass `onClose`? It also cleans up having to support the `closeBtnAriaLabel`, `closeBtnId`, and `closeBtnTextId` props.

packages/react-core/src/components/Label/__tests__/Label.test.tsx
```diff
@@ -5,11 +5,96 @@ import { Label } from '../Label';
...
+  expect(view).toMatchSnapshot();
+  const outline = shallow(
+    <Label onClose={jest.fn()} variant="outline">
+      Something
+    </Label>
+  );
+  expect(outline).toMatchSnapshot();
+});
+
+test('label with blue color', () => {
```
redallen, line 34: Maybe a forEach loop will keep these DRY?
```jsx
['blue', 'cyan', 'green', 'orange', 'purple', 'red', 'grey'].forEach(color =>
  test(`label with ${color} color`, () => {
    const view = shallow(<Label color={color}}>Something</Label>);
    const outline = shallow(
      <Label color={color} variant="outline">
        Something
      </Label>
    );
    expect(outline).toMatchSnapshot();
  });
)
```

packages/react-core/src/components/Label/Label.tsx
kmcfaul: I'll add a node property for a full custom button, and will condense the various default button props into a general 'closeBtnProps' object. That way users can pass in whatever labelling/id they want without needing multiple props or to define a full button.

packages/react-core/src/components/Label/__tests__/Label.test.tsx
kmcfaul: Much cleaner, will update!

packages/react-core/src/components/Label/Label.tsx
redallen: Works for me!

redallen:
CI is sad:
```
http://localhost:9000/documentation/react/components/label/filled
@patternfly/react-docs: ================= Violations =================
@patternfly/react-docs: {
@patternfly/react-docs:   "data": "",
@patternfly/react-docs:   "id": "button-has-visible-text",
@patternfly/react-docs:   "impact": "critical",
@patternfly/react-docs:   "message": "Element does not have inner text that is visible to screen readers",
@patternfly/react-docs:   "relatedNodes": [
@patternfly/react-docs:     {
@patternfly/react-docs:       "html": "<button class=\"pf-c-button pf-m-plain\" type=\"button\" data-ouia-component-type=\"PF4/Button\" data-ouia-safe=\"true\">"
@patternfly/react-docs:     }
@patternfly/react-docs:   ]
@patternfly/react-docs: }
```

kmcfaul:
I didn't change the documentation, so it has to be the change to the default button. Does it need an aria label maybe?

redallen:
@kmcfaul Yes, I think you need to set some default `closeBtnProps`

packages/react-core/src/components/Label/Label.tsx
```diff
@@ -1,16 +1,73 @@
...
+export const Label: React.FunctionComponent<LabelProps> = ({
+  children,
+  className = '',
+  color = 'grey',
+  variant = 'filled',
+  icon,
+  onClose,
+  closeBtn,
+  closeBtnProps = {
+    'aria-label': 'label-close-button'
```
redallen, line 47: Try:
```
closeBtnProps = {
    'aria-label': 'label-close-button',
    ...closeBtnProps
}
```
or else when a consumer specifies just one property other than `aria-label` the `aria-label` won't be set. If that syntax doesn't work, on line 56 `{...{'aria-label': 'label-close-button', ...closeBtnProps}}` should.

PR #4179, opened 2020-05-04
fix(tabs): properly check children before setting hidden

### LINKED ISSUE #4074: Tabs: Uncaught TypeError: Cannot set property 'hidden' of null
In Cost Management, we're seeing the following errors in the browser console. I believe this is due to the recent change to PatternFly's tab ref code.

Screen Shot 2020-04-14 at 1 14 16 PM

Didn't notice this issue while testing previous milestone builds, but it's a blocker for production. If a user double clicks on the same tab, the page goes blank.

chrome-capture (1)

In Cost Management, we use a set of tabs, which is separated from its content. The tabs appear in the header, while the content is in a different container. See the "Separate content" example in the patternFly docs.

We lazily load tabs based on the active index. That is, we return null as the tab content if the tab is not active. Otherwise, all API requests are invoked when entering the page, which is a big performance hit.

redallen:
**What**: Closes #4074 . Fixes lint issue in NotificationDrawerGroup.tsx.

**Additional issues**: Verified this in cost management thanks to @dlabrecq !

### DIFF
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawerGroup.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawerGroup.tsx
@@ -34,7 +34,10 @@ export const NotificationDrawerGroup: React.FunctionComponent<NotificationDrawer
   title,
   ...props
 }: NotificationDrawerGroupProps) => (
-  <section {...props} className={css(styles.notificationDrawerGroup, isExpanded && styles.modifiers.expanded)}>
+  <section
+    {...props}
+    className={css(styles.notificationDrawerGroup, isExpanded && styles.modifiers.expanded, className)}
+  >
     <h1>
       <button
         className={css(styles.notificationDrawerGroupToggle)}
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/NotificationDrawer/NotificationDrawerGroup.tsx
```diff
@@ -34,7 +34,10 @@ export const NotificationDrawerGroup: React.FunctionComponent<NotificationDrawer
   title,
   ...props
 }: NotificationDrawerGroupProps) => (
-  <section {...props} className={css(styles.notificationDrawerGroup, isExpanded && styles.modifiers.expanded)}>
```
tlabaj, line 37: What does this have to do with the tabs?  Is it just fixing a linting issue?

redallen: Yep! `className` should clearly be added there, but was removed at some point.

PR #4181, opened 2020-05-04
fix(page): make tabIndex optional

### LINKED ISSUE #4180: `tabIndex` in `<Page>` should be an optional prop
See https://github.com/patternfly/patternfly-react/blob/v4/packages/react-core/src/components/Page/Page.tsx#L159

This creates an issue if scrolling occurs on a child element instead of `<main>` where the keyboard cannot be used to scroll the scrollable element.  The workaround is to add `tabIndex={-1}` to the scrollable element; however, it seems this should not be necessary as PatternFly should provide the option to not include `tabIndex` in `<Page>`.

redallen:
**What**: Closes #4180

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -27,7 +27,9 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   /** Sets the value for role on the <main> element */
   role?: string;
   /** an id to use for the [role="main"] element */
-  mainContainerId?: string;
+  mainContainerId?: string; 
+  /** tabIndex to use for the [role="main"] element */
+  mainTabIndex?: number;
   /**
    * If true, manages the sidebar open/close state and there is no need to pass the isNavOpen boolean into
    * the sidebar component or add a callback onNavToggle function into the PageHeader component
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/Page/Page.tsx
```diff
@@ -28,6 +28,8 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   role?: string;
   /** an id to use for the [role="main"] element */
   mainContainerId?: string;
+  /** tabIndex to use for the [role="main"] element */
+  mainTabIndex?: number;
```
seanforyou23, line 32: Should this be `number | null`? To unset the attribute from the DOM users have to supply `null` right?

redallen: `undefined` works as well I believe. I can verify this in a few hours.

redallen: `undefined` (or just not passing the prop) also works to unset the attribute from the DOM. `null` can be passed as well if strict null checks are disabled by the consumer.

seanforyou23: I don't see this working as you describe in https://patternfly-react-pr-4181.surge.sh/documentation/react/components/page or locally within react-integration.

Unless I provide ` mainTabIndex={null}` - the tabindex attr is set on <main> with a value of -1, any idea what I'm missing?

redallen: TLDR; I updated the PR with your suggestion.

I'm the one missing something, not you! This is due to how default properties are set and I've demoed it in https://codesandbox.io/s/upbeat-bogdan-0fyvl . Basically, you can't set a prop explicitly to `undefined` because the way default props work is setting the property if it `=== undefined` rather than `!Object.hasOwnProperty(this, propertyName)`. This is just another case of Javascript sucking:
image
I would expect the call `test({a: undefined})` to print `undefined`, but in fact I don't think it's possible at all.

PR #4190, opened 2020-05-05
Breaking/alert

### LINKED ISSUE #4082: Alert - updates to support new design
https://github.com/patternfly/patternfly/pull/2921/

the `pf-c-alert__action` element is now only used to hold the close button. There is now a `pf-c-alert__action-group` element as the last element in the alert component that holds optional actions, and the actions should only be inline link buttons - `<Button variant="link" isInline>`

jschuler:
**What**: Closes #4082 

Alerts can now contain both the close button and one or more action links.

## Breaking changes
- **Alert**: If you were previously using the `action` prop for the close button, use the `actionClose` prop instead
- **Alert**: If you were previously using the `action` prop for a link button, use the `actionLinks` prop instead


packages/react-core/src/components/Avatar/examples/Avatar.md
```diff
@@ -15,5 +15,11 @@ import React from 'react';
 import { Avatar } from '@patternfly/react-core';
 import avatarImg from './examples/avatarImg.svg';
 
-<Avatar src={avatarImg} alt="avatar"></Avatar>
+class AvatarExample extends React.Component {
```
jschuler, line 18: made this change because i noticed the codesandbox wasn't working before. should probably go through and update all the examples that use the previous syntax @redallen ?

redallen: All you need is to name the example like this @jschuler :
```jsx
AvatarExample = () => <Avatar src={avatarImg} alt="avatar"></Avatar>
```

jschuler: codesandbox would work (after adding `const` in front), but our docs would not
```
SyntaxError: Unexpected token (1:8)
1 : return (const AvatarExample = () => <Avatar src={avatarImg} alt="avatar"></Avatar>)
```

Issue #3866, opened 2020-03-03
Form: add support for form label icon for use with popovers


mcoker:
Placeholder for the work from https://github.com/patternfly/patternfly-next/issues/2703

mcoker:
The core work for this is complete - https://github.com/patternfly/patternfly/pull/2766

The form has a couple of new structural elements. `.pf-c-form__group-label` and `.pf-c-form__group-control`. The elements that go with the label (including the new label icon/popover) go in `__group-label` and the form control elements (text input, textarea, etc) go in `__group-control`. The react component could likely be updated so that we put the children of <FormGroup> in `__group-control` , and whatever is passed as the `label` (and the new help icon/popover if that's specified) in `__group-label.`

The new icon/popover is triggered via a new element, a button with class `.pf-c-form__group-label-help`.

`pf-m-no-padding-top` now applies to `.pf-c-form__group-label` instead of `.pf-c-form__label`

`pf-m-inline` applies to `.pf-c-form__group-control` instead of `.pf-c-form__group`

Also we have removed `.pf-m-border` and `.pf-c-form__section` from the docs as they aren't defined anywhere in the CSS, so we need to confirm whether those can be deleted in the react component if they do exist.

PR #4195, opened 2020-05-06
feat(form): add support for form label icons

seanforyou23:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/3866

This PR adds a new prop `labelIcon` which can be used to supply an icon for a label inside a form group. It also applies some structural changes to FormGroup to bring it up to speed with changes from core.

Added `aria-describedby` to the [host button element](https://github.com/patternfly/patternfly-react/pull/4195/files#diff-9b8f908953d552c9a8e583f13013ec22R73) to provide more context for screen reader users. Also, by including some uniquely identifiable name of the field in the `aria-label` value, users can differentiate between many help button icons in a view. So, "More info _for field **x**_" instead of just "More info". Can we review also these changes, and if we like them can we add them to the core examples?

**Additional issues**: https://github.com/patternfly/patternfly/pull/2766


mturley:
This is a nice feature! A few times I have had to just make my own `label` as a node instead of a string, with a button in it. But that never quite felt right.

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -60,23 +63,30 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
...
-      {children}
-      {validated === ValidatedOptions.error && helperTextInvalid
-        ? inValidHelperText
-        : validated !== ValidatedOptions.error && helperText
-        ? validHelperText
-        : ''}
+
+      <div className={css(styles.formGroupControl, isInline && styles.modifiers.inline)}>
+        {children}
+        {validated === ValidatedOptions.error && helperTextInvalid
```
redallen, line 84: I'm not a fan of nested ternaries. Do you think it'd be possible to clean this up into a `helperText` variable?

redallen:
Conflicts from #4247

PR #4196, opened 2020-05-06
feat(tabs): allow hiding tab buttons

### LINKED ISSUE #4095: tabButton: misuse of refs causes tab conents to not appear when using separate content
In our app (OCM), we use the <Tabs> component from PatternFly React (currently on     `"@patternfly/react-core": "^3.134.2"`).
Most of the tabs appear on mount, but some tabs take a few seconds before we know if we can show them or not.

We also use the option to have the tabs themselves and the tab contents in different components.

This causes an issue where in some cases, a tab that appears after mount won't be usable - switching to it shows an empty space where the tab contents should be. Further investigation showed the tab content `<section>` element had the `hidden` property, even tho the tab was selected.

Then I found out that this happens because the `tabContentRef` refers to the button, instead of the tab itself.

Looks like that in TabButton.tsx, line 22, 
https://github.com/patternfly/patternfly-react/blob/v4/packages/react-core/src/components/Tabs/TabButton.tsx#L22

Instead of reusing the name `tabContentRef`, you use just `ref` - this makes React allocate this ref to the TabButton itself if it wasn't allocated yet to the section.

This is a small reproducer with two tabs. The second tab appears 500ms after mount. Note how clicking on it doesn't switch between first/second tab contents.

https://codesandbox.io/s/thirsty-leftpad-yyz0b?file=/index.js

While a workaround is theoretically possible, I haven't figured out one yet. Need to find a way to ensure the tab contents somehow claim the ref before the tab buttons...

redallen:
**What**: Closes #4095

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -15,6 +15,8 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   tabContentId?: string | number;
   /** child reference for case in which a TabContent section is defined outside of a Tabs component */
   tabContentRef?: React.RefObject<any>;
+  /** whether to render the tab or not */
+  isHidden?: boolean;
 }
 
 export const Tab: React.FunctionComponent<TabProps> = () => null;
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -246,11 +246,21 @@ class Tabs extends React.Component<TabsProps & InjectedOuiaProps, TabsState> {
           </button>
           <ul className={css(styles.tabsList)} ref={this.tabList} onScroll={this.handleScrollButtons}>
             {React.Children.map(children, (child: any, index) => {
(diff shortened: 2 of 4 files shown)


redallen:
@dlabrecq The `isHidden` prop is intended to be used like in the [Toggled separate content example:](https://patternfly-react-pr-4196.surge.sh/documentation/react/components/tabs#toggled-separate-content)
```jsx
<Tab title="Tab item 1" tabContentId="refTab1Section" tabContentRef={this.contentRef1} />
<Tab title="Tab item 2" tabContentId="refTab2Section" tabContentRef={this.contentRef2} isHidden={showTab2} />
<Tab title="Tab item 3" tabContentId="refTab3Section" tabContentRef={this.contentRef3} />
```

We can't omit the TabContent component (either on the component side or example side) because we still need to capture the refs or the current implementation of separate Tabs and TabContent crashes. Once #4179 is merged we could do something like `{!isTab2Hidden && 'Tab 2 section'}`.

tlabaj:
@redallen can you open up another PR to modify the demo app and integration test please.

redallen:
Done! I opened #4205

PR #4198, opened 2020-05-06
feat(Select): Add support for 0 and all badges

### LINKED ISSUE #4186: Add "0" and "All" selected badging for Select
OpenShift uses something similar to select, but it has badging for "0" and "All" selected. Right now, Select only supports badging for 1+ items selected. OpenShift folks were wondering if PatternFly would be willing to support the "0" and "All" badging cases.

None:
Screen Shot 2020-05-05 at 9 30 05 AM

All:
Screen Shot 2020-05-05 at 9 30 16 AM

This is an enhancement. We're going to look into passing a React node into the placeholder text so we can work around this limitation per @tlabaj's guidance for now. However, it would be great if PatternFly supported this more formally.

OpenShift 4.5+

rebeccaalpert:
Added props users can pass in to trigger badging for "none selected" and "all selected" use cases.

Fixes #4186.


packages/react-core/src/components/Select/Select.tsx
```diff
@@ -470,6 +479,27 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
...
+      let badgeText = null;
+      if (!isCheckboxSelectionBadgeHidden && selections && Array.isArray(selections)) {
+        if (enableZeroBadge && selections.length === 0) {
+          badgeText = 0;
+        }
+        if (selections.length > 0) {
+          badgeText = selections.length;
+        }
+        if (enableAllBadge && selections.length === maxNumSelections) {
+          badgeText = 'All';
```
tlabaj, line 492: this text needs to set by prop fro localization.

rebeccaalpert: Sure, I'll refactor.

rebeccaalpert: @tlabaj - I pushed a refactored version; let me know if you have other feedback.

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -522,10 +537,12 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
                   {toggleIcon && <span className={css(styles.selectToggleIcon)}>{toggleIcon}</span>}
                   <span className={css(styles.selectToggleText)}>{placeholderText}</span>
                   {!isCheckboxSelectionBadgeHidden &&
-                    selections &&
-                    (Array.isArray(selections) && selections.length > 0) && (
+                    (customBadgeText !== null ||
```
dlabrecq, line 540: What happens with an empty string or negative numbers? Perhaps we can add some minimal validation for that?

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -470,6 +475,16 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
       );
     }
 
+    const generateSelectedBadge = () => {
+      if (customBadgeText !== null) {
+        return customBadgeText;
+      }
+      if (Array.isArray(selections) && selections.length > 0) {
```
redallen, line 482: You can use the `hasAnySelections` variable instead of `Array.isArray(selections) && selections.length > 0`.

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -522,10 +537,12 @@ class Select extends React.Component<SelectProps & InjectedOuiaProps, SelectStat
                   {toggleIcon && <span className={css(styles.selectToggleIcon)}>{toggleIcon}</span>}
                   <span className={css(styles.selectToggleText)}>{placeholderText}</span>
                   {!isCheckboxSelectionBadgeHidden &&
-                    selections &&
-                    (Array.isArray(selections) && selections.length > 0) && (
+                    (customBadgeText !== null ||
+                      (selections && Array.isArray(selections) && selections.length > 0)) && (
```
redallen, line 541: You can use the `hasAnySelections` variable instead of `Array.isArray(selections) && selections.length > 0`.

packages/react-core/src/components/Select/Select.tsx
redallen, line 540: No need to add the `customBadgeText !== null ||` check, you already do that in `generateSelectedBadge`.

redallen: A consumer could always pass the string "-1". We're just rendering a passed prop, so I don't think we need to add validation.

rebeccaalpert: It's to render the surrounding div and span; otherwise you just get a string back.

rebeccaalpert:
Addressed Zack's comments.

rebeccaalpert:
@redallen - I rolled back one of the changes in response to your comments since it was causing a build error.

packages/react-core/src/components/Select/Select.tsx
tlabaj: I agree, I don't think we need to add validation here.

PR #4202, opened 2020-05-06
fix(pagination): update to match core

### LINKED ISSUE #4189: Bottom pagination updates
This is a placeholder for https://github.com/patternfly/patternfly/pull/3050. This was originally just an update to the design, but it ended up needing markup changes.

* Core's bottom pagination is now `.pf-m-bottom` instead of `.pf-m-footer` to align with react
* Any instance where we use bottom pagination, it should not be wrapped in a toolbar, and should just come after the thing it is pagination for (data-list, table, etc)
* The bottom pagination on mobile is now "sticky" to the bottom of the container it's in, so it will float on top of the table, data-list, etc that it comes after.
  * We have added a variation to disable that called `.pf-m-static`, so we will need a prop for that.
* The nav buttons in pagination (first, prev, next, last) are now wrapped in `<div class="pf-c-pagination__nav-control>` with a modifier for each `pf-m-first`, `pf-m-prev`, `pf-m-next`, `pf-m-last`.

redallen:
**What**: Bump core and close #4189

**Additional issues**: Looks like `pf-m-prev` and `pf-m-next` don't exist. I didn't diff the HTML yet, but this looks close.

## Breaking changes
- **Pagination:** Remove obsolete 'left' and 'right' variants. These should be replaced with 'top' (default) or 'bottom'.


mcoker:
Pagination looks good to me! We aren't styling `.pf-m-prev/next` so it's fine to add those when we introduce styles.

Can we:
1. add the bottom variation to the bottom pagination on https://patternfly-react-pr-4202.surge.sh/documentation/react/demos/bulkselecttable
1. move the bottom pagination out of the toolbar on https://patternfly-react-pr-4202.surge.sh/documentation/react/demos/tablecolumnmanagementdemo

And I just noticed these, they're outside of the scope of the PR. Let me know if you'd like me to open a separate issue.
1. Looks like `variant` accepts 4 values. `top` - default, `bottom` - applies `.pf-m-bottom`. What do left and right do?
1. Curious why we set the inline style `--pf-c-pagination__nav-page-select--c-form-control--width-chars:2;`? That's set in core by default. You would only define that if you want to change how many characters are shown in the nav input. That was added here - https://github.com/patternfly/patternfly-react/pull/3534

redallen:
1. Done
2. Done

I'd consider them inside the scope of this PR.
1. Not sure, I think they were copied from Core and some point and are now obsolete. I've removed them.
2. This this is a fix for #3530 which allows pagination inputs with more than 2 digits to grow:


You can see this issue for numbers greater than 2 digits in [Core today:](http://patternfly-v4.surge.sh/documentation/core/components/pagination/top)
image

mcarrano:
This looks good. I only ran into one issue that may or may not be relevant. I was testing this in the context of this demo https://patternfly-react-pr-4202.surge.sh/documentation/react/demos/bulkselecttable  Seems to be the only place we have a demo of bottom pagination. Clicking the next or back buttons after I shrink to mobile size has some unexpected behavior (scrolls to bottom of web page). This may just be an artifact of the demo sitting inside the webpage rather than opening in a new window, but just thought I'd bring it up.

mcoker:
Noticed one other thing. The options menu when the bottom pagination isn't `isStatic` should probably use the top variation so the menu expands upward instead of down out of the viewport.

Screen Shot 2020-05-07 at 10 34 17 AM
Screen Shot 2020-05-07 at 10 33 51 AM

The only caveat to that is that the pagination is only sticky to the table, so if there is content after the table and you scroll down to it, the pagination is no longer sticky. So at that point, since there is room for the options menu to expand down, it would be odd if the options menu continued to expand upward when you click on it. WDYT @mcarrano?

Ideally the options menu would know if there is enough room in the viewport height to expand down, and if not, it expands up instead.

redallen:
@mcarrano That is an effect of the Table's height changing in between pages as it loads new data It is 2234px tall on first page but 2258px on second page (and the 24px difference is much more pronounced on mobile viewports with wrapping). We can write some code to scroll to the bottom or top once the next page has finished loading. I'm not sure what the intended behavior is for tables this tall.

mcarrano:
@mcoker Good catch about the menu direction. I agree that in most cases I'd expect the menu to open up. Yes, ideally the component would detect whether there is enough space and automatically switch direction. But is that's not possible, then I'd default to opening up as this will cover the majority of cases where a table is used.

@redallen Thanks for the explanation about the scroll position of the page in the demo. Not sure if this will be a problem in a real application, so I'd leave this alone for now.

redallen:
I've updated the default menu position to be pointed up in https://patternfly-react-pr-4202.surge.sh/documentation/react/demos/bulkselecttable . Please re-review @mcoker and @mcarrano !

mcoker:
Hmm... looks like you're just applying the top variation on the options menu manually, so it would be up to the user to specify that anywhere they're using sticky bottom pagination? I was thinking we would do that by default if they're using the sticky bottom pagination, like `variant == PaginationVariant.bottom && !isStatic && dropDirection = 'up'`. Can you think of a reason not to do that?

If we're leaving it to the consumer to add that, we'll also need to update https://patternfly-react-pr-4202.surge.sh/documentation/react/demos/tablecolumnmanagementdemo/column-management-action and anywhere else there is a bottom pagination.

redallen:
Rather than update many examples, I've updated the component with the following logic:
```js
const dropDirection = dropDirectionProp || (variant === 'bottom' && !isStatic ? 'up' : 'down');
```

Just waiting for CI to finish the latest release so I can push and resolve these merge conflicts.

redallen:
Please rereview again @mcarrano and @mcoker !

PR #4209, opened 2020-05-07
chore(Table): apply core changes to Table

### LINKED ISSUE #3670: Table: better handle white-space
This is a placeholder for the work in https://github.com/patternfly/patternfly-next/issues/2205 and https://github.com/patternfly/patternfly-next/issues/2325 which may result in changes to the props we use to manipulate white space, word wrapping, etc.

nicolethoen:
**What**: Closes #3670 

- Add `TableText` component to help display text better in Grid layouts. Added example to demonstrate use.
- Replace buttons with `.pf-c-table__button` in the sortable header and the compound expansion toggle. Wrap `.pf-c-table__button` contents in `.pf-c-table__button-content`.
- Adds new decorators (`nowrap`, `truncate`, `fitContent`, `breakWord`) that can be passed to `transforms` and `cellTransforms` alongside the previously existing `wrappable` decorator. Using these new decorators apply new modifiers. Created new example to demonstrate these new modifiers in action.
- updated jest and integration tests, as well as the demos


packages/react-table/src/components/Table/TableText.tsx
```diff
@@ -2,25 +2,41 @@ import * as React from 'react';
...
 export const TableText: React.FunctionComponent<TableTextProps> = ({
   children = null,
   className = '',
-  component = 'span',
+  variant = 'span',
+  wrapModifier = null,
   ...props
 }: TableTextProps) => {
-  const Component: any = component;
+  const Component: any = variant;
```
seanforyou23, line 36: Are we able to use the `TableTextVariant | 'span' | 'div'` type here instead of `any`?

packages/react-table/src/components/Table/utils/decorators/compoundExpand.tsx
```diff
@@ -32,9 +31,14 @@ export const compoundExpand: ITransform = (
...
-      <Button variant="link" onClick={onToggle} aria-expanded={props.isOpen} aria-controls={props.ariaControls}>
+      <button
+        className={css(styles.tableButton)}
+        onClick={onToggle}
+        aria-expanded={props.isOpen}
+        aria-controls={props.ariaControls}
+      >
         {title}
-      </Button>
+      </button>
```
dlabrecq, line 41: Curious as to why we cannot use our own button here?

nicolethoen: This is probably a better question for @mattnolting since it's more css related. but we are using a `.pf-c-table__button` instead of `.pf-c-button` in the case of the compound expansion toggle.

mattnolting: @dlabrecq Although `pf-c-table__button` shares some visual styling with `pf-c-button`,  it's behavior is very different in `sort` and `compound-expansion__toggle` contexts.

redallen:
Once @mcoker or @mattnolting review this we can take it in.

mcarrano:
@nicolethoen following on my last comment, here are some design mockups that show how this should behave. Let me know if you have any questions: https://marvelapp.com/6bhc03d/screen/66159925

mcarrano:
@nicolethoen Took another look. I'm seeing the truncation now on the text in the cell for column 1, but not on the header when it truncates.

PR #4223, opened 2020-05-08
replace Toolbar with PageHeaderTools in page header

### LINKED ISSUE #4106: Toolbar layout needs to be removed from the Page component (on the black header) and on all demos that use the Page component.
This should be done to match Core. Currently core and react are out of sync.
I believe a flex layout can be used instead.
There may also be a header-tools component that can be used.

jschuler:
**What**: Closes #4106 

This PR updates the page demos to use the new `PageHeaderTools`/`PageHeaderToolsGroup`/`PageHeaderToolsItem` components

https://patternfly-react-pr-4223.surge.sh/documentation/react/demos/pagelayout

## Breaking changes
**PageHeader**: `avatar` prop was removed. Add the avatar to the `PageHeaderTools` component instead (which is passed into `PageHeader` via the `headerTools` prop.
**PageHeader**: `toolbar` prop was removed. Use the `headerTools` prop instead. Also, if you previously used the `Toolbar`, `ToolbarGroup`, or `ToolbarItem` components for the header, replace them with the `PageHeaderTools`, `PageHeaderToolsGroup`, and `PageHeaderToolsItem` components.


packages/react-core/src/demos/CardView/examples/CardView.md
```diff
@@ -621,50 +618,14 @@ class CardViewBasic extends React.Component {
...
     const Header = (
       <PageHeader
         logo={<Brand src={imgBrand} alt="Patternfly Logo" />}
-        toolbar={PageToolbar}
-        avatar={<Avatar src={imgAvatar} alt="Avatar image" />}
+        headerTools={
+          <PageHeaderTools>
+            <Avatar src={imgAvatar} alt="Avatar image" />
+          </PageHeaderTools>
+        }
```
redallen, line 628: Some breaking change notes describing this would be great!

packages/react-core/src/components/Page/examples/Page.md
```diff
@@ -6,7 +6,7 @@ typescript: true
 propComponents: ['Page', 'PageHeader', 'PageSidebar', 'PageSection']
```
nicolethoen, line 6: Should the new Components be listed here so their props are included in the docs?

packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+  hiddenOnLg,
+  visibleOnLg,
+  hiddenOnSm,
+  ...props
+}: PageHeaderToolsGroupProps) => (
+  <div
+    className={css(
+      styles.pageHeaderToolsGroup,
+      className,
+      hiddenOnLg && styles.modifiers.mobile,
```
mcoker, line 30: If we're going to structure this so that we offer header toolbar groups and items that require the user specify `visibleOnLg` for items that are only visible on desktop, `hiddenOnSm` for items that are hidden on mobile, etc, then I wonder if we should drop the current modifiers and use a hidden/visible mixin that would give us a set of styles that allow a user to specify `pf-m-hidden[-on-{breakpoint}]` and `pf-m-visible[-on-{breakpoint}]` on any header-tools-group or header-tools-item. In this PR, users can only specify `hiddenOnLg `, `visibleOnLg`, and `hiddenOnSm`.

Then users will have full control over hiding/showing things at any breakpoint. The downside to the current approach (whether we include the mixin or not) is that the responsibility is on the user to hide/show header toolbar items appropriately. The alternative to that would be to have something like 3 props on the header tools component, `mobileDropdown`, `desktopActions`, and `userMenu` or something like that, and we put the contents of those props in the appropriate classes and control the mobile, desktop, and user menu breakpoints where that content is hidden/shown.

The former seems like the better approach, but could lead to inconsistency in implementation. @mcarrano wdyt? If we're OK allowing users to hide/show any header toolbar items/groups at any breakpoint, but we show the proper way to do it in the demos and could have a masthead/header design guideline that goes into more detail, I'd like to drop the existing classes and add the full hidden/visible breakpoint classes to header toolbar groups and items.

We could keep the existing classes and add the hidden/visible classes on top, but that seems like it might be confusing to have both, and the prop names that enable the current classes seem like they would overlap.

packages/react-core/src/components/Page/PageHeaderToolsItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * This component wraps the child and applies styling such as selected on the thing passed in, example:
+ *
+ * <PageHeaderToolsItem isSelected><Button isPlain>Hello</Button></PageHeaderToolsItem>
+ * will render
+ * <button class="pf-m-plain pf-m-selected">Hello</button>
+ */
+export const PageHeaderToolsItem: React.FunctionComponent<PageHeaderToolsItemProps> = ({
```
mcoker, line 25: I'm curious how this component works. I think it would be more consistent if we had a `.pf-c-page__header-tools-item` class that you could apply `hidden/visible` to. I'm assuming that would still allow us to apply `.pf-m-selected` to a `<Button>` used in `.pf-c-page__header-tools-item` if it has the `isSelected` prop?

packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
jschuler: Yes, I was going back and forth with the prop names, initially I named them something similar to your mentioned alternative before landing on this (as it also seemed familiar with what we have in other components). 
I guess if we wanted to we could streamline this even more and provide a way for users to define specific things and we can lay it out for them:
```
<PageHeaderTools 
   mobileDropdown={<Dropdown />}
   desktopActions={
      { component: <NotificationBadge />, isSelected: false }
      { component: <Button><CogIcon /></Button>, isSelected: true }
   }
   userDropdown={
      <Dropdown />
   }
   avatar={<Avatar />}
>
   Other components here that should come before?
</PageHeaderTools>
```

Maybe this for ease-of-use, but still allow users to use the <PageHeaderTools> component without those convenience props and define the content themselves using visibility mixins?

packages/react-core/src/components/Page/PageHeaderToolsItem.tsx
jschuler: So the problem is, in HTML you can just add the `pf-m-selected` class to a `button` anywhere. If you wanted to do this in react you'd have to manually add this class name yourself i.e. 
```
<PageHeaderTools>
  <PageHeaderToolsGroup>
      <Button className="pf-m-selected">
          <CogIcon />
      </Button>
  </PageHeaderToolsGroup>
</PageHeaderTools>
```
because the selected modifier is not a prop we have available for Buttons, the class is defined in the context of page.

Wrapping the `<Button>` component in `<PageHeaderToolsItem>` basically allows me to pass the `pf-m-selected` class or any other potential modifier down to the child component.

packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
mcoker: Looking at the way it's done in master now, it looks like we import the accessibility utility classes and it's up to users to apply hidden/visible classes on their own. It seems logical then to bake this hidden/visible behavior into the page header tools, and give users the ability to hide/show header elements at any breakpoint.

mcarrano:
@jschuler I realize that the goal of this PR is not to fully implement the intended design updates for the masthead, but here is a Marvel deck that shows where we want to get to for reference: https://marvelapp.com/5cjajdg/screen/63256946  Getting the demo as close as possible to this would be great. Noe the the Notification icon never hides.

jschuler:
@mcarrano thanks for that, I'll update the demo to reflect this layout as much as possible.

@mcarrano @mcoker and @mceledonia and I met this morning and decided that core will make some changes. Core will remove the `pf-m-mobile`, `pf-m-user`, and `pf-m-icons` classes, and instead bring in the breakpoint modifier classes.
Core will also create a `pf-c-page__header-tools-item` class.

Once these changes are in the core v4 branch, we'll bring it into react and then I'll update this PR. So it'll be on hold until then!

mcoker:
@jschuler https://github.com/patternfly/patternfly/pull/3091 was merged.

* Removed `.pf-m-[icon/mobile/user]`
* Adds `.pf-c-page__header-tools-item` (as a child of `.pf-c-page__header-tools-group`)
  * `.pf-c-page_-header-tools-group` is still required for any `.pf-c-page__header-tools-item`
* Adds `.pf-m-hidden[-on-{breakpoint}]` and `.pf-m-visible[-on-{breakpoint}]` that can be applied to `__header-tools-[item/group]`

redallen:
Depends on #4243

packages/react-core/src/components/Page/PageHeaderToolsItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface InjectedPageHeaderToolsItemProps {
+  selectedClass: string;
+}
+
+export interface PageHeaderToolsItemProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered in page header tools item. If you need the selected class, use the render prop instead */
+  children?: React.ReactNode;
+  /** If you need the selected class, use this render prop instead of children */
+  render?: (props: InjectedPageHeaderToolsItemProps) => React.ReactNode;
```
jschuler, line 15: opened issue in core so this hopefully won't be necessary to get the selected class
https://github.com/patternfly/patternfly/issues/3108

redallen: We haven't really used an API like this before, but I like it and I think it satisfies the documentation part.

jschuler: Yes, I remembered there was some controversy about using `cloneElement` so render prop puts the resposibility of adding the class back to the user. anyways, this render prop will be short-lived once the mentioned core PR is available :)

packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const PageHeaderToolsGroup: React.FunctionComponent<PageHeaderToolsGroupProps> = ({
+  children,
+  className,
+  breakpointMods,
+  ...props
+}: PageHeaderToolsGroupProps) => (
+  <div
+    className={css(
+      styles.pageHeaderToolsGroup,
+      breakpointMods && formatBreakpointMods(breakpointMods, styles),
```
tlabaj, line 25: I opened an issue agains the page header.  `formatBreakpointMods` function does not work for the  2xl breakpoint.

mcoker:
This LGTM, my only comments are that the notification bell should be in a header-tools-item element, and that the cog shouldn't be selected. We can add the selected state as an example with either a drawer or notification drawer demo where a header tools button toggles a drawer.

mcoker:
@jschuler looks like we lost the header tools in the card view demo - https://patternfly-react-pr-4223.surge.sh/documentation/react/demos/cardview/basic. Otherwise LGTM!

jschuler:
@mcoker thanks, I can add those in a followup PR

redallen:
@jschuler Fix conflicts and we can merge!

Issue #3840, opened 2020-02-28
Modal title and aria-label


mcoker:
**Describe the issue. What is the expected and unexpected behavior?**

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
https://www.patternfly.org/v4/documentation/react/components/modal

Looking at the modal component, I noticed the following:

* The `title` prop value is used to populate the `aria-label`, which seems reasonable.
* If you omit `title`, you end up with an empty `<h1 class="pf-c-title">` element, which doesn't seem ideal from a code perspective, and it breaks the layout by creating extra space at the top of the modal from the empty element.
  * To get the proper layout in a modal without a title, you also have to specify the `hideTitle` prop. That seems unnecessary.
* If you want the modal to have an `aria-label`, but with no visible title, you specify the label in the `title` prop, then use `hideTitle` to remove the title from the page. This is a bit confusing.
* Since the value of `title` is also used as the `aria-label`, the title value needs to be a string type, which may not always be ideal if maybe you wanted to use a `<span>` or `<strong>` element or icon or something like that.

I wonder if it might be better to:

* If you specify a `title`, it generates the title in the markup and assigns the title an `id` that is then used with `aria-labelledby` on the modal. Since the value of the title prop is no longer also used as the `aria-label`, the title would no longer need to be a string.
* If you omit the `title`, it removes the title from the component. No need to do anything else to address the layout.
* Add the ability to specify an `ariaLabel` or `ariaLabelledById` prop (similar to the existing `ariaDescribedById`) that gives the modal a label in the case that 1) there is no `title`, or 2) when used with a `title`, the value they pass would override the `id` the component would normally generate from the title.

tlabaj:
cc @jessiehuff

Issue #4218, opened 2020-05-08
Modal title and label missing warning


jessiehuff:
**Describe the issue. What is the expected and unexpected behavior?**
Recently in the [wizard refactor](https://github.com/patternfly/patternfly-react/pull/4140), we made the title prop for modal optional. This change makes sense for a lot of reasons, including being able to reference the wizard title as the aria-label or title for the modal and allows more flexibility in usage. Also, as Coker describes in this [issue](https://github.com/patternfly/patternfly-react/issues/3840): It used to be that if you omitted title, you ended up with an empty `<h1 class="pf-c-title">` element, which wasn't ideal from a code perspective, and broke the layout by creating extra space at the top of the modal from the empty element. To get the proper layout in a modal without a title, you used to have to specify the hideTitle prop which seems unnecessary. The modal doesn’t render an empty `<h1>` if you omit the title now which makes `hideTitle` unnecessary.

However, there is an accessibility concern if a consumer chooses to omit title and aria-label. Labeling is the most common accessibility issue that I've seen, as it's very easy to forget or not consider labeling. If they have neither attribute in a modal, it gives the user very little context that they’re inside of a modal. In a basic modal without either attribute, VO will read “close button” as soon as the modal opens and without any titles or aria-labels with the wizard in the modal, focus would go to the next button and read "steps, dialog, next, button, footer". Both of which would make the modal a bad accessible experience.

Modal needs some sort of title in it  - whether it’s the proper modal title (title prop), or you use the noModalBodyWrapper and you have an `<h1>` or whatever in it. If there isn’t a header, the modal needs either `aria-label` or `aria-labelledby` that points to a title that could serve as the label for the modal. It would be ideal if we could provide a warning to consumers if they don’t provide a title or an aria-label or if there was some way to notify users of the potential issue. We have done something similar in table: https://github.com/patternfly/patternfly-react/blob/v4/packages/react-table/src/components/Table/Table.tsx#L384-L387 . 

This would work for the most part, except in the case of the wizard or just a use of noModalBodyWrapper where the user has specified their own title in the modal. We have no way of knowing whether they supplied a title in the modal or not, so if they do have a title in their modal, but didn’t use our title or `aria-label` or `aria-labelledby`, we would end up giving them a warning even though they have a title and are using the modal correctly. 

The question is how to warn or notify users of potentially serious accessibility issues. We want to allow flexibility in our components but also ensure that we're not encouraging more accessibility errors in product usage.

[...]

PR #4239, opened 2020-05-11
fix(modal): core updates and refactoring for modal

seanforyou23:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/3840

This PR applies updates from latest core changes for Modal. It also takes a step toward solving https://github.com/patternfly/patternfly-react/issues/4218 for the case where users are supplying their own header or modal content but haven't supplied an accessible name for them.

**Additional issues**:


packages/react-core/src/components/Modal/Modal.tsx
```diff
@@ -120,6 +130,11 @@ export class Modal extends React.Component<ModalProps, ModalState> {
     } else {
       target.classList.remove(css(styles.backdropOpen));
     }
+
+    if (!title && !ariaLabel && !modalContentAriaLabelledById) {
+      // eslint-disable-next-line no-console
+      console.error('Modal: Specify at least one of: title, aria-label, modalContentAriaLabelledById');
```
mcoker, line 136: 🤠👍

mcoker: should `aria-label` be `ariaLabel`?

packages/react-core/src/components/Modal/ModalBox.tsx
```diff
@@ -23,14 +25,16 @@ export const ModalBox: React.FunctionComponent<ModalBoxProps> = ({
...
+  descriptorId,
+  labelId,
   ...props
 }: ModalBoxProps) => (
   <div
     {...props}
     role="dialog"
-    aria-label={title || ariaLabel}
-    aria-describedby={id}
+    aria-label={title || ariaLabel || null}
```
mcoker, line 35: should `ariaLabel` come first so that if someone specifies both, the val for `aria-label` would override `title`?

mcoker:
* Is there any value in warning if a both `aria-label` and `aria-labelledby` are specified?
* If a user can pass `aria-label`, why not `aria-labelledby` and `aria-describedby`

jgiardino:
I only did a quick test of examples with aria-label or aria-labelledby defined. The only thing I noticed, which might be out of scope for this PR, is that the Wizard component includes a description, but I'm not seeing a way to specify that description for the `aria-describedby` on the modal. I see that there is an id for the wizard description, but how does it get there (i.e. via a consumer defined value, or via the component itself?). Ideally the user should be able to reference that ID and use it as the modal description.

Regarding this comment:
> Is there any value in warning if a both aria-label and aria-labelledby are specified?

Using both is great for cases like a Chip component, where you have a recurring button, like "Remove", on a page and you want to provide context by referencing an additional label. I'm not sure this case applies to the modal, since the modal is expected to be the only thing that's accessible on the page (i.e. its title IS the context). Maybe there's some weird edge case I'm not thinking of, though?

mcoker:
@seanforyou23 this looks really great! What do you think about adding an `id` to the title, and using that ID as the value for `aria-labelledby` to label the modal, which would allow anything to be passed as the `title`, not just a string type?

seanforyou23:
Just pushed another round of updates - thx for the feedback!

> If a user can pass aria-label, why not aria-labelledby and aria-describedby

This was possible previously by setting `modalContentAriaDescribedById` and `modalContentAriaLabelledById` - while setting aria-describedby/labelledby would set this attribute on an inner element of the modal - which isn't the desired behavior. I've reworked these props to use their normal name (aria-describedby instead of modalContentAriaDescribedById). Much better now 👍 

>not seeing a way to specify that description for the aria-describedby on the modal

Made this more clear in the Wiz example and also made sure users can set both labelledby and describedby in this same way.

>I see that there is an id for the wizard description, but how does it get there

The component does [generate an internal id](https://github.com/patternfly/patternfly-react/pull/4239/files#diff-339a05716b361dbb9c21c5ca44805b52R78-R79) for the purpose of linking to description text in the simple cases, I extended this to accept an override if the user provides one. Wiz is a good example of a use case that's best served to set their own value for this as the container body (where the description would be held) is supplied by the user. I [updated the example](https://github.com/patternfly/patternfly-react/pull/4239/files#diff-543b02a48e5e28d8a50604589db040dfR426-R431) to show how this can be achieved.

>adding an id to the title, and using that ID as the value for aria-labelledby to label the modal...

I did play this through but ended up reverting it because I found this strategy didn't work as well as setting aria-label explicitly. The problem was that the text linked via aria-labelledby was **not** announced upon launching the modal as one would expect. You can see the same behavior in the wizard example. Any ideas what causes this?

packages/react-core/src/components/Modal/ModalBoxBody.tsx
```diff
@@ -7,14 +7,17 @@ export interface ModalBoxBodyProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   /** Additional classes added to the ModalBoxBody */
   className?: string;
+  /** An id added to the ModalBoxBody */
+  descriptorId?: string;
```
redallen, line 11: Remove this prop in favor of just passing `id`.

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -74,9 +77,16 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
   }
 
   const modalBoxHeader = header ? (
-    <div className={css(titleStyles.title)}>{header}</div>
+    <ModalBoxHeader>{header}</ModalBoxHeader>
   ) : (
-    title && <ModalBoxHeader>{title}</ModalBoxHeader>
+    title && (
+      <ModalBoxHeader>
+        <>
```
redallen, line 84: No need for Fragment here.

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -39,10 +43,8 @@ export interface ModalContentProps {
   actions?: any;
   /** A callback for when the close button is clicked */
   onClose?: () => void;
-  /** Id to use for Modal Box descriptor */
-  modalBoxAriaDescribedById?: string;
   /** Id of the ModalBoxBody */
```
redallen, line 46: Update to `/** Id of ModalBoxDescription */`

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -88,7 +98,7 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
   const modalBody = hasNoBodyWrapper ? (
     children
   ) : (
-    <ModalBoxBody {...props} {...(!description && { id })}>
```
redallen, line 91: Change back to `id`.

packages/react-core/src/components/Modal/__tests__/Generated/ModalBoxBody.test.tsx
```diff
@@ -8,6 +8,6 @@ import { ModalBoxBody } from '../../ModalBoxBody';
 import {} from '../..';
 
 it('ModalBoxBody should match snapshot (auto-generated)', () => {
-  const view = shallow(<ModalBoxBody children={<>ReactNode</>} className={"''"} />);
+  const view = shallow(<ModalBoxBody children={<>ReactNode</>} className={"''"} descriptorId={'string'} />);
```
redallen, line 11: Change back to `id`.

jgiardino:
> @seanforyou23 this looks really great! What do you think about adding an id to the title, and using that ID as the value for aria-labelledby to label the modal, which would allow anything to be passed as the title, not just a string type?

I'm not sure I 100% understand what @mcoker is suggesting. I'm giving this idea my own spin, and @mcoker please speak up if you had an alternate idea. 

Some notes:

- I'm not seeing any issues with aria-labelledby working. I am able to edit the examples, specify `aria-labelledby`, and it works as expected with VO in Safari.
- It could be confusing to understand what exactly `title` is doing, relative to `header`, `aria-label`, and `aria-labelledby`. I haven't yet tested all possible combinations and what the expected outcome should be, but thinking about them makes my brain hurt 🙃 

So regarding those notes, here is my spin of what @mcoker is suggesting:

- Only one prop is used to specify the header/title. This prop can accept html or just a string. Currently, these are separate props, one for custom header, and one for just the string.  
- The attributes `aria-label`, `aria-labelledby`, and `aria-describedby` can also be set on the modal, as currently supported in this PR.
- Given that only the one header/title prop is specified, we assume this prop holds the title, and the body holds the description. We continue to automatically set `aria-describedby` to the modal body, as it's currently implemented. We additionally automatically set `aria-labelledby` to an id that we add to `header.pf-c-modal-box__header`.
- Given that the header/title prop is specified, when the user explicitly specifies `aria-labelledby` or `aria-describedby`, we use the value the user passes. 
- Given that the header/title prop is specified, when the user explicitly specifies `aria-label`, we do not automatically set `aria-labelledby` and instead only set `aria-label` to the value the user provides.
- Given that the header/title prop is not specified, then we display a warning if either `aria-label` or `aria-labelledby` is not specified by the user. 

Is something like this possible? I think this would give the consumer the flexibility to customize what they include in the header, without requiring them to specify anything extra if they just want a basic modal with text string for the title.

mcoker:
@jgiardino 

> We additionally automatically set aria-labelledby to an id that we add to header.pf-c-modal-box__header

This is basically what I was suggesting. Instead of passing the value of the `title` prop to `aria-label`, we give the title (or header as you've suggested) an `id`, and pass the `id` to `aria-labelledby` instead, which will allow `title` to be either HTML or string.

> Given that only the one header/title prop is specified, we assume this prop holds the title, and the body holds the description. We continue to automatically set aria-describedby to the modal body, as it's currently implemented. We additionally automatically set aria-labelledby to an id that we add to header.pf-c-modal-box__header.

Does this account for the the current behavior where the `.pf-c-modal-box__description` element is also in `header.pf-c-modal-box__header` (defined by the `description` prop), and when `description` is specified, the `id` of `header.pf-c-modal-box__header` is used as the value for `aria-describedby`?

jgiardino:
> Does this account for the the current behavior where the .pf-c-modal-box__description element is also in header.pf-c-modal-box__header (defined by the description prop), and when description is specified, the id of header.pf-c-modal-box__header is used as the value for aria-describedby?

Thanks for pointing this out. I totally missed this. I don't see a reason why we wouldn't continue to support the `description` prop. So if both the header/title prop and description prop are specified, then `aria-labelledby` and `aria-describedby` are still automatically set, but `aria-describedby` would then just reference the id on `.pf-c-modal-box__description`. 

But this does raise a question about the header/title prop accepting either a string or HTML. In the current preview for this PR, if a custom header is specified, then the `description` prop appears to be ignored. If we do go with this single header/title prop supporting either a string or HTML, should we just always include the `.pf-c-modal-box__description` element when `description` is defined?

seanforyou23:
@redallen the expectation should be that setting an id prop on Modal would have the effect of planting the id attribute on the outermost container element, not an inner element like ModalBoxBody. Added this capability [here](https://github.com/patternfly/patternfly-react/pull/4239/files#diff-339a05716b361dbb9c21c5ca44805b52R86). Let me know if you see any issues, thx!

mcoker:
> if a custom header is specified, then the description prop appears to be ignored.

That's because I think we're assuming users have opted to redefine the "header", which includes the title and description. 

If the purpose of the `header` prop is just to provide a custom title (which I think is the existing behavior), I don't see the harm in continuing to include `description` if it's provided. The `pf-c-modal-box__description` element has a generic padding-top that spaces it from anything that comes before it, so that should work, though either `header` or `title` would need to be defined to use `description`.

> Only one prop is used to specify the header/title. This prop can accept html or just a string.

I wasn't suggesting that initially, but do you mean something like this?

```
<div class="pf-c-modal-box__header">
   if title == string: <h1 class="pf-c-modal-box__title">{{title}}</h1>
   elseif title == html: {{title}}
```
   
The reason I'd like to see the user be able to pass html to the title is for something simple like

`title="Normal and <em>emphasized</em> text"` or `title="<span class="whatever">Title text</span>"`

I think we could still support the two `title` and `header` props (or maybe `customTitle` instead of `header`, to be more clear?), and both can be a string or HTML, and `title` would create the normal modal title, and `customTitle` would replace it. I think that's how the `title` and `header` props work in the component today.

```
<div class="pf-c-modal-box__header">
   if customTitle: {{customTitle}}
   elseif title: <h1 class="pf-c-modal-box__title">{{title}}</h1>
```

mcoker:
@seanforyou23 per this update, looks like the modal always uses `aria-labelledby` with 2 `id`s as values, and refers to the `id` of the modal as the first value, and the `id` of the title as the second value.

A couple of problems I see:

I may be overlooking the way to do this, but when specifying an `aria-label`, the `aria-labelledby` value doesn't change. Meaning `aria-labelledby` still has the `id` of the modal and `id` of the title as its values. So the value of `aria-label` will be read with the title as the label for the modal. 
  * Is it possible to simply specify a unique `aria-label` that is read on it's own, not one that is read with the title? Or are we assuming that the contents of `title` should always be read as part of the modal's label?

When specifying a custom `header`, looks like an undefined `id` is used in the modal's `aria-labelledby` attribute. Is that a bug?

Personally I liked the approach in Jenn's comment https://github.com/patternfly/patternfly-react/pull/4239#issuecomment-627568852, except if we were to keep `header` and `title` as separate props, we can automatically assign an `id` to the title if `title` is specified, since we generate the `<h1 class="pf-c-modal-box__title">` element that wraps the title, and that `id` can be used as the value for `aria-labelledby` on the modal. If the user specifies `header` instead, we do not specify `aria-label` or `aria-labelledby`, and it's on them to either specify their own label since we have no idea what they would have passed.

If they pass `aria-label`, we don't generate an `aria-labelledby` (unless they also specify `aria-labelledby`) and use the value in the `aria-label` attribute.

If they pass `aria-labelledby`, we use that as the value for the `aria-labelledby` attribute.

If they specify a `description`, we continue to assign an `id` to `.pf-c-modal-box__description` and use the `id` as the value for `aria-describedby`.

If they don't specify a `description`, we continue to assign an `id` to `.pf-c-modal-box__body` and use the `id` as the value for `aria-describedby`.

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -88,24 +106,24 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
...
+    <ModalBoxBody {...props} {...(!description && !ariaDescribedby && { descriptorId })}>
```
redallen, line 109: Change to `<ModalBoxBody {...props} {...(!description && !ariaDescribedby && { id: descriptorId })}>`.

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -63,8 +69,11 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
   onClose = () => undefined as any,
   variant = 'default',
   width = -1,
-  modalBoxAriaDescribedById = '',
-  id = '',
+  'aria-describedby': ariaDescribedby = '',
+  'aria-labelledby': ariaLabelledby = '',
+  boxId = '',
+  labelId = '',
+  descriptorId = '',
```
redallen, line 76: You can destructure `id` (even though it's unused) to avoid passing it to ModalBoxBody. Add an `// eslint-ignore-next-line @typescript-eslint/no-unused-vars` beforehand to quiet the linter.

Issue #3684, opened 2020-02-06
Chip group behavior updates


mcoker:
Placeholder for the work in https://github.com/patternfly/patternfly-next/issues/1837

mcoker:
There are behavioral issues in the issue that are beyond the changes introduced by core - https://github.com/patternfly/patternfly/issues/1837

The core work is complete https://github.com/patternfly/patternfly/pull/2961

* Changes `pf-m-toolbar` to `pf-m-category`
* The `<li>` in the top-level `<ul>` in a category/toolbar chip group now has a class `.pf-c-chip-group__list-item`
* The overflow chip now goes inside category/toolbar chip groups instead of outside

cc @christiemolloy anything else I missed?

christiemolloy:
No, I think thats good. All other changes are documented by design in the issue

mcoker:
We are still working on a restructure in core that should finish before this work begins - https://github.com/patternfly/patternfly/pull/3058/

I will update this PR when that work has merged.

mcoker:
@tlabaj the work in https://github.com/patternfly/patternfly/pull/3058 has been merged, so this is ready to work on

tlabaj:
closed by #4246

PR #4246, opened 2020-05-13
improvement(ChipGroup): Update ChipGroup to match new design

tlabaj:
**What**: Closes # #3684

## Breaking changes
1. **ChipGroup** :`withToolbar` prop has been removed.  Add the `categoryName` prop instead to add a chip group with a category.
2. **ChipGroupToolbarItem**: This component has been removed.  To Create a ChipGroup with a category, add the `categoryName` prop to `<ChipGroup>`.  All props that were on the `<ChipGroupToolbarItem>` have been added to `<ChipGroup>`.  So to convert to new API, move move all you props up to <ChipGroup> and remove `<ChipGroupToolbarItem>`.
3. **Chip**: The overflow chip no longer contains a button.  Th specify a overflow chip as a button do the following `<Chip component='button'  isOverflowChip>
            Overflow Chip name
          </Chip>`


packages/react-core/src/components/ChipGroup/Chip.tsx
```diff
@@ -14,13 +14,13 @@ export interface ChipProps extends React.HTMLProps<HTMLDivElement> {
...
   isOverflowChip?: boolean;
-  /** Flag if chip is read only */
+  /** Flag indicating if chip is read only */
   isReadOnly?: boolean;
-  /** Function that is called when clicking on the chip button */
+  /** Function that is called when clicking on the chip close button */
   onClick?: (event: React.MouseEvent) => void;
-  /** Internal flag for which component will be used for chip */
+  /** Component that will be used for chip. It is recommended that <button> or <li>  are used when the chip is an overflow chip. */
   component?: React.ReactNode;
```
jschuler, line 24: could this be more constrained to `'div' | 'li' | 'button'`?

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -17,112 +19,163 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
...
+                    return (
+                      <li className={css(styles.chipGroupListItem)} key={i}>
+                        {child}
+                      </li>
+                    );
+                  }
+                })}
+              </React.Fragment>
+            )}
+            {React.Children.count(children) > numChips && (
```
redallen, line 153: Change to `{numChildren > numChips && (`

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -17,112 +19,163 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
...
+      categoryName,
+      children,
+      className,
+      isClosable,
+      closeBtnAriaLabel,
+      'aria-label': ariaLabel,
+      onClick,
+      numChips,
+      expandedText,
+      collapsedText,
```
redallen, line 111: Add props to avoid passing to DOM
```ts
      // eslint-disable @typescript-eslint/no-unused-vars
      defaultIsOpen,
      tooltipPosition,
      // eslint-enable @typescript-eslint/no-unused-vars
```

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -17,112 +19,163 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
...
+                  if (i < numChips) {
+                    return (
+                      <li className={css(styles.chipGroupListItem)} key={i}>
+                        {child}
+                      </li>
+                    );
+                  }
+                })}
+              </React.Fragment>
+            )}
```
redallen, line 152: This can be simplified quite a bit.
```tsx
{React.Children.toArray(children)
  .filter((_child, i) => isOpen || i < numChips)
  .map((child, i) => (
    <li className={css(styles.chipGroupListItem)} key={i}>
      {child}
    </li>
  ))}
```

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -17,112 +19,163 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLDivElement> {
...
+      const renderChipGroup = (id: string) => (
+        <div className={css(styles.chipGroup, className, categoryName && styles.modifiers.category)}>
+          {categoryName && this.renderLabel(id)}
+          <ul
+            className={css(styles.chipGroupList)}
+            {...(categoryName && { 'aria-labelledby': id })}
+            {...(!categoryName && { 'aria-label': ariaLabel })}
+            // aria-labelledby={id}
+            // aria-label={ariaLabel}
+            role="list"
```
redallen, line 129: One last thing -- [after reading MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/List_role) I don't think this is necessary anymore since we're using a `<ul>` which is a list. Per @jessiehuff 's review we can remove it.

packages/react-core/src/components/ChipGroup/Chip.tsx
tlabaj: @mcoker and I talked about this yesterday and saw no need to constrain this.

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
tlabaj: @jessiehuff @mcoker this was listed as a new attribute to add in the issue.  Should it not be added?

jessiehuff: I originally thought the same thing since `<ul>` is implicitly `role="list"`, but we found after testing different browser and AT combinations that the explicit role seemed to be beneficial. I'm in favor of keeping it unless we find that it becomes better supported in the future without it. :)

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
dlabaj: Doesn't this create double the runtime of algorithm using this approach instead of the current one?

redallen: I don't think so. Both algorithms iterate through all of their children which is O(n).

mcoker:
@mcarrano re: the cut off text, @christiemolloy noticed that also and I have a PR up to fix it - https://github.com/patternfly/patternfly/pull/3110

packages/react-core/src/components/ChipGroup/examples/ChipGroup.md
```diff
@@ -69,174 +72,81 @@ class SingleChip extends React.Component {
         <Chip key="chip4" onClick={() => this.deleteItem('readonlychip')} isReadOnly>
```
mcoker, line 72: is `isReadOnly` still necessary? We don't have a variation for that in core anymore. You create a read only chip by omitting the close button, and looks like the react chip component doesn't add a close button if you don't specify `onClick()`, which seems to create all you need for a read only chip.

tlabaj: The component currently uses the the  isReadOnly flag to omit the close button.   with the current implementation, if the consumer does not provide a `onClick` function and `isReadOnly === false`, the close button will still be rendered.

mcoker: Ah yep you're right, I'm not sure why I thought it was removed if you didn't pass `onClick()`. Since `onClick()` is only used to define the function to run when the close button is clicked, it seems intuitive that omitting that would omit the close button, but maybe there are other reasons not to do that.

Issue #3834, opened 2020-02-26
Update image used in Login component example


mcoker:
https://www.patternfly.org/v4/documentation/react/components/loginpage/basic

The header image is way too big. Can we update that image to match what's in the core example? https://www.patternfly.org/v4/documentation/core/components/login/basic

mcoker:
Looks like this has been resolved.

PR #4247, opened 2020-05-13
feat(Form): added icon to form helper text

### LINKED ISSUE #4070: Form helper text - support for icon
In the login form, we display an icon inside of the form's helper text element. In core, this was only supported in the login component. We've since moved that feature out of the login component and back into the form component, so it can be used anywhere form helper text is used.

I'm not sure how this is configured in react, so just wanted to open an issue for someone to investigate. We also added an example in the form component docs next to the textarea in http://patternfly-v4.surge.sh/documentation/core/components/form#help-text

kmcfaul:
and added icon props to login and formgroup

**What**: Closes #4070

Note:
Allows icon via FormHelperText component or via additional props in FormGroup and LoginForm, which wrap FormHelperText internally. For FormGroup, helper text with icons can be implemented by passing FormHelperText components, or using the current helperText props combined with the new helperTextIcon props.

1. **LoginForm**: added `helperTextIcon` property. The example's helper text was updated to use this new property. 
2. **FormHelperText**: added `icon` property.
3. **FormGroup**: added `helperTextIcon` and `helperTextInvalidIcon` properties. `FormHelperText` components may now be passed into the `helperText` and `helperTextInvalid` properties directly.

### DIFF
--- a/packages/react-core/src/components/Form/FormGroup.tsx
+++ b/packages/react-core/src/components/Form/FormGroup.tsx
@@ -3,7 +3,6 @@ import styles from '@patternfly/react-styles/css/components/Form/form';
 import { ASTERISK } from '../../helpers/htmlConstants';
 import { css } from '@patternfly/react-styles';
 import { ValidatedOptions } from '../../helpers/constants';
-import { FormHelperText } from './FormHelperText';
 
 export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'label'> {
   /** Anything that can be rendered as FormGroup content. */
@@ -51,7 +50,7 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
   ...props
 }: FormGroupProps) => {
   const validHelperText =
-    React.isValidElement(helperText) && (helperText as React.ReactElement).type === FormHelperText ? (
+    typeof helperText !== 'string' ? (
       helperText
     ) : (
       <div
@@ -65,7 +64,7 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
     );
 
   const inValidHelperText =
-    React.isValidElement(helperTextInvalid) && (helperTextInvalid as React.ReactElement).type === FormHelperText ? (
+    typeof helperTextInvalid !== 'string' ? (
       helperTextInvalid
     ) : (
       <div className={css(styles.formHelperText, styles.modifiers.error)} id={`${fieldId}-helper`} aria-live="polite">
--- a/packages/react-core/src/components/Form/FormGroup.tsx
+++ b/packages/react-core/src/components/Form/FormGroup.tsx
@@ -3,6 +3,7 @@ import styles from '@patternfly/react-styles/css/components/Form/form';
 import { ASTERISK } from '../../helpers/htmlConstants';
 import { css } from '@patternfly/react-styles';
 import { ValidatedOptions } from '../../helpers/constants';
+import { FormHelperText } from './FormHelperText';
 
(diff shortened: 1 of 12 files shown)


packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -40,24 +45,34 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
...
-    <div
-      className={css(styles.formHelperText, validated === ValidatedOptions.success && styles.modifiers.success)}
-      id={`${fieldId}-helper`}
-      aria-live="polite"
-    >
-      {helperText}
-    </div>
-  );
+  const validHelperText =
+    helperText && (helperText as React.ReactElement).type === FormHelperText ? (
```
jschuler, line 54: why is the type check needed? i think this would break if for whatever reason the user wrapped like this
```
helperText={
<>
  < FormHelperText />
</>
}
```

redallen, line 54: Change to `React.isValidElement(helperText) && (helperText as React.ReactElement).type === FormHelperText`

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -40,24 +45,34 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
...
+      </div>
+    );
 
-  const inValidHelperText = (
-    <div className={css(styles.formHelperText, styles.modifiers.error)} id={`${fieldId}-helper`} aria-live="polite">
-      {helperTextInvalid}
-    </div>
-  );
+  const inValidHelperText =
+    helperTextInvalid && (helperTextInvalid as React.ReactElement).type === FormHelperText ? (
```
redallen, line 68: Change to `React.isValidElement(helperTextInvalid) && (helperTextInvalid as React.ReactElement).type === FormHelperText`

packages/react-core/src/components/Form/FormGroup.tsx
kmcfaul: It would double wrap it if that was passed in I think. Would it be better to check if it is passed a string to wrap it, and if it gets a Node, it just displays it?

redallen: I think that better matches what we do in other components and is a good alternative to what I suggested in my review

kmcfaul: Updated the check

kmcfaul:
Updated FormGroup to display a `helperText`/`helperTextInvalid` node directly unless it is a string, in which case it wraps the string.

mcoker:
@kmcfaul LGTM although I'm seeing a couple of bugs that may be outside of the scope of this PR.

* The SVG in the login is too big, and I'm assuming should use the same icon as we do in core on https://www.patternfly.org/v4/documentation/core/components/login/basic
* There is a `&nbsp;` string in the error message output

Screen Shot 2020-05-13 at 4 08 28 PM

kmcfaul:
The only thing that might be considered breaking is the change that when passing in a string for helperText/Invalid it will wrap into a FormHelperText-like structure, it used to do that for everything. Now if it is passed a node it will be displayed as is.

I'll remove the title for now.

kmcfaul:
@mcoker I fixed the helper text for the login example, I'm not sure about the svg icon

mcoker:
@kmcfaul thanks. I had an issue open about that and it was closed by the stale bot. Re-opened it https://github.com/patternfly/patternfly-react/issues/3834

PR #4248, opened 2020-05-13
fix(Progress): Update to match core and fix accessibility

### LINKED ISSUE #3423: Progress component html does not match core
When comparing the html of the Progress component between react and core, there are some differences. 
- Most notably, the attribute `role="progressbar"` is not on the correct element in the react component. As a result, the aria attributes that provide additional details about the status of the progress bar are not communicated by screen readers. 
- `aria-describedby` is included on the parent element in react, which isn't needed
- Other differences might exist. As part of this issue, please follow the steps below to identify all differences. **Also note,** an issue is opened in core to make changes to the aria attributes: https://github.com/patternfly/patternfly-next/issues/2539

1. Open both core and react examples of the progress bar. 
2. Open the web browser inspector for both components to view and compare the html structure.

bug

The original issue created in core — https://github.com/patternfly/patternfly-next/issues/2539 —  was noticed by a team member of the Cockpit team.

jessiehuff:
**What**: Closes #3423

**Additional issues**:
Also addresses https://github.com/patternfly/patternfly/issues/2539 on the react side

### DIFF
--- a/packages/react-core/src/components/Progress/Progress.tsx
+++ b/packages/react-core/src/components/Progress/Progress.tsx
@@ -70,13 +70,9 @@ export class Progress extends React.Component<ProgressProps> {
       valueText,
       ...props
     } = this.props;
-    const additionalProps = {
-      ...props,
-      ...(valueText ? { 'aria-valuetext': valueText } : { 'aria-describedby': `${this.id}-description` })
-    };
 
     const progressBarAriaProps: AriaProps = {
-      'aria-describedby': `${this.id}-description`,
+      'aria-labelledby': `${this.id}-description`,
       'aria-valuemin': min,
       'aria-valuenow': value,
       'aria-valuemax': max
@@ -89,7 +85,7 @@ export class Progress extends React.Component<ProgressProps> {
     const scaledValue = Math.min(100, Math.max(0, Math.floor(((value - min) / (max - min)) * 100)));
     return (
       <div
-        {...additionalProps}
+        {...props}
         className={css(
           styles.progress,
           styles.modifiers[variant],
@@ -99,7 +95,6 @@ export class Progress extends React.Component<ProgressProps> {
           className
         )}
         id={this.id}
-        role="progressbar"
       >
         <ProgressContainer
           parentId={this.id}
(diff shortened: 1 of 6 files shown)


jgiardino:
These updates look good to me!

seanforyou23:
Good stuff! I'm still reviewing but wanted to share some initial thoughts/feedback.

One thing that jumps out to me is that there isn't currently live-region support. When a progressbar's value is updated, I think unless the user is focused on the progress bar or re-focuses the progress bar, the new value is not necessarily announced to screen reader users (SRU). Maybe there could be a simple prop to turn it into a live region? I wouldn't think to make it a live region by default, but the option would be nice. This could help solve the case below from the [usage and behavior guidelines for progressbar](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/progress-bar).

>When a process needs to be reflected to a user that is happening in the background, especially when this may take a long time

The other thing I noticed which might be worth discussing was that the progress component isn't currently navigable from the two primary methods of navigating for SRU (Rotor menu and shortcut key nav), so unless they use a technique like VO cursor navigation, I don't see how a SRU can consume this content. What sort of expectation do we want to have around how this type of content is exposed? It would be great if we could speak to this in the usage & behavior guidelines.. It's not interactive in the sense that the user can directly interact with it, but it is in the sense that the content is dynamic and the user may wish to monitor the progress in some explicit way (such as focusing the element until progress is complete) - so for that reason, I kinda lean toward thinking we could consider doing whatever necessary to make it appear in the rotor menu. Not settled on this, just wanted to throw it out there for others to chime in.

redallen:
Rebased on `master` since `v4` is about to be deleted.

jessiehuff:
@christiemolloy I definitely see where you're coming from! Currently VO is using our attributes to determine what to say, we're not directly telling the screen readers what to say. I think in order to make it more specific, we'd have to add that specific wording which wouldn't be ideal for things like localization. 

Interesting points, @seanforyou23! Both of those points seem like they'd be worth more investigation, but I think they're out of scope for this PR and should probably be a separate issue.

PR #4252, opened 2020-05-14
fix(table): Specify button type to avoid unwanted form submits

### LINKED ISSUE #4251: Tables with sortable columns inside table cause unwanted submits
SortableColumn component renders html button without type specified.

Buttons without specified type attribute are treated as 'submit' buttons by browsers. This causes issues when a table with sortable columns is used inside a form as the form is submitted by clicking the sort buttons.

jtomasek:
Buttons without specified type attribute are treated as 'submit' buttons by browsers. This causes issues when a table with sortable columns is used inside a form as the form is submitted by clicking the sort buttons.

This change sets the button type to 'button' which establishes correct behaviour.

**What**: Closes #4251

### DIFF
--- a/packages/react-table/src/components/Table/SortColumn.tsx
+++ b/packages/react-table/src/components/Table/SortColumn.tsx
@@ -7,7 +7,7 @@ import styles from '@patternfly/react-styles/css/components/Table/table';
 import { SortByDirection } from './Table';
 import { TableText } from './TableText';
 
-export interface SortColumnProps extends React.HTMLAttributes<HTMLButtonElement> {
+export interface SortColumnProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   children?: React.ReactNode;
   className?: string;
   isSortedBy?: boolean;
@@ -21,6 +21,7 @@ export const SortColumn: React.FunctionComponent<SortColumnProps> = ({
   isSortedBy = false,
   onSort = null,
   sortDirection = '',
+  type = 'button',
   ...props
 }: SortColumnProps) => {
   let SortedByIcon;
@@ -30,7 +31,12 @@ export const SortColumn: React.FunctionComponent<SortColumnProps> = ({
     SortedByIcon = ArrowsAltVIcon;
   }
   return (
-    <button {...props} className={css(className, styles.tableButton)} onClick={event => onSort && onSort(event)}>
+    <button
+      {...props}
+      type={type}
+      className={css(className, styles.tableButton)}
+      onClick={event => onSort && onSort(event)}
+    >
       <div className={css(className, styles.tableButtonContent)}>
         <TableText>{children}</TableText>
         <span className={css(styles.tableSortIndicator)}>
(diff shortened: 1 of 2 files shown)


packages/react-table/src/components/Table/SortColumn.tsx
```diff
@@ -30,7 +30,12 @@ export const SortColumn: React.FunctionComponent<SortColumnProps> = ({
     SortedByIcon = ArrowsAltVIcon;
   }
   return (
-    <button {...props} className={css(className, styles.tableButton)} onClick={event => onSort && onSort(event)}>
+    <button
+      {...props}
```
mareklibra, line 34: Would it make sense to let the user change the type via SortColumn props?

jtomasek: Done

jtomasek:
I've changed SortColumn definition such that the type is 'button' by default and it can be overridden by providing new type via props.

redallen:
Rebased on `master` since `v4` is about to be deleted.

packages/react-table/src/components/Table/SortColumn.tsx
```diff
@@ -7,7 +7,7 @@ import styles from '@patternfly/react-styles/css/components/Table/table';
 import { SortByDirection } from './Table';
 import { TableText } from './TableText';
 
-export interface SortColumnProps extends React.HTMLAttributes<HTMLButtonElement> {
+export interface SortColumnProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
redallen, line 10: What does this change do exactly? I think the old type is correct for `<button {...props}>`

packages/react-table/src/components/Table/SortColumn.tsx
```diff
@@ -30,7 +30,12 @@ export const SortColumn: React.FunctionComponent<SortColumnProps> = ({
     SortedByIcon = ArrowsAltVIcon;
   }
   return (
-    <button {...props} className={css(className, styles.tableButton)} onClick={event => onSort && onSort(event)}>
+    <button
+      type="button"
```
redallen, line 34: If you want to make `type="button"` a default prop, do:
```
  sortDirection = '',
  type = 'button',
  ...props
}: SortColumnProps
```
...
`<button type={type}`
instead.

packages/react-table/src/components/Table/SortColumn.tsx
jtomasek: React.HTMLAttributes is a common subset of all html element attributes, therefore it does not include button specific ones. For better explanation check: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36505#issuecomment-549394273

packages/react-table/src/components/Table/SortColumn.tsx
jtomasek: Done

packages/react-table/src/components/Table/SortColumn.tsx
redallen: Wonderful, thanks for the explanation!

packages/react-table/src/components/Table/SortColumn.tsx
seanforyou23: Do we also need to register the new type prop in `SortColumnProps` interface? I see the default value type = 'button', but maybe we need also something like `type?: 'button' | 'submit' | 'reset';` in the interface?

redallen:
Will merge after RC2.

PR #4086, opened 2020-04-16
chore(readme): update react-core readme to new format

dgutride:
Closes https://github.com/patternfly/patternfly-react/issues/4020 and #4008 

Updates the Readme to the new format for react-core.  Contribution guidelines are being moved out to the common contribution guide and advanced usage concepts were promoted to a top level readme.

PR #4255, opened 2020-05-14
feat(master): merge v4

redallen:
**What**: For RC1

**Additional issues**: The breaking change lint is safe to ignore. We just need to make sure to omit the `* Breaking changes: Alert` when squash + merging in Github.

**Notable conflicts**: Select group, master detail demo


ADVANCED-USAGE-README.md
```diff
@@ -0,0 +1,60 @@
...
+Ensure optimization.sideEffects is set to true within your Webpack config:
+```JS
+optimization: {
+  sideEffects: true
+}
+```
+
+Use ESM module imports to enable tree shaking with no additional setup required.
+```JS
+import { TimesIcon } from '@patternfly/react-icons';
```
jschuler, line 23: is this ok, i thought we were directly importing from the file not the barrel

redallen: Consumers are free to use this syntax. This was from #4086

PR #4258, opened 2020-05-15
fix(topology): make required packages normal deps

redallen:
**What**: Fixes build in patternfly-org. Also bump the theme.

**Additional issues**: Also fix promote script to use prerelease versions since we're still prereleasing.

### DIFF
--- a/packages/react-topology/package.json
+++ b/packages/react-topology/package.json
@@ -28,28 +28,28 @@
     "clean": "rimraf dist"
   },
   "dependencies": {
+    "@types/dagre": "0.7.42",
+    "@types/webcola": "3.2.0",
     "@patternfly/react-core": "^4.15.0",
     "@patternfly/react-icons": "^4.3.0",
     "@patternfly/react-styles": "^4.3.0",
     "d3": "^5.16.0",
+    "dagre": "0.8.2",
     "lodash": "^4.17.15",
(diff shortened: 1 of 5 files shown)


packages/react-topology/package.json
```diff
@@ -28,28 +28,28 @@
     "clean": "rimraf dist"
   },
   "dependencies": {
+    "@types/dagre": "0.7.42",
```
jschuler, line 31: seems like a workaround? what's the error/issue in org?

redallen: The types don't really matter (we are inconsistent for deps like lodash and react-virtualized whether we move them to dev deps or normal deps).

What matters is that we install dagre and webcola which are used in `react-topology/src/layouts/index.ts`. Webpack will automatically try to require with an ESM import like `import { Whatever } from '@patternfly/react-topology';`.

Our options are to make them deps (what this PR does), or make them true peer deps and removing them from the main barrel file.

redallen: `tslib` _is_ a dependency because `tsc` emits files that `require('tslib')`.

As for `@types` dependencies, I don't really think there is a standard. They can be deps (and automatically installed for consumers), dev deps, peer deps, or optional deps.

redallen:
I made the `@types` packages consistently devDeps across all packages.

PR #4262, opened 2020-05-18
feat(Table): Add inline editable dropdown

### LINKED ISSUE #4003: Table inline edit - dropdown
There is a need in the OpenShift migration UI to be able to change the storage class of a given persistent volume (PV) within a table row. In the current implementation a dropdown (I guess technically, it's a PF select component.) with all of the available storage class options is displayed in each table row. See below.
Screen Shot 2020-03-30 at 4 23 01 PM
This is problematic because a "preferred" storage class is selected by default for each PV and we really don't want users messing with the selection unless they're sure they know what they're doing. A better design would be to display the "preferred" selection as text with the user having the option to edit it.

nicolethoen:
**What**: Closes #4003


packages/react-table/src/components/Table/Table.md
```diff
@@ -1549,7 +1585,7 @@ class EditableRowsTable extends React.Component {
             },
             {
               title: (value, rowIndex, cellIndex, props) => (
-                <EditableTextCell
+                <EditableSelectInputCell
```
seanforyou23, line 1588: This rename may have been by mistake?

nicolethoen: oh oops

packages/react-table/src/components/Table/EditableSelectInputCell.tsx
```diff
@@ -0,0 +1,100 @@
...
+  options?: React.ReactElement[];
+  /** Flag indicating the select input is disabled */
+  isDisabled?: boolean;
+  /** Current selected options to display as the read only value of the table cell */
+  selections?: string | SelectOptionObject | (string | SelectOptionObject)[];
+  /** Flag indicating the select menu is open */
+  isOpen?: boolean;
+  /** Event handler which fires when the select toggle is toggled */
+  onToggle?: (isExpanded: boolean) => void;
+  /** */
```
tlabaj, line 48: Is this one missing a description?

packages/react-table/src/components/Table/EditableSelectInputCell.tsx
```diff
@@ -0,0 +1,100 @@
...
+import inlineStyles from '@patternfly/react-styles/css/components/InlineEdit/inline-edit';
+import formStyles from '@patternfly/react-styles/css/components/Form/form';
+
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
+export interface IEditableSelectInputCell extends Omit<React.HTMLProps<HTMLElement | HTMLDivElement>, 'onSelect'> {
+  /** Row index of this select input cell */
+  rowIndex: number;
+  /** Cell index of this select input cell */
+  cellIndex: number;
+  /** Data structure containing:
```
tlabaj, line 13: @redallen is there a way to format these so it displays better in the docs?  It is just one line right now, not very readable.

packages/react-table/src/components/Table/EditableSelectInputCell.tsx
nicolethoen: yup

nicolethoen:
I've opened an issue to fix tooltip so that it parses content better. It is not an issue with the inline editable dropdown but with the way tooltips are applied to truncated content in the table.

packages/react-table/src/components/Table/EditableSelectInputCell.tsx
redallen: Unfortunately not -- but we can look into translating "\n" into `<br />` or something like that.

nicolethoen:
> The only issue I see is that the styling/treatment around validation may need some further work. Should we create a followup issue for it since it may require some adjustment on core side also? 

@seanforyou23 go for it!

PR #4266, opened 2020-05-19
chore(Toolbar): make id prop optional

nicolethoen:
**What**: Closes #

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Toolbar/Toolbar.tsx
+++ b/packages/react-core/src/components/Toolbar/Toolbar.tsx
@@ -132,6 +132,10 @@ export class Toolbar extends React.Component<ToolbarProps, ToolbarState> {
   };
 
   render() {
-    return <GenerateId>{randomId => this.renderToolbar(this.props.id || randomId)}</GenerateId>;
+    return this.props.id ? (
+      this.renderToolbar(this.props.id)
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Toolbar/Toolbar.tsx
```diff
@@ -129,5 +129,9 @@ export class Toolbar extends React.Component<ToolbarProps, ToolbarState> {
         </ToolbarContext.Provider>
       </div>
     );
+  };
+
+  render() {
+    return <GenerateId>{randomId => this.renderToolbar(this.props.id || randomId)}</GenerateId>;
```
dtaylor113, line 135: I find this a little confusing.  GenerateId/randomId can be an id provided by app dev in props and not random or generated?

nicolethoen: It's fair it's a little confusing...
Basically renderToolbar needs an ID, by wrapping the call in a GenerateId component, I can retrieve a randomID, or I can use the one passed by the consumer. would this be better?

```return this.props.id ? this.renderToolbar(this.props.id) : <GenerateId>{randomId => this.renderToolbar(randomId)</GenerateId>;```

Issue #4277, opened 2020-05-21
Use Victory's createContainer instead of allowZoom prop


dlabrecq:
I'm in the process of implementing a chart cursor feature (https://github.com/patternfly/patternfly-react/issues/3151) and discovered an issue with an existing property. Thus, I'd like to remove our `allowZoom` property from the `ChartContainer` component.

There are several Victory containers supporting features like `zoom`, `brush`, `cursor`, `voronoi`, etc. The problem is that these containers are often combined to support multiple features. For example, I may want `zoom` and a `cursor`. This is typically done using the Victory `createContainer` function below. I'd like to change our chart zoom example to use that approach instead. 

`const CursorVoronoiContainer = createContainer("voronoi", "cursor");`

This will allow us to create examples that combine multiple Victory features. That is, instead of only supporting a single zoom container that cannot be combined with anything else. (edited)

PR #4278, opened 2020-05-21
fix(charts): Use Victory's createContainer instead of allowZoom prop

dlabrecq:
This change uses the `VictoryZoomContainer` instead of our `allowZoom` prop. 

Ultimately, this will allow us to to combine Victory container features. For example, if a user wants both a cursor and zoom feature, they must combine containers using Victory's `createContainer` function.

Fixes https://github.com/patternfly/patternfly-react/issues/4277

## Breaking change
Use `containerComponent` instead of `allowZoom`, for example:

```
import { VictoryZoomContainer } from 'victory-zoom-container';
<Chart containerComponent={<VictoryZoomContainer />} />
```

### DIFF
--- a/packages/react-charts/src/components/Chart/Chart.tsx
+++ b/packages/react-charts/src/components/Chart/Chart.tsx
@@ -13,7 +13,6 @@ import {
   VictoryStyleInterface
 } from 'victory-core';
 import { VictoryChart, VictoryChartProps } from 'victory-chart';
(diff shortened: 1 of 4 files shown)


redallen:
We're using `allowZoom` in `packages/react-integration/demo-app-ts/src/components/demos/BarChartDemo/ColorBarZoomDemo.tsx`.

PR #4280, opened 2020-05-22
fix(wiz): add missing label for wiz-in-modal

seanforyou23:
**What**: This PR adds label and description for the case where a user is launching wizard in modal with isOpen. Currently, the modal example for Wizard throws an expected error about this.

Also, I see there's a `width` prop documented for Wizard, but setting it doesn't seem to have an impact. Seems like Wiz should just occupy the space of its container like most other things.. Should we remove the width prop?

### DIFF
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -453,7 +453,14 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
 
     if (isOpen !== undefined) {
       return (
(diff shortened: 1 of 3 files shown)


mcoker:
Hmm... I'm not sure about the `width` and `height` props. Looking at the existing v3 examples, doesn't look like those props do anything - https://www.patternfly.org/v4/documentation/react/components/wizard

Though I imagine that was just to give a custom size to the "wizard modal." 

>  Seems like Wiz should just occupy the space of its container like most other things.. Should we remove the width prop?

Agreed. If anything, we would just want to extend those props to the modal now. Looks like the modal component supports `width`. Should the modal also support a height?

PR #4284, opened 2020-05-24
fix(Table): update imports to direct imports for react table

karelhala:
**What**:

Fixes bloating build when using any react-table component. For some reason this file was forgotten when imports were changed to direct imports in react-table. When we used any component/util that used anything from Table component (for instance `sortable`) the build suddenly contained every icon and component. This PR fixes such issue, I've tried running build and publishing custom package that I later used. The build size went from 12 MB to 6 MB (might need a bit more investigation why it's still relatively that big).

### DIFF
--- a/packages/react-table/src/components/Table/EditColumn.tsx
+++ b/packages/react-table/src/components/Table/EditColumn.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
-import { Button } from '@patternfly/react-core';
-import { PencilAltIcon, CheckIcon, TimesIcon } from '@patternfly/react-icons';
+import { Button } from '@patternfly/react-core/dist/js/components/Button';
+import PencilAltIcon from '@patternfly/react-icons/dist/js/icons/pencil-alt-icon';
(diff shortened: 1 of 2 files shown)


rvsia:
Yep, the Approval-UI is now two times bigger because of this and it really makes loading times grievously long. :-/ 

It would be also nice to fix the issue of including tokens in the core here. It's the same kind of a problem and I would like to see it fixed in the version 3, as we are already using it in GA.

Hyperkid123:
I think there is one more thing that should be probably mentioned. Using the `/js` `/esm` imports directly in the code will mix the commonjs into ESM build and vice versa. This will cause duplicate code in the consumer APP. You will end up with both commonjs and module code of the Button component for instance.

We have been solving this problem using the transform import plugin in libraries that I am taking care of. We are using the shorthand imports in the codebase and then using babel to correctly transform the import paths to the respective build packages based on the build environment.

@redallen don't know if this is something you want to do in the future but it does make sense to me.

redallen:
This can be fixed. We point directly to CJS files for the types.

PR #4286, opened 2020-05-26
fix(Table): update imports to direct imports for react table [v4]

karelhala:
**What**:
Similiar to https://github.com/patternfly/patternfly-react/pull/4284, but opened to master branch.


karelhala:
@redallen if you want we have a starter app repository that can be turned into example of how treeshaking in an app can break. We have to clean some things from our plates as well (our shared frontend components are having similiar issues as PF and we are moving away from webpack to parcel), but once that's done we should update the repository and you guys can then somehow check the bundle sizes every now and then.

redallen:
Please share when you can, that would be great @karelhala !

PR #4290, opened 2020-05-27
fix(Pagination): Fix imports of tokens in pagination to direct imports

karelhala:
**What**:
Similiar to https://github.com/patternfly/patternfly-react/pull/4286. I've updaed the versions of PF packages to latest v3 and saw the drop of build chunks from ~11MB to 4MB. That is awesome! I've checked the BundleAnalyzePlugin and saw additional 800KB to be pulled from `react-tokens` which felt ood. I've then searched for direct imports for this package and found out that paginations is still using the old way of importing files. Hopefully this should shave off significant part of this dependency.

**Additional issues**:


karelhala:
I've searched for this string on master branch as well and couldn't find any reference to generic imports. Hopefully this is the last one.

PR #4293, opened 2020-05-27
chore(InputGroup): add plain button example

### LINKED ISSUE #3045: Add input group examples with plain button modifier
Core issue - https://github.com/patternfly/patternfly-next/issues/2299

Just need to add an example of the input group component with a `<Button variant="plain">` button.

nicolethoen:
**What**: Closes #3045

### DIFF
--- a/packages/react-core/src/components/InputGroup/examples/InputGroup.md
+++ b/packages/react-core/src/components/InputGroup/examples/InputGroup.md
@@ -23,7 +23,7 @@ import {
 ## Examples
 ```js title=Basic
 import React from 'react';
-import { AtIcon, QuestionCircleIcon, SearchIcon } from '@patternfly/react-icons';
+import { AtIcon, SearchIcon } from '@patternfly/react-icons';
 import {
   Button,
   InputGroup,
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/InputGroup/examples/InputGroup.md
```diff
@@ -58,6 +58,13 @@ class SimpleInputGroups extends React.Component {
             <SearchIcon />
           </Button>
         </InputGroup>
+        <br />
+        <InputGroup>
+          <TextInput name="textInput12" id="textInput12" type="search" aria-label="search input example" />
+          <Button variant="plain" aria-label="search button for search input">
```
mcoker, line 64: Can we change the `aria-label` to "popover for input"? The core example is wrong, too.

```suggestion
          <Button variant="plain" aria-label="Popover for input">
```

mcoker:
@nicolethoen after looking at the full examples page, it looks like this should go under the "with popover" examples. Also this example and the existing popover example should likely be `type="text"`. WDYT?

nicolethoen:
> @nicolethoen after looking at the full examples page, it looks like this should go under the "with popover" examples. Also this example and the existing popover example should likely be `type="text"`. WDYT?

@mcoker sounds good to me

mcoker:
@nicolethoen thanks! Can you update the `type` on both of the popover inputs to be `type="text"` and update the `aria-label` on the plain button text input to match the other popover text input? Change `search input example` to `input example with popover`

PR #4306, opened 2020-05-29
feat(Select): add maxWidth to control the panel overflow (#4305)

### LINKED ISSUE #4305: Select dropdown width can't be controlled

If one of the options displayed with a `Select` is very long, the dropdown stretches to contain it all.

https://codesandbox.io/s/qhil7

It's possible to tweak the css to add a `max-width: 100%` to `.pf-c-select__menu`, but maybe this should be supported out of the box with a `maxWidth` property, since there is a `maxHeight` one already.

riccardo-forina:
**What**: Closes #4305 

Add a `maxWidth` property to control the dropdown overflow.


riccardo-forina:
> Nice! Working as expected. We have an [example of maxHeight being used in the demo app](https://github.com/patternfly/patternfly-react/blob/2e2676ce6bc34f5e20231ca066d26ea4c3ac95e4/packages/react-integration/demo-app-ts/src/components/demos/SelectDemo/SelectDemo.tsx#L851), but not maxWidth - might be good to add this new prop there too but overall LGTM

Thank you, added it to that same demo.

riccardo-forina:
Good catch, I didn't think of trying with values below the normal width. Now it should be working as expected, `maxWidth` is internally mapped to the `width` of the dropdown, so that:

* with no overflowing item and no `maxWidth`, the dropdown's `min-width` will kick-in
* with overflowing item(s) and no `maxWidth`, the dropdown width will grow to accomodate the larger item
* with no overflowing item and with `maxWidth` set to any value, that value will be the width of the dropdown
* with overflowing item(s) and with `maxWidth` set to any value, that value will be the width of the dropdown and there will be a scrollbar

packages/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -145,35 +149,27 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
       ...props
     } = this.props;
     /* eslint-enable @typescript-eslint/no-unused-vars */
+    const overflowProps =
+      maxWidth || maxHeight
+        ? { style: { width: maxWidth, maxHeight, overflow: 'auto', minWidth: maxWidth ? 'auto' : undefined } }
```
mcoker, line 154: If an invalid `maxWidth` is used, the menu still gets `overflow: auto` and `min-width: auto`. Is that intentional? It still changes the presentation of the menu in that case.

We can always add this support in core and provide a variable for the react component (or a user) to customize.

riccardo-forina: It would be a bit hard to check for valid values, I think this kind of check isn't even done by CSS-in-JS libs, is it?

And yeah, support for this on the core css-with a modifier I guess?-would be cool.

mcoker: > And yeah, support for this on the core css-with a modifier I guess?-would be cool.

That's how I would like to approach it if we can!  Same for the `maxHeight` prop.

We could have something like

```
.pf-m-scrollable/contrain-height {
    max-height: var(--pf-c-select__menu--MaxHeight); // defaults to something reasonable specified by design. Maybe 300px?
    overflow: auto;
}

.pf-m-contrain-width {
    width: auto;
    max-width: var(--pf-c-select__menu--MaxWidth); // defaults to auto maybe? The menu would be the size of the widest item in the menu
    overflow: auto;
}
```

The component could support a new prop (`scrollable/constrainHeight`) that would only apply the `pf-m-scrollable/constrain-height` modifier. And it would automatically apply `.pf-m-scrollable/constrain-height` if `maxHeight` is used, and use something like `.setProperty('--pf-c-select__menu--MaxHeight', maxHeight)`, to not break the API.

Same for `.pf-m-constrain-width`. wdyt? cc @tlabaj

riccardo-forina: What about backward compatibility for the `maxHeight` prop? I guess we could keep the prop to override the `--pf-c-select__menu--MaxHeight` var you proposed, which I guess it's the same thing `scrollable/constrainHeight` would do?

riccardo-forina: Makes sense. How do you want to proceed? Should I open a pr to add the modifiers on the PatternFly repo?

mcoker: When are you aiming to have this feature implemented?

riccardo-forina: Sorry got distracted by other stuff. I'm checking the `patternfly` repo dev process, I have to digest it a bit. Feel free to jump in for this if you want :D

mcoker: I'm happy to do it! I was curious when you're trying to get this feature by? We have a release on June 24, but it may be hard to get this feature in by then. It may be more realistic to target this in for the release after that, which is in the middle of July. Would that be OK?

riccardo-forina: Ah, sure that would work, I have no urgent need. On Atlasmap I applied a workaround but I figured it could be a common enough case to open a PR.

tlabaj:
Closing this PR.  Once the work for https://github.com/patternfly/patternfly/issues/3177 is done in core, we will follow up with a new PR.

PR #4308, opened 2020-05-29
chore(Select): replace all uses of value as default id of SelectOption

### LINKED ISSUE #4217: Select - menu item ID and checkbox select bugs
Looks like the `id` assigned to select menu items is based off of the item text, and this leads to invalid values for `id`

Screen Shot 2020-05-07 at 2 48 58 PM
Screen Shot 2020-05-07 at 2 47 42 PM

Also an issue reported by a user is with checkbox selects (where the `pf-c-select__menu-item` element is `label.pf-c-check`), if you set a `maxHeight` on the select that is shorter than the select menu (so that the menu scrolls), and you scroll down in the menu a bit, if you click anywhere in the item that is not the `<input type="checkbox">`, the scroll position in the menu is reset and you're taken to the top of the menu. Clicking on the checkbox seems to work fine. 

May-07-2020 14-40-25

nicolethoen:
**What**: Partially addresses #4217 

replace all uses of value as default id of SelectOption


packages/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -53,31 +53,34 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
     hasInlineFilter: false
   };
 
-  extendChildren() {
+  extendChildren(randomId: string) {
     const { children, isGrouped } = this.props;
     const childrenArray: React.ReactElement[] = children as React.ReactElement[];
     if (isGrouped) {
       let index = 0;
       return React.Children.map(childrenArray, (group: React.ReactElement) =>
```
redallen, line 61: Isn't index the second parameter? `(group: React.ReactElement, index: number)`

packages/react-core/src/components/Select/SelectMenu.tsx
```diff
@@ -100,14 +103,15 @@ export class SelectMenu extends React.Component<SelectMenuProps> {
...
             >
-              {group.props.children.map((option: React.ReactElement) =>
-                React.cloneElement(option, {
-                  isChecked: checked && checked.includes(option.props.value),
-                  sendRef,
-                  keyHandler,
-                  index: index++
-                })
-              )}
+              {group.props.children &&
```
redallen, line 106: Can be `React.Children.map(group.props.children, (option: React.ReactElement) => ...)`

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -88,6 +88,8 @@ export interface SelectProps
   inlineFilterPlaceholderText?: string;
   /** Custom text for select badge */
   customBadgeText?: string | number;
+  /** */
```
redallen, line 91: Prefix for `<input id="">` of checkbox variant

PR #4310, opened 2020-05-29
fix(breakpoints): make API consistent and appear in docs

redallen:
Use consistent API for `formatBreakpointMods` by making each modifier a prop. Props use objects whose values are strings instead of enums for better documentation. Also make values consistently `camelCased` instead of `dasherized-case`.

Before:
```js
breakpointMods={[
  { modifier: 'visible' }
  { modifier: 'hidden', breakpoint: 'lg' }
]}
```

After:
```js
visibility={{
  default: 'visible',
  lg: 'hidden'
}}
```

Now the docs correctly document modifiers:
image

## Breaking changes
- **Flex:**
  - Removed `breakpointMods` prop from `Flex` in favor of `spacer`, `spaceItems`, `grow`, `shrink`, `flex`, `direction`, `alignItems`, `alignContent`, `alignSelf`, `align`, `justifyContent`, `display`, `fullWidth` and `flexWrap`
  - Removed `breakpointMods` prop from `FlexItem` in favor of `spacer`, `grow`, `shrink`, `flex`, `alignSelf`, `align`, and `fullWidth`
- **Page:**
  - Removed prop `hasNoPadding` in favor of `padding={{ default: 'noPadding' }}` to prevent conflicts when setting both `hasNoPadding` and `padding={{ default: 'padding' | 'noPadding' }}`
- **Toolbar:**
  - Removed `breakpointMods` prop from `ToolbarItem` in favor of `visiblity`, `alignment`, and `spacer`
  - Removed `breakpointMods` prop from `ToolbarToggleGroup` in favor of `visiblity`, `alignment`, `spacer`, and `spaceItems`
  - Removed `breakpoint` prop from `ToolbarToggleGroup` in favor of `show`


redallen:
@tlabaj brought up a good point on Slack:
> I am concerned that it is not intuitive for the consumer to set a modifier independent of a breakpoint. The name of the prop "breakpointMods" does not give any information. I actually had not idea that we could use the breakpointMods prop to set a modifier without a breakpoint until you told me.

I'm going to modify the API more.

nicolethoen:
I like it so far :) interested to see the name changes

redallen:
@nicolethoen I updated the PR body with the new API.

@tlabaj Please re-review!

packages/react-core/src/components/DataList/__tests__/DataList.test.tsx
```diff
@@ -163,7 +162,7 @@ describe('DataList', () => {
   test('DataListAction visibility - hide button on 2xl', () => {
     const view = shallow(
       <DataListAction
-        className={css(DataListActionVisibility.hiddenOn2Xl)}
+        visibility={{ default: 'hidden', '2xl': 'hidden' }}
```
jschuler, line 165: the default should be visible here, no?

redallen: Yeah, the `default` shouldn't be there. Will update in next commit that also bumps core.

packages/react-core/src/components/Toolbar/ToolbarToggleGroup.tsx
```diff
@@ -7,32 +7,73 @@ import { ToolbarContext, ToolbarContentContext } from './ToolbarUtils';
...
+  visiblity?: {
+    default?: 'hidden' | 'visible';
+    md?: 'hidden' | 'visible';
+    lg?: 'hidden' | 'visible';
+    xl?: 'hidden' | 'visible';
+    '2xl'?: 'hidden' | 'visible';
+  };
+  /** Controls when filters are shown and when the toggle button is hidden. */
+  show?: {
+    default?: 'show';
```
jschuler, line 25: this only has show?

redallen: Yep, it replaces the `breakpoint` prop! Similarly Flex has single `grow` and `shrink` props. I think it's fine since in most cases they're strings and to keep it consistent.

redallen:
1. The purpose is just to be consistent with other breakpoint properties in case we want to add more values.
2. I think it's React-like to use `cssVariableName` rather than `css-variable-name`. What if we replaced `_2xl` with just `2xl`?

packages/react-core/src/layouts/Flex/Flex.tsx
```diff
@@ -3,25 +3,368 @@ import { css } from '@patternfly/react-styles';
...
+    default?: 'shrink';
+    sm?: 'shrink';
+    md?: 'shrink';
+    lg?: 'shrink';
+    xl?: 'shrink';
+    '2xl'?: 'shrink';
+  };
+  /** Value to add for flex property at various breakpoints */
+  flex?: {
+    default?: 'flexDefault' | 'flexNone' | 'flex_1' | 'flex_2' | 'flex_3' | 'flex_4';
```
jschuler, line 158: There are also these flex_1 etc values and the width_33 values above, WDYT, camelCase them as well? These and the width values do look very readable, so either way is fine with me

redallen: Ah, I forgot about these guys! I think they're fine with underscores. The `width33` in particular is harder to read.

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
```diff
@@ -4,51 +4,27 @@ import { css } from '@patternfly/react-styles';
...
   /** Additional classes added to the drawer. */
   className?: string;
   /** Content to be rendered in the drawer panel. */
   children?: React.ReactNode;
   /** Flag indicating that the drawer panel should not have a border. */
   hasNoBorder?: boolean;
-  /** Default width for drawer panel */
-  breakpointMods?: DrawerBreakpointMod[];
+  /** Width for drawer panel at various breakpoints */
+  widths?: {
```
tlabaj, line 15: Can you change this to width instead.

packages/react-core/src/components/Page/PageSection.tsx
```diff
@@ -56,17 +54,15 @@ export const PageSection: React.FunctionComponent<PageSectionProps> = ({
...
   isFilled,
   ...props
 }: PageSectionProps) => (
   <section
     {...props}
     className={css(
       variantType[type],
-      hasNoPadding && styles.modifiers.noPadding,
-      formatBreakpointMods(breakpointMods, styles),
+      formatBreakpointMods(padding, styles),
```
tlabaj, line 65: can you change the name of this to describe what it is formatting.

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
redallen: No, because it will conflict with the `width?: string` from `React.HTMLProps<HTMLDivElement>` :(

packages/react-core/src/components/Page/PageSection.tsx
redallen: Do you mean change the name of the function or of the prop `padding`?

dlabrecq:
Props like `spacer`, `show`, etc. are not descriptive enough, IMO.

PR #4312, opened 2020-05-31
docs(Notification Drawer): Add streamlined notification drawer example to docs

### LINKED ISSUE #4160: NotificationDrawer: Add stripped down example of notification drawer.
Add an example of the notification drawer as used in openshift where neither read/unread nor kebabs are used

jcaianirh:
Add streamlined notification drawer example to docs (examples, integration app, and integration tests).

Provide the streamlined notification drawer to patternfly that we use in openshift. IE we don't use
the read/unread notifications so all notifications in the drawer are styled unread, all category
counts are read. Also removed kebabs in all components in this new example.

fix #4160
cc: @mcarrano 

@mcoker when moving all notifications in a category to unread, I get a vertical scrollbar. Would you mind taking a look please? See image below:

Screen Shot 2020-05-31 at 2 15 02 PM

### DIFF
--- a/packages/react-core/src/components/NotificationDrawer/examples/NotificationDrawer.md
+++ b/packages/react-core/src/components/NotificationDrawer/examples/NotificationDrawer.md
@@ -390,7 +390,7 @@ render() {
               There are currently no critical alerts firing. There may be firing alerts of other severities or silenced critical alerts however.
             </EmptyStateBody>
             <EmptyStatePrimary>
-              <Button variant="link">View all alerts</Button>
+              <Button variant="link">Action</Button>
             </EmptyStatePrimary>
            </EmptyState>
          </NotificationDrawerList>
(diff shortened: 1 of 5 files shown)


mcarrano:
@jcaianirh Thanks for adding this example to reflect the OpenShift implementation. I'm just struggling a bit with the naming and ordering of the examples in this case. Seems like the most basic example is what you are doing without any global actions or read/unread. Then we add unread notifications and grouping to give us more complex versions of the drawer. @mcoker wanted to get your thoughts on this. Should we be adding the new example to Core also and do you think we then need to refactor the examples at all or do you think this is OK?

As for the example itself, there was only one small nit I had @jcaianirh . Where there are 0 alerts to display and you are showing the empty state, I'm wondering why the 'View all alerts' action is there. What would that do?

jcaianirh:
> @jcaianirh Thanks for adding this example to reflect the OpenShift implementation. I'm just struggling a bit with the naming and ordering of the examples in this case. Seems like the most basic example is what you are doing without any global actions or read/unread. Then we add unread notifications and grouping to give us more complex versions of the drawer. @mcoker wanted to get your thoughts on this. Should we be adding the new example to Core also and do you think we then need to refactor the examples at all or do you think this is OK?

@mcarrano The most basic example labeled Basic is the notification drawer that does not have categories / groups. Its a single list of notifications. The Groups is obviously just adding grouping, and the streamlined (I hesitated to add the product because I don't think we usually reference that in pf) is the openshift product implementation of the drawer. It's really another way to implement the group notification drawer but by changing the props in a way that there are fewer actions, and no read/unread state but since its a variance from the core examples, I ordered it last.

With that, I'm open to changing names / ordering of the examples. I'm not sure how important the order is in this case because there isn't an obvious flow of least to most complex, but if you see a better ordering or naming I'm down to change it.

jcaianirh:
> As for the example itself, there was only one small nit I had @jcaianirh . Where there are 0 alerts to display and you are showing the empty state, I'm wondering why the 'View all alerts' action is there. What would that do?

@mcarrano sure, since there are multiple categories of notifications in this drawer (and depending on implementation, there may be notifications not shown in the drawer at all), so the view all will go to a list view of all notifications / alerts of all category / severity. Its part of the openshift design here: https://openshift.github.io/openshift-origin-design/designs/administrator/4.4/notification-drawer/#empty-state

mcarrano:
@jcaianirh While I think the usage of "View all alerts" in Openshift makes sense, I'm thinking it would be better to remove that since this seems to be highly product specific and seems to make sense if only if the product has a global alerts view. Or if you think it could be helpful to include a placeholder, just change it to something generic like "Action." That way we are not implying anything specific about how this should work. It's left to the consumer to customize this to their needs. Do you think that makes sense?

jcaianirh:
@mcarrano changed action text to Action as requested.

mcoker:
@jcaianirh we have a PR up to fix the unread/scrollbar issue here - https://github.com/patternfly/patternfly/pull/3149

redallen:
@tlabaj Is this ready to go in?

PR #4317, opened 2020-06-02
Topology components sync with OpenShift Console

wojta:
Another round of sync with https://github.com/openshift/console/tree/master/frontend/packages/topology


wojta:
@redallen I suppose there shouldn't be any problem with the code.

redallen:
@wojta No sir! We're just in code freeze now for the next day until RC2 gets promoted.

PR #4231, opened 2020-05-10
feat(table): support row checkbox disable

boaz0:
**What**: Closes #4233

PR #4320, opened 2020-06-02
fix(table): update property name to avoid error

seanforyou23:
**What**: A follow up from recent change to selectable transform for Table. Fixed the below error

Screen Shot 2020-06-02 at 9 54 02 AM

The DOM was getting the correct attribute anyway, just that now we don't get the console error.

**Additional issues**: https://github.com/patternfly/patternfly-react/pull/4231


nicolethoen:
well... maybe I should wait for it to finish building...

PR #4332, opened 2020-06-04
Fix broken links and references in README files caused by moving patternfly-3 packages to their own branch

mturley:
When we moved away from the `packages/patternfly-3` and `packages/patternfly-4` directory structure, we left references to it in various README files. This PR fixes those broken links and removes patternfly-3-specific info from the contributing guide since it has its own contributing guide on the patternfly-3 branch.

We may also want to fix similar references on the patternfly-3 branch.

### DIFF
--- a/packages/react-integration/README.md
+++ b/packages/react-integration/README.md
@@ -1,13 +1,14 @@
 # Integration testing for PatternFly 4 React
 
 ## Overview
+
 PatternFly 4 React provides an integration testing demo app as an environment to verify and test PatternFly 4 code changes and contributions.
 
 The demo app is written in TypeScript and includes Cypress. We use Cypress for integration testing. If you’re new to Cypress and want to learn more, you can start with their documentation.
 
-* [Cypress Documentation](https://docs.cypress.io/guides/overview/why-cypress.html#In-a-nutshell)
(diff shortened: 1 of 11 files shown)


packages/react-integration/README.md
```diff
@@ -53,10 +55,10 @@ Run the following command from the root directory to launch Cypress:
 
 You can leave both the demo app and Cypress running while you develop your tests. You may receive notifications from Cypress that ask you to refresh as you make changes. Use the “Try again” button to refresh Cypress with your updated tests.
 
-Add any new integration tests you write to the [integration folder here](https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/react-integration/cypress/integration).
-
+Add any new integration tests you write to the [integration folder here](https://github.com/patternfly/patternfly-react/tree/master/packages/react-integration/cypress/integration).
```
seanforyou23, line 58: Is it possible to use a relative path here? I think they generally work better when navigating the directory structure in someone's fork of the repo, makes it easier to preview your changes before submitting the PR.

mturley: Hmm, yeah we should be able to. I just did a find-and-replace, didn't notice some of them aren't relative.

mturley: Pushed an update to use relative paths here and in a few other places!

PR #4340, opened 2020-06-04
fix(Select): fix panel issue

### LINKED ISSUE #3768: Select: Panel example inaccessible by keyboard
When navigating to the panel example, if you press enter or space on the panel, you get the following error: 
Screen Shot 2020-02-18 at 10 34 34 AM

This is a bug.

DaoDaoNoCode:
**What**: Closes #3768

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -150,7 +150,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       this.refCollection[0] = this.filterRef.current;
     }
 
-    if (!prevState.openedOnEnter && this.state.openedOnEnter) {
+    if (!prevState.openedOnEnter && this.state.openedOnEnter && !this.props.customContent) {
       this.refCollection[0].focus();
     }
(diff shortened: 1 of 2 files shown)


DaoDaoNoCode:
@redallen PTAL, thanks.

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -548,10 +548,12 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
             hasClearButton={hasOnClear}
           >
             {customContent && (
-              <div className={css(styles.selectToggleWrapper)}>
-                {toggleIcon && <span className={css(styles.selectToggleIcon)}>{toggleIcon}</span>}
-                <span className={css(styles.selectToggleText)}>{placeholderText}</span>
-              </div>
+              <React.Fragment>
```
seanforyou23, line 551: Is the fragment necessary?

DaoDaoNoCode: No, it's not necessary. I will remove it, thanks.

Issue #2790, opened 2019-08-28
Dropdown menu in modal gets clipped or adds scrollbars to modal


christianvogt:
Within a modal, the dropdown is causing our modals to have scrollbars when ideally the menu would be shown over the modal.

It would be better if the dropdown menu was parented to the body instead of within the same container as the dropdown. Popperjs could be used for dropdown to help with positioning and alignment.

There is a similar concern with dropdowns from patternfly base:
https://github.com/patternfly/patternfly-next/issues/2194

rachael-phillips:
Hi @christianvogt thanks for filing! A few quick questions for you.

1. Please provide the steps to reproduce so we can better understand and address the issue. Feel free to link CodePen or another tool:

2. Is this bug currently blocking you or is there a workaround?

3. What is your product and what release version are you targeting?

4. Is this issue for PF3 or PF4?

rachael-phillips:
Hi @christianvogt I'm following up here. Is this still an issue after #2418 was merged?

christianvogt:
@rachael-phillips yes this is still a problem.

I don't have a simple way to create a codepen without patternfly being hosted on a CDN. So here's a snippet I tried in your live documentation by combining the modal and dropdown examples:

![dropdown in modal](https://user-images.githubusercontent.com/14068621/64967599-d438b880-d86e-11e9-9322-04a37a9b3b63.gif)


```
import React from 'react';
import { Modal, Button, Dropdown, DropdownToggle, DropdownItem, DropdownSeparator, DropdownPosition, DropdownDirection, KebabToggle } from '@patternfly/react-core';
import { ThIcon, CaretDownIcon } from '@patternfly/react-icons';

class SmallModal extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      isModalOpen: false,
      isOpen: false
    };
    this.handleModalToggle = () => {
      this.setState(({ isModalOpen }) => ({
        isModalOpen: !isModalOpen
      }));
    };
    this.onToggle = isOpen => {
      this.setState({
        isOpen
      });
    };
    this.onSelect = event => {
      this.setState({
        isOpen: !this.state.isOpen
      });
    };
  }

renderDropdown() {
    const { isOpen } = this.state;
    const dropdownItems = [
      <DropdownItem key="link">Link</DropdownItem>,
      <DropdownItem key="action" component="button">
        Action
      </DropdownItem>,
      <DropdownItem key="disabled link" isDisabled>
        Disabled Link
[...]

Issue #4210, opened 2020-05-07
Popover interaction inconsistency


seanforyou23:
**Describe the issue. What is the expected and unexpected behavior?**

Popover doesn't forward focus correctly to close btn unless you use the enter key to launch it. 

VoiceOver (VO) screen readers use `cmd+option+space` as the primary way of selecting a thing. It's not uncommon to also use the enter key, but VO instructs the former when you focus on a clickable element. What I see is that a blind user wouldn't know the Popover was correctly launched when they use the `cmd+option+space` bar to launch it. The popover content is, by design, outside of the normal page content flow. Because the popovers are not live regions, nothing is announced other than the fact that the button is now "expanded" which is just a change in state of the button - not an indication or announcement of the new dynamic content. When new dynamic content appears on the screen, we generally want to send focus to it

In this state, the user hears the popover is expanded so they know its been launched, but unable to navigate to the content except through the rotor menu. This means they have to "figure out" that they need to press enter to get it to forward the VO cursor into the popover so that they can read the content like normal, or switch to another method of navigation to find the new content.

`cmd+options+space` fires a _click_ event on the focused element. If we can listen for that also, it could bring further alignment between the sighted and accessible experiences. 

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
Click a [Popover host element](https://www.patternfly.org/v4/documentation/react/components/popover#basic), notice it doesn't send focus to the Popover close button. Then, keyboard focus that same element, and press enter on your keyboard. Notice the focus is sent to the dynamic content.

<img width="454" alt="Screen Shot 2020-05-07 at 10 54 41 AM" src="https://user-images.githubusercontent.com/5942899/81309680-35590a80-9051-11ea-9b26-8efeb6fe00a4.png">

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?** Considering we generally want alignment between sighted and unsighted experiences, I'm inclined to call this a bug.

**What is your product and what release version are you targeting?** Ongoing PF development

Ping @jessiehuff @mcarrano

PR #4348, opened 2020-06-05
Allow dropdown to be appended to a higher DOM element

jschuler:
Closes https://github.com/patternfly/patternfly-react/issues/2790

Adds a new prop to `Dropdown`: `menuAppendTo`.
Defaults to 'inline` which is the same as the previous behavior and inlines the menu next to the toggle.

Can set to another DOM element to break out of the dropdown DOM element (e.g. if the menu is causing scrollbar in the parent container or it is cut off)
Some example usages:
`<Dropdown menuAppendTo={document.body} />`
`<Dropdown menuAppendTo="parent" />`
`<Dropdown menuAppendTo="inline" />`
`<Dropdown menuAppendTo={document.getElementById('targetId')} />`
`<Dropdown menuAppendTo={() => document.querySelector('div.#target')} />`

Additionally, there was no way to handle the `Modal` component's handling of the ESC key. So if the Dropdown was open, hitting the ESC key would just close the entire modal instead of just the Dropdown menu first. Added `onEscapePress` prop to Modal to support that.

2 examples added:
https://patternfly-react-pr-4348.surge.sh/documentation/react/components/dropdown#append-menu-document-body
https://patternfly-react-pr-4348.surge.sh/documentation/react/components/modal#with-dropdown

- [x] Can add the menu to any element
- [x] Default behavior is old behavior of inlining the menu
- [x] Keyboard navigation still works, using Enter key on the toggle opens up the menu with the first item focused
- [x] ESC press while dropdown menu is open closes the menu
- [x] Can handle Modal ESC press behavior yourself


jschuler:
@redallen I am planning to look at those components next. While looking at those if I can lift out common code I will do so, but it might be more complex due to different implementations and handling of keyboard presses

packages/react-core/src/components/Dropdown/DropdownToggle.tsx
```diff
@@ -18,6 +18,8 @@ export interface DropdownToggleProps extends React.HTMLProps<HTMLButtonElement>
   onToggle?: (isOpen: boolean) => void;
   /** Element which wraps toggle */
   parentRef?: HTMLElement;
+  /** The menu element */
+  getMenuRef?: any;
```
tlabaj, line 22: More of a question here. Is this expected to be a function?  To me the name implies function. If it is a function maybe update description.  If not maybe update name to `menuRef`?  
Is the any type to allow for Element or function?

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
```diff
@@ -82,7 +100,35 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
...
+          const popoverContent = (
+            <div
+              className={css(
+                baseClass,
+                direction === DropdownDirection.up && styles.modifiers.top,
+                position === DropdownPosition.right && styles.modifiers.alignRight,
+                isOpen && styles.modifiers.expanded,
+                className
+              )}
+              style={{ position: 'absolute' }}
```
tlabaj, line 126: is the plan to add css in core for this?  Or is this a temporary solution?

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
```diff
@@ -98,28 +144,45 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
...
+              interactive
+              interactiveBorder={0}
+              maxWidth="none"
+              distance={0}
+              appendTo={menuAppendTo}
+              boundary="window"
+              flip={false}
+              placement={this.getPlacement(position, direction)}
+              hideOnClick={false}
+              theme="pf-popover"
```
tlabaj, line 176: I wonder if we should use `css(styles...)` here?

packages/react-core/src/components/Modal/Modal.tsx
```diff
@@ -44,6 +44,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
   hasNoBodyWrapper?: boolean;
   /** An ID to use for the ModalBox container */
   id?: string;
+  /** Modal handles pressing of the Escape key and closes the modal. If you want to handle this yourself you can use this callback function */
+  onEscapePress?: (event: KeyboardEvent) => void;
```
tlabaj, line 48: Can you add a this prop to a demo and add am integration test?

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
jschuler: Tippy has its own set of css and overrides on that css in react-styles which is outside the realm of PF core. I could add a class there that targets the dropdown element, but might be more trouble than it's worth

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
jschuler: not needed since it is just one class, and i cannot use styles object since tippy decided it needs to add the `-theme` suffix to the passed in theme class here. This class (`pf-popover-theme` lives in the tippy-overrides.css in react-styles

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
mcoker: What if we added a modifier to the `.pf-c-dropdown` container that allows us to target specific styles when it's used in this context (only the `.pf-c-dropdown__menu` item is inside of it), including adding this style and anything else that might come up.

FWIW, `position: absolute` fixes an alignment issue due to `.pf-c-dropdown` being `inline-block`, and then the `.pf-c-dropdown__menu` element inside of it is absolutely positioned relative to `.pf-c-dropdown`. The default alignment for an inline block element is `vertical-align: bottom;`, so it's positioning it in the bottom of the containing block. You could also fix it by adding `vertical-align: top;` to `.pf-c-dropdown`, but if the dropdown isn't being displayed inline with content, using `display: block` is probably the simplest fix here. I made a quick demo that shows the problem.  https://codepen.io/mcoker/pen/ZEQbyyY

jschuler: Hey @mcoker , thanks for taking a look! I was curious why the menu was offset unless i added absolute positioning to it. Seems like there are a variety of ways to solve this. Is it fine to keep this fix (position: absolute) temporarily until we have a core modifier class?

jschuler: using `display: block` as the temporary solution until https://github.com/patternfly/patternfly/issues/3156 can be brought into react

seanforyou23:
Nice! The only thing I wanted to mention is that it exhibits the same behavior quirk mentioned in https://github.com/patternfly/patternfly-react/issues/4210 which I _think_ is leading to the new dynamic content not properly being announced to assistive technology. Other than that it looks good to me!

PR #4362, opened 2020-06-09
fix(NavExpandable): add style.modifiers.expandable

### LINKED ISSUE #4360: Extra lines added between nav items in an expandable subnav list
The lines between the subnav items shouldn't be there. The parent `li.pf-c-nav__item` that serves as an expandable item with a subnav list inside needs the class `.pf-m-expandable`

Screen Shot 2020-06-09 at 9 35 26 AM

cc @tlabaj @mcarrano

redallen:
**What**: Closes #4360

**Additional issues**:


redallen:
@tlabaj Updated.

PR #4375, opened 2020-06-11
Chart: Tooltip legend updates

dlabrecq:
A couple updates for the chart tooltip legend. 

- Updated the stacked area example to use a legend
- Renamed `ChartLegendTooltipLegend` (sounds weird) as `ChartLegendTooltipContent`
- Tooltip flyout should  be overridable
- Added `isCursorTooltip` prop so `ChartTooltip` can be used instead of `ChartCursorTooltip`


redallen:
> Renamed ChartLegendTooltipLegend (sounds weird) as ChartLegendTooltipContent
Isn't this a breaking change? How old is ChartLegendTooltipLegend?

dlabrecq:
ChartLegendTooltipLegend is new and hasn't been promoted -- just don't like the name.

PR #4376, opened 2020-06-11
chore(charts): Doc update

dlabrecq:
Replaced old DefinitelyTyped comments with links to victory types


redallen:
I'm fixing the a11y tester right now. It's flaky.

PR #4384, opened 2020-06-15
Remove all references to FlexModifiers

DarrienG:
**What**: FlexModifiers no longer exists and results in a compile error if used in
a new project. Out it goes.

**Additional issues**:

No


DarrienG:
Issue: https://github.com/patternfly/patternfly-react/issues/4392

PR #4385, opened 2020-06-15
Set modal size in docs -> variant

DarrienG:
**What**: Modal doesn't use size anymore 🚮

**Additional issues**:

No


DarrienG:
Made issue documenting here: https://github.com/patternfly/patternfly-react/issues/4393

PR #4386, opened 2020-06-15
Ensure PagerHeaderToolsGroup visibility mods are optional

DarrienG:
**What**: 

In the official demos here:
https://www.patternfly.org/v4/documentation/react/demos/pagelayout

You can see from the horizontal nav example that only the default and lg
props are specified.

This is legal in JS and I assume is how it is expected to work. Set a
default and add exceptions for other sizes you'd like for other
properties.

In ts, there is a hard requirement on setting each of the properties
because it is typechecked unlike JS which is super tedious (especially
if you just want something to always be visible).

This brings the convenience of the JS examples to TS as well.

**Additional issues**:

No

### DIFF
--- a/packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
+++ b/packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
@@ -10,7 +10,7 @@ export interface PageHeaderToolsGroupProps extends React.HTMLProps<HTMLDivElemen
   className?: string;
   /** Visibility at various breakpoints. */
   visibility?: {
-    default: 'hidden' | 'visible';
+    default?: 'hidden' | 'visible';
     sm?: 'hidden' | 'visible';
     md?: 'hidden' | 'visible';
     lg?: 'hidden' | 'visible';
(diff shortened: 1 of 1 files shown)


DarrienG:
Added issue here: https://github.com/patternfly/patternfly-react/issues/4394

packages/react-core/src/components/Page/PageHeaderToolsGroup.tsx
```diff
@@ -11,11 +11,11 @@ export interface PageHeaderToolsGroupProps extends React.HTMLProps<HTMLDivElemen
   /** Visibility at various breakpoints. */
   visibility?: {
     default: 'hidden' | 'visible';
```
redallen, line 13: This should be optional too, no?

DarrienG: I figured the intent was that default is supposed to be enforced and exceptions specified, but now looking at how some of the other components work:

      <Flex direction={{}}>

Seems like default is generally optional. Let me update real quick...

DarrienG:
Yeah of course! It's a great library and I appreciate the work you folks have done on it as well.

Last commit should fix review comments.

PR #4387, opened 2020-06-15
Import TitleSizes in Title demo

DarrienG:
**What**: Include missing import in Title docs.

No issue for this that I know of.

Side Q - has there been any thought to having CI compile doc code? Sometimes it doesn't compile and there are warnings for it like extra imports that could be removed. I've been a little spoiled by [rust docs that do it automatically](https://doc.rust-lang.org/rustdoc/documentation-tests.html). 

**Additional issues**:

No


DarrienG:
Issue here: https://github.com/patternfly/patternfly-react/issues/4395

PR #4388, opened 2020-06-15
Fix/no anonymous functions

### LINKED ISSUE #4369: Make anonymous function components named
In all our projects, code like:
```tsx
export const AboutModalBoxCloseButton = ({
  className = '',
  'aria-label': ariaLabel = 'Close Dialog',
  ...props
}: AboutModalBoxCloseButtonProps) => ( ... );
```
gets compiled to anonymous CJS arrow functions:
```js
exports.AboutModalBoxCloseButton = (_a) => { ... };
```

However, for Enzyme (which uses CJS), we'd like our code to get compiled down to named components so we can write selectors like `view.find(AboutModalBoxCloseButton)`.

### ~~Named functions~~
This has been ruled out in favor of display names.

~~An alternative way to write our arrow functions to accomplish this is like:~~
```tsx
export function AboutModalBoxCloseButton({
  className = '',
  'aria-label': ariaLabel = 'Close Dialog',
  ...props
}: AboutModalBoxCloseButtonProps) {
  return ( ...  );
}
```
~~which gets compiled to CJS as a named function:~~
```js
function AboutModalBoxCloseButton(_a) { ... }
```

~~The types change slightly from:~~
```ts
export declare const AboutModalBoxCloseButton: React.FunctionComponent<AboutModalBoxCloseButtonProps>;
```
~~to~~
```ts
export declare function AboutModalBoxCloseButton({ className, onClose, 'aria-label': ariaLabel, ...props }: AboutModalBoxCloseButtonProps): JSX.Element;
```

~~... but these are compatible and _should_ be okay (see [blog post](https://blog.echobind.com/react-with-typescript-components-as-function-declarations-vs-function-expressions-e433ac8d6938) and [React.FC typing README](https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/README.md#function-components))~~

### Display names

An alternative way to write our arrow functions to accomplish this is like:
```tsx
export const AboutModalBoxCloseButton = ({
  className = '',
  'aria-label': ariaLabel = 'Close Dialog',
  ...props
}: AboutModalBoxCloseButtonProps) => ( ... );
AboutModalBoxCloseButton.displayName = 'AboutModalBoxCloseButton';
```

This works everywhere and the function name won't get obfuscated by Webpack's terser or other build tools so that React dev tools can pickup component names in our docs.

redallen:
**What**: Add an eslint rule to disallow anonymous functions for Enzyme and React dev tools' sake. Closes #4369

**Additional issues**:

### DIFF
--- a/packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
+++ b/packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
@@ -1,0 +1,68 @@
+// https://astexplorer.net/#/gist/60c119aa79dab3a3e438cc762d21490f/e2cdd0547688a64eff1d8964156cfd8fce55d21d
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        "Don't use anonmous functions without a displayName so Enzyme and React devtools show component name",
+      category: 'Possible Errors',
+      recommended: false
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'object',
+        properties: {},
+        additionalProperties: true
+      }
+    ]
+  },
+  create(context) {
+    return {
+      ExportNamedDeclaration(node) {
+        if (!node.declaration || node.declaration.type !== 'VariableDeclaration') {
+          return;
+        }
+        const declaration = node.declaration.declarations['0'];
+        if (!declaration || !declaration.init || declaration.init.type !== 'ArrowFunctionExpression') {
+          return;
+        }
+        const typeAnnotation = declaration.id.typeAnnotation;
+        if (
+          typeAnnotation &&
+          typeAnnotation.typeAnnotation &&
+          typeAnnotation.typeAnnotation.typeName &&
+          typeAnnotation.typeAnnotation.typeName.left &&
+          typeAnnotation.typeAnnotation.typeName.left.name === 'React' &&
+          ['FunctionComponent', 'FC', 'SFC'].includes(typeAnnotation.typeAnnotation.typeName.right.name)
+        ) {
+          context.report({
+            node,
+            message: `${declaration.id.name}: Do not use an anonymous arrow function export. Use a named function export instead.`,
+            fix(fixer) {
+              const declarationNode = context.getSourceCode().getFirstToken(node.declaration);
+              const initBody = declaration.init.body;
+              const equalsToken = context.getSourceCode().getTokenAfter(declaration.id);
+              const fixes = [
+                fixer.replaceText(declarationNode, 'function'), // export (const|var) -> export function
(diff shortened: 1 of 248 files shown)


packages/react-core/src/components/DataList/DataListToggle.tsx
```diff
@@ -21,29 +21,34 @@ export interface DataListToggleProps extends React.HTMLProps<HTMLDivElement> {
   'aria-controls'?: string;
 }
 
-export const DataListToggle: React.FunctionComponent<DataListToggleProps> = ({
+/**
```
seanforyou23, line 24: What's the value in adding so many instances of this empty comment block? If it's only to satisfy eslint, we should handle it another way. We could either actually add text to these comments so they serve a purpose, or we could exclude this from being something that eslint requires. I think either way would be better than bloating the codebase with the empty comments.

test.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+const reactDocgen = require('react-docgen');
+
+const sourcePath = require.resolve('@patternfly/react-catalog-view-extension/src/components/CatalogTile/CatalogTileBadge.tsx');
+const sourceText = fs.readFileSync(sourcePath, 'utf8');
+
+
```
seanforyou23, line 7: Can remove space

test.js
```diff
@@ -0,0 +1,15 @@
...
+const parsedComponents = reactDocgen.parse(
+  sourceText,
+  reactDocgen.resolver.findAllExportedComponentDefinitions,
+  null,
+  { filename: sourcePath }
+);
+
+console.dir(parsedComponents, { depth: 4 })
```
seanforyou23, line 15: Can we add a newline?

packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
```diff
@@ -0,0 +1,68 @@
...
+            fix(fixer) {
+              const declarationNode = context.getSourceCode().getFirstToken(node.declaration);
+              const initBody = declaration.init.body;
+              const equalsToken = context.getSourceCode().getTokenAfter(declaration.id);
+              const fixes = [
+                fixer.replaceText(declarationNode, 'function'), // export (const|var) -> export function
+                fixer.remove(typeAnnotation), // remove : React.FunctionComponent<AboutModalBoxCloseButtonProps>
+                fixer.remove(equalsToken) // remove =
+              ];
+              // () => (<div></div>) -> function() { return (<div></div>) }
```
seanforyou23, line 50: Can we remove it?

redallen:
Closing this in favor of adding `displayName`s. That should address @jschuler 's feedback in Slack that the production builds are missing component names in the React devtools and @seanforyou23 's feedback about the empty comment blocks

Issue #4311, opened 2020-05-29
Table: Rendering issues of virtualized table causing performance and display issues in OpenShift 


jcaianirh:
**Describe the issue. What is the expected and unexpected behavior?**
Specifically, pf4 tables vs pf3 tables scrolling is noticeably degraded. The PF4 table scrolling is choppy and jumpy by comparison. 

This issue is reproducible with the Patternfly sample paes. The scrolling is re-rendering rows on every pixel movement. (See the screenshot attached) This re render did not occur on pf3 tables.

**Display of rerender of every row on scroll:**
![virtualscroller](https://user-images.githubusercontent.com/35978579/83284533-c6cc2000-a1aa-11ea-93dc-b003f643d6c4.gif)

**Display of render issue on filter**
![render-issue](https://user-images.githubusercontent.com/35978579/83285132-be281980-a1ab-11ea-8b3b-914eaa44b53f.gif)

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

**Is this a bug or enhancement? 
Openshift Bugs related: https://issues.redhat.com/browse/CONSOLE-1532
https://bugzilla.redhat.com/show_bug.cgi?id=1819752
https://bugzilla.redhat.com/show_bug.cgi?id=1795305
https://bugzilla.redhat.com/show_bug.cgi?id=1805918
cc: @janwright73

kmcfaul:
Re: the re-rendering, both pf3 and pf4 source their virtualized scrolling from the `react-virtualized` library with minor changes in pf4's [copy](https://github.com/patternfly/patternfly-react/pull/2011) that shouldn't impact performance. The examples in the docs have a small overscan index of 2 so that could be why they update more often. It is possible the difference lies in Table, so I am continuing to investigate.

Re: the filtering bug, I have not been able to reproduce this behavior so far after adding a filterable table demo to the virtualized docs (can be looked at in this [WIP PR](https://github.com/patternfly/patternfly-react/pull/4389) with [surge](https://patternfly-react-pr-4389.surge.sh/documentation/react/extensions/virtualizedtable#filterable)). Are you getting any console errors/warnings when the white space appears?

jcaianirh:
@kmcfaul with the filtering issue: no console errors or warnings when this occurs. But one other clue, is that subsequent scrolling on the page rerenders the component correctly, and the blank space is removed.

tlabaj:
closed by #4389

PR #4389, opened 2020-06-15
feat(react-virtualized-extension docs): add filterable demo, fix for filtering+scrolling

kmcfaul:
**What**: fix for filtering in #4311


kmcfaul:
Updated demo to include WindowScroller to better match the Openshift use case.

Discovered an (open bug)[https://github.com/bvaughn/react-virtualized/issues/1402] in `react-virtualized` that was causing the intermittent scroll gap when filtering. Reverting `VirtualGrid` to a previous iteration fixes this issue, though I'm unsure if it will introduce other bugs in the future.

PR #4390, opened 2020-06-15
improvement(Topology): Allows a promise to be returned by the create connector callback

jeff-phillips-18:
For instances where some computation may be necessary, allow a promise to be returned to the create connector callback.

/cc @christianvogt

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/Connectors.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/Connectors.tsx
@@ -210,7 +210,13 @@ export const CreateConnector = withTopologySetup(() => {
         }
         if (kind === ModelKind.node) {
           return withCreateConnector(
-            (source: Node, target: Node | Graph, event: DragEvent, choice: ColorChoice | undefined): any[] | null => {
+            (
+              source: Node,
+              target: Node | Graph,
+              event: DragEvent,
+              dropHints: string[],
+              choice: ColorChoice | undefined
+            ): ColorChoice[] | void => {
               if (!choice) {
                 return [{ label: 'Create Annotation', color: 'red' }, { label: 'Create Binding', color: 'green' }];
               }
@@ -246,7 +252,6 @@ export const CreateConnector = withTopologySetup(() => {
                 }
               });
               controller.fromModel(model);
-              return null;
             }
           )(
             withDndDrop<Node, any, { droppable?: boolean; hover?: boolean; canDrop?: boolean }, NodeProps>({
--- a/packages/react-topology/src/behavior/withCreateConnector.tsx
+++ b/packages/react-topology/src/behavior/withCreateConnector.tsx
@@ -6,7 +6,14 @@ import Point from '../geom/Point';
 import Layer from '../components/layers/Layer';
 import { ContextMenu, ContextMenuItem } from '../components/contextmenu';
 import { Node, isNode, AnchorEnd, GraphElement, isGraph, Graph } from '../types';
-import { DragSourceSpec, DragSourceMonitor, DragEvent } from './dnd-types';
+import {
+  DragSourceSpec,
+  DragSourceMonitor,
+  DragEvent,
+  DragObjectWithType,
+  DragSpecOperationType,
+  DragOperationWithType
+} from './dnd-types';
 import { useDndDrag } from './useDndDrag';
 
 export const CREATE_CONNECTOR_OPERATION = '#createconnector#';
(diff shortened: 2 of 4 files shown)


packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/Connectors.tsx
```diff
@@ -210,9 +210,18 @@ export const CreateConnector = withTopologySetup(() => {
...
         if (kind === ModelKind.node) {
           return withCreateConnector(
-            (source: Node, target: Node | Graph, event: DragEvent, choice: ColorChoice | undefined): any[] | null => {
+            (
+              source: Node,
+              target: Node | Graph,
+              event: DragEvent,
+              dropHints: string[],
+              choice: ColorChoice | undefined
+            ): Promise<any[] | null> => {
```
christianvogt, line 219:
```suggestion
            ): ColorChoice[] | void => {
```

packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/Connectors.tsx
```diff
@@ -210,9 +210,18 @@ export const CreateConnector = withTopologySetup(() => {
...
               if (!choice) {
-                return [{ label: 'Create Annotation', color: 'red' }, { label: 'Create Binding', color: 'green' }];
+                return Promise.resolve([
+                  { label: 'Create Annotation', color: 'red' },
+                  { label: 'Create Binding', color: 'green' }
+                ]);
```
christianvogt, line 224: Remove `Promise` from this example:
```suggestion
                return [{ label: 'Create Annotation', color: 'red' }, { label: 'Create Binding', color: 'green' }];
```

packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/Connectors.tsx
```diff
@@ -246,7 +255,7 @@ export const CreateConnector = withTopologySetup(() => {
                 }
               });
               controller.fromModel(model);
-              return null;
+              return Promise.resolve(null);
```
christianvogt, line 258: Remove this return statement as we can return void.
```suggestion
```

packages/react-topology/src/behavior/withCreateConnector.tsx
```diff
@@ -37,8 +37,15 @@ type CreateConnectorWidgetProps = {
...
+    dropHints?: string[] | undefined,
     choice?: ConnectorChoice
-  ) => ConnectorChoice[] | void | undefined | null | React.ReactElement[];
+  ) =>
+    | Promise<ConnectorChoice[] | void | undefined | null | React.ReactElement[]>
+    | ConnectorChoice[]
+    | void
+    | undefined
+    | null
+    | React.ReactElement[];
```
christianvogt, line 48: Add new type. No idea why we originally had `React.ReactElement[]` in the return type, so let's remove that as well.
```
type OnCreateResult = ConnectorChoice[] | void | undefined | null;
```

```suggestion
  ) => Promise<OnCreateResult> | OnCreateResult;
```

packages/react-topology/src/behavior/withCreateConnector.tsx
```diff
@@ -87,14 +94,28 @@ const CreateConnectorWidget: React.FC<CreateConnectorWidgetProps> = observer(pro
       end: (dropResult: GraphElement, monitor: DragSourceMonitor, dragProps: any) => {
```
christianvogt, line 94: Update `dragProps: CreateConnectorWidgetProps` to remove the use of `any`.
Also for the `drag` function above.

packages/react-topology/src/behavior/withCreateConnector.tsx
```diff
@@ -87,14 +94,28 @@ const CreateConnectorWidget: React.FC<CreateConnectorWidgetProps> = observer(pro
...
         const event = monitor.getDragEvent();
         if ((isNode(dropResult) || isGraph(dropResult)) && event) {
-          const choices = dragProps.onCreate(dragProps.element, dropResult, event);
-          if (choices && choices.length) {
-            setPrompt({ element: dragProps.element, target: dropResult, event, choices });
+          const handleResults = (choices: ConnectorChoice[]) => {
```
christianvogt, line 97: This `end` function could be simplified to this I believe:
```
      end: async (dropResult: GraphElement, monitor: DragSourceMonitor, dragProps: CreateConnectorWidgetProps) => {
        const event = monitor.getDragEvent();
        if ((isNode(dropResult) || isGraph(dropResult)) && event) {
          const choices = await dragProps.onCreate(dragProps.element, dropResult, event, monitor.getDropHints());
          if (choices && choices.length) {
            setPrompt({ element: dragProps.element, target: dropResult, event, choices });
            return;
          }
        }
        setActive(false);
        dragProps.onKeepAlive(false);
      },
```

PR #4391, opened 2020-06-15
Fix/no anonymous functions2

### LINKED ISSUE #4369: Make anonymous function components named

redallen:
**What**: Write eslint rule to enforce adding display names and add `displayName`s to 411 components missing them. Closes #4369

**Additional issues**: In favor of #4388

### DIFF
--- a/packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
+++ b/packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
@@ -3,7 +3,7 @@ module.exports = {
   meta: {
     docs: {
       description:
-        "Don't use anonmous functions without a displayName so Enzyme and React devtools show component name",
+        "Don't use arrow functions or classes without a displayName so Enzyme and React devtools show component name",
       category: 'Possible Errors',
       recommended: false
     },
--- a/packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
+++ b/packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
@@ -1,0 +1,82 @@
+// https://astexplorer.net/#/gist/60c119aa79dab3a3e438cc762d21490f/e2cdd0547688a64eff1d8964156cfd8fce55d21d
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        "Don't use anonmous functions without a displayName so Enzyme and React devtools show component name",
+      category: 'Possible Errors',
+      recommended: false
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'object',
+        properties: {},
+        additionalProperties: true
+      }
+    ]
+  },
+  create(context) {
+    return {
+      ExportNamedDeclaration(node) {
+        if (!node.declaration) {
+          return;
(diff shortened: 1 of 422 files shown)


packages/eslint-plugin-patternfly-react/lib/rules/no-anonymous-functions.js
```diff
@@ -0,0 +1,82 @@
+// https://astexplorer.net/#/gist/60c119aa79dab3a3e438cc762d21490f/e2cdd0547688a64eff1d8964156cfd8fce55d21d
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        "Don't use anonmous functions without a displayName so Enzyme and React devtools show component name",
```
jschuler, line 6:
```suggestion
        "Don't use arrow functions or classes without a displayName so Enzyme and React devtools show component name",
```

jschuler: Awesome, our docs now also show the component name in react dev tools, and the snapshots as well!

Issue #3219, opened 2019-10-28
Chart tooltip with colored legend


jotak:
Compared to PF3, the chart tooltips are much simpler (and I would say less pretty).
In Kiali we had an issue about that, as it's seen as a regression: https://github.com/kiali/kiali/issues/1792 (see also discussion here: https://github.com/kiali/kiali-ui/pull/1491 )

So we've ended up with a custom implementation: https://github.com/kiali/kiali-ui/pull/1491/commits/ca7a5f15f1ea5db165dec75d991dd14d00e2d83b

Not sure what you would want for Patternfly: having a similar implementation for better homogeneity, or leave this topic to the end-user implementation?

rachael-phillips:
Hi @jotak we've seen this, and we'll get back to your soon with a response!

mcarrano:
I am going to reopen this issue after discussion with @dlabrecq. Before we can commit to this, we need to perform further investigation and possibly create a PoC to understand what can be accomplished using current Victory chart tooltips of if we need enhancements from Victory (or if there is another way to do this). 

@tlabaj @LHinson @dlabrecq what's the best path forward on this? Sounds like we should get a React spike into our roadmap?

jotak:
For what it's worth, here what we've ended up using in Kiali:

![Capture d’écran de 2020-03-18 16-28-08](https://user-images.githubusercontent.com/2153442/76977530-b8a18e00-6935-11ea-8446-fda54f956f8f.png)

https://github.com/kiali/k-charted/blob/master/web/pf4/src/components/CustomTooltip.tsx

dlabrecq:
Thank you for sharing the Kiali code, it was very helpful. It potentially could be an example we add to the docs?

A couple of notes for a POC:

1. The tooltip symbol colors appear to use the line color. This would be in sync with the legend, but the symbol itself is hard coded as 'square'.

2. If a particular legend item / line uses the 'dash' symbol; for example, how would that sync with the tooltip?

3. Not clear what happens when one or more data points are not available? 

In Cost Management, we have multiple data sets, but one may have a day or more of unavailable data. In that case, Victory does omits that data from the tooltip (i.e., when using voronoiDimension="x").

4. Does Victory provide a null value to preserve the props.text order? If not, would that result in data associated with the wrong symbol?

5. If data should be unavailable, should the tooltip still display the full legend? What would the missing value be -- zero, dash, etc.?

[...]

PR #4398, opened 2020-06-16
chore(charts): align legend tooltip with flyout edge

dlabrecq:
While testing Cost Management with the new legend tooltip feature (yet to be promoted), I noticed that the alignment is off with very long labels.

This change better aligns legend components with the tooltip flyout edge. 

**Before:**
Screen Shot 2020-06-15 at 12 21 56 PM

**After:**
chrome-capture

Related to #3219


ddonahue007:
LGTM

PR #4399, opened 2020-06-16
feat(alert,notificationdrawer): add support to truncate title

### LINKED ISSUE #4357: Truncate prop for alert title and notification drawer item title
https://github.com/patternfly/patternfly/pull/3145/files

Adds an optional `pf-m-truncate` prop for the alert title (`.pf-c-alert__title`) and notification drawer entry titles (`.pf-c-notification-drawer__list-item-header-title`) that will truncate long strings with an `...`.

ALSO! We added a super cool thing that will allow a user to specify the number of lines they want to display before truncation happens.

For the alert title, you set the set the var `--pf-c-alert__title--max-lines` to the number of lines to show.

For the notification drawer item title, `--pf-c-notification-drawer__list-item-header-title--max-lines`.

You can also set it on the group toggle title (`.pf-c-notification-drawer__group-toggle-title`). That var is `--pf-c-notification-drawer__group-toggle-title--max-lines`

boaz0:
**What**: Closes #4357 

**Additional issues**:
 - Do not merge till `@patternfly/patternfly` is updated to `4.10.33`

//cc @mcoker

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -6,6 +6,7 @@ import accessibleStyles from '@patternfly/react-styles/css/utilities/Accessibili
 import { AlertIcon } from './AlertIcon';
 import { capitalize, getOUIAProps, OUIAProps } from '../../helpers';
 import { AlertContext } from './AlertContext';
+import maxLines from '@patternfly/react-tokens/dist/js/c_alert__title_max_lines';
 
 export enum AlertVariant {
   success = 'success',
@@ -38,6 +39,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   isLiveRegion?: boolean;
   /** If set to true, the time out is 8000 milliseconds.  If a number is provided, alert will be dismissed after that amount of time in milliseconds. */
   timeout?: number | boolean;
+  /** Truncate title to number of lines */
+  truncateTitle?: number;
 }
 
 export const Alert: React.FunctionComponent<AlertProps> = ({
@@ -54,6 +57,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   ouiaId,
   ouiaSafe = true,
   timeout = false,
+  truncateTitle = 0,
   ...props
 }: AlertProps) => {
   const getHeadingContent = (
@@ -64,7 +68,13 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   );
 
   const [disableAlert, setDisableAlert] = useState(false);
-
+  const titleRef = React.useRef(null);
+  React.useEffect(() => {
+    if (!titleRef.current || !truncateTitle) {
+      return;
+    }
+    titleRef.current.style.setProperty(maxLines.name, truncateTitle.toString());
+  }, [titleRef, truncateTitle]);
   const customClassName = css(
     styles.alert,
     isInline && styles.modifiers.inline,
@@ -94,7 +104,9 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
         })}
       >
         <AlertIcon variant={variant} />
-        <h4 className={css(styles.alertTitle)}>{getHeadingContent}</h4>
+        <h4 ref={titleRef} className={css(styles.alertTitle, truncateTitle && styles.modifiers.truncate)}>
+          {getHeadingContent}
+        </h4>
         {actionClose && (
           <AlertContext.Provider value={{ title, variantLabel }}>
             <div className={css(styles.alertAction)}>{actionClose}</div>
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawerGroup.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawerGroup.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/NotificationDrawer/notification-drawer';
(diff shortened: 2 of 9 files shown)


packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -48,9 +51,18 @@ export const Alert: React.FunctionComponent<AlertProps & OUIAProps> = ({
...
   ...props
 }: AlertProps & OUIAProps) => {
+  const titleRef = React.useRef(null)
+  React.useEffect(() => {
+    if(!Boolean(titleRef.current) || !Boolean(truncateTitle)) {
+      return ;
+    }
+    titleRef.current.style.setProperty(maxLines.name, truncateTitle.toString())
+  }, [titleRef, truncateTitle])
+
```
dlabrecq, line 65: Curious why you chose to use a `titleRef` Vs applying the modifier directly to `className`?

mcoker:
Does this add support for setting `--pf-c-notification-drawer__group-toggle-title--max-lines` to limit the number of lines shown in `.pf-c-notification-drawer__group-toggle-title`? Thats an element we don't have a truncate modifier for (it truncates by default), but we still support the var to change the number of lines of truncation.

mcoker:
This lgtm! Also to follow up on my original question, it looks like if you pass `0` as a value to the max-lines var, the text will not truncate. So I think you could apply the `truncate=number` prop to `.pf-c-notification-drawer__group-toggle-title`, too, and it would set `--pf-c-notification-drawer__group-toggle-title--max-lines` (without setting `.pf-m-truncate`, like the prop on the other components), and `0` could be used to allow the title not to trucnate.

packages/react-core/src/components/Alert/Alert.tsx
boaz0: @dlabrecq the use of `titleRef` is to inject a new value to the `*-title--max-lines` variable which indicates the number of lines to show in the title.

I could not find any way to set it as `className`. If you know a way to do so, let me know. :smiley_cat:

boaz0:
Hi @mcoker 
- I added support to `--pf-c-notification-drawer__group-toggle-title--max-lines` but it seems not to work - mind taking a look what I am missing? Maybe I set it on the wrong element? You can see it in the group example of the docs.
- I updated the default value to `0` as you suggested.

mcoker:
@boaz0 looks like the title element is missing the class `pf-c-notification-drawer__group-toggle-title`

boaz0:
resolved conflict.

dlabrecq:
@boaz0 Looks like you're still working on @mcoker 's suggestion? Let me know and I'll review, again.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -48,10 +51,19 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   title,
   children = '',
   className = '',
+  truncateTitle = 0,
   ouiaId,
   ouiaSafe = true,
   ...props
 }: AlertProps) => {
+  const titleRef = React.useRef(null);
```
karelhala, line 59: Unfortunately we can't use hooks in PF components due to older versions of React.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -48,10 +51,19 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
+  React.useEffect(() => {
+    if (!titleRef.current || !truncateTitle) {
+      return;
+    }
+    titleRef.current.style.setProperty(maxLines.name, truncateTitle.toString());
```
karelhala, line 64: Also might be better to limit just the `title` from props in `getHeadingContent` if it's a string. Otherwise consumer has to do the truncation.

packages/react-core/src/components/Alert/Alert.tsx
boaz0: No problem, I don't have to use Hooks for that. :+1:

packages/react-core/src/components/Alert/Alert.tsx
boaz0: I am not sure what you meant. But I cannot do that in `getHeadingContent` as it is wrapping `title` in `React.Fragment`:

```
    <React.Fragment>
      <span className={css(accessibleStyles.screenReader)}>{variantLabel}</span>
      {title}
    </React.Fragment>
```

I guess I misunderstood you. Can you give an example?

karelhala: Here you could trim the `title` if I am reading the code correctly. But as I said, that's my preference to trim the string instead of css rule. But both are doable!

packages/react-core/src/components/Alert/Alert.tsx
tlabaj: As of the breaking change release, we updated our React dependency to `react@^16.8.0` so we can now use hooks.

packages/react-core/src/components/NotificationDrawer/NotificationDrawerGroup.tsx
```diff
@@ -32,28 +35,39 @@ export const NotificationDrawerGroup: React.FunctionComponent<NotificationDrawer
...
+      {...props}
+      className={css(styles.notificationDrawerGroup, isExpanded && styles.modifiers.expanded, className)}
+    >
+      <h1>
+        <button
+          className={css(styles.notificationDrawerGroupToggle)}
+          aria-expanded={isExpanded}
+          onClick={e => onExpand(e, !isExpanded)}
+        >
+          <div ref={titleRef}>{title}</div>
```
mcoker, line 60: I think this needs to be 

```suggestion
          <div ref={titleRef} className={css(styles.notificationDrawerGroupToggleTitle)}>{title}</div>
```

tlabaj:
@boaz0 looks like you still need to make the update the @mcoker suggested.  Can you please make that change and resolve the conflicts.  Thanks!

Issue #4364, opened 2020-06-09
Create common API for changing where components are inserted


evwilkin:
See original discussion in https://github.com/patternfly/patternfly-react/issues/2790

Components may need to be appended into areas other than the default to avoid issues such as causing scroll bars when contained within a Modal.  Find a common API to solve this issue across relevant components:

- [x] Dropdown, Split Button (#4348, https://github.com/patternfly/patternfly-react/pull/4403)

- [x] Select (https://github.com/patternfly/patternfly-react/pull/4403)

- [x] Application Launcher (https://github.com/patternfly/patternfly-react/pull/4414)

- [x] Context Selector (https://github.com/patternfly/patternfly-react/pull/4414)

- [x] Options Menu (https://github.com/patternfly/patternfly-react/pull/4414)

- [x] Tooltip (https://github.com/patternfly/patternfly-react/pull/4491)

- [x] Popover (https://github.com/patternfly/patternfly-react/pull/4491)

- [ ] Topology uses their own popper.js implementation, see if it can be aligned

christianvogt:
+1 for fixing `Select` as well. I expected both to be addressed by #2790 and never dove into the code to realize that they aren't using the same foundation code.

rebeccaalpert:
Same here; this affects our modal conversion efforts as well.

evwilkin:
Bumped to 2020.10 milestone for newly identified Topology item above

Issue #4359, opened 2020-06-09
Tooltip: should use the `pf-c-tooltip__arrow` class for the tooltip arrow


KKoukiou:
When using the Tooltip from the React component the produced html is not the same as the html for the tooltip advertised in the documentation.

Specifically, the html code when the React component is used looks like this:

``` html
<div class="tippy-tooltip pf-tooltip-theme pf-c-tooltip" id="tippy-20" data-state="visible" tabindex="-1" data-arrow="" data-animation="fade" style="max-width: 18.75rem; top: -15px; left: 0px; transition-duration: 300ms;" role="tooltip" data-placement="top">
  <div class="tippy-arrow" style="left: 137px;"></div>
  <div class="tippy-content" data-state="visible" style="transition-duration: 300ms;"><div>
   <div class="" role="tooltip" id="">
     <div class="pf-c-tooltip__content">
        <div>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla turpis.
        </div>
      </div>
    </div>
  </div>
</div>
</div>
```
Note that there is not `pf-c-tooltip__arrow` class used for the tooltip arrow as advertised in the [html documentation for the tooltip.](https://www.patternfly.org/v4/documentation/core/components/tooltip)

Also there are some extra div wrapper in the React component, and tippy-arrow, tippy-content classes are not documented.

jschuler:
We use tippy.js to position the tooltip/popover/dropdown menu relative to the triggering element. It also lets as append the element to the document body so that it is not constrained by the element flow. 

The outer divs are produced by tippy, and so what was done here was to style tippy's arrow to match the pf-core arrow. I do agree that it would be great if the produced DOM matched the core DOM closer and that the arrow div had the `pf-c-tooltip__arrow` class applied, perhaps there is a way to change the output to achieve that (https://github.com/atomiks/tippyjs-react/issues/84#issuecomment-500675007).

jschuler:
I'll explore if the produced DOM can be improved

PR #4403, opened 2020-06-16
Use popper for Dropdown and Select when menu is appended to another element

jschuler:
Also towards fixing https://github.com/patternfly/patternfly-react/issues/4359

Up until now we've relied in Tippy.js for tooltip/popover/dropdown menu positioning. Tippy.js itself uses popper.js under the hood, but adds its own flair and styling.
This PR is the first step in transitioning over to popper.js directly, since we want to be in control of the produced DOM and we use our own styling from core.
The Dropdown and Select components make use of the new Popper component when the `menuAppendTo` prop is set to something other than `inline`.

A future PR will transition Tooltip and Popover components over to use popper as well, so we can remove the tippy.js dependency and its associated styles in the react-styles package.


packages/react-core/src/helpers/Popper/FindRefWrapper.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * - class based components we can assign our own ref to
+ * - functional components that have forwardRef implemented
+ * However, there is no guarantee that is what will get passed in as trigger/toggle in the case of tooltips and popovers
+ */
+export class FindRefWrapper extends React.Component<FindRefWrapperProps> {
+  static displayName = 'FindRefWrapper';
+  componentDidMount() {
+    // eslint-disable-next-line react/no-find-dom-node
+    const root = ReactDOM.findDOMNode(this);
+    this.props.onFoundRef(root);
```
redallen, line 22: Why do we need this component? Why not add a `componentDidMount` to `Popper.tsx`?

jschuler: I need to find 2 different refs for the toggle and the menu, it's cleaner to have this reusable class component, also Popper gets to be a functional components and i can use hooks such as the `usePopper` hook

redallen: Fair enough. There are conflicts now after merging #4308

PR #4408, opened 2020-06-17
feat(Banner): add banner component

### LINKED ISSUE #4244: Introduce banner component and demo
Placeholder for when https://github.com/patternfly/patternfly/issues/2727 is complete

kmcfaul:
**What**: Closes #4244 

Also includes core version bump to 4.12.0 for banner styling

### DIFF
--- a/packages/react-core/src/components/Banner/Banner.tsx
+++ b/packages/react-core/src/components/Banner/Banner.tsx
@@ -13,21 +13,20 @@ export interface BannerProps extends React.HTMLProps<HTMLButtonElement> {
   isSticky?: boolean;
 }
 
-const variantStyle = {
-  ['default']: '',
-  ['info']: styles.modifiers.info,
-  ['danger']: styles.modifiers.danger,
-  ['success']: styles.modifiers.success,
-  ['warning']: styles.modifiers.warning
-};
-
 export const Banner: React.FunctionComponent<BannerProps> = ({
   children,
   className,
   variant = 'default',
   isSticky = false
 }: BannerProps) => (
-  <div className={css(styles.banner, variantStyle[variant], isSticky && styles.modifiers.sticky, className)}>
+  <div
+    className={css(
+      styles.banner,
+      styles.modifiers[variant as 'success' | 'danger' | 'warning' | 'info'],
+      isSticky && styles.modifiers.sticky,
+      className
+    )}
+  >
     {children}
   </div>
 );
--- a/packages/react-core/src/components/Banner/Banner.tsx
+++ b/packages/react-core/src/components/Banner/Banner.tsx
@@ -7,9 +7,9 @@ export interface BannerProps extends React.HTMLProps<HTMLButtonElement> {
   children?: React.ReactNode;
   /** Additional classes added to the banner */
   className?: string;
-  /** Adds banner variant styles */
+  /** Variant styles for the banner */
   variant?: 'default' | 'info' | 'danger' | 'success' | 'warning';
-  /** Sticks the banner to the top of its container if true */
+  /** If set to true, the banner sticks to the top of its container */
   isSticky?: boolean;
 }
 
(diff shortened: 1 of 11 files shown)


packages/react-core/src/components/Banner/__tests__/Banner.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { Banner } from '../Banner';
+import React from 'react';
+import { mount } from 'enzyme';
+
+['default', 'info', 'success', 'warning', 'danger'].forEach(
+  (variant: 'default' | 'info' | 'success' | 'warning' | 'danger') => {
```
nicolethoen, line 6: Probably worth testing the isSticky prop?

kmcfaul:
Sure I'll add a demo & test for `isSticky`

packages/react-integration/demo-app-ts/src/components/demos/BannerDemo/BannerDemo.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Banner } from '@patternfly/react-core';
+
+export class BannerDemo extends React.Component {
+  static displayName = 'BannerDemo';
+  componentDidMount() {
+    window.scrollTo(0, 0);
+  }
+
+  render() {
```
tlabaj, line 10: Can you test the isSticky prop on one of these Banners?

packages/react-core/src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,34 @@
...
+  children?: React.ReactNode;
+  /** Additional classes added to the banner */
+  className?: string;
+  /** Variant styles for the banner */
+  variant?: 'default' | 'info' | 'danger' | 'success' | 'warning';
+  /** If set to true, the banner sticks to the top of its container */
+  isSticky?: boolean;
+}
+
+const variantStyle = {
```
tlabaj, line 16: rather than having this const can we use the css 'getModifier` function?

tlabaj: Sorry.  Did not realize that had been removed.  Let's follow what is done in Alerts.

packages/react-core/src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const Banner: React.FunctionComponent<BannerProps> = ({
+  children,
+  className,
+  variant = 'default',
+  isSticky = false
+}: BannerProps) => (
+  <div
+    className={css(
+      styles.banner,
+      styles.modifiers[variant as 'success' | 'danger' | 'warning' | 'info'],
```
tlabaj, line 25: do we need to check for `variant !== default` here?

kmcfaul: If `variant === default` no class gets applied, so I don't think we need to check for it.

packages/react-core/src/demos/PageLayout/examples/BannerDemo.md
```diff
@@ -0,0 +1,528 @@
...
+          mainContainerId={pageId}
+        >
+          <Banner isSticky>
+            <Flex
+              justifyContent={{ default: 'justifyContentCenter', lg: 'justifyContentSpaceBetween' }}
+              flexWrap={{ default: 'nowrap' }}
+            >
+              <div className="pf-u-display-none pf-u-display-block-on-lg">Localhost</div>
+              <div className="pf-u-display-none pf-u-display-block-on-lg">
+                This message is sticky to the top or bottom of the page.
```
nicolethoen, line 242: I think it's just a little confusing that the two demos, which appear to have banners sticking to three different parts of the page all have the same message that says, "This message is sticky to the top or bottom of the page." Maybe the messages should each be a little more specific and different?

kmcfaul: That's how they are in the core demo. I can update them to reference their specific location.

mcoker: Agree, we should update that in the core demo, too. If you want to make that change in this demo, core can update to match.

PR #4414, opened 2020-06-19
menuAppendTo prop in several more components

jschuler:

Makes `menuAppendTo` prop available in:
- [x] Application Launcher
- [x] Context Selector
- [x] Options Menu

### DIFF
--- a/packages/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
+++ b/packages/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
@@ -9,8 +9,9 @@ import { ApplicationLauncherGroup } from './ApplicationLauncherGroup';
 import { ApplicationLauncherSeparator } from './ApplicationLauncherSeparator';
 import { ApplicationLauncherItem } from './ApplicationLauncherItem';
 import { ApplicationLauncherContext } from './ApplicationLauncherContext';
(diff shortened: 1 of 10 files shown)


mcoker:
This is something we can probably follow up on later if needed, but the menus have `min-width: 100%`, which is relative to the parent component container. The parent container's width is defined by the width of the toggle. Meaning the menu should have a min-width of the toggle width. Here's what it looks like normally:

Screen Shot 2020-06-19 at 2 25 59 PM

That's broken when the menu is moved outside of the parent container that wraps the toggle

Screen Shot 2020-06-19 at 2 25 07 PM

mcoker:
It looks like the problem above may have been solved in the select component, except that we're setting a `width` on the menu that equals the toggle width (I'm assuming), and that updates as the toggle width changes. I think that should be a `min-width` instead, to match the component when `menuAppendTo` is not used. That same logic could apply to all of these menus since they all behave the same way.

jschuler:
@mcoker where are you seeing that? looking at https://patternfly-react-pr-4414.surge.sh/documentation/react/components/dropdown#append-menu-document-body the menu item there expands to match at least the toggle width

mcoker:
Ah sorry! I didn't realize the dropdown/select changed in this PR, so I was looking at master.

PR #4417, opened 2020-06-19
feat(Table): adds sticky header prop and demo

### LINKED ISSUE #4406: Table - sticky headers
https://github.com/patternfly/patternfly/pull/3093

To enable, just add `.pf-m-sticky-header` to `.pf-c-table`

jenny-s51:
**What**: Closes #4406. Includes version bump for styling.


mcoker:
Looks like there is a bug in the core implementation where the header cells' bottom border disappears when the sticky modifier is applied. Looking into that.

mcoker:
@tlabaj @jenny-s51 that bug was fixed in https://github.com/patternfly/patternfly/pull/3197 and available in https://github.com/patternfly/patternfly/releases/tag/prerelease-v4.16.3

jenny-s51:
@tlabaj done and done - @mcoker thank you for linking the fix and the prerelease version!

PR #4424, opened 2020-06-19
fix(OptionsMenu): fix accessibility bug in plain with text disabled part

### LINKED ISSUE #4423: OptionsMenu: Accessibility bug in Plain with text disabled
When we use keyboard to control the options menu, we can still focus on the disabled plain with text button and press enter into the dropdown menu. We can also choose items in the dropdown menu.
image
image

Bug.

DaoDaoNoCode:
**What**: I pass the `isDisabled` value to the dropdown button so that it cannot be focused when it is disabled.
Closes #4423


DaoDaoNoCode:
Sorry, this bug should have been fixed in #4352 , but I didn't check the accessibility of the disabled button part. It's solved in this PR.

Issue #1588, opened 2019-03-15
Wizard - keyboard and screen reader interaction


jgiardino:
1. The navigation links currently are not accessible for either keyboard-only users or screen reader users. They can be made accessible by including a url in an `href` attribute. (**Update**: given current recommendations, it might make the most sense to make the items buttons so that they are keyboard accessible- see discussion below)
2. In the mobile layout, the navigation is accessed as a menu. The dropdown behavior pattern that I think makes sense here currently isn't implemented anywhere yet. I included comments about similar behavior in the PRs for [context selector](https://github.com/patternfly/patternfly-react/pull/1526#issuecomment-471074929) and [select](https://github.com/patternfly/patternfly-react/pull/1487#issuecomment-473037694) components, but maybe it would make sense to have a quick chat with everyone to understand if there's a simpler way to handle this interaction so that it can be easily reused across the different components?
- **Update:** Ideally, the interaction pattern would be the following: https://docs.google.com/document/d/1m9nICFRgCXnXeHYN487BMX9MHw4DjvOln2B5p8yJa70/edit#heading=h.1lr7m0rc8a6s however making the menu accessible is the most important aspect.

matthewcarleton:
Following up on this @jgiardino @jessiehuff now that we're using the wizard in a product.
What are your thoughts on [this from MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events)?
```onclick events
Anchor tags are often abused as fake buttons by setting their href to "#" or "javascript:void(0)" 
to prevent the page from refreshing, then listening for their click events .

These bogus href values cause unexpected behavior when copying/dragging links, opening 
links in a new tab/window, bookmarking, or when JavaScript is loading, errors, or is disabled. 
They also convey incorrect semantics to assistive technologies, like screen readers.

Use a <button> instead. In general, you should only use a hyperlink for navigation to a real URL.
```

jgiardino:
Hey @matthewcarleton, thanks for sharing that link. I agree with the recommendations made by MDN on when to use anchor elements. I guess I assumed that anchor elements were chosen because the expectation was that real URLs would be provided for them. But maybe that wasn't the intention, and it might be worth discussing whether the navigation in a wizard should be handled as links or buttons. For example, do we want to support the user hitting the browser Back button in the context of a wizard? Or should the user be forced to click one of the buttons instead?

matthewcarleton:
[...]

PR #4425, opened 2020-06-19
feat(wiz): make wizard nav items more easily discoverable

seanforyou23:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/1588

Updates Wiz to use more descriptive labels for landmark regions and to allow for buttons as well as anchors for the main nav items. Also adds jest/cypress tests.

It would be good to start considering these accessible names and the overall strategy of how to handle label composition as part of our ongoing content strategy (it's microcopy). I created a new issue tag `microcopy review` to help identify where we need more people to weigh in on the approach in question. @abigaeljamie ping

### DIFF
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -6,7 +6,7 @@ import { Modal, ModalVariant } from '../Modal';
 import { WizardFooterInternal } from './WizardFooterInternal';
 import { WizardToggle } from './WizardToggle';
 import { WizardNav } from './WizardNav';
-import { WizardNavItem } from './WizardNavItem';
+import { WizardNavItem, WizardNavItemProps } from './WizardNavItem';
 import { WizardContextProvider } from './WizardContext';
 import { PickOptional } from '../../helpers/typeUtils';
 import { WizardHeader } from './WizardHeader';
@@ -24,6 +24,8 @@ export interface WizardStep {
   canJumpTo?: boolean;
   /** Sub steps */
   steps?: WizardStep[];
+  /** Props to pass to the WizardNavItem */
+  stepNavItemProps?: React.HTMLProps<HTMLButtonElement | HTMLAnchorElement> | WizardNavItemProps;
   /** (Unused if footer is controlled) Can change the Next button text. If nextButtonText is also set for the Wizard, this step specific one overrides it. */
   nextButtonText?: React.ReactNode;
   /** (Unused if footer is controlled) The condition needed to enable the Next button */
@@ -66,6 +68,12 @@ export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
   startAtStep?: number;
   /** Aria-label for the Nav */
   navAriaLabel?: string;
+  /** Sets aria-labelledby on nav element */
+  navAriaLabelledBy?: string;
+  /** Aria-label for the main element */
+  mainAriaLabel?: string;
+  /** Sets aria-labelledby on the main element */
+  mainAriaLabelledBy?: string;
   /** Can remove the default padding around the main body content by setting this to true */
   hasNoBodyPadding?: boolean;
   /** (Use to control the footer) Passing in a footer component lets you control the buttons yourself */
@@ -99,7 +107,7 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
   static displayName = 'Wizard';
   private static currentId = 0;
   static defaultProps: PickOptional<WizardProps> = {
-    title: '',
+    title: null,
     description: '',
     className: '',
     startAtStep: 1,
@@ -108,7 +116,10 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     cancelButtonText: 'Cancel',
     hideClose: false,
     closeButtonAriaLabel: 'Close',
-    navAriaLabel: 'Steps',
+    navAriaLabel: null,
+    navAriaLabelledBy: null,
+    mainAriaLabel: null,
+    mainAriaLabelledBy: null,
     hasNoBodyPadding: false,
     onBack: null as WizardStepFunctionType,
     onNext: null as WizardStepFunctionType,
@@ -312,6 +323,9 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
       hideClose,
       closeButtonAriaLabel = 'Close',
       navAriaLabel,
+      navAriaLabelledBy,
+      mainAriaLabel,
+      mainAriaLabelledBy,
       hasNoBodyPadding,
       footer,
       appendTo,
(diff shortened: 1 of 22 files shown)


packages/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -26,16 +28,22 @@ export const WizardNavItem: React.FunctionComponent<WizardNavItemProps> = ({
   isDisabled = false,
   step,
   onNavItemClick = () => undefined,
-  navItemComponent = 'a'
+  navItemComponent = Button
```
jschuler, line 31: In core they use anchor and this could be a breaking change
https://pf4.patternfly.org/documentation/core/components/wizard

seanforyou23: Yep, I wanted to use the react workspace to facilitate the conversation on both ends since we can better see the components in action with these interactive examples. If we're good with it we'd definitely need to make corresponding changes in core.

seanforyou23: Seems like it needs to be configurable between link and button, and maybe favor button as the default since we're typically building SPAs these days. In case the user isn't using a single page app they'll want anchors, and in the case of SPA button seems to carry the most appropriate semantics and behavior as selecting them doesn't currently change the page URL address. We could look into hooking in routing with the nav items, but that's starting to look like a good place to break out into a separate effort.

So yea definitely is missing the anchor support now (I can add this back in with a simple prop to toggle between the two). Does that sound ok? @jschuler

packages/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -26,16 +28,22 @@ export const WizardNavItem: React.FunctionComponent<WizardNavItemProps> = ({
...
 
   return (
     <li className={css(styles.wizardNavItem)}>
       <NavItemComponent
+        variant="link"
         aria-current={isCurrent && !children ? 'page' : false}
         onClick={() => onNavItemClick(step)}
-        className={css(styles.wizardNavLink, isCurrent && 'pf-m-current', isDisabled && 'pf-m-disabled')}
+        className={css(
+          btnStyles.modifiers.inline,
```
mcoker, line 42: In general, I think we try to avoid this sort of thing where `<NavItemComponent>` is also a `<Button>`, but we just use the button component styles. Another way could be for us just to add the `<button>` html element support for use with this BEM element.

packages/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -26,16 +28,22 @@ export const WizardNavItem: React.FunctionComponent<WizardNavItemProps> = ({
...
+          styles.wizardNavLink,
+          isCurrent && 'pf-m-current',
+          isDisabled && 'pf-m-disabled'
```
mcoker, line 45: if it's a `button`, seems like we would want to use the `disabled` attr?

packages/react-core/src/components/Wizard/WizardNavItem.tsx
mcoker: If we're supporting both `<a>` and `<button>` here, I don't think that's breaking. Shouldn't be breaking unless we remove support for `a.wizardNavLink`. Though I'm not sure we would use the inline button variation here - we could also just add `border: 0;` (I think that's it?) to the `wizardNavLink` styles to support use of the `<button>` html element.

jgiardino:
These updates are great to see! Thanks for working on this. 

I just had some comments on the use of aria-label for landmark regions. I pasted below one of the screencaps shared in the description. The thing that jumps out at me here is the use of "Content area for ... main":

1. Here are guidelines on use of `<main>`: https://www.w3.org/TR/wai-aria-practices-1.2/#aria_lh_main. A label is recommended when you have more than one, but I'm not sure that would ever be the case. More than one `<main>` should only exist if nested inside a `document` or `application`, which I think is highly unlikely for most apps, or at least most instances of a wizard component. 
2. Refer to the last bullet in this section: https://www.w3.org/TR/wai-aria-practices-1.2/#general-principles-of-landmark-design. They recommend against using "Navigation" in the aria-label of a navigation region, because it's redundant. I would argue that using "Content for" is also redundant with the semantic meaning of the `main` region which is meant to provide the contents of the page. 

My general suggestion is to be very conservative about adding labels to our examples. The Navigation component is an example where we definitely want to include labels in our examples, because we vetted the labels with content strategy and design, and the strings we provide are the same for all instances of that component. But in a case like this, there is a potential risk that placeholder strings we provide in examples are never updated by the consumer. And when this happens, it has the opposite effect of what you're hoping to achieve, because an incorrect label could be very confusing to the user. In other words, if in this case the options are: 1) having a descriptive label is nice but not necessary, 2) having no label is fine, vs 3) having an incorrect label is potentially damaging to the user experience, then the safest option is to not include a label in our examples. Or alternatively, find a way to provide a descriptive label where you eliminate the possibility of it being incorrect...

To do this, a possible alternative solution is to use `aria-labelledby` instead, which references the wizard header text (and maybe also the current step?). This is something that the consumer doesn't need to handle, because we would handle that for them. So in this case, the main region would be announced as "Basic First step main" (which would hopefully make more sense with real wizard contents).  This avoids the risk I mention above while allowing us to include a label. _But_ I also consider this as a nice-to-have given my first point above about `main` not needing a label, and so maybe it isn't worth the extra work to hook this up. 🤷 

For the navigation region, we definitely need a unique label, and I can see it being really helpful to include the wizard header text as part of that label. If this is something we want to include, then I would definitely suggest aria-labelledby instead of aria-label for the navigation region, for similar reasons that I mention above (i.e. that it's really risky to include placeholder text in examples that might get ignored by consumers after copying/pasting the examples). So in this case, aria-labelledby could reference the wizard heading, and be announced as "Basic Steps navigation".

Just to clarify though, I'm not opposed to supporting the ability to provide an aria-label in these components, I'm primarily warning against doing this in examples.

wizard-accessibility-current-1

seanforyou23:
Good stuff, thx for the feedback everyone! I'll make some updates and push the changes soon as they are ready. Keep you posted!

packages/react-core/src/components/Wizard/WizardNavItem.tsx
mcoker: @seanforyou23 opened https://github.com/patternfly/patternfly/pull/3216 if you want to have a look.

seanforyou23: Made some updates, will remove the temporary inline style soon as we get your core PR merged. Let me know if you see any issues with the styling!

seanforyou23:
Alright, have another round of updates ready! Added some examples so both cases are clearer. I re-swizzled the way the labels are applied to use aria-labelledby approach, it's working quite well but I later realized a newer requirement for the in-page wizard not to have a header so for in-page wiz there's nothing to point aria-labelledby to out of the box. For these examples, I added the bare example title as aria-label to the `nav` and `main` to emulate what it's like for wiz-in-modal. WDYT?

We should now be able to override all these new props and I've done some cleanup to ensure we're not adding aria attributes in cases where none are needed.

The reason the links were not showing up to screen reader users was because they didn't have href, so for these cases we now throw an error to help prevent users from making this mistake.

I noticed the button in WizardToggle renders to the DOM for the in-page wizard, it's just hidden in the background.. is it used for anything? If not I may add a small check to remove it from the DOM for housekeeping sake.

packages/react-core/src/components/Wizard/WizardNavItem.tsx
mcoker: Oh I forgot to mention - I noticed this weird behavior in chrome where the focus outline applies to the item's pseudo element (the number), too. We can likely fix that with `:focus-within` on the parent list item. WDYT, is that worth updating?

Screen Shot 2020-06-26 at 9 25 19 AM

seanforyou23: Hmm good question. Currently the number is announced when you focus the item which seems like it may have been done by design. Would changing to focus-within stop the step numbers from being automatically announced? I think the main thing we'd want to try and preserve is the current behavior in Safari in terms of how the items are announced to SRU

mcoker: I'm not aware of any impacts to screen readers when using `:focus-within`, I believe it just changes the presentation of the focus indicator. I made a couple of demos here if you want to see if there is any sort of effect. The links with "focus within" in the text change the focus indicator so that the number doesn't get its own focus ring.

https://codepen.io/mcoker/pen/pogrRwB - using `:focus-within`
https://codepen.io/mcoker/pen/yLeogaw - just changes the link/button's `outline-style`

jgiardino:
Thanks for making updates!

I'm finally realizing that `<main>` should only be used in the context of a modal, given the requirement that there should only be one `main` region on a page, unless nested inside an `application` or `document`.  Is it possible to enable that?

I don't remember the discussion or decision behind using "Steps" as the nav region label. It might be worth circling back to when that was introduced into core, as there might be some discussion there, or if not, follow up with content strategy on what's recommended here. For all the other nav regions, we had standardized on strings to use "Global", "Local", etc... so that users could expect a level of consistency in labeling across all the applications. I would think the same applies to the nav region in the wizard.

packages/react-core/src/components/Wizard/WizardNavItem.tsx
seanforyou23: Yea both of those methods seem to announce the step number just fine. `:focus-within` sounds good to me, good catch!

mcoker:
@seanforyou23 https://github.com/patternfly/patternfly/pull/3216 was merged. We still have a weird focus outline issue, but there is a follow up to fix that later (next sprint hopefully) https://github.com/patternfly/patternfly/issues/3269

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -66,6 +68,12 @@ export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
   startAtStep?: number;
   /** Aria-label for the Nav */
   navAriaLabel?: string;
+  /** Sets aria-labelledby on Nav element */
```
tlabaj, line 71: I am wondering if these need to be sentence case...

jschuler:
Screen Shot 2020-07-13 at 11 00 04 AM
the substep style looks different than before (centered vs left-aligned) @mcoker

packages/react-core/src/components/Wizard/WizardBody.tsx
```diff
@@ -7,14 +7,26 @@ export interface WizardBodyProps {
...
-  <main className={css(styles.wizardMain)}>
-    <div className={css(styles.wizardMainBody, hasNoBodyPadding && styles.modifiers.noPadding)}>{children}</div>
-  </main>
-);
+  hasNoBodyPadding = false,
+  'aria-label': ariaLabel = null,
+  'aria-labelledby': ariaLabelledBy,
+  mainComponent = 'div'
+}: WizardBodyProps) => {
+  const MainComponent = mainComponent as React.ElementType;
```
jschuler, line 25: should the type be ElementType to begin with so you don't have to cast it?

packages/react-core/src/components/Wizard/WizardBody.tsx
```diff
@@ -7,14 +7,26 @@ export interface WizardBodyProps {
...
 export const WizardBody: React.FunctionComponent<WizardBodyProps> = ({
   children,
-  hasNoBodyPadding = false
-}: WizardBodyProps) => (
-  <main className={css(styles.wizardMain)}>
-    <div className={css(styles.wizardMainBody, hasNoBodyPadding && styles.modifiers.noPadding)}>{children}</div>
-  </main>
-);
+  hasNoBodyPadding = false,
+  'aria-label': ariaLabel = null,
```
jschuler, line 21: don't need to init to null

packages/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -26,18 +28,35 @@ export const WizardNavItem: React.FunctionComponent<WizardNavItemProps> = ({
   isDisabled = false,
   step,
   onNavItemClick = () => undefined,
-  navItemComponent = 'a'
+  navItemComponent = 'button',
+  href = null,
+  ...rest
 }: WizardNavItemProps) => {
-  const NavItemComponent = navItemComponent as any;
+  const NavItemComponent = navItemComponent as React.ElementType;
```
jschuler, line 35: maybe change the type in the props above

packages/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -17,6 +17,8 @@ export interface WizardNavItemProps {
   onNavItemClick?: (step: number) => any;
   /** Component used to render WizardNavItem */
   navItemComponent?: React.ReactNode;
```
jschuler, line 19: do we support only button and a or anything goes? if the former we could limit it to those 2

packages/react-core/src/components/Wizard/WizardNavItem.tsx
```diff
@@ -26,18 +28,35 @@ export const WizardNavItem: React.FunctionComponent<WizardNavItemProps> = ({
...
   onNavItemClick = () => undefined,
-  navItemComponent = 'a'
+  navItemComponent = 'button',
+  href = null,
+  ...rest
 }: WizardNavItemProps) => {
-  const NavItemComponent = navItemComponent as any;
+  const NavItemComponent = navItemComponent as React.ElementType;
+
+  if (navItemComponent === 'a' && !href) {
```
jschuler, line 37: might want to add also this check `process.env.NODE_ENV !== 'production'`

packages/react-core/src/components/Wizard/WizardBody.tsx
seanforyou23: Made the change, but isn't won't it be a breaking change? I was asserting to `ElementType` to improve from `any` but to also try and keep from breaking anyone. Lemme know what you think, happy to do whatever we think is best.

seanforyou23: Oops, nvm, forgot this is a new prop :)

tlabaj:
@seanforyou23 looks like the core fix went in for the alignment.  Can we rebase this and hopefully merge it.

packages/react-core/src/components/Wizard/WizardNav.tsx
```diff
@@ -26,7 +29,11 @@ export const WizardNav: React.FunctionComponent<WizardNavProps> = ({
   }
 
   return (
-    <nav className={css(styles.wizardNav, isOpen && 'pf-m-expanded')} aria-label={ariaLabel}>
+    <nav
+      className={css(styles.wizardNav, isOpen && 'pf-m-expanded')}
```
jschuler, line 33: is the class available in the styles object?

jschuler:
Also, not sure when it broke but looks like the Wizard finished step is not up-to-date anymore?
Screen Shot 2020-07-28 at 4 52 18 PM
Not sure if you want to fix it in this PR or separately

packages/react-core/src/components/Wizard/WizardNav.tsx
seanforyou23: I wasn't previously, but is now! Updated

seanforyou23:
@jschuler good catches, fixed up the emtpy state bits in the finished step example, I guess they've been out of date for a while now. Regarding the expanded class, I updated it to use dynamic style property from react-styles, but then noticed it's never applied because isOpen is always false. Does this need to be cleaned up?

The last thing was that the docs build keep failing with different errors on CI vs locally. Local error attached below.. When I fix the path to something that works like `import { CogsIcon } from '@patternfly/react-icons';` and then run `yarn lint:ts --fix` it reverts back to the further qualified path under dist. Not sure what to do here, as one way fails the lint target and the other fails the docs build. Any advice would be appreciated!

Screen Shot 2020-07-29 at 10 47 09 AM

jschuler:
@seanforyou23 you should be able to set the path to `import CogsIcon from '@patternfly/react-icons/dist/js/CogsIcon';` did that fail previously?

seanforyou23:
Yea it doesn't seem to like that. The import works fine so I just ignored it for now if that's ok.

PR #4426, opened 2020-06-21
fix(alert): add display name to the close action button

### LINKED ISSUE #4415: AlertActionCloseButton doesn't have displayName
It seems that `AlertActionCloseButton` does not have `displayName` property set.
ping @redallen

Related to: 
https://github.com/patternfly/patternfly-react/pull/4388
https://github.com/patternfly/patternfly-react/issues/4369

boaz0:
**What**: Closes #4415 

//cc @redallen @PanSpagetka

### DIFF
--- a/packages/react-core/src/components/Alert/AlertActionCloseButton.tsx
+++ b/packages/react-core/src/components/Alert/AlertActionCloseButton.tsx
@@ -14,7 +14,7 @@ export interface AlertActionCloseButtonProps extends ButtonProps {
   variantLabel?: string;
 }
 
-export const AlertActionCloseButton = ({
+export const AlertActionCloseButton: React.FunctionComponent<AlertActionCloseButtonProps> = ({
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Alert/AlertActionCloseButton.tsx
```diff
@@ -35,3 +35,4 @@ export const AlertActionCloseButton = ({
     )}
   </AlertContext.Consumer>
 );
+AlertActionCloseButton.displayName = 'AlertActionCloseButton';
```
redallen, line 38: On line 17 can we type `AlertActionCloseButton` as a `React.FunctionComponent<AlertActionCloseButtonProps>` like all our other components as well? That's why the linter didn't pick it up...

boaz0: Done :+1:

PR #4427, opened 2020-06-22
fix(expandable-section): pass expandable value to onToggle

### LINKED ISSUE #4412: ExpandableSection#onToggle should include the expand / collapsed state
`ExpandableSection` has an `onToggle` function however it does not have a parameter indicating which state the component is in: expanded or collapsed.

This is a request to supply the current state as a parameter to `onToggle`.

This would be useful to know so that clients wishing to lazy load content can do so here.

enhancement

Work around is to wrap the ExpandableSection and always use controlled state such that a new callback function with the expand / collapse param can be provided.

OpenShift Console 4.6

boaz0:
**What**: Closes #4412 

//cc @christianvogt

### DIFF
--- a/packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
+++ b/packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
@@ -81,8 +81,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
     // uncontrolled
     if (isExpanded === undefined) {
       propOrStateIsExpanded = this.state.isExpanded;
-      onToggle = isExpanded => {
-        this.setState({ isExpanded }, () => onToggleProp(this.state.isExpanded));
+      onToggle = isOpen => {
+        this.setState({ isExpanded: isOpen }, () => onToggleProp(this.state.isExpanded));
       };
     }
 
--- a/packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
+++ b/packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
@@ -18,7 +18,7 @@ export interface ExpandableSectionProps {
   /** Text that appears in the toggle when collapsed (will override toggleText if both are specified; used for uncontrolled expandable with dynamic toggle text) */
   toggleTextCollapsed?: string;
   /** Callback function to toggle the expandable content */
-  onToggle?: () => void;
+  onToggle?: (isExpanded: boolean) => void;
   /** Forces active state */
   isActive?: boolean;
 }
@@ -42,7 +42,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
     toggleText: '',
     toggleTextExpanded: '',
     toggleTextCollapsed: '',
-    onToggle: (): any => undefined,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    onToggle: (isExpanded): void => undefined,
     isActive: false
   };
 
@@ -80,9 +81,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
     // uncontrolled
     if (isExpanded === undefined) {
       propOrStateIsExpanded = this.state.isExpanded;
-      onToggle = () => {
-        onToggleProp();
-        this.setState(prevState => ({ isExpanded: !prevState.isExpanded }));
+      onToggle = isExpanded => {
+        this.setState({ isExpanded }, () => onToggleProp(this.state.isExpanded));
       };
     }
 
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
```diff
@@ -42,7 +42,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
     toggleText: '',
     toggleTextExpanded: '',
     toggleTextCollapsed: '',
-    onToggle: (): any => undefined,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    onToggle: (_isOpen): void => undefined,
```
dlabrecq, line 46: Curious why an underscore is used? I kind of expect the compiler would have a conflict here.

Can we rename this as `isOpen`?

boaz0: Yes, we can do that, it's just what WebStorm suggested me to call it.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
```diff
@@ -18,7 +18,7 @@ export interface ExpandableSectionProps {
   /** Text that appears in the toggle when collapsed (will override toggleText if both are specified; used for uncontrolled expandable with dynamic toggle text) */
   toggleTextCollapsed?: string;
   /** Callback function to toggle the expandable content */
-  onToggle?: () => void;
+  onToggle?: (isOpen: boolean) => void;
```
tlabaj, line 21: Can we just call this parameter `isExpanded`
Can we make the parameter optional, otherwise this would be a breaking change.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
```diff
@@ -80,9 +81,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
     // uncontrolled
     if (isExpanded === undefined) {
       propOrStateIsExpanded = this.state.isExpanded;
-      onToggle = () => {
-        onToggleProp();
-        this.setState(prevState => ({ isExpanded: !prevState.isExpanded }));
+      onToggle = isOpen => {
+        this.setState({ isExpanded: isOpen }, () => onToggleProp(this.state.isExpanded));
```
tlabaj, line 85: We should only do this is `isOpen/isExpanded` is provided, otherwise we should what we were doing before.

dlabrecq:
Linter failed

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
```diff
@@ -18,7 +18,7 @@ export interface ExpandableSectionProps {
   /** Text that appears in the toggle when collapsed (will override toggleText if both are specified; used for uncontrolled expandable with dynamic toggle text) */
   toggleTextCollapsed?: string;
   /** Callback function to toggle the expandable content */
-  onToggle?: () => void;
+  onToggle?: (isExpanded: boolean) => void;
```
tlabaj, line 21: In order to not break consumers this argument needs to be optional.
```suggestion
  onToggle?: (isExpanded?: boolean) => void;
```

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
```diff
@@ -80,9 +81,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
     // uncontrolled
     if (isExpanded === undefined) {
       propOrStateIsExpanded = this.state.isExpanded;
-      onToggle = () => {
-        onToggleProp();
-        this.setState(prevState => ({ isExpanded: !prevState.isExpanded }));
+      onToggle = isExpanded => {
```
tlabaj, line 84: We need to account for the case when isExpanded is not provided.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
christianvogt: This is correct:
```
 onToggle?: (isExpanded: boolean) => void;
```

You want to specify that `isExpanded` is always provided as a boolean value. By using `isExpanded?` the type becomes `boolean | undefined`. But a boolean value is always provided.

Clients are free to ignore this param as per the previous implementation.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
christianvogt: `isExpanded` here is _always_ provided because on line `110` is where this function is called and it supplies a boolean param.

tlabaj: @boaz @christianvogt I know this is a day off for some I can fix linting error so we can this into the release.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
tlabaj: Yup I see that.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
redallen: Underscores usually represent unused variables.

boaz0:
Thank you @tlabaj - I am sorry I was unavailable yesterday but I see you did a good job!

PR #4460, opened 2020-06-29
feat(Select): add option description property

### LINKED ISSUE #4430: Select - add support for item description
https://github.com/patternfly/patternfly/pull/3130

If a menu item has a description:

* The `.pf-c-select__menu-item` element adds `.pf-m-description`
* The children of `.pf-c-select__menu-item` are placed in a new element `div.pf-c-select__menu-item-main`
* `div.pf-c-select__menu-item-description` with the user's description is placed after `.pf-c-select__menu-item-main` inside of `.pf-c-select__menu-item`

kmcfaul:
**What**: Closes #4430

Applies to all variants except for `checkbox` currently due to differing structures (checkbox options are inside a label instead of a button). @mcoker Is there a design for descriptions with checkbox select?

### DIFF
--- a/packages/react-core/src/components/Select/SelectOption.tsx
+++ b/packages/react-core/src/components/Select/SelectOption.tsx
@@ -176,6 +176,7 @@ export class SelectOption extends React.Component<SelectOptionProps> {
                   styles.selectMenuItem,
                   isDisabled && styles.modifiers.disabled,
                   isFocused && styles.modifiers.focus,
+                  description && styles.modifiers.description,
                   className
                 )}
                 onKeyDown={this.onKeyDown}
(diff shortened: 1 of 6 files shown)


mcoker:
@mcarrano wdyt? We don't have an example of this in core, but I would assume the checkbox select description would just use the check component's description feature.

Screen Shot 2020-06-29 at 11 06 54 AM

regular select description for reference

Screen Shot 2020-06-29 at 11 05 41 AM

mcarrano:
Yes, that makes sense to me @mcoker .

kmcfaul:
Updated for checkbox option descriptions and removed a few demos that looked to be duplicates.

mcoker:
@mcarrano we use 12px font in the select/description/etc descriptions, and 14px in the check/radio descriptions. Should we normalize those? Or do we want to change the check description to 12px when it's used in a select?

mcarrano:
@mcoker I would rather not change the check/radio descriptions without further analysis. But if you can override that when they are used within a select, that would be a good solution, i.e. always display description as 12px when used in a Select or Dropdown.

PR #4464, opened 2020-07-01
fix(Select): handle missing select group label

### LINKED ISSUE #4447: Select group label not required, but without label it breaks the UI

When using select with SelectGroup the label is marked as optional, yet if I don't pass it it breaks the whole UI with error message `Uncaught (in promise) TypeError: Cannot read property 'replace' of undefined`

Remove label in PF examples and it breaks the component.

Rather critical bug, we can't properly upgrade to new PF because of this bug.

cloud.redhat.com

kmcfaul:
**What**: Closes #4447

If `label` is left out or set as `label=""` in `SelectGroup`, no title will be displayed in the UI. We may want to consider making the prop required later.


dlabrecq:
Tests failed

PR #4466, opened 2020-07-01
fix(NotificationDrawer): fix accessibility issues and add accessibility test cases

### LINKED ISSUE #4449: Notification drawer examples are currently inaccessible
- You are able to select items within each notification group by clicking on the items, but there is no way to do this by keyboard or screen reader. Focus shifts to each item, but pressing enter does not select the item in the same way that clicking does. I would expect in VO to be able to select this item by pressing "Control + Option + Space" but this does not select the item either.

Screen Shot 2020-06-25 at 11 06 59 AM

- Every action kebab is inaccessible by keyboard. You can shift focus to the kebab, but pressing enter doesn't open the dropdown.
Screen Shot 2020-06-25 at 11 02 03 AM

Bug

DaoDaoNoCode:
**What**: Closes #4449 
In this PR, I make these following changes:

1. Add the `onKeyDown` function for both the group title and the items in the group. When user press Enter or Space on the group title, it will toggle the whole group. When user press Enter or Space on the group items, it will call the click function to click the item.
2. Change the usage in the doc of Notification Drawer. The previous usage of kebab toggle button on the header is not correct, it will only open the dropdown menu when user click on the toggle button. The way to change it is to create a `isOpen` state and `onToggle` function for every toggle button, but not use `isActive` state.
3. Add accessibility test cases for both Notification Drawer Groups and Notification Drawer Basic. 

    - For Notification Drawer Basic: Test the accessibility of toggle buttons on notification drawer header and group list item header. Press Enter on them to see whether the dropdown menu is expanded and whether the first item in the dropdown menu is focused. Then Press Tab to see whether the dropdown menu is closed.
    - For Notification Drawer Group: Besides the test cases above, I also test the accessibility of the group title button. Press Enter on the group title to see whether the whole group is expanded.
4. According to above changes in the doc, change the code in demo app.

**Additional issues**: I am wondering whether there is a better way to fix the accessibility issues of the toggle button. If there are 10 toggle buttons in a notification drawer, users need to create 10 `isOpen` state and 10 `onToggle` functions, which is not user-friendly when they use them.


packages/react-core/src/components/NotificationDrawer/NotificationDrawerListItem.tsx
```diff
@@ -29,20 +29,29 @@ export const NotificationDrawerListItem: React.FunctionComponent<NotificationDra
...
-      className
-    )}
-    tabIndex={tabIndex}
-    onClick={e => onClick(e)}
-  >
-    {children}
-  </li>
-);
+}: NotificationDrawerListItemProps) => {
+  const onKeyDown = (event: any) => {
```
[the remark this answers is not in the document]
DaoDaoNoCode: Fixed. Thanks!

packages/react-core/src/components/NotificationDrawer/examples/NotificationDrawer.md
```diff
@@ -11,45 +22,74 @@ import { Button, Dropdown, DropdownItem, DropdownPosition, DropdownDirection, Dr
...
+  Title
+} from '@patternfly/react-core';
 
 class BasicNotificationDrawer extends React.Component {
-constructor(props) {
+  constructor(props) {
     super(props);
     this.state = {
-      isOpen: false,
-      isActive: ''
```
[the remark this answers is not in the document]
DaoDaoNoCode: Thanks for your suggestions. I think use id is a good way to simplify the whole things so I change the usage in Notification Drawer Groups example (both in the doc and in the demo app) because it has 9 toggle buttons. Here is the way I changed it.

1. Instead of using 9 `onToggle` functions and 9 `isOpen` state, I use only one `isToggle` function and `isOpen` state.
2. `isOpen` state is an object, the key of this object is the id of the ToggleButton and the value of this object is the state of the dropdown menu linked to this toggle button.
3. `onToggle` function receives 2 parameters, the first is the key and the second is the value. When it is toggled, I will set the key and value into `isOpen` state.
4. Only when `isOpen` is not null and `isOpen[id]` is true, this dropdown menu will be expanded.

After changing the usage, the example reduces about 40 lines of code, which looks much better than before. I only change the usage in Notification Drawer Groups example because it uses the most toggle buttons than other examples. I hope that our users could choose what usage they would use depending on the number of toggle buttons.

packages/react-integration/demo-app-ts/src/components/demos/NotificationDrawerDemo/NotificationDrawerBasicDemo.tsx
```diff
@@ -29,41 +32,50 @@ export class BasicNotificationDrawerDemo extends React.Component<
...
+    this.setState({
+      isOpen3
+    });
+  };
+  onToggle4 = (isOpen4: boolean) => {
+    this.setState({
+      isOpen4
+    });
+  };
+  onSelect = (event: any) => {
```
seanforyou23, line 67: Can we type this? Is it `React.SyntheticEvent<HTMLDivElement>`?

packages/react-integration/demo-app-ts/src/components/demos/NotificationDrawerDemo/NotificationDrawerGroupsDemo.tsx
```diff
@@ -42,68 +41,41 @@ export class GroupsNotificationDrawerDemo extends React.Component<
...
   };
-
-  onFocus = (id: string) => {
-    if (id) {
-      const element = document.getElementById(id);
-      element.focus();
-    }
-  };
-
   toggleFirstDrawer = (event: any, value: boolean) => {
```
seanforyou23, line 61: Can these handlers be typed as `React.SyntheticEvent<HTMLDivElement>`?

packages/react-integration/cypress/integration/notificationdrawergroups.spec.ts
```diff
@@ -80,4 +80,49 @@ describe('Drawer Groups Demo Test', () => {
       .last()
       .contains('30 minutes ago');
   });
+
+  // Accessibility test
+  it('Verify keyboard events happen correctly', () => {
+    // Verify the accessibility of toggle button on group list header
```
seanforyou23, line 86: You might consider revising these comments to be a little more descriptive since "accessibility" can mean many things. Maybe something like "verify the list header toggle button keyboard interactivity opens/closes actions menu" just as an example.

packages/react-integration/cypress/integration/notificationdrawergroups.spec.ts
```diff
@@ -80,4 +80,49 @@ describe('Drawer Groups Demo Test', () => {
...
+      cy.get('#notification-0')
+        .find('.pf-c-dropdown__menu-item')
+        .first()
+        .should('be.focused');
+      cy.wrap(toggleButton).trigger('keydown', { keyCode: 9 });
+      cy.get('#notification-0')
+        .find('.pf-c-dropdown__menu.pf-m-align-right')
+        .should('not.exist');
+    });
+    // Verify the accessibility of group header
```
seanforyou23, line 101: Similar comment here, can we describe the specific behaviors/interactions this test is verifying?

packages/react-integration/cypress/integration/notificationdrawergroups.spec.ts
```diff
@@ -80,4 +80,49 @@ describe('Drawer Groups Demo Test', () => {
...
+    cy.get('.pf-c-notification-drawer__group')
+      .first()
+      .should('have.class', 'pf-m-expanded');
+    cy.get('.pf-c-notification-drawer__group-toggle')
+      .first()
+      .trigger('keydown', { keyCode: 13 });
+    cy.get('.pf-c-notification-drawer__group')
+      .first()
+      .should('not.have.class', 'pf-m-expanded');
+    // Verify the accessibility of toggle button on list item header
```
seanforyou23, line 111: Same question here

packages/react-integration/demo-app-ts/src/components/demos/NotificationDrawerDemo/NotificationDrawerBasicDemo.tsx
DaoDaoNoCode: Fixed. I am not sure whether it is HTMLDivElement so I use `React.SyntheticEvent<HTMLElement>` instead.

packages/react-integration/demo-app-ts/src/components/demos/NotificationDrawerDemo/NotificationDrawerGroupsDemo.tsx
DaoDaoNoCode: Fixed as above.

packages/react-integration/demo-app-ts/src/components/demos/NotificationDrawerDemo/NotificationDrawerBasicDemo.tsx
seanforyou23: Cool! Yea whatever makes most sense. Ty!

packages/react-integration/cypress/integration/notificationdrawergroups.spec.ts
DaoDaoNoCode: Fixed. Thanks!

packages/react-integration/cypress/integration/notificationdrawergroups.spec.ts
DaoDaoNoCode: Fixed!

packages/react-integration/cypress/integration/notificationdrawergroups.spec.ts
DaoDaoNoCode: Fixed!

DaoDaoNoCode:
> This looks great so far, Juntao! 🙂 I'm curious why we couldn't use `isActive` here though? I'm also noticing that through mouse click, I can select list items, but not through keyboard. I also agree with your comment on the end of your PR description, it does seem like a lot of event handlers. I believe that this is the pattern I've seen with most of our examples, but I would not mind us exploring a solution for that as well.

@jessiehuff In previous usage, `isActive` is only set when user click on the toggle button. If user begins with keyboard, `isActive` will never be set, and this is why we cannot use keyboard to press enter on the toggle button as the issue #4449 mentioned.
As for the event handlers, thanks for the suggestions from @seanforyou23 , I change the usage in Notification Drawer Groups examples and uses only 1 event handler. I think our users could choose to use either method. If they have only a few toggle buttons, they can create event handlers for every toggle button. If they have lots of toggle buttons, they could refer to the new usage in Notification Drawer Groups example.

DaoDaoNoCode:
@seanforyou23 @jessiehuff Sorry I fixed the conflicts and merge it. Could you please review again?

PR #4468, opened 2020-07-02
add description to DropdownItem

### LINKED ISSUE #4429: Dropdown - add support for item description
https://github.com/patternfly/patternfly/pull/3130

If a menu item has a description:
* The `.pf-c-dropdown__menu-item` element adds `.pf-m-description`
* The children of `.pf-c-dropdown__menu-item` is placed in a new element `div.pf-c-dropdown__menu-item-main`
* `div.pf-c-dropdown__menu-item-description` is placed after `.pf-c-dropdown__menu-item`

nicolethoen:
Closes #4429 

Screen Shot 2020-07-02 at 12 43 12 PM


tlabaj:
@mcarrano Should we change the type of the description then?  Right now it is a React Node so the consumer can has the flexibility of adding an icon if they like.  Do we want to be more strict and only allow  for a string?

mcarrano:
Thanks for the clarification @tlabaj . No I would not restrict this. But maybe update the example to not show an icon here as it's not something I'd want to encourage.

nicolethoen:
> Thanks for the clarification @tlabaj . No I would not restrict this. But maybe update the example to not show an icon here as it's not something I'd want to encourage.

@mcarrano It's been updated :)

PR #4472, opened 2020-07-02
feat(Select): fix typeahead button toggle propagation

### LINKED ISSUE #4471: Select typeahead toggle button does not work with keyboard
Toggle button doesn't open or close the select menu via keyboard keys.

kmcfaul:
**What**: Closes #4471

Button was actually sending the event twice.

Since the input already opens/closes the menu via keyboard navigation, would it be better to remove the button from the tabIndex instead @jessiehuff ? Using the keyboard to open the menu via this button doesn't focus the input atm.

### DIFF
--- a/packages/react-core/src/components/Select/SelectToggle.tsx
+++ b/packages/react-core/src/components/Select/SelectToggle.tsx
@@ -254,7 +254,9 @@ export class SelectToggle extends React.Component<SelectToggleProps> {
                   onClose();
                 }
               }}
-              tabIndex={-1}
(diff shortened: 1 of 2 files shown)


kmcfaul:
@jessiehuff Updated with the change to tabIndex for typeahead variants.

I'm unsure, would this would count as breaking?

kmcfaul:
@jenny-s51 Seeing it too now. For some reason they've got a typeahead modifier applied. Investigating.

PR #4475, opened 2020-07-06
fix(windows): fix build on windows

### LINKED ISSUE #3997: patternfly-react has dev build errors on Windows
Repo cannot build successfully out of the box when built in VSCode with Windows OS. 

@redallen 
What I've noticed so far:
- Some single quotes have to be changed into escaped double quotes in the package.json of react-catalog-view-extension and react-table.
- `generateClasses.js` in react-styles outputs to dist incorrectly on Windows due to path slash differences (failing to remove the extra src/css)
- The project level `./node_modules/.bin/gatsby.js` reference in the package.json of react-docs had to be changed to point directly at the top node_modules. Unsure if this is an error or just my environment not picking up the sym link.

redallen:
**What**: Closes #3997

**Additional issues**:

### DIFF
--- a/packages/react-styles/scripts/generateClassMaps.js
+++ b/packages/react-styles/scripts/generateClassMaps.js
@@ -64,7 +64,12 @@ function generateClassMaps() {
   const srcCSSFiles = glob.sync('src/css/**/*.css');
 
   const res = {};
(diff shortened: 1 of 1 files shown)


kmcfaul:
Thanks for looking into this!

PR #4484, opened 2020-07-07
chore(dependabot): add dependabot.yml

redallen:
**What**: Our bot to bump core in React broke some time ago. This could possibly work instead.

**Additional issues**: Dependabot claims to [support monorepos](https://dependabot.com/javascript/) and the [config is valid.](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#configuration-options-for-dependabotyml)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for npm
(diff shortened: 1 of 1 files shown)


evwilkin:
@redallen looks like this failed CI due to an unrelated a11y issue, otherwise LGTM

PR #4488, opened 2020-07-07
feat(DataList): Add support for action containers in DataList

### LINKED ISSUE #3860: Use of action containers in data list

In core we use `.pf-c-data-list__action` to wrap plain button actions (ie, kebab menu) to better align the action in the row. Doesn't look like this was implemented in the react component.

Look at the rendered code where a kebab is used as an action in a data-list row in core vs react. Core will have a `.pf-c-data-list__action` container and be aligned differently compared to react.

* https://www.patternfly.org/v4/documentation/react/components/datalist
* https://www.patternfly.org/v4/documentation/core/components/datalist

----

It doesn't seem ideal to ask consumers to wrap a plain button dropdown/action with this container, since it would likely lead to inconsistencies. Is there a way we can enable this in react? We can also revisit the CSS and see if there is a way to handle this spacing.

tlabaj:
**What**: Closes #3860

Note: I opened an issue to clean up the unused functions and state in DataListAction component.

### DIFF
--- a/packages/react-core/src/components/DataList/DataListAction.tsx
+++ b/packages/react-core/src/components/DataList/DataListAction.tsx
@@ -23,7 +23,7 @@ export interface DataListActionProps extends Omit<React.HTMLProps<HTMLDivElement
     xl?: 'hidden' | 'visible';
     '2xl'?: 'hidden' | 'visible';
   };
-  /* Flag to indicate that the action is a plain button (e.g. kebab dropdown toggle) so that styling is applied to align the button */
+  /** Flag to indicate that the action is a plain button (e.g. kebab dropdown toggle) so that styling is applied to align the button */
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/DataList/DataListAction.tsx
```diff
@@ -23,6 +23,8 @@ export interface DataListActionProps extends Omit<React.HTMLProps<HTMLDivElement
     xl?: 'hidden' | 'visible';
     '2xl'?: 'hidden' | 'visible';
   };
+  /* Flag to indicate that the action is a plain button (e.g. kebab dropdown toggle) so that styling is applied to align the button */
```
nicolethoen, line 26: I think you need another * for the docs to parse this comment correctly.

Issue #3930, opened 2020-03-19
Popover is not centered when maxWidth set


rvsia:
**Describe the issue. What is the expected and unexpected behavior?**

When `maxWidth` set in the popover, the popover is not centered and it starts from the very left of the window.

![image](https://user-images.githubusercontent.com/32869456/77053234-55b00580-69ce-11ea-81bb-8a9edf533d3e.png)

For example, 30% maxWidth

![image](https://user-images.githubusercontent.com/32869456/77053436-a1fb4580-69ce-11ea-9543-62d4f7d08be2.png)

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

1. Go to demo https://www.patternfly.org/v4/documentation/react/components/popover
2. Use this code

```jsx
import React from 'react';
import { Popover, PopoverPosition, Button } from '@patternfly/react-core';

SimplePopover = () => (
<React.Fragment>
{'asdddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd sds '}
  <Popover
    headerContent={<div>Popover Header</div>}
    maxWidth="80%"
    bodyContent={
[...]

Issue #3722, opened 2020-02-11
popover display artifacts when closing


christianvogt:
This bug is easily observed in the patternfly react documentation website.

Steps to reproduce:
- Open Popover documentation: https://www.patternfly.org/v4/documentation/react/components/popover
- scroll down to the first example popover. Make sure to scroll down enough so that the popover is forced to open to the right.
- toggle the popover open and closed by clicking the `Toggle Popover` button.

Observe: 
The contents of the popover are unmounted immediately on close which triggers tippy to reposition the popover causing the popover to move positions to the top.
![popover-flicker](https://user-images.githubusercontent.com/14068621/74247660-64f7b100-4cb4-11ea-8502-987c875a05bf.gif)

![image](https://user-images.githubusercontent.com/14068621/74247083-87d59580-4cb3-11ea-9e32-d2307857ac5d.png)

Expected:
The contents of the popover should not be unmounted until the animation is complete.

rachael-phillips:
Hi @christianvogt we've seen this issue and will be back soon with a response.

christianvogt:
bump. This is still an issue.

Issue #4467, opened 2020-07-02
aria-labelledby and tabIndex: Type 'null' is not assignable


rvsia:
**Describe the issue. What is the expected and unexpected behavior?**

When compiling our project using PF, TypeScript fails:

```jsx
node_modules/@patternfly/react-core/dist/js/components/Dropdown/DropdownItem.d.ts:3:18 - error TS2430: Interface 'DropdownItemProps' incorrectly extends interface 'InternalDropdownItemProps'.
  Types of property 'tabIndex' are incompatible.
    Type 'number | null | undefined' is not assignable to type 'number | undefined'.
      Type 'null' is not assignable to type 'number | undefined'.

3 export interface DropdownItemProps extends InternalDropdownItemProps {
                   ~~~~~~~~~~~~~~~~~

node_modules/@patternfly/react-core/dist/js/components/Modal/Modal.d.ts:3:18 - error TS2430: Interface 'ModalProps' incorrectly extends interface 'HTMLProps<HTMLDivElement>'.
  Types of property ''aria-labelledby'' are incompatible.
    Type 'string | null | undefined' is not assignable to type 'string | undefined'.
      Type 'null' is not assignable to type 'string | undefined'.

3 export interface ModalProps extends React.HTMLProps<HTMLDivElement> {
```

I believe these two lines should not include the null option:

https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Dropdown/DropdownItem.tsx#L34

https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Modal/Modal.tsx#L21

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug

**What is your product and what release version are you targeting?**

"@patternfly/react-core": "^4.18.5",

Data Driven Forms

Hyperkid123:
cc @evwilkin

redallen:
@seanforyou23 and I had a previous issue about this that resulted in `null` should be allowed to not render the `tabIndex` prop in the DOM at all, whereas `undefined` might result in a default `tabIndex` being set. However, this does not match the `React.HTMLProps<HTMLDivElement>` which is just `number | undefined`.

[...]

PR #4491, opened 2020-07-07
Rewrite tooltip/popover to use popper

jschuler:
Closes https://github.com/patternfly/patternfly-react/issues/4359
Closes https://github.com/patternfly/patternfly-react/issues/3930
Closes https://github.com/patternfly/patternfly-react/issues/3722


packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -89,251 +88,231 @@ export interface PopoverProps {
...
-   * clicked or the ESC key is used
+   * clicked, Enter key was used on it, or the ESC key is used
    */
-  shouldClose?: (tip: TippyInstance) => void;
+  shouldClose?: (hideFnc?: any) => void;
+  /**
+   * Callback function that is only invoked when isVisible is also controlled. Called when the Enter key is
+   * used on the focused trigger
+   */
+  shouldOpen?: (showFnc?: any) => void;
```
redallen, line 98: Can we call these `onShouldClose` and `onShouldOpen`?

packages/react-core/src/helpers/Popper/Popper.tsx
```diff
@@ -40,16 +85,84 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
...
+    return () => {
+      onMouseEnter && triggerElement && triggerElement.removeEventListener('mouseenter', onMouseEnter);
+      onMouseLeave && triggerElement && triggerElement.removeEventListener('mouseleave', onMouseLeave);
+      onFocus && triggerElement && triggerElement.removeEventListener('focus', onFocus);
+      onBlur && triggerElement && triggerElement.removeEventListener('blur', onBlur);
+      onTriggerClick && triggerElement && triggerElement.removeEventListener('click', onTriggerClick, false);
+      onTriggerEnter && triggerElement && triggerElement.removeEventListener('keydown', onTriggerEnter, false);
+      onPopperClick && popperElement && popperElement.removeEventListener('click', onPopperClick, false);
+      onDocumentClick && triggerElement && document.removeEventListener('click', onDocumentClickCallback, false);
+      onDocumentKeyDown && triggerElement && document.removeEventListener('keydown', onDocumentKeyDown, false);
```
redallen, line 137: This code isn't very DRY. How about a helper function?
```js
function addEventListener(listener, element, event) {
  if (listener && element) {
    element.addEventListener(event, listener);
  }
}
```

You don't have to worry about the `useCapture = false` since [it's the default.](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)

packages/react-core/src/components/Tooltip/examples/Tooltip.md
```diff
@@ -61,63 +48,159 @@ OnIconTooltip = () => (
...
+  return (
+    <>
       <div>
-        <div>
-          <span style={{ paddingRight: '10px' }}>Tooltip Position</span>
-          <select
-            aria-label="Tooltip position"
-            onChange={event => {
-              this.setState({ position: event.target.value });
+        <div style={{ border: '1px solid'}}>
```
redallen, line 75: Can you remove this border?

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -50,150 +49,183 @@ export interface TooltipProps {
...
+    }
+  };
 
-  extendChildren() {
-    return React.cloneElement(this.props.children, {
-      isAppLauncher: this.props.isAppLauncher
-    });
-  }
+  const addAriaToTrigger = () => {
+    if (aria === 'describedby' && !children.props['aria-describedby']) {
```
redallen, line 201: Also needs `&& children.props` since `children` can be a `string`.

On that note, are strings fully supported or should the type be changed to `children: React.ReactElement`?

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -50,150 +49,183 @@ export interface TooltipProps {
...
+  const content = (
+    <div
+      className={css(styles.tooltip, className)}
+      role="tooltip"
+      id={id}
+      style={{
+        maxWidth: hasCustomMaxWidth ? maxWidth : null,
+        opacity,
+        transition: `opacity ${animationDuration}ms cubic-bezier(.54, 1.5, .38, 1.11)`
+      }}
```
redallen, line 170: Can we create a template string or a `getStyles()` function for this since it's shared with `Popover.tsx`?

@mcoker Letting you know this exists.

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -89,251 +88,231 @@ export interface PopoverProps {
...
+      aria-label={headerContent ? undefined : ariaLabel}
+      aria-labelledby={headerContent ? `popover-${uniqueId}-header` : undefined}
+      aria-describedby={`popover-${uniqueId}-body`}
+      onMouseDown={onContentMouseDown}
+      style={{
+        minWidth: hasCustomMinWidth ? minWidth : null,
+        maxWidth: hasCustomMaxWidth ? maxWidth : null,
+        opacity,
+        transition: `opacity ${animationDuration}ms cubic-bezier(.54, 1.5, .38, 1.11)`
+      }}
```
redallen, line 284: @mcoker Letting you know this exists.

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -89,251 +88,231 @@ export interface PopoverProps {
...
+  position = 'top',
+  enableFlip = true,
+  className = '',
+  isVisible = null as boolean,
+  shouldClose = (): void => null,
+  shouldOpen = (): void => null,
+  'aria-label': ariaLabel = '',
+  bodyContent,
+  headerContent = null as typeof PopoverHeader,
+  footerContent = null as typeof PopoverFooter,
```
redallen, line 123: Don't need `as typeof PopoverHeader` or `as typeof PopoverFooter`.

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -89,251 +88,231 @@ export interface PopoverProps {
...
-      this.tip.hide();
-    } else {
-      // notify consumer
-      this.props.shouldClose(this.tip);
+export const Popover: React.FunctionComponent<PopoverProps> = ({
+  children,
+  position = 'top',
+  enableFlip = true,
+  className = '',
+  isVisible = null as boolean,
```
redallen, line 117: Neat trick to check if the user set a default. Might help us solve #4467.

packages/react-core/src/components/Popover/Popover.tsx
jschuler: the previous API already had `shouldClose` and as i wanted to break as few things as possible i kept it that way

packages/react-core/src/helpers/Popper/Popper.tsx
```diff
@@ -25,13 +28,55 @@ export interface PopperProps {
...
+  onPopperClick?: (event?: MouseEvent) => void;
+  /** Callback function when document is clicked */
+  onDocumentClick?: (event?: MouseEvent, triggerElement?: HTMLElement) => void;
+  /** Callback function when keydown event occurs on document */
+  onDocumentKeyDown?: (event?: KeyboardEvent) => void;
+  /** Enable to flip the popper when it reaches the boundary */
+  enableFlip?: boolean;
+  /** The behavior of how the popper flips when it reaches the boundary */
+  flipBehavior?: 'flip' | ('top' | 'bottom' | 'left' | 'right')[];
+  // boundary?: 'scrollParent' | 'window' | 'viewport' | HTMLElement;
```
nicolethoen, line 79: did you mean to leave this here? or uncomment it?

packages/react-core/src/components/Tooltip/examples/Tooltip.md
jschuler: I could but I like it :P, any particular reason?

packages/react-core/src/components/Tooltip/Tooltip.tsx
jschuler: should be ReactElement, good call!

jschuler: oh, looks like that is already the case! `children: ReactElement<any>;`

packages/react-core/src/components/Tooltip/examples/Tooltip.md
redallen: It doesn't really match the rest of our examples. Our layouts used a dashed border for example, and our forms don't have a border.

jschuler: In this case it makes it more obvious that the text is an interactive element which is why I added it

packages/react-core/src/helpers/Popper/Popper.tsx
jschuler: i'll remove it, thanks

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -19,13 +16,15 @@ export enum TooltipPosition {
...
 }
 
-export interface TooltipProps {
+export interface TooltipProps extends Omit<React.HTMLProps<HTMLDivElement>, 'content'> {
   /** The element to append the tooltip to, defaults to body */
-  appendTo?: Element | ((ref: Element) => Element);
-  /** Aria-labelledby or aria-describedby for tooltip */
-  aria?: 'describedby' | 'labelledby';
-  /** If enableFlip is true, the tooltip responds to this boundary */
-  boundary?: 'scrollParent' | 'window' | 'viewport' | HTMLElement;
```
tlabaj, line 28: Isn't removing this a breaking change?  Can we just leave it for now and deprecate it?   WE can open up a followup breaking change issue for our next major release.

jschuler: it technically is, but it shouldn't be needed any longer since the popper element can be appended to any container, I tried to re-create the functionality but with the new popper version it didn't work really well. Even the recent tippy versions have removed it. 
I figured, this close to the last breaking change release it might be okay to remove this mostly useless option! 
Of course, I could add it back but not let it do anything if that's what you mean?

tlabaj: I think that would be best incase anyone is using it.

redallen: If we do we should warn that the prop is ignored.

jschuler: @tlabaj @redallen I added it back as a deprecated prop for both Tooltip and Popover and have also added a warning

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -50,150 +49,192 @@ export interface TooltipProps {
...
 
-  componentDidMount() {
-    document.addEventListener('keydown', this.handleEscKeyClick, false);
-  }
-
-  componentWillUnmount() {
-    document.removeEventListener('keydown', this.handleEscKeyClick, false);
-  }
+  const onDocumentClick = (event: MouseEvent, triggerElement: HTMLElement) => {
+    // event.currentTarget = document
```
tlabaj, line 188: can we remove these commented lines?

jschuler: it's a helpful comment to know what is on the event

PR #4500, opened 2020-07-08
fix(DrawerPanelContent): don't render children when hidden

### LINKED ISSUE #4498: Drawer should not render contents unless expanded
The contents of a `Drawer` are rendered regardless of whether the drawer is opened or closed. There should at least be an option for the drawer to not render the panel contents when collapsed. But TBH I'm not sure what use case you would ever need the contents present unless the drawer is expanded. The reason for not rendering the contents is because they could be expensive to render and / or be network chatty. This is unnecessary if the contents are not visible to the user.

A workaround is to conditionally render the `panelContents` if the drawer is expanded or not. However this has a side effect of losing the animation. The `Drawer` component should render the content immediately before opening and perform the open animation. On collapsed, the `Drawer` should animate closed and then unmount the contents.

See patternfly react documentation samples:
https://patternfly-react.surge.sh/documentation/react/components/drawer

I see this as a Bug because the workaround loses the animation.

OpenShift 4.6

redallen:
**What**: Closes #4498

**Additional issues**: The animation still works since `<div className="pf-c-drawer-panel">` is always rendered.

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -28,20 +28,24 @@ export const DrawerPanelContent: React.SFC<DrawerPanelContentProps> = ({
   ...props
 }: DrawerPanelContentProps) => (
   <DrawerContext.Consumer>
-    {({ isExpanded, isStatic }) => (
-      <div
-        className={css(
-          styles.drawerPanel,
-          hasNoBorder && styles.modifiers.noBorder,
-          formatBreakpointMods(widths, styles),
-          className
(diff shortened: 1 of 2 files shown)


christianvogt:
Without testing this yet (will wait for the test link), I'm assuming the the drawer contents will go blank for the split second the drawer is animating closed. Not a big deal, but would be nice to wait on animation end before unmounting. I wouldn't stop this PR from going in for this reason alone though.

redallen:
@christianvogt You are correct. This is also a problem with the close button:
image

We can add a timeouts to circumvent this on unmount in another PR.

PR #4503, opened 2020-07-08
fix(select): option with description on active display only value

### LINKED ISSUE #4502: Typeahead selector displays options with descriptions incorrectly
In the typeahead example: On the select menu shown clicking on the down arrow allows to go through the options.
Once an option is "active" it is colored in gray and its value is shown in the input. However, when the option has description, the description is shown with the value in the input box. In addition, the option is not colored in gray like any other option.

I believe it's a bug and no work-around exists.

3.153.3

boaz0:
**What**: Closes #4502

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -254,12 +254,17 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
   };
 
   extendTypeaheadChildren(typeaheadActiveChild: HTMLElement) {
+    let activeElement = null as HTMLElement;
+    if (Boolean(typeaheadActiveChild) && typeaheadActiveChild.classList.contains('pf-m-description')) {
+      activeElement = typeaheadActiveChild.firstElementChild as HTMLElement;
+    } else if (typeaheadActiveChild) {
+      activeElement = typeaheadActiveChild;
+    }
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Select/Select.tsx
```diff
@@ -409,6 +419,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       menuAppendTo,
       ...props
     } = this.props;
+    console.log('active state', this.state.typeaheadActiveChild)
```
dlabrecq, line 422: Please remove this console message

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -303,13 +310,16 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
         } else {
           nextIndex = getNextIndex(typeaheadCurrIndex, position, this.refCollection);
         }
+        const hasDescriptionElm = Boolean(this.refCollection[nextIndex].classList.contains('pf-m-description'));
+        const optionTextElm = hasDescriptionElm ? this.refCollection[nextIndex].firstElementChild as HTMLElement: this.refCollection[nextIndex]
+        console.log(optionTextElm)
```
jenny-s51, line 315: delete this log statement

PR #4510, opened 2020-07-09
feat(Alert): Add alert timeout functionality

### LINKED ISSUE #1654: Alerts should have an optional timeout when used as a toast
Per the PF4 design guidelines, toast alerts should time out after a certain period of time. 

"Transient toast alerts automatically dismiss after a period of time. Eight (8) seconds is a recommended timeout, but alerts should remain visible if the user is hovering over them."

Looks like this behavior is not built into the component but it should be.

tlabaj:
**What**: Closes #1654


tlabaj:
Good point at @seanforyou23!

PR #4512, opened 2020-07-10
fix(NavExpandable): converts nav expandable to button

### LINKED ISSUE #4377: NavExpandable should be a button and not a link
`NavExpandable` results in a link with `href="#"`.  Since this item is not linkable and will always have an `href` value `#`, it should not be a link.

Expected behavior to be similar to `ExpandableSection` which uses a `button`.

See example https://www.patternfly.org/v4/documentation/react/components/nav#expandable

Bugish?

OpenShift Console 4.6

jenny-s51:
**What**: Closes #4377.

**Additional issues**: Extends https://github.com/patternfly/patternfly/pull/3250 and includes version bump to pull in those  CSS changes.

### DIFF
--- a/packages/react-core/src/components/Nav/NavExpandable.tsx
+++ b/packages/react-core/src/components/Nav/NavExpandable.tsx
@@ -27,8 +27,6 @@ export interface NavExpandableProps
   id?: string;
   /** allow consumer to optionally override this callback and manage expand state externally */
   onExpand?: (e: React.MouseEvent<HTMLLIElement, MouseEvent>, val: boolean) => void;
-  /** Callback function to toggle the expandable content */
-  onClick?: () => void;
 }
 
 interface NavExpandableState {
@@ -44,8 +42,7 @@ export class NavExpandable extends React.Component<NavExpandableProps, NavExpand
     className: '',
     groupId: null as string,
     isActive: false,
-    id: '',
-    onClick: (): any => undefined
+    id: ''
   };
 
   expandableRef = React.createRef<HTMLAnchorElement>();
--- a/packages/react-core/src/components/Nav/NavExpandable.tsx
+++ b/packages/react-core/src/components/Nav/NavExpandable.tsx
@@ -28,7 +28,7 @@ export interface NavExpandableProps
   /** allow consumer to optionally override this callback and manage expand state externally */
   onExpand?: (e: React.MouseEvent<HTMLLIElement, MouseEvent>, val: boolean) => void;
   /** Callback function to toggle the expandable content */
-  onToggle?: () => void;
+  onClick?: () => void;
 }
 
 interface NavExpandableState {
@@ -45,7 +45,7 @@ export class NavExpandable extends React.Component<NavExpandableProps, NavExpand
     groupId: null as string,
     isActive: false,
     id: '',
-    onToggle: (): any => undefined
+    onClick: (): any => undefined
   };
 
   expandableRef = React.createRef<HTMLAnchorElement>();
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Nav/NavExpandable.tsx
```diff
@@ -90,9 +93,27 @@ export class NavExpandable extends React.Component<NavExpandableProps, NavExpand
   };
 
   render() {
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const { id, title, srText, children, className, isActive, groupId, isExpanded, onExpand, ...props } = this.props;
+    const {
+      onToggle: onToggleProp,
```
tlabaj, line 97: You do not need to rename this here.  I would change the name of the function on line 113 instead.  Maybe onClick?  Then you can also get rid of line 111 and declare the function on line 113 as a const.

jenny-s51: good point! will fix.

packages/react-core/src/components/Nav/NavExpandable.tsx
```diff
@@ -92,8 +95,13 @@ export class NavExpandable extends React.Component<NavExpandableProps, NavExpand
   render() {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const { id, title, srText, children, className, isActive, groupId, isExpanded, onExpand, ...props } = this.props;
+
     const { expandedState } = this.state;
 
+    const onClick = () => {
+      this.setState(prevState => ({ expandedState: !prevState.expandedState }));
```
tlabaj, line 102: You need to call onClick here if it is provided.

jenny-s51: What do you mean?

packages/react-core/src/components/Nav/NavExpandable.tsx
```diff
@@ -27,6 +27,8 @@ export interface NavExpandableProps
   id?: string;
   /** allow consumer to optionally override this callback and manage expand state externally */
   onExpand?: (e: React.MouseEvent<HTMLLIElement, MouseEvent>, val: boolean) => void;
+  /** Callback function to toggle the expandable content */
+  onClick?: () => void;
```
dlabrecq, line 31: I don't see where this `onClick` prop is called?

packages/react-integration/demo-app-ts/src/components/demos/NavDemo/NavDemo.tsx
```diff
@@ -59,7 +59,7 @@ export class NavDemo extends Component {
         <Title headingLevel="h2" size="2xl">
           Default Nav
         </Title>
-        <div className="example" style={{ border: '1px solid rgb(114, 118, 123)' }}>
+        <div className="example" style={{ backgroundColor: '#212427', border: '1px solid rgb(114, 118, 123)' }}>
```
tlabaj, line 62: why do we need this change?

packages/react-integration/demo-app-ts/src/components/demos/NavDemo/NavDemo.tsx
```diff
@@ -124,7 +124,7 @@ export class NavDemo extends Component {
         <Title headingLevel="h2" size="2xl">
           Expandable Nav
         </Title>
-        <div className="example" style={{ border: '1px solid rgb(114, 118, 123)' }}>
+        <div className="example" style={{ backgroundColor: '#212427', border: '1px solid rgb(114, 118, 123)' }}>
```
tlabaj, line 127: same question for this.

packages/react-integration/demo-app-ts/src/components/demos/NavDemo/NavDemo.tsx
jenny-s51: @tlabaj the nav demo looked like this before adding that styling: https://github.com/patternfly/patternfly/issues/3279. per @redallen, that is actually our default nav styling! He mentioned it needs to be on a dark background for it to look right.

packages/react-integration/demo-app-ts/src/components/demos/NavDemo/NavDemo.tsx
jenny-s51: see above 🙂

PR #4516, opened 2020-07-10
feat(NotificationBadge): add needs attention state and count

### LINKED ISSUE #4431: Updates to notification badge
Placeholder for https://github.com/patternfly/patternfly/issues/2994

nicolethoen:
Closes #4431 

Preview: https://patternfly-react-pr-4516.surge.sh/documentation/react/components/notificationbadge


packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
```diff
@@ -16,13 +20,21 @@ export interface NotificationBadgeProps extends ButtonProps {
...
   className,
   children,
   ...props
 }: NotificationBadgeProps) => (
   <Button variant={ButtonVariant.plain} className={className} {...props}>
-    <span className={css(styles.notificationBadge, isRead ? styles.modifiers.read : styles.modifiers.unread)}>
+    <span
+      className={css(
+        styles.notificationBadge,
+        isRead ? styles.modifiers.read : isNeedingAttention ? styles.modifiers.attention : styles.modifiers.unread
```
redallen, line 33: Isn't this a variant? Based on the [core examples](https://pf4.patternfly.org/documentation/core/components/notificationbadge) I assume a badge CANNOT be "isRead" and "isNeedingAttention."

If that is the case I think it'd be good to add a variant prop rather than `isNeedingAttention`. To not be a breaking change we can have `isRead` just set the variant to `read` or `unread` if not defined.

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
```diff
@@ -6,6 +6,10 @@ import styles from '@patternfly/react-styles/css/components/NotificationBadge/no
 export interface NotificationBadgeProps extends ButtonProps {
   /**  Adds styling to the notification badge to indicate it has been read */
   isRead?: boolean;
+  /** Adds styling to the notification badge to indicate it needs attention */
+  isNeedingAttention?: boolean;
+  /** A number displayed in the badge alongside the icon */
+  count?: number;
```
redallen, line 12: Can we just make this `number | string` and remove the `count > 0` condition? Someone might want to put something else in there.

mceledonia:
I think there are two parts to this:

1. The badge count should optional, so products can choose to show the total number of notifications or just have the badge styling changes to indicate none/new/attention. Regardless of whether or not there is a count, the badge styling for each state is the same.
2. If the user does choose to show a count, I believe we decided that it would always only show the total number of alerts rather than trying to decipher between read/unread. @mcarrano double check my recollection there! 

So all that to say the count shouldn't disappear when the drawer is activated if the user chooses to include a count.

jcaianirh:
> I don't believe the interaction is quite right here. I'm looking at https://marvelapp.com/prototype/4f0341a/screen/68220243 and it does not seem like the count is intended to be hidden entirely when the user clicks on the badge.
> 
> @mceledonia @jcaianirh can you take a look and verify what the expected behavior should be?

It looks like the behavior of the alert count reset is defined in the on click which is implemented per project...maybe a project that has the concept of read/unread would want the count to correlate to the unread alerts which would reset once the badge is selected? For openshift, we wouldn't wan the count to be removed.

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
tlabaj: @mcarrano do we want to allow a sting here?

tlabaj:
> > I don't believe the interaction is quite right here. I'm looking at https://marvelapp.com/prototype/4f0341a/screen/68220243 and it does not seem like the count is intended to be hidden entirely when the user clicks on the badge.
> > @mceledonia @jcaianirh can you take a look and verify what the expected behavior should be?
> 
> It looks like the behavior of the alert count reset is defined in the on click which is implemented per project...maybe a project that has the concept of read/unread would want the count to correlate to the unread alerts which would reset once the badge is selected? For openshift, we wouldn't wan the count to be removed.

So should our example should reflect the recommended usage then.

nicolethoen:
> > > I don't believe the interaction is quite right here. I'm looking at https://marvelapp.com/prototype/4f0341a/screen/68220243 and it does not seem like the count is intended to be hidden entirely when the user clicks on the badge.
> > > @mceledonia @jcaianirh can you take a look and verify what the expected behavior should be?
> > 
> > 
> > It looks like the behavior of the alert count reset is defined in the on click which is implemented per project...maybe a project that has the concept of read/unread would want the count to correlate to the unread alerts which would reset once the badge is selected? For openshift, we wouldn't wan the count to be removed.
> 
> @mceledonia @mcarrano So should our example should reflect the recommended usage then.

And what would that look like since there are no actual notifications in the example?

nicolethoen:
> @mcarrano @nicolethoen @mceledonia I think the read/unread state needs to be updated before this pr can be merged... when the unread state with the square box was leaked to openshift, the entire team thought there was a bug in the badge and that the focus state of the badge was always true, when it was just the unread treatment. I think we need to come up with something that doesn't get confused with focus state here.

Oh yeah, I'm also planning to put this on a dark background like its in a header.

jcaianirh:
@nicolethoen Its not the black background, its the blue square border that is the issue. See https://github.com/patternfly/patternfly-design/issues/895

mcoker:
fwiw, we have a core issue to investigate these sorts of issues - https://github.com/patternfly/patternfly/issues/3131

mcarrano:
Yes, just to follow-up on this - @jcaianirh and I discussed and we agreed that this should be put on hold until we can solve the focus vs unread state issue which will require further design investigation. In the interim, I think it would still be worth working through the logic of how the demo should behave, but would not merge this until we have updated visual styling.

cc @evwilkin @tlabaj

nicolethoen:
> @nicolethoen Its not the black background, its the blue square border that is the issue. See [patternfly/patternfly-design#895](https://github.com/patternfly/patternfly-design/issues/895)

I understand. I just wanted this PR as ready as possible for the style changes we are waiting for :P

nicolethoen:
Once this finishes building, and hopefully when surge is working again, @mcarrano @mattnolting @jcaianirh this is hopefully ready to be reviewed again.

mcarrano:
A question about the count. I believe this is intended to be the number of Unread items, correct? So if clicking the badge clears the unread items, should it also reset the count to 0? @mceledonia @jcaianirh can you confirm what the intended behavior is?

nicolethoen:
> A question about the count. I believe this is intended to be the number of Unread items, correct? So if clicking the badge clears the unread items, should it also reset the count to 0? @mceledonia @jcaianirh can you confirm what the intended behavior is?

That's how I had it originally, I thought I was told to leave them. but I can easily add it back.

nicolethoen:
Not sure surge is working.
Jul-30-2020 14-10-26

mcoker:
@nicolethoen we're going to create a custom bell icon in the same way we created the attention-bell icon for this use, and in cases where the bell and attention-bell should be used interchangeably.  https://github.com/patternfly/patternfly/issues/3350

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
```diff
@@ -2,27 +2,49 @@ import * as React from 'react';
...
-export interface NotificationBadgeProps extends ButtonProps {
+export enum NotificationBadgeVariant {
+  read = 'read',
+  unread = 'unread',
+  attention = 'attention'
+}
+
+export interface NotificationBadgeProps extends Omit<ButtonProps, 'variant'> {
   /**  Adds styling to the notification badge to indicate it has been read */
   isRead?: boolean;
```
jschuler, line 16: deprecate this in favor of variant?

nicolethoen: I'm just trying to avoid breaking changes. is there a best way to deprecate it? Should I add a warning to anyone using it?

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
jschuler: Is there no use case for a string? e.g. `${count} messages`

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
```diff
@@ -2,27 +2,49 @@ import * as React from 'react';
...
+  attentionIcon = <AttentionBellIcon />,
+  icon = <BellIcon />,
   className,
   children,
   ...props
 }: NotificationBadgeProps) => (
   <Button variant={ButtonVariant.plain} className={className} {...props}>
-    <span className={css(styles.notificationBadge, isRead ? styles.modifiers.read : styles.modifiers.unread)}>
+    <span className={css(styles.notificationBadge, styles.modifiers[variant])}>
       {children}
```
jschuler, line 45: i don;t see any examples with children being used, when I tried it just now the text was not visible

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
jschuler: maybe in the docs comment add 
`/** @deprecated Use the variant prop instead - Adds styling to the notification badge to indicate it has been read */`
and in code
```
// when you destructure isRead don't give it the default value of false
if (process.env.NODE_ENV !== 'production' && isRead !== undefined) {
  console.warn(
        'The isRead prop has been deprecated. Use the variant prop instead.'
      );
}
```

packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx
jschuler: btw, i am not sure the previous examples would continue to work which would make it breaking in this form
```
<NotificationBadge isRead={isRead} onClick={this.onClick} aria-label="Notifications">
        <BellIcon />
      </NotificationBadge>
```

PR #4517, opened 2020-07-12
feat(ContextSelector): Add ability to render content before searchbar

MariaAga:
Adds the ability to render custom content before the search bar - good for titling the context selector, adding non filterable items, adding 'add project' buttons and more.
Added a storybook item as an example for title + non filterable item

### DIFF
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -43,6 +43,8 @@ export interface ContextSelectorProps extends ToggleMenuBaseProps, OUIAProps {
   searchInputPlaceholder?: string;
   /** Function callback for when Search Button is clicked */
   onSearchButtonClick?: (event?: React.SyntheticEvent<HTMLButtonElement>) => void;
+  /** content that is inserted before the search bar filter */
+  menuHeader?: React.ReactNode;
 }
 
 export class ContextSelector extends React.Component<ContextSelectorProps> {
@@ -61,7 +63,8 @@ export class ContextSelector extends React.Component<ContextSelectorProps> {
     searchInputPlaceholder: 'Search',
     onSearchButtonClick: () => undefined as any,
     menuAppendTo: 'inline',
-    ouiaSafe: true
+    ouiaSafe: true,
+    menuHeader: null as React.ReactNode
   };
 
   parentRef: React.RefObject<HTMLDivElement> = React.createRef();
@@ -92,12 +95,15 @@ export class ContextSelector extends React.Component<ContextSelectorProps> {
       menuAppendTo,
       ouiaId,
       ouiaSafe,
+      menuHeader,
       ...props
     } = this.props;
+
     const menuContainer = (
       <div className={css(styles.contextSelectorMenu)}>
         {isOpen && (
           <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
+            {menuHeader && <ContextSelectorMenuList isOpen={isOpen}>{menuHeader}</ContextSelectorMenuList>}
             <div className={css(styles.contextSelectorMenuSearch)}>
               <InputGroup>
                 <TextInput
--- a/packages/react-core/src/components/ContextSelector/examples/ContextSelector.md
+++ b/packages/react-core/src/components/ContextSelector/examples/ContextSelector.md
@@ -6,7 +6,9 @@ ouia: true
 ---
 
 ## Examples
+
 ### Basic
+
 ```js
 import React from 'react';
 import { ContextSelector, ContextSelectorItem } from '@patternfly/react-core';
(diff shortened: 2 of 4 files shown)


packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -92,12 +95,15 @@ export class ContextSelector extends React.Component<ContextSelectorProps> {
...
       ouiaSafe,
+      beforeSearch,
       ...props
     } = this.props;
+
     const menuContainer = (
       <div className={css(styles.contextSelectorMenu)}>
         {isOpen && (
           <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
+            <ContextSelectorMenuList isOpen={isOpen}>{beforeSearch}</ContextSelectorMenuList>
```
nicolethoen, line 106: You might want to check that you have something you want to render here before you include the `ContextSelectorMenuList`. if the prop is undefined or null, then you probably don't want to include the element.

MariaAga: Yeah you're right, thanks! added a `beforeSearch && <ContextSelectorMenuList...`

MariaAga:
This doesn't have an issue assigned to it, the 'projects' title in the example is just an example of how a user can customize the `before search`, if a designer can suggest me some styling I'll gladly add it to the example :)

Rohoover:
Hey all - So I see the purpose of adding "All Projects" but don't understand why it is located before the search bar, as opposed to after?

MariaAga:
@Rohoover Because it's not an actual project name, and if it was under the search bar it might be confused with a project named "all projects". Also we can use the above search bar to link 'manage projects' option

packages/react-core/src/components/ContextSelector/examples/ContextSelector.md
```diff
@@ -82,3 +82,91 @@ class SimpleContextSelector extends React.Component {
...
+      <ContextSelector
+        toggleText={selected}
+        onSearchInputChange={this.onSearchInputChange}
+        isOpen={isOpen}
+        searchInputValue={searchValue}
+        onToggle={this.onToggle}
+        onSelect={this.onSelect}
+        onSearchButtonClick={this.onSearchButtonClick}
+        screenReaderLabel="Selected Project:"
+        beforeSearch={
```
laviro, line 154: can you rename it to something like `menuHeader`?

MariaAga: Yeah, thanks for the suggestion

MariaAga:
@jschuler Can you take a look, its for the layout refactor in Foreman? :)

nicolethoen:
What is the reasoning behind the `All Projects` option in the context selector always appearing selected when I expand the menu? Even if I had selected a different project, `All Projects` is in focus.

MariaAga:
I'm not sure where it comes from I think it's for accessibility, but without `All Projects` the search bar is shown as selected. It's only happening once you open the dropdown.
selected and un-selected search bar screenshots:
Screenshot from 2020-07-21 18-45-02
Screenshot from 2020-07-21 18-45-17

nicolethoen:
I don't know the history of this feature request, but I think the example is a little problematic with the `All Projects` option being above the search bar. Is this actually a case that Patternfly would endorse @mcarrano? Usually every example in our docs is an approved/suggested use case. Certainly the focus should stay on the search when the menu is expanded. Could there possibly be a different use case to demonstrate this new feature in the new example?

MariaAga:
I can change it to 'Manage projects' so if it's selected it would make more sense?

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -92,12 +95,15 @@ export class ContextSelector extends React.Component<ContextSelectorProps> {
...
       ouiaSafe,
+      menuHeader,
       ...props
     } = this.props;
+
     const menuContainer = (
       <div className={css(styles.contextSelectorMenu)}>
         {isOpen && (
           <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
+            {menuHeader && <ContextSelectorMenuList isOpen={isOpen}>{menuHeader}</ContextSelectorMenuList>}
```
jschuler, line 106: is this blessed by Core? am not seeing anything like that in the examples there @mcoker https://pf4.patternfly.org/documentation/core/components/contextselector

evwilkin:
@MariaAga thanks for this contribution and asking for any additional design input! After chatting with the PF design team, we'd like to send this idea through the PF design & core teams to align on the presentation and markup before pulling it into PF React.

I've created a design issue here (https://github.com/patternfly/patternfly-react/pull/4517), we can keep this PR open in the meantime and follow up on any updates needed to it from these teams. Thanks!

mcarrano:
@MariaAga @evwilkin We have completed our design investigation and you can see our proposed solution here: https://github.com/patternfly/patternfly-design/issues/897 If we want to proceed with adding this to PatternFly, then the next step will be to open a core issue to build this into the HTML structure for the Context Selector and then we can complete the React work. Let me know if I should open an issue to move ahead with this solution.

evwilkin:
@mcarrano this looks good to me

MariaAga:
@mcarrano Thanks! I'm not sure where to open an issue, but I'll be happy to implement it in React

mcarrano:
@MariaAga There is currently a core issue open here to build out the HTML structure for this: https://github.com/patternfly/patternfly/issues/3475 Once that is complete, it will be ready to pick up in React.

@evwilkin @tlabaj how do you want to proceed? This PR should probably be closed and a new React issue opened that can be picked up once the core work is complete. Thoughts?

evwilkin:
@mcarrano I agree a new issues makes sense, I've created one and referenced this issue in it for any additional context needed.

Closing in favor of https://github.com/patternfly/patternfly-react/issues/4865

PR #4520, opened 2020-07-13
feat(toolbar): filter chips can be toolbar chip group type

### LINKED ISSUE #4474: ToolbarFilter chips to accept ToolbarChipGroup type
Currently, the `chips` prop of `ToolbarFilter` accepts an array of strings, while `deleteChip` and `categoryName` accept `ToolbarChipGroup`. Ideally, `chips` would also accept the `ToolbarChipGroup` type. 

I want to display localized strings for each chip while working with IDs. I could do that with `ToolbarChip`, providing a ReactNode. However, I simply need the `name` string from `ToolbarChipGroup`.

boaz0:
**What**: Closes #4474 

@dlabrecq do you mind reviewing this to make sure I got it right. Thanks!

### DIFF
--- a/packages/react-integration/cypress/integration/toolbar.spec.ts
+++ b/packages/react-integration/cypress/integration/toolbar.spec.ts
@@ -127,3 +127,73 @@ describe('Data Toolbar Demo Test', () => {
     });
   });
 });
+
+describe('Filter by type', () => {
+  it('Filter by Node and Region and clear all', () => {
+    cy.get('.pf-c-chip-group')
+      .eq(0)
+      .get('li.pf-c-chip-group__list-item')
+      .last()
+      .should('have.text', 'Node');
+    cy.get('.pf-c-toolbar__item')
+      .eq(3)
+      .click();
+    cy.get('label.pf-c-select__menu-item')
+      .eq(2)
+      .click();
+    cy.get('.pf-c-toolbar__item')
+      .eq(3)
+      .click();
+    cy.get('.pf-c-chip-group')
+      .eq(2)
+      .get('li.pf-c-chip-group__list-item')
+      .last()
+      .should('have.text', 'Region');
+    cy.get('div.pf-c-chip-group__close')
+      .eq(1)
+      .click();
+  });
+  it('Toggle Node filter', () => {
+    cy.get('.pf-c-chip-group').should('have.length', 2);
+    cy.get('.pf-c-toolbar__item')
+      .eq(3)
+      .click();
+    cy.get('label.pf-c-select__menu-item')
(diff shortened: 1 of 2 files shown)


nicolethoen:
So this does not seem to make sense to me. I read @dlabrecq's issue, and I still don't quite understand.

A `ToolbarChipGroup` is supposed to be an object that represents data about groups of chips. I believe `deleteChip` was modified to accept `ToolbarChipGroups` so that the user could delete a whole group of chips with one click. And the name in `ToolbarChipGroup` is meant to represent the display name of a chip group category.

so I don't think it's appropriate for `chips` to be represented as `ToolbarChipGroup` objects. Plus, everything you could conceivably do with a `ToolbarChipGroup`, i believe you could also do with a `ToolbarChip`, right?

Maybe I'm misunderstanding the use case described in the issue.

boaz0:
Thanks @nicolethoen 
There might be a misunderstanding on my side. I will reach out @dlabrecq and understand what he meant.
I will update you when I have news.

dlabrecq:
When I created the issue, it wasn't clear why we had two, similar interfaces. The ReactNode type works with strings; although, I'd prefer to see one interface to cut down the number of objects created for each prop. 

If there's nothing we can do to clarify the interfaces, feel free to close. I was able to implement the filter using a ReactNode with ToolbarChip.

nicolethoen:
> When I created the issue, it wasn't clear why we had two, similar interfaces. The ReactNode type works with strings; although, I'd prefer to see one interface. I'm only working with strings and wanted to cut down the number of unique objects created for each prop.
> 
> If there's nothing we can do to clarify the interfaces, feel free to close. I was able to implement localized labels using a ReactNode with ToolbarChip.

hm... I think the two interfaces were built for two different scenarios. Basically, one was built in the case that someone wanted to have their chips to contain something more complex than a simple string. The other was built in the case that someone wanted to have their chip group's displayed category name be different than its unique identifier (because some people's APIs were using unreadable ids as filter categories).

@dlabrecq @boaz0 
Is there a clever name for this interface that would encompass both scenarios? or better ways to name the two existing interfaces to eliminate confusion?

boaz0:
@dlabrecq @nicolethoen - IMHO `ToolbarChip` is a good name but changing its `node` attribute to `label` and its type to  `React.ReactNode | string` will make it better.

I will change this PR to just have the integration tests as it's too bad to discard that.

boaz0:
Sorry, I forgot about this. I will find time to work on it this week. :open_mouth:

boaz0:
Reviewing this, I realized that the current demo is better and my changes are redundant

PR #4536, opened 2020-07-14
fix(overflowmenu-dropdown-item): allow to pass dropdown item props

### LINKED ISSUE #4372: Should be able to pass any DropdownItem props to OverflowMenuDropdownItem
Should be able to pass any DropdownItem props to OverflowMenuDropdownItem

boaz0:
**What**: Closes #4372 

//cc @jschuler

### DIFF
--- a/packages/react-core/src/components/OverflowMenu/OverflowMenuDropdownItem.tsx
+++ b/packages/react-core/src/components/OverflowMenu/OverflowMenuDropdownItem.tsx
@@ -1,10 +1,8 @@
 import * as React from 'react';
-import { DropdownItem } from '../Dropdown';
+import { DropdownItem, DropdownItemProps } from '../Dropdown';
 import { OverflowMenuContext } from './OverflowMenuContext';
 
-export interface OverflowMenuDropdownItemProps extends React.HTMLProps<HTMLDivElement> {
-  /** Any elements that can be rendered in the menu */
-  children?: any;
+export interface OverflowMenuDropdownItemProps extends DropdownItemProps {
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/OverflowMenu/OverflowMenuDropdownItem.tsx
```diff
@@ -14,12 +12,13 @@ export interface OverflowMenuDropdownItemProps extends React.HTMLProps<HTMLDivEl
...
   isShared = false,
-  index
+  index,
+  ...additionalProps
 }: OverflowMenuDropdownItemProps) => (
   <OverflowMenuContext.Consumer>
     {value =>
       (!isShared || value.isBelowBreakpoint) && (
-        <DropdownItem component="button" index={index}>
+        <DropdownItem component="button" index={index} {...additionalProps}>
```
dlabrecq, line 21: I'd rather be explicit about what props are passed down Vs anything random. The issue is only asking for DropdownItem props to be passed thru. Can we destructure the valid props here?

boaz0: I don't think we can, dropdown items props extends `HTMLDivElement` and that one contains much much more props.
Unless, we do not need to pass `HTMLDivElement` props, I believe we can do that explicitly.

dlabrecq: Looking at other components, it appears props are passed thru without any validation.

PR #4543, opened 2020-07-14
Tooltip/Popover - add tippyProps back so as not to cause breaking changes

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4542

Add deprecated `tippyProps` back for Tooltip and Popover.
~~Add tippy.js back as a dependency in case consumers were depending on it.~~

If you previously imported from `tippy.js` and were depending on it being side-loaded by react-core, you'll need to either add it as a dependency to your own library:
`"tippy.js": "5.1.2"`
or if you just need the types:
`import { Instance as TippyInstance } from '@patternfly/react-core/dist/js/helpers/Popper/DeprecatedTippyTypes';`

### DIFF
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -42,6 +42,7 @@
     "focus-trap": "4.0.2",
     "react-dropzone": "9.0.0",
     "react-popper": "2.2.3",
+    "tippy.js": "5.1.2",
     "tslib": "^1.11.1"
   },
   "devDependencies": {
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -42,7 +42,6 @@
     "focus-trap": "4.0.2",
     "react-dropzone": "9.0.0",
     "react-popper": "2.2.3",
-    "tippy.js": "5.1.2",
     "tslib": "^1.11.1"
   },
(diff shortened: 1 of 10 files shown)


packages/react-core/package.json
```diff
@@ -42,6 +42,7 @@
     "focus-trap": "4.0.2",
     "react-dropzone": "9.0.0",
     "react-popper": "2.2.3",
+    "tippy.js": "5.1.2",
```
redallen, line 45: This defeats the purpose of #4491. Can we find a way to not add this dep back?

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -14,6 +15,7 @@ import { ReactElement } from 'react';
 import { FocusTrap } from '../../helpers';
 import { Popper, getOpacityTransition } from '../../helpers/Popper/Popper';
 import { getUniqueId } from '../../helpers/util';
+import { Instance as TippyInstance, Props as TippyProps } from 'tippy.js';
```
redallen, line 18: Can we find a way to copy the `TippyInstance` and `TippyProps` definitions instead?

PR #4544, opened 2020-07-14
fix(Modal): ensure getElement does not return null

### LINKED ISSUE #4093: Modal(PF4): when not specifying appendTo property the component will crash
When using the PF modal we get an error in the console as follows:

```
TypeError: target is null Modal.js:72
    componentDidMount Modal.js:72
    React 6
    commitRootImpl self-hosted:981
    unstable_runWithPriority scheduler.development.js:653
    React 4
    performSyncWorkOnRoot self-hosted:926
    flushSyncCallbackQueueImpl React
    unstable_runWithPriority scheduler.development.js:653
    React 6
    bind_applyFunctionN self-hosted:1062
    dispatchDiscreteEvent self-hosted:1041
```

This can be avoided by passing 'appendTo={document.body}' to the component (which seems to be the default value anyway).
Apparently another way to WA this is by moving the <script> part in index.html after <body>, so when document.body is present when Modal is imported.

In any case - this is not a blocker for us (cockpit) so use the modals but it should be addressed.

nicolethoen:
Closes #4093 
@KKoukiou could you test this change and be extra sure this addresses your specific issue?

### DIFF
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -105,13 +105,10 @@ export class Modal extends React.Component<ModalProps, ModalState> {
   };
 
   getElement = (appendTo: HTMLElement | (() => HTMLElement)) => {
-    let target: HTMLElement;
     if (typeof appendTo === 'function') {
-      target = appendTo();
-    } else {
-      target = appendTo;
+      return appendTo();
     }
-    return target;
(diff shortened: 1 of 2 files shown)


nicolethoen:
@KKoukiou any word on this?

KKoukiou:
@nicolethoen works as expected. Thanks.

Issue #4530, opened 2020-07-13
Tooltip/popover - support trigger passed as selector / reference


jschuler:
Tooltip/popover have to pass the trigger as a child currently:
```
<Popover bodyContent="Some text">
  <Button>Trigger</Button>
</Popover>
```
Also support passing trigger through selector/reference:
```
const myNodeRef = React.useRef();
...
<Popover
  reference={myNodeRef.current}
```
or
```
<Popover
  reference={() => document.querySelector(...)}
```

PR #4546, opened 2020-07-14
Tooltip/Popover feature: Add `reference` prop

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4530

Adds more options to attach a Tooltip/Popover to a trigger with the `reference` prop.
Using selector:
```
<div>
          <button id="popover-selector">Popover attached via selector ref</button>
          <Popover
            bodyContent="Content"
            reference={() => document.getElementById('popover-selector')}
          />
</div>
```
Using ref:
```
const popoverRef = React.useRef();
<div>
          <button ref={popoverRef}>
            Popover attached via react ref
          </button>
          <Popover
            bodyContent="Content"
            reference={popoverRef}
          />
</div>
```


packages/react-core/src/helpers/Popper/Popper.tsx
```diff
@@ -108,20 +116,33 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
...
+      if ((reference as React.RefObject<any>).current) {
+        setRefElement((reference as React.RefObject<any>).current);
+      } else if (typeof reference === 'function') {
+        setRefElement(reference());
+      } else {
+        setRefElement(null);
+      }
+    } else {
+      setRefElement(null);
+    }
```
redallen, line 144: Instead of these two branches, can't you set the default to null like `const [refElement, setRefElement] = React.useState<HTMLElement>(null);`?

packages/react-core/src/helpers/Popper/Popper.tsx
```diff
@@ -133,22 +154,22 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
...
-    addEventListener(onFocus, triggerElement, 'focus');
-    addEventListener(onBlur, triggerElement, 'blur');
-    addEventListener(onTriggerClick, triggerElement, 'click');
-    addEventListener(onTriggerEnter, triggerElement, 'keydown');
+    addEventListener(onMouseEnter, refElement || triggerElement, 'mouseenter');
+    addEventListener(onMouseLeave, refElement || triggerElement, 'mouseleave');
+    addEventListener(onFocus, refElement || triggerElement, 'focus');
+    addEventListener(onBlur, refElement || triggerElement, 'blur');
+    addEventListener(onTriggerClick, refElement || triggerElement, 'click');
+    addEventListener(onTriggerEnter, refElement || triggerElement, 'keydown');
```
redallen, line 162: DRY would be like:
```js
const popperElement = refElement || triggerElement; // this line can be included above React.useEffect
addEventListener(onMouseEnter, element, 'mouseenter');
...
```

packages/react-core/src/helpers/Popper/Popper.tsx
```diff
@@ -198,7 +220,8 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
     }),
     [popperMatchesTriggerWidth]
   );
-  const { styles: popperStyles, attributes } = usePopper(triggerElement, popperElement, {
+
+  const { styles: popperStyles, attributes } = usePopper(refElement || triggerElement, popperElement, {
```
redallen, line 224: Replace `refElement || triggerElement` with `popperElement`

packages/react-core/src/helpers/Popper/Popper.tsx
```diff
@@ -108,20 +116,33 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
...
   const [popperElement, setPopperElement] = React.useState(null);
   const [ready, setReady] = React.useState(false);
-  const onDocumentClickCallback = React.useCallback(event => onDocumentClick(event, triggerElement, popperElement), [
-    isVisible,
-    triggerElement,
-    popperElement,
-    onDocumentClick
-  ]);
+  const onDocumentClickCallback = React.useCallback(
+    event => onDocumentClick(event, refElement || triggerElement, popperElement),
```
redallen, line 127: Replace `refElement || triggerElement` with `popperElement`.

jschuler:
@dlabrecq children and reference are mutually exclusive

packages/react-core/src/helpers/Popper/Popper.tsx
jschuler: Thanks, popperElement is the element that pops up, so i've named it instead refOrTrigger

PR #4553, opened 2020-07-15
fix(Toolbar): deleteChip expects categoryKey as string

### LINKED ISSUE #3552: DataToolbarFilter: categoryName label required to delete chips

nicolethoen:
Closes #3552

### DIFF
--- a/packages/react-core/src/components/Toolbar/ToolbarFilter.tsx
+++ b/packages/react-core/src/components/Toolbar/ToolbarFilter.tsx
@@ -67,21 +67,23 @@ export class ToolbarFilter extends React.Component<ToolbarFilterProps, ToolbarFi
   render() {
     const { children, chips, deleteChipGroup, deleteChip, categoryName, showToolbarItem, ...props } = this.props;
     const { isExpanded, chipGroupContentRef } = this.context;
+    const categoryKey = typeof categoryName === 'string' ? categoryName : categoryName.key;
+
     const chipGroup = chips.length ? (
       <ToolbarItem variant="chip-group">
         <ChipGroup
-          key={typeof categoryName === 'string' ? categoryName : categoryName.key}
+          key={categoryKey}
           categoryName={typeof categoryName === 'string' ? categoryName : categoryName.name}
           isClosable={deleteChipGroup !== undefined}
           onClick={() => deleteChipGroup(categoryName)}
         >
           {chips.map(chip =>
             typeof chip === 'string' ? (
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Toolbar/ToolbarFilter.tsx
```diff
@@ -67,21 +67,23 @@ export class ToolbarFilter extends React.Component<ToolbarFilterProps, ToolbarFi
   render() {
     const { children, chips, deleteChipGroup, deleteChip, categoryName, showToolbarItem, ...props } = this.props;
     const { isExpanded, chipGroupContentRef } = this.context;
+    const categoryKey = typeof categoryName === 'string' ? categoryName : categoryName.key;
```
dlabrecq, line 70: Thinking we should test `typeof ToolbarChipGroup` before returning `categoryName.key`? (It's possible that `key` may not be defined for the given `categoryName` object.) That way, the string could be returned as the default.

nicolethoen: The only types valid for categoryName is string or ToolbarChipGroup which requires a key. So it should be safe

dlabrecq: Although categoryName is typed, users can provide a different object using `any` (e.g., categoryName={myObject as any}). I'm just advocating that it's safer to test for the ToolbarChipGroup type first and default to string; thus, avoiding a potential undefined exception.

dlabrecq: Although `categoryName` is typed, users can still provide a different object (e.g., `categoryName={myObject as any}`). I'm just advocating that it's safer to test for the `ToolbarChipGroup` interface first (since you're already using `typeof`) and default to string; thus, avoiding a potential undefined exception.

PR #4560, opened 2020-07-16
fix(Toolbar*): change spelling of visiblity prop

### LINKED ISSUE #4501: Toolbar: Add new prop with correct spelling "Visibility" to Toolbar Components
See issue #4499. The prop `visiblity` is misspelled.  Since changing it would be a breaking change, let's add a new prop with the correct spelling and deprecate the misspelled prop.  All examples should be updated to use the correctly spelt prop

The following components all need to 

- ToolbarItem
- ToolbarContent
- ToolbarGroup
- ToolbarToggleGroup

zanewoodfin:
- deprecate visiblity prop in ToolbarItem, ToolbarContent, ToolbarGroup, and ToolbarToggleGroup
- add visibility prop in ToolbarItem, ToolbarContent, ToolbarGroup, and ToolbarToggleGroup

fix #4501

### DIFF
--- a/packages/react-core/src/components/Toolbar/ToolbarContent.tsx
+++ b/packages/react-core/src/components/Toolbar/ToolbarContent.tsx
@@ -9,6 +9,14 @@ export interface ToolbarContentProps extends React.HTMLProps<HTMLDivElement> {
   /** Classes applied to root element of the data toolbar content row */
   className?: string;
   /** Visibility at various breakpoints. */
+  visibility?: {
+    default?: 'hidden' | 'visible';
+    md?: 'hidden' | 'visible';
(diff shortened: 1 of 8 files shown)


zanewoodfin:
> Can you also please update the demo app to add the correctly spelt visibility prop.

@tlabaj The demo app does not seem to have any instances of the old, misspelled prop.  Do you want me to create some sort of new example with it?

tlabaj:
@zanewoodfin yes.  You should add the new prop and the old one as well to the demo the verify that the work as expected.  The old prop should have been in the demo.  Please test all breakpoints we have add a lot of issues where breakpoints were not working as expected.

Issue #4551, opened 2020-07-15
Update user menu example to include Avatar and not user icon


mcarrano:
This example should be showing how an Avatar component can be integrated into a dropdown to create a user menu. https://www.patternfly.org/v4/documentation/react/components/dropdown#dropdown-with-image-and-text  This will make it consistent with the core example, also.

cc @tlabaj

mcarrano:
@tlabaj Looks like this was done?

tlabaj:
closed by #4566

PR #4566, opened 2020-07-16
docs(Dropdown.md): update dropdown with image example to use an Avatar

zanewoodfin:
- update the dropdown with image example to use an Avatar instead of a UserIcon

re #4551

### DIFF
--- a/packages/react-core/src/components/Dropdown/examples/Dropdown.md
+++ b/packages/react-core/src/components/Dropdown/examples/Dropdown.md
@@ -1164,13 +1164,14 @@ class RouterDropdown extends React.Component {
 ```js title=Dropdown-with-image-and-text
 import React from 'react';
 import {
+  Avatar,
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Dropdown/examples/Dropdown.md
```diff
@@ -1164,13 +1164,14 @@ class RouterDropdown extends React.Component {
...
 import {
+  Avatar,
   Dropdown,
   DropdownGroup,
   DropdownToggle,
   DropdownItem,
   DropdownSeparator
 } from '@patternfly/react-core';
-import { UserIcon } from '@patternfly/react-icons';
+import avatarImg from '../../Avatar/examples/avatarImg.svg';
```
tlabaj, line 1174: Should we be using the Avatar component here @mcoker?

PR #4579, opened 2020-07-20
fix(Select): use width prop in select

### LINKED ISSUE #4509: Select - width prop doesn't work
Per documentation Select has a `width` prop, however changing its value does nothing. Looking at the code the value does not seem to propagate anywhere.

https://codesandbox.io/s/keen-rubin-ckic1

Bug. There is a workaround by wrapping the component in something with a set width.

N/A

MariaAga:
fix #4509

**What**: Closes #4509


tlabaj:
@jschuler can you weigh in here.  I think width may have been removed to the `menuAppendTo` changes

jschuler:
@tlabaj yes you're right, looks like it was accidentally removed in the menuAppendTo changes.

PR #4582, opened 2020-07-20
chore(packages): replace classnames with css function

redallen:
**What**: Our [`css` implementation in `@patternfly/react-styles`](https://github.com/patternfly/patternfly-react/blob/master/packages/react-styles/src/index.ts#L5-L30) is an exact copy of the `classnames` package since the major release. This means that the 15 remaining usages of `classNames()` can be replaced with `css()` to eliminate a dependency and reduce bundle sizes.

**Additional issues**:


evwilkin:
@redallen 2 approvals but can't merge due to conflicts

PR #4586, opened 2020-07-20
feat(DescriptionList): add new DescriptionList component

### LINKED ISSUE #4523: Add description list component
New component! https://github.com/patternfly/patternfly/pull/3243

cc @mattnolting

jenny-s51:
**What**: Closes #4523


packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  /** Additional classes added to the list */
+  className?: string;
+  /** Sets the description list component term and description pair to a horizontal layout. */
+  isHorizontal?: boolean;
+  /** Sets the description list to format automatically. */
+  hasAutoColumnWidths?: boolean;
+  /** Modifies the description list display to inline-grid. */
+  hasInlineGrid?: boolean;
+  /** Sets the number of columns on the description list */
+  variant?: '1Col' | '2Col' | '3Col';
```
tlabaj, line 17: Since the columns are breakpoint dependent,  would use the interface we use in other components.  Look at Tabs `inset` prop  for an example. I also would rename it.  It is not really a variant.

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  hasInlineGrid,
+  variant,
+  variantOnMd,
+  variantOnLg,
+  variantOnXl,
+  variantOn2xl,
+  ...props
+}: DescriptionListProps) => (
+  <dl
+    {...props}
```
tlabaj, line 68: Can you spread the props at the end.

packages/react-core/src/components/DescriptionList/DescriptionListText.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
```
tlabaj, line 1: I actually don't think we need this component.  I think it can be rendered internally within the `DecriptionListTerm` and `DescriptionListDescription`. 
@mattnolting  will Should everything rendered in the `<dt>` and the `<dd>` be wrapped in the `.pf-c-description-list__text` `<div>`?

packages/react-core/src/components/DescriptionList/DescriptionListText.tsx
```diff
@@ -0,0 +1,32 @@
...
+import styles from '@patternfly/react-styles/css/components/DescriptionList/description-list';
+import { css } from '@patternfly/react-styles';
+
+export interface DescriptionListTextProps extends React.HTMLProps<HTMLDivElement> {
+  /** Anything that can be rendered inside of list text */
+  children: React.ReactNode;
```
tlabaj, line 7: I think we just need the children prop here.  Then the consumer can place whatever the want in the `<dt>`.  I don't see any related classes to any sub elements  of the `<dt>` in core.

packages/react-core/src/components/DescriptionList/DescriptionListText.tsx
```diff
@@ -0,0 +1,32 @@
...
+  children = null,
+  component = 'div',
+  href,
+  icon,
+  ...props
+}: DescriptionListTextProps) => {
+  const Component: any = component;
+
+  return (
+    <Component href={href} className={'pf-c-description-list__text'} {...props}>
```
tlabaj, line 26: see comment above.

packages/react-integration/cypress/integration/descriptionlist.spec.ts
```diff
@@ -0,0 +1,23 @@
+describe('Description List Demo Test', () => {
```
tlabaj, line 1: make sure to also verify the modifiers get applied appropriately and that the column modifiers get applied at all breakpoints.

packages/react-integration/demo-app-ts/src/components/demos/DescriptionListDemo/DescriptionListDemo.tsx
```diff
@@ -0,0 +1,71 @@
...
+import React, { Component } from 'react';
+
+export class DescriptionListDemo extends Component {
+  componentDidMount() {
+    window.scrollTo(0, 0);
+  }
+
+  render() {
+    return (
+      <DescriptionList>
```
tlabaj, line 19: Make sure all props are used at least once in the demos to verify typing.

packages/react-core/src/components/DescriptionList/DescriptionListText.tsx
mattnolting: cc @mcoker we added the `__text` element for future requests, like truncation, wrapping etc. I'd expect it to be a `span` though. @mcoker wdyt?

mcoker: @mattnolting yep, agree

mcoker: @mattnolting @tlabaj should definitely be a span in the `<dt>` so it can be displayed inline with things around it and those thing wrap like content in form labels. The basic layout will be similar to a form.

Screen Shot 2020-07-21 at 2 28 58 PM


@tlabaj yep I would expect the children of the `dt` and `dd` to go in the `__text` element. We may enable some prop on them to create elements we will put adjacent to the `__text` element - for example in the `dt`, a prop that will generate a help icon with tooltip/popover text.

maryshak1996:
@mcarrano @jenny-s51 I think that this is looking good! The only thing that I think that we should update would be the default spacing in between columns from 16px to 24. The tightness of the columns makes it a little difficult to distinguish between where one description pair starts and another ends (screenshot of what I'm referring to below)
Screen Shot 2020-07-27 at 12 01 51 PM

packages/react-core/src/components/DescriptionList/DescriptionListTerm.tsx
```diff
@@ -0,0 +1,21 @@
...
+  className?: string;
+}
+
+export const DescriptionListTerm: React.FunctionComponent<DescriptionListTermProps> = ({
+  children = null,
+  className,
+  ...props
+}: DescriptionListTermProps) => (
+  <dt className={css(styles.descriptionListTerm, className)} {...props}>
+    <span>{children}</span>
```
tlabaj, line 18: You are missing the `'pf-c-description-list__text'` class here.
@mcoker @mnolting is it is core being updated so that the the child of `<dt>` is a span?

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
```diff
@@ -0,0 +1,36 @@
...
+  textComponent = 'span',
+  className,
+  href,
+  icon,
+  ...props
+}: DescriptionListDescriptionProps) => {
+  const DescriptionListText: any = textComponent;
+  return (
+    <dd className={css(styles.descriptionListDescription, className)} {...props}>
+      <DescriptionListText href={href} className={'pf-c-description-list__text'}>
```
tlabaj, line 29: Why are yo using the class name directly here instead of `styles`?
Why do we have the `href` here?  I don't thins the 'pf-c-description-list__text' is intended to be an `<a>`
@mcoker @mnolting.  Is the child of the `dd` supposed to be configurable or always a div (like in core)?

packages/react-integration/cypress/integration/descriptionlist.spec.ts
tlabaj: This still is missing.

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export interface DescriptionListDescriptionProps extends React.HTMLProps<HTMLElement> {
+  /** Anything that can be rendered inside of list description */
+  children: React.ReactNode;
+  /** Additional classes added to the DescriptionListDescription */
+  className?: string;
+  /** Indicates which component will be used for the text wrapped by the description */
+  textComponent?: React.ReactNode;
+  /** Default hyperlink location */
+  href?: string;
```
tlabaj, line 13: this is not needed.  the anchor would be. A child of the 'pf-c-description-list__text' `div` like it is in core.

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
```diff
@@ -0,0 +1,36 @@
...
+import styles from '@patternfly/react-styles/css/components/DescriptionList/description-list';
+import { css } from '@patternfly/react-styles';
+
+export interface DescriptionListDescriptionProps extends React.HTMLProps<HTMLElement> {
+  /** Anything that can be rendered inside of list description */
+  children: React.ReactNode;
+  /** Additional classes added to the DescriptionListDescription */
+  className?: string;
+  /** Indicates which component will be used for the text wrapped by the description */
+  textComponent?: React.ReactNode;
```
tlabaj, line 11: I don't think we need this.  See comment below. (line 29)

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
```diff
@@ -0,0 +1,36 @@
...
+  /** Default hyperlink location */
+  href?: string;
+  /** Icon added to the left of the label text. */
+  icon?: React.ReactNode;
```
tlabaj, line 15: You don't need this either.  The consumer will take care of adding the icon iff they choose.

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -0,0 +1,511 @@
...
+
+SimpleDescriptionList = (
+  <DescriptionList>
+    <DescriptionListGroup>
+      <DescriptionListTerm>Name</DescriptionListTerm>
+      <DescriptionListDescription>Example</DescriptionListDescription>
+    </DescriptionListGroup>
+    <DescriptionListGroup>
+      <DescriptionListTerm>Namespace</DescriptionListTerm>
+      <DescriptionListDescription textComponent="a" href="#">
```
tlabaj, line 39: this will be updated once you update the component.

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -0,0 +1,511 @@
...
+      </DescriptionListDescription>
+    </DescriptionListGroup>
+    <DescriptionListGroup>
+      <DescriptionListTerm>Labels</DescriptionListTerm>
+      <DescriptionListDescription>example</DescriptionListDescription>
+    </DescriptionListGroup>
+    <DescriptionListGroup>
+      <DescriptionListTerm>Pod selector</DescriptionListTerm>
+      <DescriptionListDescription>
+        <Button variant="link" isInline icon={<PlusCircleIcon />}>
```
tlabaj, line 50: This will also we updated once you update the component.

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
jenny-s51: I use the class name there because `styles.descriptionListText` doesn't exist -- there is no `.pf-c-description-list__text class or corresponding styles in https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/DescriptionList/description-list.scss.

You're right; looks like the <a> tag should be the child of a div -- will fix.

packages/react-integration/cypress/integration/descriptionlist.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  it('Verify text', () => {
+    cy.get('.pf-c-description-list__term').contains('Name');
+  });
+
+  it('Verify description', () => {
+    cy.get('.pf-c-description-list__description').contains('Example');
+  });
+
+  it('Verify 2 Col Description List', () => {
```
tlabaj, line 22: You should verify every breakpoint.  Take a look at the Tabs component.  We have had issues in the past where the modifier was not being applied as expected.

packages/react-integration/demo-app-ts/src/components/demos/DescriptionListDemo/DescriptionListDemo.tsx
```diff
@@ -0,0 +1,248 @@
...
+    return (
+      <StackItem isFilled>
+        <br />
+        <Title headingLevel="h2" size="2xl">
+          Two Column Description List
+        </Title>
+        <Divider component="div" />
+        <br />
+        <div className="example">
+          <DescriptionList className="2-col-description-list" columnModifier={{ default: '2Col' }}>
```
tlabaj, line 72: why not us an `id` here instead of a className to uniquely identify the element?

packages/react-core/src/components/DescriptionList/DescriptionListTerm.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export interface DescriptionListTermProps extends React.HTMLProps<HTMLElement> {
+  /** Anything that can be rendered inside of list term */
+  children: React.ReactNode;
+  /** Additional classes added to the DescriptionListTerm */
+  className?: string;
+}
+
+export const DescriptionListTerm: React.FunctionComponent<DescriptionListTermProps> = ({
+  children = null,
```
tlabaj, line 13: why do we need to default to `null` here?

packages/react-integration/cypress/integration/descriptionlist.spec.ts
jenny-s51: Sounds good, thank you for referencing the `Tabs` demo -- I can write a separate demo file to test every breakpoint.

packages/react-core/src/components/DescriptionList/DescriptionListTerm.tsx
jenny-s51: I don't think we do - will fix

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
```diff
@@ -0,0 +1,21 @@
...
+  className?: string;
+}
+
+export const DescriptionListDescription: React.FunctionComponent<DescriptionListDescriptionProps> = ({
+  children = null,
+  className,
+  ...props
+}: DescriptionListDescriptionProps) => (
+  <dd className={css(styles.descriptionListDescription, className)} {...props}>
+    <div className={'pf-c-description-list__text'}>{children}</div>
```
jschuler, line 18: is this class defined on `styles.descriptionListText`?

packages/react-core/src/components/DescriptionList/DescriptionListTerm.tsx
```diff
@@ -0,0 +1,21 @@
...
+  className?: string;
+}
+
+export const DescriptionListTerm: React.FunctionComponent<DescriptionListTermProps> = ({
+  children,
+  className,
+  ...props
+}: DescriptionListTermProps) => (
+  <dt className={css(styles.descriptionListTerm, className)} {...props}>
+    <span className={'pf-c-description-list__text'}>{children}</span>
```
jschuler, line 18: same comment

packages/react-core/src/components/DescriptionList/DescriptionListDescription.tsx
jenny-s51: No it isn't @jschuler - there is no `.pf-c-description-list__text` class or corresponding styles in https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/DescriptionList/description-list.scss, but this class is still applied in the core examples for some reason

PR #4587, opened 2020-07-20
feat(Drawer): enable users to define actions when they click outside the drawer panel

### LINKED ISSUE #4574: Drawer should be able to close when clicking outside of the drawer panel
In Drawer component, when user opens the drawer and click outside the drawer, we hope the drawer will close automatically if it is not inline. However, we can only use toggle button or close icon to close it now.
Enhancement.

DaoDaoNoCode:
**What**: Closes #4574 
To solve this problem, I forward the ref of `div` node in DrawerPanelContent for user to use. User then could use the ref to create some event listeners to detect whether they click outside the drawer panel or not and customize the actions as well.
Thus, I change the usage of the first example Basic Drawer in the docs. When user click outside of the drawer panel, the drawer will close automatically.

**Additional issues**: Now it's a little bit too complex for users to use the ref and create the auto-close event. Is there a better way to solve this problem? Can we change the code inside the component so that users could easily use it by passing a function like `onClickOutside` as a prop and don't need to create those event listeners by themselves?

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -18,13 +18,16 @@ export interface DrawerPanelContentProps extends React.HTMLProps<HTMLDivElement>
     xl?: 'width_25' | 'width_33' | 'width_50' | 'width_66' | 'width_75' | 'width_100';
     '2xl'?: 'width_25' | 'width_33' | 'width_50' | 'width_66' | 'width_75' | 'width_100';
   };
+  /** Forward the ref outside the component */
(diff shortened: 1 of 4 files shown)


DaoDaoNoCode:
I added a new example in Drawer.md to show the usage of the `ref` in DrawerPanelContent. I also reformat the code in the doc.

tlabaj:
Let's not merge this one.  There is still some discussion happening concerning the behaviour.  See the associated issue for more details.

evwilkin:
Removed milestone from this PR, added it to original issue #4574 - it was added her by mistake

tlabaj:
Closing this since the issue was closed. Please see the conversation in the issue for more details.

PR #4588, opened 2020-07-20
feat(SearchInput): add new SearchInput component

### LINKED ISSUE #4521: Add search input component
New component! https://github.com/patternfly/patternfly/pull/3264

The design/interaction docs are in the original issue - https://github.com/patternfly/patternfly/issues/3151#issuecomment-637807651 Please consult with @maryshaks1996 on any design questions.

nicolethoen:
Closes #4521

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -13,8 +13,9 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   className?: string;
   /** Value of the search input */
   value?: string;
-  /** The count of returned search results */
-  results?: string;
+  /** The number of search results returned. Either a total number of results,
+   * or a string representing the current result over the total number of results. i.e. "1 / 5" */
+  numberOfResults?: number | string;
   /** An accessible label for the search input */
   'aria-label'?: string;
   /** placeholder text of the search input */
@@ -35,7 +36,7 @@ export const SearchInput: React.FunctionComponent<SearchInputProps> = ({
   placeholder,
   onChange,
   onClear,
-  results,
+  numberOfResults,
   onNextClick,
   onPreviousClick,
   'aria-label': ariaLabel = 'Search input',
@@ -63,9 +64,9 @@ export const SearchInput: React.FunctionComponent<SearchInputProps> = ({
       </span>
       {value && (
         <span className={css(styles.searchInputUtilities)}>
-          {results && (
+          {numberOfResults && (
             <span className={css(styles.searchInputCount)}>
-              <Badge isRead>{results}</Badge>
+              <Badge isRead>{numberOfResults}</Badge>
             </span>
           )}
           {!!onNextClick && !!onPreviousClick && (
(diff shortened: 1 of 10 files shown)


packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+import TimesIcon from '@patternfly/react-icons/dist/js/icons/times-icon';
+import SearchIcon from '@patternfly/react-icons/dist/js/icons/search-icon';
+
+export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange' | 'results'> {
+  /** Additional classes added to the banner */
+  className?: string;
+  /** Value of the search input */
+  value?: string;
+  /** The count of returned search results */
+  results?: string;
```
tlabaj, line 17: hmmm, I wonder if the name is misleading. To me it does not imply it is a count.

nicolethoen:
> I think that the 3rd example is a little confusing without context to show the whole interaction (like the document you are searching).

@tlabaj but adding more context, wouldn't that break our typical documentation and example practices?
Could I leave this example out or move it to a demo instead?

@mcarrano

mcarrano:
We could add a demo in the future, but I'd still be in favor of leaving the example in. This is one of the reasons it would be helpful to have example descriptions. Maybe if we changed the name of the example to something like, "Find input with multiple matches?" That would better describe the use case this is intended for. @nicolethoen @tlabaj what do you think?

nicolethoen:
> We could add a demo in the future, but I'd still be in favor of leaving the example in. This is one of the reasons it would be helpful to have example descriptions. Maybe if we changed the name of the example to something like, "Find input with multiple matches?" That would better describe the use case this is intended for. @nicolethoen @tlabaj what do you think?

I used the same example name that was used in Core. That doesn't necessarily make it the best name, but the component name is called Search Input and it already has an example with multiple matches. The example up for debate the the example with navigable options. I think navigable is the key word...

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+import SearchIcon from '@patternfly/react-icons/dist/js/icons/search-icon';
+
+export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange' | 'results'> {
+  /** Additional classes added to the banner */
+  className?: string;
+  /** Value of the search input */
+  value?: string;
+  /** The number of search results returned. Either a total number of results,
+   * or a string representing the current result over the total number of results. i.e. "1 / 5" */
+  numberOfResults?: number | string;
```
dlabrecq, line 18: The `numberOfResults` name implies number is the type. Perhaps this be something more like "resultsCount", "resultsTotal", or "resultsText"?

tlabaj:
> Looks great! Only thing I see is that when you reach the beginning or end of the navigable result list, the appropriate nav button should be disabled.

@mcoker that functionality will be defined my the consumer.  They very well  could have the button enabled and when clicked, search from the beginning.   In that case the results string should be updated to `1/3`. 
So, my question here is which interaction do we want to show @mcarrano .

mcarrano:
> @mcoker that functionality will be defined my the consumer. They very well could have the button enabled and when clicked, search from the beginning. In that case the results string should be updated to 1/3.
So, my question here is which interaction do we want to show @mcarrano .

Not sure exactly what you mean here @tlabaj . Are you saying that clicking "^" if you are already on the first item would re-invoke the search? I think that disabling the first and last buttons would be the right thing to do in most cases, but don't feel strongly about this.

mcarrano:
@jessiehuff are you good with the keyboard interaction on this? I wasn't sure what to expect.

PR #4596, opened 2020-07-21
feat(Button): add cta/large modifier

### LINKED ISSUE #4527: Button - add CTA/large variation
https://github.com/patternfly/patternfly/pull/3214

variation is `.pf-m-display-lg` and applies to
* .pf-m-primary,
* .pf-m-secondary,
* .pf-m-tertiary,
* .pf-m-link

kmcfaul:
**What**: Closes #4527

### DIFF
--- a/packages/react-core/src/components/Button/Button.tsx
+++ b/packages/react-core/src/components/Button/Button.tsx
@@ -52,6 +52,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement>, OUIAPro
   tabIndex?: number;
   /** Adds small styling to the button */
   isSmall?: boolean;
+  /** Adds large styling to the button */
+  isLarge?: boolean;
 }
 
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/Button/Button.tsx
```diff
@@ -52,6 +52,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement>, OUIAPro
   tabIndex?: number;
   /** Adds small styling to the button */
   isSmall?: boolean;
+  /** Adds large styling to the button */
+  isLarge?: boolean;
```
tlabaj, line 56: In retrospect we probably should have combined the isLarge and isSmall.

kmcfaul: Maybe in the next breaking change release we can refactor them into a `size` property.

PR #4610, opened 2020-07-23
feat(Tile): add tile component

### LINKED ISSUE #4524: Add tile component
New component! https://github.com/patternfly/patternfly/pull/3238

Some interaction details are in the design in the original issue - https://github.com/patternfly/patternfly/issues/3100

* Only title is required
* Component is a `div` with `tabindex="0"` so it's focusable
* It can be:
  * selected via `.pf-m-selected`
  * disabled via `.pf-m-disabled`
  * can have a stacked layout via `.pf-m-stacked` on `.pf-c-tile__header`
  * set to a larger variation via `.pf-m-display-lg` (if the header is stacked)

cc @christiemolloy

kmcfaul:
**What**: Closes #4524

### DIFF
--- a/packages/react-core/src/components/Tile/Tile.tsx
+++ b/packages/react-core/src/components/Tile/Tile.tsx
@@ -1,0 +1,53 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Tile/tile';
+import { css } from '@patternfly/react-styles';
+
+export interface TileProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the banner */
+  children?: React.ReactNode;
+  /** Additional classes added to the banner */
+  className?: string;
+  /** Title of the tile */
+  title: string;
+  /** Icon in the tile title */
+  icon?: React.ReactNode;
+  /** Flag indicating if the tile is selected */
+  isSelected?: boolean;
+  /** Flag indicating if the tile is disabled */
+  isDisabled?: boolean;
+  /** Flag indicating if the tile icon is stacked */
+  isStacked?: boolean;
+  /** Flag indicating if the stacked tile icon is large */
+  isLargeIcon?: boolean;
+}
+
+export const Tile: React.FunctionComponent<TileProps> = ({
+  children,
+  title,
+  icon,
+  isStacked,
+  isSelected,
+  isDisabled,
+  isLargeIcon,
+  className,
+  ...props
+}: TileProps) => (
+  <div
+    className={css(
+      styles.tile,
+      isSelected && styles.modifiers.selected,
+      isDisabled && styles.modifiers.disabled,
+      isLargeIcon && styles.modifiers.displayLg,
+      className
+    )}
+    tabIndex={0}
+    {...props}
+  >
+    <div className={css(styles.tileHeader, isStacked && styles.modifiers.stacked)}>
+      {icon && <div className={css(styles.tileIcon)}>{icon}</div>}
+      <div className={css(styles.tileTitle)}>{title}</div>
+    </div>
+    {children && <div className={css(styles.tileBody)}>{children}</div>}
+  </div>
+);
+Tile.displayName = 'Tile';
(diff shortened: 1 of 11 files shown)


packages/react-core/src/components/Tile/examples/Tile.md
```diff
@@ -0,0 +1,110 @@
+---
```
tlabaj, line 1: It looks like the "extra content" example is missing.

kmcfaul:
@mcarrano @jessiehuff 
Re: @dlabrecq comment

I can add an `onKeyDown` handler to the tile, but which keys to we want to allow for selection? `Enter` only or both enter and `Space`?

packages/react-core/src/components/Tile/Tile.tsx
```diff
@@ -0,0 +1,53 @@
...
+  className?: string;
+  /** Title of the tile */
+  title: string;
+  /** Icon in the tile title */
+  icon?: React.ReactNode;
+  /** Flag indicating if the tile is selected */
+  isSelected?: boolean;
+  /** Flag indicating if the tile is disabled */
+  isDisabled?: boolean;
+  /** Flag indicating if the tile icon is stacked */
```
mcoker, line 18: Just a nit, but this might say "tile header is stacked" since that's the layout variation - though it does stack the icon, so that works, too :)

packages/react-core/src/components/Tile/Tile.tsx
```diff
@@ -0,0 +1,53 @@
...
+  isStacked?: boolean;
+  /** Flag indicating if the stacked tile icon is large */
+  isLargeIcon?: boolean;
```
mcoker, line 21: I would rename this to `isDisplayLarge`. This variation could introduce other styles that don't apply to the icon. Also this is a new "display" concept we've added for more compact (application) vs larger (website/marketing) designs, which is why it would be `isDisplayLarge` instead of something like `isLarge`, which could do something different, like change the width or height or something like that and be different in concept to what `isDisplayLarge` does.

mcoker: Also that matches the core class, so we will avoid name conflicts in the future since these match.

mcoker:
Looking at the extra content examples, the core examples limit the tile width, and place them in a flex layout, so the cards display side by side and are equal heights:

Screen Shot 2020-07-29 at 7 39 56 PM

Here are those examples in this PR:

Screen Shot 2020-07-29 at 7 31 41 PM

@christiemolloy I'm assuming that example in the core docs shows how cards of different heights (with differing amounts of content) should have the same height when placed in a row? If so I'm wondering if we should use a layout for that instead of custom workspace styles, and if that belongs in a demo instead? WDYT?

christiemolloy:
@mcoker thats a good point, without the workstyle style of `display: flex` the cards wont all grow to the same height. I think a demo with a flex layout could definitely solve this though

mcarrano:
@kmcfaul I'm going to pull together a quick meeting to discuss what the intended behavior should be. Tiles can be used like a checkbox or radio button group. Not sure that the examples currently illustrate that.

kmcfaul:
@mcoker @christiemolloy 
Updated `isStacked` property description.
Renamed `isLargeIcon` to `isDisplayLarge`.
Updated `Tile with extra content` to use `Flex`.

@mcarrano 
I'm not sure the Tile component examples would show this behavior, it sounds more like a demo.

kmcfaul:
@mcarrano @jessiehuff @maryshak1996

Added two tile demos to the demos section: single select allowing `space` as a selection key, and multiple select allowing `enter` as a selection key.

There is a bit of weirdness with the mouse, because the event can pick up the title div as a target instead of the outer div where the `id` is present. I'm still thinking this one through, but Tile may have to pipe down the onClick to the title div (but that would lock in using `id` or another data property for comparison).

Another point of confusion is that the styles for the focus and selected states are very similar, so it can be hard to tell what is selected versus what is focused. @mcoker

kmcfaul:
@redallen Any idea what the build error is pointing to? Docs builds fine locally.

kmcfaul:
@jessiehuff I added a note and link to the demo section in the Tile.md, let me know if that looks okay. Eventually, the demo section will live in a tab alongside the rest of the component docs, at which point it may no longer be necessary.

Issue #4615, opened 2020-07-27
Maximum call stack size exceeded in a read-only typeahead Select field when pressing enter


jerolimov:
**Describe the issue. What is the expected and unexpected behavior?**

When creating a typeahead Select field which do not support new items (no isCreatable prop) the components crashs when the user types an entry which does not exist and press enter the application crashs with an "Maximum call stack size exceeded" error.

* If you have the values aaaaa, bbbbb, .... Entering aaa and press enter works and selects the entry aaaaa.
* But when you enter eee, no entry was shown. If you then press enter the error above was thrown.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
Codesandbox to reproduce the error: https://codesandbox.io/s/elegant-hofstadter-omfke?file=/index.js

```jsx
import ReactDOM from "react-dom";
import "@patternfly/react-core/dist/styles/base.css";

import * as React from "react";
import { Select, SelectVariant, SelectOption } from "@patternfly/react-core";

const YourIssue = () => {
  const [isOpen, setIsOpen] = React.useState(false);
  const [options] = React.useState(["aaaaa", "bbbbb", "ccccc"]);
  const [value, setValue] = React.useState("aaaaa");

  const onToggle = () => setIsOpen(!isOpen);
  const onSelect = (_, selection) => {
    setValue(selection);
    setIsOpen(false);
  };
  const onClear = () => setValue(null);
  const onCreateOption = () => setValue(null);

  return (
    <div style={{ padding: "40px" }}>
      <Select
        variant={SelectVariant.typeahead}
        placeholderText="Placeholder"
        selections={value}
        isOpen={isOpen}
        isCreatable={false} // Works with true!
        onCreateOption={onCreateOption}
        onToggle={onToggle}
        onSelect={onSelect}
        onClear={onClear}
      >
        {options.map(option => (
          <SelectOption key={option} value={option}>
            {option}
          </SelectOption>
        ))}
      </Select>
    </div>
  );
};

ReactDOM.render(<YourIssue />, document.getElementById("root"));
```

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug
No workaround other than enable isCreatable

**What is your product and what release version are you targeting?**
OpenShift

jerolimov:
PR: https://github.com/patternfly/patternfly-react/pull/4616

tiagojufr:
This issue still exists in a different scenario:
[...]

PR #4616, opened 2020-07-27
Fix Maximum call stack size exceeded in a read-only typeahead Select field when pressing enter

jerolimov:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4615

**Additional issues**:

The first commit fixes the issue https://github.com/patternfly/patternfly-react/issues/4615 which is reproducible on CodeSandbox: https://codesandbox.io/s/elegant-hofstadter-omfke?file=/index.js

After fixing this I noticed that the preventDefault method was not called for an "enter" key event. Which results in the issue that the Form was automatically submitted. When the Select is open and the user presses enter this should close the select field instead of submitting the form. I added a second commit for this.

I'm not sure if this is your desired way to solve this issue. I'm open for your feedback on this fix.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -161,7 +161,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       this.refCollection[0] = this.filterRef.current;
     }
 
-    if (!prevState.openedOnEnter && this.state.openedOnEnter && !this.props.customContent) {
+    if (!prevState.openedOnEnter && this.state.openedOnEnter && !this.props.customContent && this.refCollection[0]) {
       this.refCollection[0].focus();
     }
 
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -289,15 +289,17 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
     const { isOpen, isCreatable, createText } = this.props;
     const { typeaheadActiveChild, typeaheadCurrIndex } = this.state;
     if (isOpen) {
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/Select/Select.tsx
```diff
@@ -289,15 +289,17 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
...
+        if (typeaheadActiveChild || this.refCollection[0]) {
+          this.setState({
+            typeaheadInputValue:
+              (typeaheadActiveChild && typeaheadActiveChild.innerText) || this.refCollection[0].innerText
+          });
+          if (typeaheadActiveChild) {
+            typeaheadActiveChild.click();
+          } else {
+            this.refCollection[0].click();
+          }
```
dlabrecq, line 302: It's not clear how this fixes the endless loop? Other than splitting the `if` statement to a new line, the structure looks the same to me.

jerolimov: Because `this.refCollection[0]` is null in this case the code goes into the else case. I expected that this is not the expected behavior, but as said, I'm not 100% sure here.

The endless loop then happen when `getNextIndex` was called:

https://github.com/patternfly/patternfly-react/blob/b1addef8bca3ad4c4bb851aebb4b5782c45e4ce1/packages/react-core/src/components/Select/Select.tsx#L309

getNextIndex calls getNextIndex again and again:

https://github.com/patternfly/patternfly-react/blob/b1addef8bca3ad4c4bb851aebb4b5782c45e4ce1/packages/react-core/src/helpers/util.ts#L215

dlabrecq: Thanks for the explanation

Issue #4592, opened 2020-07-21
Error when SelectGroup contains only one SelectOption


idevat:
**Describe the issue. What is the expected and unexpected behavior?**
Error when `SelectGroup` contains only one `SelectOption`

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
In page https://www.patternfly.org/v4/documentation/react/components/select
1. change `JSX` code for `Grouped single` to contain only one `SelectOption`
2. click on the rendered result above
3. the rendered select disappears and an error appears - see below:
![single-select-option](https://user-images.githubusercontent.com/16188254/88060804-fa9e4300-cb66-11ea-9132-69bdfb80f755.png)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug.

**What is your product and what release version are you targeting?**

KKoukiou:
Closes by #4622

PR #4622, opened 2020-07-28
feat(Select): allow typeahead groups, fix few bugs around groups

### LINKED ISSUE #4590: Select produces errors and warning
Component `Select` produces 
* a warning when it is opened 
* error when filter is set to a value that makes list of options empty 

When I prepare component with `Select` (I need items in groups and I also need filtering) like this...:
```typescript
import React from "react";

import {
  Select,
  SelectGroup,
  SelectOption,
  SelectOptionObject,
  SelectVariant,
} from "@patternfly/react-core";

type OptionsSource = Record<string, string[]>;

const sourceMap: OptionsSource = {
  group1: ["one"],
  group2: ["two", "three"],
};

const filterSources = (
  search: string,
  sources: OptionsSource,
): OptionsSource => {
  if (search === "") {
    return sources;
  }
  return Object.keys(sources).reduce((filteredSources, group) => {
    const items = sources[group].filter(i => i.includes(search));
    return items.length > 0
      ? { ...filteredSources, [group]: items }
      : filteredSources;
  }, {});
};

const renderOptions = (sources: OptionsSource) => {
  return Object.keys(sources).map(group => (
    <SelectGroup label={group} key={group}>
      {sources[group].map(item => (
        <SelectOption key={`${group}:${item}`} value={item} />
      ))}
    </SelectGroup>
  ));
};

export const Scratch = () => {
  const [isOpen, setIsOpen] = React.useState(false);
  const [selected, setSelected] = React.useState<string | SelectOptionObject>();

  const onSelect = (
    event: React.MouseEvent | React.ChangeEvent,
    value: string | SelectOptionObject,
  ) => {
    setSelected(value);
    setIsOpen(false);
  };

  const onClear = () => {
    setSelected(undefined);
    setIsOpen(false);
  };

  const onFilter = (event: React.ChangeEvent<HTMLInputElement>) => {
    const options = renderOptions(filterSources(event.target.value, sourceMap));
    console.log(event.target.value, options);
    return options;
  };

  return (
    <Select
      isGrouped
      onToggle={(isExpanded: boolean) => setIsOpen(isExpanded)}
      onSelect={onSelect}
      onClear={onClear}
      selections={selected}
      isOpen={isOpen}
      onFilter={onFilter}
    >
      {renderOptions(sourceMap)}
    </Select>
  );
};
```
...I get following misbehavior:

1. When I open the select I get 
```
Warning: React does not recognize the `isFocused` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isfocused` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```
warning-after-Select-open

2. When I write characters to filter options, it works when there are some matching items. When there is no matching item (see what `console.log` produces in the picture) it crashes with an error:
```
Uncaught TypeError: Cannot read property 'map' of undefined
```
error-after-filtering

It looks like a bug (but it can be misuse of component). Yes it blocks me. I don't know about any work-around.
I use `"@patternfly/react-core": "^4.32.1"` but I got this behavior also in older version `"@patternfly/react-core": "^4.23.1"`

My product is web interface for HA Cluster; next version.

kmcfaul:
**What**: Closes #4590, #4592 + a couple related bugs found while addressing the tickets

In summary:
- Allows typeahead variants to properly use `isGrouped` and `SelectGroup`. (#4590)
- Allows mixed grouped and ungrouped SelectOptions when using `isGrouped`, as typeahead variants may add ungrouped options (such as no results, and createable options).
- Fixes bug in keyboard navigation of grouped options, where the ref collection wasn't indexed properly.
- Fixes bug in keyboard navigation of typeahead options, where disabled options weren't skipped properly.
- Fixes bug where an error occurred when a SelectGroup had only one SelectOption and wasn't defined as an array. (#4592)
- Adds handling of an edge case in typeahead keyboard navigation, where a typeahead select's options could solely be a disabled option (in the case of no results).
- Adds an example of a grouped typeahead select.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -7,6 +7,7 @@ import { css } from '@patternfly/react-styles';
 import TimesCircleIcon from '@patternfly/react-icons/dist/js/icons/times-circle-icon';
 import { SelectMenu } from './SelectMenu';
 import { SelectOption, SelectOptionObject } from './SelectOption';
+import { SelectGroup } from './SelectGroup';
 import { SelectToggle } from './SelectToggle';
 import { SelectContext, SelectVariant, SelectDirection, KeyTypes } from './selectConstants';
 import { Chip, ChipGroup } from '../ChipGroup';
@@ -193,7 +194,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
   };
 
   onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const { onFilter, isCreatable, onCreateOption, createText, noResultsFoundText, children } = this.props;
+    const { onFilter, isCreatable, onCreateOption, createText, noResultsFoundText, children, isGrouped } = this.props;
     let typeaheadFilteredChildren: any;
     if (onFilter) {
       typeaheadFilteredChildren = onFilter(e) || children;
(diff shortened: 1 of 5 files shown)


kmcfaul:
The biggest thing to test is the typeahead grouped example, and then just that the keyboard navigation works.

PR #4626, opened 2020-07-29
add tooltip css back for backwards compatibility

jschuler:
**What**: Closes https://github.com/patternfly/training-scenarios/issues/307

Should hopefully fix the above issue. I think what's happening is that even though the katacoda environment wasn't updated recently and uses `"@patternfly/react-core": "^4.18.5"` in the package lock file it is pulling in 
`"@patternfly/react-styles": "^4.3.4",` which probably resolves to 4.5.0 which has those files missing.


redallen:
Even if we merge this change, katacoda will have to install the new package, which we don't seem to be able to control anyway.

Can we configure katacoda's package.json file to use the latest `@patternfly` packages?

jschuler:
@redallen looks like the package lock file messes things up for katacoda, should be regenerated. Investigating more...

jschuler:
Closing in favor of https://github.com/patternfly/training-scenarios/pull/308

Issue #4630, opened 2020-07-29
Bump Victory packages


dlabrecq:
Update the Victory packages in order to unblock https://github.com/patternfly/patternfly-react/pull/4435 -- an effort related to "auditing and Standardizing all OUIA components for WT integration".

This package bump will allow us to provide our own OUIA props to `VictoryContainer`. See `https://github.com/FormidableLabs/victory/pull/1654`.

PR #4435, opened 2020-06-23
OUIA ChartDonut

psav:
**What**: Closes #4400

PR #4633, opened 2020-07-30
Update Victory packages to support OUIA props

dlabrecq:
Updated the Victory packages in order to unblock #4435 -- an effort related to "auditing and Standardizing all OUIA components for WT integration".

This package bump will allow us to provide our own OUIA props to `VictoryContainer`. See https://github.com/FormidableLabs/victory/pull/1654.

Fixes https://github.com/patternfly/patternfly-react/issues/4630


redallen:
I verified this change does not break any known products with @dlabrecq .

PR #4640, opened 2020-07-31
feat(NotificationDrawer): Add notification drawer demo

### LINKED ISSUE #4535: Notification drawer demo
Follow on to https://github.com/patternfly/patternfly/issues/3186 for adding a notification drawer demo in React. Should demonstrate behavior of the drawer integrated into a page including the notification badge with latest updates (#4323 

cc @mcoker @tlabaj

nicolethoen:
**What**: Closes #4535 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawerHeader.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawerHeader.tsx
@@ -30,7 +30,9 @@ export const NotificationDrawerHeader: React.FunctionComponent<NotificationDrawe
     <Text component={TextVariants.h1} className={css(styles.notificationDrawerHeaderTitle)}>
       {title}
     </Text>
-    {count && <span className={css(styles.notificationDrawerHeaderStatus)}>{`${count} ${unreadText}`}</span>}
+    {count !== undefined && (
+      <span className={css(styles.notificationDrawerHeaderStatus)}>{`${count} ${unreadText}`}</span>
+    )}
     {children && <div className={css(styles.notificationDrawerHeaderAction)}>{children}</div>}
   </div>
 );
--- a/packages/react-catalog-view-extension/package.json
+++ b/packages/react-catalog-view-extension/package.json
@@ -35,7 +35,7 @@
     "clean": "rimraf dist"
   },
   "dependencies": {
-    "@patternfly/patternfly": "4.31.2",
+    "@patternfly/patternfly": "4.31.3",
     "@patternfly/react-core": "^4.38.1",
     "@patternfly/react-styles": "^4.6.1",
     "classnames": "^2.2.5",
--- a/packages/react-charts/package.json
+++ b/packages/react-charts/package.json
@@ -29,7 +29,7 @@
   },
   "homepage": "https://github.com/patternfly/patternfly-react#readme",
   "dependencies": {
-    "@patternfly/patternfly": "4.31.2",
+    "@patternfly/patternfly": "4.31.3",
     "@patternfly/react-styles": "^4.6.1",
     "@patternfly/react-tokens": "^4.8.1",
     "hoist-non-react-statics": "^3.3.0",
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -44,7 +44,7 @@
     "tslib": "^1.11.1"
   },
   "devDependencies": {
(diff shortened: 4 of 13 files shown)


mcoker:
@nicolethoen this looks great! A few things:

* The parent `.pf-c-page__header-tools-item` element with the notification bell icon should have `.pf-m-selected` when the drawer is open.
* The notification bell disappears on a mobile viewport. It should be visible on all viewports
* The grouped notification drawer is showing some old styles - namely that groups shouldn't scroll independently of the overall notification list. You can compare this demo to the [core grouped demo](https://pf4.patternfly.org/documentation/core/demos/notificationdrawer/expanded-groups). I assume that will resolve with a core bump?

nicolethoen:
> This looks great @nicolethoen . I have the same comment about scrolling behavior as @mcoker . The only other thing is that this still uses the old notification badge component. Any reason we wouldn't pull the new component into this demo?

@mcarrano 
the new react notification badge has not been merged yet. It should update automatically.

Once merged, is there any reason one of these demos should use the 'attention state' notification badge?

mcarrano:
@nicolethoen regarding the attention state, yes, it would make sense to have a demo of that. However I'm not sure that we totally defined all the rules for when this appear/disappears so maybe pick that up in a follow-on issue?

nicolethoen:
> @nicolethoen regarding the attention state, yes, it would make sense to have a demo of that. However I'm not sure that we totally defined all the rules for when this appear/disappears so maybe pick that up in a follow-on issue?

@mcarrano 
should I just use the attention state in one of the existing two examples on this PR or make a whole new example?

mcarrano:
@mceledonia Can you take a look at the behavior of the notification badge and verify that it is correct. Should the 'unread' state only display when the drawer is closed? Other than that, I think this is good to go @nicolethoen . It might be nice to show an example of the notification count in the badge, we we could always add that later.

nicolethoen:
> @mceledonia Can you take a look at the behavior of the notification badge and verify that it is correct. Should the 'unread' state only display when the drawer is closed? Other than that, I think this is good to go @nicolethoen . It might be nice to show an example of the notification count in the badge, we we could always add that later.

I think that is a result of the `pf-m-selected` modifier being applied to the badge when the drawer is open.

packages/react-core/src/components/NotificationDrawer/NotificationDrawerHeader.tsx
```diff
@@ -30,7 +30,9 @@ export const NotificationDrawerHeader: React.FunctionComponent<NotificationDrawe
     <Text component={TextVariants.h1} className={css(styles.notificationDrawerHeaderTitle)}>
       {title}
     </Text>
-    {count && <span className={css(styles.notificationDrawerHeaderStatus)}>{`${count} ${unreadText}`}</span>}
+    {count !== undefined && (
+      <span className={css(styles.notificationDrawerHeaderStatus)}>{`${count} ${unreadText}`}</span>
```
jschuler, line 34: this seems a bit too rigid of a format and won't work well with i18n

nicolethoen: What do you propose?

jschuler: Not a blocker if this was already released, but I question if the count prop here is needed, user could just pass `<NotificationDrawerHeader unreadText=`${count} unread` />`

nicolethoen: ah
`NotificationDrawerGroup` also expects a `count: number;` to display in its header.

I guess I'm not sure I see a big difference, and it would be a breaking change to remove `count` from `NotificaitonDrawerHeader`

jschuler: It's a little different in the drawer group because it's just the count there. Here we mandate that count is followed by a space and then a string which is not correct for some languages. This already existed before so not a blocker for this PR tho! But we are accruing technical debt here, since this is a beta component I am not sure if breaking changes are allowed or not.

jschuler: https://github.com/patternfly/patternfly-react/issues/4651

Issue #4641, opened 2020-07-31
Select:  Add support for divider in the grouped select


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Add support to add a divider to the select.

PR #4642, opened 2020-08-01
feat(Select): add support for favourites

### LINKED ISSUE #4528: Select - add support for favorites
https://github.com/patternfly/patternfly/pull/3238

This mimics the app launcher favorites for the most part with the exception that in the select component, the favorite/star action also has the class `.pf-m-favorite-action`. The item with class `.pf-c-select__menu-item.pf-m-action` should be considered a generic action element that can be added beside a menu item. The class `.pf-m-favorite-action` targets that particular action so we can style it with custom colors and whatever else we might want to apply to the favorite action.

Ping me with any questions.

tlabaj:
**What**: Closes #4528 
This adds the support for favourites.  React does not currently support Dividers in Select.  That will be covered in issue #4641.


tlabaj:
> @tlabaj This seems to be working as expected except for the clear (x) button. Looks like that only works when I'm typing into the field. Was it always this way? If it's not going to always clear the current selection, then it should probably be hidden. Seems like a bug.

@mcarrano that is the way the typeahead as always worked.  If you look at the other typeahead examples, you will see the same behaviour.

nicolethoen:
I think it looks good! She we be able to favorite the item using the keyboard? I was unable to figure out how to do that.

mcoker:
Looks good! A couple of comments.

In a favorites item, when we add `.pf-m-wrapper` to the list item, we need to identify the button (`.pf-c-select__menu-item`) that serves as the select item text by adding `.pf-m-link` to that item.

The favorite action icon should be wrapped in `span.pf-c-select__menu-item-action-icon`

nicolethoen:
> Looks good!
> 
> @nicolethoen @tlabaj
> Re: keyboard interaction. I remember when implementing favorites for app launcher I had to adjust the DropdownMenu refCollection to a multidimensional array [][] (adding the favorites button refs) and then adjust the arrow key handler to account for left/right. Since single/checkbox select use the same keyboard helper function, all that would need to be updated is the Select refCollection and the typeahead keyboard handler (and the getNextIndex helper function potentially).
> 
> Maybe it's worth opening a new issue to add in the keyboard navigation?

@kmcfaul @tlabaj Was an accessibility issue opened?

tlabaj:
@kmcfaul @nicolethoen @jessiehuff I have opened this follow up issue for keyboard interactions
https://github.com/patternfly/patternfly-react/issues/4650

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -481,13 +490,47 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
...
+
+    // If onFavorites is set,  add isFavorite prop to children and add a Favorites group to the SelectMenu
+    let renderableItems: React.ReactNode[] = [];
+    if (onFavorite) {
+      let favoritesGroup: React.ReactNode[] = [];
+      let renderableFavorites: React.ReactNode[] = [];
+      if (favorites.length > 0) {
+        // if variant is typeahead use filtered children
+        const tempRenderableChildren =
+          variant === 'typeahead' || 'typeaheadmulti' ? typeaheadFilteredChildren : children;
```
jschuler, line 512:
```suggestion
          variant === 'typeahead' || variant === 'typeaheadmulti' ? typeaheadFilteredChildren : children;
```

jschuler: is there a test to check if this works properly with both variants?

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -481,13 +490,47 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
...
+        renderableFavorites = createRenderableFavorites(tempRenderableChildren, isGrouped, favorites);
+        favoritesGroup = [
+          <SelectGroup key="favorites" label={favoritesLabel}>
+            {renderableFavorites}
+          </SelectGroup>
+        ];
+      }
+      // if variant is type-ahead call the extendTypeaheadChildren before adding favorites
+      const tempExtendedChildren =
+        variant === 'typeahead' || 'typeaheadMulti' ? this.extendTypeaheadChildren(typeaheadActiveChild) : children;
```
jschuler, line 522:
```suggestion
        variant === 'typeahead' || variant === 'typeaheadMulti' ? this.extendTypeaheadChildren(typeaheadActiveChild) : children;
```

packages/react-core/src/components/Select/Select.tsx
tlabaj: Yes.  in the demo app

packages/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -109,16 +121,47 @@ export class SelectOption extends React.Component<SelectOptionProps> {
...
       inputId,
+      isFavorite,
+      ariaIsFavoriteLabel = 'starred',
+      ariaIsNotFavoriteLabel = 'not starred',
       ...props
     } = this.props;
     /* eslint-enable @typescript-eslint/no-unused-vars */
     const Component = component as any;
+
+    if (!id && isFavorite !== null) {
```
jschuler, line 132: can they not use the existing `key` prop instead of adding id?

packages/react-core/src/components/Select/__tests__/Select.test.tsx
```diff
@@ -336,7 +353,7 @@ describe('typeahead multi select', () => {
     expect(view).toMatchSnapshot();
   });
 
-  test('test onChange', () => {
+  xtest('test onChange', () => {
```
jschuler, line 356: some tests are disabled, can we turn them back on?

tlabaj: So these test are just snapshot test that are not testing for anything.  They fail everytime due to the random id that is now added to the select options.

packages/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -109,16 +121,47 @@ export class SelectOption extends React.Component<SelectOptionProps> {
...
+
     return (
       <SelectConsumer>
-        {({ onSelect, onClose, variant, inputIdPrefix }) => (
+        {({ onSelect, onClose, variant, inputIdPrefix, onFavorite }) => (
           <React.Fragment>
             {variant !== SelectVariant.checkbox && (
-              <li role="presentation">
+              <li
+                id={generatedId}
```
jschuler, line 158: is this id here needed? will fail snapshot tests

packages/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -127,6 +170,7 @@ export class SelectOption extends React.Component<SelectOptionProps> {
                     isDisabled && styles.modifiers.disabled,
                     isFocused && styles.modifiers.focus,
                     description && styles.modifiers.description,
+                    isFavorite && styles.modifiers.link,
```
mcoker, line 173: It looks like the `pf-m-link` modifier is only applying to the `pf-c-select__menu-item` when it is a favorited item. The `pf-m-link` modifier should apply to the `pf-c-select__menu-item` that holds the main item text for the list item any time there are multiple actions in an item. `pf-m-link` is what differentiates the main item text `pf-c-select__menu-item` from other action `pf-c-select__menu-item`s.

Another way of saying that is that any time the parent is `pf-c-select__menu-wrapper`, the `pf-c-select__menu-item` that holds the item text needs to have `pf-m-favorite`. Re-using the test above, it seems like this could be achieved via

```suggestion
                    isFavorite !== null && styles.modifiers.link,
```

packages/react-core/src/components/Select/SelectOption.tsx
tlabaj: Yes.  It is needed.  Previously the id was generated by the menu, I moved her to facilitate favorites.  The random Id generator was not being used previously so the id was predictable and the test cases would pass. So, when a consumer creates an option in the select, the do no supply an id and it is always generated causing snapshot to fail.

kmcfaul:
@tlabaj @mcoker 

It looks like `pf-m-link` is overriding the grey highlight created by `pf-m-focus`. Or maybe the two classes just need to be swapped in order?

packages/react-core/src/components/Select/__tests__/Select.test.tsx
jschuler: Can the tests be made deterministic so that the id always stays the same?

gdoyle1:
@tlabaj this looks great!! thank you for your work here

nicolethoen:
There are a few things preventing the favorites demo from working as desired with regard to the typeahead variant.
The css needs to apply the .pf-c-select__menu-wrapper:focus-within styles to items with the pf-c-select__menu-wrapper children have the .pf-m-focus modifier applied. This is because the focus STAYS on the typeahead input even when you are using the arrow keys to navigate the menu.
This is why the ‘left’ and ‘right’ arrow keys are not shifting focus to the favorite star - they are moving the cursor left and right in the typeahead input.
So the act of keeping focus on the input is going to hinder the ability to use keyboard controls to navigate typeahead variants paired with favorites.

packages/react-core/src/components/Select/examples/Select.md
```diff
@@ -1701,3 +1701,93 @@ class SelectMenuDocumentBody extends React.Component {
...
+        });
+    };
+
+    this.options = [
+      <SelectGroup label="Status" key="group1">
+        <SelectOption id={"option-1"} key={0} value="Running" description="This is a description." />
+        <SelectOption id={"option-2"} key={1} value="Stopped" />
+        <SelectOption id={"option-3"} key={2} value="Down (disabled)" isDisabled/>
+        <SelectOption id={"option-4"} key={3} value="Degraded" />
+          <SelectOption id={"option-5"} key={4} value="Needs Maintenence" />
```
jschuler, line 1760:
```suggestion
          <SelectOption id={"option-5"} key={4} value="Needs Maintenance" />
```

kmcfaul:
Tab works well, but the disabled option in the favorites is throwing an error when you use the arrow keys (and preventing arrow down from continuing when not tabbed inside the menu). Looks like `handleTypeaheadKeys` needs to handle null refs for disabled items.

Also while focus is on the input, the 'focus' highlight is being applied only to the favorite button opposed to the whole line, and the favorites themselves aren't being highlighted properly (it is applying the highlight to the original item when you arrow through the favorites list). Tabbing into the menu, both look fine.

packages/react-core/src/helpers/util.ts
```diff
@@ -275,105 +277,66 @@ export const toCamel = (s: string) => s.replace(/([-_][a-z])/gi, camelize);
...
+ * This function is a helper for creating an array of renderable favorite items for the Application launcher or Select
  *
- * @param {string} text The text to calculate the width for
- * @param {HTMLElement} node The HTML element
+ * @param {object} items The items rendered in Select or Application aLauncher
+ * @param {boolean} isGrouped Flag indicating if items are grouped
+ * @param {any[]} favorites Array of ids of favorited items
+ * @param {boolean} isEnterTriggersArrowDown Flag indicating if we should add isEnterTriggersArrowDown to favorited item
  */
-export const getTextWidth = (text: string, node: HTMLElement) => {
```
jschuler, line 285: looks like a bad merge!

jschuler, line 285: uh oh, this is not merged correctly

mcoker (dismissed):
LGTM. A few thoughts:

* We need a "focus" selector that the react component can apply to rows to get the proper styling as a keyboard user navigates.
* I think it's kind of confusing that you have to use the tab key to access actions in the menu. I might expect just using the up/down arrows to traverse the interactive elements in the menu based on the DOM order (straight down for a normal select with 1 interactive element per item row, left to right, then down for item rows with multiple interactive children)
* When you use the keyboards to access items in the menu, if you have favorite items in the menu, those items are skipped - you can only use the up/down arrows to navigate non-favorited items. Though it's worth noting you can use tab to navigate to favorited items. Feels inconsistent, I would expect navigation by keyboard to go through all items in the menu in order that they appear.

jschuler (dismissed):
Thanks @tlabaj !

nicolethoen:
> Also while focus is on the input, the 'focus' highlight is being applied only to the favorite button opposed to the whole line, and the favorites themselves aren't being highlighted properly (it is applying the highlight to the original item when you arrow through the favorites list). Tabbing into the menu, when something is favorited, two highlights are applied and move with the arrows.

I've opened an issue in core to address this. https://github.com/patternfly/patternfly/issues/3429

kmcfaul (dismissed):

evwilkin (approved):
LGTM

jessiehuff (approved):
LGTM!

mcoker:
Tabbing doesn't seem to work. The focus goes back and forth between the typeahead and the first item, and looks like the actual focus is moving as opposed to the focus staying in the typeahead and the `pf-m-focus` being applied to the items as you tab.

kmcfaul (approved):

kmcfaul:
@mcoker That's the intended keyboard interaction for favorites (specifically, shouldn't be for normal typeahead variants). It lets keyboard users get to the favorite button via arrows

mcoker (approved):
Thanks for the clarification @kmcfaul! lgtm!

mcarrano:
This looks good. But once https://github.com/patternfly/patternfly-react/pull/4772 is merged to add dividers to Select,  the example should include a divider that separated the Favorites section from the other Select items as in the designs here: https://github.com/patternfly/patternfly/issues/3209

PR #4663, opened 2020-08-05
feat(Select): fix space/enter closing typeahead variants incorrectly

### LINKED ISSUE #4664: Typeahead regression
There is a regression in the select for typeahead variation where the space/enter closes incorrectly.

kmcfaul:
**What**: Closes #4664

Check preventing typeahead variants' key inputs reaching the other variants' open/close logic was removed.

### DIFF
--- a/packages/react-core/src/components/Select/SelectToggle.tsx
+++ b/packages/react-core/src/components/Select/SelectToggle.tsx
@@ -135,6 +135,8 @@ export class SelectToggle extends React.Component<SelectToggleProps> {
     }
 
     if (
+      variant === SelectVariant.typeahead ||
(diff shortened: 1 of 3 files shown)


jschuler:
Nice, looks like the previous regression that this fixes was accidentally introduced out of the desire to prevent the Enter key from submitting a form, e.g. if you have
```
<form onSubmit={() => alert('submitted')}>
  <Select />
</form>
```
. Perhaps that needs to be re-addressed in a future PR

kmcfaul:
@jschuler I see, possibly needs a flag to hit a `preventDefault` in that case. We have `FormSelect` for form use cases I believe as well.

Issue #4676, opened 2020-08-10
Page demos: add tertiary nav demo


mcoker:
Noticed we're missing a tertiary nav demo - https://www.patternfly.org/v4/documentation/core/demos/page/tertiary-nav

The unique part of that is the section that wraps the tertiary nav is `section.pf-c-page__main-nav`

PR #4686, opened 2020-08-12
feat(Page): add width limit

### LINKED ISSUE #4656: Add variation to page sections to limit width of content inside
https://github.com/patternfly/patternfly/pull/3352

on any of these sections, apply `.pf-m-limit-width` to the element, then wrap the children in `div.pf-c-page__main-body`

* `.pf-c-page__main-nav`
* `.pf-c-page__main-breadcrumb`
* `.pf-c-page__main-section`
* `.pf-c-page__main-wizard`

Let me know if there are any questions.

kmcfaul:
**What**: Closes #4656

@mcoker I couldn't find any use of `.pf-c-page__main-wizard`, where should I be adding the limit flag for this? Also, is my edit of the breadcrumb correct with the placement of the `div.pf-c-page__main-body`?

### DIFF
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -38,6 +38,8 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   notificationDrawer?: React.ReactNode;
   /** Flag indicating Notification drawer in expanded */
   isNotificationDrawerExpanded?: boolean;
+  /** Flag indicating if breadcrumb width should be limited */
+  isBreadcrumbWidthLimited?: boolean;
   /** Callback when notification drawer panel is finished expanding. */
   onNotificationDrawerExpand?: () => void;
   /** Skip to content component for the page */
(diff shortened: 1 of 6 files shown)


mcoker:
@kmcfaul that looks good to me, although tertiary navigation is supposed to go above breadcrumbs. How does someone add tertiary nav to a page? I noticed there is no tertiary nav demo, so I added an issue to create one - https://github.com/patternfly/patternfly-react/issues/4676

I'm guessing currently they would just add a `<PageSection><Nav variant="Tertiary">...</Nav></PageSection>` as a child of the page component?

mcoker:
To your question about the wizard, is there currently a demo or example of the wizard used in context of the page component? We have a unique section type for that. If there isn't one (meaning users likely aren't using the section we have for it in core) I have an issue in core to investigate a scrolling problem with the wizard and can see if I can get the wizard to work within a regular page section and we can deprecate that element.

kmcfaul:
I don't see a wizard demo that nests inside Page, just by itself. @jschuler Am I missing that demo somewhere?

If it's not there I'll just leave the code changes as is.

PR #4690, opened 2020-08-13
feat(topology): Enable saving current topology graph model

jeff-phillips-18:
**What**: 
This PR is aimed at allowing adopters to track the current layout of nodes so that it can be saved and restored.

Provides API for `toModel` for the current Controller/GraphElements.
Keeps track of nodes that have had their position set, do not reset positions on layout.
Adds events on individual node moves, and overall event for all node moves (debouncing individual node moves)

/cc @christianvogt

### DIFF
--- a/packages/react-topology/src/Visualization.ts
+++ b/packages/react-topology/src/Visualization.ts
@@ -45,15 +45,24 @@ export class Visualization extends Stateful implements Controller {
   }
 
   @action
-  fromModel(model: Model): void {
+  fromModel(model: Model, merge: boolean = true): void {
     const oldGraph = this.graph;
 
-    // create elements
+    // If not merging, clear out the old elements
+    if (!merge) {
+      _.forIn(this.elements, element => this.removeElement(element));
+    }
+
+    // Create the graph if given in the model
     if (model.graph) {
       this.graph = this.createElement<Graph>(ModelKind.graph, model.graph);
+      if (!merge && oldGraph) {
+        this.graph.setDimensions(oldGraph.getDimensions());
+      }
     }
-    const validIds: string[] = [];
 
+    // Create elements
+    const validIds: string[] = [];
     const idToElement: { [id: string]: ElementModel } = {};
 
     model.nodes &&
@@ -92,14 +101,16 @@ export class Visualization extends Stateful implements Controller {
     model.edges && model.edges.forEach(processElement);
 
     // remove all stale elements
-    _.forIn(this.elements, element => {
-      if (!isGraph(element) && !validIds.includes(element.getId())) {
-        this.removeElement(element);
-      }
-    });
+    if (merge) {
+      _.forIn(this.elements, element => {
+        if (!isGraph(element) && !validIds.includes(element.getId())) {
+          this.removeElement(element);
+        }
+      });
 
-    if (oldGraph && oldGraph !== this.graph) {
-      this.removeElement(oldGraph);
+      if (oldGraph && oldGraph !== this.graph) {
+        this.removeElement(oldGraph);
+      }
     }
 
     if (this.graph) {
@@ -132,6 +143,17 @@ export class Visualization extends Stateful implements Controller {
 
   getElements(): GraphElement[] {
     return _.values(this.elements);
+  }
+
+  toModel(): Model {
+    const graph = this.getGraph();
+    const nodes = this.getElements().filter(n => isNode(n)) as Node[];
+    const edges = this.getElements().filter(e => isEdge(e)) as Edge[];
+    return {
+      graph: graph.toModel(),
+      nodes: nodes.map(n => n.toModel()),
+      edges: edges.map(e => e.toModel())
+    };
   }
 
   addElement(element: GraphElement): void {
--- a/packages/react-topology/src/elements/BaseEdge.ts
+++ b/packages/react-topology/src/elements/BaseEdge.ts
@@ -160,4 +160,13 @@ export default class BaseEdge<E extends EdgeModel = EdgeModel, D = any> extends 
       this.bendpoints = model.bendpoints ? model.bendpoints.map(b => new Point(b[0], b[1])) : [];
     }
   }
+
+  toModel(): EdgeModel {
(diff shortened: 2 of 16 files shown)


christianvogt:
@jeff-phillips-18 i'll have a look at this today or tomorrow

packages/react-topology/src/types.ts
```diff
@@ -126,6 +127,7 @@ export interface Node<E extends NodeModel = NodeModel, D = any> extends GraphEle
   setBounds(bounds: Rect): void;
   getPosition(): Point;
   setPosition(location: Point): void;
+  clearPosition(): void;
```
christianvogt, line 130: I think we can remove `clearPosition`.
All other comments are related to trying to remove this function.

packages/react-topology/src/layouts/BaseLayout.ts
```diff
@@ -439,7 +439,7 @@ class BaseLayout implements Layout {
     this.edges = this.getLinks(this.graph.getEdges());
 
     // initialize new node positions
-    this.initializeNodePositions(newNodes, this.graph, initialRun);
+    this.initializeNodePositions(this.nodes, this.graph);
```
christianvogt, line 442: Let's keep `initialRun` as a param to force layout on the nodes.

packages/react-topology/src/layouts/BaseLayout.ts
```diff
@@ -359,15 +363,17 @@ class BaseLayout implements Layout {
...
-      const { x, y } = node.element.getPosition();
-      if (force || (x === 0 && y === 0)) {
+
+    nodes.forEach((node: LayoutNode) => {
+      // only init position for nodes that have not been positioned
+      if (!node.element.isPositioned()) {
         node.setPosition(cx, cy);
+      } else {
+        node.setFixed(true);
       }
```
christianvogt, line 377:
```suggestion
      if (force || !node.element.isPositioned()) {
        node.setPosition(cx, cy);
      } else if (!force) {
        node.setFixed(true);
      }
```

packages/react-topology/src/elements/BaseNode.ts
```diff
@@ -142,6 +145,19 @@ export default class BaseNode<E extends NodeModel = NodeModel, D = any> extends
 
   setPosition(point: Point): void {
     this.position = point;
+    this.positioned = true;
+    try {
+      this.getController().fireEvent(NODE_POSITIONED_EVENT, { node: this });
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
```
christianvogt, line 152: A bit concerned that this can result in a lot of event notifications during the layout which may impact performance.

I wonder if we should consider a single event notification that's debounced that can announce `model changed`.  This isn't really ideal either.

packages/react-topology/src/elements/BaseGraph.ts
```diff
@@ -133,8 +139,21 @@ export default class BaseGraph<E extends GraphModel = GraphModel, D = any> exten
...
+
+    try {
+      this.getController()
+        .getElements()
+        .filter(e => isNode(e))
+        .forEach(n => {
+          (n as Node).clearPosition();
+        });
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
```
christianvogt, line 156: We can remove this if we can remove `clearPosition` altogether.

packages/react-topology/src/elements/BaseEdge.ts
```diff
@@ -160,4 +160,13 @@ export default class BaseEdge<E extends EdgeModel = EdgeModel, D = any> extends
       this.bendpoints = model.bendpoints ? model.bendpoints.map(b => new Point(b[0], b[1])) : [];
     }
   }
+
+  toModel(): EdgeModel {
+    return {
+      ...super.toModel(),
+      source: this.getSource() ? this.getSource().getId() : '',
+      target: this.getTarget() ? this.getTarget().getId() : '',
```
christianvogt, line 168: I think we should have these as undefined:
```suggestion
      source: this.getSource() ? this.getSource().getId() : undefined,
      target: this.getTarget() ? this.getTarget().getId() : undefined,
```

packages/react-topology/src/elements/BaseNode.ts
```diff
@@ -287,6 +299,19 @@ export default class BaseNode<E extends NodeModel = NodeModel, D = any> extends
     }
   }
 
+  toModel(): NodeModel {
+    return {
+      ...super.toModel(),
+      x: this.getPosition().x,
+      y: this.getPosition().y,
```
christianvogt, line 306: Only include the position if `this.positioned === true`

packages/react-topology/src/types.ts
```diff
@@ -171,6 +173,7 @@ export interface Graph<E extends GraphModel = GraphModel, D = any> extends Graph
   getLayout(): string | undefined;
   setLayout(type: string | undefined): void;
   layout(): void;
+  stopLayout(): void;
```
christianvogt, line 176: I'd prefer to not add this function to the graph api.
They can always do `graph.getLayout()?.stopLayout()`

packages/react-topology/src/types.ts
```diff
@@ -9,6 +9,7 @@ export type PointTuple = [number, number];
 
 export interface Layout {
   layout(): void;
+  stopLayout(): void;
```
christianvogt, line 12: Call it `stop(): void` ? Since this is a function of the layout itself.

packages/react-topology/src/Visualization.ts
```diff
@@ -45,12 +45,20 @@ export class Visualization extends Stateful implements Controller {
...
 
   @action
-  fromModel(model: Model): void {
+  fromModel(model: Model, merge: boolean = true): void {
     const oldGraph = this.graph;
 
+    // If not merging, clear out the old elements
+    if (!merge) {
+      this.elements = {};
+    }
```
christianvogt, line 54: We may want to go through the routine of destroying the element for the sake of consistency.

PR #4691, opened 2020-08-13
feat(TextInput): add helper util for left trim

### LINKED ISSUE #4637: Truncate the left side content
Hi @mcarrano, will patternfly consider supporting truncate the left-side content as been shown in the screenshot? As the use case I am currently working on, the content on the right side is more important than the left side. By truncating the left side, the contents of the right side can always be shown. 
Screen Shot 2020-06-10 at 11 41 32 AM

jenny-s51:
**What**: Closes #4637 

This is WIP; working on adding this functionality to `TextInput`.


packages/react-core/src/helpers/util.ts
```diff
@@ -273,3 +273,79 @@ export const toCamel = (s: string) => s.replace(/([-_][a-z])/gi, camelize);
...
+
+
+            console.log("frac: ", frac);
+            console.log('sub', sub)
+            console.log("value length", value.length);
+            console.log("percent", frac * 100)
+            console.log("offsetwidth", row.offsetWidth)
+            console.log("scrollwidth", row.scrollWidth)
+            let newLength = frac;
+            value = '...' + value.substr(100);
```
jenny-s51: the `.substr` parameter is used as a placeholder and probably shouldn't be 100. That value needs to based on some calculation from the variables above. @dgutride  suggested using `Math.ceil(valueLength * frac)` which works, but this wasn't truncating to the full width of the input field.

packages/react-core/src/helpers/util.ts
```diff
@@ -273,3 +273,79 @@ export const toCamel = (s: string) => s.replace(/([-_][a-z])/gi, camelize);
...
+            parentVal = value;
+            console.log("parentVal", parentVal);
+            (node as HTMLInputElement).value = parentVal;
+            console.log((node as HTMLInputElement).value)
```
jenny-s51, line 316: Don't want to replace value with the truncated value ... should still be the value entered by the user. Is there a way to do this so that only the preview is truncated?

packages/react-core/src/helpers/util.ts
```diff
@@ -273,3 +273,80 @@ export const toCamel = (s: string) => s.replace(/([-_][a-z])/gi, camelize);
...
+          if (row.scrollWidth > row.offsetWidth) {
+
+
+            console.log("frac: ", frac);
+            console.log('sub', sub)
+            console.log("value length", value.length);
+            console.log("percent", frac * 100)
+            console.log("offsetwidth", row.offsetWidth)
+            console.log("scrollwidth", row.scrollWidth)
+            let newLength = Math.ceil(value.length * frac);
```
jenny-s51, line 310: This calculation was suggested by @dgutride and seems to work, but it doesn't truncate to the full width of the input box ... is there a `newLength` value that would do this?

packages/react-core/src/components/TextInput/examples/TextInput.md
```diff
@@ -26,7 +26,7 @@ class SimpleTextInput extends React.Component {
     const { value } = this.state;
 
     return (
-      <TextInput value={value} type="text" onChange={this.handleTextInputChange} aria-label="text input example" />
+      <TextInput isLeftTruncated value={value} type="text" onChange={this.handleTextInputChange} aria-label="text input example" />
```
jenny-s51, line 29: Clicking outside of the input field has different behaviors in different browsers. In chrome and safari, when you click outside the input field it returns to the start of the string that has been entered and truncates the right. In Firefox, it seems to stay at the end. Should this functionality be overwritten?

jschuler:
LGTM, I noticed a bug where if you remove isDisabled / isRead and try to type into the textbox it won't let you! But it is not something from this PR, it already exists since the last release at least. But we should fix this @tlabaj

packages/react-core/src/helpers/util.ts
```diff
@@ -273,3 +273,101 @@ export const toCamel = (s: string) => s.replace(/([-_][a-z])/gi, camelize);
...
+/**
+ * This function is a helper for truncating text content on the left, leaving the right side of the content in view
+ *
+ * @param {HTMLElement} node HTML element
+ * @param {string} value The original text value
+ */
+export const trimLeft = (node: HTMLElement, value: string) => {
+  const availableWidth = innerDimensions(node).width;
+  let newValue = value;
+  if (getTextWidth(value, node) > availableWidth) {
```
jschuler, line 361: just one bug here, can you add an `else` condition to this `if`. Noticed that if you just a little overflow and you expand the textbox until the whole text should fit again, it won't restore the value
```
 else {
    if ((node as HTMLInputElement).value) {
      (node as HTMLInputElement).value = value;
    } else {
      node.innerText = value;
    }
  }
```

PR #4701, opened 2020-08-19
feat(TreeView): add treeview, demos, tests

### LINKED ISSUE #4652: Add treeview component
https://github.com/patternfly/patternfly/pull/3354

There are a few follow-ups we're in the progress of working on - https://github.com/patternfly/patternfly/issues/3358

Reach out to @christiemolloy or myself with questions.

kmcfaul:
**What**: Closes #4652


doruskova:
@kmcfaul Thank you for this update! As was mentioned, selecting the parent node should select every child and when I select all and then I deselect some nodes > mixed state should appear. The menu with action on the right side should be black on hover/focus. And also I'm wondering about the bold text for a parent node. I'd leave the font-weight normal for a parent node ([as it's in HTML/CSS design](https://patternfly-pr-3410.surge.sh/documentation/core/components/treeview)). 
Snímek obrazovky 2020-08-24 v 12 02 27

If I understand it well, just current and previous parents are bold. But, when you go deeper into a menu, the first parent is not bold anymore. It'd make sense if every parent that I visited was bold. But, I don't think it's necessary. WDYT @mcarrano  ?
Aug-24-2020 12-24-40

mturley:
@kmcfaul to solve the checkbox state problem, I think you could take advantage of having all the checkbox state at the top level, and inspect the state of subtrees and apply changes recursively.

Since each data item knows about its children, when you map over `data` in `<TreeView>` you could determine the state of that list item's checkbox. Assuming an item can only have a full checkmark if all its children are selected, you could maybe do something like:

```ts
// Helpers
const isChecked = (dataItem) => checkedItems.includes(dataItem.id);
const areAllDescendantsChecked = (dataItem) =>
  (!dataItem.children && isChecked(dataItem)) ||
  dataItem.children.every((child) => areAllDescendantsChecked);
const areSomeDescendantsChecked = (dataItem) =>
  (!dataItem.children && isChecked(dataItem)) ||
  dataItem.children.some((child) => areSomeDescendantsChecked);

// In your `data.map()` function
const hasCheck = areAllDescendantsChecked(data);
const hasPartialCheck = areSomeDescendantsChecked(data);

```

And then you can also select all descendants by walking the data recursively too:

```ts
this.onCheck = (evt, treeViewItem) => {
  const checked = evt.target.checked;
  console.log(checked);

  const itemsToCheck = [];
  const checkItemAndDescendants = (item) => {
    itemsToCheck.push(item);
    if (item.children) {
      item.children.forEach((child) => checkItemAndDescendants(child));
    }
  };
  checkItemAndDescendants(treeViewItem);

  this.setState(
    (prevState) => ({
      checkedItems: checked
        ? [
            ...prevState.checkedItems,
            ...itemsToCheck.filter(
              (item) => !prevState.checkedItems.map((i) => i.id).includes(item.id)
            ),
          ]
        : prevState.checkedItems.filter((item) => !itemsToCheck.includes((i) => i.id === item.id)),
    }),
    () => {
      console.log('Checked items: ', this.state.checkedItems);
    }
  );
};
```

I thiiiink that would work?

kmcfaul:
@mturley Thanks so much for the help! I was starting to go in circles with all the recursion and re-mapping. Between what I had done and your code I was able to get the checkbox behavior we wanted! 😄

mturley:
Awesome! Glad I could help!

mturley:
Oof, just realized I botched my example code with a bad copy-paste in my previous edit. Fixed, but it sounds like you got the idea anyway.

jschuler:
Can you/should you be able to specify different icons for leaf nodes? For example in the icons example i would expect leaf nodes not to have a folder icon

kmcfaul:
@jschuler The icons example passes in the folder icons at the top level `TreeView`, which will pass down the icons to each item. Alternatively, an item can define their own `icon` / `expandedIcon` property, so you would be able to mix and match icons this way.

As far as the checkbox expanded/collapsing, I think the onClick on the item is also getting triggered when the input changes. @mcoker @christiemolloy Would the checkbox structure changes coming to TreeView in core potentially fix this?

jschuler:
@kmcfaul you could probably add `event.stopPropagation()` for `onCheck`

packages/react-core/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,125 @@
...
+  <TreeViewList isNested={isNested} onSearch={onSearch} searchProps={searchProps} {...props}>
+    {data.map(item => (
+      <TreeViewListItem
+        key={item.name.toString()}
+        name={item.name}
+        id={item.id}
+        defaultExpanded={defaultAllExpanded || item.defaultExpanded || false}
+        onSelect={onSelect}
+        onCheck={onCheck}
+        hasCheck={hasChecks || item.hasCheck || false}
```
jschuler, line 91: if item should be able to override the default, I think you'll need something like
`hasCheck={item.hasCheck !== 'undefined' ? item.hasCheck : hasChecks}`

packages/react-core/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,125 @@
...
+        checkProps={item.checkProps}
+        hasBadge={hasBadges || item.hasBadge || false}
```
jschuler, line 93: similar to `hasCheck` might want to give preference to the item setting over the tree default

packages/react-core/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,125 @@
...
+        badgeProps={item.checkProps}
+        activeItems={activeItems}
+        parentItem={parentItem}
+        itemData={item}
+        icon={icon || item.icon || null}
+        expandedIcon={expandedIcon || item.expandedIcon || null}
```
jschuler, line 99: same

mcoker:
@kmcfaul good point, I must have missed that initially. It's invalid to have an `<input type="checkbox">` in a `<button>`.

I think @jschuler's suggestion is probably the way to go for now. We may reconsider what happens when you click on an item with a checkbox. We could do something like have the toggle icon on the left expand the item (as a `<button>`), and adjacent to it, clicking anywhere else in the item select it (via the checkbox) (as a `<label>`).

I opened an issue to investigate more here - https://github.com/patternfly/patternfly/issues/3425

kmcfaul:
event.stopPropagation() doesn't work so it may just have to remain as it is for now?

kmcfaul:
Updated to allow the item level props to override the tree level props.

kmcfaul:
@jschuler @mcoker Fixed the checkboxes triggering the expand/collapse. Had to add the evt.stopPropagation to the `onClick` of the input (the `onCheck` function is an `onChange` handler).

packages/react-core/src/components/TreeView/TreeViewList.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const TreeViewList: React.FunctionComponent<TreeViewListProps> = ({
+  isNested = false,
+  onSearch,
+  searchProps,
+  children,
+  ...props
+}: TreeViewListProps) => {
+  const list = <ul role={isNested ? 'group' : 'tree'}>{children}</ul>;
```
jschuler, line 25: should prop be forwarded here too?

kmcfaul: No because those are the nested lists, the props are forwarded to the outermost div of the Tree View.

kmcfaul:
Will be followed up in #4726 for keyboard patterns and a11y fixes

mturley:
Nice work @kmcfaul ! Looking forward to using this.

PR #4720, opened 2020-08-24
fix(docs): use ids and h3s

redallen:
**What**: Make docs compatible with the new patternfly.org redesign:
- Use `id` instead of `title` in frontmatter
- Use `### Example title` instead of `title=Example-title`
- Add missing imports that were forgiven by previous Gatsby-mdx implementation
- Rename demos to match component names to support a "Demo" tab for some components
  - Consolidate table demos into one file

**Additional issues**: Closes #2037


packages/eslint-plugin-patternfly-react/lib/rules/import-tokens-icons.js
```diff
@@ -1,3 +1,23 @@
+/**
+ * @param specifier module specifier
+ * @param moduleName module name (@patternfly/react-tokens or @patternfly/react-icons)
+ * @returns string of non-treeshaken import
+ */
+function makeImport(specifier, moduleName) {
+  let res = `import ${specifier.local.name} from '`;
+  res += moduleName.replace(/\/dist\/(js|esm)/, '');
+  res += '/dist/js';
```
jschuler, line 9: only `/dist/js`?

redallen: This function builds a path to the CJS module.

Ex: `import { some_token } from '@patternfly/react-tokens';` -> `import some_token from '@patternfly/react-tokens/dist/js/some_token';`

PR #4741, opened 2020-08-27
[WIP] OUIA examples in docs

### LINKED ISSUE #4740: Add OUIA examples into the docs
Patternfly docs in patternfly-react.surge.sh are missing OUIA examples. 

This is an enhancement

quarckster:
**What**: Closes #4740 #4723

This PR adds OUIA examples for the following components:

- Alert
- Breadcrumb
- Button
- ContexSelector
- Dropdown
- FormSelect
- Modal
- Nav
- OptionsMenu
- Pagination
- Select
- Switch
- Table

### DIFF
--- a/packages/react-core/src/components/Alert/examples/Alert.md
+++ b/packages/react-core/src/components/Alert/examples/Alert.md
@@ -19,11 +19,11 @@ class AlertTypes extends React.Component {
   render() {
     return (
       <React.Fragment>
-        <Alert title="Default alert title" ouiaId="default"/>
-        <Alert variant="info" title="Info alert title" ouiaId="info"/>
-        <Alert variant="success" title="Success alert title" ouiaId="success"/>
-        <Alert variant="warning" title="Warning alert title" ouiaId="warning"/>
-        <Alert variant="danger" title="Danger alert title" ouiaId="danger"/>
+        <Alert title="Default alert title"/>
+        <Alert variant="info" title="Info alert title"/>
+        <Alert variant="success" title="Success alert title"/>
+        <Alert variant="warning" title="Warning alert title"/>
+        <Alert variant="danger" title="Danger alert title"/>
       </React.Fragment>
     );
   }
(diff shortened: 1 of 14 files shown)


redallen:
Hey @quarckster , while I love the idea of this PR, I'd much rather fix our OUIA implementation to have more consistent `id`s. Maintaining our examples is costly and we cannot commit to updating the OUIA examples.

Would fixing our implementation to have more reasonable default `id`s solve what you're trying to accomplish in your PR?

quarckster:
@redallen Perhaps in some examples we can use generated ids like Alerts or Buttons. But in more complex components such as Tables it would problematically to do in my opinion. We should demonstrate that a TableRow should have a specific id that reflects the value of some column. Moreover, I would like to see static values in `data-component-id` in order to run our integration tests of [widgetastic.patternfly4](https://github.com/RedHatQE/widgetastic.patternfly4/tree/master/testing/ouia) against Patternfly docs page.

redallen (changes requested):
Our doc site is first and foremost for consumers. If it so happens to be testable by downstream QA with convenient selectors that's great, but adding examples explicitly for QA is a burden to maintain and doesn't help our consumers (most of whom don't even know what OUIA is...)

If you specify `ouiaId="some-static-id"` in the `## Basic` example (rather than copy/pasting it to a new example) and give the consumer a good reason why they should an `ouiaId` using an [example caption](https://github.com/patternfly/patternfly-org/issues/1794), I'd approve this PR. I'd approve it in a hurry if you also create an `OUIA.md` file describing the spec and how to adhere to it using PatternFly and what benefit that provides the product for QA.

quarckster:
> If you specify ouiaId="some-static-id" in the ## Basic example (rather than copy/pasting it to a new example) and give the consumer a good reason why they should an ouiaId using an example caption, I'd approve this PR. I'd approve it in a hurry if you also create an OUIA.md file describing the spec and how to adhere to it using PatternFly and what benefit that provides the product for QA.

Great I will do that, thank you.

PR #4747, opened 2020-08-28
fix: Do not handle modified clicks on VerticalNav links

### LINKED ISSUE #4783: VerticalNav links ignore the modified clicks

Links in VerticalNav open in a current tab instead of in a new one when using ctrl+click. Same for new window and shift+click.

This is reproducible when we pass `onNavigate` prop which should handle the navigation:

```js
import React from "react";
import { Route, Router } from "react-router-dom";
import { createBrowserHistory } from 'history';
import { VerticalNav } from 'patternfly-react';

const Foo = (props) => <div>Foo</div>;
const Bar = (props) => <div>Bar</div>;

const items = [
  { title: 'foo', href: '/foo' },
  { title: 'bar', href: '/bar' },
]

const history = createBrowserHistory();

const App = (props) => {
 
  return (
    <React.Fragment>
      <VerticalNav
        items={items}
        onNavigate={({ href }) => history.push(href)}

      />
      <Router history={history}>
        <Route exact path="/foo" component={Foo} />
        <Route exact path="/bar" component={Bar} />
      </Router>
    </React.Fragment>
  );
};
```
`onItemClick` in `VerticalNavItemHelper` does not check for modified clicks and does not pass original click event to `onNavigate` click handler.

Bug, no workaround known.

https://projects.theforeman.org/issues/29020

xprazak2:
**What:** Closes #4783

VerticalNav should not prevent default and call onClick handler
when user wishes to open a link from vertical navigation in another tab
and uses ctrl+click.

**Additional issues**: https://projects.theforeman.org/issues/29020

### DIFF
--- a/packages/patternfly-react/src/components/VerticalNav/VerticalNavItemHelper.js
+++ b/packages/patternfly-react/src/components/VerticalNav/VerticalNavItemHelper.js
@@ -82,8 +82,12 @@ class BaseVerticalNavItemHelper extends React.Component {
 
   onItemClick = event => {
     const { primary, secondary, tertiary } = this.getContextNavItems();
-    const { isMobile, preventHref, updateNavOnItemClick, idPath } = this.props;
+    const { isMobile, preventHref, updateNavOnItemClick, idPath, preventUpdate } = this.props;
     const { onClick } = this.navItem();
+
+    if (preventUpdate(event)) {
+      return;
+    }
 
     if (preventHref && !!onClick) {
       event.preventDefault();
@@ -331,7 +335,8 @@ BaseVerticalNavItemHelper.propTypes = {
   /** anchor id */
   id: PropTypes.string,
   /** anchor data-id */
-  dataID: PropTypes.string
+  dataID: PropTypes.string,
+  preventUpdate: PropTypes.func
 };
 
 BaseVerticalNavItemHelper.defaultProps = {
@@ -341,7 +346,8 @@ BaseVerticalNavItemHelper.defaultProps = {
   isDivider: false,
   preventHref: true,
   id: null,
-  dataID: null
+  dataID: null,
+  preventUpdate: (event) => false
 };
 
 const VerticalNavItemHelper = getContext(navContextTypes)(BaseVerticalNavItemHelper);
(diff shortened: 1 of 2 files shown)


tlabaj:
Can you please link the related issue in the PattenFly Repo.

xprazak2:
I linked the related issue.

packages/patternfly-react/src/components/VerticalNav/VerticalNavItemHelper.js
```diff
@@ -80,11 +80,17 @@ class BaseVerticalNavItemHelper extends React.Component {
...
   };
 
+  eventModified = event => !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
+
   onItemClick = event => {
     const { primary, secondary, tertiary } = this.getContextNavItems();
     const { isMobile, preventHref, updateNavOnItemClick, idPath } = this.props;
     const { onClick } = this.navItem();
 
+    if (this.eventModified(event)) {
```
jschuler, line 90: instead of this can you not just set the `preventHref` prop to false?

jschuler (changes requested):

packages/patternfly-react/src/components/VerticalNav/VerticalNavItemHelper.js
xprazak2: Setting `preventHref` to false does not take care of the issue entirely. While it opens the link in new tab, it also changes location of current tab because the`onItemClick` handler continues its execution.

redallen: Right, we wrote this component to do exactly that. Can you add a new `preventUpdate` prop to do this instead? Consumers may wish for modifier keys like shift and alt to do other things.

This first lines of `onItemClick` could just look like:
```tsx
const { isMobile, preventHref, updateNavOnItemClick, idPath, preventUpdate } = this.props;
if (preventUpdate) {
  return;
}
```
and you could pass `preventUpdate` instead of `preventHref`.

redallen (changes requested):

xprazak2:
Updated, I like the idea of leaving it to consumers how the modified clicks should be handled. When passing the prop we do not know anything about the click event yet, so the `preventUpdate` needs to be a function.

redallen (approved):
LGTM! Thank you!

jschuler (approved):
thanks @xprazak2 !

PR #4755, opened 2020-08-31
feat(Skeleton): add skeleton component, examples, tests

### LINKED ISSUE #4654: Add skeleton component
https://github.com/patternfly/patternfly/pull/3353

reach out to @christiemolloy or myself with questions

kmcfaul:
**What**: Closes #4654

### DIFF
--- a/packages/react-core/src/components/Skeleton/Skeleton.tsx
+++ b/packages/react-core/src/components/Skeleton/Skeleton.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Skeleton/skeleton';
+import { css } from '@patternfly/react-styles';
+
+export interface SkeletonProps extends React.HTMLProps<HTMLDivElement> {
+  /** Additional classes added to the Skeleton */
+  className?: string;
+  /** The percentage width of the Skeleton */
+  width?: string;
+  /** The percentage height of the Skeleton */
+  height?: string;
+  /** The font size height of the Skeleton */
+  fontSize?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
+  /** The shape of the Skeleton */
+  shape?: 'circle' | 'square' | 'rectangle';
+  /** Width of the shape */
+  shapeWidth?: 'sm' | 'md' | 'lg';
+  /** Height of the shape */
+  shapeHeight?: 'sm' | 'md' | 'lg';
+}
+
+const shapeWidths = {
+  sm: styles.modifiers.widthSm,
+  md: styles.modifiers.widthMd,
+  lg: styles.modifiers.widthLg
+};
+
+const shapeHeights = {
+  sm: styles.modifiers.heightSm,
+  md: styles.modifiers.heightMd,
+  lg: styles.modifiers.heightLg
+};
+
+export const Skeleton: React.FunctionComponent<SkeletonProps> = ({
+  className = '',
+  width = '',
+  height = '',
+  fontSize,
+  shape,
+  shapeWidth,
+  shapeHeight,
+  ...props
+}: SkeletonProps) => {
+  const presetWidth = Object.values(styles.modifiers).find(key => key === `pf-m-width-${width}`);
+  const presetHeight = Object.values(styles.modifiers).find(key => key === `pf-m-height-${height}`);
+  const fontHeight = Object.values(styles.modifiers).find(key => key === `pf-m-text-${fontSize}`);
+  return (
+    <div
+      {...props}
+      className={css(
+        styles.skeleton,
+        presetWidth,
+        presetHeight,
+        fontSize && fontHeight,
+        shape === 'circle' && styles.modifiers.circle,
+        shape === 'square' && styles.modifiers.square,
+        shapeWidth && shapeWidths[shapeWidth],
+        shapeHeight && shapeHeights[shapeHeight],
+        className
+      )}
+      {...(((width && !presetWidth) || (height && !presetHeight)) && {
+        style: {
+          width: width && !presetWidth ? `${width}%` : undefined,
+          height: height && !presetHeight ? `${height}%` : undefined,
+          ...props.style
+        }
+      })}
+    ></div>
+  );
+};
+Skeleton.displayName = 'Skeleton';
--- a/packages/react-core/src/components/Skeleton/Skeleton.tsx
+++ b/packages/react-core/src/components/Skeleton/Skeleton.tsx
@@ -5,30 +5,22 @@ import { css } from '@patternfly/react-styles';
 export interface SkeletonProps extends React.HTMLProps<HTMLDivElement> {
   /** Additional classes added to the Skeleton */
   className?: string;
-  /** The percentage width of the Skeleton */
+  /** The width of the Skeleton. Must specify pixels or percentage, or may be one of "sm", "md", or "lg". */
   width?: string;
-  /** The percentage height of the Skeleton */
+  /** The height of the Skeleton. Must specify pixels or percentage, or may be one of "sm", "md", or "lg". */
   height?: string;
   /** The font size height of the Skeleton */
   fontSize?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
   /** The shape of the Skeleton */
   shape?: 'circle' | 'square' | 'rectangle';
-  /** Width of the shape */
-  shapeWidth?: 'sm' | 'md' | 'lg';
-  /** Height of the shape */
-  shapeHeight?: 'sm' | 'md' | 'lg';
 }
 
-const shapeWidths = {
-  sm: styles.modifiers.widthSm,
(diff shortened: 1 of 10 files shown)


ryelo (dismissed):
This looks great! Really looking forward to switching ours out for this one!

packages/react-integration/cypress/integration/skeleton.spec.ts
```diff
@@ -0,0 +1,22 @@
+describe('Spinner Demo Test', () => {
```
tlabaj, line 1: should be `Skeleton` not `Spinner`

packages/react-core/src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,71 @@
...
+  /** The percentage width of the Skeleton */
+  width?: string;
+  /** The percentage height of the Skeleton */
+  height?: string;
+  /** The font size height of the Skeleton */
+  fontSize?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
+  /** The shape of the Skeleton */
+  shape?: 'circle' | 'square' | 'rectangle';
+  /** Width of the shape */
+  shapeWidth?: 'sm' | 'md' | 'lg';
```
tlabaj, line 17: I wonder if we need to separate out the shape width here.  From what I see in core, I would expect one prop for the width modifiers. 
I don't see anything that says that the ` 'sm' | 'md' | 'lg' ` only applies when a shape is set. Or, that you can not use percentage for when you use a shape.  @mcoker  is this the case?
For height I would expect a similar prop.
@mcarrano @mcoker do we want to allow custom heights and widths?

tlabaj (changes requested):

packages/react-core/src/components/Skeleton/Skeleton.tsx
mcoker: @tlabaj yep and yep!

By default, the skeleton is the height of a line of text (the font-size * line-height), and will fill the width of the container it is in.

From there, a user can assign a height and/or a width. And those can be one of our preset modifiers, or they could pass something custom like `300px` or `42%` or whatever. That custom value can be used as the value for `--pf-c-skeleton--Width` and `--pf-c-skeleton--Height`. For example, `<Skeleton height="300px">...` renders `<div class="pf-c-skeleton" style="--pf-c-skeleton--Height: 300px">...`

And from there, a user can also assign a shape of either circle or square. When a shape is assigned, any custom height is discarded, and the height just matches the specified width. We could either make that part of the documentation, or we could do this if they have set a height with a circle/square shape:

* If a user has passed both a height and a width, we discard the height for them.
* If the user has specified either a custom height or one of our height modifiers, we make that the value of the width instead.

kmcfaul: @tlabaj @mcoker Updated with these changes:

- removed `shapeWidth`/`shapeHeight` => rolled into `width`/`height`
- updated custom width/height styling to use `--pf-c-skeleton--Width` and `--pf-c-skeleton--Height` (instead of style width and height)
- updated `width` and `height` to take any `%` or `px` or preset of `sm`|`md`|`lg`. But now it does need `%` or `px` in the string to work, and presets are only detected when using `%`.

mcoker:
@kmcfaul this looks great! I'm wondering if the percentage width and height modifiers are necessary, and assuming we want to keep them, if they're passed to the `width` and `height` props like they are currently, if we need to use the modifiers classes from core core or if the value could just be used directly as the value for the variable, just like if someone passed a percentage that isn't in our preset sizes.

Right now `width="50%"` assigns `.pf-m-width-50` to the element. But `width="51%"` assigns `--pf-c-skeleton--Width: 51%;`. I don't see any benefit of using `.pf-m-width-50` over `--pf-c-skeleton--Width: 51%;`. And it seems like extra logic to check for it. wdyt @kmcfaul? Is there a benefit to the way it is now?

The size modifiers are useful for non-react users, and if we wanted to support something in the future like `width-50, height-25, width-75-on-lg, width-100-on-xl, height-50-on-2xl` allowing for different sizes at prescribed breakpoints.

FWIW [Material's skeleton](https://material-ui.com/components/skeleton/) doesn't offer size variations outside of mapping to things like font-size - https://material-ui.com/components/skeleton/#inferring-dimensions.

I also wonder about the sm, md, and lg sizes. I remember talking with @mcarrano about it, and thinking those numbers were somewhat random, and it seeming unlikely those sizes would match a user's actual use case. 

@christiemolloy @mcarrano wdyt?

kmcfaul:
Right now it checks if the passed in value is one of the preset classes, if it isn't it will assign the custom height property.

tlabaj:
@kmcfaul based on what @mcoker said I think we should remove the logic for the presets.  It complicates things for no benefit.

mcoker:
@tlabaj @kmcfaul I think for the percentage based modifiers that makes sense, though I'm not sure about the other size modifiers. I'll bring it up in our morning meeting and we can chat about whether we want to continue to offer those.

packages/react-core/src/components/Skeleton/examples/Skeleton.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Percentage heights
+
+```js
+import React from 'react';
+import { Skeleton } from '@patternfly/react-core';
+
+<div style={{ height: '400px' }}>
+  <Skeleton height="25%" style={{ width: '15%', display: 'inline-block', marginRight: '4px' }} />
```
jschuler, line 48: why not use the `width` prop? Also i'd remove the inline styles from the Skeleton components and move them to the parent just so it's a bit cleaner, something like
```
<div style={{ height: '400px', display: 'flex', alignItems: 'flex-end', justifyContent: 'space-between' }}>
  <Skeleton height="25%" width="15%" />
  <Skeleton height="33%" width="15%" />
  <Skeleton height="50%" width="15%" />
  <Skeleton height="66%" width="15%" />
  <Skeleton height="75%" width="15%" />
  <Skeleton height="100%" width="15%" />
</div>;
```

packages/react-core/src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const Skeleton: React.FunctionComponent<SkeletonProps> = ({
+  className = '',
+  width = '',
+  height = '',
+  fontSize,
+  shape,
+  ...props
+}: SkeletonProps) => {
+  const fontHeightClassName = Object.values(styles.modifiers).find(key => key === `pf-m-text-${fontSize}`);
```
jschuler, line 26: can this be a function so it is not computed unless fontSize is passed in

packages/react-core/src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  /** The height of the Skeleton. Must specify pixels or percentage. */
+  height?: string;
+  /** The font size height of the Skeleton */
+  fontSize?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
+  /** The shape of the Skeleton */
+  shape?: 'circle' | 'square' | 'rectangle';
+}
+
+export const Skeleton: React.FunctionComponent<SkeletonProps> = ({
+  className = '',
```
tlabaj, line 19: this can default to undefined.

packages/react-core/src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  width = '',
```
tlabaj, line 20: Can the width and height default to undefined?

tlabaj (changes requested):
This looks great.  The API is much cleaner. A couple of small  comments.

packages/react-core/src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  /** Additional classes added to the Skeleton */
+  className?: string;
+  /** The width of the Skeleton. Must specify pixels or percentage. */
+  width?: string;
+  /** The height of the Skeleton. Must specify pixels or percentage. */
+  height?: string;
+  /** The font size height of the Skeleton */
+  fontSize?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
+  /** The shape of the Skeleton */
+  shape?: 'circle' | 'square' | 'rectangle';
```
mcoker, line 15: do you need rectangle?

jschuler (approved):

tlabaj (approved):
LGTM

mcoker (approved):
Looks great, thanks @kmcfaul. One small nit that I don't think should hold up this PR is that all of the skeletons are percentage based. I think it would be useful to show how you can use other units there - say pixels for width and height, and em/rem values for height specifically. This would also allow you to remove the inline heights from the parents of skeletons with percentage based heights, so it's a little more clear that only skeletons with a percentage based height need the parent to have a declared height.

PR #4761, opened 2020-09-01
feat(Card): add ability to customize card title and support heading element

### LINKED ISSUE #4756: Card Header should support heading element or show more accessible example
Currently our card header is a `<div>` in order to allow flexibility for whatever the user would want to include inside the card header. However, using heading elements can be beneficial for accessibility as it makes it easier to navigate for screen reader users (both in the headings menu in the rotor menu or when using keyboard shortcuts to navigate by heading in screen readers). I think it's worth considering adding the ability to pass a component prop to determine whether the header should be a div, heading element, etc. (Similar to how we handle card, button, etc.) We could also consider making the card header a heading element by default. Or at the very least, we should probably include an example of a card with a header that is a heading element instead of only including examples of divs. 

If you navigate to [card examples](https://www.patternfly.org/v4/documentation/react/components/card), you can see that all examples only have headers that are `<div>`s. If you open up VO (command + F5, then open the rotor menu with control + option + u, use the right and left arrow keys to get to the headings menu), you can see that the card headers are not in the headings menu. 

There is a workaround in which you can put a title within the card header like `<CardTitle><Title headingLevel="h2">Header</Title></CardTitle>` or something similar like `<Text component={TextVariants.h2}>Second level</Text>`, but this may not be obvious to most people. It may make more sense to add a component prop and enable the ability to do something like this: `<CardTitle component="h2">Header</CardTitle>`.

jessiehuff:
**What**: Closes #4756

### DIFF
--- a/packages/react-core/src/components/Card/CardTitle.tsx
+++ b/packages/react-core/src/components/Card/CardTitle.tsx
@@ -7,7 +7,7 @@ export interface CardTitleProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   /** Additional classes added to the CardTitle */
   className?: string;
-  /** Sets the base component to render. defaults to button */
+  /** Sets the base component to render. defaults to div */
   component?: keyof JSX.IntrinsicElements;
 }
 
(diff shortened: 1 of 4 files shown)


redallen (approved):

tlabaj (approved):
LGTM

PR #4771, opened 2020-09-02
feat(Drawer): add bottom panel variation

### LINKED ISSUE #4714: Bottom panel drawer variation
https://github.com/patternfly/patternfly/pull/3408

Adds `.pf-m-panel-bottom` as a feature of the drawer that expands the drawer panel from the bottom. It doesn't currently work with the `pf-m-inline` variation.

kmcfaul:
**What**: Closes #4714

### DIFF
--- a/packages/react-core/src/components/Drawer/Drawer.tsx
+++ b/packages/react-core/src/components/Drawer/Drawer.tsx
@@ -14,7 +14,7 @@ export interface DrawerProps extends React.HTMLProps<HTMLDivElement> {
   /** Indicates if the drawer will always show both content and panel. */
   isStatic?: boolean;
   /** Position of the drawer panel */
-  position?: 'left' | 'right';
+  position?: 'left' | 'right' | 'bottom';
   /** Callback when drawer panel is expanded after waiting 250ms for animation to complete. */
   onExpand?: () => void;
 }
@@ -49,6 +49,7 @@ export const Drawer: React.SFC<DrawerProps> = ({
         isInline && styles.modifiers.inline,
         isStatic && styles.modifiers.static,
         position === 'left' && styles.modifiers.panelLeft,
+        position === 'bottom' && styles.modifiers.panelBottom,
         className
       )}
       {...props}
--- a/packages/react-core/src/components/Drawer/__tests__/Drawer.test.tsx
+++ b/packages/react-core/src/components/Drawer/__tests__/Drawer.test.tsx
@@ -42,3 +42,28 @@ Object.values([
     expect(view).toMatchSnapshot();
   });
 });
+
+test(`Drawer expands from bottom`, () => {
+  const panelContent = (
+    <DrawerPanelContent>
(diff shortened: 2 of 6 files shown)


kmcfaul:
@mcoker It seems that at large screen widths, the hidden panel is still creating space at the bottom of the panel, any ideas what is causing this? I'm not sure if I have to do something on react differently than other drawer variations beyond applying the modifier.

mcoker:
@kmcfaul the bottom variation requires the drawer's parent element have a `height`, either explicitly via a `height: some-value;` declaration, or an implicit height via `flex-basis` or something like that.

So maybe wrap the `.pf-c-drawer` (under the button that toggles the drawer) in a workspace class that you can add something like `height: 450px` to. That's what we did with the core examples - https://www.patternfly.org/v4/documentation/core/components/drawer#expanded-panel-on-bottom

Also, can a user pass a custom size/width to the panel, of say "150px"?

kmcfaul:
@mcarrano That's the issue I'm having with the container div / styling. It incorrectly creates a white space for the hidden drawer, and when the drawer opens, it expands upwards, pushing out of the container because of the spacing.

kmcfaul:
@mcarrano @mcoker 

Added a height to the container div of the bottom panel example, it should look better now!

mcoker:
@kmcfaul looks a lot better but the workspace buttons are getting cut off

Screen Shot 2020-09-04 at 2 37 32 PM

The drawer itself needs to be wrapped in an element with a height. Can you move the button out so it sits on top of the element with the height?

mcoker (approved):
LGTM! Thanks @kmcfaul! I found a bug, but it's a core issue and doesn't need a react follow up. https://github.com/patternfly/patternfly/pull/3460

mcarrano (commented):
Looks much better @kmcfaul . My only remaining question is why when the drawer closes, it does not have a nice easing effect like it does on opening.

mcoker:
@mcarrano are you looking at this example? https://patternfly-react-pr-4771.surge.sh/documentation/react/components/drawer/panel-on-bottom/

I tested in firefox and chrome and it looks OK to me. It uses the same basic transition that the left/right variations use.

mcarrano:
That's not what I'm seeing @mcoker . I made a [video screen capture here](https://drive.google.com/file/d/1SsVCQVi2aYro8jlWNp2TpQz4bMDMACrE/view?usp=sharing) that shows the animation of the bottom vs the side drawer.

mcoker:
@mcarrano I think it's because of the extra white space in the example. Here I made a full-page example - does this look like you'd expect it to? 
Sep-08-2020 15-36-55

mcarrano:
I guess that's about the same as what I'm seeing @mcoker . Maybe it's just me, but it feels like the downward animation is faster than the upward. @mceledonia can you also take a look at this and let us know what you think? It will be easier to evaluate in a full page demo I think.

PR #4772, opened 2020-09-02
feat(Select): Added support for Divider in Select

### LINKED ISSUE #4641: Select:  Add support for divider in the grouped select

tlabaj:
**What**: Closes #4641 , #4890

See the "Single",  "Grouped single" and "Checkbox input" , "  and "Grouped typeahead"  to see the divider preview.

A check was also added to `componentDidUpdate` so that createRenderableFavorites is only called if onFavorite is provided.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -352,8 +352,8 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
         if (group.type === SelectGroup) {
           return React.cloneElement(group, {
             titleId: group.props.label && group.props.label.replace(/\W/g, '-'),
-            children: React.Children.map(group.props.children, (child: React.ReactElement) => {
-              return child.type === Divider
+            children: React.Children.map(group.props.children, (child: React.ReactElement) =>
+              child.type === Divider
                 ? child
                 : React.cloneElement(child as React.ReactElement, {
                     isFocused:
@@ -362,8 +362,8 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                         (this.props.isCreatable &&
                           typeaheadActiveChild.innerText ===
                             `{createText} "${(child as React.ReactElement).props.value}"`))
-                  });
-            })
+                  })
+            )
           });
         } else {
           return React.cloneElement(group, {
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -209,8 +209,9 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
     }
 
     if (
-      this.props.favorites.length !== prevProps.favorites.length ||
-      this.state.typeaheadFilteredChildren !== prevState.typeaheadFilteredChildren
+      this.props.onFavorite &&
+      (this.props.favorites.length !== prevProps.favorites.length ||
+        this.state.typeaheadFilteredChildren !== prevState.typeaheadFilteredChildren)
     ) {
       const tempRenderableChildren =
         this.props.variant === 'typeahead' || this.props.variant === 'typeaheadmulti'
@@ -337,7 +338,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
   };
 
   extendTypeaheadChildren(typeaheadCurrIndex: number, favoritesGroup?: React.ReactNode[]) {
-    const { isGrouped } = this.props;
+    const { isGrouped, onFavorite } = this.props;
     const typeaheadChildren = favoritesGroup
       ? favoritesGroup.concat(this.state.typeaheadFilteredChildren)
       : this.state.typeaheadFilteredChildren;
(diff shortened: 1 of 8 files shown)


packages/react-core/src/components/Select/examples/Select.md
```diff
@@ -271,6 +271,7 @@ class GroupedSingleSelectInput extends React.Component {
         <SelectOption key={3} value="Degraded" />
         <SelectOption key={4} value="Needs Maintenence" />
       </SelectGroup>,
+      <Divider/>,
```
mcoker, line 274: Would you use it by placing `<Divider/>` like this in the menu? Curious if this should align with the dropdown's `<DropdownSeparator>`. Also is there an example of how to include a divider between regular select items that are either in a single group, or in a single list with no groups?

tlabaj: By align do you mean create a component `<SelectSeparator>`?  If so that is what I started with.  I thought about it more and decided against it.  It would just be a wrapper.  I think we kept `<SelectSeparator>` once the dropdown was refactored to use Divider in core in order to minimize breaking changes.  I was thinking in our next breaking change release we could update Dropdown and any other components that wrap the Divider.   I have no strong feeling either way.

I can added a a divider to the  "Checkbox input" example.

jenny-s51 (dismissed):

packages/react-core/src/components/Select/examples/Select.md
mcoker: OK that makes sense, thanks for the clarification. I wasn't sure if having a select specific component would allow us to do smarter detection of where the divider is placed as to which element the divider should render as. In the grouped and checkbox examples, it's fine for the divider to render as an `<hr>` (the default). But in a regular list where each item is an `<li>`, the divider would need to be an `<li>`, too. Do you think we should include an example of that to illustrate that the divider element needs to have `component="li"`?

nicolethoen (commented):
It appears to me that it's breaking when I try to use the divider with the grouped typeahead example.

mcoker (approved):
lgtm!

nicolethoen (approved):

nicolethoen (changes requested):
I cannot navigate the grouped typeahead with the divider using the keyboard

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -353,36 +353,44 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
...
-        isFocused:
-          typeaheadActiveChild &&
-          (typeaheadActiveChild.innerText === (child as React.ReactElement).props.value.toString() ||
-            (this.props.isCreatable &&
-              typeaheadActiveChild.innerText === `{createText} "${(child as React.ReactElement).props.value}"`))
-      })
-    );
+    return typeaheadChildren.map((child: React.ReactNode) => {
+      const childElement = child as any;
+      return childElement.type.displayName === 'Divider'
```
redallen, line 384: Can you cast to a ReactNode instead and then check `childElement.type` like you do for all the other component logic?Like this: `(childElement as React.ReactNode).type === Divider`?

packages/react-core/src/components/Select/examples/Select.md
```diff
@@ -15,19 +15,20 @@ import CubeIcon from '@patternfly/react-icons/dist/js/icons/cube-icon';
...
-      { value: 'Dr', disabled: false },
-      { value: 'Other', disabled: false }
+       <SelectOption key={0} value="Choose..." isPlaceholder/>,
+        <SelectOption key={1} value="Mr" />,
+        <SelectOption key={2} value="Miss" />,
+        <SelectOption key={3} value="Mrs" />,
+        <SelectOption key={4} value="Ms" />,
+        <Divider component="li" key={5} />,
+        <SelectOption key={6} value="Dr" />,
+        <SelectOption key={7} value="Other" />
```
redallen, line 31: Thanks for this!

redallen (changes requested):

packages/react-core/src/components/Select/Select.tsx
tlabaj: That is what I tried first.  It did not like that.  Elsewhere the child is a ReactElement so I can check the `type`.

redallen (approved):
I don't like how complicated Select is becoming, but LGTM.

jessiehuff (commented):
Looks pretty good! It seems to work with keyboard and VO now. However, I did see an odd error on the Grouped typeahead example: 
Screen Shot 2020-09-29 at 4 09 26 PM

To get this error, I tabbed to the grouped typeahead example, selected "isCreatable", went into the select and pressed enter to choose an option. Then I tried to edit that option in the input field. I don't seem to get this error on org when I try it.

jessiehuff (approved):
LGTM!

nicolethoen (approved):

mcoker (approved):
LGTM!

mcarrano (approved):
Looks good to me!

Issue #3456, opened 2020-01-07
Select component change for Openshift Search Toolbar. Display count rather than selections.


jcaianirh:
**Describe the issue. What is the expected and unexpected behavior?**
For Openshift Search Toolbar, there is a design that requires the selections for the multi-select type-ahead select component to display only the count of selections rather than values. 
**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

**What is your product and what release version are you targeting?**

janwright73:
Related design for this issue: http://openshift.github.io/openshift-origin-design/admin-perspective/future-openshift/search/search

PR #4775, opened 2020-09-02
Add LabelGroup component

### LINKED ISSUE #3829: Labels should be group-able like chip groups and link-able
Chip Groups allow for grouping of chips.  I'd like the same features for labels, that allow group of labels, that also serve as links. 

Enhancement

Ansible Tower

evwilkin:
**What**: Closes #3829 

This PR adds the `LabelGroup` component, and ads a new `isOverflowLabel` prop to the `Label` component.
Much of the functionality and code was borrowed from the ChipGroup which has similar functionality.


mcarrano (dismissed):
This looks good @evwilkin with one exception. In the vertical group, looks like the close button is being placed at the bottom and should be in the upper right. This actually might be a core problem and I know we've recently been having some discussion relative to wrapping chip groups. @mcoker I think that the same issues apply here. Must have missed this in the HTML review.

mcoker:
@mcarrano I agree the position of the close button in the vertical group feels awkward, but I talked with Mary about it and mocked up the close button at the top and bottom, and she advised me to put it at the bottom. Do we want to change that?

mcarrano:
OK. Let's discuss then @maryshaks1996 .  My concern is this is something different than what we are doing with the chip group.

mcarrano:
@mcoker I just discussed this with @maryshaks1996 and we agreed to move the close button for the vertical groups to the top-right to be consistent with what we do it Chip groups, alerts, etc. In a sense the vertical group and a horizontal group that wraps would look identical if you made the container narrow enough. Is that a change we can make it core?

mcoker:
@mcarrano yep, would you mind creating an issue for that? Do you want to try and get it in this sprint?

mcarrano:
@mcoker I've opened https://github.com/patternfly/patternfly/issues/3456. Yes, it would be great to get this into the current sprint if possible. Thanks!

evwilkin:
@mcoker @mcarrano if we can get that change in early next week I can update this PR to include that - I'll leave this as-is for now pending that update.

packages/react-integration/cypress/integration/labelgroup.spec.ts
```diff
@@ -0,0 +1,28 @@
+describe('Chip Group Demo Test', () => {
```
mcoker, line 1: there's a bunch of stuff in here labeled 'chip' - should it be 'label'?

mcoker:
@evwilkin LGTM! I think we can merge this when it's done and then make any updates needed for https://github.com/patternfly/patternfly/issues/3456 if necessary since it's still beta. Also I'm not sure if there will be any react update from #3456

mcoker (dismissed):
🏆 👏

packages/react-integration/cypress/integration/labelgroup.spec.ts
evwilkin: Thanks @mcoker good catch - I've updated the integration tests for this component

mcoker (dismissed):

jessiehuff (commented):
For the most part it looks pretty good! 🙂 Keyboard interactions work as expected, and it sounds good in VO. Just one small nit that the `aria-labelledby` in the last example (on the `<ul>`) doesn't actually point to any valid id. It seems that that id is missing? I think it should be on the span. If you run axe, it'll flag this one to point out where I mean.

Also, I see another axe error that I've been debating over with chip group as it exists there as well. It's a little odd to have an `aria-hidden` on the same element that we have a tooltip on, but realistically it doesn't cause any issues here so it's probably fine. That's not specific to your PR, but thought it might be worth noting.

packages/react-core/src/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -0,0 +1,119 @@
+---
+id: Label group
+section: components
+cssPrefix: pf-c-label
+propComponents: ['LabelGroup', 'Label']
+isBeta: true
```
tlabaj, line 6: I am not sure why this is not showing as Beta in the preview link.

packages/react-integration/demo-app-ts/src/components/demos/LabelGroupDemo/LabelGroupWithCategoryDemo.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { Component } from 'react';
```
tlabaj, line 1: Can you add the new `isOverflowLabel` to the demo app.
Can you also make sure all the props for the `LabelGroup` are used at least once in the demo to verify typing.

packages/react-core/src/components/LabelGroup/LabelGroup.tsx
```diff
@@ -0,0 +1,185 @@
+import * as React from 'react';
```
tlabaj, line 1: So you the `Label` and `LabelGroup` in separate directories here.  @mcarrano  know y had an opinion about the `Chip` and `ChipGroup` being placed in directory  and therefore on the same page for documentation.  Do you feel that same way for the `Label` and `LabelGroup`?

tlabaj (changes requested):

packages/react-core/src/components/LabelGroup/examples/LabelGroup.md
evwilkin: Fixed - corrected from `isBeta` to `beta`

packages/react-core/src/components/LabelGroup/LabelGroup.tsx
evwilkin: Yeah I could see it going either way - combining to mirror Chips/ChipGroup, or staying separate as users may have already been using Labels and be thrown off seeing the docs for them disappear (replaced by LabelGroup). @mcarrano happy to defer to your feedback on this, currently they're listed separately.

evwilkin:
@jessiehuff thanks for taking a look! I'm not able to recreate the first issue you mentioned, I am seeing the `<span>` just before the `<ul>` has the correct ID as expected, and axe is not throwing the error for me - are you still seeing this?

Screen Shot 2020-09-11 at 1 34 38 PM

packages/react-core/src/components/LabelGroup/LabelGroup.tsx
mcarrano: I could argue either way, but I agree with your reasoning for keeping Label as a thing on it's own. It would be potentially confusing if that went away. I also think that there are many instances when labels will be used by themselves where chips would rarely be used outside of a group. So let's keep separate.

jessiehuff (dismissed):
Oh weird! Maybe the preview hadn't updated, it's not there anymore. LGTM! 🙂

kmcfaul (approved):

packages/react-core/src/components/LabelGroup/LabelGroup.tsx
evwilkin: Thanks!

evwilkin:
@tlabaj updated tests & Beta flag, and per mcarrano above keeping Label separate from LabelGroup as-is. @kmcfaul @jessiehuff if you don't mind re-reviewing, thanks!

tlabaj (approved):
LGTM

mcoker (approved):
👍👍

jessiehuff (approved):

mcarrano (approved):
I am good with this. @evwilkin so is the understanding then that we will make a core update in the next release then to fix the close button positioning issue?

Issue #3969, opened 2020-03-24
Sortable table columns don't allow for interactive children


rebeccaalpert:
**Describe the issue. What is the expected and unexpected behavior?**
Sortable table columns don't support children with popovers and tooltips.

We're looking to add an icon with a popover to a sortable table header cell:
<img width="592" alt="Screen Shot 2020-03-24 at 4 50 30 PM" src="https://user-images.githubusercontent.com/7014965/77475710-9903dd00-6def-11ea-9ab9-871e1a24a806.png">

Sortable header cells currently wrap children in a button. Tooltips and popovers don't play well with this (it breaks their behavior).

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
This is a bug.

**What is your product and what release version are you targeting?**
A design was proposed for OpenShift that requires this functionality.

mcarrano:
@rebeccaalpert I don't know that I'd call this a bug. It was never intended to work this way by design. @mcoker what would it take to do something like this? Would we require a core change to separate the sort and the info icon into separate button actions?

mcoker:
ping @mceledonia we were chatting about this recently and he was investigating the design.

rachael-phillips:
Hey @rebeccaalpert and @mcoker what is the status of this issue? https://github.com/patternfly/patternfly-react/issues/3969

rebeccaalpert:
@rachael-phillips I haven't heard anything since @mcoker's comment 25 days ago.

mceledonia:
I remember discussing this and we got stuck on refactoring the structure of the table headers to accommodate this. If this isn't something we want to natively support (@mcarrano) we should probably recommend an alternate solution like an override or perhaps a design workaround like having the whole button always trigger the tooltip on hover.

rebeccaalpert:
It looks like this has been added to Core and can be brought into React: https://github.com/patternfly/patternfly/pull/3189

rebeccaalpert:
This is going to require some modification to the way the table processes transforms (sort, etc.) for the sort-with-help case. I took a look at it on Friday with Nicole, but may not have time to take another look this week.

PR #4778, opened 2020-09-03
header cell information tooltip/popover

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/3969

Example:
https://patternfly-react-pr-4778.surge.sh/documentation/react/components/table#basic
(First help button activates a tooltip on hover over, second help button activates a popover on click)

Tooltip variant:
Screen Shot 2020-09-03 at 1 30 20 PM

Popover variant:
Screen Shot 2020-09-03 at 1 30 42 PM

### DIFF
--- a/packages/react-core/src/components/Popover/index.ts
+++ b/packages/react-core/src/components/Popover/index.ts
@@ -1,1 +1,1 @@
-export { Popover, PopoverPosition } from './Popover';
+export { Popover, PopoverPosition, PopoverProps } from './Popover';
--- a/packages/react-core/src/components/Tooltip/index.ts
+++ b/packages/react-core/src/components/Tooltip/index.ts
@@ -1,1 +1,1 @@
-export { Tooltip, TooltipPosition } from './Tooltip';
+export { Tooltip, TooltipPosition, TooltipProps } from './Tooltip';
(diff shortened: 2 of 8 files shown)


rebeccaalpert:
This doesn't cover the sort-with-help case the issue was opened to address.

jschuler:
@rebeccaalpert it should work anywhere, this is a general solution for any header cell. If you live-edit the sortable table example and add one of these in the column definition it works fine
```
columns: [
        { title: 'Repositories', transforms: [sortable], header: {
            info: {
              tooltip: 'More information about repositories',
              className: 'repositories-info-tip',
              tooltipProps: {
                isContentLeftAligned: true
              }
            }
          } },
        'Branches',
        { title: 'Pull requests', transforms: [sortable] },
        'Workspaces',
        'Last Commit'
      ],
```

nicolethoen (approved):
The approach is awesome. Better than the road I had begun when trying to frame this problem :)

Someday, not part of this PR, we should try to figure out how to really document the API for Table so all these features/props/capabilities are explained as clearly and completely as we can. The docs don't actually offer any insight into your addition beyond the example which shows it in action.

jschuler:
@nicolethoen I absolutely agree, I think it is critical we start to add documentation to the Table API, I have created an issue for this https://github.com/patternfly/patternfly-react/issues/4793

mcoker (approved):
LGTM, thanks @jschuler. Found a bug in core that I'd like to get in with this feature, though it's a separate issue so we can merge this PR if you'd like. Here's that PR - https://github.com/patternfly/patternfly/pull/3462

rebeccaalpert (approved):
Smart way to get around the transforms limitations! Looks good to me.

PR #4781, opened 2020-09-03
fix(page): hide PageSideBar onClick on mobile when isManagedSidebar

### LINKED ISSUE #4480: On mobile the navigation toolbar only disappears when the "kebab" is clicked again
When a user on mobile uses the page navigation to navigate to a particular "page" the navigation bar does not hide after the navigation is complete. Nor does it disappear when the "page" that is in the background is clicked. The only way to get rid of it is by clicking the kebab item again. This is not very user friendly and not on par with other mobile websites.

redallen:
**What**: Closes #4480. Fixes resize event listener never detaching.

**Additional issues**: Doesn't close when nav item clicked. I think that's a bit too opinionated since designs can use side navs for more than just navigation. It's also tricky to code, but can be achieved consumer-side.

### DIFF
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import ReactDOM from "react-dom";
 import styles from '@patternfly/react-styles/css/components/Page/page';
 import { css } from '@patternfly/react-styles';
 import globalBreakpointXl from '@patternfly/react-tokens/dist/js/global_breakpoint_xl';
@@ -126,10 +125,9 @@ export class Page extends React.Component<PageProps, PageState> {
     }
   }
 
-  isMobile = () => {
+  isMobile = () =>
     // eslint-disable-next-line radix
-    return window.innerWidth < Number.parseInt(globalBreakpointXl.value, 10);
-  }
+    window.innerWidth < Number.parseInt(globalBreakpointXl.value, 10);
 
   handleResize = debounce(() => {
     const { onPageResize } = this.props;
@@ -147,7 +145,7 @@ export class Page extends React.Component<PageProps, PageState> {
         this.setState({ mobileIsNavOpen: false });
       }
     }
-  }
+  };
 
   onNavToggleMobile = () => {
     this.setState(prevState => ({
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import ReactDOM from "react-dom";
 import styles from '@patternfly/react-styles/css/components/Page/page';
 import { css } from '@patternfly/react-styles';
 import globalBreakpointXl from '@patternfly/react-tokens/dist/js/global_breakpoint_xl';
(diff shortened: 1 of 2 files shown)


dtaylor113:
Hi @redallen, do you know how I could test the PR against OpenShift?   I assume this would be in an upcoming pre-release, so I could test it then, would be nice to do before this is merged though.

redallen:
You could checkout this PR, build it, and symlink all the `@patternfly` packages under openshift's `node_modules`.

jenny-s51 (dismissed):
Tested on mobile - LGTM

packages/react-core/src/components/Page/Page.tsx
```diff
@@ -107,7 +108,9 @@ export class Page extends React.Component<PageProps, PageState> {
   componentDidMount() {
     const { isManagedSidebar, onPageResize } = this.props;
     if (isManagedSidebar || onPageResize) {
-      window.addEventListener('resize', debounce(this.handleResize, 250));
+      window.addEventListener('resize', this.handleResize);
+      document.addEventListener('mousedown', this.handleClickMobile);
```
jschuler, line 112: these could be attached to the ref, something like
```suggestion
      this.outerRef && this.outerRef.current && this.outerRef.current.addEventListener('mousedown', this.handleClickMobile);
```

packages/react-core/src/components/Page/Page.tsx
```diff
@@ -116,19 +119,32 @@ export class Page extends React.Component<PageProps, PageState> {
...
+    const mobileView = this.isMobile();
     if (onPageResize) {
-      onPageResize({ mobileView, windowSize });
+      onPageResize({ mobileView, windowSize: window.innerWidth });
     }
     this.setState({ mobileView });
+  }, 250);
+
+  handleClickMobile = (ev: any) => {
+    if (this.isMobile() && this.outerRef.current) {
```
jschuler, line 142: can also add an additional check here 
```suggestion
    if (this.isMobile() && !this.state.mobileIsNavOpen && this.outerRef.current) {
```

packages/react-core/src/components/Page/__tests__/Generated/__snapshots__/Page.test.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`Page should match snapshot (auto-generated) 1`] = `
   value={
     Object {
       "isManagedSidebar": false,
-      "isNavOpen": false,
+      "isNavOpen": true,
```
jschuler, line 8: did something else happen that i missed

jschuler (changes requested):

packages/react-core/src/components/Page/Page.tsx
redallen: Sure, although we lose this feature for clicking below pages that don't take up `100vh`.

packages/react-core/src/components/Page/__tests__/Generated/__snapshots__/Page.test.tsx.snap
redallen: The first `handleResize` was debounced 250ms. I've fixed it.

jschuler (approved):

jenny-s51:
👍

jenny-s51 (approved):

jschuler (approved):

PR #4792, opened 2020-09-08
feat(notificationDrawer): add default variation to notification items

### LINKED ISSUE #4659: Notification drawer: add default variation to notification items
https://github.com/patternfly/patternfly/pull/3338

adds a `.pf-m-default` variation to `.pf-c-notification-drawer__list-item`

boaz0:
**What**: Closes #4659 

//cc @mcoker

### DIFF
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawerListItem.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawerListItem.tsx
@@ -16,7 +16,7 @@ export interface NotificationDrawerListItemProps extends React.HTMLProps<HTMLLIE
   /**  Tab index for the list item */
   tabIndex?: number;
   /**  Variant indicates the severity level */
-  variant?: 'success' | 'danger' | 'warning' | 'info';
+  variant?: 'default' | 'success' | 'danger' | 'warning' | 'info';
 }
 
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/NotificationDrawer/NotificationDrawerListItem.tsx
```diff
@@ -27,7 +27,7 @@ export const NotificationDrawerListItem: React.FunctionComponent<NotificationDra
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onClick = (event: React.MouseEvent) => undefined as any,
   tabIndex = 0,
-  variant,
+  variant = 'default',
```
mcoker, line 30: Not sure if this is a breaking change? I would consider it a bug fix as items aren't designed to work without a variant set. Also this will now align with the alert's default variant. Curious what others think.

tlabaj: This is also still a beta component so it gives us flexibility with the API.

tlabaj (changes requested):
Can you just add the variant to the demo app as well please.

boaz0:
@tlabaj I don't know why I keep forgetting this everytime :disappointed:
I will be working on this. Thanks again!

tlabaj (approved):
Thank you!  looks good.

redallen (approved):
Thanks @boaz0 !

PR #4794, opened 2020-09-08
chore(packages): bump security deps and regenerate lockfile

redallen:
**What**: Closes some security vulnerabilities by bumping packages. Unfortunately, also bumps React's types which led to some type errors caused by React's types for `React.Children` which enforce us to be more honest with type casting.

### ~~Component type changes~~
- ~~Tabs: `children: React.ReactNode` -> `children: React.ReactElement<TabProps>[]` (which is what we assume in the component logic for rendering)~~
- ~~Select: `children?: React.ReactElement[]` -> `children?: React.ReactElement<SelectGroupProps>[]` (which again we assume in the component logic)~~
- ~~SelectGroup:  `children?: React.ReactElement[]` -> `children?: React.ReactElement<SelectGroupProps>[]` (which is needed in Select's nested rendering logic)~~

I removed the type changes by casting them. These can be added in another PR.

**Additional issues**: Updated linters added some stricter rules so there are formatting changes in here as well.


packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -17,7 +18,7 @@ export enum TabsComponent {
 
 export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivElement>, 'onSelect'>, OUIAProps {
   /** Content rendered inside the tabs component. */
-  children: React.ReactNode;
+  children: React.ReactElement<TabProps>[];
```
dlabrecq, line 21: Looks ok. Don't know if this is considered a breaking change or not?

dlabrecq (approved):
Looks ok, but not clear if the type changes are breaking or not? 

Lot of code reformatting, which makes it difficult to see the real changes.

packages/react-core/src/components/Tabs/Tabs.tsx
redallen: Considering if you pass anything but `<Tab>` as a child of `<Tabs>` it doesn't render correctly I don't think it is. I get a weird spacer [when I pass a `<div>`](https://codesandbox.io/s/icy-feather-1n4fu?file=/index.js):
image

ddonahue007 (approved):
LGTM

packages/react-core/src/components/Tabs/Tabs.tsx
tlabaj: I disagree Zack.  I think this would be a API breaking change regardless of how this was rendered.  Someone may be passing something other than`<Tabs>` and fixing the rendering with CSS.  This change would break their build.

tlabaj (changes requested):

packages/react-core/src/components/Tabs/Tabs.tsx
redallen: I just checked through all the [packages in our analytics](https://github.com/rh-uxd/patternfly-analytics) and all of them use `<Tabs><Tab /></Tabs>`. Not doing so is what I'd consider undefined behavior. This is what renders:
```html
<li class="pf-c-tabs__item">
  <button class="pf-c-tabs__link" id="pf-tab-undefined-pf-15997686294791nq1760spxn" aria-controls="pf-tab-section-undefined-pf-15997686294791nq1760spxn"></button>
</li>
```

I'm not sure how or why you could save that with CSS. If you pass anything other than a `Tab` our rendering ignores what you pass and outputs a weird `<li><button>`! If I'm a typescript consumer, having a correct type that tells me this would be very welcome.

dlabaj: What if you want to make a custom tab to use inside of tabs.  I was thinking someone may want to do that and this could potential be a breaking change if they didn't use TabProps interface when they built their tab.

redallen: Yes, if a Typescript consumer did that they would have to be explicit that their implementation matches `<Tab>`'s. For example, Kiali [already casts to `React.ReactElement<TabProps>`.](https://github.com/kiali/kiali-ui/blob/master/src/components/Tab/Tabs.tsx#L46)

As it is if we add props to `Tab` we will break these apps (and they _should_ break since they depend on our internals)!

tlabaj (approved):
LGTM

ddonahue007 (approved):
LGTM

PR #4796, opened 2020-09-09
fix(tabs): conditionally render aria-controls when mountOnEnter

### LINKED ISSUE #4785: Fix aria-controls on Tab example
Since the tabs mounting on click example only renders the tab content when the tabs are clicked, the tab content id is not available to match up to the tab `aria-controls` attribute. I'm wondering if we should be including this attribute as it's only really [supported by JAWs](https://a11ysupport.io/tech/aria/aria-controls_attribute) and doesn't seem to be well supported at that. Another idea is to conditionally render this attribute. It results in the following axe issue:
Screen Shot 2020-09-04 at 2 07 59 PM

Run axe on the following example: https://www.patternfly.org/v4/documentation/react/components/tabs#children-mounting-on-click

Bug

boaz0:
**What**: Closes #4785 

Hey @jessiehuff I took the initiative and conditionally render "aria-controls" .
I also updated its spec test to make sure it is working properly.

Would you mind looking? :pray: 

Thanks.

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -262,7 +262,12 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
                   ouiaId: childOuiaId,
                   ...rest
                 } = child.props;
-
+                let ariaControls = tabContentId
+                  ? `${tabContentId}`
+                  : `pf-tab-section-${eventKey}-${childId || uniqueId}`;
+                if ((mountOnEnter || unmountOnExit) && eventKey !== activeKey) {
+                  ariaControls = undefined;
+                }
                 return isHidden ? null : (
                   <li
                     key={index}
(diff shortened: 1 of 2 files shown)


jessiehuff (commented):
Thanks so much for jumping on this!!!! 😄  🎉  
For some reason I'm still seeing `aria-controls` in the preview that point towards sections that haven't been rendered yet. It's also still showing the axe errors (it says "ARIA attributes must conform to valid values" because it's technically pointing to something that doesn't exist at that point in time in the DOM). Perhaps we need to tweak the logic?

boaz0:
Thanks @jessiehuff !
I forgot to do the same thing on unmount. I fixed the problem and ran axe-analysis to make sure this error is not shown.
Do you mind giving a second look?

Again - thanks a bunch! :bow:

jessiehuff (approved):
This looks awesome!!! Thanks, @boaz0!! 😄 🎉

tlabaj (commented):
LGTM

tlabaj (approved):

boaz0:
Thank you all for letting me work on this

PR #4801, opened 2020-09-10
feat(browserslist): adds @patternfly/browserslist-config package

sharvit:
new package to provide integration with browserslist

**What**:

The goal is to share the [supported browsers](https://www.patternfly.org/v4/get-started/about#supported-browsers) with consumers.
See: https://forum.patternfly.org/t/what-browser-are-supported-by-pf4/370/5

**Additional issues**:

1. How can we ensure the code **syntax** is compatible with the listed browsers?
> The typescript compiles the code to es2015 so it should be compatible with all modern browsers.

2. How can we ensure the code uses browsers APIs which are compatible with the listed browsers?
> 1. Use eslint-plugin-compat and configure eslint to throw errors when using unsupported browsers based on the browserslist config.
> 2. If the project would use `babel` it would be possible to use `@babel/preset-env` together with `core-js` to fill the missing gap with polyfills based on the browserslist config.

3. How can we ensure the 3rd-party modules are compatible with the supported browsers?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,124 +1,113 @@
 {
-    "env": {
-        "browser": true,
-        "node": true,
-        "es6": true
-    },
-    "plugins": [
-        "@typescript-eslint",
-        "jsdoc",
-        "react",
-        "prettier",
-        "patternfly-react"
-    ],
-    "extends": [
-        "eslint:recommended",
-        "plugin:@typescript-eslint/recommended",
-        "plugin:react/recommended",
-        "plugin:jsdoc/recommended"
-    ],
-    "parserOptions": {
-        "sourceType": "module",
-        "ecmaFeatures": {
-            "jsx": true
-        }
-    },
-    "settings": {
-        "react": {
-            "version": "16.4.0"
-        }
-    },
-    "globals": {
-        "describe": "readonly",
-        "test": "readonly",
-        "jest": "readonly",
-        "expect": "readonly",
-        "require": "readonly",
-        "global": "writable",
-        "it": "readonly",
-        "afterEach": "readonly",
-        "beforeEach": "readonly"
-    },
-    "rules": {
-        "@typescript-eslint/adjacent-overload-signatures": "error",
-        "@typescript-eslint/array-type": "error",
-        "@typescript-eslint/ban-types": "error",
-        "@typescript-eslint/camelcase": "off",
-        "@typescript-eslint/class-name-casing": "error",
-        "@typescript-eslint/consistent-type-assertions": "error",
-        "@typescript-eslint/consistent-type-definitions": "error",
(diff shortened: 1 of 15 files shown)


.eslintrc.json
```diff
@@ -1,124 +1,113 @@
...
+    "node": true,
+    "es6": true
+  },
+  "plugins": ["@typescript-eslint", "jsdoc", "react", "prettier", "patternfly-react", "compat"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react/recommended",
+    "plugin:jsdoc/recommended",
+    "plugin:compat/recommended"
```
redallen, line 13: Is it really necessary to add this to the eslint config? Most teams use `babel` or `tsc` or other build tools that take care of shims or polyfills for them.

packages/react-core/scripts/generate-supported-browsers-regex.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+DESTINATION_PATH=src/helpers/SupportedBrowsers/supportedBrowsersRegex.ts
+SUPPORTED_BROWSERS_REGEX=$(npx browserslist-useragent-regexp --allowHigherVersions)
+
+echo "export default $SUPPORTED_BROWSERS_REGEX;" > $DESTINATION_PATH
```
redallen, line 6: I like the idea of exporting `isBrowserSupported` only if we have a more permissive `browserslist` that isn't going to say two versions ago of Chrome is unsupported.

packages/browserslist-config/index.js
```diff
@@ -0,0 +1 @@
+module.exports = ['last 1 Chrome version', 'last 1 Firefox version', 'last 1 Safari version', 'last 1 Edge version'];
```
redallen, line 1: This should be the least-permissive set of focus-trap”: “4.0.2”, “react-dropzone”: “9.0.0”, and “tippy.js”: “5.1.2”. Do you have time to review these dependencies and make a more permissive set?

If you don't have the time, I recommend removing the `isSupportedBrowser` function since it will incorrectly return `false` for older versions.

redallen (changes requested):
Hey @sharvit !

The infrastructure team reviewed this PR and we found it rather heavy-handed to add an entirely new package for what could effectively be done by adding a `"browserslist": ['last 1 Chrome version', 'last 1 Firefox version', 'last 1 Safari version', 'last 1 Edge version']` to react-core/package.json. Consumers could easily access this via `const { browserslist } = require('@patternfly/react-core/package.json');` We're happy to take this in if you lighten up the PR a little.

redallen:
Closing due to staleness. Can open a new lightweight PR.

PR #4830, opened 2020-09-16
fix(charts,table): replace or use exact lodash imports

### LINKED ISSUE #4750: Remove lodash from patternfly
Recently we did an analysis on how many functions are used by patternfly from lodash.  There is less than a handful.  The size of lodash is larger then our entire component library if you do a bundle analysis.  We should remove this dependence to improve load times, reduce our foot print, and not increase consumers bundle size by including this library.

redallen:
**What**: Closes #4750. Reduces bundle sizes up to 500kb.

**Additional issues**:

### DIFF
--- a/packages/react-charts/src/components/ChartCursorTooltip/ChartCursorFlyout.tsx
+++ b/packages/react-charts/src/components/ChartCursorTooltip/ChartCursorFlyout.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Helpers, CommonProps, Path } from 'victory-core';
 import isPlainObject from 'lodash/isPlainObject';
-import assign from 'lodash/assign';
 
 const getVerticalPath = (props: any) => {
   const { pointerWidth, cornerRadius, orientation, width, height, center } = props;
@@ -90,7 +89,7 @@ const evaluateProps = (props: any) => {
   const id = Helpers.evaluateProp(props.id, props);
   const style = Helpers.evaluateStyle(props.style, props);
 
-  return assign({}, props, { id, style });
+  return Object.assign({}, props, { id, style });
 };
 
 const ChartCursorFlyout = (props: any) => {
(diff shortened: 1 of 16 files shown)


packages/react-charts/src/components/ChartCursorTooltip/ChartCursorFlyout.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Helpers, CommonProps, Path } from 'victory-core';
-import { isPlainObject, assign } from 'lodash';
+import isPlainObject from 'lodash/isPlainObject';
+import assign from 'lodash/assign';
```
KKoukiou, line 5: You can remove this and use Object.assign instead.

KKoukiou (dismissed):
Nice job!

ddonahue007 (dismissed):
LGTM

mturley (approved):
LGTM! Nice!

packages/react-charts/src/components/ChartUtils/chart-padding.ts
```diff
@@ -1,17 +1,14 @@
...
-  if (!isEmpty(padding)) {
-    return get(padding, side, 0);
-  }
-
-  if (isFinite(padding)) {
+  if (typeof padding == 'number') {
     return padding as number;
+  } else if (typeof padding == 'object') {
+    return padding[side] || 0;
   }
```
dlabrecq, line 12: I'm not sure this change is really the equivalent of what we had in place? We want to ensure padding is truly empty, not just an object, and there may be one or more properties.

dlabrecq (changes requested):

packages/react-charts/src/components/ChartUtils/chart-padding.ts
redallen: Fixed!

dlabrecq (commented):
LGTM

dlabrecq (approved):

mturley (approved):

ddonahue007 (commented):
LGTM

PR #4831, opened 2020-09-16
feat(SkipToContent): Update skip to content to support react router

### LINKED ISSUE #4809: Update skip to content component to support react router
When trying to add the skip to content component to OpenShift, I noticed an odd behavior where instead of appending #content to the end of the path in the route, it would replace it, going from `localhost:9000/dashboards#content` to `localhost:9000/#content`. OpenShift uses react router which is likely the reason this is happening. We should update the SkipToContent component to accept other router components instead of always rendering as an anchor tag. We should update the code to be similar to this one
https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Nav/NavItem.tsx#L72

Screen Recording 2020-09-11 at 9 46 47 AM

Enhancement

Trying to target the OpenShift 4.7 release

jessiehuff:
**What**: Closes #4809

### DIFF
--- a/packages/react-core/src/components/Page/__tests__/__snapshots__/Page.test.tsx.snap
+++ b/packages/react-core/src/components/Page/__tests__/__snapshots__/Page.test.tsx.snap
@@ -556,6 +556,7 @@ exports[`Check page to verify skip to content points to main content region 1`] 
   skipToContent={
     <SkipToContent
       href="#main-content-page-layout-test-nav"
+      render={null}
       show={false}
     >
       Skip to Content
@@ -569,6 +570,7 @@ exports[`Check page to verify skip to content points to main content region 1`] 
   >
     <SkipToContent
       href="#main-content-page-layout-test-nav"
+      render={null}
       show={false}
     >
       <a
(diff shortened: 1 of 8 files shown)


jschuler (approved):

tlabaj (changes requested):
Can you add  new prop to demo app please and add an integration test.

tlabaj (approved):
LGTM

redallen (changes requested):
I don't see a [SkipToContent](https://github.com/openshift/console/search?q=SkipToContent) anywhere in OpenShift, but if there was one my guess is that it has `href="/#main-content"` instead of `href="#main-content"` which is causing your undesired behavior.

All react-router does in addition to the default browser behavior for hash links of `document.getElementById('main-content').scrollTo()` is check to see which route it should render, which should always be the same if not using a HashRouter. In the case of SkipToContent we should always want the browser behavior to take place, not a router's.

I'm happy to help you debug.

jessiehuff:
Sure @redallen! Definitely interested in any insight you might have. This is the [PR](https://github.com/openshift/console/pull/6596) I had with OpenShift trying to add the Skip to Content. I was attempting to do: `href={'#content'}`

redallen:
It doesn't work because of the `<base href="/" />` tag that OpenShift includes. Try this instead:
```jsx
<SkipToContent href={`${this.props.location.pathname}#content`}>Skip to Content</SkipToContent>`
```

jschuler:
@redallen react-router has some issues with hash links afaik. see https://github.com/ReactTraining/react-router/issues/394#issuecomment-220221604
I think that is the main reason openshift uses https://www.npmjs.com/package/react-router-hash-link

PR #4837, opened 2020-09-17
feat(react-core-button): add isLoading and spinnerAriaValueText properties

### LINKED ISSUE #4816: Add progress button component
https://github.com/patternfly/patternfly/pull/3382

If a button supports the conditional showing of the spinner when it's in progress, it needs a base class `pf-m-progress`. When the button is in the "in progress" state (the spinner shows), we need to add the class `.pf-m-in-progress` to the button and add `span.pf-c-button__progress` with a medium sized spinner in it as the first child of the button.

KKoukiou:
Closes #4816

### DIFF
--- a/packages/react-core/src/components/Button/Button.tsx
+++ b/packages/react-core/src/components/Button/Button.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/Button/button';
 import { css } from '@patternfly/react-styles';
+import { Spinner, spinnerSize } from '../Spinner';
 import { useOUIAProps, OUIAProps } from '../../helpers';
 
 export enum ButtonVariant {
@@ -35,6 +36,10 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement>, OUIAPro
   isDisabled?: boolean;
   /** @beta Adds disabled styling and communicates that the button is disabled using the aria-disabled html attribute */
   isAriaDisabled?: boolean;
+  /** Adds progress styling to button */
+  isLoading?: boolean;
+  /** Aria-valuetext for the loading spinner */
+  spinnerAriaValueText?: string;
   /** @beta Events to prevent when the button is in an aria-disabled state */
   inoperableEvents?: string[];
   /** Adds inline styling to a link button */
@@ -65,6 +70,8 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
   isBlock = false,
   isDisabled = false,
   isAriaDisabled = false,
+  isLoading = null,
+  spinnerAriaValueText,
   isSmall = false,
   isLarge = false,
   inoperableEvents = ['onClick', 'onKeyPress'],
@@ -122,6 +129,8 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
         isAriaDisabled && styles.modifiers.ariaDisabled,
         isActive && styles.modifiers.active,
         isInline && variant === ButtonVariant.link && styles.modifiers.inline,
+        isLoading !== null && styles.modifiers.progress,
+        isLoading && styles.modifiers.inProgress,
         isSmall && styles.modifiers.small,
         isLarge && styles.modifiers.displayLg,
         className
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/Button/Button.tsx
```diff
@@ -64,6 +69,8 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
   isBlock = false,
   isDisabled = false,
   isAriaDisabled = false,
+  isLoading = null,
+  spinnerAriaValueText,
```
jschuler, line 73: could set this to a reasonable default, like `Loading...` maybe

packages/react-core/src/components/Button/Button.tsx
```diff
@@ -130,12 +150,15 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
       type={isButtonElement ? type : null}
       {...ouiaProps}
     >
-      {variant !== ButtonVariant.plain && icon && iconPosition === 'left' && (
-        <span className={css(styles.buttonIcon, styles.modifiers.start)}>{icon}</span>
-      )}
-      {children}
-      {variant !== ButtonVariant.plain && icon && iconPosition === 'right' && (
-        <span className={css(styles.buttonIcon, styles.modifiers.end)}>{icon}</span>
+      {isLoading ? (
```
jschuler, line 153: i'd suggest leaving the previous code and just adding this piece in the beginning, seems more readable that way to me
```
{isLoading && (
  <span className={css(styles.buttonProgress)}>
    <Spinner size={spinnerSize.md} aria-valuetext={spinnerAriaValueText} />
  </span>
)
```

packages/react-core/src/components/Button/examples/Button.md
```diff
@@ -223,3 +223,19 @@ ButtonCTAVariants = () => (
...
 ```
+
+### Progress
+```js
+import React from 'react';
+
+ButtonProgressVariants = () => (
+  <React.Fragment>
+    <Button spinnerAriaValueText="Not loading" isLoading={false} variant="primary">Primary loader</Button>{' '}
```
jschuler, line 234: can remove the `spinnerAriaValueText` props when isLoading is false since it won't get rendered anyways.

jschuler (changes requested):

packages/react-core/src/components/Button/Button.tsx
tlabaj: @jessiehuff thoughts on the default value here.  Also does this prop name follow our conventions?

mcarrano (dismissed):
Looks good from a design perspective. Thanks @KKoukiou !

packages/react-core/src/components/Button/Button.tsx
KKoukiou: So from consumer perceptive - we never use the default text values because their are not internationalized.
I guess everyone using PF also supports different languages so I am not convinced about the necessity of default text values.

KKoukiou: Regarding the property name it's aready used in the FileUploadField component.

jschuler: We already typically default to English values, in this case even if the value is not set and therefore undefined, Spinner defaults it already to "Loading..."

jschuler: So either way it doesn't matter too much in this PR

jschuler (dismissed):
LGTM, just need to make the linter happy

mcoker:
If you go from `isLoading={false}` to `isLoading`, the text should animate and slide over to make room for the spinner to display. I can't tell if that is happening by updating the prop in the workspace. Can we verify that the animation is working when switching states?

KKoukiou:
> If you go from `isLoading={false}` to `isLoading`, the text should animate and slide over to make room for the spinner to display. I can't tell if that is happening by updating the prop in the workspace. Can we verify that the animation is working when switching states?


> If you go from `isLoading={false}` to `isLoading`, the text should animate and slide over to make room for the spinner to display. I can't tell if that is happening by updating the prop in the workspace. Can we verify that the animation is working when switching states?

I think it's working, but let me adjust the example to change the button state on button click so that you can see it easier.

KKoukiou:
@mcoker I updated the example, now you can make sure the behavior is the expected easily I think.  :)

mcoker (approved):
Perfect, thanks @KKoukiou!!

mcoker:
Looks great, thanks @KKoukiou. I filed a core issue to see if we can do something about the focus ring around the spinner https://github.com/patternfly/patternfly/issues/3511

redallen (approved):

jschuler (approved):

Issue #4822, opened 2020-09-14
Data list text modifiers 


mcoker:
https://github.com/patternfly/patternfly/pull/3463/

This adds
* `.pf-m-truncate` - truncates text with ellipsis
* `.pf-m-break-word` - enforces aggressive word wrapping for long strings
* `.pf-m-nowrap` - keeps text from wrapping to multiple lines
* `span.pf-c-data-list__text` - a generic element that can be used with one of these modifiers to control specific lines of text. The element can be whatever, doesn't have to be a `span`, so should probably make it customizable with a `component` prop.

Those modifiers apply to the following elements

`.pf-c-data-list` - applies to the data-list
`.pf-c-data-list__item-row` - applies to the row
`.pf-c-data-list__cell` - applies to the cell
`.pf-c-data-list__text` - applies to whatever a user puts in the element

PR #4842, opened 2020-09-17
feat(Progress): added truncation

### LINKED ISSUE #4823: Ability to truncate progress component description
https://github.com/patternfly/patternfly/pull/3436/

You can now apply `.pf-m-truncate` to the progress component's description and it will truncate a long line with `...`

evwilkin:
**What**: Closes #4823 

This PR adds new `isTitleTruncated` prop with a new corresponding example: https://patternfly-react-pr-4842.surge.sh/components/progress#truncate-title


packages/react-core/src/components/Progress/Progress.tsx
```diff
@@ -34,6 +34,8 @@ export interface ProgressProps extends Omit<React.HTMLProps<HTMLDivElement>, 'si
   max?: number;
   /** Accessible text description of current progress value, for when value is not a percentage. Use with label. */
   valueText?: string;
+  /** Indicate whether to truncate the title */
+  isTruncated?: boolean;
```
mcoker, line 38: wdyt about calling this something like `isTitleTruncated` in case we add truncation to something else (like the label)? Then again we could always add a unique string to the other truncated thing (`isLabelTruncated`).

evwilkin: @mcoker makes sense to me, thanks for the suggestion!  Pushed this update

mcoker (dismissed):
🤠👍

KKoukiou (commented):
I feel that we need a more generic solution for adding text mofidiers to component elements.
I am currently working on https://github.com/patternfly/patternfly-react/issues/4822 
which also is supposed to add text modifier on the DataList component.

I think passing a list of available transforms, similarly how we do in the Table component is more flexible solution than adding for each modifier a property.

Imagine if next user asks for the title to be no wrapping, would we introduce then a isTitleNoWrapping property?

All in all I think we should keep this in the back of our mind and maybe offer a general solution later on. This is fine as is for me as well.

mcoker:
Hey @KKoukiou, I agree that would be really useful. We're in the middle of creating a utility class that allows for this - applying aggressive word wrapping, no-wrap, and truncation. Users could use that where they need this kind of text modification. Going forward, I think we can offer that as a suggestion. Though one issue with relying on users to insert an element to control text on their own is that it won't always work depending on where the element is and what kind of layout that element is a part of - for example if the parent is a flex/grid layout, if we apply spacing to all children of an element and inserting an additional element introduces issues with the spacing, etc. So I imagine we will still build these modifiers in with some components.

tlabaj (changes requested):
I think we want to add a tooltip here.  In general when we truncate we want to add a tooltip so the user is still able to see the truncated text.
cc @mcarrano

mcarrano:
Yes, agree with @tlabaj . I'm also unclear about what I'm seeing in the example here. I'm not seeing this truncate:

Screen Shot 2020-09-21 at 5 09 58 PM

packages/react-core/src/components/Progress/examples/Progress.md
```diff
@@ -133,3 +133,11 @@ import { Progress, ProgressSize, ProgressMeasureLocation, ProgressVariant } from
 
 <Progress value={2} title="Title" min={0} max={5} label="Step 2: Copying files" valueText="Step 2: Copying files" />
 ```
+
+### Truncate title
+```js
+import React from 'react';
+import { Progress } from '@patternfly/react-core';
+
+<Progress value={33} title="Very very very very very very very very very very very long title which should be truncated if it does not fit onto one line above the progress bar" isTruncated />
```
mcoker, line 142:
```suggestion
<Progress value={33} title="Very very very very very very very very very very very long title which should be truncated if it does not fit onto one line above the progress bar" isTitleTruncated />
```

evwilkin: Updated - thanks 👍

mcoker (approved):

nicolethoen (commented):
Okay, im sorry im late to this PR..

I am wondering if we should handle the title truncation like we handle the modal title truncation?
then we wouldn't need the 'isTitleTruncated' prop? 
[Modal title truncation logic](https://github.com/patternfly/patternfly-react/blob/66ea373f6ebad309d98a73d91b9cce93d05ad8a6/packages/react-core/src/components/Modal/ModalBoxTitle.tsx#L25)

evwilkin:
@mcarrano demo is updated - we had renamed the prop name but neglected to update the demo to reflect this.

@tlabaj - added the Tooltip component for truncated text. Do you think this approach works best or should we look at removing the `isTitleTruncated` prop completely and forcing it to be automated per @nicolethoen's comment?

tlabaj:
> @mcarrano demo is updated - we had renamed the prop name but neglected to update the demo to reflect this.
> 
> @tlabaj - added the Tooltip component for truncated text. Do you think this approach works best or should we look at removing the `isTitleTruncated` prop completely and forcing it to be automated per @nicolethoen's comment?

@mcarrano any thoughts here?

mcarrano:
I don't have a strong opinion @tlabaj . Automating sounds good but I don't know the pros and cons or how much complexity that adds.  Otherwise this is looking good.

evwilkin:
@mcarrano @tlabaj if there's no preference then I'll leave as-is, as automating it could potentially be considered a visual breaking change as users wouldn't have a way to opt out of truncation if for any reason they don't want it added where it doesn't already exist.

redallen (dismissed):
Code LGTM. We might want to support more tooltip props in the future, though.

evwilkin:
Reviews above were mistakenly removed, apologies for the confusion

mcoker:
It looks like this if you enable truncation but the content isn't long enough to truncate.

Screen Shot 2020-09-30 at 10 33 42 AM

mcoker (dismissed):
lgtm!

redallen:
I'm taking this over since @evwilkin is on PTO.

@nicolethoen That's a good idea and I tried that, but a Progress' title **can overflow** onto another line, so a consumer has to specify whether they want a tooltip or the default 2-line title.

@mcoker I've updated the PR to only show the tooltip when there is overflow.

I noticed since @jschuler 's Popover in-housing that newly-mounted Tooltips don't appear unless you move your mouse away and then back to trigger Tooltip's `onMouseEnter`. I've removed the `triggerManually` check from `Tooltip`'s `React.useEffect` so that the Tooltip properly shows on first mouseover for Progress as well as other components that use this overflow checking pattern.

mcoker (approved):
👍

tlabaj (approved):
LGTM

packages/react-core/src/components/Progress/ProgressContainer.tsx
```diff
@@ -49,14 +54,39 @@ export const ProgressContainer: React.FunctionComponent<ProgressContainerProps>
   parentId,
   label = null,
   variant = null,
-  measureLocation = ProgressMeasureLocation.top
+  measureLocation = ProgressMeasureLocation.top,
+  isTitleTruncated = false,
+  tooltipPosition
 }: ProgressContainerProps) => {
   const StatusIcon = variantToIcon.hasOwnProperty(variant) && variantToIcon[variant];
+  const [tooltip, setTooltip] = React.useState('');
```
jschuler, line 62: how about make this a boolean `const [tooltipVisible, setTooltipVisible] = React.useState(false)`
then `onMouseEnter`, set this to true if content is truncated.
Also add `onMouseLeave to set tooltipVisible to false if it's true.
And the tooltip component
```
<Tooltip position={tooltipPosition} content={title} isVisible={tooltipVisible} trigger="manual">
  {Title}
</Tooltip>
```

redallen: This is less performant (always renders a Tooltip, which previous benchmarks show is expensive) and I have to add the visibility logic not only here, but also in the table transformers and modal box.

jschuler (approved):

PR #4843, opened 2020-09-18
Remove some unused variables && properties

KKoukiou:


jenny-s51 (commented):
Hi @KKoukiou! Could you please link the corresponding issue in the description of this PR?

KKoukiou:
> Hi @KKoukiou! Could you please link the corresponding issue in the description of this PR?

There is not corresponding issue - I just found some problems while checking the codebase and fixed them.

packages/react-core/src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -12,8 +12,6 @@ export interface DropdownMenuProps {
   className?: string;
   /** Flag to indicate if menu is opened */
   isOpen?: boolean;
-  /** Flag to indicate if menu should be opened on enter */
-  openedOnEnter?: boolean;
```
redallen, line 16: I checked [patternfly-analytics](https://github.com/rh-uxd/patternfly-analytics) and none of our known consumers are using this prop.

However, since we export this component in `Dropdown/index.ts` it's better to just mark it as deprecated and destructure it so that it's not passed to the DOM. If we remove it entirely we might break some Typescript consumer who is using it, or some Javascript consumer might be confused why their MenuComponent has `openedOnEnter` passed to it which propagates down to an HTMLElement.

redallen (changes requested):

packages/react-core/src/components/Dropdown/DropdownMenu.tsx
KKoukiou: I see your point. It's sad that we have to keep a property which literally does nothing but you are completely right.

KKoukiou: @redallen btw this component is not exposed at all in the documentation website. https://www.patternfly.org/v4/documentation/react/components/dropdown

I would say then that since it's not there, it's properties change should not worry us.

redallen (dismissed):
I checked the repo from 2018 and 2019 to see if we ever used that component in an example and it looks like it's always been internal, despite us exposing it.

Because of that, I'm good with removing the prop.  Marking it as deprecated is good as well.

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
```diff
@@ -95,10 +95,9 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
             <DropdownMenu
               setMenuComponentRef={this.setMenuComponentRef}
               component={component}
-              isOpen={isOpen}
+              isOpen={isOpen || openedOnEnter}
```
jschuler, line 98: Probably better to just leave it off here

tlabaj (changes requested):
Can you please link the associated issue.

KKoukiou:
> Can you please link the associated issue.

https://github.com/patternfly/patternfly-react/pull/4843#issuecomment-694914096

KKoukiou:
> Can you please link the associated issue.

@tlabaj in patternfly do you really create an issue for each PR? Especially for this case right here creating an issue does not really make sense IMO, because it's multiple unrelated between then fixes. It's mostly a repository maintenance PR and not targeting specific issue. Opening an issue for such PRs will just create unnecessary noise. Or am I missing something?

redallen (dismissed):

jschuler (approved):

tlabaj:
@KKoukiou Yes, this is the outlined here in our [contribution guidelines](https://github.com/patternfly/patternfly-react/blob/master/CONTRIBUTING.md#creating-issues-for-bugs).

KKoukiou:
> @KKoukiou Yes, this is the outlined here in our [contribution guidelines](https://github.com/patternfly/patternfly-react/blob/master/CONTRIBUTING.md#creating-issues-for-bugs).

Thanks for the answer :) I created an issue added the fixes line and reposted.

redallen (approved):

jschuler (approved):

tlabaj (approved):

PR #4846, opened 2020-09-18
feat(DataList): add support for text modifiers

### LINKED ISSUE #4822: Data list text modifiers

KKoukiou:
Fixes #4822

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -3,6 +3,12 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 
+export enum WrapModifier {
+  nowrap = 'nowrap',
+  truncate = 'truncate',
+  breakWord = 'breakWord'
+}
+
 export interface DataListProps extends React.HTMLProps<HTMLUListElement> {
   /** Content rendered inside the DataList list */
   children?: React.ReactNode;
@@ -16,6 +22,8 @@ export interface DataListProps extends React.HTMLProps<HTMLUListElement> {
   selectedDataListItemId?: string;
   /** Flag indicating if DataList should have compact styling */
   isCompact?: boolean;
+  /** Determines which wrapping modifier to apply to the DataList */
+  wrapModifier?: WrapModifier | 'nowrap' | 'truncate' | 'breakWord';
 }
 
 interface DataListContextProps {
@@ -35,6 +43,7 @@ export const DataList: React.FunctionComponent<DataListProps> = ({
   selectedDataListItemId = '',
   onSelectDataListItem,
   isCompact = false,
+  wrapModifier = null,
   ...props
 }: DataListProps) => {
   const isSelectable = onSelectDataListItem !== undefined;
@@ -52,7 +61,12 @@ export const DataList: React.FunctionComponent<DataListProps> = ({
       }}
     >
       <ul
-        className={css(styles.dataList, isCompact && styles.modifiers.compact, className)}
+        className={css(
+          styles.dataList,
+          isCompact && styles.modifiers.compact,
+          className,
+          wrapModifier && styles.modifiers[wrapModifier]
+        )}
         aria-label={ariaLabel}
         {...props}
       >
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 
-export enum WrapModifier {
+export enum DataListWrapModifier {
   nowrap = 'nowrap',
   truncate = 'truncate',
   breakWord = 'breakWord'
@@ -23,7 +23,7 @@ export interface DataListProps extends React.HTMLProps<HTMLUListElement> {
   /** Flag indicating if DataList should have compact styling */
   isCompact?: boolean;
   /** Determines which wrapping modifier to apply to the DataList */
-  wrapModifier?: WrapModifier | 'nowrap' | 'truncate' | 'breakWord';
+  wrapModifier?: DataListWrapModifier | 'nowrap' | 'truncate' | 'breakWord';
 }
 
 interface DataListContextProps {
--- a/packages/react-core/src/components/DataList/DataListItemRow.tsx
+++ b/packages/react-core/src/components/DataList/DataListItemRow.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
-import { WrapModifier } from './DataList';
+import { DataListWrapModifier } from './DataList';
 
 export interface DataListItemRowProps extends Omit<React.HTMLProps<HTMLDivElement>, 'children'> {
   children: React.ReactNode;
   className?: string;
   rowid?: string;
-  wrapModifier?: WrapModifier | 'nowrap' | 'truncate' | 'breakWord';
+  wrapModifier?: DataListWrapModifier | 'nowrap' | 'truncate' | 'breakWord';
 }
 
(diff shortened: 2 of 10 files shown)


KKoukiou:
@jschuler I would like to get some feedback on this before I continue with the tests and such. Are you fine with the current way of exposing the transforms?

packages/react-core/src/components/DataList/utils/decorators/wrappable.tsx
```diff
@@ -0,0 +1,14 @@
+import styles from '@patternfly/react-styles/css/components/DataList/data-list';
```
jschuler, line 1: It's an interesting approach to applying a set of modifiers, but a bit different from most of our other components (apart from Table I think). 
Typically we just have individual boolean props that map directly to a className.

Here's the Button component for example:
```
className={css(
        styles.button,
        styles.modifiers[variant],
        isBlock && styles.modifiers.block,
        isDisabled && styles.modifiers.disabled,
        isAriaDisabled && styles.modifiers.ariaDisabled,
        isInline && variant === ButtonVariant.link && styles.modifiers.inline,
        isLoading !== null && styles.modifiers.progress,
        isLoading && styles.modifiers.inProgress,
        isSmall && styles.modifiers.small,
        isLarge && styles.modifiers.displayLg,
        className
      )}
```

But if you wanted to follow this pattern, I would simplify it and just depend on typescript type safety and removing imports etc. something like

And maybe change the prop to `modifiers`?

```
export type ITransform = 'truncate' | 'breakWord' | 'nowrap';
...
modifiers?: ITransform[];
...
className={css(
    modifiers && modifiers.map(key => {
      if (styles.modifiers[key]) return styles.modifiers[key];
      console.warn(`transform ${key} not found`);
    })
)}
```

Don't have strong feelings for one approach or the other. Maybe @tlabaj and @redallen can weigh in as well.

redallen: I think we should follow the long explicit pattern like we do with button. I find this pattern clever, but we've already committed to adding modifiers the long-winded typesafe way. Also, global exports like `breakword` aren't self explanatory and shouldn't be in our main index.

If we ever have a component with more than 20 modifiers we should reconsider this, though.

KKoukiou: @redallen I feel that these properties should not be exposed as individual boolean properties on all DataList sub components, ending up with so many exposed properties is not really elegant.

Take a note that the already exposed properties on the Button don't try to affect the same issue, for example isDisabled, isLoading, all affect different concepts of the button, as for disabling events (Disabled) and adding a spinner (isLoading).

On our case though with the text modifiers, they are all referring to the text transforms, and thus should be grouped like these. From what I see in the Table component, these transforms are also not supposed to be used in combinations, which is another indicators that we should have them exposed as a sigle property.

See https://github.com/patternfly/patternfly-react/blob/master/packages/react-table/src/components/Table/TableText.tsx#L27 

I am sure 2 more of these transform will come, because they already exist on the Table component, these are `wrap` && `fitContent`.

Lastly, I understand you concern, but by using the exact same way TableText does it like here https://github.com/patternfly/patternfly-react/blob/master/packages/react-table/src/components/Table/TableText.tsx#L27
I think we will solve the issue with type checking.

packages/react-core/src/components/DataList/examples/DataList.md
```diff
@@ -1050,3 +1053,37 @@ class SelectableDataList extends React.Component {
...
+  Button,
+  DataList,
+  DataListItem,
+  DataListItemRow,
+  DataListItemCells,
+  DataListCell,
+  DataListCheck,
+  DataListAction,
+  truncate,
+  breakWord
```
redallen, line 1070: Remove these unused and non-existant imports. You could import `WrapModifier` and then use `WrapModifier.truncate` if you wanted.

redallen (changes requested):
Looks much better, just two lines are off in the example!

packages/react-core/src/components/DataList/DataListItemRow.tsx
```diff
@@ -1,20 +1,23 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
+import { WrapModifier } from './DataList';
 
 export interface DataListItemRowProps extends Omit<React.HTMLProps<HTMLDivElement>, 'children'> {
```
tlabaj, line 6: Looks like the prop descriptions were missing here.  Do you mind adding those as part of your PR?

tlabaj (changes requested):

packages/react-core/src/components/DataList/DataList.tsx
```diff
@@ -3,6 +3,12 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 
+export enum WrapModifier {
```
jschuler, line 6: Would be good to prefix to `DataListWrapModifier`, otherwise it could clash if there is another WrapModifier exported somewhere else

redallen: You could also get rid of the enum altogether if you'd like since most of our consumers use our types. It's up to you.

mcoker (commented):
This looks great! Does it add a `<DataListText>` component that also supports the `wrapModifiers`? I didn't see it but I could have overlooked it.

KKoukiou:
> This looks great! Does it add a `<DataListText>` component that also supports the `wrapModifiers`? I didn't see it but I could have overlooked it.

No - i missed this - let me add it and repost with the fixes.

packages/react-core/src/components/DataList/DataListText.tsx
```diff
@@ -0,0 +1,60 @@
...
+  nav = 'nav'
+}
+
+export interface DataListTextProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered within the data list text */
+  children?: React.ReactNode;
+  /** Additional classes added to the data list text */
+  className?: string;
+  /** Determines which element to render as a data list text */
+  variant?: DataListTextVariant | 'span' | 'div';
```
mcoker, line 18: Just want to double check this shouldn't be `component`? Typically when I've seen a prop to use a different element for something, it's a `component` prop.

mcoker (dismissed):
LGTM! Everything works as expected. Just one question about the `<DataListText>`'s `variant` prop.

tlabaj (approved):
LGTM

tlabaj (changes requested):
Can you please add some integration test.

packages/react-core/src/components/DataList/DataListText.tsx
redallen: Fixed.

packages/react-core/src/components/DataList/DataListText.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <Component
+      onMouseEnter={onMouseEnter}
+      className={css(className, wrapModifier && styles.modifiers[wrapModifier], styles.dataListText)}
+      {...props}
+    >
+      {children}
+    </Component>
+  );
+
+  return tooltip !== '' ? <Tooltip content={tooltip}>{text}</Tooltip> : text;
```
jschuler, line 53: do you need to also add the `isVisible` prop here?

jschuler (approved):

mcoker (approved):

tlabaj (approved):
LGTM

redallen (approved):

PR #4847, opened 2020-09-18
feat(menu): introduces new menu component

### LINKED ISSUE #4712: Menu component
https://github.com/patternfly/patternfly/pull/3397

We created a menu component for general use. It should support most/all of the features we have with our current individual menus (dropdown, select, app launcher, etc). The end goal would be to use this menu either inside of those components, or for it to serve as the framework to build individual menus in those components with a set of shared base code.

This should inherit mouse and keyboard interactions from existing dropdown and select menus. Flyout is a new pattern - should follow guidance on implementing accessible fly-out menus here: https://www.w3.org/WAI/tutorials/menus/flyout/  This must include adding a 1 second timer on mouseout events before closing the menu.

jenny-s51:
**What**: Closes #4712


mcarrano:
@jenny-s51 @tlabaj @evwilkin Based on some conversations I've been having on the design side, I've added the "Do not merge" tag to this so we can give the approach to this component more consideration before releasing. Will keep you posted about next steps.

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,757 @@
...
+          Docker File
+        </MenuListItem>
+      </MenuList>
+    ];
+    return <Menu onSelect={this.onSelect} items={menuItems}></Menu>;
+  }
+}
+```
+
+### With expanded toggle
```
jenny-s51, line 129: opened an issue in core: https://github.com/patternfly/patternfly/issues/3599

should I remove this variant for now?

packages/react-core/src/components/Menu/MenuListItem.tsx
```diff
@@ -129,6 +132,7 @@ export const MenuListItem: React.FunctionComponent<MenuListItemProps> = ({
             <span>{description}</span>
           </div>
         )}
+        {flyoutMenuItems}
```
jenny-s51, line 135: what is the best way to implement this? seems like it needs an event listener like `onMouseEnter` or `onMouseOver`...

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -204,6 +230,9 @@ class MenuWithFlyout extends React.Component {
           to="#default-link2"
           itemId={3}
           isActive={activeItem === 3}
+          isExpandable
+          onMouseEnter={handleMenuToggle}
+          flyoutMenuItems={this.state.isSubMenuOpen ? flyoutMenuItems : ""}
```
jenny-s51, line 235: at the moment the menu also takes the full width of the container rather than how it is in the core example when the `pf-m-flyout` modifier is applied. 

is this to be expected? see https://www.patternfly.org/v4/components/menu#with-flyout

redallen (approved):

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,103 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Menu/menu';
+import { css } from '@patternfly/react-styles';
+import { getOUIAProps, OUIAProps, getDefaultOUIAId } from '../../helpers';
+import { SearchInput } from '../SearchInput';
+import { Divider } from '../Divider';
+import { MenuContext } from './MenuContext';
+
+export interface MenuProps
+  extends Omit<React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'ref' | 'onSelect'>,
```
jschuler, line 10: this can be simplified to
`export interface MenuProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'ref' | 'onSelect'>, OUIAProps`

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onFavorite?: (event: React.MouseEvent, itemId: any) => void;
+  /** A callback for when the input value changes. */
+  onSearchInputChange?: (
+    event: React.FormEvent<HTMLInputElement> | React.SyntheticEvent<HTMLButtonElement>,
+    value: string
+  ) => void;
+  /** Accessibility label */
+  'aria-label'?: string;
+  /** Indicates menu type */
+  variant?: 'default' | 'flyout' | 'singleSelect' | 'multiSelect';
```
jschuler, line 28: is this needed? I only see `variant="flyout"` used in the examples. If it's not needed maybe it's better to remove it and add a `isFlyout` boolean prop onstead

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { SearchInput } from '../SearchInput';
+import { Divider } from '../Divider';
+import { MenuContext } from './MenuContext';
+
+export interface MenuProps
+  extends Omit<React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'ref' | 'onSelect'>,
+    OUIAProps {
+  /** Anything that can be rendered inside of the Menu */
+  children?: React.ReactNode;
+  /** Additional classes added to the container */
```
jschuler, line 14:
```suggestion
  /** Additional classes added to the Menu */
```

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onSearchInputChange?: (
+    event: React.FormEvent<HTMLInputElement> | React.SyntheticEvent<HTMLButtonElement>,
+    value: string
+  ) => void;
+  /** Accessibility label */
+  'aria-label'?: string;
+  /** Indicates menu type */
+  variant?: 'default' | 'flyout' | 'singleSelect' | 'multiSelect';
+  /** Search input of menu */
+  hasSearchInput?: React.ReactNode;
```
jschuler, line 30:
```suggestion
  hasSearchInput?: boolean;
```

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+  /** Forwarded ref */
+  innerRef?: React.Ref<any>;
+}
+
+export interface MenuState {
+  typeaheadInputValue: string | null;
```
jschuler, line 36:
```suggestion
 searchInputValue: string | null;
```

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <div
+          className={css(styles.menu, variant === 'flyout' && styles.modifiers.flyout, className)}
+          aria-label={ariaLabel || variant === 'flyout' ? 'Local' : 'Global'}
+          ref={innerRef}
+          {...getOUIAProps(Menu.displayName, ouiaId !== undefined ? ouiaId : this.state.ouiaStateId, ouiaSafe)}
+          {...props}
+        >
+          {hasSearchInput && (
+            <React.Fragment>
+              <div className="pf-c-menu__search">
```
jschuler, line 77:
```suggestion
              <div className={styles.menuSearch}>
```

packages/react-core/src/components/Menu/MenuAction.tsx
```diff
@@ -0,0 +1,40 @@
...
+    aria-label={ariaLabel}
+    onClick={onActionClick}
+    ref={innerRef}
+    {...props}
+  >
+    <span className={css(styles.menuItemActionIcon)}>{icon}</span>
+  </button>
+);
+
+export const MenuAction = React.forwardRef((props: MenuActionProps, ref: React.Ref<HTMLDivElement>) => (
```
jschuler, line 37:
```suggestion
export const MenuAction = React.forwardRef((props: MenuActionProps, ref: React.Ref< HTMLButtonElement>) => (
```

packages/react-core/src/components/Menu/MenuGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+  /** Group label */
+  label?: string;
+  /** ID for title label */
+  titleId?: string;
+  /** Forwarded ref */
+  innerRef?: React.Ref<any>;
+}
+
+const MenuGroupBase: React.FunctionComponent<MenuGroupProps> = ({
+  children = [] as React.ReactElement[],
```
jschuler, line 19:
```suggestion
  children,
```

packages/react-core/src/components/Menu/MenuGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+  className = '',
+  label = '',
+  titleId = '',
+  innerRef,
+  ...props
+}: MenuGroupProps) => (
+  <section {...props} className={css('pf-c-menu__group', className)} ref={innerRef}>
```
jschuler, line 26: is this class in the styles obj?

packages/react-core/src/components/Menu/MenuListItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+import styles from '@patternfly/react-styles/css/components/Menu/menu';
+import { css } from '@patternfly/react-styles';
+import ExternalLinkAltIcon from '@patternfly/react-icons/dist/js/icons/external-link-alt-icon';
+import StarIcon from '@patternfly/react-icons/dist/js/icons/star-icon';
+import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
+import CheckIcon from '@patternfly/react-icons/dist/js/icons/check-icon';
+import { MenuContext } from './MenuContext';
+
+export interface MenuListItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'action' | 'onClick'> {
+  /** Content rendered inside the menu list item. If React.isValidElement(children) props onClick, className and aria-current will be injected. */
```
jschuler, line 11: does this comment still apply?

packages/react-core/src/components/Menu/MenuListItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+  /** Component used to render the menu item */
+  component?: React.ReactNode;
+  /** Render item as disabled option */
+  isDisabled?: boolean;
+  /** Render item with icon */
+  icon?: React.ReactNode;
+  /** Render item with action */
+  action?: React.ReactNode;
+  /** Description of the menu item */
+  description?: string | React.ReactNode;
```
jschuler, line 32:
```suggestion
  description?: React.ReactNode;
```
string is part of ReactNode already

nicolethoen (commented):
So, I am noticing that the keyboard interactions for the menu with favorites is different than that of the select component. was that on purpose? The current select menu with favorites lets you use the arrow keys to move focus around.
@jessiehuff @tlabaj

packages/react-core/src/components/Menu/MenuGroup.tsx
jenny-s51: no, just `menuGroupItem`, but no `menuGroup`

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,738 @@
...
+            <MenuListItem to="#default-link1" itemId={0} isActive={activeItem === 0}>
+              Link not in group
+            </MenuListItem>
+          </MenuList>
+        </MenuGroup>
+        <Divider />
+        <MenuGroup label="Group 1">
+          <MenuList>
+            <MenuListItem itemId={1} isActive={activeItem === 1}>
+              Link 1
```
mcoker, line 352: should these be labeled actions or updated with `to`?

jenny-s51: nice catch @mcoker - thank you

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,101 @@
...
+          className={css(styles.menu, isFlyoutMenu && styles.modifiers.flyout, className)}
+          aria-label={ariaLabel || isFlyoutMenu ? 'Local' : 'Global'}
+          ref={innerRef}
+          {...getOUIAProps(Menu.displayName, ouiaId !== undefined ? ouiaId : this.state.ouiaStateId, ouiaSafe)}
+          {...props}
+        >
+          {hasSearchInput && (
+            <React.Fragment>
+              <div className={styles.menuSearch}>
+                <SearchInput
```
mcoker, line 76: @mcarrano Jenny is using the search input component as the search filter. Typically we build the search box in these menus out using a standard form control text input. Do you think this is the case for the search input component instead?

@jenny-s51 is there an advantage of using `<SearchInput>` over `<TextInput>`?

packages/react-core/src/components/Menu/MenuGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children,
+  className = '',
+  label = '',
+  titleId = '',
+  innerRef,
+  ...props
+}: MenuGroupProps) => (
+  <section {...props} className={css('pf-c-menu__group', className)} ref={innerRef}>
+    {label && (
+      <h1 className={css(styles.menuGroupTitle)} id={titleId} aria-hidden>
```
mcoker, line 28: should this have aria-hidden?

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,738 @@
...
+  render() {
+    const { activeItem } = this.state;
+    return (
+      <Menu onSelect={this.onSelect}>
+        <MenuList>
+          <MenuListItem icon={<CodeBranchIcon />} description="Description" itemId={0} isActive={activeItem === 0}>
+            Action 1
+          </MenuListItem>
+          <MenuListItem
+            isDisabled
```
mcoker, line 407: Any interactive items in the list item should be disabled, so the `button.pf-c-menu__item` needs `disabled` here. This is missing in the core workspace, I'll create an issue to update it.

packages/react-core/src/components/Menu/MenuListItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+            href: to
+          }
+        : {};
+    return (
+      <>
+        <Component
+          onClick={(event: React.MouseEvent) => onItemSelect(event, onSelect)}
+          className={css(
+            styles.menuItem,
+            isFavorite && styles.modifiers.favorited,
```
mcoker, line 106: This item shouldn't need `pf-m-favorited`. Only the `menuItemAction` element below that has `.pf-m-favorite` would toggle `.pf-m-favorited` when it's favorited. You can see that in the core examples - https://pf4.patternfly.org/components/menu#with-favorites

mcoker (changes requested):
This looks fantastic @jenny-s51! I've left some minor feedback for when you get a chance to review.

jschuler:
@nicolethoen We can add keyboard interaction as well, similar to Dropdown, each item could register itself with the Menu and the Menu could listen to keydown events and focus the relevant item.
@mcoker is the plan to use this Menu component inside of Dropdown in the future?

mcoker:
@jschuler yep I think so, the menu could be used as the menu in any existing dropdown-esque component, or attached to a future `<Toggle />` component, or even some random button or right clicking on the page as a context menu... the possibilities are endless, really.

packages/react-core/src/components/Menu/Menu.tsx
jenny-s51: @mcoker I think I used `<SearchInput>` here because it matched the styling in the core demo. The icon variant of `<TextInput>` has  the icon on the left as opposed to the right, but for the core demo of the filterable menu variant, the search icon was on the right so I figured `<SearchInput>` was the component needed here.

mcarrano: @mcoker I don't have a preference for using one over the other. But it does seem like we should be consistent between core and React. If there is an advantage to using the Search input component I'm fine with that.

packages/react-core/src/components/Menu/MenuGroup.tsx
jenny-s51: will remove

jschuler:
@nicolethoen follow up for keyboard interaction for next milestone https://github.com/patternfly/patternfly-react/issues/5144

jschuler:
Since surge is down I've uploaded the build here https://5fb2d8b3c3384c3d74c70e04--pensive-newton-6ba48a.netlify.app/components/menu

redallen (dismissed):

kmcfaul (approved):
Thanks @jschuler for hosting the updated build while the surge is down.

Everything looks good to me! I do think we should add examples/cypress tests for the menu component to our integration app at some point but I think we can add it later (maybe as part of the keyboard interaction PR) if we want to get this in today.

mcarrano:
This looks good. The only problem I see is that the favorites button is missing hover states. But I think this is a bug on the core side. Can you confirm @mcoker? If so I'm OK to merge this and open a core issue to pick up in the next release.

mcoker:
@mcarrano correct, opened this issue to address - https://github.com/patternfly/patternfly/issues/3674

mcarrano (approved):
Looks good except for CSS issue I noted.

tlabaj (changes requested):
Is there a reason that the favorites logic is up to the consumer rather than built into the component.  Similar to what we do with the App luncher and Select?

Could you also add to the demo app and add integration test.  Since this is beta, I am ok with that being in a follow up PR.

tlabaj:
@mcoker @mcarrano Is the design that the menu Item with an action can not also have a favorites button?

jschuler:
@tlabaj yes the component is meant to be more generic and flexible and purposefully kept as simple as possible. As you can see in the example it really doesn't take much to render another group with favorites. And let's say for example, if they wanted to render their favorites below the menu it would be super simple as it's up to them where they want to render it.

mcoker:
> Is the design that the menu Item with an action can not also have a favorites button?

Nope, you should be able to have both/multiple item actions.

jschuler:
multiple actions are supported in the component, just pass something like
```
<MenuItem
                  key={itemId}
                  itemId={itemId}
                  action={
                    <>
                      <MenuItemAction icon={<BarsIcon aria-hidden />} aria-label="Expand" onActionClick={() => console.log('clicked bars icon')} />
                      <MenuItemAction
                        icon="favorites"
                        isFavorited={isFavorited}
                        aria-label={isFavorited ? 'starred' : 'not starred'}
                        onActionClick={() => console.log('clicked favorites')}
                      />
                    </>
                  }
                  {...rest}
                >
                  {text}
                </MenuItem>
```

packages/react-core/src/components/Menu/Menu.tsx
mcoker: @jenny-s51 ah! We support a variant of the `pf-c-form-control` component (used in `<TextInput>`) with `.pf-m-search` that creates the icon on the left. We use that in `<Select>` and `<TreeView>` so I tracked those down and looks like we're creating this for some reason:

`<input className={css(formStyles.formControl, formStyles.modifiers.search)} type="search" />`

https://github.com/patternfly/patternfly-react/blob/94bdfccf5cb96c1388fdf9528a145275264671e3/packages/react-core/src/components/TreeView/TreeViewSearch.tsx#L10

https://github.com/patternfly/patternfly-react/blob/94bdfccf5cb96c1388fdf9528a145275264671e3/packages/react-core/src/components/Select/Select.tsx#L709-L712

Is there any reason we shouldn't add this as a variant to `<TextInput>` and use that there instead? Is it because it might be confusing to have `type="search"` on `<TextInput>` that isn't necessarily tied to this search icon? Currently I think both `<input class="pf-m-form-control pf-m-search" type="text">` and `<input class="pf-m-form-control" type="search">` are valid - just meaning `pf-m-search` and `type="search"` don't have to be used together.

Though I agree with @mcarrano, we could use `<SearchInput>` if there is an advantage. WDYT @jenny-s51? I'll also add that to our meeting agenda for tomorrow @mcarrano.

jschuler:
@tlabaj I've made these changes which we can merge into this PR
https://github.com/jenny-s51/patternfly-react/pull/16
```
- instead of isActive and isSelected on the MenuItem level can now use activeItemId and selected on the Menu level
- renamed isFlyoutMenu => containsFlyout since that conveys a more accurate meaning
- **renamed action => actions in the MenuItem since you can add more than one action**
- added actionId to MenuItemAction which will get passed back in the onActionClick callback on the Menu
- **added isFavorited as a convenience prop to MenuItem although you could still instead pass it as a MenuItemAction**
```

packages/react-core/src/components/Menu/MenuItemAction.tsx
```diff
@@ -0,0 +1,72 @@
...
+import styles from '@patternfly/react-styles/css/components/Menu/menu';
+import { css } from '@patternfly/react-styles';
+import StarIcon from '@patternfly/react-icons/dist/js/icons/star-icon';
+import { MenuContext, MenuItemContext } from './MenuContext';
+
+export interface MenuItemActionProps extends Omit<React.HTMLProps<HTMLButtonElement>, 'type' | 'ref'> {
+  /** Additional classes added to the action button */
+  className?: string;
+  /** The action icon to use */
+  icon?: 'favorites' | React.ReactNode;
```
tlabaj, line 11: is there a reason the consumer should not just pass the favorites icons themselves?

jschuler: they still can, this is just for convenience

jschuler: also you can now just add `isFavorited` to the MenuItem and it will render the favorites icon

packages/react-core/src/components/Menu/Menu.tsx
jenny-s51: Using search input here made more sense to me intuitively, as the filterable variant uses searching functionality and we already had this component built out in react. I definitely agree that it's a bit confusing how currently `pf-m-search` and `type="search"` are used together on  `<input>` though ... I can open a new `TextInput` issue in react to address that @mcoker @mcarrano

mcoker (approved):

packages/react-core/src/components/Menu/MenuItemAction.tsx
tlabaj: If I am understanding the code correctly, `isFavorited` prop adds the favorites icon to the item.  To me the name implies that the Menu item has been favourited (so I expect it to appear in the Favorites group).

jschuler: yep you control whether the item has been favorited by adding the prop, initially it would be `isFavorited={false}` which is accurate

jschuler: and the top group (the group of favorited items) have `isFavorited={true}` which is also correct

tlabaj (approved):

jschuler (approved):
nice work @jenny-s51 !

PR #4851, opened 2020-09-21
fix(tab): Tab component properties don't show in the docs

### LINKED ISSUE #4647: Tab component is missing documentation in React
Right now, Tabs component page does not list documentation for the Tab child component. As such I cannot at a glance see what props it accepts or what the props do

KKoukiou:
Closes #4647

### DIFF
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -20,5 +20,18 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   isHidden?: boolean;
 }
 
-export const Tab: React.FunctionComponent<TabProps> = () => null;
+/** The parent <Tabs> component accecesses this component's propeties directly in order to present each Tab */
+export const Tab: React.FunctionComponent<TabProps> = ({
+  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+  children,
+  className,
+  href,
+  title,
+  eventKey,
+  tabContentId,
+  tabContentRef,
+  isHidden,
+  ...props
+}: /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+TabProps) => <React.Fragment />;
 Tab.displayName = 'Tab';


packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -19,5 +19,23 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  title,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  eventKey,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  tabContentId,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  tabContentRef,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  isHidden
```
redallen, line 39: Instead of disabling each line you can do:
```jsx
/* eslint-disable @typescript-eslint/no-unused-vars */
...
```

redallen (changes requested):
This PR should allow `react-docgen` to get `Tab`'s props. You can also just do `propComponents: [TabProps]` in the frontmatter to get them to show.

packages/react-core/src/components/Tabs/Tab.tsx
KKoukiou: > This PR should allow `react-docgen` to get `Tab`'s props. You can also just do `propComponents: [TabProps]` in the frontmatter to get them to show.

Unfortunately this PR did not fix the issue, therefore i did not remove the draft from it. I quickly checked it again now, and seems the return null is the issue. Returning anything else like an empty React.Fragment makes the props work.

KKoukiou: propComponents: [TabProps] also does not work, it does not parse the types any more with this.

redallen (approved):
It works, thanks!!
image

tlabaj (approved):

PR #4852, opened 2020-09-21
feat(DataList): add draggable feature for basic list

### LINKED ISSUE #4708: Data list: draggable rows
Placeholder for when https://github.com/patternfly/patternfly/pull/3401 is complete.

Since this is will rely on JS for the interaction and for design to validate, core and react will likely need to work on this together during the react implementation.

kmcfaul:
**What**: Closes #4708

Basic implementation of draggable. DataLists with nested state dependencies (like expandable) I am having trouble getting to work, because the whole list needs to be saved in the state in order to update, but you cannot access state variables inside of itself.

@mcoker The draggable ghost defaults to a copy of the whole row, and I haven't found a great way to change this yet, but it at least seems to preserve the spacing. The entire row is also draggable by default instead of just the chip button (with mouse), but it doesn't interfere with other actions so I'm not sure this is an issue.

For keyboard, hitting `space` or `enter` on a drag chip button begins the drag. Use the `up arrow` or `down arrow` to move the row (showing the ghost placeholder) and `enter` to confirm/finish the drag. `Escape` will cancel the drag, or hitting any other key (like `Tab` to continue tabbing through the DOM).

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -226,7 +226,8 @@ export class DataList extends React.Component<DataListProps, DataListState> {
     };
 
     const dragProps = isDraggable && {
-      onDragOver: this.dragOver
+      onDragOver: this.dragOver,
+      onDrop: this.dragOver
     };
 
     return (
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -112,7 +112,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
 
     arr.splice(oldIndex, 1);
     arr.splice(newIndex, 0, ghost);
-    onDragMove(oldIndex, newIndex);
+    onDragMove && onDragMove(oldIndex, newIndex);
     return arr;
   };
 
@@ -127,7 +127,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
         item => (item as React.ReactElement).props.id === evt.currentTarget.id
       ) as React.ReactElement
     });
-    onDragStart(evt.currentTarget.id);
+    onDragStart && onDragStart(evt.currentTarget.id);
   };
 
   dragEnd = (evt: React.DragEvent) => {
@@ -178,7 +178,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
     const dragInd = this.getIndex(dragItem.id);
 
     if (evt.key === ' ' || (evt.key === 'Enter' && !dragging)) {
-      onDragStart(dragItem.id);
+      onDragStart && onDragStart(dragItem.id);
       this.setState({
         dragging: true,
         draggedItem: React.Children.toArray(children).find(
@@ -196,7 +196,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
           data.splice(to, 1, draggedItem);
           onDragFinish(data);
         } else {
-          onDragCancel();
+          onDragCancel && onDragCancel();
         }
       } else if (evt.key === 'ArrowUp') {
         this.setState(prevState => {
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -32,8 +32,8 @@ export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, '
   isCompact?: boolean;
   /** Determines which wrapping modifier to apply to the DataList */
   wrapModifier?: DataListWrapModifier | 'nowrap' | 'truncate' | 'breakWord';
-  /** Order of items in list */
-  itemOrder: string[];
+  /** Order of items in a draggable DataList */
+  itemOrder?: string[];
 }
 
 interface DataListState {
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -9,7 +9,7 @@ export enum DataListWrapModifier {
   breakWord = 'breakWord'
 }
 
-export interface DataListProps extends React.HTMLProps<HTMLUListElement> {
+export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, 'onDragStart'> {
   /** Content rendered inside the DataList list */
   children?: React.ReactNode;
   /** Additional classes added to the DataList list */
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -98,7 +98,8 @@ export class DataList extends React.Component<DataListProps, DataListState> {
 
   move = (arr: React.ReactElement[], oldIndex: number, newIndex: number) => {
     const ghost = React.cloneElement(this.state.draggedItem, {
-      className: css(this.state.draggedItem.props.className, styles.modifiers.ghostRow)
+      className: css(this.state.draggedItem.props.className, styles.modifiers.ghostRow),
+      'aria-pressed': true
     });
 
     arr.splice(oldIndex, 1);
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -37,7 +37,6 @@ export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, '
 }
 
 interface DataListState {
-  childrenCopy: React.ReactNode;
   draggedItemId: string;
   draggingToItemIndex: number;
   dragging: boolean;
@@ -58,8 +57,7 @@ export const DataListContext = React.createContext<Partial<DataListContextProps>
   isSelectable: false
 });
 
-// Moves i1 to i2 and the rest of the array shifts down
-function moveItem(arr: string[], i1: string, toIndex: number) {
+const moveItem = (arr: string[], i1: string, toIndex: number) => {
   const fromIndex = arr.indexOf(i1);
   if (fromIndex === toIndex) {
     return arr;
@@ -68,7 +66,7 @@ function moveItem(arr: string[], i1: string, toIndex: number) {
   arr.splice(toIndex, 0, temp[0]);
 
   return arr;
-}
+};
 
 export class DataList extends React.Component<DataListProps, DataListState> {
   static displayName = 'DataList';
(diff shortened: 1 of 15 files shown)


kmcfaul:
Re: changes in DataListCheck.

The `pf-c-data-list__item-control` div was built into the `DataListCheck` component, but now with the drag button, multiple components may reside under the control div. To prevent this being a breaking change, I edited `DataListCheck` to have a new property flag which will prevent the control div from wrapping the check, and added a new `DataListItemControl` component. 

So it's not breaking currently, but a flag will have to be passed to `DataListCheck` to get the right HTML structure. Later on we can remove the flag and remove the control div from the check component.

mcarrano (changes requested):
@kmcfaul This is looking good. Just 2 problems that I'm seeing. On Firefox, there seems to be some problem with dropping the item. It sometimes is staying in the disabled state after I release the mouse. Have you tested on Firefox? Also, the cursor behavior is a bit flaky. As I drag the item, the cursor may update as I'm moving over other items below it.

kmcfaul:
@mcarrano I'll check Firefox, maybe it wants the onDrop handler over dragEnd. 

As far as the pointer, I would guess it is because the dragging doesn't interfere with css below it. I'm not sure if there's anything I can do about that, but I'll look into it. Any thoughts @mcoker ?

jessiehuff (changes requested):
Nice! This is looking great so far! The keyboard interaction works pretty well (though I wonder if we should include enter as one of the activation key presses like this [demo](https://schne324.github.io/dragon-drop/demo/)), but I think there's some a11y updates that we should add to make it accessible for screen readers. Jenn and I hashed out all of the details in the core PR starting [here](https://github.com/patternfly/patternfly/pull/3401#pullrequestreview-478826912), but essentially I think a structure like the core example would be great, potentially including the keyboard interaction instructions above the example as they did as well: 
Screen Shot 2020-09-30 at 11 16 58 AM

Some notable points from it:
- More a11y on the button: `aria-label` with "Reorder" telling users what they can do, `aria-pressed` with true/false depending on whether the button is selected, `aria-labelledby` the item, `aria-describedby` with the description text of how to use the reorder button.
- `aria-live="assertive"` as a div on the page that announces what is happening for users with the screen reader class.
- I also notice that the checkboxes are labelled by an id that doesn't exist? It looks like the id of the list item is "simple-item1" instead of "check-item1".
(I think making these changes should also fix the axe violations happening. 🙂)

mcoker:
@kmcfaul when I test on Firefox, the row I'm dragging is really small. Are you seeing that? 
Untitled 3

I'm also seeing the cursor issue. I can't see what's happening with the row in the inspector when I drag it. Can you describe what happens to the row that follows the cursor as you drag it over the data list? Is it cloned and absolutely positioned? I think we should add a class to the row that's being dragged and assign pointer styles to it. I would also like to assign a `z-index` so we can assure it is positioned on top of other things in the data list that might have their own `z-index`

kmcfaul:
@mcoker As far as I can tell, the drag image is a literal image of the row, positioned on the cursor. I don't think the cursor sees it at all? Everything it is dragging over would have to have that pointer styling (or maybe it's possible to programmatically set the styling on the body while it's dragging, and remove it when the drag ends/is cancelled). The drag image is part of HTML 5 handler events so browsers may handle it slightly differently.

Setting a custom drag image is possible, but it's expecting an image/simple element, not a React Element / clone of something.

kmcfaul:
@jessiehuff The check aria-labelledby I think I just misunderstood when I was editing examples, it should probably be labelled by the items' ids.

Should the `aria-live="assertive"` be applied to every item or is it fine to have it on the `ul` of the DataList?

mcoker:
@kmcfaul ah, gotcha! in that case we should be able to do something like this? https://codepen.io/mcoker/pen/oNxRRbe

kmcfaul:
That might work. Do we need to add a new class for it or can it be detected via the `ghostrow` modifier we already apply when dragging?

kmcfaul:
Hmm that doesn't seem to work (`e.target.style.cursor = 'grabbing'` at least, applied on dragStart and removed on dragEnd). It applies to the grabbed row but I the cursor still changes so I don't think it's getting transferred to the drag image. Maybe if it's a global css it's different?

From what I've looked up so far, it could be the text selection in Chrome just overrides the pointer? FF doesn't seem to have this issue either.

nicolethoen:
Just popping in here to say, @kmcfaul this is really cool! I'll take my time walking through to review it shortly, but playing with it is so fun!

jessiehuff:
I completely understand what you mean about the `aria-live`, it can be confusing. 🙂 When you want to be able to give screen reader users live updates about what is happening (something a sighted user can usually get visually), you can create a container with `aria-live` on it so that any information you want to give them gets sent to that container to be announced. This container has to live on the page prior to any of the actions being announced. So, for example:
Screen Shot 2020-10-01 at 11 00 11 AM
In the demo example, they have a `div` set to `aria-live` and then as you click the drag button, it announces "{item name} grabbed" or "{item name} dropped". This text gets inserted into the `aria-live` container. In this demo, this text gets removed after a certain amount of time although that may not always be the case depending on the use case. Our [toast alerts](https://www.patternfly.org/v4/components/alert-group/) look something like this: 

```<div class="pf-c-alert pf-m-success" aria-label="Success Alert" data-ouia-component-type="PF4/Alert" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-Alert-success-2" aria-live="polite" aria-atomic="false">``` When users initiate the alert, that information gets put into the `aria-live` container and then is announced by screen readers. When the toast alert is closed, then it gets removed from the container. I think it'd be preferrable to handle it the way this demo does and have it removed from the container after it's announced or some nominal amount of time, but I think that part is somewhat flexible.
Does that all make sense? Also, there's a part of me that keeps thinking to use enter to initiate the drag as well as space. Would that be something easy to add? I was talking to Jenn about what her expectations between the two would be, and she was thinking we should try to hook into whatever the typical onClick event would be. So if a browser uses enter to select buttons, enter would work, and if the browser supported/used space to select buttons, space would work.

I agree with Nicole though, this looks awesome!!! 😄

kmcfaul:
@jessiehuff Added the aria-live div and support for updating its text, let me know if I approached this the right way!

I also added logic for `Enter` to begin the dragging operation, in addition to space.

jessiehuff (commented):
Small nit in terms of the string used in dragging the items, but otherwise looking great!!! (It currently uses the index- i.e. "Dragging index 0 to 1", but I'm not sure it's completely ideal since it dynamically changes during drag? Perhaps we could just be more generalized and say "Dragging Item 1". Other than that, the keyboard interaction is awesome, and I think it works in VO (it's slightly tough to test because of the previous role that was added ("listbox") which is interfering with testing (removing in #4913). VO seems to really not like nesting form elements inside of other form elements like buttons inside of list boxes. I can pull it down though and check locally.

mcoker (dismissed):
LGTM! We can work on the cursor styles in the next sprint.

nicolethoen (commented):
in the draggable example, should I be able to check a checkbox by hitting enter when focusing on the box? I dont think that's working.

jessiehuff:
> in the draggable example, should I be able to check a checkbox by hitting enter when focusing on the box? I dont think that's working.

Technically, checkboxes use [space](https://webaim.org/techniques/keyboard/) to check and uncheck. We do have some inconsistency with this though since our Select checkboxes use enter. It seemed out of scope to mention in previous Select PRs though, but I'll double check that we have an issue for that future enhancement.

kmcfaul:
@nicolethoen I haven't touched the checkbox interactions, so they shouldn't have changed. I believe it's `space` to interact with them as Jessie mentioned.

redallen (changes requested):
Whenever I drag an item the page scrolls down:
scrollBug

kmcfaul:
@redallen I've been looking into this one. The problem is that when the list is re-ordered, it is splicing out the original index and inserting the ghost. So I believe it's a race condition where the rendering is updating and `dragOver` is getting called again on the index above the ghost.

mcarrano (dismissed):
This is looking much better @kmcfaul . Works for me on Firefox now.  I know there are still some issues to work out, but I give the OK to merge this when you feel it is ready @redallen .

kmcfaul:
If anyone has any thoughts about the unwanted upward scroll in the screen when you drag upwards (or arrow up with the keyboard while dragging), feel free to chime in. I'm having a lot of trouble pinpointing the cause of the issue. 

I could add a timeout to `dragOver` but that doesn't stop the continual scroll, just delays it by the timeout.

kmcfaul:
Found the issue!

`overflow-anchor: none;` needs to be applied to the `ul` and it will stop the scrolling (which will stop the constant dragging). Will that styling be okay to add into `.pf-c-data-list` or should we add a specific class/modifier for dragging that includes this? Will this cause other css issues @mcoker?

kmcfaul:
Inline styling will be replaced later with a new class that applies the css

mcoker (dismissed):
@kmcfaul nice find!! Created https://github.com/patternfly/patternfly/issues/3544 to follow up in core and add a class for this style.

I noted in https://github.com/patternfly/patternfly/issues/3544 but I think it might also be worthwhile to see if maybe a different way of moving the ghost row around could fix the root problem, allowing `overflow-anchor` to be enabled.

redallen (changes requested):
When I check an item and then drag it it becomes unchecked. Is this intended behavior?
dragBug

mcarrano:
Good catch @redallen . No that is not intended from a design perspective. In fact I did some more testing on this and looks like when you move an element it clears the selection not just on the element you moved but on the entire list. @kmcfaul is it possible to change this so that the state of items is maintained after something is moved?

kmcfaul:
@mcarrano the newest update that @redallen and I worked on should fix this, once it finishes building!

redallen (approved):
It works!!

kmcfaul:
Rebased

redallen (approved):
It still works!

mcoker (approved):

mcarrano (approved):
Looks great! Thanks for all the work on this @kmcfaul .

jessiehuff (approved):
- The biggest issue I see is that the Esc key no longer works in stopping the action. This makes it not a great user experience and rather confusing for keyboard users. You can tab out of it so it's not a keyboard trap from what I can tell (which is good), but it has some odd behavior and styling in this scenario: 
Screen Shot 2020-10-05 at 9 42 40 AM

- Some minor things that could be fixed are that because the ids on the page are not unique, when testing with screen reader, it thinks it's labelled by previous examples so the VO output is odd. Not urgent though as it reads fine when isolated by itself.


- We can open another issue as I don't want to hold anything up. However, the Esc key is a pretty big issue that we would need to prioritize.

PR #4854, opened 2020-09-21
fix(ToolTip/index.ts): Exports TooltipProps as a type to help babel.

### LINKED ISSUE #4853: Exported type in ToolTip
`Tooltip/index.ts` exports TooltipProps which is solely a type. When compiled this type is removed thowing an error when the `--isolatedModules` flag is used.

This is a bug but it is not blocking. Our current workaround is to use version 4.47.3 of `react-core`.

n/a

gradybknight:
TooltipProps is exported but contains only TS code. Babel removes this at compile time throwing an
error with --isolatedModules flag.

4853

**What**: Closes #4853 

**Additional issues**: n/a


KKoukiou (commented):
It seems like 'TooltipProps' was incorrectly exported in the first place.
Seems that there are more cases in the codebase the Properties types are exported - but honestly I don't see the usecase for it. 
Unless you have a usecase to use it, I would rather not export it.

KKoukiou (commented):
This won't work either - CI fails with:
packages/react-table/src/components/Table/base/types.tsx:9:10 - error TS2305: Module '"../../../../../react-core/dist/js"' has no exported member 'TooltipProps'.

9 import { TooltipProps, PopoverProps } from '@patternfly/react-core';
           ~~~~~~~~~~~~

packages/react-table/src/components/Table/HeaderCellInfoWrapper.tsx:5:36 - error TS2305: Module '"../../../../react-core/dist/js"' has no exported member 'TooltipProps'.

5 import { Button, Tooltip, Popover, TooltipProps, PopoverProps } from '@patternfly/react-core';
                                     ~~~~~~~~~~~~

and much more errors since they expect in many places that that property is globally exported.

I am wondering if the correct approach is to remove it as you did and then import from direct file:
```
import { TooltipProps } from '@patternfly/react-core/src/components/Tooltip/Tooltip
```
in all internal usages of it.


Maybe we should just not export it as a type and put it bellow with the other exports. Maybe @redallen could give a better solution.

gradybknight:
Sorry about the multiple failures. I think the first solution was probably the most benign one: my understanding is that exporting as type (e.g. `export type {TooltipProps} from './ToolTip'`). This way all the files which point to index still have access to the type but Babel can safely compile it away as needed.

KKoukiou:
> Sorry about the multiple failures. I think the first solution was probably the most benign one: my understanding is that exporting as type (e.g. `export type {TooltipProps} from './ToolTip'`). This way all the files which point to index still have access to the type but Babel can safely compile it away as needed.

Sure - sorry for misleading. Would you like to also update packages/react-core/src/components/Popover/index.ts ? 
It suffers from the same problem.

gradybknight:
Yep, will do. It's probably going to be at the end of the work day though (US time).

redallen:
Hey @gradybknight,

I'm a little confused why you are using babel on `@patternfly/react-core/src/components/Tooltip/index.ts`. We provide transpiled JS in `@patternfly/react-core/dist/esm/components/Tooltip/index.js` or `@patternfly/react-core/dist/js/components/Tooltip/index.js`. This compiled JS does not include types as exports, but rather only `export { Tooltip, TooltipPosition } from './Tooltip';`.

We can easily support the `"isolatedModules": true` option if needbe by adding it to `package/tsconfig.base.json` and fixing the resulting errors from `yarn build`. There aren't that many.

Thanks!
- Zack

jschuler (dismissed):

jschuler (approved):

KKoukiou (approved):

PR #4855, opened 2020-09-22
feat(select): pass chip group props on typeaheadmulti variant

### LINKED ISSUE #4766: TypeAhead Multi Select should provide a way to pass props to the ChipGroup of its selections
I'm building a `<Select variant="typeaheadmulti">` and only 3 selections are shown as chips before the overflow controls appear. It would be great to be able to customize that by passing my own `numChips` prop to the `<ChipGroup>` rendered here: https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Select/Select.tsx#L527

Perhaps an optional prop for `<Select>` called `chipGroupProps` could be made available that is only used if `variant === 'typeaheadmulti'`, and the chip group could be customized in other ways as well.

boaz0:
**What**: Closes #4766 

//cc @mturley

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -10,7 +10,7 @@ import { SelectOption, SelectOptionObject } from './SelectOption';
 import { SelectGroup } from './SelectGroup';
 import { SelectToggle } from './SelectToggle';
 import { SelectContext, SelectVariant, SelectDirection, KeyTypes } from './selectConstants';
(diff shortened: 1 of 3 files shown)


mturley (approved):
@boaz0 this is awesome! LGTM

mcarrano (approved):
This is a good enhancement. Thanks for adding this @boaz0 !

tlabaj (approved):
LGTM

boaz0:
Thank you all!

PR #4859, opened 2020-09-22
Feat/4820 button inline link

### LINKED ISSUE #4820: Inline link variation of button component that acts like a link
https://github.com/patternfly/patternfly/pull/3470/

This is a variation of an inline link button that behaves like regular link text. We can't use a native `<button>` html element and have text display inline and wrap to multiple lines like regular text/links, so if a user has an inline link button that is long and they want to wrap like regular text, we need to use a `<span>` instead of `<button>` with `role="button"` and `tabindex="0"` so it's focusable. When it's disabled, we need to use `pf-m-disabled` and set `aria-disabled="true"` and `tabindex="-1"`, similar to when an `<a>` is used as the lement for the button component.

evwilkin:
**What**: Closes #4820 

This PR adds the inline link as span, which uses a `span` with `role="button"` to allow text to wrap as a normal inline link would.


packages/react-core/src/components/Button/examples/Button.md
```diff
@@ -168,6 +168,22 @@ LinkButton = () => (
 );
 ```
 
+## Inline link as span
```
redallen, line 171: This should be an h3. Change to `### Inline link as span` to fix the "Untitled example":
image

packages/react-core/src/components/Button/examples/Button.md
```diff
@@ -168,6 +168,22 @@ LinkButton = () => (
...
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+
+InlineLinkAsSpan = () => (
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+    <Button variant="link" isInline component="span">
+      This is long button text that needs to be a span so that it will wrap inline with the text around it.
+    </Button>
+    Sed hendrerit nisi in cursus maximus. Ut malesuada nisi turpis, in condimentum velit elementum non.
```
redallen, line 182: Change to `{' '}Sed ...` to preserve whitespace.

redallen (changes requested):
Implementation looks good, docs just need fixing.

packages/react-core/src/components/Button/examples/Button.md
evwilkin: Updated - thanks!

packages/react-core/src/components/Button/examples/Button.md
evwilkin: Updated

yarn.lock
```diff
@@ -209,7 +209,7 @@
     "@babel/helper-explode-assignable-expression" "^7.10.4"
     "@babel/types" "^7.10.4"
 
-"@babel/helper-builder-react-jsx-experimental@^7.10.4", "@babel/helper-builder-react-jsx-experimental@^7.11.5":
```
redallen, line 212: Can you remove this file by running `git checkout upstream/master -- yarn.lock`?

redallen (changes requested):

redallen (dismissed):
LGTM!

yarn.lock
evwilkin: Updated - thanks for the help

jessiehuff (approved):
LGTM! 😄 Only thing to note here is that because it's not an actual button, the onclick event will not fire on enter and spacebar keys with a `<span>`. That means you have to add an onkeydown event and check if the enter or spacebar keys are pressed, then prevent the default behavior for the spacebar key and run the code.
Here is a [resource](https://www.deque.com/blog/accessible-aria-buttons/#:~:text=Keyboard%20Focusability,-HTML%20Native%20Controls&text=You%20need%20a%20for,to%20make%20it%20keyboard%20operable.) where Deque goes over these considerations. The section I'm referring to is "Custom ARIA `<span role=”button”>`."

evwilkin:
@tlabaj wanted to get your feedback here - is this something we should note for users, perhaps by adding a quick blurb beneath the new `Inline link as span` example?

> LGTM! 😄 Only thing to note here is that because it's not an actual button, the onclick event will not fire on enter and spacebar keys with a `<span>`. That means you have to add an onkeydown event and check if the enter or spacebar keys are pressed, then prevent the default behavior for the spacebar key and run the code.

evwilkin:
@jessiehuff @tlabaj thanks for the feedback - I've added an example & text within it explaining how to use the `onKeyDown` listener when the using the inline link as span to maintain keyboard interactions.

jessiehuff (dismissed):
LGTM! Thanks! 😄

mcoker (dismissed):
Nice!! LGTM!

redallen (dismissed):

redallen (approved):

jessiehuff (approved):

PR #4860, opened 2020-09-22
Table: Allows expanded content to have multiple cells

### LINKED ISSUE #4858: Table - Expanded content is always one cell (<td>)

I have a table that shows an expanded content that aligns with the row headers, I added the cells to the expanded content (a cells inside the row with with the correct parent id).

**Expected behavior**: All the cells show as I configured it.
**Unexpected behavior**: It always takes the first cell and applies a colSpan of the total number of headers.

Notice that this only happens when `onCollapse` is specified on the `Table` component:

**With `onCollapse`**
Screenshot_2020-09-22_10-15-39

**Without `onCollapse` (The way I want to have it minus the toggle)**
Screenshot_2020-09-22_10-14-41

`child - 1 with more rows` should have all the rows
 
https://codesandbox.io/s/musing-silence-8bc0z?file=/index.js

It could be a bug or by design (I suppose), is blocking some of my work and I don't have any workaround yet.

Insights Notifications UI

josejulio:
**What**: Closes #4858

Screenshot_2020-09-22_14-24-02

### DIFF
--- a/packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
+++ b/packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
@@ -56,7 +56,7 @@ export const expandedRow = (colSpan: number) => {
         extraParams: { contentId = 'expanded-content' }
       }
     }: IExtra
-  ): decoratorReturnType => {
+  ): decoratorReturnType | false => {
     if (rowData.hasOwnProperty('parent')) {
       const formatter: decoratorReturnType = {
         id: contentId + rowIndex,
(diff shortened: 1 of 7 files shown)


packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
```diff
@@ -56,12 +56,51 @@ export const expandedRow = (colSpan: number) => {
...
+    }: IExtra
+  ): decoratorReturnType | false => {
+    if (rowData.hasOwnProperty('parent')) {
+      const formatter: decoratorReturnType = {
+        id: contentId + rowIndex + '-' + columnIndex,
+        className: rowData.noPadding && css(styles.modifiers.noPadding)
+      };
+
+      if (!rowData.useAllCellInExpandedContent) {
+        formatter.skipCell = true;
```
josejulio, line 97: Tried using `isVisible: false`, but it always ended up being `true`. Not sure if is the way the transformation work, or if the `mapProps` (IIRC that's the name of the last transformation) does that. But i didn't set any `isVisible: true` on my code.

packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
```diff
@@ -56,12 +56,51 @@ export const expandedRow = (colSpan: number) => {
...
       }
     }: IExtra
-  ): decoratorReturnType =>
-    rowData.hasOwnProperty('parent') && {
-      // todo: rewrite this logic, it is not type safe
-      colSpan: colSpan + (!!rowData.fullWidth as any),
-      id: contentId + rowIndex,
-      className: rowData.noPadding && css(styles.modifiers.noPadding)
-    };
+  ): decoratorReturnType | false => {
```
boaz0, line 59: Can we change this to `Nullable<decoratorReturnType>` instead of `decoratorReturnType | false` and return in the end `null` instead of `false`?

boaz0 (approved):
LGTM

packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
josejulio: Is there any  `Nullable` type ? else I can just use `decoratorReturnType | null` ? 
I only see `NonNullable`.

josejulio: Or maybe you would prefer if I add `export type Nullable<T> = T | null;` to  [typeUtils](https://github.com/patternfly/patternfly-react/blob/2210ec2c142a385f85de8b0bac9c61a61521186d/packages/react-core/src/helpers/typeUtils.ts)?

redallen: I think `decoratorReturnType | null` will work fine as a type. We do this many other places in PatternFly.

boaz0 (approved):
LGTM :+1:

jschuler (changes requested):
Hi @josejulio thanks for this PR. I think we can approach this simpler without the need for additional props, can you see if this approach would work for you? https://github.com/patternfly/patternfly-react/pull/4899 - thanks

josejulio:
> Hi @josejulio thanks for this PR. I think we can approach this simpler without the need for additional props, can you see if this approach would work for you? #4899 - thanks

That would be great. I'm going to test the PR locally. Thanks!

josejulio:
Closing in favor of #4899

PR #4867, opened 2020-09-23
fix(ChipGroup): fix axe issue or decide to ignore in script

### LINKED ISSUE #4866: Chip group with tooltip creates axe issue
When an element is `aria-hidden`, as a general rule, that element should not be focusable. It looks like we did this when chip group has a tooltip in order to give focus to the tooltip for keyboard users (with `tab-index="0"`). However, we have `aria-hidden` so that screen readers don't hear the category name twice and create annoying, redundant information.

Unfortunately as it is right now, if a consumer adds a tooltip to the chip group category name, they will get axe errors which could affect their overall Axe score when reviewing their product's accessibility. Our options are to rework the structure slightly and see if we can get a proper structure or to ignore this axe rule and to tell our products to also ignore this axe rule.

Screen Shot 2020-09-23 at 11 10 40 AM

jessiehuff:
**What**: Closes #4866


redallen:
Conflicts with #4879. Might want to close in favor of that instead? I can add an aria-label there.

PR #4871, opened 2020-09-23
feat(Modal): add top align to modal

### LINKED ISSUE #4819: Top aligned modal
https://github.com/patternfly/patternfly/pull/3435

Adds `.pf-m-align-top` to the modal component that aligns the modal box to the top.

We can add a custom prop that allows a user to change the distance from the top of the modal to the top of the window by setting an inline style that sets the `--pf-c-modal-box--m-align-top--spacer` var. Though I found a bug with that and it should be fixed when https://github.com/patternfly/patternfly/issues/3480 is merged.

kmcfaul:
**What**: Closes #4819

Adds `position` property for top aligned modals. @mcoker Let me know if this property makes sense. I could change it to something like `isTopAligned` but I wasn't sure if we wanted to add other position type alignments (bottom/left/right) later on.

Adds `distanceFromTop` property to set the custom spacer class `--pf-c-modal-box--m-align-top--spacer`.

### DIFF
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -43,8 +43,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
   variant?: 'small' | 'medium' | 'large' | 'default';
   /** Alternate position of the modal */
   position?: 'top';
-  /** Custom distance from top */
-  distanceFromTop?: string;
+  /** Custom spacer distance from position */
+  distance?: string;
   /** Flag indicating if modal content should be placed in a modal box body wrapper */
   hasNoBodyWrapper?: boolean;
   /** An ID to use for the ModalBox container */
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -41,6 +41,10 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
   description?: React.ReactNode;
   /** Variant of the modal */
   variant?: 'small' | 'medium' | 'large' | 'default';
+  /** Alternate position of the modal */
+  position?: 'top';
+  /** Custom distance from top */
+  distanceFromTop?: string;
   /** Flag indicating if modal content should be placed in a modal box body wrapper */
   hasNoBodyWrapper?: boolean;
   /** An ID to use for the ModalBox container */
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -43,8 +43,8 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
   variant?: 'small' | 'medium' | 'large' | 'default';
   /** Alternate position of the modal */
   position?: 'top';
-  /** Can be any valid CSS length/percentage */
-  distance?: string;
+  /** Offset from alternate position. Can be any valid CSS length/percentage */
+  positionOffset?: string;
   /** Flag indicating if modal content should be placed in a modal box body wrapper */
   hasNoBodyWrapper?: boolean;
   /** An ID to use for the ModalBox container */
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -43,7 +43,7 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
   variant?: 'small' | 'medium' | 'large' | 'default';
   /** Alternate position of the modal */
   position?: 'top';
-  /** Custom spacer distance from position */
+  /** Can be any valid CSS length/percentage */
   distance?: string;
   /** Flag indicating if modal content should be placed in a modal box body wrapper */
   hasNoBodyWrapper?: boolean;
--- a/packages/react-core/src/components/Modal/ModalBox.tsx
+++ b/packages/react-core/src/components/Modal/ModalBox.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/ModalBox/modal-box';
+import c_modal_box_m_align_top_spacer from '@patternfly/react-tokens/dist/js/c_modal_box_m_align_top_spacer';
 
 export interface ModalBoxProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside the ModalBox. */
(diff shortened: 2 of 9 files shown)


packages/react-core/src/components/Modal/Modal.tsx
```diff
@@ -41,6 +41,10 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
   description?: React.ReactNode;
   /** Variant of the modal */
   variant?: 'small' | 'medium' | 'large' | 'default';
+  /** Alternate position of the modal */
+  position?: 'top';
+  /** Custom distance from top */
+  distanceFromTop?: string;
```
mcoker, line 47: if we support a generic `position` prop, can we have a generic `distance` prop?

kmcfaul: Sure, I'll update this!

mcoker (approved):
LGTM!

packages/react-core/src/components/Modal/ModalBox.tsx
```diff
@@ -36,10 +42,17 @@ export const ModalBox: React.FunctionComponent<ModalBoxProps> = ({
...
       className,
+      position === 'top' && styles.modifiers.alignTop,
       variant === 'large' && styles.modifiers.lg,
       variant === 'small' && styles.modifiers.sm,
       variant === 'medium' && styles.modifiers.md
     )}
+    {...(distance && {
+      style: {
+        '--pf-c-modal-box--m-align-top--spacer': distance,
+        ...props.style
```
redallen, line 53: I think you should destructure `style` instead of relying on the `{...(distance && {}}` tag being after `{...props}`.

Then you could write
```jsx
style={{
  ...(distance && { '--pf-c-modal-box--m-align-top--spacer': distance }),
  ...style
}}
```

packages/react-core/src/components/Modal/ModalBox.tsx
```diff
@@ -9,6 +9,10 @@ export interface ModalBoxProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Variant of the modal */
   variant?: 'small' | 'medium' | 'large' | 'default';
+  /** Alternate position of the modal */
+  position?: 'top';
+  /** Custom spacer distance from position */
```
redallen, line 14: Add `Can be any valid CSS distance.`

redallen (changes requested):

packages/react-core/src/components/Modal/ModalBox.tsx
mcoker: [Length](https://developer.mozilla.org/en-US/docs/Web/CSS/length) is generally how those values are referred to. Maybe "CSS length"?  Technically it would be "length/percentage" since [length-percentage](https://developer.mozilla.org/en-US/docs/Web/CSS/length-percentage) are a separate thing.

redallen (dismissed):

jessiehuff (approved):
LGTM! :)

packages/react-core/src/components/Modal/ModalBox.tsx
```diff
@@ -36,10 +42,15 @@ export const ModalBox: React.FunctionComponent<ModalBoxProps> = ({
     className={css(
       styles.modalBox,
       className,
+      position === 'top' && styles.modifiers.alignTop,
       variant === 'large' && styles.modifiers.lg,
       variant === 'small' && styles.modifiers.sm,
       variant === 'medium' && styles.modifiers.md
     )}
+    style={{
+      ...(distance && { '--pf-c-modal-box--m-align-top--spacer': distance }),
```
jschuler, line 51: you could also import `c_modal_box_m_align_top_spacer` from react-tokens and use 
`c_modal_box_m_align_top_spacer.name` as the key

tlabaj (changes requested):

packages/react-core/src/components/Modal/ModalBox.tsx
```diff
@@ -9,6 +9,10 @@ export interface ModalBoxProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Variant of the modal */
   variant?: 'small' | 'medium' | 'large' | 'default';
+  /** Alternate position of the modal */
+  position?: 'top';
+  /** Can be any valid CSS length/percentage */
+  distance?: string;
```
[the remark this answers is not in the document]
mcoker: I'm also not sure about the name "distance." I wonder if "offset" or maybe "positionOffset" might be better?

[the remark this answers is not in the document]
kmcfaul: I'll update the name, I do think `positionOffset` is more descriptive

packages/react-core/src/components/Modal/ModalBox.tsx
kmcfaul: Syntax complains when I tried that, it doesn't like the `.name` and doesn't like `['name']` when it's used to index a property for an object.

packages/react-core/src/components/Modal/ModalBox.tsx
```diff
@@ -36,10 +42,15 @@ export const ModalBox: React.FunctionComponent<ModalBoxProps> = ({
     className={css(
       styles.modalBox,
       className,
+      position === 'top' && styles.modifiers.alignTop,
       variant === 'large' && styles.modifiers.lg,
       variant === 'small' && styles.modifiers.sm,
       variant === 'medium' && styles.modifiers.md
     )}
+    style={{
+      ...(positionOffset && { '--pf-c-modal-box--m-align-top--spacer': positionOffset }),
```
jschuler, line 51: i noticed that when i pass in positionOffset, it isn't just moved down vertically, but the width also shrinks, is that supposed to happen?

packages/react-core/src/components/Modal/ModalBox.tsx
redallen: Wouldn't `[c_modal_box_m_align_top_spacer.name]: distance` work?

packages/react-core/src/components/Modal/ModalBox.tsx
kmcfaul: @mcoker ^

mcoker: @jschuler @kmcfaul does this have the latest version of core? That should have been fixed in https://github.com/patternfly/patternfly/pull/3481

kmcfaul: It did not, now it does :)

redallen (approved):
LGTM!

jessiehuff (approved):

jschuler (approved):

tlabaj (approved):
LGTM

mcoker (approved):
LGTM!!

PR #4877, opened 2020-09-24
fix(demo-app): replace CRA with webpack and ts-loader

### LINKED ISSUE #4876: Workspace is slow to `yarn install`
`yarn install` takes over a minute locally and 1-3m in CI. A big culprit is the old version of Create React App we use in `demo-app-ts`.

redallen:
**What**: Closes #4876. Notice the -4.2k LOC in yarn.lock!

**Additional issues**: Since removing Gatsby, we don't need `react-helmet` anymore. Upgrading cypress also saves ~1m per parallel test.


evwilkin (approved):
LGTM

ddonahue007 (approved):
LGTM

PR #4879, opened 2020-09-24
Fix/readd patternfly a11y

### LINKED ISSUE #4878: Add back patternfly-a11y
The script was temporarily removed to upgrade react-docs and because it was being upgraded. Now it should be added back.

redallen:
**What**: Closes #4878

**Additional issues**:
This adds the a11y tester back to CI and fixes pages that developed problems in the meantime or had previously uncaught issues:
- http://patternfly-react-pr-4879.surge.sh/components/chip-group/react/chip-groups-with-categories-removable
- http://patternfly-react-pr-4879.surge.sh/components/label-group/react/vertical-category-overflow-removable
- http://patternfly-react-pr-4879.surge.sh/components/page/react-demos/tertiary-nav
- http://patternfly-react-pr-4879.surge.sh/demos/primary-detail/react-demos/primary-detail-full-page
- http://patternfly-react-pr-4879.surge.sh/demos/primary-detail/react-demos/primary-detail-content-padding
- http://patternfly-react-pr-4879.surge.sh/demos/primary-detail/react-demos/primary-detail-data-list-in-card
- http://patternfly-react-pr-4879.surge.sh/demos/primary-detail/react-demos/primary-detail-inline-modifier
- http://patternfly-react-pr-4879.surge.sh/components/table/react-demos/bulk-select
- http://patternfly-react-pr-4879.surge.sh/components/table/react-demos/column-management
- http://patternfly-react-pr-4879.surge.sh/components/table/react-demos/filterable


jessiehuff (commented):
- It looks like chip group and label group have the same issue. This was a somewhat tricky one because the reasoning for making it `aria-hidden` with the `tabindex` had a benefit to SR users, but it's more of a hacky workaround. The issue was that we wanted the tooltip to appear when using a keyboard (so it needed focus), but without the `aria-hidden` the category name gets announced 3 times: it will say "{category name} group, {category name}, then "end of {category name} group" before getting to the list which is a lot of redundant repetition. The thought process behind #4867 was to include an additional span within the tooltip span so that the internal category name span could remain hidden while the external tooltip would be announced. This allows it to only announce "{category name}, group" then the list.
- The tertiary nav issue is happening because `aria-controls` is pointing to something that technically doesn't exist in our demo since it doesn't look like the side nav is actually in this example. In a product implementation, this is correct but aria-controls would need to point to the id of their nav. Not necessarily certain on the best path forward on that one because having `aria-controls` is correct, but the rest of the implementation doesn't seem to be built out to show that to axe.
- The primary detail full page, I'm ok with taking off the `aria-labelledby`, but I'd rather update it to announce that it's selectable in some way. I assume that's why the `aria-labelledby` was added previously (similar to the previous example, it's pointing to an id that doesn't exist), but I'm not sure if it would need to have the `aria-labelledby` pointing to the drawer or if it would need updated with additional aria. I'm not positive, and it would need some testing to determine the best path forward. Maybe we can take it off for now and open another issue addressing these concerns at a later date. (Currently a SR user can't tell that selecting that item does anything at all, they likely would just assume it was a basic list since no other functionality is announced.) --- It looks like this is true for all of the primary detail examples here.
- The table examples look good. 
Thank you for working on this!! 😄

redallen:
- I applied your ChipGroup patch to this PR and added it to LabelGroup as well.
- I added an `aria-controls` prop to `PageHeader.tsx`. I find our previous default `"page-header"` unhelpful and no known product uses it. I think adding `aria-controls` to our examples is outside the scope of this PR but could be done.
- I've added `role="listbox"` to DataList and `role="option"` to DataListItem. Let me know if this is a poor solution and I'll revert it.
- Yay! Glad it works.

jschuler (approved):

ddonahue007 (approved):
LGTM

jessiehuff (approved):
Not a bad solution to these! Technically when using `role="listbox"` we need additional aria though like  `aria-activedescendant` when an option is visually focused and `aria-selected`/`aria-multiselectable` when an item is selected as this [WAI examples](https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/listbox/listbox.html) shows, but I'm ok with us opening another issue for that if you'd like. I definitely don't want to hold up this PR. 🙂   Excited by these updates!

redallen:
@jessiehuff If you could open an issue for this that'd be great!

PR #4887, opened 2020-09-24
Removes the clearBtn button when onClear is not specified for typeahead

### LINKED ISSUE #4886: Select component (variant typeahead and multiTypeahead) shows the clear button regardless if there is onClear or not

Created a `Select` with variant `typeahead` and `multiTypeahead`, without `onClear` function.

**Expected behavior:**  There is no button to clear
**Unexpected behavior:** The button actually shows

https://codesandbox.io/s/focused-moon-6buwr

Bug

Insights Notifications UI

josejulio:
(and typeaheadMulti)

**What**: Closes #4886

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -863,7 +863,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   ref={this.inputRef}
                 />
               </div>
-              {(selections[0] || typeaheadInputValue) && clearBtn}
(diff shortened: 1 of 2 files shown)


KKoukiou (commented):
The code looks correct but I would like to wait for an answer on #4886 because it seems like this is the desired design, to always contain the clear button.

redallen (approved):
I agree with you @josejulio .

If design really wants the clear button always showing, we should be consistent across all Select variants.

kmcfaul (approved):

PR #4888, opened 2020-09-24
feat(DatePicker): Add basic date picker

### LINKED ISSUE #4730: Build a basic date picker
Use Flatpickr

See notes from CSS investigation. 
https://github.com/patternfly/patternfly/issues/3256

### LINKED ISSUE #4817: Add ability to specify a custom icon used in text input component
https://github.com/patternfly/patternfly/pull/3453

This adds a `pf-m-icon` variation to the text inputs that allows a user to specify their own icon in the input. We have added 2 extra variations to the text input that use `pf-m-icon` for use with the date & time picker. Those are `.pf-m-calendar` and `.pf-m-clock` - added along with `.pf-m-icon`, the text input will show a calendar and clock icon.

We also added an example in core of how to use `.pf-m-icon` and add a custom icon via customizing CSS vars. All that's really needed is to provide the data URI of the image as the value for `--pf-c-form-control--m-icon--BackgroundUrl`.

To replicate the core example, you could add these as inline styles:

```
  --pf-c-form-control--m-icon--BackgroundSizeX: var(--pf-global--icon--FontSize--lg);
  --pf-c-form-control--m-icon--BackgroundSizeY: var(--pf-global--icon--FontSize--lg);
  --pf-c-form-control--m-icon--BackgroundUrl: url('data:image/svg+xml;charset=utf8,%3Csvg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"%3E%3Cpath fill="%23a18fff" d="M158.87.15c-16.16-1.52-31.2 8.42-35.33 24.12l-14.81 56.27c187.62 5.49 314.54 130.61 322.48 317l56.94-15.78c15.72-4.36 25.49-19.68 23.62-35.9C490.89 165.08 340.78 17.32 158.87.15zm-58.47 112L.55 491.64a16.21 16.21 0 0 0 20 19.75l379-105.1c-4.27-174.89-123.08-292.14-299.15-294.1zM128 416a32 32 0 1 1 32-32 32 32 0 0 1-32 32zm48-152a32 32 0 1 1 32-32 32 32 0 0 1-32 32zm104 104a32 32 0 1 1 32-32 32 32 0 0 1-32 32z"/%3E%3C/svg%3E');
```

cc @nicolethoen not sure how much of this will be done with your current date & time picker work?

nicolethoen:
Closes #4730 
Closes #4817 

**Includes the new custom icon variant of the text input**

There are a few things that have evolved since this design doc, but for reference:
https://docs.google.com/document/d/1hlS0R1l7gnoe-YDu1SEHpUSrf6WMuZPejR5AwYr0a20/edit#heading=h.suoqe71652b5


packages/react-date-time/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "peerDependencies": {
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
+  },
+  "devDependencies": {
+    "rimraf": "^2.6.2",
+    "typescript": "^3.8.3",
+    "flatpickr": "^4.6.6",
+    "moment": "^2.27.0"
```
redallen, line 51: These should be listed as dependencies. Also, please pin them to exact versions instead of using the `^` semver.

I'm hesitant to add `moment`. I'm not sure how well it code-splits, and many of its functions have been replaced by browser standards. They recently stated the [project has reached legacy status.](https://momentjs.com/docs/#/-project-status/) If all we need is ISO 8601 parsing [Date.prototype.toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) and [Date.parse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) should be supported consistently in all ES5 browsers (PatternFly currently targets ES6).

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,196 @@
...
+      dateFormat,
+      locale: (FlatpickrLanguages as any)[locale],
+      minDate,
+      maxDate,
+      defaultDate: value,
+      allowInput: true,
+      static: true,
+      allowInvalidPreload: true,
+      parseDate: (datestr: string, format: string) => moment(datestr, [format, moment.ISO_8601], true).toDate(),
+      formatDate: (date: Date, format: string) => moment(date).format(format),
```
redallen, line 89: This should be possible to do only using browser standards.

packages/react-date-time/src/components/DatePicker/examples/DatePicker.md
```diff
@@ -0,0 +1,36 @@
+---
+id: Date picker
+section: components
```
redallen, line 3: Shouldn't this be listed under extensions? If we're fully committing to it as a component, we should add a message like we do with Table:
```md
Note: Table lives in its own package at [@patternfly/react-table](https://www.npmjs.com/package/@patternfly/react-table)!
```

packages/react-date-time/package.json
nicolethoen: I believe we will be adding moment when we get to adding the timepicker especially because of the desires to build in timezone pickers and such. So I can probably not rely on it for the date picker, but I felt that since we will be using it for the time picker and the date and time picker will eventually be used in tandem, it made sense to make sure the date picker was thinking like moment. but you're right that it probably isn't necessary yet.

redallen: Timezone locale formatting has come a long way since `moment` was created and is also supported in ES5 browsers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString. 

If there's a set of functions you need implemented I'm happy to find lighter alternatives to `moment`!

mcarrano (commented):
This is looking great @nicolethoen !  I only noticed one small visual issue. Looks like the text for the month field is getting chopped off for letters that extend below the line. See below:

Screen Shot 2020-09-25 at 1 50 35 PM

@maryshak1996 It would be great is you can also take a look at the visuals.

mcoker:
> the text for the month field is getting chopped off 

@mcarrano I'll take a look at that

packages/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -60,6 +60,12 @@ export interface TextInputProps
   onFocus?: (event?: any) => void;
   /** Callback function when input is blurred (focus leaves) */
   onBlur?: (event?: any) => void;
+  /** icon variant */
+  iconVariant?: 'calendar' | 'clock';
+  /** Custom icon url to set as the input's background-image */
+  customIconUrl?: string;
```
jschuler, line 66: could this work with react-icons?

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+  componentDidMount() {
+    const { dateFormat, minDate, maxDate, locale, value } = this.props;
+
+    this.calendar = flatpickr(this.inputEl.current, {
+      dateFormat,
+      locale: (FlatpickrLanguages as any)[locale],
+      minDate,
+      maxDate,
+      defaultDate: value,
+      ariaDateFormat: dateFormat,
```
redallen, line 137: @jessiehuff Is this the right thing to do for a screen reader?

redallen (changes requested):
When I click on the bottom of the input I get a flash of a Date picker and then it goes away until I click again:
image

packages/react-date-time/README.md
```diff
@@ -0,0 +1,130 @@
+# @patternfly/react-table
+
+This package provides Date and Time PatternFly components based on [PatternFly 4][patternfly-4]
```
jschuler, line 3:
```suggestion
This package provides date and time PatternFly components based on [PatternFly 4][patternfly-4]
```

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+import styles from '@patternfly/react-styles/css/components/DatePicker/date-picker';
+import { DatePickerLocales } from './DatePickerUtils';
+import { TextInput } from '@patternfly/react-core/dist/js/components/TextInput/TextInput';
+
+export interface DatePickerProps
+  extends Omit<React.HTMLProps<HTMLInputElement>, 'onChange' | 'onFocus' | 'onBlur' | 'disabled' | 'ref'> {
+  /** Additional classes added to the date time picker. */
+  className?: string;
+  /** Flag indicating whether the value of the date picker is invalid. */
+  invalid?: boolean;
```
jschuler, line 16:
```suggestion
  isInvalid?: boolean;
```

packages/react-date-time/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@patternfly/react-date-time",
+  "version": "0.0.0",
```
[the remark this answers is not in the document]
redallen: It should start at `4.0.0`.

redallen:
Forgot to leave this in my review, but now is a good time to rename the package to `react-datetime` if we want.

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+  /** Accessible label for the date picker */
+  'aria-label'?: string;
+  /** The minimum date that a user can start picking from. */
+  minDate?: string;
```
jschuler, line 20: I think we should use the same types as flatpickr
https://github.com/flatpickr/flatpickr/blob/f4478d27fce40532298a738aab428b3bbbff751c/src/types/options.ts#L165 for all the props that are forwarded

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+
+function toNumber(str: string) {
+  if (str) {
+    return Number(str);
+  }
+  return NaN;
+}
+// Supports (yyyy|mm|dd) to parse string to Date object
+function parseYYYYMMDD(date: string, format: string) {
```
jschuler, line 55: flatpickr has a `parseDate` function, i don't think we need these functions here

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+      index += formatBlock.length;
+    }
+  }
+
+  // Year and month are required
+  return new Date(year, month, day);
+}
+
+// Supports (yyyy|mm|dd) to format Date object to string
+function formatYYYYMMDD(date: Date, format: string) {
```
jschuler, line 81: flatpickr has a `formatDate` function you can use

packages/react-date-time/README.md
```diff
@@ -0,0 +1,130 @@
...
+```js
+import '@patternfly/react-core/dist/styles/base.css';
+```
+
+#### Example Component Usage
+
+```js
+import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+
+class SimpleTable extends React.Component {
```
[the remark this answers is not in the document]
nicolethoen: lol right.... thanks nice catch

packages/react-core/src/components/TextInput/TextInput.tsx
nicolethoen: I do not believe so, because it modifies the css background image which needs a url string... @mcoker does that check out?

packages/react-date-time/src/components/DatePicker/DatePickerUtils.ts
```diff
@@ -0,0 +1,52 @@
+export enum DatePickerLocales {
```
jschuler, line 1: is this needed? flatpickr exports
https://github.com/flatpickr/flatpickr/blob/f4478d27fce40532298a738aab428b3bbbff751c/src/types/locale.ts#L113

jschuler (changes requested):

packages/react-date-time/src/components/DatePicker/DatePickerUtils.ts
nicolethoen: ah good call.

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
nicolethoen: So, flatpickr works best when you require the user to choose a date from the calendar. There are a few requirements in our design - especially the one that wants the user to be able to type in their date rather than be forced to use the calendar, that begin to pose some problems.

flatpickr uses weird date formatting tokens - not standard ones, so we simplify things quite a bit.
we do our own error handling for invalid date formats since flatpickr does not do very robust error handling. 

The last example on this page: https://flatpickr.js.org/examples/ sheds the very beginning of some light.

since we are allowing the user to type in their own input, which could be invalid - we handle the basic error handling, which includes parsing and formatting, before handing dates off to flatpickr

packages/react-date-time/package.json
mturley: A good set of resources on how to replace moment is here: https://github.com/you-dont-need/You-Dont-Need-Momentjs/blob/master/README.md

If the native APIs really don't cut it, Luxon is a much more lightweight alternative made by the same developer of Moment: https://moment.github.io/luxon/

mturley: Although @redallen the concern about using a heavy library here is maybe less of an issue for the general consumer since this is going into a separate `@patternfly/react-date-time` package. If for some reason we determine we really do need something like Moment at least it won't be in the bundle for those who only use core.

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+    this.calendar = flatpickr(this.inputEl.current, {
+      dateFormat,
+      locale: (FlatpickrLanguages as any)[locale],
+      minDate,
+      maxDate,
+      defaultDate: value,
+      ariaDateFormat: dateFormat,
+      allowInput: true,
+      static: true,
```
jschuler, line 139: not sure you'd want this to be true, if I understand this option correctly it would place the datepicker inline as opposed to appending it to document.body. This would cause the picker not to break out of a Modal or Wizard for example

mturley:
> When I click on the bottom of the input I get a flash of a Date picker and then it goes away until I click again:

Hmm @redallen I can't reproduce this, are you still seeing it?

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
nicolethoen: hm... okay. I grouped it with the input based on core's request. I'll double check with them.

packages/react-date-time/src/components/DatePicker/DatePickerUtils.ts
nicolethoen: except, if we import that, there is no beautiful way to document the options. So I figured building an enum that maps to flatpickr type allowed me possibly include the options in the docs and be sure the consumer is using a valid locale code.

packages/react-date-time/README.md
```diff
@@ -0,0 +1,130 @@
+# @patternfly/react-table
```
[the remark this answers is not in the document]
jschuler: actually there is no time component at the moment, right? or is it planned to be added

redallen:
@mturley Yes, I'm still seeing it on Chrome Version 85.0.4183.121 (Official Build) (64-bit). It's close to pixel-perfect... If you keep trying to click the bottom border you'll eventually hit it.

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
mturley: Should we expose that as a prop like we do with `menuAppendTo` in Dropdown and Select?

nicolethoen: that'd take a bit more experimentation. possibly.

nicolethoen: only because I need to figure out how that could impact how flatpickr handles positioning.

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
jschuler: yes, you can listen to input changes, and call this.calendar.parseDate(inputValue, formatToCheckAgainst) simply, no?

nicolethoen: I guess the trade off is instead of writing our own parser/formatter functions, we would need to create a mapping for the datepickr formatter tokens to standard tokens. Or link to the flatpickr doumentation in our docs, but I'd rather keep flatpickr more under the hood.

nicolethoen: https://flatpickr.js.org/formatting/

packages/react-core/src/components/TextInput/TextInput.tsx
mcoker: That's right, all of the icons in text input, form select (the down arrow), and text area are background images, so you can use anything you could pass to `background-image: url();`. If there were a utility to convert a react icon to a data URI, you could...

packages/react-date-time/README.md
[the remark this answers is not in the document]
nicolethoen: its the next story all part of this epic

packages/react-date-time/src/components/DatePicker/DatePicker.tsx
nicolethoen: okay. it looks like flatpickr has incorporated a feature that will give us more fine tuned control over the positioning of the calendar relative to an element. Right now, though it gives the calendar absolute position and calculates its `top` and `left` position based on the element the consumer tells it to append to. This causes problems for us based on PatternFly's current page setup. so once the latest feature gets released, we can probably append to body or another container, but for now, this is the best way to position the calendar - inline with the input.

jschuler (dismissed):
Nice job @nicolethoen , LGTM

packages/react-datetime/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,235 @@
...
+        '<svg aria-hidden="true" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 512"><path fill="currentColor" d="M31.7 239l136-136c9.4-9.4 24.6-9.4 33.9 0l22.6 22.6c9.4 9.4 9.4 24.6 0 33.9L127.9 256l96.4 96.4c9.4 9.4 9.4 24.6 0 33.9L201.7 409c-9.4 9.4-24.6 9.4-33.9 0l-136-136c-9.5-9.4-9.5-24.6-.1-34z" class=""></path></svg>'
+    });
+
+    window.document.body.addEventListener(
+      'keydown',
+      e => {
+        if (e.code === 'Enter' && e.target === this.inputEl.current) {
+          this.validateDate((e.target as HTMLInputElement).value, e as any);
+        }
+      },
```
redallen, line 129: Future work: on pressing `Escape` call `this.calendar.close()`.

redallen (approved):
I still have the issue when clicking the bottom border that it quickly opens/closes, but it's good enough for beta.

nicolethoen:
there is a 'documentClick' function in Flatpickr that fires when you release the mouse outside the input and closes the calendar. I will open an issue to investigate ways to clean this up. Not sure what the options are...

jschuler (approved):
let's get it in before another merge ;)

jessiehuff (approved):
Looks good to me on the a11y side! 🙂

mturley:
Really nice work @nicolethoen, this will be awesome to have!

Issue #4890, opened 2020-09-27
Select component encountering 'TypeError: Cannot read property 'id' of undefined' in createRenderableFavorites


mhuth:
**Describe the issue. What is the expected and unexpected behavior?**
Getting this error when using the `<Select>` component:
```
instrument.js:110 TypeError: Cannot read property 'id' of undefined
    at util.js:298
    at Array.filter (<anonymous>)
    at Object.t.createRenderableFavorites (util.js:298)
```

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

In our application the `<Select>` component has a few child components in it, including an array of `<SelectGroup>` and `<SelectOption>`s.  However the array element doesn't have a direct 'props' property, eg:
```
<Select>
  <InputGroup key: 'null'>...</InputGroup>
  <Divider key: "inline-filter-divider" ... />
  [array of ...
    <SelectGroup>
      <SelectOption>
     ...
      </SelectOption>
     </SelectGroup>
  ]
  <Button key: "manage all tags">...</Button>
</Select>
```
This is the value of 'items' at the line causing the problem in createRenderableFavorites in helpers/utils.js:
```
e: Array(4)
0: {$$typeof: Symbol(react.element), key: null, ref: null, props: {…}, type: ƒ, …}
1: {$$typeof: Symbol(react.element), key: "inline-filter-divider", ref: null, props: {…}, type: ƒ, …}
2: [{…}]
3: {$$typeof: Symbol(react.element), key: "manage all tags", ref: null, props: {…}, type: ƒ, …}
```
... where you can see items[2] is an array without a 'props' element.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Its a bug and the workaround is to use an earlier version of react-core without the problematic code in helpers/utils.js.

Proposed fix for this issue is:
```
diff --git a/packages/react-core/src/helpers/util.ts b/packages/react-core/src/helpers/util.ts
index 106b83d..f476425 100644
--- a/packages/react-core/src/helpers/util.ts
+++ b/packages/react-core/src/helpers/util.ts
...
@@ -319,7 +320,7 @@ export const createRenderableFavorites = (
     return favoriteItems;
   }
   return (items as React.ReactElement[])
-    .filter(item => favorites.includes(item.props.id))
+    .filter(item => 'props' in item && favorites.includes(item.props.id))
[...]

PR #4891, opened 2020-09-27
Fix TypeError if missing 'props' from favorite items

mhuth:
**What**:
Patch for https://github.com/patternfly/patternfly-react/issues/4890

**Additional issues**:
There are more `item.props.id` lines in that file that might also need modifying.  This patch doesn't address them all, just the one that was affecting our application.

### DIFF
--- a/packages/react-core/src/helpers/util.ts
+++ b/packages/react-core/src/helpers/util.ts
@@ -297,7 +297,7 @@ export const createRenderableFavorites = (
         return (
           group.props.children &&
           (group.props.children as React.ReactElement[])
-            .filter(item => favorites.includes(item.props.id))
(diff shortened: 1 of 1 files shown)


packages/react-core/src/helpers/util.ts
```diff
@@ -319,7 +319,7 @@ export const createRenderableFavorites = (
     return favoriteItems;
   }
   return (items as React.ReactElement[])
-    .filter(item => favorites.includes(item.props.id))
+    .filter(item => 'props' in item && favorites.includes(item.props.id))
```
KKoukiou, line 322: React.Element should have props and it's not correct to check this here. 
I answered in the issue you opened.

tlabaj:
This issue is fixed in PR #4772

PR #4892, opened 2020-09-28
feat(ChipGroup): update html structure for category

### LINKED ISSUE #4821: Update chip group structure to show close button at top/right
https://github.com/patternfly/patternfly/pull/3445/

This adds `div.pf-c-chip-group__main` around the label and list elements in a chip group with a label (category chip group). When the category chip group is closable, the `__main` element should be adjacent to the close button (`.pf-c-chip-group__close`)

kmcfaul:
**What**: Closes #4821


kmcfaul:
@mcoker Is this a dupe of #4873 ?
Because that looks like it would cover this case, if the main chip group div is always added.

mcoker:
@kmcfaul ah thanks! I didn't know there was a PR already up for this. After looking at it a closer, it's better if `pf-c-chip-group__main` is always present in all variations. If #4873 does that, then we can close #4821

PR #4906, opened 2020-09-29
fix(react-table): use direct imports of react-core components

Hyperkid123:
Recently we have noticed that the about modal background images are being included in insights application builds, even though these applications do not use the about modal component and using direct imports from react-core. Recently there were some changes in the react-table package which are not using the direct imports, causing these images to be included. Here is a trace from webpack:

```bash

 @ ./node_modules/@patternfly/react-styles/css/components/AboutModalBox/about-modal-box.css
 @ ./node_modules/@patternfly/react-styles/css/components/AboutModalBox/about-modal-box.js
 @ ./node_modules/@patternfly/react-core/dist/esm/components/AboutModal/AboutModalBoxContent.js
 @ ./node_modules/@patternfly/react-core/dist/esm/components/AboutModal/AboutModalContainer.js
 @ ./node_modules/@patternfly/react-core/dist/esm/components/AboutModal/AboutModal.js
 @ ./node_modules/@patternfly/react-core/dist/esm/components/AboutModal/index.js
 @ ./node_modules/@patternfly/react-core/dist/esm/components/index.js
 @ ./node_modules/@patternfly/react-core/dist/esm/index.js
 @ ./node_modules/@patternfly/react-table/dist/esm/components/Table/Table.js
 @ ./node_modules/@patternfly/react-table/dist/esm/components/Table/index.js
 @ ./node_modules/@patternfly/react-table/dist/esm/components/index.js
 @ ./node_modules/@patternfly/react-table/dist/esm/index.js
 @ ./node_modules/@redhat-cloud-services/frontend-components/components/cjs/TagModal.js
 @ ./src/js/App/GlobalFilter/TagsModal.js
 @ ./src/js/App/GlobalFilter/GlobalFilter.js
 @ ./src/js/App/GlobalFilter/index.js
 @ ./src/js/App/RootApp.js
 @ ./src/js/chrome/entry.js
 @ ./src/js/chrome.js
```

After using the direct import of commonjs version of react-core, the issue is gone.

@evwilkin @redallen I would also like to again suggest using babel transform imports plugin to dynamically transform any patternfly import path to direct path from js or esm or umd build based on the current build environment. Currently, there is only one properly optimized build and that is commonjs because in multiple components we directly import components from dist/js/... and that causes duplicate asset imports in our applications. Using the babel import will not only fix duplicate assets issues but will also prevent future shorthand imports in all your react packages. That could save us a lot of debugging when our applications bloat for no apparent reason.

cc @karelhala @ryelo

### DIFF
--- a/packages/react-table/src/components/Table/BodyCell.tsx
+++ b/packages/react-table/src/components/Table/BodyCell.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import { Tooltip } from '@patternfly/react-core/dist/js/components/Tooltip/Tooltip';
 import styles from '@patternfly/react-styles/css/components/Table/table';
-import { SelectProps } from '@patternfly/react-core';
+import { SelectProps } from '@patternfly/react-core/dist/js/components/Select';
 
 export interface BodyCellProps {
   'data-label'?: string;
--- a/packages/react-table/src/components/Table/EditableSelectInputCell.tsx
+++ b/packages/react-table/src/components/Table/EditableSelectInputCell.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
-import { Select, SelectOptionObject, SelectProps } from '@patternfly/react-core';
+import { Select, SelectOptionObject, SelectProps } from '@patternfly/react-core/dist/js/components/Select';
 import inlineStyles from '@patternfly/react-styles/css/components/InlineEdit/inline-edit';
 import formStyles from '@patternfly/react-styles/css/components/Form/form';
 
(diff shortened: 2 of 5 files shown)


redallen:
When running `webpack --mode production` in the seed app with Table imports I'm unable to reproduce this bug:
image

If you could provide steps to reproduce that'd be great. When using ES imports in ESM files tree shaking should be enabled.

We don't use `babel` to transpile our TS, we use `tsc`, but you could write an ESLint plugin similar to [import-default-tokens](https://github.com/patternfly/patternfly-react/blob/master/packages/eslint-plugin-patternfly-react/lib/rules/import-tokens-icons.js). This patch is just going to get broken by our devs again who use shorthand ES imports.

Hyperkid123:
@redallen its caused by a third-party library using and exporting the table. We had the same issues with icons a while back if you can remember.

redallen:
Ah, what if you import `@redhat-cloud-services/frontend-components/components/esm/TagModal.js` instead? It seems whatever bundler you're using parses ES imports just fine and could tree shake them then.

Hyperkid123:
@redallen yeah we have done that that stack is outdated sorry. The images imports disappeared only after I've applied and build changes from this PR locally and used that bundle instead.

Hyperkid123:
@redallen I assume I can close this in favor of #4910?

Hyperkid123:
@redallen I see. I will continue with it next week. Thanks for the info.

PR #4907, opened 2020-09-29
Port react-console from PF3 branch - move to typescript and utilize PF4 components

### LINKED ISSUE #4836: Need support for react-console from PF3 in PF4
Patternfly react 3 provided a console package which provided VncConsole, SerialConsole and DesktopViewer React components to access virtual machine or server consoles. These components are needed for projects upgrading to Patternfly 4.

Feature Request

Openshift 4.7

KKoukiou:
Closes #4836


KKoukiou:
@redallen I am a bit stuck with not being able to build the docs here :( 

It fails with (node:2800) UnhandledPromiseRejectionWarning: ReferenceError: window is not defined
    at init_logging (/home/circleci/project/packages/react-docs/.cache/ssr-build/components/accessconsoles/react/index.js:43:5)

I understand that window does not exist, unless we have a browser, but I don't know how to workaround it. :/ Any idea?

KKoukiou:
> @KKoukiou You need to check `typeof window !== 'undefined'` before you use `window` so that your component can be server side rendered (SSR).

The problem is that the place where is fails is not in my code, but in an imported module's code.

KKoukiou:
> > @KKoukiou You need to check `typeof window !== 'undefined'` before you use `window` so that your component can be server side rendered (SSR).
> 
> The problem is that the place where is fails is not in my code, but in an imported module's code.

And even with a conditional import where I include the problematic module just if window is defined. It fails the same way :/

KKoukiou:
@redallen I need some help here with the docs if possible :) I have two main issues:

* the css does not seem to be included in the docs, for the demo app it works fine
* the properties of the new components do not appear in the docs either

redallen (changes requested):
Hey @KKoukiou , this looks awesome so far!

- To fix your CSS problem you need to import each piece of CSS in the components it belongs to. I'm not sure which components need `xterm/css/xterm.css` but you can import it in each component that needs it.
- To make the properties appear you need to call `sourceProps` to parse your files with `react-docgen`.

I'm not seeing an actual console with text -- just a dropdown to download a file. I believe the old react-console did this using vnc, no?
image

KKoukiou:
> Hey @KKoukiou , this looks awesome so far!
> 
>     * To fix your CSS problem you need to import each piece of CSS in the components it belongs to. I'm not sure which components need `xterm/css/xterm.css` but you can import it in each component that needs it.
> 
>     * To make the properties appear you need to call `sourceProps` to parse your files with `react-docgen`.
> 
> 
> I'm not seeing an actual console with text -- just a dropdown to download a file. I believe the old react-console did this using vnc, no?

Right, there are three types of supported consoles. VNC and XTerm are only in the demo app now, not in the actual docs because their code access window properties and makes the docs compilation fail. 

> image

packages/react-console/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "copy:sass": "shx mkdir -p dist/sass && shx cp -r sass/react-console/* dist/sass",
+    "clean": "rimraf dist"
+  },
+  "dependencies": {
+    "@patternfly/patternfly": "4.50.4",
+    "@patternfly/react-core": "^4.63.2",
+    "@spice-project/spice-html5": "^0.2.1",
+    "@types/file-saver": "^2.0.1",
+    "@types/novnc-core": "^0.1.2",
+    "blob-polyfill": "^3.0.20180112",
```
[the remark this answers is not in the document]
KKoukiou: Blob.js implements the W3C Blob interface in browsers that do not natively support it.

I checked at the list and it's really old versions of Firefox and Chrome, I guess we should not really care https://github.com/eligrey/FileSaver.js#supported-browsers

packages/react-console/sass/react-console.scss
```diff
@@ -0,0 +1,5 @@
+@import 'xterm/css/xterm.css';
+@import './react-console/AccessConsoles.scss';
+@import './react-console/DesktopViewer.scss';
+@import './react-console/SerialConsole.scss';
+@import './react-console/VncConsole.scss';
```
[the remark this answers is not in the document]
KKoukiou: I converted them to CSS.
However the copy step is still needed. We somehow need to copy the css file to the dist directory. Therefore it's easier to keep them all together and copy the css to the dist from a build script.

Please check the current code and let me know if that's fine for you.

packages/react-integration/demo-app-ts/src/components/demos/ConsolesDemo/ConsolesDemo.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { debounce } from '@patternfly/react-core';
+
+export const ConsolesDemo: React.FC = () => {
+  const [status, setStatus] = React.useState('disconnected');
+  const setConnected = React.useRef(debounce(() => setStatus('connected'), 3000)).current;
+  const ref = React.createRef<SerialConsole>();
+
+  return (
+    <div className="consoles-demo-area">
+      <AccessConsoles preselectedType="SerialConsole">
```
redallen, line 13: What error is thrown when you add this example to our main docs? I think it used to prerender fine, maybe it's fixable?

packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx
```diff
@@ -0,0 +1,140 @@
+import React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+import constants from '../common/constants';
+
+import '@patternfly/react-console/css/AccessConsoles.css';
```
redallen, line 6: Now that these are CSS, can you move them to `react-styles/src/css/components`? That way you can get types and remove the copy step.

redallen (changes requested):
This is really close!

packages/react-integration/demo-app-ts/src/components/demos/ConsolesDemo/ConsolesDemo.tsx
KKoukiou: Something like https://github.com/patternfly/patternfly-react/pull/4907#issuecomment-701343682

When did it use to prerender fine? In PF3 documentation? I don't think it was even included there.

redallen: Ah, you're right it was never on our Gatsby site https://patternfly-react-v3.surge.sh which prerendered but instead only on the Storybook at https://patternfly-react-pf3.surge.sh .

I'm looking now and the worst case is that we can disable prerendering for that page.

KKoukiou: Thanks for looking into this.

redallen: I've filed an upstream PR for `xterm`: https://github.com/xtermjs/xterm.js/pull/3134. `novnc-core` seems to only add event handling, so for SSR I'll use `null-loader` on it in our theme (https://github.com/patternfly/patternfly-org/pull/2210). With these two changes it builds!

If upstream won't accept the PR, I'll have no choice but to disable SSR for this page which means it won't be indexed by search.

redallen (changes requested):
I pushed some changes to fix the build. I say we give the upstream PR a week to be accepted before resorting to publishing our own package or disabling SSR. Other than that, I'm not sure what section of the docs this extension should be under.

packages/react-console/src/components/VncConsole/VncConsole.tsx
```diff
@@ -84,7 +84,7 @@ export class VncConsole extends React.Component<VncConsoleProps, VncConsoleState
     textDisconnected: 'Disconnected',
     textDisconnect: 'Disconnect'
   };
-  private rfb: RFB;
```
KKoukiou: ah thanks :) I was not sure how to fix this, and ts-ignore stopped working since last week!

scripts/verifyPatternflyVersions.js
```diff
@@ -47,7 +47,7 @@ async function verifyPatternflyVersions() {
   if (process.argv[2] === '--fix') {
     mismatchedVersions.forEach(([dep, versions]) => {
       const highestVersion = Object.keys(versions)
-        .sort(semver.compare)
+        .sort((a, b) => semver.compare(a.replace(/^\^/, ''), b.replace(/^\^/, '')))
```
KKoukiou, line 50: Is this change unrelated?

redallen: I added it so I could run `node scripts/verifyPatternflyVersions.js --fix` to update react-console's out-of-date `@patternfly/react-core`.

KKoukiou: Super thanks

packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx
```diff
@@ -0,0 +1,140 @@
...
+  /** Initial selection of the Select */
+  preselectedType?: string; // NONE_TYPE | SERIAL_CONSOLE_TYPE | VNC_CONSOLE_TYPE | DESKTOP_VIEWER_CONSOLE_TYPE;
+}
+
+export interface AccessConsolesState {
+  type: string;
+  isOpen: boolean;
+}
+
+export class AccessConsoles extends React.Component<AccessConsolesProps, AccessConsolesState> {
```
jeff-phillips-18, line 44: Would be good to convert this to a FunctionComponent

packages/react-console/src/components/DesktopViewer/ConnectWithRemoteViewer.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const onClickVV = () => {
+    const type = spice ? SPICE_CONSOLE_TYPE : VNC_CONSOLE_TYPE;
+    if (_console) {
+      const vv = onGenerate(_console, type);
+      return onDownload(DEFAULT_VV_FILENAME, vv.content, vv.mimeType || DEFAULT_VV_MIMETYPE);
+    }
+  };
+
+  const onClickRDP = () => {
+    if (rdp) {
```
jeff-phillips-18, line 93: the button is only rendered if `rdp` so no need to check here

packages/react-console/src/components/SerialConsole/SerialConsole.tsx
```diff
@@ -0,0 +1,147 @@
...
+  textDisconnect?: string;
+  /** Text content rendered inside the Reset button */
+  textReset?: string;
+  /* Text content rendered inside the EmptyState for when console is disconnnected */
+  textDisconnected?: string;
+  /* Text content rendered inside the EmptyState for when console is loading */
+  textLoading?: string;
+}
+
+export class SerialConsole extends React.Component<SerialConsoleProps> {
```
jeff-phillips-18, line 33: Would be good to convert to a FunctionComponent

packages/react-console/src/components/SerialConsole/XTerm.tsx
```diff
@@ -0,0 +1,127 @@
...
+  fontSize?: number;
+  /** Terminal title has been changed. */
+  onTitleChanged?: (title: string) => void;
+  /** Data to be sent from terminal to backend; (data) => {} */
+  onData?: (e: string) => void;
+  /** Terminal has been resized; backend shall be informed. (rows; cols) => {} */
+  onResize?: (rows: number, cols: number) => void;
+}
+
+export class XTerm extends React.Component<XTermProps> {
```
jeff-phillips-18, line 23: FunctionComponent

packages/react-console/src/components/SpiceConsole/SpiceConsole.tsx
```diff
@@ -0,0 +1,150 @@
...
+  textConnecting?: string;
+  textDisconnected?: string;
+  textSendShortcut?: string;
+  textCtrlAltDel?: string;
+}
+
+export interface SpiceConsoleState {
+  status: string;
+}
+export class SpiceConsole extends React.Component<SpiceConsoleProps, SpiceConsoleState> {
```
jeff-phillips-18, line 40: FunctionComponent

packages/react-console/src/components/VncConsole/VncConsole.tsx
```diff
@@ -0,0 +1,240 @@
...
+  textSendShortcut?: string;
+  /** Text content rendered inside the Ctrl-Alt-Delete dropdown entry */
+  textCtrlAltDel?: string;
+}
+
+export interface VncConsoleState {
+  status: string;
+}
+
+export class VncConsole extends React.Component<VncConsoleProps, VncConsoleState> {
```
jeff-phillips-18, line 69: FunctionComponent

jeff-phillips-18 (commented):
Looks good. Preference towards FunctionComponent, I'd be ok with not converting at this time, but it would be nice to get it done on introduction rather than converting later.

packages/react-console/src/components/SerialConsole/XTerm.tsx
KKoukiou: So this and the SerialConsole have two public method which should be accessed by the caller using a ref. In functional components refs dont work simply like that. I will need to use a forwardRef I think.
I will try to finish this today according to your comments.

jeff-phillips-18 (dismissed):
👍

KKoukiou:
> +1

Thanks for approving. I just see that the a11y tests fail but I can't see on what? Any idea?

redallen:
@KKoukiou You can check the report [here:](https://patternfly-react-pr-4907-a11y.surge.sh/):
image

packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx
```diff
@@ -100,6 +100,7 @@ export const AccessConsoles: React.FunctionComponent<AccessConsolesProps> = ({
       {React.Children.toArray(children).length > 1 && (
         <div className="pf-c-console__actions">
           <Select
+            aria-label="Select Console Input"
```
jeff-phillips-18, line 103: `aria-label={textSelectConsoleType}`

KKoukiou: Ah indeed, I am not even sure if that's what broke the a11y tests though, just a guess.

KKoukiou: It was not that

KKoukiou:
> @KKoukiou You can check the report [here:](https://patternfly-react-pr-4907-a11y.surge.sh/):
> image

https://github.com/xtermjs/xterm.js/pull/3172

Are we going to block on this? Or shall we just whitelist it somehow?

KKoukiou:
@jeff-phillips-18 I could you please re-review - I had some issues with ayy1 because of xterm and needed to repost.

jeff-phillips-18 (dismissed):

packages/react-console/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "clean": "rimraf dist"
+  },
+  "dependencies": {
+    "@patternfly/patternfly": "4.59.1",
+    "@patternfly/react-core": "^4.75.9",
+    "@spice-project/spice-html5": "^0.2.1",
+    "@types/file-saver": "^2.0.1",
+    "@types/novnc-core": "^0.1.2",
+    "file-saver": "^1.3.8",
+    "lodash": "^4.17.19",
```
redallen, line 41: Looks like react-console doesn't actually use lodash, could you remove this dependency?

packages/react-console/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+Note the build scripts for this are located in the root package.json under `yarn build`.
+
+### Testing
+
+Testing is done at the root of this repo. To only run the @patternfly/react-charts tests:
+
+```
+yarn test packages/react-charts
+```
```
redallen, line 98: Can you remove this section? Most of it is out of date.

packages/react-console/src/components/VncConsole/VncConsole.tsx
```diff
@@ -0,0 +1,213 @@
...
+import { EmptyState, EmptyStateBody, EmptyStateIcon, Spinner } from '@patternfly/react-core';
+
+import * as NovncLog from 'novnc-core/lib/util/logging';
+/** Has bad types. https://github.com/larryprice/novnc-core/issues/5 */
+import RFB from 'novnc-core';
+
+import { VncActions } from './VncActions';
+import constants from '../common/constants';
+
+import '@patternfly/react-styles/css/components/Consoles/VncConsole.css';
```
redallen, line 12: You can write `import styles from '@patternfly/react-styles...';` and then later use `styles.consoleVnc` rather than the string `pf-c-console__vnc` if you'd like.

redallen (changes requested):
Few final nits, looking good!

redallen (approved):
LGTM!

jeff-phillips-18 (approved):

PR #4913, opened 2020-09-30
Fix(datalist): update aria of datalist

### LINKED ISSUE #4896: Datalist: update aria and a11y considerations
We should take a deeper look at the datalist component and determine what the right aria for each scenario would be. Roles were added to communicate the selectable datalist, but they are currently missing `aria-activedescendant` when an option is visually focused and `aria-selected/aria-multiselectable` when an item is selected as this [WAI example](https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/listbox/listbox.html) shows. However, I question whether this should be the case for all datalist examples.

Open up VO (cmd + F5), go to datalist example and see that it considers all datalist examples as selectable now which isn't accurate. It also doesn't communicate to users when they have focus on an option or when they've selected an item on the selectable datalist (due to the missing aria).

a11y bug

jessiehuff:
**What**: Closes #4896

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -54,7 +54,7 @@ export const DataList: React.FunctionComponent<DataListProps> = ({
       <ul
         className={css(styles.dataList, isCompact && styles.modifiers.compact, className)}
         aria-label={ariaLabel}
-        role="listbox"
+        role={isSelectable? "listbox" : null}
         {...props}
       >
         {children}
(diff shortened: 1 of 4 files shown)


redallen (changes requested):
Not sure this does what you're expecting. Since `Boolean({ 'aria-selected': true }) === true` this will only add `{  'aria-activedescendant': selectedDataListItemId }` and will ignore the `aria-selected`. Did you mean to add both?
```jsx
{...(isSelectable && selectedDataListItemId === id && {
  'aria-selected': true,
  'aria-activedescendant': selectedDataListItemId
})}
```

kmcfaul (dismissed):

KKoukiou (approved):
Reading your comment and the listbox [role docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) I think your decision makes sense.

redallen (approved):

PR #4915, opened 2020-09-30
For Select chips, return the string value.

### LINKED ISSUE #4914: Select throws error when using a unknown SelectOptionObject

When using a `Select` with `SelectOptionObject`, if the option is unknown (i.e. The option is not in the default list of children of the select) the `Select` throw an error when building the chips.

It works fine if instead of `SelectOptionObject` we use plain strings.

https://codesandbox.io/s/wandering-leftpad-gszxg

Results in:
```
Objects are not valid as a React child (found: object with keys {val}). If you meant to render a collection of children, use an array instead.
```

If you change line 36 from `foobar` (unknown) to `Alabama` it works well.

Bug

Notifications

josejulio:
**What**: Closes #4914

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -517,7 +517,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       }
       return item.props.value.toString();
     }
-    return value;
+    return value.toString();
   };
 
   findText = (item: React.ReactNode) => {


redallen (approved):
Good catch.

cc @kmcfaul

nicolethoen (approved):

kmcfaul (approved):

jschuler (approved):

Issue #3095, opened 2019-10-09
[PF4] Changing selections value outside Select does not update correctly in Select component


SpyTec:
This is a duplicate of #2828. Issue was not resolved properly in my PR #2848 and still exists.

PR #4916, opened 2020-09-30
fix(select): display label and not value in typeahead

### LINKED ISSUE #4448: Select typeahead shows selected value instead of text on some renders
Hi, I'm experiencing an issue: with a `Select` of the `typeahead` (single) variant, the first render after making a selection displays the item's `value`, not its textContent, as the "typeaheadValue" inside the input. However, the *second* render replaces the value with the proper text.

From the user's perspective, it looks like this:
1. Click on the select's arrow to see the available options
2. Click on one of the options to select it → a different text (the **value**) appears as the selection in the input at the top
3. Click outside the select to close it → the proper text (the textContent between `<SelectOption>` and `</SelectOption>`) appears as the selection - this is what the user expects from the beginning
4. Alternatively, clicking the same option twice has the same result - on the second render, the option's text, not value, is rendered

I think it has something to do with this fix: https://github.com/patternfly/patternfly-react/pull/3096  
The linked change appears to "fix" setting the `selections` from outside by using the `selections` value as the type-ahead text. However, this seems to assume that the `value` and the displayed option text are always the same, which looks like an unreasonable assumption to make.

Here's a minimal example that shows the weird behavior: https://codesandbox.io/s/unruffled-chandrasekhar-idl9j?file=/src/App.tsx

This is a bug that has no work-around that I know of, as having a different option `value` and option text is absolutely necessary (values are often not human-readable).

Is there any other way to make this work? I've tried not setting `selections` and letting the Select component manage this internally, but this results in the input not updating at all (i.e. no user feedback as to which option is selected).

Our product is an admin-type GUI and we're on `react-core` 3.153.3, though this is reproducible on newer versions also (as seen in the sample code linked above).

boaz0:
**What**: Closes #4448

~~**DO NOT MERGE**~~ READY FOR MERGE
- [x] Add test verifying that this fixes #4448 
- [x] Add test verifying that it does not break #3095

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -202,12 +202,6 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       });
     }
 
-    if (prevProps.selections !== this.props.selections && this.props.variant === SelectVariant.typeahead) {
-      this.setState({
-        typeaheadInputValue: this.props.selections as string
-      });
-    }
-
     if (
       this.props.onFavorite &&
       (this.props.favorites.length !== prevProps.favorites.length ||
--- a/packages/react-integration/cypress/integration/select.spec.ts
+++ b/packages/react-integration/cypress/integration/select.spec.ts
@@ -19,6 +19,23 @@ describe('Select Test', () => {
       .should('exist');
   });
 
+  it('Verify Typeahead Select with labels', () => {
+    const find = (selector: string) =>
+      cy
+        .get('#typeahead-select-labels-id')
+        .parent()
+        .find(selector);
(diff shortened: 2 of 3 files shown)


boaz0:
//cc @knokit @rkaw92

dlabrecq (dismissed):

kmcfaul (dismissed):

boaz0:
Rebased PR

redallen (approved):
LGTM! Thanks for the test.

jschuler (approved):

ddonahue007 (approved):
LGTM

boaz0:
Thank you so much! :bow:

ddonahue007:
Your welcome :)

PR #4919, opened 2020-10-01
fix(Select): grouped Select generates invalid HTML

### LINKED ISSUE #4917: Invalid HTML in grouped menus
In grouped menus, the list items need to be wrapped in a `<ul>`. Noticed this in the grouped single and grouped typeahead examples.

Screen Shot 2020-09-30 at 2 54 05 PM

KKoukiou:
Fixes #4917

### DIFF
--- a/packages/react-core/src/components/Select/SelectGroup.tsx
+++ b/packages/react-core/src/components/Select/SelectGroup.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/Select/select';
 import { css } from '@patternfly/react-styles';
 
+import { SelectConsumer, SelectVariant } from './selectConstants';
+
 export interface SelectGroupProps extends React.HTMLProps<HTMLDivElement> {
   /** Checkboxes within group. Must be React.ReactElement<SelectOptionProps>[] */
   children?: React.ReactNode;
@@ -20,11 +22,15 @@ export const SelectGroup: React.FunctionComponent<SelectGroupProps> = ({
   titleId = '',
   ...props
 }: SelectGroupProps) => (
-  <div {...props} className={css(styles.selectMenuGroup, className)}>
-    <div className={css(styles.selectMenuGroupTitle)} id={titleId} aria-hidden>
-      {label}
-    </div>
-    {children}
-  </div>
+  <SelectConsumer>
+    {({ variant }) => (
+      <div {...props} className={css(styles.selectMenuGroup, className)}>
+        <div className={css(styles.selectMenuGroupTitle)} id={titleId} aria-hidden>
+          {label}
+        </div>
+        {variant === SelectVariant.checkbox ? children : <ul role="listbox">{children}</ul>}
+      </div>
+    )}
+  </SelectConsumer>
 );
 SelectGroup.displayName = 'SelectGroup';
--- a/packages/react-core/src/components/Select/SelectMenu.tsx
+++ b/packages/react-core/src/components/Select/SelectMenu.tsx
@@ -203,19 +203,30 @@ class SelectMenuWithRef extends React.Component<SelectMenuProps> {
                 {children}
               </div>
             )}
-            {variant !== SelectVariant.checkbox && !isCustomContent && (
-              <ul
-                ref={innerRef}
-                className={css(styles.selectMenu, className)}
-                role="listbox"
-                aria-label={ariaLabel}
-                aria-labelledby={(!ariaLabel && ariaLabelledBy) || null}
-                {...(maxHeight && { style: { maxHeight, overflow: 'auto' } })}
-                {...props}
-              >
-                {this.extendChildren(inputIdPrefix)}
-              </ul>
-            )}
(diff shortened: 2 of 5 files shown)


nicolethoen (commented):
The dividers are not wrapped in an `li` tag.
I'm a little afraid to comment that, and i don't know if it's invalid HTML or not without looking it up.
If we need to wrap the dividers, we need to make sure that the keyboard interactions still work in the typeahead variants.

redallen (approved):

packages/react-core/src/components/Select/SelectGroup.tsx
```diff
@@ -20,11 +20,11 @@ export const SelectGroup: React.FunctionComponent<SelectGroupProps> = ({
...
 }: SelectGroupProps) => (
-  <div {...props} className={css(styles.selectMenuGroup, className)}>
-    <div className={css(styles.selectMenuGroupTitle)} id={titleId} aria-hidden>
-      {label}
-    </div>
-    {children}
-  </div>
+  <li {...props} className={css(styles.selectMenuGroupTitle, className)} id={titleId}>
+    {label}
+    <ul className={css(styles.selectMenuGroup)} role="listbox">
```
jschuler, line 25: I think you should only change that `children` is wrapped by `<ul className={css(styles.selectMenuGroup)} role="listbox">`, and leave the outer elements the way they were

jschuler (changes requested):

packages/react-core/src/components/Select/SelectGroup.tsx
KKoukiou: That's what the issue description suggests to do, but if you read the specification of what can be child of UL is not contain divs.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

SelectGroups is direct child of UI, and should not contain any divs in other words.

KKoukiou:
> The dividers are not wrapped in an `li` tag.
> I'm a little afraid to comment that, and i don't know if it's invalid HTML or not without looking it up.
> If we need to wrap the dividers, we need to make sure that the keyboard interactions still work in the typeahead variants.

Good catch, it's essentially the same problem, do you want me to fix it within this PR, on send another one?
(this fix is a standalone so I suggest to not block it on fixing more things)

packages/react-core/src/components/Select/SelectGroup.tsx
jschuler: The DOM should look similar to this (copied from Core example):
Screen Shot 2020-10-01 at 1 43 16 PM

jschuler: Also the changes break the styling a bit, additional indentation etc
https://patternfly-react-pr-4919.surge.sh/components/select#grouped-checkbox-input

KKoukiou: Ack I see it now in the HTML docs. I will adjust it tomorrow.
Thanks for the review.

KKoukiou: I wonder if it's a concious decision though that nested ul li were not used in the HTML docs.

For me it makes sense that the outer items would be also a list (ul) since there indeed part of a list.

@mcarrano I guess you can answer this one :)  For example this https://patternfly-react-pr-4919.surge.sh/components/select/#grouped-single

packages/react-core/src/components/Select/SelectGroup.tsx
```diff
@@ -20,11 +20,11 @@ export const SelectGroup: React.FunctionComponent<SelectGroupProps> = ({
   titleId = '',
   ...props
 }: SelectGroupProps) => (
-  <div {...props} className={css(styles.selectMenuGroup, className)}>
```
mcoker, line 23: This was fine as it is. The structure should be

```
div.selectMenuGroup
    div.selectMenuGroupTitle
    ul
        li
        li
        ....
```

You can see the basic structure in the favorites example, except the `li`'s in the favorites are `li.pf-c-select__menu-wrapper` and they should be `li.pf-c-select__menu-item` in non-favorites lists.

packages/react-core/src/components/Select/SelectGroup.tsx
mcarrano: @mcoker can you answer the question above?

mcoker: @jschuler is right with the screenshot posted above. I left the same basic comment here - https://github.com/patternfly/patternfly-react/pull/4919#discussion_r498446371

FWIW you can find the same basic structure in the dropdown, or any menu with groups/titles https://www.patternfly.org/v4/components/dropdown/react/with-groups/

```
div/section.section
    div/h1-h6.section-title
    ul
        li
        li
        ....
```

jschuler (approved):
LGTM

KKoukiou:
> LGTM

I needed to rebase - had merge conflicts. Sorry :(

mcoker (approved):
lgtm! 🤩

redallen (approved):

PR #4925, opened 2020-10-01
    feat(table): added support for single row radio select

### LINKED ISSUE #4882: Add radio button select in Table
Introduce a new Table variant that supports radio button select rather than checkbox select to select only one row at a time. Core issue https://github.com/patternfly/patternfly/issues/3418

KKoukiou:
Closes #4882

### DIFF
--- a/packages/react-table/src/components/Table/Table.tsx
+++ b/packages/react-table/src/components/Table/Table.tsx
@@ -247,7 +247,7 @@ export interface IRow extends RowType {
   selected?: boolean;
   /** deprecated - Use disableSelection instead - Whether the row checkbox is disabled */
   disableCheckbox?: boolean;
-  /** Where the row checkbox/radio button is disabled */
+  /** Whether the row checkbox/radio button is disabled */
   disableSelection?: boolean;
 }
 
--- a/packages/react-table/src/components/Table/Table.tsx
+++ b/packages/react-table/src/components/Table/Table.tsx
@@ -33,6 +33,11 @@ export enum TableVariant {
 
 export type RowEditType = 'save' | 'cancel' | 'edit';
 
+export enum RowSelectVariant {
+  radio = 'radio',
+  checkbox = 'checkbox'
+}
+
 export interface RowErrors {
   [name: string]: string[];
 }
@@ -97,6 +102,7 @@ export interface IColumn {
     onCollapse?: OnCollapse;
     onExpand?: OnExpand;
     onSelect?: OnSelect;
+    selectVariant?: 'checkbox' | 'radio';
     onRowEdit?: OnRowEdit;
     rowLabeledBy?: string;
     expandId?: string;
@@ -237,10 +243,12 @@ export interface IRow extends RowType {
   isExpanded?: boolean;
   isFirstVisible?: boolean;
   isLastVisible?: boolean;
-  /* Whether the row checkbox is selected */
+  /** Whether the row checkbox/radio button is selected */
   selected?: boolean;
-  /* Whether the row checkbox is disabled */
+  /** deprecated - Use disableSelection instead - Whether the row checkbox is disabled */
   disableCheckbox?: boolean;
+  /** Where the row checkbox/radio button is disabled */
+  disableSelection?: boolean;
 }
 
 export interface TableProps extends OUIAProps {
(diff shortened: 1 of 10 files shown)


packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -487,6 +487,84 @@ class SelectableTable extends React.Component {
 }
 ```
 
+### Selectable Radio Input
```
redallen, line 490: Use sentence case: `Selectable radio input`

redallen (changes requested):
Probably don't want a radio button in the table header?
image

mcoker:
Aside from @redallen's comments, this LGTM. One thing I noticed is that the react table uses `pf-c-check__input` and `pf-c-radio__input` for the check and radio inputs, but we don't use those in core. They're just regular `<input type="[radio/check]">` elements. The check and radio components are designed to work as an input/label pair, and the `pf-c-[check/radio]__input` elements are expected to be children of `pf-c-[check/radio]`.

I don't think that we want to change the checkbox inputs since that could break users if they're styling them using the `pf-c-check__input` class or something like that, but I'm wondering if we should use regular radio inputs, or use the `pf-c-radio__input` element to be consistent with the table checkboxes.

KKoukiou:
> Aside from @redallen's comments, this LGTM. One thing I noticed is that the react table uses `pf-c-check__input` and `pf-c-radio__input` for the check and radio inputs, but we don't use those in core. They're just regular `<input type="[radio/check]">` elements. The check and radio components are designed to work as an input/label pair, and the `pf-c-[check/radio]__input` elements are expected to be children of `pf-c-[check/radio]`.
> 
> I don't think that we want to change the checkbox inputs since that could break users if they're styling them using the `pf-c-check__input` class or something like that, but I'm wondering if we should use regular radio inputs, or use the `pf-c-radio__input` element to be consistent with the table checkboxes.

@mcarrano amazing how you noticed this! Indeed I introduced these classes in a separate commit in this PR, and the reason I did it, it because I noticed that checboxes/radio inputs without these two classes (`pf-c-radio__input, pf-c-check__input`) do not show the black circle when being hovered over in a disable mode.

This is a bug and adding these classes fixes the issue.

So if I was to choose between fixing an existing bug by using the classes, and protecting users who are using custom CSS on these classes and might see issues, I would go with the first.
In any case, I leave it up to you to decide :)

packages/react-table/src/components/Table/Table.tsx
```diff
@@ -237,10 +243,12 @@ export interface IRow extends RowType {
   isExpanded?: boolean;
   isFirstVisible?: boolean;
   isLastVisible?: boolean;
-  /* Whether the row checkbox is selected */
+  /** Whether the row checkbox/radio button is selected */
   selected?: boolean;
-  /* Whether the row checkbox is disabled */
+  /** deprecated - Use disableSelection instead - Whether the row checkbox is disabled */
   disableCheckbox?: boolean;
+  /** Where the row checkbox/radio button is disabled */
```
jschuler, line 250:
```suggestion
  /** Whether the row checkbox/radio button is disabled */
```

redallen (approved):
Code LGTM!

mcoker:
@KKoukiou ah! Thanks for the explanation and fixing the bug. It's ultimately a misuse of the radio/check component, and will also change the positioning of the existing check input because the check component's input has a margin - https://github.com/patternfly/patternfly/blob/1b3c31608b8d7f43f3509cdeef8b7e62358b8a46/src/patternfly/components/Check/check.scss#L26-L28

I'm in favor of leaving it with the disabled input bug for the time being. We have an issue to allow the check/radio inputs to be used on their own in the check/radio components here - https://github.com/patternfly/patternfly/issues/3311

I think once we have support for  https://github.com/patternfly/patternfly/issues/3311 we can look at whether we should introduce the check/radio component input here, or just style the regular check/radio inputs in the table to support a disabled state until we're able to update the table to use the check/radio component inputs. Sound ok?

jschuler (commented):
Code LGTM, just need to clarify on the radio/checkbox styling with @mcoker

KKoukiou:
> @KKoukiou ah! Thanks for the explanation and fixing the bug. It's ultimately a misuse of the radio/check component, and will also change the positioning of the existing check input because the check component's input has a margin - https://github.com/patternfly/patternfly/blob/1b3c31608b8d7f43f3509cdeef8b7e62358b8a46/src/patternfly/components/Check/check.scss#L26-L28
> 
> I'm in favor of leaving it with the disabled input bug for the time being. We have an issue to allow the check/radio inputs to be used on their own in the check/radio components here - [patternfly/patternfly#3311](https://github.com/patternfly/patternfly/issues/3311)
> 
> I think once we have support for [patternfly/patternfly#3311](https://github.com/patternfly/patternfly/issues/3311) we can look at whether we should introduce the check/radio component input here, or just style the regular check/radio inputs in the table to support a disabled state until we're able to update the table to use the check/radio component inputs. Sound ok?

Perfect! Reverting the relevant commit.

mcoker (approved):
lgtm! Thanks @KKoukiou!!

mcoker (approved):

redallen (approved):

PR #4931, opened 2020-10-02
Select: Pass aria-label / aria-labelledby to the <ul role="listbox"> to prevent axe issues

mturley:
I noticed when using `<Select>` in the default variant (`variant !== SelectVariant.checkbox && !isCustomContent`), when the SelectMenu is opened, the `<ul role="listbox">` that is rendered does not have the `aria-label` or `aria-labelledby` props that I pass into Select. Other variants of SelectMenu do pass these props down, so I found this odd.

Arbitrary props can be passed to this element via the `{...props}` spread syntax, but these specific props are not included because they are pulled out of that `props` object in the destructure assignment at the top of SelectMenu's render method.

In our app downstream, we are seeing a react-axe error caused by this:

Screen Shot 2020-10-02 at 6 11 11 AM

I don't see another way to solve that error with the props currently available.

### DIFF
--- a/packages/react-core/src/components/Select/SelectMenu.tsx
+++ b/packages/react-core/src/components/Select/SelectMenu.tsx
@@ -208,6 +208,8 @@ class SelectMenuWithRef extends React.Component<SelectMenuProps> {
                 ref={innerRef}
                 className={css(styles.selectMenu, className)}
                 role="listbox"
+                aria-label={ariaLabel}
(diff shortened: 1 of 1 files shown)


redallen (approved):

kmcfaul (approved):

mturley:
Sweet, thanks guys!

PR #4935, opened 2020-10-02
fix(react-icons): use lying cast

redallen:
**What**: Fixes @jcaianirh 's problem in OpenShift which uses `<CheckIcon alt="" />`. This becomes `<svg alt="">` which is invalid ([should be `<svg aria-label="">`](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html)), but we'll continue lying that it's okay.

**Additional issues**:

### DIFF
--- a/packages/react-icons/src/createIcon.tsx
+++ b/packages/react-icons/src/createIcon.tsx
@@ -31,7 +31,7 @@ export interface IconDefinition {
   yOffset?: number;
 }
 
-export interface SVGIconProps extends Omit<React.SVGProps<SVGElement>, 'size' | 'ref'> {
+export interface SVGIconProps extends Omit<React.HTMLProps<SVGElement>, 'size' | 'ref'> {
   color?: string;
   size?: IconSize | keyof typeof IconSize;
   title?: string;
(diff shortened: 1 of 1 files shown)


jcaianirh (approved):
thanks @redallen

kmcfaul (approved):

spadgett:
> **What**: Fixes @jcaianirh 's problem in OpenShift which uses `<CheckIcon alt="" />`. This becomes `<svg alt="">` which is invalid ([should be `<svg aria-label="">`](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html)), but we'll continue lying that it's okay.

We want to do the right thing. I thought this was an `img` element (which requires `alt`) and didn't realize it's an `svg`. Although based on feedback from @jessiehuff, it seems a little unclear what actually works best: https://github.com/openshift/console/pull/6771#discussion_r498502679

redallen:
Hey @spadgett, I should have recommended `title` to @jcaianirh rather than `aria-label`. We support [pattern 9](https://www.deque.com/blog/creating-accessible-svgs/) in `createIcon.tsx` already, so an easy solution is to just pass the `title` prop instead of `alt`.

spadgett:
@jcaianirh Let's switch to `title`. (Sorry for the flip flop.)

jcaianirh:
Yes will update to title when the pf release drops today. Thanks for following this @spadgett

PR #4939, opened 2020-10-05
feature(Modalbox): add alert variation

### LINKED ISSUE #4824: Modal alert variation
https://github.com/patternfly/patternfly/pull/3487

Adds support for the title to have an icon, and for the modal to take an optional alert variant of default, info, success, warning, or error, which changes the icon color.

There is also hidden screen reader text added to the title when an alert variant is used. For reference, we have that same functionality in the alert component and notification drawer alerts.

KKoukiou:
Closes #4824

* [x]  Blocked on patternfly/patternfly#3563

### DIFF
--- a/packages/react-core/src/components/Modal/ModalBoxTitle.tsx
+++ b/packages/react-core/src/components/Modal/ModalBoxTitle.tsx
@@ -11,7 +11,7 @@ import InfoCircleIcon from '@patternfly/react-icons/dist/js/icons/info-circle-ic
 import BellIcon from '@patternfly/react-icons/dist/js/icons/bell-icon';
 
 export const isModalAlertVariant = (icon: any): icon is string =>
-  ['success', 'error', 'warning', 'info', 'defaut'].includes(icon as string);
+  ['success', 'danger', 'warning', 'info', 'defaut'].includes(icon as string);
 
 export interface ModalBoxTitleProps {
   /** Content rendered inside the modal box header title. */
@@ -19,7 +19,7 @@ export interface ModalBoxTitleProps {
   /** Optional alert icon (or other) to show before the title of the Modal Header
    * When the predefined alert types are used the default styling
    * will be automatically applied */
-  titleIconVariant?: 'success' | 'error' | 'warning' | 'info' | 'default' | React.ComponentType<any>;
+  titleIconVariant?: 'success' | 'danger' | 'warning' | 'info' | 'default' | React.ComponentType<any>;
   /** Optional title label text for screen readers */
   titleLabel?: string;
   /** Additional classes added to the modal box header title. */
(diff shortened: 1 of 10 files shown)


KKoukiou:
@mcoker I would need css review on this but also I would like to raise the attention on the following two issues.

* With the alert variant (similar for Banner, NotificationDrawer) we use 'danger' instead 'error' variant type. Was this rename a deliberate decision or just overlook? 

*Secondly i noticed that the spacing in the react variant which uses the SVGs for icons, instead of <i class .. is smaller than the HTML examples. Any idea why this happens?

mcoker:
> we use 'danger' instead 'error' variant type

Nice catch! That was an oversight, it should match the other components. 

>  the spacing in the react variant which uses the SVGs for icons, instead of <i class .. is smaller than the HTML examples

It's due to the HTML examples maintaining whitespace between elements and the react component removing whitespace.

Created a PR here if you would like to review - https://github.com/patternfly/patternfly/pull/3563

* Renames `pf-m-error` to `pf-m-danger`
* Adds `pf-m-icon` to `pf-c-moda-box__title` when there is an icon in the title, which switches the title to a flex layout so whitespace is removed by default, and the margin between the icon and text is the same between html/react.

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -138,7 +152,12 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
     <ModalBox
       id={boxId}
       style={boxStyle}
-      className={className}
+      className={css(
+        className,
+        isModalAlertVariant(titleIconVariant) &&
+          modalStyles.modifiers[titleIconVariant as 'success' | 'warning' | 'info' | 'danger' | 'default'],
+        titleIconVariant && modalStyles.modifiers.icon
```
mcoker, line 159: `modalStyles.modifiers.icon` should go on `<ModalBoxTitle>` when there is an icon passed. Also when there is a modal, the value of the `title` prop should be wrapped in `<span class="pf-c-modal-box__title-text">`

redallen (changes requested):
I don't see the BullHornIcon in the "Custom icon" example:
image

KKoukiou:
> I don't see the BullHornIcon in the "Custom icon" example:
> image

ups, fixed the issue in the code.

mcoker (dismissed):
LGTM! Thanks @KKoukiou

packages/react-core/src/components/Modal/ModalBoxTitle.tsx
```diff
@@ -1,12 +1,27 @@
...
+import { capitalize } from '../../helpers';
 import { Tooltip } from '../Tooltip';
+import CheckCircleIcon from '@patternfly/react-icons/dist/js/icons/check-circle-icon';
+import ExclamationCircleIcon from '@patternfly/react-icons/dist/js/icons/exclamation-circle-icon';
+import ExclamationTriangleIcon from '@patternfly/react-icons/dist/js/icons/exclamation-triangle-icon';
+import InfoCircleIcon from '@patternfly/react-icons/dist/js/icons/info-circle-icon';
+import BellIcon from '@patternfly/react-icons/dist/js/icons/bell-icon';
+
+export const isModalAlertVariant = (icon: any): icon is string =>
+  ['success', 'danger', 'warning', 'info', 'defaut'].includes(icon as string);
```
redallen, line 14: Can you rename this to `isVariantIcon`? Very clever usage of `icon is string`, I've never seen that before!

packages/react-core/src/components/Modal/ModalBoxTitle.tsx
```diff
@@ -17,25 +32,50 @@ export const ModalBoxTitle: React.FunctionComponent<ModalBoxTitleProps> = ({
   className = '',
   id,
   title,
+  titleIconVariant,
+  titleLabel = '',
   ...props
 }: ModalBoxTitleProps) => {
   const [isTooltipVisible, setIsTooltipVisible] = React.useState(false);
   const h1 = React.useRef<HTMLHeadingElement>();
+  const label =
```
redallen, line 41: Can you define `const isVariantIcon = isVariantIcon(titleIconVariant);` and then use `isVariantIcon` 3 places?

redallen (changes requested):
Just some nits, looking good!

packages/react-core/src/components/Modal/ModalBoxTitle.tsx
KKoukiou: You use the same name for both the boolean variable and the function here.
Did you have some other name in mind?

redallen: Ah sorry, you can call it something else like `isTitleIconVariant`!

KKoukiou: When I define this const and reuse it, somehow typescript is complaining about the types.
 
```
packages/react-core/src/components/Modal/ModalBoxTitle.tsx:65:46 - error TS2538: Type 'FunctionComponent<any>' cannot be used as an index type.

65           {isTitleIconVariant ? variantIcons[titleIconVariant] : <CustomIcon />}
```

I did not research how to fix it a lot to be honest, but I don't know from the top of my head.

redallen (approved):
That's okay, LGTM!

dlabrecq (approved):

mcoker (approved):

PR #4940, opened 2020-10-05
docs(catalog-view-extension): add back

redallen:
**What**: Adds Extensions -> Catalog items back.

**Additional issues**:


christianvogt:
lgtm

thanks!

KKoukiou (approved):
The change makes sense but why it was removed from the docs in the first place?

redallen:
They were removed with the docs refactor. I forgot to add extensions back.

jschuler (approved):

PR #4944, opened 2020-10-05
fix(nav): Fix ariaLabel not getting applied

### LINKED ISSUE #4507: Unable to customize aria-label for Nav

When I set `aria-label="Foo"` on the [Nav](https://www.patternfly.org/v4/documentation/react/components/nav#default) component, it applies the attribute as `aria-label="Local"`. As a result, the custom aria-labels we set in org are not coming through. Is it something we can fix? Does this prop need to be added to the NavContext by chance?

https://www.patternfly.org/v4/documentation/react/components/nav#default

Bug

KKoukiou:
The condition was wrong.

Fixes #4507


redallen (approved):
Nice catch!

tlabaj (approved):

PR #4947, opened 2020-10-05
DataListAction: convert to Functional Component

### LINKED ISSUE #4489: DataListAction should be converted to Functional Component
DataListAction. has a state that is never used.  It also has an onSelect and onToggle function that are not used.  These should be removed and the component should be functional instead os class.  This is a breaking change since the state is exported.

KKoukiou:
Fixes #4489

### DIFF
--- a/packages/react-core/src/components/DataList/DataListAction.tsx
+++ b/packages/react-core/src/components/DataList/DataListAction.tsx
@@ -27,44 +27,20 @@ export interface DataListActionProps extends Omit<React.HTMLProps<HTMLDivElement
   isPlainButtonAction?: boolean;
 }
 
-interface DataListActionState {
-  isOpen: boolean;
-}
-
-export class DataListAction extends React.Component<DataListActionProps, DataListActionState> {
-  static displayName = 'DataListAction';
-  state = {
-    isOpen: false
-  };
-
-  onToggle = (isOpen: boolean) => {
-    this.setState({ isOpen });
-  };
-
-  onSelect = () => {
-    this.setState(prevState => ({
-      isOpen: !prevState.isOpen
-    }));
-  };
-
-  render() {
-    const {
-      children,
-      className,
-      visibility,
-      /* eslint-disable @typescript-eslint/no-unused-vars */
-      id,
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/DataList/DataListAction.tsx
```diff
@@ -27,44 +27,20 @@ export interface DataListActionProps extends Omit<React.HTMLProps<HTMLDivElement
...
-}
+export const DataListAction: React.FunctionComponent<DataListActionProps> = ({
+  children,
+  className,
+  visibility,
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  id,
+  'aria-label': ariaLabel,
+  'aria-labelledby': ariaLabelledBy,
+  isPlainButtonAction,
```
jschuler, line 38: Would be great to update this section. The isPlainButtonAction prop is used and the other three either don't have to be destructured or should be explicitly mapped in the div

KKoukiou: sure, I was just wondering if these aria-labels are desired there or should be instead deprecated and not used as properties at all.
Because I think it's common to use aria-label with div element but then they need role as well.

redallen (changes requested):
LGTM, just need to fix the duplicate `id`s for [this example](https://patternfly-react-pr-4947.surge.sh/components/data-list/react/checkboxes-actions-and-additional-cells/)

packages/react-core/src/components/DataList/DataListAction.tsx
KKoukiou: So I initially implemented to pass these 'props' but the tests caught, that in the scenario, when we have 2 actions for the same item which appear conditionally on the UI according to the screen size, then we end up with duplicate Ids, which is html dom validity violation.

Then we could of course provide different 'id's per action, but that would make card the 'aria-labelled-by' asignment.

Anyway, let's take a look into that into another PR. @redallen

redallen (approved):

kmcfaul (approved):

redallen (approved):

PR #4950, opened 2020-10-06
Set of composable table components

jschuler:
**What**: Towards https://github.com/patternfly/patternfly-react/issues/2358

Adds a set of base table components which provides the option of building a table more explicitly and declaratively (compared to the current Table component that is more configuration based).

I've added corresponding examples to the build.
https://patternfly-react-pr-4950.surge.sh/components/table#tablecomposable-examples

These base components have also been incorporated within the `Table` component.

Example:
```
import { TableComposable, Thead, Tbody, Tr, Th, Td, Caption } from '@patternfly/react-table';

ComposableTableBasic = () => {
  const columns = ['Repositories', 'Branches', 'Pull requests', 'Workspaces', 'Last commit'];
  const rows = [
    ['one', 'two', 'three', 'four', 'five'],
    ['one - 2', null, null, 'four - 2', 'five - 2'],
    ['one - 3', 'two - 3', 'three - 3', 'four - 3', 'five - 3']
  ];
  return (
    <TableComposable aria-label="Simple table using composable components">
      <Caption>Simple table using composable components</Caption>
      <Thead>
        <Tr>
          {columns.map((column, columnIndex) => (
            <Th key={columnIndex}>{column}</Th>
          ))}
        </Tr>
      </Thead>
      <Tbody>
        {rows.map((row, rowIndex) => (
          <Tr key={rowIndex}>
            {row.map((cell, cellIndex) => (
              <Td key={`${rowIndex}_${cellIndex}`} dataLabel={columns[cellIndex]}>
                {cell}
              </Td>
            ))}
          </Tr>
        ))}
      </Tbody>
    </TableComposable>
  );
};
```


packages/react-table/src/components/BaseTable/BaseTableHeaderRow.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
```
redallen, line 1: Can't this and `BaseTableBodyRow` be replaced with `BaseTableRow`?

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -2,38 +2,67 @@
...
+BaseTable,
+BaseTableCaption,
+BaseTableHead,
+BaseTableBody,
+BaseTableHeaderRow,
+BaseTableBodyRow,
+BaseHeaderCell,
+BaseBodyCell,
+HeaderCellInfoWrapper
 } from '@patternfly/react-table';
```
redallen, line 66: You get all react-table imports for free from the theme, so remove this block.

redallen (changes requested):
I think the API can be cleaned up. I suggest the prefix `Simple` instead of `Base` and to use component names that map to the HTML elements.
```jsx
<SimpleTable aria-label="Countries">
  <THead>
    <TRow>
       USA
  ...
  <TBody>
    <TRow isExpanded isSelected isEditable>
      <TD>1</TD>
```

There's also merge conflicts and I'd like to review this another time.

packages/react-table/src/components/Table/Table.tsx
```diff
@@ -28,7 +25,9 @@ export enum TableGridBreakpoint {
 }
 
 export enum TableVariant {
-  compact = 'compact'
+  compact = 'compact',
+  compactBorderless = 'compactBorderless',
+  compactExpandable = 'compactExpandable'
```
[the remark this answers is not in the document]
jschuler: The `getVariant` function checks if the (now deprecated) prop `borders` is used with the `compact` variant, and if so returns the corrected variant

packages/react-table/src/components/Table/examples/ComposableTable.md
```diff
@@ -0,0 +1,831 @@
...
+      </BaseTable>
+    </React.Fragment>
+  );
+};
+```
+
+### Composable: Expandable
+
+```js
+import React from 'react';
```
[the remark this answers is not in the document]
jschuler: I did just now, you reviewed too early ;)

packages/react-table/src/components/BaseTable/BaseBodyCell.tsx
```diff
@@ -0,0 +1,194 @@
...
+  /** On toggling the expansion */
+  onCollapse?: (
+    event: React.MouseEvent,
+    rowIndex: number,
+    isOpen: boolean,
+    rowData: IRowData,
+    extraData: IExtraData
+  ) => undefined;
+  /** True to remove padding */
+  noPadding?: boolean;
```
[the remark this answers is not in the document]
jschuler: Can you clarify what you mean?

packages/react-table/src/components/Table/examples/ComposableTable.md
```diff
@@ -0,0 +1,831 @@
...
+import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
+
+ComposableTableExpandable = () => {
+  const [columns, setColumns] = React.useState([
+    'Repositories',
+    'Branches',
+    'Pull requests',
+    'Workspaces',
+    'Last commit'
+  ]);
```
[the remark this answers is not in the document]
jschuler: Yes, I'll go through the examples and remove state where not needed

jschuler:
@redallen I like your suggestion with giving the inner components simpler names that correspond with the HTML elements, i.e. `Tr` `THead` etc. 
Not sure about `SimpleTable` though, because you could argue either way which way is simpler :) It depends on your use-case I think.

packages/react-table/src/components/BaseTable/BaseTableHeaderRow.tsx
jschuler: I could, but BaseTableBodyRow has some additional props that are not applicable to rows in the header
```
/** Makes the row expandable and determines if it's expanded or not */
  isExpanded?: boolean;
  /** Whether the row is editable */
  isEditable?: boolean;
```

packages/react-table/src/components/Table/examples/ComposableTable.md
```diff
@@ -0,0 +1,1290 @@
...
+    'BaseHeaderCell',
+    'BaseBodyCell',
+  ]
+ouia: true
+beta: true
+---
+
+Note: Table lives in its own package at [@patternfly/react-table](https://www.npmjs.com/package/@patternfly/react-table)!
+
+import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
```
redallen, line 22: Do you not get these imports for free as well from the theme?

packages/react-table/src/components/Table/Table.tsx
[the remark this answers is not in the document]
redallen: Ok, that's fine then!

packages/react-table/src/components/BaseTable/BaseBodyCell.tsx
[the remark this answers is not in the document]
redallen: When I first reviewed this I didn't realize this code is shared between `Table` and `BaseTable`. I was hoping for props that set state like `allRowsSelected` and `onSelect` to be omitted from `Base` components to keep them simpler. Then the consumer can implement setting `isExpanded` or `isSelected` themselves however they please.

This is still possible, it just makes the `Base` components more complex since they have to handle being used in the Table monolith.

packages/react-table/src/components/BaseTable/BaseTableHeaderRow.tsx
redallen: What happens if you set these on a row in `thead`? If it's not undefined behavior I'm in favor of removing this component.

redallen (changes requested):
Thanks for addressing my feedback so fast!

packages/react-table/src/components/BaseTable/BaseTableHeaderRow.tsx
jschuler: OK i kept it to a single Tr component :)

packages/react-table/src/components/Table/examples/ComposableTable.md
jschuler: doesn't seem like it, if I remove them the component is not styled correctly

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1947,11 +1823,1099 @@ class EditableRowsTable extends React.Component {
...
+
+## TableComposable examples
+
+### Composable: Basic
+
+A basic example using the composable table components. The `TableComposable` component differs from the regular `Table` component, in that it allows you to compose the table by nesting the relevant `Thead`, `Tbody`, `Tr`, `Th` and `Td` components within it. For a less declarative and more implicit approach, use the `Table` component instead.
+
+Some general notes:
+
+- Provide `dataLabel` prop to the `Td` components so that in mobile view the cell has a label to provide context.
```
redallen, line 1844: I love everything about this new API except for this fact because it prevents row logic from being separate from column logic. What do you think about some clever JS to pull strings from React nodes in `Thead` and then inject `dataLabel`s in `Tbody` `Td`s? Would this break the existing Table component?

redallen (changes requested):
Code looks good except for a stray comment. I love the paragraphs explaining the examples. Some still have some quirks.

I'm excited to be able to use PatternFly tables with a much simpler API!!

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1947,11 +1823,1099 @@ class EditableRowsTable extends React.Component {
...
+import { TableComposable, Thead, Tbody, Tr, Th, Td, ExpandableRowContent } from '@patternfly/react-table';
+
+import CodeBranchIcon from '@patternfly/react-icons/dist/js/icons/code-branch-icon';
+import CodeIcon from '@patternfly/react-icons/dist/js/icons/code-icon';
+import CubeIcon from '@patternfly/react-icons/dist/js/icons/cube-icon';
+
+// https://github.com/patternfly/patternfly-react/blob/master/packages/react-table/src/components/Table/demo/DemoSortableTable.js
+import DemoSortableTable from './demo/DemoSortableTable';
+
+ComposableCompoundExpandableTable = () => {
```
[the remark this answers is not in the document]
jschuler: nice catch thanks

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1479,37 +1333,47 @@ class ModifiersWithTableText extends React.Component {
 ```
 
 ### Empty state
```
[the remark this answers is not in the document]
jschuler: probably not, removing

packages/react-table/src/components/Table/examples/Table.md
jschuler: `dataLabel` passed to `Td` could be any string that the user wants to display in mobile view, typically it would be the column header. I am not opposed to your idea, I just want to be careful not to start going down the path of the current table component with too much context and state being passed around :) maybe this can be looked at as a further enhancement in the future?

tlabaj (commented):
This is great!
Do you think additional API testing is needed in he integration app? Or should the fact that the Table is built from the compassable pieces enough?

packages/react-table/src/components/Table/examples/Table.md
redallen: I agree I don't want to have to pass around this state. I've opened https://github.com/patternfly/patternfly/issues/3617

redallen (approved):
LGTM!

jschuler:
@tlabaj I've added integration demos to make sure the types are good

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -2,72 +2,171 @@
...
-} from '@patternfly/react-table';
-import styles from '@patternfly/react-styles/css/components/Table/table';
+import { Table, TableHeader, TableBody } from '@patternfly/react-table';
 
 class SimpleTable extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      columns: ['Repositories', 'Branches', 'Pull requests', 'Workspaces', 'Last commit'],
+      rows: [
```
mturley, line 66: Yeah, like this-- this is certainly outside the scope of this PR, but IMO these shouldn't be in state at all, they can just be constants and in the other examples that mutate them for selection/expandable, etc, that state could be separated and correlated with the row at render time.

mturley (commented):
@jschuler this is excellent! Even for debugging the regular Table, it is way more readable with this stuff broken out! Great work.

I have a couple small comments just around how we present the row arrays in our examples. IMO we should decouple the underlying data the rows are presenting from the display details of those rows.

Maybe even naming them `items` instead of `rows` would drive this home. a "row" is a presentational thing (maybe).

packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableComposableDemo.tsx
```diff
@@ -0,0 +1,1012 @@
...
+        fullWidth: true,
+        noPadding: true
+      },
+      {
+        parent: ['parent 5', '2', 'b', 'four', 'five'],
+        child: [
+          {
+            title: "spans 'Repositories' and 'Branches'",
+            props: {
+              colSpan: 2
```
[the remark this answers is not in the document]
jschuler: I agree with you, I was trying to think how to map some props over into the rendered rows, might be better to add that into the JSX directly so it is not confused with having to set a specific configuration

packages/react-table/src/components/Table/examples/Table.md
jschuler: yes, would be good to update these old examples for sure! maybe outside this PR

packages/react-table/src/components/TableComposable/Td.tsx
```diff
@@ -0,0 +1,195 @@
...
+        className,
+        textCenter && styles.modifiers.center,
+        noPadding && styles.modifiers.noPadding,
+        selectParams && selectParams.className,
+        actionParams && actionParams.className,
+        expandableParams && expandableParams.className,
+        compoundParams && compoundParams.className,
+        modifier && styles.modifiers[modifier as 'breakWord' | 'fitContent' | 'nowrap' | 'truncate' | 'wrap'],
+        widthParams && widthParams.className,
+        visibilityParams && visibilityParams.className
```
karelhala, line 182: Can be simplified
```suggestion
        widthParams?.className,
        visibilityParams?.className
```

packages/react-table/src/components/TableComposable/Th.tsx
```diff
@@ -0,0 +1,140 @@
...
+            allRowsSelected: isSelected
+          }
+        }
+      })
+    : null;
+  const widthParams = width ? cellWidth(width)() : null;
+  const visibilityParams = visibility
+    ? classNames(...visibility.map((vis: keyof IVisibility) => Visibility[vis]))()
+    : null;
+  const Component: any = (sortParams && sortParams.component) || (selectParams && selectParams.component) || component;
```
karelhala, line 102: Can be simplified
```suggestion
  const Component: any = sortParams?.component || selectParams?.component || component;
```

packages/react-table/src/components/TableComposable/Th.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const transformedChildren =
+    (sortParams && sortParams.children) || (selectParams && selectParams.children) || children;
```
karelhala, line 104:
```suggestion
    sortParams?.children || selectParams?.children || children;
```

packages/react-table/src/components/TableComposable/Th.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const cell = (
+    <Component
+      onMouseEnter={tooltip !== null ? onMouseEnter : onMouseEnterProp}
+      scope="col"
+      aria-sort={sortParams ? (sortParams['aria-sort'] as 'none' | 'ascending' | 'descending') : null}
```
karelhala, line 109: I don't know about this one type, but I guess this'll work?
```suggestion
      aria-sort={(sortParams?.['aria-sort'] as 'none' | 'ascending' | 'descending' | undefined) || null}
```

packages/react-table/src/components/TableComposable/Th.tsx
```diff
@@ -0,0 +1,140 @@
...
+      ref={innerRef}
+      data-key={columnIndex}
+      className={css(
+        className,
+        textCenter && styles.modifiers.center,
+        sortParams && sortParams.className,
+        selectParams && selectParams.className,
+        modifier && styles.modifiers[modifier as 'breakWord' | 'fitContent' | 'nowrap' | 'truncate' | 'wrap'],
+        widthParams && widthParams.className,
+        visibilityParams && visibilityParams.className
```
karelhala, line 119:
```suggestion
        styles.modifiers?.[modifier as 'breakWord' | 'fitContent' | 'nowrap' | 'truncate' | 'wrap' | undefined],
        widthParams?.className,
        visibilityParams?.className
```

packages/react-table/src/components/Table/Table.tsx
```diff
@@ -415,6 +421,18 @@ export class Table extends React.Component<TableProps, {}> {
...
     });
 
+    const getVariant = () => {
+      if (variant === 'compact') {
+        if (borders === false) {
+          return 'compactBorderless';
+        } else if (onCollapse || onExpand) {
+          return 'compactExpandable';
+        }
+        return 'compact';
```
karelhala, line 431: Redundant `return`

karelhala (commented):
Looking good! Just a few improvements with safe accessor and one object mapper instead of complex if statement.

packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableComposableDemo.tsx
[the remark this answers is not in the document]
mturley: Yeah I think that sets a better example, separating row data from presentation

jessiehuff (dismissed):
This looks pretty good from an accessibility standpoint! I noticed a few things, but I think that they're all preexisting so I'm ok with solving these later. However, I thought it might be valuable to note for our tables in the future:

- Headers repeat in VO (not quite sure what is causing this, ex: "Workspaces, Workspaces column 4 of 5"), it doesn't make it inaccessible by any means but could be annoying to screen reader users.
- All truncated tooltips should have `tabindex="0"` so that keyboard users can also see truncated text.
- The sortable button could use a label to let screen reader users know what this button does.
- Testing this makes me question how we can shift screen reader user focus to a popover when it's opened. It's connected to click events and currently when using a popover we don't shift focus to the popover when it's clicked. When popovers are used in the table, I don't see them in VO at all.
- It does pass an html validator, though it does have one error in terms of the `aria-describedby`. It looks like it's pointing to something that I can't find in the DOM? I assume it's pointing to the tooltip before it exists?

packages/react-table/src/components/TableComposable/TableComposable.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const ouiaProps = useOUIAProps('Table', ouiaId, ouiaSafe);
+  const variantModifiers = () => {
+    if (variant === 'compact') {
+      return styles.modifiers.compact;
+    } else if (variant === 'compactBorderless') {
+      return css(styles.modifiers.compact, styles.modifiers.noBorderRows);
+    } else if (variant === 'compactExpandable') {
+      return css(styles.modifiers.compact, styles.modifiers.expandable);
+    }
+  };
```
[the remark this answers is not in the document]
jschuler: nice, thanks!

packages/react-table/src/components/TableComposable/Td.tsx
```diff
@@ -0,0 +1,195 @@
...
+            rowIndex,
+            column: {
+              extraParams: {}
+            }
+          }
+        )
+      : null;
+  const widthParams = width ? cellWidth(width)() : null;
+  const visibilityParams = visibility
+    ? classNames(...visibility.map((vis: keyof IVisibility) => Visibility[vis]))()
```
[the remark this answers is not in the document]
jschuler: it's already wrapped with conditional and Visibility will always be defined in our code

packages/react-table/src/components/TableComposable/Td.tsx
jschuler: I've made some changes in the code to use the pre-existing `mergeProps` function which simplified this area

packages/react-table/src/components/TableComposable/Th.tsx
jschuler: was able to remove this here since the prop will come in through the `sortable` object that comes back

jschuler:
Thanks @jessiehuff ! that is very detailed, since these issues impact both the old and the new table components I will go ahead and create a separate issue for this, thanks!!

nicolethoen (approved):

dlabrecq (approved):

PR #4951, opened 2020-10-06
fix(Dropdown): test and update dropdown to work better in JAWS

### LINKED ISSUE #3955: unable to use keyboard navigation on dropdown opened via mouse click
When using keyboard navigation to open a Dropdown menu, then first item gets selected and navigation within the dropdown menu can be done using the up and down arrow keys.

This same behavior isn't present when the Dropdown menu is opened via mouse click.

Expect Dropdown menu to behave similar to default browser select menus:
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select

https://patternfly-react.surge.sh/documentation/react/components/dropdown
Same experience with the Select component:
https://patternfly-react.surge.sh/documentation/react/components/select

- Click a Dropdown menu and press arrow keys. No navigation.
- Select a Dropdown menu and press enter key. Navigation of the menu is enabled.

Bug

OpenShift Console

### LINKED ISSUE #4946: Dropdown: test and update dropdown to work better in JAWS
It was recently reported by an OpenShift customer that the dropdown didn't work when using:
Screen reader: **JAWS 2019**
Browser: **Chrome**
Which resulted in the following OpenShift [issue](https://bugzilla.redhat.com/show_bug.cgi?id=1884629). 

**Jenn tested this in JAWS and found:**
The main issues happening with JAWS are:
- Focus stays on the toggle/popup button, and announces the button as collapsed. I think this might be that JAWS announces the state of aria-expanded before we have a chance to update it.
- When Forms mode is not engaged before interacting with the button, the menu items are announced in a confusing way. So the user hears that the button is collapsed, then moves to the next item, the Administrator menu item, but it is announced as “Administrator menu”, and then the next item as “Developer menu”

******For updates to make**, my thinking is to remove aria-haspopup, and see if that improves the experience in JAWS. The aria-haspopup is the attribute that tells JAWS that this is a form component, and it needs to go into Forms mode for allowing the user to interact with it. There are two ways that I know of for enabling Forms mode:
1. use the Tab key to shift focus to the button
2. use the Enter key while focus is on the button

For method 1, the component works as expected, because Forms mode is already enabled before the user hits Enter. For method 2, the component doesn’t work as expected, because the behavior we associate with the Enter key is overwritten by JAWS.

Essentially, JAWS owns all key commands when it’s running. So if you have focus on an element, and say that the down arrow key should jump somewhere, JAWS is going to override that with the behavior that it associates with the Down arrow key, which is to go to the next DOM element. BUT if Forms mode is enabled, that is JAWS acknowledging that in a form, the key commands provided by the form components are more important and therefore stops overriding key commands (i.e. hitting the H key while focus is in an input should enter the character “h”, not jump to the next heading).
So the reason why method 2 is wonky is because Forms mode isn’t enabled yet. And when the user hits the Enter key, JAWS overrides our behavior for the component and instead enters into Forms mode, where we now have freedom to define behavior for key presses.

***I think we could create a testing PR for options, test it in JAWS, and potentially provide it to user to determine the exact ways they are navigating and which option works best.

**For full background, when testing in JAWS her experience was:** 
- When focus is on the toggle and I hit Enter, the button is announced as collapsed, which could be confusing. I am able to use the down arrow key to shift focus, but it announces menu items as “Administrator menu” and “Developer menu”, which could also be confusing. I am able to select the one that has focus and see that the UI behaves as expected.

- Sometimes clicking Space when focus is on the toggle will successfully trigger the focus to move to the menu, but this is difficult to reproduce. When focus is in the menu, I hear “Menu, Administrator button” then “Developer button”.

- If I use the Tab key to shift focus to the component, then JAWS will automatically switch to Forms mode. And then when interacting with the Dropdown, it behaves as expected—focus moves to the menu on open, and the arrow keys let me change my selection, and menu items are announced as “Adminstrator button”.

- Currently Forms mode being enabled is required for this component to work. Upon initial interaction, it will switch to forms mode (due to the presence of aria-haspopup which tells JAWS that this is a form element), but JAWS does not switch to Forms mode in time to make the menu usable. Since this component isn’t in the context of a Form, the likeliness of a JAWS user navigating to the menu using Tab is low.

Bug

jessiehuff:
**What**: 
Closes #4946
Closes #3955 

Attempted testing with `aria-haspopup` and `aria-controls` adaptations, followed WAI ARIA [authoring practices example](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions.html), added delay, used `aria-live` - none of which seemed to work with JAWS in Chrome. 

Adding event listeners after click allows JAWS users to interact with dropdown after its opened. It also allows users to switch easier between mouse and keyboard.


packages/react-core/src/components/Dropdown/Toggle.tsx
```diff
@@ -109,7 +109,7 @@ export class Toggle extends React.Component<ToggleProps> {
     event.preventDefault();
     if ((event.key === 'Tab' || event.key === 'Enter' || event.key === ' ') && this.props.isOpen) {
       this.props.onToggle(!this.props.isOpen, event);
-    } else if ((event.key === 'Enter' || event.key === ' ') && !this.props.isOpen) {
+    } else if ((event.key === 'Enter' || event.key === ' ' || event.keyCode === 40) && !this.props.isOpen) {
```
redallen, line 112: Change to `event.key === 'ArrowDown`

redallen (changes requested):
Globally handling `keydown` events and mucking through `classList` isn't great, but I prefer it to handling refs as state :)

packages/react-core/src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -63,6 +64,39 @@ export class DropdownMenu extends React.Component<DropdownMenuProps> {
...
+    ) {
+      const refs = this.refsCollection;
+      const collectionLength = refs.length;
+      const lastFocusTargetCollection = refs.slice(collectionLength - 1, collectionLength);
+      const lastFocusTarget = lastFocusTargetCollection && lastFocusTargetCollection[0];
+      if (lastFocusTarget[0] && lastFocusTarget[0].focus) {
+        setTimeout(() => lastFocusTarget[0].focus());
+      }
+    }
+  };
```
redallen, line 98: This can be refactored to be DRYer and more readable as well:
```ts
  static validToggleClasses = [styles.dropdownToggle, styles.dropdownToggleButton] as string[];
  static focusFirstRef = (refCollection: HTMLElement[]) => {
    if (refCollection && refCollection[0] && refCollection[0].focus) {
      setTimeout(() => refCollection[0].focus());
    }
  };

  onKeyDown = (event: any) => {
    if (
      !this.props.isOpen ||
      !Array.from(document.activeElement.classList).find(className =>
        DropdownMenu.validToggleClasses.includes(className)
      )
    ) {
      return;
    }
    const refs = this.refsCollection;
    if (event.key === 'ArrowDown') {
      const firstFocusTargetCollection = refs.find(ref => ref && ref[0] && !ref[0].hasAttribute('disabled'));
      DropdownMenu.focusFirstRef(firstFocusTargetCollection);
    } else if (event.key === 'ArrowUp') {
      const collectionLength = refs.length;
      const lastFocusTargetCollection = refs.slice(collectionLength - 1, collectionLength);
      const lastFocusTarget = lastFocusTargetCollection && lastFocusTargetCollection[0];
      DropdownMenu.focusFirstRef(lastFocusTarget);
    }
  };
```

packages/react-core/src/components/Dropdown/Toggle.tsx
jessiehuff: I tried this first, but for some reason it was triggering Shift to open the dropdown as well. Using the keyCode made it so that only the down arrow key worked for it.

packages/react-core/src/components/Dropdown/DropdownMenu.tsx
jessiehuff: Thank you!!! 😄  I knew it needed refactored since it was pretty dirty so I really appreciate your input on that!

jgiardino:
Great job on this, @jessiehuff! 

I tested in Chrome using JAWS 2019, and this seems to work as expected. In cases where the menu item to receive focus is dynamically determined (i.e. not the "With initial selection" example), the menu opens but focus stays on the toggle when using JAWS, which made the menu previously inaccessible in this SR/browser combo. This fix let's me use the Up/Down arrow key to interact with the menu, which also addresses issue #3955 that you referenced earlier. 🎉 🎉 

One thing to note about the keyboard-only experience, different browsers have different behaviors with regard to handling focus for mouse interactions. And this fix doesn't alter that default browser behavior. For example, in Chrome, when I click a button, focus is also placed on the button. So in this case, I can then use Up/Down arrow keys to interact with the menu. But in Firefox, focus is not placed on the item that the user clicks using a mouse, so using the Up/Down arrow keys will not interact with the menu. I think this is fine—I think these differences in behavior are reasons why users choose to use one browser over the other, and we should generally allow these differences in browser behavior to exist rather than override them.

redallen (approved):

tlabaj (approved):
LGTM

PR #4967, opened 2020-10-06
TreeView: Replace button wrapper around actions with a div

### LINKED ISSUE #4799: Tree view: add support for action to not be wrapped in a <button>

Currently, the `action` property in a `TreeViewDataItem` creates the children for a `<button>` that is rendered internally. You can pass props to that button with `actionProps`, but you can't replace it. I'd like to render a `<Dropdown>` in place of that button, since it provides its own toggle button. Currently when I do that I get a button inside a button.

Render a `<Dropdown>` inside the `action` property of a tree view item.

Enhancement

https://github.com/konveyor/mig-ui (Migration Toolkit for Containers), no specific target release.

mturley:
Resolves #4799.
Relevant core change: https://github.com/patternfly/patternfly/pull/3522

Allows any arbitrary JSX to be used in the `action` prop on a TreeViewListItem, wrapping it in a plain div with `.pf-c-tree-view__action` instead of a button. Replaces the existing examples which pass bare icons to this prop with buttons containing those icons.

This makes it possible to render a `<Dropdown>` in the action container of a tree view item without resulting in a button-in-a-button.

BREAKING CHANGES: existing consumers who want to keep the previous behavior must wrap the contents of their `action` in a `<Button>` and pass the values of their `actionProps` directly to the button.

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeView.tsx
+++ b/packages/react-core/src/components/TreeView/TreeView.tsx
@@ -23,10 +23,8 @@ export interface TreeViewDataItem {
   hasBadge?: boolean;
   /** Additional properties of the tree view item badge */
   badgeProps?: any;
-  /** Action of a tree view item, nested inside a button */
+  /** Action of a tree view item, can be a Button or Dropdown */
   action?: React.ReactNode;
-  /** Additional properties of the tree view item action button */
-  actionProps?: any;
 }
 
 export interface TreeViewProps {
@@ -98,7 +96,6 @@ export const TreeView: React.FunctionComponent<TreeViewProps> = ({
         icon={item.icon !== undefined ? item.icon : icon}
         expandedIcon={item.expandedIcon !== undefined ? item.expandedIcon : expandedIcon}
         action={item.action}
-        actionProps={item.actionProps}
         compareItems={compareItems}
         {...(item.children && {
           children: (
--- a/packages/react-core/src/components/TreeView/__tests__/TreeView.test.tsx
+++ b/packages/react-core/src/components/TreeView/__tests__/TreeView.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import { TreeView } from '../TreeView';
+import { Button } from '@patternfly/react-core';
 import { FolderIcon, FolderOpenIcon, FlagIcon } from '@patternfly/react-icons';
 
 const options = [
@@ -92,7 +93,11 @@ const flagOptions = [
     name: 'Cost Management',
     id: 'Cost',
     hasBadge: true,
-    action: <FolderIcon />,
+    action: (
+      <Button>
+        <FolderIcon />
+      </Button>
+    ),
     children: [
       {
         name: 'Application 3',
--- a/packages/react-core/src/components/TreeView/__tests__/TreeView.test.tsx
+++ b/packages/react-core/src/components/TreeView/__tests__/TreeView.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { mount } from 'enzyme';
 import { TreeView } from '../TreeView';
 import { Button } from '@patternfly/react-core';
-import { FolderIcon, FolderOpenIcon, FlagIcon } from '@patternfly/react-icons';
+import { FolderIcon, FolderOpenIcon } from '@patternfly/react-icons';
 
 const options = [
   {
@@ -94,7 +94,7 @@ const flagOptions = [
     id: 'Cost',
     hasBadge: true,
     action: (
-      <Button>
+      <Button variant="plain" aria-label="Folder action">
         <FolderIcon />
       </Button>
     ),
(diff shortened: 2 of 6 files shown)


mcoker:
@christiemolloy can you verify the structure for the action element? Looks like the action element is in `__node`, which requires `__node` be a `div` (or something non-interactive), and also means the taborder focus goes on the whole node (including the action) first, then the action inside.

Looking at select/app-launcher favorites (https://www.patternfly.org/v4/components/application-launcher#with-favorites-and-search), the main item and associated action are siblings. So focus goes from one to the other without one being nested in the other. It seems like the tree view component is setup to do that - then we could get rid of the `__action`'s negative margins and `__node` could continue to be a `button` (or interactive element) as long as it doesn't include a checkbox. Is there any reason we couldn't do that?

packages/react-core/src/components/TreeView/__tests__/TreeView.test.tsx
```diff
@@ -92,7 +93,11 @@ const flagOptions = [
     name: 'Cost Management',
     id: 'Cost',
     hasBadge: true,
-    action: <FolderIcon />,
+    action: (
+      <Button>
```
mcoker, line 97: Can you make this a plain button and add an `aria-label`?

```suggestion
      <Button variant="plain" aria-label="Action">
```

mcoker: And update this in any other examples of a tree view action? By default `<Button>` renders as a primary variant, which we don't want to use here.

redallen (dismissed):
Code LGTM!

packages/react-core/src/components/TreeView/__tests__/TreeView.test.tsx
mturley: Got it, will do.

mturley:
@mcarrano I resolved comments re: variant and aria-label and rebased.

mturley:
@redallen can you re-approve?

redallen (approved):

packages/react-core/src/components/TreeView/TreeView.tsx
```diff
@@ -23,10 +23,8 @@ export interface TreeViewDataItem {
   hasBadge?: boolean;
   /** Additional properties of the tree view item badge */
   badgeProps?: any;
-  /** Action of a tree view item, nested inside a button */
+  /** Action of a tree view item, can be a Button or Dropdown */
   action?: React.ReactNode;
-  /** Additional properties of the tree view item action button */
-  actionProps?: any;
```
dlabrecq, line 29: I know TreeView is relatively new, but isn't this a breaking change?

dlabrecq (changes requested):

christiemolloy:
> @christiemolloy can you verify the structure for the action element? Looks like the action element is in `__node`, which requires `__node` be a `div` (or something non-interactive), and also means the taborder focus goes on the whole node (including the action) first, then the action inside.
> 
> Looking at select/app-launcher favorites (https://www.patternfly.org/v4/components/application-launcher#with-favorites-and-search), the main item and associated action are siblings. So focus goes from one to the other without one being nested in the other. It seems like the tree view component is setup to do that - then we could get rid of the `__action`'s negative margins and `__node` could continue to be a `button` (or interactive element) as long as it doesn't include a checkbox. Is there any reason we couldn't do that?

Yeah that makes sense. Originally it was inside because the design specified that you would hover over the node and the action item would appear.

mcoker:
@christiemolloy great, thanks! I'll submit a core PR to fix the layout in this PR. Issue - https://github.com/patternfly/patternfly/issues/3592

Screen Shot 2020-10-13 at 10 21 46 AM

packages/react-core/src/components/TreeView/TreeView.tsx
mturley: Yeah it is. It's a beta component though, I thought breaking changes were ok / covered by the disclaimer?

mturley: @redallen should I make a note/label somewhere indicating the breaking changes, or will that screw up semantic release versioning since it's beta?

kmcfaul (approved):

packages/react-core/src/components/TreeView/TreeView.tsx
redallen: Just make note of it in the OP of this PR for the release notes.

dlabrecq: Ok, didn't realize this was still beta

dlabrecq (approved):

mturley:
Great, thanks @mcoker !

@redallen should I bump the pf core version here? not sure if there's a process for that, do we normally do version bumps like that as standalone PRs?

redallen:
@mturley Usually we let the bot do it (#4983) but you're welcome to do it in this PR as well. Bump it in one package and then run `node scripts/verifyPatternflyVersions --fix`. I think it's semver comparison package might have broken recently, fixes appreciated!

tlabaj (approved):
LGTM

PR #4971, opened 2020-10-07
feat(FormSection): Add FormSection component

### LINKED ISSUE #4960: Form sections
https://github.com/patternfly/patternfly/pull/3483/files

Adds a `.pf-c-form__section` element that can be used to group `<FormGroup>`s

KKoukiou:
Closes #4960

### DIFF
--- a/packages/react-core/src/components/Form/FormSection.tsx
+++ b/packages/react-core/src/components/Form/FormSection.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Form/form';
+import { css } from '@patternfly/react-styles';
+
+export interface FormSectionProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the section */
+  children?: React.ReactNode;
+  /** Additional classes added to the section */
+  className?: string;
+}
+
+export const FormSection: React.FunctionComponent<FormSectionProps> = ({
+  className = '',
+  children,
+  ...props
+}: FormSectionProps) => (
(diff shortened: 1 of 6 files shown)


redallen (approved):

packages/react-core/src/components/Form/FormSection.tsx
```diff
@@ -0,0 +1,21 @@
...
+  /** Additional classes added to the section */
+  className?: string;
+}
+
+export const FormSection: React.FunctionComponent<FormSectionProps> = ({
+  className = '',
+  children,
+  ...props
+}: FormSectionProps) => (
+  <section {...props} className={css(styles.formSection, className)}>
```
mcoker, line 17: We may build in the ability to pass a `component` prop so an element other than `<section>` can be used. WDYT @jessiehuff? Related, now that we have this, I have opened an issue to look at supporting the `<fieldset>` element in our form layouts https://github.com/patternfly/patternfly/issues/3589

dlabrecq (approved):

packages/react-core/src/components/Form/FormSection.tsx
jessiehuff: I can see that as a benefit. I think it's worth noting that section elements should always have a label. I think it would make sense that when the user has a heading, the section would be labelled by that heading. If a user isn't providing a heading, we could default that html element to a div. I agree with opening the issue that you did. I think it would make sense to explore fieldset and legend with this (which also applies to our current basic form example. The checkboxes used should likely be grouped/associated).

PR #4976, opened 2020-10-08
feat(ContextSelector): add footer to selector

### LINKED ISSUE #4865: Add actions footer to the Context Menu
React follow-up issue to Core issue https://github.com/patternfly/patternfly/issues/3475.

Add an optional footer to the Context menu component. The footer can hold a button right aligned in the footer according to the design in https://github.com/patternfly/patternfly/issues/3475.

Replacement for original contribution #4517

@jessiehuff to investigate accessibility

kmcfaul:
**What**: Closes #4865

### DIFF
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -43,6 +43,8 @@ export interface ContextSelectorProps extends ToggleMenuBaseProps, OUIAProps {
   searchInputPlaceholder?: string;
   /** Function callback for when Search Button is clicked */
   onSearchButtonClick?: (event?: React.SyntheticEvent<HTMLButtonElement>) => void;
+  /** Footer of the Context Selector */
+  footer?: React.ReactNode;
 }
 
 export class ContextSelector extends React.Component<ContextSelectorProps, { ouiaStateId: string }> {
@@ -61,7 +63,8 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
     searchInputPlaceholder: 'Search',
     onSearchButtonClick: () => undefined as any,
     menuAppendTo: 'inline',
-    ouiaSafe: true
+    ouiaSafe: true,
+    footer: null as React.ReactNode
   };
   constructor(props: ContextSelectorProps) {
     super(props);
@@ -98,6 +101,7 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
       menuAppendTo,
       ouiaId,
       ouiaSafe,
+      footer,
       ...props
     } = this.props;
     const menuContainer = (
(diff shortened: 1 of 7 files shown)


MariaAga (dismissed):
Looks good!

dlabrecq (approved):

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -127,6 +131,7 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
             <ContextSelectorContext.Provider value={{ onSelect }}>
               <ContextSelectorMenuList isOpen={isOpen}>{children}</ContextSelectorMenuList>
             </ContextSelectorContext.Provider>
+            {footer && footer}
```
mcoker, line 134: Why wouldn't we add `<ContextSelectorFooter>` for the user if they set `footer`, then pass the value of `footer` to `<ContextSelectorFooter>`?

mcarrano (commented):
This looks great. I just wonder whether we should be added this as a new example to make it clear that this is an optional feature. What do others think?

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
kmcfaul: We could make the `ContextSelectorFooter` ourselves, we would just need to another another prop to pass any extra props to the Footer element. Should I update it?

MariaAga (dismissed):

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
mcoker: Ah, thanks. In that case I'll leave it to your judgment. I anticipate the footer will be pretty simple. But if we add a left-aligned variation for the footer, it would be better to pass that directly to the footer, instead of `isFooterLeftAligned` on `<ContextSelector>`.

redallen: Change to `{footer}`.

redallen (changes requested):
Boolean logic says `a && a == a` for any `a`. Other than that this looks good!

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -43,6 +43,8 @@ export interface ContextSelectorProps extends ToggleMenuBaseProps, OUIAProps {
   searchInputPlaceholder?: string;
   /** Function callback for when Search Button is clicked */
   onSearchButtonClick?: (event?: React.SyntheticEvent<HTMLButtonElement>) => void;
+  /** Footer of the Context Selector */
```
tlabaj, line 46: Nit. Can you make this sentence case.

tlabaj (dismissed):
One little nit.

mcoker (approved):
LGTM!

mcarrano (approved):
Looks great. Thanks for changing the example @kmcfaul .

redallen (approved):

PR #4978, opened 2020-10-08
fix(topology): Do not update nodes on layout when fixed

jeff-phillips-18:
**What**: 
Prevent layout from updating node positions when the node is fixed. Cola layout (in particular) is updating the node position even though the node is set to fixed. Ignore the updates in this case.

Added a ColaNoForce layout to the integration app to verify the fix.

/cc @christianvogt


christianvogt:
lgtm

redallen (approved):

kmcfaul (approved):

PR #4981, opened 2020-10-09
fix(tslib): lock tslib to 1.13.0 for consumers

redallen:
**What**: Our semver of tslib was ^1.11.1, but version 1.14.0 recently released breaking changes that consumers might install if they don't have a lockfile. This PR pins it to the older version until the Typescript team fixes it or we upgrade to tslib 2.

**Additional issues**: I filed an issue upstream: https://github.com/microsoft/tslib/issues/128


ddonahue007 (approved):
LGTM

kmcfaul (approved):

Issue #4989, opened 2020-10-09
Broken stack chart example size


dlabrecq:
Under "Stack chart", the "Monthly data with responsive container" example appears much smaller than expected.

Not certain why it is so small, but it was much larger previously. It should be at least as large as the example below it.

<img width="1196" alt="Screen Shot 2020-10-09 at 2 20 32 PM" src="https://user-images.githubusercontent.com/17481322/95618206-a2c2fa00-0a3a-11eb-853e-d4def8d3d322.png">

PR #4994, opened 2020-10-09
chore(charts): fix stack chart example size

dlabrecq:
Fixes this example, where the chart appears much smaller than expected.

**Before**

Screen Shot 2020-10-09 at 2 20 32 PM

Fixes https://github.com/patternfly/patternfly-react/issues/4989

**After**

Screen Shot 2020-10-09 at 3 04 58 PM


redallen (approved):

ddonahue007 (approved):
LGTM

PR #4996, opened 2020-10-09
fix(datalist): cancel moving for invalid drops

### LINKED ISSUE #4941: Datalist: Fix draggable datalist escape key interaction
Currently the draggable datalist no longer works as expected when using the escape key. Previously you could press enter or space to start the interaction and escape to exit the interaction. Now when trying to escape the interaction, it seems to create a ghost row. You can tab away from it, but this creates some odd behavior and styling: 
Screen Shot 2020-10-05 at 9 42 40 AM

- Some minor things that also could be fixed are that because the ids on the page are not unique, when testing with screen reader, it thinks it's labelled by previous examples so the VO output is odd. Not urgent though as it reads fine when isolated by itself.

Tab to draggable datalist, press enter or space, try to escape the interaction with escape.

Bug

redallen:
**What**: Closes #4941. If the user tries to drag/drop with the mouse outside the `<ul>` cancel the drop altogether. If they use the keyboard and press Escape do the same.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -164,7 +164,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
         draggingToItemIndex: null
       });
     }
-  }
+  };
 
   dragEnd0 = (el: HTMLElement) => {
     el.classList.remove(styles.modifiers.ghostRow);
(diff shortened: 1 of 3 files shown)


jessiehuff (approved):
Keyboard interaction works as expected now, Esc key works great. Thanks for working on this! 🙂

kmcfaul (approved):
Looks good!

PR #4999, opened 2020-10-11
feat(toggleGroup): add suport to text + icon

### LINKED ISSUE #4716: Toggle group - support for text + icon
https://github.com/patternfly/patternfly/pull/3373

boaz0:
**What**: Closes #4716 

I am still not sure whether the text can be passed through the `children` or the `text` props.

WDYT?

//cc @mcoker

### DIFF
--- a/packages/react-core/src/components/ToggleGroup/ToggleGroupItem.tsx
+++ b/packages/react-core/src/components/ToggleGroup/ToggleGroupItem.tsx
@@ -2,19 +2,15 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/ToggleGroup/toggle-group';
 import ToggleGroupContext from './ToggleGroupContext';
-
-export enum ToggleGroupItemVariant {
-  icon = 'icon',
-  text = 'text'
-}
+import { ToggleGroupItemVariant, ToggleGroupItemElement } from './ToggleGroupItemElement';
 
 export interface ToggleGroupItemProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange'> {
-  /** Content rendered inside the toggle group item */
-  children?: React.ReactNode;
+  /** Text rendered inside the toggle group item */
+  text?: React.ReactNode;
+  /** Icon rendered inside the toggle group item */
+  icon?: React.ReactNode;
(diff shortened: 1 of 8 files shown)


nicolethoen (approved):
LGTM - at first i could not decide if the extra ToggleGroupItemElement was necessary - it probably isn't. But, since none of it is exposed to the user, any changes to the ToggleGroupItemElement wont be breaking, so I guess it just lets us flesh out tests better to have it.

mcoker:
> I am still not sure whether the text can be passed through the children or the text props.

@boaz0 what you have works, but if I'm reading it right, we could set it up like a button and keep the `variant === ToggleGroupItemVariant.icon` prop to indicate it's an icon/plain button and wrap `children` in the `.pf-c-toggle-grup__icon` element.

If it's not the icon variant, the check for an `icon` prop for the icon source and `iconPosition` prop for whether to put the icon before or after the text.

mcarrano (approved):
Looks good to me.

dlabrecq (approved):

tlabaj (approved):

boaz0:
@nicolethoen 

> at first i could not decide if the extra ToggleGroupItemElement was necessary - it probably isn't. But, since none of it is exposed to the user, any changes to the ToggleGroupItemElement wont be breaking, so I guess it just lets us flesh out tests better to have it.

Yes, definitely I am not happy with adding `ToggleGroupItemElement`. Maybe we can think of a better way to do it.
I think @mcoker 's idea will fix that. WDYT?

@mcoker thanks for the feedback.

> we could set it up like a button and keep the variant === ToggleGroupItemVariant.icon prop to indicate it's an icon/plain button and wrap children in the .pf-c-toggle-grup__icon element.
> If it's not the icon variant, the check for an icon prop for the icon source and iconPosition prop for whether to put the icon before or after the text and the text is what's passed as children

That's actually a good idea. Although the only disadvantage that I can think of is that it is less trivial than what we have now.

@dlabrecq  @tlabaj  and @nicolethoen

redallen (approved):
LGTM. Do you want this merged @boaz0 and @mcoker ?

redallen:
Merging. Feel free to use this thread to discuss further.

boaz0:
@mcoker do you want to discuss about it through email or slack?

PR #5002, opened 2020-10-12
feat(Modal): add help section to the modal header

### LINKED ISSUE #4883: Add optional Help button to modal header
Includes an optional help button to the left of the Close button. Core issue here: https://github.com/patternfly/patternfly/issues/3474

For keyboard interaction, the Close button should continue to have focus when the user opens the modal. @jessiehuff will further investigate overall keyboard navigation when the help button is present.

kmcfaul:
**What**: Closes #4883

### DIFF
--- a/packages/react-core/src/components/Modal/examples/Modal.md
+++ b/packages/react-core/src/components/Modal/examples/Modal.md
@@ -832,7 +832,6 @@ class HelpModal extends React.Component {
                 <div>
                   Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla
                   turpis.
-                  <Button onClick={() => console.log('click')}>click</Button>
                 </div>
               }
               footerContent="Popover Footer"
--- a/packages/react-core/src/components/Modal/examples/Modal.md
+++ b/packages/react-core/src/components/Modal/examples/Modal.md
@@ -799,7 +799,7 @@ class WithDropdown extends React.Component {
 
 ```js
 import React from 'react';
-import { Modal, Button } from '@patternfly/react-core';
+import { Modal, Button, Popover } from '@patternfly/react-core';
 import HelpIcon from '@patternfly/react-icons/dist/js/icons/help-icon';
 
 class HelpModal extends React.Component {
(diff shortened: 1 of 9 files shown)


dlabrecq (dismissed):
I see the help icon in the example, but it's not functional. I assume this would launch a popover or tooltip? It would be helpful to see how that pattern is expected to behave in the modal.

mcarrano (commented):
@mcoker @mceledonia I thought we agreed that the Help button would be right-aligned and grouped with the Close button. Am I mistaken about that?

kmcfaul:
@dlabrecq You are right that would be very helpful. I made the help a popover for the example!

mcoker:
@mcarrano I have a core issue here to right align the help icon https://github.com/patternfly/patternfly/pull/3603

It's CSS only so once it's pulled in to react, the alignment should be fixed.

redallen (approved):
Code LGTM!

packages/react-core/src/components/Modal/examples/Modal.md
```diff
@@ -793,3 +794,73 @@ class WithDropdown extends React.Component {
...
+        <Modal
+          title="Simple modal header"
+          help={
+            <Popover
+              headerContent={<div>Help Popover</div>}
+              bodyContent={
+                <div>
+                  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam id feugiat augue, nec fringilla
+                  turpis.
+                  <Button onClick={() => console.log('click')}>click</Button>
```
tlabaj, line 835: I don't know that we need this button in the example.  It looks a little off to me.  @mcarrano can you take a look at the popover content.  I am not sure that the `click` button would follow design recommendations.

tlabaj (commented):
Other than waiting for the core bump to replace that inline style and the question I had about the `click` button in the popover, it looks good.

packages/react-core/src/components/Modal/examples/Modal.md
kmcfaul: I will remove it.

Also, there is no actual `pf-c-modal-box__header-help` class, it doesn't do anything. It is a placeholder class to indicate the div.

mcarrano (approved):

redallen (approved):

jessiehuff (approved):
Small nit, but overall really great! 😄 When navigating by keyboard, when you try to hit escape to close the popover, you close the entire modal. We likely need to stop event propagation in the example (when the popover is open, escape should close the popover, when the popover is closed, escape should close the modal). Would be great to add this small aspect, but I'm ok with us adding that later!

mcoker (approved):
I ❤️ it!

PR #5003, opened 2020-10-12
Add git hook for linting

redallen:
**What**: Updates eslint and related packages. Adds `"lint:all": "yarn lint:md && yarn lint:versions && yarn lint:ts"` target.  Adds `husky` that will run `yarn lint:all --fix` on `git commit`.

**Additional issues**:


.eslintrc.json
```diff
@@ -91,7 +89,7 @@
         "no-eval": "error",
         "no-new-wrappers": "error",
         "no-prototype-builtins": "off",
-        "no-shadow": "error",
+        "no-shadow": "off",
```
dlabrecq, line 92: Do we want to turn off no-shadow -- seems fairly useful?

dlabrecq (approved):

.eslintrc.json
redallen: `no-shadow` started complaining about every enum definition being shadowed after updating eslint packages.

jschuler (approved):

.eslintrc.json
dlabrecq: Thanks for the explanation. I thought it was something like that.

Issue #5005, opened 2020-10-13
Cart legend example: Fix createContainer property order


dlabrecq:
When the chart is re-rendered, the legend tooltip stops working properly. For example, when the browser window is resized.

This can be best seen with the tooltip example below.

http://patternfly-react.surge.sh/charts/stack-chart#multi-color-unordered-responsive-container

To solve the issue, we simply change the order in which "voronoi" and "cursor" container properties are merged together. That is, we need to create a container like below, where "voronoi" properties are applied first.

`const CursorVoronoiContainer = createContainer("voronoi", "cursor");`

PR #5006, opened 2020-10-13
chore(charts): Legend tooltip example: Fix createContainer property order

dlabrecq:
This fixes the `createContainer` property order for the chart legend tooltip examples.

Ensures the tooltip continues to work after charts are re-rendered (e.g., when the browser window is resized).

Fixes https://github.com/patternfly/patternfly-react/issues/5005


ddonahue007 (approved):
LGTM

redallen (approved):

PR #5009, opened 2020-10-13
add backdrop examples

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/5008


dlabrecq (dismissed):
The example oddly shows a giant question mark. Is that expected?

redallen (approved):
Many of these image diffs are text rendering differences (probably due to the order in which CSS is loaded?). We should work to make these consistent upstream in patternfly-org later.

nicolethoen (approved):

dlabrecq (approved):

PR #5025, opened 2020-10-14
Page section enhancements

### LINKED ISSUE #4962: Page section enhancements - group sections, make top/bottom sticky, add top/bottom shadow, make scrollable
So much fun stuff packed into one PR!  🎉 🥳

https://github.com/patternfly/patternfly/pull/3516

* Adds a new `pf-c-page__main-group` element that can be used to group page sections
* Adds sticky modifiers to make something sticky to the top or bottom of the page main element
  * `pf-m-sticky-top` and `pf-m-sticky-bottom`
* Adds shadow modifiers to add a top or bottom shadow to a page section
  * `pf-m-shadow-top` and `pf-m-shadow-bottom`
* Adds `overflow: auto` (enables a scrollbar when there is overflow content) to a page section
  * `pf-m-overflow-scroll`

Here's a list of the elements that get these new modifiers

```
.pf-c-page__main-nav,
.pf-c-page__main-breadcrumb,
.pf-c-page__main-section,
.pf-c-page__main-wizard,
.pf-c-page__main-group {
  &.pf-m-sticky-top { // top sticky code
  &.pf-m-sticky-bottom { // top sticky code
  &.pf-m-overflow-scroll { // top sticky code
  &.pf-m-shadow-bottom { // top sticky code
  &.pf-m-shadow-top { // top sticky code
}
```

kmcfaul:
**What**: Closes #4962

Adds two methods for grouping to prevent potentially breaking changes.

Firstly, you can group the tertiary nav and breadcrumb by using `isTertiaryNavGrouped` and `isBreadcrumbGrouped` along with `additionalGroupedContent` and `groupProps` to provide other sections to be included in that same grouping and group props respectively. The group will be the first element, and will be followed by whatever was not grouped (by leaving out its flag), followed by the Page children.

Alternatively (and maybe the 'breaking way'), you can use the new components `PageBreadcrumb` and `PageNavigation` along with `PageGroup` directly and pass everything to the `Page` via the children. This method would also let the user control the order of components in the group containing the breadcrumb/nav.

The example section contains an demo for the new components, and the demos section contains demos for both methods.

### DIFF
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -232,13 +232,13 @@ export class Page extends React.Component<PageProps, PageState> {
 
     const isGrouped = isTertiaryNavGrouped || isBreadcrumbGrouped;
 
-    const group = (
+    const group = isGrouped ? (
       <PageGroup {...groupProps}>
         {isTertiaryNavGrouped && nav}
         {isBreadcrumbGrouped && crumb}
         {additionalGroupedContent}
       </PageGroup>
-    );
+    ) : null;
 
     const main = (
       <main
@@ -249,7 +249,7 @@ export class Page extends React.Component<PageProps, PageState> {
         tabIndex={mainTabIndex}
         aria-label={mainAriaLabel}
       >
-        {isGrouped && group}
+        {group}
         {!isTertiaryNavGrouped && nav}
         {!isBreadcrumbGrouped && crumb}
         {children}
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -73,9 +73,13 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   tertiaryNav?: React.ReactNode;
   /** Accessible label, can be used to name main section */
   mainAriaLabel?: string;
+  /** Flag indicating if the tertiaryNav should be in a group */
   isTertiaryNavGrouped?: boolean;
+  /** Flag indicating if the breadcrumb should be in a group */
   isBreadcrumbGrouped?: boolean;
+  /** Additional content of the group */
   additionalGroupedContent?: React.ReactNode;
+  /** Additional props of the group */
   groupProps?: PageGroupProps;
 }
 
@@ -231,7 +235,7 @@ export class Page extends React.Component<PageProps, PageState> {
     const group = (
       <PageGroup {...groupProps}>
         {isTertiaryNavGrouped && nav}
-        {isBreadcrumbGrouped && breadcrumb}
+        {isBreadcrumbGrouped && crumb}
         {additionalGroupedContent}
       </PageGroup>
     );
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -230,7 +230,7 @@ export class Page extends React.Component<PageProps, PageState> {
       )
     ) : null;
 
-    const isGrouped = isTertiaryNavGrouped || isBreadcrumbGrouped;
+    const isGrouped = isTertiaryNavGrouped || isBreadcrumbGrouped || additionalGroupedContent;
 
     const group = isGrouped ? (
       <PageGroup {...groupProps}>
--- a/packages/react-core/src/components/Page/Page.tsx
+++ b/packages/react-core/src/components/Page/Page.tsx
@@ -4,6 +4,7 @@ import { css } from '@patternfly/react-styles';
 import globalBreakpointXl from '@patternfly/react-tokens/dist/js/global_breakpoint_xl';
 import { debounce } from '../../helpers/util';
 import { Drawer, DrawerContent, DrawerContentBody, DrawerPanelContent } from '../Drawer';
+import { PageGroup, PageGroupProps } from './PageGroup';
 
 export enum PageLayouts {
   vertical = 'vertical',
@@ -72,6 +73,10 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   tertiaryNav?: React.ReactNode;
   /** Accessible label, can be used to name main section */
   mainAriaLabel?: string;
+  isTertiaryNavGrouped?: boolean;
+  isBreadcrumbGrouped?: boolean;
+  additionalGroupedContent?: React.ReactNode;
+  groupProps?: PageGroupProps;
 }
 
 export interface PageState {
@@ -187,6 +192,10 @@ export class Page extends React.Component<PageProps, PageState> {
       mainAriaLabel,
       mainTabIndex,
       tertiaryNav,
+      isTertiaryNavGrouped,
+      isBreadcrumbGrouped,
+      additionalGroupedContent,
+      groupProps,
       ...rest
     } = this.props;
     const { mobileView, mobileIsNavOpen, desktopIsNavOpen } = this.state;
@@ -197,6 +206,36 @@ export class Page extends React.Component<PageProps, PageState> {
       isNavOpen: mobileView ? mobileIsNavOpen : desktopIsNavOpen
     };
 
+    const nav = tertiaryNav ? (
+      isTertiaryNavWidthLimited ? (
+        <div className={css(styles.pageMainNav, styles.modifiers.limitWidth)}>
+          <div className={css(styles.pageMainBody)}>{tertiaryNav}</div>
+        </div>
+      ) : (
+        <div className={css(styles.pageMainNav)}>{tertiaryNav}</div>
+      )
+    ) : null;
+
+    const crumb = breadcrumb ? (
+      isBreadcrumbWidthLimited ? (
+        <section className={css(styles.pageMainBreadcrumb, styles.modifiers.limitWidth)}>
+          <div className={css(styles.pageMainBody)}>{breadcrumb}</div>
+        </section>
+      ) : (
+        <section className={css(styles.pageMainBreadcrumb)}>{breadcrumb}</section>
+      )
+    ) : null;
+
+    const isGrouped = isTertiaryNavGrouped || isBreadcrumbGrouped;
+
+    const group = (
+      <PageGroup {...groupProps}>
+        {isTertiaryNavGrouped && nav}
+        {isBreadcrumbGrouped && breadcrumb}
+        {additionalGroupedContent}
+      </PageGroup>
+    );
+
     const main = (
       <main
         ref={this.mainRef}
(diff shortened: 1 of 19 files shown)


mcarrano (commented):
Not sure exactly what to look for here. Is this addressing the sticky header/footer issue? Is there a specific demo I should look for here? Will defer to @mcoker to approve.

kmcfaul:
@mcarrano There is a "Group Section" example showing the new group component for sections, and two full page group demos in the demo's section of Page's documentation (showing both syntax options for JS, but visually will look the same) that also display the `sticky` and `shadow` flags.

mcarrano:
Thanks for the clarification @kmcfaul . Looks good. The only thing I wonder is whether the demo names should be different. You name the examples "Grouped tertiary nav and breadcrumb" which to me implies that we have something called a "Grouped tertiary nav" which we do not. Maybe you want to say grouped header? Grouped section?

mcoker:
Agree the "grouped" name is confusing since we have a demo called "grouped nav" that shows titles & section groups. Also "grouped tertiary nav and breadcrumb" is a group with those and the first page section with title & description. I think "Grouped sections" would work.

mcoker:
We call this basic demo "sticky section group" in core, which is probably better, since you could use a group that isn't sticky.

mcoker:
This is kind of random, but the header image is black - do you know if that file was updated recently? It loses its size on a desktop layout and doesn't show, but is visible like this on a small viewport https://patternfly-react-pr-5025.surge.sh/components/page/react-demos/grouped-tertiary-nav-and-breadcrumb-alternate-syntax/

Screen Shot 2020-10-22 at 11 13 40 AM

mcoker:
There shouldn't be all of this white space in the title/paragraph section. I put up a PR in core to fix it, I'll let you know when that's merged.
 
Screen Shot 2020-10-22 at 11 15 40 AM

packages/react-core/src/components/Page/__tests__/Page.test.tsx
```diff
@@ -80,6 +84,151 @@ test('Check page to verify breadcrumb is created', () => {
...
+  expect(view.find('.pf-c-page__main').getDOMNode().id).toBe('');
+});
+
+test('Check page to verify grouped nav and breadcrumb - new components syntax', () => {
+  const Header = <PageHeader logo="Logo" headerTools="PageHeaderTools | Avatar" topNav="Navigation" />;
+  const Sidebar = <PageSidebar isNavOpen />;
+
+  const view = mount(
+    <Page {...props} header={Header} sidebar={Sidebar}>
+      <PageGroup sticky="bottom" hasShadowTop>
```
mcoker, line 148: You don't need to specify a shadow if it's sticky. A sticky position will automatically apply the correct shadow (top sticky = bottom shadow, bottom sticky = top shadow)

```suggestion
      <PageGroup sticky="bottom">
```

mcoker (dismissed):
Code all LGTM! Great job!!

kmcfaul:
@mcoker Not sure about the patternfly icon, it uses the same framework / files as the other examples.

mcoker (dismissed):

dlabrecq (dismissed):

nicolethoen (approved):

packages/react-core/src/components/Page/examples/Page.md
```diff
@@ -299,3 +302,93 @@ class UncontrolledNavPage extends React.Component {
...
+
+```js
+import React from 'react';
+import {
+  Page,
+  PageHeader,
+  PageHeaderTools,
+  PageSidebar,
+  PageSection,
+  PageGroup,
```
jschuler, line 316: Add the new components to the props at the top of the examples file so they show up in the docs

packages/react-core/src/components/Page/Page.tsx
```diff
@@ -197,6 +210,36 @@ export class Page extends React.Component<PageProps, PageState> {
...
+          <div className={css(styles.pageMainBody)}>{breadcrumb}</div>
+        </section>
+      ) : (
+        <section className={css(styles.pageMainBreadcrumb)}>{breadcrumb}</section>
+      )
+    ) : null;
+
+    const isGrouped = isTertiaryNavGrouped || isBreadcrumbGrouped;
+
+    const group = (
```
jschuler, line 235: not sure about performance impact.. but you could add a check `isGrouped` and if false return null

packages/react-core/src/components/Page/Page.tsx
```diff
@@ -72,6 +73,14 @@ export interface PageProps extends React.HTMLProps<HTMLDivElement> {
   tertiaryNav?: React.ReactNode;
   /** Accessible label, can be used to name main section */
   mainAriaLabel?: string;
+  /** Flag indicating if the tertiaryNav should be in a group */
+  isTertiaryNavGrouped?: boolean;
```
jschuler, line 77: is it a valid use case to group content but leave the tertiary nav or breadcrumbs out of it?

packages/react-core/src/components/Page/PageBreadcrumb.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Page/page';
+
+export interface PageBreadcrumbProps extends React.HTMLProps<HTMLDivElement> {
```
jschuler, line 5: I believe `section` would just extend `HTMLElement` based on what I am seeing here
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d063d0175b3eb59a7a4cc5e64992f8a220d9a910/types/react/index.d.ts#L2749

jschuler (commented):
looks good otherwise

packages/react-core/src/components/Page/Page.tsx
kmcfaul: This gives a user flexibility at least. I could change it to an all or nothing flag?

packages/react-core/src/components/Page/Page.tsx
kmcfaul: For `group`? Sure I'll add it.

packages/react-core/src/components/Page/Page.tsx
```diff
@@ -197,6 +210,36 @@ export class Page extends React.Component<PageProps, PageState> {
...
+      isBreadcrumbWidthLimited ? (
+        <section className={css(styles.pageMainBreadcrumb, styles.modifiers.limitWidth)}>
+          <div className={css(styles.pageMainBody)}>{breadcrumb}</div>
+        </section>
+      ) : (
+        <section className={css(styles.pageMainBreadcrumb)}>{breadcrumb}</section>
+      )
+    ) : null;
+
+    const isGrouped = isTertiaryNavGrouped || isBreadcrumbGrouped;
```
jschuler, line 233: oh, i think you'll also want to add `|| additionalGroupedContent` in case they don't want to group tertiary and breadcrumb but still show some grouped sticky content at the top?

packages/react-core/src/components/Page/Page.tsx
jschuler: okay I was curious because if it was all or nothing it could simplify the API, but after playing with the demo this is fine

packages/react-core/src/components/Page/Page.tsx
kmcfaul: Will add. Because users may also use the group component directly, they could also do that instead of using these props in this case.

jschuler (approved):

packages/react-core/src/demos/Page/Page.md
```diff
@@ -1198,7 +1207,471 @@ class PageLayoutTertiaryNav extends React.Component {
...
+    const Header = (
+      <PageHeader logo={<Brand src={imgBrand} alt="Patternfly Logo" />} headerTools={headerTools} showNavToggle />
+    );
+    const pageId = 'main-content-page-layout-tertiary-nav';
+    const PageSkipToContent = <SkipToContent href={`#${pageId}`}>Skip to content</SkipToContent>;
+
+    return (
+      <React.Fragment>
+        <Page header={Header} isManagedSidebar skipToContent={PageSkipToContent} mainContainerId={pageId}>
+          <PageGroup sticky="top" hasShadowBottom>
```
mcoker, line 1613: don't need `hasShadowBottom` - comes with `sticky="top"`

```suggestion
          <PageGroup sticky="top">
```

packages/react-core/src/demos/Page/Page.md
```diff
@@ -1198,7 +1207,471 @@ class PageLayoutTertiaryNav extends React.Component {
...
+                <Text component="p">
+                  Body text should be Overpass Regular at 16px. It should have leading of 24px because <br />
+                  of it’s relative line height of 1.5.
+                </Text>
+              </TextContent>
+            </PageSection>
+          }
+          groupProps={{
+            sticky: 'top',
+            hasShadowBottom: true
```
mcoker, line 1427: not needed

```suggestion
```

packages/react-core/src/demos/Page/Page.md
```diff
@@ -1198,7 +1207,471 @@ class PageLayoutTertiaryNav extends React.Component {
...
+                <Text component="h1">Main title</Text>
+                <Text component="p">
+                  Body text should be Overpass Regular at 16px. It should have leading of 24px because <br />
+                  of it’s relative line height of 1.5.
+                </Text>
+              </TextContent>
+            </PageSection>
+          }
+          groupProps={{
+            sticky: 'top',
```
mcoker, line 1426:
```suggestion
            sticky: 'top'
```

nicolethoen (dismissed):

dlabrecq (dismissed):

nicolethoen (approved):

packages/react-core/src/components/Page/PageGroup.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface PageGroupProps extends React.HTMLProps<HTMLDivElement> {
+  /** Additional classes to apply to the PageGroup */
+  className?: string;
+  /** Content rendered inside of the PageGroup */
+  children?: React.ReactNode;
+  /** Modifier indicating if PageGroup is sticky to the top or bottom */
+  sticky?: 'top' | 'bottom';
+  /** Modifier indicating if PageGroup should have a shadow at the top */
+  hasShadowTop?: boolean;
```
jschuler, line 13: sticky top adds shadow bottom and sticky bottom adds shadow top by default right? but then you can’t disable it? then the prop name and/or description could be confusing since it seems you should be able to disable it.
It’s really not a modifier toggle, just an add, no?
so instead of `hasShadowBottom` should it be `addShadowBottom` and the same for the top?

jschuler: I don't know, maybe it's fine, it was just confusing to me in the moment. Perhaps just a description change to indicate that it is enabled by default for sticky?

mcoker: IMO the way it is now is fine. We use similar props for `pf-m-padding` and `pf-m-no-padding` to add padding to something that doesn't have it by default, and remove it from something that has it by default. I think the shadows behave the same here.

The way we handle those props in the drawer is `hasPadding` and `hasNoPadding`

https://github.com/patternfly/patternfly-react/blob/55e07e690fb8aa26288f3294f13d5819beaa961d/packages/react-core/src/components/Drawer/DrawerContentBody.tsx#L11

https://github.com/patternfly/patternfly-react/blob/55e07e690fb8aa26288f3294f13d5819beaa961d/packages/react-core/src/components/Drawer/DrawerPanelBody.tsx#L11

mcoker: You can see the use case for adding shadows to page sections here - https://pf4.patternfly.org/components/page/html-demos/overflow-scroll/

The top and bottom sections have shadows but aren't sticky. That page is 3 regular page sections, with the middle one is set to `pf-m-overflow-scroll`, meaning it enables a scrollbar for overflow content.

jschuler (approved):

mcoker (approved):

Issue #4732, opened 2020-08-26
Build Date Range picker


nicolethoen:

unlikelyzero:
@mcarrano @nicolethoen how does this relate to https://github.com/patternfly/patternfly-react/issues/4504

nicolethoen:
@unlikelyzero 
This issue is part of the Date and Time picker epic.
It is the next priority after introducing a basic Time picker (https://github.com/patternfly/patternfly-react/issues/4731) and cleaning up the calendar (https://github.com/patternfly/patternfly-react/issues/5057)

PR #5026, opened 2020-10-14
fix(datepicker): add calendar component

### LINKED ISSUE #5057: Update implementation for Date Picker
Follow-up work from initial implementation #4730 

Date Picker was initially implemented using Flatpickr library, which provided some functionality but also required battling against some of their features to get what we need.

To avoid this moving forward, replace Flatpickr with custom implementation while maintaining features from existing implementation.

### LINKED ISSUE #5114: Popover enhancements - auto width and no padding
https://github.com/patternfly/patternfly/pull/3646/

Popover now supports:
* `.pf-m-no-padding` - removes the outer padding for popover layouts that should consume the entire dialog.
* `.pf-m-auto-width` - removes the fixed width and allows the popover width to be defined by the content inside.

redallen:
**What**: Replace flatpickr with a Popover containing our new CalendarMonth component.

https://patternfly-react-pr-5026.surge.sh/components/date-picker

(Temporarily): http://preview.patternfly.org/components/date-picker

Interaction design largely taken from W3: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

Closes #5057
Closes #5114

**Additional issues**: Unblocks #4732.


nicolethoen:
This looks great! I feel like the keyboard interaction is not quite what I expect.
I _think_ I expect the calendar to open on focus rather than on enter, since if i focus on the date, then hit enter, I would not expect the calendar to open (when I feel like I'm done entering a date). That, however, is not my specialty - I could have poor instincts lol. @jessiehuff is who we should defer to for that.

mcarrano (changes requested):
@redallen There seem to be some problems with interaction between the calendar and the field. If I open the calendar and click to select a date, it's fine. But if I change the month, it looks like a new date is being selected but it's not passed to the field, Changing the month or year should keep the calendar open and not show any date selected until the user clicks on a date, I think. Does that make sense to you?

mcarrano:
I also wonder about positioning of the calendar @redallen . You chose to center the calendar to the field.

Screen Shot 2020-11-10 at 2 02 50 PM

Looks like the current component is left aligning the calendar

Screen Shot 2020-11-10 at 2 02 35 PM

I actually wonder if the calendar should be right-aligned so the popover points to the calendar icon. What do others think?

redallen:
@nicolethoen When it comes to that interaction I think we should follow the [W3 reference implementation](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html) which behaves by opening the Popover only when the icon to the right of the input is pressed. Sadly with our design the "button" on the right side of our calendar is actually just a background on the input so we can't follow that pattern unless Core refactors the input to have a trailing button.

As for the focus, I'm not sure there's an easy way to only "trap" focus if the Popover is opened via keyboard. Our current flatpickr implementation is totally inaccessible in this regard. Yes I agree it's odd to have to click twice to type, but what should the CalendarMonth do as you type? Should it flash and change as you type?

@mcarrano Sure that makes sense, I'll push the styling changes to this PR. As for the positioning, it's our default Popover positioning and I'm partial to it because I don't think there's an easy way to change the alignment, but cc @jschuler for Popper possibilities.

jessiehuff (commented):
I think you both make great points here. I definitely see what Nicole means, but I think Zack is right that you'd need a separate button to click for the date picker. I don't really see any significant a11y issues with the current behavior so if Matt is good with the interaction from a UX perspective then I think it's fine. (I imagine this would just require users to navigate to the submit button instead of a default enter submitting in the context of a form.) I did see one issue though, but maybe it's a core issue @mcoker: 
image (8)I.e. if the current selected date is visible in the previous month, the font color doesn’t change for the selected date (the color is based on the previous/next month styles).

I also notice that if I have the month dropdown open and press Esc, it closes the entire date picker, not just the menu.

In terms of the screen reader experience, I think it could be improved a bit. I wonder if the days of the week should be labeled for a screen reader to make it clearer. (For a visual user, it completely makes sense to take the lettering approach. For a screen reader, I wonder if it could be confusing to know whether you're on Tuesday or Thursday when it says T and would require navigating around every time to determine your place.) I asked Jenn on her thoughts of that aspect, and she thought it'd be preferable to have a hidden span with "Friday" and the "F" with `aria-hidden`, but hopefully all of these strings would be customizable or could be internationalization-able. I also think that it would be ideal to have the date buttons `aria-labelledby` the month and year so it would announce something like "4 March 2020". I'm open to thoughts on these aspects though. Mostly I want it to be clear to screen reader users what's happening as they're trying to navigate.

Looks pretty good so far! 🙂

mcarrano:
I took another look. Seems like this is coming along, but I do have a couple of comments:

* Looks like the way it's working now is that the calendar opens as soon as I put focus into the field. That's OK, but it looks like the field then looses focus so I can't enter a date from the keyboard without closing the calendar. That seems unexpected. I'm comparing against the current datapicker implementation with Flatpicker and there the focus remains in the field even when the calendar is opened. Can we make it work like that?

* Because we are using the same visual effect to denote focus and hover, you can get into a state like this with two dates highlighted:

Screen Shot 2020-11-11 at 1 50 36 PM

If I click, July 10 will be returned, but if I press the Enter key July 15 will be returned. Should we be using the same effect for both focus and hover? Seems confusing.

redallen:
Re @jessiehuff :

> I did see one issue though, but maybe it's a core issue

I've easily fixed this for now by not applying the previous month class if the date is selected. You should include this in a new Core issue related to CalendarMonth a11y, though.

> I also notice that if I have the month dropdown open and press Esc, it closes the entire date picker, not just the menu.

This is very complicated since we attach global event handlers for escape in Popover and Select and neither knows about the other. I'll fix this pending the decision whether to open the date picker on click which is tied to this.

> In terms of the screen reader experience, I think it could be improved a bit.

All of your suggested changes require changing the Core HTML which should be done in https://github.com/patternfly/patternfly. As for this PR, I think having _some_ keyboard accessible screen reader is better than the none that existed before. If you'd like for me to disable all keyboard functionality to be exactly like `flatpickr` that's an option too.

redallen:
Re @mcarrano :

> I'm comparing against the current datapicker implementation with Flatpicker and there the focus remains in the field even when the calendar is opened. Can we make it work like that?

This is the same issue @nicolethoen brought up. I'll answer again: we can, but then what should happen as the user types a date? Should the Calendar flicker as they enter valid dates? I find it unexpected that the `flatpickr` implementation makes you press enter to see the typed date on the Calendar month _and the calendar still doesn't close_. However, we can still make it allow typing in the input after clicking one of two ways:
1. Use a button per the [W3 reference implementation](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html) which behaves by opening the Popover only when the icon to the right of the input is pressed. could toggle between being keyboard interactive and "just allowing typing." This requires Core changes.
2. Differentiate between keyboard and mouse users by having `onClick` not send focus to the calendar or trap focus but 'onKeyDown` do so for the enter key. This creates 2 new issues, though:
    - There's still the issue of what should happen to the CalendarMonth as the user types. Additionally, what should happen to the input if the user selects the month, year, or clicks on a day?
    - Usually these pickers are used in a form where pressing enter submits the form rather than opens a date picker. Do we really want to disallow pressing enter inside a date? Granted the current implementation I have does this too but I'd consider that a fixable bug.

mcarrano:
@redallen @mcoker @jessiehuff I've scheduled some time for tomorrow for us all to talk through this together. I get that there are not easy or straight forward answers, but I want to take the time to get it right. I'm looking at the implementation done in Carbon that behaves like I'd expect: https://react.carbondesignsystem.com/?path=/story/datepicker--single Perhaps we can model off that?

Regarding the confusion between the focus and hover states, I looked back at the prior implementation and we did have a different hover state as I'm showing here:

Screen Shot 2020-11-11 at 5 11 04 PM

@mcoker did we loose that in the CSS in this iteration?

cc @evwilkin

redallen:
Per our meeting, I've:

- Used a InputGroup with a button. The icon is now OutlinedCalendarAltIcon
- Made interacting with header tools not move focus to the table
- Made closing the Select with escape not close the Popover
- Added hidden screenreader text with full week day names
- Put unique aria-labels on the Select and Input for the month/year on CalendarMonth and made the forward/backward month buttons use `aria-labelledby`
- Bumped core to capture https://github.com/patternfly/patternfly/pull/3669 which:
    - Removes default browser focus outline
    - Fixes adjacent month selected styling

jessiehuff:
Just a few notes, but looking good so far! 🙂 
- I noticed the month input doesn't have an `aria-label` and the `aria-labelledby` seems to point at itself? 
- I think we have two potential options with the labeling: One would be the original direction I was thinking which would be to add ids to the month and year fields then add `aria-labelledby` to the date numbers themselves so let's say we were on the 3rd, it would have `aria-labelledby="{3's id} {month's id} {year's id}"`. However, there is a potential drawback to this approach in that it might announce "3, select month, select year" if the labels on month and year aren't their selected values. 
Or, a second approach could be to have 3 hidden spans that have the concatenated values of the month and year values that are possible for each viewpoint. (Ex: if you're viewing May 2020, your potential spans could be "April 2020", "May 2020", "June 2020" all referenced when on their dates like `aria-labelledby="{3's id} {relevant span's id}"`). I was talking to Jenn today, and she was thinking this might be the preferable option.

redallen:
Hey @jessiehuff ,
1. Easy to fix.
2. Can you please elaborate on the second approach? Where should the 3 hidden spans go and what should be put in each of them? This sounds like the approach that is the easiest to implement (feel free to ping me on Slack).
3. I added them because you said:
> I also think that it would be ideal to have the date buttons aria-labelledby the month and year so it would announce something like "4 March 2020".

but I now I see by "date buttons" you meant the table cells. I'll remove them from the previous/next month buttons.

jessiehuff (dismissed):
This looks awesome! Thanks Zack! 🙂 🎉

kmcfaul (dismissed):

nicolethoen (commented):
I vote that the calendar button toggles the calendar? not just opens the calendar?

redallen:
I pushed a fix for this @nicolethoen

mcarrano (commented):
I gave this another look. Mostly it looks great, there just seems to be one bug I encountered with focus when I move to a new month. If I am here:

Screen Shot 2020-11-13 at 5 48 14 PM

and then I click the arrow key to advance to the next month, I get this:

Screen Shot 2020-11-13 at 5 48 28 PM

The only other thing is a small nit. Can we make the placeholder "mm/dd/yyy" or "MM/DD/YYYY" and not "MM/dd/yyyy"?

mcarrano:
@redallen This looks good to me now. @jessiehuff can you take a look at the keyboard interaction one more time to make sure this is what you expect? Thanks.

jessiehuff (approved):
Looks great to me with keyboard and VO! 🙂

mcarrano (approved):
Looks good to me, also. Thanks @redallen !

evwilkin (approved):
LGTM

nicolethoen (approved):

Issue #5027, opened 2020-10-15
Provide custom Chip render props in the multi Typehead select.


anujsingla:
**Describe the issue. What is the expected and unexpected behavior?**
We are using multi Typehead select to select one or more items from a list. We need to enable/disable (remove-icon/Readonly) the chip as per condition.
Example: Suppose the user has selected 4 items from the list, then we need to disable 2 items and 2 enable. 
Currently, There are no options to pass custom chip group in the Typehead.
It would be good if we accept this feature.

Implementation as per my suggestion:
1. Accept 1 new props (customChipRender)
2. If user pass customChipRender prop then we render user chip otherwise default chip render.

https://www.patternfly.org/v4/components/select/#multiple

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
enhancement

**What is your product and what release version are you targeting?**

PR #5031, opened 2020-10-15
#5027 - custom Chip render props in the multi Typehead select .

anujsingla:
**What**: Closes # https://github.com/patternfly/patternfly-react/issues/5027

**Additional issues**: It is improvement and added new props.


redallen (changes requested):
If you could make the API take a React.ReactNode that'd match other component and would be more flexible!

anujsingla:
@redallen Thanks for the review, I have made changes as per your suggestion. Please review.

redallen (approved):
Thanks, LGTM!

nicolethoen (approved):
LGTM

PR #5032, opened 2020-10-15
fix(Skeleton): Add screen reader text to skeleton

### LINKED ISSUE #4997: Skeleton should have accessible text
When items are still loading, we provide the visual indicator to sighted users with the skeleton component. We should also provide some kind of text for screen reader users to let them know that a section is still loading or will ideally soon have content. The approach I'd suggest is the one that [OpenShift](https://github.com/openshift/console/pull/6743/files) has taken with their status section skeletons with screen reader only text. 

I imagine we could create a prop that would determine the text inside the span (that wouldn't be required). We could create a sensible default, but I am curious what others think about the scenario of using multiple skeleton components on a page. Perhaps you would only use the prop/screen reader text on the first one? I wouldn't want to overload screen reader users with redundant information like seeing "Loading, loading, loading" a bunch of times if there's multiple skeletons.

Open VO or another screen reader with the skeleton component. It will skip right over it.

Bug - it's currently not perceivable by screen reader

jessiehuff:
**What**: Closes #4997

### DIFF
--- a/packages/react-core/src/components/Skeleton/Skeleton.tsx
+++ b/packages/react-core/src/components/Skeleton/Skeleton.tsx
@@ -13,6 +13,8 @@ export interface SkeletonProps extends React.HTMLProps<HTMLDivElement> {
   fontSize?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
   /** The shape of the Skeleton */
   shape?: 'circle' | 'square';
+  /** Text read just to screen reader users */
+  screenreaderText?: string;
 }
 
 export const Skeleton: React.FunctionComponent<SkeletonProps> = ({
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/Skeleton/Skeleton.tsx
```diff
@@ -21,6 +23,7 @@ export const Skeleton: React.FunctionComponent<SkeletonProps> = ({
   height,
   fontSize,
   shape,
+  screenreaderText = 'Loading...',
```
redallen, line 26: I understand accessible defaults, but if this should NOT be the default whenever there exists more than 1 skeleton we might want to use a React context to conditionally add the text rather than ask the consumer to pass an empty string for every Skeleton after the first.

Is this your recommended usage?

redallen (changes requested):

packages/react-core/src/components/Skeleton/Skeleton.tsx
jessiehuff: Yeah, I agree with you, I was thinking along those lines myself. I think based on our conversation, it makes more sense to avoid the complexity and just remove the default, allowing consumers to intentionally add text instead.

redallen (approved):

redallen (approved):

kmcfaul (approved):

PR #5030, opened 2020-10-15
Table - expand/collapse all demo

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4943

https://patternfly-react-pr-5030.surge.sh/components/table/react-demos#expandcollapse-all

PR #5034, opened 2020-10-19
feat(DataList): add expand/collapse all

### LINKED ISSUE #4942: Add expand/collapse all capability to Data list component
Want to introduce ability to expand or collapse all rows in an expandable data list. This feature will be controlled from the toolbar as per the designs here: https://docs.google.com/document/d/1zEtmy0z4DfTDwZNei-Wgd-04ox8H3NQxA8z18-109QU/edit#. There should also be a demo to illustrate this functionality incorporating a toolbar as per the design.

kmcfaul:
**What**: Closes #4942

### DIFF
--- a/packages/react-core/src/components/DataList/examples/DataList.md
+++ b/packages/react-core/src/components/DataList/examples/DataList.md
@@ -33,6 +33,8 @@ KebabToggle,
 DropdownItem,
 } from '@patternfly/react-core';
 import CodeBranchIcon from '@patternfly/react-icons/dist/js/icons/code-branch-icon';
+import AngleDownIcon from '@patternfly/react-icons/dist/js/icons/angle-down-icon';
+import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
 import { css } from '@patternfly/react-styles';
 
 ## Examples
@@ -491,6 +493,8 @@ import {
   KebabToggle
 } from '@patternfly/react-core';
 import CodeBranchIcon from '@patternfly/react-icons/dist/js/icons/code-branch-icon';
+import AngleDownIcon from '@patternfly/react-icons/dist/js/icons/angle-down-icon';
+import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
 
 class ExpandableDataList extends React.Component {
   constructor(props) {
@@ -499,7 +503,27 @@ class ExpandableDataList extends React.Component {
       expanded: ['ex-toggle1', 'ex-toggle3'],
       isOpen1: false,
       isOpen2: false,
-      isOpen3: false
+      isOpen3: false,
+      allExpanded: false
+    };
+
+    this.onToggleAll = () => {
+      this.setState(
+        {
+          allExpanded: !this.state.allExpanded
+        },
+        () => {
+          if (this.state.allExpanded) {
+            this.setState({
+              expanded: ['ex-toggle1', 'ex-toggle2', 'ex-toggle3']
+            });
+          } else {
+            this.setState({
+              expanded: []
+            });
+          }
+        }
+      );
     };
 
     this.onToggle1 = isOpen1 => {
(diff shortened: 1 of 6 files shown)


redallen (approved):

mcoker:
@kmcfaul we just merged a PR that adds an expand all item of the toolbar that has a toolbar element for the arrow icon. The expand all item has an expanded class that rotates and animates the arrow rotation. Would you have time to update this PR with that? https://github.com/patternfly/patternfly/pull/3601

mcarrano (approved):
This looks great @kmcfaul !

kmcfaul:
@mcoker I should be able to, when the .59 core bump gets merged. Working on it now, adding the new variant and a small wrapper class for the `span` (since the button/tooltip itself gets passed to the toolbar item).

kmcfaul:
@mcoker Actually, what do you mean by the class controlling the animation? Right now, the example passes in either the angle right or angle down icon itself.

mcoker:
@kmcfaul this creates a new item of the toolbar (similar to the bulk-select item) that is `.pf-c-toolbar__item.pf-m-expand-all` with a plain button inside with an element `<span class="pf-c-toolbar__expand-all-icon">` that holds the right-arrow icon.

You can apply `.pf-m-expanded` to `.pf-c-toolbar__item.pf-m-expand-all` and that will rotate/animate the icon. The button should also have an `aria-label="Expand all"` when collapsed, and `"Collapse all"` when expanded.

kmcfaul:
I see. The design has a tooltip as well though, not just a plain button.

kmcfaul:
Maybe this would be better as a separate issue #, if we want to create a new controlled expand/collapse component for Toolbar? I don't modify anything in Toolbar right now.

mcoker:
@kmcfaul that's fine with me. There is also an expand all table demo that @jschuler worked on https://github.com/patternfly/patternfly-react/pull/5030. Both would get this update.

I guess my question is what the value of creating the demo now is assuming we're going to update it next sprint?

kmcfaul:
I'm fine with adding it now, I just can't decide whether to add just a wrapper class for the span only, or make it a whole component itself (since the ToolbarItem ends up controlling whether the icon looks expanded or collapsed).

Right now, I have it looking like this in the demo-side of the code:
```
           <ToolbarItem variant="expand-all" isExpanded={this.state.allExpanded}>
              <Tooltip
                position="right"
                content={
                  <div>
                    {this.state.allExpanded && 'Collapse all rows'}
                    {!this.state.allExpanded && 'Expand all rows'}
                  </div>
                }
              >
                <Button
                  onClick={this.onToggleAll}
                  variant="plain"
                  aria-label={this.state.allExpanded ? 'Collapse all rows' : 'Expand all rows'}
                >
                  <ToolbarExpandIconWrapper>
                    <AngleRightIcon />
                  </ToolbarExpandIconWrapper>
                </Button>
              </Tooltip>
            </ToolbarItem>
```
Assuming I'm correct that you only have to pass in the `angle-right` icon (which then gets transformed by the expanded modifier). 

But another way to add this would be to put the whole tooltip + button into its own component, or even make it a part of ToolbarItem. Something like:
```
<ToolbarItem variant='expand-all' isExpanded={this.state.allExpanded}>
  <ToolbarExpandAll onClick={this.onToggleAll} tooltipProps={...} buttonProps={...} />
</ToolbarItem>

kmcfaul:
If the top method is fine, I can add it now.

mcoker:
> Assuming I'm correct that you only have to pass in the angle-right icon (which then gets transformed by the expanded modifier).

Correct! I'm not sure about the way it's built though, I would probably vote for whatever's most consistent with the other items.

kmcfaul:
Yeah I was waiting for the merge, and testing it now it behaves correctly still. I'm not sure what's most consistent. 

The first may be more consistent with the other variants of ToolbarItem, but the control of the icon sitting in ToolbarItem instead of the child is what throws the structure into question for me.

@tlabaj Any opinions?

mcoker:
Can we also update the table expand all demo once this is complete? I can file a separate issue if we don't want to do it in this PR.

kmcfaul:
I've got the Table demo updated with the first implementation in my working branch, just waiting on if there are objections to the first implementation.

kmcfaul:
Titani approved the first method, so updated the PR with the Toolbar changes, and also updates to the Table demo!

nicolethoen (approved):

mcoker:
Can you wrap the toolbar children in `<ToolbarContent>` in both of these demos?

mcoker (approved):
LGTM! Thanks for the last minute updates!

nicolethoen (approved):

tlabaj (approved):
LGTM

dlabrecq (approved):

mcarrano (approved):
Looks great, Thanks @kmcfaul !

PR #5044, opened 2020-10-22
Update disabled dropdown items

### LINKED ISSUE #5039: Dropdown Accessibility: Disabled items in a dropdown are not announced by a screen reader
When validating https://github.com/patternfly/patternfly-react/issues/4946 with a customer who uses JAWS, we noticed a bug in the dropdown component. When asked what items were in the Basic dropdown on the examples page, he repeated all the items present in the dropdown, but the disabled items were not repeated. When asked if he expected the disabled items in the dropdown h said that he expects the disabled items to be announced as disabled and not omitted from the list. 

Must use JAWS to reproduce this issue.
Bug

jessiehuff:
**What**: Closes #5039

Updates to determine what will allow disabled dropdown items to be announced by JAWS

### DIFF
--- a/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
+++ b/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
@@ -168,10 +168,8 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
 
     if (component === 'a') {
       additionalProps['aria-disabled'] = isDisabled;
-      // additionalProps.tabIndex = isDisabled ? -1 : additionalProps.tabIndex;
     } else if (component === 'button') {
-      additionalProps['aria-disabled'] = isDisabled;       
-      // additionalProps.disabled = isDisabled;
+      additionalProps['aria-disabled'] = isDisabled;
       additionalProps.type = additionalProps.type || 'button';
     }
     const renderWithTooltip = (childNode: React.ReactNode) =>
--- a/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
+++ b/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import { DropdownContext } from './dropdownConstants';
 import { KEYHANDLER_DIRECTION } from '../../helpers/constants';
+import { preventedEvents } from '../../helpers/util';
 import { Tooltip } from '../Tooltip';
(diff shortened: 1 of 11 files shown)


redallen (changes requested):
I like this design and don't see any issues the code. All I ask is that the same change to be made to Select for consistency:
select-disabled-item

redallen:
Ah, I forgot about options menu as well:
options-menu-disabled-item

I think that's all of them, but you should probably double check if any of our toolbar components have the same interaction!

evwilkin (dismissed):
LGTM

tlabaj (approved):
@jessiehuff @redallen can we open a new issues to track the follow up work to make the same changes in other components.

mcarrano (commented):
@jessiehuff I'm good with this, but echo @redallen 's comments about duplicating this across other components that have dropdown menus. @jenny-s51 this should also apply to the new modular menu component you are working on.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -181,6 +184,15 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
       ) : (
         childNode
       );
+    const preventedEvents = inoperableEvents.reduce(
```
seanforyou23, line 187: Nice! Since it's essentially a same capability as we use in [Button](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Button/Button.tsx#L101), and may be needed on other components as well, can we maybe abstract this out into a utility function that can be imported/shared between?

seanforyou23 (commented):
Nice job! Had one question about reusing some of the code, other than that it looks good to me!

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
jessiehuff: Yup, sure can! 🙂

jessiehuff:
@mcarrano I completely agree. Actually just checking on that, it looks like because options menu and application launcher use dropdown, they also inherit this interaction pattern so they shouldn't need any additional changes (but would love everyone to take a look and make sure there aren't any issues in another of these components). Select is kind of its own thing so that will still need to be updated separately.

seanforyou23 (approved):
😎

redallen (approved):
LGTM! Select followup is #5095.

jcaianirh (approved):

tlabaj (approved):
LGTM

PR #5052, opened 2020-10-22
fix(datalist): respect isDisabled for dragging

### LINKED ISSUE #5046: DataListDragButton is missing props in the documentation

### LINKED ISSUE #5048: bug(DataListDragButton): isDisabled add `pf-m-disabled` class but one can still move around the row

When DataListDragButton has isDisabled set one should not be able to move the row around.

redallen:
**What**: Closes #5048. Closes #5046. Prevents moving the row if `isDisabled` is passed to the button, but other rows can be moved which in turn move the disabled row.

I'm not sure what intended behavior is and I think it was a mistake to expose `isDisabled` on `DataListDragButton` which really should be on `DataListItem`. `isDisabled` on a button doesn't quite capture that the entire **row** shouldn't be draggable nor that other rows shouldn't be able to occupy its index.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DataList/DataListItem.tsx
+++ b/packages/react-core/src/components/DataList/DataListItem.tsx
@@ -3,6 +3,7 @@ import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/DataList/data-list';
 import { DataListContext } from './DataList';
 import { KeyTypes } from '../Select';
+import { DataListDragButton, DataListDragButtonProps } from './DataListDragButton';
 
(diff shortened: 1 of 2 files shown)


KKoukiou:
@redallen thanks - that implements exactly the behavior that I need/would expect.

redallen:
If this works for you I think it's a fine interim solution.

dlabrecq (approved):

KKoukiou (approved):
Amazing thanks. It works exactly as expected. And I guess it does not worth any more to adjust the exposed properties, so that isDisabled get's exposed in that DataListItem element instead of the DataListDragButton, we can just always live with this WA I guess.

PR #5058, opened 2020-10-23
fix(table): set button type on compoundExpand transform button

seanforyou23:
This PR fixes an issue where when selecting a compound expandable table cell, when the table is also "editable", would cause a full page refresh instead of expanding the cell as expected. It's a patch for https://github.com/patternfly/patternfly-react/issues/5118

This is happening because when the table is "editable" the table is wrapped in a form, and the compoundExpand buttons do not have a type attribute set, and the default action of selecting a button when inside of a form is to attempt to submit the form.

Screen Shot 2020-10-23 at 6 04 30 PM

To reproduce the issue, add `onRowEdit={() => {}}` to a compound expandable table and then select one of the expandable cells. Notice it attempts to submit the form to the current url and a "?" query param token is added to the url.

The fix is simply to add `type="button"` to those expandable button elements so that they are treated as regular buttons and not "submit" buttons. I added a cypress test to verify this, but important to note that I didn't go as far as actually building a full editable-compound-expand table functionality in the integration app. I just added enough of a demo to verify the fix.

Hoping we can accept this incremental fix as a step toward the goal of further hardening the editable table feature and make it work fully with compound expand in a follow-up PR so we can promote editable table from beta to standard feature. Any hesitation or concern with this approach? I'm happy to try and build the full solution for the integration app in this PR, but might be nice to get this fix in place first just so the bug isn't causing issues for others who may want to press ahead using both of these complex feature together before we can get a full demo ready in the main react docs.

avoid editable table form submission on selecting compound expandable cells by setting btn type
add cypress test to verify selecting editable compound table cells does not cause page refresh
update test for header width table to use correct value

Screen Shot 2020-10-23 at 6 17 20 PM

### DIFF
--- a/packages/react-integration/cypress/integration/tableeditablecompoundexpandable.spec.ts
+++ b/packages/react-integration/cypress/integration/tableeditablecompoundexpandable.spec.ts
@@ -1,0 +1,20 @@
+describe('Table Compound Expandable Test', () => {
+  it('Navigate to demo section', () => {
+    cy.visit('http://localhost:3000/');
+    cy.get('#table-editable-compound-expandable-demo-nav-item-link').click();
+    cy.url().should('eq', 'http://localhost:3000/table-editable-compound-expandable-demo-nav-link');
+  });
+
+  it('Test expandable/collapsible', () => {
+    cy.get('button.pf-c-table__button')
+      .first()
+      .click();
+
+    cy.get('button.pf-c-table__button')
+      .first()
+      .click();
+
+    // should not have changed the url
+    cy.url().should('eq', 'http://localhost:3000/table-editable-compound-expandable-demo-nav-link');
+  });
+});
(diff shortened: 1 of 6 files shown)


jschuler (dismissed):
Hi @seanforyou23 ! LGTM

jessiehuff (approved):
Nice catch! LGTM! :)

tlabaj:
Hi @seanforyou23! Is there a related issue for this one?

seanforyou23:
Hi @tlabaj! No related issue I'm aware of. Ran into this when pair programming with someone who's trying to pioneer ahead with a compound expandable editable table - not even sure it's a feature PF wants to commit to (though I think it would be awesome) so I just posted this incremental fix for along the way. Happy to raise an issue if you like, and I started to do that initially but wasn't sure if the issue would make sense if not geared toward building the entire feature. Is it ok go raise an issue just for the unexpected role=submit for these expandable buttons when in edit mode? Just let me know what's the best path forward and I'll be happy to file whatever issues we need :)

tlabaj:
@seanforyou23 Yes, please open an issue to t rack this.  Thanks!

tlabaj (approved):
LGTM.  You just need to resolve the conflicts then we can merge it.

kmcfaul (dismissed):

jessiehuff (approved):

jschuler (approved):

PR #5061, opened 2020-10-26
fix(Wizard): Update progressive wizard example to improve keyboard a11y

### LINKED ISSUE #4445: Wizard with Progressive Steps example has inaccessible radio button
When using keyboard navigation with the wizard example "Progressive Steps", only the first radio button is focusable. The second radio button is clickable but not focusable or selectable by keyboard. This is likely just an example issue.

Screen Shot 2020-06-24 at 11 23 37 AM

Bug

jessiehuff:
**What**: Closes #4445

The radio button name attributes needed to be unique in order to get keyboard focus.

### DIFF
--- a/packages/react-core/src/components/Wizard/examples/Wizard.md
+++ b/packages/react-core/src/components/Wizard/examples/Wizard.md
@@ -497,7 +497,7 @@ class ProgressiveWizard extends React.Component {
             isChecked={getStartedStepRadio === 'Create'}
             onChange={(_, event) => this.setState({ getStartedStepRadio: event.currentTarget.value })}
             label="Create a new thing"
-            name="radio-step-start"
+            name="radio-step-start-1"
             id="radio-step-start-1"
           />{' '}
           <Radio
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Wizard/examples/Wizard.md
```diff
@@ -497,15 +497,15 @@ class ProgressiveWizard extends React.Component {
...
+            name="radio-step-start-1"
             id="radio-step-start-1"
           />{' '}
           <Radio
             value="Update"
             isChecked={getStartedStepRadio === 'Update'}
             onChange={(_, event) => this.setState({ getStartedStepRadio: event.currentTarget.value })}
             label="Update an existing thing"
-            name="radio-step-start"
+            name="radio-step-start-2"
```
seanforyou23, line 508: I think when you're defining a radio group the `name` attribute is what relates the inputs in that group together. The keyboard interaction on radio has a slightly different behavior where the first radio in the group will receive focus, and then you can use arrow keys to shift focus to the other radios. So this would be like setting two radio groups.

Agree the standard tab focus is nice, but I think it might actually be behaving as expected? The best way to see is check out the Form Controls rotor menu in VoiceOver, they'll both mention that the radios are "1 of 1" in the group after it announces the name. Hope it helps!

jessiehuff: Ah, thank you for this catch! 🙂

Issue #4810, opened 2020-09-11
Table: improve truncation of headers with a help popover


mcoker:
https://github.com/patternfly/patternfly/pull/3467

In any column header with a help tooltip/popover, we need to add `.pf-m-help` to the column header `td`/`th` element.

In any non-sortable column header with the help tooltip/popover, we need to wrap the text in `span.pf-c-table__text`.

You can see the structure for sortable and non-sortable headers with a help popover in this example - https://pf4.patternfly.org/documentation/core/components/table#sortable

cc @jschuler

PR #5075, opened 2020-10-28
table info transform

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/4810

Adds the `info` transform that should be used instead of the previous `header` object to add tooltip/popover, although that will still work.
The transform will make it easier in combination with sortable to combine the classes (`pf-m-help` and `pf-m-sortable`) on the `th` and compose the children together at the same time


jschuler:
@mcoker does the DOM look correct to you? https://patternfly-react-pr-5075.surge.sh/components/table#sortable

mcoker (approved):
LGTM! Thanks @jschuler

redallen (approved):

Issue #3953, opened 2020-03-23
Add keyboard interaction accessibility tests


jessiehuff:
**Describe the issue. What is the expected and unexpected behavior?**
We should write cypress keyboard interaction accessibility tests for our components in order to prevent future regressions, focus should be given to the most popular components: 
- [x] Button
- [ ] Modal
- [ ] Card
- [ ] Form
- [ ] Alert
- [ ] Empty State
- [ ] Tooltip 

(All dropdown related components captured in a separate issue: #3108)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
This would be an enhancement to cut down on manual testing time.

thatblindgeye:
Due to inactivity I'm going to close this out for now, but moved to Parking Lot to keep in mind that it's something to revisit when we have the bandwidth.

PR #5081, opened 2020-10-30
chore(modal): add/update test cypress tests for modal

seanforyou23:
**What**: Steps toward closing https://github.com/patternfly/patternfly-react/issues/3953

This PR adds adds a couple of new tests for modal which test to ensure focustrap is working as expected as well as that pressing escape key with an open modal would close it. I spent some time trying to also verify that pressing enter on the close button (or cancel/confirm) would also close the modal, but for some reason the trigger/type functions for cypress didn't actually close the modal. I tried several combinations of the following:

`....trigger('keydown', { keyCode: 13, which: 13 });`
`....type('{enter}')`

Any idea why these won't simulate the keypress event? The functionality does obviously work, just that cypress can't seem to invoke it. I could cheap out and go for `click()` which works fine, but the main point of the issue is to test keyboard interactivity so not sure that would fully exercise the right path in the code.

In any case, what's here should help prevent regressions to keyboard interactivity in the future, would be cool to get it merged and circle back to the pressing enter issue if possible.

I also fixed a small issue where one of the examples in the test suite was missing an accessible name.

Screen Shot 2020-10-29 at 9 28 59 PM

### DIFF
--- a/packages/react-integration/cypress/integration/modal.spec.ts
+++ b/packages/react-integration/cypress/integration/modal.spec.ts
@@ -208,4 +208,41 @@ describe('Modal Test', () => {
         });
     });
   });
+
+  it('Verify focustrap for basic modal', () => {
+    cy.get('#tabstop-test').focus();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    cy.tab().click(); // click first btn to open first modal
+    cy.focused().should('have.attr', 'aria-label', 'Close');
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    cy.tab();
+    cy.focused().should('have.attr', 'data-id', 'modal-01-cancel-btn');
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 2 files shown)


jeff-phillips-18 (approved):
We should file an issue to circle back to attempt to get the return key to work.

jessiehuff (approved):
I took a look myself and attempted a few things, but I'm seeing the same thing as you. I noticed that [this issue](https://github.com/cypress-io/cypress/issues/8267) exists in the cypress repo - I wonder if they're related and if we can add to that conversation.

seanforyou23:
Yea that seems like a relevant issue to chime in on. I tried chaining the .type() and .trigger() methods off of .get(), .focused() and .tab() methods and they all had the same behavior. I'll tweak this PR a bit so there's a few lines in the cypress test that can be toggled on/off to compare how it works with click but not the keyboard events. Would that be enough to get these keyboard interaction related PRs merged, or should we leave them open? I can add the relevant info and instruction on how to reproduce in #5088 if you think that's enough - then reference that and this PR on a comment to that cypress thread. Just let me know what you prefer!

tlabaj (approved):

PR #5083, opened 2020-10-30
chore(form): add more tests around keyboard interactivity in forms

seanforyou23:
**What**: Steps toward closing https://github.com/patternfly/patternfly-react/issues/3953

This PR adds more cypress tests to the Forms demo to verify basic keyboard interactions are in place. I did notice that cypress has a hard time working with our Form controls. Things like the cypress tab plugin do not successfully tab around the typeahead multiselect (and forms in general) and using trigger/type functions to press enter were also inconsistent. Despite this, I was able to add some useful tests.

For checkbox, I wanted to verify that pressing the spacebar would check/uncheck the box but support for spacebar seems incomplete in cypress. The functionality does of course work in our components, just that after spending a couple hours trying to get cypress to verify this, I had to call a stopping point. Any help here would be great, otherwise we may be better served to raise an issue in the cypress repo.

### DIFF
--- a/packages/react-core/src/components/Divider/examples/Divider.md
+++ b/packages/react-core/src/components/Divider/examples/Divider.md
@@ -11,7 +11,7 @@ propComponents: ['Divider']
 import React from 'react';
 import { Divider } from '@patternfly/react-core';
 
-<Divider/>
+<Divider />
 ```
 
 ### Using li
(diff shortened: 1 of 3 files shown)


jessiehuff (approved):
I agree. I pulled down your code and was trying it myself, and it's so bizarre that the test works the first time checking that it becomes checked but not unchecked. I think this does verify that the keyboard functionality works though, and when submitting an issue to cypress I can see it making sense to keep this test to show where it doesn't work.
LGTM!

redallen (approved):
Cypress' lack of keyboard support is astonishing.

tlabaj (approved):
LGTM

PR #5089, opened 2020-10-30
Feat/jump link

### LINKED ISSUE #5022: Introduce Jump links component
Introduce a new Jump links component that allows users to navigate content within a long scrolling page. This includes both vertical and horizontal versions. Scrollspy behavior can be enabled/disabled via a prop. As the user scrolls manually, the Scrollspy component should reflect current position in the page.

Core issue: https://github.com/patternfly/patternfly/issues/3558
Designs: https://marvelapp.com/prototype/25f068g6

Component should be coupled with a demo that shows behavior. Clicking any link in the navigation menu will scroll the page to a corresponding anchor link. As the user scrolls manually, the Scrollspy component should reflect current position in the page.

@jessiehuff can you think of any special keyboard interaction considerations here?

redallen:
**What**: Closes #5022

**Additional issues**: I made up the demo, cc @mcarrano for review.

### DIFF
--- a/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
+++ b/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
@@ -1,0 +1,38 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/JumpLinks/jump-links';
+
+export interface JumpLinksProps {
+  'aria-label'?: string;
+  isCentered?: boolean;
+  isVertical?: boolean;
+  children?: React.ReactNode;
+  label?: React.ReactNode;
+}
+
+export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
+  isCentered,
+  isVertical,
+  children,
+  label,
+  'aria-label': ariaLabel = typeof label === 'string' ? label : null
+}: JumpLinksProps) => {
+  const ul = <ul className={styles.jumpLinksList}>{children}</ul>;
+
+  return (
+    <nav
+      className={css(styles.jumpLinks, isCentered && styles.modifiers.center, isVertical && styles.modifiers.vertical)}
+      aria-label={ariaLabel}
+    >
+      {label ? (
+        <div className={styles.jumpLinksMain}>
+          <div className={styles.jumpLinksLabel}>{label}</div>
+          {ul}
+        </div>
+      ) : (
+        ul
+      )}
+    </nav>
+  );
+};
+JumpLinks.displayName = 'JumpLinks';
--- a/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
+++ b/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/JumpLinks/jump-links';
-import { JumpLinksItem } from './JumpLinksItem';
+import { JumpLinksItem, JumpLinksItemProps } from './JumpLinksItem';
 
 export interface JumpLinksProps {
   'aria-label'?: string;
@@ -9,16 +9,11 @@ export interface JumpLinksProps {
   isVertical?: boolean;
   children?: React.ReactNode;
   label?: React.ReactNode;
+  scrollableSelector?: string;
 }
 
-export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
-  isCentered,
-  isVertical,
-  children,
-  label,
-  'aria-label': ariaLabel = typeof label === 'string' ? label : null
-}: JumpLinksProps) => {
-  const scrollItems = React.Children.toArray(children).map((child: any) => {
+const getScrollItems = (children: React.ReactNode) =>
+  React.Children.toArray(children).map((child: any) => {
     if (child.type === JumpLinksItem) {
       const scrollNode = child.props.node || child.props.href;
       if (typeof scrollNode === 'string' && typeof document !== 'undefined') {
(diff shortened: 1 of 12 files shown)


mcarrano:
@redallen This is a good start. I just have a couple of comments/questions:

* While I understand that the examples can't really do much outside of the context of a demo, I wonder whether we can at least make these a bit more interactive like the Tab examples so I can at least see a new item activated when I click on it. @tlabaj any suggestions here?

* In the demo, seems like we have situations where two items and selected/active because multiple H2's are in view. This is not what I expect from scrollspy. Checked against the website and it only recognized the first matching header for example.

* Regarding the demo, would be good to see a vertical and well as horizontal example. @christiemolloy is working on this issue to build out some full page examples/demos: https://github.com/patternfly/patternfly/issues/3560  Perhaps you guys can sync on this as I will want those demos to ultimately be implemented in React.

redallen:
@mcarrano 
1. I'd have to implement the click handlers in each example's code just for most consumers to remove because they want scroll spy behavior like the demo, not tab behavior like the example.
2. I know the bug you're talking about and I believe it's due to CSS. I'll sync with HTML.
3. For a vertical demo we'll have to invent a new horizontal page layout (or do something strange like use the side nav or a drawer). If there's a design I'm happy to implement it.

jessiehuff (commented):
Can we add keyboard interactivity? I would expect that if I was using a keyboard, and I pressed enter on the jump link, that the next focusable item would be from the point I'm jumping to. Perhaps we could add links/buttons inside of the example heading content sections so that when the user selects "Heading 3", for example, the next focusable item would be the link or button inside heading 3. Without that interaction pattern, adding a lot of jump links would be really burdensome to keyboard users if they want to interact with the page (they'd have to tab through however many links just to get to the main content).

redallen:
@jessiehuff I'd love to be able to rely on that default browser behavior (example: [Wikipedia](https://en.wikipedia.org/wiki/List_of_hunting_deities#Filipino_mythology)), but there's a few things working against that:
1. SPAs render using JS. While JS is loading Chrome eats jump link events and never executes them.
2. I'd like to support Jump links to ANY element, not just elements with `id`s.
3. Our current doc site router `@reach/router` rerenders `<Page>` when the hash changes so no `<h2 id="jump-to-me" />` exists when you click the jump link until React finishes rerendering. Even nicer browsers like FF will inconsistently jump.
4. Many consumers use a hash router, so `/page#heading-1` will actually navigate to a page called `heading-1`!

This all means I have to scroll the page myself using `element.scrollIntoView()` and leave `location.hash` alone. But then what item would you like to be focused? Our [Title](https://www.patternfly.org/v4/components/title/react/sizes/) component (which is what the Jump link is jumping to) is not focusable. There is no browser API I know of to set the next tab element in line to be the next DOM node in the tree after another specified element. If you know of one, I'm all ears.

We could create an official [`AutoLinkHeader`](https://github.com/patternfly/patternfly-org/blob/master/packages/theme-patternfly-org/components/autoLinkHeader/autoLinkHeader.js) component that has a focusable icon and focus the icon to emulate the default browser behavior. However @mcarrano would have to weigh in with a design.

I'll add a skip to content in the meantime.

jessiehuff (dismissed):
LGTM! 🙂

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
```diff
@@ -0,0 +1,134 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/JumpLinks/jump-links';
+import { JumpLinksItem, JumpLinksItemProps } from './JumpLinksItem';
+
+export interface JumpLinksProps extends Omit<React.HTMLProps<HTMLElement>, 'label'> {
+  /** Whether to center children. */
+  isCentered?: boolean;
+  /** Variant to use. */
```
tlabaj, line 9: this description does no match the prop.

packages/react-core/src/components/JumpLinks/examples/JumpLinks.md
```diff
@@ -0,0 +1,74 @@
...
+beta: true
+---
+
+## Examples
+### Basic
+```js
+import React from 'react';
+import { JumpLinks, JumpLinksItem } from '@patternfly/react-core';
+
+<JumpLinks>
```
tlabaj, line 14: I know we have the demos, but I wonder if we should also make the examples interactive so clicking a different link would make it the active link.

tlabaj (changes requested):
Could you also update the demo-app and add integration tests.

packages/react-core/src/components/JumpLinks/examples/JumpLinks.md
redallen: Again, I'd have to implement the click handlers in each example's code just for most consumers to remove because they want scroll spy behavior like the demo, not tab behavior like the examples.

tlabaj: @redallen I see what you are say.  I wonder if we should add a blurb in the example to point the consumer to the demo to see the interactivity.
@mcarrano what do you think?

mcarrano: Yes. I agree.

redallen: OK, I'll add a blurb soon!

redallen: I added this to the top:
> These examples are static because they have no element to scroll spy on that makes sense. Check out the [React demos](./react-demos) to see scroll spying in action!

tlabaj (dismissed):
LGTM

mcoker (approved):
lgtm!

mcarrano (approved):
Looks good. Thanks @redallen !

packages/react-core/src/components/JumpLinks/JumpLinksItem.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/JumpLinks/jump-links';
+
+export interface JumpLinksItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'onClick'> {
+  /** Whether this item is active. Parent JumpLinks component sets this when passed a `scrollableSelector`. */
+  isActive?: boolean;
+  /** Href for this ling */
```
evwilkin, line 8:
```suggestion
  /** Href for this link */
```

evwilkin:
@redallen CodeSandbox example isn't working, this may not be blocking but wanted to call this out

evwilkin (approved):
LGTM

redallen:
Codesandbox never works for new components in PRs because they haven't been released to NPM yet.

tlabaj (approved):
LGTM

PR #5093, opened 2020-11-02
feat(Card): add expandable card, move tests

### LINKED ISSUE #5018: Expandable card
Introduce a new expandable card variant. This will allow a card header to expand/collapse

Core issue: https://github.com/patternfly/patternfly/issues/3554
Designs: https://marvelapp.com/prototype/5e55293

kmcfaul:
**What**: Closes #5018

Adds the ability to make a Card expand/collapse.

This is done by adding `isExpanded` to the `Card` and `onExpand` to the `CardHeader`. Hidden content is put within `CardExpandableContent` and may be either hidden or not rendered unless the Card is expanded.

`onExpand` will return the `Card`'s ID by default, allowing a user to make a list of expanded IDs if many cards exist.


mcarrano (commented):
Looks good to me. @gdoyle can you also take a look at this to see if this is what you were expecting for expandable cards in OpenShift?

mcoker (commented):
LGTM, though I have a couple of questions about the a11y @jessiehuff 

Looks like we're missing `aria-expanded` in the core example and here - just want to verify we should add that?

In core we have the button's `aria-label` as "Details", used in `aria-labelledby` with the card title ID, so that the button reads "{card title} details". In this PR, the button has `aria-label="Toggle button"`. Should either of these be updated?

jessiehuff (commented):
Nice catch @mcoker! Yeah, I would recommend adding `aria-expanded="true"` when the card is expanded and "false" when it's collapsed similar to other[ expandable](https://www.patternfly.org/v4/components/expandable-section) components we have. Also, I think the core labeling makes more sense here. I tend to avoid labeling controls with the element type because it tends to make it repeat the type. So for example, if we label it "Toggle button", in VO it will announce "Toggle button, button".

kmcfaul:
Rebased and updated aria labels + aria-expanded

evwilkin (approved):
LGTM

packages/react-core/src/components/Card/CardHeader.tsx
```diff
@@ -1,21 +1,53 @@
...
   /** Content rendered inside the CardHeader */
   children?: React.ReactNode;
   /** Additional classes added to the CardHeader */
   className?: string;
+  /** ID of the card header */
+  id?: string;
+  /** Callback expandable card */
+  onExpand?: (cardId: string) => void;
+  /** Additional props for expandable toggle button */
+  toggleButtonProps?: any;
```
tlabaj, line 18: @jessiehuff Do we need an explicit aria-label prop for the toggle button?

tlabaj (changes requested):
The interface seems a little off to me.  It looks like the consumer is responsible for rendering the expandable section.  Is there a reason the component can't do that internally?

kmcfaul:
@tlabaj Hmm since we have a provider I could pass the isExpanded prop down to the `CardExpandableContent` and apply a `hidden` property. Is that more in line with what you're thinking?

packages/react-core/src/components/Card/CardHeader.tsx
jessiehuff: I think following core's lead here would make sense. In core we have the button's `aria-label` as "Details", used in `aria-labelledby` with the card title ID, so that the button reads "{card title} details".

mcoker:
`aria-expanded` looks good, but I'm seeing the toggle button's `aria-labelledby` set to the id of the card, so VO is reading it as

Screen Shot 2020-11-12 at 12 58 13 PM

kmcfaul:
First expandable example should be good now. @jessiehuff For the second example, there is not title/header to put an ID on for `aria-labelledby`, should I put an ID on the `img` itself or should `aria-labelledby` just contain the button ID in this case?

jessiehuff:
That's a good question. I think we could try adding the id to the icon and using `aria-labelledby`, but I'm not sure whether that would work or what screen reader support there is for something like that between different screen readers. So if that doesn't work, we might just want to use an `aria-label="PatternFly Details"` instead.

kmcfaul:
To be safe I'll update the aria-label. The id+aria-labelledby could be tested by editing the app code in the example, since it is all passed in.

jessiehuff (approved):
LGTM! 🙂

packages/react-core/src/components/Card/CardHeader.tsx
```diff
@@ -1,21 +1,53 @@
...
 
 export interface CardHeaderProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside the CardHeader */
   children?: React.ReactNode;
   /** Additional classes added to the CardHeader */
   className?: string;
+  /** ID of the card header */
+  id?: string;
+  /** Callback expandable card */
+  onExpand?: (cardId: string) => void;
```
jschuler, line 16:
```suggestion
  onExpand?: (id: string) => void;
```

packages/react-core/src/components/Card/CardHeader.tsx
```diff
@@ -1,21 +1,53 @@
...
+  /** Additional props for expandable toggle button */
+  toggleButtonProps?: any;
 }
 
 export const CardHeader: React.FunctionComponent<CardHeaderProps> = ({
   children = null,
   className = '',
+  id = '',
```
jschuler, line 24: maybe better if this was undefined otherwise it ends up in the DOM (with no value)

packages/react-core/src/components/Card/CardHeader.tsx
```diff
@@ -1,21 +1,53 @@
...
+  <CardContext.Consumer>
+    {({ cardId }) => (
+      <div className={css(styles.cardHeader, className)} id={id} {...props}>
+        {onExpand && (
+          <div className={css(styles.cardHeaderToggle)}>
+            <Button
+              variant="plain"
+              type="button"
+              onClick={() => {
+                onExpand(cardId);
```
jschuler, line 38: should pass the event along as well, and I think typically event should be the first parameter passed back?

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -6,6 +6,8 @@ import { useOUIAProps, OUIAProps } from '../../helpers';
 export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   /** Content rendered inside the Card */
   children?: React.ReactNode;
+  /** ID of the Card */
```
jschuler, line 9:
```suggestion
  /** ID of the Card. Also passed back in the CardHeader onExpand callback. */
```

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -20,40 +22,63 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
...
+}
+
+export const CardContext = React.createContext<Partial<CardContextProps>>({
+  cardId: '',
+  isExpanded: false
+});
+
 export const Card: React.FunctionComponent<CardProps> = ({
   children = null,
+  id = '',
```
jschuler, line 41:
```suggestion
  id,
```
just so empty div doesn't get rendered

jschuler (approved):

nicolethoen (approved):

redallen (approved):
Works well!

tlabaj (approved):

dlabrecq (commented):
LGTM

PR #5097, opened 2020-11-03
fix(table): change internal column ids to uppercase

### LINKED ISSUE #4835: Table with column 'ID' breaks React keys

Expected behavior: no errors
Actual behavior: Encountered two children with the same key, `[object Object]-row`.

https://codesandbox.io/s/youthful-noyce-s1bwg?file=/index.js

Bug. Don't know what's the workaround apart from renaming the column which one shouldn't have to do.

michpetrov:
**What**: Closes #4835 

Assuming the column data have to be properties of `rowData` (instead of having their own bucket) switching to uppercase solves the naming collision. Also has the added benefit of separating the column data from the rest of the properties.

### DIFF
--- a/packages/react-table/src/components/Table/utils/headerUtils.tsx
+++ b/packages/react-table/src/components/Table/utils/headerUtils.tsx
@@ -109,7 +109,7 @@ const mapHeader = (column: ICell, extra: any, key: number, ...props: any) => {
     property:
       (typeof title === 'string' &&
         title
-          .toLowerCase()
+          .toUpperCase()
           .trim()
           .replace(/\s/g, '-')) ||
       `column-${key}`,
--- a/packages/react-table/src/components/Table/utils/headerUtils.test.tsx
+++ b/packages/react-table/src/components/Table/utils/headerUtils.test.tsx
@@ -58,7 +58,7 @@ describe('headerUtils', () => {
       const mixed = calculateColumns(cells, {});
       cells.forEach((oneCell: ICell, key) => {
         test(`${oneCell}`, () => {
-          expect(mixed[key].property).toBe(oneCell.title || oneCell);
+          expect(mixed[key].property).toBe((oneCell.title || oneCell).toUpperCase());
           expect(mixed[key].header.label).toBe(oneCell.title || oneCell);
           expect(mixed[key].header.transforms).toHaveLength(2);
           expect(mixed[key].cell.transforms).toHaveLength(1);
@@ -68,8 +68,8 @@ describe('headerUtils', () => {
     });
 
     test('correct property', () => {
-      expect(calculateColumns(['some'], {})[0].property).toBe('some');
-      expect(calculateColumns(['some with space'], {})[0].property).toBe('some-with-space');
+      expect(calculateColumns(['some'], {})[0].property).toBe('SOME');
+      expect(calculateColumns(['some with space'], {})[0].property).toBe('SOME-WITH-SPACE');
       expect(calculateColumns([''], {})[0].property).toBe('column-0');
     });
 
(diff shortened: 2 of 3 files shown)


packages/react-table/src/components/Table/utils/headerUtils.tsx
```diff
@@ -109,7 +109,7 @@ const mapHeader = (column: ICell, extra: any, key: number, ...props: any) => {
     property:
       (typeof title === 'string' &&
         title
-          .toLowerCase()
+          .toUpperCase()
           .trim()
           .replace(/\s/g, '-')) ||
```
redallen, line 114: I wonder if we could employ a hash function?
```js
/**
 * Calculate a 32 bit FNV-1a hash
 * Found here: https://gist.github.com/vaiorabbit/5657561
 * Ref.: http://isthe.com/chongo/tech/comp/fnv/
 *
 * @param {string} str the input value
 * @param {boolean} [asString=false] set to true to return the hash value as 
 *     8-digit hex string instead of an integer
 * @param {integer} [seed] optionally pass the hash of the previous chunk
 * @returns {integer | string}
 */
function hashFnv32a(str, asString, seed) {
    /*jshint bitwise:false */
    var i, l,
        hval = (seed === undefined) ? 0x811c9dc5 : seed;

    for (i = 0, l = str.length; i < l; i++) {
        hval ^= str.charCodeAt(i);
        hval += (hval << 1) + (hval << 4) + (hval << 7) + (hval << 8) + (hval << 24);
    }
    if( asString ){
        // Convert to 8 digit hex string
        return ("0000000" + (hval >>> 0).toString(16)).substr(-8);
    }
    return hval >>> 0;
}
```

redallen (changes requested):
Inline edit extension's  jest tests are failing, but feel free to skip its failing tests by using `xit` instead of `it` in this PR. We've already deprecated inline edit extension in favor of [editable rows.](http://patternfly-react.surge.sh/components/table#editable-rows)

packages/react-table/src/components/Table/utils/headerUtils.tsx
michpetrov: Tbh, I feel like that's unnecessary obfuscation (from a debugging standpoint the rowData object becomes unreadable), the problem is with `id` and `cells` since those are single words and thus not in camelCase like the rest. If we make sure the column names are never lowercase/camelCase then we're fine.

michpetrov:
@redallen I'm not that familiar with testing, how would I do that?

packages/react-table/src/components/Table/utils/headerUtils.tsx
redallen: If you'd like it to be debuggable just append the hash to column name. Unfortunately, we can't "make sure the column names are never lowercase/camelCase" and it's likely that there are many instances of this out in the wild.

redallen:
In `packages/react-inline-edit-extension/src/components/InlineEdit/InlineEdit.api.test.tsx` you can change `test('should call correct function', ...)` to `xtest('should call correct function', ...)`.

packages/react-table/src/components/Table/utils/headerUtils.tsx
michpetrov: This is happening because of variables we're inserting into the rowData object, we have full control over that, don't we?

redallen: I don't know. @jschuler do you know?

jschuler:
Thanks @michpetrov ! we're closing this PR in favor of https://github.com/patternfly/patternfly-react/pull/5152

PR #5107, opened 2020-11-08
feat(textarea): support isDisabled

### LINKED ISSUE #5068: TextArea isDisabled property is called readOnly?
It's more of a consistency issue really, all of the components that support being "disabled" have a property called `isDisabled` except TextArea where it's called `readOnly`. Wouldn't it be better to call it `isDisabled` there as well?

boaz0:
**What**: Closes #5068 

It is a breaking change. If someone is using `disabled` in the props then after updating with this change Typescript will show an error message saying `disabled` is not a valid props of `TextArea`.

### DIFF
--- a/packages/react-core/src/components/TextArea/TextArea.tsx
+++ b/packages/react-core/src/components/TextArea/TextArea.tsx
@@ -15,6 +15,10 @@ export interface TextAreaProps extends Omit<HTMLProps<HTMLTextAreaElement>, 'onC
   className?: string;
   /** Flag to show if the TextArea is required. */
   isRequired?: boolean;
+  /** Flag to show if the TextArea is disabled. */
+  isDisabled?: boolean;
+  /** Flag to show if the TextArea is read only. */
+  isReadOnly?: boolean;
   /** Value to indicate if the textarea is modified to show that validation state.
    * If set to success, textarea will be modified to indicate valid state.
    * If set to error, textarea will be modified to indicate error state.
@@ -38,6 +42,7 @@ export class TextAreaBase extends React.Component<TextAreaProps> {
     innerRef: React.createRef<HTMLTextAreaElement>(),
     className: '',
     isRequired: false,
+    isDisabled: false,
     validated: 'default',
     resizeOrientation: 'both',
     'aria-label': null as string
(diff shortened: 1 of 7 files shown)


redallen (approved):

packages/react-core/src/components/TextArea/TextArea.tsx
```diff
@@ -10,11 +10,13 @@ export enum TextAreResizeOrientation {
   both = 'both'
 }
 
-export interface TextAreaProps extends Omit<HTMLProps<HTMLTextAreaElement>, 'onChange' | 'ref'> {
+export interface TextAreaProps extends Omit<HTMLProps<HTMLTextAreaElement>, 'disabled' | 'onChange' | 'ref'> {
```
redallen, line 13: I'm in favor of this and I think it warrants finally having a "breaking type change" discussion I've been postponing.

tlabaj (changes requested):
We should also add the `isReadOnly` prop.  That is what the consumer seems to be looking for.  Please see my comment in the issue.

kmcfaul (dismissed):

tlabaj (commented):
Can you also update the demo app please.

boaz0:
@tlabaj done! :+1:

kmcfaul (dismissed):

packages/react-integration/demo-app-ts/src/components/demos/TextAreaDemo/TextAreaDemo.tsx
```diff
@@ -124,6 +124,12 @@ export class TextAreaDemo extends React.Component<{}, TextAreaState> {
           validated={validated}
           aria-label="text area example 5"
         />
+        <Text>Disabled text area </Text>
```
tlabaj, line 127: can you update the cypress test to verify expected behaviour.

boaz0:
Hi @tlabaj, I added tests to cypress as you requested. Sorry for the late response :bow:

tlabaj (approved):

kmcfaul (approved):

redallen (approved):

boaz0:
Thank you all!

PR #5115, opened 2020-11-10
feat(Touchspin): add touchspin component

### LINKED ISSUE #5021: Touchspin component
Introduce a Touchspin component that supports numeric data entry.

Core issue and designs: https://github.com/patternfly/patternfly/issues/362

kmcfaul:
**What**: Closes #5021

### DIFF
--- a/packages/react-core/src/components/Touchspin/Touchspin.tsx
+++ b/packages/react-core/src/components/Touchspin/Touchspin.tsx
@@ -8,7 +8,7 @@ import { Button, ButtonProps } from '../Button';
 export interface TouchspinProps extends React.HTMLProps<HTMLDivElement> {
   /** Value of the touchspin */
   value?: number;
-  /** Additional classes added to the Skeleton */
+  /** Additional classes added to the touchspin */
   className?: string;
   /** Sets the width of the touchspin to a number of characters */
   widthChars?: number;
@@ -50,7 +50,6 @@ export const Touchspin: React.FunctionComponent<TouchspinProps> = ({
   const touchspinUnit = <div className={css(styles.touchspinUnit)}>{unit}</div>;
   return (
     <div
-      {...props}
       className={css(styles.touchspin, className)}
       {...(widthChars && {
         style: {
@@ -58,6 +57,7 @@ export const Touchspin: React.FunctionComponent<TouchspinProps> = ({
           ...props.style
         } as React.CSSProperties
       })}
+      {...props}
     >
       {unit && unitPosition === 'before' && touchspinUnit}
       <div className={css(styles.inputGroup)}>
@@ -66,6 +66,7 @@ export const Touchspin: React.FunctionComponent<TouchspinProps> = ({
           {...minusButtonProps}
           {...(onMinus && { onClick: evt => onMinus(inputProps.name, evt) })}
           {...(isDisabled && { isDisabled })}
+          {...(!(minusButtonProps && minusButtonProps['aria-label']) && { 'aria-label': 'Minus' })}
         >
           <span className={css(styles.touchspinIcon)}>
             <MinusIcon aria-hidden="true" />
@@ -85,6 +86,7 @@ export const Touchspin: React.FunctionComponent<TouchspinProps> = ({
           {...plusButtonProps}
           {...(onPlus && { onClick: evt => onPlus(inputProps.name, evt) })}
           {...(isDisabled && { isDisabled })}
+          {...(!(plusButtonProps && plusButtonProps['aria-label']) && { 'aria-label': 'Plus' })}
         >
           <span className={css(styles.touchspinIcon)}>
             <PlusIcon aria-hidden="true" />
--- a/packages/react-core/src/components/Touchspin/Touchspin.tsx
+++ b/packages/react-core/src/components/Touchspin/Touchspin.tsx
@@ -15,21 +15,33 @@ export interface TouchspinProps extends React.HTMLProps<HTMLDivElement> {
   /** Indicates the whole touchspin should be disabled */
   isDisabled?: boolean;
   /** Callback for the minus button */
-  onMinus?: (name: string, event: React.MouseEvent) => void;
+  onMinus?: (event: React.MouseEvent, name?: string) => void;
   /** Callback for the text input changing */
   onChange?: (event: React.FormEvent<HTMLInputElement>) => void;
   /** Callback for the plus button */
-  onPlus?: (name: string, event: React.MouseEvent) => void;
+  onPlus?: (event: React.MouseEvent, name?: string) => void;
   /** Adds the given touchspin unit to the touchspin */
   unit?: React.ReactNode;
   /** Position of the touchspin unit in relation to the touchspin */
   unitPosition?: 'before' | 'after';
+  /** Minimum value of the touchspin, disabling the minus button when reached */
+  min?: number;
+  /** Maximum value of the touchspin, disabling the plus button when reached */
+  max?: number;
+  /** Name of the input */
+  inputName?: string;
+  /** Aria label of the input */
+  inputAriaLabel?: string;
+  /** Aria label of the minus button */
+  minusBtnAriaLabel?: string;
+  /** Aria label of the plus button */
+  plusBtnAriaLabel?: string;
   /** Additional properties added to the touchspin text input */
   inputProps?: any;
   /** Additional properties added to the minus button */
-  minusButtonProps?: ButtonProps;
+  minusBtnProps?: ButtonProps;
   /** Additional properties added to the plus button */
-  plusButtonProps?: ButtonProps;
+  plusBtnProps?: ButtonProps;
 }
 
 export const Touchspin: React.FunctionComponent<TouchspinProps> = ({
@@ -42,9 +54,15 @@ export const Touchspin: React.FunctionComponent<TouchspinProps> = ({
   onPlus,
   unit,
   unitPosition = 'after',
+  min,
+  max,
+  inputName,
+  inputAriaLabel = 'Input',
+  minusBtnAriaLabel = 'Minus',
+  plusBtnAriaLabel = 'Plus',
   inputProps,
-  minusButtonProps,
-  plusButtonProps,
+  minusBtnProps,
+  plusBtnProps,
   ...props
 }: TouchspinProps) => {
   const touchspinUnit = <div className={css(styles.touchspinUnit)}>{unit}</div>;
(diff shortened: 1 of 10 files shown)


mcarrano (approved):
Looks good @kmcfaul . Thanks!

packages/react-core/src/components/Touchspin/Touchspin.tsx
```diff
@@ -0,0 +1,98 @@
...
+import styles from '@patternfly/react-styles/css/components/Touchspin/touchspin';
+import { css } from '@patternfly/react-styles';
+import MinusIcon from '@patternfly/react-icons/dist/js/icons/minus-icon';
+import PlusIcon from '@patternfly/react-icons/dist/js/icons/plus-icon';
+import { Button, ButtonProps } from '../Button';
+
+export interface TouchspinProps extends React.HTMLProps<HTMLDivElement> {
+  /** Value of the touchspin */
+  value?: number;
+  /** Additional classes added to the Skeleton */
```
tlabaj, line 11: Should be touchspin not skeleton

packages/react-core/src/components/Touchspin/Touchspin.tsx
```diff
@@ -0,0 +1,98 @@
...
+  className,
+  widthChars,
+  isDisabled = false,
+  onMinus,
+  onChange,
+  onPlus,
+  unit,
+  unitPosition = 'after',
+  inputProps,
+  minusButtonProps,
```
tlabaj, line 46: I am not sure about the aria-label being part of this prop.  @jessiehuff shouldn't we default the aria labels for the buttons to a useful string like we do elsewhere for some buttons (e.g. a close button)?
@kmcfaul are there other props that might be passed to the plus or minus buttons?

packages/react-core/src/components/Touchspin/Touchspin.tsx
```diff
@@ -0,0 +1,98 @@
...
+  plusButtonProps,
+  ...props
+}: TouchspinProps) => {
+  const touchspinUnit = <div className={css(styles.touchspinUnit)}>{unit}</div>;
+  return (
+    <div
+      {...props}
```
tlabaj, line 53: can you spread these at the end?

packages/react-core/src/components/Touchspin/examples/Touchspin.md
```diff
@@ -0,0 +1,345 @@
...
+
+  render() {
+    const { value1, value2 } = this.state;
+    return (
+      <React.Fragment>
+        <Touchspin
+          value={value1}
+          onMinus={this.onMinus1}
+          onPlus={this.onPlus1}
+          inputProps={{ onChange: this.onChange1, name: 'touchspin-input1', 'aria-label': 'touchspin input 1' }}
```
tlabaj, line 120: Shouldn't you be using the onChange prop here?

tlabaj, line 120: @mcoker is the name optional for the input?

packages/react-integration/demo-app-ts/src/components/demos/TouchspinDemo/TouchspinDemo.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  render() {
+    const { value } = this.state;
+    const minValue = 0;
+    const maxValue = 3;
+
+    return (
+      <Touchspin
+        id="touchspin1"
```
tlabaj, line 43: can you add a example and test for with `isDisabled` and `unitPosition`

packages/react-core/src/components/Touchspin/Touchspin.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /** Indicates the whole touchspin should be disabled */
+  isDisabled?: boolean;
+  /** Callback for the minus button */
+  onMinus?: (name: string, event: React.MouseEvent) => void;
+  /** Callback for the text input changing */
+  onChange?: (event: React.FormEvent<HTMLInputElement>) => void;
+  /** Callback for the plus button */
+  onPlus?: (name: string, event: React.MouseEvent) => void;
+  /** Adds the given touchspin unit to the touchspin */
+  unit?: React.ReactNode;
```
tlabaj, line 24: @mcarrano is there a use case for a unit to be something other than a string (e.g, maybe an icon)?

tlabaj (changes requested):

packages/react-core/src/components/Touchspin/examples/Touchspin.md
kmcfaul: You can use either because they to the same place, but I'll edit the example for consistency.

packages/react-core/src/components/Touchspin/Touchspin.tsx
kmcfaul: I can add defaults for the aria-labels, but I wanted to allow a user to pass them in directly if they wanted custom strings or localized strings. 

You can pass any props on our Button component in with these, as well as general props like `id`.

kmcfaul: Also used for the thresholds to pass in the Button `isDisabled`

packages/react-core/src/components/Touchspin/examples/Touchspin.md
kmcfaul: Right now the last example uses it to consolidate various handler methods, and it is passed back in the `onPlus` and `onMinus` callbacks. But it isn't "required" and could be removed if needed (because the callbacks don't really need any props).

packages/react-core/src/components/Touchspin/Touchspin.tsx
mcarrano: @tlabaj I can't really think of why that would be the case, but I'm fine with making this flexible.

packages/react-core/src/components/Touchspin/Touchspin.tsx
tlabaj: So, I wonder if we want to have explicit props for these (aria-label, isDisabled, etc). Especially if it is a common use case ti disable one of the buttons.  I don't think we do these anywhere else where we let the consumer spread props directly to button in the component.
@mcarrano what do you think about this?  Should we expose an interface for the consumer to be able to disable the plus or minus button?

packages/react-core/src/components/Touchspin/examples/Touchspin.md
mcoker: Hmm... technically it isn't a required attribute, so I don't know what we need to require it on our end, though that's what is passed in a POST/GET request so I imagine most people will use it. It should probably just match the form control component.

mcoker (approved):
lgtm!

nicolethoen (approved):

packages/react-core/src/components/Touchspin/Touchspin.tsx
mcarrano: I'm not sure @tlabaj . But I will say that I expected that instead of the consumer needing to write code to do limit checking and disable/enable buttons, I would have expected to just pass a min or max value and have the component handle this internally. That said, I'll defer to you about the right approach.

kmcfaul: I agree with `minThreshold` and `maxThreshold` properties (and will add them), but would we still want a `props` property or restrict the user to only passing in `aria-label` for each button?

jessiehuff: I think it would be good to add a default like you guys mention, but I don't see a significant issue with passing the `aria-label` this way. 🙂

kmcfaul: And `id`

jessiehuff (commented):
Looks pretty good! 🙂  Works with keyboard and VO. The only oddity that I noticed is with the thresholds example. If you don't use the incremental buttons and input a higher number than the maximum, it seems to allow you to pass the threshold. That seems a little unexpected?

packages/react-core/src/components/Touchspin/Touchspin.tsx
kmcfaul: Actually `aria-label` should have a default already (`Minus` and `Plus`), so that should be good to go.

tlabaj: In my opinion I think that this deviates from how we usually add aria attributes.  I also think that is is not as intuitive for the consumer.  If they don't look at the example, they would nit know how to add the props.

packages/react-core/src/components/Touchspin/examples/Touchspin.md
kmcfaul: I can edit it to make it optional in the sense that the callbacks don't have to return it at least. @tlabaj Should I just do that or remove it?

packages/react-core/src/components/Touchspin/Touchspin.tsx
kmcfaul: Ok, I can remove the `props` object and only let the user pass in `minusButtonAriaLabel`, `plusButtonAriaLabel`, `minusButtonId`, and `plusButtonId`, if we're sure those are the only things we want the user to customize on our Button component. What about the `inputProps`?

kmcfaul: `props` just allows greater flexibility

kmcfaul:
Ah missed that bit of validation, will add that.

packages/react-core/src/components/Touchspin/Touchspin.tsx
tlabaj: @mcarrano how flexible do we want the plus and minus buttons to be?  
@kmcfaul I don't think we typically add additional props to sub components like buttons in a component.  If we have done that in the past it was most likely an ask form a consumer.  I also think we want a prop for the input's aria-label.
this is just my opinion.

kmcfaul: Added dedicated `aria-label` properties for the buttons and input. Kept the additional props objects in, but they generally shouldn't be used unless the user wants something specific out of them (like adding IDs for testing).

packages/react-core/src/components/Touchspin/examples/Touchspin.md
kmcfaul: Updated the functions to indicate the name is optional. The final example still uses them for convenience sake (and to demonstrate how multiple touchspins can share a handler if necessary).

mcarrano:
> @mcarrano how flexible do we want the plus and minus buttons to be?

Not sure I totally understand the question @tlabaj . I can imagine what you would add beyond upper and lower limits.

tlabaj (approved):
This looks good Katie!

jessiehuff (approved):
Nice!! LGTM!

nicolethoen (approved):

redallen (approved):

dlabrecq (commented):
LGTM

PR #5116, opened 2020-11-10
feat(packages): bump Victory package

DaoDaoNoCode:
**What**: The latest release of Victory contains some accessibility function I need to use. When I tried to upgrade the version of Victory packages in `react-chart`, I found that we only need to keep the `victory` dependency in `package.json` file and we could remove all the other `victory-` dependencies. This will simplify the upgrade of `victory` package in the future.

### DIFF
--- a/packages/react-charts/package.json
+++ b/packages/react-charts/package.json
@@ -35,7 +35,22 @@
     "hoist-non-react-statics": "^3.3.0",
     "lodash": "^4.17.19",
     "tslib": "1.13.0",
-    "victory": "^35.3.5"
+    "victory": "^35.3.5",
+    "victory-area": "^35.2.0",
+    "victory-axis": "^35.2.0",
+    "victory-bar": "^35.2.0",
+    "victory-chart": "^35.2.0",
+    "victory-core": "^35.2.0",
+    "victory-create-container": "^35.2.0",
+    "victory-group": "^35.2.0",
+    "victory-legend": "^35.2.0",
+    "victory-line": "^35.2.0",
+    "victory-pie": "^35.2.0",
+    "victory-scatter": "^35.2.0",
+    "victory-stack": "^35.2.0",
+    "victory-tooltip": "^35.2.0",
+    "victory-voronoi-container": "^35.2.0",
+    "victory-zoom-container": "^35.2.0"
   },
   "peerDependencies": {
     "react": "^16.8.0",
--- a/packages/react-charts/package.json
+++ b/packages/react-charts/package.json
@@ -35,22 +35,7 @@
     "hoist-non-react-statics": "^3.3.0",
     "lodash": "^4.17.19",
     "tslib": "1.13.0",
-    "victory": "^35.2.0",
-    "victory-area": "^35.2.0",
-    "victory-axis": "^35.2.0",
-    "victory-bar": "^35.2.0",
-    "victory-chart": "^35.2.0",
-    "victory-core": "^35.2.0",
-    "victory-create-container": "^35.2.0",
-    "victory-group": "^35.2.0",
-    "victory-legend": "^35.2.0",
-    "victory-line": "^35.2.0",
-    "victory-pie": "^35.2.0",
-    "victory-scatter": "^35.2.0",
-    "victory-stack": "^35.2.0",
-    "victory-tooltip": "^35.2.0",
-    "victory-voronoi-container": "^35.2.0",
-    "victory-zoom-container": "^35.2.0"
+    "victory": "^35.3.5"
   },
   "peerDependencies": {
     "react": "^16.8.0",
--- a/packages/react-charts/src/components/ChartAxis/ChartAxis.tsx
+++ b/packages/react-charts/src/components/ChartAxis/ChartAxis.tsx
@@ -12,7 +12,7 @@ import {
   RangePropType,
   ScalePropType,
   StringOrNumberOrList,
-  TickLabelProps
+  LabelProps
 } from 'victory-core';
 import { VictoryAxis, VictoryAxisProps, VictoryAxisTTargetType } from 'victory-axis';
 import { ChartContainer } from '../ChartContainer';
@@ -338,7 +338,7 @@ export interface ChartAxisProps extends VictoryAxisProps {
       [K in keyof React.CSSProperties]: string | number | ((tick?: any) => string | number);
     };
     tickLabels?: {
-      [K in keyof TickLabelProps]: string | number | ((tick?: any) => string | number);
+      [K in keyof LabelProps]: string | number | ((tick?: any) => string | number);
     };
   };
   /**
--- a/packages/react-charts/src/components/ChartBullet/ChartBulletGroupTitle.tsx
+++ b/packages/react-charts/src/components/ChartBullet/ChartBulletGroupTitle.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { PaddingProps, Line, StringOrNumberOrCallback } from 'victory-core';
+import { PaddingProps, Line, StringOrNumberOrCallback, VictoryLabelStyleObject } from 'victory-core';
 import { ChartContainer } from '../ChartContainer';
 import { ChartLabel } from '../ChartLabel';
 import { ChartBulletStyles, ChartThemeDefinition } from '../ChartTheme';
@@ -137,10 +137,14 @@ export const ChartBulletGroupTitle: React.FunctionComponent<ChartBulletGroupTitl
   };
 
   const labelPadding = {
-    bottom: getPaddingForSide('bottom', padding, Number(theme.legend.style.labels.padding)),
-    left: getPaddingForSide('left', padding, Number(theme.legend.style.labels.padding)),
-    right: getPaddingForSide('right', padding, Number(theme.legend.style.labels.padding)),
-    top: getPaddingForSide('top', padding, Number(theme.legend.style.labels.padding))
+    bottom: getPaddingForSide(
+      'bottom',
+      padding,
+      Number((theme.legend.style.labels as VictoryLabelStyleObject).padding)
+    ),
+    left: getPaddingForSide('left', padding, Number((theme.legend.style.labels as VictoryLabelStyleObject).padding)),
+    right: getPaddingForSide('right', padding, Number((theme.legend.style.labels as VictoryLabelStyleObject).padding)),
+    top: getPaddingForSide('top', padding, Number((theme.legend.style.labels as VictoryLabelStyleObject).padding))
   };
 
   // Horizontal divider to render under the group title
--- a/packages/react-charts/src/components/ChartBullet/ChartBulletGroupTitle.tsx
+++ b/packages/react-charts/src/components/ChartBullet/ChartBulletGroupTitle.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { PaddingProps, Line, StringOrNumberOrCallback, VictoryLabelStyleObject } from 'victory-core';
+import { PaddingProps, Line, StringOrNumberOrCallback } from 'victory-core';
 import { ChartContainer } from '../ChartContainer';
 import { ChartLabel } from '../ChartLabel';
 import { ChartBulletStyles, ChartThemeDefinition } from '../ChartTheme';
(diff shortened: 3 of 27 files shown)


DaoDaoNoCode:
@dlabrecq PTAL, thanks.

DaoDaoNoCode:
There are still some building errors after upgrading. I will check them.

packages/react-charts/package.json
```diff
@@ -35,22 +35,7 @@
...
-    "victory-create-container": "^35.2.0",
-    "victory-group": "^35.2.0",
-    "victory-legend": "^35.2.0",
-    "victory-line": "^35.2.0",
-    "victory-pie": "^35.2.0",
-    "victory-scatter": "^35.2.0",
-    "victory-stack": "^35.2.0",
-    "victory-tooltip": "^35.2.0",
-    "victory-voronoi-container": "^35.2.0",
-    "victory-zoom-container": "^35.2.0"
```
dlabrecq, line 53: We're importing directly from these packages; thus, we want to maintain our own dependencies.

For example:
`import { VictoryArea, VictoryAreaProps, VictoryAreaTTargetType } from 'victory-area';`

The Victory package may install these dependencies anyway, which is why your build compiles, but we don't want to rely on dependencies from another package.

dlabrecq (changes requested):

dlabrecq:
Please file an issue for this Victory update. It would be good to explain there what new features you're looking for.

packages/react-charts/src/components/ChartAxis/ChartAxis.tsx
```diff
@@ -12,7 +12,7 @@ import {
   RangePropType,
   ScalePropType,
   StringOrNumberOrList,
-  TickLabelProps
+  LabelProps
```
DaoDaoNoCode, line 15: https://github.com/FormidableLabs/victory/commit/439e0ad8de23d1d6ed5580a90f60feaca3f09c4e#diff-1d0ded8dfdfbe48d50e6a99272c7548f81143f224f88b6528023cc87a32e9a40R129-R132.
Here they uses `LabelProps` to replace the `TickLabelProps` in Victory.

packages/react-charts/src/components/ChartBullet/ChartBulletGroupTitle.tsx
```diff
@@ -137,10 +137,14 @@ export const ChartBulletGroupTitle: React.FunctionComponent<ChartBulletGroupTitl
   };
 
   const labelPadding = {
-    bottom: getPaddingForSide('bottom', padding, Number(theme.legend.style.labels.padding)),
-    left: getPaddingForSide('left', padding, Number(theme.legend.style.labels.padding)),
-    right: getPaddingForSide('right', padding, Number(theme.legend.style.labels.padding)),
-    top: getPaddingForSide('top', padding, Number(theme.legend.style.labels.padding))
+    bottom: getPaddingForSide(
```
DaoDaoNoCode, line 140: They change the props of `labels` from `VictoryStyleObject` to `VictoryLabelStyleObject | VictoryLabelStyleObject[]`. However, we cannot find a key named `padding` in an Array. So we need to cast it to `VictoryLabelStyleObject`. (Are there any better way to solve this problem?)

packages/react-charts/src/components/ChartCursorTooltip/ChartCursorTooltip.tsx
```diff
@@ -159,7 +158,7 @@ export interface ChartCursorTooltipProps extends ChartTooltipProps {
    * values. If this prop is not provided it will be determined from the sign of the datum, and the value of the
    * horizontal prop.
    */
-  orientation?: OrientationTypes | VictoryNumberCallback;
+  orientation?: OrientationOrCallback;
```
DaoDaoNoCode, line 161: https://github.com/FormidableLabs/victory/commit/439e0ad8de23d1d6ed5580a90f60feaca3f09c4e#diff-1d0ded8dfdfbe48d50e6a99272c7548f81143f224f88b6528023cc87a32e9a40R115
Here they give a new interface named `OrientationOrCallback`, which can be used directly here.

packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
```diff
@@ -596,8 +596,12 @@ export const ChartDonutUtilization: React.FunctionComponent<ChartDonutUtilizatio
       const donutThresholds = getDonutThresholds();
       const mergeThemeProps = (i: number) => {
         // Merge just the first color of dynamic (blue, green, etc.) with static (gray) for expected colorScale
-        newTheme.pie.colorScale[0] = donutThresholds[i].color;
-        newTheme.legend.colorScale[0] = donutThresholds[i].color;
+        if (newTheme.pie.colorScale instanceof Array) {
```
DaoDaoNoCode, line 599: https://github.com/FormidableLabs/victory/commit/439e0ad8de23d1d6ed5580a90f60feaca3f09c4e#diff-1d0ded8dfdfbe48d50e6a99272c7548f81143f224f88b6528023cc87a32e9a40L317-R317
Here `colorScale` is no longer a `string` array, but the `ColorScalePropType`, which is `type ColorScalePropType =
  | "grayscale"
  | "qualitative"
  | "heatmap"
  | "warm"
  | "cool"
  | "red"
  | "green"
  | "blue"
  | string[];`. We can only use an index on Array so we need to decide whether it's an array. (Any better ideas?)

packages/react-charts/src/components/ChartLegendTooltip/ChartLegendTooltip.tsx
```diff
@@ -294,7 +293,7 @@ export const ChartLegendTooltip: React.FunctionComponent<ChartLegendTooltipProps
   theme = getTheme(themeColor, themeVariant),
   ...rest
 }: ChartLegendTooltipProps) => {
-  const pointerLength = theme && theme.tooltip ? theme.tooltip.pointerLength : 10;
+  const pointerLength = theme && theme.tooltip ? Number(theme.tooltip.pointerLength) : 10;
```
DaoDaoNoCode, line 296: https://github.com/FormidableLabs/victory/commit/439e0ad8de23d1d6ed5580a90f60feaca3f09c4e#diff-1d0ded8dfdfbe48d50e6a99272c7548f81143f224f88b6528023cc87a32e9a40R462
Here `pointerLength` is no longer `number` but `NumberOrCallback`. So we need to cast it to `number` type.

packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
```diff
@@ -596,8 +596,12 @@ export const ChartDonutUtilization: React.FunctionComponent<ChartDonutUtilizatio
...
+          newTheme.pie.colorScale[0] = donutThresholds[i].color;
+        }
+        if (newTheme.legend.colorScale instanceof Array) {
+          newTheme.legend.colorScale[0] = donutThresholds[i].color;
+        }
```
dlabrecq, line 604: If `colorScale` is not an array, then the static color will never be set. We need to set a default, for example:

```
        if (newTheme.pie.colorScale instanceof Array) {
          newTheme.pie.colorScale[0] = donutThresholds[i].color;
        } else {
          newTheme.pie.colorScale = donutThresholds[i].color;
        }
```

packages/react-charts/src/components/ChartUtils/chart-theme.ts
```diff
@@ -113,8 +113,10 @@ export const getDonutThresholdStaticTheme = (
   invert?: boolean
 ): ChartThemeDefinition => {
   const staticTheme = cloneDeep(ChartDonutThresholdStaticTheme);
-  if (invert) {
-    staticTheme.pie.colorScale = staticTheme.pie.colorScale.reverse();
+  if (staticTheme.pie.colorScale instanceof Array) {
+    if (invert) {
```
dlabrecq, line 117: Would prefer to check `invert` first.

`if (invert && staticTheme.pie.colorScale instanceof Array) {`

dlabrecq (changes requested):

packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
DaoDaoNoCode: OK, since `string` is unchangeable so I will find a method to assign a value to `string`.

packages/react-charts/src/components/ChartBullet/ChartBulletGroupTitle.tsx
dlabrecq: I don't know if casting to `VictoryLabelStyleObject` would be correct as they are very different objects? I would be tempted to cast as `any` until `padding` is supported.

dlabrecq: Casting to VictoryLabelStyleObject is probably ok, but the alternative would be to cast as `any`.

dlabrecq: Casting to VictoryLabelStyleObject is probably ok, but the alternative would be to cast as any.

packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
dlabrecq: I didn't realize `colorScale` was unchangeable?

DaoDaoNoCode: Sorry, my bad. It's changeable. Never mind.

packages/react-charts/src/components/ChartCursorTooltip/ChartCursorTooltip.tsx
dlabrecq: `OrientationOrCallback` should be equivalent to the old types

packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
dlabrecq: We will never use "grayscale" | "qualitative" | "heatmap" | "warm" | "cool" | "red" | "green" | "blue" as our `colorScale`. These values only have meaning for the raw Victory charts.

We could change our `colorScale` types from `ColorScalePropType` to `string[]`. It's possible to omit those types from `ColorScalePropType`, but might be easier to override `ColorScalePropType` via src/typings/victory.d.ts?

packages/react-charts/src/components/ChartLegendTooltip/ChartLegendTooltip.tsx
dlabrecq: In that case, we may want to do something more like this:

const pointerLength = theme.tooltip.pointerLength === 'function' ? theme.tooltip.pointerLength() : Number(theme.tooltip.pointerLength);

dlabrecq: Victory also has the helper function below we could use:

```
export { default as Helpers } from "./victory-util/helpers";
```

```
function evaluateProp(prop, props) {
  return isFunction(prop) ? prop(props) : prop;
}
```

dlabrecq: Actually, the `isFunction` comes from lodash, which is also available

packages/react-charts/src/components/ChartCursorTooltip/ChartCursorTooltip.tsx
DaoDaoNoCode: Oh actually they are different because of the callback function. It will not return a `number` in the callback function, but returns `OrientationTypes` instead.

packages/react-charts/src/components/ChartDonutUtilization/ChartDonutUtilization.tsx
dlabrecq: Created https://github.com/patternfly/patternfly-react/issues/5129 to address the `colorScale` proper type.

packages/react-console/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@patternfly/patternfly": "4.59.1",
-    "@patternfly/react-core": "^4.75.9",
+    "@patternfly/react-core": "^4.75.10",
```
dlabrecq, line 36: Are you sure we should bump this? Typically, the @patternfly versions are kept in sync and updated with each milestone.

What's in @patternfly/react-core that is needed in the charts package?

packages/react-charts/src/components/ChartUtils/chart-tooltip.ts
```diff
@@ -49,7 +50,12 @@ export const getCursorTooltipCenterOffset = ({
   offsetCursorDimensionY = false,
   theme
 }: ChartCursorTooltipCenterOffsetInterface) => {
-  const pointerLength = theme && theme.tooltip ? theme.tooltip.pointerLength : 10;
+  const pointerLength =
+    theme && theme.tooltip
+      ? isFunction(theme.tooltip.pointerLength)
+        ? theme.tooltip.pointerLength({ index: 0 })
```
dlabrecq, line 56: Although only `index` appears to be required for `CallbackArgs`, we should also provide `data`, `datum`, `x`, and, `y` since they are available.

dlabrecq (changes requested):

packages/react-charts/src/components/ChartUtils/chart-tooltip.ts
dlabrecq: Note that `index` is a valid prop for this component, so we should pass in `rest.index` first.

packages/react-console/package.json
DaoDaoNoCode: I am not sure. But if I don't bump this, it will fail the lint test on my computer. I will revert it.

dlabrecq, line 36: If @redallen is ok with bumping `@patternfly/react-core` here, then I'm good too.

redallen: Yeah, this should have been done in an earlier PR.

redallen (approved):

dlabrecq (approved):

PR #5119, opened 2020-11-11
fix(datalist): use drag-over modifier instead of css style

### LINKED ISSUE #5041: Class to fix data list shifting when rows are dragged
Currently we're setting an inline style of `overflow-anchor: none;` on the data-list once the dragging event starts

https://github.com/patternfly/patternfly-react/blob/196bcb84d8f4bcfba134a92dd8f058d55089c04b/packages/react-core/src/components/DataList/DataList.tsx#L316

We added a class `.pf-m-drag-over` that can be applied to the data-list while dragging instead.

https://github.com/patternfly/patternfly/pull/3574/files

cc @kmcfaul

boaz0:
**What**: Closes #5041 

//cc @mcoker

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -157,6 +157,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
     this.move(this.props.itemOrder);
     Array.from(this.ref.current.children).forEach(el => {
       el.classList.remove(styles.modifiers.ghostRow);
+      el.classList.remove(styles.modifiers.dragOver);
       el.setAttribute('aria-pressed', 'false');
     });
     this.setState({
@@ -182,7 +183,13 @@ export class DataList extends React.Component<DataListProps, DataListState> {
 
   dragEnd0 = (el: HTMLElement) => {
     el.classList.remove(styles.modifiers.ghostRow);
+    el.classList.remove(styles.modifiers.dragOver);
     el.setAttribute('aria-pressed', 'false');
+    this.setState({
+      draggedItemId: null,
+      draggingToItemIndex: null,
+      dragging: false
+    });
     this.props.onDragFinish(this.state.tempItemOrder);
   };
 
@@ -326,12 +333,10 @@ export class DataList extends React.Component<DataListProps, DataListState> {
             isCompact && styles.modifiers.compact,
             gridBreakpointClasses[gridBreakpoint],
             wrapModifier && styles.modifiers[wrapModifier],
+            dragging && styles.modifiers.dragOver,
             className
           )}
-          style={{
-            ...(dragging && { overflowAnchor: 'none' }),
-            ...props.style
-          }}
+          style={props.style}
           {...props}
           {...dragProps}
           ref={this.ref}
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/DataList/DataList.tsx
```diff
@@ -326,12 +326,10 @@ export class DataList extends React.Component<DataListProps, DataListState> {
             isCompact && styles.modifiers.compact,
             gridBreakpointClasses[gridBreakpoint],
             wrapModifier && styles.modifiers[wrapModifier],
+            dragging && styles.modifiers.dragOver,
```
mcoker, line 329: This looks good to replace the existing behavior, but I think we need to change it so that `styles.modifiers.dragOver` is only applied while dragging, meaning it's removed once the item is dropped.

kmcfaul: It looks like the `dragging` state needs to be updated in the drag end handler, it is removed properly on a drag cancel. It should work with that updated.

boaz0: As @kmcfaul mentioned. `dragging` is a part of [the `DataList` state](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/DataList/DataList.tsx#L55) which is being updated to true in [`dragStart0`](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/DataList/DataList.tsx#L143-L146) and false in [`onDragCancel`](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/DataList/DataList.tsx#L162-L166). These two are being called accordingly to the events when the user clicks and holds the element or releases the element.

kmcfaul: It should also be updated in `dragEnd0` so the style gets removed when the drag is successfully completed.

tlabaj:
@boaz It looks like there are outstanding comments here.

tlabaj:
@boaz0 this one still has outstanding comments.

boaz0:
@tlabaj  sorry, I thought I fixed it. I will work on it.
Thanks.

kmcfaul (dismissed):

tlabaj (approved):
lgtm

mcoker:
Looking at this [draggable example](https://patternfly-react-pr-5119.surge.sh/components/data-list#draggable), I'm still seeing `.pf-m-drag-over` on the data list after I've stopped dragging.

boaz0:
Um, maybe I am missing something. But this is what I have on Firefox.


https://user-images.githubusercontent.com/2453279/105348877-9b8eb900-5bf1-11eb-9453-197a9a3e37f7.mp4

mcoker:
@boaz0 ah ok, I was testing in chrome - this is what I'm seeing


https://user-images.githubusercontent.com/35148959/105370748-830aa880-5bc9-11eb-9f0e-98a7087d7204.mp4

boaz0:
Thanks @mcoker I will look at this more. :laughing:

boaz0:
@mcoker do you mind review this again. I hope this fixes the problem.

boaz0:
@mcoker do you mind reviewing this again. I hope this fixes the problem.

mcoker (approved):
LGTM!!

tlabaj (approved):

kmcfaul (approved):
lgtm!

PR #5120, opened 2020-11-11
feat(DualListSelector): add new component

### LINKED ISSUE #5019: Dual-list selector component
Introduce a Dual list selector as a new component.

Core issue: https://github.com/patternfly/patternfly/issues/3556
Design: https://marvelapp.com/prototype/12e47bhj

@jessiehuff this might be a good one to review for keyboard interaction.

nicolethoen:
**What**: Closes #5019


mcarrano (changes requested):
Great start @nicolethoen . Just a couple comments:

* Arrow buttons should be disabled when there is nothing selected to move.
* In the second example, the text above the list boxes ("# available" or "#chosen") does not seem right. Is there a reason this should be different than for the first example?

@mmenestr would be good if you can also review.

nicolethoen:
@mcarrano In the second example I was just demonstrating that the default status can be overwritten and consumer defined. By default it behaves like the first example

evwilkin:
@nicolethoen there are 2 things in the examples that are a bit confusing to me - first, when deselecting an item, the background color remains on the item until you click away (the text color does change from blue to black as expected):
Screen Shot 2020-11-11 at 4 43 17 PM

Also - in the second example, it looks like the count for the "Chosen options" is mirroring the state for the "Available options" instead of actually giving the count for the chosen options:
Screen Shot 2020-11-11 at 4 47 28 PM

packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -0,0 +1,186 @@
...
+    ];
+    
+    return (
+      <DualListSelector
+        isSearchable
+        availableOptions={this.state.availableOptions}
+        availableOptionsStatus={`${this.state.availableOptions.length} available`}
+        availableOptionsActions={availableOptionsActions}
+        chosenOptions={this.state.chosenOptions}
+        chosenOptionsStatus={`${this.state.availableOptions.length} chosen`}
```
evwilkin, line 174:
```suggestion
        chosenOptionsStatus={`${this.state.chosenOptions.length} chosen`}
```

nicolethoen:
@evwilkin 
nice catch on the second note. that's my bad.
as far as the background color - I think that's because the button still has focus after you deselect it. so the grey background is a result of the focus being on the item even though it's been deselected. 
@mcoker do you know if that's an implementation problem on my end? or is that a css issue as far you can tell?

mcarrano:
> @mcarrano In the second example I was just demonstrating that the default status can be overwritten and consumer defined. By default it behaves like the first example

Can we note/document that as part of the example then? In looking at this, not sure I'd understand why the text is different. Maybe even just rename the example "Custom header with filter" or something like that?

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
```diff
@@ -0,0 +1,319 @@
...
+  availableOptionsStatus?: string;
+  /** Actions to be displayed above the available options pane. */
+  availableOptionsActions?: React.ReactNode[];
+  /** Title applied to the chosen options pane. */
+  chosenOptionsTitle?: string;
+  /** Options to display in the chosen options pane. */
+  chosenOptions?: React.ReactNode[];
+  /** Status message to display above the chosen options pane.*/
+  chosenOptionsStatus?: string;
+  /** Actions to be displayed above the available options pane. */
```
kmcfaul, line 29: Description should say `chosen`

evwilkin:
@nicolethoen it looks like there's a bug in the keyboard interaction, I can't select either of the `<` or `<<` options with the keyboard unless one of the "Available options" is selected.  The buttons will appear active as expected and can be clicked, but cannot be accessed via keyboard.

In the top example below I can't navigate down inside the controls panel, and in the bottom example I can only navigate down to the first control (`>>`) unless I select an "Available option", then I can use the keyboard to navigate between all available controls.

Screen Shot 2020-11-12 at 11 33 35 AM

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
```diff
@@ -0,0 +1,319 @@
...
+          id={`${id}-available-pane`}
+        />
+        <div
+          className={css(styles.dualListSelectorControls)}
+          tabIndex={0}
+          ref={this.controlsEl}
+          role="listbox"
+          aria-label="Selector controls"
+        >
+          <div className={css('pf-c-dual-list-selector__controls-item')} role="option">
```
kmcfaul, line 259: Can we use the `styles` object for these (also on lines 270, 281, 292)? Or are they placeholder classes?

mmenestr:
To add to the discussion on the customizable text underneath the filters, I feel like we should keep them the same for both examples. Even if they are customizable, we want to encourage people to stick with the same wording of "X out of X selected", so I would keep it the same throughout the examples. 

It goes back to the debate of how much freedom/customizability we want to give people and I don't think that should be one that we should "encourage" customizability for. I can add some text in the design guidelines i'm drafting up for this to say that it's customizable but I'd still recommend keeping it as designed!

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
```diff
@@ -0,0 +1,202 @@
...
+      /* eslint-disable @typescript-eslint/no-unused-vars */
+      filterOption,
+      onOptionSelect,
+      ...props
+    } = this.props;
+    const { input, focusedOption } = this.state;
+
+    return (
+      <div
+        className={css(styles.dualListSelectorPane, isChosen ? styles.modifiers.chosen : 'pf-m-available', className)}
```
kmcfaul, line 137: `styles.modifiers.available`? Or placeholder css / not on the styles object?

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
```diff
@@ -0,0 +1,202 @@
...
+        {...props}
+      >
+        {title && (
+          <div className={css(styles.dualListSelectorHeader)}>
+            <div className="pf-c-dual-list-selector__title">
```
kmcfaul, line 142: Same question for `styles`

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
```diff
@@ -0,0 +1,202 @@
...
+        )}
+        {(actions || isSearchable) && (
+          <div className={css(styles.dualListSelectorTools)}>
+            {isSearchable && (
+              <div className={css(styles.dualListSelectorToolsFilter)}>
+                <input
+                  className={css(formStyles.formControl, formStyles.modifiers.search)}
+                  type="search"
+                  onChange={this.onChange}
+                  aria-label={`${isChosen ? 'Chosen' : 'Available'} search input`}
```
kmcfaul, line 155: Should we allow for users to pass in aria-labels directly? For localization purposes

kmcfaul: Also applicable to the aria-labels in `DualListSelector`.

tlabaj:
I agree with Evan.  It is a confusing to me that the the item that is not selected but is hovered on has the same background color as a selected item.

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
nicolethoen: exactly, @kmcfaul

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
nicolethoen: yep, placeholder classes.

nicolethoen:
@jessiehuff Tell me how you think this sounds: 

As I understand it, the relationship between the html structure as written in core and the expectations of the VO tools is a little misaligned.

the VO tools are expecting `role='listbox'` to be tracking focus on its `role='option'` children which in the case of a list, should be the `ul` and `li` elements. But our list has buttons within the `li` and the buttons are what are getting the focus when the user uses the keyboard to navigate.

i propose we open a core issue to do two things:

1. consider removing the buttons in each `li` so we can add the event handlers to the `li` and move focus around the list items like [this example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/listbox/listbox-rearrangeable.html#ex2_label)
2. add the appropriate aria attributes to the core examples so we can be sure that the html structure matches what the VO tools expect. This way the relationship between html and VO expectations will be better defined before adding js interactions.

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
nicolethoen: I'll double check on all these hardcoded classes, but yeah lots of them are placeholders.

mcoker:
@nicolethoen @evwilkin @mcarrano 

> when deselecting an item, the background color remains on the item until you click away (the text color does change from blue to black as expected)
> do you know if that's an implementation problem on my end? or is that a css issue as far you can tell?

Yep that's a CSS issue because the element is still focused after you click on it, even when you've deselected it. Maybe we should update to remove the focus styles or make them unique?

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
```diff
@@ -0,0 +1,361 @@
...
+  /** Title applied to the available options pane. */
+  availableOptionsTitle?: string;
+  /** Options to display in the available options pane. */
+  availableOptions?: React.ReactNode[];
+  /** Status message to display above the available options pane. */
+  availableOptionsStatus?: string;
+  /** Actions to be displayed above the available options pane. */
+  availableOptionsActions?: React.ReactNode[];
+  /** Title applied to the chosen options pane. */
+  chosenOptionsTitle?: string;
```
mcoker, line 24: Do we want to have a default for this or should it be required by the user to give the panes a name?

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
```diff
@@ -0,0 +1,206 @@
...
+              aria-multiselectable="true"
+              aria-labelledby={`${id}-status`}
+              aria-activedescendant={focusedOption}
+            >
+              {options.map((option, index) => {
+                if (this.displayOption(option, input)) {
+                  return (
+                    <DualListSelectorListItem
+                      key={index}
+                      isSelected={selectedOptions.indexOf(index) !== -1}
```
mcoker, line 187: This should go on the button (`<DualListSelectorItem>`)

mcoker (changes requested):
Looks really good! The selected class just needs to go on the item `<button>`, not the parent `<li>`.

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
nicolethoen: the default for this is set on line 78. It defaults to "Chosen options" but the user can override it.

mcoker: Oh sorry, I was asking if we should just make this required instead of providing a default title. I wasn't expecting the title to come with the component by default, though maybe it's fine. wdyt @mcarrano, if you don't pass a title to either the available or chosen panes, should we provide this title by default?

mcarrano: Yes, I think that's fine.

jessiehuff (commented):
This is seriously awesome, Nicole!!!! Really nice job with this so far, this one is a complicated one. The keyboard interaction is mostly good, but I noticed one odd behavior. Once all of the options are to the right, I'm having trouble accessing the arrow buttons. If I were to choose all the options and then change my mind or notice that one of the options weren't right, I wouldn't be able to move it back over to the available options section: 
Screen Shot 2020-11-13 at 9 45 13 AM
Also, I'm still struggling to use this with VO at all. It doesn't announce when items are selected (always says "not selected"), and I can't seem to perform any of the actions. I suspect it's the same issue that we've been talking about with the structure of the li and buttons. We may need to determine what the right structure and aria attributes are needed in react and then update core so that we can ensure that it's the right interaction. As long as we're able to potentially change the structure in the future, I'm ok with this going in and following up with another issue if you're unable to fix it here.

evwilkin:
@nicolethoen looks like everything I saw mentioned above is fixed, just one new issue - on first load, the sort button in the second example doesn't work.  If you move any/all items from Available to Chosen, then the sort button works as expected moving forward.

dlabrecq (commented):
In the basic example, I cannot move any of the available options to the selected options list. It's a little confusing because I was expecting this would be "Basic" functionality.

evwilkin (approved):
LGTM!

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+export class DualListSelector extends React.Component<DualListSelectorProps, DualListSelectorState> {
+  static displayName = 'DualListSelector';
+  private controlsEl = React.createRef<HTMLDivElement>();
+  static defaultProps: PickOptional<DualListSelectorProps> = {
+    availableOptions: [] as React.ReactNode[],
+    availableOptionsTitle: 'Available options',
+    availableOptionsSearchAriaLabel: 'Available search input',
+    chosenOptions: [] as React.ReactNode[],
+    chosenOptionsTitle: 'Chosen Options',
```
tlabaj, line 80: This should be sentence case

packages/react-core/src/components/DualListSelector/DualListSelectorPane.tsx
tlabaj: @mcoker should these style `pf-m-available` be added to the component?

packages/react-integration/demo-app-ts/src/components/demos/DualListSelector/DualListSelectorDemo.tsx
```diff
@@ -0,0 +1,173 @@
...
+        toggle={<KebabToggle onToggle={isOpen => this.onToggle(isOpen, 'chosen')} id="toggle-id-6" />}
+        isOpen={this.state.isChosenKebabOpen}
+        isPlain
+        dropdownItems={dropdownItems}
+        key="chosenDropdown"
+      />
+    ];
+
+    return (
+      <DualListSelector
```
tlabaj, line 157: Can we add example to make sure we use all the props the verify typing and expected behaviour?

tlabaj (changes requested):

packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -0,0 +1,194 @@
...
+} from '@patternfly/react-core';
+import PficonSortCommonDescIcon from '@patternfly/react-icons/dist/js/icons/pficon-sort-common-desc-icon';
+import PficonSortCommonAscIcon from '@patternfly/react-icons/dist/js/icons/pficon-sort-common-asc-icon';
+
+class ManagedDualListSelector extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      availableOptions: [
+        <span>Option 1</span>,
```
mcoker, line 72: Curious why these are in a span?

mcarrano (dismissed):
Looks good. Thanks for all your work on this @nicolethoen !

mcoker (dismissed):
LGTM!

packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
nicolethoen: The most common case I expect will be that the options will all be strings, but the type of the prop is an array of ReactNodes, so this example is showing how you can wire up all the pieces to work if you are passing something more complicated that a string.

kmcfaul:
I noticed on the "Basic with actions" example that if you change the sorting on one of the panes and then move options into it, they aren't sorted until you toggle the sort button again. Maybe the example should be sorting them again whenever the options get moved?

Aside from that the code lgtm

evwilkin:
@nicolethoen the fixes look good but I caught one keyboard bug - if your search returns a subset of the options, it doesn't look like there's a way to keyboard in to select any of the options.  Note that if I try a search that returns all options (such as searching the word 'Option') it appears to work just fine.

Screen Shot 2020-11-16 at 2 39 20 PM

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
```diff
@@ -0,0 +1,394 @@
...
+      }
+      event.preventDefault();
+    }
+    if (moveFocus && controls[currentIndex]) {
+      (controls[currentIndex] as HTMLElement).focus();
+    }
+  };
+
+  componentDidMount() {
+    window.addEventListener('keydown', this.handleKeys);
```
evwilkin, line 267: This is killing all arrow key functionality for me as it calls `event.preventDefault()` on line 259 above for any arrow key - even outside of the component this is disabling the arrow keys, maybe @jessiehuff knows a better way to attach this listener

nicolethoen: oh you're right, the preventDefault is in the wrong place :)

kmcfaul:
I'm noticing now that clicking the sort button the first time doesn't sort the options. If you toggle it a second time, it sorts, but is backwards to what it should be as indicated by the button. Moving options between the lists does sort them now, and will 'fix' the sort order if its backwards, but then hitting the sort button again after that will again not sort the options.

It seems like the code thinks the list is already sorted properly for the first sort button press?

nicolethoen:
> It seems like the code thinks the list is already sorted properly for the first sort button press?

@kmcfaul Do you think I should default the list to be unsorted by default? I think the icons are facing the correct direction based on their names.
if you click it while it is a 'descending icon' it makes the items descend highest to lowest.
It may be a little confusing on the first click because it sorts them ascending (and the icon is the ascending icon) but by default the list is already ascending.

kmcfaul:
Not sure how to clip a video, so here's what I was doing to see this behavior:

From a fresh page, clicking the sort icon on the Available pane doesn't change the order. If you move any items out to the Chosen pane, or move all options to Chosen and back, the lists' sorting gets updated. If you click the Available sort button again, it again doesn't change the order. If you click the button a second time, the order is changed, but it is backwards from what the icon indicates. If you move the options, the sorting is updated.

I think it could be this line:
` if (prevState.availableDescending) returnValue = returnValue * -1;` in the `onSort` function. 
Changing `prevState.availableDescending` to `!prevState.availableDescending` appears to fix it for me (and the same change in the chosen pane counterpart).

nicolethoen:
I simplified it. I removed the changing asc/desc icon since it wasnt clear if clicking an ascending icon would make sort it in ascending order, or if it was already in ascending order. 

Now in the beginning the list is not sorted so that clicking the button clearly sorts it in ascending order. And as you move items around, the items are not automatically sorted, but you can click the button to sort them in ascending order.

is this approach better?

tlabaj (approved):

evwilkin (approved):
LGTM!

kmcfaul (approved):
LGTM the example sorts now 👍

mcarrano (approved):
Looks good! Thanks @nicolethoen .

jessiehuff (approved):
LGTM! Really awesome job with this, Nicole!! 🏆  🎉

I created an [issue](https://github.com/patternfly/patternfly-react/issues/5149) for the follow up screen reader work. 🙂

nicolethoen:
> LGTM! Really awesome job with this, Nicole!! 🏆 🎉
> 
> I created an [issue](https://github.com/patternfly/patternfly-react/issues/5149) for the follow up screen reader work. 🙂

@jessiehuff ah i meant to do that, thank you!!

mcoker (approved):

PR #5128, opened 2020-11-11
(fix)TreeView: fix various bugs

### LINKED ISSUE #4872: Tree View bugs
Comparing PatternFly Core with PatternFly React. Some of the new updates that we made in Core need to be updated in React as per this PR: https://github.com/patternfly/patternfly/pull/3410 

- spacing between the toggle and the text is smaller in react
- In the check example: tree-view-node should be a div not a button, so that you can click on the text to turn the check on and off which currently is happening in Core not React
- For current items, only the text should be blue not the arrow
- the text next to the check should be a label not a span
- in react there should be a toggle wrapper around the toggle icon

cc: @mcoker @tlabaj

### LINKED ISSUE #5165: Tree View indentation is messed up (even more after being fixed in core), especially with checkboxes
The spacing issues in core (https://github.com/patternfly/patternfly/issues/3532) were fixed (https://github.com/patternfly/patternfly/pull/3661), but in react-core the TreeView indentation is still a bit of a mess.

This is especially evident with checkboxes. The things I'm looking for are:
* Each level of the tree is indented by the same amount
* All siblings at the same tree depth have their checkboxes aligned horizontally, whether they are expandable or not

This is looking correct here: https://patternfly-pr-3661.surge.sh/components/tree-view#with-checkboxes

Screen Shot 2020-11-19 at 10 14 32 AM

But compare that with the latest pf-react example: http://patternfly-react.surge.sh/components/tree-view#with-checkboxes

Screen Shot 2020-11-19 at 10 15 34 AM

In that example you can see these issues:
* The indentation between top-level items and the depth-1 items is huge, and the indentation to depth-2 is inconsistent
* Looking at the sibling items under ApplicationLauncher -> Application 2, the "Settings" item is not aligned with the "Loader" item even though they are siblings

My guess is we need to double check that the HTML structure rendered by this component matches the new core HTML structure.

tlabaj:
**What**: Closes #4872  
closes #5165

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeView.tsx
+++ b/packages/react-core/src/components/TreeView/TreeView.tsx
@@ -1,4 +1,6 @@
 import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/TreeView/tree-view';
 import { TreeViewList } from './TreeViewList';
 import { TreeViewListItem } from './TreeViewListItem';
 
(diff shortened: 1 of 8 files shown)


mcoker:
Here is the PR that implemented the button - https://github.com/patternfly/patternfly/pull/3665

It updates `.pf-c-tree-view__node-toggle` so that it is a `<button>` in checkbox select, and `<div>` everywhere else. We will likely always use a button element there when there are interactive things in the node (like checkboxes) and `.pf-c-tree-view__node` can not be a button because you can't put something interactive in something interactive.

mturley:
Ah, I didn't see this PR was open when I opened https://github.com/patternfly/patternfly-react/issues/5165. Looks like this will address that when it's ready.

One thing I notice here is that if there are sibling items where one is expandable and the other isn't, it looks like the expandable item is one level deeper. Under ApplicationLauncher -> Application 2, the "Settings" and "Loader" items are both direct children of "Application 2", but "Loader" has an expand arrow that pushes it to the right. The content after the arrow should be aligned, so basically if there isn't an arrow the contents should have padding to compensate. It's more evident when looking at the checkboxes example (the checkboxes on those 2 sibling items should be aligned). This is working now in core, so I'm not sure what changed

mcoker:
@mturley I think the main issue is that any toggle now needs to be wrapped in `.pf-c-tree-view__node-toggle`. I wrapped the "Loader" toggle icon in `__node-toggle` and it aligns. What the `__node-toggle` element does is remove it from the node layout (it's absolutely positioned), meaning it doesn't take up space in the node row and is positioned by default at the same spot as where the checkbox/text (regular node content) starts. Then we move that element to to the left by it's own width so it appears beside the node content.

mturley:
Aha! Thanks @mcoker, that sounds right. @tlabaj do you want to fix that as part of this PR or consider it out of scope? I'd be happy to send a fix either as a commit here or as a separate PR

mturley:
@tlabaj if you're intending to also fix #5165 in this PR, the spacing is still a little bit off (although much better). Not sure what the difference in resulting HTML/CSS is compared to core now.

In the core example, all tree items are evenly vertically spaced including across parent/child boundaries, and the indentation is such that a child item's expand/collapse icon will line up with its parent's checkbox. In this PR there is some extra margin around each sub-list of items and the indentation is a bit less.

Core: https://patternfly-pr-3661.surge.sh/components/tree-view/#with-checkboxes

Screen Shot 2020-11-30 at 3 29 51 PM

This PR: https://patternfly-react-pr-5128.surge.sh/components/tree-view/#with-checkboxes

Screen Shot 2020-11-30 at 3 29 56 PM

christiemolloy (commented):
theres a weird space occuring between the first node and the second 

Screen Shot 2020-12-02 at 2 22 42 PM

christiemolloy:
The indentation for the tree view is incorrect, the correct implementation should be like the first picture

Screen Shot 2020-12-02 at 2 26 18 PM

Screen Shot 2020-12-02 at 2 26 25 PM

mturley:
@christiemolloy's review matches what I'm seeing. The indentation is more obvious-looking in the variant with checkboxes.

tlabaj:
@mturley @christiemolloy  I think I fixed the vertical spacing between the node.  The checkbox spacing still seems to be off though.  I am still debugging that piece.

christiemolloy:
@tlabaj vertical spacing is fixed but the node spacing still off

mturley:
@tlabaj the indentation and vertical spacing look perfect to me now, but it does seem that the space between the expand toggle and the checkbox / title is still slightly larger here than in core

mturley:
But I think you can add "closes https://github.com/patternfly/patternfly-react/issues/5165" to the description here, this will address that when merged now

tlabaj:
> @tlabaj the indentation and vertical spacing look perfect to me now, but it does seem that the space between the expand toggle and the checkbox / title is still slightly larger here than in core

@mcoker could this also have to do with the icon sizes?

mcarrano (approved):
Looks good to me pending @mturley 's question about the spacing between the caret and text.

mcoker:
@mturley  @tlabaj the spacing looks the same to me.

React:
Screen Shot 2020-12-07 at 12 35 27 PM
Screen Shot 2020-12-07 at 12 34 22 PM
Screen Shot 2020-12-07 at 12 34 15 PM


Core:
Screen Shot 2020-12-07 at 12 34 47 PM
Screen Shot 2020-12-07 at 12 34 58 PM
Screen Shot 2020-12-07 at 12 35 37 PM

mturley:
Ah, sorry @mcoker I was comparing against an older version of core. LGTM!

mcoker:
@tlabaj not a big deal, but should this kebab button in the action example be a dropdown?

Screen Shot 2020-12-07 at 12 43 40 PM

mturley (approved):

mcoker (dismissed):
LGTM!!! 🥳

tlabaj:
@mcoker.  I made the update to add the dropdown instead of a button.

mcoker (dismissed):

jschuler (changes requested):
LGTM but you have changes to yarn.lock and RELEASE-NOTES.md, are those intended?

tlabaj:
@jschuler good catch!  I will revert those

mturley (dismissed):

jschuler (approved):

evwilkin (approved):
LGTM

mcoker (approved):

PR #5134, opened 2020-11-13
feat(SimpleList): Uncontrolled Simple list support

### LINKED ISSUE #4918: SimpleListItem's isCurrent prop not behaving as expected
SimpleListItem accepts `isCurrent` that is described as "Indicates if the link is current/highlighted", allowing manual control. In practice though, `isCurrent` is only the initial state and is overwritten by SimpleListItem's internal state when something on the list is clicked. 

An Item in a SimpleList's  `isCurrent` prop can be hardcoded to true, making it highlighted, then click an item. It will behave as if the prop was never set and only the clicked on Item will be highlighted.

Bug, not blocking.

wojta:
**What**: Closes #4918

* Added support for the `SimpleList` component to be uncontrolled. By default it remains controlled. It's possible to set property `controlled` to `false` to make it uncontrolled.
* Propery `isCurrent` of `SimpleListItem` was renamed to `isActive` to match similar functionality in `NavList`
* Property `itemId` of `SimpleListItem` was added to identify the item.


packages/react-core/src/components/SimpleList/SimpleListItem.tsx
```diff
@@ -15,7 +17,7 @@ export interface SimpleListItemProps {
   /** Additional props added to the SimpleList <a> or <button> */
   componentProps?: any;
   /** Indicates if the link is current/highlighted */
-  isCurrent?: boolean;
```
nicolethoen, line 18: Unfortunately, changing the API - changing prop names - on public facing components constitutes a breaking change. So I agree this name change is a good idea, but we cannot be making breaking changes to our components unless it's a breaking change release.

packages/react-core/src/components/SimpleList/SimpleList.tsx
```diff
@@ -14,6 +14,8 @@ export interface SimpleListProps extends Omit<React.HTMLProps<HTMLDivElement>, '
     ref: React.RefObject<HTMLButtonElement> | React.RefObject<HTMLAnchorElement>,
     props: SimpleListItemProps
   ) => void;
+  /** Indicates whether component is controlled by its internal state (by default it's true) */
+  controlled?: boolean;
```
nicolethoen, line 18: Use `isControlled` for the flag to match the other flags' naming pattern.

nicolethoen (commented):
I'm noticing a weird behavior when after selecting an item from Group 2 in the second example.
I can then not select an item from your new third example?

It looks like trying to choose an item from Group 2 in the grouped example is sort of breaking all the simple lists? Worth a look? Or open an issue if it's not a straight forward fix?

packages/react-core/src/components/SimpleList/SimpleListItem.tsx
wojta: It's a beta component. I don't know if it's allowed, but @tlabaj told me so. I can of course revert the name back.

wojta: @nicolethoen That problem is caused by `component="a" href="#"` in the `SimpleListItem`. I don't know what was the reasoning behind this, it causes refresh of the page, so state is lost...

tlabaj: Yes, this is still a beta component so we have flexibility to make some improvements to the API.

nicolethoen:
From what I can see, I can still not select the second item from your controlled list example

wojta:
@nicolethoen Sorry, I forgot that href on one more place... Thanks for being thorough. 👍

packages/react-core/src/components/SimpleList/SimpleList.tsx
```diff
@@ -14,6 +14,8 @@ export interface SimpleListProps extends Omit<React.HTMLProps<HTMLDivElement>, '
     ref: React.RefObject<HTMLButtonElement> | React.RefObject<HTMLAnchorElement>,
     props: SimpleListItemProps
   ) => void;
+  /** Indicates whether component is controlled by its internal state (by default it's true) */
```
tlabaj, line 17: You can remove the "(by default it's true)". The documentation lists the default values.

packages/react-core/src/components/SimpleList/SimpleListItem.tsx
```diff
@@ -4,6 +4,8 @@ import styles from '@patternfly/react-styles/css/components/SimpleList/simple-li
 import { SimpleListContext } from './SimpleList';
 
 export interface SimpleListItemProps {
+  /** id for the item. */
+  itemId?: number | string;
```
tlabaj, line 8: Can we just call this prop `id`?

tlabaj (dismissed):
the a11y test are also still failing.

packages/react-core/src/components/SimpleList/SimpleListItem.tsx
wojta: not sure, I would rather not to avoid confusion with HTML `id` property. It's the same as in the `NavItem` component.

packages/react-core/src/components/SimpleList/SimpleList.tsx
wojta: ok

wojta:
> the a11y test are also still failing.

I don't know what's the process there. It is in the separate project. So something has chaged in the properties and this  causes the test to fail. I would need to update the test for this as well.

kmcfaul:
Unsure if this is the source of the a11y issue (it doesn't look like the tests even ran according to CircleCI), but it seems like `itemID` isn't being assigned to anything in SimpleListItem. It just catches the prop and doesn't do anything with it. I would expect it to be assigned in the `li`?

wojta:
```@patternfly/react-docs:       TS2769: No overload matches this call.
@patternfly/react-docs:   Overload 1 of 2, '(props: Readonly<SimpleListItemProps>): SimpleListItem', gave the following error.
@patternfly/react-docs:     Type '{ children: any[]; key: any; className: string; isCurrent: boolean; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<SimpleListItem> & Pick<Readonly<SimpleListItemProps> & Readonly<...>, never> & Partial<...> & Partial<...>'.
@patternfly/react-docs:       Property 'isCurrent' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<SimpleListItem> & Pick<Readonly<SimpleListItemProps> & Readonly<...>, never> & Partial<...> & Partial<...>'.
```, this is because I renamed the property. Should I make PR for `patternfly-a11y`?

**Comment by redallen** [comment]
@wojta Yep!

https://github.com/patternfly/patternfly-a11y

Once you make a PR we can release a new version and you can bump it to that version in this PR.

**Review by redallen — DISMISSED** [review]

File: packages/react-core/src/components/SimpleList/SimpleListItem.tsx (hunk 1 above)
Comment (dlabrecq) [inline packages/react-core/src/components/SimpleList/SimpleListItem.tsx:18]: Isn't this a breaking change? I see this property being used in the existing examples. Thinking it needs to be deprecated instead?

```
<SimpleListItem key="item1" isCurrent>
  List item 1
</SimpleListItem>,
```

dlabrecq (changes requested):

packages/react-core/src/components/SimpleList/SimpleListItem.tsx
wojta: ok, I can do that, but I think that shouldn't be necessary for the beta component.

redallen (approved):

packages/react-core/src/components/SimpleList/SimpleListItem.tsx
ddonahue007: @wojta I think you make a good point here. If it is a Beta component then I believe backwards compatibility goes out the window.  people should use Beta components at their own risk and expect them to change.

ddonahue007 (approved):
LGTM

dlabrecq:
Didn't realize this was still beta

dlabrecq (approved):

kmcfaul (approved):

PR #5138, opened 2020-11-13
[WIP] feat(ChartBar): add accessibility support

### LINKED ISSUE #5131: Add a11y demo for Bar Chart
Since Victory added the support for change the role of `<svg>`, now we can first add an a11y demo to bar chart that support focus and screen reader.

Enhancement.

DaoDaoNoCode:
**What**: Closes #5131 .
Since now we can define the `role` of a SVG image in Victory, we are able to add accessibility function to the chart.
The first demo I add is in the Bar Chart. There are 2 aspects that need to be done if a user wants to add a11y feature.

1. Add a11y feature to the whole Bar Chart
    Simply pass 2 props `role='group'` and `tabIndex={0}` to `<ChartVoronoiContainer>`. The first prop is to make sure we can access the contents in the SVG image. The second prop is to make sure we can focus on the whole Bar Chart. Here, VO will read `ariaTitle` and `ariaDesc` we defined before.
2. Add a11y feature to each individual Bar
    Here I add a new prop for `<ChartBar>` named `ariaLabel`, which will read the string array of `aria-label` that given by users. In the demo, I also show how a user could map the data to generate the formatted `aria-label` they want. In `<ChartBar>` component, I will check whether user pass the `ariaLabel` prop into the component. If so, I will create an `aria-label` and `tabIndex={0}` for every individual Bar. If not, users will not be able to focus on the content inside the Bar Chart. Here, VO will read out the aria-label for every Bar when we press Tab to focus on them one by one.

**Additional issues**:
This is only a draft version and needs a lot of reviews and suggestion. There are several additional issues.

1. When I Tab among the bars, it will be in the render order. As you can see in the demo, if I start from `2015 Cats` and press Tab, it will go to `2016 Cats -> 2017 Cats -> 2018 Cats -> 2015 Dogs...`. However, what we expected and what looks more reasonable is I start from `2015 Cats` and press Tab it will go to `2015 Dogs -> 2015 Birds -> 2015 Mice -> 2016 Cats...`. Is there a good way to solve the problem? Although we can still use VO to read them no matter what order it is. If we change the `tabIndex` of each bar, it may change the Tab behavior of the whole page.
2. When the Bar Chart is extremely huge with a lot of data, user may not want to use Tab to go over them because they need to hit Tab for many times. Is there a good solution for this? (@dlabrecq mentioned we may use `Skip to Content` component, but what's the condition to trigger that component?)
3. This is only a demo for Bar Chart because it's simple, any idea on how to read out those complex charts?
4. Any better way to pass `ariaLabel` of every individual bar? Now it's a little bit complex for our users to map the data. Can we do the mapping procedure for them?

### DIFF
--- a/packages/react-charts/src/components/ChartBar/ChartBar.tsx
+++ b/packages/react-charts/src/components/ChartBar/ChartBar.tsx
@@ -19,7 +19,7 @@ import {
   StringOrNumberOrList,
   VictoryStyleInterface
 } from 'victory-core';
-import { VictoryBar, VictoryBarAlignmentType, VictoryBarProps, VictoryBarTTargetType } from 'victory-bar';
+import { Bar, VictoryBar, VictoryBarAlignmentType, VictoryBarProps, VictoryBarTTargetType } from 'victory-bar';
 import { ChartContainer } from '../ChartContainer';
 import { ChartThemeDefinition } from '../ChartTheme';
 import { getTheme } from '../ChartUtils';
@@ -422,12 +422,20 @@ export interface ChartBarProps extends VictoryBarProps {
    * @example 'last_quarter_profit', () => 10, 1, 'employees.salary', ["employees", "salary"]
    */
   y0?: DataGetterPropType;
+  /**
+   * Use ariaLabel prop to determine how screen reader will read the contents in the chart.
+   * If you don't define this prop, the content in the chart will not be accessible.
+   *
+   * @example ['There are 7 cats in 2015', 'There are 5 cats in 2016', 'There are 8 cats in 2017', 'There are 2 cats in 2018']
+   */
+  ariaLabel?: string[];
 }
 
 export const ChartBar: React.FunctionComponent<ChartBarProps> = ({
   containerComponent = <ChartContainer />,
   themeColor,
   themeVariant,
+  ariaLabel,
 
   // destructure last
   theme = getTheme(themeColor, themeVariant),
@@ -440,7 +448,14 @@ export const ChartBar: React.FunctionComponent<ChartBarProps> = ({
   });
 
   // Note: containerComponent is required for theme
-  return <VictoryBar containerComponent={container} theme={theme} {...rest} />;
+  return (
+    <VictoryBar
+      containerComponent={container}
+      theme={theme}
+      dataComponent={ariaLabel && <Bar ariaLabel={({ index }) => ariaLabel[Number(index)]} tabIndex={0} />}
+      {...rest}
+    />
+  );
 };
 ChartBar.displayName = 'ChartBar';
 
--- a/packages/react-charts/src/components/ChartBar/ChartBar.tsx
+++ b/packages/react-charts/src/components/ChartBar/ChartBar.tsx
@@ -45,6 +45,13 @@ export interface ChartBarProps extends VictoryBarProps {
    */
   animate?: boolean | AnimatePropTypeInterface;
   /**
+   * Use ariaLabel prop to determine how screen reader will read the contents in the chart.
+   * If you don't define this prop, the content in the chart will not be accessible.
+   *
+   * @example 'There are 7 cats in 2015'
+   */
+  ariaLabel?: string;
+  /**
    * The barRatio prop specifies an approximate ratio between bar widths and spaces between bars.
    * When width is not specified via the barWidth prop or in bar styles, the barRatio prop will
    * be used to calculate a default width for each bar given the total number of bars in the data series
@@ -422,13 +429,6 @@ export interface ChartBarProps extends VictoryBarProps {
    * @example 'last_quarter_profit', () => 10, 1, 'employees.salary', ["employees", "salary"]
    */
   y0?: DataGetterPropType;
-  /**
-   * Use ariaLabel prop to determine how screen reader will read the contents in the chart.
-   * If you don't define this prop, the content in the chart will not be accessible.
-   *
-   * @example ['There are 7 cats in 2015', 'There are 5 cats in 2016', 'There are 8 cats in 2017', 'There are 2 cats in 2018']
-   */
-  ariaLabel?: string[];
 }
 
 export const ChartBar: React.FunctionComponent<ChartBarProps> = ({
@@ -452,7 +452,7 @@ export const ChartBar: React.FunctionComponent<ChartBarProps> = ({
     <VictoryBar
       containerComponent={container}
       theme={theme}
-      dataComponent={ariaLabel && <Bar ariaLabel={({ index }) => ariaLabel[Number(index)]} tabIndex={0} />}
+      dataComponent={ariaLabel && <Bar ariaLabel={ariaLabel} tabIndex={0} />}
       {...rest}
     />
   );
--- a/packages/react-charts/src/components/ChartBar/examples/ChartBar.md
+++ b/packages/react-charts/src/components/ChartBar/examples/ChartBar.md
@@ -34,7 +34,7 @@ import { Chart, ChartAxis, ChartBar, ChartGroup, ChartVoronoiContainer } from '@
     ariaDesc="Average number of pets"
     ariaTitle="Bar chart example"
     containerComponent={<ChartVoronoiContainer labels={({ datum }) => `${datum.name}: ${datum.y}`} constrainToVisibleArea />}
-    domain={{y: [0,9]}}
+    domain={{ y: [0, 9] }}
     domainPadding={{ x: [30, 25] }}
     legendData={[{ name: 'Cats' }, { name: 'Dogs' }, { name: 'Birds' }, { name: 'Mice' }]}
     legendOrientation="vertical"
@@ -144,7 +144,7 @@ import { Chart, ChartBar, ChartVoronoiContainer } from '@patternfly/react-charts
     ariaDesc="Average number of pets"
     ariaTitle="Bar chart example"
     containerComponent={<ChartVoronoiContainer labels={({ datum }) => `${datum.name}: ${datum.y}`} constrainToVisibleArea />}
-    domain={{y: [0,9]}}
+    domain={{ y: [0, 9] }}
     domainPadding={{ x: [30, 25] }}
     legendData={[{ name: 'Cats' }]}
     legendOrientation="vertical"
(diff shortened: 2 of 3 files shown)


packages/react-charts/src/components/ChartBar/ChartBar.tsx
```diff
@@ -440,7 +448,14 @@ export const ChartBar: React.FunctionComponent<ChartBarProps> = ({
   });
 
   // Note: containerComponent is required for theme
-  return <VictoryBar containerComponent={container} theme={theme} {...rest} />;
+  return (
+    <VictoryBar
+      containerComponent={container}
+      theme={theme}
+      dataComponent={ariaLabel && <Bar ariaLabel={({ index }) => ariaLabel[Number(index)]} tabIndex={0} />}
```
dlabrecq, line 455: Curious why `tabIndex` is always set to zero? I believe we should allow devs to set that value -- use a default if necessary.

packages/react-charts/src/components/ChartBar/ChartBar.tsx
```diff
@@ -422,12 +422,20 @@ export interface ChartBarProps extends VictoryBarProps {
    * @example 'last_quarter_profit', () => 10, 1, 'employees.salary', ["employees", "salary"]
    */
   y0?: DataGetterPropType;
+  /**
+   * Use ariaLabel prop to determine how screen reader will read the contents in the chart.
+   * If you don't define this prop, the content in the chart will not be accessible.
+   *
+   * @example ['There are 7 cats in 2015', 'There are 5 cats in 2016', 'There are 8 cats in 2017', 'There are 2 cats in 2018']
+   */
+  ariaLabel?: string[];
```
dlabrecq, line 431: I don't believe we want to pass in an array of strings here. In Victory's Bar example, they provide the datum prop as shown below:

ariaLabel={({ datum }) => x: ${datum.x}}

We follow this same pattern for our tooltips, for example.

The `datum` property can contain a lot of information. Besides just the `x` prop shown above, devs can add their own custom properties, which is very useful for creating custom content.

dlabrecq: This should also be in alphabetical order.

packages/react-charts/src/components/ChartBar/ChartBar.tsx
dlabrecq: The tab order is odd. Instead of tabbing to each bar in the year 2015, it skips to the next year.

For example, cats 2015, cats 2016, cats 2017, cats 2018, dogs 2015, dogs 2016, etc.

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
```diff
@@ -163,20 +163,105 @@ import { Chart, ChartBar, ChartVoronoiContainer } from '@patternfly/react-charts
 </div>
 ```
 
+### Basic with right aligned legend (Accessibility Demo)
+
+This demo demonstrate the accessibility function in a Bar chart. You can now focus on the whole chart or the contents in the chart. At the same time, you could use screen reader like VoiceOver to read the chart and contents as you defined in format.
+
```
dlabrecq, line 169: I believe this demo would be best in its own a11y section? Similar to functionality for tooltips, legend, etc. We could build more complex examples there.

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
```diff
@@ -163,20 +163,105 @@ import { Chart, ChartBar, ChartVoronoiContainer } from '@patternfly/react-charts
...
+      { name: 'Birds', x: '2017', y: 9 },
+      { name: 'Birds', x: '2018', y: 7 }
+    ],
+    mice: [
+      { name: 'Mice', x: '2015', y: 3 },
+      { name: 'Mice', x: '2016', y: 3 },
+      { name: 'Mice', x: '2017', y: 8 },
+      { name: 'Mice', x: '2018', y: 5 }
+    ]
+  };
```
dlabrecq, line 200: The particular example only has 16 data points. I'd like to see an example with more data points to see how well tabbing works.

The example below has 90 data points.
http://patternfly-react.surge.sh/charts/stack-chart#monthly-data-with-responsive-container

In Cost Management; for example, we have 6 sets of data with a total of 180 data points. However, there are up to 9 charts on a single page. 

My concern is that it may be annoying to tab thru all of that?

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
```diff
@@ -163,20 +163,105 @@ import { Chart, ChartBar, ChartVoronoiContainer } from '@patternfly/react-charts
...
+  );
+  console.log(ariaLabel);
+  return (
+    <div style={{ height: '250px', width: '600px' }}>
+      <Chart
+        ariaDesc="Average number of pets"
+        ariaTitle="Accessible Bar chart example"
+        containerComponent={
+          <ChartVoronoiContainer
+            role="group"
```
dlabrecq, line 214: It's not clear that `role=group` is required to make tabbing possible. Perhaps this is something we want to set by default?

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
```diff
@@ -163,20 +163,105 @@ import { Chart, ChartBar, ChartVoronoiContainer } from '@patternfly/react-charts
...
+            tabIndex={0}
```
dlabrecq, line 215: Curious why tabIndex is always set to zero? 

How is this used, when `VictoryBar` is hard coded to zero?

dlabrecq (changes requested):
I see that `VictoryAccessibleGroup` allows `aria-label` and `tabIndex` to be set. Devs can use that to change the `role` and even the `g` tag itself.

Did you explore the possibility of overriding `groupComponent` instead?

packages/react-charts/src/components/ChartBar/ChartBar.tsx
DaoDaoNoCode: > Curious why `tabIndex` is hard coded to zero? I believe we should allow devs to set that value -- use a default if necessary.

If we don't set `tabIndex` here, it will be set to default `-1` in Victory. And in the global style file in Victory, if an element has `-1` as `tabIndex` value, we cannot focus on it. So if we want to focus on an individual bar in our chart, we have to assign it a value.

DaoDaoNoCode: > The tab order is odd. Instead of tabbing to each bar in the year 2015, it skips to the next year.
> 
> For example, cats 2015, cats 2016, cats 2017, cats 2018, dogs 2015, dogs 2016, etc.

Yes, this is the additional issue I mentioned in the PR description. I am also waiting for some suggestions about how to deal with the problem. We can of course set every bar a different `tabIndex` value to control how users Tab through the bars. However, if there are other charts on the same page uses the same `tabIndex`, it will performs like jumping from a chart to the other chart.

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
DaoDaoNoCode: Will do. Thanks!

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
DaoDaoNoCode: I also noticed this point and mentioned it in the additional issues. Still need more suggestions about how we could implement this!

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
DaoDaoNoCode: It's `tabIndex` but not `role=group` that makes it possible for us to tab and focus the whole chart. However, it's `role=group` that makes it possible to access the contents inside the chart. If we set it to default `role=img`, we can never tab into the chart and focus on the individual bar.

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
DaoDaoNoCode: It's also used to focus on the chart. It should be on the `svg` tag along with the `role`.
Screen Shot 2020-11-16 at 2 18 19 PM

packages/react-charts/src/components/ChartBar/ChartBar.tsx
DaoDaoNoCode: Fixed, thanks!

packages/react-charts/src/components/ChartBar/examples/ChartBar.md
jessiehuff: I think this is a really good point - I think it would be tedious to tab through 90-250k data points. I've been trying to do some research on how other accessible charts do this, and it doesn't seem like there's a firm standard. However, my favorite chart I've seen so far in terms of how they handle this situation is this [Highcharts demo](https://www.highcharts.com/demo/bar-basic). 

What I like about their implementation is that they have two axes/directions for navigation, columns and rows. Left/right arrow keys navigate bars within a group (and only the one group), and up/down arrow keys navigate the same category from group to group, then starts over with the next category. The tab key itself navigates to the main regions (like starting data point then legend then to the next element). This would account for a case where you have a ton of data points because you only tab a few times per chart but can use arrow keys if you want to dig into the data points deeper. I'm guessing we could use event handlers to manage this focus similar to how we've handled [dropdown](https://github.com/patternfly/patternfly-react/blob/21a343deedf3c30f6dc63f34b23f10cf93103611/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx#L113) keyboard interaction, [select](https://github.com/patternfly/patternfly-react/blob/6abd540e61ad0226b359e2f91f3be6042dd76f2a/packages/react-core/src/components/Select/SelectOption.tsx#L104), etc.

dlabrecq: Using the up/down arrow keys would certainly be better than tabbing. Although, keep in mind that the Highcharts demo is also a simple bar chart with limited (~12) data points.

I'd also like to explore the use of a hidden HTML table. The idea is that screen readers could choose to navigate thru all the data if desired, but in a more standard way, via a traditional HTML table. This would free typical users from having all those data points added to the tab index.

DaoDaoNoCode: Thanks @jessiehuff @dlabrecq , I will also do some research on both Highcharts and hidden HTML table.

PR #5140, opened 2020-11-15
feat(alert): add support to custom icon

### LINKED ISSUE #5125: Add custom Alert type
Add the ability to pass a custom icon to an Alert via a prop. Reference: https://github.com/patternfly/patternfly/pull/3641

boaz0:
**What**: Closes #5125 

//cc @mcoker

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -46,6 +46,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   truncateTitle?: number;
   /** Position of the tooltip which is displayed if text is truncated */
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
+  /** Set a custom icon to the Alert. If not set the icon is set according to the variant */
+  customIcon?: React.ReactNode;
 }
 
 export const Alert: React.FunctionComponent<AlertProps> = ({
@@ -65,6 +67,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   onTimeout,
   truncateTitle = 0,
   tooltipPosition,
+  customIcon,
   ...props
 }: AlertProps) => {
   const ouiaProps = useOUIAProps(Alert.displayName, ouiaId, ouiaSafe, variant);
@@ -128,7 +131,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
           'aria-atomic': 'false'
         })}
       >
-        <AlertIcon variant={variant} />
+        <AlertIcon variant={variant} customIcon={customIcon} />
         {isTooltipVisible ? (
           <Tooltip content={getHeadingContent} position={tooltipPosition}>
             {Title}
--- a/packages/react-core/src/components/Alert/AlertIcon.tsx
+++ b/packages/react-core/src/components/Alert/AlertIcon.tsx
@@ -20,13 +20,15 @@ export interface AlertIconProps extends React.HTMLProps<HTMLDivElement> {
   variant: 'success' | 'danger' | 'warning' | 'info' | 'default';
   /** className */
   className?: string;
+  /** A custom icon. If not set the icon is set according to the variant */
+  customIcon?: React.ReactNode;
 }
 
-export const AlertIcon = ({ variant, className = '', ...props }: AlertIconProps) => {
+export const AlertIcon = ({ variant, customIcon, className = '', ...props }: AlertIconProps) => {
   const Icon = variantIcons[variant];
   return (
     <div {...props} className={css(styles.alertIcon, className)}>
-      <Icon />
+      {customIcon || <Icon />}
     </div>
   );
 };
--- a/packages/react-core/src/components/Alert/__tests__/Alert.test.tsx
+++ b/packages/react-core/src/components/Alert/__tests__/Alert.test.tsx
@@ -4,6 +4,7 @@ import { mount } from 'enzyme';
 import { Alert, AlertVariant } from '../Alert';
 import { AlertActionLink } from '../AlertActionLink';
 import { AlertActionCloseButton } from '../AlertActionCloseButton';
+import { UsersIcon } from '@patternfly/react-icons';
 
 test('default Alert variant is default', () => {
   const view = mount(<Alert title="this is a test">Alert testing</Alert>);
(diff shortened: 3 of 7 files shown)


packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -9,6 +9,13 @@ import { AlertContext } from './AlertContext';
 import maxLines from '@patternfly/react-tokens/dist/js/c_alert__title_max_lines';
 import { Tooltip } from '../Tooltip';
 
+export interface SVGIconProps extends Omit<React.HTMLProps<SVGElement>, 'size' | 'ref'> {
+  color?: string;
+  size?: 'sm' | 'md' | 'lg' | 'xl';
+  title?: string;
+  noVerticalAlign?: boolean;
```
mcoker, line 16: @boaz0 what does this block do?

mcoker (approved):
LGTM! Everything works as expected, though I did have a question about icon props.

packages/react-core/src/components/Alert/Alert.tsx
boaz0: I think there is a problem with the linter. I hardcoded the SVGIconProps from `react-icons` to make sure the value passed to the `customIcon` is an icon and not just anything else. The problem is the linter tells me that I cannot export SVGIconProps in that way and I need to change it to something else. However, running `yarn lint:ts --fix` fixes the lint error but the build fails because the file it imports SVGIconProps does not exist.

I hardcoded it to verify that this is the problem. I will probably have to fix the linter before merging this because it's stupid to have this duplication.

Thanks for having a look :laughing:

jschuler: The linting rule sees that the import is in braces `import { SVGIconProps }` and thinks you're trying to import from a barrel file which is why it's throwing the error. You can add an ignore comment since it's a bug with the rule
```
// eslint-disable-next-line patternfly-react/import-tokens-icons
import { SVGIconProps } from '@patternfly/react-icons/dist/js/createIcon';
```

boaz0: I did that but the linter insists on updating it to `'@patternfly/react-icons/dist/js/icons/s-v-g-icon-props';`.
I had to tweak the linter a little bit to treat this particular source path to what you suggested (here: https://github.com/patternfly/patternfly-react/pull/5163)

jschuler: Thanks @boaz0 , that PR is merged now so you can rebase this PR :)

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -65,6 +68,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   onTimeout,
   truncateTitle = 0,
   tooltipPosition,
+  customIcon = undefined,
```
jschuler, line 71:
```suggestion
  customIcon,
```
it's undefined by default

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -46,6 +47,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   truncateTitle?: number;
   /** Position of the tooltip which is displayed if text is truncated */
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
+  /** Set a custom icon to the Alert */
```
jschuler, line 50: Can you also expand on the description here `If not set the icon is set according to the variant`

tlabaj (commented):
@boaz0 looks like there are still outstanding comments on this one.

tlabaj (changes requested):
This on needs to be added to demo app .

mcarrano (dismissed):
Looks great. Thanks @boaz0 !

mcoker (dismissed):

jschuler (approved):

kmcfaul (dismissed):

packages/react-core/src/components/Alert/AlertIcon.tsx
```diff
@@ -20,10 +21,12 @@ export interface AlertIconProps extends React.HTMLProps<HTMLDivElement> {
...
+  customIcon?: React.ComponentClass<SVGIconProps>;
 }
 
-export const AlertIcon = ({ variant, className = '', ...props }: AlertIconProps) => {
-  const Icon = variantIcons[variant];
+export const AlertIcon = ({ variant, customIcon, className = '', ...props }: AlertIconProps) => {
+  const Icon = customIcon || variantIcons[variant];
   return (
     <div {...props} className={css(styles.alertIcon, className)}>
       <Icon />
```
jschuler, line 32: on second thought... we would be limiting the icon here to only allow react icons. how about we open this up to accept `children` so you're not limited:
```
<div {...props} className={css(styles.alertIcon, className)}>
  {children || <Icon />
</div>
```

jschuler (changes requested):

redallen (approved):

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -46,6 +47,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   truncateTitle?: number;
   /** Position of the tooltip which is displayed if text is truncated */
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
+  /** Set a custom icon to the Alert. If not set the icon is set according to the variant */
+  customIcon?: React.ComponentClass<SVGIconProps>;
```
redallen, line 51: We should probably just make this `React.ReactNode` in the future.

redallen, line 51: To piggyback on @jschuler 's change of mind let's make this `React.ReactNode` to allow anything.

redallen (changes requested):

packages/react-core/src/components/Alert/Alert.tsx
tlabaj: That was my initial thought as well. That is what we do elsewhere.  
@boaz was there a reason for the `React.ComponentClass<SVGIconProps>` type here?

boaz0: The `createIcon` method returns `React.ComponentClass<SVGIconProps>` and I assumed we want the user to pass only icons.
If this is not an issue then I am OK with going with the `React.ReacNode` type.

redallen: It's not an issue. We want to offer the user the most flexibility.

boaz0:
@tlabaj @jschuler @redallen - done please review this PR again.
Thanks for the comments.

redallen (approved):
LGTM!

jschuler (approved):
LGTM, thank you!

boaz0:
Thank you all
 @jschuler @redallen @tlabaj @mcoker @mcarrano

PR #5145, opened 2020-11-17
fix(FilterSidePanelCategoryItem): fix typings

### LINKED ISSUE #5139: FilterSidePanelCategoryItem has wrong type
Reference:
https://github.com/patternfly/patternfly-react/blob/master/packages/react-catalog-view-extension/src/components/FilterSidePanel/FilterSidePanelCategoryItem.tsx#L7

Trying to use FilterSidePanelCategoryItem in `tsx`, getting the following tserror:

`Type 'X' is missing the following properties from type 'FilterSidePanelCategoryItemProps': context, setState, forceUpdate, render, and 3 more.`

Looking at the typings
`export interface FilterSidePanelCategoryItemProps extends React.Component<Checkbox>`

looks wrong. AFAIK it should be `React.HTMLProps<HTMLDivElement>`

glekner:
**What**: Closes #5139

### DIFF
--- a/packages/react-catalog-view-extension/src/components/FilterSidePanel/FilterSidePanelCategoryItem.tsx
+++ b/packages/react-catalog-view-extension/src/components/FilterSidePanel/FilterSidePanelCategoryItem.tsx
@@ -4,7 +4,7 @@ import { Checkbox } from '@patternfly/react-core';
 
 import { getUniqueId } from '../../helpers/util';
 
-export interface FilterSidePanelCategoryItemProps extends React.Component<Checkbox> {
+export interface FilterSidePanelCategoryItemProps extends React.HTMLProps<HTMLDivElement> {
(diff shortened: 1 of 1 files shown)


redallen (changes requested):
This may cause breakage in [Openshift](https://github.com/openshift/console/blob/master/frontend/public/components/utils/tile-view-page.jsx#L809) since the `title` prop on `<Checkbox>` gets passed to an `<input>` which I think makes more since than passing it to an outer `<div>`.

If you could add a title prop that gets passed to the `<Checkbox>` still this could be non-breaking for Openshift which is the only other known consumer of react-catalog-view-extension.

glekner:
done @redallen

redallen (approved):
LGTM!

kmcfaul (approved):
lgtm!

PR #5148, opened 2020-11-17
Fix TS type for isChecked prop of Checkbox so it allows null (indeterminate)

mturley:
Our Checkbox component supports a null value for `isChecked`, in order to show the indeterminate (partially-checked) state. We demonstrate this in the Controlled example of the checkbox docs page.

However, when trying to pass this null value into the prop in a TypeScript app, TS gives an error since the type of this prop is `boolean` (null is not assignable). This doesn't cause an issue in our examples since they are plain JS.

This PR simply changes the `isChecked` prop type to `boolean | null` so TS won't complain when a null value is passed.


redallen (approved):

jessiehuff (approved):
LGTM! :)

PR #5150, opened 2020-11-17
TreeView: add support for indeterminate checkboxes without requiring consumer to use refs

mturley:
Currently, in order to show an indeterminate (partially-checked) checkbox in a TreeViewListItem, the consumer has to use a ref to set the element's `indeterminate` prop. This is what we already do internally in our regular Checkbox when the consumer passes a null value for `isChecked`. This PR adds support for rendering this indeterminate state just by passing a null value for `checkProps: { checked }` on a tree item instead of using a ref. I updated the existing tree view example to use this approach instead of the ref.

This isn't a breaking change: if a consumer is already using their own ref to set this property, that will still work.

I considered refactoring the tree view to use our `<Checkbox>` component instead of a native `<checkbox>` so we wouldn't have to duplicate this logic in both components, but the tree view checkbox needs to be pretty bare for the styles to work correctly and it seemed overkill to add props to `<Checkbox>` that disabled a bunch of its features and styles.

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
@@ -6,6 +6,10 @@ import AngleDownIcon from '@patternfly/react-icons/dist/js/icons/angle-down-icon
 import { TreeViewDataItem } from './TreeView';
 import { Badge } from '../Badge';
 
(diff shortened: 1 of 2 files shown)


redallen (approved):
Clever. LGTM!

kmcfaul (approved):
lgtm!

PR #5160, opened 2020-11-18
fix(datepicker): forward popoverprops

### LINKED ISSUE #5159: Date picker misses menuAppendTo prop
According to [docs](https://www.patternfly.org/v4/components/date-picker) and my own experiments, `DatePicker` components does not have `menuAppendTo` prop. That means that when I use this component in `Modal`, upon opening the calendar picker it just expands the modal body but I need to scroll to see it. The same happens with `Select`, but when I specify `menuAppendTo="parent"` it behaves as expected.

enhancement/missing feature

"@patternfly/react-datetime": "4.1.0"

redallen:
**What**: Closes #5159

**Additional issues**:

### DIFF
--- a/packages/react-datetime/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-datetime/src/components/DatePicker/DatePicker.tsx
@@ -5,7 +5,7 @@ import styles from '@patternfly/react-styles/css/components/DatePicker/date-pick
 import buttonStyles from '@patternfly/react-styles/css/components/Button/button';
 import { Locales, Locale } from '../../helpers';
 import { TextInput } from '@patternfly/react-core/dist/js/components/TextInput/TextInput';
-import { Popover } from '@patternfly/react-core/dist/js/components/Popover/Popover';
+import { Popover, PopoverProps } from '@patternfly/react-core/dist/js/components/Popover/Popover';
 import { InputGroup } from '@patternfly/react-core/dist/js/components/InputGroup/InputGroup';
 import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/js/icons/outlined-calendar-alt-icon';
 import { parse, format, isValid as isNotNull } from 'date-fns';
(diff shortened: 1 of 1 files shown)


nicolethoen (dismissed):

jschuler:
Maybe worthwhile to add both `menuAppendTo` as well (and `popoverProps` after) to cover the most common use-case?

redallen:
@jschuler What should happen then if a user passes `menuAppendTo` and `popoverProps: { menuAppendTo }`?

jschuler:
@redallen since `{...popoverProps}` is spread last into the Popover that should take priority over any prior props.

jschuler:
You could also make `popoverProps` only allow a subset of props from Popover by using `pick`

redallen:
I've added a convenience `appendTo` prop.

jschuler (approved):

jschuler:
@redallen I forgot until now but for other menu components we also let users pass `inline` and `parent` for `menuAppendTo` as a convenience, see https://github.com/patternfly/patternfly-react/blob/c9b183473e2aa366ef4da032caa73fa34b9d482c/packages/react-core/src/components/Select/Select.tsx#L936

redallen:
@jschuler This isn't a `menuAppendTo` prop but rather just `appendTo`. About modal, Tooltip, Popover, Popper, and Wizard all don't support `inline`/ `parent`. I think adding that support is outside the scope of this PR.

jschuler:
oh right, date picker always uses a Popover whereas the other components only use it optionally

nicolethoen:
@jschuler if the DatePicker popover defaults the appendTo to document.body by default, what other values could the consumer want to set? Im having trouble dreaming up a test case I can use to make sure the positioning is all still correct.

jschuler:
@nicolethoen having the Popover default to document.body is probably good enough most of the time. There are edge cases of content wrapped in shadow-dom or inside an iframe where document.body might not work, e.g. in the PR preview if you change the first example to 
```
<iframe width="100%" src="https://patternfly-react-pr-5160.surge.sh/components/date-picker"></iframe>
```
and then open up the datepicker from within the iframe you'll see. Changing the example inside the iframe to use `appendTo={() => window.parent.document.body}` would help.

nicolethoen (approved):

PR #5161, opened 2020-11-18
fix(datepicker): remove date-fns and allow more customization

### LINKED ISSUE #5142: Can @patternfly/react-datetime DatePicker component be a controlled component?
I am trying to use the date picker component (`@patternfly/react-datetime`) to scroll through a certain range of dates. I have another component that is just a left and right arrow button to go forward or backward a day. From what I have tried and looking through the code, it seems like the date picker can't be controlled and the value attribute for the date picker is just used as a default value. I was wondering if it would be possible to make this a controlled component as well? I made a codesandbox below giving a simple example of how I am using the date picker and then I made a simple native html date picker (`<input type="date"/>`) that can be controlled to show a working example. 

https://codesandbox.io/s/busy-worker-gtzls?file=/src/App.js

Enhancement

@patternfly/react-datetime@lateset

### LINKED ISSUE #5156: Unable to use DatePicker
**Describe the issue. What is the expected and unexpected behavior?** DatePicker is not working for me. I'm trying to use it in the most basic fashion as captured in the documentation, like the below snippet for example. 

`<DatePicker minDate="04/09/1988" maxDate="11/18/2020" onChange={onChange} />`

If I use an early version of the package (`4.1.0`) the component does render as expected and the app styling is without issue, but there is no `onChange` method exposed.

Screen Shot 2020-11-18 at 1 41 01 AM

I then tried to use it as a basic JavaScript component, thinking the type was just wrong, but selecting a date entry still does not fire any onChange event. Seems to be related to https://github.com/patternfly/patternfly-react/issues/5001. Strangely, pressing the space bar with focus in the input fires an event, but it's not usable from an experience standpoint, and the values don't match what the documentation says anyway.

Screen Shot 2020-11-18 at 1 54 22 AM

I then came across https://github.com/patternfly/patternfly-react/pull/5011 which looked hopeful, so I re-tried the package at version `4.2.19`, which is actually what I tried initially, but it breaks the styling across the entire app. I thought I was just doing something wrong, but after quite a bit of back and fourth I finally had to come to a stopping point as I realized there is no test in the jest test suite that exercises `onChange` nor is there a demo of this in `react-integration`, so it seems like it might be more than just a small issue. Here below is an screenshot of how the app styling breaks completely when importing and using the newer version.

Screen Shot 2020-11-18 at 2 36 59 AM

Removing the `@patternfly/react-datetime` package makes the app styling go back to normal.

- Install the DatePicker package
- Try to use it
- Notice the `onChange` does not work
- Notice styling issues that arise when using the latest version on npm (4.2.19 as of time of writing)

I put together an example that illustrates the problem using the patternfly-seed project

https://github.com/seanforyou23/patternfly-react-seed/tree/datepicker-issue-illustration

This is a blocking bug, I have tried everything I can think of but there doesn't seem to be any work-around I can find.

Personal projects and also can be seen in the [patternfly-seed](https://github.com/patternfly/patternfly-react-seed) project.

redallen:
**What**: Remove `date-fns` and amend API to be more flexible given feedback from @nicolethoen and @karelhala .

- Change default formatting/parsing to `yyyy-MM-dd`.
  - Allows relying on default browser formatting/parsing in DatePicker via native date methods like `date.toISOString().substring(0, 10)` and ``new Date(`${val}T00:00:00`)``.
  - Allows CalendarMonth to render any ReactNodes via format functions:
     ```js
     monthFormat = date => date.toLocaleDateString(locale, { month: 'long' }),
     weekdayFormat = date => date.toLocaleDateString(locale, { weekday: 'narrow' }),
     longWeekdayFormat = date => date.toLocaleDateString(locale, { weekday: 'long' }),
     buttonFormat = date => date.getDate()
     ```
      - `locale` defaults to the browser's locale giving decent i18n out of the box. We can force `en-US` to make our Calendar look consistent regardless of language.
- Expose new `weekStart` prop
- Remove `minDate`, `maxDate`, `invalidFormatErrorMessage`, `dateOutOfRangeErrorMessage`, `beforeMinDateErrorMessage`, and `afterEndDateErrorMessage` in favor of `validators?: ((date: Date) => string)[];`. This is more elegant and allows greater flexibility moving forward.

Closes #5142
Closes #5156

**Additional issues**:

### DIFF
--- a/packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
@@ -9,9 +9,17 @@ import styles from '@patternfly/react-styles/css/components/CalendarMonth/calend
 import commonStyles from '@patternfly/react-styles/css/base/patternfly-common';
 import { getUniqueId } from '@patternfly/react-core/dist/js/helpers/util';
 
-export interface CalendarProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange'> {
-  /** Month/year to base other dates around */
-  date: Date;
+export enum Weekday {
+  Sunday = 0,
+  Monday,
+  Tuesday,
+  Wednesday,
+  Thursday,
+  Friday,
+  Saturday
+}
+
+export interface CalendarFormat {
   /** How to format months in Select */
   monthFormat?: (date: Date) => React.ReactNode;
   /** How to format week days in header */
@@ -19,11 +27,16 @@ export interface CalendarProps extends Omit<React.HTMLProps<HTMLDivElement>, 'on
   /** How to format days in header for screen readers */
   longWeekdayFormat?: (date: Date) => React.ReactNode;
   /** How to format days in buttons in table cells */
-  buttonFormat?: (date: Date) => React.ReactNode;
+  dayFormat?: (date: Date) => React.ReactNode;
   /** If using the default formatters which locale to use. Undefined defaults to current locale. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation */
   locale?: string;
   /** Day of week that starts the week. 0 is Sunday, 6 is Saturday. */
-  weekStart?: 0 | 1 | 2 | 3 | 4 | 5 | 6;
+  weekStart?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | Weekday;
+}
+
+export interface CalendarProps extends CalendarFormat, Omit<React.HTMLProps<HTMLDivElement>, 'onChange'> {
+  /** Month/year to base other dates around */
+  date: Date;
   /** Callback when date is selected */
   onChange?: (date: Date) => void;
   /** Functions that returns if a date is valid and selectable */
@@ -65,8 +78,8 @@ export const CalendarMonth = ({
   monthFormat = date => date.toLocaleDateString(locale, { month: 'long' }),
   weekdayFormat = date => date.toLocaleDateString(locale, { weekday: 'narrow' }),
   longWeekdayFormat = date => date.toLocaleDateString(locale, { weekday: 'long' }),
-  buttonFormat = date => date.getDate(),
-  weekStart = 0,
+  dayFormat = date => date.getDate(),
+  weekStart = 0, // Use the American Sunday as a default
   onChange = () => {},
   validators = [() => true],
   className,
@@ -200,7 +213,7 @@ export const CalendarMonth = ({
                 const isFocused = isSameDate(day, focusedDate);
                 const isValid = validators.every(validator => validator(day));
                 const isAdjacentMonth = day.getMonth() !== focusedDate.getMonth();
-                const dayFormatted = buttonFormat(day);
+                const dayFormatted = dayFormat(day);
 
                 return (
                   <td
--- a/packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
@@ -12,13 +12,13 @@ import { getUniqueId } from '@patternfly/react-core/dist/js/helpers/util';
 export interface CalendarProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange'> {
   /** Month/year to base other dates around */
   date: Date;
-  /** How to format months in Select according to date-fns */
+  /** How to format months in Select */
   monthFormat?: (date: Date) => React.ReactNode;
-  /** How to format week days in header according to date-fns */
+  /** How to format week days in header */
   weekdayFormat?: (date: Date) => React.ReactNode;
-  /** How to format days in header for screen readers according to date-fns */
+  /** How to format days in header for screen readers */
   longWeekdayFormat?: (date: Date) => React.ReactNode;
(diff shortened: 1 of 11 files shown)


nicolethoen (commented):
exposing third party libraries is something we'd normally avoid, right?

redallen:
Yes and the reason for that is we don't want consumers depending too heavily on 3rd party deps that we might modify or swap in the future.

In this particular case unless we pick a date string formatting spec and abstract `date-fns` away with our own `format` and `parse` methods I don't think we can modify or swap `date-fns` in the future. I don't foresee a reason to swap or modify it since they support a wide range of strings and locales.

FWIW, we're already "exposing" the Locales from `date-fns`. If we really wanted to not expose this third party library we'd have to allow the consumer to define their own `format` and `parse` functions which also handle locales for both CalendarMonth and DatePicker. I'm not opposed to that given that many older PatternFly consumers use Moment, but it is an annoyance that'd require 2 new props and to hook up an existing or new library.

karelhala (approved):
Hella sweet! I like how you cleaned the `DatePicker`, I know it brings the heavy lifting to consumer, but this feels more robust and straight forward. I'd just change the comments above format props from `{propName} according to date-fns` to for instance `Format month from given Date, used in Select`

packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
```diff
@@ -6,42 +6,41 @@ import ArrowLeftIcon from '@patternfly/react-icons/dist/js/icons/arrow-left-icon
...
+  /** How to format week days in header */
+  weekdayFormat?: (date: Date) => React.ReactNode;
+  /** How to format days in header for screen readers */
+  longWeekdayFormat?: (date: Date) => React.ReactNode;
+  /** How to format days in buttons in table cells */
+  buttonFormat?: (date: Date) => React.ReactNode;
+  /** If using the default formatters which locale to use. Undefined defaults to current locale. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation */
+  locale?: string;
+  /** Day of week that starts the week. 0 is Sunday, 6 is Saturday. */
+  weekStart?: 0 | 1 | 2 | 3 | 4 | 5 | 6;
```
jschuler, line 26: Might be nicer to see `SUN | MON | TUE | WED | THU | FRI | SAT` for less mental mapping

packages/react-datetime/src/components/DatePicker/DatePicker.tsx
```diff
@@ -3,165 +3,108 @@ import { css } from '@patternfly/react-styles';
...
   /** Props to pass to the Popover */
   popoverProps?: Omit<PopoverProps, 'appendTo'>;
+  /** How to format months in Calendar */
+  monthFormat?: (date: Date) => React.ReactNode;
+  /** How to format week days in Calendar */
+  weekdayFormat?: (date: Date) => React.ReactNode;
+  /** How to format days in Calendar */
+  longWeekdayFormat?: (date: Date) => React.ReactNode;
+  /** How to format days in Calendar */
+  buttonFormat?: (date: Date) => React.ReactNode;
```
jschuler, line 47: Same here, `dayFormat`?

jschuler (commented):
CalendarMonth and DatePicker share many props, I think they can extend out a common interface

packages/react-datetime/src/components/DatePicker/DatePicker.tsx
```diff
@@ -3,165 +3,108 @@ import { css } from '@patternfly/react-styles';
...
-  dateFormat = 'MM/dd/yyyy',
+  locale = undefined,
+  dateFormat = (date: Date) => date.toISOString().substring(0, 10),
+  dateParse = (val: string) => new Date(`${val}T00:00:00`),
   isDisabled = false,
-  locale = Locales.enUS,
-  placeholder = 'MM/dd/yyyy',
-  value: valueProp = '',
+  placeholder = 'yyyy-MM-dd',
+  value: valueProp = dateFormat(new Date()),
```
nicolethoen, line 65: I think it should be empty rather than have it default to the current date if no default date is provided.

nicolethoen (commented):
How hard would it be for someone to configure it so that the date is more readable like most users are accustomed? I think we should be sure that it's very easy, since it will be very common for people to expect the date format they see most often.

packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
redallen: I've added a `Weekday` enum.

packages/react-datetime/src/components/DatePicker/DatePicker.tsx
redallen: Done.

redallen:
@jschuler I made a common "CalendarFormat" interface that appears in the docs, renamed `buttonFormat` -> `dayFormat`, and added a `Weekday` enum.

@nicolethoen I added an "American" example where I specify a simple `dateFormat` and `dateParse` using native APIs that works for 4-digit years. I made the default value an empty string.

jschuler (approved):

nicolethoen:
nit: can you make maybe the basic date picker example - or at least one of the examples - not have a default date? :)

jschuler (approved):

nicolethoen (approved):

PR #5166, opened 2020-11-19
feat(react-core): Update react-core copyStyles.js

sjd78:
Update react-core's `copyStyles.js` generate script to generate both the pure PF4 `base.css` file and the PF3+PF4 no reset
`base-no-reset.css` file.  This will allow a project using both PF3-react and PF4-react on the same page without resorting to using sass or falling back to PF4's `patternfly-no-reset.css`.

Before this PR, if a project uses PF3 and PF4, and doesn't want to adopt sass, the css imports would be:
```jsx
import 'patternfly-react/dist/css/patternfly-react.css' // PF3-react
import '@patternfly/patternfly/patternfly-no-reset.css' // PF4
```

After this PR is merged, projects just using PF4 can still import the `base.css` normally:
```jsx
import '@patternfly/react-core/dist/style/base.css'
```

But, now projects using PF3 and PF4 can use the new react-core `base-no-reset.css` that includes the optimizations of the current `base.css`:
```jsx
import 'patternfly-react/dist/css/patternfly-react.css' // PF3-react
import '@patternfly/react-core/dist/style/base-no-reset.css' // PF4-react
```

### DIFF
--- a/packages/react-core/scripts/copyStyles.js
+++ b/packages/react-core/scripts/copyStyles.js
@@ -4,37 +4,15 @@ const { resolve, dirname, join } = require('path');
 const { parse: parseCSS, stringify: stringifyCSS } = require('css');
 /* eslint-enable @typescript-eslint/no-var-requires */
 
-const baseCSSFilename = 'patternfly-base.css';
 const stylesDir = resolve(__dirname, '../dist/styles');
-const pfDir = dirname(require.resolve(`@patternfly/patternfly/${baseCSSFilename}`));
-
-const css = readFileSync(join(pfDir, baseCSSFilename), 'utf8');
-const ast = parseCSS(css);
+const pfDir = dirname(require.resolve('@patternfly/patternfly/patternfly.css'));
 
 const unusedSelectorRegEx = /(\.fas?|\.sr-only)/;
 const unusedKeyFramesRegEx = /fa-/;
 const unusedFontFamilyRegEx = /Font Awesome 5 Free/;
 const ununsedFontFilesRegExt = /(fa-|\.html$|\.css$)/;
 
-// Core provides font awesome fonts and utlities. React does not use these
-ast.stylesheet.rules = ast.stylesheet.rules.filter(rule => {
-  switch (rule.type) {
-    case 'rule':
-      return !rule.selectors.some(sel => unusedSelectorRegEx.test(sel));
-    case 'keyframes':
-      return !unusedKeyFramesRegEx.test(rule.name);
-    case 'charset':
-    case 'comment':
-      return false;
-    case 'font-face':
-      // eslint-disable-next-line no-case-declarations
-      const fontFamilyDecl = rule.declarations.find(decl => decl.property === 'font-family');
-      return !unusedFontFamilyRegEx.test(fontFamilyDecl.value);
-    default:
-      return true;
-  }
-});
-
+// Copy assets
 copySync(join(pfDir, 'assets/images'), join(stylesDir, 'assets/images'));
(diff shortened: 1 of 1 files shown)


redallen:
Unfortunately `@patternfly/patternfly/patternfly-no-reset.css` is not a "base" 40Kb stylesheet like `@patternfly/patternfly/patternfly-base.css` but rather 850Kb of every PatternFly component. This means after this PR if you write:
```
import '@patternfly/react-core/dist/style/base-no-reset.css' // PF4-react
import { Button } '@patternfly/react-core'
```
you'd get styles for ALL components on the first line and `Button.css` styles a second time on the second line.

@mcoker Can we make a `patternfly-base-no-reset.css` upstream?

mcoker:
@redallen sure! https://github.com/patternfly/patternfly/issues/3735

dlabrecq (commented):
I prefer not to have any PF3 styles in order to keep is as small as possible for my app.

redallen:
@dlabrecq This doesn't add PF3 styles. It just reexports the `no-reset` styles from Core which don't override existing PF3 styles.

sjd78:
@redallen - To add to the fun, at least on ovirt, extensions are loaded into an iframe.  We do a React portal to the parent frame to render the modal.  Part of the portal element creation is pushing the css to the parent frame so the Modal gets styled properly.  The only way I've found to make sure everything PF4 gets styled properly is to use the full style sheet.

Unless you have some ideas on how to let the dynamic style loading the components do reflect in the iframe's parent?

packages/react-core/scripts/copyStyles.js
```diff
@@ -4,42 +4,51 @@ const { resolve, dirname, join } = require('path');
...
   filter(src) {
     return !ununsedFontFilesRegExt.test(src);
   }
 });
-writeFileSync(join(stylesDir, 'base.css'), stringifyCSS(ast));
+
+// Copy css
+const baseCssFiles = {
+  'base.css': 'patternfly-base.css',
+  'base-no-reset.css': 'patternfly-no-reset.css'
```
redallen, line 27: Once #5219 is in just change this to `patternfly-base-no-reset.css`.

redallen (changes requested):
If you need the whole stylesheet your existing `import '@patternfly/patternfly/patternfly-no-reset.css'` works well and I'd recommend [using `null-loader` on `react-styles/css/**/*.css`](https://github.com/patternfly/patternfly-org/blob/master/packages/theme-patternfly-org/scripts/webpack/webpack.client.config.js#L82) to only load CSS once.

The reexporting base styles from react-core is just for convenience using our current "only load the CSS you need" model and we aren't aiming to reexport all of `@patternfly/patternfly`. I still find it useful to reexport a `base-no-reset` for consumers transitioning from PF3 that want to use this model.

sjd78:
> If you need the whole stylesheet your existing `import '@patternfly/patternfly/patternfly-no-reset.css'` works well and I'd recommend [using `null-loader` on `react-styles/css/**/*.css`](https://github.com/patternfly/patternfly-org/blob/master/packages/theme-patternfly-org/scripts/webpack/webpack.client.config.js#L82) to only load CSS once.

Oh that's a good idea!

> 
> The reexporting base styles from react-core is just for convenience using our current "only load the CSS you need" model and we aren't aiming to reexport all of `@patternfly/patternfly`. I still find it useful to reexport a `base-no-reset` for consumers transitioning from PF3 that want to use this model.

Will do. Hopefully it'll be useful for us on ovirt-web-ui (which is mostly PF3) soon(ish).

redallen (approved):
LGTM, thanks @sjd78 !

evwilkin (approved):
LGTM

Issue #5171, opened 2020-11-20
Form component demo needs tidying


seanforyou23:
**Describe the issue. What is the expected and unexpected behavior?**
There are a few small issues with the Form component documentation/demo page. The first thing is that the Popover for the help icon launches when the enter key is pressed for any of the fields. Seems the event bubbles and eventually triggers the popover to launch unexpectedly. Next is that the whitespace within the popover is collapsing which causes some of the words to run together.

<img width="400" alt="Screen Shot 2020-11-19 at 9 58 56 PM" src="https://user-images.githubusercontent.com/5942899/99752810-14b74600-2ab3-11eb-9e6d-27e2626ff30e.png">

Last thing is that with the "Limit width" form uses the same id's as the simple form, which can cause issues.

<img width="400" alt="Screen Shot 2020-11-19 at 9 33 06 PM" src="https://user-images.githubusercontent.com/5942899/99752929-4b8d5c00-2ab3-11eb-89f2-ea63042e2dd5.png">

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

- Navigate to the email field in the [Form component demo page](https://www.patternfly.org/v4/components/form/)
- Type "asdf"
- Press the enter key
- Notice the Popover is launched

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug

**What is your product and what release version are you targeting?**
Ongoing PF development

PR #5172, opened 2020-11-20
fix(form-demo): tidy form demo examples

seanforyou23:
**What**: This PR updates the Form component demo page to fix a small issue in the demo where event bubbling causes the Popover to launch unexpectedly when pressing enter key while focus is on certain form controls within the form by adding `type="button"` to the button element that launches the Popover. It also adds some missing whitespace so certain words are not jumbled, and finally, updates id's to be unique across multiple form examples to clear the console warnings.

Screen Shot 2020-11-19 at 10 07 57 PM

Should help close https://github.com/patternfly/patternfly-react/issues/5171

@jschuler ping - and thanks for the help nailing down that Popover issue!

### DIFF
--- a/packages/react-core/src/components/Form/examples/Form.md
+++ b/packages/react-core/src/components/Form/examples/Form.md
@@ -67,7 +67,7 @@ class SimpleForm extends React.Component {
             <Popover
               headerContent={
                 <div>
-                  The <a href="https://schema.org/name">name</a> of a <a href="https://schema.org/Person">Person</a>
+                  The{' '}<a href="https://schema.org/name" target="_blank">name</a>{' '}of a{' '}<a href="https://schema.org/Person" target="_blank">Person</a>
                 </div>
               }
               bodyContent={
@@ -85,9 +85,10 @@ class SimpleForm extends React.Component {
               }
             >
               <button
+                type="button"
                 aria-label="More info for name field"
                 onClick={e => e.preventDefault()}
-                aria-describedby="simple-form-name"
+                aria-describedby="simple-form-name-01"
                 className="pf-c-form__group-label-help"
               >
                 <HelpIcon noVerticalAlign />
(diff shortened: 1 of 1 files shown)


redallen (approved):

packages/react-core/src/components/Form/examples/Form.md
```diff
@@ -67,7 +67,7 @@ class SimpleForm extends React.Component {
             <Popover
               headerContent={
                 <div>
-                  The <a href="https://schema.org/name">name</a> of a <a href="https://schema.org/Person">Person</a>
+                  The{' '}<a href="https://schema.org/name" target="_blank">name</a>{' '}of a{' '}<a href="https://schema.org/Person" target="_blank">Person</a>
```
jschuler, line 70: seems our documentation is stripping out these whitespaces? I don't see that on codesandbox @redallen

jschuler (approved):

packages/react-core/src/components/Form/examples/Form.md
redallen: I see this in our docs:
```jsx
headerContent={
  <div>
    The{' '}<a href="https://schema.org/name" target="_blank">name</a>{' '}of a{' '}<a href="https://schema.org/Person" target="_blank">Person</a>
  </div>
}
```
which matches codesandbox:
```jsx
headerContent={
  <div>
    The{' '}<a href="https://schema.org/name" target="_blank">name</a>{' '}of a{' '}<a href="https://schema.org/Person" target="_blank">Person</a>
  </div>
}
```

What's the discrepancy @jschuler ?

jschuler: This PR is adding `{' '}` to have spaces show in the headerContent, because without it the space is ignored in the PF docs. But even without them the space should be visible (which it is if you look at codesandbox for example)

redallen: Ah you're right. I'll see if I can fix this upstream.

Issue #5132, opened 2020-11-12
Popover not dismissing on click of an action button inside the popover


darrenrevans10:
**Describe the issue. What is the expected and unexpected behavior?**

The Docs state that...
Popovers are triggered by a click and dismissed in one of the following ways:

The user clicks the exit icon
**The user clicks an action button inside the popover**
The user clicks anywhere on the screen outside the popover

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
You can see this unexpected behavior here, https://www.patternfly.org/v4/components/popover/react. The first example shows a button within the popover that doesn't automatically dismiss the popover.

jschuler:
@mcarrano is this guideline still accurate? If so do we differentiate between types of action buttons? For example the datepicker has buttons but you wouldn't expect the popover to auto-close. Should there be a special type of button that can be used within a Popover to make it auto-close?

mcarrano:
@jschuler I agree that clicking on an action should not always close the popover. But I think it should also be possible to include a button besides the close button that will trigger an action and close the popover, like a modal. Is this feasible? @mmenestr anything to add? I believe you authored the guideline document.

mmenestr:
Yes, I can definitely change it according to what we decide here!
When writing that guideline I was referring to actions like "Learn more" that would link you out to external documentation for more information, and assumed that clicking on it would close the popover. But let me know if that's not the case, I don't feel too strongly about this!

Also, could you give me more background on the use case where you would have a date picker inside a popover? Like what would the scenario be?

mcarrano:
> Also, could you give me more background on the use case where you would have a date picker inside a popover? Like what would the scenario be?

@mmenestr I think this was just referencing the fact that the date picker calendar uses the popover as a container for the calendar. It's all integrated as part of the date picker and not something a designer using the date picker needs to be concerned with.

jschuler:
@mcarrano @mmenestr thanks, sounds like we need to provide some way to add elements into the popover that can auto-close it. I'll take a look at this and see what we can do in that regard!

jschuler:
@mmenestr maybe the docs could be updated to reflect that actions buttons may close the popover? (not always)

PR #5173, opened 2020-11-20
Popover - close from content

jschuler:
Can close the popover from within the content
**What**: Closes https://github.com/patternfly/patternfly-react/issues/5132


evwilkin (approved):
LGTM

nicolethoen:
Looks good, should there be tests - probably at least cypress tests?

jessiehuff (commented):
For the most part it looks pretty good, but it looks like I can't use the "Without header/footer/close and no padding" example with a keyboard. When I try to press enter I get the following error: "Error: You can't have a focus-trap without at least one focusable element"
Also, it's a little odd in VO because you have to use the enter key instead of the typical cmd + opt + space, but that is likely related to this [issue](https://github.com/patternfly/patternfly-react/issues/4210).

redallen (changes requested):
I'm a bit confused -- don't we already have the `isVisible` prop to do exactly this? I'd imagine this should be implemented something like:
```jsx
const Basic = () => {
  const [isVisible, setIsVisible] = React.useState(false);
  return (
    <div style={{ margin: "100px" }}>
      <Popover
        headerContent="Header"
        bodyContent={<Button onClick={() => setIsVisible(false)}>Close</Button>}
        footerContent="Popover Footer"
        isVisible={isVisible}
        position="bottom"
      >
        <Button onClick={() => setIsVisible(!isVisible)}>Toggle Popover</Button>
      </Popover>
    </div>
  );
};
```

jschuler:
@redallen it's more convenient that way if you don't already control the Popover with the `isVisible` prop

redallen:
Right, but if you do control it with the `isVisible` prop you can get into a confusing state where `isVisible={true}` but the Popover is hidden because you called `hide`. Rerendering your component would cause the Popover to be visible again.

To me this is a strange API to avoid setting `isVisible` which isn't all that hard!

jschuler:
@redallen it's just providing one more way to tap into the visibility state handled by the component itself. I agree it's not that hard to control the visibility manually, but you also have to listen to when the Close icon is pressed, and if a user clicks outside the popover. When the component handles the state it already does all that for you.

redallen:
> but you also have to listen to when the Close icon is pressed, and if a user clicks outside the popover.

Ah yes, I forgot that part. However it's as simple as adding `shouldClose={() => setIsVisible(false)}`.

I believe as a consumer you should either control the `isVisible` state entirely yourself or not at all. This API is a hook into partially controlling `isVisible` yourself which can lead to two different sources of truth:
```js
import React from 'react';
import { Popover, PopoverPosition, Button } from '@patternfly/react-core';

ClosePopoverFromContent = () => {
  const [isVisible, setIsVisible] = React.useState(false);
  return (
    <div style={{ margin: '50px' }}>
      <pre>isVisible: {isVisible.toString()}</pre>
      <Popover
        aria-label="Popover with button in the body that can close it"
        headerContent={<div>Popover header</div>}
        bodyContent={hide => (
          <div>
            <div>
              All the content props (headerContent, bodyContent, footerContent) can take a function which the Popover
              component passes the hide function to which can be used to close the Popover after some user interaction.
            </div>
            <div>
              <Button onClick={hide}>Close popover</Button>
            </div>
          </div>
        )}
        footerContent="Popover footer"
        isVisible={isVisible}
      >
        <Button onClick={() => setIsVisible(!isVisible)}>Toggle Popover</Button>
      </Popover>
    </div>
  );
}
```

Here when you use the `hide` function in bodyContent you'll end up with with `isVisible=true` but the Popover being hidden! That means you have to remember to keep your `isVisible` state in-sync with `Popover`'s `isVisible`.

jschuler:
@redallen That is correct, if you do end up mixing `isVisible` with using the render prop function you need to keep them in sync. But I imagine this wouldn't be a regular occurrence. You'd usually either fully control visibility yourself, or let it handle state itself.
Imagine the case where you have to render multiple Popovers (for example in table rows) and some action within the content should also close it. Handing state for each one will become more complex. If you can let the component handle it it will be so much easier.

redallen:
> You'd usually either fully control visibility yourself, or let it handle state itself.

Maybe in most cases, but with this PR we wouldn't be enforcing that.

> Imagine the case where you have to render multiple Popovers (for example in table rows) and some action within the content should also close it. Handing state for each one will become more complex. If you can let the component handle it it will be so much easier.

Since you can only have one Popover at a time I think it'd be pretty simple to just include a single state variable, no?
```jsx
const [popoverOpen, setPopoverOpen] = React.useState(-1);
return items.map((cell, index) =>
  <Popover
    key={index}
    isVisible={popoverOpen === index}
    shouldClose={() => setPopoverOpen(-1)}
    // Is a single source of truth better than a fancy close function as part of render?
    bodyContent={<Button onClick={() => setIsVisible(false)}>Close</Button>}
  >
    <div onClick={() => setPopoverOpen(index)}>
      Something that has a popover and you control what triggers and closes it!
    </div>
  </Popover>
);
```

vs

```jsx
return items.map((cell, index) =>
  <Popover
    key={index}
    // Or is a fancy render function worth not using `isVisible`?
    bodyContent={hide => <Button onClick={() => hide()}>Close</Button>}
  >
    <div>
      Something that has a popover and you just want some content to close it!
    </div>
  </Popover>
);
```

Looking at the 2 APIs, I don't find the latter much easier than the former. I like that the former lets me control exactly how the Popover is opened/closed.

If it helps with our own implementation of Table I suppose the former is fine, but I fear the precedent it sets of multiple sources of truth for state. Dealing with state in just your own app is hard enough, let alone a component library.

jschuler:
@redallen i don't really see the problem here, it's an API addition, you can use it or not. It does simplify usage if you don't want to manually control the visibility state. Examples show usage.

redallen:
The problem I see is that we're creating a _partially_ controlled API through a render function. This is an unusual precedent to set that allows undefined behavior if used with the `isVisible` prop. I also personally don't like passing a callback that controls state to a render function since we haven't done that anywhere else that I know of in PatternFly.

If you put a disclaimer not to use the `isVisible` prop in conjuction with the new `((hide: () => void) => React.ReactNode)` API in the example caption I'll still approve, though.

packages/react-core/src/components/Popover/examples/Popover.md
```diff
@@ -25,7 +25,43 @@ import { Popover, PopoverPosition, Button } from '@patternfly/react-core';
...
+        footerContent="Popover footer"
+      >
+        <Button>Toggle Popover</Button>
+      </Popover>
+    </div>
+  );
+};
+```
+
+### Close popover from content (uncontrolled)
```
redallen, line 65: Add warning like: "When using the `hide` callback with `isVisible` be sure to keep the two in-sync."

redallen (changes requested):

redallen (approved):
LGTM, thanks @jschuler !

nicolethoen (approved):

PR #5178, opened 2020-11-23
fix(react-charts): bump victory and move victory dep to demo app

redallen:
**What**: Towards https://github.com/openshift/console/pull/7287

**Additional issues**:


jcaianirh (approved):
lgtm thanks @redallen

jschuler (approved):

dlabrecq (commented):
Should we fix the import in chartLine.md?

import { VictoryZoomContainer } from 'victory';

PR #5184, opened 2020-11-24
feat(date-picker): bugfixes and add range picker demo

### LINKED ISSUE #4732: Build Date Range picker

redallen:
**What**: Closes #4732

Bugfixes:
- Remove `'@patternfly/patternfly/patternfly-date-picker.css'` left over from `flatpickr` implementation and remove unused `@patternfly/patternfly` dependency
- Allow `value={null}` or other invalid dates in DatePicker and CalendarMonth
- Memoize weeks, days, and validity of days to only recompute when the year, month, or weekStart changes. Scrolling between months and years is noticeably faster.
- Only validate DatePicker `onBlur` per @mcarrano's suggestion
  - Don't validate if field was untouched per @nicolethoen's suggestion
- Clone dates before passing them to consumers (before consumers could mutate component state since dates are mutable)
- Add `ariaLabel` props to `CalendarMonth` per @nicolethoen 

Demo https://patternfly-react-pr-5184.surge.sh/components/date-picker/react-demos
- Expose `isValidDate` helper
- Add hovered date support to CalendarMonth for the range picker

**Additional issues**:

### DIFF
--- a/packages/react-datetime/package.json
+++ b/packages/react-datetime/package.json
@@ -35,7 +35,6 @@
     "clean": "rimraf dist"
   },
   "dependencies": {
-    "@patternfly/patternfly": "4.65.5",
     "@patternfly/react-core": "^4.79.4",
     "@patternfly/react-icons": "^4.7.18",
     "@patternfly/react-styles": "^4.7.16",
--- a/packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-datetime/src/components/CalendarMonth/CalendarMonth.tsx
@@ -80,8 +80,10 @@ const isSameDate = (d1: Date, d2: Date) =>
 
 export const isValidDate = (date: Date) => Boolean(date && !isNaN(date as any));
 
+const today = new Date();
+
 export const CalendarMonth = ({
-  date: dateProp = new Date(),
+  date: dateProp = today,
   locale = undefined,
   monthFormat = date => date.toLocaleDateString(locale, { month: 'long' }),
   weekdayFormat = date => date.toLocaleDateString(locale, { weekday: 'narrow' }),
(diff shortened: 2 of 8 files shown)


jessiehuff (commented):
Looks pretty good to me! Still works with keyboard, and it's pretty solid in VO. The only enhancement I'd make is changing the `aria-label` on the input fields from "Date picker" on both: `<input type="text" aria-label="Date picker" placeholder="yyyy-MM-dd" class="pf-c-form-control" value="1993-01-01" aria-invalid="false">`
to "Start date" and "End date". That's the model that Google calendar follows, and I think that makes the most sense. 🙂

nicolethoen (commented):
First thought - can we allow for custom aria labels throughout the component for localization purposes?

nicolethoen:
Also, when I focus on the first date input in the range picker demo, then tab out without typing a date, I get an invalid date error. I dont think that should happen

redallen:
@nicolethoen In DatePicker yes but in CalendarMonth no. I'll add a slew of props now.

redallen:
@nicolethoen I've added pristine state tracking as well. It will only show error text on blur if you've typed into it.

nicolethoen:
idk if this is more a css thing, but It doesn't look very pretty to me when I have an invalid date error message:
can we keep both pickers in the range on the same line? even if one has an error message?
Screen Shot 2020-11-30 at 1 50 51 PM

jessiehuff (approved):
It looks great from an accessibility standpoint. 👍 I see what Nicole means on the invalid date error though.

redallen:
@nicolethoen I've been working since last week with @mcoker to make a layout that addresses that issue.

mcoker:
If I pick a start date, then open the end date calendar and go to the next month, I end up with this - seems like the `pf-m-end-range` class isn't being applied correctly. It fixes itself if I hover over the cell.

Screen Shot 2020-12-01 at 9 43 28 AM

nicolethoen (approved):
If they don't already exist, could you be sure to create issues to add cypress and unit tests for the calendar month and datepicker components?

redallen:
Good point @mcoker . What should the calendar look like after clicking the previous/next month arrows from the following state @mcarrano ?
image

jessiehuff (dismissed):
Looks good from the a11y side! 🙂

mcarrano (approved):
Looks good to me. Thanks @redallen !

mcoker (approved):
lgtm!

PR #5189, opened 2020-11-29
feat(breadcrumb): add render prop to breadcrumb item

### LINKED ISSUE #5121: BreadcrumbItem issue with React Router Links
`<BreadcrumbItem>` doesn't work well with [React Router Links](https://reactrouter.com/web/api/Link). There's three issues that makes these two not work well together

1. Using `<BreadcrumbItem to="/foobar" component={Link}>Foobar</BreadcrumbItem>` doesn't generate a link as BreadcrumbItem adds to `prop` as `href` prop to`<Link>` - which it doesn't have
2. Using `<BreadcrumbItem><Link to="/foobar">Foobar</Link></BreadcrumbItem>` means `BreadcrumbItem` doesn't apply correct classes as `to` prop is not provided
3. Removing `BreadcrumbItem` and using class names like such `<li className="pf-c-breadcrumb__item"><Link to="/foobar" className="pf-c-breadcrumb__link">Foobar</Link>` doesn't work well either as `Breadcrumb` component adds `showDivider` prop to its children

There's one possible fix for this, a wrapper for Link that moves `href` to the `to` prop
```js
const MapLink: React.FC = ({href, target, children, ...props}) => (
  <Link to={href} {...props}>{children}</Link>
);
```

@jschuler suggested to update the `BreadcrumbItem` component with a render prop

Enhancement

Telltale; Any version

boaz0:
**What**: Closes #5121 

//cc @SpyTec

### DIFF
--- a/packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
+++ b/packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
@@ -3,6 +3,11 @@ import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-ic
 import styles from '@patternfly/react-styles/css/components/Breadcrumb/breadcrumb';
 import { css } from '@patternfly/react-styles';
 
+export interface BreadcrumbItemRenderArgs {
+  className: string;
+  ariaCurrent: 'page' | undefined;
+}
+
 export interface BreadcrumbItemProps extends React.HTMLProps<HTMLLIElement> {
   /** Content rendered inside the breadcrumb item. */
   children?: React.ReactNode;
@@ -18,33 +23,34 @@ export interface BreadcrumbItemProps extends React.HTMLProps<HTMLLIElement> {
   target?: string;
   /** Sets the base component to render. Defaults to <a> */
   component?: React.ElementType;
+  /** A render function to render the component inside the breadcrumb item. */
+  render?: (props: BreadcrumbItemRenderArgs) => React.ReactNode;
 }
 
 export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
   children = null,
-  className = '',
+  className: classNameProp = '',
   to = null,
   isActive = false,
   showDivider,
   target = null,
   component = 'a',
+  render = null,
   ...props
 }: BreadcrumbItemProps) => {
   const Component = component;
+  const ariaCurrent = isActive ? 'page' : undefined;
+  const className = css(styles.breadcrumbLink, isActive && styles.modifiers.current);
   return (
-    <li {...props} className={css(styles.breadcrumbItem, className)}>
+    <li {...props} className={css(styles.breadcrumbItem, classNameProp)}>
       {showDivider && (
         <span className={styles.breadcrumbItemDivider}>
           <AngleRightIcon />
         </span>
       )}
-      {to && (
-        <Component
-          href={to}
-          target={target}
-          className={css(styles.breadcrumbLink, isActive && styles.modifiers.current)}
-          aria-current={isActive ? 'page' : undefined}
-        >
+      {render && render({ className, ariaCurrent })}
+      {to && !render && (
+        <Component href={to} target={target} className={className} aria-current={ariaCurrent}>
           {children}
         </Component>
       )}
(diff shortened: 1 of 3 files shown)


SpyTec (dismissed):
LGTM! Though `showDivider` and `isActive` props won't be visible in type-definitions for the render function

boaz0:
@SpyTec what do you mean? :smile: Am I missing something?

SpyTec:
Yes, that would help

tlabaj (commented):
Can you please update the demo-app and integration tests.

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -18,6 +26,8 @@ export interface BreadcrumbItemProps extends React.HTMLProps<HTMLLIElement> {
   target?: string;
   /** Sets the base component to render. Defaults to <a> */
   component?: React.ElementType;
+  /** A render function to render the component inside the breadcrumb item. */
+  render?: (props: BreadcrumbItemRenderProps) => JSX.Element;
```
tlabaj, line 30: This is inconsistent with how we defined render prop in the Label component.
@jshuler any thoughts on this vs the other implementation?

tlabaj (changes requested):

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
boaz0: No problem, I can change the return value to `React.ReactNode`. 
If there are other props that are missing or redundant I can also update as well.

SpyTec: Looking at this SO it seems ReactNode would be more accurate for a render function

https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -38,7 +49,8 @@ export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
           <AngleRightIcon />
         </span>
       )}
-      {to && (
+      {to && render && render({ href: to, children, target, isActive, showDivider, ...props })}
```
jschuler, line 52: I think it would be more useful to use the render prop to return things like the PF classname and some aria attributes, like this:
```
<BreadcrumbItem isActive render={({ className, ariaCurrent }) => {
  <Link to="your-link" className={className} aria-current={ariaCurrent}>Router item</Link>
}} />
```

and in the code (but you'd want to hold classname etc in variables so it is not duplicated across the code)
```
render && render({ className: css(styles.breadcrumbLink, isActive && styles.modifiers.current), ariaCurrent: isActive ? 'page' : undefined,...props })}
```

boaz0:
@tlabaj & @jschuler I updated the PR and added a test to the specs. Mind reviewing it again?
Thanks!

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -18,33 +27,34 @@ export interface BreadcrumbItemProps extends React.HTMLProps<HTMLLIElement> {
...
         </span>
       )}
-      {to && (
-        <Component
-          href={to}
-          target={target}
-          className={css(styles.breadcrumbLink, isActive && styles.modifiers.current)}
-          aria-current={isActive ? 'page' : undefined}
-        >
+      {to && render && render({ className, ariaCurrent, to, isActive, target, showDivider, ...props })}
```
jschuler, line 55: thanks, this is almost good, but I don't think we need to require that the user pass `to` to use the render prop. All of those props (`to`, `isActive`, etc.) just get mirrored back to the user and doesn't provide more information. IMO the only useful props are className and ariaCurrent since we return something that the user didn't have to begin with.

jschuler (approved):
awesome, thanks @boaz0

tlabaj (approved):

boaz0:
Thank you for your feedback @jschuler & @tlabaj and thank you @SpyTec for suggesting this idea.

PR #5193, opened 2020-11-30
feat(Drawer): add splitter feature

### LINKED ISSUE #5050: Add Drawer and MasterDetails demos(?)/variants(?) which incorporate a sliding splitter
Create interactive demos that show usage of the Splitter component. A core issue has been opened here to build the required layouts: https://github.com/patternfly/patternfly/issues/3610

kmcfaul:
**What**: Closes #5050 

To enable resize, `isResizable` is passed to the `DrawerPanelContent` component.

Other additions:
- `minSize` / `maxSize` properties to customize resize limits. Defaults to the bounds of the parent container.
- `increment` property to customize keyboard size jump per keystroke. Defaults to `5px`.
- `resizeAriaLabel` / `resizeAriaDescribedBy` properties to customize labelling of the splitter div.

Keyboard interaction:
Tab over to the splitter and press `space` to begin resizing using the arrow keys. Press `enter` or `escape` to stop resizing.

@mcoker I've implemented the resizing by overriding the `--pf-c-drawer__panel--FlexBasis` css variable, since it appears height/width won't change without it. Is that all right or would there be a better class/css variable? 
There's also an odd effect when resizing the left drawer, where the drawer contents look like they drag a little. This is because `margin-right` has to be updated in sync with the size, so I'm not sure if there's anything I can do for that.

Direct link: https://patternfly-react-pr-5193.surge.sh/components/drawer#resizable-on-right

### DIFF
--- a/packages/react-core/src/components/Drawer/Drawer.tsx
+++ b/packages/react-core/src/components/Drawer/Drawer.tsx
@@ -23,12 +23,14 @@ export interface DrawerContextProps {
   isExpanded: boolean;
   isStatic: boolean;
   onExpand?: () => void;
+  position?: string;
 }
 
 export const DrawerContext = React.createContext<Partial<DrawerContextProps>>({
   isExpanded: false,
   isStatic: false,
-  onExpand: () => {}
+  onExpand: () => {},
+  position: 'right'
 });
 
 export const Drawer: React.SFC<DrawerProps> = ({
@@ -41,7 +43,7 @@ export const Drawer: React.SFC<DrawerProps> = ({
   onExpand = () => {},
   ...props
 }: DrawerProps) => (
-  <DrawerContext.Provider value={{ isExpanded, isStatic, onExpand }}>
+  <DrawerContext.Provider value={{ isExpanded, isStatic, onExpand, position }}>
     <div
       className={css(
         styles.drawer,
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -11,6 +11,18 @@ export interface DrawerPanelContentProps extends React.HTMLProps<HTMLDivElement>
   children?: React.ReactNode;
   /** Flag indicating that the drawer panel should not have a border. */
   hasNoBorder?: boolean;
+  /** Flag indicating that the drawer panel should be resizable. */
+  isResizable?: boolean;
+  /** The minimum size of a resizable drawer, in pixels. Defaults to the starting width of the drawer. */
+  minSize?: number;
+  /** The maximum size of a resizable drawer, in pixels. Defaults to the max width of the parent container. */
+  maxSize?: number;
+  /** The increment amount for keyboard drawer resizing, in pixels. */
+  increment?: number;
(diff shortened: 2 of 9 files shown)


nicolethoen (commented):
This is awesome!!

What is the reasoning for the 'space' to be what allows the arrow keys to move the splitter? Do they arrow keys have another function if the browser focus is on the splitter? Could the arrow keys just work without needing to hit space?

I also think if possible, the interaction/callbacks with the splitter should be tested in the integration tests

kmcfaul:
Having the arrow keys control the resizing directly without having to enter/exit the interaction would work. I was mostly following what we've done for other interaction.

mcoker:
@kmcfaul the flex-basis var is perfect, and just merged a fix for the left drawer content shifting https://github.com/patternfly/patternfly/pull/3719

tlabaj (commented):
This looks great Katie!  I agree with Nicole, it would be nice to test some of the interactions in the cypress tests.

tlabaj (approved):
LGTM

nicolethoen (approved):

mcarrano (approved):
Looks great @kmcfaul !

mattnolting (approved):
LPTM!

dlabrecq (approved):

mcoker (approved):
Very nice!

jessiehuff (commented):
This looks really good so far. I find the keyboard interaction somewhat unexpected though. I would expect it to follow [these keyboard interactions](https://www.w3.org/TR/wai-aria-practices-1.1/#windowsplitter). A keyboard user likely wouldn't think to press space before using a splitter, they would likely expect to tab to it then press the arrow keys left/right for vertical splitters and up/down for horizontal splitters similar to this [demo](https://codepen.io/captainKAIO/full/yLYPPxX). I also would think that it would follow the pattern that when enter is pressed, the drawer would close. 

When using `aria-describedby`, it's typical to pass it an id that points to a description like [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute). So I'd recommend updating the keyboard interaction, dropping the `aria-describedby` and adding the aria properties listed in the [aria authoring practices](https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-25): 
- The element that serves as the focusable splitter has role separator.
- The separator element has the `aria-valuenow` property set to a decimal value representing the current position of the separator.
- The separator element has the `aria-valuemin` property set to a decimal value that represents the position where the primary pane has its minimum size. This is typically 0.
- The separator element has the `aria-valuemax` property set to a decimal value that represents the position where the primary pane has its maximum size. This is typically 100.
- If the primary pane has a visible label, it is referenced by `aria-labelledby` on the separator element. Otherwise, the separator element has a label provided by `aria-label`.
- The separator element has `aria-controls` referring to the primary pane.

I can't seem to use it with VO, but I think that's because it's not quite the expected keyboard and aria. I don't want to hold it back so if we want to complete this in a follow up issue, I'm ok with that, but we should update these when we can. 🙂

tlabaj:
@jessiehuff  Let's open a follow up issue.  If you are ok with that we can merge this.

Issue #5181, opened 2020-11-23
Add a prop to Switch component to let user choose whether to show CheckIcon or not


DaoDaoNoCode:
**Describe the issue. What is the expected and unexpected behavior?**
According to issue #5174 , the Switch design in OpenShift has the CheckIcon when switch button is checked.
<img width="105" alt="Screen Shot 2020-11-23 at 4 58 21 PM" src="https://user-images.githubusercontent.com/37624318/100020091-3b27fa80-2dad-11eb-8c35-f2f38e88316c.png">
However, when we choose to show label in the Switch component, there is no option for us to show the CheckIcon on the switch button. Only if we use the switch button with no label it will show the CheckIcon. Should we add a prop like `checkedIcon` for users to decide whether they would like to show the icon?

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Enhancement.

PR #5196, opened 2020-11-30
fix(Toolbar): add log viewer toolbar example

### LINKED ISSUE #5174: Toolbar POC: create a new toolbar example to support a console log viewer toolbar
Create a new example to support a console log viewer toolbar. Provide both a desktop and compressed version. The following design provides context into the details of the toolbar requirements:
http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/log%20toolbars/

Enhancement
OpenShift v4.8

DaoDaoNoCode:
**What**: Closes #5174 .
Here I create a new example for console log viewer toolbar, based on the design here http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/log%20toolbars/.


packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -390,6 +393,272 @@ class ToolbarGroupTypes extends React.Component {
...
+        </ToolbarItem>
+        <ToolbarItem>
+          <Switch
+            id="switch-2"
+            label="Wrap lines"
+            isChecked={secondSwitchChecked}
+            onChange={this.handleSecondSwitchChange}
+          />
+        </ToolbarItem>
+        <ToolbarItem>
```
jcaianirh, line 598: According to the design: http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/log%20toolbars/

The wrapped lines marks the end of the first group of components on the toolbar which are left justified, and the last group are right justified. @nicolethoen might be able to help here, she showed us how it could be done during a meeting about this component.

This is how the component currently looks at a larger size format:

Screen Shot 2020-11-30 at 4 39 43 PM

packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -390,6 +393,272 @@ class ToolbarGroupTypes extends React.Component {
...
+
+    return (
+      <Toolbar id="toolbar-group-types">
+        <ToolbarContent>{items}</ToolbarContent>
+      </Toolbar>
+    );
+  }
+}
+```
+
```
jcaianirh, line 661: We also need to take into account the compressed version of this component. The design located here: http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/log%20toolbars/#compressed-toolbar

jcaianirh (changes requested):

packages/react-core/src/components/Toolbar/examples/Toolbar.md
DaoDaoNoCode: I would like to discuss how the users would like to use the compressed version in tomorrow's meeting, so I would implement this after tomorrow.

packages/react-core/src/components/Toolbar/examples/Toolbar.md
nicolethoen: Yeah I believe there is an `alignment` prop that can go on the `ToolbarContent`, the `ToolbarGroup`, or the `ToolbarItem` depending on the structure. And it can set the alignment at specific or all viewport widths.

nicolethoen: I think you'll want to group the left aligned items and the right aligned items separately in different ToolbarGroups

packages/react-core/src/components/Toolbar/examples/Toolbar.md
nicolethoen: I do believe that you will make use of the `ToolbarToggleGroup` and/or the `OverflowMenu`
I'm happy to help with this, too if need be. I'm not as familiar with the `OverflowMenu` but there may be opportunity for multiple of them in this case.

nicolethoen (commented):
I wonder if this is actually a situation which calls for a Demo rather than a component example - it's sort of like a specific toolbar use case, so I think that's probably actually a Demo. @mcarrano do you have an opinion?

mcarrano:
I agree with @nicolethoen . This is a pretty specific use case and is more appropriate as a demo than as a new component example. I think there are more toolbar demos we could add in the future to illustrate more usage specific instances.

DaoDaoNoCode:
@nicolethoen @mcarrano Thanks, I will move this use case into a demo.

mcarrano (changes requested):
This looks much better as a demo @DaoDaoNoCode . Just a couple of design changes I'd like to see made:

* Since the label associated with the switch does not change for 'Timestamp' or 'Wrap lines', you would use the switch styling that includes an embedded checkmark as shown here: https://www.patternfly.org/v4/components/switch/react#without-label.

* Unless they are used as a link, icon buttons that trigger an action should be gray and not blue. That would apply to the full-screen button at the right of the toolbar. So it would be button variant "plain" rather than "link." @mceledonia can you confirm that?

DaoDaoNoCode:
@mcarrano Thanks, but I got a question. Since we use the Switch without label, what component should we use to show the labels? Should we use `Text`? Or I will just simply put a string next to the Switch component.

mcarrano:
Good question @DaoDaoNoCode . I think you should be able to just put a string inline, but maybe @mcoker can provide some advise. Take a look at this demo: https://patternfly-react-pr-5196.surge.sh/components/toolbar/react-demos/console-log-viewer-toolbar-demo/  Basically what I'm asking is that we use the switch variation with an embedded checkmark since the label will not change based on the state of the switch. Would we just place the label following the switch component?

mcoker:
@mcarrano @DaoDaoNoCode Looking at the designs, they're using the switch check icon + the switch label. When either "timestamp" or "wrap lines" is switched on, the text is black, and when they're switch off, the text is grey.

The problem with just putting inline text next to the switch is that the switch and text beside it will be top-aligned, since the switch is bigger than regular text. If we don't want clicking on the text to toggle the switch, and for the text color to change as the switch is toggled on/off, you could use a level layout for the switch + label, with either a gutter or a custom margin on the switch/label text to create the space between the switch and label. 

Screen Shot 2020-12-02 at 5 44 17 PM
Screen Shot 2020-12-02 at 5 44 24 PM

Or if you wanted more control over the spacing without writing custom styles, you could use a flex layout with align-items set to center and a custom spacer.

mcarrano:
@DaoDaoNoCode I looked at this again and see that you opened this issue that was closed: https://github.com/patternfly/patternfly-react/issues/5181 Adding the prop would solve this problem, I think. Do you know why that was closed?

DaoDaoNoCode:
@mcoker That makes sense, thanks! I will use the gutter method.

DaoDaoNoCode:
@mcarrano As you can see in the design doc here http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/log%20toolbars/. There is a check icon inside the switch button. I was thinking of using Switch with label at that time so I hoped to set the check icon by myself. However, there is no prop for me to choose whether to show the check icon or not. If we use Switch with label it will not show the check icon. Only if we choose to use Switch without label it will show the icon. Then I opened the issue.
After that, I asked @andybraren about this design. He said it's totally fine without the check icon, so I closed the issue after that because it may violates the original intention of the design.
After reading your comment above I found actually It's my fault because I am not fully understand the design of Switch component. If we the label will not changed along with the Switch button, then we should use the Switch without label and set the label separately, then there should be a check icon on it. There is nothing wrong with the design doc.
Or do you think we need to add a prop for our users to choose whether to show the check icon or not? If so, I would reopen the issue and implement it.

mcarrano:
@DaoDaoNoCode I'm not sure how best to answer this because the design you are working with deviates from the intended usage of the switch component. The label following the switch is intended to reflect the state of the setting, on or off by default. Since the label does not change in this case, I felt like we should be using the version with a checkmark to provide that feedback. The expected usage would be something more like this with a label in front of the switch:

Screen Shot 2020-12-03 at 3 50 59 PM

@andybraren do you think it makes sense to swap the order of the switch and label to be more consistent with our guidance here: https://www.patternfly.org/v4/components/switch/design-guidelines?

DaoDaoNoCode:
@mcarrano @jcaianirh I have implemented the responsive layout of the demo. Now if you change the window size to 'xl' or less, it will show the icon buttons and toggles. Could you please review it?

andybraren:
@mcarrano @DaoDaoNoCode Ah, we missed that deviation in our design review over the summer, thanks for pointing that out! The guidance tweaks in https://github.com/patternfly/patternfly-org/pull/2287 seem to clarify this a bit, mostly focusing on the space-saving factor.

Swapping the order of the switch and label and using the checkmark makes sense to me. 👍

DaoDaoNoCode:
Thanks @andybraren , that makes sense. So do you think we need to change what's said in the design doc and put the label to the left of the switch button?

nicolethoen:
when i view the demo on the surge site, I see this on the demo page:
Screen Shot 2020-12-08 at 10 15 04 AM
And I see this when i open the demo to view on a full page:
Screen Shot 2020-12-08 at 10 14 44 AM
The compact view looks like its appearing as desired.

DaoDaoNoCode:
@nicolethoen Thank you. Do you have suggestions for anything I could improve for this demo like code structure or layout?

andybraren:
> Thanks @andybraren , that makes sense. So do you think we need to change what's said in the design doc and put the label to the left of the switch button?

@DaoDaoNoCode Moving the label to the left certainly makes sense. Oftentimes the forward-looking documentation we merge into the OpenShift Design repo ends up being more of a guideline than a strict spec, so some differences are totally cool. After the implementation merges in OpenShift we can move that design from `4.x` to the version we release in and add a list to the top of the various tweaks that were made during the implementation phase. 👍

nicolethoen:
> @nicolethoen Thank you. Do you have suggestions for anything I could improve for this demo like code structure or layout?

There are ways you can manipulate the spacing between items and groups of items. And i don't see the design having a `2/3` in the search input, so maybe you can remove that. I don't totally know. In reality it looks like each component in the log view toolbar is actually larger than the design expects. That may be something to take back to the designers.

mcarrano (commented):
@DaoDaoNoCode could you adjust the breakpoints so you don't get this happening?

Screen Shot 2020-12-08 at 3 48 14 PM

Seems like we'd want to break earlier.

Also, and this is more of a design question for @andybraren , if when the window shrinks we change the switches to checkboxes, does it make better sense to just use checkboxes in the full width toolbar? That would provide a consistent interaction and save some space, I think.

DaoDaoNoCode:
@mcarrano Of course I can adjust the breakpoint.
About the Switch component I just discussed this with @andybraren , we both think that maybe checkboxes in a dropdown menu instead of switch button is a better idea. It's scalable and users can add more functions in the future. I will implement it soon.

DaoDaoNoCode:
@mcarrano @andybraren @nicolethoen What do you think of the design now? I put the 2 switch buttons into a toggle menu named Option. You can even add more functions into the menu like clearing the logs.

Screen Shot 2020-12-09 at 2 14 31 PM

Also, in the mobile version, I use the cog icon toggle button to replace the previous kebab toggle button, which help users to understand this is a setting/options menu.

nicolethoen:
I think they look good as far as layout! 
I think that you might need a ToolTip on hover for the buttons that loose their label on the small view?

DaoDaoNoCode:
@nicolethoen That's a nice suggestion! It could help users to know the function of button/dropdown instead of guessing the functions only be the icon!

andybraren:
Looking good @DaoDaoNoCode, and agreed! Tooltips on the icon-only buttons makes sense.

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,518 @@
...
+            isOpen={containerExpanded}
+            customContent={selectDropdownContent}
+            placeholderText={selectToggleContent({ showText: true })}
+          />
+        </ToolbarItem>
+        <ToolbarItem visibility={{default: 'hidden', '2xl': 'visible'}}>
+          <Dropdown
+            toggle={
+              <DropdownToggle id='option-toggle-desktop' onToggle={this.onOptionToggle} icon={<CogIcon />}>
+                Option
```
andybraren, line 340:
```suggestion
                Options
```

DaoDaoNoCode: Fixed, thanks!

mcarrano (approved):
I like it @DaoDaoNoCode !

jcaianirh (approved):
lgtm thanks @DaoDaoNoCode

nicolethoen (approved):

tlabaj (approved):
LGTM

DaoDaoNoCode:
@mcarrano @andybraren @nicolethoen @jcaianirh  Could you please review the design again? After last meeting @andybraren and I discussed about how we could implement the search input on the mobile view. Now we add a search icon and expand the search input below the icon if users click on it. Could you give me some feedbacks on this design?

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+              }
+              isOpen={optionExpandedMobile}
+              dropdownItems={optionDropdownItems}
+            />
+          </Tooltip>
+        </ToolbarItem>
+        <ToolbarItem visibility={{default: 'visible', '2xl': 'hidden'}}>
+          <Tooltip
+            position="top"
+            content={<div>{isPaused? "Start log" : "Pause log"}</div>}
```
andybraren, line 415:
```suggestion
            content={<div>{isPaused? "Resume log" : "Pause log"}</div>}
```

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+              <DropdownToggle id='option-toggle-desktop' onToggle={this.onOptionToggle} icon={<CogIcon />}>
+                Options
+              </DropdownToggle>
+            }
+            isOpen={optionExpanded}
+            dropdownItems={optionDropdownItems}
+          />
+        </ToolbarItem>
+        <ToolbarItem visibility={{default: 'hidden', '2xl': 'visible'}}>
+          <Button variant={isPaused? "plain": "link"} onClick={this.pauseOrStart}>
```
andybraren, line 373: I like how the mobile tooltips you added to this button - `Pause log` and `Start log` - are more descriptive and likely more accessible than the current "Log Stream" label in the button itself. Maybe the button label should toggle between those two strings when clicked as well? WDYT?

`Pause log` and `Resume log` would actually align well with similar toggles we've started to use elsewhere in OpenShift too:

image

image

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+          <Dropdown
+            onSelect={this.onExternalSelect}
+            toggle={
+              <DropdownToggle
+                splitButtonItems={[
+                  <DropdownToggleAction key="action-7" onClick={this.onExternalClick}>
+                    External Logs
+                  </DropdownToggleAction>
+                ]}
+                splitButtonVariant="action"
```
andybraren, line 447: Non-split, basic dropdowns may work a bit better for the External Logs and Download dropdowns here. I see now that the mockup had a divider line which was deceiving, but I can't think of a scenario (at least in OpenShift) where we'd need these to be split.

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+        <React.Fragment>
+          <Badge>{this.firstOptions[containerSelected].type}</Badge>
+          {showText && ` ${containerSelected}`}
+        </React.Fragment>
+      );
+    };
+
+    const LogsSearchInput = (
+      <SearchInput 
+        placeholder='find'
```
andybraren, line 338:
```suggestion
        placeholder='Search logs'
```

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+        </ToolbarItem>
+      </React.Fragment>
+    );
+
+    const rightAlignedItemsMobile = (
+      <React.Fragment>
+        <ToolbarItem visibility={{default: 'visible', lg: 'hidden'}}>
+          <Tooltip
+            position="top"
+            content={<div>search logs</div>}
```
andybraren, line 481:
```suggestion
            content={<div>Search Logs</div>}
```

Maybe this should turn into `Close search` when in that state?

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+        </ToolbarItem>
+      </React.Fragment>
+    );
+
+    const rightAlignedItems = (
+      <React.Fragment>
+        {rightAlignedItemsDesktop}
+        {rightAlignedItemsMobile}
+        <ToolbarItem>
+          <Button variant="plain" aria-label="expand">
```
andybraren, line 527: Should this have a tooltip as well? `Expand`?

packages/react-core/src/demos/Toolbar.md
```diff
@@ -0,0 +1,585 @@
...
+        </ToolbarItem>
+      </React.Fragment>
+    );
+
+    const leftAlignedItemsMobile = (
+      <React.Fragment>
+        <ToolbarItem visibility={{default: 'visible', '2xl': 'hidden'}}>
+          <Tooltip
+            position="top"
+            content={<div>Select Pod</div>}
```
andybraren, line 386:
```suggestion
            content={<div>Select Container</div>}
```

nit, not that this matters much at all for the demo 🙂

andybraren (commented):
I really like the new search-on-mobile experience @DaoDaoNoCode, great idea! I played with the demo more and came across a couple of other small nitpicks and details to consider.

packages/react-core/src/demos/Toolbar.md
DaoDaoNoCode: +1. I will change `Log Stream` to `Resume Log` and `Start Log`.

DaoDaoNoCode:
@andybraren Thank you for your suggestions! I have modified all the details you mentioned above. The only problem is the tooltip of `Select Container`. It is too long to show on the top of the icon toggle so it will show on the right of it instead like this. Do you think it's OK? It will block the Options toggle and Resume/Pause button when hover it.
Screen Shot 2020-12-16 at 5 20 22 PM

mcarrano:
@DaoDaoNoCode Is this using the toggle group approach as defined here? https://patternfly-react-pr-5196.surge.sh/components/toolbar/react#examples-with-toggle-groups-and-filters That's the approach we should be using to show/hide elements in a separate panel when you shrink the viewport. Here is a core example of that used to show/hide filters: https://www.patternfly.org/v4/components/toolbar/html-demos#toolbar-attribute-value-search-filter-on-mobile In this case the icon used to toggle the group does not change.

DaoDaoNoCode:
@mcarrano Thank you for the hint on using `ToolbarToggleGroup`. This is a great solution and much easier to implement the responsive search input! I have pushed the changes and could you please review it again?

mcarrano (approved):
Looks good to me. Thanks for updating this @DaoDaoNoCode .

jcaianirh (approved):
lgtm

andybraren (approved):
LGTM, thank you @DaoDaoNoCode!

@mcarrano Playing around with the [core toggle group example](https://www.patternfly.org/v4/components/toolbar/html-demos/toolbar-attribute-value-search-filter-on-mobile/) you linked, is there a reason that `.pf-c-toolbar__expandable-content` has a `box-shadow` below the`992px` breakpoint?

I'm referring to the slight shadow seen here:

image

This PR's toolbar example consumes that component so it includes the shadow as well, but thinking ahead to what this'll look like with an eventual log viewer below it (mocked up below) I'm not sure the shadow adds much. Maybe this is a visual question @mceledonia?

Artboard

mcarrano:
@andybraren It was a decision to make the toggle panel feel like an overlay on top of the content below. I do agree that the shadow looks a bit odd in the screenshot above. What do you think @mceledonia ? We would need to revisit that in the core implementation. There might be an easy way to turn the shadow off without making a breaking change.

tlabaj (approved):

nicolethoen (approved):

PR #5201, opened 2020-12-01
feat(react-core): move react-datetime to react-core

redallen:
**What**: Now that react-datetime has no 3rd party deps it can move in with react-core to be consumed more easily.

**Additional issues**:


tlabaj (approved):
LGTM

evwilkin (approved):
LGTM

Issue #5203, opened 2020-12-02
Chart: Interactive legend example tooltips stop working


dlabrecq:
When a data series is hidden via the interactive legend example, the tooltip stops working. The tooltips will begin working again if the user moves the cursor away from the chart.

In order to fix the example, the chart's container should not be recreated from scratch each time the chart is rendered.

![before](https://user-images.githubusercontent.com/17481322/100829338-1e379b00-342f-11eb-8b1f-2395ccbf6175.gif)

dlabrecq:
This example has been fixed

PR #5204, opened 2020-12-02
charts(chore): Interactive legend example tooltips stop working

dlabrecq:
When a data series is hidden via the interactive legend example, the tooltip stops working. The tooltips will begin working again if the user moves the cursor away from the chart.

In order to fix the example, the chart's container should not be recreated from scratch each time the chart is rendered.

https://github.com/patternfly/patternfly-react/issues/5203

**Before**
before

**After**
after


ddonahue007 (approved):
LGTM

tlabaj (approved):
LGTM

Issue #5205, opened 2020-12-02
Menu component: Let consumer add SearchInput/TextInput and remove hasSearchInput prop


jschuler:
Copying from conversation with @karelhala, @jenny-s51 and @mcoker in slack:
> While I was going trough the component itself I think it could be improved, I don't think we need to include the hasSearchInput prop at all, we could allow passing custom components as consumers require, here is an example - https://codesandbox.io/s/cocky-yonath-65nuf?file=/index.js

we could provide more components to control that level, i.e. by adding a MenuInput and MenuContent container
```
<Menu>
  <MenuInput>
    <SearchInput />
  </MenuInput>
  <Divider>
  <MenuContent>
    <MenuList>
      <MenuListItem />
      <MenuListItem />
    </MenuList>
  </MenuContent>
</Menu>
```

Related: https://github.com/patternfly/patternfly-react/issues/5153

karelhala:
Just a note that this should support multiple `Inputs` and spread around the menu itself, for instance to support searching in group
```
<Menu onSelect={this.onSelect} activeItemId={activeItem}>
  <MenuGroup>
    <MenuInput>
      <SearchInput />
    </MenuInput>
    <MenuContent>
      <MenuList>
      <MenuList>
    </MenuContent>
  </MenuGroup>
  <MenuGroup>
    <MenuInput>
      <SearchInput />
    </MenuInput>
    <MenuContent>
      <MenuList>
      <MenuList>
    </MenuContent>
  </MenuGroup>
</Menu>
```

Ideally it would be really good if we could have inputs as `MenuItem`s as well, but I would mark that as a stretch goal and might not actually be needed.

Also, while doing this change, I would reommend removing the `hasSearchInput` altogether. I'd be more happy if we didn't have to do some conditions to render searches and instead give consumers direct message that they can use whatever input they want to.

jschuler:
@mcoker is that something we can already with with the Core CSS? Add inputs at the group level

mcoker:
Technically you can, there are no styles associated with `<MenuGroup>` or `<MenuContent>`.

<img width="1254" alt="Screen Shot 2020-12-03 at 10 36 59 AM" src="https://user-images.githubusercontent.com/35148959/101059302-baf35900-3553-11eb-8a15-c7323b7ca710.png">

cc @mcarrano

jenny-s51:
[...]

PR #5211, opened 2020-12-03
Menu: Let consumer add SearchInput/TextInput

### LINKED ISSUE #5153: Menu component: replace SearchInput with TextInput
This issue follows the discussion here: https://github.com/patternfly/patternfly-react/pull/4847#discussion_r523275762

For the original issue that was opened to implement the menu component, see #4712.

The `<SearchInput>` component should be replaced with `<TextInput>` in the filterable menu variant, per @mcoker and @mcarrano.

jenny-s51:
**What**: Closes #5153 and #5205

**Additional issues**: #4712


packages/react-core/src/components/TextInput/TextInput.tsx
```diff
@@ -180,7 +180,7 @@ export class TextInputBase extends React.Component<TextInputProps> {
           styles.formControl,
           validated === ValidatedOptions.success && styles.modifiers.success,
           validated === ValidatedOptions.warning && styles.modifiers.warning,
-          (iconVariant || customIconUrl) && styles.modifiers.icon,
+          (iconVariant !== 'search' || customIconUrl) && styles.modifiers.icon,
```
jschuler, line 183: Also needs a check to see if it is defined otherwise it will always add the icon style
```suggestion
          ((iconVariant && iconVariant !== 'search') || customIconUrl) && styles.modifiers.icon,
```

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -76,25 +73,7 @@ class MenuBase extends React.Component<MenuProps, MenuState> {
...
-                    this.setState({ searchInputValue: '' });
-                    onSearchInputChange(event, '');
-                  }}
-                />
-              </div>
-              <Divider />
-            </React.Fragment>
-          )}
-          <div className={css(styles.menuContent)}>{children}</div>
+          <MenuContent>{children}</MenuContent>
```
jschuler, line 76: Doesn't the input have to be outside of the `MenuContent` ? If so then this should just be `{children}` and consumer will have to pass
```
<Menu>
  <MenuInput><TextInput /></MenuInput>
  <MenuContent>
     <MenuList>...</MenuList>
  </MenuContent>
</Menu>
```

jschuler (changes requested):

packages/react-core/src/components/Menu/MenuInput.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export interface MenuInputProps extends React.HTMLProps<HTMLElement> {
+  /** Items within input */
+  children?: React.ReactNode;
+  /** Forwarded ref */
+  innerRef?: React.Ref<any>;
+}
+
+const MenuInputBase: React.FunctionComponent<MenuInputProps> = ({ children }: MenuInputProps) => (
+  <div className={styles.menuSearch}>{children}</div>
```
jschuler, line 12: I think I also saw Core nesting `pf-c-menu__search-input` here, is that needed @mcoker ?

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -174,19 +174,79 @@ class MenuWithFiltering extends React.Component {
...
+      <Menu onSelect={this.onSelect} activeItemId={activeItem}>
+        <MenuInput>
+          <TextInput
+            value={input}
+            aria-label="filterable-example-with-text-input"
+            iconVariant="search"
+            type="search"
+            onChange={value => this.handleTextInputChange(value, 'input')}
+          />
+        </MenuInput>
```
jschuler, line 211: Also needs the `Divider` back in both examples

packages/react-core/src/components/Menu/MenuInput.tsx
mcoker: @jschuler it's not necessary, but having that element makes it easier for us to style the text input relative to other things we might add to the menu search element.

jschuler (approved):
snapshots needs to be fixed, otherwise LGTM

packages/react-core/src/components/Menu/MenuContent.tsx
```diff
@@ -0,0 +1,19 @@
...
+  innerRef?: React.Ref<any>;
+}
+
+const MenuContentBase: React.FunctionComponent<MenuContentProps> = ({ children }: MenuContentProps) => (
+  <div className={css(styles.menuContent)}>{children}</div>
+);
+
+export const MenuContent = React.forwardRef((props: MenuContentProps, ref: React.Ref<HTMLElement>) => (
+  <MenuContentBase {...props} innerRef={ref} />
+));
```
karelhala, line 18: You are not doing anything with the `innerRef` prop, this way we'd just block the ref forwarding. Plus any additional props will be ignored and for instance custom class would be impossible to add. You can also simplify this by using the `div` directly
```suggestion
export const MenuContent = React.forwardRef((props: MenuContentProps, ref: React.Ref<HTMLElement>) => (
  <div {...props} className={css(styles.menuContent, props.className)} ref={ref} />
));
```

packages/react-core/src/components/Menu/MenuInput.tsx
```diff
@@ -0,0 +1,18 @@
...
+);
+
+export const MenuInput = React.forwardRef((props: MenuInputProps, ref: React.Ref<HTMLElement>) => (
+  <MenuInputBase {...props} innerRef={ref} />
+));
```
karelhala, line 17: Same as above, not really needed to add new component
```suggestion
export const MenuInput = React.forwardRef((props: MenuInputProps, ref: React.Ref<HTMLElement>) => (
  <div {...props} className={css(styles.menuSearch, props.className)} ref={ref} />
));
```

karelhala (commented):
Looking good! I like the example with plain TextInput it shows the prefered way and allows for easy example that consumers can use whatever they want to show there.

jenny-s51:
Thank you for your helpful feedback @karelhala ! I've applied your suggestions

karelhala (approved):

tlabaj (approved):
This looks good Jenny!

jschuler (approved):

Issue #5020, opened 2020-10-14
Support "favorites" in a table


mcarrano:
Add the ability to mark table items (rows) as favorites and then sort the table by favorites. Interaction should be same as for favoriting in the Select or Application Launcher expect that items and not moved or copied when they are marked/unmarked as favorities.

Core issue and designs: https://github.com/patternfly/patternfly/issues/1860

PR #5213, opened 2020-12-04
Table favorites

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/5020

https://patternfly-react-pr-5213.surge.sh/components/table#favoritable


mcarrano:
This looks great @jschuler . There's just one small styling issue that I wonder if we can fix. When the favorites star in the header is not selected to sort, it should be a slightly darker gray per the design. Here's what the original design looked like:

Screen Shot 2020-12-04 at 9 38 04 AM

and here's the implementation:

Screen Shot 2020-12-04 at 9 39 08 AM

Right now it feels like the column is disabled. @mcoker do we need a CSS update for that? @mceledonia can you tell us what gray value we should be using there? I don't see it specified in the design. Thanks.

mcoker:
@mcarrano I want to say that was something we changed while implementing the PR, but I can make it darker.

mceledonia:
Yeah it should be our secondary text/global color #6A6E73 in order for it to pass contrast requirements.

mcarrano (changes requested):
@jschuler the styling looks right now. In looking at this again, there's just one thing I was hoping you can change. When I first click the star in the header to sort by favorities, the initial sort is to put the favorited items at the end. I need to click again to move them to the top. That's exactly opposite of what I expect. If a user sorts by favorities, it's most likely because they want to pop all of the favorited items to the top of the table so they can take some action on them. Is is possible to just turn that around?

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1812,6 +1816,171 @@ class EditableRowsTable extends React.Component {
...
+          onChange={this.toggleFavsSort}
+          aria-label="toggle select all checkbox"
+          id="toggle-select-all"
+          name="toggle-select-all"
+        />
+        <Table
+          // using this prop enables the favorites column
+          onFavorite={this.onFavorite}
+          // if the onSort prop is detected, favorites can be sorted
+          // if you want to exclude favorites from sorting you can use this prop with a value of `false`
```
tlabaj, line 1966: These comments are great.  I wonder if it makes sense to add them to the prop docs?

packages/react-table/src/components/TableComposable/Td.tsx
```diff
@@ -63,6 +64,16 @@ export interface TdProps extends BaseCellProps, Omit<React.HTMLProps<HTMLTableDa
...
   };
+  favorites?: {
+    /** Whether the corresponding row is favorited */
+    isFavorited: boolean;
+    /** Callback on clicking the favorites button */
+    onFavorite?: OnFavorite;
+    /** The row index */
+    rowIndex?: number;
+    /** Additional props forwarded to the FavoritesCell */
+    props?: any;
```
tlabaj, line 75: DO you think we should make the name more specific here?

packages/react-table/src/components/Table/examples/Table.md
jschuler: yes, I added a description for the Favorites example that explains how to enable favoriting, and in the props I added
```
/**
   * Enables favorites column
   * Callback triggered when a row is favorited/unfavorited
   */
  onFavorite?: OnFavorite;
  /** Along with the onSort prop, enables favorites sorting, defaults to true */
  canSortFavorites?: boolean;
```

tlabaj (commented):
Looks good. Can we add some cypress interaction test as well.

mcarrano (approved):
Looks great. Thanks for making that change @jschuler .

kmcfaul (approved):
lgtm code wise!

packages/react-table/src/components/TableComposable/Td.tsx
jschuler: To use it you need to wrap it in the `favorites` object which gives it context, e.g.
```
favorites: {
  props: { 'aria-label': 'favs' }
}
```

tlabaj (approved):
LGTM

mcoker (approved):
LGTM! 🤩

PR #5216, opened 2020-12-07
feat(TimePicker): Add basic time picker

### LINKED ISSUE #4731: Build basic time picker
Does not include timezones.

nicolethoen:
**What**: Closes #4731 

https://patternfly-react-pr-5216.surge.sh/components/time-picker/


packages/react-core/src/components/TimePicker/TimePickerConst.tsx
```diff
@@ -0,0 +1,164 @@
...
+  `20${delimiter}30`,
+  `21${delimiter}00`,
+  `21${delimiter}30`,
+  `22${delimiter}00`,
+  `22${delimiter}30`,
+  `23${delimiter}00`,
+  `23${delimiter}30`
+];
+
+export const hourMappings24: { [key: string]: number } = {
```
redallen, line 103: Can't this just be `(n: string) => Number(n) * 2`?

packages/react-core/src/components/TimePicker/TimePickerConst.tsx
```diff
@@ -0,0 +1,164 @@
...
+  '17': 34,
+  '18': 36,
+  '19': 38,
+  '20': 40,
+  '21': 42,
+  '22': 44,
+  '23': 26
+};
+
+export const hourMappings12: { [key: string]: number } = {
```
redallen, line 140: Can't this just be `(n: string) => (n % 12) * 2;`?

packages/react-core/src/components/TimePicker/examples/TimePicker.md
```diff
@@ -0,0 +1,55 @@
...
+import { TimePicker } from '@patternfly/react-datetime';
+
+SimpleTimePicker = () => {
+  const onChange = (time) => {
+    console.log(time);
+  };
+  
+  return (
+    <React.Fragment>
+        <TimePicker defaultTime="3:35pm" onChange={onChange}/>
```
redallen, line 24: I'm seeing `3:35am` instead of `3:35pm`
image

redallen (changes requested):
Clicking and then scrolling with the arrow keys doesn't scroll to the list item:
arrow-keys

packages/react-core/src/components/TimePicker/examples/TimePicker.md
```diff
@@ -0,0 +1,45 @@
...
+beta: true
+---
+
+import { TimePicker } from '@patternfly/react-core';
+
+## Examples
+### Basic 12 hour
+```js
+import React from 'react';
+import { TimePicker } from '@patternfly/react-datetime';
```
redallen, line 15: Replace `@patternfly/react-datetime` with `@patternfly/react-core` in all examples.

nicolethoen:
FYI I'm working on the clear button :)

redallen (changes requested):
What exactly is the point of this button when you can just click inside the `<input>`? 
image

Maybe it'd be good to follow the interaction pattern of [Typeahead select](https://patternfly-react-pr-5216.surge.sh/components/select#typeahead) and have clicking on the button just focus the input?

nicolethoen:
and im removing the selected state

nicolethoen:
also i know i need a little more clean up with the am/pm logic

im going to remove the selected check and selected state.

nicolethoen:
i need to validate the hours are in range

nicolethoen:
> What exactly is the point of this button when you can just click inside the `<input>`?

we are going to remove the clear button so it looks more like the date picker. and then i can use the time icon variant of the `TextInput` component

jessiehuff (commented):
So far, this looks pretty good! I agree with Zack above about the clock button, it seems odd to have that interaction behavior on both. Based on your comment, it sounds like you're removing the clear button so that clears up adding the keyboard interaction as well. It sounds fairly solid at the moment in VO as well, and I'm able to navigate everything. 🙂 👍

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,509 @@
...
+    } else {
+      this.setState({
+        time,
+        focusedIndex: null
+      });
+    }
+  };
+
+  validTimeFormat = (value: string) => {
+    const regexp = new RegExp(`\\b\\d\\d?${this.props.delimiter}?[0-5]?\\d?\\s?([AaPp][Mm])?\\b`);
```
evwilkin, line 347: Not sure if this is related to the "need to validate the hours are in range" comment, but this allows hours up to 99:59pm

nicolethoen: yeppp. gotta clean that up :)

tlabaj (changes requested):
Can you also open follow up issue for demo and integration tests?

jessiehuff (dismissed):
LGTM! 🙂

mcoker (dismissed):
Fantastic!

packages/react-core/src/components/TimePicker/TimePickerConst.tsx
```diff
@@ -0,0 +1,101 @@
+export const times12Hr = (delimiter: string) => [
```
redallen, line 1: Here's a function that allows for a custom step:
```ts
function makeOptions(start: Date, stepMinutes: number, hour12: boolean, delimiter: string) {
  let res = [];
  let iter = new Date(start);
  let iterDay = iter.getDay();
  while (iter.getDay() === iterDay) {
    let hour = iter.getHours();
    let suffix = 'am';
    if (hour12) {
      if (hour === 0) {
        hour = 12; // 12am
      }
      else if (hour >= 12){
        suffix = 'pm';
      }
      if (hour > 12) {
        hour %= 12;
      }
    }
    res.push(
      hour.toString().padStart(2, '0') +
      delimiter +
      iter.getMinutes().toString().padStart(2, '0') +
      (hour12 ? suffix : '')
    )

    iter.setMinutes(iter.getMinutes() + stepMinutes);
  }

  return res;
}
```

redallen (changes requested):
It works but with some cleanup can support a few more props and be easier to maintain in the future.

mcarrano (dismissed):
Looks great @nicolethoen . Thanks for all your iteration on this!

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,457 @@
...
+      time = `${splitTime[0]}${delimiter}${minutes}${new Date().getHours() > 11 ? 'pm' : 'am'}`;
+    } else if (
+      variant === '12hr' &&
+      splitTime.length > 1 &&
+      splitTime[1].length === 2 &&
+      !time.toLowerCase().includes('am') &&
+      !time.toLowerCase().includes('pm')
+    ) {
+      time = `${time}${new Date().getHours() > 11 ? 'pm' : 'am'}`;
+    }
```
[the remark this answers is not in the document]
nicolethoen: I'll pull out the code they share into a new `appendMissingAmPm` function.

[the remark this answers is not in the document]
nicolethoen: I lied - they don't have any shared code/logic. It looks similar, probably because it's betraying something about the way my brain works, but there is enough difference that they actually dont share any scenarios.

`parseTime` parses ISO 8601 date strings, and appends am/pm to 12 hour times that are otherwise valid.

`scrollToSection` finds the option that matches the current partial or complete time input and scrolls to it. In the case that there is not exact match, it determines the closest match and scrolls to it.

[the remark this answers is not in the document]
redallen: That's fine. I wonder if we could embed each option's time in a data attribute in the `<li>` to avoid this in the future.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,517 @@
...
+      const amPm =
+        variant === '12hr'
+          ? splitTime[1].toLowerCase().includes('a')
+            ? 'am'
+            : splitTime[1].toLowerCase().includes('p')
+            ? 'pm'
+            : new Date().getHours() > 11
+            ? 'pm'
+            : 'am'
+          : '';
```
redallen, line 247: I find this hard to read and have opened #5238. You don't have to fix it.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,517 @@
...
+    }
+  };
+
+  onBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    this.setState({
+      isInvalid: !this.validateTime(this.parseTime(event.currentTarget.value))
+    });
+  };
+
+  makeOptions = (stepMinutes: number, hour12: boolean, delimiter: string) => {
```
redallen, line 369: This can be static.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,517 @@
...
+  };
+
+  appendMissingAmPm = (time: string) =>
+    this.props.variant === '12hr' &&
+    this.validateTime(time) &&
+    time !== '' &&
+    !time.toLowerCase().includes('am') &&
+    !time.toLowerCase().includes('pm')
+      ? `${time}${new Date().getHours() > 11 ? 'pm' : 'am'}`
+      : time;
```
redallen, line 294: Looks like this function is unused, remove.

packages/react-core/src/components/TimePicker/examples/TimePicker.md
```diff
@@ -0,0 +1,45 @@
...
+SimpleTimePicker = () => {
+  const onChange = (time) => {
+    console.log(time);
+  };
+  
+  return (
+    <React.Fragment>
+      <TimePicker defaultTime="3:35am" onChange={onChange}/>
+    </React.Fragment>
+  );
```
redallen, line 26: Remove `<React.Fragment>`

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,517 @@
...
+  componentDidMount() {
+    const { defaultTime } = this.props;
+    document.addEventListener('mousedown', this.onDocClick);
+    document.addEventListener('touchstart', this.onDocClick);
+    document.addEventListener('keydown', this.handleGlobalKeys);
+    if (defaultTime !== '') {
+      this.setState({
+        time: this.parseTime(defaultTime)
+      });
+    }
```
redallen, line 111: Why is this not in the constructor?

redallen (changes requested):
Looking good but needs a little more cleanup

packages/react-core/src/components/TimePicker/TimePicker.tsx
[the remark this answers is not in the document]
nicolethoen: how would that change this situation? it would just be the same as the option's value?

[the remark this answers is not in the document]
redallen: I meant that you could put the Date object into a data attribute and then compare the time part to what the user has input.

`<li data-date="2020-12-10T15:00:00.000Z ... />` -> `new Date(ref.getAttribute('data-date'))`

This way entering `1:5` could snap to `02:00` instead of `01:00` like it does now.

redallen (approved):
It's pretty and functional!

tlabaj (approved):
Look great! Nice work Nicole!

PR #5223, opened 2020-12-07
fix(jumplinks): allow hrefs with spaces

### LINKED ISSUE #5218: JumpLinksItem doesn't support spaces in href
Need to support spaces or add documentation to not include spaces.

redallen:
**What**: Closes #5218

**Additional issues**: Previously wasn't working because `id`s [can't include spaces.](https://www.quora.com/jQuery-How-can-I-select-a-div-whose-id-contains-a-space)

### DIFF
--- a/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
+++ b/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
@@ -24,7 +24,12 @@ const getScrollItems = (children: React.ReactNode, hasScrollSpy: boolean) =>
   React.Children.toArray(children).map((child: any) => {
     if (hasScrollSpy && typeof document !== 'undefined' && child.type === JumpLinksItem) {
       const scrollNode = child.props.node || child.props.href;
-      if (typeof scrollNode === 'string' && typeof document !== 'undefined') {
+      if (typeof scrollNode === 'string') {
+        if (scrollNode.startsWith('#')) {
+          // Allow spaces and other special characters as `id`s
+          // https://stackoverflow.com/questions/70579/what-are-valid-values-for-the-id-attribute-in-html
+          return document.getElementById(scrollNode.substr(1)) as HTMLElement;
+        }
         return document.querySelector(scrollNode) as HTMLElement;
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/JumpLinks/JumpLinks.tsx
```diff
@@ -24,7 +24,12 @@ const getScrollItems = (children: React.ReactNode, hasScrollSpy: boolean) =>
   React.Children.toArray(children).map((child: any) => {
     if (hasScrollSpy && typeof document !== 'undefined' && child.type === JumpLinksItem) {
       const scrollNode = child.props.node || child.props.href;
-      if (typeof scrollNode === 'string' && typeof document !== 'undefined') {
+      if (typeof scrollNode === 'string') {
+        if (scrollNode.startsWith('#')) {
+          // Allow spaces and other special characters as `id`s
```
KKoukiou, line 29: Hmm? The link you have here mentions that spaces are not allowed to be used in IDs.

redallen: Yes, but sadly Chrome allows them:
image

I think using `docuement.getElementById(id)` instead of `document.getSelector('#' + id)` is more permissive and better.

KKoukiou: I think more or less every docs I could find was  clear about spaces not being allowed in IDs.

https://www.w3.org/TR/REC-html40/types.html#type-name or https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id 

Of course allowing them here does not break anything, so you can without any problem merge it, but it's slightly making the code more complicated and harder to read, with the benefit of offering the users the ability to add spaces in their hrefs, which is an undocumented behavior.

If it was up to me, I would not implement this corner case, to favor code simplicity and if any time we get a report about this not working as expected I would just reply with the above doc links.

KKoukiou (approved):
@redallen as I commented I don't find this change necessary but I will approve and leave it up to you.

dlabrecq (approved):

PR #5236, opened 2020-12-09
Fix/focus trap dep

### LINKED ISSUE #4421: Select - FocusTrap issue causes scrollable menu to jump
reported by a user is with checkbox selects (where the pf-c-select__menu-item element is label.pf-c-check), if you set a maxHeight on the select that is shorter than the select menu (so that the menu scrolls), and you scroll down in the menu a bit, if you click anywhere in the item that is not the <input type="checkbox">, the scroll position in the menu is reset and you're taken to the top of the menu. Clicking on the checkbox seems to work fine.

image

FocusTrap has a fix in the works but it has not yet been released. [FocusTrap issue #104](https://github.com/davidtheclark/focus-trap/issues/104)
@redallen was made a co-maintainer of FocusTrap to get the fixed released.

redallen:
**What**: Closes #4421

**Additional issues**:

### DIFF
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -38,7 +38,7 @@
     "@patternfly/react-icons": "^4.7.21",
     "@patternfly/react-styles": "^4.7.20",
     "@patternfly/react-tokens": "^4.9.21",
-    "focus-trap": "4.0.2",
+    "@patternfly/focus-trap": "^4.0.0",
     "react-dropzone": "9.0.0",
     "tippy.js": "5.1.2",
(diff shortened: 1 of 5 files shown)


nicolethoen (approved):

jschuler:
Nice, should we fix it in this PR or as a followup? There are more components (like the Popover) that can benefit from this.

redallen:
@jschuler A followup issue would be great since I don't know which components those are.

jschuler (approved):

Issue #5248, opened 2020-12-11
When a node is added Topology with cola no force layout does not place the new items well


jeff-phillips-18:
**Description**
The topology cola layout w/o force does not place new items well. The items are centered in the graph and show on top of other existing items. The graph should adjust to accommodate the new items.

**Steps to Reproduce**
Run the demo-app
Navigate to the topology demo and then Topology Package
Select ColaNoForce layout
Click `Add Node` multiple times

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug, blocking

**What is your product and what release version are you targeting?**
OpenShift 4.7

PR #5249, opened 2020-12-11
fix(topology): adjust topology node layout when new nodes are added

jeff-phillips-18:
**What**: 
Closes https://github.com/patternfly/patternfly-react/issues/5248

**Description**
Update topology cola layout to run one round of force simulation after the cola layout when nodes are added. This will slightly adjust the graph to accommodate new nodes preventing initial overlap.

/cc @christianvogt


redallen (approved):

ddonahue007 (approved):
LGTM

PR #5251, opened 2020-12-11
fix(calendar-month): rely on base styles from react-core/dist/styles/base.css

redallen:
**What**: In some cases may fix base styles being loaded incorrectly:
image

**Additional issues**: React-styles shouldn't reexport `base` styles because `react-core/dist/styles/base.css` already exports them. Including them from react-styles just includes the CSS twice which can cause issues.


jschuler (approved):

evwilkin (approved):
LGTM

dlabrecq (commented):
LGTM

PR #5256, opened 2020-12-14
fix(Wizard): set a custom width to the wizard modal

### LINKED ISSUE #5254: Width prop is not used in the Wizard component
Width prop is not used in the Wizard component. Should probably be used like height:
https://github.com/patternfly/patternfly-react/blob/13dea26dc2ea6bcb6f873c145ec16fed909e6092/packages/react-core/src/components/Wizard/Wizard.tsx#L437

boaz0:
**What**: Closes #5254 

//cc @MariaAga

### DIFF
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -429,12 +429,17 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
       activeStep
     };
 
+    const divStyles = {
+      ...(height ? { height } : {}),
+      ...(width ? { width } : {})
+    };
+
     const wizard = (
       <WizardContextProvider value={context}>
         <div
           {...rest}
           className={css(styles.wizard, activeStep && activeStep.isFinishedStep && 'pf-m-finished', className)}
-          {...(height && { style: { height } })}
+          style={Object.keys(divStyles).length ? divStyles : undefined}
         >
           {title && (
             <WizardHeader
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -479,6 +479,7 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     if (isOpen !== undefined) {
       return (
         <Modal
+          width={width !== null ? width : undefined}
           isOpen={isOpen}
           variant={ModalVariant.large}
           aria-labelledby={this.titleId}
--- a/packages/react-integration/cypress/integration/wizard.spec.ts
+++ b/packages/react-integration/cypress/integration/wizard.spec.ts
@@ -5,9 +5,10 @@ describe('Wizard Demo Test', () => {
     cy.url().should('eq', 'http://localhost:3000/wizard-demo-nav-link');
   });
 
-  it('Verify wizard in modal launches in a dialog', () => {
+  it('Verify wizard in modal launches in a dialog and has a custom width', () => {
     cy.get('#launchWiz').click();
     cy.get('#modalWizId.pf-c-wizard').should('exist');
+    cy.get('.pf-c-modal-box').should('have.attr', 'style', 'width: 710px;');
     cy.focused().click();
   });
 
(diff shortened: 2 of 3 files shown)


MariaAga:
@boaz0 Thanks for the quick fix! It looks like this only changes the width of the wizard if it's in a modal. But a user still can't set custom width for the wizard like they can set the height.

boaz0:
Um, sorry I didn't realized that.

Can you elaborate more what is the use case? If you have mocks or screenshots that will be helpful too.

Thanks.

MariaAga:
I saw that there were a width prop that wasn't used so it was weird.
And about my use case, I'm not sure if the solution it max-width prop or width prop but my wizard looks a bit funny being so long :)
Screenshot from 2020-12-14 11-41-10
Screenshot from 2020-12-14 11-41-04

boaz0:
Uh huh. I see.
I don't think that is what `width` was supposed to do. 
Nonetheless, I will add this to the PR and let @tlabaj @jschuler & @redallen to review it as well.

Anyway, what I do to fix this is using the `flex` layout or set width to `360px` in the `Form`'s style.

Thanks.

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -429,12 +429,17 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
+      ...(width ? { width } : {})
+    };
+
     const wizard = (
       <WizardContextProvider value={context}>
         <div
           {...rest}
           className={css(styles.wizard, activeStep && activeStep.isFinishedStep && 'pf-m-finished', className)}
-          {...(height && { style: { height } })}
+          style={Object.keys(divStyles).length ? divStyles : undefined}
```
mcoker, line 442: We have CSS vars you can use here or via `setProperty()` to control the width and max-width. That's usually preferred as we may use that variable to help calculate some other property value that references the modal's width.

`--pf-c-modal-box--Width` - sets the width
`--pf-c-modal-box--MaxWidth` - sets the max-width

Helpful if you want the modal to be say, 50%, but not exceed 800px (`--pf-c-modal-box--Width: 50%; --pf-c-modal-box--MaxWidth: 800px;`)

boaz0: does it apply on the wizard when it's not displayed inside a modal?

mcoker: Oh sorry @boaz0, I read this wrong, I thought the width applied to the modal. My suggestion would not change anything with the wizard at all - it only change the modal. When the wizard is used in a modal, it is set to fill the full width and height of the modal. So by updating the modal's dimensions, the wizards dimensions automatically change to match the modal's.

If this feature is only to change the width of the wizard when in a modal, personally I think we should change the width of the modal. And instead of using the var I suggested, I just looked up the modal docs and `<Modal>` supports a `width` prop.

But if we also want the ability for someone to pass a width to the wizard when used outside of a modal, the approach in the PR makes sense. Another approach we could take for that, assuming the wizard is used in a page section, is to allow a custom `max-width` on page sections, and allow the user to set the max width on the page section the wizard is in.

redallen (commented):
> ...if we also want the ability for someone to pass a width to the wizard when used outside of a modal, the approach in the PR makes sense.

@MariaAga Is this what you are asking for?

MariaAga:
@redallen Yes as I saw that in the Wizard documentation here: https://www.patternfly.org/v4/components/wizard#props
```
width | number | string | No | null | Custom width of the wizard
```
But as mcoker said I can use `--pf-c-modal-box--Width` for that (or max-width).

redallen (approved):
Ah I see the issue -- we left behind an unused `width` prop that used to be for the modal but should now be for the Wizard itself.

This change LGTM although I'd like to deprecate the `width` and `height` props in favor of passing `style={{ width: '', height: '' }}`.

nicolethoen (approved):

mcoker (approved):
LGTM, thanks @boaz0!

tlabaj (approved):

PR #5258, opened 2020-12-14
fix(Select): fix crash when opened using Enter with disabled options

### LINKED ISSUE #5257: <Select> causes a crash when opened using enter key (this.refCollection[0][0] is null)
When a `<Select>` has a bunch of disabled options on the top of the options list, and you try to open it using keyboard navigation (tab to focus, then Enter to open), it crashes with the exception `TypeError: this.refCollection[0][0] is null`.

It only happens if there are disabled options, and only when opened using keyboard navigation.

https://codesandbox.io/s/red-cdn-ywe14?file=/index.js

1.  Prepare a `<Select>` with a bunch of disabled options, like the code sandbox link above
2.  Press tab to focus on the Select
3.  Hit enter. It'll crash.

When trying to debug this in my app, I noticed that for a Select with disabled options, `this.refCollection` looks like the following:
image
As you can see, the disabled options have no refs.

The offending line seems to be https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Select/Select.tsx#L199 , it tries to focus on the first option when opened using the Enter key, but it appears that because refs aren't saved for disabled options, it fails.

I suggest that instead of assuming `this.refCollection[0][0]` would always exist, it'd be better to have something like this, to focus on the first enabled item:
```js
this.refCollection.find((item) => item[0] !== null)[0].focus()
```

It's a bug. I haven't found a workaround yet, and it makes keyboard navigation in one of our forms useless. However, the fix should hopefully be simple enough as described above.

OpenShift Cluster Manager, currently using `"@patternfly/react-core": "^4.63.3"` (I don't think we're "targeting" a specific release)

elad661:
**What**: Closes #5257

When a `<Select>` has a disabled options on top of the options list,
`this.refCollection[0]` will be an array in the shape of `[null, null]`.
This means that trying to call `this.refCollection[0][0].focus()`
will fail with TypeError in this case.

This patch fixes this issue, by finding the first non-null ref in the
collection and focusing on it, instead of assuming the first ref in the
collection is valid.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -195,8 +195,11 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       this.refCollection[0][0] = this.filterRef.current;
     }
 
-    if (!prevState.openedOnEnter && this.state.openedOnEnter && !this.props.customContent && this.refCollection[0]) {
-      this.refCollection[0][0].focus();
+    if (!prevState.openedOnEnter && this.state.openedOnEnter && !this.props.customContent) {
+      const firstRef = this.refCollection.find(ref => ref !== null);
+      if (firstRef && firstRef[0]) {
+        firstRef[0].focus();
+      }
     }
 
     if (prevProps.children !== this.props.children) {


nicolethoen (approved):
Thanks for catching this!

elad661:
Reviewers:

To observe the bug, go to https://www.patternfly.org/v4/components/select and change the example code of the single select to have `isDisabled` on the first item, instead of `isPlaceholder`. Use keyboard navigation to focus on it, and then hit Enter, and it'll crash (the Select will disappear).

To test the fix, go to https://patternfly-react-pr-5258.surge.sh/components/select , and do the same thing, and the Select will open properly.

jschuler (approved):

kmcfaul (approved):

tlabaj (approved):

PR #5271, opened 2020-12-16
fix(topology): Prevent context sub menus from overflowing viewport

jeff-phillips-18:
**What**:
Fixes an issue where a topology context menu's sub menus would grow the viewport causing a window scrollbar to appear when the submenu would not fit on the screen.

**Solution**:
Fix the class on the sub menu dropdown and change the overflow prevention to be on the viewport rather than window.
Added context menus to the TopologyPackage demo for testing purposes.
Also fixes the submenu indicator padding so load order is not important.

/cc @christianvogt


kmcfaul (approved):

kmcfaul (approved):

ddonahue007 (approved):
LGTM

christianvogt:
lgtm

PR #5272, opened 2020-12-16
feat(ChipGroup): Add event handler prop for overflow-chip click 

### LINKED ISSUE #5190: Provide event for click on collapsedText or expandedText button in ChipGroup.

Currently, there is no event is fired when user click on collapsedText or expandedText in ChipGroup.
We should provide handler to the parent component. It will helps to perform some operation on the click of collapsedText or expandedText button.

Enhancement

Enhancement

joeldmyers:
**What**: Closes #5190 

Adds an optional prop called `onOverflowChipClick` that, when used, calls the provided handler function when the overflow (toggle expand/collapse) chip is clicked.  

There is already an onClick prop, hence naming this differently.  Let me know if you have a better idea for a name for this prop.

//cc @tlabaj

### DIFF
--- a/packages/react-core/src/components/ChipGroup/ChipGroup.tsx
+++ b/packages/react-core/src/components/ChipGroup/ChipGroup.tsx
@@ -33,7 +33,7 @@ export interface ChipGroupProps extends React.HTMLProps<HTMLUListElement>, OUIAP
   /** Function that is called when clicking on the chip group close button */
   onClick?: (event: React.MouseEvent) => void;
   /** Function that is called when clicking on the toggle expand/collapse button */
-  onToggleCollapseClick?: (event: React.MouseEvent) => void;
+  onOverflowChipClick?: (event: React.MouseEvent) => void;
   /** Position of the tooltip which is displayed if the category name text is longer */
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
 }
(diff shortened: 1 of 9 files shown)


packages/react-core/src/components/ChipGroup/ChipGroup.tsx
```diff
@@ -156,7 +160,14 @@ export class ChipGroup extends React.Component<ChipGroupProps, ChipGroupState> {
               ))}
               {numChildren > numChips && (
                 <li className={css(styles.chipGroupListItem)}>
-                  <Chip isOverflowChip onClick={this.toggleCollapse} component="button">
+                  <Chip
+                    isOverflowChip
+                    onClick={event => {
+                      this.toggleCollapse();
+                      onOverflowChipClick(event);
```
tlabaj, line 167: Shouldn't calling this be  this be conditional on wether it as an overflow chip?

tlabaj (changes requested):
Could you also please update the react-integration demo app and add the new prop to verify the API typing.  And can you also add a integration test.

packages/react-core/src/components/ChipGroup/ChipGroup.tsx
joeldmyers: From what I understand, line 156 in ChipGroup.tsx is generating all the non-overflow chips passed through as children, and then this last Chip is by definition always the overflow chip This chip is only rendered if numChildren > numChips (i.e., if we have overflow).  Here, there is the `isOverflowChip` which is shorthand for `isOverflowChip={true}`. 

So this code here should be only applicable to the overflow chip.

nicolethoen (commented):
Once the prop is included in relevant tests, it looks good :)

joeldmyers:
OK I added a demo component with this, as well as an integration test.  Hopefully I understood everything you wanted correctly; please lmk if any other changes are needed.  Thanks!

tlabaj (approved):
Lgtm

nicolethoen (approved):

tlabaj (approved):

PR #5275, opened 2020-12-17
fix(icons): removing inline styles from SVG icons

### LINKED ISSUE #4767: Breadcrumbs dividers has an inline style for vertical alignment
The svg of the Breadcrumb right arrow, has a 'vertical-align: -0.125em;' inline style applied.
Projects with strict CSP which forbid inline styles, have misaligned arrows looking like this:

Screen Shot 2020-09-02 at 11 57 55

wojta:
**What**: 
Closes #4767

**How**
* removing inline style with `vertical-align: -0.125em;`
* replacing with CSS style

**Additional issues**:
https://github.com/patternfly/patternfly/issues/3774


packages/react-styles/src/css/base/patternfly-svg.css
```diff
@@ -0,0 +1,5 @@
+@charset "UTF-8";
+
+.pf-svg-vertical-align {
```
wojta, line 3: I don't know exactly where to put this CSS class. It would require manually importing this file in the every application for this to work properly.

redallen: Yes, this belongs in https://github.com/patternfly/patternfly or existing applications will not get the styles.

mcoker: @wojta we have a global PF class stylesheet here where you could add this class https://github.com/patternfly/patternfly/blob/master/src/patternfly/base/_common.scss. Let me know if you'd like me to create the issue/PR.

wojta: @mcooker added there https://github.com/patternfly/patternfly/pull/3775

mcoker (commented):
I'm not seeing the styles applied - looks like this needs a core bump?

wojta:
Yes, that's needed.

tlabaj:
The core version has been updated.  You can rebase this PR now to get the styles

wojta:
Rebased, demos look good, styles are there now.

tlabaj:
this one has a conflict.  Can you please rebase again.

packages/react-icons/src/createIcon.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { css } from '@patternfly/react-styles';
```
redallen, line 2: Remove import.

packages/react-icons/src/createIcon.tsx
```diff
@@ -66,13 +67,11 @@ export function createIcon({
...
       const hasTitle = Boolean(title);
       const heightWidth = getSize(size);
-      const baseAlign = -0.125 * Number.parseFloat(heightWidth);
-      const style = noVerticalAlign ? null : { verticalAlign: `${baseAlign}em` };
       const viewBox = [xOffset, yOffset, width, height].join(' ');
 
       return (
         <svg
-          style={style}
+          className={noVerticalAlign ? undefined : css('pf-svg-vertical-align')}
```
redallen, line 74: Change `css('pf-svg-vertical-align')` to `'pf-svg-vertical-align'`

mcoker:
Everything looks good visually, though a couple of discrepancies I see that I'd like to confirm:

* On the [icons page](https://patternfly-react-pr-5275.surge.sh/icons), the `vertical-align` value is different. The inline style is currently `-0.375em`, while the `.pf-svg-vertical-align` class is fixed at `-0.125em`. However, the icons page is the only place I've seen anything other than `-0.125em` used as the vertical align value.
* Spot checking a few components, it looks like we're currently applying `vertical-align` inline on svg's, yet in this PR, we aren't applying the `.pf-svg-vertical-align` class. Examples are empty state, button, and data-list.

redallen (changes requested):
Since you put 'pf-svg-vertical-align' in the common styles there's no to introduce `@patternfly/react-styles` as a dependency. Someone might want to use icons without styles.

packages/react-icons/src/createIcon.tsx
wojta: ok

packages/react-icons/src/createIcon.tsx
wojta: ok

packages/react-icons/src/createIcon.tsx
```diff
@@ -66,13 +66,11 @@ export function createIcon({
...
       const hasTitle = Boolean(title);
       const heightWidth = getSize(size);
-      const baseAlign = -0.125 * Number.parseFloat(heightWidth);
-      const style = noVerticalAlign ? null : { verticalAlign: `${baseAlign}em` };
       const viewBox = [xOffset, yOffset, width, height].join(' ');
 
       return (
         <svg
-          style={style}
+          className={noVerticalAlign ? undefined : 'pf-svg-vertical-align'}
```
redallen, line 73: To fix the few issues @mcoker mentioned destructure `className` prop and change to ```className={`${className}${noVerticalAlign ? '' : 'pf-svg-vertical-align'}`}```.

redallen (changes requested):
Sorry I missed @mcoker 's comment yesterday!

redallen (approved):

mcoker (commented):
This looks good, though I'm wondering if the reason the the icons on the [icons page](https://patternfly-react.surge.sh/icons) are `0.375em` is because of this code: 

https://github.com/patternfly/patternfly-react/blob/2c77e2676a24b184befbc5c9174b8dc8537cbcd5/packages/react-icons/src/createIcon.tsx#L69-L70

That shouldn't apply to most of the patternfly components, since we set the icon size in the CSS... but from looking at the [icon page in this PR](https://patternfly-react-pr-5275.surge.sh/icons), I'm assuming this change will apply a fixed vertical align via `.pf-svg-vertical-align` instead of a dynamic one based on the icon size for users who have chosen a different icon size?

wojta:
Ok, now I understand. We can get the behavior back when height is set for the icon. But this would include inline `style`.

mcoker:
@wojta since the size serves as a multiplier for `0.125em`, we can just set the size as an inline style, and use a `calc()` function in the CSS to do the multiplication. Here's an example - https://codepen.io/mcoker/pen/xxREaBQ

Also just thinking about this change - could react-icons be used on its own before this change? I'm wondering if this is breaking at all since the vertical-align style now comes from a PF stylesheet, so it's a dependency where it wasn't before.

wojta:
@mcoker I think we would need to set inline style in that case as well. Can we use `calc()` somehow to calculate it from the element height? Because we have 4 sizes allowed, what about having separete CSS classes `pf-svg-vertical-align-sm`, `pf-svg-vertical-align-md`,`pf-svg-vertical-align-lg`,pf-svg-vertical-align-xl`?

redallen:
> I'm wondering if this is breaking at all since the vertical-align style now comes from a PF stylesheet, so it's a dependency where it wasn't before.

Yes, this is technically breaking. To get around that we'd have to move the CSS from the common stylesheet to another stylesheet (not sure where?) so `react-styles` can make a separate CSS file for us to include.

wojta:
sorry, accidentally pushed wrong version, will fix tomorrow

wojta:
This needs to be merged first https://github.com/patternfly/patternfly/pull/3871, then I'll update the dependencies.

tlabaj:
@wojta can you please rebase this one.

tlabaj (approved):

kmcfaul (approved):

packages/react-icons/src/createIcon.tsx
```diff
@@ -7,18 +7,18 @@ export enum IconSize {
...
-      return '2em';
+      return 'pf-svg-size-lg';
     case IconSize.xl:
-      return '3em';
+      return 'pf-svg-size-xl';
     default:
-      return '1em';
+      return 'pf-svg-size-sm';
   }
 };
```
redallen, line 23: I can't find the definition for these classes. Did you mean to add them to [base/_common.scss](https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/_common.scss)?

redallen (changes requested):
It looks like you want to remove the `height` and `width` attributes in addition to `style`. However, I can't find definitions for the classes you're trying to replace them with.

You can either settle for just removing the `style` attribute with `pf-svg-vertical-align` or add the size classes upstream.

packages/react-icons/src/createIcon.tsx
mcoker: @redallen we have an open core PR to add the classes - https://github.com/patternfly/patternfly/pull/3871

dlabrecq (approved):

jenny-s51:
Reopening as part of #5682

packages/react-icons/src/createIcon.tsx
```diff
@@ -62,20 +62,16 @@ export function createIcon({
...
     render() {
-      const { size, color, title, noVerticalAlign, ...props } = this.props;
+      const { size, color, title, noVerticalAlign, className, ...props } = this.props;
 
       const hasTitle = Boolean(title);
-      const heightWidth = getSize(size);
-      const baseAlign = -0.125 * Number.parseFloat(heightWidth);
-      const style = noVerticalAlign ? null : { verticalAlign: `${baseAlign}em` };
       const viewBox = [xOffset, yOffset, width, height].join(' ');
+      const classes: String[] = [className, noVerticalAlign ? '' : getSizeClass(size)].filter(t => t);
```
mcoker, line 69: Can you explain like I'm 5 what this does? 🙃 Just want to make sure it allows you to set a size class and `noVerticalAlign` independently (both set or not, or either set and the other not)

mcoker (commented):
I have a few questions about this issue. 

* Looks like the size modifiers aren't working, probably because this file needs to be imported - https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/patternfly-icons.scss. That's the result of this PR: https://github.com/patternfly/patternfly/pull/3871/files
  * That file imports [_icons.scss](https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/_icons.scss), which pulls in our class-based icon font rules for font awesome and the custom pficons set. I'm assuming that isn't needed? Here's the compiled CSS for that file - https://unpkg.com/browse/@patternfly/patternfly@4.219.2/base/patternfly-icons.css
  * I'm also wondering if we want to have size modifiers at all, and if so, if they should align with core's size modifiers. We have the icon component now that can do that. FWIW, core's icons do not support size modifiers, they're controlled by inheriting the text size (react-icons are, too).

packages/react-icons/src/createIcon.tsx
jenny-s51: Ahh I see - it looks like this logic was not setting a size class if `noVerticalAlign` was set. Just pushed a fix for this

jenny-s51 (commented):
Thank you for looking at this @mcoker!

> * Looks like the size modifiers aren't working, probably because this file needs to be imported - https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/patternfly-icons.scss. That's the result of this PR: https://github.com/patternfly/patternfly/pull/3871/files
 
>   * That file imports [_icons.scss](https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/_icons.scss), which pulls in our class-based icon font rules for font awesome and the custom pficons set. I'm assuming that isn't needed? Here's the compiled CSS for that file - https://unpkg.com/browse/@patternfly/patternfly@4.219.2/base/patternfly-icons.css

I noticed this as well and have been looking into this! Currently we cannot import from the `base` directory as it does not seem to get generated in the build, so importing this file directly into `createIcon.tsx` causes some issues. This could 
potentially require some build-related updates. Investigating...
>   * I'm also wondering if we want to have size modifiers at all, and if so, if they should align with core's size modifiers. We have the icon component now that can do that. FWIW, core's icons do not support size modifiers, they're controlled by inheriting the text size (react-icons are, too).

This is a good question! Is https://github.com/patternfly/patternfly-react/issues/8263 the related issue for this?

nicolethoen (approved):

mcoker:
@jenny-s51, yep - https://github.com/patternfly/patternfly-react/issues/8263 is the issue for whether react-icons needs a size prop at all, and whether the sizing should align with PF's icon sizes or if they should stay the way they are now (1em, 1.5em, 2em, 3em) as something more generic.

A couple of questions:

* If this update makes requires core styles be imported into an app as a dependency for react-icons - do we want that? If possible, we could define style classes in this package, and if they need to align with core, we can import the tokens from core instead and use them to do something like this (horrible code, just to illustrate). It gives a PF core var that can 1) come from PF if the PF styles are imported, and 2) be customized/themed by the user, and sets a default if the var isn't present that comes from the stylesheet in core. This is sorta how react-charts works.
  ```
  @import icon-size-sm from '@patternfly/react-tokens/dist/esm/[icon-sm-token]';

  <style>
  .react-icon-sm {
    width: var({icon-size-sm.var}, {icon-size-sm.value});
    height: var({icon-size-sm.var}, {icon-size-sm.value});
  }
  </style>
  ```
* I'm wondering if by default, these icons can be 1(w) x 1(h) x -0.125em(lh) without a size class. Right now we assign a small class if no size is passed. IMO no class would be better (default size - 1x1), and if we maintain a size prop, users can assign a small, medium, large, or custom size.
  * That would imply these icons get some kind of base class to identify what they are, and they have a default size, with modifiers for size variations.
* Should we remove "breaking change" from the commit title? I'm not sure what our semantic release process will do with that 😅

jenny-s51 (commented):
In our recent team discussion regarding this PR at PF office hours, we decided that we will need to do the following. @mcoker Please let me know if there is anything I missed 🙂

- [x] Open an issue in core to remove the size variations we added (`pf-svg-size-[value]`), and instead replace them with a new class that would attach to all of the icons with the same attribute values  (`height="1em", style="vertical-align: -0.125em;" width="1em"`)

- [ ] Remove `size`, `noVerticalAlign`, and `color` props from `createIcon.tsx`

- [ ] Update the README of react-icons to say that react-styles is a dependency, and that the svg styling comes from there. Then provide the generic styles (same values, see above) which is the recommendation font-awesome makes.

nicolethoen (approved):

gitdallas (approved):

thatblindgeye (approved):

packages/react-core/src/components/FileUpload/examples/FileUploadCustomPreview.tsx
```diff
@@ -29,7 +29,7 @@ export const CustomPreviewFileUpload: React.FunctionComponent = () => {
     >
       {value && (
         <div className="pf-u-m-md">
-          <FileUploadIcon size="lg" /> Custom preview here for your {value.size}-byte file named {value.name}
+          <FileUploadIcon width="2em" height="2em" /> Custom preview here for your {value.size}-byte file named {value.name}
```
srambach, line 32: I'm not sure why this change, but I'm also not a React dev :-)

jenny-s51: The `size` prop was removed in `createIcon.tsx`, so I had to set this custom styling here to match the styling we had before @srambach - not sure if there is a better way to do this

srambach: Ok, I see that now. 🤷‍♀️

srambach (approved):

mcoker (approved):
Nice job @jenny-s51! 🚀

tlabaj (approved):
LGTM Jenny!
Do we have a code mod issue for this?

tlabaj (commented):
@jenny-s51 We can merge this one once it is rebased

PR #5292, opened 2020-12-21
fix(radio/checkbox): show error message if id is missing

### LINKED ISSUE #5269: bug(Radio/Checkbox): when 'id' property is not used the component are not accessible
The documentation is mentioning that this is a required property but it's really easy to miss. I am familiar, but I just noticed that many of our PF4 Radio components in cockpit were inaccessible because I was not using 'id's.

Is there a way that rendering the component would throw an error if the required properties are not passed?

boaz0:
**What**: Closes #5269


tlabaj (approved):
LGTM

dlabrecq (approved):

jessiehuff (approved):
LGTM! :)

boaz0:
Thank you all! :bow: :smile:

PR #5293, opened 2020-12-22
Feat(PageHeader) - Changing logoComponent to revert to span if it is a and no href is provided

### LINKED ISSUE #4665: Masthead: default element for MastheadBrand
https://github.com/patternfly/patternfly/issues/3332

> My thought would be to only make the logo an `a` tag if there's an `href`. In this context an `<a>` tag without an `href` doesn't make much sense.

Referencing the [PageHeader component](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Page/PageHeader.tsx) - since `logoComponent` is an `a` by default, it needs some sort of link attribute for that to be the appropriate element to use there. I wonder if there is a good way to detect whether the element is suitable as an `<a>` or some other element? As in if it has an `href` or sets a location on click, it's rendered as an `<a>`? If it just has a click handler, maybe it should be a `<button>` instead, or `role="button"`, since [an `<a>` without a `href` has no implicit role](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Properties). Otherwise it's rendered as a `<span>`?

I see you can pass a `logoComponent` to achieve these variations, but was curious if there is a way we could do this detection for the user and default to a `<span>` 

UPDATE: The PageHeader and it's `logoComponent` prop are deprecated in v5. But we would want to make this same change to the MastheadBrand component, so that it's only an anchor when an href is provided, otherwise it'd default to a div (and can be overriden by MastheadBrand's `component` prop)

joeldmyers:
Proposed - adds a default href='/' in logoProps. 

#4665 pointed out that it doesn't make sense for the logoComponent to default to an `a` if it doesn't have a link attribute.  

I see three options for fixing this: 

1.  (done in this PR) - leave the logoComponent default to "a", and have a default href="/" in logoProps.  This seemed the cleanest and most straightforward fix for this to me.  I reviewed a number of sites and the vast majority have their header logo link to their home page. 

2. Default the logoComponent to  be a `span`, with no default href - this may risk breaking any settings who are using the default "a" but have an href or other in logoProps.

3. As suggested in #4665, have some kind of detection of what is being passed in and conditionally show it as span or a depending.  This could possibly be achieved with something like this - 

```
const getDefaultLogoComponentType = (logoProps: { [key: string]: any }): string => {
  let logoComponentType = 'span';
  if (logoProps.href || logoProps.onClick) {
    logoComponentType = 'a';
  }
  return logoComponentType;
};


export const PageHeader: React.FunctionComponent<PageHeaderProps> = ({
  ...
  logoComponent = getDefaultLogoComponentType(logoProps),
  ...props
}: PageHeaderProps) => {

```

However, testing out all the various ways this could be set up, with different router types etc., could get messy, and  to me, somewhat defeats the purpose of having defaultProps.  

Please LMK thoughts on this front.

**What**: Closes #4665

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Page/PageHeader.tsx
+++ b/packages/react-core/src/components/Page/PageHeader.tsx
@@ -6,13 +6,18 @@ import BarsIcon from '@patternfly/react-icons/dist/js/icons/bars-icon';
 import { Button, ButtonVariant } from '../../components/Button';
 import { PageContextConsumer, PageContextProps } from './Page';
 
+interface LogoProps {
+  href?: string;
+  ['key']: any;
+}
+
 export interface PageHeaderProps extends React.HTMLProps<HTMLDivElement> {
   /** Additional classes added to the page header */
   className?: string;
   /** Component to render the logo/brand, use <Brand /> */
   logo?: React.ReactNode;
   /** Additional props passed to the logo anchor container */
-  logoProps?: object;
+  logoProps?: LogoProps;
   /** Component to use to wrap the passed <logo> */
   logoComponent?: React.ReactNode;
   /** Component to render the header tools, use <PageHeaderTools />  */
@@ -36,7 +41,7 @@ export interface PageHeaderProps extends React.HTMLProps<HTMLDivElement> {
 export const PageHeader: React.FunctionComponent<PageHeaderProps> = ({
   className = '',
   logo = null as React.ReactNode,
-  logoProps = { href: '/' },
+  logoProps = null as LogoProps,
   logoComponent = 'a',
   headerTools = null as React.ReactNode,
   topNav = null as React.ReactNode,
@@ -49,7 +54,12 @@ export const PageHeader: React.FunctionComponent<PageHeaderProps> = ({
   'aria-controls': ariaControls = null,
   ...props
 }: PageHeaderProps) => {
-  const LogoComponent = logoComponent as any;
+  let detectedLogoComponent = logoComponent;
+  if (logoComponent === 'a' && !logoProps?.href) {
+    detectedLogoComponent = 'span';
+  }
+  const LogoComponent = detectedLogoComponent as any;
+
   if ([false, true].includes(deprecatedIsManagedSidebar)) {
     console.warn(
       'isManagedSidebar is deprecated in the PageHeader component. To make the sidebar toggle uncontrolled, pass this prop in the Page component'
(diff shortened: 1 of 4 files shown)


mcoker:
@joeldmyers I think this is ideally how the component could be setup, I'm just concerned about introducing something that may be a breaking change. Looks like with this, we would be inserting a link where there may have not been one previously, and we can't ensure the link location is valid. Also if a user is currently passing `logoComponent="span"`, the component will render as `<span href="/">`.

I'm curious to hear what others think.

joeldmyers:
Hey @mcoker, after some thought, I agree with you that adding the href might result in undesired breaking changes. I switched this to now have a simple function that changes the logoComponent to span if no href is provided. 

I am not 100% sure that this resolves the concern around breaking changes, so please take a look and let me know thoughts.

mcoker:
@joeldmyers this seems better, though I suppose you could still have a situation where we unexpectedly change the element type, if someone has passed something like `logoProps = { [id/name]: 'some-id' }` expecting it to render as `<a [id/name]="some-id">`. While that would still work with a `<span>`, that would likely still be a breaking change.

I'm also wondering if there are other ways a person could want to use an `<a>` here without passing an `href`.

joeldmyers:
@mcoker good point about the name or id, though I'm a bit conflicted about the ID - isn't it possible for someone to add an ID but not an href here, and mean for the ID to be specifically for CSS (i.e., not intend for it to be clickable)? Maybe remote, but not sure.  I also am not sure how many use cases there may be for IDs for logo links.  

I pushed a change, andn added in more checks for name, target, and onClick.  Let me know thoughts about whether I should add ID or anything else.  

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

Also this was probably thought about but if the core issue is UX concern about having `cursor: pointer` for <a> elements with no href, maybe something like this would be a better fix? 

```
.pf-c-page__header-brand-link:not([href]) {
  cursor: default;
}
```

According to this, an a without href is still valid - https://www.w3.org/TR/2011/WD-html5-20110525/text-level-semantics.html#the-a-element

Let me know thoughts. Thanks!

packages/react-core/src/components/Page/PageHeader.tsx
```diff
@@ -6,13 +6,21 @@ import BarsIcon from '@patternfly/react-icons/dist/js/icons/bars-icon';
...
+}
+
 export interface PageHeaderProps extends React.HTMLProps<HTMLDivElement> {
   /** Additional classes added to the page header */
   className?: string;
   /** Component to render the logo/brand, use <Brand /> */
   logo?: React.ReactNode;
   /** Additional props passed to the logo anchor container */
-  logoProps?: object;
+  logoProps?: LogoProps;
```
tlabaj, line 23: Changing the type here might break consumers.

tlabaj (commented):
This is a good solution.  However, it would break the consumer.  I think we will need to hold of on making this update until we have our next breaking change release.

joeldmyers:
OK sounds good, thanks!

tlabaj:
Closing this one for now since it is a breaking change.

Issue #4795, opened 2020-09-08
Code editor component


mcarrano:
Introduce the Code snippet component per https://github.com/patternfly/patternfly/issues/3419

Design: https://marvelapp.com/prototype/b284chd/screen/70964526

@jessiehuff to provide input keyboard interaction. Reference this OS issue for more background: https://issues.redhat.com/browse/PD-598

jessiehuff:
It looks like there's currently a common practice among some code editors, Monaco included, that do something called "[Tab Trapping](https://github.com/microsoft/monaco-editor/wiki/Monaco-Editor-Accessibility-Guide#tab-trapping)". As you're tabbing, you become trapped in the code editor, and while there is a way out, it's not easily discoverable. (You can toggle the trapping of Tab with Ctrl+M on Windows and Linux and with Ctrl+Shift+M on OSX, and subsequent Tab keys will move focus out of the editor.) 

Depending on what we're capable of doing here, I think we should either:
- Consider a new interaction pattern: an idea I think we could follow is that you would tab through the page, focus goes to the editor itself (as opposed to immediately entering the editor), and you would press enter to interact with it. Then, similar to other interaction patterns like a modal or dropdown, pressing Esc would exit the inside of the editor and put focus back on the entire editor/outer div. This appears to me to be a little more intuitive, but that's up to debate as most interactions can be. I couldn't easily find a standard for code editors specifically, but I will continue to research this.
- Make the instructions on how to leave the editor significantly clearer and nearly require products to provide this documentation to users. Otherwise, it would appear to completely violate WCAG 2.1.2 [No Keyboard Trap](https://www.w3.org/WAI/WCAG21/quickref/#no-keyboard-trap) which would likely negatively affect products if they ever go to create a VPAT.

nicolethoen:
- forward ref to editor, maybe editordidmount function
- prop to hide language
- auto fill parent container
- configure style?
- toggle padding?
- does monaco need to know the language to read the file type? does the uploaded file need to be the right extension?

PR #5294, opened 2020-12-22
feat(CodeEditor): add new code editor

nicolethoen:
Closes: #4795

https://patternfly-react-pr-5294.surge.sh/components/code-editor/


mcarrano:
@nicolethoen I took a quick look at this today. It's looking good. I just had a couple of questions/comments:

* In read-only mode, I'm still able to get a text cursor even though I can't enter anything. Per the design here, https://marvelapp.com/prototype/b284chd/screen/70964527, when in read-only mode the user should see a gray background and not be able to interact with the editor. There is a core example here: https://www.patternfly.org/v4/components/code-editor#read-only
* All tool buttons should have tooltips per the design.
* I can't seem to be able to make upload work. I can choose a file, but not seeing anything loaded into the editor. Should this work?

Let me know if you have any questions after looking at the Marvel document. @gdoyle worked on this and can also be consulted with any design questions.

cc @evwilkin

nicolethoen:
@mcarrano 

> * In read-only mode, I'm still able to get a text cursor even though I can't enter anything. Per the design here, https://marvelapp.com/prototype/b284chd/screen/70964527, when in read-only mode the user should see a gray background and not be able to interact with the editor. There is a core example here: https://www.patternfly.org/v4/components/code-editor#read-only

- the read only styling will need to be handled by core, i think. I've applied the readOnly styles provided, but they do not currently impact anything inside the actual editor.

> * All tool buttons should have tooltips per the design.

- i can do that!

> * I can't seem to be able to make upload work. I can choose a file, but not seeing anything loaded into the editor. Should this work?

- Core and I still need to create/implement error states, but the uploaded file type needs to match the specified language, that may be why it is not working.

evwilkin:
@nicolethoen the tooltip isn't displaying if you click on the icon, only if you click in the blank space around the icon.

> I can't seem to be able to make upload work. I can choose a file, but not seeing anything loaded into the editor. Should this work?

After trying several file types a `.txt` file uploaded & opened fine. Per your suggestion I added the `language="javascript"` prop and then the `.txt` file wouldn't upload but the `.js` file started working as expected for upload 👍  I think that if this is the expected behavior we should call it out in the docs somewhere.

mcarrano:
> the uploaded file type needs to match the specified language, that may be why it is not working.
Yes, thanks for the tip. It works when I correctly set the prop. Agree with @evwilkin that we should document this somewhere.

> the read only styling will need to be handled by core, i think. I've applied the readOnly styles provided, but they do not currently impact anything inside the actual editor.

@mcoker can you look at/comment on this?

mcoker:
This looks good to me. The component is pretty simple on the core side. We allow for an HTML `<pre>` tag that uses our monospace font "Liberation Mono", which I'm assuming would be used to represent a static code block (not using the monaco editor). That looks like this:

Screen Shot 2021-01-04 at 5 54 46 PM

The same block in the monaco editor looks like this

Screen Shot 2021-01-04 at 5 54 40 PM

I'm not sure if we would want to create some sort of parity in the font family, size, etc?

It also seems like it might look better if the padding around the editor were changed. Here is what it looks like without the padding. You achieve that by not including the `pf-c-code-editor__code` element. I think our intent with that element is for when static code is placed in the editor, though we may want to change that so that static code and monaco editor code look more similar.

Screen Shot 2021-01-04 at 6 02 14 PM

> the read only styling will need to be handled by core

Assuming there isn't something out of the box we can use built in to the editor, and we don't want to go another route to indicate that the editor is read-only (couple of ideas below), we probably want to look at what options are available to style/theme it, and apply some sort of read only styles. 

Screen Shot 2021-01-04 at 6 17 02 PM
Screen Shot 2021-01-04 at 6 14 49 PM

packages/react-code-editor/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "tslib": "1.13.0"
+  },
+  "peerDependencies": {
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
+  },
+  "devDependencies": {
+    "@monaco-editor/react": "3.7.2",
+    "monaco-editor": "0.21.2",
+    "react-dropzone": "9.0.0",
```
redallen, line 45: Move these 3 to dependencies.

packages/react-code-editor/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@patternfly/react-code-editor",
+  "version": "0.0.0",
```
redallen, line 3: Change to 4.0.0

packages/react-code-editor/package-lock.json
```diff
@@ -0,0 +1,35 @@
+{
```
redallen, line 1: Remove this file.

redallen (changes requested):
What's our plan for language support?

nicolethoen:
> What's our plan for language support?

I know of no particular plan. 
The editor has a fairly extensive list of languages it supports. I built some enums and mappings for languages/extensions, so right now the consumer is limited to the languages I specified there. We can add more. I know that'd need to be documented.

I'm open to suggestions or insights. My initial approach was a bit of a stab in the dark to get this PR up.

redallen:
Oh I completely missed the `language` prop which works splendidly!

I'm noticing that this PR loads 4 files from `cdn.jsdelivr.net`:
image

PatternFly has never required an external CDN... do you know if it's possible to import these files directly instead?

gdoyle1:
@nicolethoen is there a way we can make the react code flexible enough to include custom actions? I was assuming that download copy and upload were just examples... depending on the use case, they may need to be different. Here's an example of what I mean where the terminal action would run that command in a web terminal. Created this by just editing the HTML component that's been released
Screen Shot 2021-01-08 at 11 07 46 AM

nicolethoen:
@gdoyle1 I am happy to try! I'll let you know if I run into any issues

mcarrano:
@gdoyle1 Do you think we should have some actions that are permanent/standard? Like Upload or Copy to Clipboard? Or should all the actions just be passed through a prop? I guess that would require the consumer to pass an icon, and callback function, and possibly a tooltip for each action. Thoughts @nicolethoen ?

nicolethoen:
@mcarrano i could do it so that the actions are customizable, i think. But then the example which implements the upload/download/copy would probably need to be a demo because the user would need to build the empty state and the dropzone in that case. is that okay?

mcarrano:
I don't know @nicolethoen . I kind of like that these features are built into the component.

@gdoyle1 do you see any reason you would want to remove or customize any of the standard features? What is we made it so you could add additional actions which could be custom?

gdoyle1:
> I don't know @nicolethoen . I kind of like that these features are built into the component.
> 
> @gdoyle1 do you see any reason you would want to remove or customize any of the standard features? What is we made it so you could add additional actions which could be custom?

@mcarrano Allowing additional actions that are custom would be fine - as long as you can optionally remove a standard feature (if you didn't offer uploading in the code editor, you can remove that action), then I think we can just add a "custom" action additionally

gdoyle1:
@nicolethoen here's the OpenShift keyboard shortcut accessibility add-on feature I was referring to in our meeting
Screen Shot 2021-01-14 at 11 45 43 AM
Screen Shot 2021-01-14 at 11 45 38 AM

christianvogt:
image

nicolethoen:
@redallen  
> PatternFly has never required an external CDN... do you know if it's possible to import these files directly instead?

Would this solution work in PF?
If it looks promising to you i'll reach out so you can help me understand it.
https://github.com/suren-atoyan/monaco-react#config

redallen:
Yes that looks promising! I'm happy to dig into the source to figure out how to load from node_modules instead.

kmcfaul:
Everything seems good to me. Only thing I noticed is for the keyboard interaction, `Shift + Tab` takes you out of the editor while `Escape` does nothing. I think either keybind works, just curious if `Escape` should also have that behavior?

nicolethoen:
@kmcfaul I will look into it. 
I do know that the editor uses many key controls already so I'll have to investigate if the escape button already has some function in the editor. The editor's key controls are extensive so I'm not sure.

kmcfaul:
Yeah understandable. I didn't notice it do anything but perhaps there is an edge case. But in any case I'm fine with the current keyboard handling, I don't think this is a blocker or anything.

jessiehuff (commented):
I was actually just about to chime in about that too haha. 🙂   I think that the current keyboard implementation is SO much better than what is defaulted by Monaco. I'm not sure what key commands that the Monaco editor uses though so I'm not positive what's available to us and what's not. I checked with Jenn on this one as well because there's not really a common standard on this pattern so I wanted to make sure that we talked about it. Jenn agreed that this is currently pretty awesome. 

When talking it out though, we wondered if Shift + Tab would conflict with anything in Monaco? We talked about how maybe we could also implement using Enter to get into the editor as well as the down arrow key, and like Katie's comment, we wondered if the Esc key would feel more natural to escape than the Shift + Tab. Though, truthfully, if it stayed the way it is now, I still think it's a massive improvement to the "Tab trapping" that Monaco does. But I think which interaction we use is definitely open for debate. It's a subjective thing that not a lot of people have done before from what I can find. (Another cool way that we're paving the way for accessibility! 😄 )

We also discussed whether we should include instructions in some way for users as part of the design because we don't know what keyboard users would expect (since it's not a common or firm standard). However, I know that's more of a design question. 🙂

nicolethoen:
@jessiehuff 
so, in the first example, if you start typing, you'll see that (since the language is set to javascript) as you type, an autocomplete or intellisense feature that pops up. Escape closes that box.

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,455 @@
...
+  /** Name of the file if user downloads code to local file */
+  downloadFileName?: string;
+  /** Flag to add upload button to code editor actions. Also makes the code editor accept a file using drag and drop*/
+  isAllowUpload?: boolean;
+  /** Flag to add download button to code editor actions*/
+  isAllowDownload?: boolean;
+  /** Flag to add copy button to code editor actions*/
+  isAllowCopy?: boolean;
+  /** Flag to include a label indicating the currently configured editor language */
+  isLanguageLabel?: boolean;
```
tlabaj, line 69: How about `isLanguageLabelVisible`

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,455 @@
...
+  /** Content to display in space of the code editor when there is no code to display */
+  emptyState?: React.ReactNode;
+  /** Name of the file if user downloads code to local file */
+  downloadFileName?: string;
+  /** Flag to add upload button to code editor actions. Also makes the code editor accept a file using drag and drop*/
+  isAllowUpload?: boolean;
+  /** Flag to add download button to code editor actions*/
+  isAllowDownload?: boolean;
+  /** Flag to add copy button to code editor actions*/
+  isAllowCopy?: boolean;
```
tlabaj, line 67: how about `isCopyButtonVisible`

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,455 @@
...
+  toolTipMaxWidth: string;
+  /** The position of tooltips on all buttons */
+  toolTipPosition: 'auto' | 'top' | 'bottom' | 'left' | 'right';
+  /** A single node or array of nodes - ideally CodeEditorControls - to display above code editor */
+  customControls?: React.ReactNode | React.ReactNode[];
+  /** Callback which fires after the code editor is mounted containing
+   * a reference to the monaco editor and the monaco instance */
+  onEditorDidMount?: (editor: any, monaco: any) => void;
+  /** Flag to add the minimap to the code editor */
+  isDisplayMinimap?: boolean;
```
tlabaj, line 98: maybe `isMinmapDisplayed` or `isMinmapVisible`

redallen (approved):
LGTM from the build side!

tlabaj (dismissed):
LGTM

tlabaj:
Looks like you have a conflict and linting errors.

kmcfaul (dismissed):
Looks good! Just need to resolve the conflict

mcarrano (commented):
Looks good @nicolethoen . The only question I have from a design perspective is what this example is intended to show: https://patternfly-react-pr-5294.surge.sh/components/code-editor#with-custom-control

This is what I see, even though it looks like you are declaring that a Play button should be added to the toolbar.

Screen Shot 2021-01-25 at 11 16 07 AM

nicolethoen:
> Looks good @nicolethoen . The only question I have from a design perspective is what this example is intended to show: https://patternfly-react-pr-5294.surge.sh/components/code-editor#with-custom-control
> 
> This is what I see, even though it looks like you are declaring that a Play button should be added to the toolbar.
> 
> Screen Shot 2021-01-25 at 11 16 07 AM

@mcarrano if you begin putting code in the editor, the controls appear. It's a custom `execute` command, which only makes sense if there is code in the editor to execute.

mcarrano (approved):
> @mcarrano if you begin putting code in the editor, the controls appear. It's a custom execute command, which only makes sense if there is code in the editor to execute.

Thanks for the clarification @nicolethoen . I would prefer that the buttons be visible and disabled in this case, but I will not hold up the release for this. I'm OK to open a follow-up issue for this.

tlabaj (approved):

kmcfaul (approved):

redallen (approved):

Issue #5295, opened 2020-12-22
ChartLegendTooltip shown with no data


dlabrecq:
The `ChartLegendTooltip` may be shown incorrectly when there is no visible data for a given coordinate. 

In the example below, all but one legend item is active. When the user hovers over the chart, in an area with no visible data, only the tooltip title is shown. In this case, the flyout's height and width were not set because there were no active and visible data series. 

Note that we calculate the tooltip height and width based on the datum text length and visible legend label. Ideally, the tooltip would not be shown at all for this scenario.

![before](https://user-images.githubusercontent.com/17481322/102923930-9a396780-445e-11eb-951a-2d4a20034000.gif)

dlabrecq:
In Cost Management, we work around this by providing a null value for missing / unavailable data. That way, we can return a "no data" string to display in the tooltip instead.

![final](https://user-images.githubusercontent.com/17481322/102939399-37ef5f80-447c-11eb-80ec-604c5d5f4a32.gif)

PR #5296, opened 2020-12-22
fix(charts): ChartLegendTooltip shown with no data

dlabrecq:
The ChartLegendTooltip may be shown incorrectly when there is no visible data for a given coordinate.

In the "before" example below, all but one data series / legend item is active. When the user hovers over the chart, in an area with no visible data, only the tooltip title is shown because there is no data / legend labels to display here. The tooltip should not be shown at all for this scenario.

Fixes https://github.com/patternfly/patternfly-react/issues/5295

**Before**
before

**After**
after


ddonahue007 (dismissed):
LGTM

redallen:
Merge conflicts in `chart-tooltip.ts`.

dlabrecq:
Fixed the merge conflict

redallen (approved):
Nice fix!

ddonahue007 (approved):
LGTM

PR #5298, opened 2020-12-23
fix(table): empty state should be consistent across demos and examples

### LINKED ISSUE #5297: Rendering empty states in Tables are not consistent
In the Table demos/examples I see two ways that are used to render empty state. I am not sure what is the right way:
- Rendering Empty State below Table: https://patternfly-react.surge.sh/components/table/react-demos#bulk-select
  
image

- Rendering Empty State as a row in Table: https://patternfly-react.surge.sh/components/table/react-demos#loading
image

Looking at the `patternfly-core` it seems like we are supposed to do the latter.

boaz0:
**What**: Closes #5297


kmcfaul (approved):

mcarrano (commented):
This looks fine to me visually. @mcoker can you also take a look to confirm that this is correct?

mcoker (approved):
👍👍

tlabaj (approved):

Issue #5300, opened 2020-12-27
Chart: theme font must use token value


dlabrecq:
In order to position chart labels properly  (e.g., within legends and tooltips), Victory must know the height and width of the text shown. 

To achieve an approximate size for `RedHatText` and `Overpass` fonts, we made use of Victory's `characterConstant` style prop. Although this gave us an approximation of the character average size, it was not perfect (e.g., if the font is changed, the constant may not be correct).

Victory no longer supports the `characterConstant` style property in favor of referencing exact character sizes. Victory's `approximateTextSize` function is now based on measuring specific character widths for the font defined by the given style or theme -- see https://bl.ocks.org/tophtucker/62f93a4658387bb61e4510c37e2e97cf. 

To help resolve the issue, I contributed a PR upstream so Victory can support `RedHatText` and `Overpass` font character widths -- see `https://github.com/FormidableLabs/victory/pull/1761`. However, Victory will still defaults to using Helvetica font character widths. For Victory to properly use `RedHatText` and `Overpass` fonts, our theme must provide a react-token value as Victory's functions do not work with variables.

For example:

```
const TYPOGRAPHY_FONT_FAMILY = chart_global_FontFamily.value;
const TYPOGRAPHY_LETTER_SPACING = chart_global_letter_spacing.value;
```

PR #5301, opened 2020-12-27
fix(charts): theme font must use token value

dlabrecq:
In order to position chart labels properly (e.g., within legends and tooltips), Victory must know the height and width of the text shown. 

Victory no longer supports the `characterConstant` style property in favor of referencing exact character sizes. Victory's `approximateTextSize` function is now based on measuring specific character widths for the font defined by the given style or theme.

I contributed an upstream PR, so Victory can support `RedHatText` and `Overpass` font character widths -- see `https://github.com/FormidableLabs/victory/pull/1761`. However, Victory will still defaults to using Helvetica font character widths. For Victory to properly use `RedHatText` and `Overpass` fonts, our theme must provide a react-token value as Victory's functions do not work with variables.

Fixes https://github.com/patternfly/patternfly-react/issues/5300

**After**
Screen Shot 2020-12-26 at 11 06 05 PM


ddonahue007 (approved):
LGTM

redallen (approved):
I wish the CSS variable could still be used but understand the direction Victory is trying to go.

Issue #5129, opened 2020-11-11
Charts: Incorrect colorScale prop type


dlabrecq:
The chart `colorScale` props use Victory's `ColorScalePropType` type, which was previously defined as a string array. The `ColorScalePropType` type is now defined as:

`"grayscale", "qualitative", "heatmap", "warm", "cool", "red", "green", "blue" | string[]`

PatternFly charts will never use these color string types as they only have meaning for Victory's 'material' and 'grayscale' themes. 

We should change our `colorScale` type to be `string[]` and update the property description to reflect that.

Note: It may be possible to override the `ColorScalePropType` type via src/types/victory.d.ts?

PR #5303, opened 2020-12-28
fix(charts): Incorrect colorScale prop type

dlabrecq:
The chart `colorScale` props use Victory's `ColorScalePropType` type, which was previously defined as a string array. The `ColorScalePropType` type is now defined as:

`"grayscale", "qualitative", "heatmap", "warm", "cool", "red", "green", "blue" | string[]`

PatternFly charts will never use these color string types as they only have meaning for Victory's 'material' and 'grayscale' themes.

Note: `colorScale` still accepts a string array, just as before. Just not the unused strings; "grayscale", "qualitative", "heatmap" etc.

Fixes https://github.com/patternfly/patternfly-react/issues/5129


ddonahue007 (approved):
LGTM

redallen (approved):

PR #5308, opened 2021-01-04
fix(dualListSelector): do not pass custom handlers to root div

### LINKED ISSUE #5307: Warning: Unknown event handler property `onOptionSelect`.
Please see https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/DualListSelector/DualListSelector.tsx#L544

`onOptionSelect` prop is not catched and it's being spread to the div.

It throws:

```jsx
    Warning: Unknown event handler property `onOptionSelect`. It will be ignored.
        in div (created by DualListSelector)
        in DualListSelector (created by DualList)
```

rvsia:
**What**: Closes #5307


nicolethoen (approved):
Nice catch!

redallen (approved):

PR #5312, opened 2021-01-07
fix(Nav): Make nav items keyboard and screen reader accessible

### LINKED ISSUE #5260: Update navigation demos to show accessible experience
Currently when you navigate to the navigation demos by keyboard, you are unable to navigate to the side panel and click other links. This is likely due to the fact that the links do not have href's. However, when href's are added and the user tabs through the page, keyboard focus disappears off the page. You can determine from the link previews at the bottom left that focus is still going to these links even when the panel is closed. I believe that we could fix this by altering the items to display none when the panel is collapsed. However, we need to account for transitioned events so that if we ever add transitions, the panel items don't just disappear before the transition.

This issue was first noticed in OpenShift. When comparing it to the react demos, we noticed that our demos have the same issue. It would make sense to solve this issue in the context of PF first.

Open [any navigation demo](https://www.patternfly.org/v4/components/navigation/react-demos/default-nav/) and attempt to navigate by keyboard. 

Bug

jessiehuff:
**What**: Closes #5260

### DIFF
--- a/packages/react-core/src/components/Nav/NavItem.tsx
+++ b/packages/react-core/src/components/Nav/NavItem.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/Nav/nav';
 import { css } from '@patternfly/react-styles';
 import { NavContext, NavSelectClickHandler } from './Nav';
+import { PageSidebarContext } from '../Page/PageSidebar';
+import { PageContextConsumer, PageContextProps } from '../Page';
 import { useOUIAProps, OUIAProps } from '../../helpers';
 
 export interface NavItemProps extends Omit<React.HTMLProps<HTMLAnchorElement>, 'onClick'>, OUIAProps {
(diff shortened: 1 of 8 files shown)


packages/react-core/src/components/Page/PageSidebar.tsx
```diff
@@ -19,33 +19,49 @@ export interface PageSidebarProps extends React.HTMLProps<HTMLDivElement> {
...
   className = '',
   nav,
   isNavOpen = true,
   theme = 'dark',
   ...props
 }: PageSidebarProps) => (
-  <PageContextConsumer>
-    {({ isManagedSidebar, isNavOpen: managedIsNavOpen }: PageSidebarProps) => {
-      const navOpen = isManagedSidebar ? managedIsNavOpen : isNavOpen;
+  <PageSidebarContext.Provider value={{ isNavOpen }}>
```
jschuler, line 43: I would bring this down a level and wrap the `{nav}` more directly, this way you would also have the definitive answer to whether the nav is open, i.e. (leaving some code out for brevity)
```
<PageContextConsumer>
  <div id="page-sidebar">
     <div className={css(styles.pageSidebarBody)}>
        <PageSidebarContext.Provider value={{ isNavOpen: navOpen }}>
          {nav}
        </PageSidebarContext.Provider>
      </div>
  </div>
</PageContextConsumer>
```

packages/react-core/src/components/Page/PageSidebar.tsx
```diff
@@ -19,33 +19,49 @@ export interface PageSidebarProps extends React.HTMLProps<HTMLDivElement> {
   theme?: 'dark' | 'light';
 }
 
+export interface PageSidebarContextProps {
+  className: string;
+  nav: React.ReactNode;
+  isNavOpen: boolean;
```
jschuler, line 25: are the other ones needed? I think only `isNavOpen` is used

packages/react-core/src/components/Nav/NavItem.tsx
```diff
@@ -43,19 +45,27 @@ export const NavItem: React.FunctionComponent<NavItemProps> = ({
...
-      <Component
-        href={to}
-        onClick={(e: any) => context.onSelect(e, groupId, itemId, to, preventLinkDefault, onClick)}
-        className={css(styles.navLink, isActive && styles.modifiers.current, className)}
-        aria-current={isActive ? 'page' : null}
-        {...props}
-      >
-        {children}
-      </Component>
+      <PageContextConsumer>
```
jschuler, line 52: don't need PageContextConsumer since you have the definitive answer to isNavOpen if you make the changes in my other comment :)

packages/react-core/src/components/Page/PageSidebar.tsx
jessiehuff: Yup, you're right. Wasn't sure if I should leave the others in in case of future stuff, but just isNavOpen is probably better for now.

packages/react-core/src/components/Page/PageSidebar.tsx
jessiehuff: Thank you!! 😄

nicolethoen (approved):

jschuler (approved):

kmcfaul (approved):

tlabaj (approved):
lgtm

Issue #5283, opened 2020-12-18
Resizable drawer: splitter disappears on parent resize


evwilkin:
Resizing parent width to be smaller when drawer is already at min width makes splitter disappear

kmcfaul:
@mcoker Do you mind taking a quick look at this issue from the core-side? From what I can tell, the panel width doesn't matter, it's just when the parent size is narrowed, after a point the `::after` styling that forms the splitter disappears.

mcoker:
@kmcfaul submitted https://github.com/patternfly/patternfly/pull/3788 which should fix this.

kmcfaul:
Closed in #5313

Issue #5280, opened 2020-12-18
Resizable drawer: splitter disappears after resize to 0 width then toggling drawer


evwilkin:
Resize down to 0 width, then toggle drawer open / closed or resize the parent container, the splitter disappears.

kmcfaul:
The splitter isn't disappearing, it's being scrolled out of view. From what I can tell, when the content of the panel is wider/taller than the panel size, it scrolls to the content when re-toggled because the content receives focus.

@mcoker I'm not sure it's possible, but do you happen to know if this scrolling can be stopped?

Setting `minSize` to something larger would resolve the issue, so maybe instead of defaulting to 0 we should default to 25% (the smallest width available of the `widths` prop? Or I can keep the panel content hidden when the width is small (maybe lower than 5%, or through a new property). Would that be preferable or is raising the minSize enough? @evwilkin

kmcfaul:
@mcoker I think you mentioned a PR on the core side that would address this?

mcoker:
@kmcfaul yep, it's https://github.com/patternfly/patternfly/pull/3778 - @DaoDaoNoCode is working on it and it should be merged soon.

kmcfaul:
This is also being caused by our examples, which manually focus the panel content on expansion. I added to the min/max drawer size PR to address this (by giving the panels a minimum size where the focus doesn't scroll the panel).

kmcfaul:
Closed by #5313

PR #5313, opened 2021-01-07
fix(Drawer): initial width doesn't respect min/max widths

### LINKED ISSUE #5278: Resizable drawer: initial size doesn't adhere to min/max size props

kmcfaul:
**What**: Closes #5278 & #5280

Adds a new `defaultSize` prop to control starting width for resizable drawers. 

A resizable drawer will start at `defaultSize`, or `minSize` if `defaultSize` is undefined and `minSize` is given, or the default drawer width if neither property is given.

Added default and min sizes to the examples to address #5280 (because our examples, which focus the drawer content, was also causing trouble for the scrolling).


tlabaj (changes requested):
Looks good. Can you update the integration app and add a test for this please.

jschuler (approved):

tlabaj (dismissed):

jessiehuff (approved):
LGTM! :)

kmcfaul:
Added default and min sizes for our resizable examples because the code, which focuses the drawer panel content on expansion, was also causing the bug in #5280 by scrolling. 

@mcoker I noticed while adding min sizes that for a left expanding resizable drawer, if the min size is too small, the splitter will disappear into the panel before hitting the min width. I think this is because the splitter is after the panel content (which gets the min width)? If you decrease the left-drawer's `minSize` to `150` from `200`, you can see the behavior.

tlabaj (approved):

mcoker:
@kmcfaul 

> If you decrease the left-drawer's minSize to 150 from 200, you can see the behavior.

It looks like if we add the `.pf-c-drawer__panel-main` element adjacent to the splitter in resizable drawers, that should fix this problem.

Also for resizable drawers, we need to update the var we use to customize the width from `--pf-c-drawer__panel--FlexBasis` to `--pf-c-drawer__panel--m-resizable--FlexBasis`. We only apply changes to that var if they're above the mobile layout breakpoint.

kmcfaul:
For clarity, the changes @mcoker suggests I have ready to go pending another core bump, but these don't impact this PR and can be added in a follow up if necessary. This PR should be ready to go in.

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
```diff
@@ -17,8 +17,10 @@ export interface DrawerPanelContentProps extends React.HTMLProps<HTMLDivElement>
   isResizable?: boolean;
   /** Callback for resize end. */
   onResize?: (width: number, id: string) => void;
-  /** The minimum size of a resizable drawer, in pixels. Defaults to the starting width of the drawer. */
+  /** The minimum size of a resizable drawer, in pixels. Defaults to 0. */
   minSize?: number;
+  /** The starting size of a resizable drawer, in pixels. Defaults to minSize. */
```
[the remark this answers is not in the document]
kmcfaul: Oh nice catch for this bug. The code isn't setting the style because 0 === false for the check to add it. I'll fix the logic.

kmcfaul:
Last big update for this PR, pulling in the latest core changes which fixes #5283 as well as updates how min and max size are set and controlled.

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
```diff
@@ -17,17 +17,19 @@ export interface DrawerPanelContentProps extends React.HTMLProps<HTMLDivElement>
   isResizable?: boolean;
   /** Callback for resize end. */
   onResize?: (width: number, id: string) => void;
-  /** The minimum size of a resizable drawer, in pixels. Defaults to the starting width of the drawer. */
+  /** The minimum size of a drawer, in pixels. */
   minSize?: number;
-  /** The maximum size of a resizable drawer, in pixels. Defaults to the max width of the parent container. */
+  /** The starting size of a resizable drawer, in pixels. */
+  defaultSize?: number;
```
mcoker, line 23: is there a requirement for this to be in pixels? The var doesn't care what the unit is - px, %, em vw, etc... and that works with the CSS that keeps it within the `minSize`/`maxSize`

kmcfaul: I could change it to a string property. The user would have to specify `px` or `%` in that case. Would that be okay?

kmcfaul: I had it in pixels so I could limit how much the `mousemove` would calculate and set a new size (throwing out the sizes outside the min/max). But since that's being done in CSS I can probably remove them.

kmcfaul:
Updated `minSize`, `maxSize`, and `defaultSize` to `string` type, to support % boundaries. Removed code-side checking of boundaries because CSS maintains them. Updated docs.

mcoker:
Just a nit but looks like there are 2 "Stacked content body elements" examples. Also I believe the "basic" and "panel on right", and "inline" and "inline panel on right" and are the same?

mcoker (approved):
LGTM! Great work @kmcfaul!! 🥳

tlabaj:
Just need a rebase than it can go in.

tlabaj (approved):

evwilkin (approved):
LGTM

PR #5314, opened 2021-01-07
Feat/jumplink subsection

### LINKED ISSUE #5185: Add jump link subsections
Add the ability to have subsections in a list of vertical jump links per https://github.com/patternfly/patternfly/issues/3683

### LINKED ISSUE #5186: Page enhancements to support Jump links demo
Add a vertical page section to support building a jump links demo with a sticky sidebar that stays static while the page scrolls. See https://github.com/patternfly/patternfly/issues/3653 for more details.

### LINKED ISSUE #5344: Responsive/expandable jump links
https://github.com/patternfly/patternfly/pull/3802

We added 3 variations.

`.pf-m-expandable{-on-[breakpoint]}` - allows you to specify the responsive layout on optional breakpoints
`.pf-m-non-expandable{-on-[breakpoint]}` - specifies the regular layout on optional breakpoints
`.pf-m-expanded` - expands the jump links list when in the responsive layout

Also adds a `__header` element that goes around the label, and a `__toggle` element we can conditionally render if the jump links are expandable. And a `__toggle-icon` element in the toggle for the icon. 

Also if the user doesn't want a label when the jump links are not in the responsive/expandable view, we can add their responsive label as `.pf-c-jump-links__label.pf-m-toggle`, and it will only show up when jump links switches to the responsive layout.

cc @mcarrano

redallen:
**What**: Closes #5185 Closes #5186 Closes #5344

**Additional issues**: ~~Blocked on https://github.com/patternfly/patternfly/issues/3653 which will properly implement a layout for the demo.~~

~~Blocked on core fixing the Stacked Sidebar layout to not be sticky to the top of the scroll container.~~ https://github.com/patternfly/patternfly/pull/3807 has been merged


mcarrano (approved):
Looks good to me. Thanks @redallen

mcoker:
@redallen why did we change the markup? It doesn't follow the same structure as items with nested sections we have in other components (tree view and nav as examples)

redallen:
@mcoker We changed it because I believe this JSX:
```jsx
<JumpLinks>
  <JumpLinksItem href="#a">Section title</JumpLinksItem>
  <JumpLinksList>
    <JumpLinksItem href="#a-a">Subsection title</JumpLinksItem>
  </JumpLinksList>
</JumpLinks>
```
is simpler than the alternative which effectively requires `JumpLinksList` to inherit all of `JumpLinksItem`'s props in order to create the first `<a>` tag inside of it:
```jsx
<JumpLinks>
  <JumpLinksList href="#a" label="Section title"> // Should props spread on the ul or a? Which prop should make the <a> appear? How should we document that this effectively contains a JumpLinksItem?
    <JumpLinksItem href="#a-a">Subsection title</JumpLinksItem>
  </JumpLinksList>
</JumpLinks>
```

This is also problematic because my scrollspy logic is specialized to spy only on `JumpLinksItem`.

jessiehuff (commented):
So I think the structure looks pretty good. The only thing that I noticed is off is that the demo's keyboard and mouse functionality seems off. When you click the jump links or use the keyboard, they don't take you to the one you selected.

redallen:
@jessiehuff I know what's wrong, will fix tomorrow.

jessiehuff (dismissed):
LGTM!

kmcfaul (dismissed):
lgtm. One thought though, do we want to add maybe a transition or animation to the scrolling?

mcoker:
> Blocked on core fixing the Stacked Sidebar layout to not be sticky to the top of the scroll container.

@redallen if we can't use `position: sticky` for when jump links float on top of the things the jump links anchor to, it's going to require more of a rework of the sidebar than I think we can get in this release. That said, I don't think that should hold up releasing subsections if that is a priority. We can just hold off on building the demos for now, and potentially the sidebar component until we sort out the best way to do this.

redallen:
@mcoker Agreed, I can remove the demo that uses subsections and the Sidebar component from this PR if subsections alone is a priority.

mcarrano (approved):
This looks good to me @redallen . The only thing I wondered was if the jump links menu should be open or closed by default in the mobile configuration. Hard to say what people would want. Is that something that can be controlled by the consumer? If so, I think we're good.

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
```diff
@@ -10,6 +16,8 @@ export interface JumpLinksProps extends Omit<React.HTMLProps<HTMLElement>, 'labe
   isVertical?: boolean;
   /** Label to add to nav element. */
   label?: React.ReactNode;
+  /** Flag to always show the label when using `expandable` */
+  alwaysShowLabel?: boolean;
```
mcoker, line 20: Just my $0.02, I would expect that if a label is defined, it's shown, even when the jump links are expandable, and we would have a prop to only show the label as the toggle label in the expandable state (ie, hide it when the vertical tabs aren't expandable). At least that's how the core component is built - the class `.pf-m-toggle` is the opt-in to make the label only show as the toggle text when it's in the expandable layout. Wdyt @mcarrano @tlabaj?

mcarrano: The consumer should be able to control this. Is that what the prop `alwaysShowLabel` does?

packages/react-core/src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,31 @@
...
+  hasGutter,
+  hasNoBackground,
+  ...props
+}: SidebarProps) => (
+  <div
+    className={css(
+      styles.sidebar,
+      hasGutter && styles.modifiers.gutter,
+      hasNoBackground && styles.modifiers.noBackground,
+      variant !== 'default' && styles.modifiers[variant]
```
mcoker, line 24: We don't support this now, but I plan on supporting responsive stack, split, gutter, and panel-right/panel-left variants in the future. So like `stack-on-md, split-on-lg`, etc.... `panel-left-on-md, panel-right-on-lg`, etc... `gutter-on-md, no-gutter-on-lg`, etc... would that change how this is built, or should we deal with that when we get there? `panel-right` doesn't seem like a variant with `stack` and `split`, either. I would expect to be able to do `split, stack-on-md, split-on-lg` and also `panel-left-on-lg`

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
mcoker: Yep. I'm just asking what the default should be, and what the prop opts into now. Currently in this PR if you create jump links that are expandable on small screens, but regular vertical jump links on large screens, the jump links label you define will only show up in the expandable UI as the toggle text. It will not be visible on large screens.

I was suggesting that we leave the label as it is now, and if it's defined, it shows up as both the expandable toggle on small screens and the jump links label on large screens, and we create a prop to hide the label on large screens so that the label only serves as the expandable toggle text.

packages/react-core/src/components/Sidebar/Sidebar.tsx
redallen: Split into orientation and `isPanelRight` props.

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
redallen: I've set `alwaysShowLabel` to true by default.

In the future we could allow a different label for responsive or non-responsive views.

jessiehuff (approved):

kmcfaul (approved):

dlabrecq (approved):

mcoker (approved):

mcoker:
@mcarrano in the demo - https://patternfly-react-pr-5314.surge.sh/components/jump-links/react-demos/scrollspy-with-subsections/

if you switch to horizontal, shrink the screen and expand the jump links, the list is horizontal. Is that what you would expect? If the list is just going to display the way it was, I'm not sure what the point is in collapsing/expanding it. Should the jump links be vertical on a small screen like that if they're expandable? Is there value in having horizontal jump links be expandable, or should expandable apply to vertical jump links?

tlabaj (approved):

PR #5319, opened 2021-01-08
feat(Drawer): add resize callback

### LINKED ISSUE #5282: Resizable drawer: unable to return user to prior size
no way to “control” the resize so that we can capture and return the user back to their previous size

kmcfaul:
**What**: Closes #5282

Adds `onResize` callback which will return the new size of the drawer after resizing as well as the ID of the panel (if it was given). Adds prop description for `id`.

Combined with #5313 a user may store a predetermined width and pass it with `defaultSize` to set the width of the panel when first opened. Using the `onResize` callback, this width may be updated and persisted if the page is refreshed. Or the width may remain the same to reset the panel width.

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -7,12 +7,16 @@ import { formatBreakpointMods } from '../../helpers/util';
 export interface DrawerPanelContentProps extends React.HTMLProps<HTMLDivElement> {
   /** Additional classes added to the drawer. */
   className?: string;
+  /** ID of the drawer panel */
+  id?: string;
   /** Content to be rendered in the drawer panel. */
   children?: React.ReactNode;
   /** Flag indicating that the drawer panel should not have a border. */
   hasNoBorder?: boolean;
   /** Flag indicating that the drawer panel should be resizable. */
   isResizable?: boolean;
+  /** Callback for resize end. */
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
```diff
@@ -154,12 +164,14 @@ export const DrawerPanelContent: React.FunctionComponent<DrawerPanelContentProps
...
           panel.current.style.setProperty('--pf-c-drawer__panel--FlexBasis', (newSize / max) * 100 + '%');
         }
+        currWidth = newSize;
       }
     }
   };
   const hidden = isStatic ? false : !isExpanded;
   return (
     <div
+      id={id}
```
tlabaj, line 174: Is the id required for the callback to work?  Should we either generate an id when `onResize` is provided. or throw a console warning?

tlabaj (changes requested):

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
kmcfaul: ID isn't required for the callback to work, if it isn't provided the function will return undefined for that parameter. I added it only as a way for a user to keep track of resizing if there are multiple resizable drawers in a single page.

tlabaj (approved):

dlabrecq (approved):

Issue #5024, opened 2020-10-14
Add multi-level drill down menu


mcarrano:
Introduce a multi-level drill down menu per these designs: https://marvelapp.com/prototype/66b0059/screen/72079939

Core issue: https://github.com/patternfly/patternfly/issues/2635

@jessiehuff Thoughts about keyboard interaction for this?

Requesting projects: Multiple (cloud platform, app services)
Date needed: March
Contacts:  @ajacobs21e @nlcwong @aprilma419

jessiehuff:
The only keyboard accessibility concern I have with a multi level drill down menu is making sure that the submenus are not immediately opened when using the tab key to navigate through the menu. I would expect that the user would tab through menu options, select the option with the submenu with space or enter (using either the parent or the button as the toggle) and then be able to tab through those options. Otherwise if the user is trying to get to the menu option after one with a submenu, they could be forced to go through the entire submenu first. This [article](https://www.w3.org/WAI/tutorials/menus/flyout/) explains the accessibility functionality/options pretty well.

PR #5321, opened 2021-01-11
feat(Menu): Drilldowns

jschuler:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/5024

https://patternfly-react-pr-5321.surge.sh/components/menu#with-drilldown


packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -26,6 +26,10 @@ export interface MenuProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'r
   'aria-label'?: string;
   /** Indicates if menu contains a flyout menu */
   containsFlyout?: boolean;
+  /** Indicates if menu contains a drilldown menu */
+  containsDrilldown?: boolean;
+  /** Indicates if a menu is drilled into */
+  drilledIn?: boolean;
```
tlabaj, line 32: perhaps rename to `isMenuDrilledin`

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -40,6 +42,12 @@ export interface MenuItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'onC
   flyoutMenu?: React.ReactNode;
   /** Callback function when mouse leaves trigger */
   onShowFlyout?: (event?: any) => void;
+  /** Drilldown menu */
+  drilldownMenu?: MenuProps;
+  /** Sub menu direction */
+  direction?: 'down' | 'up';
+  /** True if item is on current selection path */
+  onPath?: boolean;
```
tlabaj, line 50: `isOnPath`?

tlabaj (commented):
can you add integration test for the new props.

mcarrano (commented):
This is looking good @jschuler . Just two comments/questions:

* I love the animation when you open a new level. I see this when I go from level 1 to 2, but not when I drill down to the third level. Was it intended that these behave differently?

* I'm wondering what the alignment of the description text should be in the breadcrumb, or if it should display there at all. I'm referring to this-

Screen Shot 2021-01-20 at 2 32 59 PM

@doruskova Please also take a look at this. Interested in your thoughts about the description text question above.

doruskova:
@mcarrano I think it should be aligned under the name of the category. I can imagine that the description can be helpful if your data can be described in 1 sentence. 
Snímek obrazovky 2021-01-21 v 17 27 19

mcarrano:
I agree with that @doruskova .

@jschuler is it possible to make that change?

kmcfaul:
@mcoker I think the update to get the description in-line with the group title might be a core change? Thoughts?

I've got the PR pulled and will make any updates

mcarrano:
@kmcfaul @mcoker I'm not even sure it's necessary to have the description line in the title like this. If it's not in Core, then I'm guessing it wasn't part of the initial design. It must be possible for menu items to have descriptions, however. @doruskova what are your thoughts? Do you think this adds value or clutter?

mcoker:
@mcarrano @kmcfaul it would require a core change if we decide to keep the description in the drill up item.

mcarrano:
@kmcfaul @mcoker In that case I'd just rather not carry the description into the "drill-up" label. Can we do that?

mcoker:
@kmcfaul can we just not show the description in a drill up item in the react component?

kmcfaul:
Yes I updated to remove the menu header's descriptions in my PR addition (which is referenced above on my fork).

Also, do you mind taking a look at the drilldown's transition animation @mcoker? It only appears to be working for the first drilldown but not for subsequent inner menus. I'm not sure if I'm missing a styling, or if we need to re-apply a class somewhere (because once `pf-m-drilled-in` is applied it remains there for the following drilled in menus without change).

mcarrano:
Thanks for removing the description from the header @kmcfaul . I had the same question about the animation when I looked at this first a month ago. I think either it should be made to work for all levels or we should remove it.

tlabaj:
closing this PR.  The issue will be resolved in PR #5485

PR #5330, opened 2021-01-14
fix(Drawer): stop panel content unmounting before collapse

### LINKED ISSUE #5290: Resizable drawer: content unmounts early on collapse
When collapsing, content should be unmounted only after the animation completes. Ideally the content width should remain constant so that during the collapse animation the content isn’t being squished but rather simply sliding out of view.

kmcfaul:
**What**: Closes #5290

Removes the extra logic to not render the content when a panel is hidden. The panel is already hidden via the `hidden` attribute so the content should not have to also be un-rendered.


tlabaj (approved):

jessiehuff (approved):
LGTM!

evwilkin (approved):
LGTM

dlabrecq (approved):

Issue #5287, opened 2020-12-18
Resizable drawer: no touch support


evwilkin:
Does not work with touch device (however only tested using chrome dev tools on desktop)

DaoDaoNoCode:
@nicolethoen When I am working on my issue #5279 , I realized maybe this issue could also be fixed. Should we use `PointerEvent` instead of `MouseEvent` to support touch?

nicolethoen:
very possible! I'll look into it!

PR #5331, opened 2021-01-15
fix(drawer): fix dragging problem on iframe when resizing drawer

### LINKED ISSUE #5279: Resizable drawer: resize stops working when mouse over iframe

DaoDaoNoCode:
**What**: Closes #5279 
Add a new class set all the iframe in that class to `pointer-events: none`. When dragging happens, add that class to the drawer and remove that class when dragging ends.

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/Drawer/drawer';
+import '@patternfly/react-styles/css/components/Drawer/DrawerIframe';
 import { css } from '@patternfly/react-styles';
 import { DrawerContext } from './Drawer';
 import { formatBreakpointMods } from '../../helpers/util';
@@ -55,6 +56,7 @@ export const DrawerPanelContent: React.FunctionComponent<DrawerPanelContentProps
     e.preventDefault();
     document.addEventListener('mousemove', callbackMouseMove);
     document.addEventListener('mouseup', callbackMouseUp);
+    panel.current.parentElement.classList.add('pf-m-drawer-resizing');
     isResizing = true;
   };
 
@@ -98,6 +100,7 @@ export const DrawerPanelContent: React.FunctionComponent<DrawerPanelContentProps
     if (!isResizing) {
       return;
     }
+    panel.current.parentElement.classList.remove('pf-m-drawer-resizing');
     isResizing = false;
     document.removeEventListener('mousemove', callbackMouseMove);
     document.removeEventListener('mouseup', callbackMouseUp);
--- a/packages/react-docs/patternfly-docs.css.js
+++ b/packages/react-docs/patternfly-docs.css.js
@@ -5,6 +5,7 @@ import '@patternfly/react-styles/src/css/components/Topology/topology-side-bar.c
 import '@patternfly/react-styles/src/css/components/Topology/topology-view.css';
 import '@patternfly/react-styles/src/css/layouts/Toolbar/toolbar.css';
 import '@patternfly/react-styles/src/css/components/Popper/Popper.css';
+import '@patternfly/react-styles/src/css/components/Drawer/DrawerIframe.css';
 
 import '@patternfly/react-styles/src/css/components/Consoles/AccessConsoles.css';
 import '@patternfly/react-styles/src/css/components/Consoles/DesktopViewer.css';
--- a/packages/react-styles/src/css/components/Drawer/DrawerIframe.css
+++ b/packages/react-styles/src/css/components/Drawer/DrawerIframe.css
@@ -1,0 +1,3 @@
+.pf-m-drawer-resizing iframe {
+  pointer-events: none !important;
+}
--- a/packages/react-styles/src/css/components/Drawer/DrawerIframe.css
+++ b/packages/react-styles/src/css/components/Drawer/DrawerIframe.css
@@ -1,3 +1,7 @@
-.pf-m-drawer-resizing iframe {
+.pf-m-drawer-resizing {
   pointer-events: none !important;
+}
+
+.pf-m-drawer-resizing .pf-c-drawer__splitter {
+  pointer-events: auto !important;
 }


mcoker:
@DaoDaoNoCode I see that resizing doesn't work in the [codesandbox example](https://codesandbox.io/s/kwshg?file=/index.js) but I don't understand why. Can you explain what the problem is?

DaoDaoNoCode:
@mcoker I believe it is because we cannot use code sandbox to test this because it still uses the released version of PatternFly React components so the changes I made will not be reflected in the sandbox.

There is another way to test.

Copy this

```
import React from 'react';
import {
  Drawer,
  DrawerPanelContent,
  DrawerContent,
  DrawerContentBody,
  DrawerPanelBody,
  DrawerHead,
  DrawerActions,
  DrawerCloseButton,
  Button,
  Title

class ResizableDrawer extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      isExpanded: false,
      isResizing: false
    };
    this.drawerRef = React.createRef();

    this.onExpand = () => {
      this.drawerRef.current && this.drawerRef.current.focus();
    };

    this.onClick = () => {
      const isExpanded = !this.state.isExpanded;
      this.setState({
        isExpanded
      });
    };

    this.onCloseClick = () => {
      this.setState({
        isExpanded: false
      });
    };
  }

  render() {
    const { isExpanded } = this.state;
    const panelContent = (
      <DrawerPanelContent isResizable>
        <DrawerHead>
          <span tabIndex={isExpanded ? 0 : -1} ref={this.drawerRef}>
            drawer-panel
          </span>
          <DrawerActions>
            <DrawerCloseButton onClick={this.onCloseClick} />
          </DrawerActions>
        </DrawerHead>
        <iframe src="https://www.redhat.com" style={{height: '100%', width: '100%'}} />
      </DrawerPanelContent>
    );

    const drawerContent =
      'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus pretium est a porttitor vehicula. Quisque vel commodo urna. Morbi mattis rutrum ante, id vehicula ex accumsan ut. Morbi viverra, eros vel porttitor facilisis, eros purus aliquet erat,nec lobortis felis elit pulvinar sem. Vivamus vulputate, risus eget commodo eleifend, eros nibh porta quam, vitae lacinia leo libero at magna. Maecenas aliquam sagittis orci, et posuere nisi ultrices sit amet. Aliquam ex odio, malesuada sed posuere quis, pellentesque at mauris. Phasellus venenatis massa ex, eget pulvinar libero auctor pretium. Aliquam erat volutpat. Duis euismod justo in quam ullamcorper, in commodo massa vulputate.';

    return (
      <React.Fragment>
        <Button aria-expanded={isExpanded} onClick={this.onClick}>
          Toggle Drawer
        </Button>
        <Drawer isExpanded={isExpanded} onExpand={this.onExpand}>
          <DrawerContent panelContent={panelContent}>
            <DrawerContentBody><iframe src="https://www.ibm.com" style={{height: '100%', width: '100%'}} /></DrawerContentBody>
          </DrawerContent>
        </Drawer>
      </React.Fragment>
    );
  }
}
```

into the inline editor of this [example](https://patternfly-react-pr-5331.surge.sh/components/drawer#resizable-on-right). You will see it works like this.

Screen Shot 2021-01-18 at 11 41 32 AM

Here, the drawer components has been modified and you will see it's ok to drag resizable drawer in `iframe`.

mcoker:
@DaoDaoNoCode thanks! I'm just asking what the original problem is and how `pointer-events: none` on the `iframe` fixes it.

DaoDaoNoCode:
@mcoker Oh sorry I misunderstood. In my opinion, the problem is because the `iframe` in `DrawerContentBody` will swallow the mouse event when we drag over it. For example, if we want to drag the drawer to a bigger size, our mouse will definitely move over the IBM `iframe`. Thus, the target of the mouse event will transit from the splitter to the IBM page. so we cannot drag it anymore.

The way I fix it is to add this `pointer-events: none` to all `iframe` in the drawer when the drag begins and remove it when drag ends. Thus, when we start drag, the iframe will no longer be set to target and it will not swallow the mouse event anymore.

mcoker:
Hmm... I guess I'm still not sure what's special about the iframe that would cause a problem. I'm also wondering if some other kind of element could cause this same problem, and if we will need to add more styles to prevent mouse events from interfering in this way.

Also could we just apply `pointer-events: none;` to both the drawer content and panel if the drawer is being resized, regardless what's inside it?

DaoDaoNoCode:
@mcoker Actually I am not sure about this, either. I haven't met other elements which could swallow mouse events, except `iframe`.
Here I got another idea to fix this issue, do you know `setPointerCapture` function of an element? Maybe I could try that to set the Pointer always on the splitter. We need to wait for #5287 , it's very possible that in that issue we should replace all `MouseEvent` to `PointerEvent`. Then we could try to use `setPointerCapture` function instead of adding new classes to `iframe` or other elements.

mcoker:
@DaoDaoNoCode I guess I'm not sure why the iframe is capturing the pointer events if the splitter is a separate element. When you hover the splitter, you're not hovering the `iframe` as far as I can tell.  And I'm not familiar with `setPointerCapture`.

Another thing is that `pf-c-drawer__body` is required in the drawer, meaning the panel or content element's contents need to be wrapped in `pf-c-drawer__body`. In the codesandbox link, the `iframe` is a direct child of `pf-c-drawer__panel`. While that works, it's not how the drawer was designed to be used. In that example, we would just need to wrap the `iframe ` in `pf-c-drawer__body` since it's valid to have multple `__body` elements.

kmcfaul:
@mcoker @DaoDaoNoCode My understanding is `iframe` is effectively a separate window/document, so events won't transfer between them and the rest of the page. 

Does this fix work with iframes outside of the drawer panel?

DaoDaoNoCode:
@kmcfaul Yes, it will simply disable all the pointer events in drawer (except splitter) when dragging. Do you mean will this fix work with iframes outside the whole drawer? My answer maybe no. It will need us to disable the pointer event of the whole page. So I am considering using `setPointerCapture` to implement this if we could change all the mouse events to pointer events.

DaoDaoNoCode:
Thanks @nicolethoen 's try on `PointerEvents`, it's not good to use it here. Currently fixing by adding classname is probable the best way.

kmcfaul (approved):
It's probably more straightforward to handle this on the app-side of the code, on a per iframe basis, since we do not know if the drawer contains an iframe or if the mouse is being dragged over an iframe. But I'm fine with handling iframes within a drawer at least. 

Code looks good to me

nicolethoen (approved):

PR #5357, opened 2021-01-21
fix(Drawer): enable touch controls for resizable drawer

### LINKED ISSUE #5287: Resizable drawer: no touch support

nicolethoen:
Closes #5287


DaoDaoNoCode:
I am wondering could we use `PointerEvents` API here? As is mentioned in this article https://mobiforge.com/design-development/html5-pointer-events-api-combining-touch-mouse-and-pen.

> The Pointer Events API specifies a way to handle mouse, touch and pen inputs at the same time, without having to code for separate sets of events, mouse events, and touch events.

So we can remove both `MouseEvents` and `TouchEvents`. Instead of using 4 event listeners here, we could only use 2 event listeners and also avoid writing duplicated code.

nicolethoen:
> https://mobiforge.com/design-development/html5-pointer-events-api-combining-touch-mouse-and-pen

right! You mentioned that in the issue! Idk why I read it as mouseEvent. I guess I got tunnel visioned and didn't reference your comment again. one moment and I'll test/update with pointerevents

nicolethoen:
@DaoDaoNoCode 
There is an issue with the pointerup event not being fired after the mouse has dragged.

evwilkin:
@nicolethoen I believe we'll need to stop the event bubbling, as the touch interactions are still scrolling the page up/down (with top/bottom drawer) and left/right (triggering my browser to go back to the previous page) when dragging the resizable drawer.

kmcfaul:
There is a `e.stopPropagation` call, but we might need a `e.preventDefault` to stop the scroll/navigation. Also has a conflict but other than those two things the code looks good.

nicolethoen:
@kmcfaul There is no `e.preventDefault` on touchEvents, that's why i haven't solved this yet. I need to chase down some alternative solutions.

kmcfaul:
Tested on my iPad and touch doesn't seem to work at all for me. Can someone confirm? Maybe @evwilkin 

I'm also seeing empty space for left/bottom/inline drawers where the panel would be if its pulled up (but right drawer is fine). As in, the text is pushed to the side and not taking up the full space like it should. I'm not sure if that's related to this PR or if its something going on with mobile/iPad @mcoker

nicolethoen:
hm... wild.
@kmcfaul it's working my phone - I dont have a tablet. And it's working on the various devices when I toggle between them using the chrome browser tools. I don't think I have any other means to test this.

kmcfaul:
Brought my personal laptop back to life and tested it out again and it works. I think my iPad didn't work because its old and probably a lot of its apps are out of date (I don't use it much). 

Had a bit of lag, but I think that was on my end, not PF.

kmcfaul (approved):

evwilkin (approved):
LGTM - tested on tablet and resizing drawer no longer causes page to scroll as well!

tlabaj (approved):

PR #5358, opened 2021-01-22
feat(Slider): Added new slider component

### LINKED ISSUE #5187: Add Slider component
Introduce a new Slider component.

Core issue and designs: https://github.com/patternfly/patternfly/issues/296 

There are two flavors of the slider: Continuous and Discrete. 

* For the continuous slider, the value returned should simply be based on the position between the start and end values. This can be fractional.

* For the discrete slider, when the user releases the mouse (or presses an arrow key to move the thumb) the thumb will always snap to the nearest tick mark.

tlabaj:
**What**: Closes #5187

TODO: integration tests


KKoukiou (commented):
Thanks for working on this, we look forward to this feature in cockpit. 
I quickly tried it out, I have some pointers for improvement.

* Keyboard interactions don't work
* Clicking on random place in the slider should move the current position there

Tried it on firefox-83.0-13.fc33.x86_64.

mcarrano (changes requested):
This is looking good @tlabaj . Just had a couple of issues/questions.

* For the continuous slider I notice that it never returns fractional values. Should we allow the consumer to set precision via a prop? There might be cases where you don't want the returned value always rounded to a whole number.
* For the "Thumb value input" example, I was able to get the thumb and value input to disappear from view by entering a value beyond the upper limit of the slider. Seems like there should be some value checking to make sure the slider value can never exceed the upper or lower limits.
* The keyboard interaction worked for me except on the last example of the slider with actions. @KKoukiou how were you expecting the interaction to work. We specified that the slider value should only change upon pressing Enter or tabbing out of the field. Not on every keystroke.
* Agree with @KKoukiou that you should also be able to update the value by clicking on the slider bar.

KKoukiou:
> This is looking good @tlabaj . Just had a couple of issues/questions.
> 
>     * For the continuous slider I notice that it never returns fractional values. Should we allow the consumer to set precision via a prop? There might be cases where you don't want the returned value always rounded to a whole number.
> 
>     * For the "Thumb value input" example, I was able to get the thumb and value input to disappear from view by entering a value beyond the upper limit of the slider. Seems like there should be some value checking to make sure the slider value can never exceed the upper or lower limits.
> 
>     * The keyboard interaction worked for me except on the last example of the slider with actions. @KKoukiou how were you expecting the interaction to work. We specified that the slider value should only change upon pressing Enter or tabbing out of the field. Not on every keystroke.

I expect that clicking on the circle of the slider and then using my right and left keyboard keys, should move the slider value left and right accordingly. This does not happen for me at least. 
I just tried on all browsers i have available, firefox, chrome, edge and epiphany, its' the same everywhere.
Let me know if you still can't reproduce it.

> 
>     * Agree with @KKoukiou that you should also be able to update the value by clicking on the slider bar.

tlabaj:
Thanks @KKoukiou and @mcarrano You are right, you have to tab into the thumb to be able to move it.  I will work on that update and clicking on the slider to update the value.

tlabaj:
@mcarrano 

> This is looking good @tlabaj . Just had a couple of issues/questions.
> 
> * For the continuous slider I notice that it never returns fractional values. Should we allow the consumer to set precision via a prop? There might be cases where you don't want the returned value always rounded to a whole number.

It does return the fractional value.  The demo code takes the floor and displays that.  I can update that if you like.

> * The keyboard interaction worked for me except on the last example of the slider with actions. @KKoukiou how were you expecting the interaction to work. We specified that the slider value should only change upon pressing Enter or tabbing out of the field. Not on every keystroke.

For this one, are you talking about both the sliders in the action example or the very last one.  I could not reproduce an issue with the keyboard interaction other than what was mentioned about clicking on the slider and clicking the thumb to apply focus.

jessiehuff (commented):
Really nice job with this so far, Titani!!! 🎉 ⭐ From my perspective, I think the keyboard interaction looks pretty good so far, and I think handling the aria position with the `aria-valuetext` was a decent way to do it. 

The only thing I've noticed that I thought was a little odd is that I can't seem to get the slider to 0 for a lot of the examples by keyboard, though it looks like this is true by mouse as well for some. Is it intentional to enforce a minimum value like that? Wondering if some of them may be a tad odd though. For example, if you're looking at the discrete example and the slider thumb starts at 2 and you drag it to 0, the slider value above it stays at 2 (I noticed this is true if it starts at any number and you drag it to 0, it will always stay at the previous number).

mcarrano:
> For this one, are you talking about both the sliders in the action example or the very last one. I could not reproduce an issue with the keyboard interaction other than what was mentioned about clicking on the slider and clicking the thumb to apply focus.

Yes, I'm talking about the very last one. If I change the value in the number field and hit Enter or tab out of the field, nothing happens. I'm having the same problem with the second example under Value input.

jessiehuff (dismissed):
LGTM! Looks like you fixed the thing I was seeing before. 🙂

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,302 @@
...
+        type="number"
+        value={localInputValue}
+        aria-label={inputAriaLabel}
+        onKeyDown={handleKeyPressOnInput}
+        onChange={onChangeHandler}
+        onClick={onInputFocus}
+        onFocus={onInputFocus}
+        onBlur={onBlur}
+      />
+      {inputLabel && <span className={css(styles.inputGroupText, 'pf-m-plain')}>{inputLabel}</span>}
```
mcoker, line 253: Technically an `<InputGroup>` is only needed here if this text exists, but it doesn't hurt to leave it the way it is.

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,302 @@
...
+      thumbRef.current.style.setProperty('--pf-c-slider--value', `${newValue}%`);
+      setValue(newValue);
+      if (onValueChange) {
+        onValueChange(newValue);
+      }
+    }
+  };
+
+  const inputGroup = (
+    <div className={css(styles.inputGroup)}>
```
mcoker, line 240: I'm curious why you don't use `<InputGroup>` here?

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,302 @@
...
+      <input
```
mcoker, line 241: Curious why you don't use `<TextInput>`?

packages/react-core/src/components/Slider/Slider.tsx
mcoker: Also curious why you don't use `<InputGroupText>` for this element? Looks like we would need an enhancement for `isPlain` or something to apply `.pf-m-plain`.

I can open a separate issue for that enhancement if we don't want to do it here. It was added as part of https://github.com/patternfly/patternfly/pull/3711/

mcoker (dismissed):
LGTM! Left some questions, but they shouldn't hold up merging this PR.

mcarrano (approved):
Thanks for making those updates @tlabaj . Looks good to me!

packages/react-core/src/components/Slider/Slider.tsx
tlabaj: I have updated this to use iInputGroup, `InputGroupText` and `TextInput`.  It was an oversight.  

The `InputGroupText` component  still need to be updated for the plain variation.  @mcarrano do you want me to open a new issue to make that update?  The style is still applied inline here so it still displays correctly.

mcarrano: Sure. Go ahead and open a core issue @tlabaj .

tlabaj: @mcarrano it is a react issue.  Core already made the update.
here is the new issue
https://github.com/patternfly/patternfly-react/issues/5367

kmcfaul:
Looks good, just two interaction questions.

In the `Thumb input value` example, you can move the slider but the text input doesn't update with the current slider value. The other slider examples do update so I'm not sure if this is intentional.

In the `Actions` example, you can "lock" the slider but only the text input is disabled, you can still drag the slider and change the value that way. Is the lock meant only for the text input?

tlabaj:
@kmcfaul 
> Looks good, just two interaction questions.
> 
> In the `Thumb input value` example, you can move the slider but the text input doesn't update with the current slider value. The other slider examples do update so I'm not sure if this is intentional.

The input is working fine for me

> 
> In the `Actions` example, you can "lock" the slider but only the text input is disabled, you can still drag the slider and change the value that way. Is the lock meant only for the text input?
@mcarrano @mcoker there is no styling to disable a slider correct?

mcarrano:
> In the Actions example, you can "lock" the slider but only the text input is disabled, you can still drag the slider and change the value that way. Is the lock meant only for the text input?
@mcarrano @mcoker there is no styling to disable a slider correct?

Right. There is no disabled styling for the Slider. We could add that but I'm hesitant as there is not a product use case for it. Can you just disable events on the slider for now @tlabaj to make this example work better?

tlabaj:
> > In the Actions example, you can "lock" the slider but only the text input is disabled, you can still drag the slider and change the value that way. Is the lock meant only for the text input?
> > @mcarrano @mcoker there is no styling to disable a slider correct?
> 
> Right. There is no disabled styling for the Slider. We could add that but I'm hesitant as there is not a product use case for it. Can you just disable events on the slider for now @tlabaj to make this example work better?

@mcarrano I could.  However, I don't think that is a good idea.  I think that could be  confusing to the consumer since there is no visual indication that the slider is disabled.

mcarrano:
@tlabaj @kmcfaul Do you think it's better to leave as is or to remove the example? Again, I see this as a "nice to have" feature that we don't have a hard requirement for right now.

kmcfaul (approved):
I think I'd be fine with it either way. Since this will be a beta component we can update the example later if we leave it in.

jessiehuff (commented):
So I only see this on the actions example so it might just be that example's issue, but I noticed that when I try to navigate the actions example, it doesn't quite behave right (tring to use the right arrow key on the thumb doesn't work, only the left) and when I try to use the increase/plus button, it adds on a bizarre increment: 
Screen Shot 2021-01-25 at 1 21 09 PM

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,311 @@
...
+}
+
+export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange'> {
+  /** Additional classes added to the spinner. */
+  className?: string;
+  /** Current value  */
+  currentValue?: number;
+  /** Flag indicating if the slider is is discrete */
+  isDiscrete?: boolean;
+  /** Array of slider step objects (value and label of each step) for the slider. If this is provided, the numSteps prop will be ignored.*/
```
nicolethoen, line 25: is there still a `numSteps` prop? I dont see it.

nicolethoen:
we should also open an issue to make the sliders work on mobile.

jessiehuff (approved):
LGTM on the accessibility side! 🙂

tlabaj:
@nicolethoen there is already an issue open for that.

packages/react-core/src/components/Slider/Slider.tsx
tlabaj: nope. that was removed

nicolethoen (approved):

kmcfaul (approved):

PR #5359, opened 2021-01-22
feat(Modal): support disabling the 'Close' button on the modal header

### LINKED ISSUE #5210: Modal: should we support disabling the 'Close' button on the modal header when cancel is disabled?
This is more of a discussion topic/design guideline question:
If you agree it's a nice feature, I am happy to work on it in the react side at least - otherwise just close it.

So in Cockpit we have many modal dialogs, which perform operations that can't be cancelled when the primary button is clicked. For example deleting a VM's disk, formatting a storage device and so on. In this case, we keep the modal open, and disable all ways to close it, while the operation is in progress. (of course these operations don't take very long time, it's a matter of some seconds)
That means: 

* We disable close button
* In these kind of modals we disable Esc close and also hide the [x] button in the header

That results in a inconsistency in our project, where some modals have the [x] button and some not.
The question here is, should be maybe show the [x] button similarly to the 'Cancel' button and just disable it in this case?

If you 're positive about this, maybe we should transfer this issue to the react repository, and add support for this through a property.

KKoukiou:
When onCloseDisabled is set the modal can't be closed using Escape key
either.

Fixes #5210
Closes #5359

### DIFF
--- a/packages/react-core/src/components/Modal/Modal.tsx
+++ b/packages/react-core/src/components/Modal/Modal.tsx
@@ -31,6 +31,9 @@ export interface ModalProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
   'aria-label'?: string;
   /** Id to use for Modal Box descriptor */
   'aria-describedby'?: string;
+  /** Flag to disable the close button in the header area of the modal.
+   * When set pressing of the Escape key will not close the modal either */
+  isCloseDisabled?: boolean;
   /** Flag to show the close button in the header area of the modal */
   showClose?: boolean;
   /** Custom footer */
@@ -87,6 +90,7 @@ export class Modal extends React.Component<ModalProps, ModalState> {
     titleIconVariant: null,
     titleLabel: '',
     'aria-label': '',
+    isCloseDisabled: false,
     showClose: true,
     'aria-describedby': '',
     'aria-labelledby': '',
@@ -116,7 +120,7 @@ export class Modal extends React.Component<ModalProps, ModalState> {
 
   handleEscKeyClick = (event: KeyboardEvent): void => {
     const { onEscapePress } = this.props;
-    if (event.keyCode === KEY_CODES.ESCAPE_KEY && this.props.isOpen) {
+    if (event.keyCode === KEY_CODES.ESCAPE_KEY && this.props.isOpen && !this.props.isCloseDisabled) {
       onEscapePress ? onEscapePress(event) : this.props.onClose();
     }
   };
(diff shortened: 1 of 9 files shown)


KKoukiou:
Rebased to solve conflict.

mcarrano (approved):
Looks good to me. Thanks @KKoukiou .

tlabaj (changes requested):
Look good.  Can you just add the new prop to the demo-app and add an integration test for this.

jessiehuff (commented):
I find the example here slightly confusing. From an accessibility standpoint, if you trap focus there has to be a clear way to get out and to have that communicated to users. Here, the user can still get out of the modal by confirming or cancelling so there's a way out, but it seems confusing to me to see the disabled close button without understanding why it's disabled. In this example, why wouldn't we just not show the close button? Should there be some way to communicate that there's a process happening or that input is required or some design guideline that helps users to know when this will be used and how to include a11y in that? I also noticed that there's not a core example of this. Should there be an issue to create that example in core as well?

KKoukiou:
> I find the example here slightly confusing. From an accessibility standpoint, if you trap focus there has to be a clear way to get out and to have that communicated to users. Here, the user can still get out of the modal by confirming or cancelling so there's a way out, but it seems confusing to me to see the disabled close button without understanding why it's disabled. In this example, why wouldn't we just not show the close button? Should there be some way to communicate that there's a process happening or that input is required or some design guideline that helps users to know when this will be used and how to include a11y in that? I also noticed that there's not a core example of this. Should there be an issue to create that example in core as well?

Good point, I was not aware about the accessibility issues this might cause.

The use case I have is that many modals (in Cockpit) when submitted perform operations that can't be stopped. For these modals, 
* I disable the `cancel` button,
*  show a spinner in the submit button
* disable all input fields

Of course Esc and [x] have to also be disabled or hidden though for the modal to no unexpectedly close.
 
For consistency with the rest of the modals in our project, which don't have non-interruptible operations, and therefore can be stopped by closing the modal, I thought we should always have the close button available, just sometimes disabled when the 'Cancel' button is also disabled.

Another approach would be to always allow closing the modal by 'cancel' or [x] or Esc but have a warning that the operation started will continue on the background.

Either of the two approaches would work for me. So if some designer here feels that they want to advice me to follow one of the two approaches, I am happy to proceed accordingly. 

I just find important that all modals within a project don't deviate visually from one another.

mcarrano:
@KKoukiou @jessiehuff I at first found this example confusing also since there no context. Should Cancel as well as Close be disabled? Maybe Confirm should change to Exit? I don't think the close button should be hidden since it might be available prior to the action.

I agree that there a design question here that goes beyond the scope of this example. By holding the user in this dialog with no way to close it, there is no way out if the operation hangs or takes too long, for example. I guess you could allow the dialog to be closed and then use a toast alert to signal the user when the process is complete. But this is really a product design question that I don't think we can enforce via PatternFly.

KKoukiou:
@mcarrano @jessiehuff thanks for taking time to answer here. I am fine with handling it as @mcarrano suggested, in that case we can close this.

tlabaj:
closing this one as a result of the discussion.

PR #5375, opened 2021-01-26
fix(Select): adjust Select component so that it also works with children changes while it's mounted

### LINKED ISSUE #5374: Select Typeahead: when the children are updated while the component is mounted existing value of the typeahead select will not filter them
In Cockpit [1] we utilize this component in order to create a filesystem
browser field inside dialogs.

Initially the values of the SelectOptions are the subdirs of the root
('/') directory:

So:
```
<Select>
    <SelectOption value="bin" />
    <SelectOption value="boot" />
    etc.
</Select>
```
When the user selects one of the directories/SelectOptions we basically
navigate them inside the directory they selected, therefore we update all
the SelectOptions in order to reflect the selected directory content.

Of course it's expected that the current value of the typeahead select
input will still act as a filter for the new children but in the current code this is ignored.

Initially I though I can workaround it by adding a key property to the Select component, which changes everytime the chidren (options) change but that crashes the component.

KKoukiou:
Fixes #5374
Blocked on https://github.com/patternfly/patternfly-react/pull/5370


KKoukiou:
I think I have broken the create variant of the select typeahead with my changes, integration tests caught it. If someone knows how to properly address this issue feel free to take over.

kmcfaul (approved):

redallen (approved):

dlabrecq (commented):
LGTM

tlabaj:
we need to open up a follow up issue to add the new prop to the demo app.

tlabaj:
I opened issue #5403

Issue #5377, opened 2021-01-27
Bar chart remains muted with interactive legend


dlabrecq:
When hovering over an interactive legend item, all other datasets in the chart are muted (i.e., an opacity is applied to make them look slightly grayed out). When the cursor moves away from the legend item, the muted styles are removed. This works fine with line and area charts, but there is a problem with bar charts. It's more of an edge case, but under certain conditions, bars may remain muted.

To recreate the scenario, all datasets are first hidden by clicking each legend item. Then, a single data series is shown by clicking on a legend item associated with the bar chart. Typically, this will result in most bars remaining muted.

Note: Victory's `onMouseOver` events provide an index when associated with legend items, which can be used to hide / show data in the chart. However, the same events don't always provide a usable index for bars shown in the chart. The index works with line charts because it matches the ordered data series given for legend items. For a bar chart, this is not the case as the index represents each bar.

Basically, we need a way to prevent the bars from being muted if the data series is hidden.

![chrome-capture](https://user-images.githubusercontent.com/17481322/105925622-7f828180-600e-11eb-94e6-5236f2a6580f.gif)

PR #5378, opened 2021-01-27
Bar chart remains muted with interactive legend

dlabrecq:
When hovering over an interactive legend item, all other datasets in the chart are muted (i.e., an opacity is applied to make them look slightly grayed out). When the cursor moves away from the legend item, the muted styles are removed. This works fine with line and area charts, but there is a problem with bar charts. It's more of an edge case, but under certain conditions, bars may remain muted.

Fixes https://github.com/patternfly/patternfly-react/issues/5377

Cost Management
capture-good


ddonahue007 (approved):
LGTM

redallen (approved):

PR #5386, opened 2021-01-27
chore: bump patternfly version

nicolethoen:


redallen:
Duplicate of #5385 . Let's use that instead since it's further along in CI...

PR #5389, opened 2021-01-27
feat(card): add pf-m-display-lg and pf-m-rounded

### LINKED ISSUE #5334: Large card variant
Add a new variant of the card component per https://github.com/patternfly/patternfly/pull/3793 . New card has increased spacing for marketing content (Large card)

redallen:
**What**: Closes #5334

**Additional issues**: I made an interactive modifiers example to showcase it

### DIFF
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -14,7 +14,7 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   component?: keyof JSX.IntrinsicElements;
   /** Modifies the card to include hover styles on :hover */
   isHoverable?: boolean;
-  /** Modifies the card to include compact styling */
+  /** Modifies the card to include compact styling. Should not be used with isLarge. */
   isCompact?: boolean;
   /** Modifies the card to include selectable styling */
   isSelectable?: boolean;
@@ -24,7 +24,7 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   isFlat?: boolean;
   /** Modifies the card to include rounded styling */
   isRounded?: boolean;
-  /** Modifies the card to be large */
+  /** Modifies the card to be large. Should not be used with isCompact. */
   isLarge?: boolean;
   /** Flag indicating if a card is expanded. Modifies the card to be expandable. */
   isExpanded?: boolean;
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -59,6 +59,10 @@ export const Card: React.FunctionComponent<CardProps> = ({
 }: CardProps) => {
   const Component = component as any;
   const ouiaProps = useOUIAProps(Card.displayName, ouiaId, ouiaSafe);
+  if (isCompact && isLarge) {
+    console.warn('Card: Cannot use isCompact with isLarge. Defaulting to isCompact');
+    isLarge = false;
+  }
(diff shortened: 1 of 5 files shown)


KKoukiou (approved):
Nice parametrization of the examples code @redallen!

Your change it of course totally fine codewise, I just wonder why PF wants to expose to its consumers that isRounded property. It causes a really subtle difference, and I feel that the PF designers should decide if the cards should have rounded corners or not and not let the consumers make decisions about it.

After all, the reason that we all use PF and not custom styling is in order that all RH products have the same Ui experience.

So, this PR is fine to go in but I really don't see the usecase for it.

redallen:
@KKoukiou This thread has the context behind pf-m-rounded: https://github.com/patternfly/patternfly/issues/3670 . I'd ask @mcarrano its purpose.

mcarrano:
You ask a valid question @KKoukiou . This originated from the web properties design team wanting to use PatternFly to present more marketing oriented content. @mmenestr do you know if we have plans to add this guidance to our design guidelines for Cards? I agree that we don't want designers to just arbitrarily pick styling. Products and web applications should stick with the squared corners.

mmenestr:
@mcarrano I don't believe we do. This is the first I've heard of this! Just to make sure I have the info right, is it just a modifier to give cards the capability of having rounded edges? And we want to recommend that this only be done for marketing purposes and that products keep using non-rounded cards?

mcoker (commented):
`isLarge` and `isCompact` shouldn't be used together - do we need to do anything special for that?

redallen:
@mcoker Unfortunately since we've already released `isCompact` there isn't much we can do. I've added a warning and will default to `isCompact` if both are specified.

kmcfaul:
Can we also add something to both prop descriptions (isLarge and isCompact) for clarity? Like `Should not be used together with {X}`. Or maybe the console warning is enough?

redallen:
@kmcfaul I added that to the prop description.

mcoker (approved):
L👍TM!

jessiehuff (commented):
I really like the modifier example! It's really neat to be able to play with all of the different modifiers. Can we add the keyboard interaction to the selectable example though? I currently can't select it by keyboard or screen reader the way I could with the previous selectable example. (I'm hoping that by adding the keyboard interaction, it'll work through screen reader as well.)

redallen:
@jessiehuff I've removed the `isSelectable` and `isSelected` to avoid confusion and keep the demo simple.

mcarrano (dismissed):
Looks good to me. Thanks @redallen .

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -47,12 +51,19 @@ export const Card: React.FunctionComponent<CardProps> = ({
...
+  isRounded = false,
+  isLarge = false,
   ouiaId,
   ouiaSafe = true,
   ...props
 }: CardProps) => {
   const Component = component as any;
   const ouiaProps = useOUIAProps(Card.displayName, ouiaId, ouiaSafe);
+  if (isCompact && isLarge) {
+    // eslint-disable-next-line no-console
```
tlabaj, line 63: can you add a test for this

tlabaj (changes requested):
looks good.  Can you just update the demo app and add test for new props.

packages/react-core/src/components/Card/Card.tsx
redallen: Done

kmcfaul (approved):

PR #5399, opened 2021-01-29
fix(timepicker): fix issue when clicking dropdown menu does not trigger onchange function

### LINKED ISSUE #5398: onChange handler of Time Picker component not working as expected
onChange handler works only when typing into the inputbox, it's not called on selecting a time from the dropdown

https://codesandbox.io/s/tqg6s 
If typing into the input box, console logs are displayed, but if selecting a time value from the dropdown, the console log is not displayed, meaning the onChange handler is not invoked

This is a blocking bug

Red Hat Advanced Cluster Management - release 2.3

DaoDaoNoCode:
**What**: Closes #5398

**Reason**: The `onChange` function is only called in `onInputChange` function, which is only triggered when users input something. However, in `onSelect` function, we don't use `onChange` function.

To optimize the code structure, I called `onInputChange` function inside `onSelect` function because they did almost the same thing, and I removed those unused props and redundant setState functions. It works for both mouse events and keyboard events.

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimeOption.tsx
+++ b/packages/react-core/src/components/TimePicker/TimeOption.tsx
@@ -14,7 +14,7 @@ export interface TimeOptionProps extends Omit<React.HTMLProps<HTMLLIElement>, 'o
   /** Flag forcing the focused state */
   isFocused?: boolean;
   /** Optional callback for click event */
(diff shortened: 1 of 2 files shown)


nicolethoen (approved):
Awesome!

tlabaj (approved):
Thank you!  looks good.

PR #5406, opened 2021-02-02
bug(TableComposable): add dataLabel property

### LINKED ISSUE #5405: bug(TableComposable): missing dataLabel property for <Th> cells in case these are used as part of the row
According to the [spec](https://www.w3.org/TR/REC-html40/struct/tables.html#edef-TH) `<th>` can be used as part of `<tr>` elements. For that we need the dataLabel property exposed to `<th>` exactly how we do for `<td>`.

This property also needs to be mandatory when `<Th>` is direct child of `<Tr>` so that the mobile mode works as expected.

KKoukiou:
Closes #5405

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableComposableDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableComposableDemo.tsx
@@ -15,7 +15,16 @@ import {
   OnCollapse,
   OnSort
 } from '@patternfly/react-table';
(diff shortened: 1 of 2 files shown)


redallen (dismissed):

tlabaj:
Can you please add the new prop to the demo app.

KKoukiou:
I can do it - it's fine.

kmcfaul (approved):

evwilkin (approved):
LGTM

jessiehuff (approved):
LGTM!

dlabrecq (approved):

PR #5414, opened 2021-02-03
fix(TreeView): expand/collapse on filtering

### LINKED ISSUE #5351: Treeview: Using search with treeview should reveal leaf nodes matching the search criteria
When using treeview to search a string in the tree, the branch nodes are correctly filtered, however the leaf nodes aren't expanded when there is a match. The user would have to expand manually to see the matched string.

Bug
Openshift, unscheduled

cc: @evwilkin

kmcfaul:
**What**: Closes #5351

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/TreeView/tree-view';
 import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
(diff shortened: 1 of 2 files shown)


mcarrano (approved):
Looks good @kmcfaul !

@jcaianirh is this also what you were expecting?

jcaianirh (commented):
@kmcfaul this is great, and yes this @mcarrano this is exactly what we were looking for. One bug i'm experiencing is that when i search for current, and start typing c all leaf nodes are expanded even though they don't contain the letter c. This doesn't, however, happen when i start typing settings as only the settings nodes are expanded

nicolethoen (approved):
very cool!

kmcfaul:
Re: the search with `c` everything gets expanded because the filter is matching the string anywhere in the text, so all folders/leaves have `c` in them somewhere (like "Application 1" gets matched so the whole folder expands). Filtering behavior can be updated in the `filterItems` function of the example.

tlabaj (approved):

dlabrecq (commented):
LGTM

PR #5415, opened 2021-02-03
feat(Progress): Add warning when title and aria-label not present

### LINKED ISSUE #5396: bug(Progress): when 'title' property is not used in the Progress component there is an accessibility issue
The progress bar uses `aria-labelledby` pointing to some ID having the title. When the title is not used this is empty. This is an accessibility issue.

Current workaround is to use `aria-label` directly passed on the Progress bar, but of course the progress bar still has an `aria-labelledby` pointing to the empty div.

jessiehuff:
**What**: Closes #5396


tlabaj (changes requested):
Thanks for taking this one Jessie.  Looks good!  Can you just add a unit test to verify the warning gets generated as expected.

tlabaj (approved):
LGTM

evwilkin (approved):
LGTM

dlabrecq (commented):
LGTM

Issue #5418, opened 2021-02-04
Drawer: splitter is in tab order even when panel hidden


kmcfaul:
If you tab through the drawer, it will expand an empty panel because tab will take you to the splitter, even when the panel is closed. The splitter shouldn't be in the tab order when hidden.

This might be fixed if we go back to unmounting the panel content after it closes (opposed to just hiding the panel).

tlabaj:
closed by #5426

Issue #5412, opened 2021-02-02
Drawer: resizable css cleanup & transition property conflict


kmcfaul:
The current `transition` property being set on the panel is conflicting with other transitions. The property needs to be removed on `mouseup` and also can be set to `none` instead of the current curve (it seems that the other transitions were slowing the drag op rather than the op needing a transition). 

After talking with @mcoker, we should move the `transition` property to a class that will be added/removed instead of setting it on the style object. The local CSS stylesheet added the classes for the pointer-events will also be moved to core. So this issue should be updating the transition & local css references to the upcoming core classes.

tlabaj:
closed by #5426

PR #5426, opened 2021-02-08
feat(Drawer): update transitions and content mounting

### LINKED ISSUE #5419: Drawer: panel content should unmount on collapse, but after the transition animation
Panel content should still be unmounting instead of just hidden when the panel collapses, but it shouldn't unmount until after the transition animation ends.

kmcfaul:
**What**: Closes #5419 , #5418 , #5412

Removes the transition from the style object on the panel and adds in the new CSS class to handle the property. 
Updates the local CSS class for iframes to the new core side CSS class.
Panel content will now unmount after the transition to close ends, and re-mount when the panel opens. This solves the splitter being in the tab order because it no longer exists when closed.

### DIFF
--- a/packages/react-core/src/components/Drawer/Drawer.tsx
+++ b/packages/react-core/src/components/Drawer/Drawer.tsx
@@ -24,16 +24,18 @@ export interface DrawerContextProps {
   isStatic: boolean;
   onExpand?: () => void;
   position?: string;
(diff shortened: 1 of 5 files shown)


nicolethoen (approved):

tlabaj (approved):
lgtm

Issue #5095, opened 2020-11-03
Update Select to focus on disabled items


jessiehuff:
**Describe the issue. What is the expected and unexpected behavior?**
We should update Select to match the dropdown interaction so that disabled items are focusable and JAWS will announce them.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
 
Try to open the select in JAWS, it does not function well in JAWS. Also, key through select, you will see that disabled items are skipped.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug

nicolethoen:
Part of the epic: https://github.com/patternfly/patternfly-react/issues/6014

tlabaj:
This will be addressed in the next version of the select

PR #5427, opened 2021-02-08
fix(selectmenu): don't use focus trap when no tabbable children

### LINKED ISSUE #5430: Checkbox select without enabled child crashes
When a checkbox select has children but none are tabbable it crashes due to FocusTrap: https://codesandbox.io/s/eager-fast-jo7bb?file=/index.js

redallen:
**What**: Closes #5430

**Additional issues**: I'm not sure why we were using FocusTrap to begin with in this component. Please make sure I wasn't missing anything @kmcfaul !

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -743,6 +743,16 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   this.handleMenuKeys(0, 0, 'left');
                 } else if (event.key === KeyTypes.ArrowRight) {
                   this.handleMenuKeys(0, 0, 'right');
+                } else if (event.key === KeyTypes.Tab && variant === SelectVariant.checkbox) {
+                  // More modal-like experience for checkboxes
+                  // Let SelectOption handle this
+                  if (event.shiftKey) {
+                    this.handleMenuKeys(0, 0, 'up');
+                  } else {
+                    this.handleMenuKeys(0, 0, 'down');
+                  }
+                  event.stopPropagation();
+                  event.preventDefault();
                 }
               }}
               ref={this.filterRef}
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -780,7 +780,8 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
           variantProps = {
             checked: selections,
             isGrouped,
-            hasInlineFilter
+            hasInlineFilter,
+            openedOnEnter
           };
           variantChildren = filterWithChildren;
           break;
--- a/packages/react-core/src/components/Select/SelectMenu.tsx
+++ b/packages/react-core/src/components/Select/SelectMenu.tsx
@@ -6,7 +6,6 @@ import { SelectOptionObject, SelectOption } from './SelectOption';
 import { SelectConsumer, SelectVariant } from './selectConstants';
 import { PickOptional } from '../../helpers/typeUtils';
 
-import { FocusTrap } from '../../helpers';
 import { SelectGroup } from './SelectGroup';
 import { Divider } from '../Divider/Divider';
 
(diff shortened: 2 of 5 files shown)


kmcfaul:
We were using FocusTrap to allow tabbing through the checkbox children. But since we now have more keyboard interaction (which is working for the checkbox variant still), and FocusTrap was causing an issue with scrolling, we might be able to remove it? @jessiehuff 

The only issue is that this may be considered a breaking change, since it's changing the interaction pattern (Tab will now close the menu instead of tabbing through). I'm unsure if anyone was using Tab to go through the options or not.

tlabaj:
ist there an associated issue with this one?

jgiardino:
@redallen Thanks for looking into this issue. There are some updates that were recently made to the dropdown that I think are relevant to this discussion:
- changing the `disabled` attribute to `aria-disabled` so that we could manage focus and make disabled menu items discoverable to screen reader users
- enabling the use of Down arrow keys to shift focus from the toggle to the menu (this was make the dropdown usable in a recent version of JAWS with chrome)

These updates were made to the dropdown component based on feedback from a screen reader user while investigating a bug. Later, @jessiehuff created https://github.com/patternfly/patternfly-react/issues/5095 to implement similar behavior for the Select menu which is still open (and I've heard it's not an easy issue to fix).

In my opinion, consistency with Dropdown behavior is important, given similar affordances between the some of the select variants and the Dropdown. But ensuring accessibility for all users is also really important for this project. We can put a workaround in place in our project to avoid the error until it is resolved. But ultimately, we would prefer that these components are accessible when we provide a public eval in April. 

I suspect that if we were to fix https://github.com/patternfly/patternfly-react/issues/5095, then this issue would no longer exist, because disabled menu items would be able to receive focus. For that reason and the reasons mentioned above, I'd like to propose investigating a fix for that issue. 

**This is following up from our chat on slack here...** 
Here was my original comment:
> I remember a previous update made for the Dropdown to use aria-disabled on menu items, so that they could still receive focus. I think the use case for why that change was made still applies to the Select component (i.e. to make disabled menu items discoverable by screen reader users). If we applied this same change to the Select component, then that might resolve this issue. (cc @jessiehuff)

But in your reply, I’m not understanding what you mean by “No select items except the checkbox select can receive tab focus when opened using the mouse”. 

When I look at the Select component we have today, and use the mouse to click the toggle, the first checkbox option (which is enabled) does receive focus. Using the attribute `aria-disabled` shouldn't prevent focus, because it doesn’t affect functionality, it only communicates the semantics of the native `disabled` attribute. This is the behavior that we currently have in the Dropdown (e.g. the disabled items can receive focus).

I completely understand this is complicated, and appreciate the time you spent looking into this. But when I look at the preview from this PR, I notice that when I click the toggle for checkbox input, the menu displays, but I can’t access it with the keyboard. This was something we had to implement in the dropdown to make it accessible, and so I suspect this update could create the same a11y issue with the Select.

jschuler:
@kmcfaul I saw you mentioned another issue with focus trap and scrolling, I added this attribute some time ago that you can enable if you come across it
https://github.com/patternfly/patternfly-react/blob/6ccd7fdeb62b60ca41575bcd37cc293675bb8580/packages/react-core/src/helpers/FocusTrap/FocusTrap.tsx#L11

redallen:
> (and I've heard it's not an easy issue to fix)

I think you only need to ask @jschuler or I for help and we could figure out a way.

> I suspect that if we were to fix #5095, then this issue would no longer exist, because disabled menu items would be able to receive focus.

The issue would still exist because being able to receive focus is different from being tabbable (that is, having `tabindex >= 0`). FocusTrap doesn't care if an item is focusable, only if it is tabbable with a `tabindex >= 0`. Our dropdown does NOT have this functionality for disabled items and I think it makes little sense to add it since we want the tab key to close the dropdown.

> But in your reply, I’m not understanding what you mean by “No select items except the checkbox select can receive tab focus when opened using the mouse”.

The only thing that _DOES_ have `tabindex >= 0` in our select component is the `<input type="checkbox">`. Because disabled items don't have a `tabindex >= 0` in select (or dropdown) no items can receive tab focus when opening the toggle using the mouse.

> But when I look at the preview from this PR, I notice that when I click the toggle for checkbox input, the menu displays, but I can’t access it with the keyboard.

This matches the current behavior for ALL select examples. If you look at [the way it was before](https://www.patternfly.org/2020.16/components/select#checkbox-input) using tab after opening the select with your mouse is buggy at best and matches none of the other variants.

> This was something we had to implement in the dropdown to make it accessible

The dropdown has the same behavior when you click on the toggle and then press tab of closing the dropdown and focusing the toggle. The only difference is pressing the down arrow does nothing for the dropdown but scrolls the page for the select. Again, this is existing behavior.

jgiardino:
Thanks for the clarification, @redallen! This helps my understanding a lot more. :-D

> The only thing that DOES have tabindex >= 0 in our select component is the `<input type="checkbox">`. Because disabled items don't have a tabindex >= 0 in select (or dropdown) no items can receive tab focus when opening the toggle using the mouse *if the checkbox is disabled. That's what causes the crash in #5430.

This is where I'm wondering if using `aria-disabled` instead of `disabled` attribute would solve this issue. You're right that I can't tab to a disabled checkbox input, but I can if it has `aria-disabled`. It does require us to handle the functionality of disabling the input, which we get for free with the `disabled` attribute, but don't with `aria-disabled`. And this would be consistent with our updates to the dropdown that I described in my previous comment.

jessiehuff (approved):
After our conversation earlier, I'm good with this change. I like our planned approach of working on making it more accessible for different screen readers after this as Jenn's comments touch on too. 🙂 👍

dlabrecq (approved):

PR #5431, opened 2021-02-09
feat(Table): update column management demo to include label

### LINKED ISSUE #5273: Column Management Demo: Columns should also be selectable, not just the checkbox.
In the column management demo, selecting the column name should check the box. Currently you have to put the cursor in the checkbox.

Bug
OpenShift

jessiehuff:
**What**: Closes #5273

### DIFF
--- a/packages/react-table/src/demos/Table.md
+++ b/packages/react-table/src/demos/Table.md
@@ -911,15 +911,15 @@ class ColumnManagementAction extends React.Component {
                 id="check1"
                 onChange={this.handleChange}
               />
-              <label htmlFor="check1">
               <DataListItemCells
                 dataListCells={[
                   <DataListCell id="table-column-management-item1" key="table-column-management-item1">
-                    Repositories
+                    <label htmlFor="check1">
+                      Repositories
+                    </label>
                   </DataListCell>
                 ]}
               />
-              </label>
             </DataListItemRow>
           </DataListItem>
           <DataListItem aria-labelledby="table-column-management-item2">
@@ -931,15 +931,15 @@ class ColumnManagementAction extends React.Component {
                 id="check2"
                 onChange={this.handleChange}
               />
-              <label htmlFor="check2">
               <DataListItemCells
                 dataListCells={[
                   <DataListCell id="table-column-management-item2" key="table-column-management-item2">
-                    Branches
+                    <label htmlFor="check2">
+                      Branches
+                    </label>
                   </DataListCell>
                 ]}
               />
-              </label>
           </DataListItemRow>
           </DataListItem>
           <DataListItem aria-labelledby="table-column-management-item3">
@@ -951,15 +951,15 @@ class ColumnManagementAction extends React.Component {
                 id="check3"
                 onChange={this.handleChange}
               />
-              <label htmlFor="check3">
               <DataListItemCells
                 dataListCells={[
                   <DataListCell id="table-column-management-item3" key="table-column-management-item3">
-                    Pull requests
+                    <label htmlFor="check3">
+                      Pull requests
+                    </label>            
                   </DataListCell>
                 ]}
               />
-              </label>            
             </DataListItemRow>
           </DataListItem>
           <DataListItem aria-labelledby="table-column-management-item4">
@@ -971,15 +971,15 @@ class ColumnManagementAction extends React.Component {
                 id="check4"
                 onChange={this.handleChange}
               />
-              <label htmlFor="check4">
               <DataListItemCells
                 dataListCells={[
                   <DataListCell id="table-column-management-item4" key="table-column-management-item4">
-                    Workspaces
+                    <label htmlFor="check4">
+                      Workspaces
+                    </label>
                   </DataListCell>
                 ]}
               />
-              </label>
             </DataListItemRow>
           </DataListItem>
           <DataListItem aria-labelledby="table-column-management-item5">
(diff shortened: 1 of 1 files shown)


packages/react-table/src/demos/Table.md
```diff
@@ -908,15 +908,18 @@ class ColumnManagementAction extends React.Component {
                 aria-labelledby="table-column-management-item1"
                 isChecked={this.state.check1}
                 name="check1"
+                id="check1"
                 onChange={this.handleChange}
               />
+              <label htmlFor="check1">
```
KKoukiou, line 914: I feel that's something that the component should handle instead of the consumer. When DataListCheck is used the DataListItemCells component should be wrapped by a label. This is an implementation detail of DataListCheck that the users should not be bothered with. 
Secondly this behavior exists also in the expandable rows for table. The toggle button has to be clicked directly, I would expect that clicking anywhere on the row would expand the row. (not sure if i's the same issue)

nicolethoen (commented):
I think this looks and works better for sure! Would it be worth making sure that if the demo allows you to rearrange the order of the columns in the modal, that the rearranging would be reflected in the table? It may be better to remove that ability from the modal?

packages/react-table/src/demos/Table.md
jessiehuff: Hey @KKoukiou! Thanks for taking a look, that was definitely my first thought as well. I actually brought that up to the other team leads to determine if this should be a component or example level approach, but the difficulty here is all of the interactive elements that could be in the datalist. We wouldn't want clicking anywhere within a row to select the row, in general, since you could have other interactive elements there and the different ways it could be implemented. So the idea was to give control to consumers if they want to add a label in a situation like this or if they don't want that functionality in a more complex example like[ this example](https://www.patternfly.org/v4/components/data-list#checkboxes-actions-and-additional-cells). It's definitely really tough for us to balance all of the different potential use cases/allowing for flexibility while not over burdening the consumer. I'm very open to continuing this conversation on it though! @mcarrano, is there anything I missed in terms of why we were thinking of handling this at the demo/example level instead of at the component level?

KKoukiou: I can see you point here, I suggest to leave then how you have implemented it here, after all it's just an example, not an API that we are bound to, and if we figure out something else that works for the use case you describe above we can improve it in the future!

jessiehuff:
Hey @nicolethoen! That was definitely the plan that I got snagged on yesterday haha. So I pushed what I had so that Joachim could take a look. I think he helped me find a good solution for that aspect that I just pushed. 🙂

nicolethoen (approved):
looks good to me!

packages/react-table/src/demos/Table.md
```diff
@@ -906,85 +906,875 @@ class ColumnManagementAction extends React.Component {
...
             <DataListItemRow>
               <DataListCheck
                 aria-labelledby="table-column-management-item2"
-                isChecked={this.state.check2}
+                checked={this.state.check2}
                 name="check2"
+                id="check2"
                 onChange={this.handleChange}
               />
+              <label htmlFor="check2">
```
mcoker, line 934: While wrapping the entire cell in a label, there are parent > child relationships in the data list row component elements that this will break. For example, `<DataListItemRow>` is a flex layout, and it expects `<DataListItemCells>` to be a child (it generates `.pf-c-data-list__item-content` [here](https://github.com/patternfly/patternfly-react/blob/de2b31168ee79151ffef751fa26953b0a8d6cb3d/packages/react-core/src/components/DataList/DataListItemCells.tsx#L21)), and `.pf-c-data-list__item-content` is set to `flex-grow` to fill the available space in the row. When the label is introduced to wrap `.pf-c-data-list__item-content`, it breaks that flex parent/child layout, so the `.pf-c-data-list__item-content` no longer fills the row.

Here's what it should look like
Screen Shot 2021-02-11 at 11 20 28 AM

And here's what it looks like now
Screen Shot 2021-02-11 at 11 20 13 AM

All just a long winded explanation for why I would move this `<label>` into `<DataListCell>` and wrap the word(s) you want to serve as the label. After updating, you should see that  `.pf-c-data-list__item-content` grows to fill the entire row.

mcoker (changes requested):

packages/react-table/src/demos/Table.md
jessiehuff: Really great catch, I would not have realized that!! And thanks for the explanation too. 🙂  Just updated!

mcarrano (changes requested):
Oddly, I can only get the drag-and-drop to work on chrome, not Firefox. On Firefox I can drag the row, it creates the ghost row, but when I release the mouse it goes back to it's starting position.

mcoker (approved):
👀👍2️⃣🤠!!

jessiehuff:
> Oddly, I can only get the drag-and-drop to work on chrome, not Firefox. On Firefox I can drag the row, it creates the ghost row, but when I release the mouse it goes back to it's starting position.

@mcarrano That's really odd. I can't seem to replicate this- it seems to work for me on Firefox. :(

evwilkin:
> Oddly, I can only get the drag-and-drop to work on chrome, not Firefox. On Firefox I can drag the row, it creates the ghost row, but when I release the mouse it goes back to it's starting position.

I'm seeing the same result as @mcarrano - I tried in both Firefox 79 and Firefox 72, and in both the interactions look perfect until I release the mouse, then the dragged column snaps back to where it started.

jessiehuff:
@mcarrano @evwilkin I'm wondering if this might be a separate issue with the draggable datalist? I hadn't noticed it before, but when I try the [draggable datalist on Firefox](https://www.patternfly.org/v4/components/data-list/#draggable), I also have the same issue.

mcarrano (approved):
@jessiehuff Yes, this seems to be a draggable data list bug. So I will remove the block on this PR then and get the datalist issue into the queue for the next sprint.

tlabaj (approved):

PR #5438, opened 2021-02-10
feat(Navigation): Added support for spreading props to the NaveExpand…

### LINKED ISSUE #5384: cannot add custom props to NavExpandable button
We have a feature which highlights certain parts of the UI to guide the user based on attaching custom `data-` attributes. When we add a `data-` attribute to the `NavItem` the prop gets passed down the actual `button` / `a` tag. However when we do the same to the `NavExpandable` the attribute gets added to the `li` instead.

The result is we cannot reference the interactive button and the result is we highlight the entire list item grouping instead. The work around is for us to do a child selector however would prefer to stick to using react component APIs.
image

The request here is to make it consistent in the way props are passed down to the children so that the referenced element is always the interactive button. Or otherwise provide a way to add props to this component for `NavExpandable`.

`<NavItem data-foo="bar"...` => `<button data-foo="bar"...`
vs
`<NavExpandable data-foo="bar"...` => `<li data-foo="bar"...`

Enhancement / inconsistency bug

OpenShift Console 4.7

tlabaj:
…able button

**What**: Closes #5384

### DIFF
--- a/packages/react-core/src/components/Nav/NavExpandable.tsx
+++ b/packages/react-core/src/components/Nav/NavExpandable.tsx
@@ -30,7 +30,7 @@ export interface NavExpandableProps
   id?: string;
   /** allow consumer to optionally override this callback and manage expand state externally */
   onExpand?: (e: React.MouseEvent<HTMLLIElement, MouseEvent>, val: boolean) => void;
-  /** Additional props added to the NaveExpandable <button> */
+  /** Additional props added to the NavExpandable <button> */
   buttonProps?: any;
 }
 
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/Nav/NavExpandable.tsx
```diff
@@ -30,6 +30,8 @@ export interface NavExpandableProps
   id?: string;
   /** allow consumer to optionally override this callback and manage expand state externally */
   onExpand?: (e: React.MouseEvent<HTMLLIElement, MouseEvent>, val: boolean) => void;
+  /** Additional props added to the NaveExpandable <button> */
```
nicolethoen, line 33: small typo in comment

kmcfaul (approved):

nicolethoen (approved):

PR #5439, opened 2021-02-10
feat(TreeView): add keyboard interaction

### LINKED ISSUE #4726: Tree View: keyboard interactions and a11y
Follow up issue to implement custom keyboard interaction patterns and a11y to the TreeView component, as well as any clean up.

#4701

kmcfaul:
**What**: Closes #4726

Keyboard interaction:
- `ArrowUp` and `ArrowDown` move through the list of displayed tree items
- `ArrowRight` and `ArrowLeft` respectively open and close expandable folders. Focus is moved to the first child item when a folder is opened, and focus is moved to the parent folder if the `ArrowLeft` is triggered while within a folder, or remains on the folder if already on the expanded folder.
- `ArrowRight` and `ArrowLeft` iterate through a node for the checkbox variant, and do not expand/collapse a folder (left to `Space`).
- `Space` acts as a click to expand/collapse a folder or select a leaf item (or checkbox for checkbox variants)

@jessiehuff Let me know if there's anything I'm missing or anything I should add for a11y!

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeViewRoot.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewRoot.tsx
@@ -15,6 +15,12 @@ export class TreeViewRoot extends React.Component<TreeViewRootProps> {
 
   componentDidMount() {
     window.addEventListener('keydown', this.props.hasChecks ? this.handleKeysCheckbox : this.handleKeys);
+    if (this.props.hasChecks) {
+      (this.treeRef.current.getElementsByClassName('pf-c-tree-view__node-toggle')[0] as HTMLElement).tabIndex = 0;
+      (this.treeRef.current.getElementsByTagName('INPUT')[0] as HTMLElement).tabIndex = 0;
+    } else {
+      (this.treeRef.current.getElementsByClassName('pf-c-tree-view__node')[0] as HTMLElement).tabIndex = 0;
+    }
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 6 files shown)


mcarrano (commented):
I will defer to @jessiehuff on whether this is the interaction pattern we should be using. I did some testing and this seems to work well for the basic tree, but it gets confusing once checkboxes are introduced. I'm not really sure what's happening or is expected to happen in that case.

kmcfaul:
Hmm I'll check into that

kmcfaul:
Fix up for trees not getting the correct event propagation. I still have to update the handler to account for the checkbox variant since the structure is different.

kmcfaul:
Updated to handle the checkbox variant TreeView. However the interaction pattern is a tiny bit different.

`ArrowUp` and `ArrowDown` behave the same (taking the user up and down through the node-toggles or the checkboxes themselves), while `ArrowLeft` and `ArrowRight` iterate through the item (going between the toggle and the checkbox). `Space` still acts as a click, to expand/collapse a folder or select a checkbox.

Main difference is that the right/left arrows no longer expand a folder for this variant, and this handling requires the `hasChecks` flag on the `TreeView` (so individually added checkboxes to single items and not every item are not handled right now).

mcarrano (commented):
This works for me @kmcfaul . But I will defer to @jessiehuff for final approval.

jessiehuff (commented):
Awesome job on this!!! You really made those tabIndex changes quickly! I will say that I still see the left/right arrows go the checkboxes instead of expanding/collapsing in the checkbox example. However, I'm thinking this with some of the other updates we talked about might be a problem for another day!

jessiehuff (approved):

packages/react-core/src/components/TreeView/TreeViewRoot.tsx
```diff
@@ -0,0 +1,197 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/TreeView/tree-view';
+
+export interface TreeViewRootProps {
```
tlabaj, line 5: looks good.  just curios why you made this a new component?

tlabaj (approved):
looks good!

PR #5448, opened 2021-02-11
DualListSelector: use a more complicated example for complex options

### LINKED ISSUE #5436: Dual List Selector: enhancement to allow more user control over the dual list selector
I have been trying to use the [dual list selector component](https://www.patternfly.org/v4/components/dual-list-selector/#using-more-complex-options-with-actions) but am unable to get it to work with nested react components. Looking over the documentation I see the option to pass `React.ReactNode[]` for the available and chosen options. The example uses a list of spans. I would like to be able to pass in a more complex react component such as a list of:

```
<TextContent>
    <Text>"Name"</Text>
    <Text>"Description"</Text>
</TextContent>
```
I would like to have more control over the onListChange and onOptionSelect functions. Currently, I am able to pass in my desired list of elements but if I click on anything the component breaks.

This is a request for enhancement.

My product is Image Builder.

Thank you for any assistance you can offer.

KKoukiou:
Fixes #5436

### DIFF
--- a/packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
+++ b/packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
@@ -85,28 +85,50 @@ class BasicDualListSelectorWithSearch extends React.Component {
 
 ```js
 import React from 'react';
-import { Button, ButtonVariant, Dropdown, DropdownItem, DualListSelector, KebabToggle } from '@patternfly/react-core';
+import { Button, ButtonVariant, Dropdown, DropdownItem, DualListSelector, KebabToggle, Text, TextContent } from '@patternfly/react-core';
 import PficonSortCommonAscIcon from '@patternfly/react-icons/dist/js/icons/pficon-sort-common-asc-icon';
 
 class ComplexDualListSelector extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      availableOptions: [<span>Option 1</span>, <span>Option 3</span>, <span>Option 4</span>, <span>Option 2</span>],
+      availableOptions: [
+        <TextContent key={1}>
+          <span className="pf-c-dual-list-selector__item-text">Option 1</span>
+          <Text component={TextVariants.small}>Description 1</Text>
+        </TextContent>,
+        <TextContent key={2}>
+          <span className="pf-c-dual-list-selector__item-text">Option 2</span>
+          <Text component={TextVariants.small}>Description 2</Text>
+        </TextContent>,
+        <TextContent key={3}>
+          <span className="pf-c-dual-list-selector__item-text">Option 3</span>
+          <Text component={TextVariants.small}>Description 3</Text>
+        </TextContent>,
+        <TextContent key={4}>
+          <span className="pf-c-dual-list-selector__item-text">Option 4</span>
+          <Text component={TextVariants.small}>Description 4</Text>
+        </TextContent>
+      ],
       chosenOptions: [],
       isAvailableKebabOpen: false,
       isChosenKebabOpen: false
     };
 
     this.onSort = panel => {
+      const customSort = (a, b) => {
+        let returnValue = 0;
+        if (a.props.children[0].props.children > b.props.children[0].props.children) {
+          returnValue = 1;
+        }
+        if (a.props.children[0].props.children < b.props.children[0].props.children) {
+          returnValue = -1;
+        }
+        return returnValue;
+      }
       if (panel === 'available') {
         this.setState(prevState => {
-          const available = prevState.availableOptions.sort((a, b) => {
-            let returnValue = 0;
-            if (a.props.children > b.props.children) returnValue = 1;
-            if (a.props.children < b.props.children) returnValue = -1;
-            return returnValue;
-          });
+          const available = prevState.availableOptions.sort(customSort);
           return {
             availableOptions: available
           };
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -103,8 +120,12 @@ class ComplexDualListSelector extends React.Component {
         this.setState(prevState => {
           const available = prevState.availableOptions.sort((a, b) => {
             let returnValue = 0;
-            if (a.props.children > b.props.children) returnValue = 1;
-            if (a.props.children < b.props.children) returnValue = -1;
+            if (a.props.children[0].props.children > b.props.children[0].props.children) {
+              returnValue = 1;
+            }
+            if (a.props.children[0].props.children < b.props.children[0].props.children) {
+              returnValue = -1;
```
nicolethoen, line 127: you need to make this same change to the logic onSort in the chosen panel. :)

nicolethoen (commented):
Looks good other than the missing logic in the onSort function!

packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
KKoukiou: Good catch, sorry I missed it!

mcarrano (changes requested):
If we are going to create this example using descriptions following each option, then the spacing should match what we use in other menus where we have an item + description. See below. You'll see that the spacing makes it clearer that item labels and descriptions are paired. @mcoker can you provide some advice on the appropriate styling to use here?

Screen Shot 2021-02-11 at 9 27 33 AM

KKoukiou:
Hey @mcarrano I now replaced `<p>` with `<span>` which removed the extra padding above the description, I think it looks better now. If it's still not optimal we could always have another example, the reason I chose this is because it was mentioned in the issue. 
Now it looks like:
Screen Shot 2021-02-11 at 16 58 40

kmcfaul:
Did we change the sort direction at some point @nicolethoen ? I remember initially having both descending and ascending sort for the panes, but only one remains now (here and on our docs). 

Everything in the PR LGTM, but I was curious about the sort.

nicolethoen:
@kmcfaul from my memory, there was confusion around the sort directions and the icons, so for simplicity's sake, we just made it so the icons sorted in ascending order and didn't toggle between ascending and descending in the example.

nicolethoen (approved):

mcarrano (approved):
Definitely looks better. Thanks @KKoukiou !

mcoker:
@mcarrano while the example works, the way it's built really isn't ideal.  If we want to support that type of content, could we add support for a description to the items?

packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -85,28 +85,50 @@ class BasicDualListSelectorWithSearch extends React.Component {
...
 import PficonSortCommonAscIcon from '@patternfly/react-icons/dist/js/icons/pficon-sort-common-asc-icon';
 
 class ComplexDualListSelector extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      availableOptions: [<span>Option 1</span>, <span>Option 3</span>, <span>Option 4</span>, <span>Option 2</span>],
+      availableOptions: [
+        <TextContent key={1}>
+          <span className="pf-c-dual-list-selector__item-text">Option 1</span>
```
mcoker, line 97: This element isn't really intended to be nested this way. If we don't want to build in a description prop for these elements for this kind of layout, we could pass this to the option instead

```
Option 1
<TextContent>
  <Text component={TextVariants.small}>Description 1</Text>
</TextContent>
```

That would generate this markup

Screen Shot 2021-02-11 at 3 43 00 PM

And it will also retain the component item's intended font-size, which you can see in the first option here, compared to other options that have `<TextContent>` wrapping all of the item text:

Screen Shot 2021-02-11 at 3 43 04 PM

Or if you need the `key` on something that wraps the whole item, just add a `<span key={...}>` around all of that.

mcarrano:
@mcoker I thought about the same thing, but wasn't sure what the goal is for the PR. @KKoukiou is it specifically to add descriptions to list items or just to demonstrate how to pass a complex structure to a list item. If the former, I agree that there might be a better approach, but it depends on what the goal is.

KKoukiou:
> @mcoker I thought about the same thing, but wasn't sure what the goal is for the PR. @KKoukiou is it specifically to add descriptions to list items or just to demonstrate how to pass a complex structure to a list item. If the former, I agree that there might be a better approach, but it depends on what the goal is.

@mcarrano @mcoker thanks for taking the time to review this. My goal was to add a more complicated react node for the list items, because this changes the handlers and that was not clear to the person who created the assosiated issue. If you have another example that I can use, I am happy to change it. but it has to be something more than a plain `<div>` or a `<span>`. 

@mcoker shall I use your code snippet from above or change completely the item? That was not clear from the comments yet.

mcoker:
> shall I use your code snippet from above or change completely the item?

@KKoukiou I'm inclined to say we should add something to the item that wouldn't be confused with a standard enhancement we could make to the component. As in, if we add something that looks like a description, I wonder if we should just add an enhancement for a description? Otherwise someone might look at the example as the way to add a description, which may not be our actual recommendation.

Though I can't think of much else to add here, especially because the text element where the item content should go isn't setup hold much else other than the item text 😄

KKoukiou:
> > shall I use your code snippet from above or change completely the item?
> 
> @KKoukiou I'm inclined to say we should add something to the item that wouldn't be confused with a standard enhancement we could make to the component. As in, if we add something that looks like a description, I wonder if we should just add an enhancement for a description? Otherwise someone might look at the example as the way to add a description, which may not be our actual recommendation.
> 
> Though I can't think of much else to add here, especially because the text element where the item content should go isn't setup hold much else other than the item text smile

@mcoker I can also completely drop this PR, I mostly sent it because from https://github.com/patternfly/patternfly-react/issues/5436 it was apparent that the example is not super helpful and users might have trouble understanding how to use this component with more complicated scenarios. But you 're right, we should also be careful about not offering in example code something that is not really part of the component. That said, I the issue reporters saw the example code I pasted here, so we can close this issue and hope that not other similar reports will come. In the case you 'd rather keep it open feel free to reopen.

mcoker:
Sorry for the late reply! Closing for now sounds good to me 👍

PR #5450, opened 2021-02-11
Select(typeahead): remove onFilter function from example with custom objects

### LINKED ISSUE #4361: Using a empty string as a SelectOption with SelectVariant.typeahead crashes
@patternfly/react-core 3.158.1

This crashes when you start interacting with the Select component: https://codesandbox.io/s/green-water-3vl2t

On Firefox the error is `this.getDisplay(...) is undefined`.

It's the typeahead example from here https://www.patternfly.org/v4/documentation/react/components/select with an extra `{ value: '' }` entry in `options`.

Empty strings are not really useful as select options, but they should at least not crash the component, IMO. They could be filtered out, maybe.

### LINKED ISSUE #5449: Typeaheadselect example with custom object crashes on user input on docs page
https://www.patternfly.org/v4/components/select/#multiple-with-custom-objects

Just type anything and check the browser console. I did not check if it's a wrong example code or issue with the component itself, probably the first.

KKoukiou:
* fixes the crash on typeahead select component when an empty string option was
    passed (issue#4361)
* fixes the crash on typeahead select with custom objects example
    (issue#5449)
    
Fixes #5449
Fixes #4361

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -306,7 +306,24 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       } else {
         typeaheadFilteredChildren =
           typeaheadInputValue.toString() !== ''
-            ? childrenArray.filter(child => this.getDisplay(child.props.value.toString(), 'text').search(input) === 0)
+            ? childrenArray.filter(child => {
+                const valueToCheck = child.props.value;
+                // Dividers don't have value and should not be filtered
+                if (!valueToCheck) {
+                  return true;
+                }
+
+                const isSelectOptionObject =
+                  typeof valueToCheck !== 'string' &&
+                  (valueToCheck as SelectOptionObject).toString &&
+                  (valueToCheck as SelectOptionObject).compareTo;
+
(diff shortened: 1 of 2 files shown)


KKoukiou:
This needs more work - it still crashes on something else now.

KKoukiou:
Ok this is ready for review.

kmcfaul (approved):

tlabaj (approved):
Looks great.

rkaw92:
Hey, thank you for a prompt fix! One minor question from me (I requested the reopening of #4361 ). I see now that the fix prevents calling `getDisplay()` with an empty value, which was causing it to return `undefined`. However, if some other caller relies on `getDisplay` in the same way, they may encounter the same problem with it returning undefined. Shouldn't we always return a string from `getDisplay()`, making it easier and more predictable for consumers of this method? I'm not sure what the rationale for the guard clause `if (!value)` was, originally, but at the moment it looks like a bug generator since the "undefined" leaks to callers.

KKoukiou:
@rkaw92 I will reopen the issue and take a look again. Thanks for reporting back!

PR #5451, opened 2021-02-11
fix(Select): fix typeahead variant crashing when empty string option is passed

### LINKED ISSUE #4361: Using a empty string as a SelectOption with SelectVariant.typeahead crashes

KKoukiou:
Fixes #4361

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -306,7 +306,9 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       } else {
         typeaheadFilteredChildren =
           typeaheadInputValue.toString() !== ''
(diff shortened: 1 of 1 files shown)


KKoukiou:
Closes in favor of, https://github.com/patternfly/patternfly-react/pull/5450 they would have merge conflict anyway

Issue #5332, opened 2021-01-15
Attribute-value search input


mcarrano:
Enhance the Search input component to support advanced attribute-value search use cases that allow the user to enter a query string directly in the input field or fill out a search form. 

Designs: https://marvelapp.com/prototype/276i4086

Core: https://github.com/patternfly/patternfly/pull/3783

For keyboard accessibility - treat similar to other dropdown components. Need to manage focus shift into and out of the form.

Requesting project: KeyCloak
Date needed: February
Contacts: @yih-wang, @amysueg

nicolethoen:
@mcarrano do we want to include the autocomplete feature as part of this issue? Or make that a separate issue?

mcarrano:
@nicolethoen I think that could be a separate issue. What do you prefer?

nicolethoen:
I think we'd at least need to meet about what the interaction and configuration expectations would be. Like what sorts of information would be needed from the consumer and when the auto complete would be invoked.

mcarrano:
@nicolethoen I've opened a new issue here so we can elaborate on the requirements for this: https://github.com/patternfly/patternfly-react/issues/5387  I'll add this to the next sprint as follow-on work and can set up a time to talk further bout what's involved. Thanks for bringing this up.

PR #5454, opened 2021-02-11
feat(SearchInput): add advanced search feature

nicolethoen:
Closes: #5332 

Note: tests coming very soon - just wanted to get the PR up for review

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -7,12 +7,29 @@ import AngleDownIcon from '@patternfly/react-icons/dist/js/icons/angle-down-icon
 import AngleUpIcon from '@patternfly/react-icons/dist/js/icons/angle-up-icon';
 import TimesIcon from '@patternfly/react-icons/dist/js/icons/times-icon';
 import SearchIcon from '@patternfly/react-icons/dist/js/icons/search-icon';
+import CaretDownIcon from '@patternfly/react-icons/dist/js/icons/caret-down-icon';
+import ArrowRightIcon from '@patternfly/react-icons/dist/js/icons/arrow-right-icon';
+import { ActionGroup, Form, FormGroup } from '../Form';
+import { InputGroup } from '../InputGroup';
+import { TextInput } from '../TextInput';
+import { GenerateId } from '../../helpers';
+
+export interface SearchAttribute {
+  attr: string;
+  display: React.ReactNode;
+}
 
 export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange' | 'results' | 'ref'> {
   /** Additional classes added to the banner */
   className?: string;
   /** Value of the search input */
   value?: string;
+  /** Array of attribute values */
+  attributes?: string[] | SearchAttribute[];
+  /** */
+  hasWordsAttrLabel?: string;
+  /** */
+  advancedSearchDelimiter?: string;
   /** The number of search results returned. Either a total number of results,
    * or a string representing the current result over the total number of results. i.e. "1 / 5" */
   resultsCount?: number | string;
@@ -22,6 +39,12 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   placeholder?: string;
   /** A callback for when the input value changes. */
   onChange?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
+  /** A callback for when the search button clicked changes. */
+  onSearch?: (
+    value: string,
+    event: React.SyntheticEvent<HTMLButtonElement>,
+    attrValueMap: { [key: string]: string }
+  ) => void;
   /** A callback for when the user clicks the clear button */
   onClear?: (event: React.SyntheticEvent<HTMLButtonElement>) => void;
   /** Function called when user clicks to navigate to next result */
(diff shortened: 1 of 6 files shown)


kmcfaul:
Pressing `space` in the username/first name fields adds a space to the `has words` field, which I found a little odd. I understand space not doing anything for the first two fields (though maybe it should work in the first name field, for compound names?) but I don't think it should be editing anything for a different field.

mcarrano (changes requested):
This is looking great @nicolethoen ! I encountered the same issue as @kmcfaul . The only other nit I had was that we should not assume a default for the delimiter character but make the consumer declare it explicitly. Not sure that we have an accepted standard inside Red Hat.

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -30,70 +53,245 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>,
...
+          </>
         )}
-      </div>
+      </InputGroup>
+      {attributes.length > 0 && showSearchMenu && (
+        <div className={css(styles.searchInputMenu)}>
+          <Form>
+            {buildFormGroups()}
+            <ActionGroup>
+              <Button variant="primary" onClick={onSearchHandler}>
```
mcoker, line 283: Random but should this have `type="submit"`?

```suggestion
              <Button type="submit" variant="primary" onClick={onSearchHandler}>
```

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -30,70 +53,245 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>,
...
+                {submitSearchButtonLabel}
+              </Button>
+              {!!onClear && (
+                <Button variant="link" onClick={onClear}>
```
mcoker, line 287: `type="reset"`?

```suggestion
                <Button  type="reset" variant="link" onClick={onClear}>
```

packages/react-core/src/components/SearchInput/examples/SearchInput.md
```diff
@@ -169,3 +169,67 @@ TextInputSelectAll = () => {
   );
 };
 ```
+
+### Advanced
+```js
+import React from 'react';
+import { Checkbox, SearchInput, Stack, StackItem } from '@patternfly/react-core';
```
mcoker, line 176: Do you need `Stack` and `StackItem`? Doesn't look like they're used.

```suggestion
import { Checkbox, SearchInput } from '@patternfly/react-core';
```

mcoker (commented):
Looks great! 👍 Left a few comments for review.

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -7,12 +7,29 @@ import AngleDownIcon from '@patternfly/react-icons/dist/js/icons/angle-down-icon
...
 import SearchIcon from '@patternfly/react-icons/dist/js/icons/search-icon';
+import CaretDownIcon from '@patternfly/react-icons/dist/js/icons/caret-down-icon';
+import ArrowRightIcon from '@patternfly/react-icons/dist/js/icons/arrow-right-icon';
+import { ActionGroup, Form, FormGroup } from '../Form';
+import { InputGroup } from '../InputGroup';
+import { TextInput } from '../TextInput';
+import { GenerateId } from '../../helpers';
+
+export interface SearchAttribute {
+  attr: string;
```
tlabaj, line 18: can you add descriptions to these and expose it in the docs.

packages/react-core/src/components/SearchInput/examples/SearchInput.md
```diff
@@ -169,3 +169,67 @@ TextInputSelectAll = () => {
   );
 };
 ```
+
+### Advanced
```
tlabaj, line 173: Add `SearchAttribute` on line 5

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -7,12 +7,29 @@ import AngleDownIcon from '@patternfly/react-icons/dist/js/icons/angle-down-icon
...
   /** Additional classes added to the banner */
   className?: string;
   /** Value of the search input */
   value?: string;
+  /** Array of attribute values */
+  attributes?: string[] | SearchAttribute[];
+  /** */
+  hasWordsAttrLabel?: string;
+  /** */
+  advancedSearchDelimiter?: string;
```
tlabaj, line 32: this is also missing a description

tlabaj (changes requested):

kmcfaul (approved):
I'm still seeing the odd behavior with `space` but I don't think its a blocker and can be addressed later on. Everything else looks good.

tlabaj (approved):
LGTM

evwilkin:
I'm seeing the space bar issues @kmcfaul mentioned, and also seeing that if you add a space in the middle of the text for Username or First name fields it moves the text following the space from that field down to the Has words field - not sure if that's intentional or not

mcoker (approved):
L🍕TM!! 🥳

nicolethoen:
@evwilkin @kmcfaul 
There is possibly a lot more follow up work to be done to flesh out how to parse the search strings, possibly using quotation marks(?) or other special characters.

For now, the attribute/value pair cannot have any spaces because I have no way to know when a naked value is associated with a particular attributes, or if it's a stand alone value (and thus belongs in the 'has words' field). The `space` is what delimits one attribute/value pair from another. So if there is no attribute name paired (without a space) with a value, I am assuming it is a stand alone value.

Basically, I'm using the `space` to delimit attribute/value pairs.

jessiehuff (commented):
This looks pretty good so far, Nicole!! It seems pretty decent in VO. I don't notice anything odd on Chrome, but I did find it weird that in Safari when VO gets to the input it just reads "first name John" instead of reading the aria label. I'm not sure why that would happen though.
My only concern is the keyboard interaction. I would expect it to follow more of a dropdown interaction pattern based on the design, and here I have to press space or enter on the button to get to the advanced and then tab to that dropdown before I'm able to interact with the form. I would think that the focus would shift into the dropdown like other implementations. Also, while using Tab to navigate in the dropdown here isn't too bad, typically in a dropdown you'd use up/down arrow keys. The general rule of thumb is that Tab typically navigates between components on a page or from region to region and arrow keys navigate within a component.

mcarrano (approved):
I am also good to go ahead and merge this. The space issue is indeed a bit odd. @nicolethoen maybe we can have more conversation about that and open a follow-on issue if needed.

@evwilkin I think that your other issue is expected behavior as I don't expect we would recognize a space as part of an attribute vale/name.

evwilkin (approved):

nicolethoen:
@jessiehuff I can make it so when the menu toggles open, focus goes to the first element in the menu. no problem!

This is actually a form more than a dropdown - even though it technically drops down.... gah!

tlabaj:
@nicolethoen can you please open up a follow up issue for the accessibility concerns and we can then merge this one.

jessiehuff:
> This is actually a form more than a dropdown - even though it technically drops down.... gah!

Yeah, some of these components that combine different interaction patterns can be confusing to know the right path forward. In the situation of a form in a modal, Tab is typically used. In a dropdown, up/down arrows are typically used. So combining them can leave us with a subjective decision on that. I think when Jenn and I discussed it before we were leaning towards arrow keys, but I'm ok with merging for now and maybe just following up later with any updates we iron out. The most pressing issue though is the shift in focus. I think a keyboard user would expect that if they select the advanced options button, they want to interact with it.

jessiehuff (approved):

nicolethoen:
@tlabaj issue opened!

PR #5465, opened 2021-02-15
feat(CodeEditor): Make 'code' prop controlled

### LINKED ISSUE #5421: Enhancement: Code editor as a controlled component
I've been evaluating the beta Code Editor component for use in Ansible Tower, but it's lacking a feature we would need in order to use it. Ideally the component would be a controlled component, meaning that if the `code` prop changes, the component would update with the new code displayed.

In our use case, we need this specifically because we have some inputs that can be toggled between JSON and YAML code. When the user toggles the view, we translate the code between the two and update the prop. In the current version of the code editor, this updated prop is ignored.

Related issue in AWX project: https://github.com/ansible/awx/issues/8905

nicolethoen:
**What**: Closes #5421

### DIFF
--- a/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
@@ -177,6 +177,13 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
     this.setState({ value });
   };
 
+  componentDidUpdate(prevProps: CodeEditorProps) {
+    const { code } = this.props;
+    if (prevProps.code !== code) {
+      this.setState({ value: code });
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -177,6 +177,13 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
     this.setState({ value });
   };
 
+  componentDidUpdate(prevProps: CodeEditorProps) {
+    const { code } = this.props;
```
tlabaj, line 181: could we add a test for this?

tlabaj (approved):

dlabrecq (approved):

PR #5468, opened 2021-02-15
feat(SearchInput): add isDisabled and clean up focus handling

### LINKED ISSUE #5462: SearchInput: Move focus into advance search menu when it toggles open
I would expect that the focus would shift into the dropdown when the user toggles the advance search menu open, especially when using the keyboard to open the menu.

### LINKED ISSUE #5463: feat(SearchInput): add isDisabled
It would be nice if the SearchInput component would provide an isDisabled property and according state.

My use case of instance is:
I've a toolbar and a table, while the data is still loading I'd like to disable the items in the toolbar, since no interaction with the table makes any sense. 

It's possible for the Button and TextInput component, but not for the SearchInput. That feels inconsistent to me.

nicolethoen:
**What**: Closes #5463 and Closes #5462

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -63,6 +63,8 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   submitSearchButtonLabel?: string;
   /** Label for the button which opens the advanced search form menu */
   openMenuButtonAriaLabel?: string;
+  /** Flag indicating if search input is disabled */
+  isDisabled?: boolean;
 }
 
 const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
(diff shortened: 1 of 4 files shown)


mcarrano (commented):
@nicolethoen I just finished a meeting with @mcoker about disabled styling for the slider. He will be putting up a core PR for that. Do you think it makes sense to hold on this and integrate the disabled styling into a new example?

nicolethoen:
> @nicolethoen I just finished a meeting with @mcoker about disabled styling for the slider. He will be putting up a core PR for that. Do you think it makes sense to hold on this and integrate the disabled styling into a new example?

@mcarrano I guess that depends on what you and @mcoker decided. This search input is like a text input in that the input and all buttons on it become disabled if you set the isDisabled prop to true. So it felt very straight forward to me.

mcarrano:
@nicolethoen behaviorally that's correct. But we also want to define a new set of styles to visually "gray out" the slider. @mcoker will be putting up a core PR to add those. So in addition to not accepting events, we'd want to show those styles.

nicolethoen:
are these disabled styles going to also apply to the search input? I can wait until they get released in core and pull the new styles into this PR

jessiehuff (commented):
This looks great so far, Nicole! 🙂  I've been thinking about this interaction more, and I think that this use case does fit more of a modal interaction pattern than a dropdown so I think using the Tab key to navigate inside the advanced search form is fine. The only thing I'd love for us to add is the Esc key being able to close the advanced search and return focus to the button that opened it.

mcarrano:
My bad! I guess I should read things more carefully! There was a separate issue filed to add an `isDisabled` prop to the Slider so I had that on the brain, so please disregard my earlier comment @nicolethoen . This should be fine.

jessiehuff (commented):
It looks like the Esc key works, but now the Tab key is also closing it instead of navigating through the form. 🙁

jessiehuff (approved):
Awesome! 🙂

tlabaj (commented):
Can you please update integration app?

tlabaj (approved):
lgtm

jessiehuff (approved):

PR #5469, opened 2021-02-16
DualListSelect: add onAvailableOptionsSelectInputChanged and onChosenOptionsSelectInputChanged

### LINKED ISSUE #5436: Dual List Selector: enhancement to allow more user control over the dual list selector

KKoukiou:
The use case can be as seen in the linked issue, that the cunsumer might
need to use the left search bar to query an API for available options.

This is understandable as the list of available options can get
really long in some use-cases and in that case the API calls should only
fetch the options that are relevant to the user input.

Fixes #5436

### DIFF
--- a/packages/react-core/src/components/DualListSelector/DualListSelector.tsx
+++ b/packages/react-core/src/components/DualListSelector/DualListSelector.tsx
@@ -69,8 +69,12 @@ export interface DualListSelectorProps {
   isSearchable?: boolean;
   /** Accessible label for the search input on the available options pane. */
   availableOptionsSearchAriaLabel?: string;
+  /** A callback for when the search input value for available options changes. */
+  onAvailableOptionsSearchInputChanged?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
   /** Accessible label for the search input on the chosen options pane. */
   chosenOptionsSearchAriaLabel?: string;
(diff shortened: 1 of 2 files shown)


KKoukiou:
I did not add the new properties in the demo or example, because I think it's a corner case usage the example from the linked issue, and we should not promote the usage of these properties.

jkozol:
I reviewed this in the preview and it meets the needs of #5436. Thank you @KKoukiou. 

This feature is urgently needed by Image Builder.

nicolethoen (approved):
LGTM - i like it :) works well

redallen (approved):

dlabrecq (commented):
LGTM

PR #5470, opened 2021-02-16
fix(Progress): make proper usage of accessibility properties     

### LINKED ISSUE #5472: Progress: aria-labelledby should be customizable
There are use cases in which the Progress component does not have a Title but has text nearby that labels it. For example, in Cockpit they have some progress bars inside tables: 

image

In this scenario, the user should be able to customize their `aria-labelledby` to point to the text. (In this example, the text that the `aria-labelledby` would point to would be "CPU" and "Memory".)

KKoukiou:
Closes #5472

### DIFF
--- a/packages/react-core/src/components/Progress/Progress.tsx
+++ b/packages/react-core/src/components/Progress/Progress.tsx
@@ -38,8 +38,10 @@ export interface ProgressProps extends Omit<React.HTMLProps<HTMLDivElement>, 'si
   isTitleTruncated?: boolean;
   /** Position of the tooltip which is displayed if title is truncated */
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
-  /** Adds accessible text to the ProgressBar. Required when title not used */
+  /** Adds accessible text to the ProgressBar. Required when title not used and there is not any label associated with the progress bar */
   'aria-label'?: string;
+  /** Associates the ProgressBar with it's label for accessibility purposes. Required when title not used */
+  'aria-labelledby'?: string;
 }
 
 export class Progress extends React.Component<ProgressProps> {
@@ -58,7 +60,8 @@ export class Progress extends React.Component<ProgressProps> {
     valueText: null as string,
     isTitleTruncated: false,
     tooltipPosition: 'top' as 'auto' | 'top' | 'bottom' | 'left' | 'right',
-    'aria-label': null as string
+    'aria-label': null as string,
+    'aria-labelledby': null as string
   };
 
   id = this.props.id || getUniqueId();
@@ -81,24 +84,33 @@ export class Progress extends React.Component<ProgressProps> {
       isTitleTruncated,
       tooltipPosition,
       'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledBy,
       ...props
     } = this.props;
 
     const progressBarAriaProps: AriaProps = {
-      'aria-labelledby': title ? `${this.id}-description` : null,
-      'aria-label': ariaLabel ? ariaLabel : null,
       'aria-valuemin': min,
       'aria-valuenow': value,
       'aria-valuemax': max
     };
 
+    if (title || ariaLabelledBy) {
+      progressBarAriaProps['aria-labelledby'] = title ? `${this.id}-description` : ariaLabelledBy;
+    }
+
+    if (ariaLabel) {
+      progressBarAriaProps['aria-label'] = ariaLabel;
+    }
+
     if (valueText) {
       progressBarAriaProps['aria-valuetext'] = valueText;
     }
 
-    if (!title && !this.props['aria-label']) {
+    if (!title && !ariaLabelledBy && !ariaLabel) {
       /* eslint-disable no-console */
-      console.warn('An accessible aria-label is required when using the progress component without a title.');
+      console.warn(
+        'One of aria-label or aria-labelledby properties should be passed when using the progress component without a title.'
+      );
     }
 
     const scaledValue = Math.min(100, Math.max(0, Math.floor(((value - min) / (max - min)) * 100)));
--- a/packages/react-core/src/components/Progress/__tests__/Generated/__snapshots__/Progress.test.tsx.snap
+++ b/packages/react-core/src/components/Progress/__tests__/Generated/__snapshots__/Progress.test.tsx.snap
@@ -12,7 +12,6 @@ exports[`Progress should match snapshot (auto-generated) 1`] = `
     parentId="''"
     progressBarAriaProps={
       Object {
-        "aria-label": null,
         "aria-labelledby": "''-description",
         "aria-valuemax": 100,
         "aria-valuemin": 0,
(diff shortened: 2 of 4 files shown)


nicolethoen:
I like it!

Does this constitute a breaking change? Is it possible that we could leave the aria-label as well as add aria-labelledby?

And is there an issue to link to this PR?

KKoukiou:
> I like it!
> 
> Does this constitute a breaking change? Is it possible that we could leave the aria-label as well as add aria-labelledby?

We could leave the aria-label id needed but removing it makes it cleaner what should be used. if it's not a breaking change I 'd rather drop it.

> 
> And is there an issue to link to this PR?

The original issue was this https://github.com/patternfly/patternfly-react/issues/5396 but it was closed, in my opinion the resolution should be what I do in this PR. So I can either link the old issue or create a new one, but it's basically the same.

nicolethoen:
> We could leave the aria-label id needed but removing it makes it cleaner what should be used. if it's not a breaking change I 'd rather drop it.

Since it's a valid html attribute, it might not be a breaking change - someone could still include it and it could possibly still be applied. Actually, if you put back the `aria-label` into the `progressBarAriaProps` it's possible that even if you remove the `aria-label` prop from the Progress props, it'll still be applied to the ProgressBar correctly if someone was using it. This may prevent the breaking change.

and linking the old issue is probably fine.

KKoukiou:
> > We could leave the aria-label id needed but removing it makes it cleaner what should be used. if it's not a breaking change I 'd rather drop it.
> 
> Since it's a valid html attribute, it might not be a breaking change - someone could still include it and it could possibly still be applied. Actually, if you put back the `aria-label` into the `progressBarAriaProps` it's possible that even if you remove the `aria-label` prop from the Progress props, it'll still be applied to the ProgressBar correctly if someone was using it. This may prevent the breaking change.
> 
> and linking the old issue is probably fine.

I will still repost for the failing tests.

jessiehuff (commented):
I'm a little confused by your suggestion. Currently, if there is a title we use `aria-labelledby` to point to that title text. However, in scenarios in which there was no title used, there was nothing for this `aria-labelledby` to point to.

In the scenario in which there is no text, I would recommend using an `aria-label` over an `aria-labelledby` because if there's no text, I can't think of anything appropriate for the progress component to point to (which was why that prop was added in the previous PR). If something else has been added to be pointed to, I can see the argument for `aria-labelledby` (for example, pointing at a heading element). 

However, even in that scenario it would make sense to me to still keep `aria-label` as a prop here. The reason being, if the user adds an `aria-label` without the prop, this label gets added to the outer div, creating a less ideal experience in a screen reader. It reads as "Title, group" then when you continue forward it reads "33 percent, progress indicator" then you move forward again before reading "end of group". When we add the prop to the progress bar itself, it reads "Title, 33 percent, progress indicator" which I think is a lot cleaner and a more straight forward experience. 

I'm ok with adding `aria-labelledby` in addition to `aria-label` if we provide documentation on adding something for it to point to (though I wonder if that's a design question of showing an example of some other kind of text we could point to?). Otherwise, if users use `aria-labelledby` and don't know how to properly use it, they will end up with the same error that the original issue discussed.

KKoukiou:
@jessiehuff thanks for giving some useful accessibility feedback here. So my idea is the following:

Every user of that component should provide somehow an associated label, that's either by using the 'title' property or externally.

For example in cockpit we have some progress bars inside tables.
Screen Shot 2021-02-16 at 16 42 34

aria-label would be useful if there was not label, but for this component this should never be there case! The progress bar needs some description about what it is about, therefore we need to use some kind of label. 

So the users should not really provide aria-label because there would always be a label associated with this component.
Keeping it just for the case that someone uses it, is not a satisfying reason for me either. The reason is that every property we expose should have a reason for being used, in this case aria-label does not add some functionality that it's missing.

So I suggest still to not expose aria-label, unless it's a breaking change, since that might misguide the users to use extra unnecessary accessibility properties.

jessiehuff:
Ah ok, I understand your use case/reasoning now! Based on the fact that the [Single line example](https://www.patternfly.org/v4/components/progress#single-line) was pre-existing and the description in the original issue, I was thinking the use case was when there's a visual affordance instead of any visible text. I double checked with Matt Carrano and Jenn on potential use cases, and it seems like there still might be cases where using `aria-label` might be necessary. Jenn pointed out that there might also be situations where the consumer might not be able to define an ID on a visible label. It's always hard to predict all potential use cases. Nicole's point is correct as well, which Matt echoed, removing the prop now would be considered a breaking change so we likely shouldn't completely remove it. Matt also stated that he believed the single line instance is valid so that should probably be added back.

I definitely see your point now though, and I agree that customizing your `aria-labelledby` makes the most sense in your use case. (I wonder if an example should be added to show a customized `aria-labelledby`? Just a thought though.) Perhaps we can just add the `aria-label` back to account for all potential use cases and have both available for customization. Based on the discussion though, it sounds like this one is a good candidate for better documentation in the future. It should be clear to users when `aria-labelledby` should be used and when `aria-label` is the preferred attribute.

Thanks for sharing your use case, that definitely helps! I'm going to create an issue for you to point to with this PR. 🙂

KKoukiou:
@jessiehuff sorry for the late response here :) I updated the PR according to our discussion, hope it looks fine.

KKoukiou:
Ehm @jessiehuff I am not sure what' the issue with the failing a11y test.

nicolethoen:
@KKoukiou I think the a11y test is failing because the last example has an element `.pf-c-progress__bar` which has an `aria-labelledby="progress-label"` which should map to an element with an `id="progress-label"` but there is no element on the page with that `id`. I think the `DescriptionListTerm` needs to have an unique `id` and that `id` is what should be set as the value of the `aria-labelledby` attribute.

KKoukiou:
> @KKoukiou I think the a11y test is failing because the last example has an element `.pf-c-progress__bar` which has an `aria-labelledby="progress-label"` which should map to an element with an `id="progress-label"` but there is no element on the page with that `id`. I think the `DescriptionListTerm` needs to have an unique `id` and that `id` is what should be set as the value of the `aria-labelledby` attribute.

Oh thanks a lot :) I will repost shortly.

nicolethoen (approved):

jessiehuff (approved):
LGTM! Thanks so much for taking a look at this. I appreciate all of the a11y work, @KKoukiou!! 🙂

packages/react-core/src/components/Progress/Progress.tsx
```diff
@@ -81,24 +84,33 @@ export class Progress extends React.Component<ProgressProps> {
...
+    }
+
     if (valueText) {
       progressBarAriaProps['aria-valuetext'] = valueText;
     }
 
-    if (!title && !this.props['aria-label']) {
+    if (!title && !ariaLabelledBy && !ariaLabel) {
       /* eslint-disable no-console */
-      console.warn('An accessible aria-label is required when using the progress component without a title.');
```
tlabaj, line 101: can you add a jest test to verify console warning?

KKoukiou: I think there is already one, see: 'progress component generates console warning when no accessible name is provided'

KKoukiou:
@tlabaj ping. I think the test code covers already the console warning. Let me know if I understand something wrong.

tlabaj (approved):

Issue #5333, opened 2021-01-15
Menu component: add drill-down with breadcrumbs


mcarrano:
Add optional bread crumbs to drill down menu. See https://github.com/patternfly/patternfly/issues/3491 for designs and core implementation https://github.com/patternfly/patternfly/pull/3797.

When using the collapsible breadcrumb, the consumer should be able to set how many levels to display before collapsing via a prop.

Requesting projects: Multiple (cloud platform, app services)
Date needed: March
Contacts: @ajacobs21e @nlcwong @aprilma419

tlabaj:
closed by #5548

Issue #5144, opened 2020-11-16
Menu component: Add keyboard interaction


jschuler:
Add arrow key interaction to the menu component

tlabaj:
cc @mcarrano

jessiehuff:
I think another point that would be worth investigating here is if we can support the two main variants of menu: 
- A menu of actions (these are our components that would semantically fit `role="menu"` like dropdown, application launcher, etc.) which I would think we could follow a pattern like [this](https://www.w3.org/TR/wai-aria-practices-1.2/examples/menu-button/menu-button-actions-active-descendant.html ) and consider using `aria-activedescendant`
- A menu of options (this is any component that would semantically fit `role="listbox"` like select)
I think it's important here to make sure that the keyboard interaction works within a screen reader and that the semantics are communicated to a user to communicate how to interact with it.

jessiehuff:
This one can be a bit tricky since there are so many potential variations of menu. I think I would expect up/down arrow keys to navigate between menu items, right/left to navigate through the drill down aspects. In general, I'm thinking this menu component might be most like [these expectations](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-12). 

With the fly out use case, we currently have that menu appear on hover. If it stays on hover and arrow keys are implemented, this menu should appear on focus as well for keyboard users. If using Esc on a drill down variation, focus would go to whatever toggle triggered that menu (this could be the previous menu, the original button that opened the button, etc).

I detailed breadcrumb keyboard expectations [here](https://github.com/patternfly/patternfly-react/issues/5361#issuecomment-790985061), but to summarize:
Interactive breadcrumb options would be navigated through Tab, the badge button uses enter and should be treated as a modal interaction, arrow keys within that modal, Esc to move focus back to badge button, Tab to the menu, then arrow keys should be used between the menu items.

Action or favoriting variations should be handled similarly to the Selectable favorites with left/right arrow keys. The filtering variation could be very tricky if we expect to support a similar interaction as the Typeahead Select (being able to type text into the input, see potential suggestions in the menu and then use arrow keys to navigate them). Perhaps we'd want to take a similar approach as we did with Select and use `aria-activedescendant`, but that probably needs to be tested. 

I tried to cover all of our specific use cases, but let me know if I missed any particular interaction!

PR #5485, opened 2021-02-23
Menu drilldown

### LINKED ISSUE #5024: Add multi-level drill down menu

kmcfaul:
**What**: Follows up PR #5321. Closes #5024 and begins to address #5333 (+breadcrumbs, not yet on breadcrumb dropdowns)


kmcfaul:
@jschuler I re-added the other menu demos, is that fine or was there a reason they were removed?

jschuler:
@kmcfaul thanks for adding them back in, I had just removed them temporarily to work on this feature :)

mcoker (approved):
LGTM!

kmcfaul:
Fixed the animation bug for sub-level drilldowns.
Also added the first round for drilldown breadcrumbs in a separate demo. Doesn't cover breadcrumb dropdowns yet.

Went with using our current Breadcrumb components over making a container "DrilldownMenuHeader" component. This way gives the user finer control over what the Breadcrumb/BreadcrumbItem should do and look like (opposed to passing in an array of data to an aggregate component) and later down the line will let the user control where in the chain the breadcrumb dropdown should be. 

I did have to update Breadcrumb to get the structure we use in core for the link styling, if using the `component` prop to do this interferes with anything else I can take it back out and make a wrapper component in Menu.

@mcoker I do have a question about using the link version of BreadcrumbHeading, which is in the core example. Is it ok just to pass a string for the drilldown? An href of "#" will reset the whole drilldown (because the page reloads).

mcarrano (commented):
This is looking great @kmcfaul . Love the animation effects. My only question/ask if if we can expose the label used for the Root node via a prop. The consumer might want to replace this with something more germane to their use case. @doruskova can you also take a look when you have a chance?

kmcfaul:
@mcarrano The text in the breadcrumbs is defined on the app-side so it should be no problem to change it. Keyboard interaction is not in yet beyond any default handling by the browser, we will likely need a follow up issue for it.

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -692,3 +700,535 @@ class MenuOptionMultiSelect extends React.Component {
...
+    const rootMenu = () => (
+      <Menu
+        id="rootMenu"
+        containsDrilldown
+        isMenuDrilledIn={menuDrilledIn.includes('rootMenu')}
+        onSelect={this.onSelect}
+      >
+        <MenuContent
+          getHeight={height => this.setHeight('rootMenu', height)}
+          style={{ '--pf-c-menu__content--Height': `${menuHeights[activeMenu]}px` }}
```
redallen, line 782: If we want setting the menu height based off the drilled down MenuItem to be the default behavior can we store that state in MenuContent instead?

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -692,3 +700,535 @@ class MenuOptionMultiSelect extends React.Component {
...
+  MenuItemStorage,
+  MenuItemAppGrouping,
+  MenuItemLabels
+} from './Submenus.js';
+
+class MenuWithDrilldown extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      activeItem: null,
```
redallen, line 725: It doesn't look like `activeItem` is used in render or logic?

redallen (changes requested):
Wow this component is getting complicated! I wonder if there's more ways than what I've mentioned to make implementing drilldown menus simpler for the consumer?

I'm thinking about the API more broadly now and I think it might be possible for the consumer to not assign `itemId`s, `onClick`s, `isOnPath`, and `isMenuDrilledIn` at all. Couldn't we do that using `React.Children` in `Menu` if it has `containsDrilldown` and use index-based `itemId`s instead? Then if a consumer wants manual control of the drilled down menu they can pass `drilldownPath=[1,2,3]` to indicate the 3rd MenuItem of the 2nd MenuItem of the 1st MenuItem is drilled in.

packages/react-core/src/components/Menu/examples/Menu.md
kmcfaul: `activeItem` indicates the selected item (set by onSelect)

redallen: Right, but I only see it used 3 places in the 2 drilldown examples:
```js
this.state = {
      **activeItem: null,
      menuDrilledIn: [],
      drilldownPath: [],
      menuHeights: {},
      activeMenu: 'rootMenu'
    };
 ...
 this.onSelect = (event, itemId) => {
      if (itemId.startsWith('group:')) {
        console.log(`selected sub-menu: ${itemId.split('group:')[1]}`);
      } else {
        console.log(`selected: ${itemId}`);
        this.setState({
          **activeItem: itemId
        });
      }
    };
 ...
 const { **activeItem, menuDrilledIn, drilldownPath, activeMenu, menuHeights } = this.state;
```
Look like it's not used anywhere in the render logic and can be safely removed.

redallen (changes requested):
This is looking much cleaner!! I still think `activeItem` and `onSelect` can be removed. Could the `setHeight`s be moved to a different example or is it a designer necessity? I don't like how the menu changes height and potentially scrolls off the screen.

packages/react-core/src/components/Menu/examples/Menu.md
kmcfaul: Yeah I can remove them

kmcfaul:
@mcarrano @mcoker Any thoughts about the changing height?

I think it may require a core update, I tried removing the height and it appears to make longer lists within the sub-menus truncate and un-scrollable.

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -26,8 +28,24 @@ export interface MenuProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'r
   'aria-label'?: string;
   /** Indicates if menu contains a flyout menu */
   containsFlyout?: boolean;
+  /** Indicates if menu contains a drilldown menu */
+  containsDrilldown?: boolean;
+  /** Indicates if a menu is drilled into */
+  isMenuDrilledIn?: boolean;
+  /** Indicates the drilldown path of menu itemIds */
+  drilldownPath?: string[];
+  /** Indicates the menus that are drilled in */
```
tlabaj, line 37: I think the description is a little confusing.  May just "array of menus that are drilled into".

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -26,8 +28,24 @@ export interface MenuProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'r
   'aria-label'?: string;
   /** Indicates if menu contains a flyout menu */
   containsFlyout?: boolean;
+  /** Indicates if menu contains a drilldown menu */
+  containsDrilldown?: boolean;
+  /** Indicates if a menu is drilled into */
+  isMenuDrilledIn?: boolean;
+  /** Indicates the drilldown path of menu itemIds */
+  drilldownPath?: string[];
```
tlabaj, line 36: Maybe `drilldownPaths`?  The current name implies one path to me but the type is an array.

kmcfaul: My interpretation is that it is a path of nodes. Maybe `drilldownItemPath`?

packages/react-core/src/components/Menu/Menu.tsx
```diff
@@ -58,15 +77,45 @@ class MenuBase extends React.Component<MenuProps, MenuState> {
...
+      onDrillOut,
+      onGetMenuHeight,
+      parentMenu = null,
       activeItemId = null,
       innerRef,
       ...props
     } = this.props;
-
+    const _isMenuDrilledIn =
+      (isMenuDrilledIn && isMenuDrilledIn) || (drilledInMenus && drilledInMenus.includes(id)) || false;
```
jschuler, line 93: This looks a bit funny, should it be `isMenuDrilledIn !== undefined && isMenuDrilledIn || ...`

jschuler (commented):
Thanks for picking this up @kmcfaul ! Since this "variation" adds a bunch of new props you could also consider making it a dedicated component instead and not rolling the props into the Menu component.
For example you could create a new InternalMenu component that has ALL the props, but the public facing DrilldownMenu and Menu components would only surface a few of them.

Anyways, example works great!

packages/react-core/src/components/Menu/Menu.tsx
kmcfaul: That should be equivalent

mcarrano:
> @mcarrano @mcoker Any thoughts about the changing height?

Not sure what to do about this. Would it work to set a max-height after which a scrollbar is exposed? Ideas @mcoker ?

kmcfaul:
Do you think its a blocker or is this something we can open a follow up issue for? @mcarrano

mcarrano:
I think we can address the height issue as follow-up @kmcfaul .

redallen (dismissed):
In that case @mcarrano this LGTM!

packages/react-core/src/components/Menu/DrilldownMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ...props
+}: DrilldownMenuProps) => (
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  <MenuContext.Consumer>
+    {({ menuId, parentMenu, ...context }) => (
+      <Menu
+        id={id}
+        parentMenu={menuId}
+        isMenuDrilledIn={isMenuDrilledIn}
+        {...props}
```
tlabaj, line 32: is there a reason we don't spread the props at the end?

jessiehuff (dismissed):
Currently there's still an [open issue for the keyboard interaction](https://github.com/patternfly/patternfly-react/issues/5144) for this component. I think there are also some refinements that could be made in the future for screen reader users (for example, there's currently nothing announcing to users if a menu item is able to be expanded so it's hard to tell when a screen reader user can continue to drill into the menu). However, I don't think that's specific to the drilldown and might exist with the flyout variations of menu too so probably doesn't need to be addressed in this PR.

tlabaj (approved):
LGTM

mcarrano (dismissed):
Looks good. I will add the keyboard interaction issue to the coming sprint and open a new issue for addressing the height.

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -692,3 +693,459 @@ class MenuOptionMultiSelect extends React.Component {
...
+        onGetMenuHeight={this.setHeight}
+      >
+        <MenuContent menuHeight={`${menuHeights[activeMenu]}px`}>
+          <MenuList>
+            <MenuItem
+              itemId="group:start_rollout"
+              direction="down"
+              drilldownMenu={
+                <DrilldownMenu id="drilldownMenuStart">
+                  <MenuBreadcrumb>
```
mcoker, line 975: In core this is constructed a bit differently. The breadcrumbs aren't a list item adjacent to the other items in the dropdown menu list. `.pf-c-menu__breadcrumb` (`<MenuBreadcrumb>`) is a direct child of `.pf-c-menu` (`<Menu>`). After that is a `<Divider>` (`<hr>`), then `.pf-c-menu__content` (`<MenuContent>`) comes after the divider. Here's a screenshot:

Screen Shot 2021-03-02 at 5 47 53 PM

That said, we can update this if the markup needs to be different. I'm curious what @jessiehuff thinks. At a minimum though, if we keep the markup in this PR and update core to match, `<MenuBreadcrumb>` needs to be an `<li>` since it's a child of `<MenuList>` (a `<ul>`)

cc @mattnolting

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -48,13 +48,18 @@ export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
           <AngleRightIcon />
         </span>
       )}
+      {component === 'button' && (
+        <button className={className} aria-current={ariaCurrent}>
```
mcoker, line 52: this needs `type="button"`

mcoker:
> @mcoker I do have a question about using the link version of BreadcrumbHeading, which is in the core example. Is it ok just to pass a string for the drilldown? An href of "#" will reset the whole drilldown (because the page reloads).

@kmcfaul sorry I just saw this. Could it be a button instead like the other items? While what you have works, the parent item to the text and divider is a flex layout (or it will be once this bug is fixed - https://github.com/patternfly/patternfly-react/issues/5508), and ideally all of the item text needs to be wrapped in an element. I suppose the `component` used for the breadcrumb item link could be a `<span>`, too. 

I created https://github.com/patternfly/patternfly/issues/3896 as a follow up for whatever we implement in this PR.

kmcfaul:
- Added similar button setup to BreadcrumbHeading to match BreadcrumbItem
- Added `type="button"` to both
- Changed `MenuBreadcrumb` from `div` to `li`

tlabaj (dismissed):

packages/react-core/src/components/Menu/examples/Menu.md
mattnolting: The back/up links follow the menus as you drillin and are relative to them, which I believe is the desired effect. 

`<MenuBreadcrumb>` is a child of `pf-c-menu`, rather than `<MenuList>`, so it should be a `<div>`. Otherwise you see a new set of breadcrumbs scroll in with each drilldown, which I don't think is the desired effect. @mcarrano @doruskova wdyt?

mattnolting:
@kmcfaul Sorry, I'm just catching up with this one. This is looking great! `MenuBreadcrumb` isn't a child of a`<ul>` intentionally, so a new set of `Breadcrumb`s isn't scrolled in with each drill down.

mcoker (changes requested):
It looks like the last breadacrumb item is missing the `.pf-c-breadcrumb__link` class, which is causing these border styles.

Screen Shot 2021-03-03 at 3 49 45 PM


Also I'm noticing I either get sent into the wrong menu or the breadcrumbs look incorrect when I'm navigating. Here's a video of going into "start rollouts", then "pause rollouts", then back to "start rollouts" but the breadcrumbs say "pause rollouts"


https://user-images.githubusercontent.com/35148959/109877630-e102c500-7c38-11eb-85b3-6ba8f4655e1a.mp4

kmcfaul:
@mcoker Fixed the incorrect drill out (paths weren't getting updated properly). Also fixed the styling on the BreadcrumbHeading not getting applied

tlabaj:
@kmcfaul if you pushed any changes last night,  I accidentally overwrote them.

mattnolting (commented):
Great work! Here are my observations about `breadcrumb` placement. 

If you think about the relationship of a back button vs breadcrumbs, the back button is relative to the current menu, so it makes sense that that button would scroll with the menu. Breadcrumbs, on the other hand, allow you to traverse all the way back to the root of the tree, so they shouldn’t be relative to individual sub-lists. Also, if the breadcrumbs were to wrap (edge case as they should become a dropdown after several level) it alters the height of the menu because the line wrap throws of the height calculation. If all the animation is isolated to the lists and sublists, the calculation is more accurate/predictable as list items truncate text.

Here's a screenshot of that with a width of 350px, which is a reasonable expected width: 
Screen Shot 2021-03-04 at 2 42 23 PM

kmcfaul:
@tlabaj Updated to resolve the conflict

mcarrano (approved):
This behaves much better with the changes to the breadcrumb. Thanks @kmcfaul @tlabaj !

mattnolting (approved):
It's perfect 🤩!! Great work, thank you!

redallen (approved):

nicolethoen (approved):

PR #5487, opened 2021-02-23
fix(datepicker): only show non-whitespace error text

### LINKED ISSUE #5486: Error styles shown without error text in Datepicker
When passing 2 validators that return empty strings (`validators={[afterValidator, yearValidator]}`), error styles are still shown:
image

redallen:
**What**: Closes #5486

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-core/src/components/DatePicker/DatePicker.tsx
@@ -199,7 +199,7 @@ export const DatePicker: React.FunctionComponent<DatePickerProps> = ({
         </div>
       </Popover>
       {helperText && <div className={styles.datePickerHelperText}>{helperText}</div>}
-      {errorText && <div className={css(styles.datePickerHelperText, styles.modifiers.error)}>{errorText}</div>}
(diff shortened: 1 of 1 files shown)


boaz0 (approved):
LGTM :+1:

packages/react-core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -199,7 +199,7 @@ export const DatePicker: React.FunctionComponent<DatePickerProps> = ({
         </div>
       </Popover>
       {helperText && <div className={styles.datePickerHelperText}>{helperText}</div>}
-      {errorText && <div className={css(styles.datePickerHelperText, styles.modifiers.error)}>{errorText}</div>}
+      {errorText.trim() && <div className={css(styles.datePickerHelperText, styles.modifiers.error)}>{errorText}</div>}
```
jschuler, line 202: errorText can not be null or undefined?

jschuler: OK i looked at where it's coming from, state defaults it to empty string so this is fine

jschuler (approved):

jessiehuff (approved):

dlabrecq (commented):
LGTM

PR #5503, opened 2021-03-01
feat(TreeView): add flag for expand/collapse all

### LINKED ISSUE #4797: Tree view: add capability for an "expand/collapse all" toggle

Currently, the expanded state of a subtree is handled internally in a `useState` hook in `TreeViewListItem.tsx`. We are trying to add a toggle button to expand/collapse everything in the tree, but we can't control that state from outside the component.

I propose a new prop `allExpanded` that is passed down the tree similarly to `defaultAllExpanded`, and if it is defined its value is used instead of the internal expanded state. a `null/undefined` value would allow the internal state to work, a `true` value would force the entire tree to expand, and a `false` value could force the entire tree to collapse. When the user clicks one of the toggles in the tree itself, it could be set to `null` again so that the internal state kicks back in.

Maybe instead of an external prop, that `allExpanded` behavior could be handled internally at the top level, and a button could be shown to the left of the search bar (or in place of it if search is disabled) if a `showExpandAll` boolean is passed. Maybe that would be better, to keep it consistent with the existing internal state.

Observable in the workspace examples.

Enhancement / nice to have.

https://github.com/konveyor/virt-ui (Red Hat Migration Toolkit for Virtualization), no particular release target.

cc @kmcfaul

kmcfaul:
**What**: Closes #4797

- Adds `allExpanded` prop to TreeView, allowing the override of default and internal flags for the expanded state. Set to `null` or do not pass in to allow the `defaultExpanded` prop to set the initial state.
- Adds a button to the default demo to showcase this prop, and updates the search demo to use the new prop (over `defaultExpanded`).

### DIFF
--- a/packages/react-core/src/components/TreeView/examples/TreeView.md
+++ b/packages/react-core/src/components/TreeView/examples/TreeView.md
@@ -14,23 +14,30 @@ import { FolderIcon, FolderOpenIcon, EllipsisVIcon, ClipboardIcon, HamburgerIcon
 
 ```js
 import React from 'react';
-import { TreeView, TreeViewDataItem } from '@patternfly/react-core';
+import { TreeView, TreeViewDataItem, Button } from '@patternfly/react-core';
 
 class DefaultTreeView extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = { activeItems: {} };
+    this.state = { activeItems: {}, allExpanded: null };
 
     this.onClick = (evt, treeViewItem, parentItem) => {
       this.setState({
         activeItems: [treeViewItem, parentItem]
       });
     };
+
+    this.onToggle = evt => {
+      const { allExpanded } = this.state;
+      this.setState({
+        allExpanded: allExpanded !== undefined ? !allExpanded : true
+      });
+    };
   }
 
   render() {
-    const { activeItems } = this.state;
+    const { activeItems, allExpanded } = this.state;
 
     const options = [
       {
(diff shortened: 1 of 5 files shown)


redallen (approved):
Mighty convenient!

packages/react-core/src/components/TreeView/examples/TreeView.md
```diff
@@ -14,23 +14,30 @@ import { FolderIcon, FolderOpenIcon, EllipsisVIcon, ClipboardIcon, HamburgerIcon
...
 import React from 'react';
-import { TreeView, TreeViewDataItem } from '@patternfly/react-core';
+import { TreeView, TreeViewDataItem, Button } from '@patternfly/react-core';
 
 class DefaultTreeView extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = { activeItems: {} };
+    this.state = { activeItems: {}, allExpanded: null };
```
dlabrecq, line 23: Why `null`?

dlabrecq (dismissed):
Can remove this debug message from the default example:

`console.log('All expand?', allExpanded);`

nicolethoen (dismissed):

packages/react-core/src/components/TreeView/examples/TreeView.md
kmcfaul: `null` allows the `defaultExpanded` prop to function properly (if you don't want to force the state for everything before you start filtering/etc). It can be set to `true` or `false` if `defaultExpanded` is not used or it is fine to override it.

tlabaj (changes requested):
Can you please add to the demo-app and add an integration test

tlabaj (approved):

redallen (approved):
Still mighty convenient!

PR #5504, opened 2021-03-01
fix(timepicker): use different am/pm suffixes

### LINKED ISSUE #5401: Fix needed for 12-hour format in Time picker component
12-hour format of time picker is written incorrectly [here](https://www.patternfly.org/v4/components/time-picker) in an interactive dropdown.

Instead of '3:35am' it should be written '3:35 AM' - using a space after the numbers and the capital letters for AM and PM. 
This change has been consulted with Alana from the content team.

redallen:
**What**: Closes #5401

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -9,7 +9,7 @@ import { TimeOption } from './TimeOption';
 import { KeyTypes, SelectDirection } from '../Select';
 import { InputGroup } from '../InputGroup';
 import { TextInput } from '../TextInput';
-import { parseTime, validateTime, makeTimeOptions } from './TimePickerUtils';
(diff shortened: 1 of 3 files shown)


nicolethoen (commented):
So, I think it was an existing bug, but made more obvious by this change.
If I type a time plus a space and then any string that starts with and 'a' or 'p' then focus out of the time picker using the tab or a space then it appends an additional AM or PM after the invalid string. 
Screen Shot 2021-03-02 at 1 55 09 PM

redallen:
@nicolethoen I've made `parseTime` return a consistently formatted time and tightened up `timeRegex` a tad to fix strange inputs. It's whitespace friendly.

redallen:
@nicolethoen I've updated `scrollToSelection`

nicolethoen (approved):

tlabaj (approved):

PR #5507, opened 2021-03-02
remove unused dep

redallen:
**What**: As far as I can tell, charts and table don't need to use `@patternfly/patternfly`. Including it is causing the core workspace not to reflect changes in `/src/patternfly/assets`

**Additional issues**: Towards https://github.com/patternfly/patternfly/issues/3910


dlabrecq (approved):

packages/react-charts/package.json
```diff
@@ -29,7 +29,6 @@
   },
   "homepage": "https://github.com/patternfly/patternfly-react#readme",
   "dependencies": {
-    "@patternfly/patternfly": "4.90.0",
```
dlabrecq, line 32: I don't believe we're using it to build, but it is a run-time dependency.

dlabrecq (changes requested):
Should this be a peer dependency?

redallen:
@dlabrecq `@patternfly/patternfly` is not a runtime dependency unless you choose to use null-loader on `@patternfly/react-styles/css/*` (in which case you could argue its a peer dependency, but `@patternfly/react-*` releases on the same cadence `@patternfly/patternfly` does). I don't think optional peer dependencies belong in dependencies.

The only dependencies for styles for `@patternfly/react-table` and `@patternfly/react-charts` are `@patternfly/react-styles` and `@patternfly/react-tokens`. We can safely remove `@patternfly/patternfly`.

dlabrecq:
Ok. I see that `@patternfly/patternfly` isn't a dependency of `@patternfly/react-core` either.

dlabrecq (approved):

tlabaj (approved):

PR #5509, opened 2021-03-03
[wip] recursive index for DropdownItem

jschuler:
Problem:

If DropdownItem / ApplicationLauncherItem (and other items that extend out from DropdownItem) are wrapped, then keyboard interaction doesn't find them.

Example with no problem:
```
const dropdownItems = [
      <DropdownItem key="link">Link</DropdownItem>,
        <DropdownItem key="action" component="button">
          Action
        </DropdownItem>,
      <DropdownItem key="another_item">
        Another Item
      </DropdownItem>
    ];
    return (
      <Dropdown
        onSelect={this.onSelect}
        toggle={
          <DropdownToggle id="toggle-id" onToggle={this.onToggle} toggleIndicator={CaretDownIcon}>
            Dropdown
          </DropdownToggle>
        }
        isOpen={isOpen}
        dropdownItems={dropdownItems}
      />
    );
```

Example with problem:
**Note how one of the items is wrapped in a fragment**
```
const dropdownItems = [
      <DropdownItem key="link">Link</DropdownItem>,
      <React.Fragment>
        <DropdownItem key="action" component="button">
          In a fragment
        </DropdownItem>
      </React.Fragment>,
      <DropdownItem key="another_item">
        Another Item
      </DropdownItem>
    ];
    return (
      <Dropdown
        onSelect={this.onSelect}
        toggle={
          <DropdownToggle id="toggle-id" onToggle={this.onToggle} toggleIndicator={CaretDownIcon}>
            Dropdown
          </DropdownToggle>
        }
        isOpen={isOpen}
        dropdownItems={dropdownItems}
      />
    );
```

Solution currently would be to either assign `index` manually to the nested DropdownItem (in this case `index={1}`) or do something like in this PR where you recursively assign index to all the children...

### DIFF
--- a/packages/react-core/src/components/Dropdown/DropdownMenu.tsx
+++ b/packages/react-core/src/components/Dropdown/DropdownMenu.tsx
@@ -152,11 +152,25 @@ export class DropdownMenu extends React.Component<DropdownMenuProps> {
         return React.cloneElement(group, props);
       });
     }
-    return React.Children.map(children, (child, index) =>
-      React.cloneElement(child as React.ReactElement, {
-        index
-      })
-    );
+    return this.recursiveMap(children);
+  }
+
+  recursiveMap(children: any) {
+    return React.Children.map(children, (child, index) => {
+      if (!React.isValidElement(child)) {
+        return child;
+      }
+
+      if ((child.props as any).children) {
+        const obj = {
+          children: this.recursiveMap((child.props as any).children),
+          index
+        };
+        child = React.cloneElement(child, obj);
+      }
+
+      return child;
+    });
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


jschuler:
@jessiehuff 
@kmcfaul

packages/react-core/src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -152,11 +152,25 @@ export class DropdownMenu extends React.Component<DropdownMenuProps> {
...
+  recursiveMap(children: any, index: number) {
+    return React.Children.map(children, child => {
+      if (!React.isValidElement(child)) {
+        return child;
+      }
+
+      if ((child.props as any).children) {
+        const obj = {
+          children: this.recursiveMap((child.props as any).children, index),
+          index
```
TheRealJon, line 167: Should we do something like:

```suggestion
          index: child.props.index ?? index
```

So that we don't overwrite an existing index prop?

TheRealJon: Also, if downstream user uses their own `index` prop on a Dropdown item or any of its children, would this cause keyboard nav to misbehave?

PR #5513, opened 2021-03-03
fix(Radio, Checkbox): Added standalone modifier when no label is prov…

### LINKED ISSUE #5435: Add standalone variation of checkbox and radio components
https://github.com/patternfly/patternfly/pull/3821/

Adds `.pf-m-standalone` as a variation for the checkbox and radio components that remove the requirement for a label. As a side note, the input should have an `aria-label` or `aria-labelledby` since there is no longer a label.

Also updates the checkboxes in these places to use the standalone checkbox

* dual list selector
* tree view
* card view demo
* and anywhere else we use a regular `<input type="checkbox/radio">` that is a beta component or example/demo that it's OK for us to change.

tlabaj:
…ided

**What**: Closes #5435

### DIFF
--- a/packages/react-core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/react-core/src/components/Checkbox/Checkbox.tsx
@@ -78,7 +78,7 @@ export class Checkbox extends React.Component<CheckboxProps> {
 
     checkedProps.checked = checkedProps.checked === null ? false : checkedProps.checked;
     return (
-      <div className={css(styles.check, className)}>
+      <div className={css(styles.check, !label && styles.modifiers.standalone, className)}>
         <input
           {...props}
           className={css(styles.checkInput)}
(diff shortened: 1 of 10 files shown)


mcoker (approved):
LGTM!

jessiehuff (commented):
This looks good to me! I'm wondering though if this use case is a good candidate for providing a console warning when a radio is standalone and the user doesn't provide an `aria-label` or `aria-labelledby` like we did with the progress component? This use case seems at a risk for being used in an inaccessible way since it relies on users to add accessibility.

tlabaj:
@jessiehuff there is already a warning.  React already supported not adding a label.

jessiehuff (approved):
Ah you're right, I didn't see that. 🙂  Ok, awesome!

nicolethoen (approved):

PR #5518, opened 2021-03-03
fix(searchinput): add searchInputMenuBody class

### LINKED ISSUE #5516: Search input - add body element for expanded menu contents
This is part of https://github.com/patternfly/patternfly/pull/3892/ that needs to be updated in react at the same time as core.

It just adds a new `div.pf-c-search-input__menu-body` that is used to wrap contents of the search input menu when you want padding in the menu. So the advanced attribute search form should be wrapped in this new element.

redallen:
**What**: Closes #5516

**Additional issues**:


jessiehuff (approved):

nicolethoen:
@mcoker was something changed in the css so that the `x` clear button no longer has an bottom border? It look strange in the advanced search example here.

mcoker:
@nicolethoen yep, good catch! That's a separate bug and I have a PR here to fix - https://github.com/patternfly/patternfly/pull/3905

mcoker (approved):
LGTM! Thanks for doing this one so quickly.

nicolethoen (approved):

PR #5522, opened 2021-03-04
fix(Switch): fix error message regarding accessibility usage

### LINKED ISSUE #5521: Switch: accessibilty properties check is wrong, id should not be required the label should be instead
There is currently a console error:
Switch: Switch requires either an id or aria-label to be specified

The id is autogenerated when no provided. What we need to check here is that there is a label which describes the switch.

KKoukiou:
When the user has not provided an 'id', `getUniqueId` is used to
generate a random one. That id is used after for the aria-labelled-by
property. Therefore when ensuring correct usage of accessibility
properties we need to check that a label is provided, not an id.

Closes #5521


jessiehuff (approved):
LGTM, nice catch! 🙂

tlabaj (approved):

PR #5524, opened 2021-03-05
feat(Slider): add isDisabled property

### LINKED ISSUE #5496: feat(Slider): add isDisabled property
Add the ability to disable the Slider using the styles defined in Core. https://github.com/patternfly/patternfly/pull/3879

cc @KKoukiou

KKoukiou:
Closes #5496

### DIFF
--- a/packages/react-core/src/components/Slider/Slider.tsx
+++ b/packages/react-core/src/components/Slider/Slider.tsx
@@ -22,7 +22,7 @@ export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onCh
   currentValue?: number;
   /** Flag indicating if the slider is is discrete */
   isDiscrete?: boolean;
-  /** Adds disabled styling and disables the slider */
+  /** Adds disabled styling and disables the slider and the input component is present */
   isDisabled?: boolean;
   /** Array of slider step objects (value and label of each step) for the slider. */
   steps?: SliderStepObject[];
@@ -38,8 +38,6 @@ export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onCh
   inputLabel?: string | number;
   /** Position of the input */
   inputPosition?: 'aboveThumb' | 'right';
-  /** Flag indicating input is disabled */
-  isInputDisabled?: boolean;
   /** Value input callback.  Called when enter is hit while in input filed or focus shifts from input field */
   onChange?: (value: number) => void;
   /** Value change callback. This is called when the slider value changes */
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -316,10 +316,11 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
           aria-valuenow={value}
           aria-valuetext={findAriaTextValue()}
           aria-label={thumbAriaLabel}
-          onMouseDown={handleMouseDown}
-          onTouchStart={handleToucheStart}
-          onKeyDown={handleThumbKeys}
-          onClick={onThumbClick}
+          aria-disabled={isDisabled}
```
mcoker, line 319: Also need to modify the `tabIndex` above to `-1` (or remove it) when it's disabled so it won't take keyboard focus.

mcoker (changes requested):
Looks great! Just one comment about the `tabindex` on the thumb

mcarrano (approved):
Looks good to me. Thanks @KKoukiou .

packages/react-core/src/components/Slider/Slider.tsx
KKoukiou: Thanks for catching this! Updated with fix.

nicolethoen:
I think this looks great! @jessiehuff I wonder if we don't actually want to change the tabindex, since now the slider is completely undiscoverable to the screen reader? Is there a different way to make sure the screen reader knows its there?

mcoker:
FWIW this models tabindex being removed from other disabled form elements like button, text input, switch, etc - also a [disabled HTML range input/slider](https://codepen.io/mcoker/pen/WNoWbJe). Since it's `aria-disabled`, but not hidden/`aria-hidden`, it should still be discoverable and dimmed by screen readers the way other disabled elements are.

jessiehuff (commented):
So I double checked this in VO, JAWS, and NVDA to make sure it worked, and it seems that all of them recognized the disabled slider. VO calls it "dimmed", JAWS and NVDA say "unavailable." There are times that we've managed focus when something interferes with the screen reader seeing it (for example, disabled dropdown items), but otherwise it seems to handle it ok. The only thing I noticed that was confusing in testing was there was no `aria-valuetext` on the disabled example. It was instead announced as 50 which may not be what is intended if a sighted user is seeing 4. I would look into why that isn't getting added with this example.

mcoker (approved):
LGTM! 🥳

tlabaj (approved):

Issue #5291, opened 2020-12-18
Resizable drawer


evwilkin:
OpenShift 4.8 is looking to adopt the resizable drawer in 3 Epics for ODC (Dev Console: Quick Starts, Web Terminal, & Topology side panel) but has identified the included issues to be resolved:

https://docs.google.com/document/d/1BlVATH7MWtHzWLOMWl6YmUy4ZrOqOm2AwLG5Ilm_rR8/edit

evwilkin:
@mcarrano @kmcfaul FYI set up this Epic to track the Resizable drawer issues 👍

tlabaj:
closing this since all issues in the epic have been addressed

PR #5533, opened 2021-03-10
feat(topology): Use Drawer for topology side panel and allow resizing

jeff-phillips-18:
**What**: Addresses portions of #5291

### DIFF
--- a/packages/react-integration/cypress/integration/codeeditor.spec.ts
+++ b/packages/react-integration/cypress/integration/codeeditor.spec.ts
@@ -10,7 +10,7 @@ describe('Context Selector Demo Test', () => {
       .first()
       .then(code => expect(code).to.have.text('test'));
     cy.get('#reset-code').click();
-    cy.wait(100);
+    cy.wait(1000);
     cy.get('.view-lines')
       .first()
       .then(code => expect(code).to.have.text('resetcode'));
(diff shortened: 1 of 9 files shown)


redallen (approved):
Nice drawers 👖

packages/react-integration/cypress/integration/codeeditor.spec.ts
```diff
@@ -10,7 +10,7 @@ describe('Context Selector Demo Test', () => {
       .first()
       .then(code => expect(code).to.have.text('test'));
     cy.get('#reset-code').click();
-    cy.wait(100);
+    cy.wait(1000);
```
KKoukiou, line 13: Why was this necessary timeout increase? Just out of interest!

KKoukiou (dismissed):
Nice cleanup!

KKoukiou:
Just for the record:

Before: 
Screen Shot 2021-03-11 at 09 44 03


After:
Screen Shot 2021-03-11 at 09 41 13

packages/react-integration/cypress/integration/codeeditor.spec.ts
jeff-phillips-18: Tests were failing. This was recently updated to `100` to fix the flakes, but apparently it wasn't enough.

jeff-phillips-18:
@KKoukiou Added a pass thru `onSideBarResize` so applications can track the user desired side bar size for topology.

KKoukiou (commented):
In patternfly we rarely encourage the user to pass size/with/height related properties to components. We usually have some preselected sizes, (sm, md, lg) and let the user choose one of them. How do you feel about having preselected sizes here as well?

jeff-phillips-18:
> In patternfly we rarely encourage the user to pass size/with/height related properties to components. We usually have some preselected sizes, (sm, md, lg) and let the user choose one of them. How do you feel about having preselected sizes here as well?

I'm following the `Drawer` parameters here. I think being flexible here is necessary.

redallen (approved):

nicolethoen (approved):
Ran it, looks good to me! 
Interesting about needing to up the wait on the code editor test - I feel like the code editor plug in just isn't playing well with the cypress tests... hopefully this fixes the flakiness.

Issue #5537, opened 2021-03-11
Topology control bar buttons don't look disabled when the are


jeff-phillips-18:
**Description**
Setting the control bar buttons to disabled does not visually change the background color to indicate that it is disabled.
 
**Steps to reproduce**
Set the state of a control button to disabled 

**Screen Shot**
Each of the buttons are disabled:
![image](https://user-images.githubusercontent.com/11633780/110857439-da063300-8286-11eb-9c67-7d487296e01c.png)

PR #5538, opened 2021-03-11
fix(topology): Use disabled background for control buttons when disabled

jeff-phillips-18:
**What**:
closes https://github.com/patternfly/patternfly-react/issues/5537

### DIFF
--- a/packages/react-styles/src/css/components/Topology/topology-controlbar.css
+++ b/packages/react-styles/src/css/components/Topology/topology-controlbar.css
@@ -6,10 +6,12 @@
 .pf-topology-control-bar__button.pf-c-button.pf-m-tertiary {
   margin-right: var(--pf-global--spacer--xs);
   margin-top: var(--pf-global--spacer--xs);
-  background-color: var(--pf-global--BackgroundColor--100);
   border: none;
   border-radius: var(--pf-global--BorderRadius--sm);
   box-shadow: var(--pf-global--BoxShadow--sm);
+}
+.pf-topology-control-bar__button.pf-c-button.pf-m-tertiary:not(.pf-m-disabled) {
+  background-color: var(--pf-global--BackgroundColor--100);
 }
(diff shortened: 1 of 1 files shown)


packages/react-styles/src/css/components/Topology/topology-controlbar.css
```diff
@@ -11,6 +11,9 @@
   border-radius: var(--pf-global--BorderRadius--sm);
   box-shadow: var(--pf-global--BoxShadow--sm);
 }
+.pf-topology-control-bar__button.pf-c-button.pf-m-tertiary.pf-m-disabled {
```
KKoukiou, line 14: Shouldn't the Button CSS rules be included when you use the `<button>` component from the react code? 

ITW there is a pf-c-button.pf-m-disabled rules in the button.css which theoretically is imported automatically when you use the button.

jeff-phillips-18: The rule above for `pf-topology-control-bar__button.pf-c-button.pf-m-tertiary` overrides the `. pf-c-button.pf-m-disabled` rule (more specificity). Added this rule to override that rule.

KKoukiou: Oh got it.. Ideally I would nest the `background-color` rule above, in a :not(.pf-m-disabled) block. In this way you just allow the pf-m-disabled to be applied, instead of overwriting *2* times. 

Anyway, I will approve, and leave it up to you to decide.

KKoukiou (dismissed):

packages/react-styles/src/css/components/Topology/topology-controlbar.css
jeff-phillips-18: Agreed, and updated. Thanks!

KKoukiou (approved):

redallen (approved):

PR #5559, opened 2021-03-22
feat(Checkbox, Radio): Added support for custom body content

### LINKED ISSUE #5519: Radio/Check - add support for custom body content
https://github.com/patternfly/patternfly/pull/3884

Adds a new element `div.pf-c-check__body` that comes after the description that can accept whatever a user wants to put in it.

tlabaj:
**What**: Closes #5519 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/react-core/src/components/Checkbox/Checkbox.tsx
@@ -98,7 +98,7 @@ export class Checkbox extends React.Component<CheckboxProps> {
             {label}
           </label>
         )}
-        {description && <div className={css(styles.checkDescription)}>{description}</div>}
+        {description && <span className={css(styles.checkDescription)}>{description}</span>}
         {body && <span className={css(styles.checkBody)}>{body}</span>}
       </div>
     );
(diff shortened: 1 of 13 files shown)


redallen (dismissed):

mcarrano (dismissed):
Looks good to me. Thanks @tlabaj .

mcoker (changes requested):
This looks great! Thanks @tlabaj 

One small issue I see is the use of the description and body when using the `isLabelWrapped` prop. That makes the `pf-c-check/radio` parent container a `<label>` element, and the [permitted content of a label element is phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#technical_summary), so the description and body elements should be a `<span>` instead of a `<div>` when they're the child of a `<label>` element.

That said, we can just update the default element for the description/body to be a `<span>` so that the HTML is valid and we don't need to change the element depending on whether it's a child of a `<label>` or not.

Can you change those elements to `<span>`s? I've put up a core PR here - https://github.com/patternfly/patternfly/pull/3945

jessiehuff (approved):
LGTM!

mcoker (changes requested):

nicolethoen (dismissed):

kmcfaul (dismissed):

kmcfaul:
lgtm just need to update tests

mcoker (dismissed):
LGTM!! 👍

jessiehuff (approved):

mcoker (approved):

dlabrecq (approved):

jessiehuff (approved):

PR #5566, opened 2021-03-23
feat(alert): don't timeout if isMouseOver

### LINKED ISSUE #5489: do not hide toast alert when mouse is over
Alerts used as a toast notification can have a timeout which will hide the toast after the timeout expires. This request is to halt the timeout while the user's mouse is over the toast. If the user is trying to read the message or copy something from the toast, it shouldn't disappear out from under their mouse.

Enhancement

OpenShift Console 4.8

redallen:
**What**: Wait until the Alert is timed out AND the mouse is not over it AND the keyboard isn't focusing an element inside of it. After the mouse exits wait at least `timeoutAnimation` seconds. Closes #5489

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -40,6 +40,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   isLiveRegion?: boolean;
   /** If set to true, the timeout is 8000 milliseconds. If a number is provided, alert will be dismissed after that amount of time in milliseconds. */
   timeout?: number | boolean;
+  /** If the user hovers over the Alert and `timeout` expires, this is how long to wait before finally dismissing the Alert */
+  timeoutAnimation?: number;
   /** Function to be executed on alert timeout. Relevant when the timeout prop is set */
   onTimeout?: () => void;
   /** Truncate title to number of lines */
@@ -64,6 +66,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   ouiaId,
   ouiaSafe = true,
   timeout = false,
+  timeoutAnimation = 3000,
   onTimeout = () => {},
   truncateTitle = 0,
   tooltipPosition,
@@ -80,8 +83,11 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
     </React.Fragment>
   );
 
+  const divRef = React.useRef<HTMLDivElement>();
   const [timedOut, setTimedOut] = useState(false);
+  const [timedOutAnimation, setTimedOutAnimation] = useState(true);
   const [isMouseOver, setIsMouseOver] = useState(false);
+  const [containsFocus, setContainsFocus] = useState(false);
   const [isTooltipVisible, setIsTooltipVisible] = useState(false);
   const titleRef = React.useRef(null);
   React.useEffect(() => {
@@ -97,9 +103,29 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   React.useEffect(() => {
     timeout = timeout === true ? 8000 : Number(timeout);
     if (timeout > 0) {
-      setTimeout(() => setTimedOut(true), timeout);
+      const timer = setTimeout(() => setTimedOut(true), timeout);
+      return () => clearTimeout(timer);
     }
   }, []);
+  React.useEffect(() => {
+    const onDocumentFocus = () => {
+      if (divRef.current) {
+        if (divRef.current.contains(document.activeElement)) {
+          setContainsFocus(true);
+          setTimedOutAnimation(false);
+        } else if (containsFocus) {
+          setContainsFocus(false);
+          setTimeout(() => setTimedOutAnimation(true), timeoutAnimation);
+        }
+      }
+    };
+
+    document.addEventListener('focus', onDocumentFocus, true);
+
+    return () => document.removeEventListener('focus', onDocumentFocus, true);
+  }, [containsFocus]);
+  const dismissed = timedOut && timedOutAnimation && !isMouseOver && !containsFocus;
+  React.useEffect(onTimeout, [dismissed]);
   const Title = (
     <h4
       {...(isTooltipVisible && { tabIndex: 0 })}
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -91,12 +94,12 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
-  const customClassName = css(
-    styles.alert,
-    isInline && styles.modifiers.inline,
-    variant !== AlertVariant.default && styles.modifiers[variant as 'success' | 'danger' | 'warning' | 'info'],
-    className
-  );
+  React.useEffect(() => {
+    timeout = timeout === true ? 8000 : Number(timeout);
+    if (timeout > 0) {
+      setTimeout(() => setTimedOut(true), timeout);
```
christianvogt, line 100: You need to clear the timeout on unmount otherwise you'll end up updating state on an unmounted component if the component is unmounted before the timeout completes.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -107,49 +110,54 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
-      </div>
-    );
-  } else {
+  const myOnMouseLeave = (ev: React.MouseEvent<HTMLDivElement>) => {
+    setIsMouseOver(false);
+    onMouseLeave(ev);
+  };
+
+  if (timedOut && !isMouseOver) {
+    onTimeout();
```
christianvogt, line 124: Do not execute side effects in `render`. Use an effect.

christianvogt (commented):
I'd prefer to see an approach where if the user interacts with the alert, the current timeout is first cleared. When the user mouses away, a timeout resumes but at a much lower time. Perhaps `Math.max(timeout / 2, 3)` or even just a small fixed time. The reason for this is to not immediately hide the alert due to a mouse slip.

Also, we shouldn't hide the alert if the user focuses on an action via keyboard. You'll need to track focus in addition to mouse events.

redallen:
I discussed this with @mcarrano and we decided on a small 3s timeout to prevent mouseslips. I addressed all your feedback, the keyboard focus was rather tricky.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -91,12 +100,32 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
+        }
+      }
+    };
+
+    document.addEventListener('focus', onDocumentFocus, true);
+
+    return () => document.removeEventListener('focus', onDocumentFocus, true);
+  }, [containsFocus]);
+  const dismissed = timedOut && timedOutAnimation && !isMouseOver && !containsFocus;
+  React.useEffect(onTimeout, [dismissed]);
```
christianvogt, line 128: You should only call `onTimeout` when `dismissed` is `true` otherwise you notify on first render.
```suggestion
  React.useEffect(() => dismissed && onTimeout(), [dismissed]);
```

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -107,49 +136,57 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
-    );
-  } else {
+  const myOnMouseLeave = (ev: React.MouseEvent<HTMLDivElement>) => {
+    setIsMouseOver(false);
+    setTimeout(() => setTimedOutAnimation(true), timeoutAnimation);
+    onMouseLeave(ev);
+  };
+
+  if (dismissed) {
+    onTimeout();
```
christianvogt, line 152: You can remove this now that it's done in an effect.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -107,49 +136,57 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
-          </AlertContext.Provider>
-        )}
-        {children && <div className={css(styles.alertDescription)}>{children}</div>}
-        {actionLinks && <div className={css(styles.alertActionGroup)}>{actionLinks}</div>}
-      </div>
-    );
-  } else {
+  const myOnMouseLeave = (ev: React.MouseEvent<HTMLDivElement>) => {
+    setIsMouseOver(false);
+    setTimeout(() => setTimedOutAnimation(true), timeoutAnimation);
```
christianvogt, line 147: This timer needs to be cleared on mouse enter and restarted on mouse leave.
You also need to cancel the timer on component unmount.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -91,12 +100,32 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
+  }, []);
+  React.useEffect(() => {
+    const onDocumentFocus = () => {
+      if (divRef.current) {
+        if (divRef.current.contains(document.activeElement)) {
+          setContainsFocus(true);
+          setTimedOutAnimation(false);
+        } else if (containsFocus) {
+          setContainsFocus(false);
+          setTimeout(() => setTimedOutAnimation(true), timeoutAnimation);
```
christianvogt, line 118: Cancel timer on unmount.

packages/react-core/src/components/Alert/Alert.tsx
redallen: > restarted on mouse leave
Not sure how to do this part. I don't think it needs to be restarted on mouse leave since `const dismissed = timedOut && timedOutAnimation && !isMouseOver && !containsFocus;`

christianvogt: looks like you changed the implementation here, so this won't be needed now

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -91,12 +96,57 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
...
-    styles.alert,
-    isInline && styles.modifiers.inline,
-    variant !== AlertVariant.default && styles.modifiers[variant as 'success' | 'danger' | 'warning' | 'info'],
-    className
-  );
+
+  const [timedOut, setTimedOut] = useState(false);
+  const [timedOutAnimation, setTimedOutAnimation] = useState(true);
+  const [isMouseOver, setIsMouseOver] = useState<boolean | undefined>();
+  const [containsFocus, setContainsFocus] = useState<boolean | undefined>();
```
christianvogt, line 103:
```suggestion
  const [isMouseOver, setIsMouseOver] = useState<boolean>(false);
  const [containsFocus, setContainsFocus] = useState<boolean>(false);
```

christianvogt:
tested it out and it seems to work as expected now

packages/react-core/src/components/Alert/Alert.tsx
redallen: This won't work because then this `useEffect`'s condition will needlessly happen on first render:
```
  React.useEffect(() => {
    if (containsFocus === false || isMouseOver === false) {
      const timer = setTimeout(() => setTimedOutAnimation(true), timeoutAnimation);
      return () => clearTimeout(timer);
    }
  }, [containsFocus, isMouseOver]);
```

christianvogt: tricky tri-state booleans!

nicolethoen (approved):

mcarrano (approved):
Looks good to me! Thanks @redallen .

tlabaj (approved):
lgtm

PR #5568, opened 2021-03-24
feat(textarea): add auto resizing

### LINKED ISSUE #5497: Add option to auto-size a text area
Want to enable optional auto-sizing of a text-area according to the following behavior/requirements:
*   starts at the default size
*   would be useful if this was configurable (ie rows or min-height?)
*  when the user reaches the end of the height of the area, the next 'enter' grows the box to support the new line
*   when the user removes a line, it shrinks back down (until it hits the "default" size)

https://github.com/patternfly/patternfly/pull/3883

redallen:
**What**: Closes #5497 using https://github.com/patternfly/patternfly/pull/3883 and https://gomakethings.com/automatically-expand-a-textarea-as-the-user-types-using-vanilla-javascript/

**Additional issues**: https://patternfly-react-pr-5568.surge.sh/components/text-area#auto-resizing

### DIFF
--- a/packages/react-core/src/components/TextArea/TextArea.tsx
+++ b/packages/react-core/src/components/TextArea/TextArea.tsx
@@ -19,6 +19,8 @@ export interface TextAreaProps extends Omit<HTMLProps<HTMLTextAreaElement>, 'onC
   isDisabled?: boolean;
   /** Flag to show if the TextArea is read only. */
   isReadOnly?: boolean;
+  /** Flag to modify height based on contents. */
+  autoResize?: boolean;
   /** Value to indicate if the textarea is modified to show that validation state.
    * If set to success, textarea will be modified to indicate valid state.
    * If set to error, textarea will be modified to indicate error state.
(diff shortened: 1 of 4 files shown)


nicolethoen:
I am not seeing any auto-resizing when I test it...
Mar-24-2021 13-23-48

redallen:
@nicolethoen if you touch the borrom-right resizer it will stay whatever size you adjust it to. If you just type in the input you should see it autoresize, though:
autoresize

nicolethoen (approved):
ah yes

tlabaj (changes requested):
looks good so far.  Can you add to demo-app and cypress test.

mcarrano (approved):
This looks great @redallen .

@andrewballantyne Does this behave as you expected?

andrewballantyne:
> This looks great @redallen .
> 
> @andrewballantyne Does this behave as you expected?

Yup! That performs just as expected. We use `rows={1}` prop as well in our use-case and I just tested that out here and it works well. Thanks for the support of the prop!

tlabaj (approved):
Looks great!

PR #5569, opened 2021-03-24
feat(Form): Added support for field groups

### LINKED ISSUE #5023: Add expandable field groups
Want to add the ability to include expandable, nested field groups to the Form component per the designs here: https://marvelapp.com/prototype/5e57if3/screen/72963176

Core issue: https://github.com/patternfly/patternfly/issues/3557

Reference current Expandable section component for keyboard interactions.

Requesting projects: OpenShift, App Services
Date needed:
Contacts: @itsptk @lboehling

tlabaj:
**What**: Closes #5023 

**Additional issues**:


jessiehuff (commented):
Looking good so far! 🙂  The only thing I noticed is that I think you're trying to include the toggle in the `aria-labelledby`, but the id you used to associate it is on the outer div (the `id="form-field-group-toggle140"`). If you want the "Details" `aria-label` on the button to be announced, the id should be on the button and then use that id in the `aria-labelledby`. So the `aria-labelledby` would still look like this: `aria-labelledby="field-group1-titleText-id form-field-group-toggle140"`, but the id would be on the button where the action is happening and that "Details" `aria-label` is.

nicolethoen (approved):

mcarrano (approved):
Looks great @tlabaj !

jessiehuff (approved):

packages/react-core/src/components/Form/FormFieldGroupToggle.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <div className={css(styles.formFieldGroupToggleButton)}>
+      <Button
+        variant="plain"
+        aria-label={ariaLabel}
+        onClick={onToggle}
+        aria-expanded={isExpanded}
+        aria-labelledby={ariaLabelledby}
+        id={toggleId}
+      >
+        {isExpanded ? <AngleDownIcon aria-hidden="true" /> : <AngleRightIcon aria-hidden="false" />}
```
mcoker, line 42: If you wrap this in `span.pf-c-form__field-group-toggle-icon`, then you only need to include the right angle, and the CSS will rotate it with the expanded class on the field group parent element. That will also enable the transition animation of the icon rotating when toggling the expanded state.

Also why is `aria-hidden="false"` with one icon and not the other?

mcoker (changes requested):
Looks great @tlabaj! Just one comment about the toggle icon.

mcoker (dismissed):

packages/react-core/src/components/Form/examples/Form.md
```diff
@@ -770,3 +781,433 @@ class SimpleForm extends React.Component {
...
+            isRequired
+            id="form-expandable-field-groups-label1"
+            name="form-expandable-field-groups-label1"
+            value={value1}
+            onChange={value => {
+              this.handleTextInputChange({ value1: value });
+            }}
+          />
+        </FormGroup>
+        <FormGroup label="Label 2" isRequired fieldId="form-expandable-field-groups-label2 ">
```
jessiehuff, line 863: Here's one of the spaces causing one of the errors. On the `fieldId`.

packages/react-core/src/components/Form/examples/Form.md
```diff
@@ -770,3 +781,433 @@ class SimpleForm extends React.Component {
...
+            isRequired
+            id="form-expandable-field-groups-label3"
+            name="form-expandable-field-groups-label3"
+            value={value21}
+            onChange={value => {
+              this.handleTextInputChange({ value21: value });
+            }}
+          />
+        </FormGroup>
+        <FormGroup label="Label 4" isRequired fieldId="form-expandable-field-groups-label4 ">
```
jessiehuff, line 1198: I think this is the other space issue, on the `fieldId` again.

packages/react-core/src/components/Form/examples/Form.md
```diff
@@ -770,3 +781,433 @@ class SimpleForm extends React.Component {
...
+                    <TrashIcon />
+                  </Button>
+                }
+              />
+            }
+          >
+            <FormGroup label="Label 1" isRequired fieldId="form-expandable-field-groups-field-group1-label1">
+              <TextInput
+                isRequired
+                id="form-expandable-field-groups-field-group1-label"
```
jessiehuff, line 907: This is where one of the issues is coming from. See how the id and name don't match up with the fieldId for the label 1? You can see how the FormGroup below it matches. I think if you get these matching (maybe "form-expandable-field-groups-field-group1-label1), it'll fix that issue.

jessiehuff (commented):
Just a few updates to get the a11y tests working and fix the screen reader issues. 🙂

tlabaj:
Thanks Jessie!

jessiehuff (approved):
LGTM!

kmcfaul:
`FormFieldGroup` props should match `InternalFormFieldGroup` props so our docs show them (since the internal component isn't exported/picked up by our docs). Unless those extra props are all just for internal use?

kmcfaul (approved):

mattnolting:
Looks good! Side note, all the `input` s in https://patternfly-react-pr-5569.surge.sh/components/form#basic have the same ID.

dlabrecq (approved):

mattnolting (approved):
👍 👍

PR #5570, opened 2021-03-24
feat(Menu): keyboard interaction first pass

### LINKED ISSUE #5144: Menu component: Add keyboard interaction

kmcfaul:
**What**: Closes #5144

First pass for keyboard interaction with the Menu component. May need further revision based on feedback/other edge cases I may have missed.

Generally speaking:
- `ArrowUp` and `ArrowDown` navigate menu items
- Disabled items are skipped
- `ArrowLeft` and `ArrowRight` navigate items within a menu (for actions and breadcrumbs)
- `Enter` and `Space` select menu items and control drill in/drill out for drilldown menus and drilldown breadcrumbs

For filterable menus, the filter box and menu are swapped between via using `Tab`.
For flyout menus, the flyout is opened from the flyout item using `Enter`, `Space` or `ArrowRight` then traversed normally with arrow keys. Flyout menus are closed with `Escape` or `ArrowLeft`.

Re: breadcrumbs @jessiehuff. I missed the blurb about using Tab to go between menu/breadcrumb nav so right now it uses the arrow keys. I can probably change this back to Tab if it should be that way instead. Let me know!

### DIFF
--- a/packages/react-core/src/components/Dropdown/Toggle.tsx
+++ b/packages/react-core/src/components/Dropdown/Toggle.tsx
@@ -103,13 +103,19 @@ export class Toggle extends React.Component<ToggleProps> {
     if (event.key === 'Tab' && !this.props.isOpen) {
       return;
     }
-    if (!this.props.bubbleEvent) {
-      event.stopPropagation();
-    }
-    event.preventDefault();
     if ((event.key === 'Tab' || event.key === 'Enter' || event.key === ' ') && this.props.isOpen) {
+      if (!this.props.bubbleEvent) {
+        event.stopPropagation();
+      }
+      event.preventDefault();
+
       this.props.onToggle(!this.props.isOpen, event);
     } else if ((event.key === 'Enter' || event.key === ' ' || event.key === 'ArrowDown') && !this.props.isOpen) {
+      if (!this.props.bubbleEvent) {
+        event.stopPropagation();
+      }
+      event.preventDefault();
+
       this.props.onToggle(!this.props.isOpen, event);
       this.props.onEnter();
     }
--- a/packages/react-core/src/components/Menu/DrilldownMenu.tsx
+++ b/packages/react-core/src/components/Menu/DrilldownMenu.tsx
@@ -29,6 +29,7 @@ export const DrilldownMenu: React.FunctionComponent<DrilldownMenuProps> = ({
         id={id}
         parentMenu={menuId}
         isMenuDrilledIn={isMenuDrilledIn}
+        isRootMenu={false}
         ref={React.createRef()}
         {...context}
         {...props}
--- a/packages/react-core/src/components/Menu/Menu.tsx
+++ b/packages/react-core/src/components/Menu/Menu.tsx
@@ -158,8 +158,28 @@ class MenuBase extends React.Component<MenuProps, MenuState> {
       activeElement.classList.contains('pf-c-breadcrumb__link') ||
       activeElement.classList.contains('pf-c-dropdown__toggle');
 
-    if (key === 'Tab' && !event.shiftKey) {
+    if (!isFromBreadcrumb && key === 'Tab' && !event.shiftKey) {
       event.preventDefault();
+    }
+
+    if (key === 'Tab') {
+      if (isFromBreadcrumb) {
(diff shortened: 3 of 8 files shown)


jessiehuff (commented):
This is looking great so far! 🙂 

A few thoughts though: I noticed that Tab still navigates through all the items as well as the arrow keys which I found unexpected. Typically in other menu-like components (dropdown, app launcher, etc.) Tab will close the menu so I don't know if it makes sense to navigate with Tab here. However, it does remind me that because of how we're handling this component, it's impossible to account for how users will incorporate this into their products so I wonder if a warning on keyboard a11y should be on the page. (For example, I would expect that when I'm inside of a menu that I can press the Esc key and leave the menu. I'm not sure if that's possible here because of how it might be used, right?) If we can add these typical expectations, I think we should, but if we can't we should note that so that users know that they need to do that when using this component.

I was concerned at first about the disabled items not receiving focus (because we saw that issue with dropdown causing issues for a JAWS user), but I tested this on VO, NVDA, and JAWS, and they all seem to recognize the disabled items so I think that part is ok.

For the breadcrumb example, I do think Tab makes more sense to navigate through the breadcrumb items and arrow keys for the menu. Conceptually Tab is usually used to navigate between UI elements and arrow key are used within UI elements. When Jenn and I talked about this one, we saw the breadcrumb on top of the menu and the menu itself as distinct from each other and while they're still in the same component, they seem more like different UI elements. I also found that example a little bizarre to try to navigate. 
When I just tried to use arrow keys, it never shifted focus into the menu but did scroll the container. When I tried to use Tab, focus disappeared for a few key presses but then eventually showed up but still ends up shifting the container and focus disappears again for a few key presses before leaving the menu. I can send a screen recording but for some reason every time I try to add it to this comment, it turns into a jpg haha. 


Some other things I noticed that are unrelated to this PR: I think the flyout examples should have some label on the toggle to signify that they're different than other menu items. Currently they're all announced as buttons so there's no way for a screen reader user to know that they fly out. Also, the filtering with text input example has a weird `aria-label` on the input. It's `"filterable-example-with-text-input"` which doesn't make much sense when hearing it in the screen reader.

nicolethoen (dismissed):
It definitely will take some practice to learn all these key controls ;) but I did not find any broken edge cases

kmcfaul:
@jessiehuff Which menu were you trying to navigate where you were getting the scroll? It shouldn't be doing that (because the arrow events prevent the default behavior).

kmcfaul:
For now I removed default `Tab` behavior. You can still Tab into and Shift+Tab out of a menu. I also updated the unhelpful aria label and added an aria label to the flyout item, so hopefully that should indicate it has a flyout now.

After investigating the breadcrumb focus issue you were having @jessiehuff , I believe it is because the whole drilldown menu is present in the DOM. The default Tab behavior was tabbing to something off screen. Continuing tabbing would eventually bring focus back to the current menu. Since Tab is prevented at the moment, it shouldn't be an issue now. If we add Tab back in for breadcrumbs, I will program in boundaries. But I'm still confused about how focus will shift between breadcrumb and menu if Tab would already be used for inter-breadcrumb navigation.

jessiehuff (commented):
So it looks like by removing the default Tab behavior, it's creating a keyboard trap within the menu. Is it possible to instead leave the menu if the user presses Tab like we have in dropdown-like components except in cases of the breadcrumb? That's the only one that I can imagine Tab being expected. We're combining a lot of different types of interactions which makes this really tricky.

I'm also seeing the weird scrolling thing on the drilldown breadcrumbs example. I'll send you a message with the recording of it since I've been having trouble including it in my comments.

kmcfaul:
The tricky bit with that is I have nowhere to send the focus after a Tab. In Dropdown, when the menu is tabbed, the menu closes and the focus returns to the toggle. I suppose I can try to find the next focusable element?

mcarrano (commented):
I will defer to @jessiehuff on this one. But let me know if there are any specific issues that you would like an opinion about.

jessiehuff (approved):
This is awesome!!! 🏆 🎉 You're a keyboard rockstar! 🤩 🎹

dlabrecq (approved):

PR #5572, opened 2021-03-24
fix(Topology): Allow applications to choose using resizable drawer for topology side bar

jeff-phillips-18:
**What**:
Provides the ability to opt in for the resizable topology side bar. Some applications may not want another Drawer inside the topology view due to layouts which cause the view to scroll rather than only the side bar contents scroll.


redallen (approved):

KKoukiou (approved):
It works as expected.

KKoukiou:
For the record it looks like this; 
Screen Shot 2021-03-25 at 16 24 42

PR #5573, opened 2021-03-25
feat(Table): Add Tree Table variant

### LINKED ISSUE #5016: Tree Table implementation
We would like to introduce tree table support based on the Cloud Platform implementation here: https://github.com/RedHatInsights/frontend-components/tree/master/packages/components/src/Components/TreeTable. You can see a live demo here: https://cloud.redhat.com/docs/api

Some questions to investigate include:

* Whether this is a new component or can be introduced as a new variant of the existing React table. Would like to inherit styling options from the existing table, but concerned about continued bloat of the Table component.
* Can we build using lightweight table approach that is currently being exposed (composable table) to alleviate size/complexity concerns?

There are currently known issues with responsive behavior. I have opened a corresponding design issue here to recommend an approach: https://github.com/patternfly/patternfly-design/issues/931

cc @evwilkin @mattnolting @karelhala

nicolethoen:
**What**: Closes #5016


nicolethoen:
I know there is still some clean up to do around aria attributes, and possibly documentation.
That's what I'll be working on while I wait for more feedback.

nicolethoen:
Im looking into allowing a partially selected checkbox state

nicolethoen:
confirmed - a partially checked checkbox is possible. We could build an example or demo in a future story to show how to wire up the checkboxes to cascade up.

packages/react-table/src/components/Table/Table.tsx
```diff
@@ -243,7 +248,8 @@ export class Table extends React.Component<TableProps, {}> {
           role={role}
           variant={variant}
           borders={borders}
-          className={className}
+          className={css(isTreeTable && 'pf-m-tree-view', className)}
```
tlabaj, line 251: Is this a placeholder classname?

jessiehuff (approved):
This looks pretty good, Nicole! 🙂  I tested with VO, and it sounded great. It doesn't quite meet the keyboard interaction expectations that are typically expected for a tree view (arrow key functionality in particular), but I know that we're trying to get this in so perhaps we can open another issue for the keyboard functionality? I detailed some of the keyboard expectations [here](https://github.com/patternfly/patternfly-react/issues/5016#issuecomment-782338541), and I would expect it to follow these [guidelines](https://www.w3.org/TR/wai-aria-practices/examples/treegrid/treegrid-1.html).

mattnolting:
This looks great @nicolethoen! 

Two things, can we wrap the primary or title `th` with `pf-c-table__tree-view-title-cell` and can we remove `pf-m-grid-md` (bc that will never really work for tree view).

mattnolting (changes requested):
This looks great @nicolethoen!

Two things, can we wrap the primary or title th with pf-c-table__tree-view-title-cell and can we remove pf-m-grid-md (bc that will never really work for tree view).

mattnolting (dismissed):
Excellent work, @nicolethoen! LPTM! 🎉🎉🎉

packages/react-table/src/components/Table/Table.tsx
```diff
@@ -243,7 +246,6 @@ export class Table extends React.Component<TableProps, {}> {
           role={role}
           variant={variant}
           borders={borders}
-          className={className}
```
jschuler, line 246: should this be removed?

packages/react-table/src/components/Table/TableTypes.ts
```diff
@@ -67,6 +68,19 @@ export type OnFavorite = (
   rowData: IRowData,
   extraData: IExtraData
 ) => void;
+export type OnTreeRowCollapse = (
+  event: any,
+  rowIndex: number,
+  title: string | React.ReactNode,
```
jschuler, line 74: ReactNode encompasses string already

packages/react-table/src/components/Table/TableTypes.ts
```diff
@@ -67,6 +68,19 @@ export type OnFavorite = (
...
+  rowData: IRowData
+) => void;
+export type OnCheckChange = (
+  event: React.FormEvent<HTMLInputElement>,
+  isChecked: boolean,
+  rowIndex: number,
+  title: string | React.ReactNode,
```
jschuler, line 81: same here

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1994,6 +1994,156 @@ class FavoritesTable extends React.Component {
...
+          branches: 'Branch eight',
+          pullRequests: 'Pull request eight',
+          workspaces: 'Workplace eight'
+        }
+      ],
+      expandedRows: ['Repositories one', 'Repositories six'],
+      checkedRows: []
+    };
+    
+    this.buildRows = ([x, ...xs], level, posinset, isHidden = false) => {
```
jschuler, line 2081: can you add a comment block to this function explaining what the params are and what is being returned?

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1994,6 +1994,156 @@ class FavoritesTable extends React.Component {
...
+        const isChecked = this.state.checkedRows.includes(x.repositories);
+        return [
+          {
+            cells: [x.repositories, x.branches, x.pullRequests, x.workspaces],
+            props: {
+              isExpanded: isExpanded,
+              isHidden,
+              level: level,
+              posinset: posinset,
+              setsize: x.children ? x.children.length : 0,
```
jschuler, line 2093: can this be camel cased everywhere? `setSize`

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1994,6 +1994,156 @@ class FavoritesTable extends React.Component {
...
+        const isExpanded = this.state.expandedRows.includes(x.repositories);
+        const isChecked = this.state.checkedRows.includes(x.repositories);
+        return [
+          {
+            cells: [x.repositories, x.branches, x.pullRequests, x.workspaces],
+            props: {
+              isExpanded: isExpanded,
+              isHidden,
+              level: level,
+              posinset: posinset,
```
jschuler, line 2092: camel case? `posInset`

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1994,6 +1994,156 @@ class FavoritesTable extends React.Component {
...
+    
+    this.buildRows = ([x, ...xs], level, posinset, isHidden = false) => {
+      if (x) {
+        const isExpanded = this.state.expandedRows.includes(x.repositories);
+        const isChecked = this.state.checkedRows.includes(x.repositories);
+        return [
+          {
+            cells: [x.repositories, x.branches, x.pullRequests, x.workspaces],
+            props: {
+              isExpanded: isExpanded,
```
jschuler, line 2089:
```suggestion
              isExpanded,
```

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1994,6 +1994,156 @@ class FavoritesTable extends React.Component {
...
+              isHidden,
+              level: level,
```
jschuler, line 2091:
```suggestion
              level,
```

packages/react-table/src/components/Table/utils/decorators/treeRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+  { rowIndex, rowData }: IExtra
+) => {
+  const { isExpanded, level, setsize, toggleAriaLabel, checkAriaLabel, isChecked, checkboxId } = rowData.props;
+  const text = (
+    <div className={css(stylesTreeView.tableTreeViewContent)}>
+      <span className="pf-c-table__text">{value.title || value}</span>
+    </div>
+  );
+  const onChange = (isChecked: boolean, event: React.FormEvent<HTMLInputElement>) => {
+    onCheckChange(event, isChecked, rowIndex, value.title, rowData);
```
jschuler, line 20: above you check for `value.title || value`, the same here then?
Maybe set a new variable `const content = value.title || value` and use that everywhere

packages/react-table/src/components/Table/utils/decorators/treeRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+    component: 'th',
+    className: 'pf-c-table__tree-view-title-cell',
+    children:
+      level !== undefined ? (
+        <div className={css(stylesTreeView.tableTreeViewMain)}>
+          {setsize > 0 && (
+            <span className={css(stylesTreeView.tableToggle)}>
+              <Button
+                variant="plain"
+                onClick={event => onCollapse && onCollapse(event, rowIndex, value.title, rowData)}
```
jschuler, line 33: same question here about value.title

packages/react-table/src/components/TableComposable/Td.tsx
```diff
@@ -74,6 +84,16 @@ export interface TdProps extends BaseCellProps, Omit<React.HTMLProps<HTMLTableDa
     /** Additional props forwarded to the FavoritesCell */
     props?: any;
   };
+  treeRow?: {
+    /** */
```
jschuler, line 88: comments missing

packages/react-table/src/components/TableComposable/Tr.tsx
```diff
@@ -11,7 +11,8 @@ export interface TrProps extends React.HTMLProps<HTMLTableRowElement>, OUIAProps
   className?: string;
   /** Forwarded ref */
   innerRef?: React.Ref<any>;
-
+  /** */
```
jschuler, line 14: comment missing

jschuler (commented):
Nice job @nicolethoen looks pretty good! just some minor things here.
Also, i noticed in the demo that when i checked a parent, the children were not automatically checked, i would expect them to be

packages/react-table/src/components/Table/Table.tsx
nicolethoen: I had changed the props a handful of times while working on this, so in the end while cleaning stuff up, I guess I realized that `className` was being pulled out of `props` unnecessarily on line 161. I can put it back, but it's now passed when I spread props to the Provider on line 230.

packages/react-table/src/components/Table/examples/Table.md
nicolethoen: it could, but the aria attribute is not camel cased (i.e. `aria-posinset`, `aria-setsize`), so I figured it was less confusing to format it the same way. I could also just use the actual aria attribute name, but number of quotes i'd need to wrap the attribute name annoyed me.

What do you vote? @tlabaj @kmcfaul?

nicolethoen:
I will have opened three follow up issues - 

1. Add the roving tabindex to enhance the keyboard accessibility #5596
2. Update the example to demonstrate how the consumer can use group checking, where checking a checkbox will check all children rows, and checking a children row will partially check a parent row #5597
3. Allow an icon to be included in each row to make the core examples. #5598

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -1994,6 +1994,166 @@ class FavoritesTable extends React.Component {
...
+### Tree table
+To enable a tree table:
+1. Pass the `isTreeTable` prop to the `Table` component
+2. Add `role="tree"` to the `TableBody` component
+3. Pass the following props to each row:
+    - `isExpanded` - Flag indicating the node is expanded and its children are visible
+    - `isHidden` - Flag indicating the node's parent is expanded and this node is visible
+    - `level` - number representing how many levels deep this node is nested
+    - `posinset` - number representing where in the order this node sits amongst its siblings 
+    - `setsize` - number representing the number of children this node has
```
dlabrecq, line 2006: I know this is an example, but would prefer to see `aria-posinset` and `aria-setsize` here, mainly because there is nothing to indicate these are aria attributes.

packages/react-table/src/components/Table/examples/Table.md
kmcfaul: I think using `'aria-posinest'` and `'aria-setsize'` with quotes would be the clearest. Or at least having "aria" in the property name to be clear what the prop is intended for ( i.e. `ariaPosInset` / `ariaSetSize` )

packages/react-table/src/components/Table/TreeRowWrapper.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Table/table';
+import { RowWrapperProps } from './RowWrapper';
+import { Tr } from '../TableComposable';
+
+export const TreeRowWrapper: React.FunctionComponent<RowWrapperProps> = ({
+  className,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  rowProps,
```
kmcfaul, line 10: Is `rowProps` something used in non-composable tables? Why is it being withheld from spreading to the Tr?

jessiehuff (commented):
Unfortunately the new roles seem to break accessibility for screen readers. 😞 When testing in VO, it reads "Tree Table, table, 4 columns 1 row", I'm able to get it to read each of the column headers, but then it says "end of table" without ever getting to the tree table content itself. I don't think the tree and treeitem roles are working well with the table elements, but that's a theory. I tested in JAWS and NVDA as well. JAWS was at least able to make slightly more of a guess, but it separated the headers and the tree table content so it announced the top part as a table and then only viewed the content as a tree view component (gave no features of a table, for example it never announced the column headers as you're navigating the content, just announces it as "tree view" then reads out all of the text). In NVDA, I had difficulty navigating it. It seemed to kind of work at times, other times it would jump around. When trying to expand sections, it would check the previous section, etc. I would describe the interaction as odd for lack of a better word. 

If we want to keep the tests, I'd rather remove the attributes getting flagged and add them back later, but ideally I'd love for axe-core to turn on support for these attributes so we can follow [the W3 example](https://www.w3.org/TR/wai-aria-practices/examples/treegrid/treegrid-1.html). There's already[ an open issue for it with Axe](https://github.com/dequelabs/axe-core/issues/2794) that Sarah Higley (a big name/thought leader in the a11y community) opened for the same reason we've run into. I can comment on that issue and mention our use case to hopefully get that pushed along so we can add these attributes as well.

nicolethoen:
@jessiehuff
so, the roles getting flagged by the tests are attached to the styles. For example if it has aria-level='3', the css knows to indent it three times. at this point we cannot remove the problematic roles.

I think our only option for this PR is to try and skip the examples in the testing... 
@jschuler could you help me figure out how to modify the script to do that?

packages/react-table/src/components/Table/TreeRowWrapper.tsx
nicolethoen: I am not totally sure all the reasons the rowData prop exists. At the very least it keeps track of rowIndex and rowKey for the sake of certain callback functions. But, I know that it's already being withheld in the RowWrapper, and if I don't withhold it in the TreeRowWrapper, it spreads all the way to the `<tr>` element and causes a console warning.

packages/react-docs/patternfly-a11y.config.js
```diff
@@ -16,5 +16,5 @@ module.exports = {
   urls: Object.keys(fullscreenRoutes),
   ignoreRules: 'color-contrast,page-has-heading-one,scrollable-region-focusable,bypass',
   ignoreIncomplete: true,
-  skip: /^\/charts\//
+  skip: /(^\/charts\/)|(^\/components\/table\/react\/composable-tree-table)|(^\/components\/table\/react\/tree-table)/
```
jschuler, line 19: you could make this more concise if you like
`skip: /^\/charts\/|tree-table$/`
maybe also add a comment why they are skipped?

jessiehuff (approved):
This looks awesome, Nicole!!!! 🎉 🏆  Thank you so much for all of your effort on this, you did a phenomenal job. I just tested in VO, NVDA, and JAWS, and all of them work like a dream with this structure. 😄

dlabrecq (approved):

kmcfaul (approved):

PR #5605, opened 2021-03-30
fix(card): add empty state to card demo

### LINKED ISSUE #5337: Add empty state to card demo
Add a New card/empty state card to the Card View demo here: https://www.patternfly.org/v4/components/card/react-demos#card-view

Designs: https://marvelapp.com/prototype/f869h61/screen/74368154

https://github.com/patternfly/patternfly/pull/3794

DaoDaoNoCode:
**What**: Closes #5337 

Also change all cards to `compact` to make it the same as HTML demo.

### DIFF
--- a/packages/react-core/src/demos/Card/Card.md
+++ b/packages/react-core/src/demos/Card/Card.md
@@ -8,7 +8,7 @@ import CogIcon from '@patternfly/react-icons/dist/js/icons/cog-icon';
 import FilterIcon from '@patternfly/react-icons/dist/js/icons/filter-icon';
 import TrashIcon from '@patternfly/react-icons/dist/js/icons/trash-icon';
 import HelpIcon from '@patternfly/react-icons/dist/js/icons/help-icon';
-import AddCircleOIcon from '@patternfly/react-icons/dist/js/icons/add-circle-o-icon';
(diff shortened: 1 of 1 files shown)


mcarrano (changes requested):
@DaoDaoNoCode Looks like you are using a different icon than the design here: https://marvelapp.com/prototype/f869h61/screen/73766143 Should be the filled version.

Also, @mmenestr do we care if the Add is the first or last card? I recall some discussion on this and seemed like it could be either, but want to make sure the demo does not contradict design guidance.

mcarrano:
Thanks @mmenestr . I don't know about having both demos, but I do agree that the current text does not feel right given the context. Can we just make updates to the icon and text on the empty state card in that case @DaoDaoNoCode ?

DaoDaoNoCode:
@mcarrano Will update soon!

mcarrano (approved):
Looks good. Thanks @DaoDaoNoCode .

mcoker (approved):
LGTM! I created a follow up issue for core to update the empty state text. Feel free to take it if you'd like @DaoDaoNoCode 

https://github.com/patternfly/patternfly/issues/3982

tlabaj (approved):
Look great!

jessiehuff (approved):
LGTM!

PR #5621, opened 2021-04-05
feat(Form): Add stacked variant and section title

### LINKED ISSUE #5286: Form enhancements
Add form sections and the stack layout that allows you to stack checkboxes, radios, text inputs, etc in a single group per core changes made here: https://github.com/patternfly/patternfly/pull/3767.

@mcoker please add details as necessary.

nicolethoen:
Closes #5286

Add isStacked to FormGroup so the FormGroupControls can be stacked.
Add title to FormSection

### DIFF
--- a/packages/react-core/src/components/Form/FormGroup.tsx
+++ b/packages/react-core/src/components/Form/FormGroup.tsx
@@ -23,6 +23,8 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   validated?: 'success' | 'warning' | 'error' | 'default';
   /** Sets the FormGroup isInline. */
   isInline?: boolean;
+  /** Sets the FormGroupControl to be stacked */
+  isStacked?: boolean;
   /** Removes top spacer from label. */
   hasNoPaddingTop?: boolean;
   /** Helper text regarding the field. It can be a simple text or an object. */
@@ -48,6 +50,7 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
   validated = 'default',
   isInline = false,
   hasNoPaddingTop = false,
+  isStacked = false,
   helperText,
   isHelperTextBeforeField = false,
   helperTextInvalid,
@@ -107,7 +110,13 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
         </div>
       )}
 
-      <div className={css(styles.formGroupControl, isInline && styles.modifiers.inline)}>
+      <div
+        className={css(
+          styles.formGroupControl,
+          isInline && styles.modifiers.inline,
+          isStacked && styles.modifiers.stack
+        )}
+      >
         {isHelperTextBeforeField && helperTextToDisplay}
         {children}
         {!isHelperTextBeforeField && helperTextToDisplay}
--- a/packages/react-core/src/components/Form/FormGroup.tsx
+++ b/packages/react-core/src/components/Form/FormGroup.tsx
@@ -24,7 +24,7 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   /** Sets the FormGroup isInline. */
   isInline?: boolean;
   /** Sets the FormGroupControl to be stacked */
-  isStacked?: boolean;
+  isStack?: boolean;
   /** Removes top spacer from label. */
   hasNoPaddingTop?: boolean;
   /** Helper text regarding the field. It can be a simple text or an object. */
@@ -50,7 +50,7 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
   validated = 'default',
   isInline = false,
   hasNoPaddingTop = false,
-  isStacked = false,
+  isStack = false,
   helperText,
   isHelperTextBeforeField = false,
   helperTextInvalid,
@@ -111,11 +111,7 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
       )}
 
       <div
-        className={css(
-          styles.formGroupControl,
-          isInline && styles.modifiers.inline,
-          isStacked && styles.modifiers.stack
-        )}
+        className={css(styles.formGroupControl, isInline && styles.modifiers.inline, isStack && styles.modifiers.stack)}
       >
         {isHelperTextBeforeField && helperTextToDisplay}
         {children}
--- a/packages/react-core/src/components/Form/FormSection.tsx
+++ b/packages/react-core/src/components/Form/FormSection.tsx
@@ -9,16 +9,19 @@ export interface FormSectionProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   className?: string;
   /** Title for the section */
   title?: React.ReactNode;
+  /** Element to wrap the section title*/
+  titleElement?: 'div' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 }
 
 export const FormSection: React.FunctionComponent<FormSectionProps> = ({
   className = '',
   children,
   title = '',
+  titleElement: TitleElement = 'div',
   ...props
 }: FormSectionProps) => (
   <section {...props} className={css(styles.formSection, className)}>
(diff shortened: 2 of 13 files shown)


packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -107,7 +110,13 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
         </div>
       )}
 
-      <div className={css(styles.formGroupControl, isInline && styles.modifiers.inline)}>
+      <div
+        className={css(
+          styles.formGroupControl,
+          isInline && styles.modifiers.inline,
+          isStacked && styles.modifiers.stack
```
mcoker, line 117: wdyt about `isStack`? It matches the modifier name and the layout and usage of `pf-m-stack` in the sidebar. We do have `pf-m-stacked` in the tile component but I'm wondering if we should change that. @mcarrano do you know if we have any kind of guideline for this?

packages/react-core/src/components/Form/FormSection.tsx
```diff
@@ -2,19 +2,23 @@ import * as React from 'react';
...
 }
 
 export const FormSection: React.FunctionComponent<FormSectionProps> = ({
   className = '',
   children,
+  title = '',
   ...props
 }: FormSectionProps) => (
   <section {...props} className={css(styles.formSection, className)}>
+    {title && <div className={css(styles.formSectionTitle, className)}>{title}</div>}
```
mcoker, line 21: Should we allow `titleComponent` here (and update the core docs) to allow a heading or other element to be used?

mcoker:
I'm wondering if the checkbox options in the horizontal example are part of the same (stack) group or if they're separate form groups.

Separate groups:
Screen Shot 2021-04-05 at 11 14 21 AM

Same/stack group:
Screen Shot 2021-04-05 at 11 18 42 AM

@mcarrano @mceledonia wdyt? Another way of looking at it would be, if the checkbox options were labelled, would they have the same label in the form, eg "Settings" or would they have separate labels?

FWIW in the core demos, I updated the vertical and horizontal form examples to be the same, with the difference just being the form layout - the vertical form layout uses an inline group for the checkbox options and the horizontal form uses a stack group.

https://www.patternfly.org/v4/components/form/html-demos

mcoker (commented):
Looks good to me! Left some comments for review. Main question is around the prop name for the `pf-m-stack` variation.

packages/react-core/src/components/Form/FormSection.tsx
nicolethoen: That'll be an easy enough change for a follow up after the update to core is made, but I can also open a core issue and pause on this until its done. either way

mcoker: PR is here - https://github.com/patternfly/patternfly/pull/3973

packages/react-core/src/components/Form/FormGroup.tsx
mcarrano: Not sure I understand the question. Do you mean when to use "stack" vs "stacked?"

mcoker: @mcarrano yes, I noticed that we may have an inconsistent use of stack & stacked.

* `pf-l-stack` layout 
* `pf-m-stack` variation of the sidebar layout
* `pf-m-stack` variation of form control groups (to stack inputs on top of one another)
* `pf-m-stacked` variation for tiles

mcarrano:
Looks good to me also.

> @mcarrano @mceledonia wdyt? Another way of looking at it would be, if the checkbox options were labelled, would they have the same label in the form, eg "Settings" or would they have separate labels?

Regarding this question, yes, I think they would have one label and use this new Stack layout.

One question I have is why do the checkboxes so badly align with the labels. Here's what I see in Firefox:

Screen Shot 2021-04-05 at 1 54 30 PM

 Is this a browser thing? Any way to make this better short of introducing a custom Check component?

packages/react-core/src/components/Form/FormGroup.tsx
mcarrano: I see. Yes, it would be nice to be consistent. @mcoker what are the implications of changing this in Tile? Won't that potentially break people?

mcoker: @mcarrano it's a breaking change, yes. We'll just make sure the usage in this PR is consistent with our intent, then update any other component we need in a breaking change release.

packages/react-core/src/components/Form/FormSection.tsx
mcoker: @nicolethoen the docs were updated in core

kmcfaul (approved):
lgtm on the JS side.

kmcfaul:
@mcarrano I'm not seeing the small checkboxes in Firefox, it looks fine for me. The example uses the PF `Checkbox` component so it should be styled.

packages/react-core/src/components/Form/FormSection.tsx
```diff
@@ -2,19 +2,26 @@ import * as React from 'react';
...
 import { css } from '@patternfly/react-styles';
 
-export interface FormSectionProps extends React.HTMLProps<HTMLDivElement> {
+export interface FormSectionProps extends Omit<React.HTMLProps<HTMLDivElement>, 'title'> {
   /** Content rendered inside the section */
   children?: React.ReactNode;
   /** Additional classes added to the section */
   className?: string;
+  /** Title for the section */
+  title?: React.ReactNode;
```
tlabaj, line 11: @jessiehuff   are there any accessibility concerns if the title is not provided for the section?

jessiehuff: I think the most important thing is that the inputs have labels, but I suppose it depends on how we expect form sections to be used. If we're expecting this to be used similar to how a `<section>` element is used then it should probably have a label, as it's [required](https://www.w3.org/WAI/tutorials/page-structure/sections/#:~:text=Be%20careful%20not%20to%20overuse,should%20also%20always%20be%20labeled.) for section elements. Titles could definitely give added context if it's available, but I wonder if the idea of a form section will always have that kind of context?

jessiehuff (dismissed):
This looks good from a keyboard and VO standpoint, though Titani did bring up an interesting question which made me wonder about the use case of form sections. However, since we haven't used titles before on form sections, I'm inclined to think that the titles just add context if it's available and may not really impact the accessible experience too much if it's not there.

tlabaj (commented):
this just needs an update to the demo-app and a test.

tlabaj (changes requested):
demo app need updating

packages/react-integration/demo-app-ts/src/components/demos/FormDemo/FormDemo.tsx
```diff
@@ -165,7 +165,7 @@ export class FormDemo extends Component<FormProps, FormState> {
               ))}
             </Select>
           </FormGroup>
-          <FormSection>
+          <FormSection title="Title" titleElement="h4">
```
tlabaj, line 168: how about the new isStack prop?

nicolethoen: Just for my own understanding, I usually tend to think of the cypress tests as testing interactivity and state changes. 
Both of these updates are just layout updates, so is that not what the snapshots and jest tests are for? All i'll be doing is adding the flags and writing 'does this classname exist on the page' sort of cypress tests. Is that what you mean?

tlabaj: yup.  that is true.  the demo-app also gives us a way to test the correct typing of the api.  In this case it is just a boolean so it is not a big deal.  We have had a lot of typing issues in the past that we have only been able to to find in the demo-app.  That was more so before we converted to typescript.

tlabaj (dismissed):

jessiehuff (dismissed):

tlabaj (approved):

kmcfaul (approved):

Issue #5626, opened 2021-04-06
React-Virtualized-Ext: Demos have incorrect table cell data.


jcaianirh:
**Describe the issue. What is the expected and unexpected behavior?**
Demo data is just repeated for each cell in the table instead of cycling through the column and row data.
**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

http://patternfly-react.surge.sh/extensions/virtual-scroll-table
**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
bug
**What is your product and what release version are you targeting?**

jcaianirh:
closing as this was fixed

PR #5627, opened 2021-04-06
fix(react-virtualized): fix render of demo data so the table cells don't repeat static text

jcaianirh:
**What**: Closes # https://github.com/patternfly/patternfly-react/issues/5626

Before:
Screen Shot 2021-04-06 at 12 33 19 AM

After:
Screen Shot 2021-04-06 at 12 32 53 AM

**Additional issues**:

### DIFF
--- a/packages/react-virtualized-extension/src/components/Virtualized/examples/VirtualizedTable.md
+++ b/packages/react-virtualized-extension/src/components/Virtualized/examples/VirtualizedTable.md
@@ -87,25 +87,24 @@ class VirtualizedExample extends React.Component {
 
     const rowRenderer = ({ index, isScrolling, key, style, parent }) => {
       const { rows, columns } = this.state;
-      const text = rows[index].cells[0];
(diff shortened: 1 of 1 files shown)


jschuler (approved):
Thanks for fixing the demo!

jessiehuff (commented):
Looks good! The issues that I've seen with a11y still exist, but they're unrelated to your PR. :) The only thing is that I noticed there's a ReferenceError with the Selectable example that I don't remember seeing before.

jcaianirh:
> Looks good! The issues that I've seen with a11y still exist, but they're unrelated to your PR. :) The only thing is that I noticed there's a ReferenceError with the Selectable example that I don't remember seeing before.

No thats a bug! Thanks fixing it now.

jschuler (approved):

jessiehuff (approved):
Awesome, thanks Joe! 🙂

PR #5629, opened 2021-04-06
feat(ClipboardCopy): add inline variant

### LINKED ISSUE #5555: Add inline copy to clipboard variant
Add an inline variant of Clipboard copy per the core PR here: https://github.com/patternfly/patternfly/pull/3933

----

Adds an inline variation via `.pf-m-inline` and displays an inline version of copy to clipboard that wraps and behaves like regular text. You can also apply the modifier `.pf-m-block`, and keep the inline styling, but make it take up a block layout. 

The clipboard copy component has 2 children:
* `.pf-c-clipboard-copy__text`
  * holds the text to copy. This either a `<span>` for use with regular text (default), or can be `<code>` for use with code. When code, it also needs `.pf-m-code`, which changes the font.
* `.pf-c-clipboard-copy__actions`
  * holds any actions. Actions are plain buttons wrapped in `.pf-c-clipboard-copy__actions-item`. There is a copy to clipboard action as default, then the ability to add more actions if you want.

kmcfaul:
**What**: Closes #5555

- Utilizes `inline` value of `variant` property to toggle between inline/expand/default types of ClipboardCopy.
- Reuses internal `ClipboardCopyButton` for inline variant, changing the button styling from control to plain. (@mcoker so the html structure for the button differs slightly from core, since the current button has a tooltip. Let me know if I should keep/remove the tooltip).
- Exposes `ClipboardCopyAction` as a wrapper for additional actions added to inline ClipboardCopy via the `additionalActions` property.

### DIFF
--- a/packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
+++ b/packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
@@ -22,7 +22,8 @@ export const clipboardCopyFunc = (event: React.ClipboardEvent<HTMLDivElement>, t
 
 export enum ClipboardCopyVariant {
   inline = 'inline',
-  expansion = 'expansion'
+  expansion = 'expansion',
+  inlineCompact = 'inline-compact'
 }
 
 export interface ClipboardCopyState {
@@ -51,7 +52,7 @@ export interface ClipboardCopyProps extends Omit<React.HTMLProps<HTMLDivElement>
   /** Flag to determine if inline clipboard copy should be block styling */
   isBlock?: boolean;
   /** Adds Clipboard Copy variant styles. */
-  variant?: typeof ClipboardCopyVariant | 'inline' | 'expansion';
+  variant?: typeof ClipboardCopyVariant | 'inline' | 'expansion' | 'inline-compact';
   /** Copy button popover position. */
   position?: PopoverPosition | 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Maximum width of the tooltip (default 150px). */
@@ -68,7 +69,7 @@ export interface ClipboardCopyProps extends Omit<React.HTMLProps<HTMLDivElement>
   onChange?: (text?: string | number) => void;
   /** The text which is copied. */
   children: React.ReactNode;
-  /** Additional actions for inline clipboad copy. Should be wrapped with ClipboardCopyAction. */
+  /** Additional actions for inline clipboard copy. Should be wrapped with ClipboardCopyAction. */
   additionalActions?: React.ReactNode;
 }
 
@@ -90,7 +91,7 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
     isReadOnly: false,
     isExpanded: false,
     isCode: false,
-    variant: null,
+    variant: 'inline',
     position: TooltipPosition.top,
     maxWidth: '150px',
     exitDelay: 1600,
@@ -153,14 +154,14 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
       <div
         className={css(
           styles.clipboardCopy,
-          variant === 'inline' && styles.modifiers.inline,
+          variant === 'inline-compact' && styles.modifiers.inline,
           isBlock && styles.modifiers.block,
           this.state.expanded && styles.modifiers.expanded,
           className
         )}
         {...divProps}
       >
-        {variant === 'inline' && (
+        {variant === 'inline-compact' && (
           <GenerateId prefix="">
             {id => (
               <React.Fragment>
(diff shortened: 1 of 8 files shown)


mcoker (dismissed):
LGTM! 🥳

mcarrano (approved):

packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
```diff
@@ -86,7 +90,7 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
     isReadOnly: false,
     isExpanded: false,
     isCode: false,
-    variant: 'inline',
+    variant: null,
```
tlabaj, line 93: We can't change the API like this.  We are redefining what we were calling inline before.  this could break consumers that where already using the "inline" variation.  we will have to call the inline variation something else. We can refactor when we have a breaking change release...
cc @mcoker @mcarrano

packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
```diff
@@ -66,6 +68,8 @@ export interface ClipboardCopyProps extends Omit<React.HTMLProps<HTMLDivElement>
   onChange?: (text?: string | number) => void;
   /** The text which is copied. */
   children: React.ReactNode;
+  /** Additional actions for inline clipboad copy. Should be wrapped with ClipboardCopyAction. */
```
tlabaj, line 71: clipboard typo.

tlabaj (changes requested):

packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
kmcfaul: I'll add a variant value `inline-compact` for the new inline styling for the interim. We can update properly (`inline`, `expansion` and maybe `block` for the variant values) when we have a breaking change release.

kmcfaul:
Reverted the default `variant` and added the new `inline-compact` variant for the inline styling. Shouldn't be breaking anymore.

packages/react-core/src/components/ClipboardCopy/examples/ClipboardCopy.md
```diff
@@ -90,6 +97,79 @@ import { ClipboardCopy, ClipboardCopyVariant } from '@patternfly/react-core';
...
+### Inline
+
+```js
+import React from 'react';
+import { ClipboardCopy } from '@patternfly/react-core';
+
+<ClipboardCopy variant="inline">2.3.4-2-redhat</ClipboardCopy>;
+```
+
+### Inline code
```
tlabaj: I would suggest we change the name of this example so the consumers are not confused.  What we are calling "inline" is what Core has as "Basic". @mcoker @mcarrano   Core should probably be updated as well for consistency.

jessiehuff (commented):
When testing this in VO I noticed that the label on the copy to clipboard button wasn't being announced for any of the new examples. They were all being read as "button." When looking at the markup, it looks like there's a mismatch between the id and the aria-labelledby:
Screen Shot 2021-04-13 at 2 13 23 PM
See how the id in this first example, compact inline, is `id="text-124"` but the `aria-labelledy="copy-button-124 text-input-124"`? It's trying to find "text-input-124" but the id being used is "text-124". This looks like it's the case for each of these examples.

kmcfaul:
ah I think I know what's causing that, I will update

jessiehuff:
It looks like an extra dash got added, but it's so close!! 🙂 
Screen Shot 2021-04-14 at 9 30 17 AM

kmcfaul:
Ah my bad, the variable for the prefix had already included the hyphen. Updated! @jessiehuff

jessiehuff (dismissed):
Awesome as always! 😄

mcoker:
@mcarrano can you verify that the updated example names from "inline" to "inline compact" look good in this PR? We also need to match those example names in core, which I have a PR up for already.

The examples in this PR are here - https://patternfly-react-pr-5629.surge.sh/components/clipboard-copy#compact-inline

The updated examples in [the core PR](https://github.com/patternfly/patternfly/pull/3998) are here - https://patternfly-pr-3998.surge.sh/components/clipboard-copy#compact-inline

mcarrano:
@kmcfaul @mcoker I think we previously agreed to call these "inline compact" but the examples are labeled "Compact inline." That's what we used in the design doc here: https://patternfly-org-pr-2446-v4.surge.sh/v4/components/clipboard-copy/design-guidelines

Can we update?

mcoker (dismissed):

mcarrano (approved):

tlabaj (commented):
Looks good.  Can you just add additional actions to the demo app please.

tlabaj (approved):

jessiehuff (approved):

PR #5631, opened 2021-04-07
Select: change arguments of the onFilter callback

### LINKED ISSUE #5625: Incorrect type for `onFilter` prop of Select component - can break an app when Select children are updated

The TypeScript type of Select's `onFilter` prop is `(e: React.ChangeEvent<HTMLInputElement>) => React.ReactElement[]`, implying that the event passed to it will always be defined/truthy. It appears that after https://github.com/patternfly/patternfly-react/pull/5375, `onFilter` gets called with a `null` value for its event argument when the component's children change. See https://github.com/patternfly/patternfly-react/blame/fe2da75b77e4cbf92ef11b71e87aef998fa8dd46/packages/react-core/src/components/Select/Select.tsx#L209

When I upgraded PF, this caused an error in my code which passes a handler to `onFilter` expecting the event to always be defined and non-null. The example in our docs doesn't hit it because it has a try-catch block when it references the event.

See this CodeSandbox: https://codesandbox.io/s/laughing-forest-o01j4?file=/index.js
It is based on the one from [the Custom Filtering example](https://www.patternfly.org/v4/components/select#custom-filtering) in the docs, with the try-catch in `onFilter` removed and a new state property introduced that forces the children to change after the component mounts.

This is a bug. The workaround is to use a try-catch or otherwise defensively code around the possibility of the event being null, but TypeScript could catch this for consumers if the `onFilter` prop type was changed to `(e: React.ChangeEvent<HTMLInputElement> | null) => React.ReactElement[]`.

Migration Toolkit for Virtualization. No particular target release.

KKoukiou:
When the children of the select are updated while the component is
mounted make sure to apply the filter as well.

Add a demo to demostrate the new functionality.

Fixes #5625

### DIFF
--- a/packages/react-core/src/components/Select/examples/Select.md
+++ b/packages/react-core/src/components/Select/examples/Select.md
@@ -589,8 +589,7 @@ class FilteringCheckboxSelectInput extends React.Component {
       }
     };
 
-    this.onFilter = evt => {
-      const textInput = evt.target.value;
+    this.onFilter = (_, textInput) => {
       if (textInput === '') {
         return this.options;
       } else {
@@ -695,8 +694,7 @@ class FilteringCheckboxSelectInputWithPlaceholder extends React.Component {
       }
     };
 
-    this.onFilter = evt => {
-      const textInput = evt.target.value;
+    this.onFilter = (_, textInput) => {
       if (textInput === '') {
         return this.options;
       } else {
@@ -809,8 +807,7 @@ class FilteringCheckboxSelectInputWithBadging extends React.Component {
       }
     };
 
-    this.onFilter = evt => {
-      const textInput = evt.target.value;
+    this.onFilter = (_, textInput) => {
       if (textInput === '') {
         return this.options;
       } else {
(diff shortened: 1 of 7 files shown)


packages/react-core/src/components/Select/examples/Select.md
```diff
@@ -589,8 +589,7 @@ class FilteringCheckboxSelectInput extends React.Component {
       }
     };
 
-    this.onFilter = evt => {
```
mturley, line 592: These `this.onFilter` methods in the examples also need to be changed to have `textInput` as the second arg now I think

mturley:
Thanks for fixing this so quickly @KKoukiou !

packages/react-core/src/components/Select/examples/Select.md
KKoukiou: Good catch, thanks.

mturley (approved):
LGTM!

tlabaj (approved):
looks good!

kmcfaul (approved):

PR #5643, opened 2021-04-09
feat(ExpandableSection): add detached variant, update tests

### LINKED ISSUE #5581: Expandable section - support detaching the toggle from the expandable content
This feature will be used in the code block component, and was done in core as part of the code block component work - https://github.com/patternfly/patternfly/pull/3937/files

Basically you have 2 `.pf-c-expandable-section` elements, and each has the class `.pf-m-detached`. One has the expandable section toggle, and the other has the expandable section content.

A unique `id` is added to the `.pf-c-expandable-section__content` element in the expandable section with the content, and that `id` is used as the value for `aria-controls` on the toggle in the other expandable section element. 

We've also added the class `.pf-m-expand-top` to the `.pf-c-expandable-section__toggle-icon` element in the toggle to tell the arrow which direction to point when the toggle is expanded. Technically this can apply to any expandable section, not just the detached variation. And we should leave room to support other directions, too.

cc @mcarrano

kmcfaul:
**What**: Closes #5581

- Adds `ExpandableSectionToggle` component for detached expandable sections.
- Adds `isDetached` flag to `ExpandableSection`. This removes the built in toggle and adds the appropriate styling.
- Adds `contentId` property to both components to populate `aria-controls` label. There might be a better name for this property?
- `isExpanded` must be passed in to both components (there is no uncontrolled variation for detached section).

Also updated tests + added to integration/cypress.

### DIFF
--- a/packages/react-core/src/components/ExpandableSection/ExpandableSectionToggle.tsx
+++ b/packages/react-core/src/components/ExpandableSection/ExpandableSectionToggle.tsx
@@ -14,6 +14,8 @@ export interface ExpandableSectionToggleProps extends React.HTMLProps<HTMLDivEle
   onToggle?: (isExpanded: boolean) => void;
   /** ID of the toggle's respective expandable section content. */
   contentId?: string;
+  /** Direction the toggle arrow should point when the expandable section is expanded. */
+  direction?: 'up' | 'down';
 }
 
 export const ExpandableSectionToggle: React.FunctionComponent<ExpandableSectionToggleProps> = ({
@@ -22,6 +24,7 @@ export const ExpandableSectionToggle: React.FunctionComponent<ExpandableSectionT
   isExpanded = false,
   onToggle,
   contentId,
+  direction = 'down',
   ...props
 }: ExpandableSectionToggleProps) => (
   <div
@@ -40,7 +43,12 @@ export const ExpandableSectionToggle: React.FunctionComponent<ExpandableSectionT
       aria-controls={contentId}
       onClick={() => onToggle(!isExpanded)}
     >
-      <span className={css(styles.expandableSectionToggleIcon, isExpanded && styles.modifiers.expandTop)}>
+      <span
+        className={css(
+          styles.expandableSectionToggleIcon,
+          isExpanded && direction === 'up' && styles.modifiers.expandTop
+        )}
+      >
         <AngleRightIcon aria-hidden />
       </span>
       <span className={css(styles.expandableSectionToggleText)}>{children}</span>
--- a/packages/react-core/src/components/ExpandableSection/ExpandableSectionToggle.tsx
+++ b/packages/react-core/src/components/ExpandableSection/ExpandableSectionToggle.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/ExpandableSection/expandable-section';
+import { css } from '@patternfly/react-styles';
+import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
+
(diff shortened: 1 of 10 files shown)


nicolethoen (dismissed):
Looks great!

jessiehuff (dismissed):
So theoretically the `aria-controls` attribute should associate the toggle with the content for the screen reader so that when the button is pressed the user should be aware of the content. However, I had a hard time getting this to work in testing. I looked up support for the attribute and since it's supposed to be supported strongest in JAWS, I tested in JAWS 2019 with Chrome and Firefox and couldn't seem to see the effect. So it might be worth looking into in the future if there's anything else we can do, but theoretically we're associating everything correctly. 


I also attempted using `aria-labelledby` like we do with Tab, but that overrides the button text and makes much less sense. We could also add an `aria-label` (like `aria-label="Show more above"`) that would override the button text to make it clearer with the direction, but it's possible users will understand this already in context. I think this one might just depend on the use case and context. In our code editor, for example, it might already make sense that when you're selecting the show more button, the text should appear above in the code editor. This one feels difficult/subjective to say without talking to actual users. So I think we can move forward with this for now until we have more information.

kmcfaul:
Thanks for looking into this! @jessiehuff

mcarrano (commented):
This looks good to me, although I'm not really sure I would understand what this was for without further documentation. @kmcfaul @mcoker @tlabaj do you think the term "Detached" will be meaningful to developers on it's own?

tlabaj:
@mcarrano this is in support of the expandable content of the Code block component.

tlabaj (approved):
looks great Katie!

mcarrano:
> @mcarrano this is in support of the expandable content of the Code block component.

I understand that. Just not sure how I would know that by looking at the example. Are there any other use cases where this might be useful?

tlabaj:
@mcarrano got it.  Detached makes sense to me from development perspective.  I would be interested in other opinions as well.  
Do you think that we need to had more verbiage so designers would also know the intention?

kmcfaul:
Detached makes sense to me, because the `ExpandableToggle` is not attached to the `ExpandableSection` and floats separately in the DOM. 

`Separated` would also make sense to me if we're looking for other terms. But maybe we can expand the `isDetached` prop description if it's unclear?

mcarrano (dismissed):
@kmcfaul @tlabaj if both of you think "Detached" makes sense then I am ok to stick with it.

packages/react-core/src/components/ExpandableSection/ExpandableSectionToggle.tsx
```diff
@@ -0,0 +1,51 @@
...
+    )}
+  >
+    <button
+      className={css(styles.expandableSectionToggle)}
+      type="button"
+      aria-expanded={isExpanded}
+      aria-controls={contentId}
+      onClick={() => onToggle(!isExpanded)}
+    >
+      <span className={css(styles.expandableSectionToggleIcon, isExpanded && styles.modifiers.expandTop)}>
```
mcoker, line 43: `styles.modifiers.expandTop` should probably be added via a prop that the user sets to indicate that the expanded content is above the toggle. By default, it's fair to assume the content is below the toggle (just like in the default use of the component), so the toggle arrow should point down when expanded.

mcoker (changes requested):
This looks great! Just one comment about the toggle arrow direction. The toggle arrow shouldn't change direction by default - should only change if the user has specified they want to point the toggle arrow up because the expanded content is above the toggle.

tlabaj (approved):

mcoker (approved):
Looks great!

nicolethoen (approved):

Issue #5644, opened 2021-04-09
lots of warnings with webpack 5: export '...' (imported as 'Table') was not found in 'reactabular-table'


martinpitt:
**Describe the issue. What is the expected and unexpected behavior?**

I am moving https://github.com/cockpit-project/cockpit-machines from webpack 4 to 5. This uses a recent @patternfly/react-core 4.106.2.

Building does work, but it now gives 23 warnings of this form:
```
WARNING in ./node_modules/patternfly-react/dist/esm/components/Table/Table.js 21:0-31
export 'actionHeaderCellFormatter' (imported as 'Table') was not found in 'reactabular-table' (possible exports: Body, BodyRow, Header, Provider, __esModule, columnsAreEqual, evaluateFormatters, evaluateTransforms, mergeProps, resolveRowKey)

WARNING in ./node_modules/patternfly-react/dist/esm/components/Table/Table.js 22:0-38
export 'customHeaderFormattersDefinition' (imported as 'Table') was not found in 'reactabular-table' (possible exports: Body, BodyRow, Header, Provider, __esModule, columnsAreEqual, evaluateFormatters, evaluateTransforms, mergeProps, resolveRowKey)

WARNING in ./node_modules/patternfly-react/dist/esm/components/Table/Table.js 23:0-25
export 'defaultSortingOrder' (imported as 'Table') was not found in 'reactabular-table' (possible exports: Body, BodyRow, Header, Provider, __esModule, columnsAreEqual, evaluateFormatters, evaluateTransforms, mergeProps, resolveRowKey)
[...]
```

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

```
git clone -b webpack5 https://github.com/martinpitt/cockpit-machines
cd cockpit-machines
make
```

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

bug, but not fatal

redallen:
Many Webpack 5 projects use patternfly packages without these errors. We haven't depended on `reactabular-table` in quite a while, my guess is that your old `patternfly-react` package is causing this error.

PR #5645, opened 2021-04-09
fix(inline-edit): remove unused dep

redallen:
**What**: Just noticed an unused dep laying around. Related to #5644

**Additional issues**:


tlabaj (dismissed):

nicolethoen (approved):
@tlabaj can we be sure that we note this removed dependency in the release notes? Just in case people randomly were referencing it in non-patternfly code? I have a vague memory of removing a dependency like this cause problems once upon a time.

dlabrecq (dismissed):

redallen:
Merging before more merge conflicts come in from the next merge.

PR #5651, opened 2021-04-12
fix(stalebot): add closeComment

### LINKED ISSUE #3992: Switch Component should fire change event without id attribute
Currently, if a Switch Component does not have an id attribute, the `onChange` handler is not fired. This breaks previous behaviour, where the `onChange` handler is fired regardless of having an id attribute or not.

redallen:
**What**: Closes #3992. Should fix the bot which I don't think is running. Not sure why it was [added](https://probot.github.io/apps/stale/) on Probot's side.

**Additional issues**:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -21,5 +21,8 @@ markComment: >
   This issue has been automatically marked as stale because it has not had
   recent activity. It will be closed if no further activity occurs.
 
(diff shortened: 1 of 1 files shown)


kmcfaul (approved):

jschuler (approved):

dlabrecq (commented):
LGTM

PR #5656, opened 2021-04-12
feat(CodeBlock): add component

### LINKED ISSUE #5556: Add Code Block component
Code Block is a new component for presenting a snippet of read-only code. It is lighter weight than the Code Editor in both appearance and code footprint. It supports optional actions but not editing.

See https://github.com/patternfly/patternfly/pull/3937

----

The whitespace in the `<pre><code>...</code></pre>` block needs to be managed carefully. The whitespace in the core examples is incorrect, and that's artifact of the workspace example/code formatter, and we're working to resolve that. This is how the blocks should be formatted

Screen Shot 2021-03-25 at 4 27 21 PM

Also there is an expand/collapse feature in the code block that is using a new enhancement to the expandable section component - https://github.com/patternfly/patternfly-react/issues/5581

kmcfaul:
**What**: Closes #5556

- Adds `CodeBlock`, `CodeBlockAction`, `CodeBlockCode` components
- `CodeBlockAction` styles buttons for the header and multiple actions may be passed to `CodeBlock` via the `actions` property.
- `CodeBlockCode` styles code content for the `CodeBlock`. Detached expandable sections can be combined with `CodeBlockCode` to create expandable code sections (example is in, but commented out until #5643 merges).

Docs: https://patternfly-react-pr-5656.surge.sh/components/code-block

@mcoker Is `.pf-c-code-block__actions-item` a placeholder class? I couldn't find it in the styles object so I have it hardcoded atm.

### DIFF
--- a/packages/react-core/src/components/CodeBlock/CodeBlockAction.tsx
+++ b/packages/react-core/src/components/CodeBlock/CodeBlockAction.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+import { css } from '@patternfly/react-styles';
+
+export interface CodeBlockActionProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the code block action */
+  children?: React.ReactNode;
+  /** Additional classes passed to the code block action */
+  className?: string;
+}
+
+export const CodeBlockAction: React.FunctionComponent<CodeBlockActionProps> = ({
+  children = null,
+  className,
+  ...props
+}: CodeBlockActionProps) => (
+  <div className={css('pf-c-code-block__actions-item', className)} {...props}>
+    {children}
+  </div>
+);
+
+CodeBlockAction.displayName = 'CodeBlockAction';
--- a/packages/react-core/src/components/CodeBlock/examples/CodeBlock.md
+++ b/packages/react-core/src/components/CodeBlock/examples/CodeBlock.md
@@ -48,7 +48,7 @@ class BasicCodeBlock extends React.Component {
         this.timer = window.setTimeout(() => {
           this.setState({ copied: false });
           this.timer = null;
-        }, 2000);
+        }, 1000);
       });
     };
   }
@@ -94,11 +94,19 @@ url: https://raw.githubusercontent.com/Azure-Samples/helm-charts/master/docs`;
 }
 ```
 
-<!-- ### Expandable
+### Expandable
 
 ```js
 import React from 'react';
-import { CodeBlock, CodeBlockAction, CodeBlockCode, Button } from '@patternfly/react-core';
+import {
+  CodeBlock,
+  CodeBlockAction,
+  CodeBlockCode,
+  ClipboardCopyButton,
+  ExpandableSection,
+  ExpandableSectionToggle,
+  Button
+} from '@patternfly/react-core';
 import CopyIcon from '@patternfly/react-icons/dist/js/icons/copy-icon';
 import PlayIcon from '@patternfly/react-icons/dist/js/icons/play-icon';
 
@@ -138,7 +146,7 @@ class BasicCodeBlock extends React.Component {
         this.timer = window.setTimeout(() => {
           this.setState({ copied: false });
           this.timer = null;
-        }, 2000);
+        }, 1000);
       });
     };
   }
@@ -151,6 +159,14 @@ kind: HelmChartRepository
 metadata:
 name: azure-sample-repo
 spec:
+connectionConfig:
+url: https://raw.githubusercontent.com/Azure-Samples/helm-charts/master/docs`;
+
+    const code = `apiVersion: helm.openshift.io/v1beta1/
+kind: HelmChartRepository
+metadata:
+name: azure-sample-repo`;
+    const expandedCode = `spec:
 connectionConfig:
 url: https://raw.githubusercontent.com/Azure-Samples/helm-charts/master/docs`;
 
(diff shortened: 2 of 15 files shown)


packages/react-core/src/components/CodeBlock/CodeBlockAction.tsx
```diff
@@ -0,0 +1,21 @@
...
+  /** Additional classes passed to the code block action */
+  className?: string;
+}
+
+export const CodeBlockAction: React.FunctionComponent<CodeBlockActionProps> = ({
+  children = null,
+  className,
+  ...props
+}: CodeBlockActionProps) => (
+  <div className={css('pf-c-code-block__actions-item', className)} {...props}>
```
mcoker, line 16: > Is .pf-c-code-block__actions-item a placeholder class? 

Yep. I suspect we will want that element and it will help us avoid a breaking change and be more consistent with the way we might use it by adding it now as opposed to later.

packages/react-core/src/components/CodeBlock/examples/CodeBlock.md
```diff
@@ -0,0 +1,128 @@
...
+            <PlayIcon />
+          </Button>
+        </CodeBlockAction>
+      </React.Fragment>
+    );
+    return (
+      <CodeBlock actions={actions}>
+        <CodeBlockCode>
+          apiVersion: helm.openshift.io/v1beta1/
+          <br />
```
mcoker, line 46: Ideally we wouldn't need any markup for formatting in this block. `<CodeBlockCode>` is configured to preserve white space, so that a line break in the markup would produce a line break in the browser.

However, due to this, if we add structural elements inside of this block (like the expandable content in [this core example](https://pf4.patternfly.org/components/code-block#expandable)), we will need to remove new lines and space characters so that we don't create unnecessary space like you see in the linked core example.

So basically we need to preserve all space/line breaks within the code a user passes, but remove the leading/trailing space/line breaks around it and around any structure elements we add within `<CodeBlockCode>`.

mcoker (changes requested):
This looks great! Just one bit of feedback around white space.

packages/react-core/src/components/CodeBlock/examples/CodeBlock.md
kmcfaul: Sounds tricky. Is there anything I should update for the moment on my end?

mcoker: Are we able to preserve white space within the children of `<CodeBlockCode>`?

kmcfaul: Yes, conditionally. If you took the whole block of text and sent it in as a formatted/spaced string it would work. If you pass in just the text directly without the `<br />` tags though, it compiles down to a single line.

Something like:
```
const codeStr = `apiVersion: helm.openshift.io/v1beta1/
kind: HelmChartRepository
metadata:
name: azure-sample-repo
spec:
connectionConfig:
url: https://raw.githubusercontent.com/Azure-Samples/helm-charts/master/docs`;
```

I'm not sure how you would break up the formatted string when it comes to expandable sections, maybe you'd go back to markup.

mcoker: @kmcfaul that works great! Based off of these comments, maybe we should update our examples to pass the text in as a formatted string?

https://github.com/patternfly/patternfly/pull/3937#issuecomment-809395629
https://github.com/patternfly/patternfly/pull/3937#issuecomment-809437273

kmcfaul: Sure thing. I will update

kmcfaul: I will also update the detached expandable section example once that goes in.

mcoker (approved):
Looks great! 🤩

mcarrano (commented):
This is looking great @kmcfaul . The only thing I wonder is if we should have the same tooltip behavior here on the Copy action as for Clipboard Copy. Is that what you were expecting also, @lwrigh?

kmcfaul:
I see the buttons as a way to show users how to add actions to CodeBlock (since any amount/type of action can be passed in) more than showcasing specific behaviors, but I can add the copy behavior in if that's desired.

Let me know!

lwrigh:
> This is looking great @kmcfaul . The only thing I wonder is if we should have the same tooltip behavior here on the Copy action as for Clipboard Copy. Is that what you were expecting also, @lwrigh?

If the copy action is attached to the code block I am assuming if would have the same tooltip behavior as the Clipboard Copy component. The actions are customizable for the code block but I would expect the copy action to be a default one.

kmcfaul:
The copy action is not internal to the code block because the content of the codeblock can be an arbitrary structure (in the case of CodeBlock with the detached expandable section), not only a single string. I can still add the behavior, but it would live in the example code, not the component.

mcarrano:
I think that would be OK @kmcfaul . It would be useful for consumers to see the example of how to wire this up, since as @lwrigh stated, we expect that the Copy action will always be present. Thanks.

kmcfaul:
@mcarrano @tlabaj 
Added copy functionality to the codeblock example (and detached expandable section example when that goes in). 

I did export `ClipboardCopyButton` to get the already styled button w/ Tooltip if that is all right to expose to users? I can remove it in favor of manually styling if not.

kmcfaul:
@redallen I'm not seeing the integration error with OverflowMenu locally, any idea what's up? Does the integration have to be run again?

redallen:
Not sure why it's failing like this:
```
<testcase name="OverflowMenu Demo Test Persist OverflowMenu X-Large Dropdown view Verify dropdown menu expanded" time="0" classname="Verify dropdown menu expanded">
  <failure message="Timed out retrying: Expected to find element: `#persist-overflow-menu .pf-c-dropdown`, but never found it." type="AssertionError"><![CDATA[AssertionError: Timed out retrying: Expected to find element: `#persist-overflow-menu .pf-c-dropdown`, but never found it.
    at Context.eval (http://localhost:3000/__cypress/tests?p=cypress/integration/overflowmenu.spec.ts:187:65)]]></failure>
</testcase>
 ```
 
 Might be a flake, I'll rekick.

mcarrano (changes requested):
Thanks for adding the copy functionality @kmcfaul . Can you just make the tooltip a little wider so the word "successfully" will fit on one line? Thanks.

mcoker:
Noticed something else, seems like an edge case. Seems like the tooltip displays the success message for a period of time, and that period of time is longer than the tooltip stays visible. So if you click the copy button and get the success message, after the tooltip disappears, hover over the button again and it still displays the success message, then changes back to the copy message.

I think a better interaction might be that once the tooltip disappears, the tooltip text is reset to the default "copy to clipboard"?


https://user-images.githubusercontent.com/35148959/114938466-c1001d00-9e04-11eb-8490-2e0282836873.mp4

mcoker (dismissed):
Nice!! Left a comment about the tooltip text, but it's good the way it is, too!

nicolethoen (approved):

kmcfaul:
Added in the expandable example.

@mcoker The tooltip text changes based on a short timer, this is how the ClipboardCopy is set up as well. I fiddled with the delay timers to maybe make the timing better.

mcarrano (approved):
Everything looks great @kmcfaul !

tlabaj (approved):
LGTM!

nicolethoen (approved):

jessiehuff (approved):
I couldn't find any specific accessibility requirements for code blocks, but since we're using native html elements here I think it's ok. Perhaps if we find additional information or receive feedback in the future we can make any updates, but for now I think this is good. 👍

Issue #5341, opened 2021-01-19
Create Log viewer component


evwilkin:
Follow-up to https://github.com/patternfly/patternfly-design/issues/955

Core issue: https://github.com/patternfly/patternfly/issues/3811

Create a new Log viewer component

See [logger demo](https://github.com/redhatinsights/logger-demo), [logger toolbar POC](https://github.com/patternfly/patternfly-react/issues/5174)

MVP requirements:

- [ ] Show line numbers - the user can show/hide line numbers associated with log entries via a setting. Default = show line numbers.  (I think it’s just coming directly from the pod’s output? Joseph Caiani?)
- [ ] Download log(s) - the user can download the current log to an external file. If multiple logs are available, a ‘Download all’ option will be available.
- [ ] Launch external viewer - If enabled, allows the log file to be opened in another application (as a separate browser window/tab.
- [ ] Log tailing - the log viewer will optionally support log tailing to display real-time log data. When log tailing is enabled, the user may pause/resume the log display. When real-time data display is enabled, scrolling will be supported within a fixed-height container.
- [ ] Full-screen display - Full screen display will be supported.
- [ ] Switch between log sources - the left-hand dropdown to switch between a few different sources of logs

evwilkin:
@arivepr FYI here is the PatternFly React issue for Log viewer

evwilkin:
@kmcfaul @arivepr an additional request from @jcaianirh was to be able to support custom actions in the toolbar, similar to [this request](https://github.com/patternfly/patternfly-react/pull/5294#issuecomment-756843350) for the Code editor component.  Not sure that this is needed for the MVP requirements, but wanted to document it here case it affects the build from the start.

arivepr:
Update Regarding MVP Status:

- [ ] Show line numbers - Working, matching source correctly
- [ ] Downloading Logs - pending
- [ ] Launch External - Currently opening source in new tab
- [ ] Log Tailing - Adjusted behavior to highlight whatever line is the latest to have been added to the logger output itself, have tested with dummy delays, hasn't been tested in live server responses scenario. (Not sure if this meets requirement)
- [ ] Switch Between Log Sources - pending
- [ ] Full Screen Support - Current implementation is tied up with custom toolbar.
- [ ] Custom Toolbar - Active wip

evwilkin:
FYI - OpenShift line wrapping solution WIP: https://github.com/openshift/console/pull/8683

PR #5657, opened 2021-04-12
Feat/log viewer [WIP]

arivepr:
Toward https://github.com/patternfly/patternfly-react/issues/5341

_**React-log-viewer**_
Adding component for outputting logs in a formatted string. 
This component is being developed to satisfy:
https://marvelapp.com/prototype/bia3j2d/screen/75984271

_**Current [WIP] Items**_

1. Patternfly rebuild of the css (currently the individual rows are disconnected from the custom styling defined in files (not in PF standards)


packages/react-log-viewer/src/LogViewer/LoggerContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+  setRowInFocus?: (index: number) => void;
+  setSearchedInput: (input: string) => void;
+  setFoundWordIndex?: (index: number) => void;
+  setHighlightedRowIndexes: (indexes: Array<number>) => void;
+  setSearchedWordIndexes: (indexes: Array<number>) => void;
+};
+
+const LoggerContext = createContext<LoggerContextInterface | null>(null);
+
+export const LoggerContextProvider = ( children:React.ReactNode) => {
```
redallen, line 26: Change to `({ children:React.ReactNode })`

redallen (changes requested):
Once this is building I'll give you a full review.

packages/react-log-viewer/src/LogViewer/LoggerContext.tsx
arivepr: it's currently up because need some eyes to get it building. 

The suggested children:ReactNode causes the issue in the picture.

redallen: Unfortunately only very keen eyes can get this building without checking it out locally. You can schedule time with a Typescript expert (@kmcfaul, @jschuler , or me) who can peer program with you to get it there.

kmcfaul: Did you update the syntax to include the brackets?

You might also try changing the type to `React.ReactElement`, or defining the states and `LoggerContext.Provider` inside `LogViewer` instead of using the exported const.

arivepr: Thanks for the advice/offer @redallen. I had also tried with the brackets prior to the pr but the errors were different. `{children}:any` seemed to calm it all down and allow me to build, but is it a viable option or should I still define the context in the root logger file?

packages/react-log-viewer/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "homepage": "https://github.com/patternfly/patternfly-react/tree/master/packages/react-log-viewer#readme",
+  "scripts": {
+    "clean": "rimraf dist"
+  },
+  "dependencies": {
+    "@patternfly/react-core": "^4.106.4",
+    "@patternfly/react-icons": "^4.9.10",
+    "@patternfly/react-styles": "^4.9.5",
+    "@types/react-window": "^1.8.1",
+    "classnames": "^2.2.6",
```
redallen, line 36: Use `import { css } from '@patternfly/react-styles';` instead of `classnames`.

packages/react-log-viewer/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "memoize-one": "^5.1.0",
```
redallen, line 37: Use [`React.useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo) instead.

packages/react-log-viewer/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "react-window": "^1.8.6",
+    "react-window-infinite-loader": "^1.0.7",
+    "tslib": "1.13.0"
+  },
+  "peerDependencies": {
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0",
+    "react-monaco-editor": "^0.41.2"
```
redallen, line 45: Remove `react-monaco-editor` from peer dependencies. Remember to run `yarn install` to update the `yarn.lock` after changing `package.json`.

packages/react-log-viewer/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "tag": "prerelease"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/patternfly/patternfly-react.git"
+  },
+  "keywords": [
+    "react",
+    "patternfly",
+    "code-editor"
```
redallen, line 20: Change to `log-viewer`.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /* String that wil be processed into the logger for output */
+  data?: string[];
+  /* Flag for including in the toolbar */
+  // includesFullscreen?: boolean;
+  /* Flag for including in the toolbar */
+  // includesExternalOpen?: boolean;
+  /* Flag for including in the toolbar */
+  // includesPlayPause?: boolean;
+  /* Flag for including in the toolbar */
+  // includesDownload?: boolean;
```
redallen, line 18: Remove unused code.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /* Grouping for custom actions that the user can provide. This is where they would reside */
+  customKabob?: React.ReactNode | React.ReactNode[];
```
redallen, line 20: Change to just `React.ReactNode` (which already includes arrays).

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /* Titles users can provide for their Data Sources */
+  dataSourceTitles?: string[] | null[]; // Array<string | null | undefined>;
+  /* Link to be opened via router whenever a user wants to open the component on new tab */
+  externalOpenRouteLink?: string;
+  /* User defined method for handling fullscreen action */
+  onHandleFullScreen?: () => void;
+  /* User defined function to pausing playing information flow into the component */
+  onHandlePlayPause?: () => void;
+  /* This describes custom items devs can add to the integrated search/toolbar */
+  customToolbarActions?: () => React.ReactNode | React.ReactNode[];
```
redallen, line 36: Change to just React.ReactNode (which already includes arrays).

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  // includesExternalOpen = false,
+  // includesPlayPause = false,
+  // includesDownload = false,
+  // customToolbar = false,
+  dataSourceTitles = ['Default']
+  // className = '',
+  // customToolbarActions = () => {},
+  // onOpenExternally = () => {},
+  // onDownloading = () => {},
+  // onHandleFullScreen = () => {},
```
redallen, line 56: Remove unused code.

packages/react-log-viewer/src/LogViewer/examples/data/dataTest.yml
```diff
@@ -0,0 +1,26 @@
+{
```
redallen, line 1: Remove this file.

packages/react-log-viewer/src/LogViewer/examples/LogViewer.md
```diff
@@ -0,0 +1,41 @@
+---
+id: Log viewer
+beta: true
+section: components
+---
```
redallen, line 5: Add `propComponents: [LogViewer]`

packages/react-log-viewer/src/LogViewer/logger.tsx
```diff
@@ -0,0 +1,117 @@
...
+      setRowInFocus,
+      searchedInput,
+      highlightedRowIndexes,
+      setHighlightedRowIndexes,
+      currentDataSource,
+      setSearchedWordIndexes
+    } = loggerContext;
+    const DEFAULT_SEARCH_INDEX = 0;
+    const loggerRef = React.createRef<any>();
+    const dataToRender = createLoggerDataItem(
```
redallen, line 44: Can't you just use `React.useMemo` instead?

packages/react-log-viewer/src/LogViewer/loggerToolbar.tsx
```diff
@@ -0,0 +1,177 @@
...
+      return null;
+    }
+
+    setCurrentSearchItemCount(adjustedSearchCount);
+    setFoundWordIndex(--temp);
+    scrollToRow(searchedWordIndexes[--temp]);
+  };
+
+  /* Way to make this logic much simpler? */
+  const toolbarItems = (): React.ReactNode => (
```
redallen, line 112: Make this an expression like `const toolbarItems = <>{stuff}</>` instead of a function like `const toolbarItems = () => {}`.

It's more performant for state and simpler.

packages/react-styles/src/css/components/LogViewer/loggerToolbar.styles.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.toolbar__searchbar {
+    width: 500px;
+}
+
+.searchbar__btn {
+    width: 44px;
+    height: 36px;
+}
```
redallen, line 21: Name all your CSS selectors according to BEM standards. You can use [AccessConsoles](https://github.com/patternfly/patternfly-react/tree/master/packages/react-styles/src/css/components/Consoles) as an example. Everything should be scoped to a `.pf-c-logger` so as not to collide with consumer class names.

packages/react-log-viewer/src/LogViewer/loggerToolbar.tsx
```diff
@@ -0,0 +1,177 @@
...
+            onChange={input => setSearchedInput(input)}
+            resultsCount={`${currentSearchItemCount} / ${searchedWordIndexes.length}`}
+            className={styles.toolbarSearchbar}
+          />
+        </ToolbarGroup>
+      </ToolbarItem>
+      <ToolbarGroup alignment={{ default: 'alignRight' }}>
+        <ToolbarItem>
+          {includesPlay && (
+            <Button variant="plain" className="searchbar__btn">
```
redallen, line 140: Replace all your `className="hardcoded-string"` with `import styles from '@patternfly/react-styles/css/path/to/your/style.css';` and `className={styles.hardodedString}`.

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
```diff
@@ -0,0 +1,72 @@
+export const isArrayOfString = (array: string[]) => {
+  for (const str in array) {
+    if (typeof str !== 'string') {
+      console.log(`Sadly ${str} is not a string :c`); // eslint-disable-line
```
redallen, line 4: Remove this line.

packages/react-log-viewer/src/LogViewer/logger.tsx
```diff
@@ -0,0 +1,117 @@
...
+    );
+
+    /* Parsing depending on whether the data given is a string or a full array of strings. */
+    useEffect(() => {
+      if (typeof data === 'string') {
+        setParsedData(parseConsoleOutput(data));
+      } else if (isArrayOfString(data)) {
+        setDataSourcesAmount(data.length);
+        setParsedData(parseConsoleOutput(data[currentDataSource]));
+      }
```
redallen, line 59: Handle error case here with `console.error`.

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    if (keywordIndexPosition !== -1) {
+      searchResults.push(rowIndexCounter);
+    }
+
+    rowIndexCounter++;
+  }
+
+  if (searchResults.length > 0) {
+    return [...searchResults];
```
redallen, line 34: Why are you cloning `searchResults` before returning it?

redallen (changes requested):
Lots of cleanup that can be done. Haven't played around with the API yet.

Not sure if this weird scrollbar becoming out of sync with my cursor is avoidable?
scrollbar

Edit: Here's another oddity with line spacing:
image

nicolethoen:
For reference, here is OpenShift's log viewer using line wrapping. Not sure if it's helpful.
https://github.com/openshift/console/pull/8683

arivepr:
> Lots of cleanup that can be done. Haven't played around with the API yet.
> 
> Not sure if this weird scrollbar becoming out of sync with my cursor is avoidable?
> scrollbar
> 
> Edit: Here's another oddity with line spacing:
> image

So when it comes to the first, my theory is that it has to do with how it calculates the individual row sizes as it's rendering them. You can see the same thing happen in the api docs for varaible sized rows [here](https://react-window.vercel.app/#/examples/grid/variable-size)

Regarding the line spacing, it's the issue i've been mentioning regarding styling being disconnected from the individual rows. Broke when I changed how i was importing and changing to css from scss. Still trying to figure that one out, in reality, the intended row spacing is: 
Screen Shot 2021-04-21 at 9 08 09 AM

packages/react-log-viewer/src/LogViewer/logger.tsx
arivepr: I was following the documentation on memoizing items going into the individual row renders, but I haven't found a way to memoize an object containing fields(duh) rather than memoizing a whole function with useMemo. 

If you know how to cook something like that up I could happily switch it, just haven't found the answer.

Would using memoize-one's helper function be a nono?

redallen: It's not a nono, it's just preferred to not use unnecessary 3rd party deps. Is this not the same?
```js
const dataToRender = React.useMemo(() => ({
    parsedData,
    loggerRef,
    rowInFocus,
    highlightedRowIndexes,
    setHighlightedRowIndexes
}), [
    parsedData,
    loggerRef,
    rowInFocus,
    highlightedRowIndexes,
    setHighlightedRowIndexes
]);

**Comment by mattnolting** [comment]
@arivepr I'll look at the toolbar

File: packages/react-log-viewer/src/LogViewer/loggerToolbar.tsx [hunk 19]
```diff
@@ -0,0 +1,177 @@
...
+  includesFullScreen?: boolean;
+  includesLaunchExternal?: boolean;
+  includesDownload?: boolean;
+  dataSourcesAmount?: number;
+  dataSourceTitles?: string[] | null[];
+  customToolbarActions?: () => React.ReactNode | React.ReactNode[];
+  scrollToRow: (searchedRow: number) => void;
+}
+
+const LoggerToolbar: React.FC<LoggerToolbarProps> = ({
```
nicolethoen, line 24: Same with FC here

packages/react-log-viewer/src/LogViewer/loggerToolbar.tsx
```diff
@@ -0,0 +1,177 @@
...
+      <ToolbarGroup alignment={{ default: 'alignRight' }}>
+        <ToolbarItem>
+          {includesPlay && (
+            <Button variant="plain" className={css(styles.searchbarBtn)}>
+              <PlayIcon />
+            </Button>
+          )}
+        </ToolbarItem>
+        <ToolbarItem>
+          {includesFullScreen && (
```
nicolethoen, line 146: you should only render the `ToolbarItem` if it's contents are going to be rendered. You don't want empty toolbaritems, but most of these wont be necessary for the initial PR.

packages/react-log-viewer/src/LogViewer/loggerToolbar.tsx
```diff
@@ -0,0 +1,177 @@
...
+    setCurrentSearchItemCount(adjustedSearchCount);
+    setFoundWordIndex(--temp);
+    scrollToRow(searchedWordIndexes[--temp]);
+  };
+
+  /* Way to make this logic much simpler? */
+  const toolbarItems = (
+    <>
+      <ToolbarItem>
+        <ToolbarDropdown
```
nicolethoen, line 115: I think the consumer should handle which datasource is being displayed.

packages/react-log-viewer/src/LogViewer/loggerToolbar.tsx
```diff
@@ -0,0 +1,177 @@
...
+  includesDownload?: boolean;
+  dataSourcesAmount?: number;
+  dataSourceTitles?: string[] | null[];
+  customToolbarActions?: () => React.ReactNode | React.ReactNode[];
+  scrollToRow: (searchedRow: number) => void;
+}
+
+const LoggerToolbar: React.FC<LoggerToolbarProps> = ({
+  scrollToRow,
+  // customToolbarActions,
```
nicolethoen, line 26: you are not rendering any customToolbarActions? It's not necessary for this PR anyway, so we can remove this prop and any references to it in other components for now I think its okay if the only item in the toolbar with this PR is the search input

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  customToolbarActions?: () => React.ReactNode;
+  /* Dev defined method for downloading the output data from logger */
+}
+
+export const LogViewer: React.FC<LoggerProps> = ({
+  data = [''],
+  hasSearchbar = true,
+  dataSourceTitles = ['Default']
+}) => (
+  <LoggerContextProvider>
```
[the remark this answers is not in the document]
nicolethoen: and possibly even combine it with `Logger`, too. 
I'm happy to help you do that

jessiehuff (commented):
On the accessibility side I have a few thoughts:

- I would expect the log to be scrollable by keyboard. I think you should be able to add this by adding a `tabIndex="0"` to the containing div. (However, I'm curious if we ever expect interactive content to be inside the logger or will it always be static?)

- From VO, it reads out everything very literally. It reads out the dropdown and search input then it reads each line "1 {first line content}", etc. Is this the final semantics? Do you know if core is planning on making any changes to use more semantic elements @mattnolting? When looking into this type of component before, I wondered if [`role="log"`](https://www.digitala11y.com/log-role/) would make the most sense here with `<li>'s` and potentially an `aria-live` region if we're planning on this dynamically being rendered as new logs occur (similar to these examples: [W3 example](https://w3c.github.io/wcag/working-examples/aria-role-log/serverlog.html), [rip tutorial example](https://riptutorial.com/html/example/9165/role--log-), [deque's live region playground](https://dequeuniversity.com/library/aria/liveregion-playground)). However, I think it probably depends on the final structure we're planning on using here and what interactions we expect.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { css } from '@patternfly/react-styles';
+import { LoggerRow } from './LogViewerRow';
+import { LogViewerToolbar } from './LogViewerToolbar';
+import { LOGGER_HEIGHT, DEFAULT_FOCUS, DEFAULT_SEARCH_INDEX } from './utils/constants';
+import { searchForKeyword, parseConsoleOutput } from './utils/utils';
+import { VariableSizeList as List, areEqual } from 'react-window';
+import styles from '@patternfly/react-styles/css/components/LogViewer/base';
+import '@patternfly/react-core/dist/styles/base.css';
+
+interface LoggerProps {
```
nicolethoen, line 12: This needs to be renamed.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    return (
+      <LogViewerContext.Provider
+        value={{
+          parsedData
+        }}
+      >
+        <div className={css(styles.insCLogger)}>
+          <LogViewerToolbar
```
nicolethoen, line 78: should this be wrapped in your `hasToolbar` prop?

packages/react-log-viewer/src/LogViewer/LogViewerToolbar.tsx
```diff
@@ -0,0 +1,108 @@
...
+            onNextClick={() => handleNextSearchItem()}
+            onPreviousClick={() => handlePrevSearchItem()}
+            onClear={() => handleClear()}
+            onChange={input => setSearchedInput(input)}
+            resultsCount={`${currentSearchItemCount} / ${searchedWordIndexes.length}`}
+            className={styles.toolbarSearchbar}
+          />
+        </ToolbarGroup>
+      </ToolbarItem>
+      {customControls && <ToolbarItem>{customControls}</ToolbarItem>}
```
nicolethoen, line 98: Ideally, I think if all your controls should be `ToolbarItems` then the wrapping of the controls should be done in the `CustomControl` component

packages/react-log-viewer/src/LogViewer/examples/Logger.md
```diff
@@ -0,0 +1,64 @@
+---
+id: Logger
```
nicolethoen, line 2: you should change these reference and the file name to LogViewer

nicolethoen: or remove them? what is the difference between the two markdown files you have here?

packages/react-log-viewer/src/LogViewer/examples/LogViewer.md
```diff
@@ -0,0 +1,107 @@
...
+id: Log viewer
+beta: true
+section: components
+propComponents: [LogViewer]
```
nicolethoen, line 5: You should add `LogViewerCustomControl` to `propComponents`. The components that the consumer will be using should be listed here so the docs contain all the props and comments associated with all props so they know how yo use them.

packages/react-log-viewer/src/LogViewer/LogViewerToolbar.tsx
```diff
@@ -0,0 +1,108 @@
+import React, { useState, useEffect } from 'react';
+import { NUMBER_INDEX_DELTA } from './utils/constants';
+import { SearchInput, Toolbar, ToolbarItem, ToolbarContent, ToolbarGroup } from '@patternfly/react-core';
+import styles from '@patternfly/react-styles/css/components/LogViewer/loggerToolbar.styles';
+
+export interface LoggerToolbarProps extends React.HTMLProps<HTMLDivElement> {
+  hasToolbar?: boolean;
```
nicolethoen, line 7: We will need to add comments to describe what each of these props does. In all component files.

packages/react-log-viewer/src/LogViewer/toolbarDropdown.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
```
nicolethoen, line 1: This file needs to be removed

packages/react-log-viewer/src/LogViewer/LogViewerToolbar.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setFoundWordIndex(--temp);
+    scrollToRow(searchedWordIndexes[--temp]);
+  };
+
+  const toolbarItems = (
+    <>
+      <ToolbarItem>
+        <ToolbarGroup>
+          <SearchInput
+            placeholder="Search"
```
tlabaj, line 87: we need a prop for this placeholder text so we can localize it.

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return negativeResults;
+  }
+};
+
+export const parseConsoleOutput = (data: string) => {
+  const stringToSplitWith = '\n';
+  const stringSplitting = data.toString();
+  const cleanString = stringSplitting.split(stringToSplitWith);
+
+  // if (parseData) {
```
tlabaj, line 47: this can be removed

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const inputValidation = (inputString: string) => {
+  if (inputString.length >= 3) {
+    return true;
+  } else {
+    return false;
+  }
+};
+
+export const initStringTitles = (itemAmount: number): string[] => {
```
tlabaj, line 62: Is this used anywhere?  If not,, it can be removed.  If it is used, we can nit have hard coded strings in the component.  e need to be able to change them for localization.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
tlabaj: renamed to `LogViewerProps`

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { LoggerRow } from './LogViewerRow';
+import { LogViewerToolbar } from './LogViewerToolbar';
+import { LOGGER_HEIGHT, DEFAULT_FOCUS, DEFAULT_SEARCH_INDEX } from './utils/constants';
+import { searchForKeyword, parseConsoleOutput } from './utils/utils';
+import { VariableSizeList as List, areEqual } from 'react-window';
+import styles from '@patternfly/react-styles/css/components/LogViewer/base';
+import '@patternfly/react-core/dist/styles/base.css';
+
+interface LoggerProps {
+  /* String data being sent by the consumer*/
```
tlabaj, line 13: Comments are not formatted properly here.  Should be: 
`/** prop description */`

packages/react-log-viewer/src/LogViewer/LogViewerCustomControl.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Button } from '@patternfly/react-core';
+import { ButtonProps } from '@patternfly/react-core/src';
+
+export interface LoggerCustomControlProps extends Omit<ButtonProps, 'onClick'> {
+  /* Icon for the user defined button */
+  children?: React.ReactNode;
+  /* Callback func. with exposure for the data. */
+  onClick?: (event?: any, value?: string[]) => void;
+  /* Toggle to turn button visibility on toolbar on and off */
+  isVisible?: boolean;
```
tlabaj, line 12: this prop is not needed.  The `LogViewerToolbar`  will  control visibility.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,108 @@
...
+  data?: string;
+  /* Consumer may turn off the visibility on the toolbar */
+  hasToolbar?: boolean;
+  /* Consumers may define their own controls to feed to the toolbar for custom functionality. */
+  customControls?: React.ReactNode | null;
```
tlabaj, line 18: ReactNode includes null,  you can remove the ` | null`

packages/react-log-viewer/src/LogViewer/LogViewerRow.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState, useEffect, memo } from 'react';
+import { LOGGER_LINE_NUMBER_INDEX_DELTA } from './utils/constants';
+import classNames from 'classnames';
+import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/LogViewer/loggerRow.styles';
+
+interface LoggerRowProps extends React.Props<HTMLElement> {
```
tlabaj, line 7: Can we rename this to LogViewerRowProps?

packages/react-log-viewer/src/LogViewer/LogViewerRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+  style?: React.CSSProperties;
+  data: {
+    parsedData: any;
+    rowInFocus: number;
+    highlightedRowIndexes: number[];
+    setHighlightedRowIndexes: (indexes: number[]) => void;
+  };
+}
+
+export const LoggerRow: React.FunctionComponent<LoggerRowProps> = memo(({ index, style, data }) => {
```
tlabaj, line 18: Can we rename to LogViewerRow

mcarrano (changes requested):
As I see this currently, there are two major UI issues that need to be addressed.

* the way that the rows format is not really responsive since it seems like the rows are set to a fixed height, but this doesn't really work with line wrapping.
* The lack of text formatting in the log entries make them difficult to read.
* The Find feature does not behave as intended. It works more like a filter vs a find.

For more guidance on what this should look like, see the design document here: https://marvelapp.com/prototype/bia3j2d

jcaianirh:
@nicolethoen @evwilkin @arivepr This was the additional log viewer content that came up recently for the console. WIP design attached to the jira story: https://issues.redhat.com/browse/CONSOLE-2223

mcarrano:
The search behavior is better, but still not quite right. All instances of a match should be highlighted with the current instance (per the navigation buttons in the search input) having the darker highlight.

packages/react-log-viewer/src/LogViewer/Styles/log-viewer.css
```diff
@@ -0,0 +1,148 @@
+.pf-c-log-viewer {
```
kmcfaul, line 1: Is this local `Styles` folder still necessary? It should be pulling styling from the updated core now.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,157 @@
...
+    };
+
+    return (
+      <LogViewerContext.Provider
+        value={{
+          parsedData
+        }}
+      >
+        <div className={css(styles.logViewer) + (hasLineNumbers ? ' pf-m-line-numbers' : '')}>
+          <LogViewerToolbar
```
kmcfaul, line 122: Should be wrapped in `hasToolbar` still. Or is this to change later on in a followup?

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,157 @@
...
+        setRowInFocus(DEFAULT_FOCUS);
+        // setCurrentSearchedItemCount(DEFAULT_INDEX);
+        // setSearchedWordIndexes([]);
+      }
+    }, [searchedInput]);
+
+    const scrollToRow = (searchedRowIndex: number) => {
+      setRowInFocus(searchedRowIndex);
+      loggerRef.current.scrollToItem(searchedRowIndex, 'center');
+      return true;
```
kmcfaul, line 103: Not sure you need to return anything here.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,157 @@
...
+    };
+
+    const loggerWidth = loggerRef.current ? loggerRef.current.clientWidth : 800;
+    const guessRowHeight = (rowIndex: number) => {
+      const rowText = parsedData[rowIndex];
+      const textWidth = getTextWidth(rowText, 'Liberation Mono');
```
kmcfaul, line 109: @nicolethoen Do you think it's worth making the font a prop, in case styling overrides happen? Or possibly the whole guessRowHeight function.

packages/react-log-viewer/src/LogViewer/LogViewerCustomControl.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useContext } from 'react';
+import { LogViewerContext } from './LogViewerContext';
+import { Button, ToolbarItem } from '@patternfly/react-core';
+import { ButtonProps } from '@patternfly/react-core/src';
+
+export interface LoggerCustomControlProps extends Omit<ButtonProps, 'onClick'> {
+  /* Icon for the user defined button */
```
kmcfaul, line 7: Should still be updated to use `/** text */` to have the docs pick up the descriptions.

packages/react-log-viewer/src/LogViewer/LogViewerRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+    setIsHiglighted(!isHiglighted);
+  };
+
+  /** Helper function for applying the correct styling for styling rows containing searched keywords */
+  const handleRowHighlight = (index: number): string => {
+    if (searchedWordIndexes.includes(index)) {
+      if (rowInFocus === index) {
+        return css(styles.logViewerListItem) + ' pf-m-current';
+      }
+      return css(styles.logViewerListItem) + ' pf-m-match';
```
kmcfaul, line 59: Should be able to use `css(styles.logViewerListItem, styles.modifiers.match)` and `css(styles.logViewerListItem, styles.modifiers.current)` here instead of hardcoded strings.

packages/react-log-viewer/src/LogViewer/utils/constants.ts
```diff
@@ -0,0 +1,20 @@
+// Logger Dimensions
+export const LOGGER_HEIGHT: number = 600;
```
kmcfaul, line 2: Height, columns amount, all IDs, column widths, row height (and/or the row height function), min search input length might all be better off as props instead of constants. Anything a user may want to change on demand.

Width already is a prop so I'm not sure this constant gets used, same for row height which right now is calculated.

mcarrano (dismissed):
The search works as expected now. Thanks @arivepr .

I'm good to merge this. We can consider what further follow-up is needed separately.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+      if (adjustedSearchedInput !== '' && adjustedSearchedInput.length < 3) {
+        setRowInFocus(DEFAULT_FOCUS);
+        setCurrentSearchedItemCount(DEFAULT_INDEX);
+        setSearchedWordIndexes([]);
+      }
+
+      if (adjustedSearchedInput === '') {
+        setRowInFocus(DEFAULT_FOCUS);
+        // setCurrentSearchedItemCount(DEFAULT_INDEX);
```
tlabaj, line 101: Is this a placeholder or can we remove these comments

tlabaj (changes requested):

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,162 @@
+import React, { useState, useEffect, memo } from 'react';
+import { LogViewerContext } from './LogViewerContext';
+import { css } from '@patternfly/react-styles';
+import { LogViewerRow } from './LogViewerRow';
+import { LogViewerToolbar } from './LogViewerToolbar';
+import { DEFAULT_FOCUS, DEFAULT_SEARCH_INDEX, DEFAULT_INDEX } from './utils/constants';
+import { searchForKeyword, parseConsoleOutput, escapeString } from './utils/utils';
+import { VariableSizeList as List, areEqual } from '../react-window';
+import styles from '@patternfly/react-styles/css/components/LogViewer/log-viewer';
+import '@patternfly/react-core/dist/styles/base.css';
```
kmcfaul, line 10: Shouldn't need this import

arivepr: do you mean base.css?

kmcfaul (approved):
Probably good to merge for now, we can open some follow ups

nicolethoen (approved):

tlabaj (approved):

redallen (approved):
This won't work in non fixed-width use cases, but we can clean that up later.

PR #5666, opened 2021-04-14
feat(DatePicker): Add date and time range demo

### LINKED ISSUE #4737: Build date+time range picker

nicolethoen:
**What**: Closes #4737

@redallen Do you know why the rangeStart value is not updating on the second Date picker in my new demo? 

After I figure this out, I just need to make sure the error message is displayed at the right time - other than that, it's ready for review

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
@@ -34,14 +34,13 @@ export const makeTimeOptions = (stepMinutes: number, hour12: boolean, delimiter:
   return res;
 };
 
-export const parseTime = (time: string, timeRegex: RegExp, delimiter: string, is12Hour: boolean) => {
-  time = time.trim();
+export const parseTime = (time: string | Date, timeRegex: RegExp, delimiter: string, is12Hour: boolean): string => {
   const date = new Date(time);
 
   // if default time is a ISO 8601 formatted date string, we parse it to hh:mm(am/pm) format
-  if (!isNaN(date.getDate()) && time.includes('T')) {
+  if (!isNaN(date.getDate()) && (time instanceof Date || time.includes('T'))) {
     const hours = is12Hour
-      ? `${date.getHours() > 11 ? date.getHours() - 12 : date.getHours()}`
+      ? `${date.getHours() > 12 ? date.getHours() - 12 : date.getHours()}`
       : `${date.getHours()}`.padStart(2, '0');
     const minutes = `${date.getMinutes()}`.padStart(2, '0');
     let ampm = '';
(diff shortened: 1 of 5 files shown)


redallen:
@nicolethoen It's hard to debug without a preview, but is `from` updated correctly and a valid date object?

nicolethoen:
@redallen Lol I'll fix the build and get a preview - but I believe it is

mcarrano:
@nicolethoen I took a quick look at this. Visually it looks good and seems to behave as expected as long as I use the mouse to select dates and time. Seem to be some bugs when I attempt to enter dates and/or times from the keyboard. Here's some of the states I was able to get into:

Screen Shot 2021-04-14 at 11 07 31 AM

This seems like a problem with validation where it thinks the "to" date is before the "from" date.

Screen Shot 2021-04-14 at 11 09 52 AM

 Got this when I tried to just enter from the keyboard. Seems like I can't activate the "from" fields unless I use the mouse.

packages/react-core/src/demos/DatePicker.md
```diff
@@ -54,3 +54,90 @@ DateRangePicker = () => {
...
+      const updatedToDate = new Date(to);
+      updatedToDate.setHours(hour);
+      updatedToDate.setMinutes(minute);
+      console.log(updatedToDate);
+      setTo(updatedToDate);
+    }
+  };
+
+  return (
+    <Flex direction={{default: 'column', lg: 'row'}}>
```
mcoker, line 109: 👍

mcoker (approved):
LGTM! One note is that if you pull in the latest version of core, this issue with the button focus ring being hidden behind the element on the right should be fixed.

Screen Shot 2021-04-14 at 2 51 26 PM

jessiehuff (dismissed):
LGTM!

kmcfaul (approved):
LGTM!

mcoker (dismissed):

mcarrano (changes requested):
I still seem to be having trouble getting the "from" fields to enable after I enter dates and times in the "to" fields. Here I picked a date and time using the mouse and you can see that second set of fields remains disabled>

Screen Shot 2021-04-16 at 3 12 55 PM

I can get them to enable if I got back and edit using the keyboard.

packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
```diff
@@ -34,14 +34,13 @@ export const makeTimeOptions = (stepMinutes: number, hour12: boolean, delimiter:
...
 };
 
-export const parseTime = (time: string, timeRegex: RegExp, delimiter: string, is12Hour: boolean) => {
-  time = time.trim();
+export const parseTime = (time: string | Date, timeRegex: RegExp, delimiter: string, is12Hour: boolean): string => {
   const date = new Date(time);
 
   // if default time is a ISO 8601 formatted date string, we parse it to hh:mm(am/pm) format
-  if (!isNaN(date.getDate()) && time.includes('T')) {
+  if (!isNaN(date.getDate()) && (time instanceof Date || time.includes('T'))) {
```
redallen, line 41: Why the `time.includes('T')` check?

packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
```diff
@@ -90,3 +92,16 @@ export const validateTime = (time: string, timeRegex: RegExp, delimiter: string,
   // empty string is valid
   return time === '' || (timeRegex.test(time) && validHours);
 };
+
+export const getHours = (time: string, delimiter: string) => {
+  const parts = time.match(new RegExp(`(\\d+)${delimiter}(\\d+)(\\s?)(\\w*)`));
```
redallen, line 97: Can you add capture groups to `TimePicker.tsx:getRegExp` and pass that to `getHours` and `getMinutes` instead? No need for yet another regex.

redallen (changes requested):
Some library code cleanliness

packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
nicolethoen: You may have a better solution, but it's because ISO 8601 strings which include a time always have a T in them to delineate the time. 

if you pass any string of numbers to a new Date(<number>) it will return a valid date that is not reflective of the intended time. It basically is checking that the passed in time is either a valid Date object, or it is a string of numbers that is written in ISO 8601 format. In either of those cases, it's parsed using javascript date functions. Otherwise, it's parsed as a string.

mcarrano (approved):
Looks good now from a UX perspective. Thanks @nicolethoen

packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
redallen: Ah I didn't notice you changed the function signature to `time: string | Date`.

redallen (approved):
Thanks @nicolethoen !

kmcfaul (approved):

Issue #5529, opened 2021-03-09
bug(Slider): dragging the thumb on a continuous slider does not trigger the onValueChanged


KKoukiou:
Moving the slider with the keyboard will update the value as expected.

michpetrov:
The current implementation assumes that when steps are defined the slider is also discrete and vice versa.

A related issue is that the slider doesn't snap to 0 due to the same checks.

PR #5669, opened 2021-04-14
fix, feat(Slider): Update Slider API to be more like the range input API

### LINKED ISSUE #5464: Slider - use standard input element
I found several issues with the [Slider](https://www.patternfly.org/v4/components/slider) component.

1. Use standard `type="range"` input variant

Why a custom div is used when sliders are supported by default in HTML? The component should also support the same API as the input element (`step`, `max`, `min`, `value`).

2. `isDiscrete={false}` variant does not trigger change events when used together with `steps`

You can see it in the first example - if you remove the `isDiscrete` prop, then the value is never changed.

3. `onChange` and `value` should serve as the main props to be consistent with all other form elements

There should be no `currentValue` vs `inputValue`, `onChange` vs `onValueChange`. It's not clear at all.

cc @Hyperkid123

### LINKED ISSUE #5564: TimePicker: no way to know if they provided time is valid
When using the component in forms, we need to block submission of the form if the TimePicker input is invalid. We currently do that by disabling the component's validation, by setting invalidFormatErrorMessage='' and doing the validation on our side on submit.  Ideally we would like to let the component do the validation, but have a way to know if the inserted time is valid.

I suggest that we add another parameter to the onChange, related to the validity.

tlabaj:
**What**: Closes #5464, #5529 

- Added `min`, `max` and `step` props to match the HTML range input API 
- changed prop name `steps` to `customSteps` to be less confusing.
- renamed `currentValue` prop to `value` to match  HTML rang input
- Added `showTicks` prop so the consumer can choose to show ticks when they use the `min`, `max` and `step` props
- Removed the `isDiscrete ` prop and added `areCustomStepsContinuous` prop.  A slider with out custom steps is continuous by default.    A slider with custom steps is discrete unless the `areCustomStepsContinuous` is set to true. This allows the consumer to add labels to a continuous slider.
- Updated Slider to have only one `onChange` callback
- Fixed onChange not getting called for continuous sliders.

### DIFF
--- a/packages/react-core/src/components/Slider/Slider.tsx
+++ b/packages/react-core/src/components/Slider/Slider.tsx
@@ -136,14 +136,6 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
   };
 
   const handleThumbDragEnd = () => {
-    if (snapValue !== undefined && customSteps) {
-      thumbRef.current.style.setProperty('--pf-c-slider--value', `${snapValue}%`);
-      setValue(snapValue);
-      if (onChange) {
-        onChange(snapValue);
-      }
-    }
-
     document.removeEventListener('mousemove', callbackThumbMove);
     document.removeEventListener('mouseup', callbackThumbUp);
     document.removeEventListener('touchmove', callbackThumbMove);
@@ -216,9 +208,6 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
       snapValue = Math.round((newValue - min) / step) * step + min;
       thumbRef.current.style.setProperty('--pf-c-slider--value', `${snapValue}%`);
       setValue(snapValue);
-      if (onChange) {
-        onChange(snapValue);
-      }
     }
 
     /* If custom steps are discrete, snap to closest step value */
@@ -234,13 +223,16 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
           snapValue = customSteps[stepIndex].value;
         }
       }
+      setValue(snapValue);
     }
 
     // Call onchange callback
     if (onChange) {
-      if (areCustomStepsContinuous && customSteps) {
-        onChange(newValue);
-      }
+      if (snapValue !== undefined) {
+        onChange(snapValue);
+      } else {
+        onChange(newValue)
+      } 
     }
   };
 
--- a/packages/react-core/src/components/Slider/Slider.tsx
+++ b/packages/react-core/src/components/Slider/Slider.tsx
@@ -136,7 +136,7 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
   };
 
   const handleThumbDragEnd = () => {
-    if (snapValue !== undefined) {
+    if (snapValue !== undefined && customSteps) {
       thumbRef.current.style.setProperty('--pf-c-slider--value', `${snapValue}%`);
       setValue(snapValue);
       if (onChange) {
@@ -211,6 +211,16 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
     const newValue = (newPercentage * (max - min)) / 100 + min;
     setValue(newValue);
 
+    if (!customSteps) {
+      // snap to new value if not custom steps
+      snapValue = Math.round((newValue - min) / step) * step + min;
+      thumbRef.current.style.setProperty('--pf-c-slider--value', `${snapValue}%`);
+      setValue(snapValue);
+      if (onChange) {
+        onChange(snapValue);
+      }
+    }
+
     /* If custom steps are discrete, snap to closest step value */
     if (!areCustomStepsContinuous && customSteps) {
       const stepIndex = customSteps.findIndex(stepObj => stepObj.value >= newPercentage);
@@ -224,14 +234,11 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
           snapValue = customSteps[stepIndex].value;
         }
       }
-    } else if (!customSteps && step !== 1) {
-      // snap to new value if not custom steps and step is not 1
-      snapValue = Math.round((newValue - min) / step) * step + min;
     }
 
-    // Call value change callback unless custom steps are discrete or no custom steps and step is 1.  The value will snap after the thumb is released.
+    // Call onchange callback
     if (onChange) {
-      if ((areCustomStepsContinuous && customSteps) || (!customSteps && step === 1)) {
+      if (areCustomStepsContinuous && customSteps) {
         onChange(newValue);
       }
     }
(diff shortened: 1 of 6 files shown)


evwilkin:
@tlabaj Jest tests failing, may need to update snapshots?

packages/react-core/src/components/Slider/examples/Slider.md
```diff
@@ -51,8 +65,16 @@ class DiscreteInput extends React.Component {
     const displayValue = step ? step.label : 0;
```
evwilkin, line 65: Line above need `this.state.value` changed to `this.state.value1`:
`const step = this.steps.find(step => step.value === this.state.value1);`

evwilkin:
Typo in line 164 of Slider.tsx: `const handleToucheStart` => `const handleTouchStart`

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -196,26 +207,34 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
...
+      snapValue = Math.round((newValue - min) / step) * step + min;
+      // snapValue = Math.round(newValue / step) * step;
     }
 
-    // Call value change callback
-    if (onValueChange && !isDiscrete && !steps) {
-      onValueChange(newPercentage);
+    // Call value change callback unless custom steps are discrete or no custom steps and step is !== 1.  The value will snap after the thumb is released.
+    if (onChange) {
+      if ((areCustomStepsContinuous && customSteps) || (!customSteps && step !== 1)) {
```
evwilkin, line 235: The `step !== 1` condition here is preventing the `onChange` from running if `step={1}` or the prop is omitted completely and it defaults to 1, which in the example has the effect of allowing the slider to be moved with the mouse but not updating the value displayed.  Note that using the keyboard still works, I believe because `handleKeyPressOnInput` does not have this same condition before triggering the `onChange`.

mcarrano (changes requested):
For the most part this looks great @tlabaj. I just uncovered some bugs with a couple of the examples. In both the third Value Input example and in the Thumb Value Input example. Entering a value into the input field and pressing Enter does not move the slider.

mcarrano (approved):
Looks to be working now. Thanks @tlabaj !

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -196,26 +207,33 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
...
+      // snap to new value if not custom steps and step is not 1
+      snapValue = Math.round((newValue - min) / step) * step + min;
     }
 
-    // Call value change callback
-    if (onValueChange && !isDiscrete && !steps) {
-      onValueChange(newPercentage);
+    // Call value change callback unless custom steps are discrete or no custom steps and step is 1.  The value will snap after the thumb is released.
+    if (onChange) {
+      if ((areCustomStepsContinuous && customSteps) || (!customSteps && step === 1)) {
```
evwilkin, line 234: Now the `onChange` is only being called if `step={1}` or the prop is omitted and it defaults to 1, so in the examples the values do not update until the mouse is released if the steps are anything above 1 (the opposite issue of what it was previously).  Is there a reason to limit `onChange` from firing based on the step?  It seems like that would prevent the user from seeing their current value without releasing the mouse.

packages/react-core/src/components/Slider/examples/Slider.md
```diff
@@ -69,25 +91,34 @@ class ContinuousInput extends React.Component {
...
   }
 
   render() {
     return (
       <>
         <Text component={TextVariants.h3}>Slider Value is: {this.state.value.toFixed(2)}</Text>
-        <Slider currentValue={this.state.value} onValueChange={this.onValueChange} />
+        <Slider value={this.state.value} onChange={this.onChange} />
         <br />
         <Slider
```
evwilkin, line 117: This is the only example that doesn't display the selected value, not sure if that was intentional?

packages/react-core/src/components/Slider/Slider.tsx
tlabaj: Hmmmm. I think I had issue with that.  I think it was more with the interaction of dragging the thumb not looking smooth  if you update the value and snap before releasing the thumb.  That is just my opinion though.  I can update it and see which interaction is better.  I will keep it the same for custom steps so we can see the difference.

tlabaj: Or would you still not expect the slider thumb to snap until after release?
cc @mcarrano

mcarrano: I'm not sure what the issue here is @tlabaj . It's currently working as I'd expect it to. We're talking about this example, right? https://patternfly-react-pr-5669.surge.sh/components/slider#thumb-value-input

nicolethoen (dismissed):
Works on my phone - API makes sense - looks good to me!

packages/react-core/src/components/Slider/Slider.tsx
tlabaj: @mcarrano @evwilkin, I updated the slider to change value as you drag for non custom steps as seen in the second and third discrete examples.  The first discrete example with custom steps snaps and updates the value once the thumb is released.  Which is the correct and logic interaction?

mcarrano: OK. Now I better understand the question @tlabaj . I think we should go with the implementation in the first example - update only after the thumb is released. If I understand correctly, this is also how the standard HTML range input works? I tested this on my phone also and it seems to work well there.

FWIW, I initially thought that the second example was broken because the steps are far apart and the slider does not move until you get close enough to the next discrete value. I doubt that users would understand that.

tlabaj: So the standard html range actual updates as you drag it. Like in the second example (seen in first clip).   Heres is a recording of HTML example.  This make sense like Evan said, especially if you are displaying the value as you drag the thumb.  

https://user-images.githubusercontent.com/39532947/115035338-7d57f280-9e9a-11eb-8e34-e45b06445462.mov

mcarrano: OK. I misunderstood. Then yes, I think we should follow what the standard HTML slider does. However, I would then either change the second example to have tick marks or remove it entirely. From a design perspective I would not recommend not using tick marks when the slider has discrete steps. As a user I have no way of telling what the increments are without them.

mcarrano (approved):
Looks great. Thanks for making those changes @tlabaj .

nicolethoen (approved):

dlabrecq (commented):
LGTM

PR #5675, opened 2021-04-14
feat(Select): Added invalid state to select

### LINKED ISSUE #5579: Select - add invalid state
https://github.com/patternfly/patternfly/pull/3940/

Adds an invalid state to the select component via
* `.pf-m-invalid` on the parent select element
* adding `span.pf-c-select__toggle-status-icon` with the FA exclamation-circle icon in it after `.pf-c-select__toggle-wrapper` and before `.pf-c-select__toggle-arrow`
* `aria-invalid="true"` on the select toggle

cc @mcarrano

tlabaj:
**What**: Closes #5579 

See the added `Validated` example 

**Additional issues**:


mcarrano (commented):
@tlabaj @mcoker This looks good with one exception. Looks like it's built to only show an error state (red). Per the original request, it should be possible to apply any condition to this like we do for the Text Input that has a `validated` prop to set the color and icon.

Is this something that needs to go back to Core or can we make this work like Text Input by just adding the extra prop to apply the correct styles?

mcoker:
@mcarrano my mistake, from the issue title I assumed we were just adding invalid states for the time being. I've added the warning and success states in this core PR - https://github.com/patternfly/patternfly/pull/4008.

@tlabaj I'm assuming this would be a small update in this PR to add the additional 2 states?

nicolethoen:
this is waiting for a core bump, right? The success/warning colors are not green/yellow

tlabaj:
@nicolethoen yes!

jessiehuff (changes requested):
This looks really similar to our form validations, and we should probably follow a similar approach. Currently the example just changes the icon shown depending on whether it's validated, and this icon is hidden from a screen reader so these users will have no idea whether or not their selection is validated or not. I find it a little odd to communicate the validation state with just an icon instead of an explanation like we do with [our form validation](https://www.patternfly.org/v4/components/form#invalid).

However, if we're going with just an icon, we should at least wrap the icon span in a div with `aria-live="polite"` and then include another span inside that div (sibling to the icon span) with screen reader only text like this: `<span class="pf-u-screen-reader">{validation state}</span>`, similar to the skeleton component approach. However, that's an idea that would need to be tested.

tlabaj:
Thanks for the review @jessiehuff. We would need to make any updates in core as well correct?

cc @mcoker

kmcfaul (approved):
lgtm! Though the example name confused me for a second (`With validation` would make more sense to me). Not a blocker or anything though

tlabaj:
@kmcfaul I used the same name as examples in ur other components (Form, FormSelect, etc)
@mcarrano any thoughts on the name?

kmcfaul:
It's not a blocker. I figured out it showed all three validation states as I flipped through the options. In my mind `validated` = `success` when I glanced at it, but I saw after a moment `validated` = `select that is validated` in this case.

mcarrano:
@tlabaj  I guess that I also thought that "Validated" sounded strange, but it is what we are using elsewhere. I'd be OK to consider this in a future issue, but let's leave alone for now.

kmcfaul (dismissed):

jessiehuff (commented):
I've found two approaches that seem to work when I implement them. The quickest fix is to add an `aria-labelledby` to the Select pointing to an id of the `aria-live` div. (I realized after commenting though that there's already an `aria-labelledby` on the Select so this approach wouldn't work as it'd override the previous `aria-labelledby`.)

However, I also tried implementing the approach that we have with our form validation and that seems to follow the [error handling guidance](https://docs.google.com/document/d/1MY8v7KqSjYOn_SH5C3rnthaAk0BCg4-6_oAHrBFx-eY/edit#heading=h.bxwc6twatzev) Jenn and I discussed previously. This approach also seemed to work when I tested it locally and is probably a better long term solution. We would add an `aria-describedby` prop and an `aria-invalid` prop to Select and then use these two props. The `aria-describedby` would point to the alert's id and the `aria-invalid` would be true only if the validation == 'error'. So the md file would look like this: 
```
       <Select
          variant={SelectVariant.single}
          placeholderText="Select an option"
          aria-label="Select Input with validation"
          onToggle={this.onToggle}
          onSelect={this.onSelect}
          selections={selected}
          isOpen={isOpen}
          aria-labelledby={titleId}
          isDisabled={isDisabled}
          validated={validated}
          aria-describedby="validated-helper"
          aria-invalid={validated == 'error' ? true : false}
        >
          {this.options}
        </Select>
        <div id="validated-helper" aria-live="polite" hidden>{validated}</div>
```

I imagine we have to expose these props if we're going to hook it up later to a validation helper text component.

What do you think, @jgiardino? I've still been wrapping my head around the different use cases and approaches and what works when I test.

jessiehuff (approved):
Looks like you need to update your snapshots, but I just tested in VO and it sounds a lot better! 😄

jessiehuff (approved):
Awesome! 😄 👍 🎉

kmcfaul (approved):
LGTM! :D

nicolethoen (approved):

dlabrecq (approved):

mattnolting (approved):
Looks great to me! We'll need to update Core to reflect the accessibility features.

PR #5678, opened 2021-04-15
AccessConsoles: clear selection when the selected option is a placeholder

KKoukiou:
Without this fix one could actually select the placeholder which creates the following unexpected UI:

Screen Shot 2021-04-16 at 10 44 18

### DIFF
--- a/packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx
+++ b/packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx
@@ -108,10 +108,15 @@ export const AccessConsoles: React.FunctionComponent<AccessConsolesProps> = ({
             toggleId="pf-c-console__type-selector"
             variant={SelectVariant.single}
             onSelect={(_, selection, isPlaceholder) => {
-              setType(Object.keys(items).find(key => items[key] === selection));
-              setIsOpen(isPlaceholder ? false : !isOpen);
+              if (isPlaceholder) {
+                setType(null);
+              } else {
+                setType(Object.keys(items).find(key => items[key] === selection));
+              }
+
+              setIsOpen(false);
             }}
-            selections={items[type]}
+            selections={type ? items[type] : null}
             isOpen={isOpen}
             onToggle={onToggle}
           >


jeff-phillips-18 (approved):
One could argue that the Select component should not allow selection of the placeholder. IMO, selection of the placeholder should simply close the dropdown w/o calling the `onSelect` callback.

KKoukiou:
> One could argue that the Select component should not allow selection of the placeholder. IMO, selection of the placeholder should simply close the dropdown w/o calling the `onSelect` callback.

I actually agree but this would be a breaking change, no? meaning that it would get a year or so to get merged.

jeff-phillips-18:
> > One could argue that the Select component should not allow selection of the placeholder. IMO, selection of the placeholder should simply close the dropdown w/o calling the `onSelect` callback.
> 
> I actually agree but this would be a breaking change, no? meaning that it would get a year or so to get merged.

Maybe, if applications really want to be able to select the placeholder. I don't see what the benefit would be but I guess there could be a case for reset. As a user I'm not sure I'd think you could select the placeholder, it doesn't make much sense.

This PR is fine. I'll log an issue so others can think about it as well.

tlabaj (approved):

martinpitt:
In https://github.com/cockpit-project/cockpit-machines/pull/115 we got an (automated) update to version 4.5.3. It still has the same bug, at least in cockpit-machines the console selector still has "Select console type" as choosable option.

KKoukiou:
> In [cockpit-project/cockpit-machines#115](https://github.com/cockpit-project/cockpit-machines/pull/115) we got an (automated) update to version 4.5.3. It still has the same bug, at least in cockpit-machines the console selector still has "Select console type" as choosable option.

Just to avoid confusion, this is not a bug, see comment here: https://github.com/cockpit-project/cockpit-machines/pull/115#issuecomment-823030118

If we end up deciding this behavior is not expected we can use another component for the console selection.

KKoukiou:
> > In [cockpit-project/cockpit-machines#115](https://github.com/cockpit-project/cockpit-machines/pull/115) we got an (automated) update to version 4.5.3. It still has the same bug, at least in cockpit-machines the console selector still has "Select console type" as choosable option.
> 
> Just to avoid confusion, this is not a bug, see comment here: [cockpit-project/cockpit-machines#115 (comment)](https://github.com/cockpit-project/cockpit-machines/pull/115#issuecomment-823030118)
> 
> If we end up deciding this behavior is not expected we can use another component for the console selection.

After some discussion, I agree with Martin this is an unexpected behavior. I will send a followup PR with the approach seen here https://www.patternfly.org/v4/components/select/#single-with-description

PR #5683, opened 2021-04-15
fix(workspace): remove codecov

redallen:
**What**: Codecov had a recent security vulnerability. Their service is spotty and best and we don't use them for much, so might as well remove them!

**Additional issues**:


redallen:
Merging to test new credentials before EOD.

PR #5688, opened 2021-04-16
fix(Accordion): set button type, update prop extension

### LINKED ISSUE #5661: Button in AccordionToggle submits form

In `AccordionToggle`, the Button isn't explicitly set with type "button", which makes it a "submit" type by default (may depend on browsers). It is of course not expected that expanding or collapsing an accordion item would submit a parent form.

https://codesandbox.io/s/epic-dirac-o87of?file=/index.js

Expanding or collapsing accordion would trigger the "submitted" alert.

It's a bug. There's a workaround, since the `AccordionToggle` props are passed down to button, it's possible to write:

```jsx
            <AccordionToggle
              onClick={() => {
                onToggle("ex-toggle1");
              }}
              isExpanded={this.state.expanded === "ex-toggle1"}
              id="ex-toggle1"
              type="button"
            >
```

In typescript however, the AccordionToggleProps wouldn't allow to pass an extra props, so it has to be cast as `any` first.

Product is the openshift console. Tested with several versions: 4.101.2 (used in openshift console) and 4.106.2 in codesandbox.
Tested on Firefox 87 and Chrome 89.

kmcfaul:
**What**: Closes #5661

@tlabaj I'm not sure, but would this be considered a breaking change? It sets a default, but the default probably shouldn't be `submit` for this component.

### DIFF
--- a/packages/react-core/src/components/Accordion/AccordionToggle.tsx
+++ b/packages/react-core/src/components/Accordion/AccordionToggle.tsx
@@ -4,7 +4,8 @@ import styles from '@patternfly/react-styles/css/components/Accordion/accordion'
 import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
 import { AccordionContext } from './AccordionContext';
 
-export interface AccordionToggleProps extends Omit<React.HTMLProps<HTMLButtonElement>, 'type'> {
+export interface AccordionToggleProps
+  extends React.DetailedHTMLProps<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement> {
   /** Content rendered inside the Accordion toggle  */
   children?: React.ReactNode;
   /** Additional classes added to the Accordion Toggle  */
@@ -33,8 +34,9 @@ export const AccordionToggle: React.FunctionComponent<AccordionToggleProps> = ({
           <button
             id={id}
             className={css(styles.accordionToggle, isExpanded && styles.modifiers.expanded, className)}
-            {...props}
             aria-expanded={isExpanded}
+            type="button"
+            {...props}
           >
             <span className={css(styles.accordionToggleText)}>{children}</span>
             <span className={css(styles.accordionToggleIcon)}>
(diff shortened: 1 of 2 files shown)


kmcfaul:
Added a FileUpload demo to integration @tlabaj 

Upon testing, it looks like the `onClick` passed to the FileUploadField actually does nothing because it gets set on the div (and wasn't triggering). I adjusted the `onClick` to pass to FileUploadField's text area directly as a callback, and added a property to FileUploadField as well. @metalice Is this the behavior of the onClick you were looking for?

tlabaj (approved):
This should be fine.  We have made the same update in other components.

tlabaj:
@mattnolting core may need a similar update

packages/react-core/src/components/Accordion/AccordionToggle.tsx
```diff
@@ -4,7 +4,8 @@ import styles from '@patternfly/react-styles/css/components/Accordion/accordion'
 import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
 import { AccordionContext } from './AccordionContext';
 
-export interface AccordionToggleProps extends Omit<React.HTMLProps<HTMLButtonElement>, 'type'> {
+export interface AccordionToggleProps
+  extends React.DetailedHTMLProps<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement> {
```
nicolethoen, line 8: I have not seen this before but a quick google search says something about DetailedHTMLProps including ref and key as props? Should we be extending this in all our function components? Can you point me to why you made this update?

kmcfaul: I made the update because of the `type` prop. Without it passing 'submit' or 'reset' or 'button' gives an error (it's parsing as a string instead of one of the enum). I've seen it in a couple of components, but I'm unsure which extension we should use over.

nicolethoen: okay. I'll keep it in my back pocket as a possible remedy for other typescript errors/warnings in the future.

nicolethoen (approved):

dlabrecq (approved):

mattnolting (approved):
Perfect! 🎉

PR #5689, opened 2021-04-16
fix(FileUpload): allow user to override onClick

### LINKED ISSUE #5417: File upload: Component override onClick prop
Fileupload component is overring the prop of onClick.

This event is very helpful and cannot be used because of the override.

in file [Fileupload.tsx](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/FileUpload/FileUpload.tsx) around line 131 is the override:
```javascript
<FileUploadField
          {...getRootProps({
            ...props,
            refKey: 'containerRef',
            onClick: event => event.preventDefault() // Prevents clicking TextArea from opening file dialog
          })}
          tabIndex={null}
.....
```

I'm suggesting to change it to something like that:

```javascript
onClick: event => { 
    event.preventDefault();
    props?.onClick(event); 
}
```

kmcfaul:
**What**: Closes #5417

### DIFF
--- a/packages/react-core/src/components/FileUpload/FileUpload.tsx
+++ b/packages/react-core/src/components/FileUpload/FileUpload.tsx
@@ -27,6 +27,8 @@ export interface FileUploadProps
       | React.ChangeEvent<HTMLTextAreaElement> // User typed in the TextArea
       | React.MouseEvent<HTMLButtonElement, MouseEvent> // User clicked Clear button
   ) => void;
+  /** Callback for clicking on the FileUploadField text area. By default, prevents a click in the text area from opening file dialog. */
+  onClick?: (event: React.MouseEvent) => void;
   /** Additional classes added to the FileUpload container element. */
   className?: string;
   /** Flag to show if the field is disabled. */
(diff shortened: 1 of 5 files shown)


tlabaj (commented):
Looks good.  Can you just add this to the demo-app please.

kmcfaul:
Added a FileUpload demo to integration @tlabaj 

Upon testing, it looks like the `onClick` passed to the FileUploadField actually does nothing because it gets set on the div (and wasn't triggering). I adjusted the `onClick` to pass to FileUploadField's text area directly as a callback, and added a property to FileUploadField as well. @metalice Is this the behavior of the onClick you were looking for?

tlabaj:
@kmcfaul I see you have not received a response to your question.  Maybe make a comment on the issue.

nicolethoen (dismissed):

tlabaj:
@kmcfaul, if you rebase this we can merge it .  I think the fix should be ok. if not another issue can be opened.

tlabaj (approved):
LGTM

nicolethoen (approved):

Issue #5682, opened 2021-04-15
Mobile Issues 


christiemolloy:
Some bugs were recently filed in the Managed Kafka repo, particularly with the Control Plane UI. I have tested these bugs myself in BrowserStack and was able to reproduce some of them and also verify that they were happening in PatternFly: 

**Describe the issue. What is the expected and unexpected behavior?**

1. On most new Android Devices: 
- E.g. Samsung: Galaxy S21
PatternFly Toolbar component with Text Input Filter inside of the Toggle Group Expanded has a bug where you go to type into the filter field and the filter input immediately loses focus and the toggle group closes.

2. On Android devices:
PatternFly Toolbar component inside the toggle group expanded, the filter options and inputs positions are moving up and down depending on the selected filter type. The gap between the input and the applied filters is too big 

<img width="568" alt="Screen Shot 2021-04-15 at 11 51 39 AM" src="https://user-images.githubusercontent.com/20118816/114899815-4a9df380-9de1-11eb-9dc3-4c1e109b27a1.png">
<img width="582" alt="Screen Shot 2021-04-15 at 11 51 48 AM" src="https://user-images.githubusercontent.com/20118816/114899829-4d98e400-9de1-11eb-8c72-6e13fa6ee921.png">

3. on iPadOS 14
1536 x 2048 px screen, iPadOS 14 and 13, vertical orientation - table with kafkas in the row legend "name" is broken. Horizontal orientation is ok. Happens in both Safari and Chrome.

<img width="568" alt="Screen Shot 2021-04-15 at 12 04 47 PM" src="https://user-images.githubusercontent.com/20118816/114901496-da906d00-9de2-11eb-9a07-7b7723fea351.png">

4. on iPadOS 14
iPadOS 14, shadows of selected buttons are left after clicking in the place where the button was. This Issue was observed only on our iPad Air 2, not observed on iPad Air 4, iPad Pro and iPhone 12 Pro on Browserstack.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

- You can use this link to see the bugs in Managed Kafka: https://cloud.redhat.com/beta/application-services/streams/kafkas 

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug

**What is your product and what release version are you targeting?**

By Summit, but ideally as soon as possible. A lot of users at Summit will be using their mobile phones.

evwilkin:
[...]

PR #5692, opened 2021-04-19
fix(Toolbar): Collapse expandable content when crossing breakpoint

nicolethoen:
**What**: Towards #5682 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Toolbar/Toolbar.tsx
+++ b/packages/react-core/src/components/Toolbar/Toolbar.tsx
@@ -5,6 +5,18 @@ import { css } from '@patternfly/react-styles';
 import { ToolbarContext } from './ToolbarUtils';
 import { ToolbarChipGroupContent } from './ToolbarChipGroupContent';
 import { formatBreakpointMods } from '../../helpers/util';
+import mdBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_md';
+import lgBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_lg';
+import xlBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_xl';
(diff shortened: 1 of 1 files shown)


nicolethoen:
@evwilkin I tested it on my phone (android) and I think it fixes the closing keyboard issue.

evwilkin:
@christiemolloy this is the proposed fix for the first issue here https://github.com/patternfly/patternfly-react/issues/5682

evwilkin:
@nicolethoen this looks to fix the issue on Android, but it does look to break the responsive behavior that you mentioned is the reason for the resize observer existing.

christiemolloy:
Thanks Nicole and Evan! I tested on BrowserStack Android and the keyboard issue has been solved. What was the responsive behavior that is breaking now @evwilkin ?

evwilkin:
@christiemolloy on the docs site, if you shrink the screen down to mobile & open the filters, then expand back to desktop size the view will revert back to desktop styling:
image

In this example, with the expandable content not closing on resize, the mobile view remains if the browser is sized back up to desktop:
image

nicolethoen:
I updated it with a simpler fix which should get it all working for summit. But i did open an issue to fix some technical debt to make this resize event handler work by taking into account breakpoints - as it really should.

evwilkin (approved):
LGTM

evwilkin:
@christiemolloy mind taking a second look?  Updated to address existing responsive functionality, but the Android fix should still work as expected

tlabaj (commented):
You can add this to tech debt issue; we should test that the tool bar works when `windowWidth` is set by consumer.

christiemolloy (approved):
Confirmed on Android keyboard is working, thank you Nicole and Evan!

tlabaj (approved):

PR #5693, opened 2021-04-20
react-console(AccessConsoles): hide placeholder option after first selection

KKoukiou:


KKoukiou:
Sorry for the back and forth, I was really confused from the different behavior of the examples here https://www.patternfly.org/v4/components/select/#single-with-description

The two first options have namely a different behavior regarding the placeholder. We decided to go with the second.

redallen (approved):

nicolethoen (approved):

Issue #5073, opened 2020-10-28
Composable menu POC


mcarrano:
Create a proof of concept for composable dropdown menus per the design document here: https://docs.google.com/document/d/1GyVgog_SbErqbYyF08_N2rfLDUgwAVbdKP5udBjdxYI/edit?usp=sharing

The PoC must demonstrate building a basic dropdown menu and select menu by combining a toggle button and menu. Should also explore more complex usage to determine the best architectural approach. Ultimate goal is to be able to compose a variety of dropdown and select variants from composable elements to be introduced in parallel with existing Dropdown and Select component or possibly leading to a refactoring or deprecation of those components.

Results of this PoC will drive additional work.

mcarrano:
@tlabaj @mcoker @jschuler @kmcfaul I added the above description based on our discussion. Please feel free to add any additional information that you think is relevant here.

mcarrano:
Keep open

mcarrano:
Opened https://github.com/patternfly/patternfly-react/issues/5532 to add the Menu toggle component as a prerequisite to delivering this POC.

mcarrano:
This is closed per https://github.com/patternfly/patternfly-react/pull/5722

PR #5722, opened 2021-04-27
feat(MenuToggle): add menu toggle / Add composable menu POCs

### LINKED ISSUE #5532: Introduce the Menu toggle component
Introduce the Menu toggle component to PF React as implemented in core here: https://pf4.patternfly.org/components/menu-toggle  This component will be required to create composable dropdowns using the Menu component.

cc @tlabaj

kmcfaul:
**What**: Closes #5532 and #5073

- Adds `MenuToggle` 
- Adds 3 POC examples combining `MenuToggle`, `Menu`, and `Popper` (dropdown, select, drilldown examples)
- POCs live in the `Demos` section for the moment (@mcarrano Let me know if they should live somewhere else)
- POCs have some additional basic key/click handlers to handle closing the menu
- Exports `Popper` helper component to enable POC (@redallen @jschuler Let me know if this is acceptable to expose, I'm unsure if we had a reason to not export Popper)

MenuToggle surge link: https://patternfly-react-pr-5722.surge.sh/components/menu-toggle
Composable menu POC surge link: https://patternfly-react-pr-5722.surge.sh/demos/composable-menu


mcarrano:
@kmcfaul great start. I have a couple of issues/comments:

* Looks like I need to click on the label of the Menu toggle to get the menu to open. If I click on the "v" nothing happens. I should get the action by clicking from either place.
* The way these elements size is a bit unexpected. Should the toggle and menu keep the same width like the standard dropdown and select do now? I'm not completely sure but the way it is now does not seem exactly right. Here are some screen grabs.

Screen Shot 2021-04-27 at 3 26 58 PM

Screen Shot 2021-04-27 at 3 28 07 PM

Thoughts?

I think keeping this in demos makes sense. But is there a way to link to this demo from either the Menu toggle or Menu component. Otherwise I'm concerned that no one will find it.

kmcfaul:
@mcarrano I'll look into the dead zone on the click. Not quite sure why that's happening since the whole thing should be a button (the icon is just in a span).

For the menu/toggle width matching, that is actually an option of Popper (`popperMatchesTriggerWidth` prop). I had turned it off because the menus were getting truncated due to the thin toggle width, but I will update them to have a wider width and match.

kmcfaul:
It looks like when an svg is inside of a button it doesn't bubble pointer events? Adding `pointer-events:none` to the svg CaretDownIcon inside the controls makes the click work again, but I'm not sure if adding that would have other side effects. @mattnolting @mcarrano @mcoker Thoughts?

jessiehuff (commented):
This looks pretty good so far!! 🙂  It also sounded good in VO! The only concern I had when testing it was that the keyboard interaction seemed a bit off. For example it took pressing tab to get into the basic example and then could switch to arrow keys. In other menu-like components Tab would close the menu and move focus to the next item. The left/right arrow keys also didn't function the way we had it in the menu examples. It would be awesome if these matched those, but if we want to get this PR in quicker, I'm ok with us handling this in another issue.

kmcfaul:
@jessiehuff I'll take another look. I think I can overwrite tab behavior from the POC like I do for Escape. I can also add an example that uses the Left/Right arrow keys, none of the current POC examples used menus that had that behavior.

kmcfaul:
@jessiehuff Updated the keyboard behavior a bit. `Tab` while in an opened menu now closes the menu, and `ArrowDown` while on the toggle with an opened menu will enter into the menu. I also added another example with menu actions to demo the `ArrowRight` and `ArrowLeft` part of the menu naivgation.

jessiehuff (approved):
This is definitely looking better! I'd still love to see the arrow keys added at some point for the drilldown menu example only because I know that consumers will likely follow our examples so they might implement something without the keyboard interaction set up the way a user would likely expect. But I'm ok with approving for now and updating later if we need to. 🙂

nicolethoen:
I feel like the clickable area for the drill down menus is so small in the demo - like i kept selecting the item rather than drilling down.

Also, is it truly the case that any ReactNodes are valid content for the MenuToggle? I anticipate that the first thing people are going to ask about is putting an input and controls to build something like a typeahead select. Would that be possible the way it's built?

kmcfaul:
@nicolethoen Yes you can pass anything to MenuToggle. I just tested locally passing a simple `input` element with an `onChange` handler. To wire up a typeahead with selection the user would just have to take the `input` value and then filter and re-pass the menu options to `Menu`.

packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -0,0 +1,665 @@
...
+      }
+
+      if (event.target === this.toggleRef.current && this.state.isExpanded) {
+        if (event.key === 'ArrowDown') {
+          this.menuRef.current.querySelector('button, a').focus();
+        }
+      }
+    };
+    this.handleMenuClick = event => {
+      if (![...event.target.classList].some(c => /pf-c-menu.*/.test(c)) && this.state.isExpanded) {
```
mcoker, line 478: @kmcfaul re: clicking on the SVG not toggling the menu, would this regex match with the class of the target be the culprit? The SVG doesn't have a `pf-c-menu.*` class, but if you add `pointer-events: none;` to the SVG, then your click target will be the element behind it, which is probably `.pf-c-menu-toggle__toggle-icon`

kmcfaul: Oh that would probably be the culprit. I'll look at adjusting the match to account for an svg target.

kmcfaul: Looking at it again, that function is handling clicking off the menu. The actual onClick callback for `MenuToggle` isn't working with the svg, even though the svg is contained inside the toggle `button`.

I could make the outside function handle clicking the svg as well, but I'm curious how we handled this case for Dropdown since it does work with the basic onClick. @mcoker

packages/react-core/src/components/MenuToggle/MenuToggle.tsx
```diff
@@ -0,0 +1,92 @@
...
+  /** Content rendered inside the toggle */
+  children?: React.ReactNode;
+  /** Additional classes added to the toggle */
+  className?: string;
+  /** Flag indicating the toggle has expanded styling */
+  isExpanded?: boolean;
+  /** Flag indicating the toggle is disabled */
+  isDisabled?: boolean;
+  /** Flag indicating the toggle should have primary styling */
+  isPrimary?: boolean;
```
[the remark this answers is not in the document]
kmcfaul: I think whatever is applied last would win out on the css if both happened to be applied, but I don't think they're intended to be used together. Our Dropdown component has these as separate props too, but I'm not sure if that was just one being added later / to avoid a breaking change.

I can update these props into `variant` instead, though I'm curious if `variant` should itself be an optional prop or if it should it have a `default` value as well for basic use cases that don't need either plain/primary styling? @tlabaj

packages/react-core/src/demos/ComposableMenu.md
kmcfaul: Playing around with it further I believe I fixed the svg click issue. It had something to do with the event propagating. @mcoker

packages/react-core/src/components/MenuToggle/MenuToggle.tsx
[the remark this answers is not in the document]
tlabaj: I think the variant is the right way.  And yes, we can add the default state to the variant.

nicolethoen (approved):

redallen (approved):

mcoker (commented):
@kmcfaul this looks great! re: the composable examples, the last one is going out of the browser window, do you think we should orient the menu to come from the top of the toggle?  

Screen Shot 2021-05-11 at 12 49 39 PM

nicolethoen (approved):

redallen (approved):

PR #5723, opened 2021-04-28
feat(Card): add isToggleRightAligned prop

### LINKED ISSUE #5583: Expandable card - support expandable toggle on the right
On an expandable card, adding `.pf-m-toggle-right` to `.pf-c-card__header` will align the toggle to the right side of the header.

This was added as part of new card demos https://github.com/patternfly/patternfly/pull/3848/files

cc @mcarrano

evwilkin:
**What**: Closes #5583 

This PR adds the `isToggleRightAligned` prop to the expandable Card to move the carrot icon to the right of the `CardHeader` component. Note that spacing is to be fixed [per this comment](https://github.com/patternfly/patternfly/pull/3848/files#r591988545).

Updated the `Expandable` example allowing to toggle this prop on & off.

### DIFF
--- a/packages/react-core/src/components/Card/CardHeader.tsx
+++ b/packages/react-core/src/components/Card/CardHeader.tsx
@@ -16,6 +16,8 @@ export interface CardHeaderProps extends React.HTMLProps<HTMLDivElement> {
   onExpand?: (event: React.MouseEvent, id: string) => void;
   /** Additional props for expandable toggle button */
   toggleButtonProps?: any;
+  /** Whether to right-align expandable toggle button */
+  isToggleRightAligned?: boolean;
 }
 
 export const CardHeader: React.FunctionComponent<CardHeaderProps> = ({
@@ -24,11 +26,16 @@ export const CardHeader: React.FunctionComponent<CardHeaderProps> = ({
   id,
   onExpand,
   toggleButtonProps,
+  isToggleRightAligned,
   ...props
 }: CardHeaderProps) => (
   <CardContext.Consumer>
     {({ cardId }) => (
-      <div className={css(styles.cardHeader, className)} id={id} {...props}>
+      <div
+        className={css(styles.cardHeader, isToggleRightAligned && styles.modifiers.toggleRight, className)}
+        id={id}
+        {...props}
+      >
         {onExpand && (
           <div className={css(styles.cardHeaderToggle)}>
             <Button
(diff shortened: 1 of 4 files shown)


tlabaj (dismissed):

jessiehuff (commented):
So this one's a bit odd with keyboard. The tab order (and DOM structure) don't really match the visual presentation which makes for a kind of jumpy experience. Is there a way that we can put the toggle in the DOM after the actions? Then the DOM order, tab order, and visual order would all match. I'm not sure if this is something we'd have do do in core though?

mcarrano (commented):
I guess this is OK, other than @jessiehuff 's issue. However I'm still not clear what the use case is for this. I would probably never recommend placement on the right side of the card header like the example. I believe this enhancement originated with @mattnolting . Matt can you explain when/why we need to move the toggle to the right and if this is the best way to illustrate?

mattnolting:
> I guess this is OK, other than @jessiehuff 's issue. However I'm still not clear what the use case is for this. I would probably never recommend placement on the right side of the card header like the example. I believe this enhancement originated with @mattnolting . Matt can you explain when/why we need to move the toggle to the right and if this is the best way to illustrate?

Primarily for stacked card title alignment. Stacked cards look misaligned w/the title to the right of the toggle. @kybaker

mattnolting:
Hey @evwilkin, I opened https://github.com/patternfly/patternfly/pull/4045 to support this. I think we can move the toggle to the right in the DOM per @jessiehuff's response.

mcoker:
> I'm still not clear what the use case is for this. I would probably never recommend placement on the right side of the card header like the example. I believe this enhancement originated with @mattnolting . Matt can you explain when/why we need to move the toggle to the right and if this is the best way to illustrate?

@mcarrano sorry if this has already been covered, but this was part of the [dashboard card demos](https://github.com/patternfly/patternfly/issues/3278). We discussed whether we should include this enhancement, and @mceledonia confirmed it's a valid use.

tlabaj (approved):
LGTM

kmcfaul (approved):
lgtm!

jessiehuff (approved):

mcoker:
> Note that spacing is to be fixed per this comment. https://github.com/patternfly/patternfly/pull/3848/files#r591988545

@evwilkin I think https://github.com/patternfly/patternfly/pull/3848/files#r591988545 will be a breaking change, so I'd like to hold off on that issue for now. I added https://github.com/patternfly/patternfly/pull/4050 as an interim fix to handle the actions/toggle spacing in the header when the toggle is right aligned.

mcoker (approved):
LGTM, https://github.com/patternfly/patternfly/pull/4050 will fix the toggle alignment.

mcarrano (approved):
Looks good to me

PR #5730, opened 2021-04-30
Bug fix: change key generation field for tree view

### LINKED ISSUE #5558: Allow an id/key to be passed down to the TreeViewList to avoid react warning

Inside the TreeViewList component, I need to be able to pass in an id/key so that react doesn't warn about a missing key.

image

ibolton336:
Since name is of type Reac.ReactNode, I propose we use the item.id to generate the key. 

**What**: Closes #5558

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeView.tsx
+++ b/packages/react-core/src/components/TreeView/TreeView.tsx
@@ -87,7 +87,7 @@ export const TreeView: React.FunctionComponent<TreeViewProps> = ({
     <TreeViewList isNested={isNested} onSearch={onSearch} searchProps={searchProps}>
       {data.map(item => (
         <TreeViewListItem
-          key={item.name.toString()}
+          key={item.id.toString()}
           name={item.name}
           id={item.id}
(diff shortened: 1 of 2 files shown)


redallen (approved):

redallen (dismissed):

packages/react-core/src/components/TreeView/TreeView.tsx
```diff
@@ -87,7 +87,7 @@ export const TreeView: React.FunctionComponent<TreeViewProps> = ({
     <TreeViewList isNested={isNested} onSearch={onSearch} searchProps={searchProps}>
       {data.map(item => (
         <TreeViewListItem
-          key={item.name.toString()}
+          key={item.id.toString()}
```
nicolethoen, line 90: The only issue I see here, is that the id is not a required field, but the name is a required field on the TreeViewDataItem. So maybe it can use the id unless it's undefined? And if so, then use the name?

redallen (approved):

PR #5733, opened 2021-05-03
fix(packages): check canUseDOM before using DOM

### LINKED ISSUE #5732: [BUG] Toolbar window reference prevents the component to be used with SSR
Browser window reference in constructor prevents the component to be used with SSR

Use the component in SSR app

It is a bug and can be fixed by setting the value in `useEffect`/`componentDidMount` lifecycle.
Here is the line: https://github.com/patternfly/patternfly-react/commit/73a0f60ebb64235a1d63b7d01d15da324b790f14#diff-36c890e98584934106f358ac37c47371f6ea68605720b1a5f156446ac53a4d13R60

CRC UI docs, anyone using server-side rendering.

redallen:
**What**: Closes #5732

**Additional issues**:

### DIFF
--- a/packages/react-console/src/components/SerialConsole/XTerm.tsx
+++ b/packages/react-console/src/components/SerialConsole/XTerm.tsx
@@ -3,7 +3,7 @@ import React, { useImperativeHandle } from 'react';
 import { Terminal } from 'xterm';
 import { FitAddon } from 'xterm-addon-fit';
 
-import { debounce } from '@patternfly/react-core';
+import { debounce, canUseDOM } from '@patternfly/react-core';
 
 export interface XTermProps {
   /** The number of columns to resize to */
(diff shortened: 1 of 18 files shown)


kmcfaul:
If canUseDOM returns false, wouldn't that make the page broken (no event listeners) without any warnings? Or is canUseDOM expected to be true most of the time?

redallen:
@kmcfaul If `canUseDOM` is false the page is non-interactive. `canUseDOM` just checks that `window` exists and that React can work via `createElement`.

If `canUseDOM` is false you're probably using SSR to render components in a fake DOM using a node context. In that case you don't care about event listeners and just want an initial rendering to later [hydrate.](https://reactjs.org/docs/react-dom.html#hydrate)

kmcfaul:
Ah I see, thanks for clarifying!

kmcfaul (dismissed):

evwilkin (approved):
LGTM

nicolethoen (approved):

PR #5734, opened 2021-05-03
feat(Dropdown): add alignment breakpoints, fix ref console warning

### LINKED ISSUE #5347: Dropdown support responsive left/right alignment
https://github.com/patternfly/patternfly/pull/3746

Now you can apply:

`.pf-m-align-left{-on-[breakpoint]}`
`.pf-m-align-right{-on-[breakpoint]}`

cc @mcarrano @DaoDaoNoCode

kmcfaul:
**What**: Closes #5347

- Adds `alignments` property that dictates left/right alignment on specific breakpoints
- Fixed console warning where `ref` was getting improperly assigned to DropdownSeparator function component (ref isn't used for these).

### DIFF
--- a/packages/react-core/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react-core/src/components/Dropdown/Dropdown.tsx
@@ -18,6 +18,14 @@ export interface DropdownProps extends ToggleMenuBaseProps, React.HTMLProps<HTML
   isPlain?: boolean;
   /** Indicates where menu will be aligned horizontally */
   position?: DropdownPosition | 'right' | 'left';
+  /** Indicates how the menu will align at screen size breakpoints */
+  alignments?: {
+    sm?: 'right' | 'left';
+    md?: 'right' | 'left';
+    lg?: 'right' | 'left';
+    xl?: 'right' | 'left';
+    '2xl'?: 'right' | 'left';
+  };
   /** Display menu above or below dropdown toggle */
   direction?: DropdownDirection | 'up' | 'down';
   /** Flag to indicate if dropdown has groups */
@@ -38,6 +46,7 @@ export const Dropdown: React.FunctionComponent<DropdownProps> = ({
   ref, // Types of Ref are different for React.FC vs React.Component
   ouiaId,
   ouiaSafe,
+  alignments,
   ...props
 }: DropdownProps) => (
   <DropdownContext.Provider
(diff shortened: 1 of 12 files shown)


nicolethoen (commented):
This looks good. 
I was trying to figure out what the default alignment is - should that be in the docs?

kmcfaul:
Oh you know, default probably isn't relevant here. In other breakpoints, default is used to see a universal modifier (so it would be `pf-m-align-right` in this case) but I'm not sure we need it because we already have the `position` property. I'll remove it.

packages/react-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -18,6 +18,14 @@ export interface DropdownProps extends ToggleMenuBaseProps, React.HTMLProps<HTML
   isPlain?: boolean;
   /** Indicates where menu will be aligned horizontally */
   position?: DropdownPosition | 'right' | 'left';
+  /** Indicates how the menu will align at screen size breakpoints */
+  alignments?: {
```
mattnolting, line 22: Is this missing `default`?

mattnolting (commented):
Looks great, just the one question

mattnolting:
@mcoker Should we add `top | bottom`?

mcoker:
> Should we add top | bottom?

@mattnolting fine with me, though I would probably be in favor of adding it if we got a request for it. Left/right alignment work a lot better with viewport width mq's than top/bottom would.

nicolethoen:
I think that since there is a default alignment, the props or docs should at least say that for clarity

packages/react-core/src/components/Dropdown/Dropdown.tsx
kmcfaul: `default` is effectively the `position` attribute. However `alignments` is set on a different element than `position`, so I left `default` out since I'm not sure it would work the same.

kmcfaul:
Added to the property description to indicate default alignment should use the `position` property.

packages/react-core/src/components/Dropdown/Dropdown.tsx
tlabaj: I think dow the road if we have a breaking change we will want to combine the two props.  To me it seems a little confusing having the two.

tlabaj (approved):

nicolethoen (approved):

packages/react-core/src/components/Dropdown/Dropdown.tsx
mcoker: > However alignments is set on a different element than position

@kmcfaul do you mean the classes `alignments` add are on a different element than the classes added by `position`? If so, they should be the same. Looks like `position` adds classes to the `.pf-c-dropdown` element, but doesn't need to.

mcoker (approved):
LGTM! The only confusing thing is that `position` sets the default alignment, and `alignments` is for responsive alignment, but if that's how it needs to be until we can combine them in a breaking change, that works for me!

dlabrecq (approved):

PR #5735, opened 2021-05-03
feat(Gallery): add min/max widths at breakpoints

### LINKED ISSUE #5055: Gallery layout supports customizable min/max-width
PRs:
* https://github.com/patternfly/patternfly/pull/3549/
* https://github.com/patternfly/patternfly/pull/3703/

You can override the column width by setting `--pf-l-gallery--GridTemplateColumns--min{-on-[breakpoint]}` and `--pf-l-gallery--GridTemplateColumns--max{-on-[breakpoint]}`

Examples are on https://pf4.patternfly.org/layouts/gallery

kmcfaul:
**What**: Closes #5055

- Adds `minWidths` and `maxWidths` properties that specify their respective width modifier at different breakpoints. The `default` will set a min or max width in general (not at any breakpoint).

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/GalleryDemo/GalleryDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/GalleryDemo/GalleryDemo.tsx
@@ -9,12 +9,19 @@ export class GalleryDemo extends React.Component {
         hasGutter
         minWidths={{
           default: '100%',
+          sm: '80px',
           md: '100px',
-          xl: '300px'
+          lg: '150px',
+          xl: '300px',
+          '2xl': '500px'
         }}
         maxWidths={{
-          md: '200px',
-          xl: '1fr'
+          default: '100%',
+          sm: '80px',
+          md: '120px',
+          lg: '150px',
+          xl: '1fr',
+          '2xl': '500px'
         }}
       >
         <GalleryItem>Gallery Item</GalleryItem>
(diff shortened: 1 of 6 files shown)


nicolethoen (dismissed):

mattnolting (dismissed):
LPTM, great work @kmcfaul!

packages/react-integration/demo-app-ts/src/components/demos/GalleryDemo/GalleryDemo.tsx
```diff
@@ -5,7 +5,18 @@ export class GalleryDemo extends React.Component {
   static displayName = 'GalleryDemo';
   render() {
     return (
-      <Gallery hasGutter>
+      <Gallery
```
tlabaj, line 8: I would add demo for each breakpoint.  We have instances in the past where certain breakpoints styles were not being applied as expected.

tlabaj (changes requested):

packages/react-integration/demo-app-ts/src/components/demos/GalleryDemo/GalleryDemo.tsx
kmcfaul: I'm confused, by each breakpoint do you mean one demo for `minWidths` and one demo for `maxWidths` or do you mean one demo per size (sm/md/lg/etc) per property? The current test is testing the whole style object to ensure the classes are applied properly. Is there a different method of testing I should use in the cypress tests instead?

tlabaj: @kmcfaul I would just add the missing breakpoints to what you have (xl and 2xl etc) for both the max and min widths to make sure they are applied correctly.

kmcfaul: Oh I understand, gotcha!

kmcfaul:
Updated to check all breakpoints in integration

tlabaj (approved):
Looks great!

nicolethoen (approved):

Issue #5647, opened 2021-04-09
Add View more and footer to Select and Menu components


mcarrano:
Add two new features to Select and Menu.

* A **View more** action can be used to load more items when the options list is only partially loaded. A spinner appears while the menu is loading.

* An actions footer allows for the placement of one or more actions in a sticky footer.

Core PR: https://github.com/patternfly/patternfly/pull/3968
Designs: https://marvelapp.com/prototype/c49ebi3/screen/77975699

mcoker:
## Menu
* Adds `.pf-m-load` to `.pf-c-menu__list-item` with a standard menu item in it for the "load more"
* Adds `.pf-m-loading` to `.pf-c-menu__list-item` with a large spinner for "loading"
* Adds `div.pf-c-menu__footer` for the footer that comes after `.pf-c-menu__content` that can take any kind of action, but in our example we should include an inline link button.

## Select
* Adds `.pf-m-load` to `.pf-c-select__menu-item` for the "load more" item
* For the loading element, we add a new class to the `<li>` - `.pf-c-select__list-item`. Then we add `.pf-m-loading` to that element, and place a large spinner in it.
* For the footer, we change `.pf-c-select__menu` to `<div>` and remove the aria and role from it, then create a new `<ul>` that we move the aria/role from `.pf-c-select__menu` to, and give that `<ul>` the class `.pf-c-select__menu-list`. Then after that, we add `div.pf-c-select__footer` with the same notes as the menu component footer above.

nicolethoen:
Priority wise, adding this feature to the select component is higher priority than adding it to the menu component.

jessiehuff:
So this is a tough one on the a11y side as I don't think there are a lot of others doing this type of interaction in a very a11y friendly way. However, there are some things we could probably do based on research to improve the a11y experience. The difficulty is that when a user presses the load more button, keyboard users would likely lose their spot and have to shift focus through every option again (probably going backwards if focus remains on the load more button), and screen reader users would likely have no idea that the page has updated and would have a similar experience of needing to navigate through all the options again (if they even get rendered correctly in the screen reader as I've seen issues with this before so I wouldn't be surprised if we ended up with some kind of difficulty in that realm).

[...]

PR #5736, opened 2021-05-03
feat(Select): Added footer and view more options

### LINKED ISSUE #5698: Add view more and footer to Select

tlabaj:
**What**: Closes #5698

Added:
- select footer
- support for loading more items (marked prop to support this as beta)
- hid internal props

TODO: keyboard interactions for the checkbox view more button

@mcarrano @jessiehuff the keyboard interactions probably still need refinement.


mcarrano (commented):
@tlabaj This is looking good, but just have a couple of questions. The issue called for two things, a fixed footer as shown here, and the ability to add a "View more" option at the end of the menu to load more items as shown in the core example here: https://www.patternfly.org/v4/components/select/html#view-more. Is that here somewhere?

Also, for the footer example, is it possible to set a fixed menu height in the example so someone can see that the footer remains fixed when you scroll the menu?

nicolethoen (dismissed):
API and examples make sense to me!

packages/react-core/src/components/Select/SelectOption.tsx
```diff
@@ -209,22 +217,26 @@ export class SelectOption extends React.Component<SelectOptionProps> {
                 className={css(
                   styles.selectMenuWrapper,
```
mcoker, line 218: Hmm! The view more item should not change its background on hover, but this `.pf-c-select__menu-wrapper` class on the parent `<li>` is causing that background color to appear.

This class should only apply to an `<li>` if it is wrapping a favorites item - in that case, its children will be `.pf-c-select__menu-item.pf-m-link` (the favorite item text) and `.pf-c-select__menu-item.pf-m-action` (the favorite/star button). It looks like this line was added in https://github.com/patternfly/patternfly-react/pull/4642/files#diff-93c821c98c60ca88c0e8281f4ab9674ee65af75b0a4785592ab6f19dca05d6edR191.

We should remove it if possible. My only hesitation with removing it is a user may be targeting it with custom CSS since it's there. Maybe it would be OK to remove it and mention in the release notes that it was removed to fix a bug? WDYT?

tlabaj: I updated it to not apply `.pf-c-select__menu-wrapper` when it is the view more button or the spinner.  The spinner however to still has has the background change on hover.

tlabaj:
@mcoker @jessie I added support for adding "view more". to checkbox selects.  Given the structure the `View more` button is outside of the list.  What is the expected keyboard interaction here.  I would think it would be similar to the footer were you need to tab.  On the other hand that would be inconsistent with the other select variants.

packages/react-core/src/components/Select/SelectOption.tsx
mcoker: @tlabaj looks like the spinner in https://patternfly-react-pr-5736.surge.sh/components/select/react/view-more/ and https://patternfly-react-pr-5736.surge.sh/components/select/react/view-more-with-checkboxes/ are loading differently. The item with the spinner should be appropriate for whatever kind of parent it's in (`<li>` if it's in a `<ul>`, `<div>` if it isn't) with the class `.pf-c-select__menu-item.pf-m-loading`. That will also center it, which you can see in the core docs - https://pf4.patternfly.org/components/select/#loading

tlabaj: @mcoker So I updated the non checkbox select spinner.  The checkbox spinner parent element is a div, so I think that is right.  
I tired both `.pf-c-select__menu-item.pf-m-loading` and `.pf-c-select__list-item.pf-m-loading` for the checkbox  but neither style centered the spinner.

jessiehuff (commented):
I would say on the footer keyboard interaction, the only thing I might change is instead of shifting focus to the footer container, I would think it'd go to the first interactive item inside the container (in this situation the button).

The view more example is ok, except I think it should follow [this comment](https://github.com/patternfly/patternfly-react/issues/5647#issuecomment-827050634) and when pressed should shift focus to the next item in the list (for example, if the last item is Miss, when you press the view more button, focus would go to Mrs).

In terms of the view more with checkboxes, that's a good question. With it looking like it's a part of the list visually, part of me understands and would think to use arrow keys, but I agree that they should be consistent. Perhaps we could use the Tab for both of them for now and decide if there's a better approach later. I'd rather they both be accessible, and currently I can't get to the view more with checkboxes by keyboard at all so at least Tab would get them there. Perhaps we should have the divider on both if we're going to use Tab though?

mcarrano:
@tlabaj @jessiehuff Just to throw in my two cents on the keyboard interaction, I have to admit to having a hard time figuring out how to make this work. But I'm not that familiar with what the interaction pattern should be. The last example with the checkbox select is an odd one IMO. I'm glad if we can support this, but I also think it is an unlikely use case. Most of the time, the View More feature will be used with really long lists of items, like a list of object names for example, where there might be a performance hit to load the entire list. The checkbox select is more useful in filtering use cases where you want to filter items across a set of facets that are fixed and could probably be loaded at once. So I guess we should do our best to make this work well, but I don't consider it critical.

kmcfaul (commented):
LGTM aside from having trouble getting to the View More button via the keyboard for the checkbox variant.

mcoker (changes requested):
* The `.pf-c-select__menu-footer` element is in the `.pf-c-select__menu` parent, but the `__menu` parent is a `<ul>` and the `__menu-footer` element is a `<div>`. If we keep the footer in the menu like that, it will need to be an `<li>` to pass accessibility. That said, I don't think it belongs in the list, and having it in the list will cause issues with being able to scroll the list independently of the footer (allowing the footer to remain fixed when the list scrolls). I would expect it to be structured similar to the context selector if we can - https://www.patternfly.org/v4/components/context-selector#with-footer
* It looks like the "view more with checkboxes" example is causing the loading spinner to have a background color on hover, but I'm having a hard time determining why because I can't inspect the element in time before the spinner disappears and the additional items are loaded. When I click on "view more" in either of the view more examples, the menu closes - is that on purpose? That's part of why I'm unable to inspect that element in time. If the menu can stay open after clicking "view more" I should be able to inspect and see where the hover background is coming from.

packages/react-core/src/components/Select/SelectOption.tsx
mcoker: @tlabaj hmm... it's not working now because it is `.pf-c-select__menu-item.pf-m-loading`, but changing it to `.pf-c-select__list-item.pf-m-loading` seems to work for me in the browser. 

`.pf-c-select__menu-item.pf-m-loading`
Screen Shot 2021-05-12 at 5 26 28 PM


`.pf-c-select__list-item.pf-m-loading`
Screen Shot 2021-05-12 at 5 24 40 PM

tlabaj:
@jessiehuff I have opened issue #5772 to address keyboard interactions fro the footer and view more.

mcoker (approved):
LGTM! 🥳

nicolethoen (approved):

dlabrecq (approved):

kmcfaul (approved):
LGTM!

mcarrano (approved):
Yep, except for the keyboard interactions, this looks great. Can handle that as a follow-on issue.

PR #5740, opened 2021-05-04
core(Slider): showBoundaries should be independent from showTicks

### LINKED ISSUE #5739: Slider: showBoundaries works only when showTicks is set

showBoundaries is a property to show the slider boundaries. It should be not dependent on the showTicks property.

/bug

Cockpit

KKoukiou:
Add examples for the valid combinations.

Closes #5739

### DIFF
--- a/packages/react-core/src/components/Slider/Slider.tsx
+++ b/packages/react-core/src/components/Slider/Slider.tsx
@@ -315,6 +315,7 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
           key={i}
           value={stepValue}
           label={i.toString()}
+          isTickHidden={!showTicks}
           isLabelHidden={(i === min || i === max) && showBoundaries ? false : true}
           isActive={i <= localValue}
         />
@@ -343,7 +344,7 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
             ))}
           </div>
         )}
-        {!customSteps && showTicks && (
+        {!customSteps && (showTicks || showBoundaries) && (
           <div className={css(styles.sliderSteps)} aria-hidden="true">
             {buildSteps()}
           </div>
--- a/packages/react-core/src/components/Slider/SliderStep.tsx
+++ b/packages/react-core/src/components/Slider/SliderStep.tsx
@@ -9,6 +9,8 @@ export interface SliderStepProps extends Omit<React.HTMLProps<HTMLDivElement>, '
   value?: number;
   /** Step label **/
   label?: string;
+  /** Flag indicating that the tick should be hidden */
+  isTickHidden?: boolean;
   /** Flag indicating that the label should be hidden */
   isLabelHidden?: boolean;
   /** Flag indicating the step is active */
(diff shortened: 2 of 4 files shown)


tlabaj:
@mcarrano do we want to show ticks for the min max labels?

KKoukiou:
@mcarrano thanks for the quick answer. Here is a current dialog that uses sliders in cockpit, not PF4 yet.
Screen Shot 2021-05-06 at 08 44 05

There are around 10000 points there, showing ticks does not make sense. However I still want to show the boundaries.

KKoukiou:
Here is another dialog with wide range of input values, where ticks don't belong, however boundaries would make sense. 
Screen Shot 2021-05-06 at 08 48 08

mcarrano:
Are these examples discrete or continuous sliders @KKoukiou . If they are continuous, then I agree that tick marks at either end are not needed. If they are discrete sliders, shouldn't we always have the tick marks to indicate slider steps?

I think this example makes sense:

Screen Shot 2021-05-05 at 12 09 18 PM

But I question whether this one is needed:

Screen Shot 2021-05-05 at 12 09 30 PM

KKoukiou:
@mcarrano the step in our examples is always `1`. So I guess it's continuous, did not realize the difference till now to be honest.
Why do we even need the continuous variant in the end? 
To me it's the same like:
```
        <Slider value={this.state.value3} onChange={this.onChange3} min={-25} max={75} step={1}  />
```

KKoukiou:
@mcarrano so actualy the first usage scenario I sent (VM memory) has min value != 0, so it's not complying with the continuous variant as far as I understand.

mcarrano:
I didn't think that the continuous example required the min value to be 0, but perhaps I'm mistaken. I guess the difference between a slider with a step value = 1 and a continuous slider is that the later can return a fractional value, but maybe we don't need both. What are your thoughts @tlabaj ? If someone wants a slider that can only return whole number values, what you you expect them to use?

In any case, I am fine with the essence of this PR. I just want to make sure that the examples are reflecting the way we want people to use this.

mcarrano:
Can't the min and max values for the continuous slider be set to anything @tlabaj ?

tlabaj:
@mcarrano @KKoukiou in case where you have custom steps, the `areCustomStepsContinuous` prop will calculate the value based on the position on the slider given the first and last steps.  if you are using the min/max/step api, the default `step` is 1.  You can change that so it is less than 1 if you wish.   I think @KKoukiou want to show the min max without having to use the `customSteps` interface which I think is a a valid use case. 
This example in core show ticks with a min and max value.  In that case you do not want to snap to the value of the step.  That is why the `areCustomStepsContinuous` would be used in that use case.  That is also why I asked if we want to show ticks for the min max values.
https://pf4.patternfly.org/components/slider#continuous

mcarrano:
I'm fine with whatever seems to work best here. Regarding your original question, no I don't think it's necessary to show tick marks for the min and max values @tlabaj .

tlabaj (approved):

mcarrano (approved):
Looks good to me.

nicolethoen (commented):
I am seeing that the starting values are NaN in the new examples:

Screen Shot 2021-05-13 at 2 57 40 PM

nicolethoen:
@KKoukiou if you address this latest comment, I can merge this :)

tlabaj (approved):

nicolethoen (approved):

PR #5744, opened 2021-05-04
feat(Table): allow for more customizable Table actions

### LINKED ISSUE #4612: Table: provide a way to have more complicated actions as Table's  last column cells
In [cockpit](cockpit-project.org/) we have a Table whose last column incorporates some actions.
See the screenshot which is self explanatory.
Screen Shot 2020-07-24 at 10 11 20

I was trying to implement that we the `actions` or `actionResolver` properties of the Table component without luck since that's limiting us to use kebab only. So please either allow actions to be more generic or provide a way to add custom classes to rows cells, so that we can add the `pf-c-table__action` class to that last column cells.

nicolethoen:
**What**: Closes #4612

https://patternfly-react-pr-5744.surge.sh/components/table#actions-and-first-cell-in-body-rows-as-th

I added two checkboxes to demonstrate the feature:
- Use custom actions toggle
- Add extra actions

They demonstrate how to change the dropdown toggle to make it something other than a kebab toggle, and how to move some actions out of the dropdown.


kmcfaul (approved):
lgtm!

tlabaj (approved):

dlabrecq (commented):
LGTM

PR #5746, opened 2021-05-05
fix(tooltip): change default entry and exit delay

### LINKED ISSUE #5745: Add built in delay in tooltips
Currently, tooltips do not have a delay on hover, by default. If an element has a tooltip associated to it, the tooltip shows up immediately, with no delay which in many use cases is not a desired interaction. For example. if multiple rows in a table, or multiple cards in a card view have tooltips on hover, the user would run into a situation where they would be overwhelmed with tooltips as they scroll through the page.

I brought this issue up at PatternFly design share and everyone agreed that we should probably **add a second of delay to tooltips by default.** The only exception would be if the tooltip is associated to an icon.

@mcarrano @redallen

redallen:
**What**: Closes #5745

**Additional issues**:


KKoukiou:
@redallen I tried it interactively now, I am not really happy with this change. I feel that if I was a user, hovering over a button, I would not leave my mouse there for more than a second to see that.

In addition, In cockpit I can't think of any scenario where this change makes sense, so we would need to change the properties to the previous state. 

So, since the properties are exposed, the reporter of the issue can adjust them for their scenario. And I honestly don't see the severeness of the reported issue, if the user scrolls with the mouse, and tooltips get activated, they can move the mouse slightly to the side.

redallen:
The way tooltips work in most design systems is that there is some delay. This solves bugs when scrolling through tooltips like this:
image

I'm thinking about adding a check to see if there is a single Button/Icon element and setting it to 0 in that case.

redallen:
@KKoukiou I've added the check. Are there cases other than buttons or icons where you'd expect no delay?

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -97,6 +98,19 @@ export interface TooltipProps extends Omit<React.HTMLProps<HTMLDivElement>, 'con
 // id for associating trigger with the content aria-describedby or aria-labelledby
 let pfTooltipIdCounter = 1;
 
+function isButtonOrIcon(children: React.ReactNode) {
+  if (React.isValidElement(children)) {
+    if (['button', Button as any, 'svg', 'g'].includes(children.type)) {
```
KKoukiou, line 103: Checkbox, Radio we also use tooltips to 'explain' why elements are disabled. 

So when we disable a checkbox, we always wrap it with a tooltip and say 'Please blabla to enable this feature'

KKoukiou (commented):
I will think about thoroughly tomorrow if there is no need for this to be merged today.

kmcfaul (commented):
Would this be considering breaking as the default values are changing?

tlabaj (changes requested):
@redallen I agree with Katie.  I believe this is a breaking change.  I would be ok changing the defaults for the beta components only.

redallen:
When would this lead to unexpected behavior @tlabaj or @kmcfaul ? I consider it a bug that PatternFly tooltips show up instantly when in most operating systems and JS frameworks they have some delay. This leads to weird scenarios where pages can show more than one tooltip.

If you can think of cases where you expect a no-delay tooltip I can add them to the `isButtonOrIcon` check to leave them untouched.

mcarrano:
After discussing further with @redallen @tlabaj and @kmcfaul we have decided to set the default delay before showing a tooltip on hover to be 500ms. As was stated above, this is expected behavior for tooltips and is consistent with what is seen for most web applications and operating systems. The initial delay of 1 second (1000ms) was probably too long so at half a second it's enough time to sense that the user has stopped mouse movement over an element while preventing tootips from firing as the user is just moving the mouse over the screen. While I recognize that this is technically breaking I view the current behavior as an oversight vs intended.

tlabaj (approved):

kmcfaul (approved):

nicolethoen (approved):

mcarrano (approved):
Looks great. Thanks @redallen !

PR #5766, opened 2021-05-10
fix(Dropdown, Select): fix scrollbar click closing menu

### LINKED ISSUE #5703: Select/Dropdown closes when clicking scrollbar in order to scroll down and see all options

This may not be fixable due to the nature of click events on the document. I have a form with a Select that is near the bottom of the window, so the user needs to scroll down after opening it to see all the options. If they want to scroll by clicking and dragging the scrollbar, this triggers a `mousedown` event on the document, which causes the `onToggle` event to trigger and close the menu. They can't simply open it again after scrolling down, since its position at the bottom of the document means that when it is closed, the page shrinks a bit and scrolls back up. So the user must use a scroll wheel or gesture to scroll down and see the menu.

If possible, it would be ideal to find a way to detect whether that mousedown event is targeting the browser's scrollbar, and if so prevent the `onToggle` handler from being called. However, I see no clear way to make this detection, in my testing the event triggered when clicking the scrollbar has `event.target` set to the element underneath the scrollbar as if you clicked through it (this is on mac, where scrollbars are semi-hidden and float over the page). Solutions I've found online involve measuring whether the click location is in that scrollbar zone, which feels hacky to me.

Suggestions welcome.

Can be reproduced on all of the Select and Dropdown examples in our docs. Open a menu, use the scrollbar by dragging with the mouse to scroll down, observe that the menu closes.

I'd consider it a bug, this can potentially make using the scrollbar with the mouse pointer not viable in certain contexts.

Migration Toolkit for Virtualization. No particular target release.

kmcfaul:
**What**: Closes #5703

Replaces `mousedown` event handler trigger with `click` in Dropdown and Select toggles, to prevent clicking on the scrollbar from closing the menu. 

@tlabaj The behavior otherwise should remain the same. Was there a particular reason we went with `mousedown` at the time? I can't recall.

### DIFF
--- a/packages/react-core/src/components/Dropdown/Toggle.tsx
+++ b/packages/react-core/src/components/Dropdown/Toggle.tsx
@@ -61,13 +61,13 @@ export class Toggle extends React.Component<ToggleProps> {
   };
 
   componentDidMount = () => {
-    document.addEventListener('mousedown', this.onDocClick);
+    document.addEventListener('click', this.onDocClick);
     document.addEventListener('touchstart', this.onDocClick);
     document.addEventListener('keydown', this.onEscPress, { capture: true });
   };
 
   componentWillUnmount = () => {
-    document.removeEventListener('mousedown', this.onDocClick);
+    document.removeEventListener('click', this.onDocClick);
     document.removeEventListener('touchstart', this.onDocClick);
     document.removeEventListener('keydown', this.onEscPress, { capture: true });
   };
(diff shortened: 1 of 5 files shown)


kmcfaul:
Updated cypress. Some tests were toggling back open select menus when they were already closed by clicking actions.

mturley (approved):
LGTM!

nicolethoen (approved):

redallen (approved):

redallen:
We're releasing this week. There will be a prerelease available within the hour.

PR #5769, opened 2021-05-11
feat(LogViewer): refactor toolbar functionality

kmcfaul:
**What**: #5738

**In this PR**:
- Removes `LogViewerCustomControl`, `LogViewerToolbar`
- Adds `LogViewerSearch` containing search input
- Adds `toolbar` property to `LogViewer` (replacing `hasToolbar` and `customControls`)
- Adds `LogViewerToolbarContext` to wire up `LogViewerSearch` wherever it is passed within a toolbar

**Notes**:
- The `toolbar` property could possibly be named `header` instead, as this prop represents all content passed under the `pf-c-log-viewer__header`. Any preferences?
- `LogViewerSearch` will only function properly when passed into `LogViewer` via the `toolbar` prop due to the position of the context used to pass relevant data.

**Todos** (either here or in a follow-up):
- Allow persisted `searchedInput`
- Better complex toolbar example

https://patternfly-react-pr-5769.surge.sh/components/log-viewer


packages/react-log-viewer/src/LogViewer/examples/LogViewer.md
```diff
@@ -2,70 +2,102 @@
...
-    </React.Fragment>
-  );  
+  return (
+    <LogViewer
+      data={data.data}
+      toolbar={
+        <Toolbar>
+          <ToolbarContent>
+            <ToolbarItem>
+              <ToolbarGroup>
```
nicolethoen, line 41: I dont think you need to wrap the SearchInput in a ToolbarGroup

packages/react-log-viewer/src/LogViewer/examples/LogViewer.md
```diff
@@ -2,70 +2,102 @@
...
+      toolbar={
+        <Toolbar>
+          <ToolbarContent>
+            <ToolbarItem>
+              <Button onClick={onActionClick} variant="control">
+                Test Action
+              </Button>
+            </ToolbarItem>
+            <ToolbarItem>
+              <ToolbarGroup>
```
nicolethoen, line 88: same here

kmcfaul:
Extends SearchInputProps so user can hook into the component a bit better. Props are spread before several functions to allow users to hook into change/click events without overriding internal wireup.

tlabaj (approved):

nicolethoen (approved):

nicolethoen (commented):
the searches actually are not working atm. The values do not update

nicolethoen (changes requested):
the searches are not working atm in the examples

nicolethoen:
Looks good - only question is why does it automatically put focus into the second example's search?

kmcfaul:
Do you mean the automatic scroll to the first match? It did that before, but I think we can adjust it if that isn't the desired behavior.

nicolethoen (approved):

tlabaj (approved):

PR #5773, opened 2021-05-12
POC(Alert): Make dynamic alerts more accessible

### LINKED ISSUE #5731: POC: Make dynamic Alerts more accessible
@jessiehuff and @nicolethoen have been brainstorming a possible solution to making it easier for products to understand and implement dynamic alerts accessibly. The proposed solution needs to be implemented to be tested. It goes as follows:

Updates to Alert component:
- Title prop not required?
- Conditionally display the alert only when the title is provided
- When there is no title, only display the live region - empty - no visual alert or discoverable regions on the page, only in the html. Need to check other attributes on the .pf-c-alert are conditionally rendered properly when no title provided.
- Core update needs to hide empty alerts
- React update needs to adjust use of title attr

Update AlertGroup
- Make them live regions. If you don’t know how many alerts you will have on the page, put them in an AlertGroup

Things to communicate in docs:
- If a dynamic Alert will not be rendered in an AlertGroup: 2 options
  - it needs the isLiveRegion flag and CANNOT be rendered conditionally - the title must be passed conditionally.
  - It is rendered inside a live region. I.e. aria-live=”...”
- Dynamic alerts can be built conditionally and passed to an AlertGroup - the AlertGroup serves as the live region in this case. These alerts should not have the isLiveRegion flag.
- A static alert (no dynamic updates) does not need isLiveRegion

Update form examples to be using alerts correctly (i.e. everything needs a live region on page load)

[Here are the notes](https://docs.google.com/document/d/1bqWklSBG4H-biCknIFttvL5vGRiggwTOg8lPonSiSeU/edit) from Jessie's research and our brainstorming.

jessiehuff:
**What**: Closes #5731 

The theory behind this POC is that instead of consumers toggling the Alert itself, the Alert always exists on the page and they can toggle the title to make the alert appear. This allows the live region a11y attributes to exist on the page on load so that screen readers can hear the updates. Currently when consumers toggle the alerts, screen reader users (particularly of JAWS) don't hear the updates/alerts.

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -50,6 +50,7 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Set a custom icon to the Alert. If not set the icon is set according to the variant */
   customIcon?: React.ReactNode;
+  handleDismiss?: boolean;
 }
 
 export const Alert: React.FunctionComponent<AlertProps> = ({
@@ -73,6 +74,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   customIcon,
   onMouseEnter = () => {},
   onMouseLeave = () => {},
+  handleDismiss = false,
   ...props
 }: AlertProps) => {
   const ouiaProps = useOUIAProps(Alert.displayName, ouiaId, ouiaSafe, variant);
@@ -101,13 +103,18 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   const [timedOutAnimation, setTimedOutAnimation] = useState(true);
   const [isMouseOver, setIsMouseOver] = useState<boolean | undefined>();
   const [containsFocus, setContainsFocus] = useState<boolean | undefined>();
-  const [isDismissed, setIsDismissed] = useState<boolean | false>();
+  const dismissed = handleDismiss ? false : timedOut && timedOutAnimation && !isMouseOver && !containsFocus;
 
   React.useEffect(() => {
-    if (title && !isDismissed) {
-      setIsDismissed(timedOut && timedOutAnimation && !isMouseOver && !containsFocus);
+    timeout = timeout === true ? 8000 : Number(timeout);
+    if (timeout > 0) {
+      const timer = setTimeout(() => {
+        setTimedOut(true);
+        handleDismiss && onTimeout();
+      }, timeout);
+      return () => clearTimeout(timer);
     }
-  }, [timedOut, timedOutAnimation, isMouseOver, containsFocus, title]);
+  }, [title, handleDismiss]);
 
   React.useEffect(() => {
     timeout = timeout === true ? 8000 : Number(timeout);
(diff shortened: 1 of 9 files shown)


nicolethoen:
Looking so good! The timer should pause when the user hovers over the alert or puts focus on the alert.

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -23,7 +23,7 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   /** Flag to indicate if the Alert is inline */
   isInline?: boolean;
   /** Title of the Alert  */
-  title: React.ReactNode;
+  title?: React.ReactNode;
```
tlabaj, line 26: I am not understanding why the title is no longer required?

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -50,6 +50,7 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Set a custom icon to the Alert. If not set the icon is set according to the variant */
   customIcon?: React.ReactNode;
+  handleDismiss?: boolean;
```
tlabaj, line 53: this os missing a comment

packages/react-core/src/components/Alert/Alert.tsx
jessiehuff: So the theory behind this POC is that instead of consumers toggling the Alert itself, the Alert always exists on the page and they can toggle the title to make the alert appear. This allows the live region a11y attributes to exist on the page on load so that screen readers can hear the updates. Currently when consumers toggle the alerts, screen reader users (particularly of JAWS) don't hear the updates/alerts.

seanforyou23: >allows the live region a11y attributes to exist on the page on load so that screen readers can hear the updates

I think this was the main use case for `AlertGroup`. You're correct that screen readers often require live regions to exist on page load when the AOM is first evaluated, otherwise changes to those regions may not be communicated. This in large drove the creation of AlertGroup (that plus the need for "toast" alert positioning).

See this [example](https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html#ARIA19-examples) and notice the error(s) container they use (`<p id="errors" role="alert" aria-atomic="true"></p>`) gets the live region - this is analogous to our AlertGroup component - they add and remove from that container chunks of markup that look like `<p>Please enter your first name.</p>` which is like an Alert. It's a container that generally carries the live region properties. For example, it probably wouldn't make too much sense to set [aria-relevant](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-relevant) on a single Alert.. I think this is at the heart of the distinction.

Making the Alert do all of the above could work, but I'm not sure it makes sense to have a unique live region for every Alert on a page. Consider the case where there's a Notification drawer with many dynamic content changes, there should probably be a single live region to house them all. IMO any given page may commonly have a handful of high-level areas designated as live regions (think live regions for a form's validation, an entire notification drawer, an active countdown timer, etc.).

When you think of how overlapping alerts can easily interrupt each other and become "chatty", it makes a good case for being conservative on the total number of live regions we recommend users to create on any single page. Also, consider potential performance impacts from the browser listening to and monitoring many [live] regions on a page as opposed to only a few.

>Currently when consumers toggle the alerts, screen reader users (particularly of JAWS) don't hear the updates/alerts.

Do you know if AlertGroup happens to be missing from the equation in cases where you noticed the above? If so, might be that we're not explaining AlertGroup well enough so people are not using it when they need to be. I don't think we mention anywhere that AlertGroup should always exist on a page even when there are no Alerts.. Just another point of view to consider, hope it helps 🙂.

PR #5787, opened 2021-05-14
fix(Dropdown, Select): fix focus on toggle after click off menu

### LINKED ISSUE #5788: Toolbar: selecting the text in a text filter in the toolbar moves focus to the first item in the toolbar.
Before updating to pf 2021-6:

https://user-images.githubusercontent.com/35978579/118334761-392d3180-b4dc-11eb-9534-5263658a9ed2.mov

After updating to pf 2021-6:

https://user-images.githubusercontent.com/35978579/118334786-44805d00-b4dc-11eb-88e8-fa6e13fa76e1.mov

Bug
OpenShift 4.8

kmcfaul:
**What**: Closes #5788 

Focus shouldn't need to be manually set after a `click` event fires outside of an open menu, the clicked element should receive focus rather than the toggle. The menu should still close when a user clicks outside of the menu.

@jessiehuff Removing this shouldn't cause any accessibility issues right? This should be pure mouse behavior, where `click` already focuses the event target.

### DIFF
--- a/packages/react-core/src/components/Dropdown/Toggle.tsx
+++ b/packages/react-core/src/components/Dropdown/Toggle.tsx
@@ -79,7 +79,6 @@ export class Toggle extends React.Component<ToggleProps> {
     const clickedWithinMenu = menuRef && menuRef.contains && menuRef.contains(event.target as Node);
     if (isOpen && !(clickedOnToggle || clickedWithinMenu)) {
       onToggle(false, event);
-      this.buttonRef.current.focus();
     }
   };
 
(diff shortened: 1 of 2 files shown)


kmcfaul:
If there is a use case for manually focusing the toggle instead of whatever element is clicked outside the menu, both components can adjust their `onToggle` (both) or `onClose` (Select) callbacks.

tlabaj (approved):

nicolethoen (approved):

jessiehuff (approved):
I think this is probably fine. I assume this will allow the browser to decide what to do with focus on mouse click, though it may be worth noting that often different browsers have different behaviors with how they handle focus and mouse clicks so we should probably still keep our eye on this in the future.

PR #5789, opened 2021-05-17
feat(menu): context menu flyout

### LINKED ISSUE #5642: Menu - flyout menus do not flip sides to prevent clipping
Flyout menus should be able to flip sides/direction depending on the available space in the window.

This is follow up work from the core issue https://github.com/patternfly/patternfly/issues/3938 and PR https://github.com/patternfly/patternfly/pull/3977 which can be used as a reference for positioning.

redallen:
**What**: Closes #5642

**Additional issues**:


redallen:
This is pending further discussion on flyouts behaving like context menus. For example, instead of going off the bottom of the page, should flyouts open partially upwards like menu items 4 and 5 do here?
context-menu-clipping

mcarrano:
@christianvogt just wanted to update you on this issue that the React team is currently working on. It is a follow-up to the request here: https://github.com/patternfly/patternfly/issues/3938. See @redallen 's comment above. Turns out that this is  more complex than first thought. I know that you were looking for a fix to include in OS 4.8 which has a code freeze next week, I believe? That looks inloikely at this point. I wanted to check back with you about the urgency of this and try to understand exactly the impact of not having this fix available would be. In any case, we will keep working on this and should be able to have a more robust flyout solution in time for OpenShift 4.9.

christianvogt:
We don't require this change for 4.8 because we chose not adopt the menu component until this was addressed. We will continue to use our own menus until this one can be adopted in 4.9.

mcarrano:
Thanks @christianvogt .

mcarrano:
This is looking good @redallen

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -77,20 +80,80 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
...
+      }
+    }
+  }, [flyoutVisible, flyoutMenu]);
+
+  React.useEffect(() => {
+    const flyoutMenu = ref.current.lastElementChild as HTMLElement;
+    if (flyoutMenu && flyoutMenu.classList.contains(styles.menu)) {
+      flyoutMenu.style.right = flyoutXDirection === 'right' ? 'auto' : `calc(100% - ${offsetX}px`;
+      flyoutMenu.style.left = flyoutXDirection === 'left' ? 'auto' : `calc(100% - ${offsetX}px`;
+      flyoutMenu.style.top = `calc(-${offsetY}px + var(--pf-c-menu--m-flyout__menu--Top))`;
```
mcoker, line 149: Can you try doing this via 

```suggestion
      flyoutMenu.style.setProperty = `--pf-c-menu--m-flyout__menu--top-offset: ${offsetY}px`;
```

I don't know if this syntax is right, but basically passing the offset to the var we created and allowing core to handle the `calc()` - https://github.com/patternfly/patternfly/blob/6fe562844cf6e99990603c2885de1911a37fd9b7/src/patternfly/components/Menu/menu.scss#L10

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -77,20 +80,80 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
...
+        }
+      }
+    }
+  }, [flyoutVisible, flyoutMenu]);
+
+  React.useEffect(() => {
+    const flyoutMenu = ref.current.lastElementChild as HTMLElement;
+    if (flyoutMenu && flyoutMenu.classList.contains(styles.menu)) {
+      flyoutMenu.style.right = flyoutXDirection === 'right' ? 'auto' : `calc(100% - ${offsetX}px`;
+      flyoutMenu.style.left = flyoutXDirection === 'left' ? 'auto' : `calc(100% - ${offsetX}px`;
```
mcoker, line 148: see if you can use `--pf-c-menu--m-flyout__menu--left-offset` here

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -77,20 +80,80 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
...
+          }
+        }
+      }
+    }
+  }, [flyoutVisible, flyoutMenu]);
+
+  React.useEffect(() => {
+    const flyoutMenu = ref.current.lastElementChild as HTMLElement;
+    if (flyoutMenu && flyoutMenu.classList.contains(styles.menu)) {
+      flyoutMenu.style.right = flyoutXDirection === 'right' ? 'auto' : `calc(100% - ${offsetX}px`;
```
mcoker, line 147: try `--pf-c-menu--m-flyout__list-item--m-menu-left__menu--right-offset`

mcoker (changes requested):

mcarrano (approved):
Looks good from a UX perspective. Thanks @redallen .

packages/react-core/src/components/Menu/MenuItem.tsx
redallen: Unfortunately left/right offsets don't work well since I'm not using the `pf-m-menu-left` class. I can't use the class since if I do apply it _to the parent_ the consumer might just override it with some bit of state inside the menu.

packages/react-core/src/components/Menu/MenuItem.tsx
redallen: Done.

packages/react-core/src/components/Menu/MenuItem.tsx
mcoker: Ah! Sounds like I could just create generic left/right offset vars that aren't attached to a class?

Though `--pf-c-menu--m-flyout__menu--left-offset` isn't attached to a class, so I would expect that to work. `--pf-c-menu--m-flyout__list-item--m-menu-left__menu--right-offset` works via `.pf-m-menu-left`, but I can update that to just be `--pf-c-menu--m-flyout__menu--right-offset`

redallen: I think I could use the CSS vars and add the class if this selector was just `.pf-c-menu.pf-c-menu-left`: https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Menu/menu.scss#L184

dlabrecq (approved):

tlabaj (approved):
Looks great Zack!

christianvogt:
Found some issues:
- Left arrow does not close a menu item when a group is selected.
- Esc does not close a menu item when a group is selected.
- Esc closes all menus on my mac but only closes 1 sub menu in this. What's the expectation with PF.
- Mouse out closes the entire menu even if you mouse out then quickly return. My mac menus remain even when you mouse out. Not sure which direction PF is going but it's easy to slip off the menus.
- Focus is not transferred to the menu item when first opened and therefore cannot use keyboard navigation.

redallen:
> Left arrow does not close a menu item when a group is selected.
> Esc does not close a menu item when a group is selected.

This was a bug in existing code, fixed.

> Esc closes all menus on my mac but only closes 1 sub menu in this. What's the expectation with PF.

In Chrome it closes only one submenu and that's what I expect.

> Mouse out closes the entire menu even if you mouse out then quickly return. My mac menus remain even when you mouse out. Not sure which direction PF is going but it's easy to slip off the menus.

I spent about an hour playing with this and it's possible if we refactor the `flyoutVisible` state to be in `Menu` rather than `MenuItem`. I estimate the refactor will take a day since all the keyboard logic will have to be moved along with it. @mcarrano is it worth it in this PR?

> Focus is not transferred to the menu item when first opened and therefore cannot use keyboard navigation.

It's outside the scope of this PR to make dropdown use menus, but I've added this functionality to the example for now.

mcarrano:
>    Mouse out closes the entire menu even if you mouse out then quickly return. My mac menus remain even when you mouse out. Not sure which direction PF is going but it's easy to slip off the menus.

@redallen this seems like the opposite of what I'm seeing. I tested on Firefox and Chrome and the flyouts remain open once triggered, even if I mouse off onto another menu item. Did something change?

redallen:
@mcarrano I prototyped @christianvogt 's suggestion but missed a line while reverting, just pushed a fix that keeps the behavior of onMouseLeave hiding the menu.

mcarrano:
Thanks @redallen . I do see what you mean now about the menu dismissing on mouse_leave @christianvogt . This is one of the big reasons I'm not a big fan of flyout menus, in general. 

@redallen could we try to address this by adjusting the dwell timer a bit rather than rewriting this? I think you said it's currently 200ms. Maybe try increasing that a bit to see if that decreases the likelihood of dismissing the menu accidentally?

redallen:
I've increased it to 1000ms, preview should deploy shortly.

mcarrano:
I think this is better @redallen . Maybe a tad too long, but you also need to somehow cancel the mouse_leave event if the user re-enters the region of the menu before the timer expires. Right now if I mouse away and immediate come back, the menu dismisses, anyway.

redallen:
I've added a 700ms timeout for the user to `onMouseEnter` after `onMouseLeave`. Let me know what you think @christianvogt and @mcarrano.

christianvogt:
When switching to menus items within the current menu (be it an action or another sub menu), I think it should be immediate. Otherwise you can end up with multiple open menus.

pf-menu

redallen:
Right, that's the downside of having a timeout. The way to fix it is to have each menu know about its submenus. I'll spend the day to refactor later this week...

redallen:
I've fixed the bug @christianvogt mentioned. Instead of a timeout, the context menu stays open until you mouse over a sibling menu element. Please rereview @mcarrano and @christianvogt !

christianvogt:
Do we really want `esc` to only close one sub menu instead of the entire thing? left arrow can already be used to close one sub menu.

I found an issue where if you press the esc or left arrow key, the mouse out timeout no longer functions.

redallen:
The `esc` behavior matches what Chrome does with its context menus. I'll defer to @mcarrano for whether it should close *all* submenus, but that's something I've left alone in this PR.

The fix in my last commit was to remove usage of `onMouseExit` and instead rely on `onMouseOver` of sibling elements to close a flyout menu. This makes mouse slips more difficult since you'd have to slip to a sibling element and again matches Chrome's behavior.

> I found an issue where if you press the esc or left arrow key, the mouse out timeout no longer functions.

There aren't any timeouts in use I'm aware of, can you share a screencapture?

christianvogt:
My mistake. I got confused between the previous change and the latest where I expected the menu to close after some time away.

mcarrano:
@redallen @christianvogt While I can see value in having `esc` dismiss the entire menu, I'm also OK with maintaining consistency with Chrome. @jessiehuff can you also give an opinion on this? Are there any keyboard accessibility standards we should be following?

Also, I'm good with the removal of the timeout on `onMouseExit` however I should still be able to close the menu by clicking away from it or tabbing out of it, I think.

redallen:
I've gone ahead and put the example `<Menu>` in a proper `<Dropdown>` (I only needed minor changes to get it working). This provides the keyboard and mouse interaction you expect from a Dropdown menu @mcarrano and shortens the example so you don't have to handle basic keyboard interaction (except `onEnter` focusing the first item) @christianvogt .

As a side effect of Dropdown's existing interaction @christianvogt has been granted his wish of Escape closing the entire menu once opened. I can see the argument that this menu is a `Dropdown` that should continue with the existing behavior or the argument that this menu is a new class of ContextMenu and Escape should only close a single submenu.

Whatever the case, let us know your verdict @jessiehuff and I'll make changes if necessary.

mcarrano:
This looks great. Thanks @redallen !

redallen:
Updated with latest core changes. Please rereview @mcarrano @mcoker and @christianvogt !

jessiehuff (commented):
I'm curious what the approach would be on mobile or on touch for this? When testing this on my phone, I couldn't figure out a way to navigate the menu. It makes me wonder if Enter/click should open the submenu, move focus to the submenu item, and have the menu be persistent? This would follow the [WAI ARIA practices](https://www.w3.org/TR/wai-aria-practices-1.2/#menu) that are common for a menu and solve that issue for mobile. However, if we do enable the on click behavior to open a submenu then we should also switch off the behavior we have where the submenu would disappear when you hover over others. (So when the user presses Enter, that submenu stays open even when hovering over other menu items.) I think this might also help users with limited motor control that would struggle to move the mouse in very specific patterns. This would make sure that [WCAG 2.5.1 ](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=251#pointer-gestures)(All functionality that uses multipoint or path-based gestures for operation can be operated with a single pointer without a path-based gesture) isn't an issue.

In terms of the Esc key, I think it makes sense to follow the WAI ARIA practices here as well that it would "Close the menu that contains focus and return focus to the element or context, e.g., menu button or parent menuitem, from which the menu was opened. So based on this, I think it makes the most sense that pressing Esc would close each submenu that you are currently in.

When testing with JAWS and NVDA, I'm not able to navigate into the menu at all. I wonder if adding the enter/click interaction will enable this for these screen readers. For VO, it automatically enters into the submenus making it confusing to navigate. I'm thinking that if we follow the [properties suggested in the WAI ARIA](https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties-14) if it'll help announce to screen readers that there is a menu. (i.e. parent menus having `aria-haspopup="true"` and `aria-expanded={submenu isOpen}`)

My two cents anyway. :)

redallen:
> I'm curious what the approach would be on mobile or on touch for this?

I can make `touchEnd` do the same thing as focusing and pressing enter. I think this addresses the rest of your paragraph.

This doesn't follow the Chrome or OS interaction of context menus, but it's easy to implement if @mcarrano thinks it correct. 

> So based on this, I think it makes the most sense that pressing Esc would close each submenu that you are currently in.

Okay, I'll override Dropdown's behavior to do that. Sorry @christianvogt .

> i.e. parent menus having aria-haspopup="true" and aria-expanded={submenu isOpen}

I can add these.

Changes will come later this week but those I tagged above can still review non-a11y parts.

mcarrano:
@redallen I reviewed this morning and everything looked good excepting @jessiehuff 's comments.

As for what to do about touch interactions, I think that we should follow WAI best practices as much as possible. Not sure how that may differ from what Chrome is doing, but once you make those changes, we can review again.

mcoker (commented):
Looks like you're still using inline `left` and `right` styles to position the menu to the left/right. Are you able to use the class `.pf-c-menu.pf-m-left` now for left flyout (and no class for right flyout)? https://github.com/patternfly/patternfly/blob/5df38ca12dd3816dccd996d35d002db4b2981dfd/src/patternfly/components/Menu/menu.scss#L175-L178

Weird thing is the [core surge site](https://pf4.patternfly.org/components/menu) is listing the old modifier (`.pf-m-menu-left`).

Otherwise this looks great to me!

redallen:
@mcoker Fixed.

mcoker (dismissed):
L🥳TM!!

redallen:
@jessiehuff I've added the aria attributes and changed the escape key behavior to close submenus rather than the whole menu. For the demo I've made clicking (or touching) a MenuItem with a flyout not close the menu.

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -77,20 +84,85 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
...
+            flyoutMenu.style.setProperty(rightOffset.name, `-${xOffset}px`);
+          } else {
+            flyoutMenu.style.setProperty(leftOffset.name, `-${xOffset}px`);
+          }
+
+          const spaceLeftBot = window.innerHeight - origin.y - rect.height;
+          const spaceLeftTop = window.innerHeight - rect.height;
+          if (spaceLeftTop < 0 && spaceLeftBot < 0) {
+            // working idea: page can usually scroll down, but not up
+            // TODO: proper scroll buttons
```
kmcfaul, line 153: Is this going to be implemented now or in a follow up?

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -113,54 +113,88 @@ class MenuIconsList extends React.Component {
...
-      <Menu containsFlyout onSelect={this.onSelect} activeItemId={activeItem}>
+  const onSelect = (_ev, itemId) => {
+    console.log('select', itemId);
+    if (!itemId.includes('next-menu')) {
+      setIsOpen(false);
+    }
+  };
+
+  return (
+    <Dropdown
```
kmcfaul, line 166: Would it be worth following our composable menu demos (using MenuToggle & Popper) instead of using Dropdown, and moving this example to the composable menu page? The other Menu examples don't use a toggle so this example looks slightly out of place

packages/react-core/src/components/Menu/MenuItem.tsx
redallen: There are no designs or HTML/CSS for vertical menu scroll buttons so I'd say in a followup if a future issue is opened.

packages/react-core/src/components/Menu/examples/Menu.md
redallen: I was unaware of that section and I agree flyout fits better there and should use MenuToggle & Popper instead of `<Dropdown>`. I'll address that in my next commit.

jessiehuff (commented):
It looks like you solved the mobile issue, and a lot of the keyboard interactions look good! However, I'm guessing that without the ability to click on the menu item with a flyout, it seems like NVDA and JAWS aren't able to navigate into any of the flyouts. (In testing I can't seem to get the submenus to open.) Would it be possible to keep the menu persistent when it's clicked as well? Also, the aria attributes should be on the buttons with flyout [menus](https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties-14).
(I meant parent menu item buttons that have submenus in my previous comment - I'm sorry! I should have been clearer :) ) 

> - The element with role button has aria-haspopup set to either menu or true.
> - When the menu is displayed, the element with role button has aria-expanded set to true. When the menu is hidden, it is recommended that aria-expanded is not present. If aria-expanded is specified when the menu is hidden, it is set to false.

redallen:
I addressed @kmcfaul 's feedback and moved the example to the composable menu section. I refactored all those examples to have focus the first entry on open as expected and fixed the new KeyboardHandler helper to work.

I addressed @jessiehuff 's feedback and removed those aria attributes which seemed to confuse voiceover. To get Voiceover to behave like https://pf4.patternfly.org/components/menu/html/with-flyout/ I removed some of the `tabindex` code used for drilldown while making sure it still behaved the same on Voiceover.

redallen:
After meeting with @jessiehuff and @jgiardino today I added [aria menubar roles](https://www.w3.org/TR/wai-aria-practices-1.2/examples/menubar/menubar-1/menubar-1.html) and added back @kmcfaul 's roving which have worked for Voiceover and NVDA for me locally. Awaiting rereview...

kmcfaul (approved):
Looks great!

I did notice that tab/shift-tab to get between drilldown breadcrumbs and drilldown menu items is a bit wonky but it's the same on org so may be better as a follow up.

jessiehuff (approved):
Absolutely awesome job with this!!! 🏆 🎉  It works in JAWS, NVDA, and VO (as well as keyboard)! Woot woot! Thanks for all of your work on this! 😄

PR #5793, opened 2021-05-18
feat(select): allow filtering in single variant

### LINKED ISSUE #5764: Select: enable hasInlineFilter for single select
Enable `hasInlineFilter` for single select / grouped single select variants in addition to the checkbox select variant.

### LINKED ISSUE #5847: Accessibility bug in the Select (Grouped checkbox input with filtering) can not be navigated with the arrow keys and component doesn't move cursor to search input on click
The `<Select>` component [Grouped checkbox input with filtering](https://www.patternfly.org/v4/components/select/#grouped-checkbox-input-with-filtering) does not insert the cursor in the `input` field on click. And thus keyboard accessibility is broken, cannot be navigated by the arrow keys. 

https://codesandbox.io/s/bi6pb
Click on "Filter by status". Cursor focus should automatically be within filter input, allowing the user to start typing to filter list or use the arrow keys to traverse list.

Bug

OpenShift switched to use the `<Select>` component recently https://github.com/openshift/console/pull/8877 
version 4.8

redallen:
**What**: Refactor `SelectMenu.tsx` to allow `hasInlineFilter` to work for non-checkbox variants.
Closes https://github.com/patternfly/patternfly/issues/4047
Closes #5764
Closes #5847

https://patternfly-react-pr-5793.surge.sh/components/select#grouped-single-with-filtering

Bug fixes:
- Stop entire page scrolling when pressing down arrow when focused on filter
- Clicking on typeahead variant's input while open now closes them
- Focus inline filter input when opening with mouse

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -145,7 +145,7 @@ export interface SelectProps
 }
 
 export interface SelectState {
-  openedOnEnter: boolean;
+  focusFirstOption: boolean;
   typeaheadInputValue: string | null;
   typeaheadFilteredChildren: React.ReactNode[];
   favoritesGroup: React.ReactNode[];
@@ -211,7 +211,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
   };
 
   state: SelectState = {
-    openedOnEnter: false,
+    focusFirstOption: false,
     typeaheadInputValue: null,
     typeaheadFilteredChildren: React.Children.toArray(this.props.children),
     favoritesGroup: [] as React.ReactNode[],
@@ -230,7 +230,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       this.refCollection[0][0] = this.filterRef.current;
     }
 
-    if (!prevState.openedOnEnter && this.state.openedOnEnter && !this.props.customContent) {
+    if (!prevState.focusFirstOption && this.state.focusFirstOption && !this.props.customContent) {
       const firstRef = this.refCollection.find(ref => ref !== null);
       if (firstRef && firstRef[0]) {
         firstRef[0].focus();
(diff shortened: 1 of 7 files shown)


mcarrano (commented):
@redallen @tlabaj what should I be looking for here? Is there an example that shows this in action?

redallen:
@mcarrano Please review this example: https://patternfly-react-pr-5793.surge.sh/components/select#grouped-single-with-filtering

mcarrano (commented):
Looks good to me @redallen .

@gdoyle is this what you expected/

nicolethoen (commented):
This looks good to me - I think we should probably have at least one test for our built in filtering capabilities?

jessiehuff (commented):
I noticed that keyboard interaction doesn't work on this example. :( Can we add that? I would assume that it would work like the other grouped filtering examples.

redallen:
@jessiehuff I'm seeing with **all** filtering examples pressing the down or up arrow while on the filter field scrolls the entire page. On my example specifically the down arrow skips the first item ("Running") which I can fix. What else "doesn't work"?

keyboard-working

jessiehuff (commented):
I was trying to test out different variations of filtering, and it appeared to me that when using the filtering that the arrow keys weren't working as expected. Trying it out again to record, I think it's just the skipping of the first item that made it appear to not filter and still function.
https://user-images.githubusercontent.com/28281340/119692822-bbd9b900-be19-11eb-9901-cd2a8372718f.mov

kmcfaul:
LGTM. For the keyboard blip, I think there's some extra logic to add the input to the refCollection Select uses to navigate items. It's possible something is going wrong there? Or its overriding the first slot instead of injecting the input ref.

sg00dwin:
@jessiehuff I just opened this related issue. https://github.com/patternfly/patternfly-react/issues/5847 

On click the cursor focus does not move to the input field and thus the arrow keys do not traverse the list.  Should this function as such?

select-arrow-bug

redallen:
@sg00dwin You checked before the preview updated. Try again.
keyboard-working

kmcfaul (approved):
LGTM!

dlabrecq (approved):

sg00dwin:
> @sg00dwin You checked before the preview updated. Try again.

Yep. awesome thanks!

tlabaj:
Looks good code wise. Should the keyboard interaction for moving from the input to the menu match the checkbox select (tab rather than arrow down)? I will differ to @jessiehuff  on that.

tlabaj (approved):

gdoyle1 (approved):
LGTM!

jessiehuff (approved):
This looks pretty good to me! In an ideal world I believe Tab would go between the main regions (the search and the menu) while arrow keys would function within the region (in this case in the menu), and this would probably use a roving tabIndex approach like our menu does. However, I think this is fine for now, especially without making this PR get too complicated. So LGTM! 😄

Issue #5586, opened 2021-03-26
Interactive legend example for pie chart


dlabrecq:
As a user, I would like to see an interactive legend example for the pie chart. Pie and donut charts don't use a Chart container like ChartArea, ChartLine, etc. Therefore, there isn't a parent container to attach events to.

dlabrecq:
See https://github.com/patternfly/patternfly-react/pull/5795

PR #5795, opened 2021-05-18
Interactive legend example for pie chart

dlabrecq:
This provides an example showing how to use a pie chart with an interactive legend. 

Note that pie charts are not typically wrapped by a `Chart` component, but in this case we need something to attach `onMouseOver` events to.

https://github.com/patternfly/patternfly-react/issues/5586

chrome-capture


ddonahue007 (approved):
LGTM

redallen (approved):

PR #5808, opened 2021-05-19
feat(Button): add danger modifier to link/secondary buttons

### LINKED ISSUE #5578: Button - add danger variant to secondary and link variations
re: https://github.com/patternfly/patternfly/pull/3932/

Adds `.pf-m-danger` as a variant to `.pf-m-secondary` and `.pf-m-link` buttons.

cc @mcarrano

kmcfaul:
**What**: Closes #5578

Adds `isDanger` flag. When the button variant is `link` or `secondary`, danger styling will be added to the button with the new flag.

Should the `danger` variant be removed in the next breaking change release? 
And should the `isDanger` flag be made to add danger styling to any variant or is there a reason to limit the flag to just link/secondary?

### DIFF
--- a/packages/react-core/src/components/Button/Button.tsx
+++ b/packages/react-core/src/components/Button/Button.tsx
@@ -60,6 +60,8 @@ export interface ButtonProps extends React.HTMLProps<HTMLButtonElement>, OUIAPro
   isSmall?: boolean;
   /** Adds large styling to the button */
   isLarge?: boolean;
+  /** Adds danger styling to secondary or link button variants */
+  isDanger?: boolean;
 }
 
 export const Button: React.FunctionComponent<ButtonProps> = ({
@@ -71,6 +73,7 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
   isDisabled = false,
   isAriaDisabled = false,
   isLoading = null,
+  isDanger = false,
   spinnerAriaValueText,
   isSmall = false,
   isLarge = false,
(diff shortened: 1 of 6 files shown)


kmcfaul:
Also updated the examples to have consistent spacing, prettier is unhelpfully trying to be helpful on an autosave with our settings, removing the `{' '}` spacing for some buttons but not others.

mcoker (approved):
lgtm!

mcarrano (commented):
I think that the way you implemented this is fine, but you raise some good questions about this @kmcfaul .  Not sure I can think of other places where the danger flag would apply. For example, applying it to the Tertiary button would make appearance the same as a secondary danger button, but maybe we should allow that for parity? As you say, this leaves us with two distinct ways this is handled in the API unless and until we can make a breaking change to have everything use the `isDanger` flag. Not sure if that will be confusing to developers. @tlabaj @nicolethoen what are your thoughts?

kmcfaul:
Adding the `isDanger` flag to work with all variants would not be breaking (as the flag was added in this PR), but removing `danger` from the `variant` property would be breaking. The question is mainly whether `isDanger` makes sense to work with any variant.

mcarrano:
@kmcfaul in that case, I think it should. Do you agree? So I could apply it to any button if I want. I guess applying it to the primary would be the same result as using the current Danger variant, but I think that's OK. Does that make sense to you?

nicolethoen (approved):
I agree that it doesn't seem to be incredibly clear from the docs when using `isDanger` will apply danger styles to a `Button`. The component is relatively simple and playing with it for a few moments, I understand now - plus the prop description is clear. 

In the long run it probably makes sense to avoid having a `danger` variant AND an `isDanger` prop, since that feels like a conflict, but we also need to avoid breaking changes. 

Life is all about trade offs.

kmcfaul:
@mcarrano I tested the `isDanger` flag with other variants and it was slightly inconsistent. Primary buttons changed to look like the current Danger button. Tertiary buttons still had a black border (but did change to a red background), and Warning/Control/Plain buttons did not change at all.

I think for now we should leave the flag for the `secondary` and `link` button variants, as the prop description states currently. We can always extend support for other variants later on in a follow up if desired.

mcarrano (approved):
@kmcfaul I'm fine with that approach if others on the dev side agree.

redallen (approved):

dlabrecq (commented):
LGTM

PR #5812, opened 2021-05-20
fix(jumplinks): check for stale DOM node

### LINKED ISSUE #5811: JumpLinks can point to stale item
In kafka-ui JumpLinks grabs DOM nodes which are later unmounted and the component does not fetch new ones leading to the scroll spy breaking.

redallen:
**What**: Closes #5811

**Additional issues**:


srambach (approved):
Fixed the problem when we tested; thank you!

KKoukiou (approved):

PR #5824, opened 2021-05-24
feat(Menu): filter tree demo

### LINKED ISSUE #5655: Filter tree demo
Create a new React demo that shows how a Tree view can be placed inside of a dropdown to satisfy the use case described here: https://github.com/patternfly/patternfly-design/issues/912  - see comments on desired solution and behavior.

kmcfaul:
**What**: Closes #5655

First pass on filter tree demo. 
PR also:
- Allows TreeView leaf nodes to have badges if `customBadgeContent` is defined
- Fixes a console error for TreeView if a flat tree (no folders/children) is rendered
- Updates to cleaner logic check for composable menu toggling closed via click/keyboard
<br />

@mcoker I think this will need a css update as TreeView currently doesn't expect leaf nodes on the top level (so there isn't any spacing to keep the checkboxes inline). If we can add a modifier/flag for top level leaf nodes, I can flag them individually (as the data is known prior to rendering).

@jessiehuff Keyboard navigation in this demo is currently wonky due to the multiple TreeViews. The keyboard handling works the same per single checkbox TreeView, but that already uses Tab to go between the toggle/checkbox elements which isn't ideal in a Menu context. You currently also have to `Tab` to get to a following TreeView block. I'm still playing around with it, but I think the demo may have to implement its own handler for everything. Time depending, this might be a follow up issue.

link: https://patternfly-react-pr-5824.surge.sh/demos/composable-menu


mcarrano (commented):
@kmcfaul this is looking good with the exception of the issues you mentioned. @mcoker is it a quickly fix for the menu alignment problem or do we need to open a new issue for that?

mcoker:
@kmcfaul @mcarrano I've put up a core PR with the change here - https://github.com/patternfly/patternfly/pull/4104

packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -396,3 +396,257 @@ class SelectComposableMenu extends React.Component {
...
 }
 ```
+
+### Filter tree menu
+
+```js
+import React from 'react';
+import { MenuToggle, Menu, MenuContent, MenuGroup, MenuList, MenuItem, Popper, TreeView } from '@patternfly/react-core';
+
+class BasicComposableMenu extends React.Component {
```
tlabaj, line 406: the name of the class should be changed

tlabaj (changes requested):

kmcfaul:
Updated the demo name and rebased with master, which contains the version bump to fix the styling.

mcoker (approved):
LGTM. I'd like to review the markup and see if we need some sort of container for components that can go in the menu. Opened a core issue here for that - https://github.com/patternfly/patternfly/issues/4116

mcarrano (approved):
This looks great. Thanks @kmcfaul .

tlabaj (approved):

dlabrecq (commented):
LGTM

Issue #5597, opened 2021-03-29
Table: Tree table - update checkbox behavior in examples


nicolethoen:
Update the example to demonstrate how the consumer can use group checking, where checking a checkbox will check all children rows, and checking a children row will partially check a parent row.

Since both use cases have been described by products, we should probably allow the user to toggle between both the current and desired checkbox behaviors.

It should behave the same way it does in the TreeView component.

PR #5827, opened 2021-05-24
fix(Table): change checkbox behavior in tree table examples

nicolethoen:
Closes: #5597

Also fixes a console warning coming from the Composable actions example.

### DIFF
--- a/packages/react-table/src/components/Table/examples/Table.md
+++ b/packages/react-table/src/components/Table/examples/Table.md
@@ -30,7 +30,7 @@ import LeafIcon from '@patternfly/react-icons/dist/js/icons/leaf-icon';
 import FolderIcon from '@patternfly/react-icons/dist/js/icons/folder-icon';
 import FolderOpenIcon from '@patternfly/react-icons/dist/js/icons/folder-open-icon';
 
-import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
+import { Checkbox, ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
 
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Table/table';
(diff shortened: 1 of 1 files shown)


mcarrano (approved):
This looks good to me @nicolethoen . I tested with the mouse, only. Should we have @jessiehuff verify keyboard behavior? I'm not sure what the expected keyboard interaction is or if it is affected by this change.

jessiehuff (commented):
I would think that the keyboard interaction would match the tree view component. (An example to take a look at would be the ["with actions" tree view example](https://www.patternfly.org/v4/components/tree-view#with-action-items).) I believe that Nicole was working on coming up with a solution for this, but I think there were some complications involved in the complexity of table and all the possibilities of what can be in the table.

nicolethoen:
I don't think the keyboard behavior has changed from what's currently been released for the tree table, but we are still working on enhancements in a separate issue. @mcarrano

kmcfaul (approved):

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -30,7 +30,7 @@ import LeafIcon from '@patternfly/react-icons/dist/js/icons/leaf-icon';
 import FolderIcon from '@patternfly/react-icons/dist/js/icons/folder-icon';
 import FolderOpenIcon from '@patternfly/react-icons/dist/js/icons/folder-open-icon';
 
-import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
+import { Checkbox, ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
```
redallen, line 33: You don't need to import anything from `@patternfly/react-core` or `@patternfly/react-table` since the theme does it for you automatically (but it doesn't hurt).

redallen (approved):
Thanks @nicolethoen !

PR #5845, opened 2021-05-26
feat(Tabs) Added support for expandable vertical tabs 

### LINKED ISSUE #5520: Expandable vertical tabs
https://github.com/patternfly/patternfly/pull/3836/

Adds  3 variations
* `.pf-m-expandable` - This applies to the vertical variation and can be applied at different breakpoints via `.pf-m-expandable-on[breakpoint]`
* `.pf-m-non-expandable` can be used to go back to regular vartical tabs via something like `.pf-m-expandable.pf-m-non-expandable-on-lg` to make vertical tabs expandable by default, and non-expandable on large viewports.
* `.pf-m-expanded` - applies to `.pf-m-expandable` for the expanded state.

Also adds a new `.pf-c-tabs__toggle` element with a button and toggle text in it that comes before the `.pf-c-tabs__list` when tabs are set to expandable.

tlabaj:
**What**: Closes #5520

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -10,7 +10,9 @@ import { TabButton } from './TabButton';
 import { TabContent } from './TabContent';
 import { TabProps } from './Tab';
 import { TabsContextProvider } from './TabsContext';
+import { Button } from '../Button';
 import { getOUIAProps, OUIAProps, getDefaultOUIAId, canUseDOM } from '../../helpers';
+import { GenerateId } from '../../helpers/GenerateId/GenerateId';
 
 export enum TabsComponent {
   div = 'div',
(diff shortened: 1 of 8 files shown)


mcarrano (commented):
This looks good to me. But I will defer to @mcoker on whether this gives us everything we need to support a responsive tab solution.

jessiehuff (commented):
For the most part this looks pretty good, but I'm confused what the scroll left/right button does? I don't see anything happen when trying to use it by mouse, keyboard, or screen reader so I wasn't sure what exactly it's supposed to do.

tlabaj:
@jessie,  I am not sure I am following.

these are the two examples that were added.  neither has a scroll button.   Just the toggle button.

https://patternfly-react-pr-5845.surge.sh/components/tabs#vertical-expandable-uncontrolled
https://patternfly-react-pr-5845.surge.sh/components/tabs#vertical-expandable

jessiehuff:
I first noticed it when using the keyboard. The focus disappeared for one tab stop so I inspected and noticed two hidden buttons. I called them scroll left/right buttons because of the `aria-label`.

5558353B-1478-43E7-AA5E-DF73AE0B4192_1_105_c

mcoker (commented):
Looks great! The only thing is the toggle button doesn't have an `aria-label` - should it since the button's `id` is used with `aria-labelledby`?

jessiehuff:
That's a good point, @mcoker. This looks like a similar pattern as [the expandable section](https://www.patternfly.org/v4/components/expandable-section). That makes me curious about the two different structures. Is there a reason that the span with the text is outside of the button itself? Are they meant to be two separate sections? In the expandable section, it's inside the button. If it were inside the button, it looks like we wouldn't need the `aria-labelledby` at all because the screen reader would view the text as part of the button. Though if these are intended to be separate then we can either remove the button id from the `aria-labelledby` or add an `aria-label` to the toggle button.

mcoker:
@jessiehuff The design called for the toggle to just be the icon, and not include the text. This and the expandable jump links are built the same way, so we should probably be consistent.

jessiehuff:
Do you know if the user will ever expect the toggle text to be clickable @mcarrano? I'm curious how we decide when to keep the toggle button and toggle text separate and when it's together? It seems like we have both patterns with the expandable section and jump links. Also, do we expect use cases where only the toggle exists and there is no toggle text?

mcarrano:
That's a good question @jessiehuff . I don't remember if there is a reason jump links were implemented this way, but I would expect that the user could click on either the text or icon to expand the items below.

mcoker:
https://github.com/patternfly/patternfly/pull/4333/files

@tlabaj this should be good to work on now, it moves the `pf-c-tabs__toggle-text` element into the button, after the `__toggle-icon`

tlabaj:
Opened PR #6342 to address this.

Issue #5867, opened 2021-06-02
Chart legend triangle symbols


dlabrecq:
For Cost Management, we want to use left pointing triangles as legend symbols. Unfortunately, Victory only supports up and down pointing triangles, not left and right. A left pointing triangle would better represent a forecast's "cone of confidence" in our tooltips.

<img width="1063" alt="triangle" src="https://user-images.githubusercontent.com/17481322/120542782-81dc5a00-c3b9-11eb-8b71-f567dde6bbd4.png">

dlabrecq:
See https://github.com/patternfly/patternfly-react/pull/5869

PR #5869, opened 2021-06-02
feat(charts): Add left and right triangle legend symbols

dlabrecq:
For Cost Management, we want to use left pointing triangles as legend symbols. Unfortunately, Victory only supports up and down pointing triangles, not left and right. A left pointing triangle would better represent a forecast's "cone of confidence" in our tooltips.

triangle

https://github.com/patternfly/patternfly-react/issues/5867


redallen (approved):

ddonahue007 (approved):
LGTM

PR #5878, opened 2021-06-04
 Slider: in the scenario that we show boundaries but not ticks don't pollute the DOM with steps

KKoukiou:
Previously Slider had terrible performance when we had a few thousands of steps even when showTicks was unset.


nicolethoen (dismissed):

KKoukiou:
@nicolethoen I noticed there was a mrge conflict to rebased and reposted.

dlabrecq (approved):

nicolethoen (approved):

PR #5884, opened 2021-06-07
fix(DualListSelector): add/remove all should respect filtering

### LINKED ISSUE #5854: Dual list bug for moving filtered items
We got the following message in our Patternfly slack from Eric Gustavsson:

> With dual list I was notified of another UX concern from our application. Where if a user filters for something and sees something in the list, they might click select all button instead. They expect that only the items in the filtered results are moved over, not everything

I personally would agree with this statement. If you filter for "Option 4" and then click the double arrow, I'd expect that only "Option 4" would get moved over to the other side.

See gif below:
patternfly-dual-list-move-all

cc @mcarrano

kmcfaul:
**What**: Closes #5854

Add/remove all should now adhere to the current filter (if applicable) and behave the same without any filter. This applies to both normal and tree dual list selectors.


nicolethoen (commented):
super well done. 

The tree view variant is still not quite applying the filtering correctly. 
If I filter using `5` as the search string, and I add all, `Option 6` gets added to the selected pane though it was filtered out of the available pane. 
The same thing happens if I'm filtering on the string `5` and I select `Folder 2`. Then if I add selected items, then `Option 6` still gets added, even though it was filtered out.

kmcfaul:
I'm not seeing the first behavior. When I filter by `5` and add `all` I only see Option 5, Folder 2 > Option 5. 

The second behavior is because of the way tree table selection works. When you select a folder it checks everything under it. I'll look into having that respect the filtering too.

kmcfaul:
The move selected action is bugged and I'm looking at it. But I'm not seeing extra items being moved with the tree view + `5` in the filter + move all action.

kmcfaul:
@nicolethoen Updated the way the check handler looks for sub-items. Now it won't "check" items that are under a folder but not visible due to filtering.

nicolethoen (approved):

dlabrecq (approved):

redallen (approved):
I couldn't find a way to break it!

dlabrecq (commented):
LGTM

PR #5886, opened 2021-06-07
fix(Table): empty state cell should not have data-label

### LINKED ISSUE #5379: Table/Empty state: the Empty State table from the react demos is breaking on mobile
The examples from the docs page looks like this in mobile:

Screen Shot 2021-01-27 at 13 42 54

kmcfaul:
**What**: Closes #5379

Removes the header cell data label if a BodyCell child is either an `EmptyState` component or a `Bullseye` component with an `EmptyState` direct child. Visually only changes the mobile view (where column headers are now unseen).


KKoukiou (approved):
Works as expected! Thanks for fixing this :)

tlabaj (approved):

nicolethoen (approved):

PR #5903, opened 2021-06-09
feat(HelperText): add helper text component

### LINKED ISSUE #5818: Helper text component
Introduce new Helper text component per this core issue: https://github.com/patternfly/patternfly/issues/4017 This will allow a variety of help text configurations to be attached to a Text input or other form component. This is useful for dynamic validation of entered text for password validation of similar use cases. Consumer should be able to manipulate the list per validation of the input text  per validation rules external to the component.

kmcfaul:
**What**: Closes #5818

Users should be able to update individual HelperTextItem properties (like `variant`) to reflect a validation update. Perhaps a more detailed demo showcasing a validated input field with a HelperText list would be a good follow up?

### DIFF
--- a/packages/react-core/src/components/HelperText/HelperText.tsx
+++ b/packages/react-core/src/components/HelperText/HelperText.tsx
@@ -1,0 +1,27 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/HelperText/helper-text';
+import { css } from '@patternfly/react-styles';
+
+export interface HelperTextProps extends React.HTMLProps<HTMLDivElement | HTMLUListElement> {
+  /** Content rendered inside the helper text container. */
+  children?: React.ReactNode;
+  /** Additional classes applied to the helper text container. */
+  className?: string;
+  /** Component type of the helper text container */
+  component?: 'div' | 'ul';
+}
+
+export const HelperText: React.FunctionComponent<HelperTextProps> = ({
+  children,
+  className,
+  component = 'div',
+  ...props
+}: HelperTextProps) => {
+  const Component = component as any;
+  return (
+    <Component className={css(styles.helperText, className)} {...props}>
+      {children}
+    </Component>
+  );
+};
+HelperText.displayName = 'HelperText';
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -28,7 +28,7 @@ const variantStyle = {
   indeterminate: styles.modifiers.indeterminate,
   warning: styles.modifiers.warning,
   success: styles.modifiers.success,
-  error: styles.modifiers.invalid
+  error: styles.modifiers.error
 };
 
 export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -15,7 +15,7 @@ export interface HelperTextItemProps extends React.HTMLProps<HTMLDivElement | HT
   component?: 'div' | 'li';
   /** Variant styling of the helper text item. */
   variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'invalid';
-  /** Icon prefixing the helper text. */
+  /** Icon prefixing the helper text. This property will override the default icon paired with a dynamic helper text. */
   icon?: React.ReactNode;
   /** Flag indicating the helper text item is dynamic. */
   isDynamic?: boolean;
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -14,7 +14,7 @@ export interface HelperTextItemProps extends React.HTMLProps<HTMLDivElement | HT
   /** Sets the component type of the helper text item. */
   component?: 'div' | 'li';
   /** Variant styling of the helper text item. */
-  variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'invalid';
+  variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'error';
   /** Icon prefixing the helper text. This property will override the default icon paired with a dynamic helper text. */
   icon?: React.ReactNode;
   /** Flag indicating the helper text item is dynamic. */
@@ -28,7 +28,7 @@ const variantStyle = {
   indeterminate: styles.modifiers.indeterminate,
   warning: styles.modifiers.warning,
   success: styles.modifiers.success,
-  invalid: styles.modifiers.invalid
+  error: styles.modifiers.invalid
 };
 
 export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
@@ -57,7 +57,7 @@ export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
           {(variant === 'default' || variant === 'indeterminate') && <MinusIcon />}
           {variant === 'warning' && <ExclamationTriangleIcon />}
           {variant === 'success' && <CheckIcon />}
-          {variant === 'invalid' && <TimesIcon />}
+          {variant === 'error' && <TimesIcon />}
         </span>
       )}
       <span className={css(styles.helperTextItemText)}>{children}</span>
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -15,7 +15,7 @@ export interface HelperTextItemProps extends React.HTMLProps<HTMLDivElement | HT
   component?: 'div' | 'li';
   /** Variant styling of the helper text item. */
   variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'error';
-  /** Icon prefixing the helper text. This property will override the default icon paired with a helper text. */
+  /** Custom icon prefixing the helper text. This property will override the default icon paired with each helper text variant. */
   icon?: React.ReactNode;
   /** Flag indicating the helper text item is dynamic. */
   isDynamic?: boolean;
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -15,11 +15,11 @@ export interface HelperTextItemProps extends React.HTMLProps<HTMLDivElement | HT
   component?: 'div' | 'li';
   /** Variant styling of the helper text item. */
   variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'error';
-  /** Icon prefixing the helper text. This property will override the default icon paired with a dynamic helper text. */
+  /** Icon prefixing the helper text. This property will override the default icon paired with a helper text. */
   icon?: React.ReactNode;
   /** Flag indicating the helper text item is dynamic. */
   isDynamic?: boolean;
-  /** Flag indicating the dynamic helper text should have an icon */
+  /** Flag indicating the helper text should have an icon. Dynamic helper texts include icons by default while static helper texts do not. */
   hasIcon?: boolean;
 }
 
@@ -37,8 +37,8 @@ export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
   component = 'div',
   variant = 'default',
   icon,
-  isDynamic,
-  hasIcon = true,
+  isDynamic = false,
+  hasIcon = isDynamic,
   ...props
 }: HelperTextItemProps) => {
   const Component = component as any;
@@ -52,7 +52,7 @@ export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
           {icon}
         </span>
       )}
-      {isDynamic && hasIcon && !icon && (
+      {hasIcon && !icon && (
         <span className={css(styles.helperTextItemIcon)} aria-hidden>
           {(variant === 'default' || variant === 'indeterminate') && <MinusIcon />}
           {variant === 'warning' && <ExclamationTriangleIcon />}
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -1,0 +1,52 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/HelperText/helper-text';
+import { css } from '@patternfly/react-styles';
+
+export interface HelperTextItemProps extends React.HTMLProps<HTMLDivElement | HTMLLIElement> {
+  /** Content rendered inside the helper text item. */
+  children?: React.ReactNode;
+  /** Additional classes applied to the helper text item. */
+  className?: string;
+  /** Sets the component type of the helper text item. */
+  component?: 'div' | 'li';
+  /** Variant styling of the helper text item. */
+  variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'invalid';
+  /** Icon prefixing the helper text. */
+  icon?: React.ReactNode;
+  /** Flag indicating the helper text item is dynamic. */
+  isDynamic?: boolean;
+}
+
+const variantStyle = {
+  default: '',
+  indeterminate: styles.modifiers.indeterminate,
+  warning: styles.modifiers.warning,
+  success: styles.modifiers.success,
+  invalid: styles.modifiers.invalid
+};
+
+export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
+  children,
+  className,
+  component = 'div',
+  variant = 'default',
+  icon,
+  isDynamic,
+  ...props
+}: HelperTextItemProps) => {
+  const Component = component as any;
+  return (
+    <Component
+      className={css(styles.helperTextItem, variantStyle[variant], isDynamic && styles.modifiers.dynamic, className)}
+      {...props}
+    >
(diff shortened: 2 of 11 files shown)


packages/react-core/src/components/HelperText/HelperTextItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export interface HelperTextItemProps extends React.HTMLProps<HTMLDivElement | HTMLLIElement> {
+  /** Content rendered inside the helper text item. */
+  children?: React.ReactNode;
+  /** Additional classes applied to the helper text item. */
+  className?: string;
+  /** Sets the component type of the helper text item. */
+  component?: 'div' | 'li';
+  /** Variant styling of the helper text item. */
+  variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'invalid';
```
KKoukiou, line 13: maybe call the invalid state error, to be same as the:

validated?: 'success' | 'warning' | 'error' | 'default';

From FormGroup?

Not vetoing here, just thinking out loud, feels like we have at least three words which generated error like components:
- error
- danger
- now invalid :)

kmcfaul: Yeah it may be more consistent to use `error`. The actual modifier is called `invalid` though. @mcarrano Any thoughts on the naming here?

mcarrano (commented):
@kmcfaul I'm also torn between naming of the failure state `error` or `invalid`. It becomes a matter of context. @mcoker what do you think?

Otherwise this looks good. But I agree that we should follow up with a demo to show how to use this component. Not sure it would be obvious by just looking at the static examples.

mcoker:
> I'm also torn between naming of the failure state error or invalid. It becomes a matter of context.

@mcarrano yeah this is a bit confusing. Here's the breakdown of our usage of invalid, error, and danger.

#### Invalid
* Form control elements (text input, textarea, form select) 
* Select component

#### Error
* Form component helper text (existing helper text this will likely replace)
* Date picker helper text

#### Danger
* Progress
* Alert
* Button
* Notification drawer item (similar to alert)

This naming seems appropriate to me - form elements are invalid, helper text is some kind of error message (usually associated with an invalid form element, but not limited to that use), and the other components represent a dangerous threshold/action.

Though to @KKoukiou's point, I think we could also update the usage of `pf-m-error` on the existing helper text elements to be `pf-m-invalid` with the expectation they'll pair with a `pf-m-invalid/aria-invalid` form element, if we wanted to reduce the number of variation names. That was my intent with naming this `pf-m-invalid` in core. And in a breaking change release, we'll replace the existing use of `pf-m-error` in the form/date picker helper text with this component.

mcarrano:
So if we are using `error` for the existing helper text with a form component, doesn't it make sense to use `error` rather than `invalid` here since it's also helper text?

mcoker:
> So if we are using error for the existing helper text with a form component, doesn't it make sense to use error rather than invalid here since it's also helper text?

@mcarrano yeah I agree, updated my comment.

packages/react-integration/demo-app-ts/src/components/demos/HelperText/HelperTextDemo.tsx
```diff
@@ -0,0 +1,43 @@
...
+          <HelperTextItem id="success" isDynamic variant="success" icon={<CheckCircleIcon />}>
+            This is success helper text
+          </HelperTextItem>
+          <HelperTextItem id="invalid" isDynamic variant="invalid" icon={<ExclamationCircleIcon />}>
+            This is invalid helper text
+          </HelperTextItem>
+          <HelperTextItem id="indeterminate" isDynamic variant="indeterminate" icon={<QuestionIcon />}>
+            This is indeterminate helper text
+          </HelperTextItem>
+        </HelperText>
```
mcoker, line 39: Should we add an example and test for `isDynamic` and usage of `component`?

mcoker (commented):
Looks great! I left a few questions and have PR up to change `pf-m-invalid` to `pf-m-error` here - https://github.com/patternfly/patternfly/pull/4140

packages/react-core/src/components/HelperText/HelperText.tsx
```diff
@@ -0,0 +1,27 @@
...
+import styles from '@patternfly/react-styles/css/components/HelperText/helper-text';
+import { css } from '@patternfly/react-styles';
+
+export interface HelperTextProps extends React.HTMLProps<HTMLDivElement | HTMLUListElement> {
+  /** Content rendered inside the helper text container. */
+  children?: React.ReactNode;
+  /** Additional classes applied to the helper text container. */
+  className?: string;
+  /** Component type of the helper text container */
+  component?: 'div' | 'ul';
```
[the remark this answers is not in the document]
kmcfaul: The core examples use `div` except in the case of a dynamic list of helper text items, which uses a `ul`/`li` combo. Let me know if I should add anything else like `ol`.

packages/react-integration/demo-app-ts/src/components/demos/HelperText/HelperTextDemo.tsx
kmcfaul: There is an item with `isDynamic` already, unless you mean something else? I'm a little confused on what that flag is used for in the react side. 

I'll add a `component` example to the demo.

mcoker: Oh sorry I was looking at the cypress tests and saw we aren't looking for the dynamic class or list elements using `component`. 

> I'm a little confused on what that flag is used for in the react side

What that's supposed to do is allow the helper text item to switch states (warning, error, etc) as the field its associated with is being validated - which will update the icon, icon color, and potentially the font weight if the state changes between error and not-error. We have an issue to build a [login password strength demo](https://github.com/patternfly/patternfly/issues/4018) that would show how that works - here's the design for that, which shows how those elements would update on the fly as a field is being validated - https://marvelapp.com/prototype/6jehce7/screen/78511266

packages/react-core/src/components/HelperText/HelperText.tsx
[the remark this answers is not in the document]
mcoker: If we don't open it to just be whatever the user wants to pass in, this works just fine.

packages/react-integration/demo-app-ts/src/components/demos/HelperText/HelperTextDemo.tsx
kmcfaul: Hmm I think I get it. I'm just a bit confused because the helper text component itself is stateless, its styling is determined by its props. The app would change an app-side state to update those props in a demo like that. In which case, does `isDynamic` need to be an exposed flag, or would it be something that always exists on a helper text item. Is there a downside to applying that modifier all the time?

mcoker: > Is there a downside to applying that modifier all the time?

Yeah, the helper text has 2 basic types so far. First is the solid color/regular state. In that state, it's just helper text that you would use associated with a field of some sort, and it may or may not change depending on the state of the field. For example, it could be normal/black text helper text that's used to describe a field. That text may or may not be replaced with error/red text if the field is invalid.

The other state is `isDynamic` which is an indication that it's a live type of helper text that will update as you update the field. It has grey text (or black when error/invalid) and a state colored icon and the icon and text changes as the field is validated.

I could see allowing the user to control the states/text that's displayed, but I could also see allowing the user to pass an object to the helper text item with the default, warning, error, and success state text to display, and as the helper text item `variant` is updated, we display the correct text or something like that?

cc @mcarrano is this inline with how you see this used?

mcarrano: Yes, I think so.

kmcfaul: Ah I understand, thanks for explaining :)

packages/react-core/src/components/HelperText/HelperTextItem.tsx
mcoker: @kmcfaul renamed `pf-m-invalid` to `pf-m-error`, you'll get it in core v4.110.4 - https://github.com/patternfly/patternfly/pull/4140

mcoker:
@mcarrano looking at this again, it's up to the user to specify the icon with the helper text, both with the regular variation as well as the dynamic variation. I think that seems reasonable with regular helper text, but would we want to include the icon to use in the dynamic text (assuming it has an icon by default, and the states should all use the same icon) - and allow the user to either override the image with a prop and potentially disable the image with another prop?

nicolethoen:
I like that idea @mcoker, I think that's a pattern we've used before - like in the notification badge

kmcfaul:
@mcoker @nicolethoen Icons are now built into the component with the `isDynamic` flag. Passing `icon` will override the icon for a dynamic helper text, and icons may be turned off via a new `hasIcon` property.

mcarrano (approved):
@kmcfaul @mcoker @nicolethoen This all looks good to me.

kmcfaul:
@mcoker @mcarrano What was the final decision re: the variant naming between `error` and `invalid`? If the state will eventually be named `invalid` should I leave it as is or should I update to `error`?

mcoker:
@kmcfaul it was renamed to `pf-m-error` - https://github.com/patternfly/patternfly/pull/4140

mcoker (dismissed):
Looks great!! 🥇

packages/react-integration/demo-app-ts/src/components/demos/HelperText/HelperTextDemo.tsx
```diff
@@ -0,0 +1,54 @@
+import { HelperText, HelperTextItem } from '@patternfly/react-core';
```
tlabaj, line 1: can you als test out the `hasIcon` and `icon/customIcon` interfaces

tlabaj (changes requested):

packages/react-core/src/components/HelperText/HelperTextItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  /** Content rendered inside the helper text item. */
+  children?: React.ReactNode;
+  /** Additional classes applied to the helper text item. */
+  className?: string;
+  /** Sets the component type of the helper text item. */
+  component?: 'div' | 'li';
+  /** Variant styling of the helper text item. */
+  variant?: 'default' | 'indeterminate' | 'warning' | 'success' | 'error';
+  /** Icon prefixing the helper text. This property will override the default icon paired with a dynamic helper text. */
+  icon?: React.ReactNode;
```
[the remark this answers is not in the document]
kmcfaul: I think `icon` makes sense here, not sure it needs the extra qualifier since both helper text cases (generic/dynamic) will set an icon if the property is passed. What do you think?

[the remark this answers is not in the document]
tlabaj: the `Alert` component has a `customIcon` prop.  This seems like a similar use case.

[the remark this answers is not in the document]
tlabaj: I just did a search and notification drawer just has "icon". so we are not consistent across components.

mcoker (dismissed):

nicolethoen (commented):
What does it mean that it's dynamic? I don't see any style changes for the dynamic helper text.

If it means that it's something that is dynamically generated, we probably need to incorporate something with aria-live in this example, so that the screen readers will be able to anticipate the possibly incoming text. That's something @jessiehuff would be able to evaluate very well. If we want to merge this now, we should open an issue to verify this with Jessie.

kmcfaul:
The component isn't dynamically generated with the flag (just pairs a specific default icon with the text based on the variant). The styling itself updates the text color/icon color with a variable instead of being static I believe. 

A more complex demo would certainly include an aria-live region still, but the basic demos don't change.

nicolethoen:
Okay - I'll put a mental flag in this example as something for Jessie to make sure she understands and approves. I know the component will be in beta for a bit, so we just need to make sure use cases and 'rules' for implementing dynamic hints are well defined in the long run.

nicolethoen (approved):

dlabrecq (approved):

packages/react-core/src/components/HelperText/HelperTextItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  hasIcon = true,
+  ...props
+}: HelperTextItemProps) => {
+  const Component = component as any;
+  return (
+    <Component
+      className={css(styles.helperTextItem, variantStyle[variant], isDynamic && styles.modifiers.dynamic, className)}
+      {...props}
+    >
+      {icon && (
```
tlabaj, line 50: So the `icon` vs `hasIcon` interface still does not make sense to me.  I don't understand why the static helper text can have an icon if it is custom but not dynamic?  It is confusing that the consumer needs to provide an icon when the helper text is static.

@mcoker my understanding is that dynamic requires an icon but for that static it is optional.  is t hat correct?

kmcfaul: `icon` is an optional property that inserts an icon prefixing a helper text (affects both static and dynamic helper text, its just dynamic has default icons that this property would override). 
`hasIcon` is an optional property that controls whether the dynamic default icon is present (in cases where the included dynamic icon isn't desired).

Would `hasDynamicIcon` make more sense instead, or should static helper text also include the default icons?

mcoker: > my understanding is that dynamic requires an icon but for that static it is optional. is t hat correct?

@tlabaj correct, since the text color is the same between multiple states in the dynamic variation, the icon is used to convey the state (success, warning, etc).

mcoker: > Would `hasDynamicIcon` make more sense instead

I think so, assuming we want to keep the icon defaults as they are - regular has an optional icon that the user specifies, and dynamic comes with a specific icon you can either change or disable.

> should static helper text also include the default icons?

@mcarrano wdyt about this? I think this could be good in enforcing some consistency with icons used for the various states of helper text, though I don't know if the default use case for helper text should include an icon or not. I do know regular helper text (black text) wouldn't have an icon.

Another option could be to have `hasIcon` as a boolean and it's false for regular helper text and true for dynamic. Setting it to true with regular helper text would enable the default icons, and the defaults could be overridden by an `icon` prop (passing an `icon` prop would also just add an icon without having to set `hasIcon`). There would still be the use case of regular help text - what icon would `hasIcon` use? As far as I've seen, we haven't added an icon to regular helper text.

kmcfaul: The only thing I dislike about `hasDynamicIcon` to me, is that it implies the icon changes in some way. There might be a better name for the variable. Though if we end up going the route of including default icons on the static helper text we can keep `hasIcon`.

Another alternative is to get rid of `hasIcon` altogether, and have the component remove dynamic helper text icons if it is explicitly passed `icon={null}`?

mcarrano: >  should static helper text also include the default icons?

I'm looking at the current examples we have for Forms and the default help text does not have an icon (https://www.patternfly.org/v4/components/form/html#help-text). I think we should keep it that way and let the user add an icon only when needed. Is that you question, @mcoker ? Or are you just asking whether the default status incons should be available in both versions? If the later, I agree that might lend to consistency. But the default for static helper text should still be no icon.

Does that make sense or did I confuse things more? :)

mcoker: > I think we should keep it that way and let the user add an icon only when needed. Is that you question, @mcoker

@mcarrano yep!

kmcfaul: Should I update `hasIcon` to work for static helper text then, leaving its default to `false`? And would the default icons be the same for each variant as the dynamic helper text?

mcarrano: @kmcfaul Yes. Please make `hasIcon` valid for both versions but defaulting to false for the static variant. Regarding what default set of icons should be, I agree that they should be consistent. @mcoker now it looks like there are different icons for static and dynamic variants in the examples. Am I seeing that right? Do you know why these are different? Is it just that the designers gave you different icons?

packages/react-core/src/components/HelperText/HelperTextItem.tsx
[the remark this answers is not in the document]
tlabaj: I think we should at least add the word `custom`  to the description.

mcarrano (approved):
I think this looks good for now. We can clear up the icons and perhaps simplify the examples somewhat in a future release but I think we have the structure right.

tlabaj (approved):

packages/react-core/src/components/HelperText/HelperTextItem.tsx
kmcfaul: Followed up in slack: icons across both static/dynamic will match dynamic default icons for the moment. Might be updated later.

redallen (approved):
Maybe `FormHelperText` could be refactored to use this new component in the future? They seem quite similar.

jessiehuff (commented):
I completely agree with @nicolethoen here about the dynamic examples. I can understand if we want to just add the accessibility to a demo and leave the component customization to the consumer, but I think we at least need some kind of documentation or explanation of how to actually implement it accessibly. (And if we're creating a demo, at bare minimum pointing to the demo as how to implement it so consumers don't just look at these examples and add it without any a11y interaction.)

PR #5917, opened 2021-06-15
feat(Text): Added support for visited link styling

### LINKED ISSUE #5857: Visted style for links
re: https://github.com/patternfly/patternfly/pull/4067

In the content component, you can apply `.pf-m-visited` to either a `.pf-c-content` block (to style all links in the block), or an `<a>` element in a `.pf-c-content` (to style an individual link), and visited styling will be applied when a link has been visited.

tlabaj:
**What**: Closes #5857

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Text/Text.tsx
+++ b/packages/react-core/src/components/Text/Text.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
+import styles from '@patternfly/react-styles/css/components/Content/content';
 import { useOUIAProps, OUIAProps } from '../../helpers';
 
 export enum TextVariants {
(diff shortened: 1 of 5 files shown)


nicolethoen (commented):
I could be wrong, but I imagined that the flag would mean 
'if this link has been visited, it will be styled with the visited styles' 
rather than 'style this link as though it has been visited' 
is that true?

mcoker (changes requested):
This looks great to me when defining a link, but we also support `.pf-m-visited` on `.pf-c-content`. The use case is when you have a lot of links in a content block that you want to have visited styling. Can you add support for this on `<TextContent>`, too?

mcoker:
@nicolethoen yep that's true. If the prop is set and the browser has determined that you've visited the link, you get the visited styling (purple text)

nicolethoen:
> @nicolethoen yep that's true. If the prop is set and the browser has determined that you've visited the link, you get the visited styling (purple text)

Sweet - I think the distinction should be made clear in the comment/docs

mcoker (approved):
Nice!!

nicolethoen (approved):

kmcfaul (approved):

dlabrecq (commented):
LGTM

PR #5921, opened 2021-06-15
feat(Table): add draggable table rows demo

### LINKED ISSUE #5817: Table: add draggable rows
Add draggable row capability to Table as a follow up to: https://github.com/patternfly/patternfly/issues/4065  This feature may be added to the composable table if that seems like the best approach.

cc @jessiehuff accessibility considerations here?

kmcfaul:
**What**: Closes #5817

Basic dragging functionality. Keyboard interaction may have to be added in a follow-up.

@nicolethoen 

Surge direct link: https://patternfly-react-add_draggable_table.surge.sh/components/table/#composable-draggable-table


kmcfaul:
@nicolethoen The onClick handler for the draggable cell isn't needed for the draggable events at this time. The problem with putting the event handlers on the drag button instead of the whole row is that the 'ghost' you see when dragging would be incorrect (seeing only the button) and the re-ordering gets a bit wonky. 

I think we could update this later, maybe after revisiting the draggable implementation overall. But right now the whole row is getting dragged like in DataList. I don't think that should interfere with any control buttons in the row.

nicolethoen:
I know in my code i deleted all the other examples - I don't think you should do that ;)

kmcfaul:
Oh that would be bad haha. I didn't notice that. Will update the PR

kmcfaul:
Updated, hopefully. Rolled back to the commit before your initial commit from the file history then re-applied the draggable table example.

nicolethoen:
surge site?: patternfly-react-add_draggable_table.surge.sh

kmcfaul:
> surge site?: [patternfly-react-add_draggable_table.surge.sh](patternfly-react-add_draggable_table.surge.sh)

Yeah that does work. I'm thinking it's maybe because the branch is on the PF repo already but I'm unsure. Regardless, I'll direct link to the example at the top of the PR.

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -3713,20 +3731,228 @@ class TreeTable extends React.Component {
...
+            </TreeRowWrapper>
+          ))}
+        </Tbody>
+      </TableComposable>
+    );
+  }
+}
+```
+
+### Composable: Draggable table
```
tlabaj, line 3763: Maybe add "row" to example name.

packages/react-table/src/components/Table/DraggableCell.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export interface DraggableCellProps {
+  id: string;
+  className?: string;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+  'aria-label'?: string;
+}
+
+export const DraggableCell: React.FunctionComponent<DraggableCellProps> = ({
+  className = '',
```
tlabaj, line 13: this can default to undefined.

packages/react-table/src/components/Table/DraggableCell.tsx
```diff
@@ -0,0 +1,32 @@
...
+  'aria-label': ariaLabel,
+  id,
+  ...props
+}: DraggableCellProps) => (
+  <Button
+    id={id}
+    variant="plain"
+    className={className}
+    type="button"
+    aria-label={ariaLabel || `Draggable row draggable button`}
```
tlabaj, line 24: @jessiehuff  does this make sense for default aria-label?

nicolethoen: Jessie is out - can you think of an alternative?

kmcfaul: For what its worth, DataList has no default aria-label, but its example code uses "Reorder" as it's `aria-label`, paired with `aria-labelledby` with the list item id.

mcoker (changes requested):
This looks great! Though the parent `<td>` for the draggable button needs the class `.pf-c-table__draggable`. That will fix the button's spacing and cursor.

And core is missing a background color on the ghost row, so I put up a PR for that - https://github.com/patternfly/patternfly/pull/4159

packages/react-table/src/components/Table/DraggableCell.tsx
```diff
@@ -19,7 +21,7 @@ export const DraggableCell: React.FunctionComponent<DraggableCellProps> = ({
   <Button
     id={id}
     variant="plain"
-    className={className}
+    className={css(styles.tableDraggable, className)}
```
mcoker, line 24: oh sorry, this needs to go on the parent `<td>` for this button.

kmcfaul: Ah I missed that detail. Updated!

mcarrano (commented):
This looks good @kmcfaul . I wasn't completely clear about the discussion around whether to allow the whole row of just the drag handle to be draggable. I like how this works now for this example, but could see where we might get into trouble if there were other active elements in the row. Did you say this is or isn't consistent with how data list works? I guess that the issues would be the same.

kmcfaul:
It is the same as DataList

mcarrano (dismissed):
I'm good then. Thanks @kmcfaul .

packages/react-table/src/components/Table/DraggableCell.tsx
tlabaj: I think it is ok to leave as is and get Jessie's feedback once she returns.

tlabaj (dismissed):

nicolethoen (dismissed):

dlabrecq (approved):

kmcfaul:
Waiting on an additional core css class for drag-over styling

kmcfaul:
@mcoker Updated with the new class. The `pf-m-drag-over` modifier is being added/removed along with the ghost row modififer.

mcoker (approved):
L🔥TM!!! Fantastic work on getting this in so quickly!! 🏆🥇

tlabaj (approved):

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -3713,20 +3731,228 @@ class TreeTable extends React.Component {
...
+      draggingToItemIndex: null,
+      dragging: false,
+      itemOrder: ['row1', 'row2', 'row3'],
+      tempItemOrder: []
+    };
+    this.bodyRef = React.createRef();
+
+    this.getIndex = id => Array.from(this.bodyRef.current.children).findIndex(item => item.id === id);
+
+    this.onDragStart = evt => {
```
redallen, line 3783: Pushing this to the consumer is good for now.

redallen (approved):
Way to go Katie!

PR #5923, opened 2021-06-15
fix(Menu): update tabIndex when children change

### LINKED ISSUE #5915: Menu filtering with text input, user cannot get to menu via keyboard
This relates to [menu items with text input](https://www.patternfly.org/v4/components/menu#filtering-with-text-input).  When a user filters out options, the menu is no longer available via the keyboard.

Method 1
 - Navigate to the filter box.  
 - Enter text that will filter out the first option, but leave others.  In the example linked above, enter `2` to filter out `Action 1` and `Action 3`.
 - Click on 'tab'.  This should move the user to the menu, but the menu is completely skipped and the user has no way to get to the menu via the keyboard.
 
 Method 2
- Navigate to the filter box
- Enter text that will filter out all options.  In the example linked above, enter `no more options` to filter out all three options.
- Hit the delete key until the filter box is empty and all options are visible.
- Click on 'tab'.  This should move the user to the menu, but the menu is completely skipped  and the user has no way to get to the menu via the keyboard.
 

Bug - blocking.  There doesn't appear to be a way for the user to get to the menu via the keyboard.

Openshift console.  Targeting Openshift console 4.9

kmcfaul:
**What**: Closes #5915

Put the fix in the component-side as `tabIndex` was already being mapped internally, but let me know @tlabaj @mcarrano if it would make more sense for the fix to live in the example code instead because filtering is done on the app-side.

This updates the behavior of Menu such that whenever Menu's children change, Menu will remap the `tabIndex` of all options so only the first option is tabbable.

### DIFF
--- a/packages/react-core/src/components/Menu/Menu.tsx
+++ b/packages/react-core/src/components/Menu/Menu.tsx
@@ -81,14 +81,12 @@ class MenuBase extends React.Component<MenuProps, MenuState> {
       window.addEventListener('transitionend', this.props.isRootMenu ? this.handleDrilldownTransition : null);
     }
 
-    let ref = this.menuRef;
-    if (this.props.innerRef) {
-      ref = this.props.innerRef as React.RefObject<HTMLDivElement>;
-    }
+    this.setFirstTabIndex();
+  }
 
-    const firstItem = ref.current.querySelector('button, a');
-    if (firstItem) {
-      (firstItem as HTMLElement).tabIndex = 0;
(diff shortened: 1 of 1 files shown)


mcarrano (approved):
Looks good to me @kmcfaul . I'm not sure about the right answer to your question, however. I'll defer to @tlabaj or you can solicit an opinion from the OpenShift devs who are using this.

tlabaj (approved):
Looks great!

nicolethoen (approved):

Issue #5802, opened 2021-05-19
LogViewer: Highlight just search string matches - not whole line


nicolethoen:
Follow up work for https://github.com/patternfly/patternfly/issues/4081

Issue #5774, opened 2021-05-12
LogViewer: enable responsive width sizing


kmcfaul:
LogViewer should resize its width based on screen size changes.

PR #5927, opened 2021-06-17
feat(LogViewer): Make logviewer responsive and change search result styling

nicolethoen:
Closes: #5774 Closes: #5802 

@mcoker can we possibly update the core styles for the pf-m-current and pf-m-match modifiers in log viewer? Right now they are expected to be placed on the `pf-c-log-viewer__list-item` component, but with this update, they are placed on spans inside the pf-c-log-viewer__text.


packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -56,18 +59,53 @@ export const LogViewer: React.FunctionComponent<LogViewerProps> = memo(
...
+      if (resizeTimer) {
+        clearTimeout(resizeTimer);
+      }
+      resizeTimer = setTimeout(() => {
+        setLoading(false);
+      }, 500);
+    };
+
+    useEffect(() => {
+      if (loading) {
```
kmcfaul, line 94: `!loading` seems to work a bit better testing locally, to make the width re-update after the load finishes instead of lagging behind the resize event (I noticed sometimes the width would update to a size captured before the resize event ends or between resize events if you move the mouse slowly, and not update after letting go of the mouse).

kmcfaul:
Updating `getDerivedStateFromProps` in `createListComponent.ts` to the following signature will get rid of the console error. Not sure if that makes anything behave better but clears the console at least.

```
static getDerivedStateFromProps<T>(nextProps: Props<T>, prevState: State): State | null
```

mcoker:
@nicolethoen we will either need to create new elements for the search/match text (`.pf-c-log-viewer__text-match` and `.pf-c-log-viewer__text-current`) or we can create a more generic element we use for string highlighting (`.pf-m-log-viewer__string`) and add modifiers to that (`.pf-m-match` and `.pf-m-current`). WDYT?

nicolethoen:
> @nicolethoen we will either need to create new elements for the search/match text (`.pf-c-log-viewer__text-match` and `.pf-c-log-viewer__text-current`) or we can create a more generic element we use for string highlighting (`.pf-m-log-viewer__string`) and add modifiers to that (`.pf-m-match` and `.pf-m-current`). WDYT?

In case there is a future where we might do something like wrap date/time strings in the log viewer with specific styles, maybe it makes sense that it'd be the more generic `pf-c-log-viewer__string` and then we can add secondary future imaginary modifiers in that case as well.

mcoker (approved):
LGTM!

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -18,14 +18,16 @@ interface LogViewerProps {
   placeholder?: string;
   /** Flag to enable or disable line numbers on the log viewer. */
   hasLineNumbers?: boolean;
-  /** Width in pixels of the log viewer */
-  width: number;
-  /** Height in pixels of the log viewer */
+  /** width in pixels of the log viewer. */
```
tlabaj, line 21: should be sentence case

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -18,14 +18,16 @@ interface LogViewerProps {
...
+  width?: number;
+  /** Height in pixels of the log viewer. */
   height?: number;
   /** Rows being rendered outside of view. The more rows are rendered, the higher impact on performance */
   overScanCount?: number;
   /** Toolbar rendered in the log viewer header */
   toolbar?: React.ReactNode;
+  /** content displayed while the log viewer is loading */
```
tlabaj, line 29: sentence case

tlabaj (dismissed):
Two small sentence case issues.  We can update later.

kmcfaul (approved):

redallen (approved):
Nicely done!

PR #5946, opened 2021-06-23
feat(AlertGroup): Make dynamic alerts more accessible

### LINKED ISSUE #5731: POC: Make dynamic Alerts more accessible

jessiehuff:
**What**: Alternative approach/attempt to PR#5773
Closes #5731

This PR/POC is meant to test an alternative approach to making dynamic alerts more accessible. It turns AlertGroup into a pre-existing live region for screen readers to be notified that alerts will be rendered inside of it.

### DIFF
--- a/packages/react-core/src/components/Alert/examples/Alert.md
+++ b/packages/react-core/src/components/Alert/examples/Alert.md
@@ -194,7 +194,7 @@ class StaticLiveRegionAlert extends React.Component {
 ### Dynamic live region alert
 ```js
 import React from 'react';
-import { Alert, InputGroup } from '@patternfly/react-core';
+import { Alert, AlertGroup, InputGroup } from '@patternfly/react-core';
 
 class DynamicLiveRegionAlert extends React.Component {
   constructor() {
@@ -213,7 +213,6 @@ class DynamicLiveRegionAlert extends React.Component {
       addAlert({
         title: 'Single Success Alert',
         variant: 'success',
-        isLiveRegion: true,
         key: getUniqueId()
       });
     };
@@ -221,9 +220,6 @@ class DynamicLiveRegionAlert extends React.Component {
       addAlert({
         title: 'Single Info Alert',
         variant: 'info',
-        ariaLive: 'polite',
-        ariaRelevant: 'additions text',
-        ariaAtomic: 'false',
         key: getUniqueId()
       });
     };
@@ -231,9 +227,6 @@ class DynamicLiveRegionAlert extends React.Component {
       addAlert({
         title: 'Single Danger Alert',
         variant: 'danger',
-        ariaLive: 'assertive',
-        ariaRelevant: 'additions text',
-        ariaAtomic: 'false',
         key: getUniqueId()
       });
     };
@@ -251,17 +244,15 @@ class DynamicLiveRegionAlert extends React.Component {
             Add Single Danger Alert
           </button>
         </InputGroup>
+        <AlertGroup isLiveRegion aria-live="polite" aria-relevant="additions text" aria-atomic="false">
         {this.state.alerts.map(({ title, variant, isLiveRegion, ariaLive, ariaRelevant, ariaAtomic, key }) => (
           <Alert
             variant={variant}
             title={title}
-            isLiveRegion={isLiveRegion}
-            aria-live={ariaLive}
-            aria-relevant={ariaRelevant}
-            aria-atomic={ariaAtomic}
             key={key}
           />
         ))}
+        </AlertGroup>
       </React.Fragment>
     );
   }
@@ -271,7 +262,7 @@ class DynamicLiveRegionAlert extends React.Component {
 ### Async live region alert
 ```js
 import React from 'react';
-import { Alert, InputGroup } from '@patternfly/react-core';
+import { Alert, AlertGroup, InputGroup } from '@patternfly/react-core';
 
 class AsyncLiveRegionAlert extends React.Component {
   constructor() {
@@ -297,7 +288,6 @@ class AsyncLiveRegionAlert extends React.Component {
         addAlert({
           title: `This is a async alert number ${this.state.alerts.length + 1}`,
           variant: 'info',
-          isLiveRegion: true,
           key: getUniqueId()
         });
       }, 1500);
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/AlertGroup/AlertGroup.tsx
```diff
@@ -10,6 +10,8 @@ export interface AlertGroupProps extends Omit<React.HTMLProps<HTMLUListElement>,
   children?: React.ReactNode;
   /** Toast notifications are positioned at the top right corner of the viewport */
   isToast?: boolean;
+  /** Dyanmic alerts need to be rendered in pre-existing live region */
```
seanforyou23, line 13: Instead of documenting how another component is related to this option, let's consider just explaining what this option does for this component. For example "isDynamic AlertGroup's turn the container into a live region so that changes to content within the AlertGroup, such as appending an Alert, are reliably announced to assistive technology".

packages/react-core/src/components/AlertGroup/AlertGroupInline.tsx
```diff
@@ -7,9 +7,14 @@ export const AlertGroupInline: React.FunctionComponent<AlertGroupProps> = ({
   className,
   children,
   isToast,
+  isDynamic,
   ...rest
 }: AlertGroupProps) => (
-  <ul className={css(styles.alertGroup, className, isToast ? styles.modifiers.toast : '')} {...rest}>
+  <ul
+    aria-live={isDynamic ? 'polite' : null}
```
seanforyou23, line 14: In [Alert](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Alert/Alert.tsx#L173-L176), we use the "belt and suspenders" approach, in case certain user agent doesn't default to the same implicit aria roles. Can we explicitly set aria-atomic here so it's consistent with the Alert strategy? Users can still override that default by passing aria-atomic in addition to isDynamic/isLiveRegion. The benefit is that they are getting sensible defaults without having to do a lot of research/testing on the subject.

seanforyou23 (commented):
I'm for this change. I think it draws attention to a critical aspect of the overall AX and provides a nice hook for ensuring Alerts are announced consistently.

The only thing I wonder is if the extra API part is needed at all, and it's not just a documentation/illustration exercise. For example, setting `isDynamic` is just like typing `aria-live="polite"`. Are we missing an opportunity to expose developers to the aria API by doing it for them? Just asking that question for conversation's sake, I think we should use it but just rename it to be more declarative.

What's missing may be an explanation in the documentation page about how it's required for the live region container to exist on the page prior to the content being changed for assistive technology to "notice" the update which in turn is what allows the changes to be announced. For example..

Under "Static alert group" example heading, maybe mention

> These alerts are discoverable from within the normal page content flow, and will _not_ be announced individually/explicitly to assistive technology.

Under "Toast alert group" example heading(or whichever example uses isDynamic/isLiveRegion), maybe mention something like 

> Alerts asynchronously appended into dynamic AlertGroups will be announced to assistive technology at the moment the change happens, following the strategy used for aria-atomic, which defaults to false, which means only changes of type "addition" will be announced. Of course, this is dependent on the user agent in question.

nicolethoen:
This is great! Should we update the Dynamic alert examples as part of this effort, too?

nicolethoen:
@jessiehuff should we make a follow up issue to update Dynamic alerts examples and dynamic helper text examples?

kmcfaul (dismissed):

nicolethoen (approved):

packages/react-core/src/components/Alert/examples/Alert.md
```diff
@@ -314,9 +304,11 @@ class AsyncLiveRegionAlert extends React.Component {
             Stop Async Info Alerts
           </button>
         </InputGroup>
-        {this.state.alerts.map(({ title, variant, isLiveRegion, key }) => (
-          <Alert variant={variant} title={title} isLiveRegion={isLiveRegion} key={key} />
+        <AlertGroup isLiveRegion aria-live="polite" aria-relevant="additions text" aria-atomic="false">
```
seanforyou23, line 307: It's great to show the user that they can override default values, how about since this example is very similar to the "Async alert group" example, we make a couple small changes so that they each represent an optimal use case?

So, for this example, the only thing I'd suggest changing would be to remove the redundant aria-live/relevant/atomic properties since they are already set by default in isLiveRegion and we're not overriding them. Using the "polite" announcement strategy here makes sense because they are _info_ alerts and not as critical to the users immediate attention. Also, consider increasing the timeout from 1500ms to ~4500 so that each announcement finishes before the next starts. That way the example is less rappid fire and more realistic.

Then, for AlertGroup "[Async](https://github.com/jessiehuff/patternfly-react/blob/fix/live-alert-group/packages/react-core/src/components/AlertGroup/examples/AlertGroup.md#async-alert-group)" example we could add `aria-live="assertive"` as an override and switch those alerts from info to danger (also increasing timeout for those). This would perfectly simulate how the alert might be announced to the user in a real app setting. Using the "assertive" announcement strategy here makes sense because they are _danger_ alerts which are more likely to require the users immediate attention from a UX point of view.

Beyond that, the only other thing I can think of that would useful would be to add a small description of what's going on in each of these examples and why the different a11y-related properties were used and what the end result is on the AX.

seanforyou23 (commented):
These changes are great! I have one small recommendation on how to make the Alert/AlertGroup async examples more helpful. Hope it helps!

seanforyou23 (approved):
Nice work!

nicolethoen (approved):

kmcfaul (approved):

Issue #5933, opened 2021-06-21
Component Tree View: enhance with a more flexible toolbar/header  


gildub:
Following initial discussion from https://github.com/patternfly/patternfly-react/issues/4797#issuecomment-739988375, we'd like to continue on the idea to "make the toolbar/header area flexible to pass in anything you want including a search input, actions, or anything that makes sense"

tlabaj:
cc @mcarrano can we get this on the roadmap please.

mcarrano:
Sure. I will get this on our roadmap. Is there a timeframe when this is needed? Right now I'm scheduling into Q4.

gildub:
@mcarrano, 

The konveyor/forklift-ui project has this feature on the target list for it's next release.
I'm currently working on a PR, so hopefully that would be available sooner than later.

mcarrano:
OK. Our next scheduled release is August 4.

gildub:
@mcarrano, great, working on it to get it pushed asap. Thank you.

PR #5959, opened 2021-06-28
feat(TreeView): Add toolbarItems and toolbarLeft props

gildub:
Add optional `toolbarItems` prop. When present it wraps the search field.

Add optional `toolbarLeft` prop. By default `toolbarItems` positions its content on the left of search item.
This allow the `toolbarItems` to be positioned on the right.

Closes https://github.com/patternfly/patternfly-react/issues/5933

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeViewList.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewList.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import { TreeViewSearch } from './TreeViewSearch';
 import { Divider } from '../Divider';
+import { Toolbar, ToolbarContent, ToolbarItem } from '../Toolbar';
 
 export interface TreeViewListProps {
   /** Flag indicating if the tree view is nested under another tree view */
@@ -10,6 +11,10 @@ export interface TreeViewListProps {
   onSearch?: (event: React.ChangeEvent<HTMLInputElement>) => void;
   /** Additional props for search input */
   searchProps?: any;
+  /** Optional ToolbarItem, when used it wraps search field  */
+  toolbarItems?: React.ReactNode[];
+  /** By default toolbarItems are on left */
+  toolbarLeft?: boolean;
   /** Child nodes of the current tree view */
   children: React.ReactNode;
 }
@@ -18,11 +23,39 @@ export const TreeViewList: React.FunctionComponent<TreeViewListProps> = ({
   isNested = false,
   onSearch,
   searchProps,
+  toolbarItems,
+  toolbarLeft,
   children,
   ...props
 }: TreeViewListProps) => (
   <>
-    {onSearch && (
+    {toolbarItems && onSearch && toolbarLeft && (
+      <React.Fragment>
+        <Toolbar>
+          <ToolbarContent>
+            {toolbarItems.map((element, index) => (
+              <ToolbarItem key={index}>{element}</ToolbarItem>
+            ))}
+            <TreeViewSearch onChange={onSearch} {...searchProps} />
+          </ToolbarContent>
+        </Toolbar>
+        <Divider />
+      </React.Fragment>
+    )}
+    {toolbarItems && onSearch && !toolbarLeft && (
+      <React.Fragment>
+        <Toolbar>
+          <ToolbarContent>
+            <TreeViewSearch onChange={onSearch} {...searchProps} />
+            {toolbarItems.map((element, index) => (
+              <ToolbarItem key={index}>{element}</ToolbarItem>
+            ))}
+          </ToolbarContent>
+        </Toolbar>
+        <Divider />
+      </React.Fragment>
+    )}
+    {!toolbarItems && onSearch && (
       <React.Fragment>
         <TreeViewSearch onChange={onSearch} {...searchProps} />
         <Divider />
--- a/packages/react-core/src/components/TreeView/TreeView.tsx
+++ b/packages/react-core/src/components/TreeView/TreeView.tsx
@@ -65,6 +65,8 @@ export interface TreeViewProps {
   compareItems?: (item: TreeViewDataItem, itemToCheck: TreeViewDataItem) => boolean;
   /** Class to add to add if not passed a parentItem */
   className?: string;
+  toolbarItems?: React.ReactNode[];
+  toolbarLeft?: boolean;
 }
 
 export const TreeView: React.FunctionComponent<TreeViewProps> = ({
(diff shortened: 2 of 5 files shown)


packages/react-core/src/components/TreeView/TreeViewList.tsx
```diff
@@ -18,11 +23,35 @@ export const TreeViewList: React.FunctionComponent<TreeViewListProps> = ({
...
+        <Toolbar>
+          <ToolbarContent>
+            <TreeViewSearch onChange={onSearch} {...searchProps} />
+            {toolbarItems}
+          </ToolbarContent>
+        </Toolbar>
+        <Divider />
+      </React.Fragment>
+    )}
+    {!toolbarItems && onSearch && (
```
mturley, line 54: @gildub Instead of having one `toolbarItems` prop and a boolean for whether to put it on the left or right, it might be easier to have `toolbarItemsLeft` and `toolbarItemsRight` props. That way we can support having items on both the left and the right, and we don't need two conditions here.

We also need to support the case where we have toolbar items but no search bar.

packages/react-core/src/components/TreeView/TreeViewList.tsx
```diff
@@ -18,11 +23,35 @@ export const TreeViewList: React.FunctionComponent<TreeViewListProps> = ({
...
   children,
   ...props
 }: TreeViewListProps) => (
   <>
-    {onSearch && (
+    {toolbarItems && onSearch && toolbarLeft && (
+      <React.Fragment>
+        <Toolbar>
+          <ToolbarContent>
+            {toolbarItems}
```
nicolethoen, line 36: In a `Toolbar` each item needs to be wrapped in a `ToolbarItem` component.

nicolethoen (commented):
Thanks so much for updating the tests! Only one comment :)

packages/react-core/src/components/TreeView/TreeViewList.tsx
gildub: @nicolethoen, I'm not following you as we don't have control of the `toolbarItems` content, so how could we handle its details and break it down by items.

nicolethoen: Possibly have the items passed in an array and loop over the array and wrap each item in a `ToolbarItem` component?
That is just how our structure and styles for `Toolbar` are applied according to the design specs. 
https://www.patternfly.org/v4/components/toolbar#items

mturley: We could do that or just require that the consumer put `<ToolbarItem>` around them all themselves, in case there is some custom JSX they need in the future with a variant of `<ToolbarItem>` or a special className on it or something. The consumer has to wrap their items that way in normal Toolbar usage anyway.

nicolethoen: that'd be fine- it'd just need to be included in the docs

gildub:
@nicolethoen, Sorry for delays, pto and sickies. I'm back onto it.

gildub:
@nicolethoen, please review when you get a chance.

nicolethoen:
@kmcfaul @tlabaj I'm looking at this and it will likely work, but I wonder if the TreeView would benefit more from letting the user define their own toolbar like we do for the logviewer. If we need to couple the functionality of the treeView with a search input, we can build a specified treeViewSearch like we did for logViewer, but I don't even know if that's necessary. What are your thoughts?

tlabaj:
@nicolethoen I think that is a great idea. I however am concerned that that might allow too much flexibility.  We would need to export the `TreeViewSearch` component and even if we do so, they may not use it.  
@mcarrano do you have any opinions on this?

mturley:
Better too much flexibility than not enough, in my opinion... within reason of course. I think if we did export `TreeViewSearch` we could include it in examples and demonstrate its standard use, that would fit with other components we require the consumer to use themselves (e.g. Pagination isn't part of Table).

mcarrano:
@nicolethoen @tlabaj I'm not entirely sure what the right answer is here, but I generally not concerned about providing too much flexibility. Its hard to anticipate all the possible use cases that might come up, so a flexible approach might be best.

kmcfaul:
@nicolethoen I agree to follow the pattern with what we did for LogViewer (export the search then use a Context to wire up any dependent functionality). That way the user can define their custom toolbar separately from the component.

mturley:
@kmcfaul I think we don't even need Context here because `onSearch` and `searchProps` are already lifted out and passed in via props. Search logic already has to be performed by the consumer, so this just hands render logic to them as well which makes sense to me.

nicolethoen:
@gildub we have a code freeze today for our release next week. Our next release will be three weeks after that.
It seems from conversations on this PR that we might want to allow the user to build their own toolbar and pass it into the TreeView - to do that we would create a TreeViewSearch component which basically wraps the Search in some context so that it can communicate with the TreeView but also be placed anywhere the consumer would like in their custom Toolbar.

You could see a similar pattern in action in this LogViewer example: https://www.patternfly.org/v4/components/log-viewer#with-complex-toolbar

You are welcome to play around with this proposed solution and there is no rush - but if you'd also like to pass it off to the PatternFly team, we could incorporate the issue and proposal into one of our upcoming milestones as well.

mturley:
Aside from state management, does the TreeViewSearch do anything special the user couldn't do with a regular Search? I guess it has some specific classNames, but I wonder why. Maybe a question for later.

If we wire up context I think that'll be redundant with the `onSearch` / `searchProps` the user already passes into `TreeView`, they can just pass those into their Search as well probably. I don't think we can fully abstract those away because we don't know how the consumer needs to perform their search logic.

kmcfaul:
@mturley You are right about the externally defined search behavior, that is different than LogViewer. And TreeViewSearch doesn't do anything special outside of setting css (one tree view specific wrapping div and other styles from Form). 

We should be able to expose TreeViewSearch itself as a component and stop passing in `onSearch`/`searchProps` to TreeView. All a consumer should have to do to mimic what TreeView does for them currently with search is to render TreeViewSearch and a Divider above the TreeView.

kmcfaul:
Alternatively, using other components like `SearchInput` would also work since the whole search behavior would be moved to the user side. Having `TreeViewSearch` exported & used would just maintain the current styling shown in the examples.

mturley:
@kmcfaul that sounds ideal :)

gildub:
@nicolethoen, I'm happy to pass it to the PF team, not only because I'm going to be away for few weeks but also because the requirements are changing. Thanks for your help.

nicolethoen:
@gildub 100% Thanks for this work! It prompted good conversation!

nicolethoen:
Closing to switch to new approach, handled in this PR: https://github.com/patternfly/patternfly-react/pull/6133

PR #5978, opened 2021-07-01
docs: Move some repeated demo code to wrapper

### LINKED ISSUE #5865: docs: build reusable masthead/navigation codeblock for demos
In full page demos, much of the framework of the page is the same so we can likely build/export a wrapper component that contains the masthead/nav part of the page.

kmcfaul:
**What**: Closes #5865

Moves repeated demo code (Page, PageHeader, PageSidebar, Nav) to a wrapper/helper component to reduce the bulk of some full page demos that are not showcasing these components. Shown in PrimaryDetail and CardView demos for the moment. If it looks all right I will continue refactoring.

@redallen I would love to import the Brand/Avatar images directly into the wrapper component but I was having trouble getting the code to work/recognize the svg files (tried through import and require). Is there a good way to import svgs directly into a tsx file? Otherwise I've got them passed via properties, so it's not a huge deal.

### DIFF
--- a/packages/react-core/src/helpers/DashboardWrapper.tsx
+++ b/packages/react-core/src/helpers/DashboardWrapper.tsx
@@ -1,0 +1,215 @@
+import React from 'react';
+import {
+  Avatar,
+  Brand,
+  Breadcrumb,
+  BreadcrumbItem,
+  Button,
+  ButtonVariant,
+  Dropdown,
+  DropdownGroup,
+  DropdownToggle,
+  DropdownItem,
+  KebabToggle,
+  Nav,
+  NavItem,
+  NavList,
+  Page,
+  PageHeader,
+  PageSidebar,
+  SkipToContent,
+  PageHeaderTools,
+  PageHeaderToolsGroup,
+  PageHeaderToolsItem
+} from '../components';
+import CogIcon from '@patternfly/react-icons/dist/js/icons/cog-icon';
+import HelpIcon from '@patternfly/react-icons/dist/js/icons/help-icon';
+
+export interface DashboardWrapperProps {
+  /** Content rendered inside of Page */
+  children?: React.ReactNode;
+  /** ID of the Page */
+  mainContainerId?: string;
+  /** Custom breadcrumb of the Page */
+  breadcrumb?: React.ReactNode;
+  /** Custom header of the Page */
+  header?: React.ReactNode;
+  /** Custom sidebar of the Page */
+  sidebar?: React.ReactNode;
+  /** Brand image of the Page Header */
+  imgBrand?: string;
+  /** Avatar image of the Page Header */
+  imgAvatar?: string;
+}
+
+export interface DashboardWrapperState {
+  isDropdownOpen: boolean;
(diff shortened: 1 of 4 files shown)


packages/react-core/src/helpers/DashboardWrapper.tsx
```diff
@@ -0,0 +1,215 @@
+import React from 'react';
```
redallen, line 1: I love this class! Can it move to `packages/react-core/src/demos/DashboardWrapper.js`? Having it exported from `react-core` for consumers is strange...

Yes, you'll have to lose the types until we support TS examples, sorry!

redallen (commented):
> Is there a good way to import svgs directly into a tsx file?

I think you'd be able to import them if you follow my DashboardWrapper.js suggestion. Currently there's no good way to import SVGs from `react-core`... we use `react-icons` for that!

packages/react-core/src/helpers/DashboardWrapper.tsx
kmcfaul: Ah I agree, it is a bit weird to get the file from `react-core` since it's just demo code. I'll move it and tweak it to a `js` file.

kmcfaul: Was also getting circular dependencies having it in helpers so this should resolve that as well

kmcfaul:
Placeholder to update = list of to-be-refactored full page demos

kmcfaul:
@nicolethoen updated list of full page demos that can benefit from refactoring

redallen (dismissed):
LGTM! Thanks for dealing with the `examples` bug/feature Katie!

kmcfaul:
Exported a `DashboardHeader` and `DashboardBreadcrumb` (and used both inside `DashboardWrapper`) to re-use code and allow the substitution of just the header and breadcrumb components (for demos like `Nav` where aspects of Page are directly showcased/modified but header/breadcrumb may remain the same).

kmcfaul:
Looking over `NotificationDrawer` in more detail, the only code we'd save would be the breadcrumb as the demo modifies the Header/Page directly to show how to implement the drawer. Does anyone think that's worth refactoring the tiny bit out, or should I leave it as is?

redallen:
If it's mostly the same page structure I'd say it's worth refactoring.

kmcfaul:
It's mostly the same structure, but I'm not sure it's worth obfuscating where the properties get hooked into by passing them to a wrapper component instead of showing directly how they're passed in?

redallen:
You could make every page wrapper in a drawer if that's less confusing. Whatever the case, the point of refactoring is to make code more maintainable. If it's less maintainable by future @kmcfaul , then don't do it!

kmcfaul:
I think I will leave NotficationDrawer as is for now, same as Page, because both directly showcase Page and its components while the other full page demos only use Page as a framework to display something else.

tlabaj (approved):

nicolethoen (approved):

redallen (approved):

PR #5984, opened 2021-07-02
chore(Table): clean up and expand table docs

### LINKED ISSUE #2999: PF4 Table: The documentation needs improvements
- [ ] Clearly define the type of the `cells` and `rows` properties of the `Table` component
    - [ ] What is `IRow` - All properties here should be mentioned in the documentation
    - [ ] What is `ICell` - All properties here should be mentioned in the documentation
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-table/src/components/Table/Table.tsx#L127..L138
- [ ] What is `RowWrapper` and `BodyWrapper` - the props description mentions only that it's a Function.. 
- [ ] `cellFormatters`, `columnTransforms`, `cellTransforms`, `formatters`, `transforms` are not documented at all - there is issue about it but let's keep it here as well since this issue is more extended https://github.com/patternfly/patternfly-react/issues/2990
- [ ] The `TableBody` and `TableHeader` components are supposed to be used with properties? The examples never use them (Just use them without properties - and I guess the properties we pass to Table are propagated internally)
But it really makes sense for the consumers to be able to use `TableBody` so we need examples for that.
     - [ ] trying to use `TableBody` outside of table and without properties passed (which should be fine since the docs don't have any marked as required props gives error: TypeError: headerData is null 
- [ ] The link to the properties takes you to wrong place when clicked https://www.patternfly.org/v4/documentation/react/components/table/#tablebody-props
- [ ] `rowLabeledBy` property does not make sense to be part of the table (or I understand wrong what it is about?) It's row should be able to have it's only `rowLabeledBy` property.
- [ ] I could not find anywhere documented how to apply custom classes on Rows (in our use case we want to mark rows with a color that signals a warning if the resource on that row has some issue) 
- [ ] Document `TableVariant`

Note:  I will keep editing that inline as I find more issues.

nicolethoen:
Closes #2999

### DIFF
--- a/packages/react-table/src/components/Table/EditableSelectInputCell.tsx
+++ b/packages/react-table/src/components/Table/EditableSelectInputCell.tsx
@@ -1,31 +1,17 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
-import { Select, SelectOptionObject, SelectProps } from '@patternfly/react-core';
+import { Select, SelectOptionObject } from '@patternfly/react-core';
 import inlineStyles from '@patternfly/react-styles/css/components/InlineEdit/inline-edit';
 import formStyles from '@patternfly/react-styles/css/components/Form/form';
+import { EditableSelectInputProps } from './base';
 
 export interface IEditableSelectInputCell extends Omit<React.HTMLProps<HTMLElement | HTMLDivElement>, 'onSelect'> {
   /** Row index of this select input cell */
   rowIndex: number;
   /** Cell index of this select input cell */
   cellIndex: number;
-  /** Data structure containing:
-   * value - to display in the cell,
-   * name - of the select input,
-   * isSelectOpen - flag controlling isOpen state of select,
-   * selected - string or SelectOptionObject, or an array of strings or SelectOptionObjects representing current selections
-   * options - Array of react elements to display in the select menu,
-   * editableSelectProps - props to be passed down to the Select component housed inside this editable select input cell
-   * arbitrary data to pass to the internal select component in the editable select input cell */
-  props: {
-    name: string;
-    value: string | string[];
-    isSelectOpen: boolean;
-    selected: string | SelectOptionObject | (string | SelectOptionObject)[];
-    options: React.ReactElement[];
-    editableSelectProps?: SelectProps;
-    [key: string]: any;
-  };
+  /** Props to build the select component */
+  props: EditableSelectInputProps;
   /** Event handler which fires when user selects an option in this cell */
   onSelect: (
     newValue: string | SelectOptionObject,
--- a/packages/react-table/src/components/Table/EditableTextCell.tsx
+++ b/packages/react-table/src/components/Table/EditableTextCell.tsx
@@ -3,6 +3,7 @@ import { TextInput } from '@patternfly/react-core/dist/js/components/TextInput';
 import inlineStyles from '@patternfly/react-styles/css/components/InlineEdit/inline-edit';
 import formStyles from '@patternfly/react-styles/css/components/Form/form';
 import { css } from '@patternfly/react-styles';
+import { EditableTextCellProps } from './base';
 
 export interface IEditableTextCell extends React.HTMLProps<HTMLDivElement> {
   /** The current value of the text input */
@@ -11,15 +12,8 @@ export interface IEditableTextCell extends React.HTMLProps<HTMLDivElement> {
   rowIndex: number;
   /** Cell index of this text cell */
   cellIndex: number;
-  /** Data structure containing:
-   * value - to display in the cell,
-   * name - of the text input,
-   * arbitrary data to pass to the internal text input in the editable text cell */
-  props: {
-    name: string;
-    value: string;
-    [key: string]: any;
-  };
+  /** Props to build the input */
+  props: EditableTextCellProps;
   /** Event handler which fires when user changes the text in this cell */
   handleTextInputChange: (
     newValue: string,
--- a/packages/react-table/src/components/Table/base/types.tsx
+++ b/packages/react-table/src/components/Table/base/types.tsx
@@ -6,13 +6,33 @@
  */
 
 import * as React from 'react';
-import { TooltipProps, PopoverProps } from '@patternfly/react-core';
+import {
+  TooltipProps,
+  PopoverProps,
+  DropdownPosition,
+  DropdownDirection,
+  SelectOptionObject,
+  SelectProps
+} from '@patternfly/react-core';
 import { TableComposable } from '../../TableComposable/TableComposable';
 import { Thead } from '../../TableComposable/Thead';
 import { Tbody } from '../../TableComposable/Tbody';
 import { Tr } from '../../TableComposable/Tr';
 import { Th } from '../../TableComposable/Th';
 import { Td } from '../../TableComposable/Td';
+import {
+  IActions,
+  ISortBy,
+  OnCheckChange,
+  OnCollapse,
+  OnExpand,
+  OnFavorite,
+  OnSelect,
+  OnSort,
+  OnToggleRowDetails,
+  OnTreeRowCollapse
+} from '../TableTypes';
+import { CustomActionsToggleProps } from '../ActionsColumn';
 
 // Cell Type
 export interface CellType {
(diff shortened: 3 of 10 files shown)


packages/react-table/src/components/Table/EditableSelectInputCell.tsx
```diff
@@ -1,31 +1,17 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
-import { Select, SelectOptionObject, SelectProps } from '@patternfly/react-core';
+import { Select, SelectOptionObject } from '@patternfly/react-core';
 import inlineStyles from '@patternfly/react-styles/css/components/InlineEdit/inline-edit';
 import formStyles from '@patternfly/react-styles/css/components/Form/form';
+import { EditableSelectInputProps } from './base';
```
tlabaj, line 6: These do not appear to be included in the md file.  Can we do that so they show up in the docs.

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -7,21 +7,28 @@ propComponents:
...
     'TableComposable',
     'Thead',
     'Tbody',
     'Tr',
     'Th',
     'Td',
     'Caption',
+    'TableText',
+    'EditableSelectInputCell',
+    'EditableTextCell'
```
tlabaj, line 19: add `EditableSelectInputProps`, `EditableTextCellProps`, `TdActionsType`, `TdSelectType`, `TdTreeRowType`,  `IActions`, `TdCompoundExpandType`, `TdFavoritesType`, `TdDraggableType`, `ThInfoType`,  and `TdExpandType` here.

packages/react-table/src/components/Table/EditableTextCell.tsx
```diff
@@ -11,15 +12,8 @@ export interface IEditableTextCell extends React.HTMLProps<HTMLDivElement> {
...
-   * value - to display in the cell,
-   * name - of the text input,
-   * arbitrary data to pass to the internal text input in the editable text cell */
-  props: {
-    name: string;
-    value: string;
-    [key: string]: any;
-  };
+  /** Props to build the input */
+  props: EditableTextCellProps;
```
tlabaj, line 16: can 'EditableTextCellProps' also be added to md file.

packages/react-table/src/components/Table/base/types.tsx
```diff
@@ -22,6 +42,80 @@ export interface CellType {
...
+  disable?: boolean;
+  /** Actions dropdown position */
+  dropdownPosition?: DropdownPosition;
+  /** Actions dropdown direction */
+  dropdownDirection?: DropdownDirection;
+  /** Custom toggle for the actions menu */
+  actionsToggle?: (props: CustomActionsToggleProps) => React.ReactNode;
+}
+
+export interface TdExpandType {
```
tlabaj, line 73: Can we also add this to md file?

packages/react-table/src/components/Table/base/types.tsx
```diff
@@ -22,6 +42,80 @@ export interface CellType {
...
+  isExpanded: boolean;
```
tlabaj, line 74: this prop is missing a description.

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -2789,6 +2985,36 @@ ComposableTableSelectableRadio = () => {
 
 This example demonstrates adding actions as the last column. The header's last cell is an empty cell, and each body row's last cell is an action cell.
 
+To make a cell an action cell, pass a `TdActionsType` object via the `actions` prop on a rows's last `Td`.
+
+```
+interface TdActionsType {
```
tlabaj, line 2991: add this one to md file too.

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -2789,6 +2985,36 @@ ComposableTableSelectableRadio = () => {
...
+  disable?: boolean;
+  /** Actions dropdown position */
+  dropdownPosition?: DropdownPosition;
+  /** Actions dropdown direction */
+  dropdownDirection?: DropdownDirection;
+  /** Custom toggle for the actions menu */
+  actionsToggle?: (props: CustomActionsToggleProps) => React.ReactNode;
+}
+
+interface IAction {
```
tlabaj, line 3004: this should also me added to MD file

packages/react-table/src/components/Table/base/types.tsx
```diff
@@ -22,6 +42,80 @@ export interface CellType {
...
+  /** The row index */
+  rowIndex: number;
+  /** The column index */
+  columnIndex?: number;
+  /** On toggling the expansion */
+  onToggle?: OnCollapse;
+}
+
+export interface TdCompoundExpandType {
```
tlabaj, line 83: All these interfaces that the consumer needs should be included in the MD file so the are documented.

tlabaj (changes requested):
This is great Nicole.  A few comments.

tlabaj:
FYI interfaces that added to MD file are not currently showing the types.  Zack has a fix up https://github.com/patternfly/patternfly-org/pull/2586

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -2439,12 +2550,30 @@ ComposableTableBasic = () => {
...
-
 - If you add the `noWrap` prop to `Thead`, it won't wrap it if there is no space
 - You can add the `textCenter` prop to `Th` or `Td` to center the contents
 - You can pass `onClick`, `className`, `style` and more to `Tr`
 
+To add a header tooltip or popover to `Th`, pass a `ThInfoType` object via the `info` prop.
+
+`ThInfoType` (excerpt): 
+```
+interface ThInfoType {
```
tlabaj, line 2561: I don't know that these excerpts are needed once you add the interfaces to the MD file.  Maybe a link to the interface docs?

kmcfaul (approved):
This looks great @nicolethoen! 

My only concern is that the docs page is getting very long (and will be longer once all the interfaces are shown). It's not a blocker, but I'm curious if we could separate Table & Composable Table somewhere down the line? Maybe making a Table section outside of the components folder on the docs (to show both table implementations and demos perhaps).

dlabrecq (commented):
We have a similar scenario with Victory chart types. It's a real pain to look up what the `AnimatePropTypeInterface` type represents, for example.

The `EditableTextCellProps` props feels like the same. That is, I need to look up `EditableTextCellProps` to see that it accepts a name / value pair.

For charts, we're using the `@proptype` annotation Zack implemented for PatternFly docs. This eliminates users from having to lookup prop types.

```
  /**
   * The animate prop specifies props for VictoryAnimation to use.
   * The animate prop should also be used to specify enter and exit
   * transition configurations with the `onExit` and `onEnter` namespaces respectively.
   *
   * @propType boolean | object
   */
  animate?: boolean | AnimatePropTypeInterface;
```

nicolethoen:
> That said, I could not find where `EditableTextCellProps` is used in the docs. It appears that`IEditableSelectInputCell` uses it, but even that doesn't appear in the docs.

Oh this is super helpful! I'll look at @propType and make sure that I am not mixing up type references around editable table cells.

nicolethoen:
hm... @dlabrecq 
I'm conflicted because the `Td` and `Th` components are part of `TableComposable` situation and all those extra types (like `TdExpandType`, `TdFavoritesType`, etc) are objects with very specific  definitions and fields. And usually the user is only using one or two of those props at a time to transform a cell. So I don't think I can replace any prop definitions with a simple `object`. I'd want to spell out the prop types explicitly somewhere.

In the case of `EditableTextCellProps`, that is an object passed to `EditableTextCell`. I also added a reference to it in the text above the Editable Rows example.

nicolethoen:
I've separated the ComposableTable and Table docs onto two different tabs (a change to patternfly-org will be needed to put the new tabs in the right order).

What do people think of the names of the tabs? Do we like this?

redallen (dismissed):
I love it!

tlabaj (commented):
I love the new tabs!  could the composable tab go first?
Also maybe add Rect to the name for consistency  or `React Composable Table`and `React Legacy Table`.

nicolethoen:
> could the composable tab go first?

@tlabaj yes, I will have to wire the tab order up in org and pull in the new theme changes to see it reflected in the react workspace.

nicolethoen:
@kmcfaul @mcarrano do you like `React Composable Table` and `React Legacy Table` for tab names?

mturley:
+1 for calling the old table "legacy" and discouraging its use in new projects 🙃

mturley:
@nicolethoen this looks fantastic!!

I wonder if we should put the "React demos" tab last, after the composable and legacy tabs? They are currently using the legacy table, and in most places where we have demos that tab comes last.

nicolethoen:
@mturley indeed - I will have to wire the tab order up in org and pull in the new theme changes to see it reflected in the react workspace.

jschuler:
Nice job on this! Definitely better having them in separate tabs

dlabrecq:
> hm... @dlabrecq
> I'm conflicted because the `Td` and `Th` components are part of `TableComposable` situation and all those extra types (like `TdExpandType`, `TdFavoritesType`, etc) are objects with very specific definitions and fields. And usually the user is only using one or two of those props at a time to transform a cell. So I don't think I can replace any prop definitions with a simple `object`. I'd want to spell out the prop types explicitly somewhere.
> 
> In the case of `EditableTextCellProps`, that is an object passed to `EditableTextCell`. I also added a reference to it in the text above the Editable Rows example.

I think adding links to the interfaces would be a big help. See https://github.com/patternfly/patternfly-org/issues/2608

tlabaj:
Yes.  The links is a great idea.

mcarrano (approved):
Looks good. Thanks @nicolethoen .

redallen (approved):
I still love it! :tada:

kmcfaul (commented):
Looks great in separate tabs! 

For the tab names, I think it may be clearer to say `Composable Table` and `Legacy Table` instead of `React composable` and `React legacy`? Having the table page default to the composable table docs instead of the demos page would be nice as well, but not a major issue.

nicolethoen:
@kmcfaul the page will default to the Composable Table docs once I make a subsequent update in org.

tlabaj (approved):
This is great Nicole!

PR #5985, opened 2021-07-02
docs(Pagination): Export interfaces in md file so prop descriptions are documented

### LINKED ISSUE #1882: Pagination - add prop for page size menu toggle label

tlabaj:
**What**: Closes #1882

Added the interfaces to the md file so that the props are documented for the user.  Updated a few prop descriptions @jessiehuff @jgiardino  the `titles.optionsToggle` is what sets the toggles aria label.

### DIFF
--- a/packages/react-core/src/components/Pagination/OptionsToggle.tsx
+++ b/packages/react-core/src/components/Pagination/OptionsToggle.tsx
@@ -9,7 +9,7 @@ import { DropdownToggle } from '../Dropdown';
 export interface OptionsToggleProps extends React.HTMLProps<HTMLDivElement> {
   /** The type or title of the items being paginated */
   itemsTitle?: string;
-  /** The text to be displayed on the Options Toggle */
+  /** Accessible label for the Options Toggle */
   optionsToggle?: string;
   /** The Title of the Pagination Options Menu */
   itemsPerPageTitle?: string;
@@ -40,7 +40,7 @@ export interface OptionsToggleProps extends React.HTMLProps<HTMLDivElement> {
 let toggleId = 0;
 export const OptionsToggle: React.FunctionComponent<OptionsToggleProps> = ({
   itemsTitle = 'items',
-  optionsToggle = 'Select',
+  optionsToggle = 'Items per page',
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   itemsPerPageTitle = 'Items per page',
   firstIndex = 0,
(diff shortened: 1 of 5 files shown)


nicolethoen (approved):

jgiardino:
When I tested the following prop and changed the `optionsToggle` to "Testing", I didn't see that reflected in the html. This is the prop that I added to the first example:

`titles={ items: '', page: '', itemsPerPage: 'Items per page', perPageSuffix: 'per page', toFirstPage: 'Go to first page', toPreviousPage: 'Go to previous page', toLastPage: 'Go to last page', toNextPage: 'Go to next page', optionsToggle: 'Testing', currPage: 'Current page', paginationTitle: 'Pagination', ofWord: 'of' }`

It's very possible I'm applying this prop incorrectly, though, because I'm not seeing any values I define in the `titles` prop display in the component.

kmcfaul (approved):

jessiehuff (commented):
I agree with @jgiardino here. I can't seem to get the `optionsToggle` prop to reflect as an `aria-label` on the pagination toggle. Is there a way to do this that we're missing?

tlabaj:
@jessiehuff @jgiardino I thin you have the syntax wrong.  You need to pass the `titles` prop an object.  try adding this to the example to see the toggle aria label updated `titles={{optionsToggle: "Items per page aria-label test"}}`

jgiardino:
Thanks for pointing that out, @tlabaj! I was missing the extra `{}` when I tested it earlier. I tested again and it works! 🎉

jessiehuff (approved):
Ah ok, I also had a syntax issue. Cool! 😄  Do you think we need to add this to one of the examples or do you think it's already clear?

redallen (approved):

PR #5987, opened 2021-07-06
Remove PatternFly 3 as a dependency from react-catalog-view-extension

rhamilto:
**What**: 
react-catalog-view-extension includes PatternFly 3 as a dependency because of a couple of Bootstrap 3 classes that are in use as well as a dependence on some PatternFly 3 LESS variables.  As a result, any consumers of react-catalog-view-extension are resulting in PatternFly 3 installing as a dependency.  As part of https://issues.redhat.com/browse/CONSOLE-2361, we are trying to *remove* PatternFly 3 as a dependency from OpenShift, but are blocked by react-catalog-view-extension's dependence on it.  

This PR removes PatternFly 3 as a dependency from react-catalog-view-extension, but results in near-identical rendering *assuming* any consumers of the package had PatternFly 3 in use as a dependency.  If PatternFly 3 is *not* in use as a dependency, there will be one noticeable change in the spacing between `FilterSidePanelCategoryItem`s (it will increase).  I opted for this route as my assumption is that any consumers of react-catalog-view-extension have PatternFly 3 in use because react-catalog-view-extension depended on it.

cc: @janwright73 @jcaianirh @sg00dwin @spadgett @christianvogt

### DIFF
--- a/packages/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
+++ b/packages/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
@@ -1,6 +1,6 @@
 .catalog-tile-pf {
   &.featured {
-    border-top: 2px solid $color-pf-blue-300;
+    border-top: 2px solid $pf-color-blue-300;
   }
 
   &:active,
@@ -22,22 +22,22 @@
   padding-bottom: 16px;
 
   .catalog-tile-pf-title {
-    font-size: ($font-size-base + 1);
+    font-size: 15px;
     font-weight: 400;
   }
 
   .catalog-tile-pf-subtitle {
-    color: $color-pf-black-500;
-    font-size: ($font-size-base - 1);
+    color: $pf-color-black-500;
+    font-size: 13px;
     font-weight: initial;
 
     a {
-      color: $color-pf-blue-300;
+      color: $pf-color-blue-300;
       text-decoration: none;
     }
 
     a:hover {
-      color: $color-pf-blue-200;
+      color: $pf-color-blue-200;
       text-decoration: none;
     }
   }
--- a/packages/react-catalog-view-extension/README.md
+++ b/packages/react-catalog-view-extension/README.md
@@ -24,13 +24,11 @@ import { Component } from 'react-catalog-view-extension';
 
 #### Styling:
 
-Example with LESS:
+Example with SCSS:
 
 ```
-@import "~bootstrap/sass/variables";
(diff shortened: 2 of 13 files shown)


redallen (approved):
> I opted for this route as my assumption is that any consumers of react-catalog-view-extension have PatternFly 3 in use because react-catalog-view-extension depended on it.

I'd say that's a safe assumption since it's listed as a direct dependency.

I briefly checked every example which still seems to be working, so LGTM!

jcaianirh:
@mcoker would you mind reviewing this pr

mcoker (approved):
I can't verify how this looks in the PR preview build, but verified w/ @redallen that it looks like OS is the only product using catalog view, so I'll assume everything is fine.

jcaianirh (approved):

jcaianirh:
@mcoker I can preview this after the package version is generated

rhamilto:
> @mcoker I can preview this after the package version is generated

I tested the changes in my local OpenShift environment while developing this PR, but a second set of eyes certainly won't hurt!

packages/react-catalog-view-extension/sass/react-catalog-view-extension/_catalog-tile.scss
```diff
@@ -22,22 +22,22 @@
...
 
     a {
-      color: $color-pf-blue-300;
+      color: $pf-color-blue-300;
       text-decoration: none;
     }
 
     a:hover {
-      color: $color-pf-blue-200;
+      color: $pf-color-blue-200;
```
mcoker, line 40: [PF3 is `#7dc3e8`](https://github.com/patternfly/patternfly-3/blob/1d5ef31f3ccdae807ee9215b3f420d4243be1e64/src/sass/converted/patternfly/_color-variables.scss#L14)

[PF4 is `#73bcf7`](https://github.com/patternfly/patternfly/blob/9cc522b54163558ca0dfce4c35e2c80cd81cc2ab/src/patternfly/sass-utilities/colors.scss#L16)

mcoker (commented):
Pulled the changes in locally. Just a few observations that the colors are slightly different. I'm assuming that's a good thing since you're using PF4 colors now.

jcaianirh:
@mcoker thanks for calling those out!

rhamilto:
> Pulled the changes in locally. Just a few observations that the colors are slightly different. I'm assuming that's a good thing since you're using PF4 colors now.

Yeah.  I assumed that was intentional since those colors maintained the same variable name values, but values shifted slightly.

PR #5991, opened 2021-07-06
fix: 🐛 Tooltip: clean up timeouts on unmount to prevent warning

mturley:
If a Tooltip is unmounted while one of its animation/delay timers is still running, it will result in a console warning from React about updating state on an unmounted component:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

This can be reproduced by triggering a Tooltip and then causing it to unmount while its hide animation is playing. This is easier to reproduce if you increase the animationDelay, but I have seen it in normal usage as well.

CodeSandbox to reproduce the issue: https://codesandbox.io/s/serverless-bush-0k3pi?file=/index.js

Screen capture of issue being reproduced:
Gs1qSckq9h

This PR adds a `useEffect` to clean up those timers when the Tooltip unmounts. It also factors all those conditional `clearTimeout` calls into a helper `clearTimeouts` to shorten the code in this new effect and other places the timeouts are cleared.

### DIFF
--- a/packages/react-core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react-core/src/components/Tooltip/Tooltip.tsx
@@ -143,6 +143,23 @@ export const Tooltip: React.FunctionComponent<TooltipProps> = ({
   const transitionTimerRef = React.useRef(null);
   const showTimerRef = React.useRef(null);
   const hideTimerRef = React.useRef(null);
+
+  const clearTimeouts = (timeoutRefs: React.RefObject<any>[]) => {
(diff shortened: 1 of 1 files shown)


mturley:
Blocking low-priority downstream issue: https://github.com/konveyor/forklift-ui/issues/553

dlabrecq (approved):

redallen (approved):
I've been wondering where those warnings are coming from, thanks!

Issue #6021, opened 2021-07-13
LogViewer: Empty lines and line height


mfocko:
**Describe the issue. What is the expected and unexpected behavior?**

I have been trying to use log viewer in our dashboard, I have encountered 2 issues:

- [ ] Line height: which seems to be a bit of an overkill, since it is (at least looks like) calculated by the max height (after wrapping) and applies for **all** lines. I would expect this to be flexible and change the line height only if necessary.
- [X] Empty lines in `data`: I have noticed strange behavior with empty lines included in the logs which result in pretty big dead space in the viewer, that is followed by the rest of the logs. (can be seen in the sandbox as the 2nd line)

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

https://codesandbox.io/s/friendly-liskov-724b8?file=/index.js

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

bug; related to packit/dashboard#120 (haven't been planned to be worked on yet)

nicolethoen:
@mfocko it's great to know you're attempting to consume log-viewer!
The virtualized nature of the log viewer does provide some limitations that have forced some quirky solutions thus far. But ideally we will work to iron those out as we go. Line height has definitely been proven to be a complicated problem. Definitely let us know if you see opportunities in the code to improve the logic. Hopefully the team can reevaluate the line height logic shortly.

mfocko:
Seems to be fixed by #6201

PR #6033, opened 2021-07-15
fix(LogViewer): Do not overestimate size for empty lines

mfocko:
- Do not estimate 600 rows for empty lines => wrong priority for ‹||›
- After fixing previous bug, do not estimate 0 height for empty lines,
  i.e. there is always at least one row in the log

Signed-off-by: Matej Focko <mfocko@redhat.com>

**What**: Part of #6021

**Additional issues**:


nicolethoen (approved):
Thank you for the fix!

redallen (approved):

PR #6037, opened 2021-07-15
SearchInput: allow hasWordAttrLabel to be React.ReactNode as well

KKoukiou:
In cockpit we want to use this component variant, but the 'Has words' field needs explanation. For this reason we want to use a helpicon popover as seen in the screenshot.

search-box-label

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -33,7 +33,7 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
    * The new form elements can be wrapped in a FormGroup component for automatic formatting */
   formAdditionalItems?: React.ReactNode;
   /** Attribute label for strings unassociated with one of the provided listed attributes */
-  hasWordsAttrLabel?: string | React.ReactNode;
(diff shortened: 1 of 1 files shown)


nicolethoen (dismissed):

dlabrecq (dismissed):

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -30,7 +30,7 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>,
   /** Array of attribute values */
   attributes?: string[] | SearchAttribute[];
   /** Attribute label for strings unassociated with one of the provided listed attributes */
-  hasWordsAttrLabel?: string;
+  hasWordsAttrLabel?: string | React.ReactNode;
```
tlabaj, line 33: the type `React.ReactNode` includes `string` so you can remove that here

tlabaj (changes requested):

tlabaj (dismissed):

redallen (dismissed):
Looks good, just need to fix conflicts.

nicolethoen (approved):

redallen (approved):

PR #6038, opened 2021-07-15
feat(Dropdown): Allow for tooltips on aria-disabled items

### LINKED ISSUE #5826: Support for tooltips on disabled dropdown items (aria-disabled variation)
From a discussion in this issue - https://github.com/patternfly/patternfly-react/issues/5127

We added support for a `.pf-m-aria-disabled` variation on dropdown items to support adding tooltips on disabled items

https://github.com/patternfly/patternfly/pull/4072

cc @jschuler @mcarrano

nicolethoen:
**What**: Closes #5826 

You can see it working in the very first example - one of the disabled items is now using aria-disabled with a simple tooltip.

### DIFF
--- a/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
+++ b/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
@@ -267,11 +267,9 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
               role={role}
               onKeyDown={this.onKeyDown}
               onClick={(event: any) => {
-                if (!isDisabled) {
+                if (!isDisabled || !isAriaDisabled) {
                   onClick(event);
-                  if (!isAriaDisabled) {
-                    onSelect(event);
-                  }
+                  onSelect(event);
                 }
               }}
               id={id}
--- a/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
+++ b/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
@@ -267,9 +267,11 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
               role={role}
               onKeyDown={this.onKeyDown}
               onClick={(event: any) => {
-                if (!isDisabled || !isAriaDisabled) {
+                if (!isDisabled) {
                   onClick(event);
-                  onSelect(event);
+                  if (!isAriaDisabled) {
+                    onSelect(event);
+                  }
                 }
               }}
               id={id}
--- a/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
+++ b/packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
@@ -267,7 +267,7 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
               role={role}
               onKeyDown={this.onKeyDown}
               onClick={(event: any) => {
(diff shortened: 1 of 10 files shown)


mcarrano (approved):
Looks great @nicolethoen !

mcoker (approved):
LGTM!

dlabrecq (dismissed):

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
```diff
@@ -264,7 +267,7 @@ export class InternalDropdownItem extends React.Component<InternalDropdownItemPr
               role={role}
               onKeyDown={this.onKeyDown}
               onClick={(event: any) => {
-                if (!isDisabled) {
+                if (!isDisabled && !isAriaDisabled) {
```
kmcfaul, line 270: Should this be an `||` instead? Or does `isDisabled` also set `isAriaDisabled` if `isAriaDisabled` isn't passed in.

jessiehuff (commented):
Hmm what's odd is that this seems to work in JAWS, but not in VO and NVDA. I'm not quite sure why yet.

nicolethoen:
@jessiehuff did you do any more investigation into this? I can try playing with it in VO this afternoon.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
tlabaj: Good question.  Maybe add a test for this scenario.

jessiehuff:
I'm wondering if what Katie pointed out above is affecting the screen reader experience. It seems like if it's just `isAriaDisabled` this might not work on click which the screen reader may rely on? It's also probably worth noting what we did in [button to prevent events](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Button/Button.tsx#L122). Maybe try logging if it gets the onClick and onKeyPress. I would think that the tooltip shouldn't have a problem firing if those are getting triggered.

tlabaj (commented):
Can you update the demo app.

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
nicolethoen: I may be going cross eyed - but i think it's true that if the dropdown item is disabled OR ariaDisabled, then we don't want to fire the passed in onClick or onSelect handlers. So this logic, with the `&&` is right.

nicolethoen:
I opened https://github.com/patternfly/patternfly-react/issues/6109 because tooltips on dropdown items are not screenreader accessible.

tlabaj (approved):

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
kmcfaul: Oh you're right, the double negative with the && got me

kmcfaul (approved):

packages/react-core/src/components/Dropdown/InternalDropdownItem.tsx
redallen: Can rewrite as `isDisabled || isAriaDisabled` to be more clear. de Morgan's theorem states `!A && !B = A || B`.

redallen (approved):

PR #6039, opened 2021-07-15
fix(datepicker): change default parse format to y-m-d from y(-m)?(-d)?

### LINKED ISSUE #6036: Incomplete date entered by text input is parsed as valid date for DatePicker

When we enter in a date into the DatePicker by text input, the date value we're provided as a second argument automatically converts it to the first valid date it sees, which doesn't necessarily come from an ISO-formatted date. While this might be its expected behaviour, I don't think it's useful when DatePicker mainly uses ISO-formatted dates when selected with the picker.
The expected behaviour would be for it to wait until you have entered in a full ISO date to be valid.

https://github.com/patternfly/patternfly-react/blob/aac2e7cf90ea46337ed5bd630beb84209b8551cd/packages/react-core/src/components/DatePicker/DatePicker.tsx#L100

Here's a link to a sandbox containing the issue. 
https://codesandbox.io/s/optimistic-shaw-egk7w?file=/index.js
If you type in 2021 into the datepicker box, with this.state.date tracking the current value, we see that the date outputted is 2021-01-01.

This is a bug. The workaround is just to use the raw value (the first parameter in https://github.com/patternfly/patternfly-react/blob/aac2e7cf90ea46337ed5bd630beb84209b8551cd/packages/react-core/src/components/DatePicker/DatePicker.tsx#L101), but I feel like that makes the second parameter obsolete.

redallen:
**What**: Closes #6036

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-core/src/components/DatePicker/DatePicker.tsx
@@ -53,7 +53,7 @@ export const DatePicker: React.FunctionComponent<DatePickerProps> = ({
   className,
   locale = undefined,
   dateFormat = yyyyMMddFormat,
-  dateParse = (val: string) => new Date(`${val}T00:00:00`),
+  dateParse = (val: string) => val.split('-').length === 3 && new Date(`${val}T00:00:00`),
   isDisabled = false,
   placeholder = 'yyyy-MM-dd',
   value: valueProp = '',


dgutride:
LGTM!

dgutride (approved):

nicolethoen (approved):

dlabrecq (approved):

tlabaj (approved):

PR #6043, opened 2021-07-16
feat(SearchInput): Allow passing custom form elements after the attributes

KKoukiou:
Can be used to link to docs pages for example.

This is for cockpit, we want to use it like this:
Screen Shot 2021-07-16 at 10 17 08

See the last item, it's basically the docs link and and CopyClipboard with the CLI command equivalent generated by the user's input.

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -29,6 +29,9 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   value?: string;
   /** Array of attribute values */
   attributes?: string[] | SearchAttribute[];
+  /* Additional elements added after the attributes in the form.
+   * The new form elements can be wrapped in a FormGroup component for automatic formatting */
+  formAdditionalItems?: React.ReactNode;
   /** Attribute label for strings unassociated with one of the provided listed attributes */
   hasWordsAttrLabel?: string;
   /** Delimiter in the query string for pairing attributes with search values.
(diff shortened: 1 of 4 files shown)


redallen (dismissed):

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -29,6 +29,8 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>,
   value?: string;
   /** Array of attribute values */
   attributes?: string[] | SearchAttribute[];
+  /* Additional elements added after the attributes in the form */
```
nicolethoen, line 32: Could you add the suggestion to wrap additional form elements in a FormGroup component for formatting?

nicolethoen (approved):
We may also just want to update the advanced form unit test to include testing this new prop.

nicolethoen (commented):
Sorry, I meant to leave a comment, not approve quite yet. Wrong button.

packages/react-core/src/components/SearchInput/SearchInput.tsx
KKoukiou: Fixed, let me know if you have a better wording suggestion :)

nicolethoen: Looks good. Could you also update the advanced form unit test to include testing this new prop?

KKoukiou: definitely, forgot about the tests, sorry !

redallen (approved):

nicolethoen (approved):

Issue #6027, opened 2021-07-14
Charts: Victory tooltip performance update


dlabrecq:
Victory has a pending PR to improve tooltip performance. Currently, all tooltips are being re-rendered upon every interaction -- the fix only renders an active tooltip. For large datasets, there would likely be a significant gain in performance.

Would like to update Victory packages to the latest version when the fix below is merged. PatternFly is currently using version 35.4.4, while the latest is 35.9.0.

> Hello! I just wanted to provide an update here. We are doing some performance auditing and improvement right now. You can expect to see some small changes soon, and some plans for larger performance improvements in subsequent releases. Keep us posted as you observe changes in performance, or if you have any specific recommendations!

Related: 
`https://github.com/FormidableLabs/victory/pull/1901`
`https://github.com/FormidableLabs/victory/issues/1669`

PR #6049, opened 2021-07-17
Charts: Victory tooltip performance update

dlabrecq:
This updates Victory packages to the latest version, which includes a performance enhancement for tooltips.

Currently, all tooltips are being re-rendered upon every interaction -- the update only renders an active tooltip. For large datasets, there would likely be a significant gain in performance.

Closes https://github.com/patternfly/patternfly-react/issues/6027


ddonahue007 (approved):
LGTM

nicolethoen (approved):

redallen (approved):

Issue #5819, opened 2021-05-21
Form: field validation/password strength demo


mcarrano:
Add the ability to implement password strength validation on the Login page.

Core issue: https://github.com/patternfly/patternfly/issues/4018 

Design: https://marvelapp.com/prototype/6jehce7/screen/78511266 

--

This demo relies on https://github.com/patternfly/patternfly-react/issues/6002

nicolethoen:
We need to be sure this demo utilizes a 'live-region' correctly. May need to collaborate with @jessiehuff some.

tlabaj:
closed by #6053 6053

PR #6053, opened 2021-07-19
feat(Form): add form group label info & docs(Form): add password strength demo

### LINKED ISSUE #6002: Form - add info text to group label
https://github.com/patternfly/patternfly/pull/4172

Adds a new element `.pf-c-form__group-label-info` to the form component.

When a `.pf-c-form__group-label-info` is defined, we
* Add `.pf-m-info` to `.pf-c-form__group-label` for that form group
* Wrap `.pf-c-form__label` with `.pf-c-form__group-label-main`
* Place `.pf-c-form__group-label-info` after `.pf-c-form__group-label-main`

kmcfaul:
**What**: Closes #6002 & #5819

Adds `labelInfo` property to `FormGroup`.
Adds new unit test, example and integration test.
Also updates some duplicate ids in the examples.

Update: Adds password strength demo

### DIFF
--- a/packages/react-core/src/components/Form/FormHelperText.tsx
+++ b/packages/react-core/src/components/Form/FormHelperText.tsx
@@ -13,6 +13,8 @@ export interface FormHelperTextProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Icon displayed to the left of the helper text. */
   icon?: React.ReactNode;
+  /** Component type of the Form Helper Text */
+  component?: 'p' | 'div';
 }
 
 export const FormHelperText: React.FunctionComponent<FormHelperTextProps> = ({
@@ -21,19 +23,23 @@ export const FormHelperText: React.FunctionComponent<FormHelperTextProps> = ({
   isHidden = true,
   className = '',
   icon = null,
+  component = 'p',
   ...props
-}: FormHelperTextProps) => (
-  <p
-    className={css(
-      styles.formHelperText,
-      isError && styles.modifiers.error,
-      isHidden && styles.modifiers.hidden,
-      className
-    )}
-    {...props}
-  >
-    {icon && <span className={css(styles.formHelperTextIcon)}>{icon}</span>}
-    {children}
-  </p>
-);
+}: FormHelperTextProps) => {
+  const Component = component as any;
+  return (
+    <Component
+      className={css(
+        styles.formHelperText,
+        isError && styles.modifiers.error,
+        isHidden && styles.modifiers.hidden,
+        className
+      )}
+      {...props}
+    >
+      {icon && <span className={css(styles.formHelperTextIcon)}>{icon}</span>}
+      {children}
+    </Component>
+  );
+};
 FormHelperText.displayName = 'FormHelperText';
--- a/packages/react-core/src/components/Form/FormHelperText.tsx
+++ b/packages/react-core/src/components/Form/FormHelperText.tsx
@@ -13,7 +13,7 @@ export interface FormHelperTextProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Icon displayed to the left of the helper text. */
   icon?: React.ReactNode;
-  /** Component type of the Form Helper Text */
+  /** Component type of the form helper text */
   component?: 'p' | 'div';
 }
 
--- a/packages/react-core/src/demos/PasswordStrength.md
+++ b/packages/react-core/src/demos/PasswordStrength.md
@@ -12,6 +12,8 @@ import CheckCircleIcon from '@patternfly/react-icons/dist/js/icons/check-circle-
 
 ### Basic validation
 
+Note, the validation and password strength rules are only examples, demonstrating the changes in the UI when certain conditions are met. Currently the password strength is determined by how often validation rules are met.
+
 ```js
 import React from 'react';
 import {
--- a/packages/react-core/src/demos/PasswordStrength.md
+++ b/packages/react-core/src/demos/PasswordStrength.md
@@ -12,7 +12,7 @@ import CheckCircleIcon from '@patternfly/react-icons/dist/js/icons/check-circle-
 
 ### Basic validation
 
-Note, the validation and password strength rules are only examples, demonstrating the changes in the UI when certain conditions are met. Currently the password strength is determined by how often validation rules are met.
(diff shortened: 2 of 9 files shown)


kmcfaul:
Updated to include the password strength demo.

@mattnolting @mcoker I'm not sure why, but nesting a `HelperText` element inside `pf-c-form__group-label pf-m-info` overrides the color granted by the HelperText variant (error/warning/success), instead always showing black. Any thoughts?

@mcarrano Is there a set way to calculate password strength that we want for this demo? It was unclear in the design what the rules are for the strength of a password. Currently I have it counting how many times the 3 minimum required rules are being met and with more uppercase/special characters or digits the stronger the password becomes.

More of an open question to anyone: `FormHelperText` (`.pf-c-form__helper-text`) is a `<p>` element while the core demo uses a `<div>` instead. This is causing a console error at the moment because `HelperText` nests a `<ul>` beneath `FormHelperText`. Is there any reason I should not add a property to change the component type of `FormHelperText`?

mcarrano:
> @mcarrano Is there a set way to calculate password strength that we want for this demo? It was unclear in the design what the rules are for the strength of a password. Currently I have it counting how many times the 3 minimum required rules are being met and with more uppercase/special characters or digits the stronger the password becomes.

@kmcfaul I think what you are doing is probably fine to demo this. We don't have any particular standard for predicting password strength that we are recommending through PatternFly. @ajacobs21e is there any particular rules you would recommend applying for the purpose of this demo?

nicolethoen (approved):

kmcfaul:
Is it worth bringing in another library for just a demo? The code shows how the UI will change and users can implement their own validation/strength algorithm in place of the demo function.

mcarrano:
@kmcfaul I agree that it does not make sense to pull in an external library. Maybe there should just be mention that this demo is only to show how to reflect status in the UI, but does not intend to specify any particular validation technique.

@ajacobs21e do you think it would make sense to reflect the InfoSec recommendation in the design doc that relates to this or do you think PatternFly should just steer clear of that?

kmcfaul:
@mcarrano Updated the example docs to include a short note. Let me know if that looks good.

mcarrano:
@kmcfaul how about just adding one more sentence: "We expect consumers will substitute their own, more robust, validation algorithm."

ajacobs21e:
Yeah that would be great to mention their recommendation. thanks!

kmcfaul:
@mcarrano @ajacobs21e Updated the example disclaimer

mcarrano (approved):
This looks great. Thanks for adding the disclaimer @kmcfaul

packages/react-core/src/components/Form/FormHelperText.tsx
```diff
@@ -13,6 +13,8 @@ export interface FormHelperTextProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Icon displayed to the left of the helper text. */
   icon?: React.ReactNode;
+  /** Component type of the Form Helper Text */
```
tlabaj, line 16: nit.  sentence text here.

packages/react-core/src/demos/PasswordStrength.md
```diff
@@ -0,0 +1,170 @@
...
+        return;
+      }
+
+      if (password.length < 14) {
+        this.setState({ ruleLength: 'error' });
+      } else {
+        this.setState({ ruleLength: 'success' });
+      }
+
+      if (/redhat/gi.test(password) || /red/gi.test(password) || /hat/gi.test(password) || /rh/gi.test(password)) {
```
mcoker, line 65: @mcarrano fwiw, the validation here checks for "redhat", "red", "hat", and "rh". So the password validation fails with words that contain those strings, which seems confusing.

Screen Shot 2021-07-29 at 4 03 25 PM
Screen Shot 2021-07-29 at 4 03 16 PM
Screen Shot 2021-07-29 at 4 05 23 PM

Any problem with updating the helper text to say `Cannot contain the word "redhat"` instead of `Cannot contain any variation of the word "redhat"`?

mcoker (commented):
LGTM! Just one comment about the validation text.

packages/react-core/src/demos/PasswordStrength.md
kmcfaul: @mcarrano Updated the second validation rule to 'cannot contain the word redhat' instead of variations. Let me know if that's okay or if I should change it back.

mcoker (dismissed):

tlabaj (dismissed):

packages/react-core/src/demos/PasswordStrength.md
mcarrano: Yes, this is fine @kmcfaul . I just noticed a typo in your helper text string, however. It reads, "Cannot contain any the word "redhat" Please remove the work "any." Thanks.

nicolethoen (dismissed):

dlabrecq (dismissed):

mcarrano (dismissed):
Thanks @kmcfaul !

jessiehuff (commented):
I think this makes sense for mouse and keyboard users, but I think that we should follow the same pattern that we have for our [validated form example](https://www.patternfly.org/v4/components/form#validated), particularly for screen reader users.

The input should have an `aria-describedby` attribute that points to the helper text, and it should be `aria-invalid` when validation isn't met. It looks like your example has that, but the `aria-describedby` points to an id that doesn't exist, and it also looks like the `aria-invalid` never updates. Then the helper text should be inside of an `aria-live="polite"` to let the screen reader know that the text will change/update.

kmcfaul:
@jessiehuff Updated with your feedback. Let me know if the aria-live region is updating properly (I wrapped the HelperText with the div).

tlabaj (approved):

jessiehuff (commented):
I tested in VO, JAWS, and NVDA, and I'm seeing kind of unreliable results. I was only able to get the helper text to announce once in VO. When I compare the demo to our form invalid example, I think the things that might be affecting it are that `aria-invalid` never updates (it's always "false"), and the id that the `aria-describedby` points to is on a different element than the one with the `aria-live`. Perhaps that's causing the issue? I don't want to hold up the release though for this so we could create an issue to refine this more. At least the helper text seems discoverable in the normal DOM order on its own.

kmcfaul:
Pushed one last update moving the aria-live and fixing the aria-invalid not being updated. Hopefully that puts the PR in a decent state to merge, we can open a followup to better fit the screenreaders if needed

tlabaj (approved):

nicolethoen (approved):

jessiehuff (approved):
LGTM! Great job! 😄

mcoker (approved):

PR #6055, opened 2021-07-20
fix(Checkbox): fix warning about ouiaSafe property

KKoukiou:
This slipped in https://github.com/patternfly/patternfly-react/pull/5949

### DIFF
--- a/packages/react-core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/react-core/src/components/Checkbox/Checkbox.tsx
@@ -73,6 +73,7 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
       description,
       body,
       ouiaId,
+      ouiaSafe,
       ...props
     } = this.props;
     if (!props.id) {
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -29,8 +29,10 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>,
   value?: string;
   /** Array of attribute values */
   attributes?: string[] | SearchAttribute[];
+  /* Additional elements added after the attributes in the form */
```
redallen, line 32: Can you revert the SearchInput changes? They're not quite the same as #6043

redallen (changes requested):

packages/react-core/src/components/SearchInput/SearchInput.tsx
KKoukiou: ups sorry, did not realize by branch was dirty,

redallen (approved):

kmcfaul (approved):

PR #6064, opened 2021-07-23
feat(LogViewer): Add demo with new controls

### LINKED ISSUE #5803: LogViewer: make full screen variant
Follow up work for https://github.com/patternfly/patternfly/issues/4082

### LINKED ISSUE #5804: LogViewer: Build a demo with working toolbar
1. Move the log viewer toolbar demo out from Toolbar
2. implement a LogViewer demo with a working Toolbar to match the Openshift design

nicolethoen:
Closes #5804 and closes #5803

Adds a new React demo to Log Viewer.

Makes adjustments to allow for/demonstrate

- tailing log
- downloading logs
- full screen log viewer
- switching data sources


mcoker (changes requested):
Looks like `requestFullscreen()` isn't supported in Safari. From https://j.hn/native-fullscreen-javascript-api-plus-jquery-plugin/, it looks like Safari needs `webkitRequestFullScreen()`. I verified this works in https://codepen.io/mcoker/pen/rNyjwqo - you'll need to comment `requestFullscreen()` and uncomment `webkitRequestFullScreen()`

Also this is probably out of scope, but noticed if you put 3 characters in the search box, it will show a badge of "1/1" even though there are no matches.

Screen Shot 2021-07-23 at 10 40 26 AM

packages/react-log-viewer/src/LogViewer/examples/realTestData.js
```diff
@@ -1,6 +1,5 @@
 export const data = {
   data: `Copying system trust bundle
-
```
mfocko, line 3: I have added that empty line in #6033 as an easy way to spot regression :)

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
```diff
@@ -12,12 +12,12 @@ export const isArrayOfString = (array: string[]) => {
   It should take an array, and return an array of indexes where the searchedInput is found throughout the data array. 
   Should always be searching an array of strings. Look into lazy log for ideas.
 */
-export const searchForKeyword = (searchedInput: string, parsedData: string[]) => {
+export const searchForKeyword = (searchedInput: string, parsedData: string[], itemCount: number) => {
   const searchResults: number[] = [];
 
   parsedData.map((row, index) => {
     const regex = new RegExp(searchedInput, 'gi');
```
mfocko, line 19: I would probably take it out of the closure, I'm not sure if it gets optimized.

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -108,40 +114,63 @@ export const LogViewer: React.FunctionComponent<LogViewerProps> = memo(
...
         }
       }
 
       if (adjustedSearchedInput !== '' && adjustedSearchedInput.length < 3) {
-        setRowInFocus(DEFAULT_FOCUS);
+        if (scrollToRow) {
+          setRowInFocus(scrollToRow);
+        } else {
+          setRowInFocus(DEFAULT_FOCUS);
+        }
```
mfocko, line 159: Isn't this possible? 🤔 Can't find `DEFAULT_FOCUS` definition, so I'm not sure.

```suggestion
        setRowInFocus(scrollToRow || DEFAULT_FOCUS);
```

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
nicolethoen: I'm not sure what you mean by this

packages/react-log-viewer/src/LogViewer/examples/realTestData.js
nicolethoen: okay. I moved it down a few lines so there isn't a blank space on the loading view. I don't want that to confuse people.

packages/react-log-viewer/src/LogViewer/utils/utils.tsx
mfocko: iirc building the regex pattern takes some time, so if you use the same pattern it makes sense to have it shared across all iterations of the `map`, like the `searchResults`.

By optimalization I meant that with compiled languages the compiler can optimize it like I mentioned above.

mcarrano (changes requested):
@nicolethoen this is great. Can I just ask that you update the toolbar UI to more closely reflect the designs here? https://marvelapp.com/prototype/bia3j2d  Let me know if you have questions.

mcarrano:
@nicolethoen Looks great. Thanks for making these changes. There was just one issue that I found. When I expand to full screen, I would expect that clicking the Expand icon again will take me back, but it does not. The only way to get out of full screen was to press the ESC key.

dlabrecq (approved):

dlabrecq:
The sandbox isn't loading for me, but otherwise looks good.

Screen Shot 2021-07-28 at 10 23 07 AM

nicolethoen:
I will open an issue to fix the codesandbox examples. The code sandbox needs to be given the javascript files with the large data. Right now it's not included. But i think that will require some investigation.

kmcfaul (dismissed):
Looks great!

mcarrano (approved):
Looks great. Thanks @nicolethoen

mcoker (approved):
LGTM! I don't know if this is by design or not but it seems to skip some string matches in the search box.

Screen Shot 2021-07-30 at 1 28 22 PM

jessiehuff (commented):
I think it would make sense to make the container for the log viewer focusable so that the user can scroll through the logs with up/down arrow keys (just needs `tabIndex="0"`). While it doesn't look like there's anything interactive in the log viewer, they should probably still have access to information after it's left view. I also can't seem to get into the dropdown in the toolbar with keyboard? 

I question whether we would want this as a live region for screen reader users, but I think that could be debatable (i.e. do we want this information to interrupt or interject into a user's current workflow like alerts?)

tlabaj (approved):

mcoker (approved):
🥳

jessiehuff (commented):
I was thinking that we could put the tabIndex on the entire container for scrolling. Here we have each of the logs focusable which I think would make sense if the contents are going to be interactive. Do you know what the expected use case for this will be? Will these logs ever be interactive for users? If they'll always just be static info, I think we could just keep the tabIndex on the outer container. Also, I still can't seem to get into the dropdown (the one listed as "container-1"). Everything else looks awesome though! 😄

mcarrano:
I can't really think of a use case for interacting with individual log entries. Typically this is just a stream of data that is chunked into time stamped entries.

kmcfaul (dismissed):

nicolethoen:
@jessiehuff To track updating the scroll behavior with the keyboard I opened this issue: https://github.com/patternfly/patternfly-react/issues/6110 
Would it be better to keep it with the updates I have for now, so that the logviewer can be controlled with the keyboard, even if it's not ultimately the solution we want? Or is it better to revert?

jessiehuff (approved):
Looks awesome!!! 😄

tlabaj (approved):

Issue #6056, opened 2021-07-20
demos: Create label from text demo


kmcfaul:
Add a demo showcasing how to create `Label` components from text in a modal (follows design from https://github.com/patternfly/patternfly-design/issues/753).

mcarrano:
@tlabaj @kmcfaul I am going to close this with the assumption that I will open a new issue once the functionality proceeds through design and core. We are currently close to having finalized designs.

PR #6065, opened 2021-07-26
feat(Label): add editable label

### LINKED ISSUE #5875: Editable label group
Implement an editable label group. Core issue and requirements: https://github.com/patternfly/patternfly/issues/3741

Allows user to click on a label to enable it for editing. Click enter to commit changes.

kmcfaul:
**What**: Closes #5875

Label:
- Adds `isEditable`, `onEditComplete`, `onEditCancel`, `editableProps`
- `onEditComplete` is triggered when `Enter`, `Tab` is pressed while on an active editable div, or when the mouse clicks outside of an active editable div
- `onEditCancel` is triggered when `Escape` is pressed while on on active editable div. This will also reset the editable div text to the initially passed children. The callback also passes back the previous text as a property (but it isn't be required to set the text state in this callback). 
- Optionally, passing `onInput` and/or `onBlur` to `editableProps` can let users hook directly into the changing text div

LabelGroup:
- Adds `isEditable`, `hasEditableTextArea`, `editableTextAreaProps`
- `hasEditableTextArea` will append a text area to the end of the label group list. This is in preparation for a future demo featuring creating a label from text.

### DIFF
--- a/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
+++ b/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
@@ -170,7 +170,7 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
                 </Label>
               </li>
             )}
-            {hasEditableTextArea && (
+            {isEditable && hasEditableTextArea && (
               <li className={css(styles.labelGroupListItem, styles.modifiers.textarea)}>
                 <textarea className={css(styles.labelGroupTextarea)} rows={1} tabIndex={0} {...editableTextAreaProps} />
               </li>
--- a/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
+++ b/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
@@ -35,6 +35,12 @@ export interface LabelGroupProps extends React.HTMLProps<HTMLUListElement> {
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Flag to implement a vertical layout */
   isVertical?: boolean;
+  /** Flag indicating contained labels are editable. Allows spacing for a text input after the labels. */
+  isEditable?: boolean;
+  /** Flag indicating the editable label group should be appended with a textarea. */
+  hasEditableTextArea?: boolean;
+  /** Additional props passed to the editable textarea. */
+  editableTextAreaProps?: any;
 }
 
 interface LabelGroupState {
@@ -65,7 +71,9 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
     closeBtnAriaLabel: 'Close label group',
     tooltipPosition: 'top',
     'aria-label': 'Label group category',
-    isVertical: false
+    isVertical: false,
+    isEditable: false,
+    hasEditableTextArea: false
   };
 
   componentDidMount() {
@@ -123,6 +131,9 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
       defaultIsOpen,
       tooltipPosition,
       isVertical,
+      isEditable,
+      hasEditableTextArea,
+      editableTextAreaProps,
       /* eslint-enable @typescript-eslint/no-unused-vars */
       ...rest
     } = this.props;
@@ -159,6 +170,11 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
                 </Label>
               </li>
             )}
+            {hasEditableTextArea && (
+              <li className={css(styles.labelGroupListItem, styles.modifiers.textarea)}>
+                <textarea className={css(styles.labelGroupTextarea)} rows={1} tabIndex={0} {...editableTextAreaProps} />
+              </li>
+            )}
           </ul>
         </React.Fragment>
       );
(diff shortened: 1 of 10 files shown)


nicolethoen (commented):
I'm noticing that clicking the close button puts the label in edit mode. I see that you are not creating a close function, but I just dont think it should toggle the edit state.

kmcfaul:
@nicolethoen Updated so that events coming from a button within the label should not trigger the active edit state.

kmcfaul:
Opened a followup for an editable label with icon interactions

nicolethoen (dismissed):

mcarrano (commented):
This looks really great @kmcfaul . I only have one issue/question. It looks like if I click outside of an existing label, a text field will be opened where I can enter a new label. But right now I can enter that text but not do anything with it. I saw your note about that being follow-up functionality, but should we enable that at all now if it only partially works? I'm afraid someone will think that's a bug. Thoughts?

@bgliwa01 can you also take a look and see if this works as you expect?

kmcfaul:
I can remove the text field from the label group in the example, would that work for now?

mcarrano:
@kmcfaul Yes, I think so. Do others agree? So for this release, we would say this supports edit-in-place by clicking on an existing label, but you cannot yet add labels from here.

mcarrano (approved):
Thanks @kmcfaul . Looks great. Will the issue you opened (https://github.com/patternfly/patternfly-react/issues/6088) cover the use case to create a label in place or do we need another one? I will get this into the upcoming sprint.

kmcfaul:
@mcarrano I created #6056 earlier in the sprint to cover the create label demo, which I think shouldn't require any updates to the component

jessiehuff (commented):
So this looks pretty good to me! I think the big question that I always have is just how will users know to interact with this label? I tested with the different screen readers, and I can interact with the label. They typically say that it's editable text which is nice. 

Can you make the `aria-label` something a little more readable? Right now it looks like it's "label-close-button" which in a screen reader, it sounds odd haha. Can we just make this something like "Close {label name}"?

The only other weirdness is that when using a screen reader, the user has to press enter twice to enter into the editable field. If I try editing after one enter, it actually starts typing into the preview/org's search input at the top of the page.

In general though, my biggest concern with all of it is just how do users know what to expect with this interaction pattern- for mouse, keyboard, and screen reader users (it says that it's editable for screen readers, but I'm not sure that they would think to press enter twice)? I don't want to hold this PR up so perhaps we can look into this with a further issue or more discussion down the line?

jessiehuff:
Oof, just noticed that that `aria-label` is currently on all of the labels and not related to your PR. I guess a simple but unrelated issue haha.

packages/react-core/src/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -104,6 +128,81 @@ VerticalCategoryLabelGroupOverflowRemovable = () => {
...
+          onEditComplete={newText => this.onEditComplete(newText, 'label1')}
+          isEditable
+          editableProps={{
+            'aria-label': 'Editable text',
+            id: 'editable-label-1'
+          }}
+        >
+          {this.state.label1}
+        </Label>
+        <Label color="green">Static label</Label>
```
tlabaj, line 177: Is there a reason to make this a different color? At first glance someone may think the color indicates whether the label is editable or static.

packages/react-core/src/components/LabelGroup/LabelGroup.tsx
```diff
@@ -159,6 +170,11 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
                 </Label>
               </li>
             )}
+            {hasEditableTextArea && (
```
tlabaj, line 173: should we also check for isEditable?

mattnolting (dismissed):
This looks great @kmcfaul! Now that I see the editable label in action, I wonder if we should update the presentation of the editable state, but that's a question for design @mcarrano @mceledonia and should be addressed upstream. LPTM!

mcarrano:
> So this looks pretty good to me! I think the big question that I always have is just how will users know to interact with this label?

I think that @jessiehuff and @mattnolting bring up a good question about whether there should be some affordance to indicate when a label is editable. I've opened a design issue to explore this further. But I think this is OK for now.

packages/react-core/src/components/LabelGroup/examples/LabelGroup.md
kmcfaul: I was matching core's example, should I change this to blue?

kmcfaul:
> The only other weirdness is that when using a screen reader, the user has to press enter twice to enter into the editable field. If I try editing after one enter, it actually starts typing into the preview/org's search input at the top of the page.

Hmm I'm not seeing this myself on the surge, using only the keyboard. Enter should make the field editable and set the typing cursor to the end of the text (maybe the selection range logic is the issue?). Is there any reason a screenreader would change keyboard triggers?

packages/react-core/src/components/LabelGroup/examples/LabelGroup.md
mcarrano: It didn't bother me that these are different. It's true that the color is not meaningful here, but it might be a valid way to differentiate them.

tlabaj (approved):

mcarrano (approved):
This looks good for now with the expectation that we will do more work to address affordances and accessibility issues.

nicolethoen (approved):

mattnolting (approved):

dlabrecq (approved):

jessiehuff (approved):
LGTM! Created the follow-up screen reader issue [here](https://github.com/patternfly/patternfly-react/issues/6096).

PR #6072, opened 2021-07-27
fix(VerticalTabs): expose href to allow links

### LINKED ISSUE #5957: Expose `href` prop in VerticalTabs component
VerticalTabs component should expose the `href` prop on the component. Since `a` tag is being used and [href](https://github.com/patternfly/patternfly-react/blob/master/packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx#L50) value is hardcoded to `#`. Open link in new tab doesn't work properly.

1. Right-click on the element that is using VerticalTabs Component.
2. Click on `Open link in new Tab`
3. Notice that it takes to `<url>/#` instead of the desired URL which is being set through `onClick` Handler

Bug

Openshift DevConsole

CooperRedhat:
**What**: Eventually closes #5957, some additional changes may be necessary.

**Additional issues**: Solves the issue being addressed. However, when a tab is also a link, there is no additional styling applied. Seems there should be some indication.

### DIFF
--- a/packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx
+++ b/packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx
@@ -18,6 +18,8 @@ export interface VerticalTabsTabProps extends Omit<React.HTMLProps<HTMLLIElement
   shown?: boolean;
   /** Callback function when the tab is activated */
   onActivate?: () => void;
+  /** HREF location */
+  href?: string;
 }
 
 export const VerticalTabsTab: React.FunctionComponent<VerticalTabsTabProps> = ({
@@ -29,6 +31,7 @@ export const VerticalTabsTab: React.FunctionComponent<VerticalTabsTabProps> = ({
   hasActiveDescendant = false,
   shown = false,
   onActivate = null,
+  href,
   ...props
 }: VerticalTabsTabProps) => {
   const classes = css('vertical-tabs-pf-tab', { active, 'active-descendant': hasActiveDescendant, shown }, className);
@@ -47,7 +50,7 @@ export const VerticalTabsTab: React.FunctionComponent<VerticalTabsTabProps> = ({
 
   return (
     <li className={classes} {...props}>
-      <a className={linkClasses} onClick={e => handleActivate(e)} href="#">
+      <a className={linkClasses} onClick={e => handleActivate(e)} href={href}>
         {title}
       </a>
       {children}
--- a/packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx
+++ b/packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
+import textUtilities from '@patternfly/react-styles/css/utilities/Text/text';
+import ExternalLinkAltIcon from '@patternfly/react-icons/dist/js/icons/external-link-alt-icon';
 
 export interface VerticalTabsTabProps extends Omit<React.HTMLProps<HTMLLIElement>, 'title'> {
   /** Child tab nodes (VerticalTabsTab's) */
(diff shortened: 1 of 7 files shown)


CooperRedhat:
image
https://patternfly-react-pr-6072.surge.sh/extensions/catalog-view-vertical-tabs
@mcarrano I added the functionality to add links to the `VerticalTab`s but it seems like some styling should be applied to indicate that to the user. Thoughts?

mcarrano:
@CooperRedhat I think that makes sense, but I need to understand more about this issue before I can make a recommendation. Is this a change to vertical tabs as shown here: https://patternfly-react-pr-6072.surge.sh/components/tabs#vertical or something to do with the Catalog View extension?

CooperRedhat:
@mcarrano It is for the Catalog View extension.
https://patternfly-react-pr-6072.surge.sh/extensions/catalog-view-vertical-tabs/

mcarrano:
@CooperRedhat looks like this is currently not using any standard PatternFly menu component. I'd like to see this design get refreshed at some point to be more aligned with PatternFly 4. That said, if the intent is to add external links to the sidebar, then our typical affordance for that is to add the external link icon to the right of the label. See an example here: https://www.patternfly.org/v4/components/button/design-guidelines#link-button

redallen (approved):

packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx
```diff
@@ -29,14 +35,19 @@ export const VerticalTabsTab: React.FunctionComponent<VerticalTabsTabProps> = ({
...
-  const linkClasses = css({
-    'no-wrap': wrapStyle === 'nowrap',
-    truncate: wrapStyle === 'truncate'
-  });
+  const linkClasses = css(
+    {
+      'no-wrap': wrapStyle === 'nowrap',
+      truncate: wrapStyle === 'truncate'
+    },
+    href && textUtilities.linkColor
```
mcoker, line 49: This update should probably be made in the extension CSS. Our utilities add `!important` to all of the rules, which is a pretty bad practice for regular component styles - those are for users to use if they prefer styling that way.

From the looks of it, maybe this could be solved with attribute selectors? Do these examples help - https://codepen.io/mcoker/pen/wvdmVxB

mcoker (changes requested):

packages/react-catalog-view-extension/src/components/VerticalTabs/VerticalTabsTab.tsx
CooperRedhat: I dropped the idea of enforcing styling for now. See comment below

redallen (approved):

CooperRedhat:
Removed enforced styling when Tabs are also links. Original request was for the ability to right-click to open a link in a new tab, that remains. 

If the consumer has a need for special styling or even external link icons, the `title` attribute can take a React node, where those can be applied.

Will open another issue referencing this one to make sure this and others are updated to Patternfly 4, and we can revisit the issue of enforcing styling on specific types of `Tab`s

mcoker (approved):

tlabaj (commented):
Looks good Cooper!  Can you please add the prop to the demo app and add an integration test to verify.

packages/react-catalog-view-extension/src/components/VerticalTabs/examples/VerticalTabs.md
```diff
@@ -61,9 +61,9 @@ class MockVerticalTabsExample extends React.Component {
         {renderTab('all', 'All', null, { shown: true })}
         {renderTab('one', 'Tab One', [
           renderTab('one-one', 'Tab One-One', [
-            renderTab('one-one-one', 'Tab One-One-One'),
-            renderTab('one-one-two', 'Tab One-One-Two'),
-            renderTab('one-one-three', 'Tab One-One-Three')
+            renderTab('one-one-one', 'About Modal', null, {href: "../components/about-modal"}),
+            renderTab('one-one-two', 'Accordion', null, {href: "../components/accordion"}),
+            renderTab('one-one-three', 'Redhat', null, {href: "http://www.redhat.com"})
```
CooperRedhat, line 66: @tlabaj Is this what you mean by adding the prop to the demo app?

redallen: I think she means add it under `packages/react-integration/demo-app-ts` and write a test under `packages/react-integration/cypress/integration` that clicks the link and makes sure the browser points to the specified `href`.

redallen (approved):

nicolethoen (approved):

PR #6074, opened 2021-07-27
[WIP] add experimental drag and drop demo

### LINKED ISSUE #6020: Spike: Draggable data list demos
Could we use a third party library to implement a draggable data list and build a demo to demonstrate?
Ideally, we are not building our own draggable implementation, and we are not building the Data List component to have a 3rd party library built in.

redallen:
**What**: Closes #6020

**Additional issues**: I like the API and smoothness of `react-beautiful-dnd` but it seems rather heavyweight for what it does:

- It's 105kB and includes an entire copy of redux
- It adds a ton of non-patternfly attributes and style tags to the DOM
- It requires every draggable component expose refs (few components currently do)
- It doesn't support dragging handles
- It doesn't support left-behind ghost rows per core design
- It requires additional work to render server side (which is why the PR fails to build)

### DIFF
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -38,6 +38,7 @@
     "@patternfly/react-styles": "^4.11.4",
     "@patternfly/react-tokens": "^4.12.5",
     "focus-trap": "6.2.2",
+    "react-beautiful-dnd": "13.1.0",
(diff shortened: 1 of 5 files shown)


redallen:
Closing in favor of #6258

PR #6078, opened 2021-07-28
feat(ExpandableSection): Added props to support disclosure variation

### LINKED ISSUE #5931: Expandable section: Add Disclosure panel variant
Adds a new Disclosure variant to the Expandable section component. This differs from the default by updating styles but should be no change in behavior. See core issue here: https://github.com/patternfly/patternfly/issues/3952

tlabaj:
**What**: Closes #5931


nicolethoen (dismissed):

mcarrano (commented):
This looks good @tlabaj . I only have one question. Looks like this style is being set through the property `isDisplayLarge` is that right? Curious why that naming since the example is called Disclosure. Should we at least document that DisplayLarge and Disclosure mean the same thing? Let me know your thoughts.

mcoker (dismissed):
LGTM!

tlabaj:
> This looks good @tlabaj . I only have one question. Looks like this style is being set through the property `isDisplayLarge` is that right? Curious why that naming since the example is called Disclosure. Should we at least document that DisplayLarge and Disclosure mean the same thing? Let me know your thoughts.
@mcarrano  I think calling the prop `disclosure` is very specific to how the consumer intends to use the component.  If someone external was using the component library, that might have no have meaning to them.  I would actually prefer to rename the example to "Display large" or something more indicative of what it is showing.
cc @mcoker

mcarrano:
@tlabaj I get your point, but I would like to keep the example named Disclosure as that was something specifically requested to sync our library with PatternFly Elements. I'm OK with keeping the prop name as is, however. Could we just add a mention in the documentation that this is set to create the Disclosure styling, or something like that?

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
```diff
@@ -109,6 +117,8 @@ export class ExpandableSection extends React.Component<ExpandableSectionProps, E
           propOrStateIsExpanded && styles.modifiers.expanded,
           isActive && styles.modifiers.active,
           isDetached && styles.modifiers.detached,
+          isDisplayLarge && styles.modifiers.displayLg,
```
kmcfaul, line 120: Everything looks good to me, but I'm wondering if we will ever add different 'display size' type props in the future? If we may plan to do that a `displaySize` prop with an enum may be more future proof

kmcfaul (approved):

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
mcoker: @mcarrano wdyt about that - we use `display-large` as the name for the marketing/web variations of components. Do you anticipate we would have more variations of `display-[whatever]` in the future?

mcarrano:
I guess my only complaint with just calling this "large" is that in this case we've done more than just increase the font size and spacing. It's actually more like an expandable panel with the drop shadow and all.

packages/react-core/src/components/ExpandableSection/ExpandableSection.tsx
tlabaj: Good question @kmcfaul .

mcarrano: I guess if past history is a predictor, the only thing I could possibly think of is that we would get a request for a compact version with smaller text and tighter spacing.

mcoker: Those are typically called `pf-m-compact`, considering they're a regular variation of a component. A little background, one of the reasons we went with `.pf-m-display-large` was that we could have both a `.pf-m-large` variation as a regular modifier for a component intended for product use, and a `.pf-m-display-large` variation specifically for web/marketing use.

tlabaj: I will make the prop a variant. I can do the same for the accordion PR.

mcarrano (approved):
Looks good. Thanks @tlabaj .

mcoker (approved):

dlabrecq (approved):

kmcfaul (approved):

Issue #6082, opened 2021-07-28
Donut chart shows gap


dlabrecq:
The donut chart shows a gap when there is only one data series with a value greater than zero.

In order to show a white separator between pie segments, the PatternFly theme overrides Victory's `padAngle` property.
Even though a particular value may be zero, Victory will attempt to render the data series. This is why there is still white separator and an available tooltip for the data series with zero values.

As a workaround, users may override the pie chart's `padAngle` prop and set it to zero. Alternatively, the data series with zero values may be omitted from the pie chart.

That said, it may be possible to modify the pie chart to test if there is more than one data series with a value greater than zero before applying the `padAngle`.

![chrome-capture](https://user-images.githubusercontent.com/17481322/127392336-3d54bc14-4cf2-433f-823e-1620e9e9e548.gif)

PR #6084, opened 2021-07-28
fix(charts): donut chart shows gap

dlabrecq:
The donut chart currently shows a gap when there is only one data series with a value greater than zero.

In order to show a white separator between pie segments, the PatternFly theme overrides Victory's `padAngle` property. This change ensures `padAngle` is only applied to a given data series with a non-zero value.

Closes https://github.com/patternfly/patternfly-react/issues/6082

fixed


ddonahue007 (approved):
LGTM

redallen (approved):

PR #6085, opened 2021-07-29
feat(Accordion): Added support for bordered, display large …

### LINKED ISSUE #5498: Add new variant of Accordion for marketing use cases
Want to add a new variant of the Accordion component to support marketing use cases. This includes a display-lg variation, bordered variation, and support for multiple item bodies.

Details in core issue here: https://github.com/patternfly/patternfly/issues/3865

PR - https://github.com/patternfly/patternfly/pull/3888

tlabaj:
…and multiple body content

**What**: Closes #5498

### DIFF
--- a/packages/react-core/src/components/Accordion/AccordionContent.tsx
+++ b/packages/react-core/src/components/Accordion/AccordionContent.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Accordion/accordion';
 import { AccordionContext } from './AccordionContext';
+import { AccordionExpandedContentBody } from './AccordionExpandedContentBody';
 
 export interface AccordionContentProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside the Accordion  */
@@ -18,6 +19,8 @@ export interface AccordionContentProps extends React.HTMLProps<HTMLDivElement> {
   'aria-label'?: string;
   /** Component to use as content container */
   component?: React.ElementType;
+  /** Flag indicating content is custom. Expanded content Body wrapper will be removed from children.  This allows multiple bodies to be rendered as content. */
+  isCustomContent?: React.ReactNode;
 }
 
 export const AccordionContent: React.FunctionComponent<AccordionContentProps> = ({
@@ -26,6 +29,7 @@ export const AccordionContent: React.FunctionComponent<AccordionContentProps> = 
   id = '',
   isHidden = false,
   isFixed = false,
+  isCustomContent = false,
   'aria-label': ariaLabel = '',
   component,
   ...props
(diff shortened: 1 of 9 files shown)


mcarrano (approved):
Looks good to me. Thanks @tlabaj !

packages/react-core/src/components/Accordion/AccordionContent.tsx
```diff
@@ -18,6 +19,8 @@ export interface AccordionContentProps extends React.HTMLProps<HTMLDivElement> {
   'aria-label'?: string;
   /** Component to use as content container */
   component?: React.ElementType;
+  /** Flag indicating content is custom. Expanded content Body wrapper will be removed from children.  This allows multiple bodies to be rendered as content. */
+  isCustomContent?: React.ReactNode;
```
nicolethoen, line 23: This is an interesting way to handle this. I think we don't usually conditionally expect the consumer to wrap elements in a body component. I wonder if it's be more consistent if we allowed people to just pass a single child, or an array of children? and we wrap things in body components for them?

tlabaj: @nicole I was debating both approaches.   I initially stared with the single Node or array.  I  was thinking in the long run it would be nice to handle this similar to other components with bodies (Card, Drawer) where the consumer composes it themself,  that is why I took this approach.

mcoker (approved):
LGTM!!

packages/react-core/src/components/Accordion/AccordionContent.tsx
nicolethoen: Would it be better then, to always tell the consumer to wrap content in a body?

kmcfaul: Maybe we could update to always wrapping the content in a body in the next breaking release?

tlabaj: @nicolethoen that would break consumers who are already using the component.
@kmcfaul that is what I was thinking long term.

I am ok with either approach here... Like I said, I was debating which one to go with.

nicolethoen: I think either way will work, as long as it's clear.

packages/react-core/src/components/Accordion/examples/Accordion.md
```diff
@@ -5,42 +5,45 @@ cssPrefix: pf-c-accordion
 propComponents: ['Accordion', 'AccordionItem', 'AccordionContent', 'AccordionToggle']
```
nicolethoen, line 5: Could you add `AccordionExpandedContentBody` here

packages/react-core/src/components/Accordion/AccordionContent.tsx
tlabaj: @kmcfaul what do you think?

kmcfaul: I think having the custom prop is good until we can refactor 👍

mcarrano (approved):
Looks good!

dlabrecq (approved):

kmcfaul (approved):

PR #6108, opened 2021-08-02
Improve IAction type

josejulio:
`IAction` type is hard to use if you haven't seen the code (I don't think I found a lot of documentation regarding it as well).

Reason is that it cover 4 different things:
- An action Separator, where the `isSeparator` property is always true. 
- DropdownItem
- Button
- Custom component

I tried making it easier to consume by split into 4 different types (with the main features of each one) so that the types don't mix and is clear what properties will be used.

- `IActionSeparator` where `isSeparator` is *always* true (always false or undefined on the other types). 
- `IActionDropdownItem` where the `title` is a `ReactNode`, has the `onClick` and any property from `DropdownItemProps`
- `IActionOutsideItemButton` where the `title` is a string, has the `onClick`, `isOutsideDropdown` is true (false and undefined on others) and any property from `ButtonProps`
- `IActionOutsideItemCustom` `title` is a `ReactNode` minus `string` (to avoid overlap with previos) , has `onClick` and `isOutsideDropdown` as true and anything else you want (This could probably be worked to be a generic). 

All of these together are the `IAction` type.

I inspected the code and the usage seems more or less like that, there where a couple of castings as `any` that I'm removing now with this structure.

### DIFF
--- a/packages/react-table/src/components/Table/ActionsColumn.tsx
+++ b/packages/react-table/src/components/Table/ActionsColumn.tsx
@@ -10,7 +10,14 @@ import {
   DropdownPosition
 } from '@patternfly/react-core/dist/esm/components/Dropdown/dropdownConstants';
 
-import { IAction, IExtraData, IRowData } from './TableTypes';
+import {
+  IAction,
+  IActionDropdownItem,
+  IActionOutsideItemButton,
+  IActionOutsideItemCustom,
+  IExtraData,
+  IRowData
+} from './TableTypes';
 
 export interface CustomActionsToggleProps {
   onToggle: (isOpen: boolean) => void;
@@ -82,10 +89,16 @@ export class ActionsColumn extends React.Component<ActionsColumnProps, ActionsCo
       <KebabToggle isDisabled={isDisabled} onToggle={this.onToggle} />
     );
 
+    const outsideDropdownItems = items.filter(a => a.isOutsideDropdown) as (
+      | IActionOutsideItemButton
+      | IActionOutsideItemCustom
+    )[];
+
+    const dropdownItems = items.filter(a => !a.isOutsideDropdown) as IActionDropdownItem[];
+
     return (
       <React.Fragment>
-        {items
-          .filter(item => item.isOutsideDropdown)
+        {outsideDropdownItems
           // eslint-disable-next-line @typescript-eslint/no-unused-vars
           .map(({ title, itemKey, onClick, isOutsideDropdown, ...props }, key) =>
             typeof title === 'string' ? (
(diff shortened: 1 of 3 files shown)


redallen (approved):
Clean!

kmcfaul (approved):
lgtm

nicolethoen (commented):
Okay, this looks great! I'm trying to process all these updates and have two questions/requests:

1. Anyone currently using the `actions` prop on the Table or the `actions` prop on the TableComposable are not going to have conflicting types after this update. We cannot incorporate these changes if they will break any existing Table implementations.
2. Could you add the definition of `IActions` and the four others (`IActionSeparator`, `IActionDropdownItem`, `IActionOutsideItemButton`, `IActionOutsideItemCustom`) in the notes in the markdown above the `Composable: Actions` example on the `Composable Table page and the `Actions and first cell in body rows as th` example on the `Legacy Table` page. If it's easier, you could just add a link to the github page with the type definitions in the examples so they can look it all up themselves.

josejulio:
> 1. Anyone currently using the `actions` prop on the Table or the `actions` prop on the TableComposable are not going to have conflicting types after this update. We cannot incorporate these changes if they will break any existing Table implementations.

Let me think more about this, but on first though, typescript users using the action in a unsupported way will get conflicts. i.e. users that are using a separator and assigning more variables.
```ts
{
  isSeparator: true,
  class: 'foo' // <--- this class won't really pass to the separator 
}
```

Other things like mixing button and outside item props, the custom props of an outside item won't work on a button. I'll try to get more examples, but I had this issue. I didn't know why the props I was passing didn't work and I had to inspect the code to realize how this works.

> 2. Could you add the definition of `IActions` and the four others (`IActionSeparator`, `IActionDropdownItem`, `IActionOutsideItemButton`, `IActionOutsideItemCustom`) in the notes in the markdown above the `Composable: Actions` example on the `Composable Table page and the `Actions and first cell in body rows as th`example on the`Legacy Table` page. If it's easier, you could just add a link to the github page with the type definitions in the examples so they can look it all up themselves.

Sure!

josejulio:
Sorry for the long delay, I added a link to the [TableTypes.tsx](https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/components/Table/TableTypes.tsx) file and a note to look for the `IAction` Type

nicolethoen (approved):

kmcfaul (approved):
Looks good to me 

@josejulio Could you rebase to the latest? Then we can get this in

nicolethoen (approved):

kmcfaul (approved):

nicolethoen:
This PR was reverted because KeyCloak was reporting a type issue that was holding up the release.

the error was 
```
[INFO] [15:55:03] [@snowpack/plugin-typescript] src/components/table-toolbar/KeycloakDataTable.tsx(313,14): error TS2339: Property 'onClick' does not exist on type 'Action<T>'.
[INFO]   Property 'onClick' does not exist on type 'IActionSeparator & { onRowClick?: ((row: T) => void | Promise<boolean>) | undefined; }'.
```

They are using the IAction type the following way
```
export type Action<T> = IAction & {
  onRowClick?: (row: T) => Promise<boolean> | void;
};
```

PR #6111, opened 2021-08-03
feat(Breadcrumb): Spread props over custom component passed to BreadcrumbItem

jonkoops:
Allows other props to be passed to the `BreadcrumbItem` for example when using a React Router link:

```tsx
import { BreadcrumbItem } from '@patternfly/react-core'
import { Link } from 'react-router-dom'

<BreadcrumbItem component={Link} to={toMyRoute()} />
```

### DIFF
--- a/packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
+++ b/packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
@@ -59,7 +59,7 @@ export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
       {isDropdown && <span className={css(styles.breadcrumbDropdown)}>{children}</span>}
       {render && render({ className, ariaCurrent })}
       {to && !render && (
-        <Component {...props} href={to} target={target} className={className} aria-current={ariaCurrent}>
+        <Component href={to} target={target} className={className} aria-current={ariaCurrent} {...props}>
           {children}
         </Component>
       )}
--- a/packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
+++ b/packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
@@ -59,7 +59,7 @@ export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
       {isDropdown && <span className={css(styles.breadcrumbDropdown)}>{children}</span>}
       {render && render({ className, ariaCurrent })}
       {to && !render && (
-        <Component href={to} target={target} className={className} aria-current={ariaCurrent}>
+        <Component {...props} href={to} target={target} className={className} aria-current={ariaCurrent}>
           {children}
         </Component>
       )}


packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -59,7 +59,7 @@ export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
       {isDropdown && <span className={css(styles.breadcrumbDropdown)}>{children}</span>}
       {render && render({ className, ariaCurrent })}
       {to && !render && (
-        <Component href={to} target={target} className={className} aria-current={ariaCurrent}>
+        <Component {...props} href={to} target={target} className={className} aria-current={ariaCurrent}>
```
tlabaj, line 62: Looks great.  Can you spread the props at the end?

jonkoops: Done :+1:

jenny-s51 (approved):

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -59,7 +59,7 @@ export const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
       {isDropdown && <span className={css(styles.breadcrumbDropdown)}>{children}</span>}
       {render && render({ className, ariaCurrent })}
       {to && !render && (
-        <Component href={to} target={target} className={className} aria-current={ariaCurrent}>
+        <Component href={to} target={target} className={className} aria-current={ariaCurrent} {...props}>
```
kmcfaul, line 62: This will spread `props` over two places (here and the `li` above). Adding another property with `any` type such as `componentProps` and spreading that may make more sense than overloading `props`.

kmcfaul (changes requested):

stevekuznetsov:
I was told to use this instead:


```tsx
        const link = (props: BreadcrumbItemRenderArgs): React.ReactNode => {
            return <Link to={something}
                         onClick={onOpenClick}
                         className={props.className}
                         aria-current={props.ariaCurrent}>text</Link>;
        }
        return <BreadcrumbItem render={link}/>;
```

packages/react-core/src/components/Breadcrumb/BreadcrumbItem.tsx
jonkoops: Good catch! I actually didn't notice these props were already being spread over the root element, which arguably makes a lot more sense.

jonkoops:
@stevekuznetsov Yeah, I think I will just use this `render` prop instead to accomplish what I need. What I find stranger is that the `ariaCurrent` prop doesn't follow the convention of React's `aria-current`. So the props cannot be spread over the element used in the `render` prop.

For example:
```tsx
<BreadcrumbItem
  render={(props) => (
    <Link {...props} to={toRealmSettings({ realm, tab: "keys" })}>
      {t("keys")}
    </Link>
  )}
/>
```
Is not valid because of this.

jonkoops:
Closing this as my problem is resolved.

PR #6114, opened 2021-08-03
fix(menu): forward refs to fix openshift types

redallen:
**What**: Enables OpenShift's workaround of `translate="no"` for Menu/MenuList to keep working. cc @jcaianirh 

**Additional issues**:


kmcfaul:
Why do we need to forward these refs? Is it just Menu or do other components need the forwardRef as well?

redallen:
Openshift is relying on `Menu` and `MenuItem` having forwaded refs which require passing an extra `translate="no"` prop due to broken types in Openshift's specific version of `@types/react`. They wish to keep passing this prop when upgrading PatternFly.

redallen:
Closing since Openshift will remove the `translate="no"` props.

jcaianirh:
@redallen thanks for doing this exercise !

PR #6128, opened 2021-08-05
fix(jumplinks): improve perf

### LINKED ISSUE #6127: Improve jumplinks performance
4fps :(

image

redallen:
**What**: Fixes #6127

smooth-scrollspy

**Additional issues**: `|| newScrollItems[0].offsetParent !== null` was the big offender which was always true and caused a render loop. A possible solution could be to flip the condition ([per MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) it's null when we can't track `offsetTop` anyway) but a similar performance penalty would be incurred for hidden elements. I've removed the condition for now.

### DIFF
--- a/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
+++ b/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
@@ -104,47 +104,50 @@ export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
   React.useEffect(() => setIsExpanded(isExpandedProp), [isExpandedProp]);
   const navRef = React.useRef<HTMLElement>();
 
-  if (hasScrollSpy) {
-    React.useEffect(() => {
-      if (!canUseDOM) {
-        return;
+  const scrollSpy = React.useCallback(() => {
+    if (!canUseDOM || !hasScrollSpy) {
+      return;
+    }
+    const scrollableElement = document.querySelector(scrollableSelector) as HTMLElement;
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/JumpLinks/JumpLinks.tsx
```diff
@@ -104,47 +104,50 @@ export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
...
 
-  if (hasScrollSpy) {
-    React.useEffect(() => {
-      if (!canUseDOM) {
-        return;
+  const scrollSpy = React.useCallback(() => {
+    if (!canUseDOM || !hasScrollSpy) {
+      return;
+    }
+    const scrollableElement = document.querySelector(scrollableSelector) as HTMLElement;
```
evwilkin, line 111: Can we pull this variable out of the `scrollSpy` function and just set it in the `useEffect` below so `scrollSpy` can just use the reference to the queried element?  It shouldn't need to be queried on every scroll action.

evwilkin (changes requested):
One comment about running document.querySelector on every triggered scroll event, otherwise looks good

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
redallen: Good idea, done!

evwilkin (approved):
LGTM - thanks!

nicolethoen (approved):

Issue #3966, opened 2020-03-24
File Upload component exposes dropzoneProps from react-dropzone, should instead abstract away those features


mturley:
**Describe the issue. What is the expected and unexpected behavior?**

The new [beta FileUpload component](https://www.patternfly.org/v4/documentation/react/components/fileupload) uses the `react-dropzone` library internally to handle the file browse and drag/drop behavior. The basic use case (single text file, no restrictions) is handled directly by our FileUpload props (e.g. value, filename, onChange, onReadStarted/Finished, etc) which determine what simple props are passed into the Dropzone.

All other features of Dropzone are made available by exposing an object prop called `dropzoneProps` which allows the consumer to pass whatever props they need to that are supported by Dropzone. The "Text file with restrictions" example in our docs demonstrates this, by passing the `accept`, `maxSize` and `onDropRejected` props directly in `dropzoneProps`, since we don't expose our own props for these things.

I was discussing this with @redallen and he made the point that exposing their props interface directly may not be a good idea, since we will be stuck with that interface and dependent on their documentation for any features that require it. If we ever wanted to move away from react-dropzone and use something else internally, or remain on an older version if we disagree with changes they make in the future, we'd be stuck with consumers using their old props and we'd have to make a large breaking change or have people confused why they can't use newer props from new react-dropzone docs.

His proposed solution is to remove the `dropzoneProps` escape hatch, and expose our own props matching a subset of the props they provide. We could copy and paste their prop descriptions from here: https://github.com/react-dropzone/react-dropzone/blob/master/src/index.js#L66

This way, all of the available props and documentation would be in-house, we would no longer need to depend on react-dropzone's docs, and if we needed to depart from react-dropzone's future changes we would be able to without necessarily breaking consumers.

The main drawback here would be that if there is some use case the dropzone props support that we choose not to expose, consumers would not be able to use that unless we update our props. Also, we would complicate our props table in the docs with a bunch of stuff many people will never use.

I'm leaning towards agreeing with @redallen here, but feedback is welcome.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

N/A

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug, sort of, more like potential technical debt.

**What is your product and what release version are you targeting?**

N/A

mturley:
I think we could decide on equivalent props to expose and have dropzoneProps override them if it is still passed. That would be non-breaking

PR #6139, opened 2021-08-06
fix(FileUpload): Improving events in the FileUpload

### LINKED ISSUE #5545: FileUpload does not fire change events from the file input element
The `FileUpload` component takes an interesting approach to the `onChange` callback. It fires several different events depending on the action taken by the user. However at the end of the day there is only a single source of truth in the DOM for change events related to the file input and that is the `input type="file"`(hidden) element. As a developer I don't care whether the clear button was pressed or the user dragged a file vs browsed. I only care that the file input has changed.

When working with 3rd party form libraries such as `Formik`, the expectation is that change events originate always from the input.

Current `onChange` event:
```
  onChange?: (
    value: string | File,
    filename: string,
    event:
      | React.DragEvent<HTMLElement> // User dragged/dropped a file
      | React.ChangeEvent<HTMLTextAreaElement> // User typed in the TextArea
      | React.MouseEvent<HTMLButtonElement, MouseEvent> // User clicked Clear button
  ) => void;
```

As a developer, I want component APIs to match native DOM as closely as possible. Any extra params should be in addition to the event.
```
  onChange?: (
    event: React.ChangeEvent<HTMLInputElement>,
    // extra value add params here
  ) => void;
```

The `value` is only valid for text files. I suggest a separate callback for notifying of the text changes. This keeps things clean for users who want to simply replace a standard html file input type with the patternfly component.

It seems that patternfly has adopted a pattern of placing the event as the last parameter. I think this is incorrect and should mimic the DOM first with value add as secondary. However adjusting all inputs is a larger task and in this case we can work around this by wrapping the component and normalizing the event.

Furthermore the `Clear` button doesn't actually clear the form field, it only notifies a change. But if you inspect the DOM, the hidden input still has the previous file as a value.

Look at `FileUpload` API and observable at https://patternfly-react.surge.sh/components/file-upload/

Bug

OpenShift Console 4.8

wojta:
**What**: Closes #5545

# What changes?
* `onChange` is now split in more events, old event remain there for the compatibility
* `onTextChange` - when text area changes (@mturley edit: renamed from `onTextChanged`)
* `onFileInputChange` - called when internal `<input>` field emits its own `onChange` DOM event (@mturley edit: renamed from `onInputChange`)
* `onDataChange` - when data is loaded in the component (@mturley edit: renamed from `onDataChanged`)
* `onClearClick` - when clear button is clicked (@mturley edit: renamed from `onClearClicked`)
* react-dropzone upgraded to the `10.2.2`

### DIFF
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -38,7 +38,7 @@
     "@patternfly/react-styles": "^4.19.3",
     "@patternfly/react-tokens": "^4.21.3",
     "focus-trap": "6.2.2",
-    "react-dropzone": "11.3.4",
+    "react-dropzone": "10.2.2",
     "tippy.js": "5.1.2",
     "tslib": "^2.0.0"
   },
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -38,7 +38,7 @@
     "@patternfly/react-styles": "^4.19.3",
     "@patternfly/react-tokens": "^4.21.3",
     "focus-trap": "6.2.2",
-    "react-dropzone": "9.0.0",
+    "react-dropzone": "11.3.4",
     "tippy.js": "5.1.2",
     "tslib": "^2.0.0"
   },
--- a/packages/react-core/src/components/FileUpload/FileUpload.tsx
+++ b/packages/react-core/src/components/FileUpload/FileUpload.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react';
-import Dropzone, { DropzoneProps, DropEvent, DropzoneInputProps, FileWithPath } from 'react-dropzone';
+import Dropzone, { DropzoneProps, DropEvent, DropzoneInputProps } from 'react-dropzone';
 import { FileUploadField, FileUploadFieldProps } from './FileUploadField';
 import { readFile, fileReaderType } from '../../helpers/fileUtils';
-import { fromEvent } from 'file-selector'
+import { fromEvent } from 'file-selector';
 export interface FileUploadProps
   extends Omit<
-  FileUploadFieldProps,
-  'children' | 'onBrowseButtonClick' | 'onClearButtonClick' | 'isDragActive' | 'containerRef'
+    FileUploadFieldProps,
+    'children' | 'onBrowseButtonClick' | 'onClearButtonClick' | 'isDragActive' | 'containerRef'
   > {
   /** Unique id for the TextArea, also used to generate ids for accessible labels. */
   id: string;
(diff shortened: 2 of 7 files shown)


nicolethoen:
So far the new properties make sense to me! :)

wojta:
I agreed with @christianvogt on adding `onInputChanged` that will do same as the `input`'s `onChange`. No I fully understand the goal of the task is to provide replacement for HTML input type=file , but keeping the same implementation approach.

mturley:
@wojta do you have an opinion on https://github.com/patternfly/patternfly-react/issues/3966 ? I think it might make sense to also remove the `dropzoneProps` and expose new props to meet the use cases we were solving by opening those props up. The consumer shouldn't necessarily need to know/care about Dropzone.

packages/react-core/package.json
```diff
@@ -38,7 +38,7 @@
     "@patternfly/react-styles": "^4.11.8",
     "@patternfly/react-tokens": "^4.12.9",
     "focus-trap": "6.2.2",
-    "react-dropzone": "9.0.0",
+    "react-dropzone": "11.3.4",
```
redallen, line 41: [`react-dropzone@11.0.0`](https://github.com/react-dropzone/react-dropzone/releases/tag/v11.0.0) is a breaking change for [AMQ](https://github.com/EnMasseProject/enmasse/blob/master/console/console-init/ui/src/components/UploadFile/UploadFile.tsx#L40) (will produce a type error for `File` vs `FileRejection`).

Openshift and Kogito also use `dropzoneProps` but don't rely on `onDropRejected` event types.

packages/react-core/src/components/FileUpload/FileUpload.tsx
```diff
@@ -113,41 +131,60 @@ export const FileUpload: React.FunctionComponent<FileUploadProps> = ({
...
-  const onDropRejected: DropFileEventHandler = (rejectedFiles: File[], event: React.DragEvent<HTMLElement>) => {
+  const onDropRejected = (rejectedFiles: FileRejection[], event: DropEvent) => {
     if (rejectedFiles.length > 0) {
-      onChange('', rejectedFiles[0].name, event);
+      onChange('', rejectedFiles[0].file.name, event);
     }
     dropzoneProps.onDropRejected && dropzoneProps.onDropRejected(rejectedFiles, event);
   };
 
+  let textInput: any;
```
redallen, line 141: Instead of using this value that defaults to `ref: ''` (which relies on React's legacy ref system) can you use `const inputRef = inputProps.ref || React.useRef<HTMLInputElement>();` instead?

redallen (changes requested):

packages/react-core/package.json
wojta: Does it mean that I should leave the old version there?

redallen: If you don't want to break AMQ leave it at 9.0.0. If you hear from @drcoolsanjeev that he's okay updating react-dropzone and changing the type [in AMQ](https://github.com/EnMasseProject/enmasse/blob/master/console/console-init/ui/src/components/UploadFile/UploadFile.tsx#L40) then you can upgrade to v11.

redallen (approved):

redallen:
Thanks for your perseverance @wojta !

mturley:
@redallen I pushed commits to resolve my comments. I also renamed the `onTextChanged` and `onDataChanged` to `onTextChange` and `onDataChange` respectively, to be consistent with the `onFileInputChange` and `onChange` naming tense.

nicolethoen:
@kmcfaul @tlabaj this is ready for a final review

nicolethoen (approved):

packages/react-core/src/components/FileUpload/FileUpload.tsx
```diff
@@ -18,15 +18,19 @@ export interface FileUploadProps
   value?: string | File;
   /** Value to be shown in the read-only filename field. */
   filename?: string;
-  /** A callback for when the file contents change. */
+  /** *(deprecated)* A callback for when the file contents change. Please use rather the individual events. */
```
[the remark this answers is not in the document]
tlabaj: can we change the wording here a bit.  `The Please use rather the individual events` is a bit confusing.

tlabaj (changes requested):
Looks great.  Jut a couple of small nits

packages/react-core/src/components/FileUpload/FileUpload.tsx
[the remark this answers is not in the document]
mturley: How about "A callback for when the file contents change. Please instead use onFileInputChange, onClearClicked, onTextChange, onDataChange individually."?

I'm also realizing I didn't propagate some of these changes to the examples markdown. Fixing now as well.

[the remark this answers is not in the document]
mturley: @tlabaj I pushed fixes for these and rebased.

tlabaj (approved):

PR #6141, opened 2021-08-06
feat(ContextSelector): Add plain variant

### LINKED ISSUE #5932: Context selector: Add plain variant
Add a "plain" variant of the context selector. Core issue: https://github.com/patternfly/patternfly/issues/3954

PR: https://github.com/patternfly/patternfly/pull/4095

Adds both a plain (`pf-m-plain`) and text (`pf-m-text`) variation, similar to the options menu. Though there isn't a use case for the plain variation by itself, so core didn't include an example of that in the docs - just the plain text variant - https://www.patternfly.org/v4/components/context-selector/html#plain-with-text

CooperRedhat:
**What**: Closes #5932


packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -104,7 +107,13 @@ export class ContextSelectorToggle extends React.Component<ContextSelectorToggle
         {...props}
         id={id}
         ref={this.toggle}
-        className={css(styles.contextSelectorToggle, isActive && styles.modifiers.active, className)}
+        className={css(
+          styles.contextSelectorToggle,
+          isActive && styles.modifiers.active,
+          isPlainText && styles.modifiers.plain,
+          isPlainText && styles.modifiers.text,
```
redallen, line 114: Would there ever be a time you'd want just `styles.modifiers.text?`?

redallen (approved):

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
CooperRedhat: Not at the moment, cleaned that up to a one liner

packages/react-core/src/components/ContextSelector/examples/ContextSelector.md
```diff
@@ -85,6 +86,70 @@ class SimpleContextSelector extends React.Component {
...
+  function onSelect(event: any, value: string) {
+    setSelected(value);
+    setOpen(!isOpen);
+  };
+
+  function onSearchInputChange(value: string) {
+    setSearchValue(value);
+  };
+
+  function onSearchButtonClick(event: any) {
```
nicolethoen, line 127: The event type here should match the API. The type is `event: React.SyntheticEvent<HTMLButtonElement>`

packages/react-core/src/components/ContextSelector/examples/ContextSelector.md
```diff
@@ -85,6 +86,70 @@ class SimpleContextSelector extends React.Component {
...
+  'AWS',
+  'Azure',
+  'My Project 2',
+  'OpenShift Cluster ',
+  'Production Ansible 2 ',
+  'AWS 2',
+  'Azure 2'
+];
+
+const PlainTextContextSelector: React.FC = () => {
```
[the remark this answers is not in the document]
CooperRedhat: Looks like there isn't a difference: `type FC<P = {}> = FunctionComponent<P>;` but will change for consistency. Did lead me to read this article, however: https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/

nicolethoen (dismissed):

mcoker (commented):
Example looks great. The options menu has this variation as combining `isPlain` and `isText`. Should this prop match?

We don't have a use case for a plain context selector, so we didn't include an example for it or an entry in the core docs for `.pf-m-plain`, though technically you can apply `.pf-m-plain` to a context selector toggle and it works as you would expect it.

CooperRedhat:
> Example looks great. The options menu has this variation as combining `isPlain` and `isText`. Should this prop match?
> 
> We don't have a use case for a plain context selector, so we didn't include an example for it or an entry in the core docs for `.pf-m-plain`, though technically you can apply `.pf-m-plain` to a context selector toggle and it works as you would expect it.

@mcoker What options menu are you talking about? I can separate the two flags and potentially give them a variant alias like is being discussed [here](https://github.com/patternfly/patternfly-react/pull/6172). Either way I think we should be consistent across components.

mcarrano (approved):
Looks good to me! Thanks @CooperRedhat .

mcoker:
@CooperRedhat this is the options menu with the plain/text props - https://www.patternfly.org/v4/components/options-menu#plain-with-text

nicolethoen (approved):

mcoker (approved):

tlabaj:
Looks good.  Can you just add. the new props to the demo app please.

kmcfaul (approved):

tlabaj (approved):

kmcfaul (approved):

PR #6143, opened 2021-08-06
fix(Toolbar): use categoryName.key as filter counter index

### LINKED ISSUE #6131: Toolbar filters issues if filter label changes

When using a Toolbar with filters, we encountered an issue where, after clearing all filters, the "Clear All Filters" button remained on screen. After a bit of digging, I found it this was caused because we were dynamically changing the label of some of our filters when they were added/removed.

Looking at the internals of the Toolbar, I noticed the active filter count is being tracked by the `label` of the filters, rather than the (unchanging) `key`:

Screen Shot 2021-08-04 at 10 34 14 AM

In the screenshot, you can see both "Limit" and "Limit (job__limit)", which were both the same filter (same for "Name" and "Name (name__icontains)"). This caused the count of active filters to fall out of sync. I would expect the key to be used, since that's more "id" like than a text label.

Bug. We were able to work around it by ensuring our labels don't change (https://github.com/ansible/awx/pull/10819). In our particular use-case, this is fine, but I could imagine scenarios where the label might legitimately need to be changed dynamically.

nicolethoen:
**What**: Closes #6131

### DIFF
--- a/packages/react-core/src/components/Toolbar/ToolbarFilter.tsx
+++ b/packages/react-core/src/components/Toolbar/ToolbarFilter.tsx
@@ -55,13 +55,23 @@ export class ToolbarFilter extends React.Component<ToolbarFilterProps, ToolbarFi
 
   componentDidMount() {
     const { categoryName, chips } = this.props;
-    this.context.updateNumberFilters(typeof categoryName === 'string' ? categoryName : categoryName.name, chips.length);
+    this.context.updateNumberFilters(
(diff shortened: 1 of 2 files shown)


nicolethoen:
@CooperRedhat We do. Well, the user can choose to use a string for a category name, or they can use a `ToolbarChipGroup` object which requires a name and a key for basically exactly the reason mentioned in the issue.

tlabaj (approved):

Issue #5963, opened 2021-06-28
Search input with submit button


mcoker:
https://github.com/patternfly/patternfly/pull/4180

A new variation of the search input that includes a submit button

cc @mcarrano

PR #6145, opened 2021-08-09
feat(SearchInput): submit the search when hitting Enter in the main input field

KKoukiou:
@mcarrano do you agree with this change in the behavior of this component?


mcarrano:
@KKoukiou yes, I agree. There is currently an active issue open to create a variant of the search input with a submit button: https://github.com/patternfly/patternfly-react/issues/5963 I will make a note that pressing the enter key should also trigger the search.

tlabaj:
Thanks @KKoukiou 
@mcarrano I think it is ok to have these PR go in first.  No need to wait on the submit button issue.

nicolethoen (approved):

PR #6149, opened 2021-08-09
feat(Navigation): add horizontal subnav variant

### LINKED ISSUE #6045: Navigation: Introduce horizontal sub-nav variant and demo
Want to introduce a new sub-nav variant for updated styling as an alternative to existing Teritiary nav. 

Core issue: https://github.com/patternfly/patternfly/issues/3935
Design: https://marvelapp.com/prototype/fj0941h/screen/77719995

Also want to update related demos to be consistent with core updates.

kmcfaul:
**What**: Closes #6045

In this PR:
- Adds new `horizontal-subnav` variant option to `variant`
- Updates examples, tests, and integration
- Fixes some other nav jest tests where variant was being applied to NavList instead of Nav


nicolethoen (commented):
This looks good - I notice that Core also added a horizontal sub nav demo with the nav in a page. It may be good to add that as well for parity?

mcoker (commented):
Variant looks good to me! Though part of the core PR is adding a new page section `.pf-c-page__main-subnav` to hold this component, which is needed before adding any examples using the page component. That could be a separate issue though.

Also need to rename the tertiary demo to "Legacy tertiary nav" since we're encouraging using this one instead of the existing tertiary nav.

nicolethoen:
It's looking good - I am noticing that the code editor is missing from below each of the demos. Not sure if it's because the screenshots are missing?
I think when the site is running you can use `yarn screenshots` in a second console and it will generate the new screenshots. Can you do that and see if that fixes (both the missing screenshots and the missing code editor)?

kmcfaul:
Opened #6159 as it appears every example/demo with a screenshot is missing the code block

mcarrano:
This is looking great @kmcfaul . Just one more ask. Can we also implement this demo layout that has a vertical nav plus the new subnav? https://www.patternfly.org/v4/components/navigation/html-demos#horizontal-subnav  It's probably the most likely use case for this. Sorry that I missed it on my first review.

mcoker (approved):
LGTM! One nit I noticed - on page load, the first item in the horizontal subnav example is scrolled to the right by a tiny bit, so the left scroll button is enabled - is that intended?

kmcfaul:
faed@mcarrano Updated with horizontal subnav + default nav demo

nicolethoen (approved):

mcarrano (approved):
This looks great. Thanks @kmcfaul !

CooperRedhat (approved):

PR #6152, opened 2021-08-09
Fix menu crash if current menuRef value is not found.

### LINKED ISSUE #6151: Menu component crashes with latest core update
`Menu` component crashes in some cases with latest PF update. To be fair, the exact reason is not well known, but it looks like it was introduced with #5789.

Crash and call stack:
```
TypeError
Cannot read property 'querySelector' of null
Call Stack
 MenuBase.allowTabFirstItem
 MenuBase.componentDidUpdate
```

Problem could be solved when null-checking `this.menuRef.current` in new `allowTabFirstItem` function.

Have tried to create a codesandbox, but could not create a minimal version which can reproduce this crash. But our automated tests can reproduce this very well.

Bug. Yes this is blocking us and the only known workaround is using an older PatternFly version.

OpenShift Console

jerolimov:
**What**: Closes #6151

**Additional issues**:
Blocks https://github.com/openshift/console/pull/9279, see https://github.com/openshift/console/pull/9767#issuecomment-895607590

This PR fixes that `Menu` component can crash if menuRef is null. Other code in `handleDrilldownTransition` already checks if the current value is set. The new `allowTabFirstItem` function does this with this PR as well.

Crash and call stack:
```
TypeError
Cannot read property 'querySelector' of null
Call Stack
 MenuBase.allowTabFirstItem
 MenuBase.componentDidUpdate
```

### DIFF
--- a/packages/react-core/src/components/Menu/Menu.tsx
+++ b/packages/react-core/src/components/Menu/Menu.tsx
@@ -90,11 +90,12 @@ class MenuBase extends React.Component<MenuProps, MenuState> {
 
   allowTabFirstItem() {
     // Allow tabbing to first menu item
-    const first = this.menuRef.current.querySelector('ul').querySelector('button, a') as
-      | HTMLButtonElement
-      | HTMLAnchorElement;
-    if (first) {
-      first.tabIndex = 0;
+    const current = this.menuRef.current;
+    if (current) {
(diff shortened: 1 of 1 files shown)


redallen (approved):

jessiehuff (approved):
Thanks! 🙂

PR #6155, opened 2021-08-10
feat(Popover): support diagonal positions

### LINKED ISSUE #6075: Popover support for diagonal positioning
https://github.com/patternfly/patternfly/pull/4226

Popover now supports the following classes to reposition the arrow when the popover is placed in one of these positions

* `.pf-m-top-left`
* `.pf-m-top-right`
* `.pf-m-right-top`
* `.pf-m-right-bottom`
* `.pf-m-bottom-left`
* `.pf-m-bottom-right`
* `.pf-m-left-top`
* `.pf-m-left-bottom`

CooperRedhat:
**What**: Closes #6075

New positioning can be seen in Advanced section of examples

### DIFF
--- a/packages/react-core/src/components/Popover/Popover.tsx
+++ b/packages/react-core/src/components/Popover/Popover.tsx
@@ -71,7 +71,22 @@ export interface PopoverProps {
    * ['top', 'right', 'left']. Since there is no space to the top, it checks if right is available. There's also no
    * space to the right, so it finally shows the popover on the left.
    */
-  flipBehavior?: 'flip' | ('top' | 'bottom' | 'left' | 'right')[];
+  flipBehavior?:
+    | 'flip'
+    | (
+        | 'top'
+        | 'bottom'
+        | 'left'
+        | 'right'
+        | 'top-start'
+        | 'top-end'
+        | 'bottom-start'
+        | 'bottom-end'
+        | 'left-start'
+        | 'left-end'
+        | 'right-start'
+        | 'right-end'
+      )[];
   /**
    * Footer content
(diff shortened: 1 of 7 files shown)


nicolethoen:
I'm don't see any updates made to any of the example markdown files?

mcoker (dismissed):
Positions and styles LGTM!

CooperRedhat:
> I'm don't see any updates made to any of the example markdown files?

@nicolethoen Thanks for picking that up. I had originally added the new positions to the PopoverPostion enum defined in `Popover.jsx`, but two other components, `ClipboardCopy` and `CodeEditorControl` were importing it and I ran into build errors. Interestingly it was only those components and the example in `Popover.md` that used the enum.

nicolethoen:
> I ran into build errors.

What sorts of errors? I think importing the enum in the example is actually wise since we'd love people to know of and possibly use the enum. small updates to `clipboardCopy` and `CodeEditorControl` may be necessary - that'd be okay

mcoker (dismissed):
🥳

nicolethoen:
@CooperRedhat I think I'd rather not have an array in the examples and we should use the enum in the examples per my comment from last week :) So we should figure out what the build errors are coming from

CooperRedhat:
> @CooperRedhat I think I'd rather not have an array in the examples and we should use the enum in the examples per my comment from last week :) So we should figure out what the build errors are coming from

@nicolethoen I Removed the unused positions array, example iterates over PopoverPosition. The errors came from me using the enum improperly, but are gone now.

nicolethoen (approved):
Looks great!

mcoker (approved):

kmcfaul (approved):

PR #6170, opened 2021-08-12
Masthead

### LINKED ISSUE #6119: Add new Masthead component

kmcfaul:
**What**: Closes #6119

In this PR:
- Adds `Masthead`, `MastheadToggle`, `MastheadBrand`, `MastheadMain`, `MastheadContent`
- Adds `isFullHeight` flag to `Dropdown` and `ContextSelector` (for demo)
- Adds Masthead unit tests, Masthead integration demo

Open issues:
- `Masthead` is missing `inset` modifiers on css (core issue open) but the property is in place
- Styling on demo context selector + dropdown is off due to `ToolbarContent` having `align-items: center` (which is limiting the full height). @mattnolting @mcoker Should I override the property on `pf-c-toolbar__content`? I noticed that this element is not in the HTML demo at all but our react component bundles `pf-c-toolbar__content` and `pf-c-toolbar__content-section` together under `ToolbarContent`.


packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
```diff
@@ -126,6 +127,7 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
                 direction === DropdownDirection.up && styles.modifiers.top,
                 position === DropdownPosition.right && styles.modifiers.alignRight,
                 isOpen && styles.modifiers.expanded,
+                isFullHeight && styles.modifiers.fullHeight,
```
nicolethoen, line 130: Are we going to get any warnings in consoles about props spreading without any extra checks here before passing isFullHeight to BaseComponent?

packages/react-core/src/components/Masthead/examples/Masthead.md
```diff
@@ -0,0 +1,225 @@
...
+---
+
+import BarsIcon from '@patternfly/react-icons/dist/js/icons/bars-icon';
+import { Link } from '@reach/router';
+
+## Examples
+
+### Basic
+
+```js
```
nicolethoen, line 15: Not sure how feasible it is, but you didn't just create a whole bunch of typescript examples in the demo app - should you be making these examples using typescript?

packages/react-core/src/components/Masthead/MastheadBrand.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { css } from '@patternfly/react-styles';
+
+export interface MastheadBrandProps
+  extends React.DetailedHTMLProps<React.HTMLProps<HTMLAnchorElement>, HTMLAnchorElement> {
+  /** Content rendered inside of the masthead brand. */
+  children?: React.ReactNode;
+  /** Additional classes added to the masthead brand. */
+  className?: string;
+  /** Component type of the masthead brand. */
+  component?: React.ElementType<any> | React.ComponentType<any>;
```
nicolethoen, line 12: Will this allow for people to use react-router with a logo?

nicolethoen (commented):
Did you or could you open an issue to convert the reusable demo header/page you created to use the masthead?

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
kmcfaul: I should be catching `isFullHeight` above (line 69) so it shouldn't be passed to `BaseComponent` via `props`.

packages/react-core/src/components/Masthead/MastheadBrand.tsx
kmcfaul: I believe so, I copied the typing used by `Button` that allows `Link`. Replacing the `"Logo"` string with an icon should result in an icon Link. Should I update the example code to show this or leave the string?

packages/react-core/src/components/Masthead/examples/Masthead.md
kmcfaul: I can swap the `js` to `ts`, though it won't change much for these examples. Is that all I have to do to get the docs to parse these as TypeScript?

packages/react-core/src/components/Masthead/MastheadBrand.tsx
nicolethoen: Not a bad idea - it will avoid a few questions in the future, i'd imagine. I think would expect that the type of 'component' would need to also allow for `React.ReactNode` for an image or icon to be passed. but I might not totally understand the types you chose.

packages/react-core/src/components/Masthead/examples/Masthead.md
nicolethoen: @redallen

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
nicolethoen: oh yeah - i see that now. In hind sight this was a silly question. this is in the css function... Mondays....

packages/react-core/src/components/Masthead/examples/Masthead.md
redallen: Yes that's correct @kmcfaul .

CooperRedhat:
image


I took the `MastheadBrand` out of its `MastheadMain` wrapper and it broke the layout. From the docs it's not immediately clear that it needs that wrapper (however the examples make it clear). Should we enforce that wrapper in code, or mention it in the docs? Not sure the best practice in this case.

Otherwise LGTM

kmcfaul:
Yeah the relationship between `MastheadMain` and `MastheadBrand` is a little loose due to Masthead's composable structure. The initial idea was to leave it open to the user to include other elements with the `MastheadBrand` (or replace it altogether) under `MastheadMain`. I'll update the documentation to make it clear that `Masthead` expects `MastheadToggle`, `MastheadMain`, and `MastheadContent`, and that `MastheadMain` will typically contain a `MastheadBrand`. 

The other option is to roll `MastheadMain` into `MastheadBrand` and not let anything else be put in the main div. Anyone have opinions?

kmcfaul:
Added the expected structure to the docs

mcarrano (commented):
Looks good @kmcfaul but I did have a question about the demo you are showing. We had defined unique styles for the dropdowns placed in the masthead but I don't see those applied here. @mattnolting can you comment regarding what's needed to pull those in? They are seen in the core demo here: https://www.patternfly.org/v4/components/masthead

nicolethoen:
> The other option is to roll `MastheadMain` into `MastheadBrand` and not let anything else be put in the main div. Anyone have opinions?

In my memory, keeping the two components separate left room for someone to have a clickable logo AND a non-clickable text or company name or something like that. I'm not sure if thats likely or something we really want to allow, but I'm fine either way.

kmcfaul:
@mcarrano The new modifiers are in place for Dropdown/ContextSelector, but the difference in height is caused by a difference in HTML. Our `ToolbarContent` component contains two divs with the classes `pf-c-toolbar__content` and `pf-c-toolbar__content-section` while the HTML demo only uses `pf-c-toolbar__content-section`. The `align-items` css property on `pf-c-toolbar__content` is causing the height to shrink. 

I need guidance on how we want to resolve the difference, I can either override the `align-items` property or manually create the single `pf-c-toolbar__content-section` div inside the demo code, or we need a styling/html update.

@mcoker @mattnolting

mattnolting:
@kmcfaul I think we need to make a [styling update in Core](https://github.com/patternfly/patternfly/issues/4303).

packages/react-core/src/components/Masthead/MastheadBrand.tsx
kmcfaul: @redallen Will that typing also support images/icons? That's how it was defined in `Button` to allow for `Link` but it may be a different case here.

redallen: `React.ElementType<any>` covers images/icons via `component={() => <WhateverConsumerWants />}`.

packages/react-core/src/demos/Masthead.md
```diff
@@ -0,0 +1,205 @@
...
+                  <Dropdown
+                    onSelect={this.onDropdownSelect}
+                    toggle={
+                      <DropdownToggle id="toggle-id" onToggle={this.onDropdownToggle} toggleIndicator={CaretDownIcon}>
+                        Dropdown
+                      </DropdownToggle>
+                    }
+                    isOpen={isDropdownOpen}
+                    dropdownItems={dropdownItems}
+                    isFullHeight
```
mattnolting, line 185: Don't think this would be `isFullHeight`

packages/react-core/src/demos/Masthead.md
```diff
@@ -0,0 +1,205 @@
...
+                <ContextSelector
+                  toggleText={selected}
+                  onSearchInputChange={this.onSearchInputChange}
+                  isOpen={isOpen}
+                  searchInputValue={searchValue}
+                  onToggle={this.onToggle}
+                  onSelect={this.onSelect}
+                  onSearchButtonClick={this.onSearchButtonClick}
+                  screenReaderLabel="Selected Project:"
+                  isFullHeight
```
mattnolting, line 167: 👍

mattnolting (approved):
This looks great @kmcfaul! My only nit is that, in the demo, I wouldn't expect the `dropdown` to be `pf-m-full-height`, just the context selector.

packages/react-core/src/demos/Masthead.md
kmcfaul: The HTML demo has the dropdown with `pf-m-full-height` though?

packages/react-core/src/components/Masthead/examples/Masthead.md
```diff
@@ -0,0 +1,233 @@
...
+propComponents: ['Masthead', 'MastheadToggle', 'MastheadMain', 'MastheadBrand', 'MastheadContent']
+---
+
+import BarsIcon from '@patternfly/react-icons/dist/js/icons/bars-icon';
+import { Link } from '@reach/router';
+import pfIcon from './pf-logo-small.svg';
+
+`Masthead` should contain the following components to maintain proper layout and formatting: `MastheadToggle`, `MastheadMain`, and `MastheadContent`.
+
+`MastheadMain` represents the smaller area taken up by a logo, and will typically contain a `MastheadBrand`. `MastheadContent` represents the main portion of the masthead area and will typically contain a `Toolbar` or other menu-like content such as `Dropdown`.
```
CooperRedhat, line 14: Nice!

CooperRedhat (approved):

nicolethoen (dismissed):

packages/react-core/src/components/Masthead/MastheadBrand.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const MastheadBrand: React.FunctionComponent<MastheadBrandProps> = ({
+  children,
+  className = '',
```
tlabaj, line 17: this can default to undefined.

packages/react-core/src/components/Masthead/MastheadContent.tsx
```diff
@@ -0,0 +1,21 @@
...
+export interface MastheadContentProps extends React.DetailedHTMLProps<React.HTMLProps<HTMLDivElement>, HTMLDivElement> {
+  /** Content rendered inside of the masthead content block. */
+  children?: React.ReactNode;
+  /** Additional classes added to the masthead content. */
+  className?: string;
+}
+
+export const MastheadContent: React.FunctionComponent<MastheadContentProps> = ({
+  children,
+  className = '',
```
tlabaj, line 14: this can default to undefined

packages/react-core/src/components/Masthead/MastheadMain.tsx
```diff
@@ -0,0 +1,21 @@
...
+export interface MastheadMainProps extends React.DetailedHTMLProps<React.HTMLProps<HTMLDivElement>, HTMLDivElement> {
+  /** Content rendered inside of the masthead main block. */
+  children?: React.ReactNode;
+  /** Additional classes added to the masthead main. */
+  className?: string;
+}
+
+export const MastheadMain: React.FunctionComponent<MastheadMainProps> = ({
+  children,
+  className = '',
```
tlabaj, line 14: default to undefined

packages/react-core/src/components/Masthead/MastheadToggle.tsx
```diff
@@ -0,0 +1,21 @@
...
+export interface MastheadToggleProps extends React.DetailedHTMLProps<React.HTMLProps<HTMLDivElement>, HTMLDivElement> {
+  /** Content rendered inside of the masthead toggle. */
+  children?: React.ReactNode;
+  /** Additional classes added to the masthead toggle. */
+  className?: string;
+}
+
+export const MastheadToggle: React.FunctionComponent<MastheadToggleProps> = ({
+  children,
+  className = '',
```
tlabaj, line 14: undefined

tlabaj (commented):
Looks great Katie.  Just a few nits.

tlabaj (approved):

kmcfaul:
Will have 1 minor update when core bump gets merged to fix the toolbar demo css

mcarrano (dismissed):

CooperRedhat (approved):
LGTM, demo works! 

Approving but have a question: 
I noticed you added the demo to the `react-integration` package, but not any integration tests. Is that because the functionality is tested in integration tests for `Toolbar`, `ContextSelector`, etc?

dlabrecq (dismissed):

kmcfaul:
Yeah @CooperRedhat there's not really any functionality built into Masthead itself. Once Page is refactored to use Masthead over PageHeader then there should be something to test in integration

tlabaj (approved):

tlabaj (changes requested):
Can you mark this beta

tlabaj (dismissed):

packages/react-core/src/demos/Masthead.md
```diff
@@ -0,0 +1,202 @@
...
+    return (
+      <Masthead id="basic">
+        <MastheadToggle>
+          <Button variant="plain" onClick={() => {}} aria-label="Global navigation">
+            <BarsIcon />
+          </Button>
+        </MastheadToggle>
+        <MastheadMain>
+          <MastheadBrand>
+            <img src={imgBrand} alt="Patternfly logo" />
```
mcoker, line 148: FWIW, the core demo uses the `<Brand>` component here. I suppose you could use whatever you want though.

packages/react-core/src/demos/Masthead.md
```diff
@@ -0,0 +1,202 @@
...
+                        Dropdown
+                      </DropdownToggle>
+                    }
+                    isOpen={isDropdownOpen}
+                    dropdownItems={dropdownItems}
+                    isFullHeight
+                  />
+                </ToolbarItem>
+                <ToolbarItem>
+                  <Dropdown
```
mcoker, line 186: since this is always on the right edge of the screen, should we use `position={DropdownPosition.right}`?

mcoker (commented):
Looks great! Left a couple of questions. Only thing I think needs to be updated is the dropdown menu alignment since it's going off screen currently.

Screen Shot 2021-08-20 at 11 04 53 AM

packages/react-core/src/components/Masthead/MastheadToggle.tsx
```diff
@@ -0,0 +1,21 @@
...
+  /** Additional classes added to the masthead toggle. */
+  className?: string;
+}
+
+export const MastheadToggle: React.FunctionComponent<MastheadToggleProps> = ({
+  children,
+  className,
+  ...props
+}: MastheadToggleProps) => (
+  <div className={css(styles.mastheadToggle, className)} {...props}>
```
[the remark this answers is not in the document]
kmcfaul: Left as a `div` in this commit. Should it be changed to a span?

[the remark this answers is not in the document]
mcoker: Let's leave it as it is - as far as the layout/structure goes, it doesn't matter. We can follow up later if needed.

tlabaj (dismissed):

mcoker (dismissed):
LGTM!

packages/react-core/src/demos/Masthead.md
mattnolting: Yes, you're right. I should change that to [dropdown with image and text](https://www.patternfly.org/v4/components/dropdown/html#dropdown-with-image-and-text). [Issue 4320](https://github.com/patternfly/patternfly/issues/4320)

nicolethoen:
I'm noticing that the Logo link is not navigable using the keyboard. Not sure why that is...

kmcfaul:
Updated the MastheadBrand to have a tabIndex

nicolethoen (approved):

tlabaj (approved):

PR #6172, opened 2021-08-16
feat(TreeView): add guides + compact variants

### LINKED ISSUE #6046: Tree-view: adds compact-bordered variant
Adds a new variant of the tree view component.

Core issue: https://github.com/patternfly/patternfly/issues/4177
Design:https://marvelapp.com/prototype/73043j4/screen/80030545

CooperRedhat:
**What**: Closes #6046


nicolethoen (commented):
Couple thoughts: 
Maybe instead of two flags - `isCompact` and `compactNoBackground` - it might make sense to have a `variant` prop so that they only have to juggle one prop? It can be `variant?: 'default' | 'compact' | 'compactNoBackground'` and maybe leave the door open for more style variations in the future?
Also, I'm wondering about the use of the word `compact` since that is a term that is used in DataList and Table to a very different type of style adjustment. @mcarrano @mattnolting I'm not sure what you think about that?

CooperRedhat:
> Couple thoughts:
> Maybe instead of two flags - `isCompact` and `compactNoBackground` - it might make sense to have a `variant` prop so that they only have to juggle one prop? It can be `variant?: 'default' | 'compact' | 'compactNoBackground'` and maybe leave the door open for more style variations in the future?
> Also, I'm wondering about the use of the word `compact` since that is a term that is used in DataList and Table to a very different type of style adjustment. @mcarrano @mattnolting I'm not sure what you think about that?

@nicolethoen Using a `variant` prop for combined styles sounds like an easier experience for our users! @mattnolting How does that sound to you? And should `guides` or `withGuides` be added to that variant list?

mcarrano (commented):
This looks good, except I don't see any visual difference between the "Compact" and "Compact no background" examples. Am I missing something? 

As for the question about variants and naming, I am OK with creating a `variant` prop. However we wanted compact options and guides to be applied independently, so I like having `hasGuides` being it's own prop.

CooperRedhat:
> This looks good, except I don't see any visual difference between the "Compact" and "Compact no background" examples. Am I missing something?

No, just pushed a fix that should be visible shortly.

mcarrano (dismissed):
Looks great. Thanks @CooperRedhat !

nicolethoen (approved):

tlabaj (changes requested):
Looks good! Can you just update the demo with new props and update  integration test please.

kmcfaul (dismissed):
Looks good!

tlabaj (approved):
Looks great cooper!

kmcfaul (approved):

PR #6175, opened 2021-08-17
feat(Menu): add app-launcher mockup

### LINKED ISSUE #6018: Make a demo of ApplicationLauncher using Menu Toggle and Menu

kmcfaul:
**What**: Closes #6018

In this PR:
- Adds mockup of ApplicationLauncher (using Menu components) to the Composable Menu demo page
- Changes typing on MenuItem `component` property to allow router `Link`
- Sets `id` on `MenuItem` if given `itemId` (manually setting `id` directly should override `itemId` if both are passed)

The demo covers most of the examples shown in ApplicationLauncher. The CSS overall is similar but there are slight differences in spacing. I think the next step would be to decide how we want to handle those changes, and decide if we want to leave this as a demo or make a new AppLauncher wrapper component for Menu.


packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -869,14 +868,240 @@ MenuWithFlyout = () => {
...
+    </Menu>
+  );
   return (
-    <Popper
-      trigger={toggle}
-      popper={menu}
-      isVisible={isOpen}
-      popperMatchesTriggerWidth={false}
-    />
+    <Popper trigger={toggle} popper={menu} isVisible={isOpen} popperMatchesTriggerWidth={false} enableFlip={false} />
```
CooperRedhat, line 1104: Why no `enableFlip`?

CooperRedhat (approved):
LGTM

packages/react-core/src/demos/ComposableMenu.md
kmcfaul: Ah that was just me forcing the menu to expand downwards (since its last on the page). I can re-enable it.

kmcfaul:
Removed `itemId` setting `id` on MenuItem - `itemId` is specific to return calls and not necessarily tied to `id`. I could re-add this if others think it's worth having `itemId` set `id` automatically (without needing to pass both `itemId` and `id`).

mcarrano (commented):
This looks good @kmcfaul . The only problem I see is that the filtering does not seem to be working. For whatever I type I get "No results found."

nicolethoen (commented):
I'm noticing a weird keyboard interaction:

1.  Use `enter` to open the menu
2. Focus moves to the search
3. `tab` exits the search and moves to the menu items where I can navigate no problem with arrow keys
4. `tab` does not move the focus to the next navigable item on the page that i'd expect. It moves it off the page to the browser address bar. 

It seems that the keyboard navigation is being confused by the menu being appended to the end of the DOM rather than between the app launcher toggle and the example's typescript toggle button...

mcoker:
This looks pretty good. Some things I noticed:

* Group title bottom padding differs - 8px on app launcher, 0px on menu
  * You can see that by hovering over a menu item and noting how close the background color is to the title
  * app launcher
    * Screen Shot 2021-08-17 at 5 43 58 PM
  * menu
    * Screen Shot 2021-08-17 at 5 44 05 PM


* Search box top/bottom padding differs - 16px top/bottom on app launcher, 24px top/bottom on menu
* We impose height/width/alignment restrictions on app launcher icon. No restrictions on the menu icon.
  * The app launcher icon box is 24x24px, then we size down the largest edge (of the icon/image/whatever the user passes to the icon box) to 24px and allow the other edge to scale down.
  * It's up to the user to use the right/consistent sized image in a menu item icon. This is probably fine if they're using react-icons, not so much if using custom icons that may be sized differently.

@mcarrano do we want to update any of this in the menu component CSS to be more inline with app launcher? The app launcher should represent what we'll see in other dropdown type components.

kmcfaul:
@mcarrano Fixed the search, it was failing to filtered properly because options lacked `id` (after I removed `itemId` setting `id`). 

@nicolethoen Appended the popper to a container div to get the proper tab order. I could also remove the popper but then the example container grows/shrinks when the menu is opened/closed. Keyboard interaction is a bit odd for this example specifically, because menu search is outside of the `ul` options and isn't a part of the built in navigation (same with menu footer, you can see in the Menu examples). I removed `Tab` closing the menu for this example to allow access to the options. I can try playing with the interaction a bit more, I can maybe build arrow navigation for the input in the demo code.

mcoker:
@kmcfaul @mcarrano re: [this comment](https://github.com/patternfly/patternfly-react/pull/6175#issuecomment-900682527), I looked closer and I think it's just a markup issue:

* `<MenuList>` shouldn't wrap the whole menu - it should only wrap `<MenuItems>`
* Unless the `<Divider>` is a direct child of a `<MenuList>` (it's used to separate items in a single list), it should be an `<hr>` - doesn't need the `component` prop. Looks like all of the dividers in this menu should be `<hr>`s since they separate groups of things, not items.

Here is an example of the search box markup just showing the elements for the search box + divider - https://pf4.patternfly.org/components/menu#with-filtering

Screen Shot 2021-08-18 at 1 34 17 PM

And here is an example of the favorites markup for what to put in `<MenuContent>` - https://pf4.patternfly.org/components/menu#with-favorites

Screen Shot 2021-08-18 at 1 31 42 PM

kmcfaul:
@mcoker Updated the markup & dividers

packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -869,14 +868,259 @@ MenuWithFlyout = () => {
...
+        <TextInput
+          aria-label="Filter menu items"
+          iconVariant="search"
+          type="search"
+          onChange={value => onTextChange(value)}
+        />
+      </MenuInput>
+      <Divider />
+      <MenuContent>
+        <MenuList>
```
mcoker, line 1102: This needs to wrap the `<MenuItem>`s. I'm assuming it would wrap `{filteredFavorites}` a few lines below this?

FWIW `<MenuList>` generates a `<ul>` element. The only valid children of a `<ul>` are `<li>` elements (what the `<MenuItem>` generates here). And the inverse is similar, `<li>` elements can only be a child of `<ul>` or `<ol>`.

packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -869,14 +868,259 @@ MenuWithFlyout = () => {
...
+          {filteredFavorites.length > 0 && (
+            <MenuGroup key="favorites-group" label="Favorites">
+              {filteredFavorites}
+              <Divider key="favorites-divider" />
```
mcoker, line 1106: This should go between `<MenuGroup>`s.

mcoker (changes requested):

nicolethoen:
I like the updates to the keyboard interaction. It's not ideal but I think the issue of a11y when it comes to items appended to the DOM outside of their apparent parent is something we might have to solve on a larger scale as well.

Thumbs up from me once you address Coker's comments! :)

packages/react-core/src/demos/ComposableMenu.md
kmcfaul: `filteredFavorites` and `filteredItems` are both collections of `MenuItem` / `MenuGroup` so it should be good.

packages/react-core/src/demos/ComposableMenu.md
kmcfaul: If it's outside of `MenuGroup` that would place it directly under `MenuList`, would it need the `component='li'` in that case then?

mcoker (dismissed):
LGTM! Cool to see this come together!!

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -102,7 +102,10 @@ export const MenuItem: React.FunctionComponent<MenuItemProps> = ({
     setFlyoutRef,
     disableHover
   } = React.useContext(MenuContext);
-  const Component = component || to ? 'a' : 'button';
+  let Component = !component && to ? 'a' : 'button';
+  if (component) {
+    Component = component as any;
+  }
```
redallen, line 108: You can write `const Component = (!component && to ? 'a' : 'button') as any;`

packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -4,6 +4,8 @@ section: demos
 beta: true
 ---
 
+import { Link } from '@reach/router';
```
redallen, line 7: Conflicts in this file.

redallen (changes requested):

packages/react-core/src/components/Menu/MenuItem.tsx
kmcfaul: Will that preserve a defined `component` though? The first statement will fall through to setting `button` as the component. But if a `component` is given it should override that assignment.

packages/react-core/src/demos/ComposableMenu.md
kmcfaul: What do you mean?

packages/react-core/src/components/Menu/MenuItem.tsx
redallen: Oh sorry I missed that logic, you could do `const Component = (component || to ? 'a' : 'button') as any;` in that case.

tlabaj: I also found this a little confusing.  Also why don't we just default component to `button` on line 80?

kmcfaul: I was trying to preserve the logic that let `to` set `component` to `a`. 

`component || to ? 'a' : 'button'` wouldn't keep a given `component` either since it's a if-else.

I could use `(to ? 'a' : component) as any` and default `component` to `button`.

kmcfaul:
@redallen @tlabaj 

Updated the menuitem logic for `component`. It looks like something got updated for the menu/popper CSS though, the width is now the full page? @mcoker @mattnolting

kmcfaul:
Styling will get another quick update for the menu width, but shouldn't have any changes on the react-side for it

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -102,7 +102,7 @@ export const MenuItem: React.FunctionComponent<MenuItemProps> = ({
     setFlyoutRef,
     disableHover
   } = React.useContext(MenuContext);
-  const Component = component || to ? 'a' : 'button';
+  const Component = (to ? 'a' : component) as any;
```
redallen, line 105: I like the readability!

redallen (approved):

mcoker (dismissed):
LGTM!

tlabaj (approved):

CooperRedhat:
https://user-images.githubusercontent.com/87719785/130282146-747e46a3-0fa9-4eb5-bb23-97b156635eaa.mov

Just saw this on surge

kmcfaul:
@CooperRedhat Ah good catch, I need to update the filtering logic since the HTML structure has changed a bit since it was initially built. Will have an update out soon.

CooperRedhat (approved):

tlabaj (approved):

PR #6178, opened 2021-08-17
feat(Search input): Added support for submit button

### LINKED ISSUE #5963: Search input with submit button

tlabaj:
**What**: Closes #5963

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -329,42 +329,29 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
               )}
             </span>
           )}
-          {!!onSearch && (
-            <Button
-              type="submit"
-              variant={ButtonVariant.control}
-              aria-label={submitSearchButtonLabel}
-              onClick={onSearchHandler}
-              isDisabled={isDisabled}
-            >
-              <ArrowRightIcon />
-            </Button>
-          )}
         </div>
         {attributes.length > 0 && (
-          <>
-            <Button
-              className={showSearchMenu && 'pf-m-expanded'}
-              variant={ButtonVariant.control}
-              aria-label={openMenuButtonAriaLabel}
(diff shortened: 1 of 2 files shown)


mcarrano (commented):
Looks good @tlabaj . Is this coded so that either pressing enter or clicking the new submit button will trigger the search?

nicolethoen:
the a11y tests are failing because `aria-expanded="true"` is being applied to the nodes that don't have children and are not actually expandable. That may be a bug related to the `defaultAllExpanded` prop. It should only use `aria-expanded` if it has children to expand/collapse.

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -322,6 +322,17 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
               )}
             </span>
           )}
+          {!!onSearch && (
+            <Button
```
mcoker, line 326: This should be a child of the `<InputGroup>`, not `<div className={css(styles.searchInputBar)}>`

mcoker (changes requested):
Looks great! Just one change to match the core example - https://pf4.patternfly.org/components/search-input/#with-submit-button

nicolethoen (dismissed):
Once the tests pass and the structure is updated, looks good to me!

tlabaj:
> Looks good @tlabaj . Is this coded so that either pressing enter or clicking the new submit button will trigger the search?
@mcarrano Katerina added that in PR #6145

kmcfaul (approved):

mcarrano (approved):
Looks great. Thanks @tlabaj !

mcoker (approved):
LGTM!!

evwilkin (approved):
LGTM

dlabrecq (commented):
LGTM

Issue #6124, opened 2021-08-04
dual list selector does not allow for sub folder selection


ddonahue007:
**Describe the issue. What is the expected and unexpected behavior?**
When there is a nested folder you can not select the subfolder and have its children selected. It is not possible to select the subfolder at all. 

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
This is easily reproduced in the examples here
https://patternfly-react.surge.sh/components/dual-list-selector/#expandable-options-with-checkboxes

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
bug
**What is your product and what release version are you targeting?**
Cost Management

tlabaj:
closed by #6180

Issue #5828, opened 2021-05-25
Dual List Selector tree variant ignores initial chosenOptions


rvsia:
**Describe the issue. What is the expected and unexpected behavior?**

When [Dual List Selector](https://www.patternfly.org/v4/components/dual-list-selector) tree variant has initial `chosenOptions`, it ignores them - when moved to available, they are not moved and the selected disappear. When some available are moved to chosen, the chosen are completely rewritten.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

https://codesandbox.io/s/recursing-murdock-kw4ci?file=/index.js

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug

**What is your product and what release version are you targeting?**

Data Driven Forms, Cost Management

rvsia:
cc @ddonahue007

ddonahue007:
Also there is certainly a bug here. If you try to move the initial options back to the left pane you will notice that they are remove from the list completely.

ddonahue007:
This should be fixed, please

tlabaj:
closed by #6180

PR #6180, opened 2021-08-17
fix(DualListSelector): respect initial chosen opts, select subfolders, reset selected button on all button

### LINKED ISSUE #6179: Dual List Selector doesn't reset add/remove selected state when add/remove all is pressed
When an option is selected and then add/remove all is pressed, the add/remove selected button is still active with an empty pane.

kmcfaul:
**What**: Closes #6179, #6124, #5828

In this PR:
- Fix to respect initial chosen options while moving options between panes of tree DLS
- Fix to allow checking of tree DLS subfolders
- Fix to reset selected states when options are moved via add/remove all (leaving an empty pane + active move selected button). Only affected tree variant.
- Update tree DLS example to contain initial chosen options

@nicolethoen The fix for selecting subfolders is possibly a regression on #5763 but I was unable to reproduce the bug from that PR after changing the code (locally tested - selecting a child of a subfolder and then moving selected does not move other children of the same subfolder).

### DIFF
--- a/packages/react-core/src/components/DualListSelector/DualListSelector.tsx
+++ b/packages/react-core/src/components/DualListSelector/DualListSelector.tsx
@@ -138,7 +138,8 @@ export class DualListSelector extends React.Component<DualListSelectorProps, Dua
     removeSelectedAriaLabel: 'Remove selected',
     removeAllAriaLabel: 'Remove all'
   };
-  private originalCopy = this.props.availableOptions;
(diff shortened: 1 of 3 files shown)


nicolethoen:
I haven't looked at this closely yet, but you're my hero for getting to this!

nicolethoen (approved):
Love everything I'm seeing :)

CooperRedhat (approved):

tlabaj (approved):

PR #6197, opened 2021-08-19
fix(NumberInput): call 'onPlus' and 'onMinus' if set only

jonkoops:
The `onPlus` and `onMinus` props are optional, if they are not present they should not be called.


packages/react-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -83,7 +83,7 @@ export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
           variant="control"
```
redallen, line 83: Can you modify lines 52 and 54 to read like `onMinus = () => {},` and `onPlus = () => {}` instead? This is more consistent with our style across other components and generates nicer code.

redallen (changes requested):

redallen (approved):

tlabaj (approved):
Looks great!

PR #6201, opened 2021-08-19
feat(LogViewer): refactor the component to get rid of the gaps between items

### LINKED ISSUE #6200: LogViewer: Refactor react-window code to make log viewer resize dynamically
There are large gaps between every item in the log viewer, it's because we calculate the height of each item based on the estimated size. However, due to [a branch of react-window](https://github.com/bvaughn/react-window/blob/issues/6/src/DynamicSizeList.js), we can create a dynamic size list, so that all the items in the list will be calculated automatically.

Enhancement.

DaoDaoNoCode:
**What**: Closes #6200 
This PR includes:
1. Refactor the react-window repo, use `DynamicSizeList` instead of `VariableSizeList`, which gets rid of the gaps between items.
2. Fix a small bug in the search bar, the badge doesn't reset when the search input changes. Now it's fixed and will be set to default.
3. Add a prop named `headerComponent`, which allows users to render a React component at the header of the log viewer window (between window and toolbar). Also, add a demo to show how to use the header component.


kmcfaul (approved):
Looks great!

nicolethoen (approved):

PR #6228, opened 2021-08-25
chore(avatar): convert examples to typescript

### LINKED ISSUE #6118: Chore: Convert Avatar example to typescript
Convert the Avatar example to typescript to establish work flow for converting  react-core examples.

cc @redallen

redallen:
**What**: Closes #6118

**Additional issues**:


nicolethoen (approved):
Do you need to do anything extra so the cypress tests use these examples after removing the demos from the demo-app?

redallen:
@nicolethoen I talked to @tlabaj and we agreed we can remove integration tests that only test class names as long as we continue to test our breakpoint utility.

But yes, I'll have to do something extra for CI to use our doc build instead of demo app.

tlabaj (approved):

Issue #5833, opened 2021-05-25
Slider: max and customSteps dont play well together


mdujava:
In Slider, specifying `max` lower then `100` and  `customSteps` will not show as intended.

first slider specify both, second only max is specified.

![image](https://user-images.githubusercontent.com/5232235/119569440-070ab200-bdaf-11eb-81c0-852d1b98863f.png)

```
 const freq_steps = [
    { value: 0, label: "A" },
    ...
    { value: 4, label: "E", isLabelHidden: true },
    { value: 5, label: "F" }
  ];
...
// labels are shown, but max is ignored 
  <Slider
      onChange={onChangeFrequency}
      value={frequencyOfTask}
      max={5}
      showTicks
      customSteps={freq_steps}
  />
// max is set correct, but no labels
 <Slider
      onChange={onChangeFrequency}
      value={frequencyOfTask}
      max={5}
      showTicks
  />
```

https://codesandbox.io/s/hopeful-nash-38mnu?file=/index.js

mdujava:
When CustomSteps are applied code uses stepObj.value directly in SliderStep https://github.com/patternfly/patternfly-react/blob/8c671c470af714f80e852afe8be20e608c80342d/packages/react-core/src/components/Slider/Slider.tsx#L344-L346
Compared to automatic steps that calculates relative position (honoring min and max values) https://github.com/patternfly/patternfly-react/blob/8c671c470af714f80e852afe8be20e608c80342d/packages/react-core/src/components/Slider/Slider.tsx#L312-L323

PR #6232, opened 2021-08-26
fix(Slider): update slider to respect min/max constraints

### LINKED ISSUE #5707: Slider: the TextInput when used with the Slider should respect the max value of the slider
This is the example from the docs:
Screen Shot 2021-04-23 at 09 49 08
In the first slider, one can see that the slider max value is 8, the user can however insert in the text input a value bigger than that.

I did not check if it's an implementation issue of the example code, or the component implementation in the react-core.

### LINKED ISSUE #5893: Bug: Value input in Slider component doesn't respect max/min constraints.
If I try to enter a value that's larger than the prescribed maximum, it should be constrained to what's shown on the slider. 

Constraints work if I try to enter a value directly to the value input - after clicking enter a value that exceeds the maximum change to the maximum allowed value. What seems to be an issue is that prescribed constraints don't work when I use carets to increase or decrease the value by set increments over the maximum/ minimum.
Screenshot 2021-06-08 at 9 48 42

**Thumb value input**
What's more, if I enter a value lesser than a prescribed minimum to a thumb value input it let me do that (see the picture below)

Screenshot 2021-06-08 at 9 50 13

But after clicking enter, a thumb and a thumb value input moved far left from 0 (minimum) - out of the sight. It appears again if you click anywhere on a slider. This is the case for both direct inputs and for value changes made by carets as well.

Value entered -4
Screenshot 2021-06-08 at 10 01 28

Value entered -14
Screenshot 2021-06-08 at 9 50 01

Allowed maximimum is 100 (this bugcan be reproduce by using carets but sometimes also by direct input)
Screenshot 2021-06-08 at 10 08 57

Check all the value inputs on [this page](https://www.patternfly.org/v4/components/slider) as none of them seems to be working correctly.

evwilkin:
Closes #5893 
Closes #5707 

**Additional issues**: Related issue #5833 

This PR fixes the issue of being able to enter a value outside slider's min/max via the text input, by passing `setLocalInputValue` to the user through `onChange`.
Steps to recreate issue:
- Set a slider either to max value, or above-max value using the text input.
- `value` and `inputValue` are correctly set to the max value, component state updates & internally updates `localInputValue` variable which controls the value displayed in the text input.
- Immediately enter another value above the max, `value` and `inputValue` remain at the max value so `setState` has no effect and the `localInputValue` is not updated causing above-max value to remain.

This PR also fixes the issue of tabbing through the first two "Value input" examples causes the text input to change to 0, and tabbing through a second time causes the slider to reflect this 0 value, by adding missing `newInputValue` assignment in example code.

### DIFF
--- a/packages/react-core/src/components/Slider/Slider.tsx
+++ b/packages/react-core/src/components/Slider/Slider.tsx
@@ -49,7 +49,11 @@ export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onCh
   /** Position of the input */
   inputPosition?: 'aboveThumb' | 'right';
   /** Value change callback. This is called when the slider value changes */
-  onChange?: (value: number, inputValue?: number) => void;
+  onChange?: (
+    value: number,
+    inputValue?: number,
+    setLocalInputValue?: React.Dispatch<React.SetStateAction<number>>
+  ) => void;
   /** Actions placed to the left of the slider */
   leftActions?: React.ReactNode;
   /** Actions placed to the right of the slider */
@@ -109,7 +113,7 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
     if (event.key === 'Enter') {
       event.preventDefault();
       if (onChange) {
-        onChange(localValue, localInputValue);
+        onChange(localValue, localInputValue, setLocalInputValue);
       }
     }
   };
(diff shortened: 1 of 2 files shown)


nicolethoen (approved):

tlabaj (approved):
Looks great!

PR #6233, opened 2021-08-26
feat(Table): Add hoverable and selectable rows

### LINKED ISSUE #5444: Table: Add example for selectable rows
In the Table component we should show as example of a table with selectable rows. We currently have this example that illustrates use of a Row Click Handler (https://www.patternfly.org/v4/components/table#row-click-handler-and-header-cell-tooltipspopovers) but it doesn't show the corresponding styles for selectable rows that were added per this core PR: https://github.com/patternfly/patternfly/pull/3835

cc @mcoker 

----

https://github.com/patternfly/patternfly/pull/3835 adds:

* `.pf-m-hoverable` - this class can be added to either a `<tr>` for a non-expandable individual table row or a `<tbody>` for an expandable row.
    * those elements need `tabindex="0"` so they can be interacted with via keyboard.
* `.pf-m-selected` to the same elements for when the row or expandable row is in the selected state.

nicolethoen:
**What**: Closes #5444 

Changes:

- I removed the rowClickHandler piece from a different TableComposable example
- I added a new `Composable: Row click handler, hoverable & selected rows` example to TableComposable
- I updated the Legacy Table's `Row click handler and header cell tooltips/popovers` example to include hoverable/selectable features.
- I added keyboard handling to the `onRowClick` prop so that the event can be triggered using keyboard and not just mouse click

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableRowClickDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TableDemo/TableRowClickDemo.tsx
@@ -24,13 +24,19 @@ export class TableRowClickDemo extends React.Component<TableProps, ITableRowClic
       columns: [{ title: 'Repositories' }, 'Branches', { title: 'Pull requests' }, 'Workspaces'],
       rows: [
         {
-          cells: ['Repositories one', 'Branches one', 'Pull requests one', 'Workspaces one']
+          cells: ['Repositories one', 'Branches one', 'Pull requests one', 'Workspaces one'],
+          isHoverable: true,
+          isSelected: true
         },
         {
-          cells: ['Repositories two', 'Branches two', 'Pull requests two', 'Workspaces two']
+          cells: ['Repositories two', 'Branches two', 'Pull requests two', 'Workspaces two'],
+          isHoverable: true,
+          isSelected: false
         },
         {
-          cells: ['Repositories three', 'Branches three', 'Pull requests three', 'Workspaces three']
+          cells: ['Repositories three', 'Branches three', 'Pull requests three', 'Workspaces three'],
+          isHoverable: true,
+          isSelected: false
         }
       ]
     };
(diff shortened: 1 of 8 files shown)


mcoker (dismissed):
LGTM!

kmcfaul (dismissed):

jessiehuff (commented):
Looks pretty good to me! I think the only question I have is how to notify screen reader users that the item is selectable or selected. I've often questioned this in other components as well such as with the [Selectable Datalist](https://www.patternfly.org/v4/components/data-list#selectable-rows). With that one though, we had it in the name of the example so it at least had some indication that it was selectable (perhaps not an ideal solution though). I'm thinking there might be a better way to go about it but might have to research to see if there's an established practice. Open to suggestions for it though. :)

nicolethoen:
I opened issue: https://github.com/patternfly/patternfly-react/issues/6295 to track communicating hoverable/selectable rows for screen readers.

CooperRedhat (changes requested):
Can you change the title of the legacy example to indicate it includes hoverable/selectable?

Otherwise LGTM with accessibility issue opened.

kmcfaul (approved):
lgtm with the follow up opened for the remaining accessibility changes

mturley:
In the description for the existing checkbox-selectable example we use the term "selectable" with the assumption that we mean checkboxes:

> To make a row selectable, the table needs a selection column.

Should we maybe be a little bit more clear about the distinction? Maybe we can add a note to the new example saying something like "There are two ways to make a row selectable: by applying a style on the whole row with `isRowSelected` or by using checkboxes (see example below)" and then we could change "To make a row selectable" to "To make rows selectable with checkboxes".

mcarrano:
@mturley The primary use case for this was in something like a primary-detail view where you would select a row that represents some object and then inspect it's details in a drawer or similar container. This should not be confused with checkbox select to select multiple rows for taking action. It is a bit confusing the way this is presented now. Not sure what the best way is to clarify.

mcarrano:
@nicolethoen Can we do this to try and clarify a bit?

1. Move this example to fall AFTER the "Composable: selectable with checkbox" and Composable: selectable with radio input" examples since those represent the more typical support for selectable rows.

2. Add descriptive text to this example that reads, "Use when a table is used to present a list of objects in a Primary-detail view." Or something like that.

kmcfaul (approved):

CooperRedhat (approved):

mcarrano (approved):
Looks great! Thanks for making those changes @nicolethoen .

tlabaj (approved):

Issue #5820, opened 2021-05-21
Add expandable inline alert


mcarrano:
Add expandable inline alert variant to Alert component. The user should be able to show/hide the contents of the alert by clicking the expansion caret.

https://github.com/patternfly/patternfly/issues/4056

mcarrano:
@jessiehuff can you look at from an a11y perspective?

jessiehuff:
I think that we'll likely just have to test this to make sure that focus shifts properly to the toggle and allows the user to expand the alert. I also think that it would probably require some documentation to explain the ways that this should and shouldn't be used. It also makes me wonder if this would be a good candidate to create a demo with. I think at one point we talked about creating a demo of an alert in a form, maybe this would fit into that?

tlabaj:
closed by #6236

PR #6236, opened 2021-08-26
Alert: Add plain variant, add expandable inline

### LINKED ISSUE #6164: Alert: add 'plain' variant
Add a plain variant to the Alert that removes the border and background fill.

Core PR: https://github.com/patternfly/patternfly/pull/4262

CooperRedhat:
**What**: Closes #6164, #5820

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -8,6 +8,8 @@ import { capitalize, useOUIAProps, OUIAProps } from '../../helpers';
 import { AlertContext } from './AlertContext';
 import maxLines from '@patternfly/react-tokens/dist/esm/c_alert__title_max_lines';
 import { Tooltip } from '../Tooltip';
+import { Button, ButtonVariant } from '../Button';
+import AngleRightIcon from '@patternfly/react-icons/dist/esm/icons/angle-right-icon';
 
 export enum AlertVariant {
   success = 'success',
@@ -52,6 +54,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Set a custom icon to the Alert. If not set the icon is set according to the variant */
   customIcon?: React.ReactNode;
+  /** Flag indicating that the Alert is expandable */
+  isExpandable?: boolean;
 }
 
 export const Alert: React.FunctionComponent<AlertProps> = ({
@@ -74,6 +78,7 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
   truncateTitle = 0,
   tooltipPosition,
   customIcon,
+  isExpandable = false,
   onMouseEnter = () => {},
   onMouseLeave = () => {},
   ...props
@@ -138,6 +143,11 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
     dismissed && onTimeout();
   }, [dismissed]);
 
+  const [isExpanded, setIsExpanded] = useState(false);
+  const onToggleExpand = () => {
+    setIsExpanded(!isExpanded);
+  };
+
   const myOnMouseEnter = (ev: React.MouseEvent<HTMLDivElement>) => {
     setIsMouseOver(true);
     setTimedOutAnimation(false);
@@ -169,6 +179,8 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
         styles.alert,
         isInline && styles.modifiers.inline,
         isPlain && styles.modifiers.plain,
+        isExpandable && styles.modifiers.expandable,
+        isExpanded && styles.modifiers.expanded,
         styles.modifiers[variant as 'success' | 'danger' | 'warning' | 'info'],
         className
       )}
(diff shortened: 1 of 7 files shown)


packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -178,6 +194,20 @@ export const Alert: React.FunctionComponent<AlertProps> = ({
       onMouseLeave={myOnMouseLeave}
       {...props}
     >
+      {isExpandable && (
+        <div className={css(styles.alertToggle)}>
+          <Button
+            variant={ButtonVariant.plain}
+            onClick={onToggleExpand}
+            aria-expanded={isExpanded}
+            aria-label={`Toggle ${ariaLabel} description`}
```
nicolethoen, line 203: We should not hardcode anything in english. So ideally, we'd expose a separate aria-label for the toggle button and use a well written value as the default value.

packages/react-core/src/components/Alert/examples/Alert.md
```diff
@@ -156,6 +156,129 @@ class InlineAlertVariations extends React.Component {
...
+          actionClose={<AlertActionCloseButton onClose={() => alert('Clicked the close button')} />}
+        />
+        <Alert isInline isPlain variant="success" title="Success alert title" />
+      </React.Fragment>
+    );
+  }
+}
+```
+
+### Expandable
```
nicolethoen, line 237: I think it's worth adding a note below the title here. Something like, "It is not recommended to use an expandable alert within a toast Alert group. In such a case, the Alert could timeout before users would have time to interact and view the entire Alert."

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -22,6 +24,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   variant?: 'success' | 'danger' | 'warning' | 'info' | 'default';
   /** Flag to indicate if the Alert is inline */
   isInline?: boolean;
+  /** Flag to indicate if the Alert is plain */
```
tlabaj, line 27: sentence case here.  `alert`

packages/react-core/src/components/Alert/Alert.tsx
```diff
@@ -50,11 +54,14 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
   /** Set a custom icon to the Alert. If not set the icon is set according to the variant */
   customIcon?: React.ReactNode;
+  /** Flag indicating that the Alert is expandable */
+  isExpandable?: boolean;
```
tlabaj, line 58: sentence case

mcoker:
@mcarrano from the [core plain alert issue](https://github.com/patternfly/patternfly/issues/4220), it just says:

> Unlike the full inline alert, it would not have other actions associated with it. See https://marvelapp.com/prototype/852jaj9/screen/75866265 for an example of what appearance and placement might look like.

So it says to exclude actions, but doesn't call out descriptions specifically. I think we built the examples based off of the designs, which do not include a description.

packages/react-core/src/components/Alert/Alert.tsx
mcoker: In core we give the toggle button `aria-label="Details"`, then give an `id` to both the toggle button and alert title, then use `aria-labelledby="[id of title] [id of toggle button]"`

mcoker (approved):
Other than @mcarrano's question about the inline examples and @nicolethoen's feedback about the expandable toggle `aria-label`, lgtm!!

mcarrano:
> So it says to exclude actions, but doesn't call out descriptions specifically. I think we built the examples based off of the designs, which do not include a description.

I think that we definitely should remove actions from the examples, including the close/dismiss action. I guess we hadn't really considered the description but I think it could be useful to have a plain alert with a description. What do you think @mcoker @mceledonia ?

jessiehuff (dismissed):
I agree with Nicole's comments, but otherwise it looks good! I tested in VO, JAWS, and NVDA, and all seem to be able to navigate and understand everything in the alert. I do worry about future variations/uses of expandable alerts (like Nicole mentions), but these examples look great!

packages/react-core/src/components/Alert/Alert.tsx
CooperRedhat: I updated the aria label to be customizable and fallback on a default similar to `AlertActionCloseButton`, also put the toggle button in its own component. So the main difference between the two buttons at this point is that the user can not specify their own callback for toggling, like the onClose action in  `AlertActionCloseButton`.

CooperRedhat:
> > So it says to exclude actions, but doesn't call out descriptions specifically. I think we built the examples based off of the designs, which do not include a description.
> 
> I think that we definitely should remove actions from the examples, including the close/dismiss action. I guess we hadn't really considered the description but I think it could be useful to have a plain alert with a description. What do you think @mcoker @mceledonia ?

I removed the actions from inline plain examples, let me know if it looks good, thanks! @mcarrano

packages/react-core/src/components/Alert/AlertToggleExpandButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  ...props
+}: AlertToggleExpandButtonProps) => {
+  const { title, variantLabel: alertVariantLabel } = React.useContext(AlertContext);
+  return (
+    <Button
+      variant={ButtonVariant.plain}
+      onClick={onToggleExpand}
+      aria-expanded={isExpanded}
+      aria-label={ariaLabel === '' ? `Toggle ${variantLabel || alertVariantLabel} alert: ${title}` : ariaLabel}
+      aria-labeledby={`${title} toggle`}
```
mcoker, line 33: fyi this attr is spelled `aria-labelledby`, and it takes the value of an `id` attribute of an element that labels this element (it can take multiple `id` values, if multiple elements label it). The value shouldn't be a string though. An `aria-label` may be enough here, or just whatever's consistent with where we're doing this in other places.

mcarrano (approved):
This looks good. Thanks for making the updates @CooperRedhat !

packages/react-core/src/components/Alert/AlertToggleExpandButton.tsx
CooperRedhat: Thanks, I was trying to get parity with core, but in this case we are not even exposing that `id`  to be customizable for the user, so I will just remove it.

tlabaj (commented):
Looks like you need to update your snapshots, otherwise it looks good!

kmcfaul (approved):

tlabaj (approved):

PR #6239, opened 2021-08-26
fix(code-editor): remove uneeded peer dep

### LINKED ISSUE #6240: Hard to install react-code-editor
Takes time to figure out correct monaco deps because there's an extra peer dep.

redallen:
**What**: Closes #6240. [react-monaco-editor](https://www.npmjs.com/package/react-monaco-editor) already depends on [monaco-editor](https://www.npmjs.com/package/monaco-editor), so we don't need to ask consumers to specify it manually.

**Additional issues**:

### DIFF
--- a/packages/react-code-editor/package.json
+++ b/packages/react-code-editor/package.json
@@ -37,7 +37,6 @@
     "tslib": "^2.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


evwilkin (approved):
LGTM

Issue #6243, opened 2021-08-27
Select: The select field dropdown shows the Create option even when there's no input


debsmita1:
**Describe the issue. What is the expected and unexpected behavior?**
When the Select field of type `multiple` (https://www.patternfly.org/v4/components/select#multiple) has no options initially and you add a new option for the first time the newly added option along with the Create option keeps showing till you create another option. 
Also, the field allows the user to create duplicate options.

Expected behavior: 
- The Create option shouldn't be shown if there's no input.
- Users shouldn't be allowed to add duplicate entries.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
make sure there are no options and try adding new options
![select](https://user-images.githubusercontent.com/22490998/131078254-06201565-8e45-4016-9871-3c4b9c6e20b3.gif)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
This is a bug. Couldn't come up with a work around

**What is your product and what release version are you targeting?**
OCP 4.9

PR #6248, opened 2021-08-30
fix(Select): fixed issues in Select typeaheadmulti

debsmita1:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/6243

- added string value to key instead of `0` for the `No Results found` & `Create <text>` Select options
- added check for duplicate entries. If the option already exists for the entered text then the Create option will not be shown

GIF:
select_fix

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -408,16 +408,26 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
     if (typeaheadFilteredChildren.length === 0) {
       !isCreatable &&
         typeaheadFilteredChildren.push(
-          <SelectOption isDisabled key={0} value={noResultsFoundText} isNoResultsOption />
+          <SelectOption isDisabled key="no-results" value={noResultsFoundText} isNoResultsOption />
         );
     }
     if (isCreatable && typeaheadInputValue !== '') {
       const newValue = typeaheadInputValue;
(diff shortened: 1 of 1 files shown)


nicolethoen (dismissed):
This looks sound to me!
I'm always nervous with updates to the typeahead select, so I tagged @tlabaj and @kmcfaul who are also familiar with the typeahead select behavior/logic. I just want to make sure there isn't a reason I'm missing to check that props exist on each typeaheadFilteredChildren?
Also, @mcarrano, I just wanted you to double check this behavior makes sense to you and make you aware of this update :)

mcarrano (commented):
I agree with this change, however when testing it, I found that the check for an exact string match is case is case sensitive. See below:

Screen Shot 2021-08-31 at 10 37 01 AM

In this case, I did not expect to need to exactly type "Florida" as filtering the list is not case sensitive.

debsmita1:
> I agree with this change, however when testing it, I found that the check for an exact string match is case is case sensitive. See below:
> 
> Screen Shot 2021-08-31 at 10 37 01 AM
> 
> In this case, I did not expect to need to exactly type "Florida" as filtering the list is not case sensitive.

Fixed
fix-case

mcarrano (approved):
Looks good. Thanks @debsmita1 !

nicolethoen (approved):

kmcfaul (approved):
Looks good!

tlabaj (approved):

PR #6253, opened 2021-09-01
feat(Nav): add flyout

### LINKED ISSUE #6161: Navigation: add Flyout variation
Add new flyout variation to the Navigation component.

Core issue: https://github.com/patternfly/patternfly/issues/4245
Designs: https://www.figma.com/proto/AJi4XmCkrdGEK532YmbAqH/tertiary-nav?node-id=0%3A2&scaling=min-zoom&page-id=0%3A1

kmcfaul:
**What**: Closes #6161

In this PR:
- Adds `subnav` variant to `Nav`
- Adds `flyout` property to `NavItem`. `flyout` is a full `Nav` component with the `subnav` variant.

Still have some questions about architecture and interaction so marking as draft for now.

### DIFF
--- a/packages/react-core/src/components/Nav/Nav.tsx
+++ b/packages/react-core/src/components/Nav/Nav.tsx
@@ -35,7 +35,7 @@ export interface NavProps
   /** Indicates which theme color to use */
   theme?: 'dark' | 'light';
   /** For horizontal navs */
-  variant?: 'default' | 'horizontal' | 'tertiary' | 'horizontal-subnav';
+  variant?: 'default' | 'horizontal' | 'tertiary' | 'horizontal-subnav' | 'subnav';
 }
 
 export const NavContext = React.createContext<{
@@ -55,9 +55,15 @@ export const NavContext = React.createContext<{
   onToggle?: (event: React.MouseEvent<HTMLButtonElement>, groupId: number | string, expanded: boolean) => void;
   updateIsScrollable?: (isScrollable: boolean) => void;
   isHorizontal?: boolean;
+  flyoutRef?: React.Ref<HTMLLIElement>;
+  setFlyoutRef?: (ref: React.Ref<HTMLLIElement>) => void;
+  disableHover?: boolean;
 }>({});
 
-export class Nav extends React.Component<NavProps, { isScrollable: boolean; ouiaStateId: string }> {
+export class Nav extends React.Component<
+  NavProps,
+  { isScrollable: boolean; ouiaStateId: string; flyoutRef: React.Ref<HTMLLIElement> | null; disableHover: boolean }
+> {
   static displayName = 'Nav';
   static defaultProps: NavProps = {
     onSelect: () => undefined,
@@ -68,7 +74,9 @@ export class Nav extends React.Component<NavProps, { isScrollable: boolean; ouia
 
   state = {
     isScrollable: false,
-    ouiaStateId: getDefaultOUIAId(Nav.displayName, this.props.variant)
+    ouiaStateId: getDefaultOUIAId(Nav.displayName, this.props.variant),
+    flyoutRef: null as React.Ref<HTMLLIElement>,
+    disableHover: false
   };
 
   // Callback from NavItem
@@ -116,6 +124,7 @@ export class Nav extends React.Component<NavProps, { isScrollable: boolean; ouia
       ...props
     } = this.props;
     const isHorizontal = ['horizontal', 'tertiary'].includes(variant);
+    const Component = variant === 'subnav' ? 'section' : 'nav';
 
     return (
       <NavContext.Provider
@@ -136,12 +145,16 @@ export class Nav extends React.Component<NavProps, { isScrollable: boolean; ouia
           onToggle: (event: React.MouseEvent<HTMLButtonElement>, groupId: number | string, expanded: boolean) =>
             this.onToggle(event, groupId, expanded),
           updateIsScrollable: (isScrollable: boolean) => this.setState({ isScrollable }),
-          isHorizontal: ['horizontal', 'tertiary', 'horizontal-subnav'].includes(variant)
+          isHorizontal: ['horizontal', 'tertiary', 'horizontal-subnav'].includes(variant),
+          flyoutRef: this.state.flyoutRef,
+          setFlyoutRef: flyoutRef => this.setState({ flyoutRef }),
+          disableHover: this.state.disableHover
         }}
       >
-        <nav
+        <Component
           className={css(
-            styles.nav,
+            variant !== 'subnav' && styles.nav,
+            variant === 'subnav' && styles.navSubnav,
             theme === 'light' && styles.modifiers.light,
             isHorizontal && styles.modifiers.horizontal,
             variant === 'tertiary' && styles.modifiers.tertiary,
@@ -154,7 +167,7 @@ export class Nav extends React.Component<NavProps, { isScrollable: boolean; ouia
           {...props}
         >
           {children}
-        </nav>
+        </Component>
       </NavContext.Provider>
     );
   }
--- a/packages/react-core/src/components/Nav/Nav.tsx
+++ b/packages/react-core/src/components/Nav/Nav.tsx
@@ -57,12 +57,11 @@ export const NavContext = React.createContext<{
   isHorizontal?: boolean;
   flyoutRef?: React.Ref<HTMLLIElement>;
   setFlyoutRef?: (ref: React.Ref<HTMLLIElement>) => void;
-  isHoverDisabled?: boolean;
 }>({});
 
 export class Nav extends React.Component<
   NavProps,
-  { isScrollable: boolean; ouiaStateId: string; flyoutRef: React.Ref<HTMLLIElement> | null; isHoverDisabled: boolean }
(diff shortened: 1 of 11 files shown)


kmcfaul:
- @nicolethoen @tlabaj Should `subnav` (subnav css with `section` element instead of `nav`)  be a variant of `Nav` or a separate/new component? Currently this section structure exists internally only inside `NavExpandable`.
- @jessiehuff `Nav` currently uses `tab` to traverse options, so I've kept that as the main navigation method. Currently, `Space`/`ArrowRight` will open a flyout (letting a user tab to the sub menu) while `Escape`/`ArrowLeft` will close a flyout and focus the parent element. `Enter` is already used to select/click a link so it does nothing for a flyout. Does this make sense for keyboard navigation?
- @redallen @mcoker @mattnolting  I lifted the `useIsomorphicLayoutEffect` from `MenuItem` and it appears to be applying the classes and styles correctly but the nav flyouts aren't properly bouncing back when the screen is too thin. Am I missing something logic wise or is it just missing styles/support for a style? The `pf-m-left` modifier doesn't exist on `nav` css but using the menu style isn't working.

mcarrano (commented):
@kmcfaul this is looking good. Have we considered how this would behave on a touch device? Can I open the menus either on hover or tap/click?

mattnolting:
hover

- The hover behavior doesn't seem to be working (even on large screens). 

> @redallen @mcoker @mattnolting I lifted the useIsomorphicLayoutEffect from MenuItem and it appears to be applying the classes and styles correctly but the nav flyouts aren't properly bouncing back when the screen is too thin.

@kmcfaul There's is no current support, that I now of, for flyouts in mobile devices or narrow containers

kmcfaul:
The flyout behavior currently matches Menu's flyout. What is the expected behavior for Nav?

mcarrano:
@kmcfaul Sorry to take a while to get back to you. I think I actually flyouts better when they stick and you need to click away from them to get them to dismiss. That's how the flyout menu works. Or if we are going to keep the behavior of dismissing on mouse_leave, then there should be a delay so they don't dismiss immediately when the user slips off the menu. Thoughts?

kmcfaul:
@mcarrano Figured out a click listener that works. Flyout state is managed internally so it's a listener per flyout though - not sure if that impacts performance much.

I tried adding a timeout to the `onMouseLeave` but was having trouble getting the timeout to cancel properly if the mouse moved back in (so the flyout would still close / have inconsistent behavior when mousing in and out rapidly). Since I added in the click handler I removed the `onMouseLeave` for now. `onMouseOver` still will toggle a flyout if mousing between items inside the flyouts.

kmcfaul:
@nicolethoen Any thoughts why a table composable cypress is failing? I don't think my changes should have impacted the tooltip in that test.

mcarrano (commented):
The mouse interaction feels right now, @kmcfaul . Not sure about the keyboard interaction however. Looks like I need to use Tab to navigate the menu? Not the arrorw keys. Right arrow to open works as expected, but if I use the left arrow to close the flyout, it seems to close all levels and put focus back on the topmost node. Is that expected?

@jessiehuff would be interested in your take. I'm not quite sure what to expect. Thanks.

kmcfaul:
Left arrow should take you back to the top of the immediate outer flyout (tabbing out to Link 7 and pressing left arrow works). 

Regarding `tab` I wasn't sure if we wanted to implement arrow navigation or not, because `Nav` doesn't use arrow keys right now.

kmcfaul:
@nicolethoen I fixed the failing cypress test (TableComposable - tooltip was being found) but it shouldn't be related to my changes so I'm not sure why it was suddenly failing? I'm not sure if cypress was just waiting for the tooltip to render before moving on to the tests, because forcing the event makes it work.

mattnolting (dismissed):
Looks great, thanks @kmcfaul!

packages/react-core/src/components/Nav/Nav.tsx
```diff
@@ -55,9 +55,15 @@ export const NavContext = React.createContext<{
   onToggle?: (event: React.MouseEvent<HTMLButtonElement>, groupId: number | string, expanded: boolean) => void;
   updateIsScrollable?: (isScrollable: boolean) => void;
   isHorizontal?: boolean;
+  flyoutRef?: React.Ref<HTMLLIElement>;
+  setFlyoutRef?: (ref: React.Ref<HTMLLIElement>) => void;
+  disableHover?: boolean;
```
tlabaj, line 60: Maybe `isHoverDisabled` for naming consistency.

packages/react-core/src/components/Nav/NavItem.tsx
```diff
@@ -39,12 +54,148 @@ export const NavItem: React.FunctionComponent<NavItemProps> = ({
...
+            flyoutMenu.style.setProperty(rightOffset.name, `-${xOffset}px`);
+          } else {
+            flyoutMenu.style.setProperty(leftOffset.name, `-${xOffset}px`);
+          }
+
+          const spaceLeftBot = window.innerHeight - origin.y - rect.height;
+          const spaceLeftTop = window.innerHeight - rect.height;
+          if (spaceLeftTop < 0 && spaceLeftBot < 0) {
+            // working idea: page can usually scroll down, but not up
+            // TODO: proper scroll buttons
```
tlabaj, line 114: do we need issue for this?

packages/react-core/src/components/Nav/NavItem.tsx
```diff
@@ -39,12 +54,148 @@ export const NavItem: React.FunctionComponent<NavItemProps> = ({
...
+    if (!flyoutVisible && show) {
+      setFlyoutRef(ref);
+    } else if (flyoutVisible && !show) {
+      setFlyoutRef(null);
+    }
+    onShowFlyout && show && onShowFlyout();
+  };
+
+  useIsomorphicLayoutEffect(() => {
+    if (hasFlyout && ref.current && canUseDOM) {
```
tlabaj, line 82: how similar is this to the menu?  is there any way to share some come functionality in s utility?  If so we can follow up on that.

tlabaj (changes requested):
Can we add to demo and have a integration test.

tlabaj:
To me the styling looks a little bit off?  The flyout seems to overlap the menu that triggers it more so the in the core example.

packages/react-core/src/components/Nav/NavItem.tsx
kmcfaul: This is also in our Menu flyouts, I don't know if we have a feature to track it yet

packages/react-core/src/components/Nav/NavItem.tsx
kmcfaul: This is the same function from Menu (determining when to flip the flyout direction), but it doesn't work quite right for Nav at the moment. The styles it applies to flip the direction left doesn't work with the Nav item. We could make this a utility, or I could remove it for now.

redallen: +1 for keeping flyout code DRY. It was very difficult to write the first time around and I think it hard enough to maintain a single copy.

Maybe a set of our own `useFlyout` hooks would work?

kmcfaul:
@tlabaj It appears to overlap more because the react demo's background container div (which sets the color) is slightly wider than the nav due to padding. The actual positioning should be the same as core - it's shown better with the second flyout containing "Link 7".

packages/react-core/src/components/Nav/Nav.tsx
kmcfaul: Checking again, this prop is used internally by Menu to disable mouseover when some keyboard behavior kicks in. I copied over the whole context initially for Nav, but I actually can probably remove this prop for now since the value never changes.

packages/react-core/src/components/Nav/NavItem.tsx
kmcfaul: The classes applied would be different (menu styles for Menu, currently nothing for Nav) but I can abstract the logic of the function into a util

kmcfaul: Or can you pass properties to hooks directly?

redallen: You can pass whatever you want to custom hooks! See https://reactjs.org/docs/hooks-custom.html

kmcfaul: I'm half inclined to remove the whole useEffect for Nav though, because it doesn't actually work currently. Opinions?

redallen: Less code is better!

kmcfaul:
Removing isomorphicLayoutEffect from Nav for now and opened #6296 to follow up. I think it will need some eyes from core as well as react to get it working properly like it does for Menu.

jessiehuff (commented):
This looks pretty good! I think the only thing that seemed odd to me is that when I use the arrow keys to open the subnav, the focus doesn't automatically shift, and I have to press Tab to get into it. It surprised me to need two keys to go into the subnav. Is it possible to shift the focus automatically so it just requires one key press with the left/right arrow keys?

tlabaj (approved):
LGTM

kmcfaul:
@jessiehuff Added the focus handling in!

jessiehuff (approved):

tlabaj (approved):

PR #6254, opened 2021-09-01
fix(Select): filter out Dividers when using typeahead

### LINKED ISSUE #6005: bug(Select): noResultsFoundText will never show in the typeahead variant when one of the options is a Divider component
Issue as described in the tittle. The empty state text should be shown if none of the options match, regardless if divider is used.

https://codesandbox.io/s/competent-cherry-kiiqi?file=/index.js

bug

Cockpit, no workaround.

michpetrov:
**What**: Closes #6005 

Added a method to remove superfluous Dividers when using typeahead, i.e. it removes all Dividers from start and end and replaces consecutive Dividers with a single one. In case there's only Dividers the result will be empty, which fixes the reported issue.


kmcfaul (approved):

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -400,6 +400,9 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                 }
               })
             : childrenArray;
+        if (typeaheadInputValue.toString() !== '') {
```
nicolethoen, line 403: On line 386 there is a comment about not filtering out dividers. Could that be removed and the logic be modified there to filter out dividers?

michpetrov: I think it would be very convoluted, inside the `filter` callback I cannot tell which of preceding elements have been filtered out.

nicolethoen: I dont think it's more convoluted? 
If you change the comment on line 386 to say `// Dividers don't have value and SHOULD be filtered`
and then change the return statement on line 388 to `return false`

then likely you wont need another whole function call after having already iterated over every child.

nicolethoen (changes requested):

packages/react-core/src/components/Select/Select.tsx
michpetrov: Well, yes but I want to keep the dividers in, just not all of them.
I.e. instead of the filtered result looking like `abc | | abd | | | abx |` (current implementation) 
I want it to look like `abc | abd | abx`.

nicolethoen: @michpetrov I see - yeah I did find it a little confusing given the comments a few lines up. Maybe you could update the comment to say something like 'filtering out dividers is handled after determining which values are filtered out. And then add a comment before your new `if` statement to say something like 'filter out unnecessary dividers'

Issue #5910, opened 2021-06-11
bug(DataList): the draggable version of data list flickers when moving rows and these are of different size 


KKoukiou:
https://user-images.githubusercontent.com/10246/121147678-60f48880-c841-11eb-8369-fae17c78a25e.mp4

When dragging rows, over anything more than 1 row high, there's an obscene amount of flicker .

KKoukiou:
Friendly ping ;)

garrett:
Still a bug.

The dual list selector is getting a new drag and drop implementation @ https://github.com/patternfly/patternfly-react/pull/6258 which should be backported to this data list component, according to https://github.com/patternfly/patternfly-react/issues/6265#issuecomment-920871507... so a fix should be in the works.

KKoukiou:
This is fixed with the new implementation with Draggable.

Issue #5900, opened 2021-06-09
bug(DataList): the drag/move feature is not working at all on mobile on firefox


KKoukiou:
**Describe the issue. What is the expected and unexpected behavior?**

The draggable data list does not seem work on mobile using Firefox. Tested on Firefox Nightly, Android 11.
On Chrome, if you touch and hold for a while, it works... but the dragged item is not shown (blank rectangle)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug, no workaround.

**What is your product and what release version are you targeting?**

Cockpit Virtual Machines plugin.
This is one of the two components that are still blocking us from completely dropping PF3 from our dependencies.

KKoukiou:
@tlabaj Can someone look at this? This component is the last remaining in cockpit to port from PF3 to PF4 :) Plus it's a serious bug.

tlabaj:
Hi @KKoukiou . There are a number of issues open with draggable feature.  We are looking at them and and investigating solutions.

KKoukiou:
Friendly reminder :) Would be super nice to be able to drop this so that we can also drop completely PF3.

redallen:
It's fixed with the new suggested `<Draggable>` implementation: http://patternfly-react.surge.sh/components/data-list#draggable

![dragdrop-firefox](https://user-images.githubusercontent.com/47335686/136436194-49848e99-aaea-4689-b41f-8d736c1094af.gif)

The deprecated HTML5-based API may still be broken, but that's why it's deprecated. Closing!

Issue #5067, opened 2020-10-27
Draggable data list bugs


mcoker:
Looks like you can click and drag from anywhere in a row to drag it. The drag start should only fire if you click on the draggable button on the left. 

Also when the page loads, if I quickly drag a row to the right (keep the row in its place, not in another row slot) and let go, the whole list disappears.

cc @kmcfaul

PR #6258, opened 2021-09-02
Add Drag and Drop

### LINKED ISSUE #6167: Introduce third party library to support drag and drop behavior
Want to introduce [react-beautiful-dnd](https://react-beautiful-dnd.netlify.app/?path=/story/*) for drag and drop support.

### LINKED ISSUE #6257: DraggableTable missing import
Reported in Slack.

ReferenceError: styles is not defined 

https://codesandbox.io/s/djrp6?file=/index.js:829-895

redallen:
**What**: Reimplement the parts we're interested in from react-beautiful-dnd while applying our own styling. Closes #6167

What this does:
- Adds `DragDrop`, `Draggable`, and `Droppable` components
- Animates dropping in invalid zone
- Add styles for hovering over invalid zone
- Animates items moving out of the way for draggable item
- Allow draggable item to be dropped into other lists
- Natural movement offset by where item is picked up
- Provides drop API for moving dropped items

**Additional issues**: Will need integration tests and to be used in Draggable datalist

### DIFF
--- a/packages/react-core/src/components/DragDrop/DragDrop.tsx
+++ b/packages/react-core/src/components/DragDrop/DragDrop.tsx
@@ -8,22 +8,37 @@ export const DragDropContext = React.createContext({
   draggingZone: '',
   setDraggingZone: (_zone: string) => {},
   draggableZoneNodes: {} as ZoneNodes,
-  setDraggableZoneNodes: (_nodes: ZoneNodes) => {}
+  setDraggableZoneNodes: (_nodes: ZoneNodes) => {},
+  onDrop: (
+    _sourceDroppableIndex: number,
+    _sourceDraggableIndex: number,
+    _destDroppableIndex: number,
+    _destDraggableIndex: number
+  ) => {}
 });
 
 interface DragDropProps {
   /** Potentially Droppable and Draggable children */
   children?: React.ReactNode;
+  /** Callback for drop event */
+  onDrop?: (
+    sourceDroppableIndex: number,
+    sourceDraggableIndex: number,
+    destDroppableIndex: number,
+    destDraggableIndex: number
+  ) => void;
 }
 
-export const DragDrop: React.FunctionComponent<DragDropProps> = ({ children }: DragDropProps) => {
+export const DragDrop: React.FunctionComponent<DragDropProps> = ({ children, onDrop = () => {} }: DragDropProps) => {
   // Used for highlighting Droppable
   const [draggingZone, setDraggingZone] = React.useState('');
   // Used for reordering
   const [draggableZoneNodes, setDraggableZoneNodes] = React.useState({});
 
   return (
-    <DragDropContext.Provider value={{ draggingZone, setDraggingZone, draggableZoneNodes, setDraggableZoneNodes }}>
+    <DragDropContext.Provider
+      value={{ draggingZone, setDraggingZone, draggableZoneNodes, setDraggableZoneNodes, onDrop }}
+    >
       {children}
     </DragDropContext.Provider>
   );
--- a/packages/react-core/src/components/DragDrop/DragDrop.tsx
+++ b/packages/react-core/src/components/DragDrop/DragDrop.tsx
@@ -1,8 +1,14 @@
 import * as React from 'react';
 
+interface ZoneNodes {
+  [zone: string]: HTMLElement[];
+}
+
 export const DragDropContext = React.createContext({
   draggingZone: '',
-  setDraggingZone: (_zone: string) => {}
+  setDraggingZone: (_zone: string) => {},
+  draggableZoneNodes: {} as ZoneNodes,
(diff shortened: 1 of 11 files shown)


andrew-ronaldson (approved):
LGTM. Thanks!

mcarrano (commented):
@redallen I'm a little unclear what the scope of this PR is. Should I be seeing all the affordances shown in the design examples here including the drag (waffle) icon? Or is that added on top of this when applied to a specific component like a list? Said another way, do we consider drag & drop a component or just a capability that can be applied to coding interactions in other components?

cc @tlabaj @nicolethoen

andrew-ronaldson:
@mcarrano the icon helps let the user know it is moveable so I agree. I wasn't sure if that was being added with the style stuff later.

redallen:
> I'm a little unclear what the scope of this PR is.

The scope is to add generic Draggable and Droppable components to:
1. Be used by PatternFly internally to implement future drag and drop patterns (which we are now duplicating over our code base) and to do things like reimplement  Draggable data list. This will make our library more consistent and maintainable (see #5067, #6265, #6265, #5900, #5910).
2. Support dragging/dropping between **different** lists (can't quickly find the issue, but this has been requested in Slack).
3. Potentially in the future for consumers to use in their own components.

> Should I be seeing all the affordances shown in the design examples here including the drag (waffle) icon?

No. This is not a Data list example. We previously agreed in our design meeting you don't need a waffle icon for all drag and drop interactions (like table rows). When implementing these components I'm only worried about the simplest use case.

> Or is that added on top of this when applied to a specific component like a list?

If we release this component to the public it'd be great to give some design guidelines that specify when to add a drag icon.

> Said another way, do we consider drag & drop a component or just a capability that can be applied to coding interactions in other components?

That's an excellent question. The way we add complex "capability" in React is by creating components. In this case I've decided to create three components:

1. `<DragDrop>` which holds some state and provides you an `onDrop` callback,
2. `<Droppable>` to represent the zone you can drop things in (and display the border per design), and
3. `<Draggable>` where most of the logic lives for how an item should look when being dragged around between lists (complete with buttery animations per our design meeting).

I'd argue for this to be released as a component because I believe it provides significant value-add over alternative 3rd-party libraries. However, I'd like to pose the question of **why are we designing _PatternFly_ drag and drop interactions?** If we really want Red Hat products to have consistent interaction patterns reordering lists (and not just lists made up of existing PatternFly components) we really should release this as a component for consumers to use in their own products.

packages/react-core/src/components/DragDrop/DragDrop.tsx
```diff
@@ -0,0 +1,40 @@
...
+    _sourceDraggableKey: string | number,
+    _destDroppableKey: string | number | null,
+    _destDraggableKey: string | number | null
+  ) => false as boolean
+});
+
+interface DragDropProps {
+  /** Potentially Droppable and Draggable children */
+  children?: React.ReactNode;
+  /** Callback for drop event */
```
nicolethoen, line 21: Could you give a little more information about what all this function needs to do/what the various props are for? 
I think the example implementation could include comments for more insight, but this prop description could be a little more specific as well.

packages/react-core/src/components/DragDrop/DragDrop.tsx
```diff
@@ -0,0 +1,40 @@
...
+    _sourceDroppableKey: string | number,
+    _sourceDraggableKey: string | number,
+    _destDroppableKey: string | number | null,
+    _destDraggableKey: string | number | null
+  ) => false as boolean
+});
+
+interface DragDropProps {
+  /** Potentially Droppable and Draggable children */
+  children?: React.ReactNode;
```
nicolethoen, line 20: Are there any rules for the relationship between draggable and droppable zones? they each have their own keys - and the zone prop is used to enforce some sort of rule. I think I see how you implemented it in the basic examples, but I am having a hard time comprehending the limits or other ways I might use them.

packages/react-core/src/components/DragDrop/DragDrop.tsx
redallen: It's a callback to actually move the elements in whatever the source of truth is.

I'm thinking of using indexes instead of ids since the first step to moving elements is almost always finding their index.

packages/react-core/src/components/DragDrop/DragDrop.tsx
redallen: `Draggable`s go in `Droppable`s. `Droppable`s go in `DragDrop`s.

nicolethoen: Love it - can all that be explain in docs somehow?

nicolethoen (approved):

nicolethoen (approved):

tlabaj (approved):

PR #6261, opened 2021-09-02
Add Support for disabled state tabs

### LINKED ISSUE #6162: Add disabled state for Tabs
Want to support a disabled state for tabs with the ability to attach a tooltip.

Core issue and designs: https://github.com/patternfly/patternfly/issues/4257

CooperRedhat:
**What**: Closes #6162 

**Additional issues**:
- Refactored Tab into it's own component in order to forwardRef for `Tooltip` in `aria-disabled` case re: [Original Request](https://github.com/patternfly/patternfly-design/issues/1066)
- Added disabled state examples to match core PR for same issue

### DIFF
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -18,6 +18,12 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   tabContentRef?: React.RefObject<any>;
   /** whether to render the tab or not */
   isHidden?: boolean;
+  /** Adds disabled styling and disables the button using the disabled html attribute */
+  isDisabled?: boolean;
+  /** Adds disabled styling and communicates that the button is disabled using the aria-disabled html attribute */
+  isAriaDisabled?: boolean;
+  /** Events to prevent when the button is in an aria-disabled state */
+  inoperableEvents?: string[];
 }
 
 /** The parent <Tabs> component accecesses this component's propeties directly in order to present each Tab */
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -61,6 +61,14 @@ const TabBase: React.FunctionComponent<TabProps> = ({
   if ((mountOnEnter || unmountOnExit) && eventKey !== localActiveKey) {
     ariaControls = undefined;
   }
+  const isButtonElement = href !== null;
+  const getDefaultTabIdx = () => {
+    if (isDisabled) {
+      return isButtonElement ? null : -1;
+    } else if (isAriaDisabled) {
+      return -1;
+    }
+  };
   return (
     <li
       className={css(styles.tabsItem, eventKey === localActiveKey && styles.modifiers.current, childClassName)}
@@ -72,8 +80,9 @@ const TabBase: React.FunctionComponent<TabProps> = ({
           isDisabled && href && styles.modifiers.disabled,
           isAriaDisabled && styles.modifiers.ariaDisabled
         )}
-        disabled={isDisabled}
-        aria-disabled={isAriaDisabled}
+        disabled={isButtonElement ? isDisabled : null}
+        aria-disabled={!isButtonElement && isAriaDisabled}
+        tabIndex={getDefaultTabIdx()}
         onClick={(event: any) => handleTabClick(event, eventKey, tabContentRef)}
         {...(isAriaDisabled ? preventedEvents : null)}
         id={`pf-tab-${eventKey}-${childId || uniqueId}`}
--- a/packages/react-core/src/components/Tabs/Tab.tsx
+++ b/packages/react-core/src/components/Tabs/Tab.tsx
@@ -61,12 +61,12 @@ const TabBase: React.FunctionComponent<TabProps> = ({
   if ((mountOnEnter || unmountOnExit) && eventKey !== localActiveKey) {
     ariaControls = undefined;
   }
-  const isButtonElement = href !== null;
+  const isButtonElement = Boolean(!href);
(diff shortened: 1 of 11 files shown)


mcarrano (dismissed):
Looks good to me. Thanks @CooperRedhat !

nicolethoen (commented):
LGTM After a11y tests pass

packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -18,8 +22,72 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
...
+      ref={innerRef}
+    >
+      <TabButton
+        className={css(
+          styles.tabsLink,
+          isDisabled && href && styles.modifiers.disabled,
+          isAriaDisabled && styles.modifiers.ariaDisabled
+        )}
+        disabled={isDisabled}
+        aria-disabled={isDisabled || isAriaDisabled}
```
mcoker, line 76: It doesn't hurt anything, but you don't need `aria-disabled="true"` when it's a `button` with the `disabled` attribute.

mcoker (changes requested):
Looks good! Can you also add an example of when a tab button is an `<a>` (in the nav element examples - https://patternfly-react-pr-6261.surge.sh/components/tabs#using-the-nav-element)? Here are the core examples - https://pf4.patternfly.org/components/tabs#using-the-nav-element

packages/react-core/src/components/Tabs/Tab.tsx
CooperRedhat: Thanks for pointing that out, fixed it and added disabled state to nav examples

CooperRedhat:
Note: Nav examples with anchor tags/hrefs were not updating active state, added preventDefault to click handler (which is what jumplinks do), should be working now.

mcoker (changes requested):
Just one change- with a disabled nav tab/link:
* Replace the `disabled` attribute with `aria-disabled="true"` (`disabled` isn't valid on an `<a>`)
* Add `tabindex="-1"` to remove it from the tab order

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -303,46 +312,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
...
-                    tabContentRef={tabContentRef}
-                    ouiaId={childOuiaId}
-                    {...rest}
-                  >
-                    {title}
-                  </TabButton>
-                </li>
-              );
-            })}
+            {filteredChildren}
```
kmcfaul, line 315: I'm a little confused why this mapping is no longer needed, did the structure of `filteredChildren` change?

kmcfaul (commented):
Looks like the examples have some duplicate ids that are causing a11y tests to fail.

packages/react-core/src/components/Tabs/Tabs.tsx
CooperRedhat: The tab component used to return null, i.e. render nothing. So filtered children was just an array of objects. I needed a ref on the tab component for tooltips, so I had to forward ref and render an instance to attach a ref to.

CooperRedhat:
> Looks like the examples have some duplicate ids that are causing a11y tests to fail.

Looks like the duplicate IDs came from wrapping a tab in a tooltip, we'll see if it passes.

nicolethoen (dismissed):

mcoker (changes requested):
Screen Shot 2021-09-10 at 1 52 42 PM

For regular disabled nav links:
* Remove `disabled`
  * An `<a>` doesn't support the `disabled` attr.
* Add `aria-disabled="true"`
  * Since the browser doesn't disable a link for you, you have to tell screen readers its disabled
* Add `tabindex="-1"`
  * Since its disabled, you don't want users to be able to tab to it (just like tab buttons that are `disabled`)

Screen Shot 2021-09-10 at 1 53 12 PM

For aria-disabled tab links:
* Add `aria-disabled="true"`
  * Same as above
* Remove the `tabindex` attr
  * You want users to be able to tab to it to trigger the popover/tooltip

tlabaj (changes requested):
Can you had a demo and some integration test to verify the disables tabs?

jessiehuff (commented):
So in this scenario, I don't think we actually want to add a `tabIndex="-1"` to the tab because it will make the tooltip completely inaccessible by keyboard. The weird thing about `aria-disabled` elements is they're kind of like a combination of disabled and interactive elements. Adding a tooltip makes it kind of interactive, but the main action of the tab is disabled. However, by playing with the preview, even when I take the `tabIndex` off the tooltip isn't displaying on focus so I'm not positive what's going on there. I also noticed that the tooltip isn't getting announced in VO. I think we need an `aria-describedby="whatever the tooltip's id is"`. A good example to follow is our [aria-disabled button with tooltip examples](https://www.patternfly.org/v4/components/button#aria-disabled-button-with-tooltip).

mcoker (dismissed):
LGTM!

tlabaj (approved):

mcoker (approved):

kmcfaul (approved):

jessiehuff (commented):
So I noticed in reviewing that the tooltip still isn't showing up on keyboard focus. I asked Nicole if she could test that aspect too, and she noticed that if you pressed enter the tooltip will appear. That's pretty unexpected that it would appear on hover for mouse users but require interaction to appear for keyboard users. However, we also noticed that this happened with the tooltip on enabled tabs as well. Nicole came up with a good theory that it may have to do with a discrepancy between where the tooltip is put (on the li) and where the focus is (on a button). We could probably create a follow up issue to investigate.

Also, I think we only need the `tabIndex="0"` if there's something interactive on the disabled tab. We should probably create an issue to add this to our documentation as well.

PR #6196, opened 2021-08-19
fix(examples): updating example microcopy

evwilkin:
**What**: Closes #5714 

This PR updates microcopy across examples & demos as laid out in issue #5714

PR #6263, opened 2021-09-02
fix(datepicker): updated date picker microcopy

### LINKED ISSUE #6199: DatePicker - confirm correct microcopy & date format
Follow-up to [this discussion](https://github.com/patternfly/patternfly-react/pull/6196#discussion_r692265691) in https://github.com/patternfly/patternfly-react/pull/6196.

Confirm with content & design teams which date formatting should be used and update DatePicker microcopy and potentially design guidelines as needed.

evwilkin:
**What**: Closes #6199 

Follow up to #6196 

Per discussion with @abigaeljamie, this PR capitalizes the placeholder text for date picker examples and replaces slashes in American format example with dashes (03-05-2020 vs 03/05/2020).


redallen:
If we want to change the placeholder across the board I think we should change the prop's default in `DatePicker.tsx` so consumers see the change as well.

Then you'd just have to adjust the `toLocaleDateString` American format from slashes (what design asked for) to dashes.

redallen (approved):

abigaeljamie (commented):
Thank you! Looks great :)

Issue #5967, opened 2021-06-28
Layouts should support component prop


mcoker:
https://github.com/patternfly/patternfly/pull/4010

We added some demos of layouts as lists, but any (bullseye is an exception) should support different elements - `<section>`, `<article>`, etc.

mcoker:
not stale

PR #6270, opened 2021-09-06
Add component property to Flex, Gallery and Grid and the subitems. 

KKoukiou:
Relevant to #5967

### DIFF
--- a/packages/react-core/src/layouts/Flex/Flex.tsx
+++ b/packages/react-core/src/layouts/Flex/Flex.tsx
@@ -334,11 +334,14 @@ export interface FlexProps extends React.HTMLProps<HTMLDivElement> {
     xl?: string;
     '2xl'?: string;
   };
+  /** Sets the base component to render. defaults to div */
+  component?: React.ElementType<any> | React.ComponentType<any>;
 }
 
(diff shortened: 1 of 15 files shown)


nicolethoen (dismissed):
code looks good - just want to check with pf core and @mcoker that this doesn't have unforeseen consequences.

mcoker (dismissed):
LGTM! Thanks @KKoukiou. My only comment is I'm not sure what the examples should show - to use `<List/ListItem>` as the components or just `ul/li`. These 2 examples look a little weird, though it's a valid use case. 

Screen Shot 2021-09-09 at 9 47 11 AM
Screen Shot 2021-09-09 at 9 47 03 AM

KKoukiou:
@mcoker I happy to change the example code if we have any other idea here! I think most people will use directly the string though, like 'ul' and not specify the React PF component name.

mcoker:
@KKoukiou cool! I think the string would be better. That will match the core examples, too.

kmcfaul (approved):

nicolethoen (approved):

mcoker (approved):
🥳

PR #6271, opened 2021-09-06
fix(SelectToggle): use aria-label prop

MariaAga:
When the select toggle isn't for typeahead the aria-label prop is ignored. So when using `toggleAriaLabel` in the `Select` component that prop is ignored.


tlabaj:
Can you please open an issue for this one for tracking purposes.

MariaAga:
Closes https://github.com/patternfly/patternfly-react/issues/6284

nicolethoen (approved):

kmcfaul (approved):

tlabaj (approved):

jessiehuff (approved):
Nice catch!! 😄  🎉

PR #6274, opened 2021-09-07
docs: refactor demos to use Masthead

### LINKED ISSUE #6174: Refactor DashboardWrapper, Page demos to use Masthead (over PageHeader)
DashboardWrapper and full page demos that still include PageHeader should be refactored to use the Masthead component instead.

kmcfaul:
**What**: Closes #6174

Refactors:
- [x] DashboardHeader
- [ ] Page
- [ ] Nav
- [x] NotificationDrawer
- [x] Wizard

`PageHeader` is integrated with a `Page` context to provide the managed nav toggle trigger / aria-expanded state. To maintain this functionality I added a `PageToggleButton` component that extends our `Button` component and contains the context consumer.

### DIFF
--- a/packages/react-core/src/demos/NotificationDrawer/NotificationDrawer.md
+++ b/packages/react-core/src/demos/NotificationDrawer/NotificationDrawer.md
@@ -5,14 +5,17 @@ section: components
 
 import BellIcon from '@patternfly/react-icons/dist/esm/icons/bell-icon';
 import CogIcon from '@patternfly/react-icons/dist/esm/icons/cog-icon';
+import BarsIcon from '@patternfly/react-icons/dist/js/icons/bars-icon';
 import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
 import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import imgBrand from '@patternfly/react-core/src/components/Brand/examples/pfLogo.svg';
 import imgAvatar from '@patternfly/react-core/src/components/Avatar/examples/avatarImg.svg';
+import AttentionBellIcon from '@patternfly/react-icons/dist/esm/icons/attention-bell-icon';
 
 ## Demos
 
 ### Basic
+
 ```js isFullscreen
 import React from 'react';
 import {
@@ -51,7 +54,6 @@ import {
   NotificationDrawerListItemBody,
   NotificationDrawerListItemHeader,
   Page,
-  PageHeader,
   PageSection,
   PageSectionVariants,
   PageSidebar,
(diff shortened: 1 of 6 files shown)


kmcfaul:
@tlabaj @nicolethoen 

For `Page` demos do we still want to use `PageHeader`? I'm unsure whether to refactor those. 

For `Nav` demos, I'm unsure where the actual nav component that is embedded in `PageHeader` is supposed to go within `Masthead` or the masthead's `Toolbar`. Placing the nav in the toolbar causes the bottom stacked masthead section to overflow on a small screen.

kmcfaul:
Opened #6281 for Page specific refactoring. Taking another look at `Nav`.

nicolethoen (approved):
LGTM

packages/react-core/src/demos/NotificationDrawer/NotificationDrawer.md
```diff
@@ -728,64 +785,87 @@ class GroupedNotificationDrawer extends React.Component {
...
-              dropdownItems={userDropdownItems}
-            />
-          </PageHeaderToolsItem>
-        </PageHeaderToolsGroup>
-        <Avatar src={imgAvatar} alt="Avatar image" />
-      </PageHeaderTools>
+    const headerToolbar = (
+      <Toolbar>
+        <ToolbarContent>
+          <ToolbarGroup alignment={{ default: 'alignRight' }}>
```
mattnolting, line 791: need to add spaceItems={{ default: 'spacerNone' }}

packages/react-core/src/demos/NotificationDrawer/NotificationDrawer.md
```diff
@@ -728,64 +785,87 @@ class GroupedNotificationDrawer extends React.Component {
...
+                  isPlain
+                  position="right"
+                  onSelect={this.onDropdownSelect}
+                  isOpen={isDropdownOpen}
+                  toggle={<DropdownToggle onToggle={this.onDropdownToggle}>John Smith</DropdownToggle>}
+                  dropdownItems={userDropdownItems}
+                />
+              </ToolbarItem>
+            </ToolbarGroup>
+            <Avatar src={imgAvatar} alt="Avatar image" />
```
mattnolting, line 849: Avatar should be part of the dropdown `https://www.patternfly.org/v4/components/dropdown#dropdown-with-image-and-text`

mattnolting (changes requested):
Looking good, @kmcfaul! In each of the examples, `avatar` should be inside the dropdown.

mattnolting (dismissed):
LPTM!

nicolethoen (approved):

kmcfaul:
Nav refactoring will need additional work - masthead currently doesn't quite support containing a nav. Putting the nav inside the toolbar prevents the scrolling/shrinking from behaving properly, and putting it outside the toolbar but also in the content yields most of the space to the toolbar instead of the nav, making it look very squished (but scroll buttons do work). I tried to get the alignment and spacing working with `Flex` but couldn't get all the way there. It looks like `PageHeader` used a grid layout with a separate container div, but MastheadContent is already a flex so I'm unsure if we can/want to nest those.

mcarrano:
@kmcfaul As we discussed the other day, I'm fine with keeping the Horizontal Nav demo (https://www.patternfly.org/v4/components/navigation#horizontal-only-in-pageheader) as is for now unless @mattnolting you can think of a solution to this problem. AFAIK this pattern is rarely used, if at all. @redallen are you able to tell if anyone is using the Horizontal Nav variant? It might be a good candidate to deprecate.

mcarrano (approved):
Looks great!

tlabaj (changes requested):
Since we have the new `PageToggleButton` component,  I would had that to the demo app.  I am ok with doing that in a follow up PR.  
I however would add the component to the `Page.md` file propComponents so that it is exposed in the docs.

kmcfaul:
@tlabaj Added `PageToggleButton` to the props docs and updated #6281 to include adding a Page-Masthead-PageToggleButton demo to integration

tlabaj (approved):

mturley (approved):
LGTM

nicolethoen (approved):

PR #6279, opened 2021-09-08
Composable search with form

### LINKED ISSUE #6212: Composable search input with form
Demonstrate how a search form can be attached to a Search input to give the user the option of performing a visual search. The result should look like this:

Comosable-visual-search

nicolethoen:
**What**: Closes #6212 

I pulled out the dynamically built advanced search into a separate (unexposed) component.
I updated the docs for the dynamically built advanced search example for clarity.
I added a demo to demonstrate building a custom composable advanced search form.

### DIFF
--- a/packages/react-core/src/components/SearchInput/AdvancedSearchMenu.tsx
+++ b/packages/react-core/src/components/SearchInput/AdvancedSearchMenu.tsx
@@ -28,7 +28,7 @@ export interface AdvancedSearchMenuProps extends Omit<React.HTMLProps<HTMLDivEle
   /** Function called to toggle the advanced search menu */
   onToggleAdvancedMenu?: (e: React.SyntheticEvent<HTMLButtonElement>) => void;
   /** Flag for toggling the open/close state of the advanced search menu */
-  showSearchMenu?: boolean;
+  isSearchMenuOpen?: boolean;
   /** Label for the buttons which reset the advanced search form and clear the search input */
   resetButtonLabel?: string;
   /** Label for the buttons which called the onSearch event handler */
@@ -59,7 +59,7 @@ export const AdvancedSearchMenu: React.FunctionComponent<AdvancedSearchMenuProps
   onClear,
   resetButtonLabel = 'Reset',
   submitSearchButtonLabel = 'Search',
-  showSearchMenu,
+  isSearchMenuOpen,
   onToggleAdvancedMenu
 }: AdvancedSearchMenuProps) => {
   const firstAttrRef = React.useRef(null);
@@ -75,13 +75,13 @@ export const AdvancedSearchMenu: React.FunctionComponent<AdvancedSearchMenuProps
   });
 
   React.useEffect(() => {
-    if (showSearchMenu && firstAttrRef && firstAttrRef.current) {
+    if (isSearchMenuOpen && firstAttrRef && firstAttrRef.current) {
       firstAttrRef.current.focus();
       setPutFocusBackOnInput(true);
-    } else if (!showSearchMenu && putFocusBackOnInput && parentInputRef && parentInputRef.current) {
+    } else if (!isSearchMenuOpen && putFocusBackOnInput && parentInputRef && parentInputRef.current) {
       parentInputRef.current.focus();
     }
-  }, [showSearchMenu]);
+  }, [isSearchMenuOpen]);
 
   React.useEffect(() => {
     document.addEventListener('mousedown', onDocClick);
@@ -97,7 +97,7 @@ export const AdvancedSearchMenu: React.FunctionComponent<AdvancedSearchMenuProps
 
   const onDocClick = (event: Event) => {
     const clickedWithinSearchInput = parentRef && parentRef.current.contains(event.target as Node);
-    if (showSearchMenu && !clickedWithinSearchInput) {
+    if (isSearchMenuOpen && !clickedWithinSearchInput) {
       onToggleAdvancedMenu(event as any);
     }
   };
@@ -105,7 +105,7 @@ export const AdvancedSearchMenu: React.FunctionComponent<AdvancedSearchMenuProps
   const onEscPress = (event: KeyboardEvent) => {
     const keyCode = event.keyCode || event.which;
     if (
-      showSearchMenu &&
+      isSearchMenuOpen &&
       keyCode === KEY_CODES.ESCAPE_KEY &&
       parentRef &&
       parentRef.current.contains(event.target as Node)
@@ -122,7 +122,7 @@ export const AdvancedSearchMenu: React.FunctionComponent<AdvancedSearchMenuProps
     if (onSearch) {
       onSearch(value, event, getAttrValueMap());
     }
-    if (showSearchMenu) {
+    if (isSearchMenuOpen) {
       onToggleAdvancedMenu(event as any);
     }
   };
@@ -199,7 +199,7 @@ export const AdvancedSearchMenu: React.FunctionComponent<AdvancedSearchMenuProps
     return formGroups;
   };
 
-  return showSearchMenu ? (
+  return isSearchMenuOpen ? (
     <div className={styles.searchInputMenu}>
       <div className={styles.searchInputMenuBody}>
         <Form>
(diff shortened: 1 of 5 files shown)


CooperRedhat:
This is some super complex functionality and everything seems to be working well! Awesome. 

I clicked around for a while and here is what I noticed:
- [Original issue](https://github.com/patternfly/patternfly-react/issues/6212) shows saving search state using chips, examples are using strings. Is that going to be added later, or was it dropped?
- In the demo: 
  - the advanced search keeps the values of form elements when opening/closing the `Popper`, with exception of the actual date value
  - After filling in values for `has the words`, `Date`, and `Date within`, the string that represents the search state in the input box above only shows the `has the words` value, but the example shows all of the delimited search parameters.
  - There seems to be a lot of work handed to the user in terms of handling focus, stopping event propagation, etc. So it's great to have this complex demo for them. In the long term is there a way to simplify that interface? Or could you provide some documentation in the demo code to explain why/how you handle focus, key events, etc?

CooperRedhat (commented):
See comment above

kmcfaul (commented):
The submit button causes a null warning in the console for the click handler, but looks great for everything else! Awesome work making it composable.

nicolethoen:
@CooperRedhat we will need to build a new component to allow for the 'input with chips look' and that is on the road map. So this story is just to demonstrate that advanced search forms can be built composably. I did try to add more comments within the demo code, as well as a few notes above the demo. Do you think that's helpful? I think I've addressed the rest of your comments

I added the after:date and to:date to the input if it's supplied in the advanced form, and then if you begin manually typing into the input still only updates the 'has words' field since I don't think it's worth the time/effort to be building a parser for the words 'after:' and 'to:' and more date parsing to this simple/small demo(?) @mcarrano Let me know what you think.

CooperRedhat (dismissed):
LGTM

CooperRedhat:
> @CooperRedhat we will need to build a new component to allow for the 'input with chips look' and that is on the road map. So this story is just to demonstrate that advanced search forms can be built composably. I did try to add more comments within the demo code, as well as a few notes above the demo. Do you think that's helpful? I think I've addressed the rest of your comments
> 
> I added the after:date and to:date to the input if it's supplied in the advanced form, and then if you begin manually typing into the input still only updates the 'has words' field since I don't think it's worth the time/effort to be building a parser for the words 'after:' and 'to:' and more date parsing to this simple/small demo(?) @mcarrano Let me know what you think.

Yes the comments definitely help give some context to what's going on. Should help our users modify your example more easily to fit their needs! nice

mcarrano (commented):
@nicolethoen this is definitely better. I still had a bit of a problem figuring out the logic of how the 'Date within' and 'Date' fields were being used. Could we make a slight change to the labeling on the form to be something like "Date within" for the first field and "of" in front of the Date picker? So the whole thing reads like "date within 2 weeks of 09/09/2021" for example. I think that would clarify the example a bit. Thanks.

mcarrano (approved):
Looks great. Thanks @nicolethoen !

CooperRedhat (approved):
LGTM

tlabaj (changes requested):
Great work!  this looks awesome.  Just a few comments.

CooperRedhat (dismissed):

packages/react-core/src/components/SearchInput/AdvancedSearchMenu.tsx
```diff
@@ -0,0 +1,223 @@
...
+    attrValueMap: { [key: string]: string }
+  ) => void;
+  /** A callback for when the user clicks the clear button */
+  onClear?: (event: React.SyntheticEvent<HTMLButtonElement>) => void;
+  /** A callback for when the input value changes */
+  onChange?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
+  /** Function called to toggle the advanced search menu */
+  onToggleAdvancedMenu?: (e: React.SyntheticEvent<HTMLButtonElement>) => void;
+  /** Flag for toggling the open/close state of the advanced search menu */
+  showSearchMenu?: boolean;
```
tlabaj, line 31: Maybe  `isSearchMenuOpen`

packages/react-core/src/components/SearchInput/examples/SearchInput.md
```diff
@@ -194,6 +194,12 @@ TextInputSelectAll = () => {
 
 ### Advanced
 
+The search input component can be used to dynamically build a one to one attribute-value advanced search.
```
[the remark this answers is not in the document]
kmcfaul: Same!

kmcfaul (approved):
I agree with Titani that `isSearchMenuOpen` might be a bit more consistent. Everything else looks great! Love the descriptions/comments in the docs.

tlabaj (approved):
looks great!

kmcfaul (dismissed):

mturley (approved):
This looks awesome!

packages/react-core/src/demos/SearchInput.md
```diff
@@ -0,0 +1,299 @@
...
+  const dateWithinToggle = (
+    <MenuToggle ref={dateWithinToggleRef} onClick={toggleDateWithinMenu} isExpanded={isDateWithinOpen} style={{width: "100%"}}>
+      {dateWithin}
+    </MenuToggle>
+  );
+  
+  const advancedForm = (
+    <div ref={advancedSearchPaneRef} role="dialog" aria-label="Advanced search form">
+      <Form>
+        <Card>
```
mcoker, line 244: The card component doesn't match the search input styling. Can we expose the `.pf-c-search-input__menu` and `.pf-c-search-input__menu-body` elements and use those instead? https://github.com/patternfly/patternfly-react/blob/06b7d17f1b48996f17fd1021f1834eaeebd7bc18/packages/react-core/src/components/SearchInput/SearchInput.tsx#L358

packages/react-core/src/demos/SearchInput.md
```diff
@@ -0,0 +1,299 @@
...
+                  id="datePicker" 
+                  style={{width: "100%"}} 
+                  value={date} 
+                  onChange={setDate} 
+                  appendTo={() => document.querySelector("#datePicker")}
+                />
+              </FormGroup>
+            </Grid>
+          </CardBody>
+          <CardFooter>
```
mcoker, line 275: The spacing for card footer actions doesn't match the spacing for form actions. If we swap the card out with the search input menu/menu-body elements, we can use a form action group for the right spacing here. If we aren't going to do that, I think this form should just be in a `<CardBody>`, and the form action group used for the actions.

mcoker (changes requested):
Looks good, left a couple of comments.

packages/react-core/src/demos/SearchInput.md
nicolethoen: In this case, the menu is not a child of search input - so the styles are not applied correctly to the form if they are in orphaned `.pf-c-search-input__menu` and `.pf-c-search-input__menu-body`components.

packages/react-core/src/demos/SearchInput.md
nicolethoen: I can do that!

nicolethoen: well - actually.  I will play with it - I think i was getting into sticky territory where the actions need to be in a `Form` and so the `Form` was wrapping the `CardBody` and `CardFooter`. I'll take another stab at it

packages/react-core/src/demos/SearchInput.md
mcoker: Hmm... I'm not sure I understand. The rendered markup all seems to support the styling working, I just renamed the card elements to search input menu/menu-body and moved the actions into the form as an action group:

Screen Shot 2021-09-10 at 1 38 04 PM

tlabaj (dismissed):
Still looks good!

kmcfaul (approved):

jessiehuff (commented):
Looks pretty good to me! The only weird thing with the demo is that when I'm using keyboard (or screen reader), and I select something for the "Date within" option, the focus is completely lost, and I have to navigate all the way back to it which would be pretty frustrating for keyboard and screen reader users. I'm guessing this is just a demo issue though. :)

nicolethoen:
good catch, @jessiehuff  I'll fix that!

tlabaj (approved):

mcoker (approved):

mcarrano (approved):
Looks good. Thanks @nicolethoen !

PR #6282, opened 2021-09-08
Introduce BackToTop component and demos

### LINKED ISSUE #6166: Introduce 'Back to top' component and demo
Back to top is a floating button that, when clicked, scrolls the user back to the top of the page. The component should include a demo that shows how this behaves in the context of a page.

Core issue: https://github.com/patternfly/patternfly/issues/4255

CooperRedhat:
**What**: Closes #6166 

**Overview**:
- Adds BackToTop component and demo
- By default the component is hidden, until the user scrolls `400px` (default scroll distance can be changed and/or exposed to be customizable)
- Any query selector can be used, demo shows using `name`


nicolethoen (commented):
Is there a way to navigate to the `Back to top` using the keyboard? On the first demo, I have not been able to figure out how.

kmcfaul:
It looks like it may need a `tabIndex` (at least testing in the first demo), but I'm not sure why it isn't working be default because it is a button already.

mcarrano (commented):
This is looking good @CooperRedhat . I think it raises interesting questions about intended interaction by keyboard and for screen reader users. Not sure what to expect in those cases. @jessiehuff can you also take a look?

CooperRedhat:
> It looks like it may need a `tabIndex` (at least testing in the first demo), but I'm not sure why it isn't working be default because it is a button already.

For some reason using a button with `component="a"` causes the button to not get focus in both this case. I'm using that because that is what the core example is using. As a standard button, you can tab to the button. 

Simple button examples with `component="a"` don't have this issue, could be something with the wrapper.

mcoker (dismissed):
LGTM, though not sure about whether this should be an `<a>` or `<button>`.

CooperRedhat:
- I changed the button to a normal button, apparently an anchor element without an href does not get focus.
- Current implementation is using javascript to handle scrolling. The core example is using `#`+id href to scroll, and @mcoker expressed that it would be nice to handle scrolling via CSS.
  - I believe that requires modifying the `scroll-behavior` property of the scrolling component to smooth the animation.

Remaining questions:
- Do we want to use JS or CSS to scroll? 
- The main JS functionality of this component is to show/hide based off the scroll position of a certain element, with an onClick handler to actually scroll.
  - In the effort of composability, could that functionality be extracted to a wrapper component around a `Button` with rounded corners and a drop shadow, and an onClick handler? In other words does this need to be its own component?
  - i.e.`<ShowOnScroll scrollingComponentSelector="" showAfter="400px"><Button isBackToTop icon={<ArrowUp/> onClick={scrollToTop} /></ShowOnScroll>`
  - The missing bit is positioning the button properly, which will change based on use case.

mcoker:
> apparently an anchor element without an href does not get focus.

Why wouldn't it have an `href`?

> I believe that requires modifying the scroll-behavior property of the scrolling component to smooth the animation.

Scrolling is one part we could handle with CSS, though I was referring more to any sort of animation we apply to the button as it comes into/out of view.

CooperRedhat:
> Why wouldn't it have an `href`?

If we're using JS to scroll, no need for an anchor tag or href. I originally used anchor because that is what the core example had.

> > I believe that requires modifying the scroll-behavior property of the scrolling component to smooth the animation.
> 
> Scrolling is one part we could handle with CSS, though I was referring more to any sort of animation we apply to the button as it comes into/out of view.

Ah that makes sense and is likely simpler than also changing scroll behavior of other components.

mcarrano (commented):
@CooperRedhat Looks good. But not sure I understand the purpose of the second demo.

packages/react-core/src/components/BackToTop/BackToTop.tsx
```diff
@@ -0,0 +1,92 @@
...
+  /** Additional classes added to the back to top. */
+  className?: string;
+  /** Title to appear in back to top button, defaults to 'Back to top'. */
+  title?: string;
+  /** Forwarded ref */
+  innerRef?: React.Ref<any>;
+  /** Selector for the scrollable element to spy on. Not passing a selector defaults to spying on window scroll events. */
+  scrollableSelector?: string;
+  /** Flag to always show back to top button, defaults to false. */
+  alwaysShow?: boolean;
```
tlabaj, line 18: We usually start our Boolean prop names with `is`.  Maybe `isAlwaysVisible` or something along those lines.

nicolethoen:
I think we will need to add a tabindex to the Page section with the overflow scroll - so that the keyboard user can scroll the page section and reveal the back to top. Otherwise, there is no way to read the contents of the page section as a keyboard user.

also, the whole left nav is not keyboard accessible in the demo either...

packages/react-core/src/components/BackToTop/BackToTop.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  const handleClick = () => {
+    scrollElement.scrollTo({ top: 0, behavior: 'smooth' });
+  };
+
+  return (
+    <div
+      className={css(styles.backToTop, hidden && styles.modifiers.hidden, className)}
+      {...props}
```
tlabaj, line 78: can you spread the props at the end.

packages/react-integration/demo-app-ts/src/components/demos/BackToTopDemo/BackToTopDemo.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
tlabaj, line 1: It is probably also a good idea to have a demo of the `alwaysShow` case.  Could we also add a test for the interaction?

tlabaj (changes requested):
Looks good!  Just a few comments.

packages/react-core/src/components/BackToTop/BackToTop.tsx
```diff
@@ -0,0 +1,92 @@
...
+import styles from '@patternfly/react-styles/css/components/BackToTop/back-to-top';
+import { css } from '@patternfly/react-styles';
+import AngleUpIcon from '@patternfly/react-icons/dist/esm/icons/angle-up-icon';
+import { canUseDOM } from '../../helpers/util';
+import { Button } from '../Button';
+
+interface BackToTopProps extends React.DetailedHTMLProps<React.HTMLProps<HTMLDivElement>, HTMLDivElement> {
+  /** Additional classes added to the back to top. */
+  className?: string;
+  /** Title to appear in back to top button, defaults to 'Back to top'. */
```
tlabaj, line 11: It is not necessary to list the default sine there is a column for that in the docs.

jessiehuff (commented):
I agree with Nicole that the scrollable content should be focusable so that users can still read all of the content by keyboard. Also, the nav items need an href to be keyboard accessible as I think Nicole might have already mentioned at one point as well, even if it's just `href="#"`. Because there's nothing interactive in the main section of the page, the back to top never appears on the screen, and it seems that it's only focusable when it's visible. What are we doing to hide the back to top component when it's not in view? It seems odd to me that focus doesn't go to it even when not in view. If you add a regular button at the bottom of the page, even if it's not in view, focus will shift to it when you press Tab. It seems to sound ok in VO though.

mcoker:
> What are we doing to hide the back to top component when it's not in view?

@jessiehuff there is a hidden class from core that applies `display: none;`. I assumed we would want the keyboard/screenreader experience to be the same so I used `display: none;` which should hide it from screen readers. I can use something else instead to maintain the ability to focus on it when it isn't visible though!

jessiehuff:
> @jessiehuff there is a hidden class from core that applies `display: none;`. I assumed we would want the keyboard/screenreader experience to be the same so I used `display: none;` which should hide it from screen readers. I can use something else instead to maintain the ability to focus on it when it isn't visible though!

Ah ok, I getcha. Hmm I feel like I could argue either way here. Screen reader users won't "see" the back to top component in the context of the page until they actually get to it so that's probably fine for them. It seems like it being hidden affects the keyboard users more. It feels like the same thing as the skip to content component but in reverse. The skip to content is still tab-able even when hidden. How are we handling that one when it comes to visually hiding it?

mcoker:
>  The skip to content is still tab-able even when hidden. How are we handling that one when it comes to visually hiding it?

We move it off screen, then bring it into the viewport when it's focused. We can use the `.pf-screen-reader` class instead of `.pf-m-hidden` to hide it visually until we decide to show it instead.

packages/react-integration/demo-app-ts/src/components/demos/BackToTopDemo/BackToTopDemo.tsx
CooperRedhat: I added a toggle to the demo, let me know if that works!

tlabaj (approved):
LGTM!

mcarrano (dismissed):
Looks great!

kmcfaul (approved):

mcoker (approved):

jessiehuff (commented):
It looks like the demo could still use a few tweaks to help make it a little more accessible. It sounds like we can discuss the tabbable aspect of the button in a different PR, but I think we still need:
- the nav links should have an `href="#"` to make them accessible/interactive in the same way it is by click
- The "scrolling-section" should have a `tabIndex="0"` so that keyboard users can scroll this content the way that mouse users can. (I tried to play with the example TS and JS code to see which React element it is exactly, but I couldn't seem to make any changes. I don't know if this section is now read only? I also tried the CodeSandbox, but it was having trouble too.)

We can add these in another PR if we need to, but they're pretty much just all demo related I think at this point.

CooperRedhat:
> * the nav links should have an `href="#"` to make them accessible/interactive in the same way it is by click

I think what you are referring to isn't directly related to this demo, but a wrapper used in many demos. Does this [fix](https://github.com/patternfly/patternfly-react/pull/6297) it?


> * The "scrolling-section" should have a `tabIndex="0"` so that keyboard users can scroll this content the way that mouse users can. (I tried to play with the example TS and JS code to see which React element it is exactly, but I couldn't seem to make any changes. I don't know if this section is now read only? I also tried the CodeSandbox, but it was having trouble too.)

Interesting you had trouble adding the tabIndex, I will take a look into it and try to get it into this demo. However, such a change wouldn't only be necessary here, but in all cases where there is a scrolling section. So a PR and/or some extra documentation to might be necessary to enforce/suggest adding a tabIndex on scrolling portions of a page.

jessiehuff:
> I think what you are referring to isn't directly related to this demo, but a wrapper used in many demos. Does this [fix](https://github.com/patternfly/patternfly-react/pull/6297) it?

Yes, I didn't know that one was open. Awesome! :) 

> Interesting you had trouble adding the tabIndex, I will take a look into it and try to get it into this demo. However, such a change wouldn't only be necessary here, but in all cases where there is a scrolling section. So a PR and/or some extra documentation might be necessary to enforce/suggest adding a tabIndex on all scrolling portions of a page.

Yeah, I think my issue might have been a surge or doc site issue. And I completely agree on your point!! I would love to find a solution that would automatically add the ability to focus by keyboard on any scrollable content. I'm not sure where or how we'd add that ability.

CooperRedhat:
@jessiehuff Enforcing tabIndex might look something like [this](https://github.com/patternfly/patternfly-react/pull/6308) hooking into the hasOverflowScroll prop we use in a few components.

kmcfaul (approved):

jessiehuff (approved):
Nice!! 😄

nicolethoen (approved):

PR #6283, opened 2021-09-08
(Menu): Added context selector demo using Menu

### LINKED ISSUE #6019: Make a demo of ContextSelector using Menu Toggle and Menu

tlabaj:
**What**: Closes #6019 
link to demo: https://patternfly-react-pr-6283.surge.sh/demos/composable-menu#context-selector-using-menu-components

### DIFF
--- a/packages/react-core/src/demos/ComposableMenu.md
+++ b/packages/react-core/src/demos/ComposableMenu.md
@@ -1175,25 +1175,25 @@ import {
   Divider,
   TextInput,
   InputGroup,
-  Button,
+  Button
 } from '@patternfly/react-core';
 import { Link } from '@reach/router';
 import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import pfIcon from './examples/pf-logo-small.svg';
 
-MenuContextSelector = () => {    
+MenuContextSelector = () => {
   const items = [
-      'My project',
-      'OpenShift cluster',
-      'Production Ansible',
-      'AWS',
-      'Azure',
-      'My project 2',
-      'OpenShift cluster ',
-      'Production Ansible 2 ',
-      'AWS 2',
-      'Azure 2'
-    ];
+    'My project',
+    'OpenShift cluster',
+    'Production Ansible',
+    'AWS',
+    'Azure',
+    'My project 2',
+    'OpenShift cluster ',
+    'Production Ansible 2 ',
+    'AWS 2',
+    'Azure 2'
+  ];
   const [isOpen, setIsOpen] = React.useState(false);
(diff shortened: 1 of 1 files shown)


packages/react-core/src/demos/ComposableMenu.md
```diff
@@ -1155,3 +1156,172 @@ MenuAppLauncher = () => {
...
+            aria-label="Search menu items"
+            id="pf-context-selector-search-button-id-1"
+            onClick={onSearchButtonClick}
+          >
+            <SearchIcon aria-hidden="true" />
+          </Button>
+        </InputGroup>
+      </MenuInput>
+      <Divider/>
+      <MenuContent>
```
mcoker, line 1305:
```suggestion
      <MenuContent  maxMenuHeight="200px">
```

mcoker (changes requested):
LGTM, only difference I see is that the context menu has a max height of 200px on the menu list.

nicolethoen:
LGTM - Just wondering in VO how to get to the menu items? I can use the tab + arrow keys easily, but VO does not prompt me to do that and I cannot seem to find a VO controls path into the menu - @jessiehuff is it set up like you'd expect?

mcarrano (approved):
Looks good other than the issues/questions that have already been raised.

kmcfaul (approved):
lgtm other than the VO question. Keyboard navigation looks good

mcoker (approved):

CooperRedhat (approved):
LGTM

jessiehuff (commented):
I tested with VO, JAWS, and NVDA, and I'm able to get into the menu items in each as I'd expect. With VO, I navigated with ctr + option + arrow keys, and I think it was ok. I'm curious what your experience was, @nicolethoen! Maybe you can screen record and send it to me to see if I can replicate?

The only thing that I saw that I would consider changing is the keyboard interaction. Based on [the keyboard interaction doc](https://docs.google.com/document/d/1m9nICFRgCXnXeHYN487BMX9MHw4DjvOln2B5p8yJa70/edit#heading=h.1lr7m0rc8a6s) that Jenn and I created awhile back, menu items should be navigated by arrow keys instead of Tab. I know that we already have an [open issue](https://github.com/patternfly/patternfly-react/issues/3205) for that though so we can do these separately if we want.

jessiehuff (approved):

nicolethoen (approved):
I tried again - I mistakenly thought I had to have the VO buttons pressed while trying to navigate the menu and was finding that the VO focus was getting trapped in the button. I'm still learning the controls. When I use the arrow keys without having the VO buttons pressed, i'm able to move the VO focus into the menu no problem.

PR #6297, opened 2021-09-10
fix: wrapper nav should have hrefs

kmcfaul:
**What**: Closes #

Adds hrefs to `Nav` in `DashboardWrapper` to make it tabbable.


kmcfaul:
This change makes me realize our examples "active" NavItem in this context isn't really useful (because the href will redirect the page and clear the activeItem state). We may want to update the logic to parse/grab the url and set the activeItem based on that in the future.

nicolethoen (approved):

CooperRedhat (approved):
Thank you!

jessiehuff (approved):

Issue #4395, opened 2020-06-15
TitleSizes needs to be imported in docs for Title examples to work


DarrienG:
**Describe the issue. What is the expected and unexpected behavior?**

The latest PF release changes the API for Titles and TitleSizes needs to be imported for the examples to work.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

Visible here: https://www.patternfly.org/v4/documentation/react/components/title

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug in the docs. You can import in your own code to make it work.

**What is your product and what release version are you targeting?**

    "@patternfly/patternfly": "^4.10.31",
    "@patternfly/react-core": "4.18.5",

DarrienG:
Fixed via #4387

PR #6326, opened 2021-09-16
fix(TreeView): isExpanded should override defaultExpanded

### LINKED ISSUE #6314: TreeView: `allExpanded` prop doesn't apply recursively, doesn't take effect if `true` on first mount. `defaultAllExpanded` affects more than initial state.

The new `allExpanded` prop introduced in https://github.com/patternfly/patternfly-react/pull/5503 and the existing `defaultAllExpanded` prop seem to behave differently than I would expect.

* If `defaultAllExpanded` is not passed, changing `allExpanded` to true only expands the immediate children of the root node(s). It does not behave the same way as `defaultAllExpanded`, which expands the tree fully (all descendants down to the leaves). However, if `defaultAllExpanded` is passed, changing `allExpanded` to true DOES fully expand the tree correctly regardless of current state when it changes.
* If I set `allExpanded` to `true` when the component first mounts, it does not start with the tree expanded. Using state to flip it to false and back to true does expand the tree. It appears I need to also set `defaultAllExpanded` if I want this behavior, but that feels weird to me, maybe we should deprecate the `defaultAllExpanded` prop if we can get the initial value of `allExpanded` to apply this way?
* Worth noting, when `defaultAllExpanded` is true, expanding any node in the tree fully expands all its descendants. I would expect it to only affect the initial state (toggling an individual node should only ever affect that node's level in the tree).

The behavior I'm looking for is: the tree starts fully recursively expanded, there is an expand/collapse all control that will fully expand the tree, but expanding individual nodes does not expand their descendants. I can't see a way to do that with the props as they behave now.

You can see this behavior by changing these two props on the [default TreeView example](https://github.com/patternfly/patternfly-react/pull/5503) in the docs.

Bug, I think? Unless I'm misunderstanding the intent of these props.

Migration Toolkit for Virtualization, no particular target release.

cc @kmcfaul

kmcfaul:
**What**: Closes #6314, #4395

`allExpanded` (TreeView) and `isExpanded` (TreeViewListItem) should override `defaultAllExpanded`/`defaultExpanded`.

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
@@ -85,14 +85,10 @@ export const TreeViewListItem: React.FunctionComponent<TreeViewListItemProps> = 
   useEffect(() => {
     if (isExpanded !== undefined && isExpanded !== null) {
       setIsExpanded(isExpanded);
-    }
-  }, [isExpanded]);
-
-  useEffect(() => {
-    if (defaultExpanded !== undefined && defaultExpanded !== null) {
+    } else if (defaultExpanded !== undefined && defaultExpanded !== null) {
       setIsExpanded(internalIsExpanded || defaultExpanded);
     }
-  }, [defaultExpanded]);
+  }, [isExpanded, defaultExpanded]);
 
   const Component = hasCheck ? 'div' : 'button';
   const ToggleComponent = hasCheck ? 'button' : 'div';


nicolethoen (approved):

mturley:
Thanks @kmcfaul !

I'm trying to figure out if this fully closes #6314. I'm still noticing something weird: if either `allExpanded` or `defaultAllExpanded` is true and then you collapse some nodes, expanding any node in the tree will recursively expand its descendants. The behavior I'm trying to implement is an expand/collapse-all toggle that acts "one time", and after expanding the full tree i can expand and collapse individual nodes without causing descendants to expand. It's weird because `allExpanded` stays set to true when collapsing some nodes, but doesn't accurately reflect the state of the expansions anymore.

I think the way to implement this might be to somehow have a callback that lets me set `allExpanded` back to null if the user expands/collapse any individual node, but since that state is handled inside the component we don't expose a callback for when it changes. We're in this weird position where state is mostly handled internally but can be overridden by this external prop. I think maybe we need to do one of these:

* Expose a callback for `onExpand` so the consumer can adjust `allExpanded` if necessary
* Re-evaluate managing expanded state internally in the tree... although making it entirely driven via props sounds like a bit of a nightmare and certainly a breaking change.

Also, I wonder if `defaultAllExpanded` should be deprecated now that we can achieve its behavior with an initial `allExpanded`?

Sorry to ramble. We could address these in a followup PR maybe.

kmcfaul:
Would it be possible to use `defaultAllExpanded` to set the 'one-time' full expansion instead? `allExpanded` and `isExpanded` are meant to override the internal state, but `defaultAllExpanded` and `defaultExpanded` should only set the initial state.

mturley:
Hm. So you mean use `defaultAllExpanded` for the toggle button instead of `allExpanded`? If that only sets the initial state I don't know if it would work.

Maybe we can have `allExpanded` only set the internal state when it changes, instead of its value affecting the behavior when toggling expansion on a node normally? That's the part that feels buggy to me. Whether or not I clicked "expand all" or "collapse all" first shouldn't change the behavior of clicking toggles in the tree, I would think.

I still think maybe the callback would be helpful because it feels weird to have the toggle text still read "collapse all" after I've manually collapsed some nodes. I was thinking it should behave like the select-all checkbox in a table: when you select all the box is checked and becomes an unselect-all button, but then if you unselect some rows the box is unchecked and goes back to being a select-all button.

Sorry if I'm overthinking it.

mturley:
I don't mean to hold up the PR, if we want to merge it I can open a followup issue and/or PR.

redallen (approved):
Followup is good.

kmcfaul:
@mturley Yes let's open a followup for that, I do agree adding a callback to the internal expanded state for informational purposes is a good idea. 

I meant to say that `defaultAllExpanded` should only be used to set the default value of the expanded state of a node and so we should potentially remove the `useEffect` that listens to that prop updating, since we have `allExpanded` for that. (And their item-level counterparts `defaultExpanded` and `isExpanded`). Or we should remove `defaultAllExpanded` completely and only use `allExpanded`, using that to set the initial state as well.

mturley:
Ah I see. Yeah, I think maybe removing `defaultAllExpanded` makes the most sense, but either of those solutions sounds good to me.

PR #6328, opened 2021-09-17
fix(dragdrop): remove React.createPortal, add noWrap

redallen:
**What**: Putting data list rows (or any arbitrary list items that expect style based on a parent) leads to a situation where when moved into a portal it's styled completely differently. For this reason dragging from the main content to Sidebars, Drawers, and PageHeaders has been removed.

The `noWrap` prop was also added so that `<div>`s need not wrap the component. This was accomplished by using event targets instead of refs to the div and passing event handlers via `React.cloneElement`.

Towards https://github.com/patternfly/patternfly/issues/4376

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -27,15 +27,15 @@ export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, '
   className?: string;
   /** Adds accessible text to the DataList list */
   'aria-label': string;
-  /** Optional callback to make DataList selectable, fired when DataListItem selected */
+  /** @deprecated Optional callback to make DataList selectable, fired when DataListItem selected */
   onSelectDataListItem?: (id: string) => void;
-  /** Optional callback to make DataList draggable, fired when dragging ends */
+  /** @deprecated Optional callback to make DataList draggable, fired when dragging ends */
   onDragFinish?: (newItemOrder: string[]) => void;
-  /** Optional informational callback for dragging, fired when dragging starts */
+  /** @deprecated Optional informational callback for dragging, fired when dragging starts */
   onDragStart?: (id: string) => void;
-  /** Optional informational callback for dragging, fired when an item moves */
+  /** @deprecated Optional informational callback for dragging, fired when an item moves */
   onDragMove?: (oldIndex: number, newIndex: number) => void;
-  /** Optional informational callback for dragging, fired when dragging is cancelled */
+  /** @deprecated Optional informational callback for dragging, fired when dragging is cancelled */
   onDragCancel?: () => void;
   /** Id of DataList item currently selected */
   selectedDataListItemId?: string;
@@ -45,7 +45,7 @@ export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, '
   gridBreakpoint?: 'none' | 'always' | 'sm' | 'md' | 'lg' | 'xl' | '2xl';
   /** Determines which wrapping modifier to apply to the DataList */
   wrapModifier?: DataListWrapModifier | 'nowrap' | 'truncate' | 'breakWord';
-  /** Order of items in a draggable DataList */
+  /** @deprecated Order of items in a draggable DataList */
   itemOrder?: string[];
 }
 
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -93,6 +93,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
     wrapModifier: null
   };
   dragFinished: boolean = false;
+  html5DragDrop: boolean = false;
   arrayCopy: React.ReactElement[] = React.Children.toArray(this.props.children) as React.ReactElement[];
   ref = React.createRef<HTMLUListElement>();
 
@@ -106,7 +107,8 @@ export class DataList extends React.Component<DataListProps, DataListState> {
   constructor(props: DataListProps) {
     super(props);
 
-    if (props.onDragFinish || props.onDragStart || props.onDragMove || props.onDragCancel) {
+    this.html5DragDrop = Boolean(props.onDragFinish || props.onDragStart || props.onDragMove || props.onDragCancel);
+    if (this.html5DragDrop) {
       // eslint-disable-next-line no-console
       console.warn("DataList's onDrag API is deprecated. Use DragDrop instead.");
     }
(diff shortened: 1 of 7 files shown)


packages/react-core/src/components/DataList/examples/DataList.md
```diff
@@ -1125,6 +1125,10 @@ import {
 
 ### Draggable
 
+Draggable data lists used to have their own HTML5 API based drag and drop. Upon initial implementation we thought this implementation sufficient for all future drag and drop functionality. We were wrong.
```
nicolethoen, line 1128: I think that this statement - "Upon initial implementation we thought this implementation sufficient for all future drag and drop functionality" is always true - and obviously often proven untrue. So I wonder if it'd be more constructive to say something like
"The initial implementation has been unable to fulfill all requirement requests."

packages/react-core/src/components/DataList/examples/DataList.md
```diff
@@ -1135,129 +1139,93 @@ import {
...
+  const [items, setItems] = React.useState(getItems(10));
+  const [liveText, setLiveText] = React.useState('');
 
-  const onDragStart = newId => {
-    setId(newId)
-    setLiveText(`Dragging tarted for item id: ${newId}.`);
-  };
+  function onDrag(source) {
+    setLiveText(`Started dragging ${items[source.index].content}`);
+    return true;
```
nicolethoen, line 1167: I think it'd be helpful to add a comment about why you are returning 'true' even though the DragDrop docs explain it.

packages/react-core/src/components/DragDrop/DragDrop.tsx
```diff
@@ -17,13 +18,18 @@ interface DragDropProps {
   children?: React.ReactNode;
   /** Callback for drag event. Return true to allow drag, false to disallow. */
   onDrag?: (source: DraggableItemPosition) => boolean;
+  /** Callback for drag event. */
```
nicolethoen, line 21: how is this intended to be used differently from `onDrag`. That clarity should be included in the description.

packages/react-core/src/components/DragDrop/Droppable.tsx
```diff
@@ -10,19 +10,29 @@ interface DroppableProps extends React.HTMLProps<HTMLDivElement> {
   zone?: string;
   /** Id to be passed back on drop events */
   droppableId?: string;
+  /** Don't wrap the component in a div. Requires passing a single child. */
+  noWrap?: boolean;
```
nicolethoen, line 14: noWrap is a css term - so I think you might need a clearer variable name - like noWrapper

nicolethoen (changes requested):

packages/react-core/src/components/DataList/examples/DataList.md
nicolethoen: wait, did you do this? I don't see it.

nicolethoen (changes requested):
I know we haven't implemented keyboard accessibility for dragdrop yet, but when i just played with the keyboard to see what happened, I was able to get it to error out by focusing on a drag icon and hitting enter. I think we should at least disable keyboard handling if we are not going to fully implement it so we dont end up in an error state.

redallen:
@nicolethoen I didn't do the liveText. It was done in #5783 and I copied it over.

redallen:
I've disabled the old HTML5-based API DataLisatDragButton onClick handler.

packages/react-core/src/components/DragDrop/Draggable.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from 'react';
-import { createPortal } from 'react-dom';
 import { DroppableContext } from './DroppableContext';
 import { DragDropContext } from './DragDrop';
 
 export interface DraggableProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside DragDrop */
   children?: React.ReactNode;
+  /** Don't wrap the component in a div. Requires passing a single child. */
+  noWrapper?: boolean;
```
tlabaj, line 9: maybe hasNoWrapper to more consistent?

packages/react-core/src/components/DataList/examples/DataList.md
redallen: OH I just now see what you mean. I added it in DragDrop.md but not DataList.md, I'll add it now.

packages/react-core/src/components/DataList/examples/DataList.md
```diff
@@ -1125,6 +1125,10 @@ import {
 
 ### Draggable
 
+Draggable data lists used to have their own HTML5-based API for drag and drop. This was unable to fulfill requirements like custom styling on items as they are being dragged.
+
+Since then we've written generic new `DragDrop`, `Draggable`, and `Droppable` components for this purpose. This should be used instead of the old deprecated (and buggy!) HTML5-based API.
+
```
abigaeljamie, line 1131:
```suggestion
Draggable data lists used to have their own HTML5-based API for drag and drop, which wasn't able to fulfill requirements such as custom styling on items being dragged. So we wrote generic `DragDrop`, `Draggable`, and `Droppable` components for this purpose. Use those new components instead of the deprecated (and buggy!) HTML5-based API.
```

abigaeljamie: A few tiny tweaks for clarity :) Please feel free to edit however you see fit!

redallen: I love it, thanks!

tlabaj (changes requested):
It looks like you removed support for the old drag and drop.  Unfortunately we can not do that.  We still need to support so we don't break people who are using it.  I think it is good that you have the console warning that it is deprecated, but it still needs to work with deprecated interface for now.

redallen:
Where did you see that I removed support? All the same code is still there and the demo in the integration test still works using the old code.

packages/react-core/src/components/DataList/DataList.tsx
```diff
@@ -308,13 +313,12 @@ export class DataList extends React.Component<DataListProps, DataListState> {
     } = this.props;
     const { dragging } = this.state;
     const isSelectable = onSelectDataListItem !== undefined;
-    const isDraggable = onDragFinish !== undefined;
```
tlabaj, line 311: So here, why don't we still need to check for onDragFinish.  I think this is needed to make it draggable.  I assumed you were not checking because you were planning on no longer supporting it.  If the drag and drop still woks when not defined, this should be ok.

redallen: Oh, instead of checking if `onDragFinish !== undefined` in the render I now check `Boolean(props.onDragFinish || props.onDragStart || props.onDragMove || props.onDragCancel)` in the constructor. This is more permissive.

tlabaj: Great! They way I am understanding the original code thought is that onDragFinished is the only one of drag props required.  since it is the only one checked.  I can play around with it and see what happens if I don't supply it.

tlabaj (approved):

nicolethoen (approved):

kmcfaul (approved):
LGTM!

PR #6339, opened 2021-09-21
feat(ProgressStepper): add progress stepper

### LINKED ISSUE #6165: Introduce new Progress Stepper component
Introduces a new component to track progress of a sequential task.

Core issue: https://github.com/patternfly/patternfly/issues/4214
Designs: https://marvelapp.com/prototype/gfa4i5b/screen/80215316

Notes:
Need to consider how to reflect status of a node to a screen reader user.

kmcfaul:
**What**: Closes #6165

In this PR:
- Adds `ProgressStepper` and `ProgressStep` components, docs, tests & demos


kmcfaul:
Couple open questions - 

1. `isCenterAligned` property - Does `isCentered` sound better? Or should we use an `alignment` property instead (if we want to support right alignment in the future)?
2. `isVertical`, `isCompact`, and `isCenterAligned` don't play nice together when 1+ are applied to the same `ProgressStepper`. Are these only meant to be used separately?
3. VoiceOver won't read the progress stepper for me unless I add a `tabIndex` to `ProgressStepper`, am I missing a way to get VO to read non-interactive text or does `ProgressStepper` need a `tabIndex`? In the same vein, I added `aria-polite` to the more interactive demo but I'm not noticing anything being announced.

mcarrano (commented):
This is looking good @kmcfaul . Regarding your question about using these options together. I wouldn't expect them to be, i.e., it's not valid to have a center-aligned or compact vertical stepper. Not sure there is something to do to present that or if that's more of a design convention.

It was very useful having the demo. But one thing this shows is that there is a shift when the selected step changes to bold. We've run into this problem before with other similar things. @mcoker @mceledonia @shirimordechay what do you think? https://patternfly-react-pr-6339.surge.sh/components/progress-stepper/react-demos

mcoker (approved):
LGTM! @srambach can you also review please? Looks like there is a small difference in how the warning icons are displayed (react on the left, core on the right)
Screen Shot 2021-09-23 at 11 52 48 AM

srambach:
> LGTM! @srambach can you also review please? Looks like there is a small difference in how the warning icons are displayed (react on the left, core on the right)

I noticed that. I know there's some difference between how core and react do icons - my adjustment is on `.pf-c-progress-stepper__step-icon .fa-exclamation-triangle` but that class isn't on this icon. Interestingly, the other adjustment for `.pf-c-progress-stepper__step-icon .pficon` is also not applied to those icons (like the hourglass) but they appear centered in the react examples.

I'm not sure how to address this, so first I'd ask design (@mceledonia ?) to weigh in on whether it's worth it. The triangle here is not as far off as it seemed in core.

redallen (dismissed):
Code looks good, just a test nit.

jessiehuff (commented):
To have the steps announce the status of each, I think it would make sense to add an `id` on both the li with the status `aria-label` and the step title div. Then you can add an `aria-labelledby` on the step title div with the id of the li then the id of the step title div (`aria-labelledby={status-id title-id}` so it will get announced like "completed step, First step".

For the demo, the live region isn't announcing the changes because technically speaking no content is actually changing in the DOM, only attributes are changing. Even if we alter the `aria-relevant` to the live region (what the screen reader is listening for), none of them listen to attributes specifically as `additions` listens to element nodes, `removals` listens to deletion of nodes, and `text` listens for existing textual content nodes to change. So my suggestion here is to add a hidden div as a sibling of the ol with `aria-live="polite"` that contains screen reader text only that updates with the status and step name. 

Looks pretty good so far though! :)

redallen (approved):

jessiehuff (commented):
That demo is seriously awesome, Katie!! 😄 My only nit on it is that there's really no need for an `aria-label` on the buttons since they have visible text. Also, for the progress stepper examples, can you add an id on both the li with the status aria-label and the step title div? Then you can add an aria-labelledby on the step title div with the id of the li then the id of the step title div (aria-labelledby={status-id title-id} so it will get announced like "completed step, First step".

jessiehuff (dismissed):
Nice job as always! 😄 🏆 🎉

nicolethoen (changes requested):
Can you add a 'beta' tag to the component?

Also, @redallen or @tlabaj can correct me, but I think that since this does not have any cypress tests, we probably dont need to create a demo in the demo-app.

redallen:
That's correct @nicolethoen !

packages/react-core/src/components/ProgressStepper/examples/ProgressStepper.md
```diff
@@ -0,0 +1,303 @@
...
+propComponents: ['ProgressStepper', 'ProgressStep']
+beta: true
+---
+
+import InProgressIcon from '@patternfly/react-icons/dist/esm/icons/in-progress-icon';
+import PendingIcon from '@patternfly/react-icons/dist/esm/icons/pending-icon';
+
+## Examples
+
+Progress steppers have defeault icons associated with the `variant` property, and may be overriden and customized using the `icon` property.
```
srambach, line 14: Typo :) "default"

srambach:
I defer to @jessiehuff on screenreading, but I did notice that in core I had set it up to read both the state (completed, current, pending) and status (step with warning, step with failure, etc). I don't think this is doing that so it's missing some information that you get when viewed - e.g. in the example with the failed current state, it only reads "danger step" so you don't know that it's actually the current step or if it's failed and the process has moved on.

kmcfaul:
The aria label could be customized to dynamically input the state probably, would that work better?

srambach:
> The aria label could be customized to dynamically input the state probably, would that work better?

If I understand what you mean, that's what I did: `aria-label="current step,step with failure,"` And I added a comma so that the punctuation/pause makes it easier to parse out the two labels as separate things.

kmcfaul:
Oh I see! I'll update that

packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
```diff
@@ -0,0 +1,82 @@
...
+      {...props}
+    >
+      <div className={css(styles.progressStepperStepConnector)}>
+        <span className={css(styles.progressStepperStepIcon)}>{_icon && _icon}</span>
+      </div>
+      <div className={css(styles.progressStepperStepMain)}>
+        <div
+          className={css(styles.progressStepperStepTitle)}
+          id={titleId}
+          {...(props.id !== undefined && titleId !== undefined && { 'aria-labelledby': `${props.id} ${titleId}` })}
```
tlabaj, line 73: Should the `aria-labeled` be on the `<li> since that is what is being labeled? `id` should be added there as well if that is the case. 
Should we generate the id her if non is supplied? If we don't, will this be accessible?  If so, we should probably generate warning if step is missing a title since it will not be accessible.

tlabaj (changes requested):

packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
jessiehuff: So my concern with putting the `aria-labelledby` on the li is that I don't think the screen reader really shifts focus there consistently. The visible text will get focus every time which made me think it would be more consistent to put it there. I did notice that at times the li's `aria-label` wasn't always announced (depending on the screen reader) so we could investigate further the best way to label these. In terms of the title though, I do think a warning could be valuable.

kmcfaul: What do you mean by warning?

tlabaj: @kmcfaul  similar to what we do with other components.  Log a console warning if the props required to make the component accessible  are missing.

kmcfaul: Added a console warning for missing either `id` or `titleId` (which are used to create the `aria-labelledby`).

packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
```diff
@@ -0,0 +1,90 @@
...
+  icon,
+  titleId,
+  ...props
+}: ProgressStepProps) => {
+  const _icon = icon !== undefined ? icon : variantIcons[variant];
+
+  if (props.id === undefined || titleId === undefined) {
+    /* eslint-disable no-console */
+    console.warn(
+      'The titleId and id properties are required to make this component accessible, and one or both of these properties are missing.'
```
tlabaj, line 60: Can you add append the component name to the beginning of the warning. `ProgressSteppe: The titleID and...`

kmcfaul: I used `ProgressStep` as that is the component in question that is made accessible by those props.

tlabaj (approved):
LGTM!

nicolethoen (approved):

mcarrano (approved):

srambach (approved):
Looks great!! 👍

PR #6340, opened 2021-09-21
Improve cypress perf

redallen:
**What**: Try to improve CI times

- Upgrade cypress to latest version
  - Remove typescript plugin (now comes with cypress)
  - Fix some tests by changing `not.visible` to `not.exists`
    - Remove tests that don't actually test integration
      - Class names are already captured by jest snapshots
- Revert flyout nav which caused performance issues in demo app
- Only render nav in demo app once
- Decrease parallelism to 2 to allow 2 builds to run simultaneously in CircleCI (we have 4 free executors). This provides for slightly slower but more consistent integration test run times
- Upgrade CI to node 14

**Additional issues**:


nicolethoen:
I'm not sure we want to just pull out all the flyout code from nav - we do need to refactor some of it, but need to release a flyout nav this upcoming release. @kmcfaul do you have thoughts about how much of this code we need to try and keep for ease of reimplementation?

nicolethoen (approved):

kmcfaul (approved):

packages/react-integration/cypress/integration/datalist.spec.ts
```diff
@@ -16,28 +16,19 @@ describe('Data List Demo Test', () => {
...
-    cy.get('#row2.pf-m-selected').should('not.be.visible');
+    cy.get('#row1.pf-m-selected').should('exist');
+    cy.get('#row2.pf-m-selected').should('not.exist');
 
     cy.get('#row2').click();
-    cy.get('#row1.pf-m-selected').should('not.be.visible');
-    cy.get('#row2.pf-m-selected').should('be.visible');
-  });
-
-  it('Verify actions ', () => {
```
evwilkin, line 34: Don't we want to keep this to test the interaction?

packages/react-integration/cypress/integration/datalist.spec.ts
```diff
@@ -61,22 +52,6 @@ describe('Data List Compact Demo Test', () => {
...
   });
-
-  it('Verify rows selectable', () => {
-    cy.get('#row1.pf-m-selectable').should('be.visible');
-    cy.get('#row2.pf-m-selectable').should('be.visible');
-
-    cy.get('#row1.pf-m-selected').should('not.be.visible');
-    cy.get('#row2.pf-m-selected').should('not.be.visible');
-
-    cy.get('#row1').click();
```
evwilkin, line 72: Don't we want to keep this to test the interaction?

packages/react-integration/cypress/integration/dropdown.spec.ts
```diff
@@ -35,23 +27,6 @@ describe('Dropdown Demo Test', () => {
     cy.get('#dropdown').should('not.have.class', 'pf-m-expanded');
   });
 
-  it('Verify dropdown menu items disabled', () => {
```
evwilkin, line 38: Don't we want to keep this to test the interaction?

evwilkin:
@redallen the majority of this looks g2g, but I'm not following a few of the integration tests we're pulling out that actually are testing the interactions - thanks

packages/react-integration/cypress/integration/dropdown.spec.ts
redallen: This interaction is invalid. You cannot click on an item with `pointer-events: none` and cypress now enforces this.

packages/react-integration/cypress/integration/datalist.spec.ts
redallen: We already test this interaction in Dropdown.

packages/react-integration/cypress/integration/datalist.spec.ts
redallen: I missed that this test clicks rows. The `be.visible` could be changed to `exists`.

I'd argue it's still okay to remove. We need not test that classes are added or removed for simple interactions because we know our `isSelectable && styles.modifiers.selectable` logic is good! If it wasn't this test wouldn't catch it anyways because it was added at the same time.

redallen:
@evwilkin This PR is a foretaste of another to come where tests without interactions or only simple interactions or already tested interactions are removed in favor of jest tests which are more performant.

All of the tests removed in this PR were failing because they were invalid tests. This PR removes the egregious ones and refactors the good ones.

evwilkin (approved):
LGTM

PR #6345, opened 2021-09-22
Chore/prune integration tests

redallen:
**What**: Towards #6349. Remove some tests in order to reduce integration test time which is the current CI bottleneck. After this PR a11y tests should once again be the bottleneck:
image

Changes:
- Go directly to demo instead of clicking link in nav
- Application launcher: Remove basic test which was superseded by the favorites test
- LabelGroup: Remove basic test which was superseded by the vertical test
- Page: Remove managed sidebar test which was superseded by managed sidebar closed test
- Alert: Added `cy.clock()` to save ~30s of waiting
- Select: Added `cy.clock()` to save 32s of waiting for "view more"

Removed for not testing interactions:
- Charts
- Bullseye
- Checkbox
- Stack
- ActionsList
- Backdrop
- Badge
- Banner
- Brand
- CodeBlock
- Divider
- EmptyState
- Flex
- Gallery
- Grid
- HelperText
- Hint
- Level
- List
- MenuToggle
- Navigation (was empty)
- OUIA
- Progress
- Skeleton
- Spinner
- Split
- Text
- Tile
- Title
- Vertical tabs

Some checks removed for not testing interactions:
- Card
- Accordion
- ClipboardCopy: expanded demo
- DescriptionList: keep only help text check
- Form: horizontal, field group, select
- Label
- Modal: remove duplicated open/close checks
- Nav: remove classname only check
- Page: breadcrumb
- Tabs: inset
- Table: breakpoints, compact, controltext, first cell as header, headers wrappable, table actions (superseded by tablesimpleactions)

**Additional issues**:


CooperRedhat:
Thanks Zack! A few questions to help me review this:
- What were the criteria you used to decide that a spec did not test interactions? I'm assuming some sort of automation there so would like to help make sure you got everything and didn't do too much.
- I'm concerned that we have been relying on the integration tests to do things that belonged in unit tests. So removing all of these checks could cause future changes to break things w/o a warning during build time. Is that a valid concern, if not why not? If so, how could we mitigate?

redallen:
Hey Cooper!

1. If a spec simply checked for classes and had no `.click()` or `.focus()` testing our JS then I considered it as not testing interactions. I didn't use any automation, I just went through all 160 spec files by hand and erred on the side of caution.
2. We're already well covered. Each component already has snapshot tests using Jest to check class names. Typescript + react-styles checks that those class names actually exist upstream, which is why I feel comfortable removing them. As far as I know, the tests I've removed never have failed the build in the past 2 years.

nicolethoen (commented):
I generally feel like we were using the demos to be a safe guard for verifying our components were type safe, since our examples were not written in typescript. Now that we are beginning to convert examples, that may not be an issue, but it does leave us vulnerable in the interim.

packages/react-integration/cypress/integration/modal.spec.ts
```diff
@@ -1,104 +1,6 @@
...
   it('Navigate to Modal section', () => {
-    cy.visit('http://localhost:3000/');
-    cy.get('#modal-demo-nav-item-link').click();
-    cy.url().should('eq', 'http://localhost:3000/modal-demo-nav-link');
-  });
-
-  it('Verify Default Modal', () => {
-    cy.get('#showDefaultModalButton').then((modalButton: JQuery<HTMLButtonElement>) => {
-      cy.wrap(modalButton).click();
-      cy.get('.pf-c-modal-box')
```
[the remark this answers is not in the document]
redallen: I left `Verify Half Width Modal` to test the open/close behavior. We do also test open/close in Jest.

redallen:
@nicolethoen You're correct. However, pruning these tests before converting the examples to typescript does save the work of consistently pruning _while_ converting. It also speeds up CI for all future PRs to boot.

I think the tests I removed aren't keeping us safe because:
1. The examples don't use the strictest types and often use `any`
2. Component developers understand that if a prop is of type `string` and we want to modify it, we best accept `string | anotherType`

As proof, in CI `build_demo_app` is the job with the *highest success rate* at 97% over 264 runs. I really don't think it checks much.

nicolethoen:
@tlabaj I think if you're comfortable deleting the demos from the app - given what @redallen said above, then I'm fine with it! I guess it's just important that everyone is aware (and learning how to write better demos/tests as a result of this).

CooperRedhat (approved):

packages/react-integration/cypress/integration/alert.spec.ts
```diff
@@ -64,19 +33,4 @@ describe('Alert Demo Test', () => {
         cy.wrap(tooltipLink).trigger('mouseleave');
       });
   });
-
-  it('Verify expandable alert', () => {
```
tlabaj, line 68: I think this is valid to keep.  it verifies the expand works no?

packages/react-integration/cypress/integration/applicationlauncher.spec.ts
```diff
@@ -1,34 +1,6 @@
-describe('Application Launcher Demo Test', () => {
-  it('Navigate to demo section', () => {
-    cy.visit('http://localhost:3000/');
-    cy.get('#application-launcher-demo-nav-item-link').click();
-    cy.url().should('eq', 'http://localhost:3000/application-launcher-demo-nav-link');
-  });
-
-  it('Verify application launcher button', () => {
```
tlabaj, line 8: This one looks like it is verifying the toggle interaction.

packages/react-integration/demo-app-ts/src/components/demos/ActionListDemo/ActionListDemo.tsx
```diff
@@ -1,104 +0,0 @@
-import {
-  ActionList,
-  ActionListGroup,
-  ActionListItem,
-  Button,
-  Dropdown,
-  DropdownItem,
-  DropdownSeparator,
-  KebabToggle
```
tlabaj, line 9: Are you removing these demos because the types are being verified in the example code now since you converted to typescript?

tlabaj (changes requested):
I noticed you removed some of the test that would test the breakpoint utility conversions.  Do you plan to add those to the unit tests?

packages/react-integration/demo-app-ts/src/components/demos/ActionListDemo/ActionListDemo.tsx
redallen: No, see my earlier comment here: https://github.com/patternfly/patternfly-react/pull/6345#issuecomment-926715093

We do a good job of not breaking types and these demos do a poor job of checking types because the demo app builds in CI 97% of the time over 264 runs. I think it safe to remove, especially because we will have proper example typechecking as soon as we begin conversion.

packages/react-integration/cypress/integration/alert.spec.ts
redallen: Yep! I'll add it back.

packages/react-integration/cypress/integration/applicationlauncher.spec.ts
redallen: Readded.

nicolethoen (approved):

packages/react-integration/demo-app-ts/src/components/demos/ActionListDemo/ActionListDemo.tsx
tlabaj: Maybe we keep them until we do the conversion?  We can open an issue to come back so we don't loose track.

redallen: During the conversion we'd be removing these 89 demos anyways. I think it's bad to keep demos which have no corresponding integration tests, are a copy/paste of existing demos, and offer little type safety. No type problems that we find downstream in products like Openshift are captured by these demos.

As an example, here's `TileDemo`: `<Tile title="Selected" id="selected" isSelected />`. If any of those props were to change in a PR we'd see it in the interface and in existing examples in the docs. Having this demo provides no type safety.

tlabaj: Ok. I see what you are saying.  It was definitely useful before typescript conversion.  So these should be ok to remove.
Did you see my question about testing all variations of breakpoints.  If we remove here, we should cover in unit test.  Other than that we, should be good to go.

redallen: The only breakpoint utility checks I removed are ones that only check the default viewport size. I kept the test that checks all the different viewport sizes.

PR #6347, opened 2021-09-23
feat(JumpLinks): Added toggle text

### LINKED ISSUE #6306: Jump links - move text into expandable toggle
re: https://github.com/patternfly/patternfly/pull/4352/

* Adds `.pf-c-jump-links__toggle-text` element for the toggle button text that goes in the toggle button, after the toggle icon
* Leaves the `.pf-c-jump-links__label` element in place, unless `alwaysShowLabel` is set to `false`. If `false`, we don't need to render the label anymore. 
  * Removes the `.pf-m-toggle` modifier applied with `alwaysShowLabel` since we can just conditionally render the label now based on the prop.

tlabaj:
**What**: Closes #6306 

- Added text to expandable toggle and only shoe label when `alwaysShowLabel` is true.
- Bumped core version (pf-c-jump-links__header removed from core)

### DIFF
--- a/packages/react-catalog-view-extension/package.json
+++ b/packages/react-catalog-view-extension/package.json
@@ -35,7 +35,7 @@
     "clean": "rimraf dist"
   },
   "dependencies": {
-    "@patternfly/patternfly": "4.135.2",
+    "@patternfly/patternfly": "4.135.4",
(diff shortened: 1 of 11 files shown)


mcoker (approved):
LGTM! 👍

redallen (approved):

PR #6351, opened 2021-09-24
[ISSUE-6350] - allow onSelect to prevent the default behavior of toggling node collapse

### LINKED ISSUE #6350: TreeView: Allow onSelect to determine if TreeViewItem node is expanded
Currently there is no programmatic way to prevent a node from be expanded.

I would like to propose that we listen to the return of the `onSelect` callback, if provided, and if it returns a value of `false` do not expand the node.

The onClick behavior in TreeViewListItem would change as follows:
```
              onClick={(evt: React.MouseEvent) => {
                if (!hasCheck) {
                  const expand = onSelect ? onSelect(evt, itemData, parentItem) : true;
                  if (children && expand !== false) {
                    setIsExpanded(!internalIsExpanded);
                  }                  
                }
              }}
```

Further it would be nice to have the capability of `locking` the expansion of an entire tree or subtree.

[CodeSandbox](https://codesandbox.io/s/epic-volhard-lrid3?file=/index.js)

This would be an enhancement.

Catalog Manager, no particular target release

steverhoades:
The proposed usage would allow for the onSelect method to call `evt.preventDefault()` to prevent the nodes from toggling. This is useful for determining programmatically whether a node should be collapsed. Further, it allows for the tree to be expanded and locked by the integrator.

**What**: Closes #6350 

**Additional issues**: None

### DIFF
--- a/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
@@ -23,7 +23,7 @@ export interface TreeViewListItemProps {
   defaultExpanded?: boolean;
   /** Child nodes of a tree view item */
   children?: React.ReactNode;
-  /** Callback for item selection */
+  /** Callback for item selection. Note: calling event.preventDefault() will prevent the node from toggling. */
   onSelect?: (event: React.MouseEvent, item: TreeViewDataItem, parent: TreeViewDataItem) => void;
   /** Callback for item checkbox selection */
   onCheck?: (event: React.ChangeEvent, item: TreeViewDataItem, parent: TreeViewDataItem) => void;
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/TreeView/TreeViewListItem.tsx
```diff
@@ -186,10 +186,10 @@ export const TreeViewListItem: React.FunctionComponent<TreeViewListItemProps> =
               )}
               onClick={(evt: React.MouseEvent) => {
                 if (!hasCheck) {
-                  if (children) {
+                  onSelect && onSelect(evt, itemData, parentItem);
+                  if (children && evt.isDefaultPrevented() !== true) {
```
kmcfaul, line 190: If I'm following this correctly, expansion would be prevented if `evt.preventDefault()` is called within the user-defined `onSelect` right? If so, then could you please update the `onSelect` property description to reflect the new behavior interaction?

redallen (approved):

nicolethoen (approved):

kmcfaul (approved):

PR #6353, opened 2021-09-24
feat(Table): add toolbar sort demos

### LINKED ISSUE #6209: Add sortable table demo
Introduce a new React table demo that includes the responsive behavior introduced in this PR: https://github.com/patternfly/patternfly/pull/4307

kmcfaul:
**What**: Closes #6209

Adds a toolbar sort example to Composable Table and Legacy Table docs alongside the current sortable examples.


packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -31,8 +25,19 @@ import CubeIcon from '@patternfly/react-icons/dist/esm/icons/cube-icon';
...
+import {
+Toolbar,
+ToolbarItem,
+ToolbarContent,
+OptionsMenu,
+OptionsMenuItemGroup,
+OptionsMenuItem,
+OptionsMenuSeparator,
+OptionsMenuToggle
+} from '@patternfly/react-core';
```
redallen, line 40: I don't think you need any of these imports.

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
```diff
@@ -50,8 +50,9 @@ import CubeIcon from '@patternfly/react-icons/dist/esm/icons/cube-icon';
 import LeafIcon from '@patternfly/react-icons/dist/esm/icons/leaf-icon';
 import FolderIcon from '@patternfly/react-icons/dist/esm/icons/folder-icon';
 import FolderOpenIcon from '@patternfly/react-icons/dist/esm/icons/folder-open-icon';
+import SortAmountDownIcon from '@patternfly/react-icons/dist/esm/icons/sort-amount-down-icon';
 
-import { Checkbox, ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
+import { Checkbox, ToggleGroup, ToggleGroupItem, Toolbar, ToolbarItem, ToolbarContent, OptionsMenu, OptionsMenuItemGroup, OptionsMenuItem, OptionsMenuSeparator, OptionsMenuToggle } from '@patternfly/react-core';
```
redallen, line 55: I don't think you need any of these imports.

redallen (dismissed):
Works great, you might be able to remove some imports for cleanliness if you'd like!

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -365,6 +369,159 @@ class SortableTable extends React.Component {
 }
 ```
 
+### Sortable with toolbar
```
nicolethoen, line 372: I think it may be more useful to name this example something like `Sortable - Responsive` so that at the top level it's clear that THIS is the responsive solution for sortable tables.

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
```diff
@@ -351,10 +352,166 @@ ComposableTableSortable = () => {
 };
 ```
 
+### Composable: Sortable with toolbar
```
nicolethoen, line 355: Same comment here about example naming - it can still reference a toolbar since people may like to know that at the top level, too. But I think we should also emphasize the 'responsive' nature of this solution to contrast it to the older 'non-responsive' solution.

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -365,6 +369,159 @@ class SortableTable extends React.Component {
...
+
+  render() {
+    const { columns, rows, sortBy, isSortDropdownOpen } = this.state;
+
+    return (
+      <React.Fragment>
+        <Toolbar id="toolbar">
+          <ToolbarContent>
+            <ToolbarItem>
+              <OptionsMenu
```
nicolethoen, line 467: OptionsMenu! Whew! Haven't seen one of those in a while - would it be better to use a MenuToggle and Menu?

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -365,6 +369,159 @@ class SortableTable extends React.Component {
...
+  }
+
+  render() {
+    const { columns, rows, sortBy, isSortDropdownOpen } = this.state;
+
+    return (
+      <React.Fragment>
+        <Toolbar id="toolbar">
+          <ToolbarContent>
+            <ToolbarItem>
```
nicolethoen, line 466: Should we hide this toolbar item at wider screen widths? LIke only show it if the table headers are hidden?

nicolethoen (changes requested):

packages/react-table/src/components/Table/examples/Table.md
kmcfaul: I thought I needed them because the table was a separate package, so any imports from core had to be specified at the header?

packages/react-table/src/components/Table/examples/Table.md
kmcfaul: Core uses options menu. WDYT @mcoker @mattnolting

packages/react-table/src/components/Table/examples/Table.md
nicolethoen: Actually, Im noticing that this new demo lives under the html-demos for Table. Maybe it should be moved to `React demos` and toggle the control at wider screen widths? and link to the demo from the sortable example where you explain this example is not responsive

kmcfaul: What do you mean by link? Toolbar has its own docs for the customization of visibility and layout, etc so I'm not sure it's necessary here - the example shows more how to control Table sorting with a custom control

kmcfaul: The HTML demo is a bit more full-fledged than this example, having a full toolbar and sitting on a page. Maybe we can add that in a follow up?

nicolethoen: I think that the context may be a little important here. I'm imagining setting it up like this.
There are already sortable table examples - but they are not responsive. So above the sortable table examples we should state that and then link to a Table Demo where we show how to use the sortable table in a whole page, with context and with the responsive solution - like the HTML demos.
Without the context, I think this is a confusing example. What do you think @mcarrano

mcarrano: I agree. Cant we just create a react version of this demo that puts everything in context? https://www.patternfly.org/v4/components/table/html-demos/sortable/  Notice that only when the table headers are hidden the sort control is the toolbar appears.

nicolethoen: I think we could either have two demos - or just implement a composable demo.

but if we have only one demo, we should still link to it from both the legacy and composable examples so that we can still highlight the fact that neither existing examples are responsive.

mcarrano:
That looks great @kmcfaul Just missing the screenshot in the demo page.

mcarrano (approved):
Looks great @kmcfaul !

nicolethoen (changes requested):
The `OptionsMenu` on the new sortable demo is not opening/closing in the responsive view - it's throwing some sort of type error

packages/react-table/src/components/Table/examples/Table.md
tlabaj: I think in this demo we should match what core has (options menu).

packages/react-table/src/components/Table/examples/Table.md
```diff
@@ -365,6 +359,159 @@ class SortableTable extends React.Component {
...
+
+```js
+import React from 'react';
+import {
+  Table,
+  TableHeader,
+  TableBody,
+  sortable,
+  SortByDirection,
+  wrappable,
```
tlabaj, line 374: `wrappable` and  `cellWidth` don't appear to be used. They can be removed.

packages/react-table/src/demos/Table.md
```diff
@@ -2252,6 +2247,388 @@ class FilterTableDemo extends React.Component {
...
+
+ComposableTableSortable = () => {
+  const [navActiveItem, setNavActiveItem] = React.useState(0);
+  const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
+  const [isKebabDropdownOpen, setIsKebabDropdownOpen] = React.useState(false);
+  const [isFullKebabDropdownOpen, setIsFullKebabDropdownOpen] = React.useState(false);
+
+  const columns = ['Column 1', 'Column 2', 'Column 3', 'Column 4', 'Column 5'];
+  const [rows, setRows] = React.useState([
+    ['one', 'two', 'a', 'four', 'five'],
```
tlabaj, line 2311: I would make the values for column 2 four and five unique, so they can be sorted on.

tlabaj (changes requested):
Looks good.  A couple small comments.

kmcfaul:
@nicolethoen I am able to open/close the sort options menu in the surge.

Are you trying to toggle the placeholder "Name..." menu? That one doesn't have anything wired up to it. The sort menu only displays when the screen is small.

nicolethoen (commented):
You're right I was looking at the `Name` options menu filter. I was surprised when it didn't open and saw a typescript error when i tried to open it. I think for clarity, it should work or be removed.

I also think that it'd be wise to explain what you built to make the demo responsive above the demo's screenshot. Note the sort control that appears at smaller screen widths, and note what about the original implementation was not responsive.

nicolethoen (dismissed):

tlabaj (approved):

mcoker (commented):
In the legacy example, the "last commit" column is in the sort dropdown but doesn't have a sort column button - is that OK as it is or do we need to make an update? cc @mcarrano 
Screen Shot 2021-10-05 at 3 43 37 PM

mcoker (approved):
LGTM!!

tlabaj (approved):

nicolethoen (approved):

PR #6358, opened 2021-09-24
feat(LogViewer): improve log viewer performance

### LINKED ISSUE #6327: LogViewer: Scroll may cause page crash
If there are a lot of unrendered rows below, and you scroll down very quickly, the page will crash with the error "Maximum update depth exceeded".

Bug.

DaoDaoNoCode:
**What**: Closes #6327 

1. Revert `DynamicSizeList` to `VariableSizeList` because `DynamicSizeList` has a lot of calculations, which will highly impact the performance of the log viewer, and even cause page crashes sometimes.
2. Create a dummy row to get the font and `line-height` of the text in the log viewer, so that could precisely estimate the height of every item.

**Additional issue**:
The estimation is not 100% accurate, although it's very precise, it can wrongly estimate in some extreme situations, but this rarely happens. All in all, it could work pretty well.


nicolethoen (approved):
Nice!

jschuler (approved):
LGTM!

Issue #6268, opened 2021-08-19
Feedback on TreeView


ianw:
Hello,

The beta component requests feedback so I thought I would provide it.

I have recently re-implemented https://zuul.opendev.org/t/openstack/jobs using TreeView

The change is https://review.opendev.org/c/zuul/zuul/+/804956

Our CI provides a "live" view of this.  If you go to the above change and select "Zuul Summary -> zuul-build-dashboard-opendev" that will take you to a job run, and then select "artifacts -> site preview".  Then click on "Jobs" next to "openstack"

The biggest problem is probably performance for expansion.  

Firstly I wanted to emulate the old page by bringing it up already fully expanded.  That killed the browser (setting `allExpanded` by default).  If it would be helpful, I could propose a change with this enabled and using the live preview model above it could be profiled to see if there is anything that can be done.

I had to make the search only start completing after 6 characters, because trying to expand when you're typing just a few characters kills the browser.  This makes sense; I mean tying a vowel like "a" is basically the same as expanding the entire tree, the issue above.  One thought was perhaps there could be an example of taking the key inputs in the search box, and not triggering the filtering update until maybe 1 second after the last keypress or something?  basically give you time to type in your value before searching.  Or, I guess there should be a search model that is more -- synchronous?  As in it doesn't start searching until a button is pushed, and puts up a spinner. I could not really see that the "Search Input" has a way to do this, although I'm sure it's possible (an example would help :)

The other issue was the extant page has a "flat" view.  When I tried doing this -- flattening the input tree -- the browser totally died (I think it's about 5000 entries; with no children obviously in flat mode).  On the original page clicking it takes a few seconds.

If providing the underlying data as a json, or making changes and providing a site preview will help maybe diagnose anything we can do; please let me know.

On the search bar 

* it would be good to expand the documentation so that it's clear `searchProps` is the argument to the `Search Input` component.  It took me a while to figure out that I could use `placeholder` here.  It would probably be useful to add this to the example.
* I couldn't figure out how to get the `X` to clear the current search icon to come up?  I think it would be good if this just happened by default, or the examples showed how to make this work.

None of the examples showed putting `React.Fragments` in the `name` field.  So in my `onSearch` event I'm matching against the `id` (because `name` is a bunch of react objects).  Maybe this is "assumed knowledge" but again, an example of items with a badge next to them or something might be good.

I think it looks good though!  Hoping to get it through review and into production soon.

nicolethoen:
[...]

PR #6362, opened 2021-09-27
perf(TreeView): add memoization + test demo

kmcfaul:
**What**: Possible improvement for #6268

In this PR: 
- Adds memoization to TreeViewListItem
- `useMemo` and the new `Tree view` demo page under the `Demos` section are for demonstrating this memoization only and should not be merged. I will remove them if we want to proceed with the memo changes.

`TreeView` will completely re-render all nodes when the `onSelect` is called and a state update is made to `activeItems`, because a state update is being made at the top level of the recursive tree view. To help performance, I've added memoization to the TreeViewListItem to alleviate some of the unnecessary re-rendering - items should only re-render when their active state is going to change. 

The main caveat in adding this memoization logic is due to TreeView's recursive nature, parent nodes that do not see an update to their rendering will not re-render (or even evaluate) child nodes below them, and re-rendering when the children prop updates defeats the purpose as the whole tree would re-render. The current working solution has `activeItems` change slightly - instead of passing only the selected item, all nodes on the path from the top node to the selected node must be passed to this property. This doesn't change `activeItems` effective behavior (only the leaf node's css will change as usual) but it would be a breaking change to the current demos (unless the `useMemo` prop is kept and has a default value of `false`, which would disable memoization unless it is required).

Any other thoughts and ideas regarding tree view performance is welcome!


kmcfaul:
An unrelated but possible alternative for large tree view datasets would be to "mock" a tree-like view using a simple list structure instead of using recursion - that would also separate the state out from the stack. This would require a way to provide indentation to a nested `ul` based on depth, which I'm not sure about but may be possible?

nicolethoen:
I can definitely notice a difference. I'm not sure how much more noticeable it'll be with more complicated treeListItems. And I don't know if the improvement is as good as the customers will want.

As far as memoization goes, you're saying that you need to evaluate if activeItems has changed to know if the children should rerender, right? And mayyyyybe we could further improve this by finding a less expensive way to make that determination?

kmcfaul:
Yeah that's why I think it may be worth also investigating if we can make a faux-tree view with a simple list over the recursion.

For the memoization yes. `activeItems` is really the only prop that will warrant a rerender since the folder expansion is internal so that's what I used for the check logic.

Issue #5392, opened 2021-01-28
DualListSelector: Allow to control selected options


rvsia:
**Describe the issue. What is the expected and unexpected behavior?**

Selected (not added) values cannot be controlled, so for example when a custom sorting is implemented, the old indexes are still selected. 

![sorting](https://user-images.githubusercontent.com/32869456/106115108-acd74880-6150-11eb-9969-ff2dffc9607f.gif)

**Steps to reproduce**

1. Change order of options when they are selected.

**What is your product and what release version are you targeting?**

Data Driven Forms

nicolethoen:
I don't think this should be marked stale.

Issue #5231, opened 2020-12-08
Update DualListSelector with trees/checkboxes to be keyboard and screen reader accessible


jessiehuff:
**Describe the issue. What is the expected and unexpected behavior?**
Currently the DualListSelect with trees/checkboxes is inaccessible by keyboard and screen reader. I would expect the keyboard interaction to follow the [authoring practices for treeview](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-23) with multi-select using left/right arrow keys to handle expansion, down/up arrow keys to handle focus between options, and Enter/Space for node selection. See the above documentation for the full interaction expectations.

This issue is related to the [treeview keyboard interactivity issue](https://github.com/patternfly/patternfly-react/issues/4726) as they should follow the same interaction pattern/be consistent and the [dual list screen reader issue](https://github.com/patternfly/patternfly-react/issues/5149). 

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

Attempt to navigate and use the [dual list select with trees/checkboxes ](https://patternfly-react-pr-5207.surge.sh/components/dual-list-selector#expandable-options-with-checkboxes)with a keyboard or screen reader.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
bug

Issue #5149, opened 2020-11-17
Update Dual List Selector to work in screen reader


jessiehuff:
**Describe the issue. What is the expected and unexpected behavior?**
Currently the dual list selector doesn't work in VO or JAWS. 
My guess is the `<li><button>` structure is causing issues:

```<div role="listbox">
  <ul>
    <li role="option">
      <button>
```
For focus management to work, the element with `role="option"` should be getting focus, but it looks like the button is what gets focus. Also, the element with `role="listbox"` should have an `aria-describedby` that points to the text stating how many items are selected (i think, assuming this isn’t announced automatically with appropriate aria attributes).
For managing focus, I wonder if it’s a cleaner experience to use `aria-activedescendent`, and keep focus always on the `role="listbox"`.

I think part of the issue is the core example uses buttons in the list box. The WAI-ARIA example has just a `<ul><li>` DOM structure. I haven’t looked into why all of these `<span>`s are nested inside the button, if they’re needed for styling purposes, but I think we should try to match what the WAI-ARIA example has. So this would mean:
moving attributes from the parent div to the child ul, and make sure any other aria attributes needed for the ul are included
changing the button to a span, and moving it’s attributes to the parent li, and adding any other aria attributes needed for the li

However, all of these theories should be tested. We should adapt the aria attributes and the structure and see if we can get this to function in a screen reader.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

Open the dual list selector in VO and attempt to use the dual list selector.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug

tlabaj:
cc @mcoker

nicolethoen:
The tree view implementation of the dual list selector does not use a button in the `<li>` at all. At the very least consistency should be considered. That will probably help with accessibility and consistent interaction.

PR #6369, opened 2021-09-28
[WIP]feat(DualListSelector): introduce composable example and drag/drop demo

### LINKED ISSUE #6265: Dual list selector: Enable optional drag and drop reordering of selected items
Want to enable items in the right-hand list to be manually ordered via drag & drop per the design here:

image

nicolethoen:
closes #6265 #5149 #5231 #5392

- need to implement checking checkboxes on composable tree list
- need to implement moving options between tree lists
- need to update styles for drag/drop
- need to make sure keyboard navigation isn't glitchy


jessiehuff (approved):
This looks pretty good!!! Awesome job! 😄  I tested in VO, NVDA, and JAWS, and I was able to interact with it in each case. The only consistent thing I noticed being difficult is the Composable dual list selector tree. For that one, I wasn't able to expand the options for the tree view items. Interestingly enough I was able to expand them in NVDA but none of the others. I think we could have a follow up issue on that though since we'd like to get this in!

nicolethoen:
> This looks pretty good!!! Awesome job! 😄 I tested in VO, NVDA, and JAWS, and I was able to interact with it in each case. The only consistent thing I noticed being difficult is the Composable dual list selector tree. For that one, I wasn't able to expand the options for the tree view items. Interestingly enough I was able to expand them in NVDA but none of the others. I think we could have a follow up issue on that though since we'd like to get this in!

@jessiehuff I am going to remove the composable tree example since it still has a decent amount of functionality in the example code that I haven't ironed out. 
I think it's possible to build though - i've made the underlying components ready to be used composably, so I will open a follow up issue to track building a composable tree example. But given its complexity, I'm not sure if we will even have products asking for that anyway.

packages/react-core/src/components/DualListSelector/DualListSelectorListItem.tsx
```diff
@@ -9,70 +13,78 @@ export interface DualListSelectorListItemProps extends React.HTMLProps<HTMLLIEle
...
+      id={id}
+      ref={ref}
+      role="option"
+      tabIndex={-1}
+      {...props}
+    >
+      <div className={css(styles.dualListSelectorListItemRow, isSelected && styles.modifiers.selected)}>
+        {isDraggable && (
+          <div className={css(styles.dualListSelectorDraggable)}>
+            <Button isDisabled variant={ButtonVariant.plain} aria-label={draggableButtonAriaLabel}>
```
mcoker, line 69: why is this disabled? If the item is draggable, I would expect it to be enabled.

mcoker (commented):
Just one thing. I'll re-review once you rebase with the new draggable/droppable component styles.

packages/react-core/src/components/DualListSelector/DualListSelectorListItem.tsx
nicolethoen: I will play with it more - the issue is that we dont have any keyboard navigation implemented for drag/drop yet and the button trapping focus is pretty confusing. I had thought Zack also disabled this button in the draggable datalist - but i see now that he didn't - he only made sure no events could fire on the buttons - I'll see if I can do the same

packages/react-core/src/components/DualListSelector/DualListSelectorListWrapper.tsx
```diff
@@ -0,0 +1,123 @@
...
+          focusedOption,
+          ariaLabelledBy,
+          displayOption,
+          selectedOptions,
+          id,
+          options,
+          onOptionSelect
+        }}
+      >
+        {children ? children : <DualListSelectorList />}
```
tlabaj, line 110: So if there are no `children` render empty `DualListSelectorList`?  What happens if we don't render anything.

tlabaj (commented):
So, the component is Beta.  I am wondering if there is benefit for keeping the non composition (props) approached around.  Is this being heavily used?
cc @redallen

nicolethoen:
@tlabaj I'd guess that just from the number of people who have been opening issues, that the dynamically built version is used heavily enough that we likely need to keep it - and keep it working - for now? but it's handy that we can hide the props that are not used for the composable approach.

packages/react-core/src/components/DualListSelector/DualListSelectorListWrapper.tsx
nicolethoen: The list should have children in the case of the composable dual list selector.
The list uses it's options prop (passed via context) if it's a dynamically built dual list selector.
if there are no children and no options prop, it just renders an empty list.

packages/react-core/src/components/DualListSelector/DualListSelectorListWrapper.tsx
```diff
@@ -0,0 +1,123 @@
...
+        !Array.from(menuRef.current.getElementsByClassName('pf-c-dual-list-selector__menu')).includes(
+          (event.target as HTMLElement).closest('.pf-c-dual-list-selector__menu')
+        ))
+    ) {
+      return;
+    }
+    event.stopImmediatePropagation();
+    const validOptions = isTree
+      ? (Array.from(
+          menuRef.current.querySelectorAll(
```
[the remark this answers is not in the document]
nicolethoen: if menuRef.current is not defined, the function will have already returned before this line because of the logic on line 49.

redallen:
@tlabaj @nicolethoen The only in-product usage of DualListSelector is in insight's settings-rbac's [edit resource definitions](https://github.com/RedHatInsights/insights-rbac-ui/blob/4fc9dc38afb052a77d876aa4868ba64e66c6dc26/src/smart-components/role/edit-resource-definitions-modal.js#L182) which comes from the data-driven-forms [pf4-component-mapper.](https://github.com/data-driven-forms/react-forms/blob/master/packages/pf4-component-mapper/src/dual-list-select/dual-list-select.js#L96)

mcarrano (changes requested):
@nicolethoen just noticing one bug with the drag & drop. If I click directly on the drag icon (waffle) to drag the row, releasing the mouse does not drop it.

nicolethoen:
@mcarrano I am looking into that bug now - it seems to be firefox specific

@tlabaj I opened this issue: https://github.com/patternfly/patternfly-react/issues/6401 to track changing all dual list selector examples to be composable.

tlabaj (dismissed):

dlabrecq:
I know this is beta, but there are a ton of breaking changes. Cost Management recently implemented the dual-list tree component for the Insights settings page. The non-composition approach seemed to be stable for such a long time, so we felt confident it wouldn't be replaced.

cc @ddonahue007

ddonahue007:
> @tlabaj @nicolethoen The only in-product usage of DualListSelector is in insight's settings-rbac's [edit resource definitions](https://github.com/RedHatInsights/insights-rbac-ui/blob/4fc9dc38afb052a77d876aa4868ba64e66c6dc26/src/smart-components/role/edit-resource-definitions-modal.js#L182) which comes from the data-driven-forms [pf4-component-mapper.](https://github.com/data-driven-forms/react-forms/blob/master/packages/pf4-component-mapper/src/dual-list-select/dual-list-select.js#L96)

Cost management has the current Dual List in pre-production now(stage), I wondering that this rework might break us as well.

mcoker (approved):
L🎃TM!!!

PR #6374, opened 2021-09-29
feat(react-console): accept random children in AccessConsole component

KKoukiou:
So far we were supporting children of three types,
SerialConsole, DesktopViewer and VncConsole.

This component is consumed by `cockpit machines` product and it's
used for exposing consoles of Virtual Machines.
A Virtual Machines however can have multiple consoles of one type; for example 2
different serial consoles, something which was not supported until this
commit.

For this reason we need to extend this component to accept also custom
children apart from a single from each from the predefined types.

Also start respecting the order of the children to display the options
in the AccessConsoles dropdown.

**What**: Closes #

**Additional issues**:


packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx
```diff
@@ -30,7 +30,7 @@ export interface AccessConsolesProps {
   children?: React.ReactElement[] | React.ReactNode;
   /** Placeholder text for the console selection */
   textSelectConsoleType?: string;
-  /** The value for the Serial Console option */
+  /** The value for the Serial Console option. This can be overriden by the type property of the child component */
```
pcbailey, line 33: Can there only be multiples of the serial component? If the others can have multiple instances as well, then I think we should either add a general comment explaining how to override the option value for any of them or add this comment to each of them.

pcbailey (commented):
I had one question, but other than that it looks good to me.

PR #6387, opened 2021-09-30
fix(log-viewer): accurately estimate row height

### LINKED ISSUE #6386: Log viewer: Row height estimation could be inaccurate in some situations
This issue was also mentioned in #6358 

Bug.

DaoDaoNoCode:
**What**: Closes #6386 
Rewrite the estimation algorithm. Because text width is not always divisible by window width, so there are always some spaces at the end of every line, which may cause inaccurate estimation. To overcome this, we could first calculate the maximum characters in a single line, then use the characters numbers in the text to calculate how many rows are there. Thus, we could get an accurate estimation. This method requires users to use monospaced fonts. Also, it needs core version 4.141.1 to be bumped.

**Additional issues**: As mentioned above, this method requires users to use monospaced fonts.


redallen (approved):
Well done! Most terminals are implemented in terms of how many characters wide they are like this.

nicolethoen (commented):
Im curious why you changed the test data by so much? 
We were given that test data by OpenShift so that we could test more performance and the virtualized nature of the log viewer. It was a large file on purpose.

DaoDaoNoCode:
@nicolethoen Hi, the way we estimate the row height decides that we need to show all the characters inside the log text, so all the spaces at the beginning and end of the text would be shown. What I did was just remove those spaces in the test data to make it looks better in format 🙂

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -99,36 +91,37 @@ const LogViewerBase: React.FunctionComponent<LogViewerProps> = memo(
...
-          containerRef.current.removeChild(dummyText);
           setFirstMount(false);
         }
+        const dummyText = document.createElement('span');
+        dummyText.className = css(styles.logViewerText);
+        containerRef.current.appendChild(dummyText);
+        const dummyTextStyles = getComputedStyle(dummyText);
+        setLineHeight(parseFloat(dummyTextStyles.lineHeight));
+        const lineWidth = hasLineNumbers
+          ? (containerRef.current as HTMLDivElement).clientWidth - 97
```
jschuler, line 117: is 97 the width of the line numbers column? Does it change depending on zoom level? Can you save it to a variable?

DaoDaoNoCode: @jschuler Nice catch! It changes with the base font. I will update this, thanks!

redallen (approved):
Even better!

jschuler (approved):

PR #6408, opened 2021-10-06
fix(react-tokens): parse var(--pf, initialValue)

redallen:
**What**: Fixes #6400

**Additional issues**:


tlabaj (approved):
Looks good. Thanks Zack!

Issue #4416, opened 2020-06-19
Table - Tooltips applied to truncated table elements does not parse content correctly


nicolethoen:
If anything more complex than a string is rendered in the table cell, then the Tooltip does not parse the content correctly. A fix would need to be made when the Tooltip is applied to truncated BodyCell, HeaderCell, and TableText content.

![image](https://user-images.githubusercontent.com/26305082/85140841-da036600-b213-11ea-92be-6613fb23f58e.png)

PR #6416, opened 2021-10-06
feat(Navigation): add flyout with menu components

kmcfaul:
**What**: Closes #

Adds back in flyouts using `Menu`. 

- Keyboard navigation differs slightly in that `Menu` uses vertical arrow keys to traverse items while `Nav` does not, so right now only the flyouts have vertical arrow navigation.
- Includes a fix for the previous implementation's performance (event calls was setting a state unnecessarily).
- Keeps the changes to `variant` that allow a `subnav` `section`. Is just a new option and not breaking but lmk if I should remove it.
- There is CSS in the example code that sets the first menu offset in core and react - is this ok or should it be baked into a css class? Setting this positioning also appears to be breaking the isomorphic layout (bouncing back on screen) of the Menu component - @redallen Thoughts?

### DIFF
--- a/packages/react-core/src/components/Nav/examples/nav.css
+++ b/packages/react-core/src/components/Nav/examples/nav.css
@@ -12,13 +12,3 @@
 #ws-react-c-navigation-flyout {
   width: 300px;
 }
-
-#ws-react-c-navigation-flyout .pf-c-menu.pf-m-flyout {
-  position: absolute;
-  top: 0;
-  left: 100%;
-}
-
-#ws-react-c-navigation-flyout .pf-c-nav__item.pf-m-flyout {
-  position: relative;
-}
--- a/packages/react-integration/demo-app-ts/src/components/demos/NavDemo/NavDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/NavDemo/NavDemo.tsx
@@ -15,8 +15,6 @@ import {
   MenuItem
 } from '@patternfly/react-core';
 
-import './nav.css';
-
 interface SelectedItem {
   groupId: number | string;
   itemId: number | string;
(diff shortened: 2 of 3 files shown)


redallen:
> There is CSS in the example code that sets the first menu offset in core and react

Yeah I think consumers having to do that themselves is bad. If we always want to apply this rule to flyout menus in nav, I think there should be a rule upstream in Core:
```css
#ws-react-c-navigation-flyout .pf-c-menu.pf-m-flyout {
  position: absolute;
  top: 0;
  left: 100%;
}
```

> Setting this positioning also appears to be breaking the isomorphic layout (bouncing back on screen) of the Menu component

Hm, I'm not sure exactly why that's the case. The elements seem to be positioned similarly. I can start debugging tomorrow.

nicolethoen:
It looks like the 'subnav' is not being used in any examples? So it seems like it's probably not necessary. But overall, I think it looks good! 

For the sake of this release, if that css needs to be added to core in a follow-up, I think that'd be okay - but we should not keep it in the examples if we can help it.

redallen:
I don't think this example is using Popper. If we could avoid using it I think that'd be good. Usually we use Popper to display over top the page, but `pf-c-nav` already overlays the page content.

mattnolting:
With the addition of [#4416](https://github.com/patternfly/patternfly/issues/4416), the menu component will require base level styling updates as the direct child of `pf-c-nav`, meaning `pf-c-nav > pf-c-menu` will emulate all current `pf-c-nav` styling (blue left border, hover styles, etc). Nav, at that point, will essentially serve as a wrapper/theming element. We should add the demo css to `.pf-c-nav__item.pf-m-flyout` for now. 



```
.pf-c-nav
  .pf-c-menu.pf-m-drilldown
```

 all base level menu within `nav` will require base styling.

mcarrano:
Looks good other than the bug that @mcoker noted. Is that fix something that we will be able to pull in?

PR #6418, opened 2021-10-07
fix(react-tokens): fix some values by reading variables last

### LINKED ISSUE #4583: Some react tokens have wrong values
For example `global_BackgroundColor_100`, 
```
export const global_BackgroundColor_100 = {
  "name": "--pf-global--BackgroundColor--100",
  "value": "#151515",
  "var": "var(--pf-global--BackgroundColor--100)"
};
export default global_BackgroundColor_100;
```

The value should be `#fff`

redallen:
**What**: Closes #4583

**Additional issues**:

### DIFF
--- a/packages/react-tokens/scripts/generateTokens.js
+++ b/packages/react-tokens/scripts/generateTokens.js
@@ -1,5 +1,5 @@
 const glob = require('glob');
-const { dirname, basename } = require('path');
+const { dirname, basename, sep } = require('path');
 const { parse } = require('css');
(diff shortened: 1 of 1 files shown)


jschuler:
I've checked some of the changes and they looked good. Here's a diff:
https://diffy.org/diff/5d01cd6eddd3
It's technically a fix, but do you think there is the possibility someone might be incorrectly depending on the wrong value? wonder if we should consider it potentially breaking and do a major bump...

PR #6472, opened 2021-10-18
fix(Label): Make editable label more screen reader accessible

### LINKED ISSUE #6096: Investigate why JAWS and NVDA have odd interaction with editable label
When using JAWS and NVDA, they require two enter key presses in order to interact with the editable label. I'm curious why this is happening, and if there is a way to make it simpler/more straight forward for users. I worry that screen reader users wouldn't think to press the key twice to interact with it. 

My current theory is that perhaps screen readers are opening up a different mode (like forms mode) when pressing enter initially and then they have to press it again to get the interaction. Not sure how to get around this yet.

Could be argued as either. Technically it's accessible so you could argue that this is an enhancement, but I think that users who come across this would think it's a bug.

jessiehuff:
**What**: Closes #6096

This changes the editable label div to go between being a button and an input, allowing the elements to be more semantic and more accessible in a screen reader. Core updates likely needed to fix the styling.

### DIFF
--- a/packages/react-core/src/components/Label/Label.tsx
+++ b/packages/react-core/src/components/Label/Label.tsx
@@ -91,7 +91,8 @@ export const Label: React.FunctionComponent<LabelProps> = ({
   ...props
 }: LabelProps) => {
   const [isEditableActive, setIsEditableActive] = useState(false);
-  const editableDivRef = React.createRef<HTMLDivElement>();
+  const editableDivRef = React.createRef<HTMLButtonElement>();
+  const editableInputRef = React.createRef<HTMLInputElement>();
 
   React.useEffect(() => {
     document.addEventListener('click', onDocClick);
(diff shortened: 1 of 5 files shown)


nicolethoen:
It is looking good! I am noticing that when I go into edit the label and clear the contents, it reverts back to the previous value. I am not positive what I expect to happen, but I don't think I expected that. Should clearing the contents remove the label? Do you think that's right the way it is? @kmcfaul @jessiehuff

packages/react-core/src/components/Label/Label.tsx
```diff
@@ -91,7 +90,8 @@ export const Label: React.FunctionComponent<LabelProps> = ({
   ...props
 }: LabelProps) => {
   const [isEditableActive, setIsEditableActive] = useState(false);
-  const editableDivRef = React.createRef<HTMLDivElement>();
+  const editableDivRef = React.createRef<HTMLButtonElement>();
+  const editableInputRef = React.createRef<HTMLInputElement>();
```
DaoDaoNoCode, line 94: Will it be better to use `React.useRef` here so that refs will not be re-created on every render?

mcarrano (approved):
This looks good. Thanks @jessiehuff

mcarrano:
> Hmm what do you think @mcarrano? Do you think there's a scenario in which a user tries to remove the label by deleting the input? I think that the user would need to click on the close button for that?

I agree

dlabrecq (approved):

srambach:
I'm not sure why, but if you open the code window below, the label will no longer become editable with a mouse click.
Also, double clicking in some areas outside the labels can cause text within to highlight. I suspect this is somehow related to the hidden input being longer than the label...maybe?

nicolethoen:
I think the issue caused by the length of the hidden input may need to be fixed on the core side, @srambach. But I did fix the weird issue with it not always going into edit mode.

srambach:
I'm not seeing the problem with the code editor open now. However, I don't think this always meets minimum click size standards. For example, if you enter just an "i" into the field, to get it to edit again requires you hit a width of about 2 pixels. Should there be a minimum width of the clickable area to ensure a11y? I'd honestly expect anywhere within the colored area and "not too close to the close button" would put you into edit mode but there's a good bit of space around the text that doesn't trigger it.
I'm also still confused by why clicking in the left example can highlight text in the right example. Using the inspector, I don't see anything that overlaps.

(note that somehow snagit is recording the cursor a pixel or so lower than I see it live...?)

mcarrano:
> However, I don't think this always meets minimum click size standards. For example, if you enter just an "i" into the field, to get it to edit again requires you hit a width of about 2 pixels. Should there be a minimum width of the clickable area to ensure a11y? I'd honestly expect anywhere within the colored area and "not too close to the close button" would put you into edit mode but there's a good bit of space around the text that doesn't trigger it.

This is a good point @srambach . Is this something we can address? I'm also wondering is the selection issue you mention above is because there is no space between the labels as in other examples.

srambach:
If you move them onto different lines, I guess it's just the normal double-click and it selects the next thing. But it's kind of weird when they are next to each other and it looks like you are on one label when I guess it's not actually getting the click.

nicolethoen:
@srambach, @mcarrano  I think click area is something that'd have to be addressed on the core side.
@mcoker do you have any thoughts about the concerns raised?

mcoker:
> I don't think this always meets minimum click size standards.

There are a couple of things we can do about that. If you look at the link labels, the `__content` element is what becomes the link and we use a `::before` pseudo element attached to `__content` to fill the entire width/height of the label so that the whole label will trigger a click of the label (except for the close button if that is present, since the close button sits on top of the `__content` element).

Are we able to move the `button` element to `__content` - and potentially remove the need for `__editable-text`? That would make the whole label clickable. Alternatively, we could use the same `::before` pseudo element trick that we use on `__content` on the `__editable-text` element, too, doing the same thing. Either of those would make a single letter label's click target about the size of the close button.

> Also, double clicking in some areas outside the labels can cause text within to highlight.

I'm seeing that with regular labels, too, unless you're referring to something different. As a chronic double-clicker on everything as a nervous habit kind of person, I think that's just how the browser works - double clicking around something usually selects text within it.

mcarrano:
So should we open a core issue to rework this a bit? I can get that into the upcoming milestone. That said, I think we can move forward with the PR. Do others agree?

nicolethoen:
I agree! I opened this core issue for follow up work: https://github.com/patternfly/patternfly/issues/4626

tlabaj (approved):

PR #6496, opened 2021-10-25
fix(acccessconsoles): use object type Select options values in order to not mix displayed text with the option keys

KKoukiou:
Commit 7be4bc4a92e208073ca3e38415df1f64a89506ee introduced a bug where
the displayed SelectOption label is the key and not the expected text
after changing option.


pcbailey:
Looks good to me! =)

KKoukiou:
Friendy ping - could I get from someone a second review here?

nicolethoen (approved):

tlabaj (approved):

Issue #6454, opened 2021-10-14
ApplicationLauncherGroup does not forward custom props to DOM


jschuler:
ApplicationLauncherGroup does not forward custom props to DOM

Looking at the code, it forwards it to the underlying DropdownGroup component, but then DropdownGroup does not accept other props.

For example, i cannot do
```
<ApplicationLauncherGroup onMouseOver={() => console.log('mouse over)}>Group</ApplicationLauncherGroup>
```

boaz0:
@tlabaj / @mcarrano can you close this issue as #6521 was merged?
Thanks!

tlabaj:
closed by #6521

PR #6521, opened 2021-10-28
fix(DropdownGroup): Forward custom props to the DOM

Th1nkK1D:
**What**: From issue #6454, I `forward DropdownGroup` custom props to the DOM. Now custom props can also be given to `ApplicationLauncherGroup`.

**Additional issues**: This is a simple change, should not create any issues


jschuler (approved):
Thank you, LGTM

nicolethoen (approved):

PR #6522, opened 2021-10-29
Add option to hide topology connector menu on drag end

rottencandy:
**What**:
Currently the only way to manually hide connector context menu on drag end in topology is by returning null/empty array from the `onCreate` callback. But it is not possible to do so when rendering custom react elements (i.e. not using `ContextMenuItem`) for the choices.

The PR adds an additional option to the connector options which hides the connector menu on drag end when the option is set.


nicolethoen:
Hi @rottencandy! Thank you so much for the contribution. 
We like to make sure every PR we merge has an associated issue attached for tracking purposes, so I've created one and attached it to your PR :)

nicolethoen (approved):

jeff-phillips-18 (approved):

Issue #6463, opened 2021-10-15
Table: support nested headers


mcarrano:
Suppoet the ability to nest column headers in a table. Core PR: https://github.com/patternfly/patternfly/pull/4448

PR #6538, opened 2021-11-01
feat(ComposableTable): add sticky columns and nested headers

### LINKED ISSUE #6163: Table: support horizontal scrolling with sticky columns
Want to allow for horizontal scrolling of a table when the width exceeds the width of its container. This should also include the ability to make some columns sticky (i.e. they don't scroll).

Core issue with requirements: https://github.com/patternfly/patternfly/issues/4218

kmcfaul:
**What**: Closes #6163 and #6463

Sticky columns:
- Adds `isStickyColumn`, `hasRightBorder`, `stickyMinWidth`, `stickyLeftOffset` (all Th) properties
- Adds `OuterScrollContainer` and `InnerScrollContainer` wrapper components to use for scrollable tables
- Adds new examples showing a single sticky column, multiple sticky columns, and sticky columns with a sticky header + descriptions in each example for prop usage

Nested headers:
- Adds `nestedHeaderColumnSpans` (TableComposable), `isSubheader` (Th), `hasNestedHeader` (Thead), `resetOffset` (Tr) properties
- Adds new examples showing a basic nested header and a nested header with expandable rows + descriptions for prop usage

Update - added nested header support as it builds on top of properties added by sticky columns

### DIFF
--- a/packages/react-table/src/components/TableComposable/Th.tsx
+++ b/packages/react-table/src/components/TableComposable/Th.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Table/table';
+import scrollStyles from '@patternfly/react-styles/css/components/Table/table-scrollable';
 import { info } from '../Table/utils/decorators/info';
 import { sortable, sortableFavorites } from '../Table/utils/decorators/sortable';
 import { selectable } from '../Table/utils/decorators/selectable';
@@ -38,6 +39,14 @@ export interface ThProps
   info?: ThInfoType;
   /** Adds scope to the column to associate header cells with data cells*/
   scope?: string;
+  /** Indicates the header column should be sticky */
+  isStickyColumn?: boolean;
+  /** Adds a border to the right side of the cell */
+  hasRightBorder?: boolean;
+  /** Minimum width for a sticky column */
+  stickyMinWidth?: string;
+  /** Left offset of a sticky column */
+  stickyLeftOffset?: string;
 }
 
 const ThBase: React.FunctionComponent<ThProps> = ({
@@ -56,6 +65,10 @@ const ThBase: React.FunctionComponent<ThProps> = ({
   visibility,
   innerRef,
   info: infoProps,
+  isStickyColumn = false,
+  hasRightBorder = false,
+  stickyMinWidth = '100px',
+  stickyLeftOffset,
   ...props
 }: ThProps) => {
   const [showTooltip, setShowTooltip] = React.useState(false);
@@ -132,11 +145,20 @@ const ThBase: React.FunctionComponent<ThProps> = ({
       className={css(
         className,
         textCenter && styles.modifiers.center,
+        isStickyColumn && scrollStyles.tableStickyColumn,
+        hasRightBorder && scrollStyles.modifiers.borderRight,
         modifier && styles.modifiers[modifier as 'breakWord' | 'fitContent' | 'nowrap' | 'truncate' | 'wrap'],
         mergedClassName
       )}
       {...mergedProps}
       {...props}
+      {...(isStickyColumn && {
+        style: {
+          '--pf-c-table__sticky-column--MinWidth': stickyMinWidth ? stickyMinWidth : undefined,
+          '--pf-c-table__sticky-column--Left': stickyLeftOffset ? stickyLeftOffset : undefined,
+          ...props.style
+        } as React.CSSProperties
+      })}
     >
       {transformedChildren}
     </MergedComponent>
--- a/packages/react-table/src/components/TableComposable/Th.tsx
+++ b/packages/react-table/src/components/TableComposable/Th.tsx
@@ -45,7 +45,7 @@ export interface ThProps
   hasRightBorder?: boolean;
   /** Minimum width for a sticky column */
   stickyMinWidth?: string;
-  /** Left offset of a sticky column */
+  /** Left offset of a sticky column. This will typically be equal to the combined value set by stickyMinWidth of any sticky columns that precede the current sticky column. */
   stickyLeftOffset?: string;
 }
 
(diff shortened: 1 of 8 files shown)


kmcfaul:
@mattnolting There was a small spacing difference when I implemented the multiple sticky columns example. The first sticky column in core had a width of 100px while in react it was ~106px. I had to adjust the min-width and offset to prevent the sticky columns from shifting around slightly on scroll. I'm not sure what's different about the spacing and it's a small difference, but I figured I'd call it out in case anything jumped out to you

kmcfaul:
Updated demos according to core update PR https://github.com/patternfly/patternfly/pull/4489

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
```diff
@@ -2010,3 +2011,709 @@ class DraggableTable extends React.Component {
...
+                        <Th key={columnIndex} isStickyColumn modifier="truncate" {...sortParams}>
+                          {column}
+                        </Th>
+                      );
+                    } else if (columnIndex === 1) {
+                      return (
+                        <Th
+                          key={columnIndex}
+                          isStickyColumn
+                          stickyMinWidth="80px"
```
nicolethoen, line 2658: I'm noticing a gap between the first and second sticky columns in the last new example. 
Could you also add just a little more detail in the prop descriptions for the sticky minwidth and sticky left offset props? I see the explanation in the examples and it's super helpful, but even just one additional sentence to explain the props' intended uses would be helpful in the api descriptions would help a lot i think.

nicolethoen (changes requested):

mcarrano (commented):
@kmcfaul Other than he gap issue that @nicolethoen raised, this looks good. 

@dlabrecq can you also review to ensure it's what you were expecting?

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
kmcfaul: Oops missed updating the width. Will also add the explanation to the prop description.

nicolethoen (approved):

dlabrecq (commented):
@mcarrano The changes look ok, but Cost Management doesn't use the composable table. We will have to continue with our current workaround.

Ideally, this would be available for all tables?

kmcfaul:
I believe the css only supports ComposableTable currently. This will be the same case for nested column headers

packages/react-table/src/components/TableComposable/Th.tsx
```diff
@@ -56,6 +67,11 @@ const ThBase: React.FunctionComponent<ThProps> = ({
   visibility,
   innerRef,
   info: infoProps,
+  isStickyColumn = false,
+  hasRightBorder = false,
+  stickyMinWidth = '100px',
```
mattnolting, line 72: Can we increase this to `120px`?

mattnolting (approved):
@kmcfaul Looks great! Also looks like you already removed the icons, thanks! Just one comment.

nicolethoen:
> I believe the css only supports ComposableTable currently. This is also the case for nested column headers that I've added as part of this PR which rely on some properties added here.

@kmcfaul My instinct is that this should not be true - the CSS is the same for the composable and legacy tables. It is possible that implementing this in the Legacy table may be very difficult, especially without breaking anything else. It is the next step though to investigate that, @dlabrecq.

kmcfaul:
Updated the default sticky column width.

@nicolethoen Ah gotcha, that would make sense. I just vaguely remembered something about only Composable regarding one or both features, but perhaps I'm misremembering and it's instead Composable-first.

kmcfaul:
@dlabrecq Opened #6575 to track sticky columns for legacy table

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
```diff
@@ -2010,3 +2011,1106 @@ class DraggableTable extends React.Component {
...
+        // string sort
+        if (direction === 'asc') {
+          return a[index].localeCompare(b[index]);
+        }
+        return b[index].localeCompare(a[index]);
+      }
+    });
+    setRows(updatedRows);
+  };
+  console.log(colSpans);
```
nicolethoen, line 2831: I think you can remove that

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
```diff
@@ -2010,3 +2011,1106 @@ class DraggableTable extends React.Component {
...
+import {
+  TableComposable,
+  Thead,
+  Tbody,
+  Tr,
+  Th,
+  Td,
+  Caption,
+  TableText,
+  InnerScrollContainer
```
nicolethoen, line 2030: can we add `InnerScrollContainer` to the props tables at the bottom?

nicolethoen (changes requested):
Few comments

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
kmcfaul: Whoops

packages/react-table/src/components/TableComposable/examples/ComposableTable.md
kmcfaul: Added `OuterScrollContainer` as well

mcarrano:
@kmcfaul @nicolethoen @dlabrecq I think we should investigate what it will take to add this feature to the legacy table, although it's true that we are trying to focus on the composable table as our preferred solution moving forward. I'm OK with accepting that this PR applied only to the composable table. Feel free to open a new issue if you think it makes sense to investigate adding this also to the legacy table and I will get that into the roadmap for a future release.

kmcfaul:
@mcarrano I've opened #6575 to track sticky columns for legacy

nicolethoen (approved):

tlabaj (approved):
Looks great!

PR #6542, opened 2021-11-02
fix(jumplinks): fixed to always set clicked item as active

### LINKED ISSUE #6156: JumpList will not highlight last item when selected

**Expected behavior**
When clicking the last item in the JumpList we scroll to the section and it is highlighted.

**Actual behavior**
When clicking the last item in the JumpList we scroll to the section but it's not highlighted.

1. Visit the [PatternFly documentation](https://www.patternfly.org/v4/components/jump-links)
2. Click the last item in the table of contents to the right.

It's a bug, it's not blocking but just a bit annoying for the user. 

Keycloak Admin console, original issue can be found there (https://github.com/keycloak/keycloak-admin-ui/issues/960).

evwilkin:
**What**: Closes #6156 

This PR updates the JumpLinks to always set the clicked JumpLink as active when clicked.
- calls `setActiveIndex` in each JumpLink's `onClick` to set that link as active.
- adds `isLinkClicked` boolean variable which disables the scroll listener from also calling `setActiveIndex` and overriding the clicked link's active state when the linked section scrolls into place.
- this fixes the issue of the last JumpLink never showing as active when the linked section doesn't reach the top of the page.

### DIFF
--- a/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
+++ b/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
@@ -100,6 +100,8 @@ export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
   const [scrollItems, setScrollItems] = React.useState(hasScrollSpy ? getScrollItems(children, []) : []);
   const [activeIndex, setActiveIndex] = React.useState(activeIndexProp);
   const [isExpanded, setIsExpanded] = React.useState(isExpandedProp);
+  // Boolean to disable scroll listener from overriding active state of clicked jumplink
+  const isLinkClicked = React.useRef(false);
   // Allow expanding to be controlled for a niche use case
   React.useEffect(() => setIsExpanded(isExpandedProp), [isExpandedProp]);
   const navRef = React.useRef<HTMLElement>();
@@ -108,6 +110,10 @@ export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
 
   const scrollSpy = React.useCallback(() => {
     if (!canUseDOM || !hasScrollSpy || !(scrollableElement instanceof HTMLElement)) {
+      return;
+    }
+    if (isLinkClicked.current) {
+      isLinkClicked.current = false;
       return;
     }
     const scrollPosition = Math.ceil(scrollableElement.scrollTop + offset);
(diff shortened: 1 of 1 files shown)


mcarrano (approved):
Looks great @evwilkin !

nicolethoen (approved):

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
```diff
@@ -100,6 +100,8 @@ export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
   const [scrollItems, setScrollItems] = React.useState(hasScrollSpy ? getScrollItems(children, []) : []);
   const [activeIndex, setActiveIndex] = React.useState(activeIndexProp);
   const [isExpanded, setIsExpanded] = React.useState(isExpandedProp);
+  // Boolean to disable scroll listener from overriding active state of clicked jumplink
+  const isLinkClicked = React.useRef(false);
```
tlabaj, line 104: Curious why you are using a ref here instead of state for a boolean value.

kmcfaul: Also curious about this

kmcfaul (approved):

packages/react-core/src/components/JumpLinks/JumpLinks.tsx
evwilkin: @tlabaj @kmcfaul I did try using state with a boolean originally, but it wasn't working as expected - I'm not sure if either the `scrollSpy` was triggering before the state had finished updating, or if the state boolean was getting reset on rerender.

Either way, it does work as coded with `useRef` which [persists values across renders](https://reactjs.org/docs/hooks-reference.html#useref) or if I pull the boolean out of the component completely as a standalone variable, which I'm happy to switch to if that's preferred.

kmcfaul: After talking it through with Evan - this is a recommended pattern from the react docs https://reactjs.org/docs/hooks-reference.html#useref to store mutable instance fields (similar to how a private variable exists in a class). 

It can also work as a boolean stored outside the component. I'm good with either implementation

PR #6544, opened 2021-11-03
feat(Table): set default sort direction

### LINKED ISSUE #5329: Click table column header always sorts by ascending first

If you click the table header of a column that isn’t currently sorting the table, the first sort order is ascending, then the second click it’s descending. There doesn’t appear to be a way to specify the sort order for the first click to be descending.

In the context of a table view with a date column, if someone wants to change the sort order to be date created or modified, it’s going to show oldest first (i.e. ascending). However, in most cases, the user is likely more interested in seeing the newer items first (i.e. descending).

No steps to reproduce. The expected behavior doesn't appear to be supported.

enhancement

App Services, nice to have for end of March

cc @mcarrano @christiemolloy

kmcfaul:
**What**: Closes #5329

For both `Table` and `TableComposable` - the default sort direction can now be changed by setting `sortBy.defaultDirection` and will continue to default to ascending order.

Edit: updated to use a separate prop because using `sortBy.direction` meant that the starting sort order would change when toggling sort between different columns.

### DIFF
--- a/packages/react-table/src/components/Table/TableTypes.ts
+++ b/packages/react-table/src/components/Table/TableTypes.ts
@@ -139,6 +139,7 @@ export type IFormatterValueType = formatterValueType & {
 export interface ISortBy {
   index?: number;
   direction?: 'asc' | 'desc';
+  defaultDirection?: 'asc' | 'desc';
 }
 
 export interface IAction extends Omit<DropdownItemProps, 'title' | 'onClick'> {
--- a/packages/react-table/src/components/Table/TableTypes.tsx
+++ b/packages/react-table/src/components/Table/TableTypes.tsx
@@ -137,8 +137,11 @@ export type IFormatterValueType = formatterValueType & {
 };
 
 export interface ISortBy {
+  /** Index of the current sorted column */
   index?: number;
+  /** Current sort direction */
   direction?: 'asc' | 'desc';
+  /** Defaulting sorting direction. Defaults to "asc". */
   defaultDirection?: 'asc' | 'desc';
 }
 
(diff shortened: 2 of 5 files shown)


wise-king-sullyman (commented):
Is there any good way to explain how to take advantage of this functionality? As someone not familiar with Table it would've taken me a long time to figure out how to enable a default direction if I didn't have the changes in this PR to go off of.

I understand from looking at the Table docs that there's already....a lot. There may not be a good place/way to document it because of that, and it seems like there's a lot of quasi-documented stuff in Table already anyway so it might not be needed. If nothing else I feel like it would be beneficial to update the `sortBy` description.

nicolethoen:
@wise-king-sullyman That is always super helpful feedback!
I will point out, since it's not clear from the PR, that there are examples in the docs that [demonstrate sorting](https://patternfly-react-pr-6544.surge.sh/components/table/react-legacy#sortable--wrapping-column-headers) and do explicitly call out how to set a default sort behavior.

But @kmcfaul, I think that we should take the time to somehow define the `ISortBy` type listed in the `ThSortType` [props table](https://patternfly-react-pr-6544.surge.sh/components/table#thsorttype)

kmcfaul:
Added `ISortBy` to the docs as well as to the sorting demos to showcase how to set it (but uses the default value so the behavior isn't changing). I had to change `TableTypes.ts` to `TableTypes.tsx` to get the docs to pick up that file for `ISortBy`, will that be an issue?

nicolethoen:
@kmcfaul I dont think that's an issue.
You need to also add ISortBy to the legacy table docs, too.

And since you're going to expose the ISortBy - I wonder if it's worth exposing all the types in that file that are referenced in the Legacy Table docs. i.e.:
Screen Shot 2021-11-04 at 2 46 08 PM
a86b31491.png)

kmcfaul:
@nicolethoen Added `iSortBy` to the legacy docs. `IActions` and `IActionsResolver` don't appear to want to be added to docs, I think because they are defined as types instead of interfaces?

nicolethoen (approved):

wise-king-sullyman (approved):

PR #6545, opened 2021-11-03
feat(CalendarMonth): Added DateTimePicker demo

### LINKED ISSUE #5858: Spike: DateTimePicker Component
## Description

A component that allows users to select a date and time, combining functionality of the DatePicker and TimePicker.

Screenshot from 2021-05-27 13-55-20

### Use Cases

* Select a time within working hours on work days.
* Select a time within the next 12 hours possibly overlapping midnight to the next day.

## Why this component is needed

While it is possible to stitch the DatePicker and TimePicker together, it puts a lot of extra complexity on the application developer to manage changes in both components when values are validated and processed together. For example, if a value is out of range it is not that the date is invalid or the time is invalid but the combined date/time is invalid. It would also be good to have a single text box where the date + time value could be copied or pasted from the box rather than two separate text inputs.

## Specific Applications

Red Hat GPTE Service Catalog

Our specific use case is setting runtime and lifetime for environments where a user should be able to select a time within a bounded time range (ex: within the next 72 hours). We have our own work-around using a dropdown we can use for now.

gabipodolnikova:
**What**: Closes #5858

Adding new demo DateTimePicker using CalendarMonth.
https://patternfly-react-pr-6545.surge.sh/demos/datetimepicker
Just a PoC, totally up for any debate, it is very simple for now.

### DIFF
--- a/packages/react-core/src/demos/DateTimePicker.md
+++ b/packages/react-core/src/demos/DateTimePicker.md
@@ -5,8 +5,7 @@ beta: true
 ---
 
 import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/outlined-calendar-alt-icon';
-import OutlinedClockIcon from '@patternfly/react-icons/dist/esm/icons/outlined-clock-icon'; 
-import styles from '@patternfly/react-styles/css/components/DatePicker/date-picker';
+import OutlinedClockIcon from '@patternfly/react-icons/dist/esm/icons/outlined-clock-icon';
 
 ## Demos
 
@@ -18,7 +17,6 @@ The input is read only meaning that the following example does not cover pasting
 ```js
 import React from 'react';
 import { Dropdown, DropdownToggle, DropdownItem, CalendarMonth, InputGroup, TextInput, Button, Popover } from '@patternfly/react-core';
-import styles from '@patternfly/react-styles/css/components/DatePicker/date-picker';
 import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/outlined-calendar-alt-icon';
 import OutlinedClockIcon from '@patternfly/react-icons/dist/esm/icons/outlined-clock-icon';
 
@@ -71,7 +69,7 @@ DateTimePicker = () => {
       <Dropdown
         onSelect={onSelectTime}
         toggle={
-          <DropdownToggle toggleIndicator={null} onToggle={onToggleTime}>
+          <DropdownToggle toggleIndicator={null} onToggle={onToggleTime} style={{ padding: '6px 16px' }}>
             <OutlinedClockIcon />
           </DropdownToggle>
         }
@@ -85,7 +83,7 @@ DateTimePicker = () => {
       </Button>
   );
   return (
-    <div className={styles.datePicker}>
+    <div style={{ width: '300px' }}>
         <Popover
             position="bottom"
             bodyContent={calendar}
--- a/packages/react-core/src/demos/DateTimePicker.md
+++ b/packages/react-core/src/demos/DateTimePicker.md
@@ -1,12 +1,12 @@
 ---
-id: DateTimePicker
+id: Date and time picker
 section: demos
+beta: true
 ---
 
 import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/outlined-calendar-alt-icon';
-import OutlinedClockIcon from '@patternfly/react-icons/dist/esm/icons/outlined-clock-icon';
+import OutlinedClockIcon from '@patternfly/react-icons/dist/esm/icons/outlined-clock-icon'; 
 import styles from '@patternfly/react-styles/css/components/DatePicker/date-picker';
-import * as _ from 'lodash';
 
 ## Demos
 
@@ -17,7 +17,6 @@ The input is read only meaning that the following example does not cover pasting
 
 ```js
 import React from 'react';
-import * as _ from 'lodash';
 import { Dropdown, DropdownToggle, DropdownItem, CalendarMonth, InputGroup, TextInput, Button, Popover } from '@patternfly/react-core';
 import styles from '@patternfly/react-styles/css/components/DatePicker/date-picker';
 import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/outlined-calendar-alt-icon';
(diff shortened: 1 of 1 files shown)


nicolethoen:
@gabipodolnikova I think this is great and was exactly what I was thinking! 

@mcarrano do you think this works for a POC? We could share it with the creator of the issue.
Gabi did point out this [older design](https://github.com/patternfly/patternfly-design/blob/6d4eee49c05eddf3037a18738b4829e352d90dee/pattern-library/forms-and-controls/date-and-time/design/design.md#datetimepicker). We could also build something more like that - maybe with a scrollable menu for the time picker rather than a wheel - but I don't hate having both date and time in the same popper?

Probably only worth iterating on if a product requests it.

mcarrano:
@gabipodolnikova @nicolethoen Great to see this working and yes, I feel like this is serves the purpose of having a PoC to spark more discussion. The design link you provided is to the PF3 design. So I think there is still a design question concerning what form this should actually take.

@mmenestr I recall that we were talking about a similar issue and that you had some design proposals for this. Can you share them here? While what's in the demo does work, I think I prefer a solution that integrates everything into one popover like the PF3 design.

mmenestr:
@mcarrano I didn't have a design proposal as much as just an example of what we could do (see here https://github.com/patternfly/patternfly-design/issues/1082).

@nlcwong curious to hear what you think of this

nlcwong:
@mmenestr this feels like a good first step to the ultimate demo we were looking for, which was to create a date/time range picker (think of two date/time pickers interacting with one another to create a start and end date/time).

nicolethoen:
@nlcwong there is also already this demo live on the website: https://www.patternfly.org/v4/components/date-picker/react-demos#date-and-time-range-picker

So I could imagine building a demo that sort of incorporates this POC and the existing date and time range picker

packages/react-core/src/demos/DateTimePicker.md
```diff
@@ -0,0 +1,107 @@
+---
+id: DateTimePicker
+section: demos
+---
```
nicolethoen, line 4: We should add `beta: true` to the header of this file

packages/react-core/src/demos/DateTimePicker.md
```diff
@@ -0,0 +1,107 @@
...
+id: DateTimePicker
```
nicolethoen, line 2: I think we should write the id as `Date and time picker`

packages/react-core/src/demos/DateTimePicker.md
```diff
@@ -0,0 +1,107 @@
...
+## Demos
+
+In this section, we demonstrate how to use a [CalendarMonth](/components/calendar-month) and [Dropdown](/components/dropdown) to compose a simple DateTimePicker.
+The input is read only meaning that the following example does not cover pasting a copied date and time. Also the time can be picked only from the predefined options and cannot be changed in the input.
+
+### Date and time picker
+
+```js
+import React from 'react';
+import * as _ from 'lodash';
```
nicolethoen, line 20: Let's avoid using a third party library unless we absolutely need to.

nicolethoen (changes requested):

packages/react-core/src/demos/DateTimePicker.md
```diff
@@ -0,0 +1,106 @@
...
+        dropdownItems={timeOptions}
+    />
+  );
+  const calendarButton = (
+      <Button variant="control" onClick={onToggleCalendar}>
+          <OutlinedCalendarAltIcon />
+      </Button>
+  );
+  return (
+    <div className={styles.datePicker}>
```
nicolethoen, line 88: Let's remove this use of `styles` and avoid that import. 
If you want to limit the width of the input, let's just use a `style={{width: '250px' }}` or something like that instead.
Generally, consumers are not aware of our style class name imports.

packages/react-core/src/demos/DateTimePicker.md
```diff
@@ -0,0 +1,106 @@
...
+  const time = (
+      <Dropdown
+        onSelect={onSelectTime}
+        toggle={
+          <DropdownToggle toggleIndicator={null} onToggle={onToggleTime}>
+            <OutlinedClockIcon />
+          </DropdownToggle>
+        }
+        isOpen={isTimeOpen}
+        dropdownItems={timeOptions}
```
nicolethoen, line 79: Could you add `style={{padding: '0 6px'}}` to the time dropdown so it's the same width as the calendar button?

nicolethoen (changes requested):

mcarrano (approved):
Looks good to me. Thanks @gabipodolnikova !

nicolethoen (approved):

packages/react-core/src/demos/DateTimePicker.md
```diff
@@ -0,0 +1,104 @@
...
+section: demos
+beta: true
+---
+
+import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/outlined-calendar-alt-icon';
+import OutlinedClockIcon from '@patternfly/react-icons/dist/esm/icons/outlined-clock-icon';
+
+## Demos
+
+In this section, we demonstrate how to use a [CalendarMonth](/components/calendar-month) and [Dropdown](/components/dropdown) to compose a simple DateTimePicker.
```
tlabaj, line 12: @abigaeljamie do you mind giving feedback on the copy here.

dlabrecq (approved):

packages/react-core/src/demos/DateTimePicker.md
abigaeljamie:
```suggestion
In this demo, learn how to use a calendar month component and a dropdown component to compose a simple date time picker. The input is read only, meaning that the following example does not cover pasting a copied date and time. Also, you can pick the time only from the predefined options and cannot change it in the input.
``` 

I tweaked it a little bit to follow some consistency standards (writing component names in lowercase) and to use the active voice to make it a bit easier for readers to follow. I also edited the kick-off a tad from "In this section, we demonstrate how..." to "In this demo, learn how to..."  

After reading the other demos on PatternFly, I'm identifying areas we can edit the content so that each demo follows the same structure for consistency. @mcarrano , I'm thinking this is something Allison and Katie can work on when they start if you think it would be a good idea to get some content editing going for our demos so that they all have a unified structure and voice 🙂

tlabaj (approved):
LGTM!

packages/react-core/src/demos/DateTimePicker.md
mcarrano: That would be great @abigaeljamie . I suggest opening a new issue and attaching this feedback. @nicolethoen @tlabaj  can you advise on the best way to accomplish these updates as I know this test all lives in the React and/or Core repos.

PR #6547, opened 2021-11-03
fix(CodeEditor): add resizeobserver to code editor

### LINKED ISSUE #6227: CodeEditor does not respond well to viewport resizes?
Even if i set `width:100%` (which i think is the default anyway), the CodeEditor component does not respond to viewport resizes. When I used monaco-editor directly (in the past), I had to install a `ResizeObserver` to instruct monaco about resizing.

You may also want to allow for users of your component to pass in a resize observer of their own, because they may have subdivided the viewport in other ways.

nicolethoen:
**What**: Closes #6227

This also fixes a type discrepancy in the resizeObserver and chart-resize functions
@dlabrecq, I'd love if you'd make sure I didn't create more problems by fixing the issue I found.

### DIFF
--- a/packages/react-charts/src/components/ChartUtils/chart-resize.tsx
+++ b/packages/react-charts/src/components/ChartUtils/chart-resize.tsx
@@ -1,4 +1,3 @@
-import * as React from 'react';
 import { canUseDOM } from './chart-helpers';
 
 /**
@@ -11,7 +10,7 @@ import { canUseDOM } from './chart-helpers';
  * private observer: any = () => {};
  *
  * public componentDidMount() {
- *   this.observer = getResizeObserver(this.containerRef.current, this.handleResize);
+ *   this.observer = getResizeObserver(this.containerRef, this.handleResize);
  * }
  *
  * public componentWillUnmount() {
@@ -38,7 +37,7 @@ import { canUseDOM } from './chart-helpers';
  * private observer: any = () => {};
  *
  * public componentDidMount() {
- *   this.observer = getResizeObserver(this.inputRef.current, this.handleResize);
+ *   this.observer = getResizeObserver(this.inputRef, this.handleResize);
  * }
  *
  * public componentWillUnmount() {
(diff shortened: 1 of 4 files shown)


packages/react-charts/src/components/ChartUtils/chart-resize.tsx
```diff
@@ -57,17 +56,17 @@ import { canUseDOM } from './chart-helpers';
  *   );
  * }
  *
- * @param {React.RefObject} containerRef The container reference to observe
+ * @param {Element} containerRefElement The container reference to observe
  * @param {Function} handleResize The function to call for resize events
  * @return {Function} The function used to unobserve resize events
  */
-export const getResizeObserver = (containerRef: React.RefObject<any>, handleResize: () => void) => {
+export const getResizeObserver = (containerRefElement: Element, handleResize: () => void) => {
```
dlabrecq, line 63: I don't believe `Element` is the correct type (too restrictive) as this can be any `RefObject`. I see that the code editor is using `React.createRef<HTMLDivElement>();`, which is what the chart examples use. 

What was the type issue?

dlabrecq (changes requested):

dlabrecq (approved):
After some digging, I believe the `Element` type is ok. We're using `HTMLDivElement` with the `getResizeObserver` in Cost Management.

packages/react-charts/src/components/ChartUtils/chart-resize.tsx
```diff
@@ -38,7 +37,7 @@ import { canUseDOM } from './chart-helpers';
  * private observer: any = () => {};
  *
  * public componentDidMount() {
- *   this.observer = getResizeObserver(this.inputRef.current, this.handleResize);
+ *   this.observer = getResizeObserver(this.inputRef, this.handleResize);
```
dlabrecq, line 40: In the examples, we're using `this.containerRef.current`

this.observer = getResizeObserver(this.containerRef.current, this.handleResize);

dlabrecq (changes requested):

packages/react-charts/src/components/ChartUtils/chart-resize.tsx
dlabrecq: The code editor also uses 
`this.observer = getResizeObserver(this.ref.current, this.handleResize);`

packages/react-core/src/helpers/resizeObserver.tsx
```diff
@@ -38,7 +37,7 @@ import { canUseDOM } from './util';
  * private observer: any = () => {};
  *
  * public componentDidMount() {
- *   this.observer = getResizeObserver(this.inputRef.current, this.handleResize);
+ *   this.observer = getResizeObserver(this.inputRef, this.handleResize);
```
dlabrecq, line 40: The example should use `this.inputRef.current`

packages/react-charts/src/components/ChartUtils/chart-resize.tsx
nicolethoen: oh - oops - i was getting turned around. At one point i tried to keep the getResizeObverver types the same, and these changes would be appropriate, but instead I changed the param types, so I should have reverted these comments.

dlabrecq (approved):

tlabaj (approved):

Issue #6508, opened 2021-10-26
Popover should support changing the title's heading level


mcoker:
The title is always an `h6` - a user should be able to specify their own heading level that is appropriate

https://github.com/patternfly/patternfly-react/blob/cfe83416c03d0cc14d89e3ef5e0d90ed922f798e/packages/react-core/src/components/Popover/PopoverHeader.tsx#L9-L11

PR #6549, opened 2021-11-03
fix(popover): allow user to change heading level

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/6508


nicolethoen (approved):
Looks good! And @wise-king-sullyman this is a great basic example of setting up a context

tlabaj (approved):

PR #6552, opened 2021-11-05
WIP: Updated patternfly eslint package.

dlabaj:
Updated patternfly eslint package adding rules from the openshift console.  Please do not merge this as it's a WIP and for review only at the moment.

### DIFF
--- a/packages/eslint-plugin-patternfly-react/lib/config/recommended.js
+++ b/packages/eslint-plugin-patternfly-react/lib/config/recommended.js
@@ -20,6 +20,20 @@ module.exports = {
         devDependencies: true
       }
     ],
+    'promise/catch-or-return': 'error',
+    'promise/no-return-wrap': 'off',
+    'promise/param-names': 'off',
+    'promise/always-return': 'off',
+    'promise/no-native': 'off',
+    'promise/no-nesting': 'error',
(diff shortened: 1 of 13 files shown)


spadgett:
cc @christianvogt

andrewballantyne:
In one of our projects, we based off of `@redhat-cloud-services/eslint-config-redhat-cloud-services` ([npm.io description](https://npm.io/package/@redhat-cloud-services/eslint-config-redhat-cloud-services) really explains what it does best)... it seems there are a few of these linter bases. Have you had a chance to compare what differences would be in these two? I think Patternfly likely will be everyone's base moving forward assuming it covers the wide strokes, I'm just wondering if we are all trying to do the same thing maybe PF can take the best of all worlds.

christianvogt:
I'd like to see PF take ownership of the linting and expose multiple configs like [openshift/console eslint](https://github.com/openshift/console/blob/master/frontend/packages/eslint-plugin-console/index.js) so that it's useful to multiple projects.

christianvogt:
@dgutride Is there a timeframe for which a patternfly common eslint package is planned for?

We have several new repositories that are currently constructing their own eslint setup again and it would be good to get alignment sooner than later.

PR #6556, opened 2021-11-08
feat(Pagination): Added indeterminate count

### LINKED ISSUE #6497: Pagination: Add option for indeterminate count
Want to allow substitution of the word "Many" for the total number of items when fetching the total count is not possible. Core PR: https://github.com/patternfly/patternfly/pull/4428

gabipodolnikova:
**What**: Closes #6497

**Additional issues**:
I made the itemCount optional, so I hope I didn't break something.


mcarrano (approved):
Looks good. Thanks @gabipodolnikova !

kmcfaul (commented):
lgtm. Making a required prop into an optional should be fine, its the other direction that would be breaking.

Minor change I'd suggest is to remove `(item count is not known)` from the `.md` example title and instead add a sentence below the title describing that by not passing `itemCount` and passing `toggleTemplate` you can customize the toggle with text.

kmcfaul (approved):

nicolethoen (approved):

srambach:
This looks nice. I do notice one problem though - when you make the screen narrow, the text changes to "1-20 of 0" but I think it should say "1-20 of many".

packages/react-core/src/components/Pagination/PaginationOptionsMenu.tsx
```diff
@@ -66,7 +66,6 @@ export class PaginationOptionsMenu extends React.Component<PaginationOptionsMenu
     firstIndex: 0,
     lastIndex: 0,
     defaultToFullPage: false,
-    itemCount: 0,
     itemsTitle: 'items',
     toggleTemplate: ({ firstIndex, lastIndex, itemCount, itemsTitle }: ToggleTemplateProps) => (
       <React.Fragment>
```
nicolethoen, line 71: I think that you may need to update Pagination.tsx so that the `paginationTotalItems` is also consuming the user passed `toggleTemplate`. That is why the 0 is still in the smaller viewports

nicolethoen (changes requested):

srambach:
Looks good now!
 bitmojibitmoji

packages/react-core/src/components/Pagination/Pagination.tsx
```diff
@@ -201,8 +202,9 @@ export class Pagination extends React.Component<PaginationProps, { ouiaStateId:
   };
 
   getLastPage() {
-    const { itemCount, perPage } = this.props;
-    return Math.ceil(itemCount / perPage) || 0;
+    const { itemCount, perPage, page } = this.props;
+    // when itemCount is not known let's set lastPage as page+1 as we don't know the total count
+    return itemCount ? Math.ceil(itemCount / perPage) || 0 : page + 1;
```
nicolethoen, line 207: `itemCount` needs to be updated to `(itemCount || itemCount === 0)` for the case of 0 items, rather than unknown items.

packages/react-core/src/components/Pagination/Navigation.tsx
```diff
@@ -201,16 +204,20 @@ export class Navigation extends React.Component<NavigationProps, NavigationState
...
               min={lastPage <= 0 && firstPage <= 0 ? 0 : 1}
               max={lastPage}
               value={userInputPage}
               onKeyDown={event => this.onKeyDown(event, page, lastPage, onPageInput)}
               onChange={event => this.onChange(event, lastPage)}
             />
-            <span aria-hidden="true">
-              {ofWord} {pagesTitle ? pluralize(lastPage, pagesTitle) : lastPage}
-            </span>
+            {itemCount && (
```
nicolethoen, line 216: `itemCount` needs to be updated to `(itemCount || itemCount === 0)` for the case of 0 items, rather than unknown items.

nicolethoen (changes requested):

nicolethoen:
Noted above since I noticed the `no items` examples were broken

packages/react-core/src/components/Pagination/Navigation.tsx
gabipodolnikova: Oh good catch! :)

nicolethoen (approved):

tlabaj (approved):
Looks great!

PR #6562, opened 2021-11-08
Page nav flyout bug

### LINKED ISSUE #6396: bug(Nav): Cant see flyout nav when using page component
When using the `Nav` with flyout option inside `PageSidebar`, the flyout elements aren't shown on hover, they appear in the dom but are hidden

https://codesandbox.io/s/summer-star-f9h1e?file=/index.js
Screenshot from 2021-10-04 12-56-33

Product - foreman
pf version: "@patternfly/react-core": "4.155.0",

wise-king-sullyman:
**What**: Closes #6396 

**Additional issues**: N/A

This issue didn't call for a new demo, but making one simplified testing/debugging and cleaning up that test demo into something presentable didn't take long so I've included it. If you'd like me to remove the demo entirely I'm happy to do so.


nicolethoen:
I think the demo is great!
Are we planning to make the flyouts darker to match the rest of the nav?

wise-king-sullyman:
@nicolethoen from looking into it so far making the nav flyouts dark again will require a core change, as the current dark background style is applied when a flyout is nested within a nav item, which they no longer are because we now use popper. I'm honestly not sure how we would want to approach that.

kmcfaul:
@mattnolting @mcoker

We may need a new css class and flag added to `Menu` to get the dark nav-like styling since the nav flyouts are now contained in Popper outside the nav itself.

mcarrano:
+1 to having the new demo. It's very helpful so let's keep it in. And yes, we need to fix the CSS issue before we can release this. The other thing I noticed besides the flyouts being light is that when the mouse leaves the primary nav item to hover on the flyout, the primary item losses it's hover styling which is different from before.

Other than the styling issues above, the behavior looks good.

cc @mcoker

mcoker:
@wise-king-sullyman @kmcfaul @mcarrano I've opened a core issue here to move the styling out of the nav component and into the menu component under a modifier - https://github.com/patternfly/patternfly/issues/4510

kmcfaul (approved):

nicolethoen (approved):

mcoker (approved):
LGTM!!

PR #6563, opened 2021-11-09
Added requested text input group props

### LINKED ISSUE #6524: Text Input group
Add `onBlur` and `onFocus` callbacks.  Also add `type` prop to match TextInput component and the `isDisabled` prop.

wise-king-sullyman:
**What**: Closes #6524 

**Additional issues**: N/A

### DIFF
--- a/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
+++ b/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
@@ -37,11 +37,11 @@ export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps
   children,
   className,
   icon,
-  type,
-  onChange,
+  type = 'text',
+  onChange = (): any => undefined,
   onFocus,
   onBlur,
-  'aria-label': ariaLabel,
+  'aria-label': ariaLabel = 'Type to filter',
   value: inputValue,
   ...props
 }: TextInputGroupMainProps) => {
@@ -71,5 +71,4 @@ export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps
   );
 };
 
-TextInputGroupMain.defaultProps = { 'aria-label': 'Type to filter', type: 'text', onChange: (): any => undefined };
 TextInputGroupMain.displayName = 'TextInputGroupMain';
--- a/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
+++ b/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/TextInputGroup/text-input-group';
 import { css } from '@patternfly/react-styles';
+import { TextInputGroupContext } from './TextInputGroup';
 
 export interface TextInputGroupMainProps extends React.HTMLProps<HTMLDivElement> {
   /** Content rendered inside the text input group main div */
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
```diff
@@ -9,8 +10,25 @@ export interface TextInputGroupMainProps extends React.HTMLProps<HTMLDivElement>
...
+    | 'number'
+    | 'password'
+    | 'search'
+    | 'tel'
+    | 'time'
+    | 'url';
   /** Callback for when there is a change in the input field*/
   onChange?: () => void;
+  /** Callback for when the input field is focused*/
+  onFocus?: () => void;
```
nicolethoen, line 29: On the TextInput component, we add the event as a parameter for the event handlers so the consumer can use it if the want. I think that might be wise here for onChange, onFocus, and onBlur. In fact, onChange might benefit from also passing back the new value of the input via a parameter like the TextInput component does.

packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
```diff
@@ -19,25 +37,35 @@ export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps
   children,
   className,
   icon,
+  type,
   onChange,
+  onFocus,
+  onBlur,
   'aria-label': ariaLabel,
```
nicolethoen, line 44: Do you want to set a default value for this aria label? I saw you deleted it a few lines down, but you can set it inline here.

wise-king-sullyman: The default aria label value is just moved down to line 70 along with the default type value, if you'd rather me do the defaults inline here though I'm happy to do so.

packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
wise-king-sullyman: I can update the types to all include events, but adding the input value as a parameter itself will require adding a handler to TextInputGroupMain as was done in TextInput because otherwise typescript will complain about `onChange` not being assignable to the `onChange` on `input`.

I'm happy to have onChange just take `event` as a parameter, or add `event` and the `value` as paramaters and add a handler if you think that's best. I would lean towards just having `event` and letting the consumer get the value out of it if they choose though.

packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
nicolethoen: oh - i hadn't noticed that! @tlabaj do you prefer one way or the other?

packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
nicolethoen: I think keeping the types consistent with `TextInput` actually does help with upgrading and migrating to the TextInputGroup for products in the future.

`TextInput` handles the issue you mention by defining [this function](https://github.com/patternfly/patternfly-react/blob/320fde1581d401ba386289ac2b6893d567a079b8/packages/react-core/src/components/TextInput/TextInput.tsx#L106)

kmcfaul (approved):
lgtm

nicolethoen (approved):

packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
tlabaj: We don't do this anywhere else. Neither is wrong,.  For reviewing purposes, it maybe more intuitive to go with what we have elsewhere.

PR #6565, opened 2021-11-09
feat(card): added hoverable/selectable/selected-raised variant

### LINKED ISSUE #6465: Add selected-raised Card variant
Add new "selected-raised" Card variant. 

Core PR: https://github.com/patternfly/patternfly/pull/4425

This should be presented as the new/preferred variant for selectable cards. Make existing selectable cards as Legacy

wise-king-sullyman:
**What**: Closes #6465 

**Additional issues**: N/A


mcarrano (commented):
@wise-king-sullyman this looks good. @mceledonia can you take a look at the animation. Seems a little slow, but I'm OK with it if you are.

Also, did we have this as part of a demo anywhere? I thought so, but now I'm not sure.

kmcfaul (changes requested):
Just seeing a minor visual issue with mouse-only interaction. If you click the card to select it, and then click the card again to deselect it, the "hover" visual styling remains applied because the card is still focused (and the same styling is applied to ::focus). However, the visual border indicating focus isn't applied so it looks like you're hovering on the card permanently until you click something else or move focus away with the keyboard.

I'm not sure the best way to handle this (if at all) because the interaction is correct.

packages/react-core/src/components/Card/examples/Card.md
```diff
@@ -828,3 +828,116 @@ class ExpandableIconCard extends React.Component {
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      selected: null
+    };
+    this.onKeyDown = event => {
+      if (event.target !== event.currentTarget) {
+        return;
+      }
+      if ([13, 32].includes(event.keyCode)) {
```
kmcfaul, line 859: Can you also add `event.preventDefault();` for this block? `Space` will cause a page down while also selecting the card currently.

nicolethoen:
> the "hover" visual styling remains applied because the card is still focused

This is a common problem in a few different components. Such as the Table examples for selectable rows and selectable DataList examples. I am not sure if we'd need a JS solution to fix it - and if so, i'm not sure what it'd be.

I feel like it'd have to involve setting some sort of onmousemove handler - which sounds awful...

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -82,9 +91,12 @@ export const Card: React.FunctionComponent<CardProps> = ({
         className={css(
           styles.card,
           isHoverable && styles.modifiers.hoverable,
+          isHoverableRaised && styles.modifiers.hoverableRaised,
           isCompact && styles.modifiers.compact,
           isSelectable && styles.modifiers.selectable,
```
nicolethoen, line 96: I think it could be true that if someone sets `isSelectedableRaised` it sets `isSelectable` as well. So people dont NEED both props and since they are definitely related?

mcarrano:
@nicolethoen @kmcfaul Yeah, the mouse interaction issue is a tough one that effects multiple components like this since the focus and hover styles are essentially the same. I'm OK to live with this for now unless and until we come up with at better solution. Would defining a separate focus style help @mcoker ? Not sure if there is an elegant way around this.

packages/react-core/src/components/Card/Card.tsx
wise-king-sullyman: @nicolethoen I'm assuming you mean that if `isSelectedRaised` is set to true `isSelectableRaised` should also be set to true automatically?

If so, I would think that if a consumer is going to set a card as `isSelectedRaised` they would also want to set `isSelectableRaised` separately so I'm not *sure* if explicitly tying them together would improve usability, I could definitely be misunderstanding/seeing things wrong though.

I could definitely see the `isSelectedRaised && is SelectableRaised` check to apply the `selected-raised` as being unneeded though, I was just copying the `isSelected` and `isSelectable` usage.

nicolethoen: I actually meant that if `isSelectedRaised` is true then `isSelectable` would definitely need to be true, right?

nicolethoen (approved):

mcoker:
@mcarrano we can fix this using `:focus-visible`, though it isn't supported on Safari currently (though it's being worked on - https://bugs.webkit.org/show_bug.cgi?id=185859). That means Safari will get the current behavior.

We could also remove the focus styles since there is a focus outline applied in all of the browsers.

mcoker (approved):
🥳

mcarrano:
> @mcarrano we can fix this using :focus-visible (clicking won't indicate/retain focus, but keyboard navigation will), though it isn't supported on Safari currently (though it's being worked on - https://bugs.webkit.org/show_bug.cgi?id=185859). That means Safari will get the current behavior.

> We could also remove the focus styles since there is a focus outline applied in all of the browsers.

Let's see if we get more complaints about this. These are definitely two possibilities and I think we'd want to consider how to do this consistently across components.

dlabrecq (approved):

kmcfaul (approved):

tlabaj (approved):
LGTM

PR #6570, opened 2021-11-10
feat(panel): added panel component

### LINKED ISSUE #6462: Add new Panel component
Add a new Panel component to support composable component demos.

Follow up to https://github.com/patternfly/patternfly/issues/4378

wise-king-sullyman:
**What**: Closes #6462 

**Additional issues**: N/A


kmcfaul:
Looks good, though I'm a bit split on `PanelMain` and `PanelMainBody`. It's more flexible but I'm unsure if any content would go in `main` that couldn't go directly in `body`, so I'm wondering if we should wrap up both components into a single `PanelBody`. Anyone have thoughts? I'm happy either way.

nicolethoen:
@kmcfaul 
The core example demonstrates that the main body is what makes the panel appear raised (I think?) https://www.patternfly.org/v4/components/panel#no-bodyhttps://www.patternfly.org/v4/components/panel#no-body

Maybe that's something to either demo or explain explicitly in the react docs?

nicolethoen:
@kmcfaul It looks like the user can choose to forgo the main body and remove the default padding.
This may be helpful if the component they are putting inside also already has default padding.
https://www.patternfly.org/v4/components/panel#no-body

mcarrano (approved):
This looks great @wise-king-sullyman !

wise-king-sullyman:
@nicolethoen @kmcfaul if it would be beneficial to add an explanation for the Main/MainBody difference, or rework it in some way, I'm happy to do so. I can add a quick line to the "no body" example in no time.

packages/react-core/src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export interface PanelProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside Panel */
+  children?: React.ReactNode;
+  /** Class to add to outer div */
+  className?: string;
+  /** Flag to add raised styling to the panel */
+  isRaised?: boolean;
+  /** Flag to add bordered styling to the panel */
+  isBordered?: boolean;
```
mcoker, line 13: fwiw, I don't think you would ever have `isRaised && isBordered` - it would be one or the other. Assuming that's the case, would you keep these props as separate or have a prop that takes raised, bordered, etc as values?

packages/react-core/src/components/Panel/examples/Panel.md
```diff
@@ -0,0 +1,171 @@
...
+  );
+
+  return (
+    <Panel isScrollable>
+      <PanelHeader>Header content</PanelHeader>
+      <Divider />
+      <PanelMain>
+        <PanelMainBody>{Array(9).fill(mainContent)}</PanelMainBody>
+      </PanelMain>
+			<PanelFooter>Footer content</PanelFooter>
```
mcoker, line 167: nit - indentation is off

packages/react-core/src/components/Panel/examples/Panel.md
```diff
@@ -0,0 +1,171 @@
...
+```js
+import React from 'react';
+import { Panel, PanelMain, PanelMainBody } from '@patternfly/react-core';
+
+const ScrollablePanel = () => {
+  const mainContent = (
+    <>
+      Main Content
+      <br />
+      <br />
```
mcoker, line 131: nit - can you remove these extra `br`s at the end of the content? Makes it look like there is extra space at the end of the body

how it looks now:
Screen Shot 2021-11-10 at 2 37 42 PM

without the trialing br's
Screen Shot 2021-11-10 at 2 37 21 PM

mcoker (commented):
Looks great to me. Just a couple of nits but we can follow up later if needed.

Also you can set `--pf-c-panel__main--MaxHeight` with `isScrollable` and change the max height of the scrollable main element.

packages/react-core/src/components/Panel/examples/Panel.md
wise-king-sullyman: Odd, it shows properly in my editor 🤔

packages/react-core/src/components/Panel/Panel.tsx
wise-king-sullyman: I can definitely see the appeal of that approach (assuming raised and bordered will indeed never be used together), especially if it is anticipated that additional exclusive styles might be added in the future. I'm not totally sure that getting rid of one prop is worth adding the additional prop evaluation required for that approach though.

I'm happy to refactor to that approach if yall think it's the best way forward and we confirm raised and bordered will be exclusive.

nicolethoen (approved):

packages/react-core/src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Panel/panel';
+import { css } from '@patternfly/react-styles';
+
+export interface PanelProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside Panel */
```
tlabaj, line 6: can we make this sentences case.  (Lowercase for panel)

packages/react-core/src/components/Panel/Panel.tsx
tlabaj: I think combining the two would help prevent consumer error.  I don't know what a good name for the prop would be tough.

packages/react-core/src/components/Panel/PanelFooter.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Panel/panel';
+import { css } from '@patternfly/react-styles';
+
+export interface PanelFooterProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside PanelFooter */
```
tlabaj, line 6: Can we change this to 'Content rendered inside the panel footer`

packages/react-core/src/components/Panel/PanelHeader.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Panel/panel';
+import { css } from '@patternfly/react-styles';
+
+export interface PanelHeaderProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside PanelHeader */
```
tlabaj, line 6: Maybe `Content rendered inside the panel header`

packages/react-core/src/components/Panel/PanelMain.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import styles from '@patternfly/react-styles/css/components/Panel/panel';
+import { css } from '@patternfly/react-styles';
+
+export interface PanelMainProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside PanelMain */
```
tlabaj, line 6: Can we change to `Content rendered inside the panel main div`. Or something like that.

packages/react-core/src/components/Panel/PanelMainBody.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
```
tlabaj, line 1: @mcoker will we ever be able to have more than one `PanelMainBody` in the `PanelMain`?  Or will we ever want to put anything else inside `PanelMain`. if the answer is no to both questions, I don't know that is is necessary to expose this class to consumer.  We can just render the children on `PanelMain` in `PanelMainBody` internally.

tlabaj (changes requested):
Looks good!  A couple of suggestions.

packages/react-core/src/components/Panel/PanelMainBody.tsx
mcoker: @tlabaj you can put more than one `PanelMainBody` in `PanelMain`, though it wasn't designed for that. We use that in other places, like cards - https://www.patternfly.org/v4/components/card#with-multiple-body-sections

I'm not sure we will ever want to put anything else in `PanelMain`, but it's worth noting you should be able to use `PanelMain` on its own (without `PanelMainBody`) to get an empty panel without padding (the card also supports that)

packages/react-core/src/components/Panel/Panel.tsx
mcoker: @mcarrano can you confirm whether the raised and bordered styles would ever be used together? I'm assuming it's like a card where you use one or the other, but never together.

mcarrano: Yes, that's correct. I can't think of a reason to use these together.

dlabrecq (approved):

packages/react-core/src/components/Panel/examples/Panel.md
```diff
@@ -0,0 +1,181 @@
...
+  const mainContent = (
+    <>
+      Main content
+      <br />
+      <br />
+    </>
+  );
+
+  return (
+    <Panel isScrollable>
```
tlabaj, line 166: We need to add a tab index so that the scrollable content is accessible via keyboard.

cc @jessiehuff

nicolethoen (approved):

kmcfaul (approved):

packages/react-core/src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,37 @@
...
+import styles from '@patternfly/react-styles/css/components/Panel/panel';
+import { css } from '@patternfly/react-styles';
+
+export interface PanelProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the panel */
+  children?: React.ReactNode;
+  /** Class to add to outer div */
+  className?: string;
+  /** Styling to apply to the perimeter of the panel */
+  perimeterStyling?: 'raised' | 'bordered';
```
mcoker, line 11: There may be other variations that could be used as a value here that aren't necessarily tied to the border/box-shadow (perimeter styling). I wonder if we could use something more generic? `variation` is something we use in other components - could we use that, or is that too generic and/or we might want to reserve that for other types of variations we could introduce to the component?

packages/react-core/src/components/Panel/PanelMain.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const PanelMain: React.FunctionComponent<PanelMainProps> = ({
+  className,
+  children,
+  maxHeight,
+  ...props
+}: PanelMainProps) => {
+  const compiledProps = {
+    className: css(styles.panelMain, className),
+    ...(maxHeight && { style: { maxHeight, overflow: 'auto' } })
```
mcoker, line 22: Here's a better example than the select for how to use the CSS vars available to set this `maxHeight` value. Also you shouldn't need to specify `overflow` - the user should be using `isScrollable`, which will set `overflow` in the CSS.

https://github.com/patternfly/patternfly-react/blob/8c1eb5d895aa2512f7b4197654e414394ff25b30/packages/react-core/src/components/Menu/MenuContent.tsx#L32-L44

The var to set for the panel max height is `--pf-c-panel__main--MaxHeight`

mcoker (changes requested):

packages/react-core/src/components/Panel/Panel.tsx
wise-king-sullyman: @mcoker how does `variant` sound? I know that's what we use on Button for instance.

mcoker: wfm! But I'll defer to @nicolethoen, @tlabaj, @kmcfaul since they're more familiar with prop naming.

jessiehuff (approved):
Everything seems to work ok by keyboard and screen reader (VO), and I don't see any axe errors. It seems like this is meant to not be very opinionated so I imagine we wouldn't include too much a11y wise, but I'm not sure yet of all of the use cases. Perhaps it would just need documentation depending on the ways we expect it to be used. The current examples here look fine though because the panel itself looks like it's more for styling reasons and not something particular that needs to be communicated to a screen reader that they're in a panel.

tlabaj (approved):
Looks great

packages/react-core/src/components/Panel/PanelMain.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const PanelMain: React.FunctionComponent<PanelMainProps> = ({
+  className,
+  children,
+  maxHeight,
+  ...props
+}: PanelMainProps) => (
+  <div
+    className={css(styles.panelMain, className)}
+    style={{ '--pf-c-panel__main--MaxHeight': maxHeight } as React.CSSProperties}
```
mcoker, line 22: 🔥

mcoker (approved):
L🍕TM!!!!

PR #6576, opened 2021-11-10
feat(LogViewer): add wrapText prop

### LINKED ISSUE #6574: LogViewer: add wrapText props
Currently, the log viewer is set to default to wrap text. We should create a prop to let users choose to wrap text or not.

Enhancement.

DaoDaoNoCode:
**What**: Closes #6574.
Add a prop named `wrapText`, the default value is `true`. When it's set to `false`, the text will not be wrapped.
Also, clean the react-window source code to remove some unused variables and methods.


nicolethoen (approved):
LGTM

packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -29,6 +29,8 @@ interface LogViewerProps {
   scrollToRow?: number;
   /** Number of rows to display in the log viewer */
   itemCount?: number;
+  /** Flag indicating that log viewer is wrapping text or not */
+  wrapText?: boolean;
```
tlabaj, line 33: Typically in PF we start boolean prop names with "is" or "has'.

jschuler: If you really want to prefix it you probably also need to change the wording a bit to accommodate
- isTextWrapped
- hasWrappedText
- shouldWrapText

jschuler:
Screen Shot 2021-11-11 at 2 19 19 PM

Maybe only somewhat related to this PR, but I noticed that the search only really returns the number of rows that include a search term, not the actual number of search term results.

Due to this issue, when you do not wrap text, it will also not scroll to a search term that is hidden, if the same row includes the search term in the visible area

tlabaj (approved):

PR #6581, opened 2021-11-11
Add text input group demo

### LINKED ISSUE #6393: Demonstrate key-value filtering
Create a component/demo to demonstrate key-value filtering in a toolbar. 

Core issue to create text-input-group: https://github.com/patternfly/patternfly/pull/4408
Design issue: https://github.com/patternfly/patternfly-design/issues/1088

wise-king-sullyman:
**What**: Closes #6393

**Additional issues**: patternfly/patternfly-design#1088

Interaction proposal:
image

### DIFF
--- a/packages/react-core/src/demos/TextInputGroupDemo.md
+++ b/packages/react-core/src/demos/TextInputGroupDemo.md
@@ -132,28 +132,62 @@ export const KeyValueFiltering = () => {
     setMenuItemsText(data[selectedText]);
   };
 
-  /** enable key/value selection and selected key removal using keyboard events */
-  const handleKeydown = event => {
-    if (event.key === 'Enter') {
-      /** do nothing if the menu contains no real results */
-      if (menuItems.length === 1) {
-        return;
-      }
-
-      /** perform the appropriate action based on key selection state */
-      if (selectedKey.length) {
-        selectValue(menuItems[2].props.children);
-      } else {
-        selectKey(menuItems[2].props.children);
-      }
+  const handleEnter = () => {
+    /** do nothing if the menu contains no real results */
+    if (menuItems.length === 1) {
+      return;
     }
 
-    /** allow the user to backspace at the selected key name or hit escape to drop the selected key */
-    if (
-      event.key === 'Escape' ||
-      (event.key === 'Backspace' && selectedKey.length && inputValue.length === selectedKey.length + 2)
-    ) {
+    /** perform the appropriate action based on key selection state */
+    if (selectedKey.length) {
+      selectValue(menuItems[2].props.children);
+    } else {
+      selectKey(menuItems[2].props.children);
+    }
+  };
+
+  /** allow the user to backspace at the selected key name to drop the currently selected key */
+  const handleBackspace = () => {
+    if (selectedKey.length && inputValue === `${selectedKey}: `) {
       clearSelectedKey();
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


nicolethoen (changes requested):
This LOOKS great!!!

Some interaction issues, though.

I cannot select the key = Name. when I try to click it, it actually populates it with the key = Cluster. And if I type `Name: ` none of the values appear as options.

I think it'll be important to be sure that if I user wanted to interact with this, using only the keyboard - they should be able to. If they wanted to ignore the dropdown completely that should be an option. And if they wanted to use the keyboard to select a key or value, they should be able to.

mcarrano (commented):
This looks great @wise-king-sullyman !  Just a couple of questions/comments:

* The way that the chip gets cut off when it overflows feels odd. I'm talking about this:

Screen Shot 2021-11-11 at 2 25 37 PMBut unfortunately I don't have a better suggestion. Anyone have any ideas?

* I understand why this is a demo for Text input group. And while that's technically correct, I doubt anyone would look for it there. Would we move it or link to it with a more descriptive title? What do you think @nicolethoen ?

wise-king-sullyman:
@mcarrano an interaction question/suggestion was brought up by @nicolethoen in a conversation on slack:

> Should the user be able to create chips by hitting the enter key even if they are not a valid key-value pair?

This doesn't quite line up with my interpretation of the interaction proposal, but per @nicolethoen:

> I think this pattern will be used for running searches on tables - and people are going to want to use this both as an attr/value search and as your standard google search (on a dataset)

Is this functionality that you'd like to see added to this demo?

nicolethoen:
> * I understand why this is a demo for Text input group. And while that's technically correct, I doubt anyone would look for it there. Would we move it or link to it with a more descriptive title? What do you think @nicolethoen ?

@mcarrano I am always in favor of more explanation. 
@wise-king-sullyman It'd probably be super useful to place a sentence or two as a comment before any functions you define within the example (unless they are super straight forward).

Additionally, adding a few sentences before the demo to explain some of the interaction decisions you main or intended use cases you imagine this demo could satisfy would be helpful.

mcarrano:
@wise-king-sullyman @nicolethoen regarding the question below...

> Should the user be able to create chips by hitting the enter key even if they are not a valid key-value pair?

The intended use case for this that I was modeling from ACM constrained the input to predefined attribute-value pairs. So you would not be able to create a chip with an unknown value. That said. there is probably another use case that is more like an auto-complete search where the user is offered a list of suggestions but can also enter any value from the keyboard and hit Enter to create a chip. Would it be possible to create a second demo that models this? I see these as different use cases and it would demonstrate the flexibility of this approach.

mcoker (approved):
LGTM. Found a bug with the empty `<TextInputGroupUtilities>` element creating extra space that we can probably fix in core - https://github.com/patternfly/patternfly/issues/4514

wise-king-sullyman:
@mcarrano spinning up a second demo for that functionality is totally possible, though since code freeze was supposed to have happened already I'd like to do so in a new ticket next sprint if that's alright with you.

mcarrano:
That's fine @wise-king-sullyman . If you can open a new issue and tag me, I'll make sure it gets into the plan for the next milestone. Thanks.

nicolethoen (approved):
Awesome job!!

kmcfaul (approved):
lgtm!

PR #6583, opened 2021-11-11
chore: update react and react-dom types

jonkoops:
Updates the types for `react` and `react-dom` to version 17 to align with the actual versions being used. Ran into this as I needed to use some updated types for another PR and these older versions were getting in the way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
     "@octokit/rest": "^16.43.2",
     "@types/enzyme": "3.9.0",
     "@types/jest": "27.0.2",
+    "@types/react": "^17.0.0",
+    "@types/react-dom": "^17.0.0",
     "@typescript-eslint/eslint-plugin": "^4.4.1",
     "@typescript-eslint/parser": "^4.4.1",
     "@wojtekmaj/enzyme-adapter-react-17": "^0.3.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -35,6 +35,8 @@
     "@octokit/rest": "^16.43.2",
     "@types/enzyme": "3.9.0",
     "@types/jest": "27.0.2",
+    "@types/react": "^17.0.0",
```
jonkoops, line 38: I've added these types to the root package so that it is shared explicitly for all workspaces instead of implicitly trough some packages that happen to include it.

dlabrecq (approved):

nicolethoen:
@jonkoops Thanks so much for this contribution! Once we wrap up the current code freeze, I'm going to have a few products test with this updated version. There have already been issues created by products being behind our current types versions, we'd have to evaluate to what extent this could further exacerbate those problems (or not)

jonkoops:
Thanks for taking the time @nicolethoen. I think this should have little impact for end-users as the types are not shipped inside of PatternFly itself, the end-user is responsible for installing these types (if they want to of course).

It could be that some APIs that are being used in newer versions of React are not available (such as `isValidElement()`). However since the peer dependency for React (and React DOM) is `16.8` or `17` this should not be an issue.

nicolethoen (approved):
I spoke with @jcaianirh and I am good to merge this.
We will grab the pre-release version built with this change and test it in OpenShift to make sure it doesn't cause new breaking issues for them.

tlabaj (approved):
LGTM

PR #6592, opened 2021-11-12
fix(LogViewer): update search function

### LINKED ISSUE #6590: Update search function for log viewer
Based on the comment here https://github.com/patternfly/patternfly-react/pull/6576#issuecomment-966563766. We should better update the search function of the log viewer. Currently, if there are several search words on the same row, they will be treated as one result. Instead of this behavior, we should show a single result for every searched keyword, which makes more sense.

DaoDaoNoCode:
**What**: Closes #6590 

Only highlight the searched keyword, but not the keywords on the same row.

### DIFF
--- a/packages/react-log-viewer/src/LogViewer/LogViewer.tsx
+++ b/packages/react-log-viewer/src/LogViewer/LogViewer.tsx
@@ -195,7 +195,7 @@ const LogViewerBase: React.FunctionComponent<LogViewerProps> = memo(
       // if the keyword is out of the window when wrapping text
       if (!isTextWrapped) {
         setTimeout(() => {
-          const element = document.querySelector('.pf-c-log-viewer__string.pf-m-current');
+          const element = containerRef.current.querySelector('.pf-c-log-viewer__string.pf-m-current');
           element && element.scrollIntoView({ block: 'nearest', inline: 'center' });
         }, 1);
       }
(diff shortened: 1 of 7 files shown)


packages/react-log-viewer/src/LogViewer/LogViewer.tsx
```diff
@@ -188,12 +188,12 @@ const LogViewerBase: React.FunctionComponent<LogViewerProps> = memo(
...
+    const scrollToRowInFocus = (searchedRowIndex: searchedKeyWordType) => {
       setRowInFocus(searchedRowIndex);
-      logViewerRef.current.scrollToItem(searchedRowIndex, 'center');
+      logViewerRef.current.scrollToItem(searchedRowIndex.rowIndex, 'center');
       // use this method to scroll to the right
       // if the keyword is out of the window when wrapping text
-      if (!isWrapText) {
+      if (!isTextWrapped) {
         setTimeout(() => {
           const element = document.querySelector('.pf-c-log-viewer__string.pf-m-current');
```
jschuler, line 198: do you have a ref to the logviewer parent?
would be better to use querySelector on the parent ref. I am thinking of the case where there are 2 log viewers on the same page, if you search on the second one it will cause the window to jump up to the first one otherwise

jschuler (approved):
nice work @DaoDaoNoCode !

tlabaj (approved):
Thank you!

Issue #6367, opened 2021-09-28
Menu: add Plain variant


mcarrano:
Adds a `.pf-m-scrollable` variant to the menu that replaces previously having to set `--pf-c-menu__content--MaxHeight` var to make the menu scrollable. It gives the content height a default max-height. You can still set the max-height using that var if you want something other than the default.

Creates a `.pf-m-plain` variation that removes the outer box-shadow, for use in a regular container instead of a dropdown menu.

PR #6608, opened 2021-11-15
fix(Menu): allow maxMenuHeight with Scrollabe

### LINKED ISSUE #6607: Menu - should be able to use maxMenuHeight with isScrollable
https://github.com/patternfly/patternfly-react/blob/77f760305bd6eb6415fd27be973783cca4f0d97a/packages/react-core/src/components/Menu/MenuContent.tsx#L39

You should be able to use `maxMenuHeight` with `isScrollable`. That's probably the most common use case, I would guess. If the menu is ever scrollable, we want to get the class that comes with `isScrollable` so we can control the menu styling (like changing the footer's top border to a box shadow to communicate depth).

nicolethoen:
**What**: Closes #6607

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Menu/MenuContent.tsx
+++ b/packages/react-core/src/components/Menu/MenuContent.tsx
@@ -28,15 +28,15 @@ export const MenuContent = React.forwardRef((props: MenuContentProps, ref: React
   };
   return (
     <MenuContext.Consumer>
-      {({ menuId, onGetMenuHeight, isScrollable }) => (
+      {({ menuId, onGetMenuHeight }) => (
         <div
           {...rest}
           className={css(styles.menuContent, props.className)}
           ref={el => refCallback(el, menuId, onGetMenuHeight)}
           style={
             {
-              '--pf-c-menu__content--Height': menuHeight || 'auto',
-              '--pf-c-menu__content--MaxHeight': maxMenuHeight || (!isScrollable ? 'auto' : '')
+              ...(menuHeight && { '--pf-c-menu__content--Height': menuHeight }),
+              ...(maxMenuHeight && { '--pf-c-menu__content--MaxHeight': maxMenuHeight })
             } as React.CSSProperties
           }
         >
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/Menu/MenuContent.tsx
```diff
@@ -36,7 +36,7 @@ export const MenuContent = React.forwardRef((props: MenuContentProps, ref: React
           style={
             {
               '--pf-c-menu__content--Height': menuHeight || 'auto',
-              '--pf-c-menu__content--MaxHeight': !isScrollable ? maxMenuHeight || 'auto' : ''
+              '--pf-c-menu__content--MaxHeight': maxMenuHeight || (!isScrollable ? 'auto' : '')
```
mcoker, line 39: these vars are already defined as `auto` in the CSS - I would only set them if `menuHeight` and/or `maxMenuHeight` are defined. Can you think of a reason they always need to be set inline?

https://github.com/patternfly/patternfly/blob/a67b16945bec19cae3916d9688fe1e9c6cba2576/src/patternfly/components/Menu/menu.scss#L31-L32

mcoker (changes requested):

packages/react-core/src/components/Menu/MenuContent.tsx
nicolethoen: I can choose only to set them if the consumer passes a value - easy peasy

nicolethoen: I was not sure if there was a reason that originally it only sets 'auto' if it's not scrollable

mcoker: @nicolethoen I can't think of why it would be necessary, it's redundant with the component's CSS. My guess is it's a miscommunication from the original issue - https://github.com/patternfly/patternfly-react/issues/6367

> Adds a .pf-m-scrollable variant to the menu that replaces previously having to set --pf-c-menu__content--MaxHeight var to make the menu scrollable. It gives the content height a default max-height. You can still set the max-height using that var if you want something other than the default.

From the description, it kind of sounds like it's saying you would either use the scrollable class or set the var if you don't want the default that the scrollable class provides.

mcoker (approved):
LGTM!

tlabaj (approved):
LGTM

kmcfaul (approved):

PR #6610, opened 2021-11-15
fix(FileUpload): Downgrade react-dropzone back to 9.0.0 to avoid a breaking type change, fix behavior regressions

### LINKED ISSUE #6609: FileUpload: Downgrade react-dropzone back to 9.0.0 to avoid a breaking type change
react-dropzone (a dependency used by the FileUpload component) was upgraded from 9.0.0 to 10.2.2 in https://github.com/patternfly/patternfly-react/pull/6139. This upgrade causes some TypeScript types to change (in particular, the `event` argument of the `onChange` prop changed) which we have realized is a breaking change. This dependency needs to be downgraded so we can release the changes from #6139 in a non-breaking release.

mturley:
Closes #6609

In order to keep the changes from #6139 without any breaking TypeScript changes, this reverts react-dropzone to 9.0.0 and makes some tweaks so the new refs/events still work:
* The Browse button had broken because #6139 sets a `ref` on the hidden `<input>` in order to make its `onClearClick` handler work. This ref overrode the internal ref set by react-dropzone and used for the manual browse functionality. This PR sets both refs (using some tricky legacy callback ref stuff) so that both the Browse button and the `onClearClick` prop can work. The callback refs are not ideal, but can be reverted if we ever end up upgrading to react-dropzone 11+.
  * Note: the `getInputProps()` function used here does in fact return a `ref` function that we need for this, even though its TypeScript type says it does not. This is a bug fixed in newer react-dropzone versions, but I work around it here with an `as` assertion.
* The basic FileUpload example was broken in #6139: the user should not be blocked from typing in the textarea unless they had selected a file. I just needed to add an `onTextChange` prop to that example.

### DIFF
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -38,7 +38,7 @@
     "@patternfly/react-styles": "^4.25.2",
     "@patternfly/react-tokens": "^4.27.2",
     "focus-trap": "6.2.2",
-    "react-dropzone": "10.2.2",
+    "react-dropzone": "9.0.0",
     "tippy.js": "5.1.2",
     "tslib": "^2.0.0"
   },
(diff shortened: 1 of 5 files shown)


tlabaj (approved):
LGTM!

kmcfaul (approved):

PR #6623, opened 2021-11-18
Fix for no autocomplete for Typeahead Select

jeff-phillips-18:
**What**: 
Fixes https://github.com/patternfly/patternfly-react/issues/6625

Fix to not show the auto complete form for the TypeAhead Select component on Chrome

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -934,7 +934,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                 }
               }}
               ref={this.filterRef}
-              autoComplete="off"
+              autoComplete="never"
             />
           </div>
           <Divider key="inline-filter-divider" />
@@ -1128,7 +1128,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   type="text"
                   onClick={this.onClick}
                   onChange={this.onChange}
-                  autoComplete="off"
+                  autoComplete="never"
                   disabled={isDisabled}
                   ref={this.inputRef}
                 />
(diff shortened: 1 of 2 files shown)


kmcfaul (approved):

tlabaj (approved):

garrett:
This change specifically makes it so we _do_ experience browser autocomplete on top of PatternFly autocomplete. In other words, this change is a breaking one that actually does the opposite of what is intended.

There's no "never" attribute listed on any documentation or spec pages, but there is an "off", which is suggest to be used in cases where autocomplete is not wanted. (The exception is user/pass forms where password managers still work with "off", but this isn't for autocomplete in most cases. In that specific case, [it's suggested to use "new-password"](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields)... and this wouldn't be case for the typeahead select, _unless_ it's used for user selection. And it usually isn't used for user selection.)

- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
  > "off"
  >
  > The browser is not permitted to automatically enter or select a value for this field. It is possible that the document or application provides its own autocomplete feature, or that security concerns require that the field's value not be automatically entered.
- https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill
  > When wearing the autofill expectation mantle, the autocomplete attribute, if specified, must have a value that is an ordered set of space-separated tokens consisting of either a single token that is an ASCII case-insensitive match for the string "off", or a single token that is an ASCII case-insensitive match for the string "on", or autofill detail tokens.

garrett:
Note: It might have been the case that in testing, both user and automated, that it does seem to pass on the first time with "never", but then a second go would autocomplete a "never" type of autocomplete.

The correct solution would be to:
1. Revert this PR to unbreak the typeahead select in all general cases.
2. Add a workaround for the typeahead user case (or an option to switch), using either a hardcoded "new-password" or a custom value as the value instead of "off".

PR #6624, opened 2021-11-18
feat(Slider): add option to add Tooltip over thumb

### LINKED ISSUE #5955: Slider: can we provide a tooltip for the thumb like PF3 did?
Screen Shot 2021-04-23 at 11 52 21

KKoukiou:
Fixes #5955

### DIFF
--- a/packages/react-core/src/components/Slider/Slider.tsx
+++ b/packages/react-core/src/components/Slider/Slider.tsx
@@ -5,6 +5,7 @@ import { css } from '@patternfly/react-styles';
 import { SliderStep } from './SliderStep';
 import { InputGroup, InputGroupText } from '../InputGroup';
 import { TextInput } from '../TextInput';
+import { Tooltip } from '../Tooltip';
 
 export interface SliderStepObject {
   /** Value of the step. This value is a percentage of the slider where the  tick is drawn. */
@@ -44,6 +45,8 @@ export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onCh
   inputAriaLabel?: string;
   /* Aria label for the thumb */
   thumbAriaLabel?: string;
+  /* Adds a tooltip over the thumb containing the current value */
+  hasTooltipOverThumb?: boolean;
   /** Label that is place after the input field */
   inputLabel?: string | number;
   /** Position of the input */
@@ -73,6 +76,7 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
   inputLabel,
   inputAriaLabel = 'Slider value input',
   thumbAriaLabel = 'Value',
+  hasTooltipOverThumb = false,
   inputPosition = 'right',
   onChange,
   leftActions,
@@ -141,7 +145,9 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
         return matchingStep.label;
       }
     }
-    return localValue.toString();
+    // For continuous steps default to showing 2 decimals in tooltip
+    // Consider making it configurable via a property
+    return Number(Number(localValue).toFixed(2)).toString();
   };
 
   const handleThumbDragEnd = () => {
@@ -209,12 +215,12 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
 
     thumbRef.current.style.setProperty('--pf-c-slider--value', `${newPercentage}%`);
     // convert percentage to value
-    const newValue = (newPercentage * (max - min)) / 100 + min;
+    const newValue = Math.round(((newPercentage * (max - min)) / 100 + min) * 100) / 100;
     setValue(newValue);
 
     if (!customSteps) {
       // snap to new value if not custom steps
-      snapValue = Math.round((newValue - min) / step) * step + min;
+      snapValue = Math.round((Math.round((newValue - min) / step) * step + min) * 100) / 100;
       thumbRef.current.style.setProperty('--pf-c-slider--value', `${snapValue}%`);
       setValue(snapValue);
     }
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Slider/examples/Slider.md
```diff
@@ -134,13 +135,22 @@ class ContinuousInput extends React.Component {
     return (
       <>
         <Text component={TextVariants.h3}>Slider value is: {this.state.value.toFixed(2)}</Text>
```
nicolethoen, line 137: We should remove this line.

packages/react-core/src/components/Slider/examples/Slider.md
```diff
@@ -134,13 +135,22 @@ class ContinuousInput extends React.Component {
...
+          label="thumbHasTooltip"
+          isChecked={this.state.thumbHasTooltip}
+          onChange={thumbHasTooltip => this.setState({ thumbHasTooltip })} />
+        <Text component={TextVariants.h3}>Slider Value is: {this.state.value.toFixed(2)}</Text>
+        <Slider
+          thumbHasTooltip={this.state.thumbHasTooltip}
+          value={this.state.value}
+          onChange={this.onChange} />
         <br />
         <Text component={TextVariants.h3}>Slider value is: {this.state.valueCustom.toFixed(2)}</Text>
```
nicolethoen, line 148: As it is, the `Slide value is:` text in the example only has 2 values trailing the decimal, but the tooltip has many. I wonder if we should make the number of trailing values something the user can configure, and have it always default to 2? It'd make the aria label much more readable.

nicolethoen (changes requested):
Could you also add a unit test which basically verifies that a Tooltip gets rendered whenever `thumbHasTooltip` is true?

packages/react-core/src/components/Slider/examples/Slider.md
KKoukiou: Made it default to 2 decimals. Lets add the property for configuring once someone requests it if you 're fine with that.

KKoukiou:
> Could you also add a unit test which basically verifies that a Tooltip gets rendered whenever `thumbHasTooltip` is true?

Added

packages/react-core/src/components/Slider/examples/Slider.md
KKoukiou: Removed.

packages/react-core/src/components/Slider/examples/Slider.md
nicolethoen: Could you also remove the trimming in the examples if it's no longer needed?

KKoukiou: Done

mcarrano (changes requested):
This looks good, although while the tooltip in the example is constrained to 2 decimal places, the value displayed is not. Shouldn't these match? Here's what I'm seeing.

Screen Shot 2021-11-22 at 10 34 21 AM

Also, it seems like the Value input examples are broken. Maybe unrelated to your changes, but can you check these?

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -139,7 +143,9 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
         return matchingStep.label;
       }
     }
-    return localValue.toString();
+    // For continuous steps default to showing 2 decimals in tooltip
+    // Consider making it configurable via a property
+    return Number(localValue.toFixed(2)).toString();
```
nicolethoen, line 148: To piggy pack off of Matt's comment, I think we want to make this same update to the `onChangeHandler` function. Either that, or in the `handleThumbMove` function we should trim the continuous value on line 221 to use 2 trailing digits.

KKoukiou: Fixed.

KKoukiou:
> This looks good, although while the tooltip in the example is constrained to 2 decimal places, the value displayed is not. Shouldn't these match? Here's what I'm seeing.
> 
> Screen Shot 2021-11-22 at 10 34 21 AM
> 
> Also, it seems like the Value input examples are broken. Maybe unrelated to your changes, but can you check these?

This is also broken on master, just checked. Let's fix it in a separate PR as it's unrelated.

packages/react-core/src/components/Slider/examples/Slider.md
```diff
@@ -133,14 +134,23 @@ class ContinuousInput extends React.Component {
   render() {
     return (
       <>
-        <Text component={TextVariants.h3}>Slider value is: {this.state.value.toFixed(2)}</Text>
-        <Slider value={this.state.value} onChange={this.onChange} />
+        <Checkbox
+          id="thumb-has-tooltip"
```
gabipodolnikova, line 138: I would add a style={{ marginBottom: 20 }} to the Checkbox.

KKoukiou: Added.

gabipodolnikova (changes requested):
I also noticed that the Thumb value input example is broken, which it wasn't before. When I move it, I get: `TypeError: W.toFixed is not a function` Do you see it too?

KKoukiou:
> I also noticed that the Thumb value input example is broken, which it wasn't before. When I move it, I get: `TypeError: W.toFixed is not a function` Do you see it too?

Yes, sorry. I fixed it.

gabipodolnikova (approved):

nicolethoen (approved):

KKoukiou:
> Tooltip option looks good, but some of the Discrete slider examples seem to be broken - specifically the first one and the first two Vale input examples. I cannot drag the slider.

This problem exists already upstream, see on the docs pge directly without my PR https://www.patternfly.org/v4/components/slider/

mcarrano (approved):
Fair enough @KKoukiou . In that case I will approve this PR, but @nicolethoen @tlabaj do you have any idea when or why these examples broke. We should try to clean them up for the next release. Do we need a new issue for this?

KKoukiou:
@mcarrano I created an issue last week about it: https://github.com/patternfly/patternfly-react/issues/6644

packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -44,6 +45,8 @@ export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onCh
   inputAriaLabel?: string;
   /* Aria label for the thumb */
   thumbAriaLabel?: string;
+  /* Adds a tooltip over the thumb containing the current value */
+  thumbHasTooltip?: boolean;
```
tlabaj, line 49: Can we just call this `hasTooltip`.  This is more consistent with our naming practices.  There is also no other tooltip so the name will not conflict.

tlabaj (changes requested):
Looks great! One suggestion.

packages/react-core/src/components/Slider/Slider.tsx
nicolethoen: @KKoukiou @tlabaj 
I think referencing _what_ has the tooltip in the prop name is wise - maybe `hasTooltipOverThumb` is descriptive and follows the naming patterns?

KKoukiou: Fixed! Sorry for the wrong time to reply here.

tlabaj (approved):
Looks Good!  Thank you!
Looks like you need to update the snapshots and we can merge.

PR #6627, opened 2021-11-18
feat(MenuToggle): add plain with text variant

### LINKED ISSUE #6561: Menu toggle: add plain variant with text
Adds a new Menu toggle variant to be used for building composable menu demos. Core PR: https://github.com/patternfly/patternfly/pull/4491

nicolethoen:
**What**: Closes #6561

Added a new Menu Toggle example.

### DIFF
--- a/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
+++ b/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
@@ -15,7 +15,7 @@ export interface MenuToggleProps
   /** Flag indicating the toggle is disabled */
   isDisabled?: boolean;
   /** Variant styles of the menu toggle */
-  variant?: 'default' | 'plain' | 'primary';
+  variant?: 'default' | 'plain' | 'primary' | 'plainText';
   /** Optional icon rendered inside the toggle, before the children content */
   icon?: React.ReactNode;
   /** Optional badge rendered inside the toggle, after the children content */
@@ -55,7 +55,8 @@ export class MenuToggleBase extends React.Component<MenuToggleProps> {
           styles.menuToggle,
           isExpanded && styles.modifiers.expanded,
           variant === 'primary' && styles.modifiers.primary,
-          isPlain && styles.modifiers.plain,
+          (isPlain || variant === 'plainText') && styles.modifiers.plain,
+          variant === 'plainText' && styles.modifiers.text,
           className
         )}
         type="button"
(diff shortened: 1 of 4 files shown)


gabipodolnikova (approved):
LGTM!

mcarrano (commented):
This looks good. @mcoker @mceledonia I think we covered this before, but I assume we are good with the idea that there is no hover effect on these plain variations?

wise-king-sullyman (approved):

mceledonia:
What will it look like in terms of dev time if we add a hover state here to the caret? It will be subtle but consistent with our other icon action patterns. It would have the same colors/states, secondary text color (default) -> primary text color (hover) just to give it a small sense of feedback.

I know this isn't what we do on other dropdowns/selects, but it does feel like it needs some kind of state change since other similar components all do. 

That being said I am sort of on the fence here, if the juice isn't worth the squeeze I think the component will be fine without a hover state.

image


CC @mcoker @mcarrano

packages/react-core/src/components/MenuToggle/MenuToggle.tsx
```diff
@@ -55,7 +55,8 @@ export class MenuToggleBase extends React.Component<MenuToggleProps> {
           styles.menuToggle,
           isExpanded && styles.modifiers.expanded,
           variant === 'primary' && styles.modifiers.primary,
-          isPlain && styles.modifiers.plain,
+          (isPlain || variant === 'plainText') && styles.modifiers.plain,
```
tlabaj, line 58: Can you move the logic of checking the variant to line 39 where `isPlain` is set.

tlabaj (changes requested):

mcoker:
> What will it look like in terms of dev time if we add a hover state here to the caret?

@mceledonia pretty easy, we can try and get that in this release if you'd like. I'm assuming it should apply to all plain text toggles? Dropdown, options menu, select, etc?

nicolethoen:
@mcoker 
If you update the hover styles, that shouldn't require an update on the react side, right? 
I opened a core issue and you can prioritize it however @mcarrano and @mceledonia see fit.

nicolethoen:
Here is the new issue https://github.com/patternfly/patternfly/issues/4552

mcarrano:
Thanks @nicolethoen . I will add this issue to the next milestone.

packages/react-core/src/components/MenuToggle/MenuToggle.tsx
```diff
@@ -37,6 +37,8 @@ export class MenuToggleBase extends React.Component<MenuToggleProps> {
     const { children, className, icon, badge, isExpanded, isDisabled, variant, innerRef, ...props } = this.props;
 
     const isPlain = variant === 'plain';
+    const isPlainText = variant === 'plainText';
```
tlabaj, line 40: you could combine these into one then line 58 would not have to change.
`const isPlain = variant === 'plain' || variant === 'plainText'`

nicolethoen: Then i'd have to change the logic on lines 71-72. which I could do. But it's not true that if `plaintext` then `plain`. Plaintext also uses the expand/collapse arrow, unlike plain.

tlabaj (approved):

PR #6628, opened 2021-11-18
feat(Alert): Allow all heading levels for title

### LINKED ISSUE #6444: Alert title is forcing a h4 tag even if a h4 isn't appropriate in context
When I use the Alert component, the text entered at the title prop is rendered inside an `h4` tag.  This may not be the appropriate heading or a heading may not be appropriate for the content.  

For example, if my page looks like this:
```
<h2>My header</h2>
content 
ALERT GOES HERE
```

I would need the alert to render an h3 as the title with the child content below.  

In other cases, as in the inline alert with no children elements.  The content in the title attribute isn't a header at all, in that it doesn't describe the content below it. 

See WCAG 2.1  [1.3.1](https://www.w3.org/TR/WCAG21/#meaningful-sequence) and [2.4.6](https://www.w3.org/TR/WCAG21/#headings-and-labels)

**Expected**
I expect that I can override the default h4 tag with another tag that may be more appropriate in the context the alert is being used.

You can view this behavior on the [Alert demo page](https://www.patternfly.org/v4/components/alert)

I would call this a bug because it can lead to accessibility issues.  The current workaround is not to use the Alert component and use custom code.

OpenShift 4.10

nicolethoen:
**What**: Closes #6444

Added new alerts under 'Variations' example

### DIFF
--- a/packages/react-core/src/components/Alert/Alert.tsx
+++ b/packages/react-core/src/components/Alert/Alert.tsx
@@ -27,6 +27,8 @@ export interface AlertProps extends Omit<React.HTMLProps<HTMLDivElement>, 'actio
   isPlain?: boolean;
   /** Title of the alert  */
   title: React.ReactNode;
+  /** Sets the heading level to use for the alert title. Default is h4. */
+  titleHeadingLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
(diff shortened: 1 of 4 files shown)


evwilkin (approved):
LGTM

mcoker (approved):
LGTM!!! 🔥

tlabaj (approved):
LGTM

PR #6632, opened 2021-11-19
Allows DatePicker users to open the calendar programatically

### LINKED ISSUE #6336: Ability to open a DatePicker programmatically
I would like to be able to open/close the picker programatically.

I was requested a design where the user can't write on the date, so it's readonly. Additionally it's being requested that when the user clicks the text/date, the popover opens:

Peek 2021-09-20 16-58

enhancement

Notification's service for Insights

josejulio:
Adds an interface for controlling the component. Currently it allows to open/close/toggle the calendar.
It's totally optional to use it and doesn't breaks the current API.

Usage:
```ts

// In a component code
const dateRef = useRef<DatePickerRef>(null);
// ...
const onClick = () => {
    dateRef.current?.setCalendarOpen(true);
};
// ...

return <DatePicker /* ... other options ... */ ref={ dateRef }  />

```

Fixes #6336

### DIFF
--- a/packages/react-core/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-core/src/components/DatePicker/DatePicker.tsx
@@ -7,6 +7,7 @@ import { Popover, PopoverProps } from '../Popover/Popover';
 import { InputGroup } from '../InputGroup/InputGroup';
 import OutlinedCalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/outlined-calendar-alt-icon';
 import { CalendarMonth, CalendarFormat, isValidDate } from '../CalendarMonth';
+import { useImperativeHandle } from 'react';
 
 export interface DatePickerProps
   extends CalendarFormat,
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -43,169 +44,190 @@ export interface DatePickerProps
...
+      longWeekdayFormat,
+      dayFormat,
+      weekStart,
+      validators = [],
+      rangeStart,
+      style: styleProps = {},
+      inputProps = {},
+      ...props
+    }: DatePickerProps,
+    ref
```
nicolethoen, line 86: I like this solution - for ease of future maintenance, could you format the forward ref in a pattern consistent with our other components? For example, [here is how we do it for Button](https://github.com/patternfly/patternfly-react/blob/fe2d3b55de1b60cf0c5740fb5180beece366d68e/packages/react-core/src/components/Button/Button.tsx#L163).

josejulio: Yes, I'll update it, thanks!

josejulio: I did split the declaration of the forward ref. Let me know if that's what you had in mind :)

nicolethoen (approved):

tlabaj (changes requested):
Looks good.  Can you just add comments to the `DatePickerRef`  interface props and add it to the docs so consumers are aware. You do this by adding it to the `propsComponents` array in the examples md file. I think it would also be helpful to add an example.

josejulio:
> Looks good. Can you just add comments to the `DatePickerRef` interface props and add it to the docs so consumers are aware. You do this by adding it to the `propsComponents` array in the examples md file. I think it would also be helpful to add an example.

Done! Please let me know if that's what you had in mind

Screenshot_2021-11-30_09-34-27
Peek 2021-11-30 09-34

packages/react-core/src/components/DatePicker/examples/DatePicker.md
```diff
@@ -128,3 +128,27 @@ ControlledDate = () => {
   );
 }
 ```
+
+### Controlling the DatePicker calendar state
```
tlabaj, line 132: nit.  Can you please change to sentence case.
`Controlling the date picker calendar state`

tlabaj:
Looks great!  Just one little nit.

packages/react-core/src/components/DatePicker/examples/DatePicker.md
josejulio: done :)

tlabaj (approved):
LGTM!

PR #6634, opened 2021-11-19
chore(Alert): Convert examples to functional with typescript

### LINKED ISSUE #6633: Chore: Convert Alert examples to typescript

nicolethoen:
**What**: Closes #6633

### DIFF
--- a/packages/react-core/src/components/Alert/examples/Alert.md
+++ b/packages/react-core/src/components/Alert/examples/Alert.md
@@ -16,412 +16,245 @@ import LaptopIcon from '@patternfly/react-icons/dist/esm/icons/laptop-icon';
 ## Examples
 
 ### Types
-```js
+```ts
 import React from 'react';
 import { Alert } from '@patternfly/react-core';
 
-class AlertTypes extends React.Component {
-  render() {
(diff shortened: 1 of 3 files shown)


kmcfaul (approved):

gabipodolnikova:
Just one small thing: I see the `Async live region alert` example with a differently styled buttons than in the other examples. Do you also see it?

nicolethoen:
@gabipodolnikova yes! That's because it's using a toggleGroup instead of buttons. it is sort of an on/off switch, when the others are buttons... is it distracting? I did update it to use the toggleGroup - I figured it was a good use case but it is unnecessary.

gabipodolnikova:
> @gabipodolnikova yes! That's because it's using a toggleGroup instead of buttons. it is sort of an on/off switch, when the others are buttons... is it distracting? I did update it to use the toggleGroup - I figured it was a good use case but it is unnecessary.

Ah okay, it just surprised me, that it looks different than the others. it is a little bit distracting for me, but maybe get another opinion. Otherwise I am approving this :)

gabipodolnikova (approved):

PR #6639, opened 2021-11-23
chore(Card): Updated Card demos to use Select instead of Dropdown

### LINKED ISSUE #6560: Update dashboard card demos to use Select
Dashboard card demos should be using a Select and not a Dropdown component for filtering. They should be Selects without borders (plain variants). This applies to the following demos:

Utilization card 3: https://www.patternfly.org/v4/components/card/react-demos#utilization-card-3
Trend card 1: https://www.patternfly.org/v4/components/card/react-demos#trend-card-1
Log view: https://www.patternfly.org/v4/components/card/react-demos#log-view
Events view: https://www.patternfly.org/v4/components/card/react-demos#events-view

gabipodolnikova:
**What**: Closes #6560

### DIFF
--- a/packages/react-core/src/demos/CardDemos.md
+++ b/packages/react-core/src/demos/CardDemos.md
@@ -1247,11 +1247,10 @@ const UtilizationCard3: React.FunctionComponent = () => {
   const [isOpen, setIsOpen] = React.useState(false);
 
   const selectItems = [
-    <SelectOption value="Link" key="link" />,
-    <SelectOption value="Action" key="action" component="button" />,
-    <SelectOption value="Disabled link" key="disabled link" isDisabled />,
-    <Divider component="li" key="separator" />,
-    <SelectOption value="Separated link" key="separated link" />
+    <SelectOption value="Last hour" key="option1" />,
+    <SelectOption value="Last 6 hours" key="option2" />,
+    <SelectOption value="Last 24 hours" key="option3" />,
+    <SelectOption value="Last 7 days" key="option4" />
   ];
 
   return (
@@ -2434,11 +2433,10 @@ const TrendCard1: React.FunctionComponent = () => {
   const [isOpen, setIsOpen] = React.useState(false);
 
   const selectItems = [
-    <SelectOption value="Link" key="link" />,
-    <SelectOption value="Action" key="action" component="button" />,
-    <SelectOption value="Disabled link" key="disabled link" isDisabled />,
-    <Divider component="li" key="separator" />,
-    <SelectOption value="Separated link" key="separated link" />
+    <SelectOption value="Last hour" key="option1" />,
+    <SelectOption value="Last 6 hours" key="option2" />,
+    <SelectOption value="Last 24 hours" key="option3" />,
+    <SelectOption value="Last 7 days" key="option4" />
   ];
   return (
     <Gallery hasGutter minWidths={{ default: '360px' }}>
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (approved):

nicolethoen (approved):

mcarrano (changes requested):
This looks good @gabipodolnikova , but I just had one small request. Can we change the menu items in these filter select lists to be something more relevant. What about having, "Last hour," "Last 6 hours," "Last 24 hours," and "Last 7 days?" This will just emphasize that this is selection of options and not actions.

Also, I'm wondering why the plain select seems slightly misaligned with the card title. @mcoker can you check this? Is there a way to align these better?

gabipodolnikova:
> This looks good @gabipodolnikova , but I just had one small request. Can we change the menu items in these filter select lists to be something more relevant. What about having, "Last hour," "Last 6 hours," "Last 24 hours," and "Last 7 days?" This will just emphasize that this is selection of options and not actions.

Yes, will do that 👍 

> Also, I'm wondering why the plain select seems slightly misaligned with the card title. @mcoker can you check this? Is there a way to align these better?

I also notice this, it is more visible with the plain variant, which in the HTML demos they are not plain.

mcoker:
> Also, I'm wondering why the plain select seems slightly misaligned with the card title.

@mcarrano it's because the card title is designed to be 16px but we're using a large title variation in that card. If we're changing the title font size, I think it's expected you might need to change the action position to match.

We could
* update the card title to be the normal font size
* use the regular select toggle (with the border) - as @gabipodolnikova said, it's less obvious with that kind of select
* use custom styling/layout to see if we can better align the title and select in these demos

wdyt?

mcarrano:
@gabipodolnikova @mcoker I don't want to reintroduce the borders to the select menus as we had previously determined that the plain Selects look best here. Perhaps we could reduce the card titles normal size. @mceledonia do you think that the titles need to be as large as they are? Or we could live with this...

mceledonia:
I am hesitant to lower the title size of these demos since the typographic hierarchy is really what makes them. In other words the whole point of these trend cards is to have nice big numbers you can easily see at a glance. 

What does using custom styling look like to make the adjustment in terms of best practice / cleanliness?

mcoker:
@mceledonia custom styling would just be some sort of manual adjustment to the action position to match whatever font size was used for the card title. So you would push it down a bit. Also how do you think the action should align with the text? Right now the top of the title text is aligned with the top of the action text. That should scale regardless the font size used and number of lines in the title text since we don't have to know the card title font size to align it that way.

Screen Shot 2021-12-02 at 9 37 17 AM

We have a modifier to use there with non-plain buttons that removes the offset we put in place to align the top of the action text with the top of the title. Here's what that looks like with this title 

Screen Shot 2021-12-02 at 9 37 25 AM

^ it looks pretty good, but that's also kind of coincidental since it's just removing the button offset. It wouldn't work, for example, with larger text.

Screen Shot 2021-12-02 at 9 41 18 AM

Also just to confirm, the ideal alignment would be "baseline" where the bottom of the first line of title text is aligned with the bottom of the button text, as if they were on the same line?

mcoker:
@mceledonia @mcarrano I've opened an issue here if we would like to explore this alignment some more in core - https://github.com/patternfly/patternfly/issues/4553

kmcfaul (approved):
LGTM on the react-side of things

tlabaj (approved):

mcarrano (approved):
Looks good @gabipodolnikova . We can address the alignment issue in a future milestone.

PR #6645, opened 2021-11-24
chore(Tabs): added nested Tabs demo

### LINKED ISSUE #6513: Nested tabs

gabipodolnikova:
**What**: Closes #6513


packages/react-core/src/demos/Tabs.md
```diff
@@ -437,3 +437,132 @@ TabsOpenWithSecondaryTabsDemo = () => {
...
+          Overview
+        </Title>
+      </PageSection>
+      <PageSection type="tabs" variant={PageSectionVariants.light} isWidthLimited>
+        <Tabs activeKey={activeTabKey} onSelect={handleTabClick} usePageInsets id="open-nested-tabs-example-tabs-list">
+          <Tab eventKey={0} title={<TabTitleText>Cluster 1</TabTitleText>} tabContentId={`tabContent${0}`} />
+          <Tab eventKey={1} title={<TabTitleText>Cluster 2</TabTitleText>} tabContentId={`tabContent${1}`} />
+        </Tabs>
+      </PageSection>
+      <PageSection isWidthLimited variant={PageSectionVariants.light}>
```
nicolethoen, line 557: I don't think you want a light page section variant here. If you remove it, it will make the background of the main section grey like the core demo

gabipodolnikova: Fixed, thanks!

packages/react-core/src/demos/Tabs.md
```diff
@@ -437,3 +437,132 @@ TabsOpenWithSecondaryTabsDemo = () => {
...
+  Label,
+  LabelGroup,
+  Flex,
+  FlexItem
+} from '@patternfly/react-core';
+import DashboardWrapper from './examples/DashboardWrapper';
+import CheckCircleIcon from '@patternfly/react-icons/dist/js/icons/check-circle-icon';
+import InfoCircleIcon from '@patternfly/react-icons/dist/js/icons/info-circle-icon';
+
+TabsOpenNestedDemo = () => {
```
wise-king-sullyman, line 468: We're missing the declaration for this variable

wise-king-sullyman (changes requested):
I've commented on a few small things that I think should be addressed.

Additionally, there's a handful of small formatting things that I think should be fixed (for example the multi-line but blockless fat arrow functions). I elected not to litter this PR with comments pointing them each out individually though; as I believe they will be best addressed by just running a formatter on it.

Since the formatters don't see JS in the markdown code blocks (or at least I haven't figured out a way to make them do so) I would suggest either having a separate demo file (like I did in #6661) or just copy/pasting this to/from a `temp.js` file to do the formatting fixes for you.

gabipodolnikova:
> Since the formatters don't see JS in the markdown code blocks (or at least I haven't figured out a way to make them do so) I would suggest either having a separate demo file (like I did in #6661) or just copy/pasting this to/from a `temp.js` file to do the formatting fixes for you.
 
Cool thanks, I also moved it to a separate file.

nicolethoen (approved):

wise-king-sullyman (approved):

kmcfaul (approved):
Looks good!

mcarrano (approved):
This looks good to me. @mceledonia can you also take a look?

mceledonia:
These look good, are we able to offer a full width bottom border for these to sit on as an option? I recall seeing that in earlier designs and I think it would be a nice option for these nested tabs, so I want to be sure we don't miss it.

PR #6652, opened 2021-11-29
feat(DualListSelector): add composable tree example

### LINKED ISSUE #6401: DualListSelector: create composable tree example
consider using a treeview rather than its own native tree implementation
will require cypress and jest tests to be updated as well

make sure all examples are keyboard and screen reader accessible

### LINKED ISSUE #6424: DualListSelector - tree variant performance improvements
Large data sets for DualListSelector have a large amount of latency as well (same structure as TreeView). Memoization similar to TreeView (PR https://github.com/patternfly/patternfly-react/pull/6362) should be implemented here as well. Potentially also investigate using a non-recursive solution for further performance improvements.

nicolethoen:
**What**: Closes #6401, Closes #6424

### DIFF
--- a/packages/react-core/src/components/DualListSelector/examples/ComposableTree.tsx
+++ b/packages/react-core/src/components/DualListSelector/examples/ComposableTree.tsx
@@ -24,57 +24,7 @@ interface ExampleProps {
   data: FoodNode[];
 }
 
-export const ComposableDualListSelectorTree: React.FunctionComponent<ExampleProps> = ({
-  data = [
-    {
-      id: 'fruits',
-      text: 'Fruits',
-      children: [
-        { id: 'apple', text: 'Apple' },
-        {
-          id: 'berries',
-          text: 'Berries',
-          children: [
-            { id: 'blueberry', text: 'Blueberry' },
-            { id: 'strawberry', text: 'Strawberry' }
-          ]
-        },
-        { id: 'banana', text: 'Banana' }
-      ]
-    },
-    { id: 'bread', text: 'Bread' },
-    {
-      id: 'vegetables',
-      text: 'Vegetables',
-      children: [
-        { id: 'broccoli', text: 'Broccoli' },
-        { id: 'cauliflower', text: 'Cauliflower' }
-      ]
-    },
-    {
-      id: 'proteins',
-      text: 'Proteins',
-      children: [
-        { id: 'beans', text: 'Beans' },
-        {
-          id: 'meats',
-          text: 'Meats',
-          children: [
-            {
-              id: 'beef',
-              text: 'Beef'
-            },
-            {
-              id: 'chicken',
-              text: 'Chicken'
-            }
-          ]
-        },
-        { id: 'tofu', text: 'Tofu' }
-      ]
-    }
-  ]
-}: ExampleProps) => {
+export const ComposableDualListSelectorTree: React.FunctionComponent<ExampleProps> = ({ data }: ExampleProps) => {
   const [checkedLeafIds, setCheckedLeafIds] = React.useState<string[]>([]);
   const [chosenLeafIds, setChosenLeafIds] = React.useState<string[]>(['beans', 'beef', 'chicken', 'tofu']);
   const [chosenFilter, setChosenFilter] = React.useState<string>('');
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/DualListSelector/examples/ComposableTree.tsx
```diff
@@ -0,0 +1,344 @@
...
+    const descendentLeafIds = memoizedLeavesById[node.id];
+    const descendentsOnThisPane = isChosen
+      ? descendentLeafIds.filter(id => chosenLeafIds.includes(id))
+      : descendentLeafIds.filter(id => !chosenLeafIds.includes(id));
+
+    const hasMatchingChildren =
+      filterValue && descendentsOnThisPane.some(id => memoizedNodeText[id].includes(filterValue));
+    const isFilterMatch = filterValue && node.text.includes(filterValue) && descendentsOnThisPane.length > 0;
+
+    // A node is displayed if:
```
tlabaj, line 192: nit. can you update comment to say if either is true just to be more clear

packages/react-core/src/components/DualListSelector/examples/ComposableTree.tsx
```diff
@@ -0,0 +1,344 @@
...
+      nodeTexts = { ...nodeTexts, ...buildTextById(foodNode) };
+      leavesById = { ...leavesById, ...getLeavesById(foodNode) };
+      allLeaves = [...allLeaves, ...getDescendantLeafIds(foodNode)];
+    });
+    return {
+      memoizedLeavesById: leavesById,
+      memoizedAllLeaves: allLeaves,
+      memoizedNodeText: nodeTexts
+    };
+  }, [data]);
```
wise-king-sullyman, line 92: Shouldn't this be in the form of:

```js
  const memoizeNodes = (data: FoodNode[]) => {
    let leavesById = {};
    let allLeaves = [];
    let nodeTexts = {};
    data.forEach(foodNode => {
      nodeTexts = { ...nodeTexts, ...buildTextById(foodNode) };
      leavesById = { ...leavesById, ...getLeavesById(foodNode) };
      allLeaves = [...allLeaves, ...getDescendantLeafIds(foodNode)];
    });
    return {
      memoizedLeavesById: leavesById,
      memoizedAllLeaves: allLeaves,
      memoizedNodeText: nodeTexts
    };
  };

  // Builds a map of child leaf nodes by node id - memoized so that it only rebuilds the list if the data changes.
  const { memoizedLeavesById, memoizedAllLeaves, memoizedNodeText } = React.useMemo(() => memoizeNodes(data), [data]);
```
with `data` being passed as an argument to the memoization callback?

This is a genuine question as I'm not well versed with React memoization and I'm not sure how to really test this without a large dataset, I just saw that in the React docs they show usage as:

image

nicolethoen: My understanding is that `React.useMemo` works like `React.useEffect` where it watches for changes in the dependencies listed in the dependency array at the end - and it only recalculates the value of the `memoizedValue` if those dependencies change.

wise-king-sullyman: I misunderstood something from the docs, disregard :sweat_smile:

mturley: @wise-king-sullyman I think both are correct. Defining a dedicated `memoizeNodes` function at render time just before passing useMemo an anonymous function that calls it behaves the same as defining that behavior directly in the anonymous function. The important part is the `[data]` dependency array which will prevent the whole anonymous function from being run again if the data didn't change. So it's a matter of preference - I think the example in the React docs was written that way for brevity.

packages/react-core/src/components/DualListSelector/examples/ComposableTree.tsx
```diff
@@ -0,0 +1,344 @@
...
+                ? buildOptions(isChosen, node.children, isFilterMatch || hasParentMatch)
+                : undefined
+            }
+          ]
+        : []),
+      ...(!isDisplayed && node.children && node.children.length
+        ? buildOptions(isChosen, node.children, hasParentMatch)
+        : []),
+      ...(remainingNodes ? buildOptions(isChosen, remainingNodes, hasParentMatch) : [])
+    ];
```
wise-king-sullyman, line 227: This is pretty dense, I think some newlines between the spreads would improve readability if nothing else.

wise-king-sullyman (approved):
I have one small suggestion but this LGTM overall so I'm comfortable approving whether you change it or not.

tlabaj (approved):

PR #6653, opened 2021-11-30
fix(Tile): fix selecting tile in tile demo

### LINKED ISSUE #6637: Inconsistent selection in Tile component
In https://www.patternfly.org/v4/components/tile/react-demos/ Sometimes I can get the selection to stick and other times not. Maybe a result of recent changes with adding animation raised tile? Should make sure we don't have the same issue with cards. There is not a demo to test it there.

cc @nicolethoen

nicolethoen:
**What**: Closes #6637

Also cleans up some unused imports in the Card examples

### DIFF
--- a/packages/react-core/src/components/Card/examples/Card.md
+++ b/packages/react-core/src/components/Card/examples/Card.md
@@ -77,7 +77,6 @@ import {
   CardFooter,
   Checkbox,
   Dropdown,
-  DropdownToggle,
   DropdownItem,
   DropdownSeparator,
   KebabToggle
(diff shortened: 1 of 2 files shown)


kmcfaul (approved):

mcarrano (changes requested):
I still cannot get the tiles to reliably stick in the second demo example (multi-select). Sometime they will sometimes they won't. Testing with Firefox.

nicolethoen:
@mcarrano good catch! I fixed it

gabipodolnikova (approved):

mcarrano (approved):
Looks great!

tlabaj (approved):

PR #6659, opened 2021-11-30
feat(Card): Add disabled card and clean up props

### LINKED ISSUE #6567: Non-selectable card
Add a new card variant per this core PR: https://github.com/patternfly/patternfly/pull/4500

nicolethoen:
**What**: Closes #6567

### DIFF
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -18,10 +18,10 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   isCompact?: boolean;
   /** Modifies the card to include selectable styling */
   isSelectable?: boolean;
+  /** @beta Specifies the card is selectable, and applies the new raised styling on hover and select */
+  isSelectableRaised?: boolean;
   /** Modifies the card to include selected styling */
   isSelected?: boolean;
-  /** @beta Specifies the selectable styling variant */
-  selectableVariant?: 'legacy' | 'raised';
   /** @beta Modifies a selectable card to have disabled styling */
   isDisabled?: boolean;
   /** Modifies the card to include flat styling */
@@ -56,8 +56,8 @@ export const Card: React.FunctionComponent<CardProps> = ({
   isHoverable = false,
   isCompact = false,
   isSelectable = false,
+  isSelectableRaised = false,
   isSelected = false,
-  selectableVariant = 'legacy',
   isDisabled = false,
   isFlat = false,
   isExpanded = false,
@@ -83,16 +83,14 @@ export const Card: React.FunctionComponent<CardProps> = ({
   }
 
   const getSelectableModifiers = () => {
-    const isRaised = selectableVariant === 'raised';
-
+    if (isSelectableRaised) {
+      if (isDisabled) {
+        return css(styles.modifiers.nonSelectableRaised);
+      }
+      return css(styles.modifiers.selectableRaised, isSelected && styles.modifiers.selectedRaised);
+    }
     if (isSelectable || isHoverable) {
-      if (isRaised) {
-        if (isDisabled) {
-          return css(styles.modifiers.nonSelectableRaised);
-        } else {
-          return css(styles.modifiers.selectableRaised, isSelected && styles.modifiers.selectedRaised);
-        }
-      } else {
+      if (!isDisabled) {
         return css(styles.modifiers.selectable, isSelected && styles.modifiers.selected);
       }
     }
@@ -120,7 +118,7 @@ export const Card: React.FunctionComponent<CardProps> = ({
           getSelectableModifiers(),
           className
         )}
-        tabIndex={isSelectable ? '0' : undefined}
+        tabIndex={isSelectable || isSelectableRaised ? '0' : undefined}
         {...props}
         {...ouiaProps}
       >
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -12,20 +12,18 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   className?: string;
   /** Sets the base component to render. defaults to article */
   component?: keyof JSX.IntrinsicElements;
-  /** Modifies the card to include hover styles on :hover */
+  /** @deprecated Modifies the card to include hover styles on :hover */
   isHoverable?: boolean;
-  /** @beta Modifies the card to include hoverable-raised styles on :hover, this styling is included by default with isSelectableRaised */
-  isHoverableRaised?: boolean;
   /** Modifies the card to include compact styling. Should not be used with isLarge. */
   isCompact?: boolean;
   /** Modifies the card to include selectable styling */
   isSelectable?: boolean;
   /** Modifies the card to include selected styling */
   isSelected?: boolean;
-  /** @beta Modifies the card to include selectable-raised styling and hoverable-raised styling */
-  isSelectableRaised?: boolean;
-  /** @beta Modifies the card to include selected-raised styling */
-  isSelectedRaised?: boolean;
+  /** @beta Specifies the selectable styling variant */
+  selectableVariant?: 'legacy' | 'raised';
+  /** @beta Modifies a selectable card to have disabled styling */
+  isDisabled?: boolean;
   /** Modifies the card to include flat styling */
   isFlat?: boolean;
   /** Modifies the card to include rounded styling */
@@ -56,12 +54,11 @@ export const Card: React.FunctionComponent<CardProps> = ({
   className = '',
   component = 'article',
   isHoverable = false,
-  isHoverableRaised = false,
   isCompact = false,
   isSelectable = false,
   isSelected = false,
-  isSelectableRaised = false,
-  isSelectedRaised = false,
+  selectableVariant = 'legacy',
+  isDisabled = false,
   isFlat = false,
   isExpanded = false,
   isRounded = false,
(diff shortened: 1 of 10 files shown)


packages/react-integration/demo-app-ts/src/components/demos/CardDemo/CardDemo.tsx
```diff
@@ -93,7 +93,7 @@ export class CardDemo extends React.Component {
 
     return (
       <React.Fragment>
-        <Card isHoverable>
+        <Card isSelectable selectableVariant="raised">
```
mcoker, line 96: Just a thought - could we combine these so you don't need separate props? Wondering if `isSelectable="raised"` could apply the new modifier.

mcoker (commented):
LGTM! Just one question.

packages/react-integration/demo-app-ts/src/components/demos/CardDemo/CardDemo.tsx
nicolethoen: This is a good thought - I figured that adding the selectableVariant="raised" would mean that to opt into the new styling, they'd just need to add indicate they wanted to using the new prop. And eventually, when raised is the default, then it wont be necessary to specify that.

mcarrano (approved):
Looks great. Thanks @nicolethoen !

mcoker (approved):
🥳

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -79,6 +76,33 @@ export const Card: React.FunctionComponent<CardProps> = ({
...
+          (isSelectable || isHoverable) && isSelected && styles.modifiers.selectedRaised
+        );
+      }
+    } else {
+      selectableModifiers = css(
+        (isSelectable || isHoverable) && styles.modifiers.selectable,
+        (isSelectable || isHoverable) && isSelected && styles.modifiers.selected
+      );
+    }
+  }
```
wise-king-sullyman, line 104: I feel like this would be cleaner as a function rather than mutating a `let`.

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -79,6 +76,33 @@ export const Card: React.FunctionComponent<CardProps> = ({
...
+      } else {
+        selectableModifiers = css(
+          (isSelectable || isHoverable) && styles.modifiers.selectableRaised,
+          (isSelectable || isHoverable) && isSelected && styles.modifiers.selectedRaised
+        );
+      }
+    } else {
+      selectableModifiers = css(
+        (isSelectable || isHoverable) && styles.modifiers.selectable,
+        (isSelectable || isHoverable) && isSelected && styles.modifiers.selected
```
wise-king-sullyman, line 101: Also, I don't think the repeated `isSelectable || isHoverable` conditionals here are needed as they are already nested within an `if` checking that conditional, so they will always be true. I could be missing something though.

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -12,20 +12,18 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
...
   /** Modifies the card to include selectable styling */
   isSelectable?: boolean;
   /** Modifies the card to include selected styling */
   isSelected?: boolean;
-  /** @beta Modifies the card to include selectable-raised styling and hoverable-raised styling */
-  isSelectableRaised?: boolean;
-  /** @beta Modifies the card to include selected-raised styling */
-  isSelectedRaised?: boolean;
+  /** @beta Specifies the selectable styling variant */
+  selectableVariant?: 'legacy' | 'raised';
```
tlabaj, line 24: legacy implies the non raised variant will eventually be deprecated.  is that the case @mcarrano?

packages/react-integration/cypress/integration/card.spec.ts
```diff
@@ -5,12 +5,12 @@ describe('Card Demo Test', () => {
 
   it('Verify that selectable card can be selected and unselected with keyboard input', () => {
     cy.get('#selectableCard').focus();
-    cy.focused().should('have.class', 'pf-m-selectable');
```
tlabaj, line 8: So non raised card can no longer be selected or selectable?

tlabaj (changes requested):
I have a few questions about the logic.

packages/react-core/src/components/Card/Card.tsx
nicolethoen: I figured it means it's not preferred and the 'old but supported' way

packages/react-integration/cypress/integration/card.spec.ts
nicolethoen: correct - I think it's the same as disabled from my understanding

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -12,20 +12,18 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
...
   component?: keyof JSX.IntrinsicElements;
-  /** Modifies the card to include hover styles on :hover */
+  /** @deprecated Modifies the card to include hover styles on :hover */
   isHoverable?: boolean;
-  /** @beta Modifies the card to include hoverable-raised styles on :hover, this styling is included by default with isSelectableRaised */
-  isHoverableRaised?: boolean;
   /** Modifies the card to include compact styling. Should not be used with isLarge. */
   isCompact?: boolean;
   /** Modifies the card to include selectable styling */
   isSelectable?: boolean;
```
tlabaj, line 20: so, is this prop deprecated?

packages/react-integration/cypress/integration/card.spec.ts
tlabaj: I don't think that is the case.  We are still applying the style and we have an example for that use case.  We should be testing both here.

packages/react-core/src/components/Card/Card.tsx
tlabaj: Ok I see my confusion. I thought the selectable variant set the card to selectable.  That being said.  I think the API is a little confusing.  May we can rename the prop tp `selectableVariantStyling`.
I would also suggest not calling the option  `legacy`.  It is not really descriptive.

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -12,20 +12,18 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
...
   /** Modifies the card to include selected styling */
   isSelected?: boolean;
-  /** @beta Modifies the card to include selectable-raised styling and hoverable-raised styling */
-  isSelectableRaised?: boolean;
-  /** @beta Modifies the card to include selected-raised styling */
-  isSelectedRaised?: boolean;
+  /** @beta Specifies the selectable styling variant */
+  selectableVariant?: 'legacy' | 'raised';
+  /** @beta Modifies a selectable card to have disabled styling */
```
tlabaj, line 25: Can we be more explicit here that disabled only applies to non raised cards since we still support both variations.

packages/react-core/src/components/Card/Card.tsx
tlabaj: So I was thinking about this more.  Would it make sense to deprecate the `isSelectable` prop and let the selectableVariant prop not only set the styling, but set it selectable too?  I think that would make the API more straight forward.  Similar to what @mcoker  was suggesting in his comment about combining props.

packages/react-core/src/components/Card/Card.tsx
nicolethoen: As I have it, a disabled selectable card without the raised variant is basically a normal card with no hover/select styles at all. Still valid?

packages/react-core/src/components/Card/Card.tsx
kmcfaul: How about `selectStyle` or `selectableStyle` over using `variant`, since the function of the select remains the same?

packages/react-core/src/components/Card/Card.tsx
kmcfaul: Yeah I think only the disabled raised selectable card has any special styling. The flag probably shouldn't apply anything to a non-selectable non-raised card (or alternatively, maybe remove any select related props from non-raised cards?)

packages/react-core/src/components/Card/Card.tsx
kmcfaul: Are we going to keep supporting the original plain variation of selectable cards?

kmcfaul:
Code looks good to me the only outstanding question I've got is whether we want to keep supporting the original plain selectable style, as that would impact how we mark `isSelectable` and whether we want an enum prop or bool prop for the raised styling flag.

wise-king-sullyman (approved):

mcoker:
@kmcfaul I think it would be a breaking change if we didn't support it.

mcoker (approved):
LGTM! A thought about `isDisabled` - I wonder if we'll introduce a "disabled" card at some point that does something else, and if that might create a naming conflict. I don't know that we would, but figured I would mention it.

packages/react-core/src/components/Card/Card.tsx
tlabaj: Yes. I was confused at first.  I thought the original variant prop also set the card to be selectable.

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -90,22 +106,17 @@ export const Card: React.FunctionComponent<CardProps> = ({
         id={id}
         className={css(
           styles.card,
-          isHoverable && styles.modifiers.hoverable,
```
tlabaj, line 93: @mcoker, could someone apply `pf-m-hoverable`.  Without, `isSelectable`.  I think removing support completely would technically be breaking.  I think we can deprecate it, but we still need to apply the style.

mcoker: @tlabaj yep, they're separate. While you could use them together, it isn't necessary since `isSelected` has its own hoverable styles (they're the same as what `isHoverable` applies). https://github.com/patternfly/patternfly/blob/dc6cf07d69cbb80026dca7fe43dc28704aabea64/src/patternfly/components/Card/card.scss#L26-L27

mcarrano:
> Code looks good to me the only outstanding question I've got is whether we want to keep supporting the original plain selectable style, as that would impact how we mark isSelectable and whether we want an enum prop or bool prop for the raised styling flag.

@nicolethoen Yes, we still need to support the old selectable style. We didn't want to force a visual breaking change on products currently using selectable cards.

tlabaj:
@mcarrano is the intention that next time we have a breaking change release, the old styles will go away?

mcarrano:
> @mcarrano is the intention that next time we have a breaking change release, the old styles will go away?

Yes, most likely.

packages/react-core/src/components/Card/__tests__/Card.test.tsx
```diff
@@ -35,11 +35,6 @@ test('card with isHoverable applied ', () => {
   expect(view).toMatchSnapshot();
 });
 
-test('card with isHoverableRaised applied ', () => {
```
tlabaj, line 38: this test is still applicable.

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -79,6 +76,25 @@ export const Card: React.FunctionComponent<CardProps> = ({
     console.warn('Card: Cannot use isCompact with isLarge. Defaulting to isCompact');
     isLarge = false;
   }
+
+  if (isHoverable) {
+    // eslint-disable-next-line no-console
+    console.warn("Card: the 'isHoverable' prop is deprecated. Use isSelectable instead.");
```
tlabaj, line 82: I don't thinks this is necessary.  We don't do this anywhere else that I can find.

tlabaj (approved):
LGTM

PR #6660, opened 2021-11-30
fix(popover): forwarded focus to close button with VoiceOver click

### LINKED ISSUE #4210: Popover interaction inconsistency

evwilkin:
**What**: Closes #4210 

This PR:
- Forwards focus to close button within popover when triggering with VoiceOver click
  - `cmd+F5` on a Mac to enable VO, then `ctrl+option+space` on button to trigger Popover (see [Activate a link or form control](https://dequeuniversity.com/screenreaders/voiceover-keyboard-shortcuts#vo-mac-navigation))
- Removes the optional `onTriggerEnter` function passed as a prop to Popper
  - This function was duplicating the `onTriggerClick` functionality but targeted only at the `Enter` key, but that key press was triggering both functions so this seems to be unnecessary.

### DIFF
--- a/packages/react-core/src/components/Popover/Popover.tsx
+++ b/packages/react-core/src/components/Popover/Popover.tsx
@@ -334,23 +334,6 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
       }
     }
   };
-  const onTriggerEnter = (event: KeyboardEvent) => {
-    if (event.keyCode === KEY_CODES.ENTER) {
-      if (!visible) {
-        if (triggerManually) {
-          shouldOpen(show, event);
-        } else {
-          show(true);
-        }
-      } else {
-        if (triggerManually) {
-          shouldClose(null, hide, event);
-        } else {
-          hide();
-        }
-      }
-    }
-  };
   const onTriggerClick = (event: MouseEvent) => {
     if (triggerManually) {
       if (visible) {
@@ -362,7 +345,7 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
       if (visible) {
         hide();
       } else {
-        show();
+        show(true);
       }
     }
   };
@@ -451,7 +434,6 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
         distance={distance}
         placement={position}
         onTriggerClick={onTriggerClick}
-        onTriggerEnter={onTriggerEnter}
         onDocumentClick={onDocumentClick}
         onDocumentKeyDown={onDocumentKeyDown}
         enableFlip={enableFlip}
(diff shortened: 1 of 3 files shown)


evwilkin:
@jschuler I know you've worked with Popper a lot and I believe wrote the code I removed here, WDYT?  Want to make sure there's not an edge case I'm unaware of to keep the `onTriggerEnter` function.

nicolethoen (commented):
It seems to me that that in this PR and in production I cannot get VO focus into the popover when it opens. VO focus stays on the toggle, even though it reads the contents of the popover. That is fine except in the example of [close-popover-from-content-controlled](https://patternfly-react-pr-6660.surge.sh/components/popover/react/close-popover-from-content-controlled/) where I really struggled to figure out how to use VO to hit the controlled close button.

jschuler (approved):
@evwilkin Not sure why i added this :) seems to all work well without it!

jessiehuff (commented):
I agree with Nicole's comment in that there are a few that don't switch focus for me (close popover from content (controlled) and the advanced example). However, I don't think that they currently switched focus through keyboard either looking at org. I'm ok with solving these here or in another issue/PR. I'm open to what you guys think! :)

tlabaj (approved):
LGTM

PR #6662, opened 2021-12-01
fix(Slider): fixed examples with custom steps and input

### LINKED ISSUE #6644: bug(slider): slider with input is broken
Trying to control the slider with the mouse in the existing examples here http://localhost:8002/components/slider#value-input does not work any more. 

Reverting ed777b2288167bd42fe4e5564510b1ef135164fa fixes the issue. Did not debug more.

gabipodolnikova:
**What**: Closes #6644
I can't think of a quicker fix...
I also added one more example, I don't know if it is necessary, it is the example from the original issue https://github.com/patternfly/patternfly-react/issues/5833


evwilkin (approved):
LGTM - thanks for the fix!

nicolethoen (approved):

tlabaj (approved):

PR #6664, opened 2021-12-01
feat(Popover): add composable header and alert variants

### LINKED ISSUE #6267: Popover: add alert variants
Want to add new popover variations that have alert colors and icons in the header per the issue here: patternfly/patternfly-design#945 Should support all the usual status colors: error, warning, info, and success.

Core issue: https://github.com/patternfly/patternfly/issues/4058

kmcfaul:
**What**: Closes #6267

- Adds `header` as an alternative, composable way to build a popover header
- Adds `PopoverHeader`, `PopoverHeaderIcon`, `PopoverHeaderText` components to build headers
- Adds `alertVariant` prop to `Popover` to control alert styling
- Renames the previous `PopoverHeader` to `PopoverInternalHeader`. This component wasn't exported before so this shouldn't be breaking.

After some review, the initial approach was altered. Now:

- Adds `alertSeverityVariant` and `alertSeverityScreenReaderText` prop to `Popover` to control alert styling
- Adds `PopoverHeader`, `PopoverHeaderIcon`, `PopoverHeaderText` components to build headers (not exposed)
- Adds `headerIcon`
- When `headerIcon` or `alertSeverityVariant` are defined, the `headerContent` is placed in `PopoverHeaderText` rather than a `Title` component, and the whole header is wrapped in a `header` element.


mcarrano (approved):
This looks great. Thanks @kmcfaul !

nicolethoen (approved):
VO is not announcing the alerts.
I think that @evwilkin has opened a PR to address that? I think if his gets merged first, then we should rebase and double check it.

packages/react-core/src/components/Popover/examples/Popover.md
```diff
@@ -272,3 +277,85 @@ class AdvancedPopover extends React.Component {
...
+```ts
+import React from 'react';
+import { Popover, PopoverHeader, PopoverHeaderIcon, PopoverHeaderText, Button } from '@patternfly/react-core';
+import BullhornIcon from '@patternfly/react-icons/dist/esm/icons/bullhorn-icon';
+
+<div style={{ margin: '50px' }}>
+  <Popover
+    aria-label="Alert popover"
+    header={
+      <PopoverHeader>
```
mcoker, line 292:
```suggestion
      <PopoverHeader hasIcon>
```
```suggestion
      <PopoverHeader>
```

packages/react-core/src/components/Popover/examples/Popover.md
```diff
@@ -272,3 +277,85 @@ class AdvancedPopover extends React.Component {
...
+          <option value="warning">warning</option>
+          <option value="danger">danger</option>
+        </select>
+      </div>
+      <div style={{ margin: '50px' }}>
+        <Popover
+          aria-label="Alert popover"
+          alertVariant={alertVariant as 'default' | 'info' | 'warning' | 'success' | 'danger'}
+          header={
+            <PopoverHeader>
```
mcoker, line 346:
```suggestion
            <PopoverHeader hasIcon>
```
```suggestion
            <PopoverHeader>
```

mcoker (changes requested):

packages/react-core/src/components/Popover/PopoverFooter.tsx
```diff
@@ -2,6 +2,13 @@ import * as React from 'react';
 import styles from '@patternfly/react-styles/css/components/Popover/popover';
 import { css } from '@patternfly/react-styles';
 
+export interface PopoverFooterProps extends React.HTMLProps<HTMLDivElement> {
+  /** Additional classes added to the Popover Footer */
```
tlabaj, line 6: nit. sentence case

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -112,6 +112,12 @@ export interface PopoverProps {
   headerContent?: React.ReactNode | ((hide: () => void) => React.ReactNode);
   /** Sets the heading level to use for the popover header. Default is h6. */
   headerComponent?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  /** Composable header component. Use with the PopoverHeader, PopoverHeaderText, and PopoverHeaderIcon components. To manually close
+   * a popover after an action within the bodyContent you can provide a function which will receive a callback as an argument to hide the popover.
+   */
+  header?: React.ReactNode | ((hide: () => void) => React.ReactNode);
+  /** Variants for an alert popover */
+  alertVariant?: 'default' | 'info' | 'warning' | 'success' | 'danger';
```
tlabaj, line 120: So, I think the alertVariant on its own does not make sense.  Setting this prop does not make the popover an alert popover.  I wonder if we can improve on the API?
@jschuler I would be interested in your thoughts.

nicolethoen: @tlabaj how about `alertLevelVariant` or `alertSeverityVariant`

tlabaj: @nicole, those suggestions are good.  Maybe some more context in the description of what it does will also help.

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -112,6 +112,12 @@ export interface PopoverProps {
   headerContent?: React.ReactNode | ((hide: () => void) => React.ReactNode);
   /** Sets the heading level to use for the popover header. Default is h6. */
   headerComponent?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  /** @beta Composable header component. Use with the PopoverHeader, PopoverHeaderText, and PopoverHeaderIcon components. To manually close
```
tlabaj, line 115: Beat me to it.  I was going to ask for beta tag :)

tlabaj (commented):
Can you also mark new exported componets Beta please.

packages/react-core/src/components/Popover/Popover.tsx
jschuler: It would make more sense for me to find a prop like that on the `PopoverHeader` component, since it affects the icon and text color in there. But if we anticipate more changes to the entire Popover it would make sense to be here.

If the prop comment explains that it modifies the header icon and text colors, that would make it clearer what change it does.

wise-king-sullyman (approved):

tlabaj (approved):

packages/react-core/src/components/Popover/PopoverHeader.tsx
```diff
@@ -1,25 +1,24 @@
...
 }: PopoverHeaderProps) => (
-  <PopoverContext.Consumer>
-    {({ headerComponent }) => (
-      <Title headingLevel={headerComponent} size={TitleSizes.md} id={id} {...props}>
-        {children}
-      </Title>
-    )}
-  </PopoverContext.Consumer>
+  <header className={css('pf-c-popover__header', className)} {...props}>
+    <h1 className={css(styles.popoverTitle, hasIcon && styles.modifiers.icon)}>{children}</h1>
```
mcoker, line 21: Should `headerComponent` set the heading level here, too.

nicolethoen: I'm not sure - I've been trying to figure out the real differences between `PopoverHeader` and `PopoverInternalHeader` and one of the main differences is that `PopoverInternalHeader` uses a `Title` component. @mcoker Do you think they should both use `Title` components?

nicolethoen: @kmcfaul is the only reason you built this to be composable because you needed to put the `hasIcon` prop at the outermost level?

mcoker: > @mcoker Do you think they should both use Title components?

@nicolethoen no, I think the popover title is better as an element of the popover component instead of the title component. That way we can change the styling (size, weight, etc) of the title in the CSS as opposed to needing a markup change, which is on the user to make in some cases. When we introduced the new Red Hat font, for example, the change in font face made either the modal or popover title (can't remember which, but both used the `<Title>` component) look different, so we ended up styling `.pf-c-title.pf-m-lg` (or whatever the class name was) to a different font-size in that component, which is a bad practice. Having the title be an element of a component won't put us in a situation like that if the title styles need to change.

mcoker: In our next upcoming breaking change, I'd like to get rid of the `<Title>` component in the popover and make the header and title elements introduced in this PR the default structural elements for the popover's title.

tlabaj: @mcoker, I think you are right,  otherwise the heading level for the composable `PopoverHeader` is not configurable.

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -405,8 +421,13 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
...
         {headerContent && (
-          <PopoverHeader id={`popover-${uniqueId}-header`}>
-            {typeof headerContent === 'function' ? headerContent(hide) : headerContent}
+          <PopoverHeader
+            id={`popover-${uniqueId}-header`}
+            icon={headerIcon}
+            alertSeverityVariant={alertSeverityVariant}
+            titleHeadingLevel={headerComponent}
+          >
+            <React.Fragment>{typeof headerContent === 'function' ? headerContent(hide) : headerContent}</React.Fragment>
```
tlabaj, line 430: Why do we need the `<span class="pf-c-popover__title-text">`

nicolethoen: I think it's a wrapper for the title content in the flex layout which is applied when an icon is present

packages/react-core/src/components/Popover/examples/Popover.md
```diff
@@ -272,3 +277,74 @@ class AdvancedPopover extends React.Component {
   }
 }
 ```
+
+### Popover with icon in the title
+
+```ts isBeta
+import React from 'react';
+import { Popover, PopoverHeader, PopoverHeaderIcon, PopoverHeaderText, Button } from '@patternfly/react-core';
```
tlabaj, line 285: You can remove `PopoverHeaderText`, `PopoverHeaderIcon ` and `PopoverHeader`

packages/react-core/src/components/Popover/examples/Popover.md
```diff
@@ -272,3 +277,74 @@ class AdvancedPopover extends React.Component {
...
+    <Button>Toggle popover</Button>
+  </Popover>
+</div>;
+```
+
+### Alert popover
+
+```ts isBeta
+import React from 'react';
+import { Popover, PopoverHeader, PopoverHeaderIcon, PopoverHeaderText, Button } from '@patternfly/react-core';
```
tlabaj, line 305: You can remove PopoverHeaderText, PopoverHeaderIcon and PopoverHeader

tlabaj (changes requested):

packages/react-core/src/components/Popover/PopoverHeader.tsx
```diff
@@ -1,25 +1,48 @@
...
+  alertSeverityVariant?: 'default' | 'info' | 'warning' | 'success' | 'danger';
+  /** */
+  id?: string;
+}
 
 export const PopoverHeader: React.FunctionComponent<PopoverHeaderProps> = ({
   children,
+  icon,
+  className,
+  titleHeadingLevel = 'h1',
```
mcoker, line 27: I would probably expect this to be `h6` to match the existing `headerComponent` default value.

mcoker (changes requested):

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -405,7 +424,15 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
       <PopoverContent>
         {showClose && <PopoverCloseButton onClose={closePopover} aria-label={closeBtnAriaLabel} />}
         {headerContent && (
-          <PopoverHeader id={`popover-${uniqueId}-header`}>
+          <PopoverHeader
+            id={`popover-${uniqueId}-header`}
+            icon={headerIcon}
+            alertSeverityVariant={alertSeverityVariant}
+            alertSeverityScreenReaderText={
+              alertSeverityVariant ? alertSeverityScreenReaderText || `${alertSeverityVariant} alert:` : undefined
```
tlabaj, line 432: instead of undefined, can we just default it to ``${alertSeverityVariant} alert:``?

tlabaj (approved):
LGTM

mcoker (approved):
🥳

Issue #6669, opened 2021-12-02
Select: isInputValuePersisted not working with in Grouped single with filtering example


seanforyou23:
**Describe the issue. What is the expected and unexpected behavior?**

When I use `isInputValuePersisted` prop in the Grouped single with filtering example, it only seems to work if I select the option element that says `Create "xyz"`  - if I instead blur the field it doesn't retain the value like the other simple select case. Is it expected?

**Please provide the steps to reproduce.**

- Navigate to [Grouped single with filtering example](https://www.patternfly.org/v4/components/select/#grouped-single-with-filtering)
- Add the `isInputValuePersisted` prop to the Select component
- Type into the filter input "asdf"
- Click outside of the bounds of the Select input
- Notice the input value is not persisted

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Not sure if isInputValuePersisted was ever intended to work with the inline filter input. If it was, probably a bug, otherwise it's an enhancement. I did find a workaround is to couple this option with `isCreatable` and as long as the user selects the menu input itself, the rest of the experience is without issue.

**What is your product and what release version are you targeting?**

Migration toolkit for virtualization

![2021-12-02 12 48 31](https://user-images.githubusercontent.com/5942899/144476292-fb507a3d-92f8-4081-be90-7a470a035a60.gif)

nicolethoen:
Closed by: https://github.com/patternfly/patternfly-react/pull/6670

Issue #6300, opened 2021-09-10
Typeahead Select: when using `isInputValuePersisted`, blurring and refocusing the field clears the filter state


mturley:
**Describe the issue. What is the expected and unexpected behavior?**

If I'm using the `isInputValuePersisted` prop on a Typeahead Select, I would expect to be able to close and reopen the dropdown without losing any state (the persisted input value should still be applied as a filter on the options). Currently when using this prop, if you type in the field (see filtered results), click away and click back, you'll see all the options even though you still have a filter typed in.

Expected behavior: closing and reopening the field never changes which options are shown, only typing in the field should be able to change that.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

![ISze7eCtCT](https://user-images.githubusercontent.com/811963/132918198-8a6889ed-b83d-4b8a-9e31-f7227c08b734.gif)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug. Blocking https://bugzilla.redhat.com/show_bug.cgi?id=1992982. You can probably work around it with custom filtering, but I haven't tried that yet.

**What is your product and what release version are you targeting?**

Migration Toolkit for Virtualization. No particular target release.

nicolethoen:
@mturley 
I've been looking at this - and I understand your expectation...
The issue is that the default behavior of the typeahead select, with or without `isInputValuePersisted` is that after you've selected an item, it no longer filters the options - It displays all options with the current selection being marked with a green checkbox.

Given that, i'm not sure how to handle the scenario of  `isInputValuePersisted` since in that case, you've selected something that is not in the dropdown so we cannot provide that feedback, but we still want to show you all the available options in case you want to make a new selection.

I think yours is an example of a place where the typeahead select is trying to behave more like an autocomplete than a filterable select. The intention for the typeahead select is not that it'd be used as an autocomplete...
We do however have a new composable autocomplete demo which is about to be released in the next day or two. You can view it running here.
[https://patternfly-react-pr-6305.surge.sh/components/search-input/react-demos](https://patternfly-react-pr-6305.surge.sh/components/search-input/react-demos)

Is it possible this may more closely map to your use case than a filterable select input?

mturley:
Ah! Yeah, I think autocomplete is closer to our use case here. Good call, I'll take a closer look. Thanks @nicolethoen 🙂

tlabaj:
[...]

PR #6670, opened 2021-12-02
feat(Select): add ability to persist filtered children state

seanforyou23:
**What**: https://github.com/patternfly/patternfly-react/issues/6300

This PR adds the ability to preserve the filtered list of menu children in Select. Feedback welcome!

Ping @mturley @kmcfaul 

**Additional issues**:  https://github.com/patternfly/patternfly-react/issues/6669


mturley (approved):
LGTM! Thanks @seanforyou23

mturley:
@nicolethoen I'm curious about your thoughts here. This originated from the same product issue as https://github.com/patternfly/patternfly-react/issues/6300, and in your comment there (https://github.com/patternfly/patternfly-react/issues/6300#issuecomment-920367065) you were suggesting we use an autocomplete SearchInput instead. I hadn't fully followed up on that, and we decided that doesn't exactly meet the use case here either because we do want to be able to simply click on the field and use it as a Select without typing anything. So I think a prop like this PR introduces makes sense.

The use case here is weird: we want the user to select or create a target namespace for their migration. Many users will only have a few existing ones to choose from, so we don't want to force them to type anything (Select works for that). Some users will have a lot, so the typeahead is good. And the user needs to be able to easily create one (isCreatable), but testers complained when their input was lost if they didn't click the Create option (so isInputValuePersisted makes sense). We then got feedback that it was odd to clear the filtered results if the input wasn't cleared too. This prop seems to complete the puzzle to me (if they want to see all options again they can click the X button).

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -146,6 +146,8 @@ export interface SelectProps
   chipGroupComponent?: React.ReactNode;
   /** Flag for retaining keyboard-entered value in typeahead text field when focus leaves input away */
   isInputValuePersisted?: boolean;
+  /** Flag for retaining filter results on blur from keyboard-entered typeahead text applied with isInputValuePersisted */
```
nicolethoen, line 149: Can you add `@beta` at the beginning of this prop description?

seanforyou23: Sure no problem

nicolethoen:
I think this should work - marking it beta gives us the flexibility to be sure there are not other side effects of this change.
:) thanks for wiring it up!

nicolethoen (approved):

kmcfaul (approved):
LGTM!

tlabaj (approved):

PR #6681, opened 2021-12-08
feat(time picker): added min/max time constraints

### LINKED ISSUE #5853: Add maximum and minimum properties for TimePicker

It would be great if the TimePicker supported maximum and minimum properties that were used for validation as well as generating the range of options presented in the select drop-down.

n/a feature request

enhancement

any

### LINKED ISSUE #6665: TimePicker shows invalid on 12hour lowercase delimiter
When lowercase text character(s) are used as a custom delimiter without the `is24Hour` prop the delimiter is made uppercase and throws an invalid error once the time is interacted with.

1. Visit the [Custom delimiter](https://www.patternfly.org/v4/components/time-picker#custom-delimiter) example, expand the code and remove the `is24Hour` prop.
3. Click in the time picker field, notice the available selections include the lowercase `h` delimiter.
4. Select a time, notice the delimiter changes to uppercase `H` and causes the invalid error to show.

Bug

evwilkin:
**What**: Closes #5853 and closes #6665

This PR:
- adds `minTime`, `maxTime`, and `invalidMinMaxErrorMessage` props which enable restricting the options shown to a user and throwing an error if the user inputs a time outside the constraints.
- adds an example to the time picker showing these new features.
- fixes bug from #6665 where custom delimiter is always changed to uppercase upon time selection/entry when `is24Hour` prop is not present.

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
@@ -30,20 +30,10 @@ export const makeTimeOptions = (
       .toString()
       .padStart(2, '0');
     const timeOption = hour + delimiter + minutes + (hour12 ? suffix : '');
-    // if >=min && <=max, res.push
-    if (validateMinMax(minTime, maxTime, timeOption, hour12, delimiter)) {
-      // console.log({ res });
+    // time option is valid if within min/max constraints
+    if (isWithinMinMax(minTime, maxTime, timeOption, delimiter)) {
       res.push(timeOption);
     }
-    // FUNCTIONALITY MOVED ABOVE
-    // (hour12 ? hour.toString() : hour.toString().padStart(2, '0')) +
-    //   delimiter +
-    //   iter
-    //     .getMinutes()
-    //     .toString()
-    //     .padStart(2, '0') +
-    //   (hour12 ? suffix : '')
-
     iter.setMinutes(iter.getMinutes() + stepMinutes);
   }
   return res;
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
```diff
@@ -111,3 +110,34 @@ export const getMinutes = (time: string, timeRegex: RegExp) => {
...
+    return `${hours}${delimiter}${minutes}`;
+  };
+
+  // correctly format as 24hr times (12:30AM => 00:30, 1:15 => 01:15)
+  minTime = convertTo24Hour(minTime);
+  time = convertTo24Hour(time);
+  maxTime = convertTo24Hour(maxTime);
+
+  // simple string comparison for 24hr times
+  return minTime <= time && time <= maxTime;
```
nicolethoen, line 142: In your example, it looks like the max time is not included in the list. Should it be?

evwilkin: @nicolethoen it shouldn't be included as it's currently designed, but we could change that.  I'm not touching the time options that are created from the `stepMinutes` other than to remove any that are earlier than the min or greater than the max.

In the example, the `stepMinutes` defaults to 30 minute intervals and because the `maxTime` is 17:15 the last time option shown is 17:00 (as the next option, 17:30, is above the maxTime).

@mcarrano would you expect that we should add an additional time option with their max time (or min time) if it doesn't align with one of the existing time options?  In my example, where `maxTime=17:15`, rather than the last option being `17:00` there would be one more that would be `17:15`?

mcarrano:
> @mcarrano would you expect that we should add an additional time option with their max time (or min time) if it doesn't align with one of the existing time options? In my example, where maxTime=17:15, rather than the last option being 17:00 there would be one more that would be 17:15?

No I wouldn't expect this. The user could always enter that from the keyboard. Seems like pretty much an edge case.

nicolethoen (approved):

tlabaj (approved):

PR #6692, opened 2021-12-10
chore(Tabs): Tabs table demo

### LINKED ISSUE #6514: Tables and tabs

gabipodolnikova:
**What**: Closes #6514

With the current configuration the examples needs to be used as `js isFullscreen file="./examples/Tabs/TabsAndTable.tsx"`. I tried it as a ts file, but the docs build is failing.
Also to fully use TS in demos, it needs to be configured, which I was not able to do. In the end, I had to keep the DashboardWrapper and DashboardHeader in JS :(


packages/tsconfig.base.json
```diff
@@ -25,6 +25,7 @@
   "exclude": [
     "**/**.test.tsx",
     "**/**.test.ts",
-    "**/examples/**"
+    "**/examples/**",
+    "**/demos/Tabs/**"
```
nicolethoen, line 29: What purpose does adding this line serve? What error do you see if you didn't add it?

mcarrano (commented):
This looks good @gabipodolnikova . The only question I have is why there are subtle differences in the layout between this and the core demo here: https://www.patternfly.org/v4/components/tabs/html-demos/tables-and-tabs/#  They have nothing to do with the tabs, but seems like the mastheads are different? Not sure which is more correct. Can you take a look @mcoker ?

mcoker:
@mcarrano the react demo is using the masthead component, and the core demo is using the pre-existing page header in the page component. Looks like these are the places that still use the page header and not the masthead in core:

```
src/patternfly//demos/Drawer/drawer-demo-default.hbs
src/patternfly//demos/Page/page-demo-expandable-nav.hbs
src/patternfly//demos/Page/page-template-header.hbs
src/patternfly//demos/Alert/examples/page-default.hbs
src/patternfly//demos/Skeleton/examples/Skeleton.md
src/patternfly//demos/DataList/data-list-page-header.hbs
src/patternfly//demos/Wizard/examples/Wizard.md
src/patternfly//demos/Table/table-page-header.hbs
src/patternfly//demos/PrimaryDetail/primary-detail-template.hbs
src/patternfly//demos/Nav/examples/Nav.md
src/patternfly//demos/NotificationDrawer/notification-drawer-demo.hbs
src/patternfly//demos/CardView/examples/CardView.md
src/patternfly//demos/Banner/__banner-page-header.hbs
```

mcarrano (approved):
@mcoker  I opened a new core issue here to get those updated: https://github.com/patternfly/patternfly/issues/4579. I think this one is good to go then so I will approve it. Thanks @gabipodolnikova !

packages/tsconfig.base.json
gabipodolnikova: oh yeah, nothing now... removing... I moved the TS file to examples and forgot about this

nicolethoen (approved):

wise-king-sullyman (changes requested):
LGTM other than this one nit.

packages/react-core/src/demos/Tabs.md
```diff
@@ -440,4 +445,8 @@ TabsOpenWithSecondaryTabsDemo = () => {
 
 ### Nested tabs
 ```js isFullscreen file="./Tabs/NestedTabsDemo.js"
-```
\ No newline at end of file
+```
+
+### Tables and tabs
+```js isFullscreen file="./examples/Tabs/TabsAndTable.tsx"
```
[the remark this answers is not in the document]
gabipodolnikova: As I said in the PR description, it is not possible with the current config to import it like you suggested... :(

mcoker (changes requested):
The only bug I found is that clicking on either an action column kebab or an item in the kebab dropdown will trigger row selection. 

Here are some other difference between the core and react demos. wdyt of these @mcarrano?

* Core has divider between toolbar and table, react doesn't have divider
* Core has full/non-compact bottom pagination, react uses compact
* Core - drawer panel with content, progress, labels, etc - core using space-items-lg, react uses default spacing (medium)
* Core uses outline variation of labels in drawer panel, react uses default
* Core uses 33% drawer panel width, react uses default
* Core doesn't have breadcrumbs
* Core doesn't use width-limited page sections - react does.

mcarrano:
@mcoker Thanks for highlighting all of the differences between the demos. I knew they looked different but I couldn't put my finger on why.

I'm kind of on the fence here. None of these issues really have bearing on the problem at hand, but I can see people being confused about why these are different. In general I think of the core demos as defining the desired layout/appearance and the React demo just adding interaction. But maybe that's oversimplified.

@nicolethoen @tlabaj @mceledonia @mmenestr what do you think? Do you think it's important to have a direct visual match between equivalent core and React demos?

mmenestr:
Not sure where to look to find both the core and react demos to compare, but l personally think it's important to keep the visuals the same. When designers are looking at them, I don't think they'd think "focus on the interaction for react, and take the visuals from core" even though that's really the aim. They'd probably just think "why are the visuals different in core and react and which one is correct."

In general, the more we can keep things consistent and avoid small differences, the better. That way we can avoid general designer confusion as to how a component should look/be implemented!

mcarrano:
Thanks for this input @mmenestr I agree with your main points.

As I look at this more, I can see that there are both inconsistencies between core and React and also inconsistencies between the core demos themselves. For example, if you look at all the core demos for tabs, some have a breadcrumb and others do not. What would help here? Should we have a common template for all full page demos that includes common elements like the masthead, navigation, breadcrumbs, etc? Could that be used as a starter for all full page demos to make sure these inconsistencies don't creep in? It's true that even though these details seem trivial, they can lead to inconsistent implementations in product.

I welcome your thoughts on how to fix this. As for this specific issue, @gabipodolnikova can you clean up the current demo to address some of the obvious differences (like the toolbar and pagination) and we can address the larger problem as a separate issue, I think?

mcoker:
> Should we have a common template for all full page demos that includes common elements like the masthead, navigation, breadcrumbs, etc? Could that be used as a starter for all full page demos to make sure these inconsistencies don't creep in? It's true that even though these details seem trivial, they can lead to inconsistent implementations in product.

@mcarrano yep, we have templates like this we can use in any component/demo example, and we have page demo templates per component demo (like a common page template demo we use for all table demos, or data list demos, for example). So we should definitely re-visit this and make sure we're using a common template. Would you mind opning a core issue?

mcarrano:
I've opened a new core issue here: https://github.com/patternfly/patternfly/issues/4586  We can talk more about what should be in these templates and how to ensure they get used when creating new demos.

gabipodolnikova:
@mcarrano @mcoker So I covered the mentioned differences between core and react and fixed the issue with the kebab toggle.
There are still some differences, but they are also in the other demos too as we also use a templates (for the Navigation and Masthead)...

mcarrano:
Thanks @gabipodolnikova . Looks good to me!

mcoker (approved):
LGTM!

tlabaj (approved):
Looks great Gabi!

PR #6700, opened 2021-12-14
feat(ToggleGroup): support disabling all items under the parent

### LINKED ISSUE #6359: Add optional disable prop to ToggleGroup component.
**Describe the issue. What is the expected and unexpected behaviour?**
There is no way to disable `<ToggleGroup>` as a whole. There should be a prop like `disable` for ToggleGroup to do it.

Enhancement 
Currently to disable the ToggleGroup I had to disable the ToggleGroupItem individually.
e.g
https://codesandbox.io/s/hy5kv?file=/index.js

- @patternfly/react-core@4.115.2

boaz0:
**What**: Closes #6359

### DIFF
--- a/packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
+++ b/packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
@@ -10,7 +10,7 @@ export interface ToggleGroupProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Modifies the toggle group to include compact styling. */
   isCompact?: boolean;
-  /** Disable all ToggleGroup items under this component. */
+  /** Disable all toggle group items under this component. */
   disableAll?: boolean;
   /** Accessible label for the toggle group */
   'aria-label'?: string;
--- a/packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
+++ b/packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
@@ -10,7 +10,7 @@ export interface ToggleGroupProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Modifies the toggle group to include compact styling. */
   isCompact?: boolean;
-  /** Disable all ToggleGroupItems under this component. */
+  /** Disable all ToggleGroup items under this component. */
   disableAll?: boolean;
   /** Accessible label for the toggle group */
   'aria-label'?: string;
@@ -24,14 +24,11 @@ export const ToggleGroup: React.FunctionComponent<ToggleGroupProps> = ({
   'aria-label': ariaLabel,
   ...props
 }: ToggleGroupProps) => {
-  const toggleGroupItemList = [] as any[];
-  React.Children.forEach(children, child => {
+  const toggleGroupItemList = React.Children.map(children, child => {
     const childCompName = (child as any).type.name;
-    const newChild =
-      childCompName !== ToggleGroupItem.name
-        ? child
-        : React.cloneElement(child as React.ReactElement, disableAll ? { isDisabled: true } : {});
-    toggleGroupItemList.push(newChild);
+    return childCompName !== ToggleGroupItem.name
+      ? child
+      : React.cloneElement(child as React.ReactElement, disableAll ? { isDisabled: true } : {});
   });
 
   return (
--- a/packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
+++ b/packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
@@ -10,6 +10,7 @@ export interface ToggleGroupProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Modifies the toggle group to include compact styling. */
   isCompact?: boolean;
+  /** Disable all ToggleGroupItems under this component. */
   disableAll?: boolean;
   /** Accessible label for the toggle group */
   'aria-label'?: string;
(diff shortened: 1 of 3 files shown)


mcoker (approved):
L👍TM!

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
```diff
@@ -9,6 +10,7 @@ export interface ToggleGroupProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Modifies the toggle group to include compact styling. */
   isCompact?: boolean;
+  disableAll?: boolean;
```
nicolethoen, line 13: Could you add a prop description here?

nicolethoen (changes requested):

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
boaz0: Oh snap! Sorry I forgot. Sure I will do it now.

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
```diff
@@ -17,12 +20,18 @@ export const ToggleGroup: React.FunctionComponent<ToggleGroupProps> = ({
...
 }: ToggleGroupProps) => {
   const toggleGroupItemList = [] as any[];
   React.Children.forEach(children, child => {
-    toggleGroupItemList.push(child);
+    const childCompName = (child as any).type.name;
+    const newChild =
+      childCompName !== ToggleGroupItem.name
+        ? child
+        : React.cloneElement(child as React.ReactElement, disableAll ? { isDisabled: true } : {});
+    toggleGroupItemList.push(newChild);
```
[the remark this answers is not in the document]
boaz0: I totally agree. 👍

nicolethoen (approved):
LGTM! Thank you!!

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
```diff
@@ -9,6 +10,8 @@ export interface ToggleGroupProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Modifies the toggle group to include compact styling. */
   isCompact?: boolean;
+  /** Disable all ToggleGroup items under this component. */
```
wise-king-sullyman, line 13: One last nit, is there a reason you kept `ToggleGroup` rather than adding the space and changing capitalization to `toggle group`? My understanding of [the capitalization guidelines](https://www.patternfly.org/v4/ux-writing/capitalization/#capitalization-types) is that components named in text that gets exposed in the docs should only be capitalized when starting a sentence.

boaz0: Thanks for referencing me to the guidelines. I didn't know it mentions this. I will update this line.
Appreciate the feedback.

wise-king-sullyman (approved):

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
```diff
@@ -9,6 +10,8 @@ export interface ToggleGroupProps extends React.HTMLProps<HTMLDivElement> {
   className?: string;
   /** Modifies the toggle group to include compact styling. */
   isCompact?: boolean;
+  /** Disable all toggle group items under this component. */
+  disableAll?: boolean;
```
tlabaj, line 14: the naming doe not quite match our usual conventions.  maybe `areAllDisabled` or `areAllGroupsDisabled`

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
```diff
@@ -17,12 +20,15 @@ export const ToggleGroup: React.FunctionComponent<ToggleGroupProps> = ({
...
   ...props
 }: ToggleGroupProps) => {
-  const toggleGroupItemList = [] as any[];
-  React.Children.forEach(children, child => {
-    toggleGroupItemList.push(child);
+  const toggleGroupItemList = React.Children.map(children, child => {
+    const childCompName = (child as any).type.name;
+    return childCompName !== ToggleGroupItem.name
+      ? child
+      : React.cloneElement(child as React.ReactElement, disableAll ? { isDisabled: true } : {});
```
tlabaj, line 31: We try and avoid cloneElement when we can.  Can we do this with a render prop instead?

tlabaj (changes requested):

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
boaz0: In the current state, I don't think so. The closest we can do is to create ToggleGroupContext which will pass this to other ToggleGroupItems through its context.
What do you think?

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
boaz0: Yep, this is much better!

packages/react-core/src/components/ToggleGroup/ToggleGroup.tsx
tlabaj: Hmmm.   That might be a better option @boaz0 consumers have asked we avoid cloneElement as it can be problematic.

boaz0: So you want me to do go with the context idea and revert the current changes I've made?

kmcfaul: I think either method works in this case because we're type checking the child before cloning

tlabaj: @boaz0 I think your approach is fine for this use case.

tlabaj (approved):

boaz0:
Thank you all! @nicolethoen @wise-king-sullyman @tlabaj @mcoker @kmcfaul

PR #6704, opened 2021-12-15
chore(TextInputGroup): convert examples and demos to typescript=

### LINKED ISSUE #6679: TextInputGroup: Convert examples/demos from JS to TS
The examples and demos for the TextInputGroup need to be converted to TypeScript.

Related to #6594

wise-king-sullyman:
**What**: Closes #6679

**Additional issues**:

### DIFF
--- a/packages/react-core/src/demos/TextInputGroupDemo.md
+++ b/packages/react-core/src/demos/TextInputGroupDemo.md
@@ -24,7 +24,7 @@ Additionally, attributes can be selected by typing the full (case sensitive) nam
 
 Attributes can be deselected (returning you to attribute selection mode) by hitting `escape`, or by hitting `backspace` when the only text in the text input is the attribute.
 
-```js file="./examples/TextInputGroup/AttributeValueFiltering.js"
+```js file="./examples/TextInputGroup/AttributeValueFiltering.tsx"
 ```
 ### Auto-complete search
 
@@ -36,5 +36,5 @@ Hitting `escape` while focused on the input or menu will close the menu, and the
 
 When only one item remains in the suggestion list, tab can be used to auto-complete the typing of that item.
 
-```js file="./examples/TextInputGroup/AutoCompleteSearch.js"
+```js file="./examples/TextInputGroup/AutoCompleteSearch.tsx"
 ```
--- a/packages/react-core/src/demos/TextInputGroupDemo.md
+++ b/packages/react-core/src/demos/TextInputGroupDemo.md
@@ -24,7 +24,7 @@ Additionally, attributes can be selected by typing the full (case sensitive) nam
 
 Attributes can be deselected (returning you to attribute selection mode) by hitting `escape`, or by hitting `backspace` when the only text in the text input is the attribute.
 
-```js file="./examples/TextInputGroup/AttributeValueFiltering.tsx"
+```ts file="./examples/TextInputGroup/AttributeValueFiltering.tsx"
 ```
 ### Auto-complete search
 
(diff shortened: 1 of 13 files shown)


packages/react-core/src/demos/TextInputGroupDemo.md
```diff
@@ -24,7 +24,7 @@ Additionally, attributes can be selected by typing the full (case sensitive) nam
 
 Attributes can be deselected (returning you to attribute selection mode) by hitting `escape`, or by hitting `backspace` when the only text in the text input is the attribute.
 
-```js file="./examples/TextInputGroup/AttributeValueFiltering.js"
+```js file="./examples/TextInputGroup/AttributeValueFiltering.tsx"
```
kmcfaul, line 27: `ts file` I think

packages/react-core/src/demos/TextInputGroupDemo.md
```diff
@@ -36,5 +36,5 @@ Hitting `escape` while focused on the input or menu will close the menu, and the
 
 When only one item remains in the suggestion list, tab can be used to auto-complete the typing of that item.
 
-```js file="./examples/TextInputGroup/AutoCompleteSearch.js"
+```js file="./examples/TextInputGroup/AutoCompleteSearch.tsx"
```
kmcfaul, line 39: `ts file` here too

kmcfaul (changes requested):

packages/react-core/src/demos/TextInputGroupDemo.md
nicolethoen: hm... it may work both ways? aren't .ts files just for files with type declarations? I think .tsx could be correct? Now I'm second guessing everything lol

packages/react-core/src/demos/examples/TextInputGroup/AttributeValueFiltering.tsx
```diff
@@ -191,11 +195,11 @@ export const AttributeValueFiltering = () => {
   };
 
   /** close the menu when a click occurs outside of the menu or text input group */
-  const handleClick = event => {
+  const handleClick = (event: MouseEvent) => {
     if (
       menuRef.current &&
-      !menuRef.current.contains(event.target) &&
-      !textInputGroupRef.current.contains(event.target)
+      !menuRef.current.contains(event.target as Node) &&
```
nicolethoen, line 201: I'm not used to seeing this type. Does it work if it's `React.ReactNode`? Maybe that's what I'm more used to seeing?

nicolethoen: It may also be `HTMLElement`? It may not be wrong either - I think we just tend to be a little more specific...

wise-king-sullyman: TypeScript complains about `React.ReactNode`, but `HTMLElement` does work since it's an extension of an extension of `Node`. Pushing this change shortly.

nicolethoen (approved):
LGTM

wise-king-sullyman:
After the conversation about these conversions yesterday I went to update all of the examples and demos in this component and in doing so ran into an issue with eslint when it comes to the demos.

For the demos specifically, eslint throws an error for the no-anonymous-functions rule. It's specifically for the demos because an override was previously added which disables that rule in examples folders, but the demos folder is structured differently.

The clearest solution to this issue in my opinion is to extend the eslint override to ignore files in the demo folder, so that's what I will do unless anyone objects.

Other approaches would be to `disable-next-line` in the files, or to add display names. Because the last declaration is what's going into the markdown though this would require default exporting the component at the bottom of the file which is a bit of a departure from our current practices.

thatblindgeye:
As was discussed, and in regards to consistency that was mentioned during Wednesday's stand up, I'm thinking the file names for these separated examples should follow a consistent format such as `ComponentName[exampleName]`.

So instead of `BasicTextInputGroup`, we'd have `TextInputGroupBasic` and so on.

thatblindgeye (approved):
Looks good in terms of the updated example/file names!

nicolethoen (approved):

kmcfaul (approved):
lgtm!

PR #6708, opened 2021-12-20
fix(ClipboardCopy): clear timer on unmount

### LINKED ISSUE #6649: "React state update on an unmounted component" message being triggered after copying text
It happens here:
https://github.com/patternfly/patternfly-react/blob/d026d3fbbcdaa8438312b71ebe59a2f232103b4d/packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx#L197

I have this modal that renders the ClipboardCopy component. Right after copying the text if one closes the modal immediately, you'll get the warning in the browser's console saying: `Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method`

Sample consumer code:
https://github.com/openshift-assisted/assisted-ui-lib/blob/a3b9ca6a5b1895b5e3daf15b1c492a955ba6f7f4/src/common/components/clusterConfiguration/DownloadIso.tsx#L48

boaz0:
**What**: Closes #6649

### DIFF
--- a/packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
+++ b/packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
@@ -111,6 +111,12 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
     }
   };
 
+  componentWillUnmount = () => {
+    if (this.timer) {
+      window.clearTimeout(this.timer);
+    }
+  };
+
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   expandContent = (_event: React.MouseEvent<Element, MouseEvent>) => {
     this.setState(prevState => ({


packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx
```diff
@@ -111,6 +111,12 @@ export class ClipboardCopy extends React.Component<ClipboardCopyProps, Clipboard
     }
   };
 
+  componentWillUnmount = () => {
+    if (this.timer) {
+      window.clearTimeout(this.timer);
```
nicolethoen, line 116: Great catch! 
Does this mean that the `this.timer = null;` on line 204 and 357 is unnecessary?

boaz0: @nicolethoen that's a great question.
These lines were meant for a different case:
1) If the user clicked on the copy button we set a timer that on timeout we will close the "copied" popover
2) but what happens if the user clicks again on the copy button and the timer didn't timeout. A new timer will be created unnecessarily and we will have a flaky behavior in the popover display (it will be closed after `switchDelay` from the first click).
3) So what we did is check if there is a timer. Stop it and release it from the memory. and run a new timer.

In other words, these lines are necessary for a behavior of the popover and not covering the case where the component is unmounted and after the timeout we are changing the component state (we "cannot" change its state because it's not mounted!).

I hope I explained that well.
Thank you for the question - it was fun. 😄

nicolethoen: Fabulous explanation! Thank you so much!

nicolethoen (approved):

wise-king-sullyman (approved):

tlabaj (approved):
LGTM

PR #6712, opened 2021-12-22
fix(Pagination): Allow set plural title in Navigation

### LINKED ISSUE #6707: bug(Pagination): pluralize function only works for English
I'm using the pagination component and our project replaces English strings with their translation. The pagination adds "s" to the word 'page' and that only works in english (using the pluralize function) this function is only used in Navigation.tsx .
https://github.com/patternfly/patternfly-react/blob/23202e2b45e7538d5c0e6516c8b3f65431968a0a/packages/react-core/src/helpers/util.ts#L230
Can this function be removed and just use 'page' for 1 item and 'pages' for multiple? 
Otherwise we get weird results for other langues like:
Screenshot from 2021-12-17 12-02-23

boaz0:
**What**: Closes #6707 

//cc @Ron-Lavi @MariaAga @tlabaj


tlabaj (approved):
LGTM

PR #6716, opened 2021-12-22
chore(Accordion): convert examples to typescript/functional components

### LINKED ISSUE #6715: Accordion: Convert examples to TypeScript
The examples for the Accordion component need to be converted to TypeScript.

Related to #6594

wise-king-sullyman:
**What**: Closes #6715

**Additional issues**:


packages/react-core/src/components/Accordion/examples/AccordionDefinitionList.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import { Accordion, AccordionItem, AccordionContent, AccordionToggle } from '@patternfly/react-core';
+
+export const AccordionDefinitionList = () => {
+  const [expanded, setExpanded] = React.useState('ex-toggle2');
+
+  const onToggle = (id: string) => {
```
jeffpuzzo, line 7: This might turn into a broader discussion, but we might want to consider optimizing `examples` as they're converted to typescript.

Using this as an example, we could consider memoizing this function and including the anonymous callback used on line 19 as well.

@nicolethoen has added this as a topic to discuss in our next retro.

wise-king-sullyman: I'll save most of my thoughts for the discussion tomorrow, but overall I worry that would be jamming too many different things into one PR, unless I drop the functional conversion or extraction that I'm currently doing with these TS conversions.

nicolethoen (commented):
As per today's discussion, let's make sure each functional component has an explicit return type.

jeffpuzzo (approved):

thatblindgeye (approved):
In terms of the updated example/file names, looks good!

jeffpuzzo:
Suggestion for naming convention/directory structure:

Accordion/
  Accordion.md
  Accordion.tsx
  examples/
    Bordered.tsx
    DefinitionList.tsx

What I mentioned in standup doesn't look like it's necessary, but my thought was that if these needed to be used in other places within this codebase, you could add a `export * from "./examples"` to Accordion/index.ts, and if this was a use-case, you could; `import * as AccordionExamples from "components/Accordion/examples"`.

wise-king-sullyman:
@jeffpuzzo a key part of why we thought including the component and example names would be useful is that it allows instant finding/identification of exactly what example is being worked on based on the file name alone.

Additionally, it's a pattern that was already somewhat established for components, i.e. we have `Accordion`,  `AccordionContent`, `AccordionContext`, `AccordionExpandedContentBody`, etc.

That being said I wouldn't be opposed to that direction, as I recognize component/filename length could become an issue. I just wanted to clarify the reasoning that we used.

nicolethoen (approved):

tlabaj (approved):

PR #6718, opened 2021-12-22
chore(ActionList): convert examples to typescript/functional components

### LINKED ISSUE #6717: ActionList: Convert  examples to TypeScript
The examples for the ActionList component need to be converted to TypeScript.

Related to #6594

wise-king-sullyman:
**What**: Closes #6717 

**Additional issues**:


nicolethoen (approved):
LGTM

I will note that if all you're doing is declaring your examples as functions like:
`export const ActionListMultipleGroups = () => (...`
you might not even need to do that. I think the examples will also still work without the exported const. i.e.
```
<ActionList>
    <ActionListGroup>
      <ActionListItem>
        ...
      </ActionListItem>
    </ActionListGroup>
  </ActionList>
```
Some of the simple examples do that - like Alerts

But that may be a person preference thing. just an FYI if it interests you.

nicolethoen:
Another note: these examples are all valid typescript, but all your function examples don't explicitly declare the return type to be a FunctionComponent. That doesn't matter since no one is importing and using them. In this case typescript does not complain because it can infer the type. But it may be a good habit to be explicit about the return type? Or at least be aware that in most other cases you need to be explicit - not necessarily in these cases. 

I think both of the points I bring up here are small. I'll make note to chat about that at the retro so that we can have like some team agreement about style and such as we continue this conversion effort.

wise-king-sullyman:
@nicolethoen those concerns actually are related to a potential small issue I've seen. The markdown parser only identifies typescript examples as typescript when it actually has to convert the code into regular js because of explicit typing.
image

Your suggestion of explicitly declaring the return type for all components would make this a non-issue, but your suggestion of just having the plain JSX would contribute to the issue.

An alternative solution of course would be to edit the md parser to identify anything as TS if it's specified as such. I'm currently poking around in the MD parsing code to try and figure out what it would take to do so.

nicolethoen:
As per today's discussion, let's make sure each functional component has an explicit return type.

nicolethoen (approved):

jeffpuzzo (approved):

thatblindgeye (approved):

PR #6724, opened 2022-01-02
fix(Table): use colgroup and nested cols in composable table

### LINKED ISSUE #6642: Composable Table: nested headers example has a DOM nesting validation warning
Found when converting the new examples from https://github.com/patternfly/patternfly-react/pull/6538 to TypeScript for https://github.com/patternfly/patternfly-react/pull/6168.

Screen Shot 2021-11-23 at 5 16 55 PM

Will investigate as part of converting the example, but it might need fixing in a followup PR.

boaz0:
**What**: Closes #6642

### DIFF
--- a/packages/react-table/src/components/TableComposable/TableComposable.tsx
+++ b/packages/react-table/src/components/TableComposable/TableComposable.tsx
@@ -53,7 +53,7 @@ export interface TableComposableProps extends React.HTMLProps<HTMLTableElement>,
   isTreeTable?: boolean;
   /** Flag indicating this table is nested within another table */
   isNested?: boolean;
-  /** Collection of column spans for nested headers */
+  /** Collection of column spans for nested headers. Deprecated: see https://github.com/patternfly/patternfly/issues/4584 */
   nestedHeaderColumnSpans?: number[];
 }
 
@@ -71,7 +71,6 @@ const TableComposableBase: React.FunctionComponent<TableComposableProps> = ({
   ouiaSafe = true,
   isTreeTable = false,
   isNested = false,
-  nestedHeaderColumnSpans,
   ...props
 }: TableComposableProps) => {
   const tableRef = innerRef || React.useRef(null);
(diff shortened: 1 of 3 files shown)


nicolethoen (commented):
I'm noticing how the HTML mark up in the core examples for defining `colgroup` and `col` still differs from how it's written in the react.
see this example: https://www.patternfly.org/v4/components/table/html/#nested-column-headers

@mattnolting do you know if there is a better way to do that? Should the span prop go on a colgroup or a col?

mattnolting:
After discussing next steps w/PF devs, looks like we can remove the `colgroup` and `col` references altogether. [Core followup issue created](https://github.com/patternfly/patternfly/issues/4584).

nicolethoen:
okay - after the core issue is addressed, this PR can be updated to match.

boaz0:
Sounds good to me!

nicolethoen:
@boaz0 The core PR was merged! :)
Here was the core issue: https://github.com/patternfly/patternfly/issues/4584
Here is the core PR: https://github.com/patternfly/patternfly/pull/4600

So this PR can now be updated to match these changes!

mattnolting:
Hey @boaz0, thanks for tackling this issue! We've decided to remove `col/colgroup`s altogether, so feel free to make those changes.

boaz0:
CI failure is not related to my changes.

nicolethoen (approved):

mattnolting (approved):
LGTM👍

packages/react-table/src/components/TableComposable/TableComposable.tsx
```diff
@@ -71,7 +71,6 @@ const TableComposableBase: React.FunctionComponent<TableComposableProps> = ({
   ouiaSafe = true,
   isTreeTable = false,
   isNested = false,
-  nestedHeaderColumnSpans,
```
tlabaj, line 74: I believe you still need this so it does not get spread to the table and cause errors.

tlabaj (changes requested):

tlabaj:
@mattnolting So, 'nestedHeaderColumnSpans`. and removing it has no impact on consumers correct?

mattnolting:
> @mattnolting So, 'nestedHeaderColumnSpans`. and removing it has no impact on consumers correct?

If `nestedHeaderColumnSpans` only generates `colgroup/col` markup, then no, it has no impact upon consumers. If it's a shared property, it would affect consumers.

kmcfaul (approved):
LGTM. `nestedHeaderColumnSpans` is only used to generate the `col/colgroup` setup so it should be fine to deprecate.

nicolethoen (approved):

tlabaj (approved):
LGTM

Issue #6699, opened 2021-12-14
Table: add example for nested table inside an expansion


mcarrano:
This core PR (https://github.com/patternfly/patternfly/pull/4529) added a new example to show how to nest a table inside of an expanded row in a parent table. This is a pretty common use case and therefore we should add this example on the React side, also.

PR #6727, opened 2022-01-03
docs(Table): add table nested in a table example

MariaAga:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/6699


nicolethoen (changes requested):
This is great!
Ideally, visually, the example will match this html example: https://www.patternfly.org/v4/components/table/html#nested-column-headers-and-expandable-rows

MariaAga:
I started making the changes and I feel like adding all the features in that example makes the react example very complicated and draws attention from the main feature that we show, a nested table, should I instead make the data the same as the html example?

mcarrano:
@MariaAga @nicolethoen I actually thought this was the example we are trying to recreate in React. It's placing a compact table inside an expansion panel.

mcarrano:
Here is the link associated with the above comment: https://www.patternfly.org/v4/components/table/html#expandable-with-nested-table-example

nicolethoen:
@MariaAga I'm sorry I linked the wrong example!
And I agree that it's probably over kill for a documentation example to implement all the features. 
Maybe focus on making the nested table compact? Maybe just focus on the nested table itself matching the HTML example?

MariaAga:
Changed the example to fit the html one, is it similar enough?

mcarrano (changes requested):
@MariaAga I think that the new example is fine. But I'm noticing that the blue highlight border for the expandable parent row does not show when open like for the core and other expandable examples.

MariaAga:
Thanks! added the blue line

mcarrano (approved):
Looks good. Thanks @MariaAga !

nicolethoen (approved):

packages/react-table/src/components/TableComposable/examples/ComposableTableNestedTableExpandable.tsx
```diff
@@ -0,0 +1,189 @@
...
+      title: 'Third action',
+      onClick: () => console.log(`clicked on Third action, on row ${repo.name}`)
+    }
+  ];
+
+  return (
+    <TableComposable aria-label="Simple table">
+      <Thead>
+        <Tr>
+          <Th />
```
mattnolting, line 144: Empty `th` cells should be `td`s

```suggestion
          <Td />
```

mattnolting (changes requested):
Thanks @MariaAga, really great work! Just found one thing.

packages/react-table/src/components/TableComposable/examples/ComposableTableNestedTableExpandable.tsx
MariaAga: Thanks! fixed

mattnolting (approved):
Looks great, thank you!

tlabaj (approved):
LGTM

Issue #6682, opened 2021-12-09
Replace text input / select filters in Toolbar demos with SearchInput


nicolethoen:
follow up related to this comment: https://github.com/patternfly/patternfly-react/issues/5961#issuecomment-896834219

mcarrano:
@MariaAga I wanted to follow up with you on this issue. We've had more time to discuss this and did an audit of various components and demos to ensure we were presenting search inputs in a consistent way everywhere. Based on that we opened a new issue here: https://github.com/patternfly/patternfly-react/issues/6792 with a spreadsheet that calls out changes to be made across the library.

Essentially our recommendation is to use the [basic search input ](https://www.patternfly.org/v4/components/search-input#basic)as the default choice for toolbar examples and demos. If you want to continue with this issue, it would be great if you can update your existing PR based on that. If not, then we will close this out and schedule follow-up work against the newer issue. Let us know your thoughts.

cc @nicolethoen

PR #6728, opened 2022-01-04
refactor(Toolbar-docs): use SearchInput in Toolbar demos

MariaAga:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/6682

### DIFF
--- a/packages/react-core/src/components/InputGroup/examples/InputGroup.md
+++ b/packages/react-core/src/components/InputGroup/examples/InputGroup.md
@@ -5,7 +5,6 @@ cssPrefix: null
 propComponents: ['InputGroup', 'InputGroupText']
 ---
 import AtIcon from '@patternfly/react-icons/dist/esm/icons/at-icon';
-import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import DollarSignIcon from '@patternfly/react-icons/dist/esm/icons/dollar-sign-icon';
 import CalendarAltIcon from '@patternfly/react-icons/dist/esm/icons/calendar-alt-icon';
 import QuestionCircleIcon from '@patternfly/react-icons/dist/esm/icons/question-circle-icon';
@@ -28,7 +27,6 @@ import {
 ```js
 import React from 'react';
 import AtIcon from '@patternfly/react-icons/dist/esm/icons/at-icon';
-import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import {
   Button,
   InputGroup,
@@ -56,13 +54,6 @@ class SimpleInputGroups extends React.Component {
             <AtIcon />
           </InputGroupText>
           <TextInput validated={ValidatedOptions.error} id="textInput7" type="email" aria-label="Error state username example" />
-        </InputGroup>
-        <br />
-        <InputGroup>
-          <TextInput name="textInput11" id="textInput11" type="search" aria-label="search input example" />
-          <Button variant="control" aria-label="search button for search input">
-            <SearchIcon />
-          </Button>
         </InputGroup>
         <br />
         <InputGroup>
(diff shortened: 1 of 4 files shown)


jeffpuzzo:
This looks good to me. There are 2 small discrepancies with "with filters" example I'll note here, but I don't consider them blocking:
Original: 
image
image

New:
image
image

jeffpuzzo (approved):

MariaAga:
I see the original the same for with filters, not sure where to see the second screenshot with the long `risk` and `status` selects.
Original screenshot: 
Screenshot from 2022-01-04 19-49-38

jeffpuzzo:
> I see the original the same for with filters, not sure where to see the second screenshot with the long `risk` and `status` selects. Original screenshot: Screenshot from 2022-01-04 19-49-38

@MariaAga Sorry, I should have specified, that view is possible when the window's width is condensed to < 1200px. At this point, the filter icon is presented, and clicking on that will yield what's seen in those other screenshots.

MariaAga:
I don't see some of the components when using `< 1200px`, is that a known issue?
My view for https://www.patternfly.org/v4/components/toolbar/
Screenshot from 2022-01-04 19-59-17

jeffpuzzo:
> I don't see some of the components when using `< 1200px`, is that a known issue? My view for https://www.patternfly.org/v4/components/toolbar/ Screenshot from 2022-01-04 19-59-17

What you see is correct. The last step is to click this icon here, then you should see those inputs:
image

nicolethoen (changes requested):
Could you also update the text input to a search input in the following files:

https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/demos/Toolbar.md

the Filterable Table demo found here:
https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/demos/Table.md#filterable

MariaAga:
Fixed the visual discrepancies. Also changed in https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/demos/Table.md#filterable.
I didn't text input in https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/demos/Toolbar.md

nicolethoen:
> Fixed the visual discrepancies. Also changed in https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/demos/Table.md#filterable. I didn't text input in https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/demos/Toolbar.md

Thanks! You're right! Someone already went in and updated the Toolbar demo to use a Search Input

tlabaj (approved):
LGTM

nicolethoen (approved):

mcarrano:
After discussion with @mmenestr I'd like to hold on merging this for now. We realized that there are lot's of similar examples across the library (including on the core side) and making this change will just introduce more inconsistency. We will look at the bigger picture of this and make some recommendation as to what examples we want to expose in general for demonstrating toolbar capabilities. Thank you for your efforts on this to date @MariaAga . I don't think it's to waste, but I want to consider the larger problem before we move forward.

MariaAga:
Switched the input in some more components, but I'm not sure what do switch in `Masthead` and `Context Selector`

mcarrano:
@MariaAga I'm not sure what needs to change for the Masthead, either. I think we can leave the Context Selector as is since that search is baked into the component. @mmenestr can you review when you are back from PTO?

mmenestr:
The masthead would be changing the search inside of the context selector dropdown, but if you say it's baked in, i'm not sure it applies to either.

In general, it looks good! but @mcarrano I think we had discussed changing all the searches to use the "search input" component without the arrow button. So we might want to change that in this PR as well - but just want to confirm with you?

MariaAga:
Thanks for the review!
Changed the search to be the basic variant in `Toolbar` and `DataList` but in `Table` there is a functional search bar in the original demo, so I left the arrow in that demo:
http://localhost:8002/components/table/react-demos#filterable
Screenshot from 2022-02-04 13-06-37

mmenestr:
@MariaAga thank you for making the changes! All looks good to me, and I'm fine with leaving the arrow in that example, since as you say, it doesn't search until the user submits the search :) 

I did notice a random bug I wanted to flag while I was testing it out, which is that when you get the empty state about no search results, there's a floating checkbox in the left side of the empty state that shouldn't be there haha not sure if that's something we can fix in this PR, or if we'd rather open a new PR - but thought I'd flag.

image

mmenestr:
Sure, if you could open a new PR that would be great, and we can address it there! Thank you :)

MariaAga:
Opened https://github.com/patternfly/patternfly-react/pull/6898 for it

mmenestr (approved):

Issue #6655, opened 2021-11-30
Toolbar: Add demo with sticky toolbar


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Add a demo to show a use case for the sticky toolbar.  

cc @mcarrano

PR #6729, opened 2022-01-04
docs(Toolbar):  Add demo with sticky toolbar

MariaAga:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/6655

### DIFF
--- a/packages/react-core/src/components/Toolbar/examples/Toolbar.md
+++ b/packages/react-core/src/components/Toolbar/examples/Toolbar.md
@@ -239,6 +239,70 @@ class ToolbarSpacers extends React.Component {
 }
 ```
 
+### Sticky
+
+```js
+import React from 'react';
+import {
+  Toolbar,
+  ToolbarItem,
+  ToolbarContent,
+  Button,
+  ButtonVariant,
+  InputGroup,
+  TextInput,
+  Checkbox
+} from '@patternfly/react-core';
+import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
+
+const ToolbarItems = () => {
+  const [isSticky, setIsSticky] = React.useState(true);
+  const [showEvenOnly, setShowEvenOnly] = React.useState(true);
+  const array = [...Array(30).keys()];
+  const numbers = showEvenOnly ?  array.filter(number => number % 2 === 0):array;
+
+  return (
+    <React.Fragment>
+      <div style={{ overflowY: 'scroll', height: '200px' }}>
+        <Toolbar id="toolbar-spacers" inset={{ default: 'insetNone' }} isSticky={isSticky}>
+          <ToolbarContent>
+            <ToolbarItem>
+              <InputGroup>
+                <TextInput name="textInput1" id="textInput1" type="search" aria-label="search input example" />
+                <Button variant={ButtonVariant.control} aria-label="search button for search input">
+                  <SearchIcon />
+                </Button>
+              </InputGroup>
+            </ToolbarItem>
+            <ToolbarItem>
+              <Checkbox
+                label="Show only even number items"
+                isChecked={showEvenOnly}
+                onChange={setShowEvenOnly}
+                aria-label="checkbox for showing only even numbers"
+                id="showOnlyEvenCheckbox"
+              />
+            </ToolbarItem>
+          </ToolbarContent>
+        </Toolbar>
+        <ul>
+          {numbers.map(number => (
+            <li key={number}>{`item ${number}`}</li>
+          ))}
+        </ul>
+      </div>
+      <Checkbox
+        label="Is toolbar sticky"
+        isChecked={isSticky}
+        onChange={setIsSticky}
+        aria-label="checkbox for is sticky"
+        id="isStickyCheckbox"
+      />
+    </React.Fragment>
+  );
+};
+```
+
 ### Groups
 
 Often, it makes sense to group sets of like items to create desired associations and to enable items to respond together to changes in viewport width. (Note: This example does not demonstrate the desired responsive behavior of the toolbar. That is handled in later examples.)


packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -239,6 +239,71 @@ class ToolbarSpacers extends React.Component {
...
+        isChecked={isSticky}
+        onChange={setIsSticky}
+        aria-label="checkbox for is sticky"
+        id="isStickyCheckbox"
+      />
+      <div style={{ overflowY: 'scroll', height: '200px' }}>
+        <Toolbar
+          id="toolbar-spacers"
+          inset={{
+            default: 'insetNone',
```
nicolethoen, line 267: i think the example looks best when the value of the `inset` prop is:
```
inset={{default: 'insetNone'}}
```
What do you think?

nicolethoen:
There still may be need for a full page demo - under the react demos tab. But this is a simple and clear example.
@mcarrano what do you think?

mcarrano (approved):
I'm good with this. @nicolethoen I agree that ultimately we should have a full page demo. Thanks @MariaAga !

packages/react-core/src/components/Toolbar/examples/Toolbar.md
MariaAga: I agree, changed!

nicolethoen:
@mcarrano should we at least have a mock up made for a full page demo with a sticky toolbar in context? Could you open a design issue to track?

packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -239,6 +239,66 @@ class ToolbarSpacers extends React.Component {
 }
 ```
 
+### Sticky
+
+```js
+import React from 'react';
+import { Toolbar, ToolbarItem, ToolbarContent } from '@patternfly/react-core';
+import { Button, ButtonVariant, InputGroup, TextInput, Checkbox } from '@patternfly/react-core';
```
wise-king-sullyman, line 247: I see this is on the other examples here as well, but does anyone know if there's a particular reason we're splitting these imports like this?

packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -239,6 +239,66 @@ class ToolbarSpacers extends React.Component {
...
+                onChange={setShowEvenOnly}
+                aria-label="checkbox for showing only even numbers"
+                id="showOnlyEvenCheckbox"
+              />
+            </ToolbarItem>
+          </ToolbarContent>
+        </Toolbar>
+        <ul>
+          {[...Array(25).keys()]
+            .filter(number => (showEvenOnly ? number % 2 === 0 : number || number === 0))
```
wise-king-sullyman, line 291: I think this would be cleaner if we moved the array creation and filtering out of the jsx, and conditionally ran the filter rather than running the filter regardless of if `showEvenOnly` is true.

mcarrano:
> @mcarrano should we at least have a mock up made for a full page demo with a sticky toolbar in context? Could you open a design issue to track?

I've opened https://github.com/patternfly/patternfly-design/issues/1120

packages/react-core/src/components/Toolbar/examples/Toolbar.md
MariaAga: I looked up the history a bit and I think it's leftovers from when toolbar was experimental 
https://github.com/patternfly/patternfly-react/blob/6ec0ce4b75da8a1a4883ed61a3a2adb7dc910237/packages/patternfly-4/react-core/src/experimental/components/DataToolbar/DataToolbar.md?plain=1#L18-L19

I combined the imports for this pr

packages/react-core/src/components/Toolbar/examples/Toolbar.md
MariaAga: Agree! thanks, I changed it

packages/react-core/src/components/Toolbar/examples/Toolbar.md
nicolethoen: I imagine it happens because the IDE added the import weirdly and no one noticed in the initial PR

nicolethoen (approved):

packages/react-core/src/components/Toolbar/examples/Toolbar.md
wise-king-sullyman: @MariaAga I see the imports combined on lines 22/23 but 253/254 here are still split.

packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -442,7 +505,15 @@ A toggle group can be used when you want to collapse a set of items into an over
...
-import { Button, ButtonVariant, InputGroup, Select, SelectOption, SelectVariant, TextInput } from '@patternfly/react-core';
+import {
+  Button,
+  ButtonVariant,
+  InputGroup,
+  Select,
+  SelectOption,
+  SelectVariant,
+  TextInput
+} from '@patternfly/react-core';
```
wise-king-sullyman, line 516: Imports still split here as well.

packages/react-core/src/components/Toolbar/examples/Toolbar.md
```diff
@@ -239,6 +246,62 @@ class ToolbarSpacers extends React.Component {
...
+import React from 'react';
+import { Toolbar, ToolbarItem, ToolbarContent } from '@patternfly/react-core';
+import { Button, ButtonVariant, InputGroup, TextInput, Checkbox } from '@patternfly/react-core';
+import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
+
+const ToolbarItems = () => {
+  const [isSticky, setIsSticky] = React.useState(true);
+  const [showEvenOnly, setShowEvenOnly] = React.useState(true);
+  const array = [...Array(25).keys()];
+  const numbers = showEvenOnly ? array : array.filter(number => number % 2 === 0);
```
wise-king-sullyman, line 261: I think you've got this ternary reversed, shouldn't it be returning the array unchanged when `showEvenOnly` is false?

wise-king-sullyman (changes requested):

packages/react-core/src/components/Toolbar/examples/Toolbar.md
MariaAga: unsplit imports for the added demo only, before I accidentally styled some imports and pushed the changes

packages/react-core/src/components/Toolbar/examples/Toolbar.md
MariaAga: oops, fixed. thanks!

wise-king-sullyman (approved):
Looks great! Thanks!

tlabaj (changes requested):
Looks great! 
Nit. Can you just move the `isSticky` checkbox to the bottom like we typically do in our other examples.

MariaAga:
Thanks for the reviews! Moved the `isSticky` to the button.
Also made the list a bit longer 25 -> 30

nicolethoen (approved):

tlabaj (approved):
LGTM

PR #6731, opened 2022-01-04
fix(jumplinks): Updated JumpLinks demo to allow for horizontal display

### LINKED ISSUE #6580: JumpLinks demo - horizontal toggle not working
https://www.patternfly.org/v4/components/jump-links/react-demos/scrollspy-with-subsections/

interacting with the horizontal toggle does not actually toggle the `isVertical` prop correctly.
I think it's because using `expandable ` forces JumpLinks to stay vertical

jeffpuzzo:
**What**: Closes #6580

Also updated offset to be dynamic based on the current height of the masthead and the jump links nav, since the page is responsive and the height varies on window size.

### DIFF
--- a/packages/react-core/src/demos/JumpLinks.md
+++ b/packages/react-core/src/demos/JumpLinks.md
@@ -15,7 +15,7 @@ JumpLinks has a scrollspy built-in to make your implementation easier. When impl
 
 ### Scrollspy with subsections
 
-This demo expands on the previous to show the JumpLinks in a vertical layout with subsections. It scrolls the [Page](/components/page)'s `mainContainerId` with an `offset` of 76px for the masthead. The headings are given a tab index to allow the jump links to focus them.
+This demo expands on the previous to show the JumpLinks in a vertical layout with subsections. It scrolls the [Page](/components/page)'s `mainContainerId` with an `offset` calculated based on the height of the masthead and the nav list when it appears above the content. The headings are given a tab index to allow the jump links to focus on them.
 
 ```js isFullscreen
 import React from 'react';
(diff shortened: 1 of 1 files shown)


packages/react-core/src/demos/JumpLinks.md
```diff
@@ -38,6 +38,23 @@ ScrollspyH2 = () => {
...
 
   const [isVertical, setIsVertical] = React.useState(true);
+  const [offsetHeight, setOffsetHeight] = React.useState(10);
+
+  // Update offset based on the masthead and jump links nav heights.
+  React.useEffect(() => {
+    const masthead = document.getElementsByClassName('pf-c-masthead')[0];
+    const offsetForPadding = 10;
+
+    getResizeObserver(masthead, () => {
```
nicolethoen, line 48: Im surprised this works with out explicitly importing `getResizeObserver` but I imagine products will need to explicitly import it, so may be worth adding.

nicolethoen (approved):

packages/react-core/src/demos/JumpLinks.md
```diff
@@ -30,14 +30,32 @@ import {
...
+      if (isVertical) {
```
evwilkin, line 50: The vertical layout moves the sticky jumplinks header to the top on mobile and pushes the page content down below it, so we'll need to check for & accommodate the vertical mobile layout when calculating the offset.

packages/react-core/src/demos/JumpLinks.md
```diff
@@ -15,7 +15,7 @@ JumpLinks has a scrollspy built-in to make your implementation easier. When impl
 
 ### Scrollspy with subsections
 
-This demo expands on the previous to show the JumpLinks in a vertical layout with subsections. It scrolls the [Page](/components/page)'s `mainContainerId` with an `offset` of 76px for the masthead. The headings are given a tab index to allow the jump links to focus them.
+This demo expands on the previous to show the JumpLinks in a vertical layout with subsections. It scrolls the [Page](/components/page)'s `mainContainerId` with an `offset` calculated based on the height of the masthead and the nav list when it appears above the content. The headings are given a tab index to allow the jump links to focus on them.
```
evwilkin, line 18: I know you didn't write this part but it confused me that `this demo expands on the previous` when there's only one demo on the page, maybe we can reword this to avoid that confusion.

evwilkin (changes requested):
Looking good!  The offset isn't quite right on mobile with the vertical layout, and I noted one piece of copy that was already there that doesn't quite make sense.  Thanks!

packages/react-core/src/demos/JumpLinks.md
jeffpuzzo: @evwilkin, I was hoping you could elaborate on the expected behavior here. I did notice the stacking of the header when in vertical mode and with a small enough window width (< 768px to be exact). As it stands now, when expanding the list and clicking on 1 nav item, the list collapses and the correct `Heading` is shown/scrolled to.

evwilkin: @jeffpuzzo I noticed that though the page jumps to the correct spot, the scrollspy functionality isn't tracking the active section correctly (actually both on mobile & desktop).  This is a bug unrelated to these changes that [I've logged](https://github.com/patternfly/patternfly-react/issues/6749), sorry for the confusion!

evwilkin (approved):
LGTM - thanks!

packages/react-core/src/demos/JumpLinks.md
jeffpuzzo: I see what you mean now. Good catch. I'll take a look at your fix.

PR #6732, opened 2022-01-04
feat(Modal): Added optional Id for close button

### LINKED ISSUE #5856: Add ability to provide unique identifier for the Close button in a modal
I originally raised this question in https://github.com/patternfly/patternfly/issues/3913, but feedback was that this is a react issue. 

We are currently using `data-` attributes to uniquely identify elements in the DOM for the purpose of usage analytics. One element we are tracking is the Close button in a modal, but we aren't able to add any additional html attributes to this element. Currently, we're tracking this element using the class name `pf-m-plain` but this isn't ideal. 

No known release. At the moment, we're identifying this action via the pf class name provided, but this method prevents us from adding another `.pf-m-plain` button to the modal without breaking the metrics we capture.

tlabaj:
**What**: Closes #5856

Updated the ModalContent pass the `ouiaId` to the Modal close button.  The Modal close button component appends the display name to the `ouiaId` and sets that as the button's `ouiaId`.

To test this add `ouiaId="testId"` to the <Modal> in any example.  This will add "testIdt-ModalBoxCloseButton"

### DIFF
--- a/packages/react-core/src/components/Modal/ModalBoxCloseButton.tsx
+++ b/packages/react-core/src/components/Modal/ModalBoxCloseButton.tsx
@@ -7,14 +7,17 @@ export interface ModalBoxCloseButtonProps {
   className?: string;
   /** A callback for when the close button is clicked */
   onClose?: () => void;
+  /** Close button id */
+  id?: string;
 }
 
 export const ModalBoxCloseButton: React.FunctionComponent<ModalBoxCloseButtonProps> = ({
   className = '',
   onClose = () => undefined as any,
+  id,
   ...props
 }: ModalBoxCloseButtonProps) => (
-  <Button className={className} variant="plain" onClick={onClose} aria-label="Close" {...props}>
+  <Button className={className} variant="plain" onClick={onClose} aria-label="Close" id={id} {...props}>
     <TimesIcon />
   </Button>
 );
--- a/packages/react-core/src/components/Modal/ModalBoxCloseButton.tsx
+++ b/packages/react-core/src/components/Modal/ModalBoxCloseButton.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
 import { Button } from '../Button';
 import TimesIcon from '@patternfly/react-icons/dist/esm/icons/times-icon';
+import { OUIAProps } from '../../helpers';
 
-export interface ModalBoxCloseButtonProps {
+export interface ModalBoxCloseButtonProps extends OUIAProps {
   /** Additional classes added to the close button */
   className?: string;
   /** A callback for when the close button is clicked */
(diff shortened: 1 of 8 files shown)


nicolethoen (approved):

jeffpuzzo:
Looking at the initial issue, I see "We are currently using `data-` attributes to uniquely identify elements in the DOM.", where this is adding an `id` prop to the button, I'm wondering if this solves the problem for the user completely. It's certainly better than them using the className cited in the ticket though.

Should we confirm this meets the requirement?

tlabaj:
> Looking at the initial issue, I see "We are currently using `data-` attributes to uniquely identify elements in the DOM.", where this is adding an `id` prop to the button, I'm wondering if this solves the problem for the user completely. It's certainly better than them using the className cited in the ticket though.
> 
> Should we confirm this meets the requirement?

Yes.  I will as Jenn to take a look.  thanks!

jgiardino:
Thanks for picking up this issue! 

@tiffanynolan has taken over the task of redefining how we tag UI elements so that we can uniquely identify them for the purpose of gathering user metrics, so I'd like her to weigh in on the solution provided here.

One question I have is whether it's possible to enable us to specify a custom `data-` html attribute instead of using the `id` attribute? TBH, I'm not sure how important it is that the html attributes we use for identifying UI elements are the same, but wanted to ask. If that's not an option, then the `id` should support our use case.

tlabaj:
@jeffpuzzo @jgiardino @tiffanyNolan. I updated this to add a prop `closeButtonProps` to modal. It is more flexible and you can add `data-`. attributes or an `id`.

jgiardino (approved):
Thanks for updating this! This approach would definitely support our use case.

jgiardino (changes requested):
Can I take back what I said 😬 

Looking at what Tiffany proposed, she wants to use `data-ouia-component-id` as our attribute for tagging elements, which appears to already be used for this element. When I try to specify that value with this prop, it appears that I can't override the dynamic value that is generated by PF. Sorry for the back-track on this one.

packages/react-core/src/components/Modal/ModalBoxCloseButton.tsx
```diff
@@ -12,9 +13,17 @@ export interface ModalBoxCloseButtonProps {
...
+  ouiaId,
   ...props
 }: ModalBoxCloseButtonProps) => (
-  <Button className={className} variant="plain" onClick={onClose} aria-label="Close" {...props}>
+  <Button
+    className={className}
+    variant="plain"
+    onClick={onClose}
+    aria-label="Close"
+    {...(ouiaId && { ouiaId: `${ouiaId}-${ModalBoxCloseButton.displayName}` })}
```
jeffpuzzo, line 24: I'm not familiar with OUIAProps yet, but does this usage of `ouiaId` meet the user requirements to be able to specify a custom `data-ouia-component-id`?

tlabaj: @jeffpuzzo  Yes.  I explain in the description for this  [issue](https://github.com/patternfly/patternfly-react/pull/6732#issue-1093444865) how you can see this demonstrated.  
I talked to @jgiardino and Tiffany offline and what they were actually looking for is to utilize the `ouiaId` for identification purposes. The modal is already enabled to use `ouiaId`.  Since the Modal is not compassable, the consumer can not add `ouiaId` to the close button.  I pass the model's `ouiaId` to the close button and append the close button's display name to it so it is unique then  I add the `ouiaId` to the button.
You can read more about OUIA [here](https://www.patternfly.org/v4/developer-resources/open-ui-automation)

jeffpuzzo (approved):

tiffanynolan:
This LGTM! Much better and should work with our Pendo tracking.

nicolethoen (approved):

kmcfaul (approved):

dlabrecq (approved):

jgiardino (approved):
LGTM - thanks!

PR #6736, opened 2022-01-04
feat(DescriptionList): add option to fill columns first

### LINKED ISSUE #6685: Description list: Add option to fill columns first
Optionally fill columns and not rows first. Core PR: https://github.com/patternfly/patternfly/pull/4544

boaz0:
**What**: Closes #6685

### DIFF
--- a/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
+++ b/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
@@ -28,7 +28,7 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
   isCompact?: boolean;
   /** Sets a horizontal description list to have fluid styling. */
   isFluid?: boolean;
-  /** Sets the description list to fill the entire column space. */
+  /** Sets the the default placement of description list groups to fill from top to bottom. */
   isFillColumns?: boolean;
   /** Sets the number of columns on the description list */
   columnModifier?: {
--- a/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
+++ b/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
@@ -28,6 +28,8 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
   isCompact?: boolean;
   /** Sets a horizontal description list to have fluid styling. */
   isFluid?: boolean;
+  /** Sets the description list to fill the entire column space. */
+  isFillColumns?: boolean;
   /** Sets the number of columns on the description list */
   columnModifier?: {
     default?: '1Col' | '2Col' | '3Col';
@@ -74,6 +76,7 @@ export const DescriptionList: React.FunctionComponent<DescriptionListProps> = ({
   isInlineGrid,
   isCompact,
   isFluid,
+  isFillColumns,
   columnModifier,
   autoFitMinModifier,
   orientation,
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -891,3 +891,135 @@ import PlusCircleIcon from '@patternfly/react-icons/dist/esm/icons/plus-circle-i
   </DescriptionListGroup>
 </DescriptionList>;
 ```
+
+### Column fill
```
nicolethoen, line 895: Adding the isFillColumns prop to this example does not change the default behavior, right? Is this example necessary?

jeffpuzzo: In the same vein, if we are to have an example, does 1 demonstrating this prop's behavior not suffice (vs the 3 I'm seeing here)?

tlabaj: @jeffpuzzo we typical add examples to match the core implementation.  To reduce the duplicate example we could add check box to already existing `Default` and `Horizontal` example to toggle the `pf-m-fill-columns` modifier. We could also add a `Horizontal multi-column` example with a checkbox toggle since one does not exist.  Or we could add could use one of the other examples that happens to have horizontal multi columns.

cc @mcoker  thoughts on removing the example that really does not change the default behaviour.

mcarrano (commented):
@KKoukiou I believe you filed the original request on this. Can you verify that this is doing what you expect it to? Thanks.

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
mcoker: @tlabaj agree, it's confusing because it doesn't demonstrate anything. We could have a regular and horizontal example, though I'm fine with just having a regular description list with multiple columns.

mcoker (changes requested):
LGTM!

I have a core PR up here to only have a single example. Once that merges, can we update the react examples to match? https://github.com/patternfly/patternfly/pull/4593

Also can you move the example(s) to come after the "Fluid horizontal" example?

boaz0:
@mcoker sure I will do it. No problem and thanks for your help!

mcoker:
@boaz0 https://github.com/patternfly/patternfly/pull/4593 has been merged

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -28,6 +28,8 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
   isCompact?: boolean;
   /** Sets a horizontal description list to have fluid styling. */
   isFluid?: boolean;
+  /** Sets the description list to fill the entire column space. */
+  isFillColumns?: boolean;
```
tlabaj, line 32: Maybe we can say `Sets the the default placement of description list groups to fill from top to bottom`, so the description aligns more with what core notes in their example.

tlabaj (commented):
Looks good.  Just one little suggestion.

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -891,3 +891,49 @@ import PlusCircleIcon from '@patternfly/react-icons/dist/esm/icons/plus-circle-i
...
```
mcoker, line 895: @boaz0 can you move this example up so it comes after fluid horizontal? Currently it's under the auto-fit section and doesn't belong there.

mcoker (changes requested):
Just one little thing!

nicolethoen (approved):

mcarrano (approved):
Looks good. Thanks @boaz0 !

mcoker (approved):
🥳

Issue #6684, opened 2021-12-09
Button: add demo to show progress completion state


mcarrano:
The demo should illustrate how a progress button can be used to reflect status. When the user clicks the button, it should enter "in-progress" state and upon completion, the completion state is displayed. See core issue with linked PR for more detail. https://github.com/patternfly/patternfly/issues/4354

MariaAga:
Should this be the same demo as https://github.com/patternfly/patternfly/pull/4528?

mcarrano:
Yes, it's intended as the React version of that demo that shows the change in state.

PR #6738, opened 2022-01-04
docs(button): adds demo for progress button completion state

MariaAga:
**What**: Closes https://github.com/patternfly/patternfly-react/issues/6684

### DIFF
--- a/packages/react-core/src/demos/Button.md
+++ b/packages/react-core/src/demos/Button.md
@@ -1,0 +1,67 @@
+---
+id: Button
+section: components
+---
+
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
+
+## Demos
+
+### Progress button
+
+```ts
+import React from 'react';
+import { Form, FormGroup, ActionGroup, InputGroup, TextInput, Button } from '@patternfly/react-core';
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
+
+const ProgressButton: React.FunctionComponent = () => {
+  const [loginState, setLoginState] = React.useState<'notLoggedIn' | 'loading' | 'loggedIn'>('notLoggedIn');
+
+  return (
+    <Form>
+      <FormGroup label="Username" isRequired fieldId="progress-button-initial-login">
+        <TextInput
+          isRequired
+          type="text"
+          id="progress-button-initial-login"
+          name="progress-button-initial-login"
+          value="johndoe"
+          aria-label="username input"
+        />
+      </FormGroup>
+      <FormGroup label="Password" isRequired fieldId="progress-button-initial-password">
+        <TextInput
+          isRequired
+          type="password"
+          value="p@ssw0rd"
+          id="progress-button-initial-password"
+          name="progress-button-initial-password"
+          aria-label="password input"
+        />
+      </FormGroup>
+      <ActionGroup>
+        <Button
+          variant="primary"
+          onClick={
+            loginState === 'notLoggedIn'
+              ? () => {
+                  setLoginState('loading');
+                  setTimeout(() => {
+                    setLoginState('loggedIn');
+                  }, 3000);
+                }
+              : null
+          }
+          isLoading={loginState === 'loading'}
+          icon={loginState === 'loggedIn' ? <CheckCircleIcon /> : null}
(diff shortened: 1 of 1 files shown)


packages/react-core/src/demos/Button.md
```diff
@@ -0,0 +1,137 @@
+---
+id: Button
+section: components
+---
+
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
+
+## Examples
```
nicolethoen, line 8: I think this heading should be changed to say Demos

nicolethoen (changes requested):
This looks great, but the three demos should be combined into one to demonstrate the button styles changing as a user interacts with it

mcarrano:
Yep. Agree with @nicolethoen . What I would suggest is adding a timer so after the user clicks the button, the spinner spins for a few seconds and then it changes to the complete state.

tlabaj (changes requested):
I agree with Matt.  I think we really only need one example that goes through the states rather than the 3 separate ones here.  Core does not have interaction so they have to show the 3 examples.

MariaAga:
Thanks for the suggestions!
Changed to be 1 demo, after clicking on login the button will change to loading and after 3 seconds it will change to logged in.

tlabaj (approved):
Thanks for updates!  LGTM.

mcarrano (approved):
Looks great. Thanks for making these updates @MariaAga !

packages/react-core/src/demos/Button.md
```diff
@@ -0,0 +1,75 @@
...
+      <Button variant="primary" icon={<CheckCircleIcon />}>
+        Logged in
+      </Button>
+    )
+  };
+
+  return (
+    <Form>
+      <FormGroup label="Username" isRequired fieldId="progress-button-initial-login">
+        <InputGroup>
```
mcoker, line 48: No need for an `<InputGroup>` unless you want to have a group of form elements/buttons/dropdowns you want to appear beside one another.

packages/react-core/src/demos/Button.md
```diff
@@ -0,0 +1,75 @@
...
+          setLoginState('loading');
+          setTimeout(() => {
+            setLoginState('loggedIn');
+          }, 3000);
+        }}
+      >
+        Link account and log in
+      </Button>
+    ),
+    loading: (
```
mcoker, line 33: Just my $0.02, I would probably expect there to be a single button with vars as values for the `isLoading` and `icon` props, and the button text. The `onClick` function would update those vars to update the props/text. Seems like a more DRY approach. Though the way you've done it here is valid, too.

packages/react-core/src/demos/Button.md
```diff
@@ -0,0 +1,75 @@
...
+      <Button variant="primary" aria-label="Loading..." isLoading>
```
mcoker, line 34: From the existing progress example, looks like we add an `aria-label` to the spinner via `spinnerAriaValueText` instead of adding it to the button element.

```suggestion
      <Button variant="primary" spinnerAriaValueText="Loading..." isLoading>
```

mcoker (changes requested):

packages/react-core/src/demos/Button.md
MariaAga: Thanks for the explanation! removed it

packages/react-core/src/demos/Button.md
MariaAga: Sure, changed to be 1 button

packages/react-core/src/demos/Button.md
MariaAga: Thanks for the info! changed it

packages/react-core/src/demos/Button.md
```diff
@@ -0,0 +1,66 @@
+---
+id: Button
+section: components
+---
+
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
```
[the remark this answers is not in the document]
MariaAga: Thanks for noticing, forgot to add it to the button, added now

nicolethoen (approved):
This looks good to me! 
It even works for the screen reader :)

mcoker (approved):
awesome!

tlabaj (approved):
Looks great!

PR #6746, opened 2022-01-06
fix(log-viewer): update text selection behavior

### LINKED ISSUE #6745: Update Log Viewer text selection behavior
Since the log viewer is virtualized, there might be some issues when the user selects the text and drags to scroll the list because the list is refreshing, the previous selection will be cleared. Thus, the best solution is to disable scrolling when selecting the text, which limits the selection only inside the log window.

DaoDaoNoCode:
**What**: Closes #6745

### DIFF
--- a/packages/react-log-viewer/src/LogViewer/LogViewer.tsx
+++ b/packages/react-log-viewer/src/LogViewer/LogViewer.tsx
@@ -275,7 +275,7 @@ const LogViewerBase: React.FunctionComponent<LogViewerProps> = memo(
           <div className={css(styles.logViewerMain)} style={{ height, width }} ref={containerRef}>
             {loading ? <div>{loadingContent}</div> : createList(parsedData)}
           </div>
-          {footer}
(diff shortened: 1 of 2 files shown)


jschuler (commented):
Interesting solution, works well with mouse. Do we need to worry about keyboard selection as well? (SHIFT+arrow key)

jschuler:
User can also copy some section with mouse, then scroll away, and selection will be lost. I know it's expected due to virtualization, but mentioning it anyways since it's related case.

DaoDaoNoCode:
@jschuler Sounds interesting, yeah I only tested the trackpad selection situation. SHIFT+arrow key is inevitable. Do you have other better solutions? The requirement is here in this comment https://github.com/openshift/console/pull/10486#issuecomment-1005807702.

jschuler (approved):

dlabrecq (approved):

nicolethoen (approved):

PR #6750, opened 2022-01-07
fix(jumplinks): fixed scrollspy breakpoints are incorrect if offset prop changes

### LINKED ISSUE #6749: JumpLinks - scrollspy incorrect if offset changes
If the `offset` prop changes, the scrollspy functionality does not change to reflect the new offset, causing the active jumplink to update at the incorrect scroll points.  This is due to the `scrollSpy` function's `useCallback` not including the `offset` prop as a dependency to watch [here](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/JumpLinks/JumpLinks.tsx#L140).  

Note that clicking on a jumplink will still scroll to the correct spot because the `cloneChildren` function is not memoized so the correct `offset` is passed [here](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/JumpLinks/JumpLinks.tsx#L194).

Bug

evwilkin:
**What**: Closes #6749 

This PR adds the `offset` prop to the dependencies array of the `scrollSpy` function's `useCallback`, fixing a bug where the scrollspy would never update if the `offset` prop changed.

In our previous demo the `offset` was hard coded so the bug didn't present itself, but the bug can be seen in https://github.com/patternfly/patternfly-react/pull/6731 where the `offset` prop defaults to `10` but then is set dynamically when rendered.

### DIFF
--- a/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
+++ b/packages/react-core/src/components/JumpLinks/JumpLinks.tsx
@@ -137,7 +137,7 @@ export const JumpLinks: React.FunctionComponent<JumpLinksProps> = ({
         }
       }
     });
-  }, [scrollItems, hasScrollSpy, scrollableElement]);
+  }, [scrollItems, hasScrollSpy, scrollableElement, offset]);
 
   React.useEffect(() => {
     scrollableElement = document.querySelector(scrollableSelector) as HTMLElement;


jeffpuzzo (approved):
Verified this fix combined with the merged fix from #6731. Looks good!

nicolethoen (approved):

PR #6752, opened 2022-01-07
fix(Tabs): href paths not being pushed to url

### LINKED ISSUE #6533: Tabs href prop doesn't work
How to reproduce:
1. Consume Tab and Tabs PF4 components
2.  Add `href` prop on each tab with an hash path (i.e `#/path`)

Expected behavior:  Clicking on a tab should apply the `href` path
Actual behavior: Nothing happens, though inspecting the tab element shows a rendered `<a>` tag with `href` property

It seems that a `preventDefault` was added recently which prevent the href to propagate as expected

jeffpuzzo:
**What**: Closes #6533

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -147,10 +147,15 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
   handleTabClick(
     event: React.MouseEvent<HTMLElement, MouseEvent>,
     eventKey: number | string,
-    tabContentRef: React.RefObject<any>
+    tabContentRef: React.RefObject<any>,
+    href?: string
   ) {
     // When tab is an achor tag with href, cancel navigation event
-    event.preventDefault();
+    if (href) {
+      event.preventDefault();
+      window.history.pushState(null, null, href);
+    }
+
     const { shownKeys } = this.state;
     const { onSelect, defaultActiveKey } = this.props;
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -147,10 +147,15 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
...
     eventKey: number | string,
-    tabContentRef: React.RefObject<any>
+    tabContentRef: React.RefObject<any>,
+    href?: string
   ) {
     // When tab is an achor tag with href, cancel navigation event
-    event.preventDefault();
+    if (href) {
+      event.preventDefault();
+      window.history.pushState(null, null, href);
```
jonkoops, line 156: Why manipulate the history manually here? Omitting the `preventDefault()` call should allow the browser to navigate normally.

jonkoops (changes requested):

packages/react-core/src/components/Tabs/Tabs.tsx
jeffpuzzo: You can see in the `Using the nav element` example that simply removing the `preventDefault` call as I see you've done in a separate PR results in tabs not being focused. This is the reason history is being utilized above when an `href` is in play.

jeffpuzzo: I agree now that it is a problem with the Patternfly docs.

PR #6755, opened 2022-01-10
fix(Tabs): allow href navigation from tabs

### LINKED ISSUE #6533: Tabs href prop doesn't work

jonkoops:
Closes #6533

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -149,8 +149,6 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
     eventKey: number | string,
     tabContentRef: React.RefObject<any>
   ) {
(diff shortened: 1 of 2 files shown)


jeffpuzzo (approved):
Spoke [in Slack](https://patternfly.slack.com/archives/C4FM977N0/p1641819783009100) about this. Although this does introduce an problem with the Patternfly nav examples not being focused, it does seem to be specific to our environment and therefore a separate issue.

CooperRedhat (approved):

Issue #6705, opened 2021-12-16
Expandable section - add indented variation


mcoker:
re: https://github.com/patternfly/patternfly/pull/4571

`.pf-m-indented` will indent the contents

PR #6763, opened 2022-01-11
feat(expandable-section): added indented variation

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/6705

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/ExpandableSectionDemo/ExpandableSectionDemo.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/ExpandableSectionDemo/ExpandableSectionDemo.tsx
@@ -45,7 +45,7 @@ export class ExpandableSectionDemo extends React.Component<null, ExpandableSecti
           This content is visible only when the component is expanded.
         </ExpandableSection>
         <br />
-        <h1>Detached expandable section</h1>
+        <h1>Detached Expandable Section</h1>
         <Stack hasGutter>
           <StackItem>
             <ExpandableSection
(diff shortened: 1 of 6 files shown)


nicolethoen (approved):
This all looks great!

We are all, however, being encouraged to try and write all new examples using FunctionComponents rather than class components, such as is done in this example:
[AccordionList](http://patternfly-react-main.surge.sh/components/accordion#definition-list) and the code for that example lives [here](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Accordion/examples/AccordionDefinitionList.tsx)

Also, we've started including our new examples, each in their own typescript file and include them in the MD like [this](https://github.com/patternfly/patternfly-react/blob/03e3df6771bae6c856dfb1d5a48397031cbfb37a/packages/react-core/src/components/Accordion/examples/Accordion.md?plain=1#L14)

These suggestions are not blocking, since all the expandable-section examples need to be updated eventually, so this can be updated then. This is more of an FYI - and an opportunity to let you try and update it if you want :)

packages/react-integration/demo-app-ts/src/components/demos/ExpandableSectionDemo/ExpandableSectionDemo.tsx
```diff
@@ -81,6 +81,16 @@ export class ExpandableSectionDemo extends React.Component<null, ExpandableSecti
         >
           This content is visible only when the component is expanded.
         </ExpandableSection>
+        <br />
```
tlabaj, line 84: We don't really need to add this demo since there is no related test case.  We have recently changed what we are writing for demos since Zack made the effort to clean up integration test.  With the shift to TS for examples, the demo-app will eventually not be needed and we can test against our example code.

tlabaj (changes requested):

tlabaj (commented):
Looks Good! Just one comment.

mcarrano (approved):
LGTM!

tlabaj (approved):
LGTM

PR #6764, opened 2022-01-11
feat(NumberInput): call onMinus/onPlus on arrow-up/down pressed

### LINKED ISSUE #6753: NumberInput does not trigger onMinus/onPlus when cursor keys are used
Using:

```javascript
import React from 'react';
import { NumberInput } from '@patternfly/react-core';

class BasicNumberInput extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      value: 8192
    };

    this.onMinus = () => {
      //if (this.state.value - 512 < 0)  // can not be clamped here (ignored by cursor keys)
      //  return
      this.setState({
        value: this.state.value - 512
      });
    };

    this.onChange = event => {
      //if (this.state.value < 0) {   // also needs to clamp here
      //  this.setState({
      //    value: Number(0)
      //  });
      //  return;
      //}

      this.setState({
        value: Number(event.target.value)
      });
    };

    this.onPlus = () => {
      this.setState({
        value: this.state.value + 512
      });
    };
  }

  render() {
    const { value } = this.state;
    return (
      <NumberInput
        value={value}
        onMinus={this.onMinus}
        onChange={this.onChange}
        onPlus={this.onPlus}
        inputName="input"
        inputAriaLabel="number input"
        minusBtnAriaLabel="minus"
        plusBtnAriaLabel="plus"
      />
    );
  }
}
```

I would like to allow the user to easily increment a memory value in steps of `512`. Using the buttons this works, but when the user has selected the input textbox itself, the cursor keys are not regarded as a **plus** or **minus** operation and do not trigger the events.

While this can make for great for control, but it is not expected. However, the worst is that now the clamping needs to happen in the `onChange`

boaz0:
**What**: Closes #6753

### DIFF
--- a/packages/react-core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-core/src/components/NumberInput/NumberInput.tsx
@@ -4,6 +4,7 @@ import { css } from '@patternfly/react-styles';
 import MinusIcon from '@patternfly/react-icons/dist/esm/icons/minus-icon';
 import PlusIcon from '@patternfly/react-icons/dist/esm/icons/plus-icon';
 import { Button, ButtonProps } from '../Button';
+import { KEY_CODES } from '../../helpers';
 
 export interface NumberInputProps extends React.HTMLProps<HTMLDivElement> {
   /** Value of the number input */
@@ -44,6 +45,19 @@ export interface NumberInputProps extends React.HTMLProps<HTMLDivElement> {
   plusBtnProps?: ButtonProps;
 }
 
+type DefaultKeyDownHandlerArgs = Pick<NumberInputProps, 'inputName' | 'onMinus' | 'onPlus'>;
+
+const defaultKeyDownHandler = (args: DefaultKeyDownHandlerArgs) => (event: React.KeyboardEvent<HTMLInputElement>) => {
+  if (KEY_CODES.ARROW_UP === event.keyCode && args.onPlus) {
+    event.preventDefault();
+    args.onPlus(null, args.inputName);
+  }
+  if (KEY_CODES.ARROW_DOWN === event.keyCode && args.onMinus) {
+    event.preventDefault();
+    args.onMinus(null, args.inputName);
+  }
+};
+
 export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
   value = 0,
   className,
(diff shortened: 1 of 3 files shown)


mcarrano (commented):
@boaz0 I don't really understand what was changed here. Seems like this behaves as before. Can you provide more explanation?

gbraad:
When a different value is used for onPlus and/or onMinus, like let;s say `+ 2`, you do not actually use this value when cursor keys, like up and down are used. In that case it merely increases by the value of ` + 1` and vice versa.

The suggestion is to actually invoke `onPlus` and `onMinus` when the cursor keys up and down are used.

> Seems like this behaves as before. 

This means there is no regression. For more information, please see #6753

gbraad (approved):

mcarrano:
Thanks for the explanation @gbraad . I see how this is working now. 

@tlabaj @nicolethoen Do you think we should add a new example to illustrate this behavior? I think this is very useful functionality (i.e. custom increment) but it's not obvious that one can do this.

gbraad:
If you wanna describe custom increments you also need to resolve #6765 as the current min and max checks can be overrun as this is merely checked with an equals operator.

boaz0:
@mcarrano & @gbraad I decided it's better to move this code as an example and instead of binding `onKeyHandler` to `onKeyDown` I will let the user set it through `NumberInput`. That way our customers are more free to do whatever they want.

Feel free to review this now.

mcarrano:
@boaz0 Hmmm.... I see what you've done here. Yes, this does provide complete flexibility, but is that actually what we want? From a usability perspective it seems odd that the plus and minus increments would be different and that mouse and keypress events would also have different behaviors. I don't think I'd ever advise that as a designer. My ask about creating a new example was simply that while the default behavior is +/-1, it shows that the increment can be customized to take on another value and this increment would be consistently applied by either using the mouse or keyboard. @nicolethoen @tlabaj what are your thoughts?

I also agree with @gbraad that it would be good to also address https://github.com/patternfly/patternfly-react/issues/6765 as part of this PR.

tlabaj:
I agree with @mcarrano.  I would expect the clicking the buttons and using the cursor keys to be the same. 
I think the original solution and anding an example of a `custom` increment would be great.

gbraad:
Not understanding what the disagreement is, as I want both plus/minus and keypresses to do the same thing. That is the expectation and would otherwise break the mental model

boaz0:
Sure no problem. I will revert my latest changes.

tlabaj:
@boaz0 I am fine with a separate PR for the max/min issue.

boaz0:
@tlabaj :+1: done

mcarrano (approved):
Looks great. Thanks @boaz0 !

wise-king-sullyman (approved):
Looks great!

PR #6767, opened 2022-01-12
feat(Tabs): enforce restricted type for children

### LINKED ISSUE #6766: Enforce type for children of Tabs component
The `Tabs` component can only accept children of the type of the `Tab` component. This is [documented by a comment](https://github.com/patternfly/patternfly-react/blob/004cdf8648be921f15c9dafa489964f4bc4fa8d5/packages/react-core/src/components/Tabs/Tabs.tsx#L22-L23), but not enforced by TypeScript. To resolve this issue the correct type should be added so errors can be caught at compile time.

jonkoops:
Closes #6766

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -167,7 +167,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
     if (tabContentRef) {
       React.Children.toArray(this.props.children)
         .filter((child): child is TabElement => React.isValidElement(child))
-        .filter(child => child.props && child.props.tabContentRef && child.props.tabContentRef.current)
+        .filter(({ props }) => props.tabContentRef && props.tabContentRef.current)
         .forEach(child => (child.props.tabContentRef.current.hidden = true));
       // most recently selected tabContent
       if (tabContentRef.current) {
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -18,9 +18,12 @@ export enum TabsComponent {
   nav = 'nav'
 }
 
+type ChildElement = React.ReactElement<TabProps, React.JSXElementConstructor<TabProps>>;
+type Child = ChildElement | boolean | null | undefined;
+
 export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivElement>, 'onSelect'>, OUIAProps {
-  /** Content rendered inside the tabs component. Must be React.ReactElement<TabProps>[] */
-  children: React.ReactNode;
+  /** Content rendered inside the tabs component. */
+  children: Child | Child[];
   /** Additional classes added to the tabs */
   className?: string;
   /** Tabs background color variant */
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -18,9 +18,12 @@ export enum TabsComponent {
...
 }
 
+type ChildElement = React.ReactElement<TabProps, React.JSXElementConstructor<TabProps>>;
+type Child = ChildElement | boolean | null | undefined;
+
 export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivElement>, 'onSelect'>, OUIAProps {
-  /** Content rendered inside the tabs component. Must be React.ReactElement<TabProps>[] */
-  children: React.ReactNode;
+  /** Content rendered inside the tabs component. */
+  children: Child | Child[];
```
nicolethoen, line 26: The way our docs are generated from the props interface, the website is not doing much to describe what the definition of `Child` is. ([see here](https://patternfly-react-pr-6767.surge.sh/components/tabs/#tabs))
The easiest way to fix this for now would probably be to be more descriptive about the type def of `Child` in the prop description comment.

jonkoops: @nicolethoen I've pushed a commit to clarify this a bit WDYT?

tlabaj: Technically this is a breaking change.  It probably would not impact anyone but we try and not introduce anything that could be breaking like more restrictive type changes.  We can pull this update into our next breaking change release.

tlabaj (changes requested):

packages/react-core/src/components/Tabs/Tabs.tsx
nicolethoen: Based on the issue - it sounds like this type enforcement is already necessary - using any other type would already produce a broken tab

jonkoops: Although the type change is technically breaking, the runtime behavior is actually improved. Before if you'd pass in a [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value the component would attempt to render it, and since there are no props associated with these children this will cause a run-time error (see [this example](https://codesandbox.io/s/practical-elbakyan-b99oj)).

This is caused by the fact that before the children were filtered using a [boolean filter](https://github.com/patternfly/patternfly-react/pull/6767/files#diff-b19cd68b4e0b4ab51ece072430fafe67e2a064f4241b44a923867085f04f6497L312), which would allow truthy values that are not React elements. In this PR these filters are replaced by [`isValidElement()`](https://reactjs.org/docs/react-api.html#isvalidelement), which asserts the value passed is a valid React element created with [`createElement()`](https://reactjs.org/docs/react-api.html#createelement), meaning the required props will be present.

So considering that there could realistically not be any applications out there doing this, unless they created a new component that matches all of the required props of `Tab` I don't see this as a breaking change.

nicolethoen (approved):

jonkoops:
@nicolethoen @tlabaj do we want to go ahead and merge this one? (I don't have permission to do this)

jonkoops:
@nicolethoen @tlabaj do we want to go ahead and merge this one? (I don't have permission to do this)

nicolethoen:
as soon as @tlabaj gives this a thumbs up, she can merge it :)

tlabaj (approved):

jonkoops:
@nicolethoen could you re-open this issue as it's solution was reverted?

nicolethoen:
I did reopen the associated issue: https://github.com/patternfly/patternfly-react/issues/6766

jonkoops:
Thanks!

PR #6769, opened 2022-01-12
feat(TimePicker): Allow for seconds via new prop; includeSeconds

### LINKED ISSUE #5988: bug(DatePicker): the aria-label of the button for entries after the 30th of the month are wrong
Here is the datepicker when opened in July 2021:
Screen Shot 2021-07-06 at 17 00 47

The focused (with blue) item, belongs to August, and correctly set the date to 1st August if submitted, however in the DOM the aria label shows July.

```<button class="pf-c-calendar-month__date" type="button" tabindex="-1" aria-label="1 July 2021">1</button>```

### LINKED ISSUE #5998: Question: Will TimePicker component support seconds in the future?
I'm not sure about the other options to handle time on Patternfly, but the two that I'm aware of `TimePicker` and `TextField` with `time` type can't handle seconds. They're both `Time` components and for me it's kind strange to not handle seconds on it. Will it be supported in the future? Or already have any other option?

image

jeffpuzzo:
**What**: Closes #5998

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -17,7 +17,8 @@ import {
   pmSuffix,
   getHours,
   getMinutes,
-  isWithinMinMax
+  isWithinMinMax,
+  getSeconds
 } from './TimePickerUtils';
 
 export interface TimePickerProps
@@ -41,7 +42,7 @@ export interface TimePickerProps
   /** True if the time is 24 hour time. False if the time is 12 hour time */
   is24Hour?: boolean;
   /** Optional event handler called each time the value in the time picker input changes. */
-  onChange?: (time: string, hour?: number, minute?: number, isValid?: boolean) => void;
+  onChange?: (time: string, hour?: number, minute?: number, seconds?: number, isValid?: boolean) => void;
   /** Optional validator can be provided to override the internal time validator. */
   validateTime?: (time: string) => boolean;
   /** Id of the time picker */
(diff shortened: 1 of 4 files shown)


mcarrano (approved):
Looks great. Thanks @jeffpuzzo !

nicolethoen (changes requested):
I have not investigated why, but I can no longer edit the time of the first example by typing in the input.

Also, originally I could type in `3:35` and then tab out of the time picker and see the am/pm get automatically appended to the end. I think I was expecting that if `includeSeconds` was passed as a prop, then if I typed `3:35` and then tab out of the input, the `:00` would also automatically be appended after the time before the am/pm (when applicable). @mcarrano were you expecting that?

jeffpuzzo:
> I have not investigated why, but I can no longer edit the time of the first example by typing in the input.
> 
> Also, originally I could type in `3:35` and then tab out of the time picker and see the am/pm get automatically appended to the end. I think I was expecting that if `includeSeconds` was passed as a prop, then if I typed `3:35` and then tab out of the input, the `:00` would also automatically be appended after the time before the am/pm (when applicable). @mcarrano were you expecting that?

This feedback has been addressed. If we decide to not auto-fill with 00 seconds, I can revert that change.

mcarrano:
> Also, originally I could type in 3:35 and then tab out of the time picker and see the am/pm get automatically appended to the end. I think I was expecting that if includeSeconds was passed as a prop, then if I typed 3:35 and then tab out of the input, the :00 would also automatically be appended after the time before the am/pm (when applicable). @mcarrano were you expecting that?

Yes, I agree. @jeffpuzzo Looks like the first example was fixed but the second example (24 hour time) was not.

jeffpuzzo:
> > Also, originally I could type in 3:35 and then tab out of the time picker and see the am/pm get automatically appended to the end. I think I was expecting that if includeSeconds was passed as a prop, then if I typed 3:35 and then tab out of the input, the :00 would also automatically be appended after the time before the am/pm (when applicable). @mcarrano were you expecting that?
> 
> Yes, I agree. @jeffpuzzo Looks like the first example was fixed but the second example (24 hour time) was not.

To be clear, I don't believe the `Basic 24 hour time` example was meant to append anything automatically, but rather potentially remove seconds if they were included. I did update the `24 hours with seconds` (which I assume is what you were referring to) to auto-append seconds on Tab/Enter/etc.

nicolethoen (approved):
LGTM

kmcfaul (approved):

mcarrano:
> @mcarrano To be clear, I don't believe the Basic 24 hour time example was meant to append anything automatically, but rather potentially remove seconds if they were included. I did update the 24 hours with seconds (which I assume is what you were referring to) to auto-append seconds on Tab/Enter/etc.

@jeffpuzzo I agree. This looks good now!

PR #6774, opened 2022-01-12
fix(TextInputGroup): address various issues

### LINKED ISSUE #6697: Text Input Group issues
I added PatternFly's new `Text Input Group` component to Cost Management. The type-ahead is more usable than the `Select` component, but there are limitations. For example, we cannot add place holder text and an aria label. I also had to take some liberties regarding how keyboard navigation works.

Below are the observations I made while implementing the Text Input Group component.

**Component Issues**

1. Cannot add place holder text
2. . `TextInputGroupMain` does not clear input text for a value of undefined — must be an empty string

**Example issues**

1. Example doesn't use aira labels
2. Input focus box not rendered correctly without clear button — must conditionally render `TextInputGroupUtilities`
3. Input value cleared when menu has focus and escape key is pressed — should just close the menu and set focus on the text input for further editing
4. The tab key doesn't appear to work at all -- code looks like it was intended to select a menu option?

**Cost Management implementation**

1. Added “No results found” label instead of always showing “Suggestions”
2. Made the tab key clear the input value and navigate to the next tab index
3. A lot of keyboard navigation code must be added manually — painstaking to get working correctly

chrome-capture

wise-king-sullyman:
On auto complete search:

- undoes default tab behavior prevention

- undoes input clearing on escape when focused on menu

- adds input clearing on tab when focused on menu

- adds placeholder text and custom aria label

- clarifies reasoning for conditionally rendered search icon

On text input group main:

- sets falsy input values as an empty string

- exposes value and placeholder props

**What**: Closes #6697

**Additional issues**:

[Direct link to the auto complete search preview](https://patternfly-react-pr-6774.surge.sh/components/text-input-group/react-demos#auto-complete-search)

### DIFF
--- a/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
+++ b/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
@@ -31,6 +31,10 @@ export interface TextInputGroupMainProps extends Omit<React.HTMLProps<HTMLDivEle
   onBlur?: (event?: any) => void;
   /** Accessibility label for the input */
   'aria-label'?: string;
+  /** Value for the input */
+  value?: string | number;
+  /** Placeholder value for the input */
+  placeholder?: string;
 }
 
 export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps> = ({
@@ -43,6 +47,7 @@ export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps
   onBlur,
   'aria-label': ariaLabel = 'Type to filter',
   value: inputValue,
+  placeholder: inputPlaceHolder,
   ...props
 }: TextInputGroupMainProps) => {
   const { isDisabled } = React.useContext(TextInputGroupContext);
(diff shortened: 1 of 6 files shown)


dlabrecq (changes requested):
Per @nicolethoen, tab behavior should mimic the Search Input example.

Within the menu of Text Input Group, pressing the tab key moves focus to another element in the page, while the menu remained open. It seems the user must press the escape key first, which was not expected?

For Search Input, pressing the tab key returns focus to the input. From there, I can tab to the clear button. 

**Text Input Group example**

chrome-capture


**Search Input example**

chrome-capture (1)

wise-king-sullyman:
Tab now closes the menu and focuses on the text input. Additionally, I saw that this demo was missing the search input auto complete demo functionality of closing the menu when `Enter` or `Space` were hit so I added that as well to bring these demos more in line with each other.

dlabrecq (approved):
LGTM

nicolethoen:
I think if focus is on the input and user hits tab to complete the autocomplete, the menu should close when the focus moves to the next focusable element (the close button)

wise-king-sullyman:
0b266d4 was added as an alternate solution to patternfly/patternfly#4588. This commit demonstrates how consumers can conditionally render the `TextInputGroupUtilities` component itself as well as its children.

nicolethoen (approved):

tlabaj (approved):

kmcfaul (approved):

mmenestr:
LGTM!

mmenestr (approved):

tlabaj (approved):

PR #6777, opened 2022-01-13
chore(Tabs): Add demo for Tabs within Modals

### LINKED ISSUE #6772: Add Modal tabs react demo

jeffpuzzo:
**What**: Closes #6772

### DIFF
--- a/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
+++ b/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
@@ -30,7 +30,7 @@ interface Product {
 const products: Product[] = [
   {
     id: 0,
-    name: 'Patternfly',
+    name: 'PatternFly',
     description: 'PatternFly is a community project that promotes design commonality and improves user experience.'
   },
   {
@@ -104,13 +104,13 @@ export const ModalTabs: React.FunctionComponent = () => {
               <TabContent eventKey={1} id="doc-tab" hidden={activeTabKey !== 1}>
                 <List>
                   <ListItem>
-                    <a>Doc Link 1</a>
+                    <a>Doc link 1</a>
                   </ListItem>
                   <ListItem>
-                    <a>Doc Link 2</a>
+                    <a>Doc link 2</a>
                   </ListItem>
                   <ListItem>
-                    <a>Doc Link 3</a>
+                    <a>Doc link 3</a>
                   </ListItem>
                 </List>
               </TabContent>
--- a/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
+++ b/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
@@ -84,8 +84,8 @@ export const ModalTabs: React.FunctionComponent = () => {
       <DashboardWrapper mainContainerId="main-content-card-view-default-nav" breadcrumb={null}>
         <PageSection variant={PageSectionVariants.light}>
           <TextContent>
-            <Text component="h1">Products</Text>
-            <Text component="p">Click any product card to view Tabs within Modals.</Text>
+            <Text component="h1">Projects</Text>
+            <Text component="p">Click any project card to view Tabs within Modals.</Text>
           </TextContent>
         </PageSection>
         <PageSection isFilled>
--- a/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
+++ b/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
@@ -80,7 +80,7 @@ export const ModalTabs: React.FunctionComponent = () => {
   );
 
   return (
-    <>
+    <React.Fragment>
       <DashboardWrapper mainContainerId="main-content-card-view-default-nav" breadcrumb={null}>
         <PageSection variant={PageSectionVariants.light}>
           <TextContent>
@@ -137,6 +137,6 @@ export const ModalTabs: React.FunctionComponent = () => {
           </Grid>
         </Modal>
       )}
-    </>
+    </React.Fragment>
   );
 };
--- a/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
+++ b/packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
@@ -49,12 +49,21 @@ const products: Product[] = [
 export const ModalTabs: React.FunctionComponent = () => {
   const [isModalOpen, setIsModalOpen] = React.useState(false);
   const [selectedProduct, setSelectedProduct] = React.useState<Product>();
-  const [activeTabKey, setActiveTabKey] = React.useState(0);
+  const [activeTabKey, setActiveTabKey] = React.useState<string | number>(0);
 
   const onCardClick = React.useCallback(
     (product: Product) => () => {
       setSelectedProduct(product);
       setIsModalOpen(true);
+    },
+    []
+  );
+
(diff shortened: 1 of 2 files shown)


mcarrano (commented):
I am seeing a difference in appearance between the tabs in this demo and the core tabs in modal demo here: https://www.patternfly.org/v4/components/tabs/html-demos/modal-tabs/ Looks like the core demo is missing the horizontal separator between the tab bar and the content panel.

Can you look at this @mcoker ? I think that the React demo has it right. @mceledonia can you confirm? (React version for reference: https://patternfly-react-pr-6777.surge.sh/components/tabs/react-demos/modal-tabs/)

mcoker:
@mcarrano from the [tabs demos core PR](https://github.com/patternfly/patternfly/pull/3914), here is the design for tabs in a modal - https://marvelapp.com/prototype/5fba0cj/screen/73740953

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,110 @@
...
+            ))}
+          </Gallery>
+        </PageSection>
+      </DashboardWrapper>
+
+      {selectedProduct && (
+        <Modal variant={ModalVariant.small} title={selectedProduct.name} isOpen={isModalOpen} onClose={closeModal}>
+          <Tabs defaultActiveKey={0}>
+            <Tab eventKey={0} title={<TabTitleText>Details</TabTitleText>}>
+              <div className="pf-u-mt-md">{selectedProduct.description}</div>
```
mcoker, line 88: Instead of the manual margin utility class on the tab content container, the core demo separates the tab buttons from the tab content, and puts them in a `<Grid hasGutter>` layout, which creates space between the tab buttons and the tab contents. The way you've done it here works, though I'd say the layout method is more sustainable since you don't have to repeat the utility class for all tab content.

mcoker (commented):
LGTM other than @mcarrano's comment and a comment about using a single layout for the tabs + content spacing versus using a utility class to wrap each tab's content for that spacing.

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
jeffpuzzo: Thanks for the tip. Updated with a Grid/gutter & set the Tabs to secondary so the divider line is no longer visible.

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+              </Tabs>
+            </GridItem>
+            <GridItem>
+              <TabContent eventKey={0} id="details-tab" hidden={activeTabKey !== 0}>
+                {selectedProduct.description}
+              </TabContent>
+              <TabContent eventKey={1} id="doc-tab" hidden={activeTabKey !== 1}>
+                <List>
+                  <ListItem>
+                    <a>Doc Link 1</a>
```
mcoker, line 107:
```suggestion
                    <a>Doc link 1</a>
```

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  </ListItem>
+                  <ListItem>
+                    <a>Doc Link 2</a>
```
mcoker, line 110:
```suggestion
                    <a>Doc link 2</a>
```

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  </ListItem>
+                  <ListItem>
+                    <a>Doc Link 3</a>
```
mcoker, line 113:
```suggestion
                    <a>Doc link 3</a>
```

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+                <CardBody>{product.description}</CardBody>
+              </Card>
+            ))}
+          </Gallery>
+        </PageSection>
+      </DashboardWrapper>
+
+      {selectedProduct && (
+        <Modal variant={ModalVariant.small} title={selectedProduct.name} isOpen={isModalOpen} onClose={closeModal}>
+          <Grid hasGutter>
```
mcoker, line 93: 🔥

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+interface Product {
+  id: number;
+  name: string;
+  description: string;
+}
+
+const products: Product[] = [
+  {
+    id: 0,
+    name: 'Patternfly',
```
mcoker, line 33: match the case in the description

```suggestion
    name: 'PatternFly',
```

mcoker (approved):
LGTM! Just a couple of small nits about case - we use sentence case in the docs for everything except proper nouns and such.

mcarrano (approved):
> @mcarrano from the tabs demos core PR, here is the design for tabs in a modal - https://marvelapp.com/prototype/5fba0cj/screen/73740953

OK. Looks like this is per design. Thanks for making the updates @jeffpuzzo .

nicolethoen (changes requested):
The cards should be selectable (i.e. you should be able to open the modals) with a keyboard - using `Enter` or `space`

I'm also realizing that the screen reader is unaware that the cards are selectable - but that is likely an issue with the implementation of selectable cards - I opened this issue: https://github.com/patternfly/patternfly-react/issues/6798

packages/react-core/src/demos/examples/Tabs/ModalTabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+  },
+  {
+    id: 2,
+    name: 'Apache Spark',
+    description: 'This documentation page covers the Apache Spark component for the Apache Camel.'
+  }
+];
+
+export const ModalTabs: React.FunctionComponent = () => {
+  const [isModalOpen, setIsModalOpen] = React.useState(false);
```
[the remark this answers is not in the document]
jeffpuzzo: This default value infers what the type is, are we sure we want to add explicit types when types are inferred?

If we do want to mention the type regardless, I assume we would only be doing this because it's an example?

nicolethoen:
The only thing I'm scratching my head about with the demo - is that the ts>>js button in the example's code editor is throwing an error. I wonder if there is something related to the complexity of some of the types? The ts>>js parser is managed in org. We might want to investigate. 
Either, we should make clearer or simplify whatever it is throwing the error in that conversion - or we should open an issue in org to fix up the parser.

jeffpuzzo:
> The only thing I'm scratching my head about with the demo - is that the ts>>js button in the example's code editor is throwing an error. I wonder if there is something related to the complexity of some of the types? The ts>>js parser is managed in org. We might want to investigate. Either, we should make clearer or simplify whatever it is throwing the error in that conversion - or we should open an issue in org to fix up the parser.

Had to do with the usage of `<>` vs `<React.Fragment>`, however we're parsing/converting to JS requires the latter for some reason.

nicolethoen (approved):

dlabrecq (approved):

kmcfaul (approved):

Issue #6553, opened 2021-11-06
Spike: Look into Lodash usage - can we remove dependency?


evwilkin:
**Describe the issue. What is the expected and unexpected behavior?**
Lodash is a dependency of react-charts, but has been identified as adding large size to our package download size.  Review how/why it is used and if it can be removed.

@tlabaj FYI

PR #6784, opened 2022-01-13
fix(Tabs): enable tabs scroll button for small window

### LINKED ISSUE #5175: Tab overflow disabled for small windows
In Cost management, we noticed that the tab overflow controls stop working at small window sizes. When the browser window is resized a bit larger, the controls begin working, again.

I can repeatedly resize the window to see the controls are enabled / disabled. Although I can scroll side to side using the mouse, but the prev / next buttons are unusable.

Looking at the inspector, the size of the container is roughly 350px.

chrome-capture (5)

DaoDaoNoCode:
**What**: Closes #5175 
This issue is because there the button only scrolls left or right when the tab is totally inside the view. However, when the window is small, the tab content could even be longer than the list width, which makes it can never find the element in the view. To solve the problem, I created a new optional prop named `strict` and made a judgment on the element width and the container width, if the container width is smaller than the element width and the `strict` is not set, we treat it as part in view. Only when strict mode is set, we will not take the element width and container width into consideration, which brings more flexibility to the function.
Also, I update the bounding to make it more precise to judge whether an element is in view when the element width is not an integer. And add a debounce to the scroll function.

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as _ from 'lodash';
 import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
 import buttonStyles from '@patternfly/react-styles/css/components/Button/button';
 import { css } from '@patternfly/react-styles';
@@ -178,7 +179,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
     }
   }
 
-  handleScrollButtons = () => {
+  handleScrollButtons = _.debounce(() => {
     const container = this.tabList.current;
     let disableLeftScrollButton = true;
     let disableRightScrollButton = true;
@@ -201,7 +202,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
       disableLeftScrollButton,
       disableRightScrollButton
     });
-  };
+  }, 100);
 
   scrollLeft = () => {
     // find first Element that is fully in view on the left, then scroll to the element before it
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import * as _ from 'lodash';
 import styles from '@patternfly/react-styles/css/components/Tabs/tabs';
 import buttonStyles from '@patternfly/react-styles/css/components/Button/button';
 import { css } from '@patternfly/react-styles';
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -178,7 +179,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
     }
   }
 
-  handleScrollButtons = () => {
+  handleScrollButtons = _.debounce(() => {
```
jeffpuzzo, line 182: Echoing what I've been told recently here, which is lodash isn't something we want as a dependency going forward; https://github.com/patternfly/patternfly-react/issues/6553.

So I would suggest using `setTimeout` instead to accomplish the debouncing of this operation.

dlabrecq (commented):
Other than the Lodash issue, it LGTM

packages/react-core/src/components/Tabs/Tabs.tsx
DaoDaoNoCode: @jeffpuzzo Thanks, I will remove the use of lodash.

packages/react-core/src/helpers/util.ts
```diff
@@ -39,24 +39,25 @@ export function debounce(this: any, func: (...args: any[]) => any, wait: number)
  * @param {HTMLElement} container  The container to check if the element is in view of.
  * @param {HTMLElement} element    The element to check if it is view
  * @param {boolean} partial   true if partial view is allowed
+ * @param {boolean} strict    true if strict mode is set, never consider the container width and element width
  *
  * @returns { boolean } True if the component is in View.
  */
-export function isElementInView(container: HTMLElement, element: HTMLElement, partial: boolean) {
+export function isElementInView(container: HTMLElement, element: HTMLElement, partial: boolean, strict?: boolean) {
```
jeffpuzzo, line 46: I'm thinking it might be better to default this value to false, so `strict: boolean = false` so we don't have to count on undefined being checked correctly down below.

jeffpuzzo:
Do we know how this `strict` param might impact `NavList` (which is also using this helper)?

Also, do we know of any use-cases where we would want `strict` to be "true" and controllable by developers via a prop?

DaoDaoNoCode:
> Do we know how this `strict` param might impact `NavList` (which is also using this helper)?

Yes, the same problem also exists on `NavList`, you could check this example https://www.patternfly.org/v4/components/navigation#horizontal-only-in-pageheader and set the width to very small. So this could also help solve the issue in `NavList`.

> Also, do we know of any use-cases where we would want `strict` to be "true" and controllable by developers via a prop?
Currently, my answer is no. But I prefer to keep this prop. If they want to check whether an element is inside a container in the future in some other situations this could help.

packages/react-core/src/helpers/util.ts
DaoDaoNoCode: Will update, thanks!

jeffpuzzo:
Thanks for clarifying!

jeffpuzzo (approved):

nicolethoen (approved):

PR #6791, opened 2022-01-13
chore: Convert examples to typescript

jeffpuzzo:
**What**: Closes:
- #6785
- #6786
- #6787
- #6788
- #6778
- #6789
- #6790

### DIFF
--- a/packages/react-core/src/components/Banner/examples/BannerBasic.tsx
+++ b/packages/react-core/src/components/Banner/examples/BannerBasic.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { Banner } from '@patternfly/react-core';
+
+export const BannerBasic: React.FunctionComponent = () => (
+  <React.Fragment>
+    <Banner>Default banner</Banner>
+    <br />
+    <Banner variant="info">Info banner</Banner>
+    <br />
+    <Banner variant="danger">Danger banner</Banner>
+    <br />
+    <Banner variant="success">Success banner</Banner>
+    <br />
+    <Banner variant="warning">Warning banner</Banner>
+  </React.Fragment>
+);
--- a/packages/react-core/src/components/Banner/examples/BannerBasic.tsx
+++ b/packages/react-core/src/components/Banner/examples/BannerBasic.tsx
@@ -1,12 +1,16 @@
 import React from 'react';
-import { Banner, Flex } from '@patternfly/react-core';
+import { Banner } from '@patternfly/react-core';
 
 export const BannerBasic: React.FunctionComponent = () => (
-  <Flex direction={{ default: 'column' }}>
+  <React.Fragment>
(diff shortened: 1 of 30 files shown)


packages/react-core/src/components/Brand/examples/BrandBasic.tsx
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import { Brand } from '@patternfly/react-core';
+
+export const BrandBasic: React.FunctionComponent = () => <Brand src="./pfLogo.svg" alt="Patternfly Logo" />;
```
nicolethoen, line 4: did it not work if you tried to import the svg? might be worth opening an issue if we have an issue with svg imports.

jeffpuzzo: We don't support importing svg files at the moment (probably a tsconfig setting, possibly webpack config related), but importing the SVG didn't seem correct to begin with because `src` takes a `string` type (implying a path to a source, not the source itself), not an `SVGElement` or `HTMLElement` type. 

I don't perceive it as an issue unless we need to import svg's somewhere. In this case we don't. Then there's the question of, do we really want to treat them as their own modules or do we just create function components that return JSX interpretations of the svg, which is another approach which wouldn't require any tsconfig/webpack changes.

jeffpuzzo: Looking into this further, the default import of an SVG element returns a string, so a module file was added to accommodate for typescript not recognizing SVG imports as modules (`assets.d.ts` - if we want to update the naming convention here, let me know), and this example was reverted back to using that imported `pfLogo`.

nicolethoen (approved):

wise-king-sullyman (approved):

packages/react-core/src/components/Banner/examples/BannerBasic.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Banner, Flex } from '@patternfly/react-core';
+
+export const BannerBasic: React.FunctionComponent = () => (
+  <Flex direction={{ default: 'column' }}>
+    <Banner>Default banner</Banner>
```
tlabaj, line 6: Usually our examples match the core examples.  For that reason, I would add back the `<br />`

tlabaj (changes requested):
Looks great! Just one little nit.

packages/react-core/src/components/Banner/examples/BannerBasic.tsx
jeffpuzzo: Done. And updated other recent instances where I was using flexbox instead of breaks and empty spaces as well.

tlabaj (approved):
Looks great!

PR #6796, opened 2022-01-14
chore(Button): Convert examples to typescript

### LINKED ISSUE #6793: Button: Convert examples to typescript

jeffpuzzo:
**What**: Closes #6793

### DIFF
--- a/packages/react-core/src/components/Button/examples/ButtonProgress.tsx
+++ b/packages/react-core/src/components/Button/examples/ButtonProgress.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { Button, Flex } from '@patternfly/react-core';
+
+export const ButtonProgress: React.FunctionComponent = () => {
+  const [isPrimaryLoading, setIsPrimaryLoading] = React.useState(true);
+  const [isSecondaryLoading, setIsSecondaryLoading] = React.useState(true);
+
+  return (
+    <Flex>
+      <Button
+        isLoading={isPrimaryLoading}
+        variant="primary"
+        onClick={() => setIsPrimaryLoading(!isPrimaryLoading)}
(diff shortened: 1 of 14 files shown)


packages/react-core/src/components/Button/examples/ButtonProgress.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [isPrimaryLoading, setIsPrimaryLoading] = React.useState(true);
+  const [isSecondaryLoading, setIsSecondaryLoading] = React.useState(true);
+
+  return (
+    <Flex>
+      <Button
+        isLoading={isPrimaryLoading}
+        variant="primary"
+        onClick={() => setIsPrimaryLoading(!isPrimaryLoading)}
+        {...(isPrimaryLoading && { spinnerAriaValueText: 'Loading' })}
```
nicolethoen, line 14: This line is causing a problem when I hit the `ts>>js` button...

jeffpuzzo: The mechanism used to convert between TS and JS seems to break when spreading props (at least for Button). Can you point me to where we do the conversion so I can take a look?

nicolethoen: with about 70% confidence, I think that is what this line is supposed to do?
https://github.com/patternfly/patternfly-org/blob/main/packages/theme-patternfly-org/scripts/md/mdx-hast-to-jsx.js#L166

nicolethoen (approved):
Works great

issues with typescript -> javascript parsing will be resolved when[ this org PR](https://github.com/patternfly/patternfly-org/pull/2816) merges. This PR does not need to be blocked.

wise-king-sullyman (approved):

PR #6800, opened 2022-01-16
fix(table): fixed action example

### LINKED ISSUE #6799: table: Action layout is broken in both the compact variant and mobile breakpoint sizes
@garrett commented on [Wed Dec 01 2021](https://github.com/patternfly/patternfly/issues/4547)

Two related issues with the tables, regarding actions and layout:

1. When `isCompact` / `pf-m-compact` is added to the table, the padding is incorrect, especially on desktop.
2. When actions are more than just the ⋮ (kebab menu) is used in the actions area, the layout is broken in mobile. (It's arguably still problematic with just the kebab menu.)

### Normal, desktop (for reference; the bugs are not visible here)

normal table

### When compact is applied

There's not enough space above and below the buttons.

Screenshot 2021-12-01 at 14-35-19 PatternFly 4 • Table

### Mobile mode

Screenshot from 2021-12-01 14-35-47

### Mobile size with grid view on, to see the layout

Screenshot from 2021-12-01 14-36-51

### Mobile mode with other text and grid view

Sometimes actions need to be a bit longer. Even still, imagine longer text, like if you changed the language to German.

Screenshot from 2021-12-01 14-47-55

I did not show compact applied in mobile mode; the same issues show up regardless.

---

This affects us in Cockpit quite a bit, as we're using this pattern (table with action button + action menu) in several places. 

Our current workaround is to not set the last class for actions (that is: treat it as a normal row) and then add a `data-label=""`. This makes the actions show up as its own row on the bottom instead of being on the top and stealing horizontal space from all the content. It also makes compact mode work with proper spacing around the actions.

---

Suggested actions (somewhere between some to all):

- Re-evaluate if the mobile layout is what we want
- Adjust the PF CSS to better handle compact mode
- Perhaps adjust the layout for mobile mode (or handle it differently)
- Fix the demo to not have these issues (as a developer could copy it as-is and inherit these two bugs)

---

@mvollmer commented on [Thu Dec 02 2021](https://github.com/patternfly/patternfly/issues/4547#issuecomment-984560815)

I like this approach best right now: https://github.com/cockpit-project/cockpit/pull/16681

Buttons are shown when the table is in "row mode", but they move into the menu in "card mode".

mattnolting:
closes #6799 

relates to: #https://github.com/patternfly/patternfly/issues/4607

Action cells should be populated either by overflow menus or single dropdowns. Extra buttons shouldn't appear there.

### DIFF
--- a/packages/react-table/src/components/Table/examples/LegacyTableActions.tsx
+++ b/packages/react-table/src/components/Table/examples/LegacyTableActions.tsx
@@ -39,23 +39,9 @@ export const LegacyTableActions: React.FunctionComponent = () => {
   // In real usage, this data would come from some external source like an API via props.
   const repositories: Repository[] = [
     { name: 'a', branches: 'two', prs: '1', workspaces: 'four', lastCommit: 'five', singleAction: 'Start' },
-    {
-      name: 'disable actions',
-      branches: 'two',
-      prs: '3',
-      workspaces: 'four',
-      lastCommit: 'five',
-      singleAction: 'Start'
-    },
+    { name: 'disable actions', branches: 'two', prs: '3', workspaces: 'four', lastCommit: 'five', singleAction: '' },
     { name: 'green actions', branches: 'two', prs: '4', workspaces: 'four', lastCommit: 'five', singleAction: 'Start' },
-    {
-      name: 'extra action props',
-      branches: 'two',
-      prs: '5',
-      workspaces: 'four',
-      lastCommit: 'five',
-      singleAction: 'Start'
-    },
+    { name: 'extra action props', branches: 'two', prs: '5', workspaces: 'four', lastCommit: 'five', singleAction: 'Start' },
     { name: 'blue actions', branches: 'two', prs: '6', workspaces: 'four', lastCommit: 'five', singleAction: 'Start' }
   ];
 
@@ -84,14 +70,14 @@ export const LegacyTableActions: React.FunctionComponent = () => {
   ];
 
   const rows: TableProps['rows'] = repositories.map(repo => {
-    const singleActionButton = {
-      title: (
+    let singleActionButton = null;
+    if (repo.singleAction !== '') {
+      singleActionButton = (
         <TableText>
           <Button variant="secondary">{repo.singleAction}</Button>
         </TableText>
-      ),
-      cellTransforms: [fitContent]
-    };
+      );
+    }
 
     const cells = [repo.name, repo.branches, repo.prs, repo.workspaces, repo.lastCommit, singleActionButton];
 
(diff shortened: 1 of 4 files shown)


packages/react-table/src/components/Table/examples/LegacyTableActions.tsx
```diff
@@ -29,18 +30,89 @@ interface Repository {
...
-    { name: 'blue actions', branches: 'two', prs: '6', workspaces: 'four', lastCommit: 'five' }
+    {
+      name: 'a',
+      branches: 'two',
+      prs: '1',
+      workspaces: 'four',
+      lastCommit: 'five',
+      empty: {
+        title: (
+          <TableText>
```
mturley, line 49: I would argue that we shouldn't have any JSX or cellTransforms in the `Repository` objects here. The intention of this array is to represent PF-agnostic data that would be coming in via props in a real app (e.g. loaded from an API).

packages/react-table/src/components/Table/examples/LegacyTableActions.tsx
```diff
@@ -64,10 +135,12 @@ export const LegacyTableActions: React.FunctionComponent = () => {
...
     'Pull requests',
     'Workspaces',
-    'Last commit'
+    'Last commit',
+    { title: '', dataLabel: 'testing', cellTransforms: [fitContent] }
   ];
+
   const rows: TableProps['rows'] = repositories.map(repo => {
-    const cells = [repo.name, repo.branches, repo.prs, repo.workspaces, repo.lastCommit];
+    const cells = [repo.name, repo.branches, repo.prs, repo.workspaces, repo.lastCommit, repo.empty];
```
mturley, line 143: Maybe instead we can just put the whole `{ title, cellTransforms }` object here in the cells array instead of `repo.empty`, since it is the same for all rows anyway?

mturley (changes requested):
Some minor feedback about organizing the code here

packages/react-table/src/components/TableComposable/examples/ComposableTableActions.tsx
```diff
@@ -25,26 +20,67 @@ interface Repository {
...
-    { name: 'p', branches: 'two', prs: 'b', workspaces: 'four', lastCommit: 'five' },
-    { name: '4', branches: '2', prs: 'b', workspaces: 'four', lastCommit: 'five' },
-    { name: '5', branches: '2', prs: 'b', workspaces: 'four', lastCommit: 'five' }
+    {
+      name: 'one',
+      branches: 'two',
+      prs: 'a',
+      workspaces: 'four',
+      lastCommit: 'five',
+      singleAction: <Button variant="secondary">Start</Button>
```
mturley, line 37: Same here, we should render the Button as part of mapping over this array to convert it to JSX, not as a property of the array itself IMO.

mturley (commented):
LGTM, with a possibly unnecessary nitpick. Looks like you need to run Prettier as well to make eslint happy.

kmcfaul (approved):

nicolethoen (approved):

mturley (approved):

PR #6801, opened 2022-01-17
Fixes #6814: Don't trigger mobx update when node dimension or position doesn't change

### LINKED ISSUE #6814: Unnecessary rerenderings in topology nodes
While debugging performance issues in our topology I found some unnecessary rerenderings in topology nodes. It looks like the `BaseNode` sets the dimension and position parameters with a new cloned object, also if the attributes don't change.

https://github.com/patternfly/patternfly-react/blob/232be76ce3bb9ee8af7b86f3b0ae3629eb65a761/packages/react-topology/src/elements/BaseNode.ts#L41-L42

https://github.com/patternfly/patternfly-react/blob/232be76ce3bb9ee8af7b86f3b0ae3629eb65a761/packages/react-topology/src/elements/BaseNode.ts#L146-L153

`setModel` only checks if the parameters are part of the model independent of the fact if they "changed".

https://github.com/patternfly/patternfly-react/blob/232be76ce3bb9ee8af7b86f3b0ae3629eb65a761/packages/react-topology/src/elements/BaseNode.ts#L254-L289

N/A

Bug

OCP, OpenShift Developer Console

We targeting performance issues with 4.10, but I'm sure older versions are also affected.

The issue is related to this bug report https://github.com/openshift/console/issues/10337, we have also different bugzilla issues related to our performance issues.

jerolimov:
**What**: Fixes #6814. Don't update the internal `this.dimensions` and `this.position` updates if the `width`, `height`, `x` and `y` attributes of the model doesn't change.

This might improve the performance of the topology a little bit. In the mobx dev tools on Firefox I found this "rerender reason" and found via logging that this case (mode.width === this.dimensions.width) was true for some rerenders. But for sure, this doesn't solve the biggest performance reason.

See also https://github.com/patternfly/patternfly-react/blob/main/packages/react-topology/src/elements/BaseNode.ts#L33-L42

**Additional issues**:
None

### DIFF
--- a/packages/react-topology/src/elements/BaseNode.ts
+++ b/packages/react-topology/src/elements/BaseNode.ts
@@ -253,16 +253,17 @@ export default class BaseNode<E extends NodeModel = NodeModel, D = any> extends 
 
   setModel(model: E): void {
     super.setModel(model);
+
     let d: Dimensions | undefined;
     let p: Point | undefined;
 
-    if ('width' in model && model.width != null) {
+    if ('width' in model && model.width != null && model.width !== this.dimensions.width) {
       if (!d) {
         d = this.dimensions.clone();
       }
       d.width = model.width;
     }
-    if ('height' in model && model.height != null) {
+    if ('height' in model && model.height != null && model.height !== this.dimensions.height) {
       if (!d) {
         d = this.dimensions.clone();
       }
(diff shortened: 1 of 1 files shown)


jeff-phillips-18 (approved):

nicolethoen:
@jerolimov could you please open an issue and link it to this PR for tracking purposes

jerolimov:
@nicolethoen I created issue #6814, but without an example.

It's not easy to create also a working sandbox example, or do you have a working sandbox example for topology? If needed I can record a debugging video in ODC that shows the difference. I verified locally also that the fix reduces the rerenderings.

nicolethoen (approved):

Issue #6109, opened 2021-08-02
Tooltip examples are not accessible on dropdown items.


nicolethoen:
Whether aria-disabled or completely enabled, tooltips are not screen reader accessible when placed on a dropdown item.

jessiehuff:
I'm wondering if this is another one of those DOM related issues where the dynamic rendering of content (including the tooltip) is creating issues for screen readers. It might be related to [this issue](https://github.com/patternfly/patternfly-react/issues/5383) as well.

PR #6807, opened 2022-01-17
docs(Dropdown): add composable dropdown demo, convert composable menu demos to typescript

### LINKED ISSUE #6016: Refactor Dropdown to use Menu Toggle and Menu

nicolethoen:
Closes #6016 

Adds the Composable Dropdown Variants demo to the Composable Menu demos


mcarrano (commented):
I think this looks good. My only question is that since this only represents four of many possible dropdown variants we support in the Dropdown component, do we need to include a note that this is only a sample of what you can do (vs. a complete list)?

jessiehuff (approved):
This looks good to me from an a11y standpoint. It does remind me of pre-existing issues that Dropdown has though such as [tooltips on dropdown items](https://github.com/patternfly/patternfly-react/issues/6109) and the [Append menu document body issue](https://github.com/patternfly/patternfly-react/issues/6824).

packages/react-core/src/demos/ComposableMenu/examples/ComposableActionsMenu.tsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import { MenuToggle, Menu, MenuList, MenuItem, MenuGroup, MenuItemAction, Popper } from '@patternfly/react-core';
+import BarsIcon from '@patternfly/react-icons/dist/esm/icons/bars-icon';
+import ClipboardIcon from '@patternfly/react-icons/dist/esm/icons/clipboard-icon';
+import CodeBranchIcon from '@patternfly/react-icons/dist/esm/icons/code-branch-icon';
+import BellIcon from '@patternfly/react-icons/dist/esm/icons/bell-icon';
+
+export const ComposableActionsMenu: React.FunctionComponent = () => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selectedItems, setSelectedItems] = React.useState([]);
```
jeffpuzzo, line 10: I think since these are IDs from what it looks like we can call this `selectedItemIds` instead, and we can add a type to the useState of `number[]`.

packages/react-core/src/demos/ComposableMenu/examples/ComposableContextSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+  Divider,
+  TextInput,
+  InputGroup,
+  Button,
+  ButtonVariant
+} from '@patternfly/react-core';
+import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
+
+export const ComposableContextSelector: React.FunctionComponent = () => {
+  const items: any[] = [
```
jeffpuzzo, line 20: Is there no type associated with these or are they meant to be `any`?

packages/react-core/src/demos/ComposableMenu/examples/ComposableActionsMenu.tsx
dlabrecq: I agree with adding a type to `React.useState([])`. Adding boolean to `React.useState(false)` seems redundant, considering this type is inferred

dlabrecq (approved):

jeffpuzzo (approved):
My comments aren't blocking. The demo looks good to me!

wise-king-sullyman (commented):
Looks solid. I have one typing nit.

nicolethoen:
I have a pending question to @mattnolting 

I’m noticing that the drilldown menu example here: https://patternfly-react-pr-6807.surge.sh/demos/composable-menu#composable-drilldown-menu has an extra scroll bar when I begin drilling down. We are setting the menu height correctly (i think) but is there some extra padding or something that isn’t being accounted for?

nicolethoen:
This is ready to go in (i think)

The issue I noted above in an existing bug with the drilldown menu - I noted it here: https://github.com/patternfly/patternfly-react/issues/6842

wise-king-sullyman (approved):
LGTM!

nicolethoen:
@mcarrano I opened an issue to build the split button that is required to implement the other dropdown variants. I am not sure if it is reasonable at this point to assume every single example will be exactly re-implemented composably... I think collectively, these demos are telling a story of how versatile our composable approach is. and people can open issues if they want to see more examples?

mcarrano:
@nicolethoen Yes, I agree. I don't think it's necessary to re-implement every example unless we are specifically asked for it.

kmcfaul (approved):
lgtm. Just need to re-add the beta flag back to the menu drilldown example.

packages/react-core/src/demos/ComposableMenu/examples/ComposableDropdwnVariants.tsx
```diff
@@ -0,0 +1,235 @@
...
+                <CaretDownIcon />
+              </span>
+            </Badge>
+          </MenuToggle>
+        );
+      case 'image':
+        return (
+          <MenuToggle
+            ref={toggleRef}
+            icon={<Avatar src={avatarImg} alt="avatar" />}
```
mcoker, line 114: It looks like we use `.pf-c-menu-toggle__image` in core for this versus `.pf-c-menu-toggle__icon`, though I don't know that we need both classes - curious if we could just use a single class for both.

Some other thoughts:
1. `.pf-c-menu-toggle__image` shouldn't be used right now since there is a bug in the spacing - the right margin for the image isn't being applied, so the image will touch the adjacent text.
2. the [dropdown image toggle image element](https://www.patternfly.org/v4/components/dropdown/html#dropdown-with-image-and-text) has some additional CSS. It applies a top/bottom margin, too, and uses `inline-flex` and `flex-shrink: 0;`. I'm not sure about the top/bottom margin, but I think the flex properties are good adds.
  * Screen Shot 2022-01-20 at 3 02 18 PM
3. the image isn't really aligned. we could apply `line-height: 1; align-self: center` to the `__image/icon` element. This is what it looks like in the PR.
  * Screen Shot 2022-01-20 at 3 10 12 PM
 
I think we could just apply all of this to a single element that holds an icon or image. wdyt @mattnolting?

mcoker (commented):
LGTM! Just one thing. I'm inclined to keep what's in this PR and we make a couple of changes in core to make the element that will hold an icon or image in the toggle work better.

kmcfaul (approved):

packages/react-core/src/demos/ComposableMenu/examples/ComposableDropdwnVariants.tsx
mattnolting: @mcoker I actually added the `flex-shrink` prop in [4617](https://github.com/patternfly/patternfly/commit/9fec1a01f3a7edc5c284698bb9f694a6d82c8bdc#diff-27c8e819fb9e49641ffc5897a7b3e04c569a4fd870c88d937a66f0d3247ffafeR532) as the image was distorted when width became constrained. I agree, we only need one container element and it should be `__image` IMO. 

>  I'm not sure about the top/bottom margin, but I think the flex properties are good adds.

We should nix top/bottom margin

> the image isn't really aligned. we could apply line-height: 1; align-self: center to the __image/icon element. This is what it looks like in the PR.

👍 👍 👍

PR #6822, opened 2022-01-18
fix(DatePicker): update empty state validation and add onblur event

### LINKED ISSUE #6806: DatePicker: Empty state validation

When a date has already been selected and you try to clear it manually, as soon as you tab away from the field the DatePicker is marking the value as invalid.
It is basically not possible to clear a date value and leave the DatePicker in a valid empty state.

I also have a related question. Why the DatePicker does not have an `onBlur` prop?
It could be useful to detect when the user is leaving the field when he's manually entering a date in order to proceed with saving it only after he's finished typing.

- open the codesandbox for the DatePicker component: https://codesandbox.io/s/wkydl
- select a date using the picker
- click on the input field and manually delete the value
- hit the `tab` key on the keyboard
- the DatePicker is now marking the empty value as invalid 

This is likely a bug. I don't think there's a work-around at the moment.

DaoDaoNoCode:
**What**: Closes #6806 
1. Set the state of the date picker to pristine when the value is empty, so it will not trigger the validation check.
2. Add `onBlur` prop on the input so that users could define their own events when the input loses focus.
3. Fix the `displayName` of the component to make the props shown in the doc.

### DIFF
--- a/packages/react-core/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-core/src/components/DatePicker/DatePicker.tsx
@@ -30,6 +30,8 @@ export interface DatePickerProps
   invalidFormatText?: string;
   /** Callback called every time the input value changes */
   onChange?: (value: string, date?: Date) => void;
+  /** Callback called every time the input loses focus */
+  onBlur?: (value: string, date?: Date) => void;
   /** Text for label */
   helperText?: React.ReactNode;
   /** Aria label for the button to open the date picker */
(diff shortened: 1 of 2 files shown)


nicolethoen (commented):
This looks great! maybe it's worth updating one of the examples - maybe even the first one to use the new onBlur event rather than the onChange event to write to the console? or use both in the example?

DaoDaoNoCode:
@nicolethoen Makes sense, I will update the first example to use both `onBlur` and `onChange`!

nicolethoen (approved):

kmcfaul (approved):

PR #6825, opened 2022-01-18
feat(TimePicker): auto-set direction based on space

### LINKED ISSUE #6616: Timer-picker: Date-picker: direction should be auto-set according to the available space
Currently the component offers the direction property. We (cockpit) want to have the default 'down' direction, however, in  mobile mode, there is sometimes not space down. In this case we would like that the component automatically flips with the menu up. 

https://www.patternfly.org/v4/components/time-picker/

Enhancement

Cockpit

thatblindgeye:
**What**: Closes #6616  

@KKoukiou Can you let me know whether this resolves the issue you were facing with the TimePicker component? The DatePicker component looks to already have this auto-set functionality so I have not touched it for this.

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -3,9 +3,10 @@ import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Select/select';
 import datePickerStyles from '@patternfly/react-styles/css/components/DatePicker/date-picker';
 import formStyles from '@patternfly/react-styles/css/components/FormControl/form-control';
+import menuStyles from '@patternfly/react-styles/css/components/Menu/menu';
 import { getUniqueId } from '../../helpers';
 import { Popper } from '../../helpers/Popper/Popper';
-import { TimeOption } from './TimeOption';
+import { Menu, MenuContent, MenuList, MenuItem } from '../Menu';
 import { KeyTypes, SelectDirection } from '../Select';
 import { InputGroup } from '../InputGroup';
 import { TextInput, TextInputProps } from '../TextInput';
@@ -86,7 +87,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
   private parentRef = React.createRef<HTMLDivElement>();
   private toggleRef = React.createRef<HTMLDivElement>();
   private inputRef = React.createRef<HTMLInputElement>();
-  private menuRef = React.createRef<HTMLUListElement>();
+  private menuRef = React.createRef<HTMLDivElement>();
 
   static defaultProps = {
     className: '',
@@ -98,7 +99,6 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     placeholder: 'hh:mm',
     delimiter: ':',
     'aria-label': 'Time picker',
-    menuAppendTo: 'inline',
     direction: 'down',
     width: 150,
     stepMinutes: 30,
@@ -170,7 +170,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
           this.onToggle(false);
         } else if (event.key === KeyTypes.Enter) {
           if (focusedIndex !== null) {
-            this.onSelect((this.getOptions()[focusedIndex] as HTMLElement).innerText);
+            this.onSelect(event);
             event.stopPropagation();
           } else {
             this.onToggle(false);
@@ -221,7 +221,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
   };
 
   scrollToIndex = (index: number) => {
-    this.getOptions()[index].offsetParent.scrollTop = this.getOptions()[index].offsetTop;
+    this.getOptions()[index].closest(`.${menuStyles.menuContent}`).scrollTop = this.getOptions()[index].offsetTop;
   };
 
   scrollToSelection = (time: string) => {
@@ -289,7 +289,9 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
   };
 
   getOptions = () =>
-    (this.menuRef && this.menuRef.current ? Array.from(this.menuRef.current.children) : []) as HTMLElement[];
+    (this.menuRef && this.menuRef.current
+      ? Array.from(this.menuRef.current.querySelectorAll(`.${menuStyles.menuListItem}`))
+      : []) as HTMLElement[];
 
   isValidFormat = (time: string) => {
     if (this.props.validateTime) {
@@ -324,11 +326,11 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     });
   };
 
-  onSelect = (selection: string) => {
-    const { timeRegex, timeState } = this.state;
+  onSelect = (e: any) => {
+    const { timeRegex, timeState, focusedIndex } = this.state;
     const { delimiter, is24Hour, includeSeconds } = this.props;
-    const time = parseTime(selection, timeRegex, delimiter, !is24Hour, includeSeconds);
-
+    const elementText = e.type === 'click' ? e.target.textContent : this.getOptions()[focusedIndex].innerText;
+    const time = parseTime(elementText, timeRegex, delimiter, !is24Hour, includeSeconds);
     if (time !== timeState) {
       this.onInputChange(time);
     }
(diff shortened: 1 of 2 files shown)


KKoukiou:
This fixes indeed the issue I opened. Thanks :)

jessiehuff (commented):
I'm not sure which aspect affected it, but it doesn't work for me anymore with VO. :(

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -453,31 +480,12 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
...
           inputAndToggle
         ) : (
-          <Popper
-            trigger={inputAndToggle}
-            popper={popperContainer}
-            direction={direction}
-            appendTo={menuAppendTo}
-            isVisible={isOpen}
-          />
+          <Popper trigger={inputAndToggle} popper={menuContainer} isVisible={isOpen} />
```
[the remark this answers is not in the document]
thatblindgeye: I tried a couple of variations in regards to swapping out the entire `inputAndToggle` which result in the error text not visually appearing (best I got was the red exclamation point icon to the right of the input). I can look at the Popper component some more and see if anything else might work, though.

thatblindgeye:
@jessiehuff On the preview page, could you try passing in the `menuAppendTo="inline"` prop and value to one of the example TimePickers? The fix to get this component to auto-set the direction was to change the default behavior, so instead of being inline it now gets appeneded towards the bottom of the document body, so I'm wondering if that is what is causing the issue with VO.

thatblindgeye:
Per discussion with @jessiehuff , the issue with VO not working with the component is due to the already-known issue when elements are appended to the bottom of the document body. When the `menuAppendTo="inline"` prop is passed in VO picks up the menu contents.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -324,15 +354,15 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     });
   };
 
-  onSelect = (selection: string) => {
+  onSelect = (e: any) => {
```
nicolethoen, line 357: I think this type needs to be more specific.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -221,7 +226,27 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
...
+
+    this.getOptions()[index].closest(`.${menuStyles.menuContent}`).scrollTop = this.getOptions()[
+      indexToScroll
+    ].offsetTop;
+  };
+
+  focusSelection = (index: number) => {
+    const indexToFocus = index !== -1 ? index : 0;
+
+    if (this.menuRef && this.menuRef.current) {
```
jeffpuzzo, line 247: could simplify to `this.menuRef?.current`

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -221,7 +226,27 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
   };
 
   scrollToIndex = (index: number) => {
-    this.getOptions()[index].offsetParent.scrollTop = this.getOptions()[index].offsetTop;
+    // fixes issue where inline value for menutAppendTo results in initial menu item that is scrolled to being out of view
```
jeffpuzzo, line 229: I think explaining what is being done to fix the issue would help on top of calling out what issue is being fixed.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -233,6 +258,10 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     if (splitTime.length < 2) {
       time = `${time}${delimiter}00`;
       splitTime = time.split(delimiter);
+      // build time without seconds when includeSeconds is true
```
jeffpuzzo, line 261: Did you mean to write when includeSeconds is *false?

thatblindgeye: So there was an issue when seconds is included in the input and a user would try navigating to the menu. What _should_ happen is they're brought to the exact or the closest time (2:30:00 in the input should bring them to 2:30 in the menu of options). But since the contents of the menu doesn't include seconds, the user would always be brought to the first item in the list (12:00 AM).

Building that temporary time _without_ the seconds, even when they're included as a prop, basically just makes the scrolling work as intended. I will update the comment to reflect the reasoning for building the time without seconds, even when the includeSeconds is true.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -221,7 +226,27 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
...
-    this.getOptions()[index].offsetParent.scrollTop = this.getOptions()[index].offsetTop;
+    // fixes issue where inline value for menutAppendTo results in initial menu item that is scrolled to being out of view
+    let indexToScroll = index;
+    if (this.props.menuAppendTo === 'inline') {
+      if (indexToScroll > 0) {
+        indexToScroll = index - 1;
+      } else {
+        indexToScroll = 0;
+      }
+    }
```
wise-king-sullyman, line 237: Nit: I think this would be a bit cleaner/simpler with a nonmutation implementation that puts this logic in its own function rather than having a nested if/else that mutates a variable.

wise-king-sullyman (commented):
Seems to work perfectly, just have a couple of small comments.

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -401,107 +430,68 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
...
-          <div className={css(datePickerStyles.datePickerHelperText, datePickerStyles.modifiers.error)}>
-            {!isValidFormat ? invalidFormatErrorMessage : invalidMinMaxErrorMessage}
-          </div>
-        )}
-      </div>
+      <Menu ref={this.menuRef} isScrollable>
+        <MenuContent maxMenuHeight="200px">
+          <MenuList aria-labelledby={`${randomId}-input`}>
+            {options.map((option, index) => (
+              <MenuItem onClick={this.onSelect} key={index} id={`${randomId}-option-${index}`}>
```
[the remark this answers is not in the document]
wise-king-sullyman: Per our conversation on Slack this is probably fine since options `shouldn't` change; marking as resolved.

packages/react-core/src/components/TimePicker/TimePicker.tsx
nicolethoen: I was wrong! This is the event passed to MenuItem, which has an `onClick` that uses `e: any`. This needs to stay the way it is! :)

wise-king-sullyman (approved):
Looks 🤌

jeffpuzzo:
Noticing that this menu overlaps the page header when scrolling downward. It does not do this with the current implementation so this might be an issue.

image

packages/react-core/src/components/TimePicker/TimePicker.tsx
jeffpuzzo: Makes sense. Thanks for the explanation!

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -401,107 +423,68 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
...
```
jeffpuzzo, line 437: Using the randomId or the whole string used for `id` (could be stored as a constant and used in both places) would be more reliable as the value for `key`.

jeffpuzzo:
> Noticing that this menu overlaps the page header when scrolling downward. It does not do this with the current implementation so this might be an issue.
> 
> image
> 
> image

In a slack conversation it was pointed out that the Application Launcher also has this problem, so I won't consider it blocking, but 1 thing I think is different, at least I'm unable to recreate it with the Application Launcher is the following when condensing and expanding the window size with the side panel open: 
image

jeffpuzzo:
Spoke with Eric in slack and he was able to recreate the same sidebar issue mentioned here for the application launcher https://github.com/patternfly/patternfly-react/pull/6825#issuecomment-1018613444 (last screenshot); so I won't consider it blocking.

jeffpuzzo (approved):

jessiehuff (approved):
It works in VO again! 😄  I'm wondering if we should also create a separate issue for figuring out how to handle these scenarios when someone appends to the document body?

nicolethoen (approved):

Issue #6030, opened 2021-07-14
Masthead / Toolbar responsiveness issues


jschuler:
Currently the masthead and toolbar set media queries that work on the whole window width to determine when to hide toolbar items behind an overflow button, etc
This is problematic when you have something like a drawer within the window that takes space away from the masthead/toolbar. The drawer should cause the masthead/toolbar to respond sooner.
You can reproduce the issue for example at
https://cloud.redhat.com/beta/application-services/streams/resources?
by opening up a quick start, then resizing the window (or just resizing the drawer).

A possible approach to tackle this might be to add a resize observer to listen to the parent container width and not the browser width to determine when to respond.

redallen:
Maybe solvable with css-grid? https://moderncss.dev/container-query-solutions-with-css-grid-and-flexbox/#solutions

jschuler:
@mattnolting

mattnolting:
TLDR; @redallen `minmax()` in `flex` or `grid` would work if the request were to stack items when they run out of room (that's what the [Gallery Layout](https://www.patternfly.org/v4/layouts/gallery/react/adjusting-max-widths/) does). Using grid `auto-fill/fit` properties would make all direct descendants the same width, with is not what we would want for a flexible layout such as the toolbar. Using `display: flex` + `minmax()` would tell the items to wrap when they reached a minimum width. Unfortunately, the `minmax()` property can't tell the items when to `hide/show` or change variable definitions, it only affects when items wrap. 

Until we have container queries, we have to rely on our current breakpoint based css or an element (toolbar) resize observer. IMO, we should use element resize observers very sparingly, but if there were one place to use them in a page, it would be the toolbar.

redallen:
@mattnolting What if a "PageDrawer" or similar were included in our current css-grid for `pf-c-page`? Would that work?

mattnolting:
@redallen We already have `pf-c-page__drawer`. It's possible to set a flag on `.pf-c-page__drawer > .pf-c-drawer.pf-m-expanded` the shows/hides toolbar things, but the drawer could be open on a very wide screen and alter toolbar elements undesirably. Honestly, I think we should add a resize observer, then supplant when we have browser support for container queries.

redallen:
I believe you, sounds good!

mcarrano:
Per discussion on Dec 15, I am adding this to the upcoming milestone. Seems like consensus is to add a resize observer.

PR #6827, opened 2022-01-18
Page resize observer & Toolbar/PageHeaderTools responsiveness based on page width

jschuler:
**What**: Towards https://github.com/patternfly/patternfly-react/issues/6030

- Added a page resize observer to the Page component
- Adds **width**, and **getBreakpoint** to the `PageContext`. The width is the width of the Page (pf-c-page) container. Children components can use the context to base logic around that (like the Toolbar components in this PR). The **getBreakpoint** function defaults to the one in `util.ts`, but can be overwritten if desired.
- Adds a `pf-m-breakpoint-[default|sm|md|lg|xl|2xl]` class as a sibling to the Page (pf-c-page) container. This allows for CSS to target page based breakpoints [as described in this issue](https://github.com/patternfly/patternfly/issues/4622)

A new util function `getBreakpoint` was added to return the breakpoint for a given width.

The `util.ts` **formatBreakpointMods** function was updated to accept a `breakpoint` argument.
Before, you would have a string of class names that are media query based, i.e. `pf-m-hidden pf-m-visible-on-md pf-m-hidden-on-lg`, or `pf-m-spacer-none pf-m-spacer-md-on-md`, etc. If the `breakpoint` argument is supplied, we instead return now a single class based on the known width.

To test this, I added a lightbulb icon button to the masthead which toggles a drawer in this demo:
https://patternfly-react-pr-6827.surge.sh/components/page/react-demos/sticky-section-group/

**Additional issues/PRs**: https://github.com/patternfly/patternfly/issues/4622, https://github.com/patternfly/patternfly/pull/4625
**TBD**: More PRs to follow to update components that use the **formatBreakpointMods** function, or set breakpoint based modifier classes

Demo (note, this also contains the change from core here https://github.com/patternfly/patternfly/pull/4625 which is not yet merged):

https://user-images.githubusercontent.com/869106/150210012-f693ab20-2565-4470-bff1-fbcd81622862.mov


packages/react-core/src/components/Page/Page.tsx
```diff
@@ -110,15 +135,20 @@ export class Page extends React.Component<PageProps, PageState> {
...
-    const { isManagedSidebar, onPageResize } = this.props;
-    if (isManagedSidebar || onPageResize) {
+    const { isManagedSidebar, onPageResize, useResizeObserver } = this.props;
+    if (isManagedSidebar || onPageResize || useResizeObserver) {
       if (canUseDOM) {
-        window.addEventListener('resize', this.handleResize);
+        if (useResizeObserver) {
+          this.observer = getResizeObserver(this.pageRef.current, this.handleResize);
+        } else {
+          window.addEventListener('resize', this.handleResize);
```
dlabrecq, line 150: Curious why we offer `useResizeObserver` as a choice? 

Note that `getResizeObserver` will use a window event listener if `containerRefElement` is not available.

jschuler: hmm, I suppose we can just enable the observer by default without providing a prop. And will update the code to remove the extra listener, thanks!

dlabrecq (approved):

packages/react-core/src/helpers/util.ts
```diff
@@ -262,32 +262,84 @@ export const setBreakpointCssVars = (
...
-    '2xl'?: string;
-    '3xl'?: string;
-  },
+  mods: Mods,
   styles: any,
-  stylePrefix: string = ''
-) =>
-  Object.entries(mods || {})
+  stylePrefix: string = '',
+  breakpoint?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl'
```
mattnolting, line 285: My only feedback, I'd like to see `xs` -> `default` until we have an actual value for `xs`.

mattnolting (approved):
Fantastic work @jschuler! L🤖 TM!

mattnolting (approved):
👏👏👏  Love it

tlabaj (approved):
LGTM

nicolethoen (commented):
@jschuler I think that at least the stacked toolbar example is behaving differently after this update. Can you just verify that the [toolbar examples](https://patternfly-react-pr-6827.surge.sh/components/toolbar) are all behaving as you'd expect with the updates?

The stacked example is now collapsing the ToolbarToggleGroup at full page width - it didn't do that before.

packages/react-core/src/helpers/util.ts
```diff
@@ -318,7 +318,7 @@ export const formatBreakpointMods = (
  */
 export const getBreakpoint = (width: number): 'default' | 'sm' | 'md' | 'lg' | 'xl' | '2xl' => {
   if (width === null) {
-    width = canUseDOM ? window.innerWidth : 1200;
+    return null;
```
nicolethoen, line 321: This does cause a classname to get applied - `pf-m-breakpoint-null` so I feel like that's not quite ideal?

jschuler: thanks, I added a check in the page component to make sure width is defined first

nicolethoen (approved):

PR #6832, opened 2022-01-19
Add icon only variant for progress button

### LINKED ISSUE #6782: Add "Icon only" variant for progress button
Want to support the ability to change an icon button to a spinner to show progress. See core PR: https://github.com/patternfly/patternfly/pull/4594

boaz0:
**What**: Closes #6782

### DIFF
--- a/packages/react-core/src/components/Button/Button.tsx
+++ b/packages/react-core/src/components/Button/Button.tsx
@@ -131,7 +131,7 @@ const ButtonBase: React.FunctionComponent<ButtonProps> = ({
         isActive && styles.modifiers.active,
         isInline && variant === ButtonVariant.link && styles.modifiers.inline,
         isDanger && (variant === ButtonVariant.secondary || variant === ButtonVariant.link) && styles.modifiers.danger,
-        isLoading !== null && styles.modifiers.progress,
+        isLoading !== null && children !== null && styles.modifiers.progress,
         isLoading && styles.modifiers.inProgress,
         isSmall && styles.modifiers.small,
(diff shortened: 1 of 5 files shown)


kmcfaul:
Could the logic for `isIconOnly` be executed instead by checking whether `children` was null and so not require a new flag? `children` currently shouldn't be rendering when not passed in.

boaz0:
@kmcfaul sure! updated the PR. Thanks.

boaz0:
Seems like the CI failure isn't related to the change.

mcarrano (commented):
Just a couple of comments...
* It might be my imagination, but it feels like the spinner and the icon for the button do not exactly have the same x-y position. It feels like the spinner appears offset from the left of the icon.
* Also, I'm wondering if this should be broken into a separate example. If I just looked at this page, I doubt that I would understand that the spinner is really a plain button.

mcoker (changes requested):
For this variant, we leave the button's icon in the button when it's in progress, but we hide it using `opacity`, which hides it visually but leaves its layout in the page. That will retain the button's original shape when the spinner is shown. Looks like we're removing the icon when the icon's in progress in this PR.

boaz0:
@mcoker & @mcarrano thanks for your review. I see what you were talking about and am going to update the PR.

boaz0:
DONE :+1:

mcarrano (approved):
Looks great. Thanks @boaz0 !

tlabaj (approved):

mmenestr (approved):
LGTM!

PR #6834, opened 2022-01-19
chore(ApplicationLauncher): Convert examples to typescript

### LINKED ISSUE #6797: Application launcher: Convert examples to typescript

jeffpuzzo:
**What**: Closes #6797


packages/react-core/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
jeffpuzzo, line 1: Reason for this is mentioned here: https://github.com/patternfly/patternfly-react/pull/6791#discussion_r787988093

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncherBasic.tsx
```diff
@@ -0,0 +1,23 @@
...
+  <ApplicationLauncherItem key="disabled_application_4a" isDisabled>
+    Unavailable application
+  </ApplicationLauncherItem>
+];
+
+export const ApplicationLauncherBasic: React.FunctionComponent = () => {
+  const [isOpen, setIsOpen] = React.useState<boolean>(false);
+
+  const onToggle = React.useCallback((isOpen: boolean) => setIsOpen(isOpen), []);
+  const onSelect = React.useCallback((_event: any) => setIsOpen(prevIsOpen => !prevIsOpen), []);
```
wise-king-sullyman, line 20: I'm curious, what is the reasoning for the `setIsOpen(prevIsOpen => !prevIsOpen)` notation vs just negating the current state variable directly? Is it a requirement of using `useCallback` or just a style preference?

jeffpuzzo: It is not a requirement. You can do either, both work fine, but since the function is memoized, we would have to add the isOpen state variable as a dependency, so utilizing the ability to access previous state from the setter is a little less code and a little cleaner.

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncher.md
```diff
@@ -17,616 +17,50 @@ To add a tooltip, use the `tooltip` prop and optionally add more tooltip props b
...
-        menuAppendTo={() => document.body}
-        onSelect={this.onSelect}
-        onToggle={this.onToggle}
-        isOpen={isOpen}
-        items={appLauncherItems}
-      />
-    );
-  }
-}
+```ts file="./ApplicationLauncherMenu.tsx"
```
wise-king-sullyman, line 65: I see you've got some notable differences here between example names and component/file names. I can't remember if the team came to an alignment on this topic at the last meeting, unfortunately, but if not I feel like we should. I know that @thatblindgeye and I proposed a standard of `component name` + `example name` that matches 1:1, and I recall some debate on whether the name of the component was necessary but that's all I can recall at the moment.

I do understand (especially with the context of this example name) why you made the decisions you did though, and if we want to just standardize on something looser like `component name` + something that uniquely identifies the example I don't think I'd be opposed to that.

wise-king-sullyman (commented):
Neither of these are major blockers for me, just topics I'd like to discuss before this PR gets merged:

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncherBasic.tsx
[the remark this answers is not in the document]
jeffpuzzo: That's a great article for sure and a good discussion to have. I think in this case, you're right in that they're not needed and even though there's a possibility of these functions being used for side-effects that would require referential equality checks, probably best to not make that assumption so I'll update.

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncher.md
jeffpuzzo: Oh I see, I didn't realize we had said exactly the name of the example, versus something that resembles it (this leads to fewer characters in most cases), which is what I was aiming for. I'm fine with re-opening the discussion.

wise-king-sullyman: I know that was our proposal, but I can't recall if there was a consensus reached on it and regretfully I didn't take notes. I do think some examples (like the one for the file I pointed out) have names that just don't work well for this purpose.

@thatblindgeye do you have any thoughts?

thatblindgeye: I think componentName[exampleName] can work most of the time, but falling back to componentName[shortenedExampleName] would be good to have especially when it comes to something like "Basic with menu appended to document body". I think it's really just going to have to be a "use your best judgment" thing on which one to use and which best describes the example itself (in this instance, for example, I would say `ApplicationLauncherDocumentBody` or similar might be a more fitting name).

We could tweak the example name to something like "Menu appended to document body", but that would still result in a rather lengthy file/example name of `ApplicationLauncherMenuAppendedToDocumentBody` (give or take a few characters if this is shortened slightly further).

If we'd want the verbiage written down anywhere to actually reference I can write something a little more formal up as well, or help work on that.

nicolethoen (approved):

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncherMenu.tsx
```diff
@@ -0,0 +1,31 @@
...
+  </ApplicationLauncherItem>,
+  <ApplicationLauncherItem key="application_2a" component="button" onClick={() => alert('Clicked item 2')}>
+    Application 2 (button with onClick)
+  </ApplicationLauncherItem>,
+  <ApplicationLauncherItem key="disabled_application_4a" isDisabled>
+    Unavailable application
+  </ApplicationLauncherItem>
+];
+
+export const ApplicationLauncherMenu: React.FunctionComponent = () => {
```
thatblindgeye, line 16: Based on the conversation regarding naming, I'm wondering if this name should instead revert back to ApplicationLauncherDocumentBody, or if that does stray into "too long" territory.

wise-king-sullyman (approved):
At the risk of being excessively pedantic: `aligned right`, `aligned top`, and some of the `with` examples don't match the name of the examples 1:1 despite their lengths not being excessive if they did IMO.

These are just small nits though and "too long" is, of course, subjective so none of these are blockers for me. I just didn't feel right not pointing it out when I noticed it since it doesn't quite match up with my understanding of the consensus we reached on this topic earlier today.

nicolethoen (approved):

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncher.md
kmcfaul: I agree that shortening the example part of the name would be good to keep the overall length of the filename down. What do you think of abbreviating the component name where applicable as well (AppLaunch or AppLauncher in this case), or dropping the component name section altogether because it isn't directly user displayed and can be inferred from the file structure?

kmcfaul (approved):

packages/react-core/src/components/ApplicationLauncher/examples/ApplicationLauncher.md
jeffpuzzo: I had suggested the same as I thought the name of the component was redundant, but I don't recall there being much of a reception to that when discussed in standup a week ago.

thatblindgeye (approved):

Issue #6741, opened 2022-01-05
regression: TypeAhead select now has a browser-based pop-up


garrett:
In Cockpit, the TypeAhead select now has a browser-based pop-up on top of the PatternFly typeahead. (This also affects every project using the TypeAhead of PatternFly React. It's not Cockpit-specific.)

This is due to the PR https://github.com/patternfly/patternfly-react/pull/6623 landing which actually _causes_ the bug that https://github.com/patternfly/patternfly-react/issues/6625 talks about.

There is no "never" value for the autocomplete attribute. The correct value is "off" in almost every case. The only other value that makes sense is "new-password" for when an using typeahead for user names, else a password manager (including the built-in browser-based password managers) may show a pop-up.

Setting it to "never" means that it will autocomplete anything with a "never" attribute. This usually will result in no secondary popup on the first time, but a secondary popup is guaranteed on additional uses.

garrett:
I left some details @ https://github.com/patternfly/patternfly-react/pull/6623#issuecomment-1005586144; quoting here:

> This change specifically makes it so we _do_ experience browser autocomplete on top of PatternFly autocomplete. In other words, this change is a breaking one that actually does the opposite of what is intended.
> 
> There's no "never" attribute listed on any documentation or spec pages, but there is an "off", which is suggest to be used in cases where autocomplete is not wanted. (The exception is user/pass forms where password managers still work with "off", but this isn't for autocomplete in most cases. In that specific case, [it's suggested to use "new-password"](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields)... and this wouldn't be case for the typeahead select, _unless_ it's used for user selection. And it usually isn't used for user selection.)
> 
> - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
>   > "off"
>   >
>   > The browser is not permitted to automatically enter or select a value for this field. It is possible that the document or application provides its own autocomplete feature, or that security concerns require that the field's value not be automatically entered.
> - https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill
>   > When wearing the autofill expectation mantle, the autocomplete attribute, if specified, must have a value that is an ordered set of space-separated tokens consisting of either a single token that is an ASCII case-insensitive match for the string "off", or a single token that is an ASCII case-insensitive match for the string "on", or autofill detail tokens.

[...]

PR #6845, opened 2022-01-21
Revert "Fix for no autocomplete for Typeahead Select"

tlabaj:
Reverts patternfly/patternfly-react#6623

A regression was caused from this PR.  #6741

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -953,7 +953,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                 }
               }}
               ref={this.filterRef}
-              autoComplete="never"
+              autoComplete="off"
             />
           </div>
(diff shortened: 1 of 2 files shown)


kmcfaul (approved):

jeff-phillips-18:
Running the surge site above: https://patternfly-react-pr-6845.surge.sh/components/select , I get:

image

This shows the secondary (browser) popup. If you run from https://www.patternfly.org/v4/components/select you don't see it.

kmcfaul:
Chrome is very insistent about its autofill and can ignore input and form `autoComplete="off"`, unless autofill is turned off in the browser settings. Perhaps we can allow the user to customize the autoComplete attribute through a prop whose default is set to "off"? That way users can work around as needed by setting other values for autoComplete?

tlabaj:
will be fixed in https://github.com/patternfly/patternfly-react/pull/6846

PR #6846, opened 2022-01-21
fix(Select): change typeahead autocomplete when input is user

kmcfaul:
**What**: Closes second part of #6741 

When typeahead input value is "user" then the auto-complete attribute on the input field should be set to "new-user".


tlabaj (approved):

tlabaj (approved):

jeff-phillips-18:
Like https://github.com/patternfly/patternfly-react/pull/6845, this also shows the secondary (browser based) autofill:

image

tlabaj (approved):

jeff-phillips-18:
Can we just use `autoComplete="new-password"`? `off` still shows the auto-complete

kmcfaul:
@jeff-phillips-18 @garrett

Updated the PR a bit to allow customization of the autoComplete prop. It will default to "off".

Garrett, am I'm interpreting your request to change the autoComplete to "new-user" correctly in the code? I'm not 100% if you meant when the input value was "user" or something else.

jeff-phillips-18 (approved):
I'm ok with this fix as it allows the application to override if necessary for them.

kmcfaul:
@garrett Thanks for clarifying! I've updated the code so the default value for autoComplete will either be `pf-no-autocomplete-{random}` when the browser is detected as Chrome and `off` otherwise. This value may still be overridden by the user through the `inputAutoComplete` property as necessary (such as when the input is used for user/passwords so autoComplete should be set to `new-password`).

PR #6847, opened 2022-01-21
fix(CodeEditor): compare props and state to prevent re-renders

### LINKED ISSUE #6501: CodeEditor resets undo/redo stack on external change
The undo(CMD+Z) /redo (Shift+CMD+Z) functionality of CodeEditor breaks when the value to the editor is updated externally

Example:
```
import React from 'react';
import { CodeEditor, Language } from '@patternfly/react-code-editor';

class BasicCodeEditor extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      value: "Some example content"
    };
    
    this.onEditorDidMount = (editor, monaco) => {
      console.log(editor.getValue());
      editor.layout();
      editor.focus();
      monaco.editor.getModels()[0].updateOptions({ tabSize: 5 });
    };
    
    this.onChange = value => {
      console.log(value);
      this.setState({
        value
      })
    };
  }
  
  render() {
    const { value } = this.state;
    
    return (
      <>
        <Button onClick={() => this.setState({ value: `${value} some change` })}>Change code</Button>
        <CodeEditor
          code={value}
          onChange={this.onChange}
          language={Language.javascript}
          onEditorDidMount={this.onEditorDidMount}
          height='400px'
        />
      </>
    );
  }
}
```

DaoDaoNoCode:
**What**: Closes #6501 

The reason that causes this issue is that when the user defines their own `onChange` function, they will call `setState` to set the `value`, and the `value` will be passed as a prop to the CodeEditor. At the same time, in the CodeEditor component, it will first call `componentDidUpdate` to set the code to the new value, then it will also call the `onChange` function inside the CodeEditor component, where it will use `this.setState` (https://github.com/patternfly/patternfly-react/blob/main/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx#L279) to set the changed value to the state even the value is already changed, which causes the unnecessary re-renders (Unlike functional component, every time `this.setState` is called, the component will be re-rendered no matter whether the state changed or not).

The way to fix it is to override the `getDerivedStateFromProps` function and check the `prop.code` and `state.value` and don't re-render the component if they are the same. In this way, it will not re-render many times when the `onChange` function is being defined by the users.

### DIFF
--- a/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
@@ -182,6 +182,7 @@ export interface CodeEditorProps extends Omit<React.HTMLProps<HTMLDivElement>, '
 }
 
 interface CodeEditorState {
+  prevPropsCode: string;
   value: string;
   filename: string;
   isLoading: boolean;
@@ -264,6 +265,7 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
   constructor(props: CodeEditorProps) {
     super(props);
     this.state = {
+      prevPropsCode: this.props.code,
       value: this.props.code,
       filename: '',
       isLoading: false,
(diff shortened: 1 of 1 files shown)


jschuler (approved):
Thanks for the fix! I tested it out and it works well now

gitdallas (changes requested):
This is breaking the first example in CodeEditor.md or basically anywhere we aren't holding the value in a state updated onChange and passed into the codeEditor component

DaoDaoNoCode:
@gitdallas @nicolethoen I didn't really understand the function `getDerivedStateFromProps` before. It's actually just updating the state based on the props, and it will still re-render even if there is no state update. The way we do this is to move what should be done in `componentDidUpdate` into that function so it will only set the state once when the props change. What's more, I added a prop named `prevPropsCode` to record the previous code passed by the props, so we can decide whether the value is updated based on the props code or not.
No need to change the doc examples now and it should work well.

gitdallas (approved):

tlabaj (approved):

PR #6848, opened 2022-01-21
fix(Select): Do not filter out view more button on filter

### LINKED ISSUE #6693: Bug: View more will disappear when using onFilter of Select with typeahead variant

I think put view more as an option is not a good way.
It's better to move out of the  options `<ul>` . maybe could add a wrapper for it.
The view more button will always on the bottom, so move it out of the list will be easy to do
When options changed , it won't be impact any more.

```
<Select
  aria-label="Select Input"
  onToggle={this.onToggle}
  onSelect={this.onSelect}
  selections={selected}
  onFilter={this.customFilter}
  isOpen={isOpen}
  aria-labelledby={titleId}
  {...(!isLoading && numOptions < this.options.length && { loadingVariant: { text: 'View more', onClick: this.onViewMoreClick } })}
  {...(isLoading && { loadingVariant: 'spinner' })}
>
  {this.options.slice(0, numOptions)}
</Select>
```

https://codesandbox.io/s/dreamy-jackson-9z5qo?file=/index.js

* SelectVariant.typeahead
* View More as LoadingVariant
* with `onFilter`

tlabaj:
**What**: Closes #6693

To test change the variant of the view more variant to "typeahead"

Also fixed focus issue and keyboard regression.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -460,9 +460,6 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
     }
     if (!typeaheadFilteredChildren) {
       typeaheadFilteredChildren = [];
-    } else if (typeaheadFilteredChildren.length === 1 && loadingVariant !== undefined) {
-      // if view more button is only item, set typeaheadFilteredChildren to am empty array
-      typeaheadFilteredChildren = [];
     }
     if (typeaheadFilteredChildren.length === 0) {
       !isCreatable &&
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -443,6 +443,11 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   return true;
                 }
 
+                // spinner should be returned as not a match
+                if (loadingVariant === 'spinner' && valueToCheck === 'loading') {
+                  return true;
+                }
+
                 if (isSelectOptionObject) {
                   return (valueToCheck as SelectOptionObject).compareTo(typeaheadInputValue);
                 } else {
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -441,10 +441,8 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   (valueToCheck as SelectOptionObject).compareTo;
 
                 // View more option should be returned as not a match
-                if (loadingVariant !== undefined && loadingVariant !== 'spinner') {
-                  if (loadingVariant?.text === valueToCheck) {
-                    return true;
-                  }
+                if (loadingVariant !== 'spinner' && loadingVariant?.text === valueToCheck) {
+                  return true;
                 }
 
                 if (isSelectOptionObject) {
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -350,7 +350,8 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
         typeaheadFilteredChildren: React.Children.toArray(this.props.children)
       }),
       typeaheadCurrIndex: -1,
-      tabbedIntoFavoritesMenu: false
+      tabbedIntoFavoritesMenu: false,
+      viewMoreNextIndex: -1
     });
   };
 
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -385,7 +385,8 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       noResultsFoundText,
       children,
       isGrouped,
-      isCreateSelectOptionObject
+      isCreateSelectOptionObject,
+      loadingVariant
     } = this.props;
 
     if (onFilter) {
@@ -439,6 +440,13 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   (valueToCheck as SelectOptionObject).toString &&
                   (valueToCheck as SelectOptionObject).compareTo;
 
+                // View more option should be returned as not a match
+                if (loadingVariant !== undefined && loadingVariant !== 'spinner') {
+                  if (loadingVariant?.text === valueToCheck && loadingVariant.text === valueToCheck) {
+                    return true;
+                  }
+                }
+
                 if (isSelectOptionObject) {
                   return (valueToCheck as SelectOptionObject).compareTo(typeaheadInputValue);
                 } else {
@@ -449,6 +457,9 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       }
     }
     if (!typeaheadFilteredChildren) {
+      typeaheadFilteredChildren = [];
+    } else if (typeaheadFilteredChildren.length === 1 && loadingVariant !== undefined) {
+      // if view more button is only item, set typeaheadFilteredChildren to am empty array
       typeaheadFilteredChildren = [];
     }
     if (typeaheadFilteredChildren.length === 0) {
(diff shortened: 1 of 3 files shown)


dlabrecq (commented):
The first click of "View more" added two more options, but the second click did not. After the second click, the "View more" action disappeared. 

If there were no more options available, then perhaps the "View More" should have been omitted earlier (i.e., after the first click)? Or, never omitted at all?

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -428,6 +429,13 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   (valueToCheck as SelectOptionObject).toString &&
                   (valueToCheck as SelectOptionObject).compareTo;
 
+                // View more option should be returned as not a match
+                if (loadingVariant !== undefined && loadingVariant !== 'spinner') {
+                  if (loadingVariant?.text === valueToCheck && loadingVariant.text === valueToCheck) {
```
wise-king-sullyman, line 434: Isn't this checking the same thing twice?

jeffpuzzo:
Switching the `view more` example to use typeahead as mentioned, and clicking `View More` twice to see all options, it appears that the option where `View More` used to be is highlighted (grey background color) even when closing and re-opening the menu, and even when a different option is clicked. I would expect for that grey background to not be present, would you agree?

image

jessiehuff (commented):
So my guess is that the dark grey styling is because focus is on those elements. The focus aspect is intentional (so that when you select "view more", focus goes to the next available item instead of all the way back up to the top). However, if that is the case, it does seem odd that we'd be highlighting focus with the grey background and not also with the blue focus ring. I'm curious of your thoughts on the styling @mcoker. Is that gray being applied because of the focus like I'm thinking, and if so, should we include the focus ring to be clearer with it?

Also, unfortunately I noticed that I can't easily select the "view more" button with VO. It automatically closes the Select when pressed. I tested org though and saw that this is apparently a pre-existing issue so I created [an issue for that here](https://github.com/patternfly/patternfly-react/issues/6895).

nicolethoen (approved):

packages/react-core/src/components/Select/Select.tsx
gitdallas: seems like both both ifs could be replaced with a single `if (loadingVariant?.text === valueToCheck)` if it was either thing from the first condition, `loadingVariant?.text` would be `undefined` which could never be `valueToCheck` as we previously check that on 423.

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -428,6 +429,13 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
...
+                    return true;
+                  }
+                }
```
gitdallas, line 437:
```suggestion
                if (loadingVariant?.text === valueToCheck) {
                  return true;
                }
```

mcoker:
> So my guess is that the dark grey styling is because focus is on those elements. The focus aspect is intentional (so that when you select "view more", focus goes to the next available item instead of all the way back up to the top). However, if that is the case, it does seem odd that we'd be highlighting focus with the grey background and not also with the blue focus ring. I'm curious of your thoughts on the styling @mcoker. Is that gray being applied because of the focus like I'm thinking, and if so, should we include the focus ring to be clearer with it?

@jessiehuff @tlabaj I'm not sure why that's happening. We don't manipulate the focus outline for those elements, but we do apply a grey background on focus via the `:focus` pseudo class. I'm assuming it has to do with the way we're focusing on that element. For example, from the [`.focus()` docs on MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#result_2), there is no outline when the button gets programmatic focus, either. But it will apply a style change (if you add one in dev tools, since it triggers the `:focus` pseudo class.

packages/react-core/src/components/Select/Select.tsx
tlabaj: That gave me an error when I coded it that way because loading variant might not have `.text` if if is spinner.  So I had to add the check.

jeffpuzzo: So I assume it could just be `if (loadingVariant !== 'spinner' && loadingVariant?.text === valueToCheck) {`, yes?

gitdallas (approved):

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -255,7 +255,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       this.refCollection[0][0] = this.filterRef.current;
     }
 
-    // Move focus to top of the menu if state.focusFirstOption was updated to true and the menu does not have custom content
+    // Move focus to top of the menu if state.c was updated to true and the menu does not have custom content
```
nicolethoen, line 258: is this update a typo?

jessiehuff (approved):
Looks good on the a11y side. 🙂

mcoker (approved):
🤩

nicolethoen:
I think the keyboard interaction still seems a bit glitchy...

1. when I use the keyboard to navigate to the view more option, it does not appear focused
2. when I use the keyboard to view more, it updates the input to say 'view more' and does not show anything loading.
3. if I backspace to clear the 'view more' from the input, i see all the newly loaded options, but browser focus now conflicts with the 'fake focus' feature of the typeahead select.

Feb-15-2022 20-55-41

nicolethoen (changes requested):
couple notes about keyboard interactions

tlabaj:
@nicolethoen that is already captured in the separate [issue](https://github.com/patternfly/patternfly-react/issues/6923) I told you about.  This PR is to fix the filtering issue.

nicolethoen (approved):

thatblindgeye (commented):
I did notice that when typing in the View More example (with Typeahead variant) that if you type something *not* currently in the list (e.g. "Mister"), the View More button is removed with only the "No results found" text remaining. 

Is this the expected behavior based on the issue, or should the View More button *always* be visible until there are no more options to 'view more' of?

Screenshot showing the behavior:
Select component behavior

tlabaj:
@mcarrano Can you see Eric comment above.  What would be the expected behaviour?

mcarrano:
@tlabaj I wasn't sure how to reproduce this but based on the description, I think the current behavior is OK. So if I understand correctly the View more button disappears when the list becomes short enough that there is no more need for it. Is that what's happening? If so, that's what I would expect.

tlabaj:
> @tlabaj I wasn't sure how to reproduce this but based on the description, I think the current behavior is OK. So if I understand correctly the View more button disappears when the list becomes short enough that there is no more need for it. Is that what's happening? If so, that's what I would expect.

@mcarrano When the select is typeahead variant with view more, if the filter does not match any current loaded options "no more results".  Should we still display the "view more" since new items can still be loaded.

https://user-images.githubusercontent.com/39532947/154362675-aee97a4c-27cc-4ba7-b5bf-da96af0768ef.mov

mcarrano:
Oh, I see @tlabaj . The way I had assumed this would work is that when you apply the filter, it applies to the entire list of items, not just what's in view. So for example, if the original value list had 50 items and I'm only looking at the first 10, then View more is visible. But if the filter brings the list of possible values done to 5, then View more can be hidden. So the situation you describe would never occur. I would never expect "No results found" and View more to be present together.

Can it be made to work this way?

tlabaj:
@mcarrano the component does not have a full list.  We can only filter on what is loaded

thatblindgeye (approved):
Based on the conversation, it sounds like things are working as expected more or less, i.e. when typing something in the input that doesn't match any loaded options, the View More button should disappear. Just wanted to mention the situation I had noticed in case.

Other than that looks good!

mcarrano:
> @mcarrano the component does not have a full list. We can only filter on what is loaded

If that's the case @tlabaj , then I think you need to continue to show the View more button as long as there are more possible results in the list otherwise there would be no way to get to them, right?

nicolethoen (approved):

kmcfaul (approved):

mcarrano (approved):
Looks good. Thanks for making that last update @tlabaj .

Issue #6251, opened 2021-08-31
NumberInput has inline styles which make it break when used with strict CSP


KKoukiou:
**Describe the issue. What is the expected and unexpected behavior?**
NumberInput component seems to break when used with strict CSP.  In specific cockpit uses "content-security-policy": "img-src 'self' data:" and the styles for 'pf-c-input-group' don't seem to get applied. (@patternfly/patternfly/components/InputGroup/input-group.scss) 

This results in the following:
![Screen Shot 2021-08-31 at 13 20 32](https://user-images.githubusercontent.com/14921356/131493922-54307b27-e283-4143-a2d6-dca67e36bf29.png)

The problem can we workarounded by importing explicitely @patternfly/patternfly/components/InputGroup/input-group.scss in our CSS. 

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
bug

Do you do something different in the inclusion of the InputGroup's CSS in the NumberInput, compared to the NumberInput's own CSS? The latter get's applied fine.

nicolethoen:
@mcarrano could we add this to a milestone?
@mcoker I'm not sure if this has to do with how we import css styles.

PR #6850, opened 2022-01-23
fix(numberinput): import input group component instead

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/6251

@KKoukiou can you test this and see if it fixes it? Not sure why we were applying a class instead of importing the `<InputGroup>` component.


KKoukiou (approved):
Looks good! It's not easy to test it  now on top of cockpit, but I feel confident the fix is correct.

jeffpuzzo (approved):

tlabaj (approved):

PR #6865, opened 2022-01-27
fix(TimePicker): apply includeSeconds when making options

### LINKED ISSUE #6837: TimePicker: first time option is missing when seconds are included
When the newly added `includeSeconds` prop is passed in with a `true` value, the menu that expands with all of the time options is missing the first option (12:00 AM for 12 hour variant and 00:00 for 24 hour variant).

The expected behavior is for those time options to be present.

https://patternfly-react-main.surge.sh/components/time-picker#with-seconds

Bug. Excluding seconds in the time is technically a work-around, but the issue will need to be resolved for the new `includeSeconds` feature.

DaoDaoNoCode:
**What**: Closes #6837
When making time options for the dropdown menu, the `includeSeconds` props in not set. Which makes some errors when comparing option 00:00 with the min time 00:00:00 because they are not the same format.
Also, update the doc to prevent some building warnings.

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -425,7 +425,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     } = this.props;
     const { timeState, isOpen, isInvalid, minTimeState, maxTimeState } = this.state;
     const style = { '--pf-c-date-picker__input--c-form-control--Width': width } as React.CSSProperties;
-    const options = makeTimeOptions(stepMinutes, !is24Hour, delimiter, minTimeState, maxTimeState);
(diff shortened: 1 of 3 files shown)


jeffpuzzo:
I see in the description that this is fixing "errors", is the only error not that `00:00` and `12:00` were not being shown as min values in the dropdown?

In any case, this looks to have fixed the reported issue linked.

jeffpuzzo (approved):

DaoDaoNoCode:
@jeffpuzzo Calling it an error may not be accurate but I don't know which word is better to be used here. That's the incorrect comparison that could cause the option to be misjudged and not show the option because it's not comparing `00:00:00` with `00:00:00` but comparing `00:00` with `00:00:00`.

kmcfaul (approved):

nicolethoen (approved):

tlabaj (approved):

tlabaj (approved):

PR #6867, opened 2022-01-28
fix(Progress): NaN not returned when max and value are zero

### LINKED ISSUE #6733: When value and max values are not defined for `Progress` a `NaN%` is returned for the measure
### Issue
On initialization of our application, the values for the progress aren't yet initialized (as they are provided by an API). This means both `value` and `max` are `0`. This means that the measure will read as `NaN` as a division by zero happens?

```javascript
import React from 'react';
import { Progress, ProgressMeasureLocation } from '@patternfly/react-core';

<Progress value={0}
    min={0} max={Number(0)}
    measureLocation={ProgressMeasureLocation.outside} />
```

image

### Workaround
To prevent his to happen, a max value of at least `> 0` needs to be given.

```javascript
<Progress value={0}
    min={0} max={ max > 0 ? Number(1) : max }
    measureLocation={ProgressMeasureLocation.outside} />
```

This will result in a `0%` to be shown.
image

### Suggestion
Do not render `NaN` as this value is invalid.
Perhaps even allow a empty field when `NaN` is calculated.
Especially this `NaN` makes the bar jump around as it tries to fit.

Especially as:
```javascript
<Progress value={0}
    min={0}
    measureLocation={ProgressMeasureLocation.outside} />
```
shows that expected behaviour of showing `0%`.

thatblindgeye:
**What**: Closes #6733 

**Additional issues**:
N/A

### DIFF
--- a/packages/react-core/src/components/Progress/Progress.tsx
+++ b/packages/react-core/src/components/Progress/Progress.tsx
@@ -113,7 +113,10 @@ export class Progress extends React.Component<ProgressProps> {
       );
     }
 
-    const scaledValue = Math.min(100, Math.max(0, Math.floor(((value - min) / (max - min)) * 100)));
+    const scaledValue = Math.min(
+      100,
+      Math.max(0, Math.floor(((value - min) / ((max <= min ? 100 : max) - min)) * 100))
+    );
     return (
       <div
         {...props}
--- a/packages/react-core/src/components/Progress/Progress.tsx
+++ b/packages/react-core/src/components/Progress/Progress.tsx
@@ -113,10 +113,7 @@ export class Progress extends React.Component<ProgressProps> {
       );
     }
 
-    const scaledValue = Math.min(
-      100,
-      Math.max(0, Math.floor(((value - min) / ((max <= min ? 100 : max) - min)) * 100))
-    );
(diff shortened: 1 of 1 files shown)


nicolethoen (approved):

packages/react-core/src/components/Progress/Progress.tsx
```diff
@@ -113,7 +113,10 @@ export class Progress extends React.Component<ProgressProps> {
       );
     }
 
-    const scaledValue = Math.min(100, Math.max(0, Math.floor(((value - min) / (max - min)) * 100)));
+    const scaledValue = Math.min(
+      100,
+      Math.max(0, Math.floor(((value - min) / ((max <= min ? 100 : max) - min)) * 100))
```
gitdallas, line 118: I kind of feel like if we are in this situation, we should just return 0. It seems like it would only happen if there is some kind of error in whatever is using this component.. or perhaps if it's still initializing some of the values or something? At this point it might be best just to see no progress as opposed to potentially some value that doesn't make sense.

gitdallas (approved):

kmcfaul (approved):
lgtm

nicolethoen (approved):

tlabaj (approved):

PR #6868, opened 2022-01-28
feat(TextInputGroup): add ability to do typehead with textinputgroup, add to autocomplete demo

### LINKED ISSUE #6671: Add autocomplete ghosting feature to TextInputGroup
Follow up work to core issue: https://github.com/patternfly/patternfly/issues/4558

gitdallas:
**What**: Closes #6671 - adding typeahead feature to text input group. adding a demo where we use typeahead if there is a single match in the autocomplete list.

### DIFF
--- a/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
+++ b/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
@@ -64,14 +64,14 @@ export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps
       {children}
       <span className={css(styles.textInputGroupText)}>
         {inputHint && (
-                <input
-                  className={css(styles.textInputGroupTextInput, styles.modifiers.hint)}
-                  type="text"
-                  disabled
-                  aria-hidden="true"
-                  value={inputHint}
-                />
-              )}
+          <input
+            className={css(styles.textInputGroupTextInput, styles.modifiers.hint)}
+            type="text"
+            disabled
+            aria-hidden="true"
+            value={inputHint}
+          />
+        )}
         {icon && <span className={css(styles.textInputGroupIcon)}>{icon}</span>}
         <input
           type={type}
--- a/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
+++ b/packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
@@ -3,7 +3,7 @@ import styles from '@patternfly/react-styles/css/components/TextInputGroup/text-
 import { css } from '@patternfly/react-styles';
 import { TextInputGroupContext } from './TextInputGroup';
 
-export interface TextInputGroupMainProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange'> {
+export interface TextInputGroupMainProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onChange' | 'results' | 'ref'> {
   /** Content rendered inside the text input group main div */
   children?: React.ReactNode;
   /** Additional classes applied to the text input group main container */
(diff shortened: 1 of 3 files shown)


packages/react-core/src/demos/examples/TextInputGroup/AutoCompleteSearch.tsx
```diff
@@ -67,6 +68,20 @@ export const AutoCompleteSearch: React.FunctionComponent = () => {
...
+    /** The hint is set whenever there is only one autocomplete option left. */
+    if (filteredMenuItems.length === 1) {
+      const hint = filteredMenuItems[0].props.children;
+      if (!inputValue.indexOf(hint)) {
+        // the match was found in a place other than the start, so typeahead wouldn't work right
+        setHint('');
+      } else {
+        // use the input for the first part, otherwise case difference could make things look wrong
+        setHint(inputValue + hint.substr(inputValue.length));
+      }
```
thatblindgeye, line 80: Should the `inputValue` and `hint` for the `indexOf` check here be reversed? It looks like the `if` condition will always return false, which might explain the behavior I'm seeing.

thatblindgeye (commented):
When testing the demo out, I end up with what is shown in the image below when I try typing "space"; I was expecting the "No results found" option:

autocomplete for text input group

The `hint` ends up getting set as "spacepace".

thatblindgeye (changes requested):

packages/react-core/src/demos/examples/TextInputGroup/AutoCompleteSearch.tsx
```diff
@@ -64,9 +65,24 @@ export const AutoCompleteSearch: React.FunctionComponent = () => {
...
 
+    /** The hint is set whenever there is only one autocomplete option left. */
+    if (filteredMenuItems.length === 1) {
+      const hint = filteredMenuItems[0].props.children;
+      if (hint.indexOf(inputValue)) {
+        // the match was found in a place other than the start, so typeahead wouldn't work right
+        setHint('');
+      } else {
+        // use the input for the first part, otherwise case difference could make things look wrong
+        setHint(inputValue + hint.substr(inputValue.length));
```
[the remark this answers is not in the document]
gitdallas: @thatblindgeye - I would expect "name" not to provide ghosting because there would be 2 autocomplete matches at that point (name and namespace), it would be getting stopped at the `filteredMenuItems.length === 1` condition in this case.

If you type "names" you should get the single match and the ghosting. As for the `indexaOf` logic, `hint.indexOf(inputValue)` would be non-zero (truthy) if the match is not at the beginning (which would make ghosting possible).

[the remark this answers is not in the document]
thatblindgeye: Ah thanks for catching that. I had meant to say when I type "names". You're right, though, I think I was basing my suggested changes off of a combo of your original `indexOf` line and the updated one so thank you for pointing that out. 

After testing it some more it looks like I'm not getting the ghosting effect due to the difference in case ("name" vs "Name"), as when I type "Names" I get the ghosting to appear again. I have a screenshot showing the two side by side:

image

If that's the expected behavior I can mark this as resolved, though.

thatblindgeye (approved):
Compared to the SearchInput and after checking the ghosting again, looks good!

packages/react-core/src/components/TextInputGroup/TextInputGroupMain.tsx
```diff
@@ -42,6 +44,7 @@ export const TextInputGroupMain: React.FunctionComponent<TextInputGroupMainProps
   className,
   icon,
   type = 'text',
+  hint: inputHint,
```
tlabaj, line 47: Curious why this is needed?

gitdallas: Probably not needed. I had an issue where it wasn't recognizing the new `hint` property so I was trying a bunch of different things. The `inputHint` format was me following the way they had `inputPlaceholder` and `inputValue`. The real issue ended up being that my IDE was acting up. Would you like me to remove this?

nicolethoen (approved):
LGTM

mcoker (approved):
LGTM! 🥳

PR #6869, opened 2022-01-31
feat(MenuToggle): add full height variation

### LINKED ISSUE #5964: Full height variation of menu toggle
menu toggle: https://github.com/patternfly/patternfly/pull/4153

Adds `.pf-m-full-height` as a variation to the toggle that removes the top/bottom border and makes it fill the vertical space of the parent. Primary use case for this is with the new masthead component.

boaz0:
**What**: Closes #5964

### DIFF
--- a/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
+++ b/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
@@ -14,6 +14,8 @@ export interface MenuToggleProps
   isExpanded?: boolean;
   /** Flag indicating the toggle is disabled */
   isDisabled?: boolean;
+  /** Flag indicating the toggle is full height */
+  isFullHeight?: boolean;
   /** Variant styles of the menu toggle */
   variant?: 'default' | 'plain' | 'primary' | 'plainText';
   /** Optional icon rendered inside the toggle, before the children content */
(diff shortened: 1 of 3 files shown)


mcarrano (changes requested):
@boaz0 The core example here shows a tall button which demonstrates that this will grow to fill the available height: https://www.patternfly.org/v4/components/menu-toggle/html#full-height  Can we do something similar for the React example?

mmenestr:
Agreed @mcarrano, it would make it clearer what makes it differ from the other buttons! 

Also I notice the left and right spacing differs from the other menus which I assume is on purpose, so i'm wondering what the background on this issue is since it seems to focus on height and not the width? Is the idea that we just want a dropdown that can be resized any way, and therefore the text and caret is built to stay center aligned no matter the dimensions of the dropdown?

mcarrano:
Good catch @mmenestr .

@mcoker was it intentional that the horizontal alignment of the full-height dropdown should be unique?

mcoker:
@mcarrano @mmenestr the main use case for this variation is in the masthead, and I'm assuming it's just to balance the space around the dropdown toggle text. Here's what the spacing looks like

Screen Shot 2022-02-03 at 10 34 27 AM

Something that could be improved if we're OK with that spacing is to remove the space between the arrow and right edge of the toggle padding

Screen Shot 2022-02-03 at 10 35 15 AM

mcoker (approved):
Aside from @mcarrano's comment about the example height, LGTM! Core uses an `80px` height for this example.

mmenestr:
@mcoker I think removing that space like you're suggesting would be a good move! It would look more "balanced"!

mmenestr:
@boaz0 Looks great with that height! Is it normal that there's no top border, though?

mcoker:
> Is it normal that there's no top border, though?

@mattnolting can you confirm this?

nicolethoen (approved):

mattnolting:
@mcoker @boaz0 It is intentional that there's no top border for `full-height` variant, otherwise there'd be a double border top/bottom.

tlabaj (approved):
LGTM

mmenestr (approved):
In that case, LGTM!

mcarrano (approved):
Looks great!

PR #6870, opened 2022-01-31
fix(Menu): account for initial drill in state

### LINKED ISSUE #6761: Menu: setting an initial activeMenu different than the 'root' menu crashes the component (drilldown variant)
**Describe the issue. What is the expected and unexpected behaviors
I am implementing a language selection feature, which this (drilldown menu variant) Menu component seems to serve very well.

Screen Shot 2022-01-11 at 09 27 21 
Screen Shot 2022-01-11 at 09 27 13

The menu should have a preselected language, and for this reason I need to have the 'activeMenu', 'drilldownPath' and 'menuDrilledIn' properties set accordingly.

I tried doing it and the component crashes.

Could I get an advice from the designers if you think this component is the right one to use? @mcoker 

If yes, let's keep this bug report and make sure the component works with the non-root preselected menu item.

https://codesandbox.io/s/optimistic-bird-0flyu?file=/index.js

/bug

Anaconda - https://communityblog.fedoraproject.org/anaconda-is-getting-a-new-suit/#more-10593

For the record the component previously (in a GTK app) looked like this:
lanaguage-first

kmcfaul:
**What**: Closes #6761

The bug causing the crash lies in the `setHeight` callback. When `rootMenu` is not the initial `activeMenu`, out-of-view submenus of the root are not rendered and return `0` for their menu height. When `setHeight` attempts to collect menu heights, it evaluates `!0` to true and loops on updating the same submenu.

I've updated the example callback to allow the heights to update properly, but the caveat for this use case is that `menuHeights` must initially set the height of `rootMenu`. This is because the content of `rootMenu` is updated when a menu is drilled in/out, so it will keep a previously drilled in menu's height rather than updating based on the root content. 

I've updated the example code as well as added an additional example showcasing an initially drilled in menu with some extra description.


packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -718,7 +714,211 @@ class MenuWithDrilldown extends React.Component {
...
+To render an initially drilled in menu, the `menuDrilledIn`, `drilldownPath`, and `activeMenu` states must be set to an initial state. The `menuHeights` state must also be set, defining the height of the root menu. The `setHeight` function passed into the `onGetMenuHeight` property must also account for updating heights, other than the root menu, as menus drill in and out of view.
+
+```js isBeta
+import React from 'react';
+import { Menu, MenuContent, MenuList, MenuItem, Divider, DrilldownMenu } from '@patternfly/react-core';
+import StorageDomainIcon from '@patternfly/react-icons/dist/esm/icons/storage-domain-icon';
+import CodeBranchIcon from '@patternfly/react-icons/dist/esm/icons/code-branch-icon';
+import LayerGroupIcon from '@patternfly/react-icons/dist/esm/icons/layer-group-icon';
+import CubeIcon from '@patternfly/react-icons/dist/esm/icons/cube-icon';
+import { setTimeout } from 'timers/promises';
```
jeffpuzzo, line 890: It doesn't look like this setTimeout is used.

jeffpuzzo (approved):

jeffpuzzo:
The updated and new examples looks good to me. I wonder though if the new one for an initial drilldown should be made into a separate file and use typescript though. I don't think this should block the PR, but something to consider with our efforts to convert examples to use TS.

nicolethoen (commented):
This looks great!
Could you add information about the caveat and considerations to the documentation itself? Either outside the code or within it as comments?

kmcfaul:
Updated to remove the unused import and convert the new example (and the basic drilldown example as there was a lot of overlap) to TS as well as move them to separate files.

@nicolethoen I do have a description in front of the new example, is that sufficient?

nicolethoen (approved):

packages/react-core/src/components/Menu/examples/Menu.md
```diff
@@ -684,200 +680,14 @@ class MenuOptionMultiSelect extends React.Component {
...
-            </MenuItem>
-            <MenuItem itemId="edit">Edit</MenuItem>
-            <MenuItem itemId="delete_deployment">Delete deployment config</MenuItem>
-          </MenuList>
-        </MenuContent>
-      </Menu>
-    );
-  }
-}
+To render an initially drilled in menu, the `menuDrilledIn`, `drilldownPath`, and `activeMenu` states must be set to an initial state. The `menuHeights` state must also be set, defining the height of the root menu. The `setHeight` function passed into the `onGetMenuHeight` property must also account for updating heights, other than the root menu, as menus drill in and out of view.
```
tlabaj, line 688: Thank's for adding this, it is helpful.

tlabaj (approved):

PR #6872, opened 2022-01-31
feat(Menu): allow drilldown menu functions

### LINKED ISSUE #5534: Dynamic Menu drilldown
Regarding the new Menu with drilldown example. Can all those menu items be created dynamically; for example, as the user traverses an org hierarchy?

My concern is having to create all the components upfront with such a large hierarchy. It looks like the `drilldownMenu` prop of `MenuItem` doesn't accept a function?

@tlabaj suggested "we could modify the prop to accept a function"

kmcfaul:
**What**: Closes #5534

Allows `drilldownMenu` to accept a function returning a ReactNode. 

I've tested by wrapping the current example's drilldownMenus in function statements which works, but I'm not sure how well the component will respond to delays - primarily because the `MenuItem` needs to know the to-be-drilled-in menu's ID before the drilldownMenu would necessarily be returned/rendered. At present, I've invoked the drilldownMenu function in this drill in call to get the ID, but I'm wondering if it would be better to have a separate property for drilldownMenuId? Any thoughts?

### DIFF
--- a/packages/react-core/src/components/Menu/MenuItem.tsx
+++ b/packages/react-core/src/components/Menu/MenuItem.tsx
@@ -51,7 +51,7 @@ export interface MenuItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'onC
   /** @beta Callback function when mouse leaves trigger */
   onShowFlyout?: (event?: any) => void;
   /** @beta Drilldown menu of the item. Should be a Menu or DrilldownMenu type. */
-  drilldownMenu?: React.ReactNode;
+  drilldownMenu?: React.ReactNode | (() => React.ReactNode);
   /** @beta Sub menu direction */
   direction?: 'down' | 'up';
   /** @beta True if item is on current selection path */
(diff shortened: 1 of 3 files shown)


dlabrecq (commented):
Would it be possible to add an example?

The ask for a function was to allow the menu items to be added dynamically. In our case, we want to lazily add drill items as the user navigates. We have too many items to add all of that up front.

nicolethoen (approved):

jeffpuzzo:
It's not clear to me that this update addresses the problem fully as described by @dlabrecq, maybe he can confirm. 

It seems to me that the creation of menu items can be dynamic without drilldownMenu having the option to be a function.  To do this, a list of menu items or menu item props could be managed in state, updated when needed and iterated over.

dlabrecq (approved):
Don't believe we want to manage lists in state, there could be hundreds of items. 

I think the functions are more flexible and should work for us.

tlabaj (approved):

PR #6877, opened 2022-02-01
fix(Tile): add aria-selected to indicate selection state

### LINKED ISSUE #5891: Tile demo should indicate selection state
Currently when a tile is selected in the demo, the screen reader is not updated on the changed state. This should follow other selected state patterns. I also noticed that the link to the demo section in the documentation is out of date and should be updated.

Test the tile demo by screen reader and notice that nothing is announced when tile is selected.

Demo bug

boaz0:
**What**: Closes #5891

### DIFF
--- a/packages/react-core/src/components/Tile/Tile.tsx
+++ b/packages/react-core/src/components/Tile/Tile.tsx
@@ -35,6 +35,8 @@ export const Tile: React.FunctionComponent<TileProps> = ({
   <div
     role="option"
     aria-selected={isSelected ? true : false}
+    // aria-disabled={isDisabled}
+    {...(isDisabled && { 'aria-disabled': isDisabled })}
     className={css(
       styles.tile,
       isSelected && styles.modifiers.selected,
--- a/packages/react-core/src/components/Tile/Tile.tsx
+++ b/packages/react-core/src/components/Tile/Tile.tsx
@@ -34,8 +34,7 @@ export const Tile: React.FunctionComponent<TileProps> = ({
 }: TileProps) => (
   <div
     role="option"
-    aria-selected={isSelected ? true : false}
-    // aria-disabled={isDisabled}
+    aria-selected={isSelected}
     {...(isDisabled && { 'aria-disabled': isDisabled })}
     className={css(
       styles.tile,
--- a/packages/react-core/src/components/Tile/Tile.tsx
+++ b/packages/react-core/src/components/Tile/Tile.tsx
@@ -33,6 +33,7 @@ export const Tile: React.FunctionComponent<TileProps> = ({
   ...props
 }: TileProps) => (
   <div
+    role="option"
     aria-selected={isSelected ? true : false}
     className={css(
       styles.tile,
--- a/packages/react-core/src/components/Tile/Tile.tsx
+++ b/packages/react-core/src/components/Tile/Tile.tsx
@@ -33,6 +33,7 @@ export const Tile: React.FunctionComponent<TileProps> = ({
   ...props
 }: TileProps) => (
   <div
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Tile/Tile.tsx
```diff
@@ -33,6 +33,7 @@ export const Tile: React.FunctionComponent<TileProps> = ({
   ...props
 }: TileProps) => (
   <div
+    aria-selected={isSelected ? true : false}
```
jeffpuzzo, line 36: There's no need for a ternary here. All that's needed is `isSelected`.

nicolethoen:
So, actually, my original recommendation in the issue was only part of the solution.

[aria-selected](https://www.w3.org/TR/wai-aria-1.1/#aria-selected) needs to live inside of an element with one of the following roles: 

- [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell)
- [option](https://www.w3.org/TR/wai-aria-1.1/#option)
- [row](https://www.w3.org/TR/wai-aria-1.1/#row)
- [tab](https://www.w3.org/TR/wai-aria-1.1/#tab)

I think in this case, the `option` role is best for tiles, because our Tile component is designed to be a form control which functions as a radio or select input. This means the container div would need the [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) role - and if that's the case, we'd have to also make use of `listbox`'s `aria-multiselectable` (when applicable), `aria-activedescendent` and maybe a few others to be sure the screen reader knows the full context of what can be selected and what is selected.

packages/react-core/src/demos/TileDemo.md
```diff
@@ -73,33 +73,36 @@ const TileMultiSelect: React.FunctionComponent = () => {
...
         title="Tile 2"
         id="tile-2"
         onClick={onSelect}
         onKeyDown={onKeyDown}
         isSelected={selectedIds.includes('tile-2')}
       />
-      <Tile title="Tile 3" id="tile-3" isDisabled />
+      <Tile aria-multiselectable title="Tile 3" id="tile-3" isDisabled />
       <Tile
+        aria-multiselectable
```
thatblindgeye, line 105: For the `aria-multiselectable` attribute, this should go on the element that can have multiple descendants selected (the div with the role of "listbox" in this case) with a value of `true`, rather than on the descendants themselves. From there the `aria-multiselectable` attribute should also be removed from each of the Tile components.

thatblindgeye (changes requested):

thatblindgeye:
Hi @boaz0, just wanted to check on the status of the above change request.

packages/react-core/src/components/Tile/Tile.tsx
```diff
@@ -33,6 +33,10 @@ export const Tile: React.FunctionComponent<TileProps> = ({
   ...props
 }: TileProps) => (
   <div
+    role="option"
+    aria-selected={isSelected ? true : false}
+    // aria-disabled={isDisabled}
```
wise-king-sullyman, line 38: Was it intended to leave this comment in?

jessiehuff (commented):
Looks good to me! :) The only minor suggestion I'd make is labelling the listbox div with the title of the example so that it's clearer to a user what they're making a selection of.

wise-king-sullyman (commented):
Looks good to me other than the CI failures. Once those are resolved I'll be ready to approve.

wise-king-sullyman (approved):
🔥

jessiehuff (approved):
LGTM! 😎

tlabaj (approved):

PR #6884, opened 2022-02-02
fix(TimePicker): remove automatically validation from onChange event

### LINKED ISSUE #6838: TimePicker: validate time onBlur instead of after each key press
Currently when manually typing in a time, validation occurs after each key press. This results in the initial character to cause the "Invalid time format" error to render as well as when a user begins typing a time suffix ("A" in AM for example), so the error is constantly popping in and out.

Validation should occur a single time instead onBlur when the input no longer has focus.

https://patternfly-react-main.surge.sh/components/time-picker/

As you type in `12:00 PM`, you will notice the error begins rendering after the initial `1`, goes away once you've entered `12:00`, and reappears when you have entered `12:00 P`

Enhancement

DaoDaoNoCode:
**What**: Closes #6838 
When the focus is back on the time input, I disable the validation, I am not sure whether it makes more sense not to do any validation when the user is inputting the time.

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -357,7 +357,8 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
 
     this.inputRef.current.focus();
     this.setState({
-      isOpen: false
+      isOpen: false,
+      isInvalid: false
     });
   };
 
@@ -366,9 +367,6 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
       this.onToggle(true);
     }
     e.stopPropagation();
-    this.setState({
-      isInvalid: false
-    });
   };
 
   onInputChange = (newTime: string) => {
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -366,6 +366,9 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
       this.onToggle(true);
     }
     e.stopPropagation();
+    this.setState({
+      isInvalid: false
+    });
   };
 
   onInputChange = (newTime: string) => {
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/TimePicker/__tests__/TimePicker.test.tsx
```diff
@@ -72,7 +72,7 @@ describe('test isInvalid', () => {
     expect(view.find('.pf-m-error')).toHaveLength(0);
   })
 
-  test('should be invalid after onChange', () => {
+  test('should be invalid after onChange and onBlur', () => {
```
nicolethoen, line 75: isn't it only onBlur now?

DaoDaoNoCode: It's actually testing `valid` onChange and `invalid` onBlur... 🤯

packages/react-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -366,6 +366,9 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
       this.onToggle(true);
     }
     e.stopPropagation();
+    this.setState({
+      isInvalid: false
+    });
```
thatblindgeye, line 371: This might fall into the category of nitpick/personal preference, but I'm wondering whether we should only set the `isInvalid` state on blur period rather than also setting it to false on focus?

Mainly not sure whether having the error message/icon appear and disappear all the time would be visually confusing; it looks like the message/icon also appears when going from focus on the input to navigating into the menu options with a keyboard.

DaoDaoNoCode: I did so because I want to make sure when we select the option from the dropdown menu, the error message could disappear. Maybe it's better to set `isInvalid: false` when selecting the option from the dropdown?

thatblindgeye: Ah gotcha, that makes sense! Setting `isInvalid` to false when selecting from the dropdown I'm thinking could work to remove the error when a user selects what should be a valid option.

thatblindgeye (approved):
The update to remove the error after clicking a valid option in the dropdown I think works quite nicer! Overall things look good as well

nicolethoen (approved):

PR #6887, opened 2022-02-03
feat(ToolTip): Diagonal tooltip placement

### LINKED ISSUE #6568: Add support for diagonal tooltip positioning
Add props for diagonal tooltip positioning per this core PR: https://github.com/patternfly/patternfly/pull/4470

@mcoker anything to add?

gitdallas:
**What**: Closes #6568 adding support for diagonal tooltips, update other components that use tooltips

### DIFF
--- a/packages/react-core/src/components/Progress/ProgressContainer.tsx
+++ b/packages/react-core/src/components/Progress/ProgressContainer.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import progressStyle from '@patternfly/react-styles/css/components/Progress/progress';
 import { css } from '@patternfly/react-styles';
-import { Tooltip } from '../Tooltip';
+import { Tooltip, TooltipPosition } from '../Tooltip';
 import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
 import TimesCircleIcon from '@patternfly/react-icons/dist/esm/icons/times-circle-icon';
 import ExclamationTriangleIcon from '@patternfly/react-icons/dist/esm/icons/exclamation-triangle-icon';
@@ -38,7 +38,21 @@ export interface ProgressContainerProps extends Omit<React.HTMLProps<HTMLDivElem
   /** Whether string title should be truncated */
   isTitleTruncated?: boolean;
   /** Position of the tooltip which is displayed if title is truncated */
-  tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
+  tooltipPosition?:
+    | TooltipPosition
+    | 'auto'
+    | 'top'
+    | 'bottom'
+    | 'left'
+    | 'right'
+    | 'top-start'
+    | 'top-end'
+    | 'bottom-start'
+    | 'bottom-end'
+    | 'left-start'
+    | 'left-end'
+    | 'right-start'
+    | 'right-end';
 }
 
 const variantToIcon = {
--- a/packages/react-core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react-core/src/components/Tooltip/Tooltip.tsx
@@ -15,7 +15,15 @@ export enum TooltipPosition {
   top = 'top',
   bottom = 'bottom',
   left = 'left',
-  right = 'right'
+  right = 'right',
+  topLeft = 'topLeft',
+  topRight = 'topRight',
+  bottomLeft = 'bottomLeft',
+  bottonRight = 'bottomRight',
+  leftTop = 'leftTop',
+  leftBottom = 'leftBottom',
+  rightTop = 'rightTop',
+  rightBottom = 'rightBottom'
 }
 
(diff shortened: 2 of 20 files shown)


mcarrano:
@tlabaj @gitdallas I'm not really sure what to evaluate here from a UX perspective. Do you think it would make sense to modify this example to allow to setting the tooltip position? https://patternfly-react-pr-6887.surge.sh/components/tooltip#options Right now all of the examples seem to show a top positioned tooltip.

gitdallas:
@mcarrano - before while bottom/left/right were available they weren't used either. i assume it is because there is an interactive example that allows users to select whichever position they want

tlabaj:
> @tlabaj @gitdallas I'm not really sure what to evaluate here from a UX perspective. Do you think it would make sense to modify this example to allow to setting the tooltip position? https://patternfly-react-pr-6887.surge.sh/components/tooltip#options Right now all of the examples seem to show a top positioned tooltip.

@mcarrano  you can set the new positions from that example.  You need to select the position from the dropdown.
image

tlabaj (approved):
LGTM!

mcarrano (approved):
Ah, guess I missed that @tlabaj . Looks good in that case. Thanks @gitdallas !

mmenestr (approved):

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -214,7 +238,15 @@ export const Tooltip: React.FunctionComponent<TooltipProps> = ({
     top: styles.modifiers.top,
     bottom: styles.modifiers.bottom,
     left: styles.modifiers.left,
-    right: styles.modifiers.right
+    right: styles.modifiers.right,
+    'top-start': styles.modifiers.topLeft,
```
jeffpuzzo, line 242: Curious why you went with `top-start` instead of `top-left`. Same with the rest that have `start` and `end`.

packages/react-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -60,7 +68,22 @@ export interface TooltipProps extends Omit<React.HTMLProps<HTMLDivElement>, 'con
    * ['top', 'right', 'left']. Since there is no space to the top, it checks if right is available. There's also no
    * space to the right, so it finally shows the tooltip on the left.
    */
-  flipBehavior?: 'flip' | ('top' | 'bottom' | 'left' | 'right')[];
+  flipBehavior?:
+    | 'flip'
+    | (
```
jeffpuzzo, line 73: I think this would read better if abstracted into its own type (everything in the parentheses)

packages/react-core/src/components/Progress/ProgressContainer.tsx
```diff
@@ -38,7 +38,21 @@ export interface ProgressContainerProps extends Omit<React.HTMLProps<HTMLDivElem
   /** Whether string title should be truncated */
   isTitleTruncated?: boolean;
   /** Position of the tooltip which is displayed if title is truncated */
-  tooltipPosition?: 'auto' | 'top' | 'bottom' | 'left' | 'right';
+  tooltipPosition?:
+    | TooltipPosition
```
jeffpuzzo, line 42: I've seen this done in a few other places, so this is more of a general question: Why do we include the interface when it has the same values as the strings following it?

packages/react-core/src/components/Tooltip/Tooltip.tsx
gitdallas: this is the way that it is used within popper which this is based on

packages/react-core/src/components/Progress/ProgressContainer.tsx
gitdallas: interface is included for the IDE i guess, while the string values are included because they are necessary for the parser that creates the documentation

packages/react-core/src/components/Tooltip/Tooltip.tsx
gitdallas: could, but i think we'd still need to type it out for the parser anyway... @nicolethoen ?

packages/react-core/src/components/Tooltip/Tooltip.tsx
jeffpuzzo: I see. I wonder if we can leverage those Popper types more for this then, since I see `ComputedPlacement` from `Popper` is close to, if not the same as what we want to use for Tooltip.

packages/react-core/src/components/Tooltip/Tooltip.tsx
nicolethoen: You could test it to be sure. But that certainly has been the case in the past.

packages/react-core/src/components/Progress/ProgressContainer.tsx
jeffpuzzo: I do find it redundant to include the interface as well, but I won't consider this a blocking comment.

PR #6893, opened 2022-02-04
feat(Page): Added center alignment for page sections

### LINKED ISSUE #5966: Center aligned page sections
https://github.com/patternfly/patternfly/pull/4011/

Adds a `.pf-m-align-center` variation to the following page section elements. Only works with `.pf-m-limit-width` (`isWidthLimited`, `isBreadcrumbWidthLimited`, `isTertiaryNavWidthLimited`)

* `.pf-c-page__main-nav`
* `.pf-c-page__main-breadcrumb`
* `.pf-c-page__main-tabs`
* `.pf-c-page__main-section`
* `.pf-c-page__main-wizard`

tlabaj:
**What**: Closes #5966

### DIFF
--- a/packages/react-core/src/components/Page/examples/Page.md
+++ b/packages/react-core/src/components/Page/examples/Page.md
@@ -612,7 +612,7 @@ import {
 } from '@patternfly/react-core';
 import BarsIcon from '@patternfly/react-icons/dist/esm/icons/bars-icon';
 
-export const OptionsMenu: React.FunctionComponent = () => {
+CenterAlignedPageSection = () => {
   const headerToolbar = (
     <Toolbar id="toolbar">
       <ToolbarContent>
--- a/packages/react-core/src/components/Page/examples/Page.md
+++ b/packages/react-core/src/components/Page/examples/Page.md
@@ -263,11 +263,11 @@ class FillPage extends React.Component {
 
     return (
       <Page header={Header} sidebar={Sidebar}>
-        <PageSection style={{ height: '10em' }}>This section is set to the default fill variant</PageSection>
-        <PageSection style={{ height: '10em' }} isFilled={true}>
+        <PageSection>A default page section</PageSection>
+        <PageSection isFilled={true}>
           This section fills the available space.
         </PageSection>
-        <PageSection style={{ height: '10em' }} isFilled={false}>
+        <PageSection isFilled={false}>
           This section is set to not fill the available space, since the last page section is set to fill the available space by default.
         </PageSection>
       </Page>
--- a/packages/react-core/src/components/Page/examples/Page.md
+++ b/packages/react-core/src/components/Page/examples/Page.md
@@ -195,7 +195,89 @@ TertiaryPage = () => {
 };
 ```
 
-### Main Section Padding
+### With or without fill
+
+```js
+import React from 'react';
+import {
+  Page,
+  Masthead,
+  MastheadToggle,
+  MastheadMain,
+  MastheadBrand,
+  MastheadContent,
+  PageSidebar,
+  PageSection,
+  PageToggleButton,
+  Toolbar,
+  ToolbarContent,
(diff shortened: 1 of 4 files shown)


mcarrano (commented):
@tlabaj looks like this matches the core example, but it's not really clear to me what this is doing since the content does not look centered. @mcoker ?

mcoker (approved):
LGTM! Left some comments, not blockers though. Also in core we wrap the content in the centered section in a card to show where the page section box. Isn't necessary, but I think it helps show how the prop changes the position.

packages/react-core/src/components/Page/examples/Page.md
```diff
@@ -590,3 +590,59 @@ class VerticalPage extends React.Component {
   }
 }
 ```
+
+### Centered content
+
+```js
```
jeffpuzzo, line 596: Should we make this a typescript example in a new file instead?

jeffpuzzo:
Seeing the same as @mcarrano which is that the text is not centered. Inspecting some of these elements, it seems like it's due to `pf-c-page__main-body` having a width of `100%`. If this is removed, the content will be centered.

I think this the core change we want to fix both the HTML and React examples. @mattnolting can you speak to why this class mentioned above has a width of `100%`?

mattnolting:
> Seeing the same as @mcarrano which is that the content is not centered. Inspecting some of these elements, it seems like it's due to `pf-c-page__main-body` having a width of `100%`. If this is removed, the content will be centered. https://github.com/patternfly/patternfly/blob/e2e18a98dc1a99691a14bb99b55ba9ac996ce3e4/src/patternfly/components/Page/page.scss#L462-L464
> 
> I think this the core change we want to fix both the HTML and React examples. @mattnolting can you speak to why this class mentioned above has a width of `100%`?

@jeffpuzzo not sure why I set `width: 100%`. Seems like it had something to do with brochure style applications and very wide screen widths, but the implementation does not look correct in retrospect. https://github.com/patternfly/patternfly/issues/4676

mcoker:
This modifier doesn't center align the text in the page section - it center aligns the page section. If you wrap the content in a card, you can see that's what it's doing in the demo

Screen Shot 2022-02-08 at 6 17 30 PM

We set `width: 100%` because `.pf-m-align-center` sets `align-items: center` on the page section, which makes the child body element of that section only take up the width it needs by default, meaning the section will shrink to fit the content inside, which we don't want. If the section shrinks to fit the content, and the section is centered, then the text appears centered, but it's really just the section that's centered.

Screen Shot 2022-02-08 at 6 18 16 PM

To illustrate that the text isn't centered, if you fill the section with content you see the text is actually left aligned.

Screen Shot 2022-02-08 at 6 17 44 PM

The core demos with centered text are using a utility class (`.pf-u-text-align-center`) to center the text. https://www.patternfly.org/v4/components/page/html-demos/centered-content/

mattnolting:
@mcoker Thanks for the detailed explanation! Might be worth reexamining intent and rewording the modifier during a breaking change sprint.

mcarrano:
Should we either add some documentation or update the example to be more like what @mcoker shows above? I don't think any of this will be obvious to someone just looking at the example.

mcarrano:
Sounds good to me @mcoker

packages/react-core/src/components/Page/examples/Page.md
jeffpuzzo: @wise-king-sullyman That makes sense. I'm ok with either approach.

wise-king-sullyman (approved):
LGTM!

jeffpuzzo (approved):
Seems like there could be some follow-up issues considering this comment; https://github.com/patternfly/patternfly-react/pull/6893#issuecomment-1033961372 which cites some potential changes to the demo and naming, but otherwise, this looks good to me.

PR #6903, opened 2022-02-08
docs(Menu/MenuToggle): Composable Menu - Typeahead Select Demo

### LINKED ISSUE #6888: Add a select demo for typeahead variant

jeffpuzzo:
**What**: Closes #6888

As a part of this, the MenuToggle was switched from a `button` to a `div` since we cannot have a button within a button, and we want the Icon to clear typeahead input values to be inside of a button.

The build will fail and styling of this should not be tested until this core PR is merged: https://github.com/patternfly/patternfly/pull/4673

### DIFF
--- a/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
+++ b/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
@@ -60,18 +60,25 @@ export class MenuToggleBase extends React.Component<MenuToggleProps> {
     const isPlain = variant === 'plain';
     const isPlainText = variant === 'plainText';
     const isTypeahead = variant === 'typeahead';
+    const toggleControls = (
+      <span className={css(styles.menuToggleControls)}>
+        <span className={css(styles.menuToggleToggleIcon)}>
+          <CaretDownIcon aria-hidden />
+        </span>
+      </span>
+    );
 
     const content = (
       <>
         {icon && <span className={css(styles.menuToggleIcon)}>{icon}</span>}
         {isTypeahead ? children : <span className="pf-c-menu-toggle__text">{children}</span>}
         {badge && <span className={css(styles.menuToggleCount)}>{badge}</span>}
-        {!isTypeahead && (
-          <span className={css(styles.menuToggleControls)}>
-            <span className={css(styles.menuToggleToggleIcon)}>
-              <CaretDownIcon aria-hidden />
-            </span>
-          </span>
+        {isTypeahead ? (
+          <button type="button" className="pf-c-menu-toggle__button" onClick={onClick} aria-label="Menu toggle">
+            {toggleControls}
+          </button>
+        ) : (
+          toggleControls
         )}
       </>
     );
--- a/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
+++ b/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
@@ -74,7 +74,7 @@ export class MenuToggleBase extends React.Component<MenuToggleProps> {
         {isTypeahead ? children : <span className="pf-c-menu-toggle__text">{children}</span>}
         {badge && <span className={css(styles.menuToggleCount)}>{badge}</span>}
         {isTypeahead ? (
-          <button type="button" className="pf-c-menu-toggle__button" onClick={onClick} aria-label="Menu toggle">
+          <button type="button" className={css(styles.menuToggleButton)} aria-expanded={isExpanded} onClick={onClick} aria-label="Menu toggle">
             {toggleControls}
           </button>
         ) : (
--- a/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
+++ b/packages/react-core/src/components/MenuToggle/MenuToggle.tsx
@@ -4,8 +4,10 @@ import { css } from '@patternfly/react-styles';
 import CaretDownIcon from '@patternfly/react-icons/dist/esm/icons/caret-down-icon';
 import { BadgeProps } from '../Badge';
 
+export type MenuToggleElement = HTMLDivElement | HTMLButtonElement;
+
 export interface MenuToggleProps
-  extends Omit<React.DetailedHTMLProps<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>, 'ref'> {
+  extends Omit<React.DetailedHTMLProps<React.HTMLAttributes<MenuToggleElement>, MenuToggleElement>, 'ref'> {
   /** Content rendered inside the toggle */
   children?: React.ReactNode;
   /** Additional classes added to the toggle */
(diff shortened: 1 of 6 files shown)


nicolethoen (commented):
Few interaction discrepancies:

- When I use the keyboard to navigate to the typeahead select, I cannot open the menu. In the Typeahead Select example, hitting enter while focusing on the input opens the menu.
- I cannot find a way to close the menu without selecting an option using the keyboard. It should close on escape or tab. And in both cases, focus should return to the input.
- One of the most complex parts of the typeahead select's implementation is it's focus management while navigating the menu options. I'd play with the original some and realize that browser focus does not leave the input. Instead it stays in the input while the user hits the up and down arrow keys and the currently highlighted value in the menu is also placed into the input.


I also know that you have opened core issues to address styles for this demo. But they are not documented here so I'm going to include what I notice:

- The width of the MenuToggle increases when the clear button is added. That is not true of the Select component. 
- The input needs to be 'invisible' within the Menu Toggle. right now it looks like a box in a box, which does not match the typeahead select styles.
- If we are going to implement the same focus management as the old typeahead select, we likely need a 'pf-m-focus' modifier which can be applied to the options to 'fake' focus the same way.

mcarrano (changes requested):
Currently the UX is not equivalent to the existing component as shown here: https://www.patternfly.org/v4/components/select#typeahead  I think @nicolethoen did a good job in summarizing the issues.

jeffpuzzo:
> Currently the UX is not equivalent to the existing component as shown here: https://www.patternfly.org/v4/components/select#typeahead I think @nicolethoen did a good job in summarizing the issues.

nicolethoen:
@jeffpuzzo the typeahead select demo seems to have a syntaxerror in the preview. Is that new or has the preview been updated since your more recent pushes?

jeffpuzzo:
> @jeffpuzzo the typeahead select demo seems to have a syntaxerror in the preview. Is that new or has the preview been updated since your more recent pushes?

@nicolethoen Looks like examples fail and claim syntax errors when using optional function parameters, where the optional params are correct, and locally there will be typescript errors without them, but those lines have been updated.

nicolethoen (commented):
I think this is looking pretty good! - I'm not sure we want to mark an option selected just because the user has typed a matching string until they select it(?)
typeahead

nicolethoen (approved):
I think this LGTM :)
I am interested if @kmcfaul has any feedback given the similar demo work she has also done.

mcarrano (approved):
Looks good to me!

packages/react-core/src/demos/ComposableMenu/examples/ComposableTypeaheadSelect.tsx
```diff
@@ -0,0 +1,195 @@
...
+              innerRef={textInputRef}
+            />
+
+            <TextInputGroupUtilities>
+              {!!inputValue && (
+                <Button variant="plain" onClick={() => setInputValue('')} aria-label="Clear input value">
+                  <TimesIcon />
+                </Button>
+              )}
+              <Button variant="plain" onClick={toggleMenuOpen} aria-label="Toggle menu">
```
mcoker, line 167: In the [core typeahead example](https://pf4.patternfly.org/components/menu-toggle#typeahead), this button comes from the menu toggle. Maybe you could wrap the menu toggle icon in a button here? https://github.com/patternfly/patternfly-react/blob/b1132c597ce2c97c8342edcb3cf56d51c430d3b2/packages/react-core/src/components/MenuToggle/MenuToggle.tsx#L69-L75

mcoker (changes requested):
Nice work! Just one thing I noticed is different from the core example.

kmcfaul (approved):
Looks great!

The major difference between the multiple typeahead demo and this one is the toggle. I didn't use `MenuToggle` as it didn't support inputs and wasn't sure what structure we wanted, so rather than edit the component I directly used a `TextInputGroup` structure. I could update mine to use `MenuToggle` after this PR goes in, or leave it as is to showcase that any type of structure can be set up as the toggle trigger - lmk over in my PR.

packages/react-core/src/demos/ComposableMenu/examples/ComposableTypeaheadSelect.tsx
```diff
@@ -0,0 +1,195 @@
...
+        </MenuToggle>
+      }
+      popper={
+        <Menu id="select-menu" onSelect={onMenuSelect} selected={isSelected && inputValue}>
+          <MenuContent>
+            <MenuList>
+              {menuItems.map((itemProps, index) => (
+                <MenuItem
+                  key={itemProps.itemId || itemProps.children}
+                  className={css(focusedItemIndex === index && styles.modifiers.focus, itemProps.className)}
```
kmcfaul, line 181: Would it be worth introducing a property flag for `MenuItem` rather than having the user manually toggle the css through the custom classname? Not a blocker though as I'm good either way.

nicolethoen: I don't think consumers are very used to handling our `styles` object to get classnames, so it probably is more usable to add an 'isFocused' prop.

jeffpuzzo: That's a good point. I've addressed this in the latest commit.

packages/react-core/src/demos/ComposableMenu/examples/ComposableTypeaheadSelect.tsx
jeffpuzzo: This has been addressed. @mcoker can you please re-review when you have a chance?

nicolethoen (approved):

packages/react-core/src/components/MenuToggle/MenuToggle.tsx
```diff
@@ -51,24 +53,57 @@ export class MenuToggleBase extends React.Component<MenuToggleProps> {
...
-        <span className={css(styles.menuToggleControls)}>
-          <span className={css(styles.menuToggleToggleIcon)}>
-            <CaretDownIcon aria-hidden />
-          </span>
-        </span>
-      </React.Fragment>
+        {isTypeahead ? children : <span className="pf-c-menu-toggle__text">{children}</span>}
+        {badge && <span className={css(styles.menuToggleCount)}>{badge}</span>}
+        {isTypeahead ? (
+          <button type="button" className="pf-c-menu-toggle__button" onClick={onClick} aria-label="Menu toggle">
```
mcoker, line 77:
```suggestion
          <button type="button" className={css(styles.menuToggleButton)} aria-expanded={isExpanded} onClick={onClick} aria-label="Menu toggle">
```

mcoker (changes requested):
Just a small nit

tlabaj (commented):
Other than Michael's comment,LGTM

nicolethoen (approved):

mcoker (approved):
Truly a work of art @jeffpuzzo 🤩 Super cool to see you build this in both core and react.

tlabaj (approved):
LGTM!

PR #6906, opened 2022-02-08
feat(MultipleFileUpload): add multiple file upload component **draft**

### LINKED ISSUE #6683: Add Multiple File Upload component
Adds a new component that supports multiple file upload. 

Design: https://marvelapp.com/prototype/69jc198
Core PR: https://github.com/patternfly/patternfly/pull/4548

wise-king-sullyman:
**What**: Closes #6683

**Additional issues**: N/A

TODO:
1. add tests

The change in the Expandable Section component was needed for me to create a toggle with an icon as shown in the core implementation.


mcarrano (commented):
I tried to do some testing on this and have the following observations:

* For the most part this functions as designed, although I was having a hard time testing because of the speed at which the upload takes place. @wise-king-sullyman have you been able to test with really large files or by uploading over a network where the upload might take some time to complete?  I also wasn't able to figure out how to test error handling.
* Related to the above, I tried uploading a large .zip file and nothing happens. This raises two questions. Is the upload limited to the file types listed in the text? I don't think that was the intent but I can understand how this could be misinterpreted. I would expect a prop to set what file types are expected and maybe also to pass what the text should say. It also raises the question of what should happen if I try to upload an unsupported type. Seems like I should get an error.
* One visual thing - for the horizontal version, I notice that the width changes before and after I've uploaded a file. I wouldn't expect that.

@megan-hall @mmenestr would also be good if you can take a look at this.

thatblindgeye:
> I have tested with large (100MB+) files and observed the progress bar actually "filling" by doing so. Error handling I tested by editing the component to fail any submitted file, but I'm not sure how to best make that testable during the review process. If anyone has any ideas for that (or simulating a slow network connection during the "upload" process) please let me know.

As far as simulating a slow connection, I was going to suggest maybe try utilizing the throttling option in Dev Tools (in Chrome it's under the Network tab as a dropdown; default should be "No Throttling"), but it doesn't seem to affect the actual upload for me (only using a 7MB stock image, though). Otherwise, and I'm just spitballing, would adding in a function to simulate either a *really* large file size, or slow network connection, using some combination of setTimeout + "if percentage is not X after Y seconds, throw error" be viable?

> I agree, just having it block the file upload and not giving feedback isn't great, but I'm not sure what a good way to communicate that information would be. If you (or anyone else) have any ideas for that please let me know.

I'd agree with Nicole re: error handling. Only thing I could think of is, would we want to at least allow for error text to be passed in somewhere (maybe either the MultipleFileUploadMain component or as part of the specific file's status since this is multi file)? Or should that be something to hold off on for now/keep as the consumer's responsibility?

mcarrano:
@wise-king-sullyman thanks for the clarifications and all the documentation you added. It will make it easier for consumers to understand how to use the component, for sure. Sounds like you've done your due diligence on testing, so not too worried about that. A couple of further comments-

> I think I'd suggest that error handling - in terms of how to display the error should rely on the consumer as much as possible. Building in messaging about why it failed, that should be something the consumer takes care of because of the number of ways people would want to present that information. This being in beta - im not sure concerned about sorting all that out immediately.

I'm OK with that if it's possible for the consumer to capture that error and provide their own handling. File not supported sounds similar to the duplicate file use case we spoke about last week. Would it be possible to illustrate that handling in the example by opening a simple alert modal that just says this is an unsupported file type? I think that would give consumers some default handling that they could then modify to their liking. The alternative would be to build some error messaging into the drop zone which we could do in a future iteration. Thoughts?

Also, thanks for linking that core issue. I will get this into the upcoming milestone and hopefully that addresses the sizing problem.

wise-king-sullyman:
@mcarrano I added an alert to tell users they tried to upload an unaccepted file type in 8c7358f, is that about how you imagined it?

wise-king-sullyman:
@mcarrano a PF modal or something is definitely what I should've used, f997e23 updates the usage of `alert()`  to instead use the warning variant of the modal.

mcarrano (approved):
This is much better. Thanks for all the work on this @wise-king-sullyman !

mcoker (approved):
Awesome, looks perfect @wise-king-sullyman!

thatblindgeye (approved):

nicolethoen (approved):

packages/react-core/src/components/MultipleFileUpload/examples/MultipleFileUpload.md
```diff
@@ -0,0 +1,92 @@
...
+beta: true
+---
+
+import InProgressIcon from '@patternfly/react-icons/dist/esm/icons/in-progress-icon';
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
+import TimesCircleIcon from '@patternfly/react-icons/dist/esm/icons/times-circle-icon';
+
+File upload - multiple is able to:
+
+- accept one or more files via browse or drag-and-drop
```
tlabaj, line 27: Can you make these sentence case.

packages/react-core/src/components/MultipleFileUpload/examples/MultipleFileUploadBasic.tsx
```diff
@@ -0,0 +1,150 @@
...
+  return (
+    <MultipleFileUpload
+      onFileDrop={handleFileDrop}
+      dropzoneProps={{
+        accept: 'image/jpeg, application/msword, application/pdf, image/png',
+        onDropRejected: handleDropRejected
+      }}
+    >
+      <MultipleFileUploadMain>
+        <MultipleFileUploadTitle>
```
tlabaj, line 112: @mcoker Are the icon and titles flexible (they can go in any order)?  If not I would add a prop rather allowing consumer to add it themselves.

packages/react-core/src/components/MultipleFileUpload/examples/MultipleFileUploadBasic.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <MultipleFileUploadTitleIcon />
+          <MultipleFileUploadTitleText>
+            Drag and drop files here
+            <MultipleFileUploadTitleTextSeparator>or</MultipleFileUploadTitleTextSeparator>
+          </MultipleFileUploadTitleText>
+        </MultipleFileUploadTitle>
+        <MultipleFileUploadButton />
```
tlabaj, line 119: @mcoker Same question here.  I wonder if we need to enforce where the button goes.  this approach may be to flexible leaving room for errors.

packages/react-core/src/components/MultipleFileUpload/examples/MultipleFileUploadBasic.tsx
mcoker: @tlabaj technically they can, but each is assigned to a specific place in the grid currently - so while you can re-order them in the code, the order will not change visually. I suppose the only thing you could do by re-ordering them is to change where they are in the DOM for assistive tech or tab order reasons.

mcoker: @tlabaj technically they can since they're part of a grid with a gap, so you could put the icon after the title if you wanted to, but that does go against the design and I suspect we don't want to allow that. @mcarrano @mmenestr wdyt?

packages/react-core/src/components/MultipleFileUpload/examples/MultipleFileUploadBasic.tsx
mcoker: @tlabaj technically they can be in a different order, but each is assigned to a specific place in the grid currently - so while you can re-order them in the code, the order will not change visually. I suppose the only thing you could do by re-ordering them is to change where they are in the DOM for assistive tech or tab order reasons, and I it's generally a bad practice to have elements appear visually different than they are in the DOM, so I'm inclined to not allow that if we can avoid it.

packages/react-core/src/components/MultipleFileUpload/examples/MultipleFileUploadBasic.tsx
mcarrano: Agree. I wouldn't want those to be reordered.

mturley:
Really awesome work here @wise-king-sullyman !

PR #6917, opened 2022-02-09
fix(Label): editable label should have a tooltip

### LINKED ISSUE #6913: Label: Editable variant should have a tooltip when the label would cause an overflow

MariaAga:
**What**: Closes #6913


MariaAga:
will look into failing tests, looks like the input html tag doesn't get input when focused with enter

tlabaj:
@mcarrano do you think it is worth adding an example of the label with overflow.

mcarrano (approved):
@tlabaj this is probably OK as is. Looks like all label examples that have overflow text do expose a tooltip so it's implied as standard behavior. But if you think it would be clearer to call out a separate example, I'm OK with that, too.

nicolethoen (approved):

nicolethoen:
@MariaAga once the conflicts are resolved, this should be good!

nicolethoen (approved):

thatblindgeye (approved):

PR #6932, opened 2022-02-11
refactor(demo-masthead): use QuestionCircleIcon instead of HelpIcon in Masthead

### LINKED ISSUE #6926: Update masthead help icon used in all full page demos
For consistency with design guidance and current product usage, our full page demos should be using the filled question-circle `QuestionCircleIcon` in the masthead.

Result should look like this:
masthead

MariaAga:
**What**: Closes #6926

**Additional issues**:

### DIFF
--- a/packages/react-core/src/demos/Nav.md
+++ b/packages/react-core/src/demos/Nav.md
@@ -7,6 +7,7 @@ import { DashboardBreadcrumb } from './examples/DashboardWrapper';
 import DashboardHeader from './examples/DashboardHeader';
 import CogIcon from '@patternfly/react-icons/dist/esm/icons/cog-icon';
 import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+import QuestionCircleIcon from '@patternfly/react-icons/dist/esm/icons/question-circle-icon';
 import imgBrand from '@patternfly/react-core/src/components/Brand/examples/pfLogo.svg';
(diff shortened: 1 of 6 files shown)


thatblindgeye (changes requested):
There are a few examples in the Nav file that are throwing an error when I try opening the demos:

- Manual nav
- Horizontal nav with horizontal subnav
- Horizontal nav
 
I'm wondering if it's due to the QuestionCircleIcon not being imported at the top of the Markdown file like some of the other demos, and instead only imported within each example (all three also don't seem to use the DashboardHeader).

nicolethoen (approved):

mcarrano (commented):
I'm seeing the same error as @thatblindgeye . Also, it's not clear how many demos have been updated. Does this just apply to demos on the navigation page or all full page demos?

kmcfaul:
It looks like the change should apply to the docs as well as the full page demos (as DashboardHeader received the update).

I think @thatblindgeye has the right of it about the error, should be resolved by importing the icon at the top of the md file.

MariaAga:
Thanks all! Added the needed import

MariaAga:
Not sure if the test failures are related, I couldn't find where do they fail

thatblindgeye (approved):

mcarrano (approved):
Looks good. Thanks @MariaAga !

nicolethoen (approved):

PR #6949, opened 2022-02-17
fix(NumberInput): better support for min/max

### LINKED ISSUE #6765: NumberInput ignores min/max when increment overruns these values (equals comparison)
Consider the following:

```javascript
import React from 'react';
import { NumberInput } from '@patternfly/react-core';

class UnitThresholdNumberInput extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      value: 1     // <-- initial value is odd, and allows therefore to overrun the max/min
    };
    this.minValue = 0;     // <-- lower boundary to clamp; expected to allow >=0, not < 0
    this.maxValue = 10;    // <-- upper boundary to clamp; expected to allow <=10, not >10

    this.onMinus = () => {
      this.setState({
        value: this.state.value - 2    // <-- increment by 2, when started at 1 this means values are odd on plus, minus
      });
    };

    this.onChange = event => {
      // This only gets executed on a manual input, like key presses ?
      // therefore 'any' value is possible, 1, 2, 3, 4 as it doesn't increment with onPlus, onMinus
      const newValue = isNaN(event.target.value) ? 0 : Number(event.target.value);
      this.setState({
        value: newValue > this.maxValue ? this.maxValue : newValue < this.minValue ? this.minValue : newValue
      });
    };

    this.onPlus = () => {
      this.setState({
        value: this.state.value + 2
      });
    };
  }

  render() {
    const { value } = this.state;

    return (
      <React.Fragment>
        With a minimum value of 0 and maximum value of 10
        However: I can overrun (&gt;10) and underrun (&lt;0) these values when increment happens with 2 and is initially at 1.
        <br />
        <NumberInput
          value={value}
          min={this.minValue}
          max={this.maxValue}
          onMinus={this.onMinus}
          onChange={this.onChange}
          onPlus={this.onPlus}
          inputName="input"
          inputAriaLabel="number input"
          minusBtnAriaLabel="minus"
          plusBtnAriaLabel="plus"
          unit="%"
        />
      </React.Fragment>
    );
  }
}
```

When this code is used, the values increase or decrease by `2` when the buttons are clicked. When the initial value is even, like `0` or `2` this works as expected, but when an initial value of `1` is used you can overrun the max and min value and therefore have `-1` and `11` as values. It feels like the check is performed as an equal for `onPlus`/`onMinus` and min/max comparison. 

~~This is relevant, as an unexpected (like here an `odd` value can be inserted as the cursor keys do not use the expected events. This affects an earlier issue we have: #6753~~ While this is resolved in #6764, it is not impossible to get a value that is offset and does not match the `===` https://github.com/patternfly/patternfly-react/issues/6765#issuecomment-1010687125

Note: why doesn't clamp happen with an `onChange` ? it could be done by invoking the user specified `onChange` wrapped by a PatternFly `onChange` that clamps based on min and max.

boaz0:
**What**: Closes #6765

### DIFF
--- a/packages/react-core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-core/src/components/NumberInput/NumberInput.tsx
@@ -59,17 +59,17 @@ const defaultKeyDownHandler = (args: DefaultKeyDownHandlerArgs) => (event: React
   }
 };
 
-const normalizeBetween = ({ value, min, max }: Pick<NumberInputProps, 'value' | 'min' | 'max'>) => {
-  let result = value;
-  if (min !== undefined && max !== undefined) {
-    result = Math.max(Math.min(value, max), min);
-  } else if (min === undefined && value <= min) {
-    result = min;
-  } else if (max === undefined && value >= max) {
-    result = max;
-  }
-  return result;
-};
+// const normalizeBetween = ({ value, min, max }: Pick<NumberInputProps, 'value' | 'min' | 'max'>) => {
+//   let result = value;
+//   if (min !== undefined && max !== undefined) {
+//     result = Math.max(Math.min(value, max), min);
+//   } else if (max === undefined && value <= min) {
+//     result = min;
+//   } else if (min === undefined && value >= max) {
+//     result = max;
+//   }
+//   return result;
+// };
 
 export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
   value = 0,
@@ -95,21 +95,6 @@ export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
   const numberInputUnit = <div className={css(styles.numberInputUnit)}>{unit}</div>;
   const keyDownHandler =
     inputProps && inputProps.onKeyDown ? inputProps.onKeyDown : defaultKeyDownHandler({ inputName, onMinus, onPlus });
-
-  const [state, setState] = React.useState(value);
-  React.useEffect(() => {
-    setState(value);
-  }, [value]);
-
-  const onUpdate = (evt: React.FormEvent<HTMLInputElement>) => {
-    setState(Number(evt.currentTarget.value));
-    onChange(evt);
-  };
-
-  const onBlur = (evt: React.FormEvent<HTMLInputElement>) => {
-    setState(normalizeBetween({ value: Number(evt.currentTarget.value), min, max }));
-    onChange(evt);
-  };
 
   return (
     <div
@@ -138,13 +123,12 @@ export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
         <input
           className={css(styles.formControl)}
           type="number"
-          value={state}
+          value={value}
           name={inputName}
           aria-label={inputAriaLabel}
           {...(isDisabled && { disabled: isDisabled })}
-          {...(onChange && { onChange: onUpdate })}
+          {...(onChange && { onChange })}
           {...(!onChange && { readOnly: true })}
-          onBlur={onBlur}
           {...inputProps}
           onKeyDown={keyDownHandler}
         />
--- a/packages/react-core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-core/src/components/NumberInput/NumberInput.tsx
@@ -59,6 +59,18 @@ const defaultKeyDownHandler = (args: DefaultKeyDownHandlerArgs) => (event: React
   }
 };
 
+const normalizeBetween = ({ value, min, max }: Pick<NumberInputProps, 'value' | 'min' | 'max'>) => {
+  let result = value;
+  if (min !== undefined && max !== undefined) {
+    result = Math.max(Math.min(value, max), min);
+  } else if (min === undefined && value <= min) {
+    result = min;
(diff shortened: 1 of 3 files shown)


nicolethoen (approved):

packages/react-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -59,6 +59,18 @@ const defaultKeyDownHandler = (args: DefaultKeyDownHandlerArgs) => (event: React
...
+  if (min !== undefined && max !== undefined) {
+    result = Math.max(Math.min(value, max), min);
+  } else if (min === undefined && value <= min) {
+    result = min;
+  } else if (max === undefined && value >= max) {
+    result = max;
+  }
+  return result;
+};
+
```
thatblindgeye, line 73: Are the two `else if` conditions correct here? I wasn't sure if they should be `max === undefined && value <= min` and vice versa or not.

thatblindgeye (commented):
Would we be able to clamp the value to the min or max when a user clicks one of the buttons as well, rather than only `onBlur`, or is this something that we should leave up to the consumer? Currently the value only gets clamped when the input itself has and then loses focus, but the input may not always receive focus if users are just clicking the plus/minus buttons.

packages/react-core/src/components/NumberInput/NumberInput.tsx
kmcfaul: It is a little confusing, if `min` is undefined then `value <= min` wouldn't make a ton of sense.

kmcfaul:
Needs to be rebased @boaz0

packages/react-core/src/components/NumberInput/NumberInput.tsx
boaz0: maybe I can tweak this a little bit but the idea is simple. there are 4 different cases:
1) min and max aren't defined: in that case return we don't do anything but return the value.
2) min and max are defined: check that the value is in between `min` and `max` and if not return `max` or `min`.
3) only min `defined`: if value is smaller than `min` return `min`
4) only max `defined`: if value is bigger than `max` return `max`

Will `switch-case` make it better?

thatblindgeye: I think the `if...else` blocks are fine. Like @kmcfaul mentioned, though, the conditions don't seem to make sense when compared to the 4 cases you've laid out. For example the second condition is checking whether min is undefined, but also whether the value is less than/equal to min (which should be undefined).

When testing out on the build page, if I pass in a value to the `max` prop, but don't pass anything into the `min` prop, the value doesn't clamp to that max value like it's supposed to.

thatblindgeye:
Per a conversation with @tlabaj, the logic for handling the clamping behavior has been moved from being hardcoded inside the component itself to being placed in the example(s). This also adds a custom stepper and threshold example for when there is a threshold _and_ the numbers of steps is greater than 1.

This allows consumers to see one way they can implement logic when dealing with thresholds, with the logic provided aiming to "clamp" the component value. This also allows consumers to ignore the provided logic and pass in their own, should they not want the component value to simply be "clamped"

thatblindgeye:
@mcarrano If you would like to take a look at the latest changes, forgot to tag you in my initial reply above.

packages/react-core/src/components/NumberInput/examples/NumberInput.md
```diff
@@ -159,9 +159,21 @@ class UnitThresholdNumberInput extends React.Component {
...
+      let result = value;
+      if (min !== undefined && max !== undefined) {
+        result = Math.max(Math.min(value, max), min);
+      } else if (max === undefined && value <= min) {
+        result = min;
+      } else if (min === undefined && value >= max) {
+        result = max;
+      }
+      return result;
+    };
```
wise-king-sullyman, line 172: Nit: I think getting rid of the mutation and logically unneeded undefined checks in the second and third if clean this up a lot.

```suggestion
    this.normalizeBetween = (value, min, max) => {
      if (min !== undefined && max !== undefined) {
        return Math.max(Math.min(value, max), min);
      } else if (value <= min) { 
        return min;
      } else if (value >= max) { 
        return max;
      }
      return value;
    };
```

packages/react-core/src/components/NumberInput/examples/NumberInput.md
```diff
@@ -400,3 +412,66 @@ class CustomStepNumberInput extends React.Component {
...
```
wise-king-sullyman, line 442: Nit: I think getting rid of the mutation and logically unneeded undefined checks in the second and third if clean this up a lot.

```suggestion
    this.normalizeBetween = (value, min, max) => {
      if (min !== undefined && max !== undefined) {
        return Math.max(Math.min(value, max), min);
      } else if (value <= min) { 
        return min;
      } else if (value >= max) { 
        return max;
      }
      return value;
    };
```
```

mcarrano (approved):
New example looks great @thatblindgeye !

tlabaj (approved):

wise-king-sullyman (approved):

PR #6953, opened 2022-02-17
Feat(AlertGroup): 6890 alert group overflow

### LINKED ISSUE #6890: Add overflow feature to Alert Group
This feature will allow users to manage overflow on an alert group when the count of active toast alerts exceeds some threshold. Core PR: https://github.com/patternfly/patternfly/pull/4650 

Design doc: https://docs.google.com/document/d/1sW3b6wJNDCeZaxNQv895APoqZROz6BPRyBNj2D4Rmco/edit?usp=sharing

gitdallas:
**What**: Closes #6890 - ability to add an overflow alert which can be clicked, examples using this to limit number of displayed alerts and the ability to click to see those that are hidden

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
+++ b/packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
@@ -79,7 +79,7 @@ class ToastAlertGroup extends React.Component {
 ```
 
 ### Toast alert group with overflow capture
-After a specified number of alerts displayed is reached, we will see an overflow message instead of new alerts.
+After a specified number of alerts displayed is reached, we will see an overflow message instead of new alerts. Alerts asynchronously appended into dynamic AlertGroups with `isLiveRegion` will be announced to assistive technology at the moment the change happens. When the overflow message appears or is updated in AlertGroups with `isLiveRegion`, the `View 1 more alert` text will be read, but the alert message will not be read. screen reader user or keyboard user will need a way to navigate to and reveal the hidden alerts before they disappear. Alternatively, there should be a place that notifications or alerts are collected to be viewed or read later.
 ```js
 import React from 'react';
 import { Alert, AlertGroup, AlertActionCloseButton, AlertVariant, InputGroup } from '@patternfly/react-core';
@@ -93,11 +93,12 @@ class ToastAlertGroup extends React.Component {
       showAll: false
     };
     this.getOverflowMessage = (alertsNumber, showAll = this.state.showAll) => {
-      if (alertsNumber > this.state.maxDisplayed) {
+      const overflow = alertsNumber - this.state.maxDisplayed;
+      if (overflow > 0) {
         if (showAll) {
-          return "Hide " + (alertsNumber - this.state.maxDisplayed) + " alerts";
+          return "Hide " + (overflow) + " alerts";
         }
-        return "View " + (alertsNumber - this.state.maxDisplayed) + " more alerts";
+        return "View " + (overflow) + " more alerts";
       }
       return '';
     };
@@ -224,11 +225,12 @@ class SingularAdditiveAlertGroup extends React.Component {
       showAll: false
     };
     this.getOverflowMessage = (alertsNumber, showAll = this.state.showAll) => {
-      if (alertsNumber > this.state.maxDisplayed) {
+      const overflow = alertsNumber - this.state.maxDisplayed;
+      if (overflow > 0) {
         if (showAll) {
-          return "Hide " + (alertsNumber - this.state.maxDisplayed) + " alerts";
+          return "Hide " + (overflow) + " alerts";
         }
-        return "View " + (alertsNumber - this.state.maxDisplayed) + " more alerts";
+        return "View " + (overflow) + " more alerts";
       }
       return '';
     };
--- a/packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
+++ b/packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
@@ -100,7 +100,7 @@ class ToastAlertGroup extends React.Component {
         return "View " + (alertsNumber - this.state.maxDisplayed) + " more alerts";
       }
       return '';
-    }
+    };
     this.addAlert = (title, variant, key) => {
       this.setState({
         ...this.state,
@@ -123,7 +123,7 @@ class ToastAlertGroup extends React.Component {
     const addSuccessAlert = () => { this.addAlert('Toast success alert', 'success', getUniqueId()) };
     const addDangerAlert = () => { this.addAlert('Toast danger alert', 'danger', getUniqueId()) };
     const addInfoAlert = () => { this.addAlert('Toast info alert', 'info', getUniqueId()) };
-    const onOverflowClick = () => { console.log(this.state.alerts); this.setState({
+    const onOverflowClick = () => { this.setState({
       ...this.state,
       overflowMessage: this.getOverflowMessage(this.state.alerts.length, !this.state.showAll),
       showAll: !this.state.showAll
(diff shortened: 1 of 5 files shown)


packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
```diff
@@ -78,6 +78,85 @@ class ToastAlertGroup extends React.Component {
 }
 ```
 
+### Toast alert group with overflow capture
+After a specified number of alerts displayed is reached, we will see an overflow message instead of new alerts.
```
nicolethoen, line 82: Could you add something like the following to the write up:


Alerts asynchronously appended into dynamic AlertGroups with `isLiveRegion` will be announced to assistive technology at the moment the change happens. When the overflow message appears or is updated in AlertGroups with `isLiveRegion`, the `View 1 more alert` text will be read, but the alert message will not be read. screen reader user or keyboard user will need a way to navigate to and reveal the hidden alerts before they disappear. Alternatively, there should be a place that notifications or alerts are collected to be viewed or read later.

nicolethoen (changes requested):
Could you add unit and integration tests for your new props?

thatblindgeye (commented):
Implementation wise looks good! I did notice a couple of things that may not really be blockers for this PR, but figured I would mention in case before approving:

- For the Toast alert group with overflow capture example, when enough alerts are added and "view X more alerts" is clicked, the "hide X alerts" item is no longer viewable in any way. @mcoker would this be more a core/CSS issue?
- After clicking "view X more alerts" and then clearing each alert (while `showAll` is still set to true), adding more alerts afterwards results in all alerts being shown. Should `showAll` revert to `false` after clearing enough alerts?

mcarrano (commented):
This looks great @gitdallas . The only question I had is how the consumer sets the number of alerts to display before invoking overflow. I expected to see a prop for that, but I do not. Can you explain how this works?

gitdallas:
@mcarrano - the consumer is actually determining when to display the message, what the message is, and preventing as many alerts from showing. the only new properties are what message to show (if any) and a click handler for that message. an example of how it is limiting the number of alerts is here
https://github.com/patternfly/patternfly-react/pull/6953/files#diff-5ccce8e9bd2bb95cb199bfa28f6f18744917c6ec9be7ec5c5292e9f69a618bdeR140

gitdallas:
> Implementation wise looks good! I did notice a couple of things that may not really be blockers for this PR, but figured I would mention in case before approving:
> 
> * For the Toast alert group with overflow capture example, when enough alerts are added and "view X more alerts" is clicked, the "hide X alerts" item is no longer viewable in any way. @mcoker would this be more a core/CSS issue?
> * After clicking "view X more alerts" and then clearing each alert (while `showAll` is still set to true), adding more alerts afterwards results in all alerts being shown. Should `showAll` revert to `false` after clearing enough alerts?

for the overflow of alerts, that is what currently happens without this feature as well. they can always remove alerts to see what overflowed from the page.. but yea, perhaps core should look for a solution to this.

I actually had it revert to false at first, before i added the  "hide all" option.. but after that i just figured if they toggled that, perhaps they'd like it to stay that way.

mcoker:
> For the Toast alert group with overflow capture example, when enough alerts are added and "view X more alerts" is clicked, the "hide X alerts" item is no longer viewable in any way. @mcoker would this be more a core/CSS issue?

cc @mcarrano @mmenestr @mceledonia wdyt? I might expect the overflow alert to be added to the group any time there are too many alerts to be visible in the viewport.

A couple of other ideas...

We could set the page to be scrollable, but that would introduce an extra scrollbar and depending on the amount of alerts, you might scroll way outside of the height of the page/application

Screen Shot 2022-03-02 at 5 35 27 PM

We could make the alert group itself scrollable. This is a quick and dirty way to show that, though we may want to make the alerts scroll up to the top of the window, allow the shadows on the left/right to show, and maybe hide the scrollbar

Screen Shot 2022-03-02 at 5 46 37 PM

mcarrano:
@mcoker I've been trying to decide how to react to your question above about what to do when the list of alerts overflows the screen. I agree that it should not be possible for that to happen. Maybe making the alert group scrollable as you suggested is the right answer, but I agree that requires some attention to the styling.

The other issue here is that the implementation deviates from the original design intent. I must have missed that when we reviewed the component originally in core. The intent was not for that action to expand and collapse the group, but rather the idea was to open a notification drawer or navigate to some other location where the full list of notifications could be seen. 

So I guess the question becomes what to do now. First, can this feature be marked as beta? One option might be to include in the release with the expectation to further iterate. Otherwise I think we need more time to work through the issues. @nicolethoen @tlabaj any thoughts?

@terezanovotna I believe you were looking to get this feature for Insights. What is your priority or timeline at this point? Would what's currently implemented be useful for you? You can see a preview here: https://patternfly-react-pr-6953.surge.sh/components/alert-group#toast-alert-group-with-overflow-capture

gitdallas:
@mcarrano - the implementation is left up to the consumer, they can decide what message to show and what happens when someone clicks on it. the behavior you are seeing in the example is just one possibility.

as for this feature being marked as beta because of the page overflow, the page overflow is currently an issue with alerts. if anything, i believe this feature provides a work-around (leaving it up to the consumer to limit the number of alerts and what happens if clicked).

nicolethoen:
@mcarrano if we intended that clicking the link would open a notification drawer, then it should be a demo - and we likely dont need to build the feature into the AlertGroup component. We can demonstrate how a consumer can build their own 'view more' button, add it to an alert group, and open the drawer when clicked.

mcarrano:
@gitdallas @nicolethoen So based on what you are both saying, we don't need to be concerned with the page overflow issue. Is that right? I'm OK to leave that alone and let the consumer handle it and follow up with a demo and corresponding design guidelines. Should we still be marking this example as beta as a new feature of the alert group, however?

gitdallas:
@mcarrano @nicolethoen - i was playing around a bit with the notification drawer... perhaps it could look better in a full page demo... but it looks kind of weird in an example like this. thoughts?
image
after moving to drawer...
image

if in a full page demo.. would be more intuitive as the drawer could just replace the alerts (as far as location on screen)... i have it so if they click the close "x" in the drawer, it opens it back up in the alerts group

nicolethoen:
@gitdallas I think our discussion of using a Notification drawer is unclear.

We would use a notification drawer to display notifications the way it's done in the notification drawer examples/docs.
We wouldn't use the drawer to display toast notifications. 
basically, the toast notifications would appear, and overflow using the changes you made here. 
But when you click 'view more' then the drawer would open and ALL notifications would be displayed in a list in the drawer. Likely, toast notificaitons would be dismissed in this case.
I hope that's clearer.

@mcarrano we should mark it beta yes.

I think what we need to do to merge this today:

- mark the example as beta (adding `isBeta` after the example's file name after the opening ` ``` ` 
- remove the code that expands the hidden alerts when 'view more' is clicked, since that's not the intended behavior. Eventually we will demonstrate the intended behavior when 'view more' is clicked using a full page demo and a notificaiton drawer.

mcarrano:
+1 to everything @nicolethoen said. I can open a new issue to create a demo in a future milestone to illustrate how the alert group interacts with a drawer.

thatblindgeye (approved):

wise-king-sullyman (approved):

nicolethoen (approved):

packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
```diff
@@ -78,6 +78,79 @@ class ToastAlertGroup extends React.Component {
...
+      console.log('Overflow message clicked');
+    };
+    return (
+      <React.Fragment>
+        <InputGroup style={{ marginBottom: '16px' }}>
+          <button onClick={addSuccessAlert} type="button" className={btnClasses}>Add toast success alert</button>
+          <button onClick={addDangerAlert} type="button" className={btnClasses}>Add toast danger alert</button>
+          <button onClick={addInfoAlert} type="button" className={btnClasses}>Add toast info alert</button>
+        </InputGroup>
+        <AlertGroup isToast isLiveRegion onOverflowClick={onOverflowClick} maxDisplayed={this.state.maxDisplayed} overflowMessage={this.state.overflowMessage}>
```
tlabaj, line 133: Won't `maxDisplayed` get spread to the `ul` and cause an error.?

packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
```diff
@@ -78,6 +78,79 @@ class ToastAlertGroup extends React.Component {
 }
 ```
 
+### Toast alert group with overflow capture
+After a specified number of alerts displayed is reached, we will see an overflow message instead of new alerts. Alerts asynchronously appended into dynamic AlertGroups with `isLiveRegion` will be announced to assistive technology at the moment the change happens. When the overflow message appears or is updated in AlertGroups with `isLiveRegion`, the `View 1 more alert` text will be read, but the alert message will not be read. screen reader user or keyboard user will need a way to navigate to and reveal the hidden alerts before they disappear. Alternatively, there should be a place that notifications or alerts are collected to be viewed or read later.
```
tlabaj, line 82: We may want to be more descriptive of how many alerts will cause the overflow in this example for people who are not reading the code.

packages/react-core/src/components/AlertGroup/examples/AlertGroup.md
```diff
@@ -131,6 +204,81 @@ class SingularAdditiveAlertGroup extends React.Component {
 }
 ```
 
+### Singular dynamic alert group with overflow message
+This alert will appear in the page, most likely in response to a user action.
```
tlabaj, line 208: Same comment here about how to see the overflow.

tlabaj (changes requested):

kmcfaul (approved):

tlabaj (approved):
LGTM!

PR #6955, opened 2022-02-21
feat(Toolbar): allow custom chip group content

### LINKED ISSUE #5668: Add prop to Toolbar component for custom filter tag action
In automation analytics, we will be adding a "Save filter" action to allow users to save the current set of filter tags as a custom filter.  The idea is that this will live as an additional action at the filter tag level:

Screen Shot 2021-04-14 at 12 34 09 PM

In order to do so (and after talking with @nicolethoen who had this great idea), I think it would make sense for the Toolbar to have an additional prop which accepts an arbitrary JSX blob that renders next to the tags when they are shown.  In usage I think it would look something like:

```
<Toolbar
  customFilterAction={
    <Button variant="link">Save filter</Button>
  }
/>

kmcfaul:
**What**: Closes #5668

Adds `customChipGroupContent` property & corresponding example. This will remove the default "Clear all filters button" (to allow customization of order of the buttons) and will append to the filter generated chip groups.


nicolethoen (commented):
At least one unit test should be added to test this new prop.

thatblindgeye (approved):

kmcfaul:
@mcarrano Ah thanks for pointing that out, I missed that a different expandable component is used for that case so I had to add the custom content there as well. Updated

mmenestr:
@kmcfaul looks good to me!

mmenestr (approved):

mcarrano (approved):

wise-king-sullyman (approved):

nicolethoen (approved):

Issue #6945, opened 2022-02-16
Topology RFE: Provide mechanism to hide details at different zoom levels


jeff-phillips-18:
**Description**

As the user zooms out on a graph the nodes become small and the details are not distinguishable, more nodes are drawn and things slow down. There should be a mechanism in place to allow nodes to hide details as this happens to help un-clutter the graph and help with performance.

**Reproduce**
In the integration demo app, set the number of nodes high (1000+), groups and edges as well. Use the node options to add labels, decorators, and status. Select the `Fit to Screen` control.

**Target product / release**
OpenShift 4.11

/cc @christianvogt @serenamarie125

PR #6958, opened 2022-02-21
feature(Topology): Add mechanism to hide details at different zoom levels

jeff-phillips-18:
**What**: 
Closes https://github.com/patternfly/patternfly-react/issues/6945

Add scale details settings and a hook to get the current scale details level. This can be used by applications to hide details as they see fit.

Also add some performance improvements around drawing items that are not viewable.

/cc @christianvogt


packages/react-topology/src/elements/BaseGraph.ts
```diff
@@ -73,6 +85,27 @@ export default class BaseGraph<E extends GraphModel = GraphModel, D = any> exten
...
+        detailsLevel = ScaleDetailsLevel.low;
+      } else if (this.scale <= this.scaleDetailsSettings[ScaleDetailsLevel.medium]) {
+        detailsLevel = ScaleDetailsLevel.medium;
+      }
+    }
+    this.setDetailsLevel(detailsLevel);
+  }
+
+  getDetailsLevel(): ScaleDetailsLevel {
+    return this.detailsLevel;
```
christianvogt, line 102: This should be a computed value based on scale.
If we want to support overriding the value it should then fallback to a private observable that always takes precedence when set.

packages/react-topology/src/elements/BaseGraph.ts
```diff
@@ -150,6 +183,15 @@ export default class BaseGraph<E extends GraphModel = GraphModel, D = any> exten
...
       this.getController().fireEvent(GRAPH_POSITION_CHANGE_EVENT, { graph: this });
+      if (this.scaleDetailsSettings) {
+        let detailsLevel = ScaleDetailsLevel.high;
+        if (scale <= this.scaleDetailsSettings[ScaleDetailsLevel.low]) {
+          detailsLevel = ScaleDetailsLevel.low;
+        } else if (scale <= this.scaleDetailsSettings[ScaleDetailsLevel.medium]) {
+          detailsLevel = ScaleDetailsLevel.medium;
+        }
+        this.setDetailsLevel(detailsLevel);
+      }
```
christianvogt, line 194: Unnecessary here if the details level is computed.

packages/react-topology/src/types.ts
```diff
@@ -239,6 +249,9 @@ export interface Graph<E extends GraphModel = GraphModel, D = any> extends Graph
   setScaleExtent(scaleExtent: ScaleExtent): void;
   getScale(): number;
   setScale(scale: number): void;
+  setDetailsLevelSettings(settings: ScaleDetailsSettings): void;
```
christianvogt, line 252:
```suggestion
  setDetailsLevelSettings(settings: ScaleDetailsSettings | undefined): void;
  getDetailsLevelSettings(): Readonly<ScaleDetailsSettings> | undefined;
```

packages/react-topology/src/elements/BaseGraph.ts
jeff-phillips-18: Lets add the override if we determine it is wanted and remove the `setDetailsLevel` for now.

packages/react-topology/src/types.ts
```diff
@@ -239,6 +249,9 @@ export interface Graph<E extends GraphModel = GraphModel, D = any> extends Graph
   setScaleExtent(scaleExtent: ScaleExtent): void;
   getScale(): number;
   setScale(scale: number): void;
+  setDetailsLevelSettings(settings: ScaleDetailsSettings | undefined): void;
+  getDetailsLevelSettings(): Readonly<ScaleDetailsSettings> | undefined;
+  getDetailsLevel(): ScaleDetailsLevel;
```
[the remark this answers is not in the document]
jeff-phillips-18: Allowing the setting to be undefined allows the application to never be concerned with the scale details level.

christianvogt:
lgtm

nicolethoen (approved):

PR #6962, opened 2022-02-22
demo: date select dropdown

### LINKED ISSUE #6460: Date select demo
Per https://github.com/patternfly/patternfly-design/issues/1082 want to create a new "select list" component for date picking, which includes a list of relative date ranges. Additionally, when a date range is picked from the list, we would want the actual dates to be written out with the toggle itself.

image 

cc @mmenestr

kmcfaul:
**What**: Closes #6460

Adds a new date select demo to the composable menu demos page (can be moved if it's better in another place).

@mcarrano @mmenestr Does this match the design well enough re: the lighter/smaller grey text for the date? Also is the "custom" option supposed to function differently for this specific demo or will that be a follow up?

### DIFF
--- a/packages/react-core/src/demos/ComposableMenu/examples/ComposableDateSelect.tsx
+++ b/packages/react-core/src/demos/ComposableMenu/examples/ComposableDateSelect.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { MenuToggle, Menu, MenuContent, MenuList, MenuItem, Popper, Flex } from '@patternfly/react-core';
+import { MenuToggle, Menu, MenuContent, MenuList, MenuItem, Popper, Flex, FlexItem } from '@patternfly/react-core';
 
 export const ComposableSimpleDropdown: React.FunctionComponent = () => {
   const [isOpen, setIsOpen] = React.useState(false);
@@ -91,18 +91,9 @@ export const ComposableSimpleDropdown: React.FunctionComponent = () => {
   };
 
   const toggle = (
-    <MenuToggle
-      ref={toggleRef}
-      onClick={onToggleClick}
-      isExpanded={isOpen}
-      style={
-        {
-          '--pf-c-menu--Width': '250px'
-        } as React.CSSProperties
-      }
-    >
+    <MenuToggle ref={toggleRef} onClick={onToggleClick} isExpanded={isOpen} style={{ minWidth: '250px' }}>
       <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
-        {toggleText[selected]}
+        <FlexItem>{toggleText[selected]}</FlexItem>
         {dateText[selected]}
       </Flex>
     </MenuToggle>
--- a/packages/react-core/src/demos/ComposableMenu/examples/ComposableDateSelect.tsx
+++ b/packages/react-core/src/demos/ComposableMenu/examples/ComposableDateSelect.tsx
@@ -1,0 +1,115 @@
+import React from 'react';
+import { MenuToggle, Menu, MenuContent, MenuList, MenuItem, Popper } from '@patternfly/react-core';
+
+export const ComposableSimpleDropdown: React.FunctionComponent = () => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const toggleRef = React.useRef<HTMLButtonElement>();
+  const menuRef = React.useRef<HTMLDivElement>();
+  const [selected, setSelected] = React.useState<number>(0);
+
+  const handleMenuKeys = (event: KeyboardEvent) => {
+    if (!isOpen) {
+      return;
+    }
+    if (menuRef.current.contains(event.target as Node) || toggleRef.current.contains(event.target as Node)) {
+      if (event.key === 'Escape' || event.key === 'Tab') {
+        setIsOpen(!isOpen);
+        toggleRef.current.focus();
+      }
+    }
+  };
+
+  const handleClickOutside = (event: MouseEvent) => {
+    if (isOpen && !menuRef.current.contains(event.target as Node)) {
+      setIsOpen(false);
+    }
+  };
+
+  React.useEffect(() => {
+    window.addEventListener('keydown', handleMenuKeys);
+    window.addEventListener('click', handleClickOutside);
+    return () => {
+      window.removeEventListener('keydown', handleMenuKeys);
+      window.removeEventListener('click', handleClickOutside);
+    };
+  }, [isOpen, menuRef]);
+
+  const onToggleClick = (ev: React.MouseEvent) => {
+    ev.stopPropagation(); // Stop handleClickOutside from handling
+    setTimeout(() => {
+      if (menuRef.current) {
+        const firstElement = menuRef.current.querySelector('li > button:not(:disabled)');
+        firstElement && (firstElement as HTMLElement).focus();
+      }
+    }, 0);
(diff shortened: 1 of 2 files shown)


mmenestr:
This looks great and matches the design well - the only thing is that since the font is smaller on the grey text, it looks a bit misaligned - would it be possible to center align the grey text with the black text, instead of bottom aligning? Just want to see what that would look like, though I realized that might look even weirder haha

For this demo, I would just remove the "custom" option from the dropdown - it was for if we were to pair this with a date picker, and we wanted to display the date range selected in a date picker.

In terms of placement, I wonder if it would make sense to add this to the select component as well or instead? @mcarrano what do you think

mmenestr:
Ok, so maybe we just go ahead and keep it in the composable menu for now then!

nicolethoen (commented):
This looks good to me, I just wonder if you should prevent the width of the menu from changing when an option is selected - and prevent the width of the menu toggle from changing when the value is longer? 

It'd make the menu toggle much wider by default - I'm not sure what design's recommendation would be here @mcarrano @mmenestr ?

kmcfaul:
@mcoker @mattnolting Do we have a `pf-u-` variable to vertically center align the text? Trying to figure out how to adjust the small grey text in the toggle.

mcoker:
@mmenestr here's what the small text looks like centered in the toggle

Screen Shot 2022-03-01 at 10 39 04 AM

mmenestr:
In terms of toggle width, I think varying widths is fine, but I think we'd probably want to find a happy medium between the smallest and biggest size it could be, and make that the default width. So for example, let's say we decide on a default width of 250px. If a selection ends up taking more space than that width, then the toggle expands to accommodate the text. But if the text ends up taking less space than that default width, the toggle width remains 250px, and you'd just have extra space between the text and the end of the toggle. Is that something that's possible to do?

Also the vertically centered text looks great - let's go with that!

packages/react-core/src/demos/ComposableMenu/examples/ComposableDateSelect.tsx
```diff
@@ -0,0 +1,112 @@
...
+    3: (
+      <small className="pf-u-color-200">
+        ({dateString(new Date(new Date().setDate(date.getDate() - 14)))} - {dateString(date)})
+      </small>
+    )
+  };
+
+  const toggle = (
+    <MenuToggle ref={toggleRef} onClick={onToggleClick} isExpanded={isOpen}>
+      {toggleText[selected]} {dateText[selected]}
```
mcoker, line 95: @kmcfaul for the centered text, I used a flex layout to wrap these, added an `8px` ("sm") spacer, and added `align-items: center`

kmcfaul:
Updated to space and center the date string.

@mmenestr Regarding the minimum/starting width of the toggle, you can add `style={{minWidth: '250px'}}` to the `MenuToggle` in the example code to test this, but I didn't build it into the demo yet because with the current options, all it does is make the `Today` toggle very wide with white space, and all other options causes the toggle width to grow anyways. LMK if I should still add the min-width as 250px or a larger value.

mcarrano (commented):
I will defer to @mmenestr on this one.

mmenestr:
Just tried the 250px and I think it's better that way (personally), just so that when the date changes, the shift in the UI isn't as big. Also want to make sure there's going to be a 8px spacer between the black text and the smaller grey text?

nicolethoen:
many of the other demos on that page set widths using props we already expose for menu toggle and popper. I’d take a peek at that rather than explicitly using the style prop

wise-king-sullyman (approved):

mmenestr:
This is what i'm seeing in the preview so I'm confused haha there's no space between the 2 strings, and the width of the toggle is under 200px. Is it just that the changes haven't loaded yet?
image

kmcfaul:
@mmenestr My bad, I had the update fixing that committed but didn't notice the push failed. It should be rebuilding now!

nicolethoen (approved):

packages/react-core/src/demos/ComposableMenu/examples/ComposableDateSelect.tsx
```diff
@@ -0,0 +1,115 @@
...
+    3: (
+      <small className="pf-u-color-200">
+        ({dateString(new Date(new Date().setDate(date.getDate() - 14)))} - {dateString(date)})
+      </small>
+    )
+  };
+
+  const toggle = (
+    <MenuToggle ref={toggleRef} onClick={onToggleClick} isExpanded={isOpen} style={{ minWidth: '250px' }}>
+      <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
```
mcoker, line 95: this is currently a `div` (block element) - technically it should be a `span` (inline element) since it's in a `<button>` to be valid HTML. Not a big deal though, it isn't going to break anything.

Also just noticed that the text truncation for the toggle text isn't working because it doesn't know how to apply truncation to things in the flex layout. This is what it looks like now with long text.


Screen Shot 2022-03-04 at 10 05 02 AM
 
Working with what's there, we could use the `.pf-u-text-truncate` utility to truncate the text in both flex children

Screen Shot 2022-03-04 at 10 07 12 AM

Instead of a flex layout, we could just wrap the 2 blocks of text (regular text and grey date text) in a `span` and apply `vertical-align: middle;` to both - that should work...

Screen Shot 2022-03-04 at 10 42 59 AM

mcoker (commented):
LGTM, I don't know if my comment is a blocker - @mcarrano @mmenestr wdyt?

mcarrano:
> LGTM, I don't know if my comment is a blocker - @mcarrano @mmenestr wdyt?

@mcoker I think we can live with this (the truncation issue), but I'll defer to @mmenestr as she is more aware of how this will be used. Do you see any situations where the text won't fit on a single line?

mmenestr:
The only situation where I can see a space issue happening is when the toolbar is full of other elements, and this toggle would have limited space to exist. In that case developers might want to set a max width and have the text truncate to avoid the toolbar going beyond one line when in full screen. 

When truncating is necessary I would probably just truncate the end of the string, and not truncate **both** the black text AND the grey... basically I'd just truncate wherever the text cuts off and that's it. Does that make sense?

All things considered I don't see this being a common occurrence, but while we're here, might as well address it? Unless you want to push this out ASAP.

mcarrano:
I don't see a huge urgency around this. @kmcfaul @mcoker is this something that can be addressed quickly? If not, I wouldn't hold the release for it. Thoughts @nicolethoen @tlabaj ?

kmcfaul:
I can put up an update that changes to `span` and `style` to set vertical alignment and margin padding that is lost after removing the `Flex` wrapping.

tlabaj (approved):
LGTM

mcoker (approved):
LGTM! The only thing about the space between the 2 blocks of text is that since they're regular inline elements (not flex children), the white space between them should be preserved<sup>[1]</sup>. So adding a margin will add whatever the margin is plus the space character. If you wanted to be more exact, you could remove the whitespace (and any newline/return chars) between the two spans, then the margin would be exact. If you want to preserve newlines for formatting reasons you can do this in HTML, tho I bet there's a fancy react way to trim space.

```
<span>foo</span><!--
--!><span>bar</span>
```

Also instead of `8px` you could use `var(--pf-global--spacer--sm)`


[1]. Some applications remove white space, I'm assuming to cut down on page weight as part of some sort of minification step.

PR #6963, opened 2022-02-22
fix(NotificationDrawer): Drawer receives focus after opening

### LINKED ISSUE #5888: Focus should shift to notification drawer when opened
When a user opens the notification drawer, focus should shift into the notification drawer so that a keyboard or screen reader user can continue to navigate the way that a mouse user would (similar to our drawer component). I imagine that we may need to add this to the demo instead of the component since the notification drawer itself doesn't handle its opening behavior.

Go to the [notification drawer demos](https://www.patternfly.org/v4/components/notification-drawer/react-demos), Tab through the top nav, select the notification icon with enter, and notice that focus does not shift and you still have to go through the entire page to get to the notification drawer.

Example/demo bug

thatblindgeye:
**What**: Closes #5888 

- Updated the NotificationDrawer component to be similar to the Tab component by using forwardRef to handle placing focus on the NotificationDrawer
- Added instructions on the React Demos page for how to manage focus

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
@@ -8,7 +8,7 @@ export interface NotificationDrawerProps extends React.HTMLProps<HTMLDivElement>
   children?: React.ReactNode;
   /**  Additional classes added to the notification drawer */
   className?: string;
-  /** Forwarded ref */
+  /** @hide Forwarded ref */
   innerRef?: React.Ref<any>;
 }
 
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
@@ -18,7 +18,7 @@ const NotificationDrawerBase: React.FunctionComponent<NotificationDrawerProps> =
   innerRef,
   ...props
 }: NotificationDrawerProps) => (
-  <div ref={innerRef} tabIndex={0} {...props} className={css(styles.notificationDrawer, className)}>
+  <div ref={innerRef} {...props} className={css(styles.notificationDrawer, className)}>
     {children}
   </div>
 );
--- a/packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
+++ b/packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
@@ -8,15 +8,21 @@ export interface NotificationDrawerProps extends React.HTMLProps<HTMLDivElement>
   children?: React.ReactNode;
   /**  Additional classes added to the notification drawer */
   className?: string;
+  /** Forwarded ref */
+  innerRef?: React.Ref<any>;
 }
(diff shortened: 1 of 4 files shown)


nicolethoen (commented):
I was expecting the focus to move to the first interact-able element in the notification drawer. Is there a benefit to placing it on the notification drawer itself rather than on an interact-able element?

thatblindgeye:
> I was expecting the focus to move to the first interact-able element in the notification drawer. Is there a benefit to placing it on the notification drawer itself rather than on an interact-able element?

No explicit benefit per se, at least with Voice Over. It looks like making the first tabbable element focusable instead provides the same context. That is, VO announces the contents of the Notification Drawer itself instead of just something like "Actions, button" (which is something I would've been wary of).

I'll push an update that focuses that first item instead momentarily.

packages/react-core/src/components/NotificationDrawer/__tests__/Generated/__snapshots__/NotificationDrawer.test.tsx.snap
```diff
@@ -1,11 +1,12 @@
...
-  className="pf-c-notification-drawer ''"
+<NotificationDrawerBase
+  className="''"
+  innerRef={null}
 >
   <div>
     ReactNode
   </div>
-</div>
+</NotificationDrawerBase>
```
wise-king-sullyman, line 11: I'm unsure if the test here is still really testing the things that need to be tested based on the snapshot.

packages/react-core/src/components/NotificationDrawer/__tests__/__snapshots__/NotificationDrawer.test.tsx.snap
```diff
@@ -1,9 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`className is added to the root element 1`] = `"pf-c-notification-drawer extra-class"`;
+exports[`className is added to the root element 1`] = `"extra-class"`;
 
 exports[`renders with PatternFly Core styles 1`] = `
-<div
-  className="pf-c-notification-drawer"
+<NotificationDrawerBase
+  innerRef={null}
```
wise-king-sullyman, line 7: I'm also unsure if this test is still really testing what it's intending to now.

wise-king-sullyman (commented):
Overall looks great! I just have a couple questions around the snapshot test updates.

wise-king-sullyman (approved):

packages/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx
```diff
@@ -8,15 +8,21 @@ export interface NotificationDrawerProps extends React.HTMLProps<HTMLDivElement>
   children?: React.ReactNode;
   /**  Additional classes added to the notification drawer */
   className?: string;
+  /** Forwarded ref */
```
nicolethoen, line 11: can we also hide this prop?

tlabaj (approved):

Issue #6756, opened 2022-01-10
Select with variant typeahead is not updated correctly on unfocus if the list is closed


josejulio:
**Describe the issue. What is the expected and unexpected behavior?**
On `Select` variant `typeahead` the selection is not updated correctly if it was manually modified and the unfocus happens with the list not visible

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
https://codesandbox.io/s/nostalgic-heisenberg-i1u44?file=/index.js

Reproducer:

1. Select one element from the list (e.g. "One")
2. Write something or delete the content of the field
3. Click the field again if the list is showing and the list will hide (or just press tab)
4. Click away, e.g. click HERE (or just press tab)

**Expected**: Field value is set to the current selected value

**Actual**: Field is set to what was written (i.e. unchanged)

**Note**: It is more natural to reproduce with tabs, as you don't have to "click" to hide the list. Pressing tab does this.

![Peek 2022-01-10 11-51](https://user-images.githubusercontent.com/3845764/148822300-496063bb-3b44-4379-8f57-f5f08644368b.gif)

![Peek 2022-01-10 11-56](https://user-images.githubusercontent.com/3845764/148823047-d6214eb6-c92a-4b94-9e62-8217ce6ccf0b.gif)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug, not blocking

**What is your product and what release version are you targeting?**
Not a critical bug

PR #6966, opened 2022-02-22
fix(Select): prevent double enter, cover additional onclose cases

### LINKED ISSUE #6245: Select: clicking twice Enter in the typeahead variant will select the first entry

Use this example code. https://www.patternfly.org/v4/components/select/#typeahead
Select New York. Click enter (this will open the dropdown, and enter again. This will close the dropdown menu and select Alabama which is the first entry in the menu.

Unclear if this is a bug or expected behavior. 
Personally I would expect enter to open and close the menu and leave the selected option untouched, but feel free to close if this is already the desired behavior.

Cockpit Machines.

Related BZ https://bugzilla.redhat.com/show_bug.cgi?id=1994867

kmcfaul:
**What**: Closes #6245 & #6756

#6245 - the `enter` selection behavior is now prevented for typeahead until the arrow keys are used to move to an option.

#6756 - added additional `onClose` behaviors to two cases (when input div is manually clicked closed, and when tab is used to navigate off the input div) where the menu is closed but the internal state isn't reset.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -764,12 +764,14 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
               if (shiftKey) {
                 // close toggle if shift key and tab on input
                 this.onToggle(false);
+                this.onClose();
               } else {
                 // tab to first tabbable item in footer
                 if (tabbableItems[0]) {
                   tabbableItems[0].focus();
                 } else {
                   this.onToggle(false);
+                  this.onClose();
                 }
               }
             } else {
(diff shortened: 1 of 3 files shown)


thatblindgeye (commented):
Looks good! 

Would it be worth adding that `onClose` method for a typeahead variant that has a footer as well rather than just one without a footer? Currently if I add a footer to the [Typeahead example](https://patternfly-react-pr-6966.surge.sh/components/select#typeahead), the input doesn't revert to the current selection when I tab out of the footer and the toggle closes.

nicolethoen (approved):
I like Eric's question
outside of that, I cannot find any _new_ issues :)

kmcfaul:
Sure that makes sense, added. Footer logic was in another logic branch I skipped over initially for this PR.

thatblindgeye (approved):

gitdallas (approved):

PR #6974, opened 2022-02-23
feature(topology): Add basic Grid layout to Topology component

lucasponce:
Closes https://github.com/patternfly/patternfly-react/issues/6975

This PR adds a basic grid layout to the topology component:
image

It's basic, but it offers an "id" ordered grid that may be useful at scale scenarios:
image

I used it as a learning exercise to get familiar with the topology model, but I think it could be useful in the "demo-app" for layout comparisons.

I'd like to add a couple of more algorithms (in the following/separated PRs) that have good results at scale, like the concentric and breadth-first.

The rationale of this approach is that layout algorithms can introduce performance penalties at scale, so, having "basic" layouts with some tradeoffs could be a good starting point for alternatives.

P.S.: I'm not familiar with the code style of the Patternfly project, please, feel free to suggest how it could be the best way to propose these changes.

Thanks.

cc @jeff-phillips-18


jeff-phillips-18 (approved):
LGTM

nicolethoen (approved):

PR #6986, opened 2022-03-01
fix(Menu): screen reader correctly announces on view more

### LINKED ISSUE #5840: Menu - view more does not announce to screen reader focus change
When using the view more example, focus correctly changes to the next menu item, however it doesn't announce to a screen reader that the focus has changed. I played with an `aria-activedescendant` approach, however in this investigation I realized that with this approach focus would need to be handled differently (i.e. for an `aria-activedescendant` approach, focus would need to stay on the parent menu and then the attributes would communicate what menu item has focus, visually it would probably look more like the Typeahead Select). We should take a look at what ways we can use to communicate this change of focus. `aria-activedescendant` was an idea, but maybe there's an alternative that we can find to fix this issue.

Use VO and shift focus through the menu items, select the view more button, afterwards VO announces "end of list" instead of shifting focus to next item.

Bug

thatblindgeye:
**What**: Closes #5840 

In addition to resolving the linked issue, this PR also fixes an issue with the View More example where using keyboard to "click" the view more button did not work correctly (essentially resulting in an error).

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Menu/MenuItem.tsx
+++ b/packages/react-core/src/components/Menu/MenuItem.tsx
@@ -58,13 +58,15 @@ export interface MenuItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'onC
   isOnPath?: boolean;
   /** Accessibility label */
   'aria-label'?: string;
+  /** Forwarded ref */
+  innerRef?: React.Ref<HTMLAnchorElement | HTMLButtonElement>;
 }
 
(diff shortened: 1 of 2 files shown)


kmcfaul (approved):
LGTM

nicolethoen (commented):
LGTM! one nit

nicolethoen (approved):

kmcfaul (approved):

PR #6989, opened 2022-03-01
feat(CodeEditor): 6663 - allow for keyboard shortcut menu, add example

### LINKED ISSUE #6663: Code editor: demo for adding keyboard shortcuts
The Monaco editor already supports keyboard shortcuts, but want to create a demo that shows how to expose these in the toolbar. Here is the current design: https://marvelapp.com/prototype/15bb10j4/screen/81521392  This includes a popover with shortcuts. Can this be built using text and chips? I'd like to leverage existing components and not create anything unique if it can be avoided. We can iterate on the design, if needed.

gitdallas:
**What**: Closes #6663
allow for keyboard shortcut menu, added an example

**Additional issues**:


packages/react-code-editor/src/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -116,6 +116,62 @@ class BasicCodeEditor extends React.Component {
...
+        keys: ["Opt","F1"],
+        description: "Accessibility help",
+      },
+      {
+        keys: ["F1"],
+        description: "View all editor shortcuts",
+      },
+      {
+        keys: ["Ctrl", "Space"],
+        description: "Activate auto copmlete",
```
kmcfaul, line 153: spelling nit - complete

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -551,6 +569,39 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
...
                   )}
                 </div>
+              }
+              {<div className={css(styles.codeEditorHeaderMain)}>{headerMainContent}</div>}
+              {!!shortcuts.length && (
+                <div className="pf-c-code-editor__keyboard-shortcuts">
+                  <Popover
+                    aria-label="Shortcuts"
+                    bodyContent={
+                      <Grid span={6} hasGutter>
```
kmcfaul, line 579: I wonder how much customization we want to expose for this popover. Should we allow users to customize popover props like direction, or the content of the body? What do others think?

There's also some additional white space created compared to the design because of Grid breaking on the word when it hits the width, but I'm not sure there's a good way to handle that beyond setting a larger popup width or adjust the grid gap css.

mcarrano (commented):
This looks great @gitdallas . My only question is for you, @nicolethoen . I understand that these shortcuts are defined by the Monaco editor. Is there a way we should be extracting them from the editor or does the consumer just need to build the contents of the popover manually as done here? If the later, then I think this is good to go.

nicolethoen (changes requested):
This looks great! two requests:

@mcarrano is right that the three shortcuts included in the example are part of monaco - but I like that if anyone adds additional keyboard shortcuts, they get to define the contents of the popover (or if they want to display the shortcuts in a different language). My suggestion is to add a note above the example which explains that these are the shortcuts that we recommend describing in the popover since they are monaco features.

I also agree with @kmcfaul that we should allow people to define/override the popover props. probably with a `shortcutsPopoverProps` component. And they should be spread into the popover. They definitely need to be able to override the popover's aria-label if they want, for example.

gitdallas:
@kmcfaul @nicolethoen - part of the reason things are getting smushed is the close button... if I put heading (even if empty) to push it below the close button, it won't be as smushed.. of course descriptions could potentially get much bigger anyway.
image

gitdallas:
I went ahead and moved the entire popover props up to the consumer, including bodyContent

nicolethoen (approved):

mcarrano (changes requested):
@gitdallas This looks great, but somehow when you added the extra description to the example, it got added to the heading instead of below the heading.

Screen Shot 2022-03-03 at 9 50 38 AM

Can you fix?

nicolethoen (approved):

gitdallas:
> @gitdallas This looks great, but somehow when you added the extra description to the example, it got added to the heading instead of below the heading.
> 
> Screen Shot 2022-03-03 at 9 50 38 AM
> 
> Can you fix?

@mcarrano  - updated

mcarrano (approved):
Looks great. Thanks @gitdallas !

mcoker (changes requested):

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -551,6 +568,16 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
                     <CodeEditorContext.Provider value={{ code: value }}>{customControls}</CodeEditorContext.Provider>
                   )}
                 </div>
+              }
+              {<div className={css(styles.codeEditorHeaderMain)}>{headerMainContent}</div>}
+              {!!shortcutsPopoverProps.bodyContent && (
+                <div className="pf-c-code-editor__keyboard-shortcuts">
+                  <Popover {...shortcutsPopoverProps}>
+                    <Button variant={ButtonVariant.link}>
```
mcoker, line 576: This will wrap the icon in an element and apply the appropriate layout and spacing.

```suggestion
                    <Button variant={ButtonVariant.link} icon={<HelpIcon />}>
```

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -551,6 +568,16 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
...
+                      <HelpIcon /> View shortcutss
```
mcoker, line 577:
```suggestion
                      View shortcuts
```

mcoker (changes requested):

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -551,6 +567,19 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
...
                   )}
                 </div>
+              }
+              {<div className={css(styles.codeEditorHeaderMain)}>{headerMainContent}</div>}
+              {!!shortcutsPopoverProps.bodyContent && (
+                <div className="pf-c-code-editor__keyboard-shortcuts">
+                  <Popover {...shortcutsPopoverProps}>
+                    <Button variant={ButtonVariant.link}>
+                      <span className="pf-c-button__icon pf-m-start">
+                        <i className="pf-icon pf-icon-help" aria-hidden="true"></i>
```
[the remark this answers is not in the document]
gitdallas: done

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
gitdallas: oops

wise-king-sullyman (approved):

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -551,6 +568,16 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
                     <CodeEditorContext.Provider value={{ code: value }}>{customControls}</CodeEditorContext.Provider>
                   )}
                 </div>
+              }
+              {<div className={css(styles.codeEditorHeaderMain)}>{headerMainContent}</div>}
+              {!!shortcutsPopoverProps.bodyContent && (
+                <div className="pf-c-code-editor__keyboard-shortcuts">
+                  <Popover {...shortcutsPopoverProps}>
+                    <Button variant={ButtonVariant.link} icon={<HelpIcon />}>
+                      View shortcuts
```
kmcfaul, line 577: Should be a prop to allow localization. 

Everything else lgtm

kmcfaul (changes requested):

mcoker (approved):
nice!!

packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -551,6 +572,16 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
                     <CodeEditorContext.Provider value={{ code: value }}>{customControls}</CodeEditorContext.Provider>
                   )}
                 </div>
+              }
+              {<div className={css(styles.codeEditorHeaderMain)}>{headerMainContent}</div>}
+              {!!shortcutsPopoverProps.bodyContent && (
+                <div className="pf-c-code-editor__keyboard-shortcuts">
+                  <Popover {...shortcutsPopoverProps}>
+                    <Button variant={ButtonVariant.link} icon={<HelpIcon />}>
+                      {shortcutsPopoverProps.shortcutTitleText || 'View shortcuts'}
```
nicolethoen, line 581: maybe you can add 'View shortcuts' as a default value for shortcutsPopoverProps.shortcutTitleText so it's explicitly called out in the docs.

nicolethoen (changes requested):
So good - just a couple documentation notes:

Could you add 'ShortcutProps' to the propsComponents array at the top of the markdown file so it's definition is included in the API docs?

kmcfaul (approved):

tlabaj (approved):

PR #6994, opened 2022-03-02
refactor(Nav): make nav group titles optional

### LINKED ISSUE #6770: Navigation section titles
Related to: https://github.com/patternfly/patternfly/issues/4602

Need to make visible section titles optional. If section title not visible, section title text should be applied to to `aria-label`.

wise-king-sullyman:
**What**: Closes #6770

**Additional issues**: N/A

I chose to proceed with the second option I suggested, and implemented the suggestions of @thatblindgeye and @mcoker in that solution.

### DIFF
--- a/packages/react-core/src/components/Nav/NavGroup.tsx
+++ b/packages/react-core/src/components/Nav/NavGroup.tsx
@@ -5,7 +5,7 @@ import { getUniqueId } from '../../helpers/util';
 
 export interface NavGroupProps extends React.HTMLProps<HTMLDivElement> {
   /** Title shown for the group */
-  title: string;
+  title?: string;
   /** Anything that can be rendered inside of the group */
   children?: React.ReactNode;
   /** Additional classes added to the container */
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/Nav/NavGroup.tsx
```diff
@@ -20,12 +20,21 @@ export const NavGroup: React.FunctionComponent<NavGroupProps> = ({
...
-  </section>
-);
+}: NavGroupProps) => {
+  if (!title && !props['aria-label']) {
+    // eslint-disable-next-line no-console
+    console.warn("For accessibility reasons an aria-label should be specified on nav groups if a title isn't");
+  }
+
+  return (
+    <section className={css(styles.navSection, className)} aria-labelledby={id} {...props}>
```
mcoker, line 30: should `aria-labelledby` be conditionally rendered if `title` is set?

packages/react-core/src/components/Nav/examples/Nav.md
```diff
@@ -55,6 +55,8 @@ class NavDefaultList extends React.Component {
 
 ### Grouped
 
+Note: to keep nav groups accessible an aria label should be supplied if a title prop is not passed.
```
mcoker, line 58: I'm not sure if there is a convention for referring to attrs/props, but it might be good to write it this way?

```suggestion
Note: to keep nav groups accessible an `aria-label` should be supplied if the `title` prop is not passed.
```

packages/react-core/src/components/Nav/NavGroup.tsx
```diff
@@ -20,12 +20,21 @@ export const NavGroup: React.FunctionComponent<NavGroupProps> = ({
...
-}: NavGroupProps) => (
-  <section className={css(styles.navSection, className)} aria-labelledby={id} {...props}>
-    <h2 className={css(styles.navSectionTitle)} id={id}>
-      {title}
-    </h2>
-    <ul className={css(styles.navList, className)}>{children}</ul>
-  </section>
-);
+}: NavGroupProps) => {
+  if (!title && !props['aria-label']) {
```
mcoker, line 24: 🥳

packages/react-core/src/components/Nav/NavGroup.tsx
wise-king-sullyman: Yes I do believe so, good catch!

thatblindgeye (approved):

mcoker (approved):

PR #6995, opened 2022-03-03
fix(Toolbar): update test

kmcfaul:
**What**: Fixes jest error caused by merging #6955 - new test was not updated to use the new library.

### DIFF
--- a/packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
+++ b/packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
@@ -102,7 +102,7 @@ describe('Toolbar', () => {
     const customChipGroupContent = (
       <React.Fragment>
         <ToolbarItem>
-          <Button data-testid="save-button" variant="link" onClick={() => {}} isInline>
+          <Button variant="link" onClick={() => {}} isInline>
             Save filters
           </Button>
         </ToolbarItem>
@@ -124,9 +124,10 @@ describe('Toolbar', () => {
         <ToolbarContent>{items}</ToolbarContent>
       </Toolbar>
     );
-    expect(screen.findByTestId('save-button')).toBeTruthy();
-    expect(screen.findByTestId('clear-button')).toBeTruthy();
-    expect(view).toMatchSnapshot();
+    // Expecting 2 matches for text because the buttons also exist in hidden expandable ontent for mobile view
+    expect(screen.getAllByText('Save filters').length).toBe(2);
+    expect(screen.getAllByText('Clear all filters').length).toBe(2);
+    expect(view.container).toMatchSnapshot();
   });
 });
 
--- a/packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
+++ b/packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { shallow } from 'enzyme';
 import { ToolbarToggleGroup } from '../ToolbarToggleGroup';
 import { Toolbar } from '../Toolbar';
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
```diff
@@ -124,8 +124,8 @@ describe('Toolbar', () => {
         <ToolbarContent>{items}</ToolbarContent>
       </Toolbar>
     );
-    expect(view.find('#save-button').exists());
-    expect(view.find('#clear-button').exists());
+    expect(screen.findByTestId('save-button')).toBeTruthy();
+    expect(screen.findByTestId('clear-button')).toBeTruthy();
```
wise-king-sullyman, line 128: Unless there's an async need for using `findBy` queries  React Testing Library strongly leans towards using `getBy` queries for items that you expect to exist. It also prioritizes using "real" and accessible identifiers to select elements over using testIds, with `getByRole` using the `{ name: 'visible text' }` option as the ideal.

Link to the RTL docs where they talk about this: https://testing-library.com/docs/queries/about#types-of-queries

That being said when attempting to test using this query locally multiple of the button are being found, and I'm not sure why. If it's because of some async oddness I understand your use of `findBy`, but I see the duplication in the snapshot as well so I don't *think* that's the cause.

Additionally in my testing these tests seem to be false positives, as I can invalidate them by removing the testIds on the buttons and they still pass. I believe because `findBy` queries return promises.

packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
```diff
@@ -124,8 +124,8 @@ describe('Toolbar', () => {
...
     expect(view).toMatchSnapshot();
```
wise-king-sullyman, line 129: In RTL the return value from render is an object, and we need to access the value of `container` on that object for an accurate snapshot.

```suggestion
    expect(view.container).toMatchSnapshot();
```

tlabaj (approved):

packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
kmcfaul: Gotcha, I think I went with find only because that was the syntax of the previous test framework. Also updated the expect so the test fails if the elements aren't found.

packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
kmcfaul: Updated. I think that was also why the snapshot had doubled content because it was also reporting the value of the `container`.

packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
```diff
@@ -102,19 +102,19 @@ describe('Toolbar', () => {
...
       <React.Fragment>
         <ToolbarItem>
-          <Button id="save-button" variant="link" onClick={() => {}} isInline>
+          <Button variant="link" onClick={() => {}} isInline>
             Save filters
           </Button>
         </ToolbarItem>
         <ToolbarItem>
-          <Button id="clear-button" variant="link" onClick={() => {}} isInline>
+          <Button data-testid="clear-button" variant="link" onClick={() => {}} isInline>
```
wise-king-sullyman, line 110: Nit: I don't think this testId is needed any more.

```suggestion
          <Button variant="link" onClick={() => {}} isInline>
```

packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
```diff
@@ -124,9 +124,10 @@ describe('Toolbar', () => {
         <ToolbarContent>{items}</ToolbarContent>
       </Toolbar>
     );
-    expect(view.find('#save-button').exists());
-    expect(view.find('#clear-button').exists());
-    expect(view).toMatchSnapshot();
+    // Expecting 2 matches for text because the buttons also exist in hidden expandable ontent for mobile view
```
wise-king-sullyman, line 127: Nit: small typo

```suggestion
    // Expecting 2 matches for text because the buttons also exist in hidden expandable content for mobile view
```

packages/react-core/src/components/Toolbar/__tests__/ToolbarToggleGroup.test.tsx
```diff
@@ -124,9 +124,10 @@ describe('Toolbar', () => {
...
+    expect(screen.getAllByText('Save filters').length).toBe(2);
+    expect(screen.getAllByText('Clear all filters').length).toBe(2);
```
wise-king-sullyman, line 129: Nit: RTL prioritizes `getByRole` with the name property (when applicapble) over `getByText` in [their docs](https://testing-library.com/docs/queries/about/#priority).

```suggestion
    expect(screen.getAllByRole('button', { name: 'Save filters' }).length).toBe(2);
    expect(screen.getAllByRole('button', { name: 'Clear all filters' }).length).toBe(2);
```

wise-king-sullyman (commented):
Few nits.

wise-king-sullyman (approved):
🥳

PR #7043, opened 2022-03-10
fix(TimePicker): re-add ability to append to document body

### LINKED ISSUE #6942: TimePicker: menuAppendTo body not working
Today we got update in Cockpit and it seem now even with `menuAppendTo={() => document.body}` it does not overflow its parents.

Before: (these are modals)
TestSystemInfo-testTime-systime-manual-time-pixels

Now: (notice there is slider on the right as now the dropdown extends the modal)
TestSystemInfo-testTime-systime-manual-time-pixels

This smells like regression or do I need to change something?

_Originally posted by @marusak in https://github.com/patternfly/patternfly-react/issues/6825#issuecomment-1035404061_

thatblindgeye:
**What**: Closes #6942 

I also updated how the TimePicker gets appended to its parent by passing in a `parent` value to the `menuAppendTo` prop. This should work more similarly to the Dropdown component, and should both resolve the issue of scrollbars appearing in Modals and also allow access to the TimePicker menu via Voice Over. This can be tested by going to the the below page on the build workspace and adding the following snippets of code as instructed.

Link: [PatternFly Modal](https://patternfly-react-pr-7043.surge.sh/components/modal)

On line 2, import TimePicker:
```js
import { Modal, Button, TimePicker } from '@patternfly/react-core';
```

To test appending to a parent, before line 43 (this should come before the closing `</Modal>` tag) add:
```js
<div>
  <TimePicker menuAppendTo="parent" />
</div>
```

To test appending to document body, before line 43 add:
```js
<div>
  <TimePicker menuAppendTo={() => document.body} />
</div>
```

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/TimePicker/TimePicker.tsx
+++ b/packages/react-core/src/components/TimePicker/TimePicker.tsx
@@ -364,7 +364,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     });
   };
 
-  onInputFocus = (e: any) => {
+  onInputClick = (e: any) => {
     if (!this.state.isOpen) {
       this.onToggle(true);
     }
@@ -441,7 +441,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
     const menuContainer = (
       <Menu ref={this.menuRef} isScrollable>
         <MenuContent maxMenuHeight="200px">
-          <MenuList aria-labelledby={`${randomId}-input`}>
+          <MenuList aria-label={ariaLabel}>
             {options.map((option, index) => (
               <MenuItem onClick={this.onSelect} key={option} id={`${randomId}-option-${index}`}>
                 {option}
@@ -454,6 +454,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
 
     const textInput = (
       <TextInput
+        aria-haspopup="menu"
         className={css(formStyles.formControl)}
         id={`${randomId}-input`}
         aria-label={ariaLabel}
@@ -462,8 +463,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
         value={timeState || ''}
         type="text"
         iconVariant="clock"
-        onClick={this.onInputFocus}
-        onFocus={this.onInputFocus}
+        onClick={this.onInputClick}
         onChange={this.onInputChange}
         onBlur={this.onBlur}
         autoComplete="off"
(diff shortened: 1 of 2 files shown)


thatblindgeye:
@marusak Would you be able to test things in the preview build above? I included some instructions in my initial comment to test the changes to the TimePicker both for appending to `document.body` and for appending to a parent element. I would be most interested in whether appending to that parent element works as you might expect within a modal, as there's been known accessibility issues when appending dropdowns and similar menus to `document.body`, and I'm hoping this can be a step in the right direction for resolving two issues at once.

jessiehuff (commented):
So appending it to a parent seemed to work fairly well, and I only had two comments about that interaction:
- The menu's label isn't getting announced for some reason (it just says "menu")
- There seems to be no way to escape the menu easily without closing the entire modal. If you press the Esc key, it closes the modal (also true with keyboard)

With appending to document body, I had a lot more issues. The label was announced, interesting enough, but I couldn't figure out any way to get into the menu with VO. It makes me wonder if it's getting disabled like the rest of the document behind the modal?

thatblindgeye:
@jessiehuff based on prior convo, I added a warning to the examples regarding appending to `document.body`. I've also made the following changes, and would like to hear your thoughts on them:

- removed the functionality of the menu automatically opening when the input receives focus (similar to the Select typeahead variant, as well as the DatePicker not automatically opening a calendar on focus)
- in response to the above, added an `aria-haspopup` attribute to the input to notify users that a menu is attached to the input; keyboard users who do not user a screen reader would still be informed there is a menu once they start typing
- changed the menus `aria-labelledby` attribute to an `aria-label` attribute that uses the passed in `aria-label` prop; using the former attribute would cause the menu to be labelled as the input's current value, e.g. "3:35 AM" or "hh mm", whereas I think using the `aria-label` could provide better context.

There is still an issue of the menu being announced with a generic "menu" label by VO when the menu opens. Fixing that might be better for a separate issue, as well as possibly changing the behavior when the menu opens (i.e. should a menu item receive focus when the menu is opened via arrows keys).

jessiehuff:
So when testing it out, I couldn't seem to use it by keyboard or by VO. :( I think your points above make sense though.

thatblindgeye:
That's odd; I was able to get into the TimePicker via keyboard and VO. @nicolethoen would you be able to test out as well and see what behavior you get? If you and Jessie have the same behavior then it may require seeing the interaction in action.

jessiehuff (approved):
Hmm maybe my VO was acting up earlier. Appending to parent works, but as usual, not the document body. I think that's expected based on what we know right now though, right? And the other PR adds the verbiage to the TimePicker about that so I suppose this is good then. 😄

nicolethoen (approved):

tlabaj (approved):

PR #7052, opened 2022-03-10
feat(TextInput, TextArea, FormSelect): add isIconSprite prop

### LINKED ISSUE #7006: Add icon sprite variation for form control elements (TextInput, TextArea, FormSelect)
re: https://github.com/patternfly/patternfly/issues/4652

The icons in form controls are applied via a data URI as a background image. Data URIs can violate content security policies, so this provides a different way to add those icons.

This adds an external file for the icons at [assets/images/status-icon-sprite.svg](https://github.com/patternfly/patternfly/pull/4711/files#diff-612a895fc7fbb662cc96ddc7deb103a4c7b6ca00ffd8b113ccff65bf1e309528), and a modifier `.pf-m-icon-sprite` that tells the form controls to use the external file instead of a data URI.

You can verify it's working by looking at the computed background-image URL of a form control with an icon, and make sure it is set to the external file

Screen Shot 2022-03-04 at 10 53 54 AM

jenny-s51:
Closes #7006


packages/react-core/src/components/TextInput/examples/TextInput.md
```diff
@@ -139,37 +141,56 @@ TextInputSelectAll = () => {
...
+        type="text"
+        iconVariant="clock"
+        onChange={value => setClock(value)}
+        aria-label="text input example"
+      />
+      <br />
+      <br />
+      <TextInput
+        value={custom}
+        isIconSprite={isIconSprite}
```
gitdallas, line 179: This doesn't seem to have any affect on this icon with the customIconUrl.

jenny-s51: Thank you @gitdallas , nice catch! I've removed the `isIconSprite` prop from the variant that uses the `customIconUrl`.

gitdallas (approved):

packages/react-core/src/components/TextInput/examples/TextInput.md
```diff
@@ -139,37 +141,47 @@ TextInputSelectAll = () => {
 ```
 
 ### Icon variants
+
 ```js
 import React from 'react';
-import { TextInput } from '@patternfly/react-core';
+import { Checkbox, TextInput } from '@patternfly/react-core';
```
tlabaj, line 147: we are not using `Checkbox` so you can remove it here.

tlabaj (changes requested):
I think we want a new example like in core and to also add the explanation that they have for purpose of the sprite.  I would add a `TextArea` and Form Select example as well.

jenny-s51 (commented):
@tlabaj thank you for your review and for your helpful feedback! I've updated the PR based on your suggestions 🙂

packages/react-core/src/components/FormSelect/examples/FormSelect.md
```diff
@@ -258,3 +258,41 @@ class FormSelectInputGrouped extends React.Component {
...
+      { value: '', label: 'Select a number', disabled: false, isPlaceholder: true  },
+      { value: '1', label: 'One', disabled: false, isPlaceholder: false  },
+      { value: '2', label: 'Two', disabled: false, isPlaceholder: false  },
+      { value: '3', label: 'Three - the only valid option', disabled: false, isPlaceholder: false  }
+    ];
+  }
+
+  render() {
+    return (
+      <FormSelect isIconSprite value={this.state.value} onChange={this.onChange} aria-label="FormSelect Input">
```
mcoker, line 290: The contents in the select give the impression it's validated, but it doesn't seem to change state with different selections. It would be good to show the different validation states so we can see/verify the icon sprite in those states - like the other examples.

mcoker (commented):
LGTM, just a couple of questions. Also I noticed the invalid states are not using the sprite like they should, so put up a PR in core to fix that - https://github.com/patternfly/patternfly/pull/4732

mcoker (approved):
LGTM! The invalid text input and text area are still broken from core, but should be fixed once we have a release version from this PR - https://github.com/patternfly/patternfly/pull/4732.

I'm confident that will fix the issues in this PR. Should we wait to bump core in this PR to verify those issues or fixed?

mcoker:
FYI `@patternfly/patternfly: "4.183.2"` has the fix - https://github.com/patternfly/patternfly/releases/tag/prerelease-v4.183.2

jenny-s51:
Core has been bumped to 4.183.2 and I've rebased this PR to pull in those changes @mcoker 👍

mcoker:
@jenny-s51 lgtm!!

tlabaj (approved):
LGTM

PR #7053, opened 2022-03-11
feature(Topology): Add Concentric layout to Topology component

lucasponce:
This PR adds a new Concentric layout to the topology component.

Grid layouts work well to distribute nodes without taking into consideration edges,
image

but Concentric layouts have better results focused on high connectivity:
image
image

It's still experimental as there may be some corner cases not yet implemented, but it could be ready to have some feedback.

cc @jeff-phillips-18 @andrew-ronaldson

P.S: I have a pending optimized BreadthFirst algorithm that we're implementing in Kiali that will come in a future PR.


jeff-phillips-18 (approved):
LGTM. Thanks!

tlabaj (approved):

PR #7061, opened 2022-03-14
Feat(DescriptionList): 6866 set width of term column

### LINKED ISSUE #6866: DescriptionList: Add prop to set width of the term column in the horizontal variant

I have a design calling for a list of values resembling a [horizontal DescriptionList](https://www.patternfly.org/v4/components/description-list/#horizontal):

Screen Shot 2022-01-28 at 2 47 12 PM

I'm using the `isHorizontal` prop, but because one of the terms on the left side is a little long, it's looking cramped with wrapped text:

Screen Shot 2022-01-28 at 2 16 26 PM

I don't see a way to use the existing props of the React component to widen that side of the list. Per @mcoker it can be achieved with custom CSS by modifying the relevant CSS variables:

```
var(--pf-c-description-list--m-horizontal__term--width-on-2xl, var(--pf-c-description-list--m-horizontal__term--width-on-xl, var(--pf-c-description-list--m-horizontal__term--width-on-lg, var(--pf-c-description-list--m-horizontal__term--width-on-md, var(--pf-c-description-list--m-horizontal__term--width-on-sm, var(--pf-c-description-list--m-horizontal__term--width))))))
```

> If undefined, that stack resolves to the last fallback  --pf-c-description-list--m-horizontal__term--width: 12ch; - https://github.com/patternfly/patternfly/blob/d57498cdc0946095da6f605e95067df610f52787/src/patternfly/components/DescriptionList/description-list.scss#L41

``` --pf-c-description-list--m-horizontal__term--width: 12ch;```

This request is for a React prop we can use to set those variables somehow.

Enhancement

https://github.com/konveyor/crane-ui-plugin, no particular target release since we have a workaround.

### PR DESCRIPTION
**Pull request by gitdallas**
**What**: Closes #6866 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
+++ b/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
@@ -30,7 +30,7 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
   isFluid?: boolean;
   /** Sets the the default placement of description list groups to fill from top to bottom. */
   isFillColumns?: boolean;
-  /** Sets the number of columns on the description list */
+  /** Sets the number of columns on the description list at various breakpoints */
   columnModifier?: {
     default?: '1Col' | '2Col' | '3Col';
     sm?: '1Col' | '2Col' | '3Col';
@@ -39,7 +39,7 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
     xl?: '1Col' | '2Col' | '3Col';
     '2xl'?: '1Col' | '2Col' | '3Col';
   };
-  /** Indicates how the menu will align at screen size breakpoints. Default alignment is set via the position property. */
+  /** Indicates how the menu will align at various breakpoints. */
   orientation?: {
     sm?: 'vertical' | 'horizontal';
     md?: 'vertical' | 'horizontal';
@@ -47,7 +47,7 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
     xl?: 'vertical' | 'horizontal';
     '2xl'?: 'vertical' | 'horizontal';
   };
-  /** Sets the breakpoint modifiers for the autofit feature */
+  /** Sets the minimum column size for the auto-fit (isAutoFit) layout at various breakpoints. */
   autoFitMinModifier?: {
     default?: string;
     sm?: string;
@@ -56,7 +56,7 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
     xl?: string;
     '2xl'?: string;
   };
-  /** Sets the breakpoint modifiers for the horizontal width of the term column */
+  /** Sets the horizontal description list's term column width at various breakpoints. */
   horizontalTermWidthModifier?: {
     default?: string;
     sm?: string;
--- a/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
+++ b/packages/react-core/src/components/DescriptionList/DescriptionList.tsx
@@ -58,10 +58,28 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
     xl?: string;
     '2xl'?: string;
   };
+  termAutoFitMinModifier?: {
+    default?: string;
+    sm?: string;
+    md?: string;
+    lg?: string;
+    xl?: string;
+    '2xl'?: string;
+  };
 }
 
 const setAutoFitMinModifiers = (autoFitMinModifier: AutoFitModifiers) => {
   const prefix = '--pf-c-description-list--GridTemplateColumns--min';
+  const mods = autoFitMinModifier as Partial<{ [k: string]: string }>;
+  return Object.keys(mods || {}).reduce(
+    (acc, curr) =>
+      curr === 'default' ? { ...acc, [prefix]: mods[curr] } : { ...acc, [`${prefix}-on-${curr}`]: mods[curr] },
+    {}
+  );
+};
+
+const setTermAutoFitMinModifiers = (autoFitMinModifier: AutoFitModifiers) => {
+  const prefix = termWidthToken.name;
   const mods = autoFitMinModifier as Partial<{ [k: string]: string }>;
   return Object.keys(mods || {}).reduce(
     (acc, curr) =>
@@ -83,10 +101,18 @@ export const DescriptionList: React.FunctionComponent<DescriptionListProps> = ({
   isFillColumns,
   columnModifier,
   autoFitMinModifier,
+  termAutoFitMinModifier,
   orientation,
   style,
   ...props
 }: DescriptionListProps) => {
+  if (isAutoFit && (autoFitMinModifier || style)) {
+    style = { ...style, ...setAutoFitMinModifiers(autoFitMinModifier) };
+  }
+  if (termAutoFitMinModifier) {
+    style = { ...style, ...setTermAutoFitMinModifiers(termAutoFitMinModifier) };
+  }
+
   const listDivRef = React.useRef(null);
   React.useEffect(() => {
     if (termWidth && listDivRef && isHorizontal) {
@@ -109,11 +135,7 @@ export const DescriptionList: React.FunctionComponent<DescriptionListProps> = ({
         isFillColumns && styles.modifiers.fillColumns,
         className
       )}
-      style={
-        autoFitMinModifier || style
-          ? { ...(isAutoFit ? setAutoFitMinModifiers(autoFitMinModifier) : {}), ...style }
-          : undefined
-      }
+      style={style}
       {...props}
     >
       {children}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/react-core/src/components/DescriptionList/DescriptionList.tsx [hunk 1]
```diff
@@ -24,7 +25,10 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
   isAutoColumnWidths?: boolean;
   /** Modifies the description list display to inline-grid. */
   isInlineGrid?: boolean;
+  /** sets the width of the term column (horizontal mode) */
+  termWidth?: string;
   /** Sets the description list to compact styling. */
+
```
nicolethoen, line 31: We dont need the extra space here

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -269,6 +269,50 @@ import PlusCircleIcon from '@patternfly/react-icons/dist/esm/icons/plus-circle-i
...
+  Button,
+  DescriptionList,
+  DescriptionListTerm,
+  DescriptionListText,
+  DescriptionListGroup,
+  DescriptionListDescription
+} from '@patternfly/react-core';
+import PlusCircleIcon from '@patternfly/react-icons/dist/esm/icons/plus-circle-icon';
+
+<DescriptionList isHorizontal isInlineGrid termWidth="20ch">
```
nicolethoen, line 286: Is the `isInlineGrid` prop necessary for this example? If so, perhaps you should note that in the example and explain why.

nicolethoen (commented):
Perhaps it'd be worth updating the other horizontal example to use a checkbox to apply a custom width when checked since it's only changing one prop.

gitdallas:
@nicolethoen - just noticed your top level comment about a checkbox, will address that soon

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -24,6 +25,8 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
   isAutoColumnWidths?: boolean;
   /** Modifies the description list display to inline-grid. */
   isInlineGrid?: boolean;
+  /** sets the width of the term column (horizontal mode) */
```
tlabaj, line 28: nit. Sentence case here.

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -82,30 +86,39 @@ export const DescriptionList: React.FunctionComponent<DescriptionListProps> = ({
...
-      className
-    )}
-    style={
-      autoFitMinModifier || style
-        ? { ...(isAutoFit ? setAutoFitMinModifiers(autoFitMinModifier) : {}), ...style }
-        : undefined
+}: DescriptionListProps) => {
+  const listDivRef = React.useRef(null);
+  React.useEffect(() => {
+    if (termWidth && listDivRef) {
```
tlabaj, line 92: Is the intention to only define the `listDivRef` if `termWidth` is defined?  it looks like the ref is always defined?
If you can only set the width when horizontal can we check for that here?

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -269,6 +269,50 @@ import PlusCircleIcon from '@patternfly/react-icons/dist/esm/icons/plus-circle-i
 </DescriptionList>;
 ```
 
+### Horizontal with custom termWidth
```
tlabaj, line 272: nit.  `term width` instead of `termWidth`

tlabaj (changes requested):

tlabaj:
> Perhaps it'd be worth updating the other horizontal example to use a checkbox to apply a custom width when checked since it's only changing one prop.

@nicolethoen Maybe more than a checkbox is warranted, it would be ideal if the consumer can set the width.  Most of our examples with checkboxes are to apply modifiers.

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
gitdallas: the reference to the list div element will always be defined when the component mounts.  i will add a check for horizontal.

tlabaj:
@gitdallas I don't think we have a good way to add anything other than a checkbox right now without having the control look like it is part of the component.  I  this case I think it is best to have a separate example to show the functionality.  
An issue has been opened in design so we can update our example to better allow for adding controls.

gitdallas:
@tlabaj  - reverted back to separate examples

tlabaj (approved):

mcoker:
@gitdallas ah ok, I'm assuming they don't show up in react tokens since we don't define them with an initial value. We just use them in a fallback stack `--pf-c-description-list__term--width` at different breakpoints. You can see how they're used here: https://unpkg.com/browse/@patternfly/patternfly@4.183.1/components/DescriptionList/description-list.css

I imagine `termWidth` would be an object that supports `default`, `sm`, `md`, etc keys to define the size(s), similar to `maxWidths` and `minWidths` in the [gallery component](https://www.patternfly.org/v4/layouts/gallery#props)

gitdallas:
@nicolethoen @mcoker - see latest https://patternfly-react-pr-7061.surge.sh/components/description-list , i added modifiers for the width breakpoints.. it doesn't seem to have any affect. you can see they are applied
image

nicolethoen:
I think you're updating the wrong breakpoint modifier.
Your various breakpoints should be setting the following CSS vars:
```
  --pf-c-description-list--m-horizontal__term--width (this is the default width)
  --pf-c-description-list--m-horizontal__term--width-on-sm
  --pf-c-description-list--m-horizontal__term--width-on-md
  --pf-c-description-list--m-horizontal__term--width-on-lg
  --pf-c-description-list--m-horizontal__term--width-on-xl
  --pf-c-description-list--m-horizontal__term--width-on-2xl
```

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -55,6 +58,14 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
     xl?: string;
     '2xl'?: string;
   };
+  termAutoFitMinModifier?: {
```
nicolethoen, line 61: All these props need a prop description to clarify their purpose. The new prop is setting the width of the term column in horizontal mode at various breakpoints. The previous prop - `autoFitMinModifier` - also needs a prop description.

Also, the default value should likely replace the need for the new `termWidth` prop, and i think the name of your new `termAutoFitMinModifier` prop is not quite correct. I propose possibly `horizontalTermWidths`

nicolethoen (changes requested):

nicolethoen (approved):

mcoker (changes requested):
Looks great!!! Just one nit about the docs.

packages/react-core/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -55,11 +56,19 @@ export interface DescriptionListProps extends Omit<React.HTMLProps<HTMLDListElem
     xl?: string;
     '2xl'?: string;
   };
+  /** Sets the breakpoint modifiers for the horizontal width of the term column */
```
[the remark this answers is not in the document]
mcoker: > `autoFitMinModifier` - Sets the minimum column size for the `auto-fit` layout.

* `auto-fit` is the grid column size function, so I hyphenated it. You might replace that with `isAutoFit` so it's "column size for the `isAutoFit` layout."? 

The orientation docs seem to be wrong - maybe they were copied from somewhere else? Orientation just change the layout between horizontal and vertical.

And with all of those, I'd just say either include wording at the end like "modifies [such and such] _at various breakpoints_" or not, as long as it's consistent or what's ideal going forward mentioning breakpoint support for that kind of prop doc in react components.

mcoker (changes requested):
Left a comment re: the docs.

Also unrelated I noticed the `enableFlip` prop docs are missing a period between "necessary If" and at the end.

> If true, tries to keep the popover in view by flipping it if necessary If the position is set to 'auto', this prop is ignored

tlabaj (commented):
looks good other than the little issue.

mcoker (approved):
Nice!!

nicolethoen (approved):

packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -235,38 +235,97 @@ import {
   DescriptionListTerm,
   DescriptionListText,
   DescriptionListGroup,
-  DescriptionListDescription
+  DescriptionListDescription,
```
tlabaj, line 238: this comma is not needed.  I would have thought it would fail linting.

mturley:
Thanks so much everyone! This is perfect

PR #7063, opened 2022-03-14
chore: Migrate tests to RTL (6)

### LINKED ISSUE #7046: Migrate tests to RTL (6)
### Component test suites to migrate:
- AccessConsole
- DesktopViewer
- SerialConsole
- XTerm
- VncConsoleActions
- MultipleFileUpload
- MultipleFileUploadButton
- MultipleFileUploadInfo
- MultipleFileUploadMain
- MultipleFileUploadStatus
- MultipleFileUploadStatusItem
- MultipleFileUploadTitle
- MultipleFileUploadTitleIcon
- MultipleFileUploadTitleText
- MultipleFileUploadTitleTextSeparator
- Table
- ComputeElementDimensions
- SVGDefs
- SVGDefsProvider
- TopologyControlBar
- TopologySideBar
- TopologyView

jeffpuzzo:
**What**: Closes #7046


packages/react-core/src/components/MultipleFileUpload/MultipleFileUploadStatusItem.tsx
```diff
@@ -45,6 +45,8 @@ export interface MultipleFileUploadStatusItemProps extends React.HTMLProps<HTMLL
   progressAriaLabel?: string;
   /** Associates the ProgressBar with it's label for accessibility purposes. Required when title not used */
   progressAriaLabelledBy?: string;
+  /** Unique identifier for Progress. Generated if not specified. */
```
tlabaj, line 48: nit.  lowercase for progress here.  
It also looks like the id is not generated.

tlabaj (changes requested):

packages/react-core/src/components/MultipleFileUpload/MultipleFileUploadStatusItem.tsx
jeffpuzzo: It doesn't get generated in this file, but it does inside of `Progress`, where we pass `progressId` to the `id` prop and this the logic in that component: https://github.com/patternfly/patternfly-react/blob/ea14d7d9c04f63e62c8811090c5107edc18aa4d8/packages/react-core/src/components/Progress/Progress.tsx#L67.

I could rephrase and say something like `Generated inside of Progress if not specified.`?

In terms of the case of `progress`, I assumed it to be capitalized as it's referring to the component, just as the above comment on line 46 refers to `ProgressBar` in similar TitleCase.

tlabaj: The above should probably be in sentence case too, "progress bar".
I think the comment is fine as is for the progressId.

packages/react-table/src/components/Table/utils/transformers.test.tsx
```diff
@@ -98,10 +38,6 @@ describe('Transformer functions', () => {
       };
       const returnedData = selectable('', { column, rowIndex: 0, rowData: { selected: true } } as IExtra);
       expect(returnedData).toMatchObject({ className: 'pf-c-table__check' });
```
nicolethoen, line 40: here you did delete the aspect of the test that verified that the selecting event handlers are getting triggered on click. Should we either make a new test to verify that?

packages/react-table/src/components/Table/utils/transformers.test.tsx
```diff
@@ -144,54 +78,13 @@ describe('Transformer functions', () => {
...
-        expectDisabled: true
-      },
-      {
-        actionResolver: () => actions,
-        areActionsDisabled: () => true,
-        expectDisabled: true
-      }
-    ];
-
-    actionConfigs.forEach(testCellActions);
```
nicolethoen, line 192: The actions are dynamically built in the table, so testing that they are built and firing the right events at the right times is probably something we should keep.

packages/react-topology/src/components/defs/__tests__/SVGDefs.spec.tsx
```diff
@@ -16,40 +15,12 @@ describe('SVGDefs', () => {
...
     );
-    const innerWrapper = wrapper.find(SVGDefsSetter).first();
-    expect(innerWrapper.props()).toEqual({
-      ...contextProps,
-      ...props
-    });
-  });
-});
-
-describe('SVGDefsSetter', () => {
```
nicolethoen, line 32: Should we open an issue documenting that this test was removed?

packages/react-topology/src/components/defs/__tests__/__snapshots__/SVGDefs.spec.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`SVGDefs should get #addDef and #removeDef from context 1`] = `<DocumentFragment />`;
```
nicolethoen, line 3: this is an empty snapshot? is it service a purpose?

packages/react-console/src/components/SerialConsole/__tests__/SerialConsole.test.tsx
```diff
@@ -1,51 +1,74 @@
...
-);
-
-test('SerialConsole in the CONNECTED state', () => {
-  const view = shallow(connectedState);
-  expect(view).toMatchSnapshot();
-});
+  test('in the DISCONNECTED state', () => {
+    const { asFragment } = render(
+      <SerialConsole
+        onData={jest.fn()}
```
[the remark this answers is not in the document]
jeffpuzzo: Checked the props and XTermProps which are extended for SerialConsole and neither allow for an `id`;
image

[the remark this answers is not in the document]
nicolethoen: weird... ok

packages/react-table/src/components/Table/utils/transformers.test.tsx
jeffpuzzo: None of these iterated tests in `transformers` made much sense to me, so I can take another look, and I recall trying to get these to work, but might be something I make a separate story in our backlog for to address later.

packages/react-topology/src/components/defs/__tests__/SVGDefs.spec.tsx
jeffpuzzo: I'll take another look before doing so, but probably will end up doing that.

packages/react-topology/src/components/defs/__tests__/__snapshots__/SVGDefs.spec.tsx.snap
jeffpuzzo: Looked closer at the component and it always returns null. I'm going to end up deleting this test as well and adding it to the list to follow-up on. 

This wouldn't be a candidate for an RTL test since it's basically just verifying context works (provider, consumer, verify the consumer gets what the provider passes). Where yes, it's possible to setup context incorrectly, but type checking combined with already tested react functionality makes me think this is overkill and not necessary.

jeffpuzzo: Managed to come up with something for this, to bring back the Setter test, and the one from transformers. Please see updates in latest commit.

jeffpuzzo: Long story short for the SVGDefs test; Verifying functions are called on mount and on unmount implicitly prove the context props are passed down appropriately.

packages/react-table/src/components/Table/utils/transformers.test.tsx
nicolethoen: that'd be fine if you end up doing that i think

tlabaj (approved):

kmcfaul (approved):

nicolethoen (approved):

PR #7065, opened 2022-03-14
Chore: Compare html vs render result

### LINKED ISSUE #7064: Compare HTML instead of render result for RTL snapshots
As a part of the initial migration from enzyme to RTL for unit tests, some of the snapshot comparisons were made to simply check the against the render result object returned from RTL's `render` function. Instead, we can compare the html DOM itself. This is a simple change of `view.container` => `view.container.outerHTML` for those snapshot expectations.

jeffpuzzo:
**What**: Closes #7064


jeffpuzzo:
This is no longer the plan of attack for snapshots, so closing.

PR #7066, opened 2022-03-14
feat(TextInputGroup): combine values in attribute value filtering

### LINKED ISSUE #6795: Attribute-value filtering enhancement to support comma-separated approach for multiple values in 1 key
## Requesting new features, enhancements, or design changes to PatternFly 
- For the attribute-value filtering demo, currently each entry of a key-value pair turns into one chip, regardless of whether or not the pair has the same key as the others. 
- In ACM/ACS, the pattern that existed groups the values with the same keys in one chip. E.g., [cluster: name1, name2] instead of [cluster: name1] [cluster: name2]. It saves space and also is easier to locate and see the relationship when looking at the filters that one has applied

Current: 
image

- This is an enhancement on this demo: https://www.patternfly.org/v4/components/text-input-group/react-demos/

wise-king-sullyman:
**What**: Closes #6795

**Additional issues**:

Direct link to the [changed demo](https://patternfly-react-pr-7066.surge.sh/components/text-input-group/react-demos#attribute-value-filtering) for convenience.

### DIFF
--- a/packages/react-core/src/demos/TextInputGroupDemo.md
+++ b/packages/react-core/src/demos/TextInputGroupDemo.md
@@ -24,6 +24,8 @@ Additionally, attributes can be selected by typing the full (case sensitive) nam
 
 Attributes can be deselected (returning you to attribute selection mode) by hitting `escape`, or by hitting `backspace` when the only text in the text input is the attribute.
 
+Values can be individually removed from associated attributes after selection by selecting them a second time.
+
 ```ts file="./examples/TextInputGroup/AttributeValueFiltering.tsx"
 ```
 ### Auto-complete search with typeahead
--- a/packages/react-core/src/demos/examples/TextInputGroup/AttributeValueFiltering.tsx
+++ b/packages/react-core/src/demos/examples/TextInputGroup/AttributeValueFiltering.tsx
@@ -17,10 +17,15 @@ import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import TimesIcon from '@patternfly/react-icons/dist/esm/icons/times-icon';
 
 export const AttributeValueFiltering: React.FunctionComponent = () => {
+  interface chipData {
+    attribute: string;
+    values: string[];
+  }
+
   const [inputValue, setInputValue] = React.useState('');
   const [selectedKey, setSelectedKey] = React.useState('');
   const [menuIsOpen, setMenuIsOpen] = React.useState(false);
-  const [currentChips, setCurrentChips] = React.useState<string[]>([]);
+  const [currentChips, setCurrentChips] = React.useState<chipData[]>([]);
 
   interface attributeValueData {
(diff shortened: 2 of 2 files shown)


nicolethoen (commented):
I think this is working as described, but I have a slight issue with reselecting and item to remove it from a chained attribute/value(s) pair. 

I think it's tough because there is no 'selected state' to know which items I want to deselect.
plus, the values are hidden in truncation/tooltip so I cannot have those visible on hover and have the menu open at the same time.

I'd be interested in hearing what @mmenestr and/or @mcarrano thinks.
I know the design didn't call for it, but I think this set up begs for ChipGroups rather than truncated Chips.

mcarrano:
@nicolethoen interesting idea about using a chip group. Will a chip ground fit inside the field? They are a bit taller to enclose the chips.

nicolethoen:
> @nicolethoen interesting idea about using a chip group. Will a chip ground fit inside the field? They are a bit taller to enclose the chips.

I wouldn't know if they'd fit in the field without testing it. It'd likely require a core update if we want that... but my concern in terms of interaction is that the truncated chip is hiding the numerous values associated with the attribute. they are only revealed in the tooltip on hover, but I cannot select a new value from the menu AND see the tooltip with all the applied values at the same time, so it feels less than ideal.

mcarrano:
@imjoyjean please see @nicolethoen 's comments above. I think she raises vailid points about this design and I'm wondering if you have solutions for those cases in ACM that we are not thinking about here. You can see what the demo looks like here: https://patternfly-react-pr-7066.surge.sh/components/text-input-group/react-demos#attribute-value-filtering. You'll notice that the chips truncate after one or two values and selected per attribute and that there is no way to remove a value once it is selected.

kmcfaul:
@nicolethoen It looks like when you reselect the same key-value pair that already exists in the chips, it will remove the chip from the list. Is this issue fixed or are you referring to something else regarding deselection, as I don't see the PR updated from your feedback.

nicolethoen:
> are you referring to something else regarding deselection

I was referring to the following workflow, which has not yet been updated:

1. I select attr=`Label` and value=`environment`
2. I select `Label` and there is nothing about the `environment` option in the menu that indicates that I already have it selected.

I think this is especially problematic when we have chips which truncate when they overflow, because then the chip is hiding the selected values, so there is really now way to know whether I am selecting or deselecting an option when I click on it.

mcarrano:
@wise-king-sullyman @nicolethoen @tlabaj I discussed this yesterday with @mmenestr and we agreed that we need to reconsider this from a design standpoint. A [related design issue](https://github.com/patternfly/patternfly-design/issues/1143) was filed yesterday that requests a different approach and I think we need to consider how this might work in a variety of use cases. Can we put this PR on hold for the time being?

tlabaj:
Closing this PR.  please see the original issue for further discussion.

Issue #6925, opened 2022-02-10
Table Composable: Actions: rowId returns undefined


aferd:
**Describe the issue. What is the expected and unexpected behavior?**
row returns undefined. Clicking on the 'Some Action' returns 'clicked on Some action, on row: **undefined**

[sandbox](https://codesandbox.io/s/5d4ym?file=/index.js)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
bug

josejulio:
The same issue is seen in the demo: https://patternfly-react.surge.sh/components/table#composable-actions
Instead of receiving the rowId/index we get undefined.

@tlabaj could you take a look? thanks!

PR #7068, opened 2022-03-15
fix/docs(Table): add zindex docs, fix old rowIndex callback

### LINKED ISSUE #7007: Docs for handling z-index of things in sticky header/column tables
Sticky table headers/columns can conflict with the `z-index` of other things depending on use case, so we added some docs around that here. I think it would be beneficial to include this on the react side, too? 

https://github.com/patternfly/patternfly/pull/4709/files

kmcfaul:
**What**: Closes #7007, #6925

Adds description regarding potential z-index clashes for sticky table headers/columns, and broke out a few table examples into sections to put that blurb on top of all the sticky examples rather than just in the first example. LMK if it would be better inside the first sticky example and remove the sections.

Also addresses the rowIndex in the composable actions table callback. We've updated our example recently and the new example doesn't have this issue, but the old example method should now also work after adding `rowIndex` to the `Td` - `actions` object.

### DIFF
--- a/packages/react-table/src/components/TableComposable/examples/ComposableTable.md
+++ b/packages/react-table/src/components/TableComposable/examples/ComposableTable.md
@@ -311,6 +311,16 @@ To make a row draggable:
 ```ts isBeta file="ComposableTableDraggable.tsx"
 ```
 
+## Sticky table modifiers
+
+**Note:** Sticky table headers and columns have a higher `z-index` than the `z-index` used for menus (dropdown, select, etc). The intent is that the contents of a scrollable table will scroll under the sticky header/column, including any expanded menus. However, there may be use cases where a menu needs to appear on top of a sticky header/column, such as an expanded menu in a toolbar above a table with a sticky header.
+
+There are a few ways this can be handled:
+
+- Manipulate the `z-index` of the menu and/or table headers/columns manually.
+- Use the `menuAppendTo` prop in non-composable react components with menus to append the menu to an element outside of the table (e.g., the `<body>` element) so that the menu has a higher stacking context than - and can appear on top of - sticky headers/columns as well as appear outside of any scrollable content in the table.
+- In the case where the menu is outside of the table (e.g., above the table in a toolbar, or below the table and the menu expands up), assign the entire table a lower `z-index` than the `z-index` of the menu. This creates a lower stacking context for the entire table compared to the menu, while preserving the stacking context of the elements inside of the table.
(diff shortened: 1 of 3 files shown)


packages/react-table/src/components/TableComposable/examples/ComposableTable.md
```diff
@@ -311,6 +311,16 @@ To make a row draggable:
...
 ```
 
+## Sticky table modifiers
+
+**Note:** Sticky table headers and columns have a higher `z-index` than the `z-index` used for menus (dropdown, select, etc). The intent is that the contents of a scrollable table will scroll under the sticky header/column, including any expanded menus. However, there may be use cases where a menu needs to appear on top of a sticky header/column, such as an expanded menu in a toolbar above a table with a sticky header.
+
+There are a few ways this can be handled:
+
+- Manipulate the `z-index` of the menu and/or table headers/columns manually.
+- Use the `menuAppendTo` prop in non-composable react components with menus to append the menu to an element outside of the table (e.g., the `<body>` element) so that the menu has a higher stacking context than - and can appear on top of - sticky headers/columns as well as appear outside of any scrollable content in the table.
```
nicolethoen, line 321: This may no longer be the recommendation after all the work @thatblindgeye has been doing on appending to `document.body`. We should double check what is the best ancestor to append a menu to to solve this problem.

thatblindgeye: So I *believe* utilizing the `"parent"` value for the `menuAppendTo` prop would work as intended here?

This is only a thrown together attempt, but I added dropdowns to one of the Sticky examples in this PR's build below. Depending if that shows expected behavior, then it could make sense to either referring to "(e.g., the table's `parent` element)", or just omitting an example in this paragraph (other components should most likely have mention of the whole body vs parent thing).

https://user-images.githubusercontent.com/70952936/159538532-777686a9-844b-465a-b6c4-1bbccb6fa3e7.mov

kmcfaul: I'll update to reference the table's parent rather than the body for now at least.

thatblindgeye (approved):

nicolethoen (approved):

tlabaj (approved):
LGTM

mattnolting (approved):
Wonderful! Excellent work @kmcfaul 👏 👏 👏

PR #7070, opened 2022-03-15
fix(Menu): remove rogue scrollbar from drilldown menus

### LINKED ISSUE #6842: Drilldown menu: unnecessary scroll bar in sub menus

nicolethoen:
**What**: Closes #6842

unfortunately, the top/bottom padding for the MenuContent is = 0.5rem, so it makes the calculation complicated and dependent on font size. And when I try to load the value of the variable into react for possible calcluations, even that 0.5rem value is overridden by a subsequent rule so I cannot import the variable value directly into the component using react-tokens.
(i.e. `import c_menu__list_PaddingTop from '@patternfly/react-tokens/dist/esm/c_menu__list_PaddingTop';` returns the wrong value).

so for now, I use the default value of 16px, which is what the padding top and bottom defaults to with no custom overrides. Not sure if there is an alternate option right now...

### DIFF
--- a/packages/react-core/src/components/Menu/MenuContent.tsx
+++ b/packages/react-core/src/components/Menu/MenuContent.tsx
@@ -21,8 +21,31 @@ export const MenuContent = React.forwardRef((props: MenuContentProps, ref: React
   const menuContentRef = React.createRef<HTMLDivElement>();
   const refCallback = (el: HTMLElement, menuId: string, onGetMenuHeight: (menuId: string, height: number) => void) => {
     if (el) {
-      // if this menu has a parent MenuList, then we need to account for that parent list's padding.
-      const clientHeight = el.closest(`.${styles.menuList}`) ? el.clientHeight + 16 : el.clientHeight;
+      let clientHeight = el.clientHeight;
+
+      // if this menu is a submenu, we need to account for the root menu list's padding and root menu content's border.
+      let rootMenuList = null;
+      let parentEl = el.closest(`.${styles.menuList}`);
+      while (parentEl !== null && parentEl.nodeType === 1) {
+        if (parentEl.classList.contains(styles.menuList)) {
+          rootMenuList = parentEl;
+        }
+        parentEl = parentEl.parentElement;
+      }
+
+      if (rootMenuList) {
+        const rootMenuListStyles = getComputedStyle(rootMenuList);
+        const rootMenuListPaddingOffset =
+          parseFloat(rootMenuListStyles.getPropertyValue('padding-top').replace(/px/g, '')) +
+          parseFloat(rootMenuListStyles.getPropertyValue('padding-bottom').replace(/px/g, '')) +
+          parseFloat(
+            getComputedStyle(rootMenuList.parentElement)
+              .getPropertyValue('border-bottom-width')
+              .replace(/px/g, '')
+          );
+        clientHeight = clientHeight + rootMenuListPaddingOffset;
+      }
+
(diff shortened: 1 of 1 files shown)


jeffpuzzo:
I haven't tested this out, but here's something else to try since I agree in that hardcoding the pixels for padding isn't ideal:

```
const parentMenuList = el.closest(`.${styles.menuList}`);
const parentMenuListStyles = window.getComputedStyle(parentMenuList);

const paddingOffset = parseFloat(
  parentMenuListStyles.getPropertyValue('padding-top') + 
  parentMenuListStyles.getPropertyValue('padding-bottom')
);

const clientHeight = !!parentMenuList ? el.clientHeight + paddingOffset : el.clientHeight;
```

jeffpuzzo (approved):
If the suggestion mentioned doesn't work out, I did verify the existing change solves issue and am ok with it if no other alternatives to a hardcoded padding value exist.

kmcfaul (approved):

packages/react-core/src/components/Menu/MenuContent.tsx
```diff
@@ -21,8 +21,33 @@ export const MenuContent = React.forwardRef((props: MenuContentProps, ref: React
...
+
+      if (rootMenuList) {
+        const rootMenuListStyles = getComputedStyle(rootMenuList);
+        const rootMenuListPaddingOffset =
+          parseFloat(rootMenuListStyles.getPropertyValue('padding-top').replace(/px/g, '')) +
+          parseFloat(rootMenuListStyles.getPropertyValue('padding-bottom').replace(/px/g, '')) +
+          parseFloat(
+            getComputedStyle(rootMenuList.parentElement)
+              .getPropertyValue('border-bottom-width')
+              .replace(/px/g, '')
```
thatblindgeye, line 44: For these lines, is this just to take into account that 1px bottom-border that is on the example page rather than the menu itself?

Mainly asking as I was just wondering whether a top border should be taken into account as well.

thatblindgeye (approved):
Looks good! Just had a clarifying comment below.

packages/react-core/src/components/Menu/MenuContent.tsx
nicolethoen: In fact, the border is applied to the bottom of the root menu's `MenuContent` when it's in a nav.
```
.pf-c-nav .pf-c-menu:first-child > .pf-c-menu__content {
    border-bottom: var(--pf-c-nav--c-menu__item--before--BorderBottomWidth) solid var(--pf-c-nav--c-menu__item--before--BorderBottomColor);
}
```

mattnolting (approved):
👍  This looks right to me. Side note, the drilldown has a scrollbar when actively drilling in to a menu. We need to address that: https://github.com/patternfly/patternfly/issues/4748

Issue #6654, opened 2021-11-30
CalendarMonth RFEs


knokit:
Currently, _CalendarMonth_ always requires a `date` prop and it seems this prop should be optional, as it defaults to `today` if not provided.

Two other things that would be useful:
- allow selecting multiple individual dates (not a range)
- expose the month change events (for example `onMonthChange`)

The following issues are stories for this epic:
- [x] #7026
- [x] #7027

mcarrano:
We will get this on our roadmap. We are currently scheduling into Q2 2022.

tlabaj:
I opened a new issue for the first part of about the required prop #7026 since that is independent of the other requests.

tlabaj:
@knokit is the request "allow selecting multiple individual dates (not a range)" for the CalendarMonth component or the Date picker component?

tlabaj:
@mcarrano I also opened #7027 to add the missing examples in react for the date range selection.

knokit:
> @knokit is the request "allow selecting multiple individual dates (not a range)" for the CalendarMonth component or the Date picker component?

@tlabaj, it's for the CalendarMonth component.
 
> For the second bullet point, we have prop for the `onChange` call back. Is there something else you were looking for?

`onChange` is called when a date is selected and I'm looking for something different.

My idea is that `onMonthChange` (and `onYearChange`?) would be called whenever the visible month changes - for example, this would be useful for loading data for the visible month from an API and enable/disable dates based on it.

thatblindgeye:
@knokit The optional date prop issue has been resolved in PR #7077 . A quick note:

We have removed the default of "today" when a date prop is not passed in, which can be explained in the following comment chain within that PR: https://github.com/patternfly/patternfly-react/pull/7077#discussion_r831466050

Despite there not being a default _selected_ date when the date prop is not passed in, the Calendar Month component should still default to visually showing the current "today" month as well as applying a style to the "today" day.

mcarrano:
@knokit I wanted to follow up again about the request to support picking multiple dates from the calendar (as opposed to a range). DO you have a specific use case for this? Can you describe what that is? Thanks.

knokit:
I have this component:
<img alt="events calendar" src="https://user-images.githubusercontent.com/3659829/172127088-f56c0b79-1da5-4b82-85c6-49761b3b7cb8.png" width="400" />

[...]

PR #7077, opened 2022-03-16
feat(CalendarMonth): make date prop optional

### LINKED ISSUE #7026: Calendar Month: date prop should be optional
Currently, CalendarMonth always requires a date prop and it seems this prop should be optional, as it defaults to today if not provided.

enhancement

thatblindgeye:
**What**: Closes #7026 

Added internal state to the component so that an optional date prop works as intended whether it is passed in or not.

Also updated internal variables for the `buildCalendar` function, just to differentiate it from the `selectedDate` state.

**Additional issues**:
Part of the #6654 epic

### DIFF
--- a/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
@@ -90,7 +90,7 @@ export const isValidDate = (date: Date) => Boolean(date && !isNaN(date as any));
 const today = new Date();
 
 export const CalendarMonth = ({
-  date: dateProp = today,
+  date: dateProp,
   locale = undefined,
   monthFormat = date => date.toLocaleDateString(locale, { month: 'long' }),
   weekdayFormat = date => date.toLocaleDateString(locale, { weekday: 'narrow' }),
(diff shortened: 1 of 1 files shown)


nicolethoen (approved):

kmcfaul (approved):

packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
```diff
@@ -90,7 +90,7 @@ export const isValidDate = (date: Date) => Boolean(date && !isNaN(date as any));
 const today = new Date();
 
 export const CalendarMonth = ({
-  date: dateProp = today,
+  date: dateProp,
```
tlabaj, line 93: Shouldn't this be today or defaultDate?

tlabaj (changes requested):

packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
thatblindgeye: Nicole and I had a quick convo regarding that. Essentially this more follows how it works within the Date Picker component, where if no actual date is passed in then no date is "selected".

tlabaj: It looks like the component was already defaulting the date to "today" even though it is required (I am not sure why, possibly an oversight).  The ask in the original issue was to default to "today" if no date is provided.

@mcarrano do you have a n opinion on this?

nicolethoen: If we think of the DatePicker like a form element, it seems like always needing a date selected is not ideal. I'm imagining, for example, a toolbar filter which includes a datepicker, or inline calendarMonth. the user may not want a date to always be applied to the filterable data?

mcarrano: I don't feel strongly about this one, but I agree there may be instances where having a date selected by default does not make sense.

tlabaj: Ok.  I am fine with that, @thatblindgeye,  we should update the original epic issue to say why we are not defaulting to "today" since that was the ask.

tlabaj (approved):

PR #7082, opened 2022-03-17
chore(CalendarMonth): add date range example

### LINKED ISSUE #7027: CalendarMonth: Update examples to match core
Add example for date range selection to match core examples.

thatblindgeye:
**What**: Closes #7027, #7124

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
@@ -279,13 +279,12 @@ export const CalendarMonth = ({
                 const isSelected = isValidDate(dateProp) && isSameDate(date, dateProp);
                 const isFocused = isSameDate(date, focusedDate);
                 const isAdjacentMonth = date.getMonth() !== focusedDate.getMonth();
+                const isRangeStart = isValidDate(rangeStart) && isSameDate(date, rangeStart);
                 let isInRange = false;
-                let isRangeStart = false;
                 let isRangeEnd = false;
                 if (isValidDate(rangeStart) && isHoveredDateValid) {
                   if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
                     isInRange = date > rangeStart && date < hoveredDate;
-                    isRangeStart = isSameDate(date, rangeStart);
                     isRangeEnd = isSameDate(date, hoveredDate);
                   }
                   // Don't handle focused dates before start dates for now.
--- a/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
@@ -286,9 +286,9 @@ export const CalendarMonth = ({
                 const isRangeStart = isValidDate(rangeStart) && isSameDate(date, rangeStart);
                 let isInRange = false;
                 let isRangeEnd = false;
-                if (isValidDate(rangeStart) && isValidDate(selectedDate)) {
-                  isInRange = date > rangeStart && date < selectedDate;
-                  isRangeEnd = isSameDate(date, selectedDate);
+                if (isValidDate(rangeStart) && isValidDate(dateProp)) {
+                  isInRange = date > rangeStart && date < dateProp;
+                  isRangeEnd = isSameDate(date, dateProp);
                 } else if (isValidDate(rangeStart) && isHoveredDateValid) {
                   if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
                     isInRange = date > rangeStart && date < hoveredDate;
--- a/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
@@ -113,7 +113,11 @@ export const CalendarMonth = ({
   // eslint-disable-next-line prefer-const
   let [focusedDate, setFocusedDate] = React.useState(new Date(dateProp));
   if (!isValidDate(focusedDate)) {
-    focusedDate = today;
+    if (isValidDate(rangeStart)) {
+      focusedDate = rangeStart;
+    } else {
+      focusedDate = today;
+    }
   }
   const [hoveredDate, setHoveredDate] = React.useState(new Date(focusedDate));
   const focusRef = React.useRef<HTMLButtonElement>();
@@ -282,7 +286,10 @@ export const CalendarMonth = ({
                 const isRangeStart = isValidDate(rangeStart) && isSameDate(date, rangeStart);
                 let isInRange = false;
                 let isRangeEnd = false;
-                if (isValidDate(rangeStart) && isHoveredDateValid) {
+                if (isValidDate(rangeStart) && isValidDate(selectedDate)) {
+                  isInRange = date > rangeStart && date < selectedDate;
+                  isRangeEnd = isSameDate(date, selectedDate);
+                } else if (isValidDate(rangeStart) && isHoveredDateValid) {
                   if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
                     isInRange = date > rangeStart && date < hoveredDate;
                     isRangeEnd = isSameDate(date, hoveredDate);
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
@@ -17,3 +17,8 @@ beta: true
 
 ```ts file="./CalendarMonthSelectableDate.tsx"
 ```
+
+### Date range
+
+```ts file="./CalendarMonthDateRange.tsx"
+```
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
@@ -20,9 +20,9 @@ beta: true
 
 ### Date range
 
-In this example, there are two dates selected: the first is a range start date (via the `rangeStart` prop) and the second is a range end date (via the `date` prop). For the purposes of this example, these dates are static and cannot be updated.
+In this example, there are 2 dates selected: a range start date (via the `rangeStart` prop) and a range end date (via the `date` prop). Additionally, any dates prior to the range start date are disabled by passing in an array of functions to the `validators` prop. In this case a single function is passed in, which checks whether a date is greater than or equal to the range start date.
 
-Additionally, any dates prior to the range start date are disabled by passing in an array of functions to the `validators` prop. In this case a single function is passed in, which checks whether a date is greater than or equal to the range start date.
+For this example, these dates are static and cannot be updated. For an interactive demo, see our [Date picker demos](https://www.patternfly.org/v4/components/date-picker/react-demos).
 
 ```ts file="./CalendarMonthDateRange.tsx"
 ```
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
@@ -20,11 +20,7 @@ beta: true
 
 ### Date range
 
-In this example, there are two dates selected: the first is a range start date and the second is a range end date. As you hover over dates that take place _after_ the range start date, a style is applied to show which dates are included in that range. This style will persist as you continue to hover within the calendar, even after selecting a range end date.
-
-This example only allows updating of the range end date, though it does not prevent selecting a date prior to the range start date. Preventing a range end date that occurs prior to the range start date, or adding the ability to update the range start date, is functionality that would need to be added manually.
-
-The "Range Selection" text is not a part of the component and is only present for demonstration purposes.
+In this example, there are two dates selected: the first is a range start date (via the `rangeStart` prop) and the second is a range end date (via the `date` prop). For the purposes of this example, these dates are static and cannot be updated.
 
 ```ts file="./CalendarMonthDateRange.tsx"
 ```
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
@@ -20,5 +20,11 @@ beta: true
 
 ### Date range
 
+In this example, there are two dates selected: the first is a range start date and the second is a range end date. As you hover over dates that take place _after_ the range start date, a style is applied to show which dates are included in that range. This style will persist as you continue to hover within the calendar, even after selecting a range end date.
+
+This example only allows updating of the range end date, though it does not prevent selecting a date prior to the range start date. Preventing a range end date that occurs prior to the range start date, or adding the ability to update the range start date, is functionality that would need to be added manually.
+
+The "Range Selection" text is not a part of the component and is only present for demonstration purposes.
+
 ```ts file="./CalendarMonthDateRange.tsx"
 ```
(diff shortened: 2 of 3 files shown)


thatblindgeye:
> Range selection is not working at all for me. I can select a From date and then a To date, but the range does not persist on the calendar. Just leaves me with only the To date selected.

@mcarrano I see what you mean. I'm waiting on another PR to merge before changing this from a draft PR and fixing it up so I can re-requst your review once things are more up to date. But right now removing the `onChange={setDate}` prop from the example should show things working more as intended.

thatblindgeye:
@mcarrano I should have asked beforehand, but is the expectation with this example that we would be able to select both a starting *and* ending range? Right now the starting range can't actually be updated in the example, so we're only able to select a new ending date. If the expectation is to set both, that might be worth a discussion about how to go about that since I believe there's several avenues to explore.

I'm also not sure if it would help or is what might be causing some of the issues you're seeing, but I can also update the example so that the "selected" date and the rangeStart date aren't the same (which right now they are).

In terms of crossing months, when I open the preview build for that example this is the behavior I'm seeing with both keyboard and mouse selection (the styling is a bit subtle, but the range styling is being applied):

https://user-images.githubusercontent.com/70952936/159776864-c4ef5640-2ba6-48d1-ad6f-8f3094b0f870.mov

mcarrano:
> @mcarrano I should have asked beforehand, but is the expectation with this example that we would be able to select both a starting and ending range? Right now the starting range can't actually be updated in the example, so we're only able to select a new ending date. If the expectation is to set both, that might be worth a discussion about how to go about that since I believe there's several avenues to explore.

Good question @thatblindgeye . I'm not completely sure what this example is intended to show since it's being shown outside of any practical context. @tlabaj @nicolethoen what do you expect? I'm fine with keeping this as is provided that we add some documentation to explain what this example does (and doesn't do).

> In terms of crossing months, when I open the preview build for that example this is the behavior I'm seeing with both keyboard and mouse selection (the styling is a bit subtle, but the range styling is being applied):

It may be fine then. Would it be helpful to read out the dates selected in the example? It's hard to actually see what's going on when the range is crossing between months.

thatblindgeye:
> It may be fine then. Would it be helpful to read out the dates selected in the example? It's hard to actually see what's going on when the range is crossing between months.

That wouuld be a good call I think. Depending on the response from Nicole and/or Titani, I can add some verbiage for what exactly the example is/isn't doing and also add in something similar to the "Selectable date" example where a string of text gets updated to better show what's happening on changes.

mmenestr:
Mine is acting very buggy? When I have a number selected, and then start hovering around, a different number (24 in this case) acts like i've selected it even though I haven't. And when I do select a "to" number and move my cursor away from it, the range styling kind of flickers on and off? Not sure what to make of this haha

https://user-images.githubusercontent.com/51165119/159796798-6c1014d0-a02b-421d-b536-88d9fa325aff.mov

thatblindgeye:
@mmenestr Ah yeah, I can see that being confusing. I believe the intent there was essentially, if you have a "start date" selected (in this example it's November 24) and you hover over a date prior to that, the range styles don't take effect. Unfortunately it looks like the styling for that "start date" is being affected when it should probably not be.

For that range styling sort of flickering, the above would apply, too, but also it's showing what the "new range" might be as you continuue to hover around.

I think adding some more context as Matt suggested above might help with some confusion here, though another tweak or two I think could be done here.

thatblindgeye:
@mcarrano @mmenestr I made some quick updates to the example, namely adding some verbiage to describe what is actually happening in the example as well as some text that will dynamically update as you choose a new end date (similar to the "Selectable date" example).

Depending on if it's decided to add in functionality to change both the start and end dates, I can add that in as well. I wanted to provide this update right now to see if it helps clear things up.

mcarrano (approved):
Thanks for making these updates @thatblindgeye . The explanation and the reflection back about the dates really helps IMO.

nicolethoen (commented):
This is great!

The only interaction i'd have expected to be different is after i've selected an end date, i wouldnt expect the range to keep following my cursor as I move my mouse. Sort of like this screen shot of airbnb's range picker (Note: I realized after the screen grab that you can't see the grey date range between the start and end dates in this gif... but it's there.... weird)
Mar-24-2022 15-15-33

mcarrano:
> The only interaction i'd have expected to be different is after i've selected an end date, i wouldnt expect the range to keep following my cursor as I move my mouse. Sort of like this screen shot of airbnb's range picker

@nicolethoen That bugged me a bit, too. Looks like the date picker has the same behavior, but I'm not sure what I think it should do instead. How does it know if you want to change the end date? @mmenestr any thoughts?

thatblindgeye:
Would this sort of behavior be more inline with expectations? Keep in mind that without the `date` prop passed in, the calendar defaults to showing the current month so the range is going all the way from 11/2020 to today (this could either be looked into as part of this, or I can create another issue since this component is still in Beta)

https://user-images.githubusercontent.com/70952936/160002162-3e37d11e-b181-4304-941b-74b63ed1ef29.mov

wise-king-sullyman (approved):
This looks great to me!

I like the functionality shown in your last screen recording @thatblindgeye, but since it looks to me like the change to facilitate that would be on the component side and this issue is for the example I'm comfortable approving.

mmenestr:
I think i'd expect it to work the same way this date picker works; https://patternfly-react-pr-7082.surge.sh/components/date-picker/react-demos#date-range-picker. So essentially what you have, except the range highlight reflect what the new range would be if you were to select what you're hovering over!

https://user-images.githubusercontent.com/51165119/160009444-18989618-9f70-4cdb-a81b-794876e4175a.mov

thatblindgeye:
@mmenestr Ah gotcha, that should be how it currently works as of the current state of this PR 🙂 . My last screen recording above I just took locally to test it out.

@nicolethoen If you'd like I can open up a separate issue regarding the range highlighting if it's something we would want to implement or keep in the queue to discuss.

packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
```diff
@@ -280,13 +280,12 @@ export const CalendarMonth = ({
                 const isSelected = isValidDate(selectedDate) && isSameDate(date, selectedDate);
                 const isFocused = isSameDate(date, focusedDate);
                 const isAdjacentMonth = date.getMonth() !== focusedDate.getMonth();
+                const isRangeStart = isValidDate(rangeStart) && isSameDate(date, rangeStart);
                 let isInRange = false;
-                let isRangeStart = false;
                 let isRangeEnd = false;
                 if (isValidDate(rangeStart) && isHoveredDateValid) {
                   if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
                     isInRange = date > rangeStart && date < hoveredDate;
```
nicolethoen, line 288: If i change the logic in this if statement to:
```
if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
    isInRange = date > rangeStart && date < selectedDate;
    isRangeEnd = isSameDate(date, selectedDate);
}
```
then the hover behavior updates to look like this:
Mar-25-2022 08-56-48

I personally think this is more expected? What do you all think @wise-king-sullyman, @thatblindgeye, @mcarrano, @mmenestr ?

wise-king-sullyman: I do think that's more in line with what I would expect as well.

mcarrano:
> I personally think this is more expected? What do you all think @wise-king-sullyman, @thatblindgeye, @mcarrano, @mmenestr ?

I agree. This feels much confusing.

packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
```diff
@@ -280,13 +280,12 @@ export const CalendarMonth = ({
                 const isSelected = isValidDate(selectedDate) && isSameDate(date, selectedDate);
                 const isFocused = isSameDate(date, focusedDate);
                 const isAdjacentMonth = date.getMonth() !== focusedDate.getMonth();
+                const isRangeStart = isValidDate(rangeStart) && isSameDate(date, rangeStart);
                 let isInRange = false;
-                let isRangeStart = false;
                 let isRangeEnd = false;
                 if (isValidDate(rangeStart) && isHoveredDateValid) {
                   if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
```
nicolethoen, line 287: Sorry I noticed a slight issue with my original suggestion, 
Mar-25-2022 08-59-10
That is the behavior I was suggesting.

but I think the code that needs to be updated is more like
on CalendarMonth.tsx line 116
```if (!isValidDate(focusedDate)) {
    if (isValidDate(rangeStart)) {
      focusedDate = rangeStart;
    } else {
      focusedDate = today;
    }
  }
```
That will focus on the start date if it’s provided and no date prop is set
and then on line 290
```
                if (isValidDate(rangeStart) && isHoveredDateValid) {
                  if (hoveredDate > rangeStart || isSameDate(hoveredDate, rangeStart)) {
                    if (isValidDate(selectedDate)) {
                      isInRange = date > rangeStart && date < selectedDate;
                      isRangeEnd = isSameDate(date, selectedDate);
                    } else {
                      isInRange = date > rangeStart && date < hoveredDate;
                      isRangeEnd = isSameDate(date, hoveredDate);
                    }

                  }
                  // Don't handle focused dates before start dates for now.
                  // Core would likely need new styles
                }
```

thatblindgeye:
Based on @nicolethoen recent suggestion, the latest updates implement the following:

- If there is a `rangeStart` date passed in but no "end" `date` prop passed in, the CalendarMonth component will focus on the month that the `rangeStart` occurs (previously it focused on today's month, so basically the example's range was going from 2020 to 2022 more or less)
- Once and "end" `date` is passed in/valid, the range styles that occur when hovering over a date stop. This is something I think may not have a clear cut answer right now and we may need to get more input from consumers/users, as there are good arguments for both behaviors.

@mcarrano @wise-king-sullyman @mmenestr If you'd like to take a look at the latest updates when the build finishes (Matt and Austin I know you already approved but just in case)

mcarrano:
I like it! Thanks @nicolethoen for making this suggestion.

wise-king-sullyman:
Agree, this is great! 🚀

tlabaj:
> > @mcarrano I should have asked beforehand, but is the expectation with this example that we would be able to select both a starting and ending range? Right now the starting range can't actually be updated in the example, so we're only able to select a new ending date. If the expectation is to set both, that might be worth a discussion about how to go about that since I believe there's several avenues to explore.
> 
> Good question @thatblindgeye . I'm not completely sure what this example is intended to show since it's being shown outside of any practical context. @tlabaj @nicolethoen what do you expect? I'm fine with keeping this as is provided that we add some documentation to explain what this example does (and doesn't do).
> 
> > In terms of crossing months, when I open the preview build for that example this is the behavior I'm seeing with both keyboard and mouse selection (the styling is a bit subtle, but the range styling is being applied):
> 
> It may be fine then. Would it be helpful to read out the dates selected in the example? It's hard to actually see what's going on when the range is crossing between months.

@mcarrano I was actually thinking this would be more of a static example to show how a date range selection would look.  I think to show selection of end and start date, we would want more of a demo to demonstrate how you would select end and start date.  This demo may involve other control (similar to how you might choose travel date on an airline). I think as this is right now, it is little confusing for the consumer.

tlabaj (changes requested):
I think this should be a static demo.  See my comment in the thread.

packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
```diff
@@ -17,3 +17,14 @@ beta: true
 
 ```ts file="./CalendarMonthSelectableDate.tsx"
 ```
+
+### Date range
+
+In this example, there are two dates selected: the first is a range start date and the second is a range end date. As you hover over dates that take place _after_ the range start date, a style is applied to show which dates are included in that range. This style will persist as you continue to hover within the calendar, even after selecting a range end date.
```
tlabaj, line 23: So I think we can leave the example as an end date picker.  I would improve the text her to hopefully alleviate some confusion.  I think we want to emphasis the fact that the example demonstrates selection of an end range.   Ideally we would link to the Date Pick demo that shows selection of both start and end date date range (we don't have a precedent for that though).  I would definitely remove the "adding the ability to update the range start date, is functionality that would need to be added manually."  I don't think that makes sense in the context of this example.

tlabaj:
I am also noticing that the default example is allowing date selection.  I am not sure when that regression was introduced.  IT seems to work correctly on the website.

packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
tlabaj: I also think we should disable the dates before the start date.  I am not sure why we don't have disable functionality built right now (Probably an oversight when initially implemented).  The core example shows disabled dates.
https://www.patternfly.org/v4/components/calendar-month/html#range-end-date-selected-start-date-focused

tlabaj: I opened this issue to address the disabled dates
https://github.com/patternfly/patternfly-react/issues/7124

thatblindgeye:
@tlabaj latest commit should have fixed the regression introduced in the PR you had linked and updated the example, so the "Default" and "Date range" examples are static and cannot have any dates actually updated.

I kept in a mention that the "Date range" example was static and that the dates could not be changed just to alleviate some confusion of the purpose of the example.

mcoker (approved):
LGTM!

tlabaj (changes requested):
We should get this issue resolved before we merge thishttps://github.com/patternfly/patternfly-react/issues/7124

kmcfaul (approved):
LGTM once #7130 goes in

tlabaj (approved):
LGTM!  Thank you for the Updates @thatblindgeye.  I don't think we need another example for disabled dates.  
@mcarrano We can take this in once repos are open if it looks good to you

mcarrano:
@thatblindgeye @tlabaj I think this is OK, but it still feels a bit strange that when I click on a new date, it does not modify the range although the focus updating makes you think it is. Maybe we just bold this text to make sure people understand this is not intended to demonstrate full interaction? "For the purposes of this example, these dates are static and cannot be updated."

tlabaj:
@mcarrano you could technically argue the same abut the `default` example.  You can not select a date there either even if you click on a new date.  That being said, yes we can try and update the description So it is more clear what the purpose of the example is.  I still think mentioning/linking the date picker demo would be useful here.  to demonstrate an actual use case.

packages/react-core/src/components/CalendarMonth/examples/CalendarMonth.md
```diff
@@ -17,3 +17,12 @@ beta: true
 
 ```ts file="./CalendarMonthSelectableDate.tsx"
 ```
+
+### Date range
+
+In this example, there are two dates selected: the first is a range start date (via the `rangeStart` prop) and the second is a range end date (via the `date` prop). For the purposes of this example, these dates are static and cannot be updated.
```
tlabaj, line 23: @wolfeallison do you mind taking a quick look at the copy here.  We are trying to get  across the point that this is a static example and you can not change the date range.  We have a link to a deme for the date picker where we show how to chose a date range.

wolfeallison: Hi! I slightly changed some of it, like using a numeral instead of a number and made it a little more concise. "In this example, there are 2 dates selected: a range start date (via the `rangeStart` prop) and a range end date (via the `date` prop). For this example, these dates are static and cannot be updated. Additionally, any dates prior to the range start date are disabled by passing in an array of functions to the `validators` prop. In this case a single function is passed in, which checks whether a date is greater than or equal to the range start date."

wolfeallison: Because bold is used for component names or to help with grouping, I think it may be confusing to bold other parts. What if we put "For this example, these dates are static and cannot be updated" at the start of a new paragraph so the user is more likely to read and see it?

thatblindgeye: Based on the feedback I've updated the verbiage per @wolfeallison comments. I also included a line to link to the date picker demos (which I can either further update or remove if it ends up not working in the context).

mcarrano (approved):
Looks good to me. Thanks @thatblindgeye !

PR #7084, opened 2022-03-18
fix(react-styles): support esm import and export

### LINKED ISSUE #7054: Provide ESM export for @patternfly/react-styles css components.
Using bundlers that support ES Modules (such as vite) require ES Module exports. Currently, all of the component .js files under @patternfly/react-styles/css/components use `require` on .css files. It would be better to include in addition to this another file that uses `import` instead.

A nice to have in addition to this es module file would be placing the `@patternfly/react-styles/css` directory under both `@patternfly/react-styles/dist/esm` and `@patternfly/react-styles/dist/js` . This would allow for bundlers to easily determine which .js files to use for ES modules and commonjs modules.

kunyan:
The previous version are using `require()`, which is not fully
support esm

Add esm only output to support some pure esm project like vite or
snowpack

**What**: Closes #

Closes #7054

**Additional issues**:

### DIFF
--- a/packages/react-styles/scripts/writeClassMaps.js
+++ b/packages/react-styles/scripts/writeClassMaps.js
@@ -15,6 +15,15 @@ exports.default = ${JSON.stringify(classMap, null, 2)};
 `.trim()
   );
 
+const writeESMExport = (file, classMap) =>
+  outputFileSync(
+    join(outDir, file.replace(/.css$/, '.mjs')),
+    `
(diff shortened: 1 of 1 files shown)


mcoker (approved):
Spot checked a bunch of components, everything looks good to me!

jschuler (approved):

nicolethoen (approved):

PR #7089, opened 2022-03-18
docs(MenuToggle): update examples to match core

### LINKED ISSUE #6998: Fix vertical positioning of an image in Menu toggle
Follow up to https://github.com/patternfly/patternfly/pull/4701

### LINKED ISSUE #7088: MenuToggle: secondary variant
Followup to patternfly/patternfly#4498

Add the secondary variant to MenuToggle and update its examples to include the secondary example from core.

wise-king-sullyman:
**What**: Closes #6998 and also closes #7088

**Additional issues**:

Link to the first changed example for convenience: https://patternfly-react-pr-7089.surge.sh/components/menu-toggle#secondary


nicolethoen (commented):
Should this PR include any changes to the primary example with an icon? It seems misaligned as is.

mcarrano (changes requested):
I'm unclear about the way these React examples are structured. Why do we need both collapsed and expanded versions of the toggle? Shouldn't the state just toogle when I click on it? Also the Expanded toggle seems to toggle the Collapsed one when I click it. Strange.

I also feel like we should have an icon version for Secondary. @mcoker was there a reason we didn't include that in core?

mcoker:
> I also feel like we should have an icon version for Secondary. @mcoker was there a reason we didn't include that in core?

@mcarrano not that I know of. I added one here - https://github.com/patternfly/patternfly/pull/4743

wise-king-sullyman:
@nicolethoen I see what you're saying about it being misaligned, but I don't believe that anything in this PR should be changing that? It seems that all the classes are being applied as they are in core.

@mcoker @mattnolting is this something that was changed somehow in patternfly/patternfly#4701 and it now needs a corresponding change in React?

wise-king-sullyman:
> Why do we need both collapsed and expanded versions of the toggle? Shouldn't the state just toogle when I click on it?

@mcarrano I followed the [current practices being used in the React examples for this component](https://www.patternfly.org/v4/components/menu-toggle/), but if you'd like I can rework all of the examples here or just change how these new examples are presented. I'm not sure if the examples were done in this way for any particular reason though.

mcarrano:
@nicolethoen @tlabaj what's your opinion on the above? I thought it was odd to show the states separately rather than just use the on_click event to toggle them, but maybe there's a good reason for this that I'm not seeing? I agree with @wise-king-sullyman that if we are to change this, we should change all the examples to be consistent.

mcoker:
@wise-king-sullyman that's interesting, I've seen that before and I've always found a workaround though I'm not sure what the root cause is. I created an issue to track it here - https://github.com/patternfly/patternfly/issues/4746 and put together a few examples trying to figure out what's going on. Hopefully it's something we can solve in core.

nicolethoen:
> @nicolethoen @tlabaj what's your opinion on the above? I thought it was odd to show the states separately rather than just use the on_click event to toggle them, but maybe there's a good reason for this that I'm not seeing? I agree with @wise-king-sullyman that if we are to change this, we should change all the examples to be consistent.

@mcarrano  Without an attached menu to toggle opened/closed, it does make the various states a little harder to clearly demonstrate in the examples. There are three ways we could handle this:

1. As @wise-king-sullyman has done, we could simply put a MenuToggle for each of the states in the examples. Given that there are only three states, this isn't too bad.
2. Update the MenuToggle text to reflect the states of the MenuToggle, so it'd read 'Collapse' or 'Expanded' when the user clicks on it. In this case, we'd still need a second disabled state. Additionally, when scanning our docs, they wouldn't see the expanded state, nor tell without further hinting that they can reveal an expanded state by interacting with the MenuToggle examples.
3. Update the MenuToggle examples to use some sort of controls like checkboxes or ToggleGroups to allow the user to change the state of one MenuToggle.

mcarrano:
@wise-king-sullyman @nicolethoen I'm not sure what's best here, so let's keep it as is then (option #1). Once the variant for a secondary with icon is added, I will go ahead and approve this.

tlabaj:
> @nicolethoen @tlabaj what's your opinion on the above? I thought it was odd to show the states separately rather than just use the on_click event to toggle them, but maybe there's a good reason for this that I'm not seeing? I agree with @wise-king-sullyman that if we are to change this, we should change all the examples to be consistent.

@mcarrano I think what Austin has here is consistent with our other examples for the toggle. (e.g. https://www.patternfly.org/v4/components/menu-toggle#primary). This makes sense to me since the example is a toggle in isolation (not attached to a menu).  A demo would be where I would expect to see the actual interactions demonstrated with a menu).

mcarrano (commented):
@wise-king-sullyman Did we ever figure out the reason for the mis-alignment of the buttons with an icon? I see the same problem for both Primary and Secondary versions. I also see this in the current release so it's nothing being introduced by this PR, but would be nice to fix this here if we can.

Screen Shot 2022-03-23 at 7 17 51 PM

mcoker:
@mcarrano I created a core issue to see if it's something we can fix - https://github.com/patternfly/patternfly/issues/4746

Though to fix it here, we could add this local style to the example box:

```
#ws-react-c-menu-toggle-primary .pf-c-menu-toggle {
    vertical-align: bottom;
}
```

wise-king-sullyman:
> Though to fix it here, we could add this local style to the example box:

> #ws-react-c-menu-toggle-primary .pf-c-menu-toggle {
    vertical-align: bottom;
}

@mcarrano would you like me to add this styling on the React side of these examples as a temporary fix?

mcoker:
> @mcarrano would you like me to add this styling on the React side of these examples as a temporary fix?

@mcarrano fyi this is just a local workspace style, it won't show up in the react code or be a part of patternfly. Also worth noting that this alignment issue only happens if you place 2 buttons inline with one another side by side. Typically you wouldn't do that - the buttons would be part of a flex/grid layout that lays them out and manages the spacing between them.

mcarrano:
> @mcarrano would you like me to add this styling on the React side of these examples as a temporary fix?

Yes, I think we should.

packages/react-core/src/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -123,6 +137,34 @@ import { MenuToggle } from '@patternfly/react-core';
...
+
+```ts
+import React from 'react';
+import { MenuToggle, Avatar } from '@patternfly/react-core';
+import CogIcon from '@patternfly/react-icons/dist/esm/icons/cog-icon';
+
+<React.Fragment>
+  <MenuToggle icon={<CogIcon />} variant="secondary">Icon</MenuToggle>{' '}
+  <MenuToggle icon={<CogIcon />} variant="secondary" isDisabled>Icon</MenuToggle>
+</React.Fragment>
```
thatblindgeye, line 151: The disabled variant for the icon toggle here.

packages/react-core/src/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -123,6 +137,34 @@ import { MenuToggle } from '@patternfly/react-core';
...
+```ts
+import React from 'react';
+import { MenuToggle, Avatar } from '@patternfly/react-core';
+import avatarImg from './avatarImg.svg';
+
+<React.Fragment>
+  <MenuToggle icon={<Avatar src={avatarImg} alt="avatar" />}>Ned Username</MenuToggle>{' '}
+  <MenuToggle icon={<Avatar src={avatarImg} alt="avatar" />} isExpanded>Ned Username</MenuToggle>{' '}
+  <MenuToggle icon={<Avatar src={avatarImg} alt="avatar" />} isDisabled>Ned Username</MenuToggle>
+</React.Fragment>
```
thatblindgeye, line 165: The avatar toggles here

thatblindgeye (approved):
Looks good with the added temp. styling! I only had a nitpick for two examples marked below. For both, since other examples state via text in the toggle which toggle it is (e.g. whether it's disabled or expanded), it might be worth doing so on the below ones as well.

That said, the plain example can't exactly do that as easily so it's not like it's a hard rule or anything.

wise-king-sullyman:
@thatblindgeye I like that suggestion, matching with how things show in the other examples makes a lot of sense to me.

That *would* require the core examples to be updated, or the React examples to differentiate from the core examples in that way. @mcarrano do you have any thoughts on this?

mcarrano:
I don't feel like the core and React examples need to be identical in all cases as the React examples have the advantage of using Javascript to show state. I'm not so concerned at the moment about the fact the the plain and avatar examples can't be labeled. In the long term, I'd like to see us rework a page like this to use checkboxes or something like that to select state rather than having multiple instances of the same thing. But I don't think that's something we should try to accomplish with this PR.

As for the problem at hand with the mis-alignment, looks like your last change fixed that @wise-king-sullyman .

packages/react-core/src/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -123,6 +137,34 @@ import { MenuToggle } from '@patternfly/react-core';
...
+  <MenuToggle icon={<CogIcon />} variant="secondary" isDisabled>Icon</MenuToggle>
+</React.Fragment>
+```
+
+### With avatar and text
+
+```ts
+import React from 'react';
+import { MenuToggle, Avatar } from '@patternfly/react-core';
+import avatarImg from './avatarImg.svg';
```
mcoker, line 159: Could you import from the avatar component like the dashboard wrapper does? Then we're using the same image consistently. https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/demos/examples/DashboardHeader.js#L30

mcoker (changes requested):

mcoker (approved):
🥳

tlabaj (approved):
Lgtm

PR #7091, opened 2022-03-18
fix(Card): indicate card selectivity and status if using a screen reader

### LINKED ISSUE #6798: Selectable cards are not announced with screen reader
Currently, there is nothing about selectable cards that indicates to screen readers that they are selectable or interactable.

Selectable cards likely need an aria `role=option` and need to be placed in a parent container with an aria `role=listbox`
If multiple cards are selectable, then the parents container with the listbox role should also have `aria-multiselectable=true` and `aria-activedescendant` set up to contain the id of the card which currently has focus.
The parent will also need an `aria-label` if the list of cards does not have a title, or can reference a visual title using `aria-labelledby`.

When a card is selected, the card should use the `aria-selected=true`

this fix needs to be applied to the card/Selectable and selected example, the card/Legacy selectable and selected example, the Card view demo, the Tabs/modal tabs demo, Primary-detail card view demo.

wise-king-sullyman:
**What**: Closes #6798

**Additional issues**:

(I believe) I've implemented all of Nicole's suggestions from the issue and applied them to the card view demo. ~~If everything looks like it works well enough I'll update the other places Card is being used to the same pattern used here,  then convert this to a real PR.~~

Direct link to the card view demo for convenience: https://patternfly-react-pr-7091.surge.sh/demos/card-view/react-demos/card-view/

### DIFF
--- a/packages/react-core/src/components/Card/examples/Card.md
+++ b/packages/react-core/src/components/Card/examples/Card.md
@@ -68,6 +68,10 @@ import pfLogoSmall from './pf-logo-small.svg';
 
 ### Selectable
 
+Note: The listbox role on the containing div of this examples is needed because the cards have the option role via their `isOption` prop, and options are required to be inside a listbox. See the [option role documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role) from MDN for more information.
+
+The cards need to be options in order to effectively communicate that they're selectable (and what their current selection state is) for those using screen readers.
+
 ```ts file='./CardSelectable.tsx'
 ```
 
--- a/packages/react-core/src/components/Card/examples/Card.md
+++ b/packages/react-core/src/components/Card/examples/Card.md
@@ -70,7 +70,7 @@ import pfLogoSmall from './pf-logo-small.svg';
 
 Note: The listbox role on the containing div of this examples is needed because the cards have the option role via their `isOption` prop, and options are required to be inside a listbox. See the [option role documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role) from MDN for more information.
 
-The cards need to be options in order to effectively communicate that they're selectable (and what their current selection state is) for those using screen readers.
+The cards need to be options in order to effectively communicate that they're selectable (and what their current selection state is) for those using assistive technologies (such as screen readers).
 
 ```ts file='./CardSelectable.tsx'
 ```
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -36,6 +36,8 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   isPlain?: boolean;
   /** Flag indicating if a card is expanded. Modifies the card to be expandable. */
   isExpanded?: boolean;
+  /** Flag indicating that the card is an option in a listbox */
+  isOption?: boolean;
 }
 
 interface CardContextProps {
@@ -65,6 +67,7 @@ export const Card: React.FunctionComponent<CardProps> = ({
   isLarge = false,
   isFullHeight = false,
   isPlain = false,
+  isOption = false,
   ouiaId,
   ouiaSafe = true,
   ...props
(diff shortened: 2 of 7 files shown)


thatblindgeye (commented):
Looks (and sounds) good! I think the only thing is that the first Card, which is the "add card" one, is being announced as selectable when it actually isn't. Removing the `isOption` prop from it resolves that, but initially upon entering the listbox there doesn't seem to be any actual context for that Card other than "You are currently on an article."

Using `aria-describedby` on that card and linking it to its `<Title>` element kind of works, but still a bit iffy. Going to see if there's anything I could come across tomorrow, or whether anyone else has an idea.

wise-king-sullyman:
@thatblindgeye looks like having non options in a listbox isn't allowed.

I think either the card-adding card will need to be an option or the actually selectable cards will need to be placed in a separate div to serve as the listbox, and I don't like either of those ideas.

Any thoughts for a better approach than what I've come up with?

thatblindgeye:
> @thatblindgeye looks like having non options in a listbox isn't allowed.
> 
> I think either the card-adding card will need to be an option or the actually selectable cards will need to be placed in a separate div to serve as the listbox, and I don't like either of those ideas.
> 
> Any thoughts for a better approach than what I've come up with?

The second option is the best of the two, but based on how this demo is meant to be yeah neither option is really ideal.

Depending on the feedback from anyone else, perhaps the best option right now is to have all those cards as `isOption`, and making a note on the Demo page for it? Overall things work as intended with this one exception, so ultimately I would be okay approving this and having another issue created that could be investigated down the line.

nicolethoen (commented):
I agree with @thatblindgeye 
Adding back the option role to the first card is probably the closest we can get it to ideal without more dramatically altering the demo structure - which we'd want to do first in Core i think.
Once that and my note about the example including information about 'listbox', i'll approve.

packages/react-core/src/components/Card/examples/Card.md
```diff
@@ -68,6 +68,10 @@ import pfLogoSmall from './pf-logo-small.svg';
 
 ### Selectable
 
+Note: The listbox role on the containing div of this examples is needed because the cards have the option role via their `isOption` prop, and options are required to be inside a listbox. See the [option role documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role) from MDN for more information.
+
+The cards need to be options in order to effectively communicate that they're selectable (and what their current selection state is) for those using screen readers.
+
```
thatblindgeye, line 74: **Nit:** The final line could maybe be updated from "for those using screen readers" to "for those using assistive technologies (such as screen readers)." Train of thought is mainly inclusivity.

**General comment** that maybe @nicolethoen could weigh in on: Could it be worth noting that adding the `role="option"` to the add card button *isn't* best practice since it conveys that the card is selectable when it isn't?

thatblindgeye (approved):
Looks good all things considered! Only have a final nitpick and general comment, but otherwise I'll approve it.

packages/react-core/src/components/Card/examples/Card.md
wise-king-sullyman: Good callout 🔥

jessiehuff (commented):
So when testing with VO, I found the same as Eric that I couldn't change the size of the drawer. Looking at the attributes though, it doesn't seem like the `aria-valuenow` is getting updated as the splitter is resizing the drawer. I would think that it would increment up or down depending on whether the drawer is getting bigger or smaller. Not being able to use the typical VO commands could also be linked to this attribute not changing when the splitter is moved by mouse too. Often VO can be linked to click commands so that could be a factor. Basically the `aria-valuenow` should update as the splitter is moving.


(Also, unrelated to this PR side comment, but I noticed that with the "Resizable on left" example, the splitter is before the drawer panel in the DOM but looks visually on the right of the panel. This seemed a little confusing that it's out of order when using a keyboard.)

thatblindgeye:
@jessiehuff Hmm yeah I was unable to get to it as well. When VO was active, when I clicked the kebab with my mouse I saw focus briefly on the kebab but then it snapped focus back to the Card itself.

wise-king-sullyman:
@jessiehuff @thatblindgeye with VO active I can get to the kebab by hitting tab once while focused on a card, is that acceptable?

thatblindgeye:
One issue with that is whether a user using VO would know there's a kebab to Tab to; if they primarily use the VO + arrow keys to navigate, there may be no indication of a kebab.

That said, I'm wondering if that would be considered a separate issue from this, as per the original issue it does look resolved (as best as can be at this moment), or if the inability to use VO navigation to get to the kebab would block this update.

wise-king-sullyman:
Closing as it was decided that this would not be the path forward for this issue.

Issue #6835, opened 2022-01-19
Select: onToggle should pass event back


kmcfaul:
To allow event object manipulation.

PR #7093, opened 2022-03-21
feat(Select): add flag to not reset on select, add event to toggle callback

### LINKED ISSUE #6641: Typeahead Multi Select: filter is reset after select
Unexpected behavior in the following user story:
As a user working with the `Select` component example as linked below
- I enter 'new' in the type ahead input text,
- I see the filtered result,
- I select the first option (New Jersey),
- My filter has been reset, so the filter input text is removed after the first selection
image

Expected behavior:

- My filter is not reset and I can select another record in the filtered result 
image

https://www.patternfly.org/v4/components/select/#multiple-with-custom-chip-group-props

enhancement

"@patternfly/react-core": "^4.162.3"

kmcfaul:
**What**: Closes #6641 & #6835

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -122,7 +122,7 @@ export interface SelectProps
     isPlaceholder?: boolean
   ) => void;
   /** Callback for toggle button behavior */
-  onToggle: (isExpanded: boolean) => void;
+  onToggle: (isExpanded: boolean, event: React.MouseEvent | React.ChangeEvent | React.KeyboardEvent | Event) => void;
   /** Callback for typeahead clear button */
   onClear?: (event: React.MouseEvent) => void;
   /** Optional callback for custom filtering */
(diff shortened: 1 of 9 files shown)


packages/react-core/src/components/Select/Select.tsx
```diff
@@ -159,6 +159,8 @@ export interface SelectProps
   isInputValuePersisted?: boolean;
   /** @beta Flag for retaining filter results on blur from keyboard-entered typeahead text */
   isInputFilterPersisted?: boolean;
+  /** Flag indicating that the state of Select such as typeahead input should reset upon selection */
```
nicolethoen, line 162: is it the state or the value of the input? as is, this prop sounds very destructive 😳

nicolethoen (changes requested):
This looks good - the only interaction concern/possible regression I could find:
Mar-21-2022 12-23-54

It's doing weird things with the description text like the gif above.

packages/react-core/src/components/Select/Select.tsx
kmcfaul: I'll update the description to remove the state mention as its internal.

Will also look into the description popping up.

nicolethoen (approved):

tlabaj:
this would be a great candidate for a test to verify adding the flag works as expected.

wise-king-sullyman (approved):

tlabaj (approved):
Looks great!

PR #7095, opened 2022-03-21
chore(Dropdown): update examples to show different appends

### LINKED ISSUE #6824: Dropdown example (Append menu document body) isn't accessible by screen reader
When testing the Dropdown append menu document body example, I noticed that I couldn't get into the menu with VO. I suspect it's because the menu appears in a different location in the DOM as the input. We should investigate how to make this example accessible as there are likely users following our example.

Go to our [Dropdown append menu document body example](https://www.patternfly.org/v4/components/dropdown#append-menu-document-body) and try to test with any screen reader.

bug

thatblindgeye:
**What**: Closes #6824 

Similar to PR #7087, this does not resolve the issue with appending to `document.body`, but rather cautions against it and provides the alternative of appending to `"parent"`. Appending to `"parent"` should offer the same benefits of `document.body`, but also allow assistive technologies to navigate into the component's contents and also keep things in a more proper DOM order.

I also created another codesandbox to show both variants being used inside of a Modal: [Dropdown appended inside Modal](https://codesandbox.io/s/dropdown-appended-inside-modal-zfluqj). This shows the similarity between them when opened inside a Modal, and also the difference if using Voice Over to try and navigate inside the Dropdown contents.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Dropdown/examples/Dropdown.md
+++ b/packages/react-core/src/components/Dropdown/examples/Dropdown.md
@@ -1487,11 +1487,11 @@ class ImageTextDropdown extends React.Component {
 
 ### Appending document body vs parent
 
-When passing in a value to the `menuAppendTo` prop on the Dropdown component, passing in `document.body` should be avoided if possible as doing so can cause accessibility issues. These issues can include (but may not be limited to) being unable to enter the contents of the Dropdown options via assistive technologies (like keyboards or screen readers).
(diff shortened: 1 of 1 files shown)


gitdallas (approved):

thatblindgeye:
In addition to moving the verbiage I also added a quick explanation of what the differences between the two examples here are to match how I have the Select example description. That's easy enough to remove that + the anti-examples if that ends up being way to go, though.

kmcfaul (approved):

jessiehuff (approved):
I like the verbiage. 🙂

mcarrano:
I like it, too!

tlabaj (approved):

Issue #6923, opened 2022-02-09
Select: View more button not accessible via keyboard for typeahead variant


tlabaj:
**Describe the issue. What is the expected and unexpected behavior?**
Select: View more button not accessible via keyboard for typeahead variant.  The View more option shows in the toggle (which it should not), but it is not focused.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
Update the variant of the "View more" example to `typeahead`.  IT appears you are not be able to access view more button by keyboard since it does not get focused, however the "view more appears in toggle when you try and navigate to View more by keyboard.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
bug

Issue #6843, opened 2022-01-20
Select: Grouped + Checkbox + View More not navigable by keyboard


nicolethoen:
When using grouped + checkbox + view more, there is no way to use the keyboard to navigate to the 'view more' option

[codesandbox to reproduce](https://codesandbox.io/s/distracted-flower-90gyec?file=/index.js)

![Feb-16-2022 15-48-12](https://user-images.githubusercontent.com/26305082/154354447-a3cbea08-41c1-43d5-a1ba-ef621de6077f.gif)

PR #7107, opened 2022-03-22
fix(Select): view more fixes

### LINKED ISSUE #6950: Select:  View more with typeahead focus management
The focus should stay on the toggle input after the view more is clicked for typeahead variant.

cc @jessiehuff

kmcfaul:
**What**: Closes #6950, #6923, #6843

Remaining minor issue - typeahead input is not updating after view more is clicked and the list is loaded. I believe this is because the refCollection hasn't updated by the time it renders. I have a workaround for the focus highlighting on the new option. Will try to update this PR if I find a fix otherwise will open a follow up.


thatblindgeye (commented):
6923 (View More unable to be navigated to via keyboard) and 6843 (View More + Grouped + Checkboxes) look resolved.

 I'm unsure whether 6950 (focus management after View More is clicked) is, though, or I may be misunderstanding something.

When I click the View More button inside a `<Select variant={SelectVariant.typeahead}... />` with my mouse, I believe focus is still on the toggle based on the styling, though pressing Tab afterwards I'm unsure where focus actually is. Using a keyboard and using Enter to click View More, focus returns to the input (allowing me to type). If this is the expected behavior disregard, it may just be being a little confused. I included a video to help show what I'm talking about (towards the end I tried using arrow keys and then pressed Tab 1-2 times).

https://user-images.githubusercontent.com/70952936/159974922-1873ef8e-2da4-465a-ae4e-55b9ddc6260c.mov

wise-king-sullyman:
@thatblindgeye ++ I was just about to leave that same comment, other than being unsure about if 6950 is resolved this looks good to me.

wise-king-sullyman (commented):
I did also find some weird stuff happening with the menu if you use the arrow keys to move around in it while the additional options are loading in, I'm not positive if this is really a new problem introduced though, or if the just problem wasn't noticeable before since previously you would lose focus when you selected view more.

Example of what I'm talking about, while the menu is open I only hit the down arrow (until I closed it). Once I close the menu and reopen it the missing items come back.

https://user-images.githubusercontent.com/7115053/159979428-08cd034f-f6e6-4dbf-aab5-73d82d7d5f08.mov

kmcfaul:
Re: the extra grey row highlighting, if your mouse is hovering over an option that will also apply the styling. I'm not sure if that is what is causing the highlight, or it could possibly be an artifact of adding the workaround for the view more (where focus styling is applied to the index of the currently active child, if the default is 0).

Hmm focus is shifting properly on keyboard input. I think what's happening is that when you click an option with the mouse it receives focus (and then the view more option is deleted so focus disappears). I might have to open an additional follow up for this.

wise-king-sullyman:
Extra row highlighting does look like it was caused by my mouse hovering you're right. I'm not sure about the non-highlighting but I haven't been able to replicate that yet this morning.

I'm comfortable approving this if you want to open a followup for #6950.

kmcfaul:
@tlabaj wdyt about the clicking behavior? I'm noticing that in multiple typeahead without view more, focus will still shift to an option when it is clicked via the mouse. I don't think we control clicking that much compared to keyboard presses. I did fix that keyboard interaction will keep focus on the input.

tlabaj:
> @tlabaj wdyt about the clicking behavior? I'm noticing that in multiple typeahead without view more, focus will still shift to an option when it is clicked via the mouse. I don't think we control clicking that much compared to keyboard presses. I did fix that keyboard interaction will keep focus on the input.

@kmcfaul I am not seeing that.  Focus stays on the toggle. the selected option however gets that grey styling.  Am I missing something?

cc @@jessiehuff

kmcfaul:
@tlabaj When you use the mouse only, the keyboard should be working with the fix.

tlabaj:
@kmcfaul the only issue I see from testing is that the "Select: Grouped + Checkbox + View More" does not focus on the first checkbox element when I toggle it like I do with other checkbox select variations.

kmcfaul:
@tlabaj Pasting in the codesandbox from the bug (https://codesandbox.io/s/distracted-flower-90gyec?file=/index.js:238-3335) into the surge preview shows the behavior working for grouped checkbox + view more.

jessiehuff (commented):
I agree with others here that this is definitely an improvement on a number of levels! I'm in the same boat as Titani though where I don't think I'm seeing any issues with the multiple typeahead without view more other than the selected option getting grey styling. 

I'm also seeing Eric and Austin's issues. The weird view more interaction is happening when I'm just using keyboard though:

https://user-images.githubusercontent.com/28281340/160420195-cb22718d-cb6d-49e6-aa0a-e908442040bc.mov

h:

kmcfaul:
@jessiehuff Updated the PR to resolve the issue. Was an artifact of a rebase merge.

kmcfaul:
@tlabaj @thatblindgeye @wise-king-sullyman 

The new option being highlighted after the view more button is clicked and options are loaded is intentional - it indicates the current active option of the list. Focus should be remaining on the input when using keyboard to interact with the view more button (mouse only interaction shifts focus currently and isn't something we control atm). The follow up / issue I opened is to address the bug that the typeahead input should also be filled with the newly shown option.

wise-king-sullyman (approved):
Looks like everything has either been addressed or is out of scope for these issues / slated to be addressed in a followup.

thatblindgeye (approved):
Looks good with the updates! As noted in our convo, I'll open a separate issue for the view more + grouped + checkbox example and the focus management within it.

tlabaj:
@kmcfaul I see that working in the code sandbox, but it seems to not be working in your branch?  Or I am doing something wrong?

kmcfaul:
@tlabaj The codesandbox shouldn't be working, you can't navigate to the view more button via the arrow keys. Pasting that into the branch should show the arrow keys successfully traversing to the view more button.

tlabaj:
@kmcfaulI found my issue.  I forgot to add the "isGrouped" prop.  Looks good!

tlabaj (approved):
LGTM

PR #7118, opened 2022-03-24
fix(Drawer): Updated resizable drawer to make it more accessible

### LINKED ISSUE #5212: Drawer: update splitter keyboard interaction & aria labelling
Drawer splitter keyboard behavior needs updating and aria labelling needs to be refined. See [Jessie's comment](https://github.com/patternfly/patternfly-react/pull/5193#pullrequestreview-544476838)

tlabaj:
**What**: Closes #5212

### DIFF
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -104,7 +104,8 @@ export const DrawerPanelContent: React.FunctionComponent<DrawerPanelContentProps
         drawerContentRef.current.getBoundingClientRect().bottom - drawerContentRef.current.getBoundingClientRect().top;
     }
 
-    return Math.round(splitterPos / drawerSize) * 100;
+    const newSplitterPos = (splitterPos / drawerSize) * 100;
+    return Math.round((newSplitterPos + Number.EPSILON) * 100) / 100;
   };
 
   const handleTouchStart = (e: React.TouchEvent) => {
--- a/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
+++ b/packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
@@ -83,13 +83,15 @@ export const DrawerPanelContent: React.FunctionComponent<DrawerPanelContentProps
     let splitterPos;
     let drawerSize;
 
-    if (isInline) {
-      splitterPos =
-        splitterRef.current.getBoundingClientRect().right - drawerContentRef.current.getBoundingClientRect().left;
+    if (isInline && position === 'right') {
+      splitterPos = panel.current.getBoundingClientRect().right - splitterRef.current.getBoundingClientRect().left;
+      drawerSize = drawerRef.current.getBoundingClientRect().right - drawerRef.current.getBoundingClientRect().left;
+    } else if (isInline && position === 'left') {
+      splitterPos = splitterRef.current.getBoundingClientRect().right - panel.current.getBoundingClientRect().left;
       drawerSize = drawerRef.current.getBoundingClientRect().right - drawerRef.current.getBoundingClientRect().left;
     } else if (position === 'right') {
       splitterPos =
-        splitterRef.current.getBoundingClientRect().left - drawerContentRef.current.getBoundingClientRect().left;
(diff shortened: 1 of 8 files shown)


kmcfaul (approved):

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
```diff
@@ -76,6 +79,34 @@ export const DrawerPanelContent: React.FunctionComponent<DrawerPanelContentProps
...
+      drawerSize =
+        drawerContentRef.current.getBoundingClientRect().right - drawerContentRef.current.getBoundingClientRect().left;
+    } else if (position === 'bottom') {
+      splitterPos =
+        drawerContentRef.current.getBoundingClientRect().bottom - splitterRef.current.getBoundingClientRect().top;
+      drawerSize =
+        drawerContentRef.current.getBoundingClientRect().bottom - drawerContentRef.current.getBoundingClientRect().top;
+    }
+
+    return Math.round(splitterPos / drawerSize) * 100;
```
thatblindgeye, line 107: This is missing a couple of parentheses just before the `Math.round`; right now it's always returning 100 or 0, nothing ever in between:

```suggestion
    return Math.round((splitterPos / drawerSize) * 100);
```

thatblindgeye (changes requested):
Via conventional keyboard the splitter looks to work as intended (e.g. not needing to press Space to "activate" it).

When using Voice Over, however, I'm not able to actually change the size of the drawer using the VO keys + arrow key. @jessiehuff might have some better insight at the moment than me, but I can also try looking into it. That said, I am able to use *just* the arrow keys with VO active, and there's only a simple change that needs to be made below.

jessiehuff (commented):
So when testing with VO, I found the same as Eric that I couldn't change the size of the drawer. Looking at the attributes though, it doesn't seem like the aria-valuenow is getting updated as the splitter is resizing the drawer. I would think that it would increment up or down depending on whether the drawer is getting bigger or smaller. Not being able to use the typical VO commands could also be linked to this attribute not changing when the splitter is moved by mouse too. Often VO can be linked to click commands so that could be a factor. Basically the aria-valuenow should update as the splitter is moving.

(Also, unrelated to this PR side comment, but I noticed that with the "Resizable on left" example, the splitter is before the drawer panel in the DOM but looks visually on the right of the panel. This seemed a little confusing that it's out of order when using a keyboard.)

packages/react-core/src/components/Drawer/DrawerPanelContent.tsx
jessiehuff: Nice catch!!

tlabaj: I had it like that and the linter removed the brackets.

thatblindgeye: That's odd that it'd remove it. Could we try disabling eslint for that line in particular?

thatblindgeye:
Latest update looks to be updating the `aria-valuenow` on the element + getting announced to VO 💪🏼 If it matters, it does end up with decimal values, e.g. 43.11%, but that depends whether it *should* include decimals or not.

Still having issues using the VO + arrows, but I'm wondering if it's either as Jessie stated or whether it has to maybe do with the `Ctrl` key somehow?

thatblindgeye:
I am noticing the behavior Jessie just recently mentioned, with the `aria-valuenow` seeming to have the numbers backwards. Other than that, though, I'm not sure whether the ability to update the drawer with VO is something this PR should try to fix, especially because it seems like there may either be more to dig into with it or maybe it's something more on VO's side?

tlabaj:
@thatblindgeye @jessiehuff I fixed logic for drawer on the right.  I am not sure why VO is not working....

thatblindgeye (approved):
Looks (sounds) good! Not entirely sure about VO either, but if Jessie doesn't disagree then I'm thinking it may be better to open a separate issue if anything for that.

jessiehuff (approved):
LGTM! I agree with Eric on opening an issue for the weirdness with VO for us to take a look at. This is already a massive improvement so I'd love to get it in! 😄

tlabaj:
@jessiehuff @thatblindgeye  follow up issue
https://github.com/patternfly/patternfly-react/issues/7145

PR #7138, opened 2022-03-29
feature(topology): Add BreadthFirst layout to the Topology component

lucasponce:
Fixes https://github.com/patternfly/patternfly-react/issues/7152

This is another useful algorithm to provide a BreadthFirst layout to the Patternfly Topology component.

Algorithms like "cola" and "dagre" have typically a cost to calculate the best visualization of a connected graph, but in the domain sometimes with not good results in terms of clarity and hierarchy of the relations:

image
image

A BreadthFirst layout may help in these cases, providing a natural "levels" approach that can be combined with other algorithms to help users to identify the dependencies between elements:

image

Note, this first version doesn't manage the overflow of a row. That's an optimization that can be added in follow-up work.

In Kiali, the algorithm may detect when a row is pretty large compared with others and can combine a "grid + breadthfirst" in these cases.

But I think this algorithm can be useful for demos and as a first point to improve the component.

@jeff-phillips-18 @andrew-ronaldson

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/TopologyPackage.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/TopologyPackage.tsx
@@ -158,6 +158,9 @@ const TopologyViewComponent: React.FC<TopologyViewComponentProps> = ({ useSideba
             </DropdownItem>,
             <DropdownItem key={6} onClick={() => updateLayout('Concentric')}>
               Concentric
+            </DropdownItem>,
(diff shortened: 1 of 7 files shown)


nicolethoen:
@lucasponce would you mind also opening and attaching an issue to this PR for tracking/release notes purposes?

lucasponce:
> would you mind also opening and attaching an issue to this PR for tracking/release notes purposes?

Sure, thanks for the comment.

lucasponce:
Done @nicolethoen. 
Let me know if you'd need additional info on the issue.

jeff-phillips-18 (approved):

tlabaj (approved):

PR #7144, opened 2022-03-29
fix(Card): indicate card selectivity and status if using a screen reader

### LINKED ISSUE #6798: Selectable cards are not announced with screen reader

wise-king-sullyman:
**What**: Closes #6798 

**Additional issues**:
See PR #7091 for the previous attempt to resolve this issue and the issues identified with that approach.

This draft PR demonstrates a potential approach to improving Card accessibility by including hidden inputs for the purpose of screen reader navigation, this approach is similar to the one used for our [switch component](https://www.patternfly.org/v4/components/switch/html).

I believe this approach has been done in such a way that it is opt-in only and doesn't require a breaking change. The new structure is only rendered when a 'hasHiddenInput' prop is passed.

Currently, the checkboxes are all positioned at the top left corner of the page and there is no visual indication that the actual card is focused when its associated checkbox is focused. If this is identified as the best path forward for this issue I will create a core issue for those changes.

Convenience links:
[Card/selectable](https://patternfly-react-pr-7144.surge.sh/components/card#selectable)
[Card/legacy selectable](https://patternfly-react-pr-7144.surge.sh/components/card#legacy-selectable)
[Card view demo](https://patternfly-react-pr-7144.surge.sh/demos/card-view/react-demos/card-view)
[Tabs/modal tabs demo](https://patternfly-react-pr-7144.surge.sh/components/tabs/react-demos/modal-tabs/)
[Primary-detail card view demo](https://patternfly-react-pr-7144.surge.sh/demos/primary-detail/react-demos/primary-detail-card-view)

### DIFF
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -36,6 +36,10 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   isPlain?: boolean;
   /** Flag indicating if a card is expanded. Modifies the card to be expandable. */
   isExpanded?: boolean;
+  /** Flag indicating that the card should render a hidden input for a11y reasons */
+  hasHiddenInput?: boolean;
+  /** Callback that executes when the hidden input is changed */
+  onHiddenInputChange?: (labelledBy: string, event: React.FormEvent<HTMLInputElement>) => void;
 }
 
 interface CardContextProps {
@@ -67,6 +71,8 @@ export const Card: React.FunctionComponent<CardProps> = ({
   isPlain = false,
   ouiaId,
   ouiaSafe = true,
+  hasHiddenInput = false,
+  onHiddenInputChange = () => {},
   ...props
 }: CardProps) => {
   const Component = component as any;
@@ -90,6 +96,46 @@ export const Card: React.FunctionComponent<CardProps> = ({
     return '';
   };
 
+  const CardComponent = (
+    <Component
+      id={id}
+      className={css(
+        styles.card,
+        isCompact && styles.modifiers.compact,
+        isExpanded && styles.modifiers.expanded,
+        isFlat && styles.modifiers.flat,
+        isRounded && styles.modifiers.rounded,
+        isLarge && styles.modifiers.displayLg,
+        isFullHeight && styles.modifiers.fullHeight,
+        isPlain && styles.modifiers.plain,
+        getSelectableModifiers(),
+        className
+      )}
+      tabIndex={isSelectable || isSelectableRaised ? '0' : undefined}
+      {...props}
+      {...ouiaProps}
+    >
+      {children}
+    </Component>
+  );
+
+  const inputId = `${id}-input`;
+
+  const CardWithHiddenInput = (
+    <label htmlFor={inputId}>
+      <input
+        className="pf-screen-reader"
+        id={inputId}
+        type="checkbox"
+        checked={isSelected}
+        onChange={event => onHiddenInputChange(id, event)}
+        aria-labelledby={id}
+        disabled={isDisabledRaised}
+      />
+      {CardComponent}
+    </label>
+  );
+
   return (
     <CardContext.Provider
       value={{
@@ -97,26 +143,7 @@ export const Card: React.FunctionComponent<CardProps> = ({
         isExpanded
       }}
     >
-      <Component
-        id={id}
-        className={css(
-          styles.card,
-          isCompact && styles.modifiers.compact,
-          isExpanded && styles.modifiers.expanded,
-          isFlat && styles.modifiers.flat,
-          isRounded && styles.modifiers.rounded,
-          isLarge && styles.modifiers.displayLg,
-          isFullHeight && styles.modifiers.fullHeight,
-          isPlain && styles.modifiers.plain,
-          getSelectableModifiers(),
-          className
-        )}
-        tabIndex={isSelectable || isSelectableRaised ? '0' : undefined}
-        {...props}
-        {...ouiaProps}
-      >
-        {children}
-      </Component>
+      {hasHiddenInput ? CardWithHiddenInput : CardComponent}
     </CardContext.Provider>
   );
 };
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -37,11 +37,11 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   /** Flag indicating if a card is expanded. Modifies the card to be expandable. */
   isExpanded?: boolean;
   /** Flag indicating that the card should render a hidden input to make it selectable */
-  hasHiddenInput?: boolean;
-  /** Aria label to apply to the hidden input if one is rendered */
-  hiddenInputAriaLabel?: string;
-  /** Callback that executes when the hidden input is changed */
-  onHiddenInputChange?: (labelledBy: string, event: React.FormEvent<HTMLInputElement>) => void;
+  hasSelectableInput?: boolean;
+  /** Aria label to apply to the selectable input if one is rendered */
+  selectableInputAriaLabel?: string;
+  /** Callback that executes when the selectable input is changed */
+  onSelectableInputChange?: (labelledBy: string, event: React.FormEvent<HTMLInputElement>) => void;
 }
 
 interface CardContextProps {
@@ -80,9 +80,9 @@ export const Card: React.FunctionComponent<CardProps> = ({
   isPlain = false,
   ouiaId,
   ouiaSafe = true,
-  hasHiddenInput = false,
-  hiddenInputAriaLabel,
-  onHiddenInputChange = () => {},
+  hasSelectableInput = false,
+  selectableInputAriaLabel,
+  onSelectableInputChange = () => {},
   ...props
 }: CardProps) => {
   const Component = component as any;
@@ -117,18 +117,18 @@ export const Card: React.FunctionComponent<CardProps> = ({
   };
 
   React.useEffect(() => {
-    if (hiddenInputAriaLabel) {
-      setAriaProps({ 'aria-label': hiddenInputAriaLabel });
+    if (selectableInputAriaLabel) {
+      setAriaProps({ 'aria-label': selectableInputAriaLabel });
     } else if (titleId) {
       setAriaProps({ 'aria-labelledby': titleId });
-    } else if (hasHiddenInput && !containsCardTitleChildRef.current) {
+    } else if (hasSelectableInput && !containsCardTitleChildRef.current) {
       setAriaProps({});
       // eslint-disable-next-line no-console
       console.warn(
-        'If no CardTitle component is passed as a child of Card the hiddenInputAriaLabel prop must be passed'
+        'If no CardTitle component is passed as a child of Card the selectableInputAriaLabel prop must be passed'
       );
     }
-  }, [hasHiddenInput, hiddenInputAriaLabel, titleId]);
+  }, [hasSelectableInput, selectableInputAriaLabel, titleId]);
 
   return (
     <CardContext.Provider
@@ -138,14 +138,14 @@ export const Card: React.FunctionComponent<CardProps> = ({
         isExpanded
       }}
     >
-      {hasHiddenInput && (
+      {hasSelectableInput && (
         <input
           className="pf-screen-reader"
           id={`${id}-input`}
           {...ariaProps}
           type="checkbox"
           checked={isSelected}
-          onChange={event => onHiddenInputChange(id, event)}
+          onChange={event => onSelectableInputChange(id, event)}
           disabled={isDisabledRaised}
           tabIndex={-1}
         />
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -36,7 +36,7 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   isPlain?: boolean;
   /** Flag indicating if a card is expanded. Modifies the card to be expandable. */
   isExpanded?: boolean;
-  /** Flag indicating that the card should render a hidden input for a11y reasons */
+  /** Flag indicating that the card should render a hidden input to make it selectable */
   hasHiddenInput?: boolean;
   /** Aria label to apply to the hidden input if one is rendered */
   hiddenInputAriaLabel?: string;
--- a/packages/react-core/src/components/Card/Card.tsx
+++ b/packages/react-core/src/components/Card/Card.tsx
@@ -109,8 +109,11 @@ export const Card: React.FunctionComponent<CardProps> = ({
     return '';
   };
 
+  const containsCardTitleChildRef = React.useRef(false);
+
   const registerTitleId = (id: string) => {
     setTitleId(id);
+    containsCardTitleChildRef.current = !!id;
   };
 
   React.useEffect(() => {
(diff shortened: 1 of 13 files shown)


thatblindgeye:
So I think the result is comparable to the current implementation introduced in your original PR, with the added pro of being able to navigate to those inner interactive elements in the card view demo via VO. Overall I think this could be a really good alternative to using a listbox and option roles.

Some comments/things I'd like others' input on if this is the route we decide to go down:

- To get more of the same result as the original PR, at least for the demos/examples, we could wrap the set of cards in a `<section>` element with an `aria-label` to mimic VO announcing the listbo before the first option as it does in the Card view demo from your other PR.
- If possible it may be worth removing that visible checkbox from the cards in the Card view demo, but that could be a separate discussion to hash out after this issue is resolved. Mainly I'm not sure about having a visible and hidden checkbox that would both get announced to assistive technologies.
- The Tabs/modal I'm not sure would need the selectable Card variant

wise-king-sullyman:
@thatblindgeye

> To get more of the same result as the original PR, at least for the demos/examples, we could wrap the set of cards in a <section> element with an aria-label to mimic VO announcing the listbo before the first option as it does in the Card view demo from your other PR.

Yeah I feel like that would be a really good change, at least for the demos. Probably not really needed on the examples I don't think, but if you disagree with that I'm open to discussing it.

> If possible it may be worth removing that visible checkbox from the cards in the Card view demo, but that could be a separate discussion to hash out after this issue is resolved. Mainly I'm not sure about having a visible and hidden checkbox that would both get announced to assistive technologies.

Agree, having two checkboxes doesn't feel like optimal UX to me either. I defer to people with more knowledge than I have on this topic though.

> The Tabs/modal I'm not sure would need the selectable Card variant

Agree again, since the cards are never *really* selected this feels like an odd use of a selectable card. I don't think changing it would be in scope of this issue, but I could totally see us leaving it alone for now.

seanforyou23 (commented):
Nice work @wise-king-sullyman! The hidden inputs can work, but they might be a bit heavy handed in terms of implementation cost (requiring core changes) - I think some of the existing plumbing may be sufficient to pull this off and maybe we just need a nice and realistic demo/example to showcase.. I'll throw some ideas out.

Curious what you and others think about these two implementations, I'll mention the highlights of each - happy to demo it for anyone that wants

1) Actual [switch](https://github.com/seanforyou23/patternfly-react-seed/blob/selectable-cards-a11y-demo/src/app/Support/Support.tsx#L139-L142) semantics strategy. This communicates on/off states and is easy to implement. Probably makes sense if selecting the card actually toggles something on/off (like a light, or a server), but probably less useful if it's just a selection action that's needed. The AX is straight forward, and not customizable.

https://selectable-cards-a11y-demo.surge.sh/switch-role-semantics

2) [Generic](https://github.com/seanforyou23/patternfly-react-seed/blob/selectable-cards-a11y-demo/src/app/Dashboard/Dashboard.tsx#L116-L117) role semantics. This has the flexibility of being able to customize exactly what the label says, but on its own you lose the built in announcement effect that comes with changes to state in other native roles.

https://selectable-cards-a11y-demo.surge.sh/status-role-semantics

If you couple the second strategy along with providing [basic status updates](https://github.com/seanforyou23/patternfly-react-seed/blob/selectable-cards-a11y-demo/src/app/Dashboard/Dashboard.tsx#L92-L103) via [live region](https://github.com/seanforyou23/patternfly-react-seed/blob/selectable-cards-a11y-demo/src/app/Dashboard/Dashboard.tsx#L112-L114), you end up with what's seems like best of both worlds.

I think either of these approaches could work great, really just depends on what's actually supposed to happen when users select the card. I think most of the time it's just gonna be "context setting" which fits the second scenario better. I bet if we can think up a good use case for each, it would end up being a really nice showcase both approaches so people can see the impacts of doing it one way vs another. If you have a list of servers, and one strategy communicates them to be on/off vs selected/unselected - it's a very big difference in perception!

Hope it makes sense, lemme know what you think :)

nicolethoen:
@seanforyou23 these are compelling ideas we had not considered yet. I'm not sure if they buy us more than we already have?

> 1. Actual [switch](https://github.com/seanforyou23/patternfly-react-seed/blob/selectable-cards-a11y-demo/src/app/Support/Support.tsx#L139-L142) semantics strategy. This communicates on/off states and is easy to implement. Probably makes sense if selecting the card actually toggles something on/off (like a light, or a server), but probably less useful if it's just a selection action that's needed. The AX is straight forward, and not customizable.

This first example is slick to use the switch, I do like what the screen reader communicates, but I cannot get into the kabob in the first card because of the nested focusable element. The in-browser axe test is also throwing a serious level warning about it... so it's still not quite right...

> 2\. [Generic](https://github.com/seanforyou23/patternfly-react-seed/blob/selectable-cards-a11y-demo/src/app/Dashboard/Dashboard.tsx#L116-L117) role semantics. This has the flexibility of being able to customize exactly what the label says, but on its own you lose the built in announcement effect that comes with changes to state in other native roles.

This second suggestion seems less than ideal because the screen reader does not say anything when I first put focus on a selectable card about it being selectable or interactable - it only tells me I am on a group. Also, relying on aria-labels to communicate everything about the state and features of an element seems like something most product devs will have a very hard time doing.

@wise-king-sullyman I think the hidden input is pretty powerful. I'll note for @mcoker and @mcarrano in case it's interesting, that the checkbox on the selectable cards in [the carbon selectable card variant](https://www.carbondesignsystem.com/components/tile/code#live-demo) is not a real checkbox. It's a non-interactable svg that just looks like a checkbox to reflect the selected state of the card. Clicking on the checkbox does the same as clicking anywhere else in the card.

The major limitation I see to carbon's component is that they do not allow for interactable controls in the card as we often have in patternfly - such as an action dropdown. Using Austin's hidden input solution does allow for the other interactable elements in the card. But it may be worth sharing a slight concern that we are diverging from industry standards a bit it seems by allowing the whole card to be selectable while also including interactable elements within the card...

seanforyou23:
Ok pushed some updates, good recommendations - that definitely cleared things up a bit.

[Region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role#scrolling_content_areas_with_overflow_text) ends up being an even better role for the card container in case of selectability. These cards are exposed in both Form Controls and landmark rotor menus.

unselected-selected-state

>cannot get into the kabob in the first card because of the nested focusable element

I think it's accessible, but the current selectable card implementation seems to have an issue, can we open a dedicated issue for that? If the fix for that will impact markup for card, maybe we consider those changes before deciding anything final on how we can name in, in case a better/easier approach opens up.

I fixed all the axe issues except the one you mentioned about nested interactive element, yea I guess that's pretty much a deal breaker for the switch role approach - I'll remove that from the demo and edit my comment above so it's clear that approach doesn't cut it. Hope it helps, lemme know what you think or if there's anything else I can do.

thatblindgeye (commented):
With the latest updates, I think this feels like it's getting closer to the goal. Some thoughts while testing the links in the first comment on this PR:

- I like the container for the cards in the Card view demo announcing that the container is for a group of cards. I think this would be worth implementing on other examples/demos for cards, though not strictly related to this issue.
- For the single selectable card examples (Card/selectable and Card/legacy-selectable), a radio input might work better than a checkbox
- When navigating via VO, I am noticing that when VO places focus on the input for a card, it causes the card content to shift a bit, then when focus is placed on the `article` element that makes up the card, the content shifts back to its normal position (video below). This could be something that gets fixed with the core issue you mentioned opening since this is using that temporary CSS.
- For the Tabs and Primary detail card demos, I'm still unsure if these ones would benefit from having the input added onto them. The Tabs demo I'm mainly unsure if a user would actually have any way of knowing that the tab they clicked was "checked" or not, since the modal should prevent them from navigating to anything outside the modal. For the Primary detail demo I think it may just more depend on whether the intent is for users to be able to access other content on the page even when a cards' detail is expanded.

Video for the third bullet point:

https://user-images.githubusercontent.com/70952936/166558282-d22e0fb1-56bc-424d-965b-d501682597de.mov

wise-king-sullyman:
@thatblindgeye to update this PR based on our conversations from slack:

> I like the container for the cards in the Card view demo announcing that the container is for a group of cards. I think this would be worth implementing on other examples/demos for cards, though not strictly related to this issue.
    
I don't think we currently apply aria labels to the example containers from what I've seen looking for examples of that, but if I'm wrong I'd be more than update to address this.

> For the single selectable card examples (Card/selectable and Card/legacy-selectable), a radio input might work better than a checkbox

I agree, but I think adding the ability to have the input type be customizable would take a lot of thought and would be better done under a separate issue. If nobody opposes I'll spin up a new issue for that.

> When navigating via VO, I am noticing that when VO places focus on the input for a card, it causes the card content to shift a bit, then when focus is placed on the article element that makes up the card, the content shifts back to its normal position (video below). This could be something that gets fixed with the core issue you mentioned opening since this is using that temporary CSS.

I think this was just because of the temporary demo styling, it shouldn't happen now that I've removed that styling. If it is still happening please let me know as that means I misunderstood the issue.

> For the Tabs and Primary detail card demos, I'm still unsure if these ones would benefit from having the input added onto them. The Tabs demo I'm mainly unsure if a user would actually have any way of knowing that the tab they clicked was "checked" or not, since the modal should prevent them from navigating to anything outside the modal. For the Primary detail demo I think it may just more depend on whether the intent is for users to be able to access other content on the page even when a cards' detail is expanded.

I agree. But I also don't love the idea of leaving those cards as selectable but not applying the hidden input to them. I think that those demos really need a major rethink from an a11y perspective.

packages/react-core/src/components/Card/Card.tsx
```diff
@@ -36,15 +36,28 @@ export interface CardProps extends React.HTMLProps<HTMLElement>, OUIAProps {
   isPlain?: boolean;
   /** Flag indicating if a card is expanded. Modifies the card to be expandable. */
   isExpanded?: boolean;
+  /** Flag indicating that the card should render a hidden input for a11y reasons */
+  hasHiddenInput?: boolean;
+  /** Aria label to apply to the hidden input if one is rendered */
+  hiddenInputAriaLabel?: string;
+  /** Callback that executes when the hidden input is changed */
+  onHiddenInputChange?: (labelledBy: string, event: React.FormEvent<HTMLInputElement>) => void;
```
thatblindgeye, line 44: Only other nit at this point might be the prop names/descriptions. For now maybe tweaking the `hasHiddenInput` prop description to, "Flag indicating that the card should render a hidden input to make it selectable" ? Any further change may depend on further implementation of selectable cards/may be better for a breaking change release.

thatblindgeye (approved):
Agreed on your responses above. Specifically:

- Adding `aria-label` to card containers (or creating a card container component? this route may be advantageous in regards to the whole single selectable vs multi selectable card thing) could definitely be a separate discussion and/or issue
- Figuring out a radio vs checkbox implementation I'd agree can also be a separate issue and may be able to be lumped in with the above bullet point

Otherwise I do think this is definitely an improvement overall! A separate Core issue to resolve styling when the card's input receives focus will definitely be needed, and I am noticing in the Card view demo that clicking a card's kebab toggle selects the card (as you pointed out this occurs in the current implementation as well). I'm not willing to really block over those things, though.

Awesome job on all of this, by the way 💪🏼

packages/react-core/src/components/Card/Card.tsx
wise-king-sullyman: I'm also not in love with the prop names at all, if you have any suggestions for better ones please let them fly.

wise-king-sullyman: Also I really like the idea of a potential card container component, and I think that could improve a11y a *lot* while being easy for consumers to implement.

thatblindgeye: It's a bit tougher since there's already the `isSelectable` and `isSelectableRaised` props, which if anything maybe those could be used to determine whether the hidden input is rendered with some tweaking of those prop descriptions. If those are strictly meant to determine the styling only, though, maybe `selectableInput` or `hasSelectableInput`?

`selectableInputAriaLabel` and `onSelectableInputChange` *might* also work.

wise-king-sullyman: I do think we should have `isSelectable` automatically render the hidden input, but I don't think we can make that happen until the next breaking change. Changing `hiddenInput` -> `selectableInput` sounds good to me though!

jessiehuff (commented):
I had the same concerns as Eric, especially because I can imagine products not understanding what `hasHiddenInput` means and thinking it doesn't apply to them. I really like the prop suggestions though!

In testing it sounds awesome. 😎

wise-king-sullyman:
Realized that I didn't add tests for this new functionality, I've addressed that now.

tlabaj (commented):
I understand the difficulty with naming the props here.  I however think that it is not intuitive.  We may at least need some text to explain the example and highlight the intention of the props.

wise-king-sullyman:
@tlabaj do you think that this explanation text would be best:

1. added to the selectable card example
2. added to a new example specifically for demonstrating this
3. added to the a11y tab for card or
4. added to the code directly using comments

tlabaj:
> @tlabaj do you think that this explanation text would be best:
> 
> 1. added to the selectable card example
> 2. added to a new example specifically for demonstrating this
> 3. added to the a11y tab for card or
> 4. added to the code directly using comments

Option 2.

wise-king-sullyman:
@tlabaj I've added [an example](https://patternfly-react-pr-7144.surge.sh/components/card/#selectable-accessibility-highlight) to try and better explain things. Please let me know your thoughts, I'm happy to change it significantly if needed 🙂

nicolethoen (approved):
awesome work with all this, @wise-king-sullyman!!!!

packages/react-core/src/components/Card/examples/Card.md
```diff
@@ -71,6 +71,23 @@ import pfLogoSmall from './pf-logo-small.svg';
 ```ts file='./CardSelectable.tsx'
 ```
 
+### Selectable accessibility highlight
+
+This example demonstrates how the `hasSelectableInput` and `onSelectableInputChange` props enable improved accessibility for selectable cards.
```
tlabaj, line 76: @wolfeallison can you please review the copy here from lines 76-86.

wolfeallison: I think we could say "props improves accessibility for selectable cards." in line 76 and take out the "enable improved"

For line 78, I had a little trouble understanding what it was saying the first time and would switch the last sentence to " This input communicates to assistive technology users that a card is selectable, and if so, lists the current selection state." Not sure if "lists" is the right verb here or how the selection state is presented

For line 80, add a comma after "If you don't use the card title component in your selectable card,"

wise-king-sullyman: @wolfeallison for line 78, thoughts on: "This input communicates to assistive technology users that a card is selectable, and if so, it communicates the current selection state as well."?

seanforyou23 (approved):
`hasSelectableInput` prop name ties it all together for me, very intuitive :)

Nice work @wise-king-sullyman!

packages/react-core/src/components/Card/examples/Card.md
wolfeallison: @wise-king-sullyman I like that!

tlabaj (approved):

jessiehuff (approved):
LGTM! Great job on all the work on this!

kmcfaul (approved):

PR #7147, opened 2022-03-29
chore(DescriptionList): convert examples to TypeScript

### LINKED ISSUE #7143: DescriptionList - convert examples to TypeScript
DescriptionList

**Describe the feature**
The examples for the DescriptionList component need to be converted to TypeScript and functional components.

N/A

**Any other information?**
N/A

thatblindgeye:
**What**: Closes #7143 

**Additional issues**:
N/A

### DIFF
--- a/packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
+++ b/packages/react-core/src/components/DescriptionList/examples/DescriptionList.md
@@ -26,1028 +26,120 @@ import FlagIcon from '@patternfly/react-icons/dist/esm/icons/flag-icon';
 
 ### Basic
 
-```js
-import React from 'react';
-import {
-  Button,
-  DescriptionList,
-  DescriptionListTerm,
-  DescriptionListGroup,
-  DescriptionListDescription
-} from '@patternfly/react-core';
-import PlusCircleIcon from '@patternfly/react-icons/dist/esm/icons/plus-circle-icon';
-
-<DescriptionList>
-  <DescriptionListGroup>
-    <DescriptionListTerm>Name</DescriptionListTerm>
-    <DescriptionListDescription>Example</DescriptionListDescription>
-  </DescriptionListGroup>
-  <DescriptionListGroup>
-    <DescriptionListTerm>Namespace</DescriptionListTerm>
-    <DescriptionListDescription>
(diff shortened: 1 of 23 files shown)


jenny-s51 (approved):
Changes are looking great @thatblindgeye ! 

While reviewing, I noticed the horizontal description list variants extend past the width of the example container. Wondering what can be done to resolve this bug (outside the scope of this issue since it's unrelated to your changes!)... Seems like this is something that would  need to be resolved in [patternfly-org](https://github.com/patternfly/patternfly-org/blob/c7b9705a1cde6ce72229a0164a8eebd7717919b6/packages/theme-patternfly-org/components/example/example.css#L1-L4)? I'm happy to open an issue

thatblindgeye:
@jenny-s51 Do you have a specific example you're seeing this on (or a screenshot)? I'm not sure if I'm reproducing that behavior on my end (or if I am I'm just not registering it as that behavior 😆 )

jenny-s51:
Yes @thatblindgeye ! I'm seeing this on the Horizontal Responsive and Horizontal 3 Col on Lg examples:

Screen Shot 2022-04-01 at 3 46 51 PM

Screen Shot 2022-04-01 at 3 46 41 PM

jenny-s51:
After some more investigation it looks like this is only happening in Chrome -- Firefox and Safari don't display this behavior

thatblindgeye:
Ah okay, I don't believe I'm seeing it in Chrome either, though maybe due to OS/browser settings. When inspecting the page, it doesn't seem like anything is extending beyond the example container. Below is a screenshot of what I'm seeing after zooming Chrome out to 100% zoom:

Default responsive columns vs horizontal responsive columns

If it is an issue, though, I wonder if it could be something even within Core?

nicolethoen (approved):

nicolethoen:
I also didn't see that overflowing text issue in chrome.
@jenny-s51 I wonder if you were seeing some sort of style caching since there was a workspace style update to make the docs background grey and a few other things?

jenny-s51:
@thatblindgeye @nicolethoen Hmm definitely odd! Cleared my cache and updated chrome but I'm still seeing this behavior for some reason

tlabaj (approved):

PR #7159, opened 2022-03-31
fix(NumberInput): Custom increment/decrement and thresholds

### LINKED ISSUE #7069: Unable to enter a number in the number input component when using custom increment/decrement and thresholds
Was trying the example below, you are not able to enter a number between 90 - 100 in the input field. 
https://www.patternfly.org/v4/components/number-input#custom-incrementdecrement-and-thresholds

The input field will always go to 90 or 100 when you type in the field.

jenny-s51:
**What**: Closes #7069

### DIFF
--- a/packages/react-core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-core/src/components/NumberInput/NumberInput.tsx
@@ -20,6 +20,8 @@ export interface NumberInputProps extends React.HTMLProps<HTMLDivElement> {
   onMinus?: (event: React.MouseEvent, name?: string) => void;
   /** Callback for the text input changing */
   onChange?: (event: React.FormEvent<HTMLInputElement>) => void;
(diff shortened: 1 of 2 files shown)


thatblindgeye (commented):
This is looking good! I am noticing some flooring occurring in the following situations however:

- when changing the input via keyboard, pressing Tab to go to the increment button, and pressing Tab again, the input floors to the min (or max if at the max).
- when pressing the buttons to increment/decrement and then clicking outside of the component, the same behavior as above occurs.

It looks like the onBlur is being applied to the input and the buttons, so whenever the buttons lose focus it's causing that newValue to become 0 before running through the ternary when setting the state.

jenny-s51 (commented):
Awesome catch @thatblindgeye ! I've updated the component to apply `onBlur` only to the text input.

thatblindgeye (approved):

tlabaj (approved):
LGTM

PR #7163, opened 2022-04-01
fix(DataListCheck): assign checked prop to correct attribute

### LINKED ISSUE #6762: DataListCheck: decide on one of checked/isChecked and make sure the checked property is propagated on the DOM

If providing only the isChecked property to the component we get errors in console:
```
Warning: A component is changing an uncontrolled input to be controlled.
```
Passing only the 'checked' property seems to work properly, apart from the DOM not being updated, aka the check attribute never appears on the <input type=checkbox> in the DOM.

bug

Anaconda WebUI

thatblindgeye:
**What**: Closes #6762 

Based on the latest update to the DataListCheck component, I changed the attribute to which the `checked` prop was assigned. Having the `checked` attribute on the input element be _either_ of the `isChecked` or `checked` props was causing the behavior seen in the linked issue (a console warning of changing from uncontrolled to controlled, or vice versa). 

With this update, a console warning should only occur if *both* `checked` and `isChecked` props are passed in (which I've noted not to do in the prop descriptions). EDIT: this warning may only appear when testing in a local workspace and may not appear in the preview build.

 If `checked` is passed in, then it's assigned to the correct `defaultChecked` attribute (previously being assigned to the checked attribute was the reason why the checkbox wasn't updating visually, as it wasn't updating at all I believe). If neither prop is passed in, then the component should act as a normal uncontrolled checkbox.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DataList/DataListCheck.tsx
+++ b/packages/react-core/src/components/DataList/DataListCheck.tsx
@@ -9,9 +9,13 @@ export interface DataListCheckProps extends Omit<React.HTMLProps<HTMLInputElemen
   isValid?: boolean;
   /** Flag to show if the DataList checkbox is disabled */
   isDisabled?: boolean;
-  /** Flag to show if the DataList checkbox is checked */
+  /** Flag to show if the DataList checkbox is checked when it is controlled by React state.
+   * To make the DataList checkbox uncontrolled, instead use the checked prop, but do not use both.
+   */
   isChecked?: boolean;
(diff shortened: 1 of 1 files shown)


jenny-s51 (approved):
Nice updates @thatblindgeye ! Props work as expected and updated prop descriptions look great. When both `checked` and `isChecked` are used on `DataListCheck`, the console warning also makes it clear that the user should only use one of these props. 👍

tlabaj (approved):
LGTM

PR #7165, opened 2022-04-01
fix(ContextSelector): fix listener handling when using popper

### LINKED ISSUE #7104: Context Selector: onSelect via click doesn't work when appended to document body
From this comment in another issue https://github.com/patternfly/patternfly-react/issues/6824#issuecomment-1018568358 :

> Here I also append the dropdown menu to the document.body, but the first time we open the modal and toggle the dropdown menu, we cannot select the items in it. It looks like the click event happens on the body but not on the items so it will close the menu instead of selecting it. I believe this will also cause the same problem with the screen reader.

The expected behavior when clicking is to be the same when using keyboard, where clicking an option selects it rather than closing the dropdown and requiring needing to open and click the option again. 

View the codesandbox in the issue comment linked above, or:

- go to our [Context Selector](https://www.patternfly.org/v4/components/context-selector) page and update the first example so that after line 82 there is the line:

    ```javascript
    menuAppendTo={() => document.body}
    ```
- click to open the dropdown and try clicking on the `Action` item, notice how the dropdown closes without actually selecting it

Bug

michpetrov:
**What**: Closes #7104 

The issue here is that the menu closes because of ContextSelectorToggle's `mousedown` listener, since the menu is detached we need a separate reference for the detached container. 
The other problem is that with Popper `parentRef.current` is null because the menu isn't rendered and re-rendering it doesn't update the reference.

### DIFF
--- a/packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
@@ -57,16 +57,16 @@ export class ContextSelectorToggle extends React.Component<ContextSelectorToggle
 
   onDocClick = (event: any) => {
     const { isOpen, parentRef, onToggle } = this.props;
-    if (isOpen && parentRef && !parentRef.contains(event.target)) {
+    if (isOpen && parentRef?.current && !parentRef.current.contains(event.target)) {
       onToggle(null, false);
       this.toggle.current.focus();
     }
   };
 
   onEscPress = (event: any) => {
-    const { isOpen, parentRef, onToggle } = this.props;
+    const { isOpen, onToggle } = this.props;
     const keyCode = event.keyCode || event.which;
-    if (isOpen && keyCode === KEY_CODES.ESCAPE_KEY && parentRef && parentRef.contains(event.target)) {
+    if (isOpen && keyCode === KEY_CODES.ESCAPE_KEY) {
       onToggle(null, false);
       this.toggle.current.focus();
     }
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -85,6 +85,7 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
   }
 
   parentRef: React.RefObject<HTMLDivElement> = React.createRef();
+  popperRef: React.RefObject<HTMLDivElement> = React.createRef();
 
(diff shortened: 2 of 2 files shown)


packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -57,7 +57,7 @@ export class ContextSelectorToggle extends React.Component<ContextSelectorToggle
 
   onDocClick = (event: any) => {
     const { isOpen, parentRef, onToggle } = this.props;
-    if (isOpen && parentRef && !parentRef.contains(event.target)) {
+    if (isOpen && parentRef.current && !parentRef.current.contains(event.target)) {
```
nicolethoen, line 60: I think you need to check that parentRef is defined here. that's what's causing the test to fail

nicolethoen (approved):

michpetrov:
Oh yeah, missed that one. Fixed.

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -66,7 +66,7 @@ export class ContextSelectorToggle extends React.Component<ContextSelectorToggle
   onEscPress = (event: any) => {
     const { isOpen, parentRef, onToggle } = this.props;
     const keyCode = event.keyCode || event.which;
-    if (isOpen && keyCode === KEY_CODES.ESCAPE_KEY && parentRef && parentRef.contains(event.target)) {
+    if (isOpen && keyCode === KEY_CODES.ESCAPE_KEY && parentRef?.current && !parentRef.current.contains(event.target)) {
```
thatblindgeye, line 69:
```suggestion
    if (isOpen && keyCode === KEY_CODES.ESCAPE_KEY) {
```
This will allow the toggle to be closed with the Escape key.

thatblindgeye (changes requested):
Thanks for making the previous change! It looks like I was mistaken in the previous suggestion, as pressing Escape still doesn't close the toggle. Removing the `parentRef` parts of the if statement on `onEscPress` will give us the intended behavior. I included the suggested change below that'll work.

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
michpetrov: right, that makes sense, I don't see why the reference was required in the first place

thatblindgeye (approved):

kmcfaul (approved):

PR #7169, opened 2022-04-01
chore(TextInputGroup): update testing suite to new standards

### LINKED ISSUE #6929: Convert Text Input Group tests from Enzyme to React Testing Library
Related to #6794 and #6928

As part of the documentation development for our use of React Testing Library going forward, and specifically our need to convert Enzyme tests to RTL, this issue focuses on the conversion of an existing components test suite to use RTL.

Rather than strictly attempting to replicate the current testing approach, this conversion should ensure adherence to the "RTL way" of testing with an emphasis on testing in a way that mimics real user interaction.

wise-king-sullyman:
**What**: Closes #6929 

**Additional issues**:


packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  it('renders the input with class pf-c-text-input-group__text-input', () => {
+    render(<TextInputGroupMain>Test</TextInputGroupMain>);
+
+    const input = screen.getByRole('textbox');
+
+    expect(input).toHaveClass('pf-c-text-input-group__text-input');
+  });
+
+  it('renders given input icon props', () => {
```
jeffpuzzo, line 65: I think this is more-so "renders the icon" (or something like this) rather than this which reads to me essentially, "the text input group still renders when passing an icon prop". Maybe also mention that it is expected to be hidden.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  it('passes the aria-label prop to the input', () => {
+    render(<TextInputGroupMain aria-label="Test label">Test</TextInputGroupMain>);
+
+    const input = screen.getByRole('textbox');
+
+    expect(input).toHaveAccessibleName('Test label');
+  });
+
+  it('passes the value prop to the input', () => {
```
jeffpuzzo, line 79: Can we not instead visibly see the value passed to the input or is that hidden from users?

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+    render(<TextInputGroupMain value="value text">Test</TextInputGroupMain>);
+
+    const input = screen.getByRole('textbox');
+
+    expect(input).toHaveAttribute('value', 'value text');
+  });
+
+  it('passes the placeholder prop to the input', () => {
```
jeffpuzzo, line 87: same as my other comment. If we can verify visibility of the placeholder, that is more meaningful than verifying an attribute exists and makes it so we don't have to test the attribute value IMO.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+    render(<TextInputGroupMain type="search">Test</TextInputGroupMain>);
+
+    const textInput = screen.queryByRole('textbox');
+    const searchInput = screen.getByRole('searchbox');
+
+    expect(textInput).not.toBeInTheDocument();
+    expect(searchInput).toBeVisible();
+  });
+
+  it('does not disable the input when TextInputGroupContext isDisabled is false', () => {
```
jeffpuzzo, line 113: I don't think this is necessary since there are tests verifying this context is passed appropriately in the TextInputGroup suite. Same with the other TextInputGroupContext below.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+  });
+
+  it('calls the onBlur callback when the input loses focus', () => {
+    const onBlurMock = jest.fn();
+
+    render(<TextInputGroupMain onBlur={onBlurMock}>Test</TextInputGroupMain>);
+
+    const input = screen.getByRole('textbox');
+    userEvent.click(input);
+    userEvent.click(document.body);
```
jeffpuzzo, line 268: This is a good idea. I haven't thought to focus out by simply clicking the document body before.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+    expect(screen.getByRole('img', { hidden: true })).toBeInTheDocument();
+  });
+
+  it('passes the aria-label prop to the input', () => {
+    render(<TextInputGroupMain aria-label="Test label">Test</TextInputGroupMain>);
+
+    const input = screen.getByRole('textbox');
+
+    expect(input).toHaveAccessibleName('Test label');
```
jeffpuzzo, line 76: Very cool. Haven't used this before.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroup.test.tsx
```diff
@@ -1,166 +1,87 @@
...
-    const view = render(
-      <TextInputGroup className="custom-class" id="test-id">
-        <TextInputGroupMain>Foo</TextInputGroupMain>
-        <TextInputGroupUtilities />
-      </TextInputGroup>
-    );
-    expect(view.container).toMatchSnapshot();
-  });
+  it('renders children', () => {
+    render(<TextInputGroup>Test</TextInputGroup>);
```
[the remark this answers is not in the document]
wise-king-sullyman: Good point! What kind of assertion do you think would be appropriate for checking rendering without children though?

Part of me leans towards asserting that `Test` isn't visible so that we know it doesn't otherwise appear in the component in the test for children, but that doesn't really test that it renders as we expect without children. The other part of me thinks that maybe this is where the snapshot test for the component should be used.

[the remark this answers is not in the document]
jeffpuzzo: In this case I might resort to a `data-testid` which gets applied to the parent `div`. If we `queryByTestId` and verify it is null, then that should cover us verifying the entirety of the component is not rendered.

A snapshot could be used but would be less specific IMO.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
wise-king-sullyman: You can't grab it or the placeholder with byRole/byText, but after some further research I did find that RTL provides a matcher specifically for input values/placeholders! Great callout!

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
wise-king-sullyman: I disagree, the tests concerning the context in TextInputGroup are ensuring that TextInputGroup is "sending the right messages", whereas the tests here are ensuring that TextInputGroupMain is properly getting data from the context and acting in the way we expect.

thatblindgeye (commented):
Other than Jeff's comments, the only thing I could think of when looking this over was whether the test that explicitly checks for the default class should be included when the snapshots already have that class to test against.

As you mentioned in our conversation, it was decided in an RTL meeting that testing explicitly for classes should be done. I can see merit in doing so, especially if we don't want snapshots necessarily being used to test that attributes have a specific value, so I'm not willing to block over that. Mentioning it in case anyone else may have an opinion either way after seeing this in action.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroup.test.tsx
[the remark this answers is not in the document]
wise-king-sullyman: I'm a bit confused, shouldn't querying by the testid select the empty div with the associated TextInputGroup styling?

I could add a testid as you suggest and verify that it renders that div without children though, which I think would accomplish the overall goal here if I've understood you right.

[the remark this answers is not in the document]
jeffpuzzo: Yeah that's what I meant, sorry haha. I forgot when looking at your comment this morning that it was meant to render without children. What I said is if the component was meant to not be shown without children.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroupMain.test.tsx
jeffpuzzo: Ok, I can see there's a chance this component doesn't read in the `isDisabled` and apply it to the input.

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroup.test.tsx
```diff
@@ -1,166 +1,99 @@
...
-    expect(view.container).toMatchSnapshot();
-  });
+    expect(screen.getByTestId('TextInputGroup')).toBeVisible();
 
-  test('passes isDisabled=false to children via a context when isDisabled prop is not passed', () => {
-    const TestComponent: React.FunctionComponent = () => {
-      const context = React.useContext(TextInputGroupContext);
+    /** the below expectation is to verify that that the component isn't always rendering with the text used to verify
+    child rendering in the next test */
+    expect(screen.queryByText('Test')).not.toBeInTheDocument();
```
thatblindgeye, line 17: For this, would checking that the `children` length is 0 (or something similar) be viable for this test instead?

wise-king-sullyman: I can get why you would want that, since this expectation only makes sense with the context of the following test, but I feel like checking the length of the `children` property isn't a very "RTL" approach. @jeffpuzzo thoughts?

kmcfaul (approved):

packages/react-core/src/components/TextInputGroup/__tests__/TextInputGroup.test.tsx
jeffpuzzo: I don't think lines 15 through 17 should be included at all since you're already verifying the parent component is visible without children which is the objective of the test, correct?

wise-king-sullyman: I partially agree?

15-17 aren't strictly necessary to verify that it renders without children, *but* in the same way that the next test verifies chilldren are rendered (by testing that when "Test" is passed as a child it is rendered to the page), this ensures that the inverse is true (by testing that when "Test" isn't passed it isn't rendered to the page). Essentially it protects from flaws in the test suite, like the component potentially always rendering "Test".

I figure that it goes along with the other aspect of testing that it renders without children for the same reason that checking for "Test" is valid for checking that children are rendered. If you disagree though and think it's either unnecessary or should be a separate test I'm not super committed to it.

jeffpuzzo: I do disagree. We're specifically rendering this component with no children (aka a self-closing tag). All we care about is that the component still renders.

Testing that text we're purposefully not including in `render` doesn't exist doesn't make sense to me, as much as it might appear more-so as an inverse of the other test. If someone updates the test to include "Test" as a child after the fact for some reason, then we have bigger problems.

thatblindgeye: I think my concern is that what it's actually testing for doesn't exactly fit with what the intention is. I'd say comparing to a snapshot of the component rendered without children might be better, but that's less an option if we only want the single snapshot.

wise-king-sullyman: 👍 I can understand that/those viewpoints. I'll update shortly.

jeffpuzzo (approved):
Looks great!

thatblindgeye (approved):

tlabaj (approved):

tlabaj:
you just need to resolve conflicts and we can merge.

PR #7172, opened 2022-04-02
fix(a11y): various aria related updates

### LINKED ISSUE #7171: lack of appropriate aria-roles and aria-current
- Add aria-label prop to each progress step to documentation
- Add aria-current=step to current progress step
- Aria-current=location for active jump links
- Tabs - use tab role, and tablist role (we were already using tabpanel role)

nicolethoen:
**What**: Closes #7171

Notes of changes:

- Add `aria-label` prop to ProgressStep documentation
- Add `aria-current=step` to current progress step
- `Aria-current=location` for active jump links
- Tabs - use `tab` role, and `tablist` role (we were already using `tabpanel` role)


thatblindgeye (changes requested):
Looks good! Really like the addition of the `aria-current` attribute. 

Since `aria-label` has been used elsewhere in the PR, one suggestion I have is passing in something to the `aria-label` prop in the Tabs example(s). Main reason is that right now via VO there's no indication that there is a group of Tab components like there used to be (in the current implementation VO announces "list 6 items" before going to the first Tab item, whereas in this preview build VO goes right to that first tab item). Do you have any other opinion on this?

kmcfaul (approved):

jessiehuff (approved):
These all sound great in testing! 😄

PR #7173, opened 2022-04-04
docs(Dropdown): expose menuAppendTo in Dropdown prop docs

### LINKED ISSUE #6985: Dropdown - menuAppendTo is missing from docs
In the dropdown component docs, we have an [append menu document body example](https://www.patternfly.org/v4/components/dropdown#append-menu-document-body) that uses `menuAppendTo`,  but that prop is missing from the docs. I checked the select component docs and it's listed there. I'm not sure what all components support that prop but it might be good to double check the docs for those components, too.

jenny-s51:
**What**: Closes #6985

### DIFF
--- a/packages/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
+++ b/packages/react-core/src/components/ApplicationLauncher/ApplicationLauncher.tsx
@@ -9,10 +9,9 @@ import { ApplicationLauncherGroup } from './ApplicationLauncherGroup';
 import { ApplicationLauncherSeparator } from './ApplicationLauncherSeparator';
 import { ApplicationLauncherItem } from './ApplicationLauncherItem';
 import { ApplicationLauncherContext } from './ApplicationLauncherContext';
-import { ToggleMenuBaseProps } from '../../helpers/Popper/Popper';
 import { createRenderableFavorites, extendItemsWithFavorite } from '../../helpers/favorites';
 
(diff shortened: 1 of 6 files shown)


thatblindgeye (commented):
Looks like the `menuAppendTo` prop is now exposed in the prop docs for the component, but the description is blank and its Type info is "No type info".

Not exactly sure since the DropdowProps is extending the ToggleMenuBaseProps (which has the menuAppendTo prop), but would it be better to have the `menuAppendTo` declared inside the DropdownProps instead to render a proper description and type info in the docs?

jenny-s51:
Thank you @thatblindgeye ! Yes, the `menuAppendTo` prop is inherited from `ToggleMenuBaseProps`. It looks like when a prop is not explicitly declared in the component, but inherited from somewhere else, the description defaults to "No type info" [due to the way tsDocgen is currently implemented in org](https://github.com/patternfly/patternfly-org/blob/7550dc5e9046794034ee7997acd157b493c19bae/packages/theme-patternfly-org/scripts/tsDocgen.js). I've added a new optional `menuAppendTo` prop and description to each of the components that use it so the both the prop and description are now exposed in the doc!

nicolethoen (approved):

thatblindgeye (approved):
Ah thanks for the clarification on what was causing that issue. Latest update looks good!

kmcfaul (approved):

PR #7176, opened 2022-04-04
feat(Brand): allow picture with child elements and breakpoints, add example/test

### LINKED ISSUE #5637: Brand updates to allow for different images to be used via media queries
https://github.com/patternfly/patternfly/pull/3922

Part of the masthead updates

* Allows `.pf-c-brand` to be used with a `<picture>` element via `.pf-m-picture`
* Ability to define a base image to be used, then different `<source>` images attached to media queries for art direction based on viewport size.
* Can now define `--pf-c-brand--Width{-on-[breakpoint]}` and `--pf-c-brand--Height{-on-[breakpoint]}` on the brand component to control the width/height based on breakpoints. 

It might make sense to allow the user to create an object with breakpoints as keys and a width, height, and image src per breakpoint?

kmcfaul:
**What**: Closes #5637

When `children` is passed to `Brand`, `Brand` will turn into a container `picture` element.
`widths` and `heights` breakpoint properties are available only when `children` is passed (indicating it is a picture element).

~~`alt` is no longer a required property. I'm not sure about this change, because it is good to be required when `Brand` is an `img` but doesn't make a lot of sense when `Brand` is a `picture`. Should I leave it as required knowing it will not affect `picture` Brands?~~


packages/react-core/src/components/Brand/Brand.tsx
```diff
@@ -1,18 +1,70 @@
...
+      ...setBreakpointCssVars(heights, '--pf-c-brand--Height')
+    };
+  }
+
+  return (
+    /** the brand component currently contains no styling the 'pf-c-brand' string will be used for the className */
+    children !== undefined ? (
+      <picture className={css(styles.brand, styles.modifiers.picture, className)} style={style} {...props}>
+        {children}
+      </picture>
```
thatblindgeye, line 64: What do you think about having only the `<source>` elements passed in as children, and instead writing in the `` ourselves? Something along the lines of:

```suggestion
    children !== undefined ? (
      <picture className={css(styles.brand, styles.modifiers.picture, className)} style={style} {...props}>
        {children}
        <img src={src} alt={alt} />
      </picture>
```

This would allow us to keep the `alt` prop as required, which would apply to any image within the `<picture>` element. This might require tweaking of verbiage for some prop descriptions (like `src` could include "... When children are passed in, this specifies the fallback URL for the Brand." or something).

Would something along the lines of the above suggestion cause any issues do you think?

kmcfaul: I don't think it would cause any issues, and I like that this would keep `alt` as a required property. I'll update the PR for now and can always roll it back if we find an issue with it.

packages/react-core/src/components/Brand/Brand.tsx
```diff
@@ -1,18 +1,71 @@
...
+  children?: React.ReactNode;
+  /** Additional classes added to the either type of Brand. */
   className?: string;
-  /** Attribute that specifies the URL of the image for the Brand. */
+  /** Attribute that specifies the URL of a <img> Brand. For a <picture> Brand this specifies the fallback <img> URL. */
   src?: string;
-  /** Attribute that specifies the alt text of the image for the Brand. */
+  /** Attribute that specifies the alt text of a <img> Brand. For a <picture> Brand this specifies the fallback <img> alt text. */
   alt: string;
+  /** Widths are various breakpoints for a <picture> Brand. */
```
thatblindgeye, line 15: Super nitpick: "Widths at" to match the description for the heights prop

packages/react-core/src/components/Brand/examples/Brand.md
```diff
@@ -6,10 +6,23 @@ propComponents: ['Brand']
...
 ## Examples
 
 ### Basic
 
 ```ts file="./BrandBasic.tsx"
 ```
+
+### Responsive
+
+Passing `<source>` and `<img>` elements as children to `Brand` will change the component from an `<img>` to a `<picture>` element. In this form, breakpoint modifiers for width and height may be passed to `Brand`, and the `src` and `alt` properties are not applicable (though should be passed to any child `<img>` elements).
```
thatblindgeye, line 25: This would just need to be updated to match the recent update.

thatblindgeye (changes requested):
Looks good with the latest update! Just have one change request + nitpick.

packages/react-core/src/components/Brand/Brand.tsx
kmcfaul: Whoops haha

thatblindgeye (approved):
Looks good! Really like how the component works with this update.

mcoker (approved):
LGTM!

nicolethoen (commented):
It seems to me that the alt text will always be used, right? Why would we not make it required?

kmcfaul:
@nicolethoen It should still be required. I had updated the PR based on some feedback to make the  internal so the opening comment is out of date. I'll update that.

nicolethoen (approved):

PR #7183, opened 2022-04-06
docs(Tabs): add tables and tabs auto width tabs demo

### LINKED ISSUE #6860: Tables and tabs with auto width tabs

wise-king-sullyman:
**What**: Closes #6860

This PR also adds a prop to apply the `noBorderBottom` styling as needed by the associated demo, and a test to verify that prop works as expected.

**Additional issues**:

Convenience link directly to the added demo: https://patternfly-react-pr-7183.surge.sh/components/tabs/react-demos/tables-and-tabs-auto-width-tabs/

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -42,7 +42,7 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   /** Enables vertical tab styling */
   isVertical?: boolean;
   /** Enables no border bottom tab styling */
-  hasNoBorderBottom?: boolean;
+  hasBorderBottom?: boolean;
   /** Aria-label for the left scroll button */
   leftScrollAriaLabel?: string;
   /** Aria-label for the right scroll button */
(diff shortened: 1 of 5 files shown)


mcarrano (approved):
LGTM. Thanks @wise-king-sullyman !

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -41,6 +41,8 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   isBox?: boolean;
   /** Enables vertical tab styling */
   isVertical?: boolean;
+  /** Enables no border bottom tab styling */
+  hasNoBorderBottom?: boolean;
```
kmcfaul, line 45: Not a blocker but I wonder whether the prop should be called `hasBorderBottom` with a false value instead of `hasNoBorderBottom` with a true value? I know the css prop is called noBorderBottom but I feel like I don't see "no{X}" props as much.

kmcfaul (approved):
LGTM overall just a minor question about prop name which I'd be happy with either way.

packages/react-core/src/components/Tabs/Tabs.tsx
nicolethoen: I think making it feel more intuitive to the consumer might be better than making it match the core class name. Im in support of avoiding the double negative.

nicolethoen:
Could you add a sentence or two in the docs to clarify what this demo is highlighting?

kmcfaul (approved):

nicolethoen (approved):

PR #7190, opened 2022-04-07
feat(Progress stepper): Updated to use render prop for popover

### LINKED ISSUE #6658: ProgressStepper - investigate help popover using a ref callback
This would eliminate the need to set an additional reference property on the `ProgressStep` just for the `Popover`.

tlabaj:
**What**: Closes #6658

### DIFF
--- a/packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
+++ b/packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
@@ -26,7 +26,7 @@ export interface ProgressStepProps
    * step's icon, including the variant and the completed status. */
   'aria-label'?: string;
   /** Forwards the step ref to rendered function.  Use this prop to add a popover to the step.*/
-  popoverRender?: ({ stepRef }: { stepRef: React.RefObject<any> }) => React.ReactNode;
+  popoverRender?: (stepRef: React.RefObject<any>) => React.ReactNode;
 }
 
 const variantIcons = {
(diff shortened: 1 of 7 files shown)


thatblindgeye (approved):
Looks good! Other than the below comment, I opened issue #7200 after noticing something when testing the preview build (unrelated to this PR/issue)

packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
```diff
@@ -22,10 +22,8 @@ export interface ProgressStepProps
   description?: string;
   /** ID of the title of the progress step. */
   titleId?: string;
-  /** Popover for a progress step. */
-  popover?: React.ReactNode;
-  /** @hide Forwarded reference to title container */
-  innerRef?: React.Ref<any>;
+  /** Forwards the stepRef to rendered function.  Use this prop to add a popover to the step.*/
+  render?: ({ stepRef }: { stepRef: React.RefObject<any> }) => React.ReactNode;
```
wise-king-sullyman, line 26: Genuine question: why does `stepRef` need to be in an object? I pulled this branch locally and have `render` taking `stepRef` directly, and I *think* everything's still working properly.

nicolethoen:
I'm noticing that the keyboard accessibility is still missing from the example with a popover. Did you have a chance to peek at that?

packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
tlabaj: I was getting a lint error when I was using the innerRef we had before. I removed it so I no longer need an object.

packages/react-integration/demo-app-ts/src/components/demos/ProgressStepperDemo/ProgressStepperDemo.tsx
```diff
@@ -0,0 +1,72 @@
...
+          )}
+        >
+          First step
+        </ProgressStep>
+        <ProgressStep
+          variant="danger"
+          id="popover-step2"
+          titleId="popover-step2-title"
+          aria-label="completed step with popover, step with danger"
+          popoveRender={({ stepRef }) => (
```
kmcfaul, line 34: Missing 'r' here is why the build is failing atm.

wise-king-sullyman (approved):

kmcfaul (approved):

PR #7192, opened 2022-04-07
fix(Pagination): make items per page text in toggle clickable

### LINKED ISSUE #7123: [Pagination] - update the items per page toggle so text is clickable
Want to allow the text as well as the down caret to reveal the items per page options. Core PR: https://github.com/patternfly/patternfly/pull/4723

@mcoker anything to add?

thatblindgeye:
**What**: Closes #7123 

Currently this update changes the outer element of the toggle from a `div` to a `button`, and placing class names where appropriate. I also added in the `aria-haspopup` prop to match the current Core workspace.

This can also be updated to be more of an opt-in where a prop can be passed in to decide whether the entire toggle text should be clickable (button) or whether it should behave as it currently does in React where only the icon is clickable (div).

[Pagination preview build](https://patternfly-react-pr-7192.surge.sh/components/pagination) for convenience.

**Additional issues**:


nicolethoen (commented):
@tlabaj just checking whether you think changing the OptionsMenu with a Dropdown in the Pagination constitutes a breaking change? Since our tests had to be updated, I wonder if products' tests might end up in a similar situation?

tlabaj:
Yes, this would constitute a breaking change.  Products may also be using the selector to apply custom styling.

tlabaj:
Yes, this would constitute a breaking change.  Products may also be using the selector to apply custom styling.

packages/react-core/src/components/Pagination/OptionsToggle.tsx
```diff
@@ -59,41 +59,37 @@ export const OptionsToggle: React.FunctionComponent<OptionsToggleProps> = ({
...
-      styles.optionsMenuToggle,
-      isDisabled && styles.modifiers.disabled,
-      styles.modifiers.plain,
-      styles.modifiers.text
-    )}
-  >
+  <>
     {showToggle && (
-      <React.Fragment>
-        <span className={css(styles.optionsMenuToggleText)}>
```
tlabaj, line 72: Products could be applying custom styles here .  This would be a breaking change.

tlabaj (changes requested):

packages/react-core/src/components/Pagination/OptionsToggle.tsx
```diff
@@ -37,12 +37,14 @@ export interface OptionsToggleProps extends React.HTMLProps<HTMLDivElement> {
   onEnter?: () => void;
   /** Label for the English word "of" */
   ofWord?: string;
+  /** Component to be used for wrapping the toggle contents. */
+  component?: 'div' | 'button';
```
tlabaj, line 41: I think we should specify here that `button` should be used to make entire toggle text clickable.

packages/react-core/src/components/Pagination/Pagination.tsx
```diff
@@ -141,6 +141,8 @@ export interface PaginationProps extends React.HTMLProps<HTMLDivElement>, OUIAPr
   onPageInput?: (event: React.SyntheticEvent<HTMLButtonElement>, page: number) => void;
   /** Function called when user selects number of items per page. */
   onPerPageSelect?: OnPerPageSelect;
+  /** Component to be used for wrapping the toggle contents. */
+  component?: 'div' | 'button';
```
mcoker, line 145: I think this would be called something more specific to the menu, like `perPageComponent` or something like that. Otherwise it seems like it applies to the `<Pagination>`'s `<div class="pf-c-pagination">` element. It would also add a namespace conflict if we created a `component` prop for `div.pf-c-pagination`

packages/react-core/src/components/Pagination/PaginationOptionsMenu.tsx
```diff
@@ -26,7 +26,7 @@ export interface PaginationOptionsMenuProps extends React.HTMLProps<HTMLDivEleme
   perPageSuffix?: string;
   /** The type or title of the items being paginated */
   itemsTitle?: string;
-  /** The text to be displayed on the Options Toggle */
+  /** Accessible label for the Options Toggle */
```
mcoker, line 29: I just spot checked a handful of other react component props and we use the "accessible label" wording to describe the function of a `[whatever]AriaLabel` prop. This may be OK, just wanted to mention it.

mcoker (changes requested):
LGTM! Tho I think we should change the `component` prop on `<Pagination>` to be something that refers to the menu. 

Also assuming we continue to bundle this menu with the pagination component, and we keep the options menu over using a menu-toggle/menu pair, should we make a note in a breaking change release to just use the `<OptionsMenu>` component? It looks like we're reconstructing parts of it in this component. I could be reading it wrong though.

packages/react-core/src/components/Pagination/PaginationOptionsMenu.tsx
thatblindgeye: That's a fair point. My main reason for changing the description was due to "text to be displayed" didn't seem accurate either, as it seemed like the prop was only being used to add an `aria-label` to that dropdown toggle.

mcoker (approved):
LGTM! 🥳

tlabaj (approved):
LGTM

PR #7198, opened 2022-04-08
feat(Select): add onBlur callback to select toggle

### LINKED ISSUE #6882: Expose the onBlur event with the Select component
**Description**
I have a question about using the Select component with Formik and Yup validation.

We have been using the Formik touched property to only show invalid state in forms after the user has blurred the control at least once. But I can’t see how to use Formik’s onBlur handler with the Select component.

(As a workaround, I tried inserting a wrapper function which called Formik’s onBlur on the first change, but the event fires on the individual checkbox, not the Select itself.)

Before I go further into making a workaround, I wanted to see if others are doing this more elegantly.**

**Repro steps**
Hard to make a reduced test case, because it involves other 3rd-party libs, and ACS codebase is not open-source yet.
I would be glad to do a screen-share

**Is this a bug or enhancement?**
Enhancement request, either to expose the `onBlur` event with current multi-checkbox Select component, 
or add that as a requirement for the upcoming new Select component.

Red Hat Advanced Cluster Security (RHACS)
this is ongoing form work

kmcfaul:
**What**: Closes #6882

Adds `onBlur` property that is passed to the SelectToggle. Whenever focus is removed from the toggle the blur will trigger.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -123,6 +123,8 @@ export interface SelectProps
   ) => void;
   /** Callback for toggle button behavior */
   onToggle: (isExpanded: boolean, event: React.MouseEvent | React.ChangeEvent | React.KeyboardEvent | Event) => void;
+  /** Callback for toggle blur */
+  onBlur?: (event?: any) => void;
   /** Callback for typeahead clear button */
   onClear?: (event: React.MouseEvent) => void;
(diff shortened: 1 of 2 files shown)


kmcfaul:
@[vjwilson](https://github.com/vjwilson) Let me know if this won't work for your use case, I think I've added the callback to the right spot.

nicolethoen (approved):

tlabaj (approved):
LGTM

Issue #7174, opened 2022-04-04
Catalog view - update css to allow for dark theme support


nicolethoen:
Catalog View extension was updated from PF3 to PF4 colors in https://github.com/patternfly/patternfly-react/pull/5987 but a few issues remain:

1. Colors reference sass variables directly but need to be updated to CSS variables to respond to dark theme
2. The styles that ship with the package are not used in the examples, instead each example uses a custom stylesheet that includes all the styling for that example and these still reference PF3 colors
3. Additional upgrades can be made to adopt PF4, such as using PF4 variables for the spacing & borders and ultimately adopting PF4 components where possible

This issue is to address item 1 above, giving a quick path to enable dark theme within this package to enable products consuming it not to be blocked on dark theme adoption.

PR #7215, opened 2022-04-12
fix(catalog-view): updated colors to use dark theme friendly css vars

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/7174


nicolethoen (approved):
Makes sense to me. 
@evwilkin Should we open follow up issues for tracking purposes to remove the need to set css variables in the markdown examples? Maybe add props to apply modifiers or something?

tlabaj (approved):

Issue #7142, opened 2022-03-29
Core - Add support for React 18


jonkoops:
React 18 has been [released](https://reactjs.org/blog/2022/03/29/react-v18.html), code will have to be adjusted to that it is officially supported.

### Checklist
- [x] https://github.com/patternfly/patternfly-react/issues/6919
- [x] Update `@types/react` and `@types/react-dom` (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210).
- [x] #7195
- [x] #7216
- [x] #7217
- [x] #7218
- [x] #7282
- [x] #7363
- [x] #7402
- [x] Add `PropsWithChildren` types to components with children
- [ ] Apply migrations from [upgrade guide](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html).
  - [x] #8027
  - [ ] https://github.com/patternfly/patternfly-org/pull/3203
- [ ]  Check for incompatible dependencies and upgrade if necessary.
  - [x] #7706
  - [x] #7826
  - [x] #8577
- [x] #7303
- [x] #7821
- [x] #7827

tlabaj:
@jonkoops Thanks for opening this issue.  We are in the process of investigating impact of supporting  React 18 available in PF.  Our initial investigation found that enzyme adapter does not support React 18. Here is the PRhttps://github.com/patternfly/patternfly-react/pull/6899
We are currently moving our testing to react-testng-library to remove that dependency.  We have several PR already in the works to do so.  
When is your team looking to upgrade?

jonkoops:
We're looking to update as soon as possible, but we also have dependencies that are not yet ready for React 18.

I'd like to help out  with the upgrade so I've opened up a PR #7146 to do some ground work. It's open to collaborators so everyone can push commits to it. If you'd like we can coordinate our efforts there.

tlabaj:
@jonkoops that is great that you would to collaborate.  I am going to copy the "to do" list from your PR to the issue for tracking.

mohd-akram:
Any update on this?

jonkoops:
@nicolethoen could you re-open this issue? Looks like a closing keyword was used in #7706 so this was closed accidentally.

siwinski:
Is there an ETA for React 18 support?

nicolethoen:
React 18 will be supported officially with the next major release of PF, which is tentatively planned to be the first release of 2023.

[...]

PR #7216, opened 2022-04-12
Add correct types to instances of useCallback

jonkoops:
Adds correct type information to instances of `useCallback` to prevent implicit `any` errrors. This type has [been tightened](https://github.com/eps1lon/types-react-codemod#usecallback-implicit-any) in React 18 so this is needed to land #7142.


nicolethoen (approved):

nicolethoen:
@jeff-phillips-18 I tagged you for review here since there are some topology updates I'd love you to sanity check :)

jeff-phillips-18 (approved):

PR #7222, opened 2022-04-12
Fix for drop target highlight, group label position on drag

### LINKED ISSUE #7220: Bug - Topology - Nodes are not given the highlight modifier when they are a drop target
**Describe the problem**
When a node is a valid drop target for the current drag operation, it should be given the `pf-m-highlight` modifier.

Drag an edge that is droppable on some nodes

**Expected behavior**
The valid drop target nodes have the `pf-m-highlight` modifier.

Yes

OpenShift Console, 4.11

### LINKED ISSUE #7221: Bug - Topology - Group labels move during a regroup operation of a child node
**Describe the problem**
When dragging a node that belongs to a group, if you change to a regroup operation, the node moves outside the group correctly, but if you move it lower than the lowest group node, the group label moves to the dragged group.

Do a regroup drag of a node within a group and move it below the group.

**Expected behavior**
The group remains the same, the label remains positioned as if the node was in its original location.

Yes

**Screenshots**
bad-group

OpenShift Console, 4.11

jeff-phillips-18:
**What**: 
Fixes #7220 
Fixes #7221 

Adds appropriate modifiers on groups and nodes based on the current drag operation.
Don't update group label when a drag operation is in process.

/cc @christianvogt

### DIFF
--- a/packages/react-topology/src/components/groups/DefaultGroupExpanded.tsx
+++ b/packages/react-topology/src/components/groups/DefaultGroupExpanded.tsx
@@ -100,7 +100,6 @@ const DefaultGroupExpanded: React.FunctionComponent<DefaultGroupExpandedProps> =
 }) => {
   const [hovered, hoverRef] = useHover();
   const [labelHover, labelHoverRef] = useHover();
-  const [locations, setLocations] = React.useState<{ labelLocation: PointWithSize; path: string }>();
   const dragLabelRef = useDragNode()[1];
   const refs = useCombineRefs<SVGPathElement>(hoverRef, dragNodeRef);
   const isHover = hover !== undefined ? hover : hovered;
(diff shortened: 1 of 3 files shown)


packages/react-topology/src/components/groups/DefaultGroupExpanded.tsx
```diff
@@ -121,6 +121,9 @@ const DefaultGroupExpanded: React.FunctionComponent<DefaultGroupExpandedProps> =
   const hullPadding = React.useCallback((point: PointWithSize | PointTuple) => (point[2] || 0) + padding, [padding]);
 
   const points: (PointWithSize | PointTuple)[] = React.useMemo(() => {
+    if (droppable) {
+      return prevPoints.current;
+    }
```
christianvogt, line 126: If `prevPoints.current` hasn't been defined, you should still run this function completely.

nicolethoen (approved):

kmcfaul (approved):

christianvogt (approved):

PR #7244, opened 2022-04-13
chore(DragAndDrop): convert examples to TS 

### LINKED ISSUE #7237: Drag and drop: convert examples to TypeScript
Is this a new component or an extension of an existing one?
Drag and drop

Describe the feature
The examples for the Drag and drop component need to be converted to TypeScript and functional components.

Are there visuals for this feature? If applicable, please include examples for each state and for varying widths
N/A

Any other information?
N/A

jenny-s51:
**What**: Closes #7237

### DIFF
--- a/packages/react-core/src/components/DragDrop/examples/DragDrop.md
+++ b/packages/react-core/src/components/DragDrop/examples/DragDrop.md
@@ -5,7 +5,7 @@ propComponents: [DragDrop, Draggable, Droppable, DraggableItemPosition]
 beta: true
 ---
 
-You can use `DragDrop` to move items in or between lists. `DragDrop`s should contain `Droppable`s which contain `Draggable`s.
+You can use the `DragDrop` component to move items in or between lists. The `DragDrop` component should contain `Droppable` components  which contain `Draggable` components.
 
 ```ts noLive
 import React from 'react';
--- a/packages/react-core/src/components/DragDrop/examples/DragDrop.md
+++ b/packages/react-core/src/components/DragDrop/examples/DragDrop.md
@@ -5,19 +5,23 @@ propComponents: [DragDrop, Draggable, Droppable, DraggableItemPosition]
 beta: true
 ---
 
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/DragDrop/examples/DragDropMultipleLists.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+interface SourceType {
+  droppableId: string;
+  index: number;
+}
+
+interface DestinationType extends SourceType {}
+
+const getItems = (count: number, startIndex?: number) =>
```
thatblindgeye, line 16: Looks like this line is causing the syntax error in the preview build, the `?` in particular I believe?

Multiple lists example syntax error

thatblindgeye (approved):

packages/react-core/src/components/DragDrop/examples/DragDrop.md
```diff
@@ -1,175 +1,39 @@
...
-  DragDrop,
-  Draggable,
-  Droppable,
-  DraggableItemPosition
-]
+propComponents: [DragDrop, Draggable, Droppable, DraggableItemPosition]
 beta: true
 ---
 
 You can use `DragDrop` to move items in or between lists. `DragDrop`s should contain `Droppable`s which contain `Draggable`s.
```
tlabaj, line 8: I know this is outside of the scope of this PR, but I think we should change this wording a little because of the way renders in the docs.  I would say 

You can use `DragDrop` component to move items in or between lists. The `DragDrop` component should contain `Droppable` components  which contain `Draggable` components.

tlabaj (changes requested):

tlabaj (changes requested):
looks good.  Just the one issue with the comment.

jenny-s51 (commented):
Good call @tlabaj ! Updated - description looks much cleaner this way 👍

tlabaj (approved):

Issue #6792, opened 2022-01-13
Make demos using search input consistent


mmenestr:
Many of the React examples and React demos are using inconsistent search inputs. Some are using basic search inputs, while others are using input groups.

Since we are now steering away from using input groups for search inputs, we would like to remove any instance of those, and replace them with a basic search input.

Here is a spreadsheet where i've tracked demos that are inconsistent with what we're aiming for, and therefore would need to be edited:
https://docs.google.com/spreadsheets/d/1VXHUw2gfn9lodhie_I8C0nfJUy9j4tf_yRTIxhdmjzc/edit#gid=2024512193

The main two issues found were:
(1) Input group component is used instead of a search input
(2) A search field is used, but the search icon on the left is missing

GOAL: Make all demos with a search use a search input component

If the search had placeholder text in it, keep the text. If it didn't put it was part of an attribute filter, please add placeholder text that says "Filter by ___" depending on what is being filtered. In most cases, it seems like it'll be "Filter by name".

cc: @mcarrano

Task

- [x] #7241
- [x] #7248

The following tasks are in beta components.   We need to consider when the right time to make the update is:
- [x] #7977
- [x] #7245

tlabaj:
all non breaking tasks complete.  See issue #8167 for related breaking change tasks.

PR #7246, opened 2022-04-13
docs(Tree view): Updated search example to use search input component

### LINKED ISSUE #7245: Tree view: Search should use search input component

tlabaj:
**What**: Closes #7245

**Additional issues**: towards #6792 

Even though Tree view is Beta, I have left the `TreeViewSearch` component in the code for now.  We can remove it once we get a sense of impact on projects after Summit.

### DIFF
--- a/packages/react-core/src/components/TreeView/examples/TreeView.md
+++ b/packages/react-core/src/components/TreeView/examples/TreeView.md
@@ -2,7 +2,7 @@
 id: Tree view
 section: components
 cssPrefix: pf-c-treeview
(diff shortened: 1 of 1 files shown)


nicolethoen (commented):
nit: the core example did add a divider to the example below the search input to preserve the visual. We probably want to do the same for parity.

tlabaj:
The search input  needs to be wrapped in `pf-c-tree-view__search`. This would break structure. So I will close PR for now.

PR #7250, opened 2022-04-14
feat(HelperText): add react demos

### LINKED ISSUE #5941: Add HelperText demo/documentation
Our helper text component currently requires the user to add all accessibility considerations to dynamic helper text. However, this leaves a lot of room for error for users implementing it. We should create a demo of dynamic helper text that would follow the pattern used by form. I imagine the use case would be similar to our invalid form example where the helper text would need to be inside of a live region (`aria-live="polite"`). Depending on the expected use cases, there may need to be additional attributes to make it accessible (for example, in our [invalid form example](https://www.patternfly.org/v4/components/form#invalid) the input has to also have an `aria-invalid` attribute to announce that the input is invalid). I think it would also be valuable to add a little bit of documentation around this explaining what is being added/these considerations when implementing dynamic helper text.

Potential use cases/examples could be:

#### Invalid
* Form control elements (text input, textarea, form select) 
* Select component

#### Error
* Form component helper text (existing helper text this will likely replace)
* Date picker helper text

#### Danger
* Progress
* Alert
* Button
* Notification drawer item (similar to alert)

Enhancement

thatblindgeye:
**What**: Closes #5941 

[Helper text React demos](https://patternfly-react-pr-7250.surge.sh/components/helper-text/react-demos) for convenience.

These demos are based on conversation from the original issue/PR when Helper Text was created ( #5903 ), including design files, as well as some other component examples/demos on the PF site. The original issue this closes mentioned only the dynamic variant of the Helper Text component, but I think it is worth including additional demos to go over making the component accessible in other common use cases.

I also created an issue in org https://github.com/patternfly/patternfly-org/issues/2917 for a11y docs for the component. Depending on any feedback for this PR, I can transfer some of the repeated verbiage between demos to that a11y doc instead and make the demo text more succinct.

The one demo that could use more feedback is the last one, dynamic variant with static text. I based this off of the [password strength](https://www.patternfly.org/v4/demos/password-strength) demo, which I think is trying to make the helper text accessible but could use tweaking to get it closer where we want it. My implementation here was to have the `aria-describedby` on the input be dynamic, but other possibilities I tinkered with include:

- Adding a static variant Helper Text that dynamically updates ("Username must include the following:" when not valid and "Username is valid!" when valid) and having that live inside an `aria-live` region
- Tweaking the HelperTextItem component so that when it's dynamic, the accessible name is a combination of the variant type ("error", "success", etc) and whatever is passed into the `children` prop, and making that an `aria-live` region. Right now, because the icons are hidden from assistive tech, aria-live won't register an update to the dynamic variants for this demo, which having the accessible name be dynamic should resolve.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/HelperText/HelperText.tsx
+++ b/packages/react-core/src/components/HelperText/HelperText.tsx
@@ -9,17 +9,33 @@ export interface HelperTextProps extends React.HTMLProps<HTMLDivElement | HTMLUL
   className?: string;
   /** Component type of the helper text container */
   component?: 'div' | 'ul';
+  /** ID for the helper text container. The value of this prop can be passed into a form component's
+   * aria-describedby prop when you intend for all helper text items to be announced to
+   * assistive technologies.
+   */
+  id?: string;
+  /** Flag for indicating whether the helper text container is a live region. Use this prop when you
+   * expect or intend for any helper text items within the container to be dynamically updated.
+   */
+  isLiveRegion?: boolean;
 }
 
 export const HelperText: React.FunctionComponent<HelperTextProps> = ({
   children,
   className,
   component = 'div',
+  id,
+  isLiveRegion = false,
   ...props
 }: HelperTextProps) => {
   const Component = component as any;
   return (
-    <Component className={css(styles.helperText, className)} {...props}>
+    <Component
+      id={id}
+      className={css(styles.helperText, className)}
+      aria-live={isLiveRegion ? 'polite' : null}
+      {...props}
+    >
       {children}
     </Component>
   );
--- a/packages/react-core/src/components/HelperText/HelperText.tsx
+++ b/packages/react-core/src/components/HelperText/HelperText.tsx
@@ -33,7 +33,7 @@ export const HelperText: React.FunctionComponent<HelperTextProps> = ({
     <Component
       id={id}
       className={css(styles.helperText, className)}
-      aria-live={isLiveRegion ? 'polite' : null}
+      {...(isLiveRegion && { 'aria-live': 'polite' })}
       {...props}
     >
       {children}
--- a/packages/react-core/src/demos/HelperText.md
+++ b/packages/react-core/src/demos/HelperText.md
@@ -1,0 +1,69 @@
+---
+id: Helper text
+section: components
+---
+
+import MinusIcon from '@patternfly/react-icons/dist/esm/icons/minus-icon';
+import ExclamationTriangleIcon from '@patternfly/react-icons/dist/esm/icons/exclamation-triangle-icon';
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
+import ExclamationCircleIcon from '@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon';
+
+## Demos
+
+### Static variant with dynamic text
+
+In this demo, the static variant of the Helper Text component (the default) is used with the `hasIcon` prop passed in when there is an error, and the text itself dynamically updates based on the input value. When the input has a value of `johndoe` an error is rendered, while an empty input renders other helper text.
+
+The `aria-describedby` attribute is passed into the Text Input component and is linked to the `id` of the Helper Text component. This notifies assistive technologies of the helper text when the input receives focus, which can be helpful if a user navigates away from and then back to the input so that they are reminded of the helper text.
+
+An `aria-live` region is used to notify assistive technologies when there is an update to the helper text, notifying users immediately when an error or another issue occurs. Without this attribute, a user would have to navigate out of and back into the input field to have the new text announced via the `aria-describedby` attribute.
+
+The `aria-invalid` attribute is also passed into the text input, which allows assistive technologies to notify users that an input is invalid. When this attribute is true, it's important that users are notified of what is causing the input to be invalid; in this case, `aria-describedby` and `aria-live` help accomplish this.
+
+```ts
+import React from 'react';
+import { Form, FormGroup, TextInput, HelperText, HelperTextItem } from '@patternfly/react-core';
+import ExclamationCircleIcon from '@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon';
+
+const HelperTextStaticVariantDynamicText: React.FunctionComponent = () => {
+  const [value, setValue] = React.useState('');
+  const [inputValidation, setinputValidation] = React.useState('default');
+
+  const handleInputChange = inputValue => {
+    setValue(inputValue);
+  };
+
+  React.useEffect(() => {
+    if (value === '') {
+      setinputValidation('default');
+    } else if (value === 'johndoe') {
+      setinputValidation('error');
+    } else {
+      setinputValidation('success');
+    }
+  }, [value]);
+
+  return (
+    <Form>
+      <FormGroup label="Username" isRequired fieldId="login-input-helper-text1">
+        <TextInput
+          isRequired
+          type="text"
+          id="login-input-helper-text1"
+          name="login-input-helper-text1"
+          onChange={handleInputChange}
+          aria-describedby="helper-text1"
+          aria-invalid={inputValidation === 'error'}
+        />
+      </FormGroup>
+      <HelperText id="helper-text1" aria-live="polite">
+        {inputValidation !== 'success' && (
+          <HelperTextItem variant={inputValidation as any} hasIcon={inputValidation === 'invalid'}>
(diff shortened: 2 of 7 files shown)


mcarrano (approved):
Looks good to me. Thanks @thatblindgeye !

packages/react-core/src/demos/HelperText.md
```diff
@@ -0,0 +1,200 @@
...
+        )}
+      </HelperText>
+    </Form>
+  );
+};
+```
+
+### Dynamic variant with static text
+
+In this demo, the Helper Text Item components have the `isDynamic` prop passed in. While the text content of the components is static, the icons and styling will change as the validation of the input changes.
```
tlabaj, line 118: This sentence helped me understand better what the 'isDynamic` prop does.  I know it is outside the scope of the issue her, but it would be nice to see the prop description expanded upon, or even add a blurb to the example for dynamic.  I can open a separate issue?

tlabaj (changes requested):

packages/react-core/src/demos/HelperText.md
thatblindgeye: I can update the prop description along with the above updates

packages/react-core/src/demos/HelperText.md
```diff
@@ -0,0 +1,200 @@
...
+import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle-icon';
+import ExclamationCircleIcon from '@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon';
+import CheckIcon from '@patternfly/react-icons/dist/esm/icons/check-icon';
+import TimesIcon from '@patternfly/react-icons/dist/esm/icons/times-icon';
+
+## Demos
+
+### Static variant with static text
+
+In this demo, the static variant of the Helper Text Item component (the default) is used, and the text itself will always be visible to users and will never change.
```
tlabaj, line 17: @wolfeallison please review lines 17-21

packages/react-core/src/demos/HelperText.md
```diff
@@ -0,0 +1,200 @@
...
+        <HelperTextItem variant={'default'}>Enter your middle name or your middle initial</HelperTextItem>
+      </HelperText>
+    </Form>
+  );
+};
+```
+
+### Static variant with dynamic text
+
+In this demo, the static variant of the Helper Text Item component (the default) is used with the `hasIcon` prop passed in when there is an error, and the text itself dynamically updates based on the input value. When the input has a value of `johndoe` an error is rendered to simulate a username already being taken, while an empty input renders other helper text. When the input is valid, no helper text is rendered.
```
tlabaj, line 58: @wolfeallison please review lines 58-64

packages/react-core/src/demos/HelperText.md
tlabaj, line 118: @wolfeallison please review lines 120-122

packages/react-core/src/demos/HelperText.md
wolfeallison: I would insert a comma after "johndoe"

packages/react-core/src/demos/HelperText.md
wolfeallison: Slight changes on line 120 to make it easier to read: The `aria-describedby` attribute is passed into the Text Input component and is linked to the `id` attribute of a helper text item that results in an invalid input.

wolfeallison: On line 122, change "one" to "1" - we use numerics because they are easier to scan, even though some styles call for numbers under 10 to be written out, we break that rule!

packages/react-core/src/demos/HelperText.md
```diff
@@ -0,0 +1,200 @@
...
+### Static variant with static text
+
+In this demo, the static variant of the helper text item component (the default) is used, and the text itself will always be visible to users and will never change.
+
+The `aria-describedby` attribute is passed into the text input component and is linked to the `id` of the helper text component. This allows assistive technologies to notify users of the helper text content when the input receives focus, which can be helpful if a user navigates away from and then back to the input.
+
+Note that this demo does not validate the text input component. When it would need to be validated, there are other steps that would need to be taken to make it accessible, such as passing in `aria-invalid` and `aria-live` attributes to the appropriate components.
+
+```ts
+import React from 'react';
```
wise-king-sullyman, line 24: Is there a particular reason that you built the demos inside the .md rather than putting them in standalone files in `demos/examples/<component>/<demo name>.tsx`?

thatblindgeye: The component demo I was using as reference was set up that way so I think I was just following that as I was building things out. Separate demos have now been created, though. Good call out!

tlabaj (approved):
LGTM!

nicolethoen (approved):

wise-king-sullyman (approved):

packages/react-core/src/demos/examples/HelperText/HelperTextStaticVariantStaticText.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <TextInput
+          type="text"
+          id="login-input-helper-text1"
+          name="login-input-helper-text1"
+          onChange={handleInputChange}
+          aria-describedby="helper-text1"
+          value={value}
+        />
+      </FormGroup>
+      <HelperText id="helper-text1">
```
mcoker, line 23: Can you move this up after the `<TextInput>` and wrap it in `<FormHelperText>` like in the [password strength helper text](https://github.com/patternfly/patternfly-react/blame/main/packages/react-core/src/demos/PasswordStrength.md#L153-L165)? That should result in less space between the text input and helper text.

mcoker (changes requested):

mcoker (approved):
LGTM!!

jessiehuff (commented):
Looks pretty good so far! I think I agree with your suggestions on the last demo. I think it probably should be an `aria-live="polite"` similar to the password strength demo and the form examples. What is interesting is on the password strength demo, the `aria-live` is on the `ul` instead of the `div` on the form. I'm not quite sure what the decision was for that, but I do think it would help us to make sure that live region is on this demo. When I was testing, it seemed to be a touch inconsistent in announcing since I don't think we have that live region right now. I think your second suggestion could be valuable as well, but I wonder if we'd use an alert in that scenario and follow [this pattern](https://www.patternfly.org/v4/components/form/#invalid-with-form-alert)?

nicolethoen:
I just read the a11y docs you wrote, and I think you should add `aria-live, `aria-describedby`, and `id` to the docs explicitly so that they appear in the props tables since they are referenced as props in the a11y docs.

nicolethoen (changes requested):
I just read the a11y docs you wrote, and I think you should add `aria-live`, `aria-describedby`, and `id` to the component props explicitly so that they appear in the props tables since they are referenced as props in the a11y docs.

thatblindgeye:
> Looks pretty good so far! I think I agree with your suggestions on the last demo. I think it probably should be an `aria-live="polite"` similar to the password strength demo and the form examples. What is interesting is on the password strength demo, the `aria-live` is on the `ul` instead of the `div` on the form. I'm not quite sure what the decision was for that, but I do think it would help us to make sure that live region is on this demo. When I was testing, it seemed to be a touch inconsistent in announcing since I don't think we have that live region right now. I think your second suggestion could be valuable as well, but I wonder if we'd use an alert in that scenario and follow [this pattern](https://www.patternfly.org/v4/components/form/#invalid-with-form-alert)?

@jessiehuff Yeah I was also thinking adding a live region could be useful; aria-describedby is good when focusing the input field but can definitely be hit or miss when typing in it. Do you feel adding in the live-region would be better implemented with this PR, or do you think a separate issue would suffice? There are several ways to go about it, it just may require some testing/tweaking to put together a working concept.

packages/react-core/src/components/HelperText/HelperText.tsx
```diff
@@ -9,17 +9,33 @@ export interface HelperTextProps extends React.HTMLProps<HTMLDivElement | HTMLUL
...
+  isLiveRegion = false,
   ...props
 }: HelperTextProps) => {
   const Component = component as any;
   return (
-    <Component className={css(styles.helperText, className)} {...props}>
+    <Component
+      id={id}
+      className={css(styles.helperText, className)}
+      aria-live={isLiveRegion ? 'polite' : null}
```
nicolethoen, line 36: to avoid having aria-live=null, could it be
```
{...(isLiveRegion && {'aria-live': 'polite'})}
```

nicolethoen (approved):

Issue #7226, opened 2022-04-12
Form - Add prop to set max width


mcarrano:
**Is this a new component or an extension of an existing one?**
Form
**Describe the feature**
Currently when the consumer sets `isWidthLimited` the width of a form will be set to 500px. Per our design guidance, forms should be limited to 880px by default. Proposal is to change the default from 500px to 800px when this prop is set and add a new prop to expose the max width in case the consumer wants to modify this value.

Core PR: https://github.com/patternfly/patternfly/pull/4782

PR #7253, opened 2022-04-14
feat(form): added maxWidth prop

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/7226

@tlabaj just to confirm from our convo the other day, if we end up adding breakpoint support for this var/prop later on, we can use this existing prop to either accept a string (a static max-width), or an object with default, sm, md, lg, etc values for max-widths at different breakpoints without introducing a breaking change? Or there may be another way to do it without a breaking change?

### DIFF
--- a/packages/react-core/src/components/Form/Form.tsx
+++ b/packages/react-core/src/components/Form/Form.tsx
@@ -9,8 +9,10 @@ export interface FormProps extends React.HTMLProps<HTMLFormElement> {
   className?: string;
   /** Sets the Form to horizontal. */
   isHorizontal?: boolean;
-  /** Flag to limit the max-width to 500px. */
+  /** Limits the max-width of the form. */
   isWidthLimited?: boolean;
+  /** Sets a custom max-width when using isWidthLimited. */
+  maxWidth?: string;
 }
 
 export const Form: React.FunctionComponent<FormProps> = ({
@@ -18,15 +20,23 @@ export const Form: React.FunctionComponent<FormProps> = ({
   className = '',
   isHorizontal = false,
   isWidthLimited = false,
+  maxWidth = '',
   ...props
 }: FormProps) => (
   <form
     noValidate
+    {...(maxWidth && {
+      style: {
+        '--pf-c-form--m-limit-width--MaxWidth': maxWidth,
+        ...props.style
+      } as React.CSSProperties
+    })}
     {...props}
     className={css(
       styles.form,
       isHorizontal && styles.modifiers.horizontal,
       isWidthLimited && styles.modifiers.limitWidth,
+      maxWidth && styles.modifiers.limitWidth,
       className
     )}
   >
--- a/packages/react-core/src/components/Form/Form.tsx
+++ b/packages/react-core/src/components/Form/Form.tsx
@@ -11,7 +11,7 @@ export interface FormProps extends React.HTMLProps<HTMLFormElement> {
   isHorizontal?: boolean;
   /** Limits the max-width of the form. */
   isWidthLimited?: boolean;
-  /** Sets a custom max-width when using isWidthLimited. */
+  /** Sets a custom max-width for the form. */
   maxWidth?: string;
 }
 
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Form/Form.tsx
```diff
@@ -9,24 +9,34 @@ export interface FormProps extends React.HTMLProps<HTMLFormElement> {
   className?: string;
   /** Sets the Form to horizontal. */
   isHorizontal?: boolean;
-  /** Flag to limit the max-width to 500px. */
+  /** Limits the max-width of the form. */
   isWidthLimited?: boolean;
+  /** Sets a custom max-width when using isWidthLimited. */
+  maxWidth?: string;
```
tlabaj, line 15: We can update this prop to be a union later on with default, sm, md, lg, or with an object with max-widths at different breakpoints. That would not be breaking.

tlabaj (approved):
LGTM!

packages/react-core/src/components/Form/Form.tsx
```diff
@@ -9,24 +9,34 @@ export interface FormProps extends React.HTMLProps<HTMLFormElement> {
   className?: string;
   /** Sets the Form to horizontal. */
   isHorizontal?: boolean;
-  /** Flag to limit the max-width to 500px. */
+  /** Limits the max-width of the form. */
   isWidthLimited?: boolean;
+  /** Sets a custom max-width when using isWidthLimited. */
```
wise-king-sullyman, line 14: Nit: I think `isWidthLimited` should be sentence case.

packages/react-core/src/components/Form/Form.tsx
```diff
@@ -9,24 +9,34 @@ export interface FormProps extends React.HTMLProps<HTMLFormElement> {
...
+        '--pf-c-form--m-limit-width--MaxWidth': maxWidth,
+        ...props.style
+      } as React.CSSProperties
+    })}
     {...props}
     className={css(
       styles.form,
       isHorizontal && styles.modifiers.horizontal,
       isWidthLimited && styles.modifiers.limitWidth,
+      maxWidth && styles.modifiers.limitWidth,
```
wise-king-sullyman, line 39: I *think* we would want to apply limit width like `(isWidthLimited || maxWidth) && styles.modifiers.limitWidth`, as applying it in the current manner would cause the class name to be applied twice if both `isWidthLimited` and `maxWidth` are applied.

wise-king-sullyman (commented):
Looks good! I just have a couple of small tweaks to suggest.

packages/react-core/src/components/Form/Form.tsx
mcoker: Not sure if I understand - can you add the code suggestion change to the comment?

packages/react-core/src/components/Form/Form.tsx
mcoker: doh! I actually meant to remove this line before committing 😅

packages/react-core/src/components/Form/Form.tsx
mcoker: Actually I just removed the refence to `isWidthLimited` part now that `maxWidth` applies `.pf-m-limit-width`. wdyt?

wise-king-sullyman:
```suggestion
  /** Sets a custom max-width when using the is width limited prop. */
```

I don't love this phrasing but can't think of anything better at the moment. The goal was just to get away from having camel case `isWidthLimited` in the prop description, but just getting rid of the camel case makes the sentence read weirdly.

Related page from the PF guidelines: https://www.patternfly.org/v4/ux-writing/capitalization/#capitalization-across-patternfly. Props aren't explicitly mentioned though, so maybe this is fine. @tlabaj thoughts?

wise-king-sullyman (approved):
Looks perfect!

packages/react-core/src/components/Form/Form.tsx
tlabaj: I think that this is a special case and we should leave the prop name other wise it does not make sense like you said @wise-king-sullyman

nicolethoen (approved):

PR #7274, opened 2022-04-20
fix(Page): add tabindex to components with hasOverflowScroll

### LINKED ISSUE #7078: PageSection hasOverflowScroll should add a tabindex to allow for scrolling using keyboard.
PageSection hasOverflowScroll should add a tabindex to allow for scrolling using keyboard.

thatblindgeye:
**What**: Closes #7078 

Also added a tabindex to other Page components that are able to have a `hasOverflowScroll` prop passed in.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Page/PageBreadcrumb.tsx
+++ b/packages/react-core/src/components/Page/PageBreadcrumb.tsx
@@ -40,6 +40,7 @@ export const PageBreadcrumb = ({
       hasOverflowScroll && styles.modifiers.overflowScroll,
       className
     )}
+    {...(hasOverflowScroll && { tabIndex: 0 })}
     {...props}
   >
     {isWidthLimited && <div className={css(styles.pageMainBody)}>{children}</div>}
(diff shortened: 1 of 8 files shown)


nicolethoen (commented):
Recently, @seanforyou23 made a comment on some a11y docs about this sort of update and suggested adding an accessible label. This is because the non-negative tabindex will make the container register as a form-control on rotor menus or elements lists and without an explicit label, various browsers or screen readers may label it differently/incorrectly.
Comment: https://github.com/patternfly/patternfly-org/pull/2893#discussion_r841910449

That said, I dont think we need to update the example docs - I am adding this suggestion to the a11y docs. But maybe we should add an aria-label or aria-labelledby to elements we are intentionally adding a tabindex to?

thatblindgeye:
Created a separate issue for the above convo regarding `aria-label` https://github.com/patternfly/patternfly-react/issues/7350

nicolethoen (approved):

kmcfaul (approved):

tlabaj (approved):
lgtm

jessiehuff (approved):
LGTM! :)

PR #7277, opened 2022-04-20
feat(table): expandable table demo

### LINKED ISSUE #7271: Expandable

jenny-s51:
**What**: Closes #7271

### DIFF
--- a/packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
+++ b/packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Table, TableHeader, TableBody, expandable } from '@patternfly/react-table';
+import { PageSection, Table, TableHeader, TableBody, expandable } from '@patternfly/react-table';
+import DashboardWrapper from '@patternfly/react-core/src/demos/examples/DashboardWrapper';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 class ExpandCollapseAllTableDemo extends React.Component {
@@ -120,19 +121,23 @@ class ExpandCollapseAllTableDemo extends React.Component {
 
     return (
       <React.Fragment>
-        <Table
-          aria-label="Collapsible table"
-          onSelect={this.onSelect}
-          onCollapse={this.onCollapse}
-          rows={rows}
-          cells={columns}
-          canSelectAll={true}
-          canCollapseAll={true}
-          collapseAllAriaLabel={collapseAllAriaLabel}
-        >
-          <TableHeader />
-          <TableBody />
-        </Table>
+        <DashboardWrapper hasPageTemplateTitle>
+          <PageSection isWidthLimited>
+            <Table
+              aria-label="Collapsible table"
+              onSelect={this.onSelect}
+              onCollapse={this.onCollapse}
+              rows={rows}
+              cells={columns}
+              canSelectAll={true}
+              canCollapseAll={true}
+              collapseAllAriaLabel={collapseAllAriaLabel}
+            >
+              <TableHeader />
+              <TableBody />
+            </Table>
+          </PageSection>
+        </DashboardWrapper>
       </React.Fragment>
     );
   }
--- a/packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
+++ b/packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
@@ -14,7 +14,7 @@ class ExpandCollapseAllTableDemo extends React.Component {
       collapseAllAriaLabel: 'Expand all',
       columns: [
         {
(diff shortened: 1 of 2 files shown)


kmcfaul:
I think you're missing the isFullPage flag on the md as the example is still inline

wise-king-sullyman (approved):

thatblindgeye (approved):

mmenestr:
If I'm looking at the right demo: https://patternfly-react-pr-7277.surge.sh/components/table/react-demos/expandcollapse-all/ - is it normal that the first row doesn't have a checkbox?

tlabaj (changes requested):
I think we want this one to match core's demo.  The core demo has a toolbar and pagination at the top and bottom.

jenny-s51:
Thank you @tlabaj - I've added top/bottom pagination to this PR to match the core demo. Per team consensus, the rest of the toolbar features will be determined and implemented once https://github.com/patternfly/patternfly-design/issues/1157 and https://github.com/patternfly/patternfly-react/issues/7384 are complete.

Nice catch @mmenestr - removed the first checkbox to match the core demo!

nicolethoen (approved):

gitdallas (approved):

tlabaj (approved):

mcarrano (approved):
LGTM! Thanks @jenny-s51 .

kmcfaul (approved):

packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
```diff
@@ -115,24 +126,53 @@ class ExpandCollapseAllTableDemo extends React.Component {
...
-          canSelectAll={true}
-          canCollapseAll={true}
-          collapseAllAriaLabel={collapseAllAriaLabel}
-        >
-          <TableHeader />
-          <TableBody />
-        </Table>
+        <DashboardWrapper hasPageTemplateTitle>
+          <PageSection isWidthLimited>
+            {tableToolbar}
```
mattnolting, line 159: Needs to be wrapper with `Card`

packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
```diff
@@ -115,24 +126,53 @@ class ExpandCollapseAllTableDemo extends React.Component {
...
   render() {
     const { columns, rows, collapseAllAriaLabel } = this.state;
 
+    const tableToolbar = (
+      <Toolbar id="compact-toolbar">
+        <ToolbarContent>
+          <ToolbarItem variant="pagination">{this.renderPagination('top', true)}</ToolbarItem>
+        </ToolbarContent>
+      </Toolbar>
+    );
```
mattnolting, line 153: I think I'd remove the toolbar if it only has pagination as the white space on the left seems odd. @mcarrano wdyt?

mattnolting (commented):
Nice work! Left a few comments.

packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
tlabaj: We could also add some other controls to the toolbar?

mmenestr:
Last thing - is it normal that the first row doesn't have an expansion icon?

packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
mcarrano: I agree with @mattnolting . But I'm OK to either remove it or add some simple filter actions. But did we say we would remove these toolbars until we worked out the correct structure to match core, or am I confusing this with another issue @tlabaj ?

jenny-s51: +1 @mattnolting @mcarrano, I think it would make sense to remove the toolbar here as well since we need to figure out the correct table structure and data first before implementing controls that would use the table data

jenny-s51:
> Last thing - is it normal that the first row doesn't have an expansion icon?

Yes @mmenestr , the first row doesn't have an expansion icon because there currently are no designs or sample data for all the expanded states of https://pf4.patternfly.org/components/table/html-demos/expandable/

We can update this table to fully match the core demo once we have all the data/designs i.e. when
https://github.com/patternfly/patternfly-design/issues/1157 and https://github.com/patternfly/patternfly-react/issues/7384 are complete.

mattnolting (approved):
LPTM! Thanks @jenny-s51!

packages/react-table/src/demos/table-demos/ExpandCollapseAll.jsx
tlabaj: That was a different PR @mcarrano but it applies here as well.  We can add the toolbar back once we have a data set.

Issue #7227, opened 2022-04-12
Data list - Update expandable demo to match core


mcarrano:
**Is this a new component or an extension of an existing one?**
Data list

**Describe the feature**
Update the [expandable data list demo](https://www.patternfly.org/v4/components/data-list/react-demos#expandable-control-in-toolbar) to be a full page demo that matches core implementation.

Core PR: https://github.com/patternfly/patternfly/pull/4784

mcarrano:
Duplicate of #7178

PR #7279, opened 2022-04-20
fix(DataList): Remove primary button from expandable demo

### LINKED ISSUE #6933: Edit expandable data list example to remove the primary button
This demo uses a primary blue button to allow users to interact with the expand all/collapse feature: https://www.patternfly.org/v4/components/data-list#expandable

Having that functionality in the demo is great, the only issue is that people are considering the use of a primary button in this context a design suggestion. **This is not what we want to convey, so we should update that primary button to something else.**

Perhaps just a blue link button? but @mcarrano @mceledonia do you have any other ideas?

jenny-s51:
**What**: Closes #6933

### DIFF
--- a/packages/react-core/src/components/DataList/examples/DataListExpandable.tsx
+++ b/packages/react-core/src/components/DataList/examples/DataListExpandable.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import {
-  Button,
   DataList,
   DataListItem,
   DataListItemRow,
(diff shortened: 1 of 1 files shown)


mcoker (approved):
LGTM! Only other thing I think may be useful is to remove the line breaks between the toolbar and data-list in the demo - though we do have a separate issue to [update the react data list demos to match core](https://github.com/patternfly/patternfly-react/issues/7227), which could include that, too.

https://github.com/patternfly/patternfly-react/blob/cc80f358e790331d92df19b49df25488b095265b/packages/react-core/src/components/DataList/examples/DataListExpandable.tsx#L76-L77

tlabaj (approved):
LGTM

PR #7281, opened 2022-04-21
feat(CodeBlock): add class to code component

### LINKED ISSUE #7156: Bug - CodeBlockCode - Does not have a className prop similar to codeblockaction
**Describe the problem**
https://github.com/openshift/console/pull/11263/files/4672dde3878ccb18e3c7b3973b0cfe52e695baf8#r838956868
CodeBlockCode does not have a className prop similar to CodeBlockAction to add additional classes to the code block.

**Expected behavior**
have a className prop

There is a workaround to get styles to the block.

Openshift V4.11
**Any other information?**

kmcfaul:
**What**: Closes #7156

Added a class to both the wrapper and the actual code component, though I'm unsure if just the wrapper className prop would suffice. LMK if it would be clearer to remove the code classname.


jcaianirh (approved):
lgtm

tlabaj (approved):
I don't know that we need the prop on code, but I think it could be useful.

PR #7282, opened 2022-04-21
Add correct types for class components using context API

jonkoops:
In React 18 the default type for `context` is `unknown`, meaning that unless proper types are added the build will fail. This adds the correct types to these contexts when used in class component.

Needed to land #7142.


jenny-s51 (commented):
LGTM, modified components render and function as expected

wise-king-sullyman (approved):

kmcfaul (approved):

tlabaj (approved):
lgtm

PR #7284, opened 2022-04-21
fix(DataListCheck): revert checked prop

### LINKED ISSUE #7283: Bug - DataListCheck - checked prop needs to be reverted
**Describe the problem**
The DataListCheck component fix introduced in #7163 introduced an issue when trying to use the `checked` prop with a controlled DataListCheck component. The component needs to be reverted to before this fix was merged.

Provide steps to reproduce. A [codesandbox](https://codesandbox.io/s/serverless-cherry-q9t3f ) demonstrating the problem is appreciated.

**Expected behavior**
A clear and concise description of the expected behavior.

List the workaround if there is one.

**Screenshots**
If applicable, add screenshots to help explain the issue.

**What is your environment?**

**Any other information?**

thatblindgeye:
**What**: Closes #7283 

This PR somewhat reverts the component to before PR #7163 while retaining the `defaultChecked` behavior.

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/DataList/DataListCheck.tsx
+++ b/packages/react-core/src/components/DataList/DataListCheck.tsx
@@ -9,10 +9,15 @@ export interface DataListCheckProps extends Omit<React.HTMLProps<HTMLInputElemen
   isValid?: boolean;
   /** Flag to show if the DataList checkbox is disabled */
   isDisabled?: boolean;
-  /** Flag to show if the DataList checkbox is checked when it is controlled by React state. Both isChecked and checked are valid.
+  /** Flag to show if the DataList checkbox is checked when it is controlled by React state. Both isChecked and checked are valid,
+   * but only use one.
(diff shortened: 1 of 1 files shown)


nicolethoen:
I just want to be sure that if someone was using a prop thinking it was setting the defaultChecked attribute, that it will still work as before in those cases.

thatblindgeye:
> I just want to be sure that if someone was using a prop thinking it was setting the defaultChecked attribute, that it will still work as before in those cases.

If someone has started using the `checked` prop to set the defaultChecked attribute since the previous PR was merged, this revert could cause an initial issue if the `defaultChecked` prop isn't used instead. I think if anyone was trying to use `checked` to set the defaultChecked attribute prior to that, it wouldn't have been working as intended.

tlabaj (commented):
LGTM!  Thanks for making the change.

kmcfaul (approved):

tlabaj (approved):

PR #7285, opened 2022-04-21
feat(divider): add support for switching orientation at various breakpoints

### LINKED ISSUE #7225: Divider - add breakpoint support for horizontal/vertical change
Extends core work done in https://github.com/patternfly/patternfly/pull/4765

jenny-s51:
**What**: Closes #7225


packages/react-core/src/components/Divider/Divider.tsx
```diff
@@ -25,13 +25,22 @@ export interface DividerProps extends React.HTMLProps<HTMLElement> {
     xl?: 'insetNone' | 'insetXs' | 'insetSm' | 'insetMd' | 'insetLg' | 'insetXl' | 'inset2xl' | 'inset3xl';
     '2xl'?: 'insetNone' | 'insetXs' | 'insetSm' | 'insetMd' | 'insetLg' | 'insetXl' | 'inset2xl' | 'inset3xl';
   };
+  /** Indicates how the divider will display at various breakpoints. */
+  orientation?: {
```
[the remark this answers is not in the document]
mcoker: @tlabaj yep, that would be inline with how we handle these breakpoint modifiers in other components.

mcoker (changes requested):
LGTM! Though I agree with @tlabaj, we should just use this instead of `isVertical` in the examples and plan to remove that prop in our next breaking change release.

mcarrano (approved):
Looks fine from my perspective

jenny-s51 (commented):
Thank you @tlabaj @mcoker @mcarrano ! 

I've added `default` to the `orientation` prop, and updated the `isVertical` description to reflect that it is now deprecated. 

I've also updated the vertical demos to use `orientation`.

tlabaj:
@jenny-s51 I opened #7325 to remove deprecated prop when we have breaking change.

tlabaj (approved):
LGTM

packages/react-core/src/components/Divider/Divider.tsx
```diff
@@ -25,13 +25,23 @@ export interface DividerProps extends React.HTMLProps<HTMLElement> {
...
+    '2xl'?: 'vertical' | 'horizontal';
+  };
 }
 
 export const Divider: React.FunctionComponent<DividerProps> = ({
   className,
   component = DividerVariant.hr,
   isVertical = false,
   inset,
+  orientation,
```
nicolethoen, line 44: can we or should we set the default orientation alongside the other default props? Before, `isVertical` defaulted to false.

packages/react-core/src/components/Divider/Divider.tsx
```diff
@@ -25,13 +25,23 @@ export interface DividerProps extends React.HTMLProps<HTMLElement> {
     xl?: 'insetNone' | 'insetXs' | 'insetSm' | 'insetMd' | 'insetLg' | 'insetXl' | 'inset2xl' | 'inset3xl';
     '2xl'?: 'insetNone' | 'insetXs' | 'insetSm' | 'insetMd' | 'insetLg' | 'insetXl' | 'inset2xl' | 'inset3xl';
   };
+  /** Indicates how the divider will display at various breakpoints. Vertical divider must be in a flex layout. */
```
mcoker, line 28: Just a nit and outside the scope, but since we're copying this doc comment, is the bit about it needing to be in a flex layout necessary? I'm assuming so since the vertical divider doesn't really work on its own like the default/horiz variant. If we're going to leave it, technically the parent (or divider itself) just needs a height of some sort, or the parent needs to support `align-items` so that [`align-items: stretch` on the divider](https://github.com/patternfly/patternfly/blob/9da0f628740d4dc7ad78c3a11181d8615196e8ec/src/patternfly/components/Divider/divider.scss#L72) works (this is why it works in flex, but that is supported in grid, too).

Here are a few examples - https://codepen.io/mcoker/pen/ZErzBeg

This is what the [vertical divider core docs](https://www.patternfly.org/v4/components/divider/html#usage) say

> This modifier requires that the parent has an explicit or implicit height, or has a flex or grid based layout parent.

mcoker (commented):
LGTM - one nit about the docs, not sure if it's something we want to update in this PR or not.

Also looks like we use a vertical divider in other places - should we update those, too? Or anywhere else you can think we might have a vertical divider?

```
marvin:patternfly-react cmichael$ grep -ri '<Divider.*isVertical' packages/react-core/
packages/react-core//src/components/Divider/__tests__/Divider.test.tsx:      <Divider isVertical/>
packages/react-core//src/components/Divider/examples/DividerVerticalFlexInsetMedium.tsx:    <Divider isVertical inset={{ default: 'insetMd' }} />
packages/react-core//src/components/Divider/examples/DividerVerticalFlex.tsx:    <Divider isVertical />
packages/react-core//src/demos/CardDemos.md:              {content.length > 1 && index === 0 && <Divider key={`${index}_d`} isVertical />}
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />
packages/react-core//src/demos/CardDemos.md:                      <Divider className="pf-u-hidden-on-md" isVertical inset={{ default: 'insetSm' }} />

```

packages/react-core/src/components/Divider/Divider.tsx
mcoker: @jenny-s51 gotcha! I guess it's a little confusing to me since it can be used outside of a CSS flex layout, but mostly works in the `<Flex>` react layout.

packages/react-core/src/components/Divider/examples/DividerOrientationVariousBreakpoints.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Divider, Flex, FlexItem } from '@patternfly/react-core';
+
+export const DividerOrientationVariousBreakpoints: React.FunctionComponent = () => (
+  <Flex>
+    <FlexItem>first item</FlexItem>
+    <Divider
+      orientation={{
+        default: 'vertical',
+        sm: 'vertical',
```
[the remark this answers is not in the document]
jenny-s51: good point, fixed 👍

jenny-s51 (commented):
Good call on the breakpoints demo, thank you @mcoker! I've added `2xl` and updated the `orientation` prop to alternate between `vertical` and `horizontal` for each breakpt, starting with `vertical` as the default.

mcoker (approved):
🥳

PR #7292, opened 2022-04-22
feat(table): update sortable demo to align with core

### LINKED ISSUE #7272: Feat(Table): Full screen demo for sortable

jenny-s51:
**What**: Closes #7272 

**Additional issues**:


mmenestr:
Is this demo the one I'm supposed to be checking for this PR? https://patternfly-react-pr-7292.surge.sh/components/table#composable-sortable---custom-control

mcoker:
@mmenestr it's this one

https://patternfly-react-pr-7292.surge.sh/components/table/react-demos/sortable---responsive/

wise-king-sullyman (commented):
Is this the right core demo to be looking at? https://www.patternfly.org/v4/components/table/html-demos/sortable

If so I'm not sure if these differences are substantive or not, but I notice a handful of differences between the demos still.

Areas of difference:
- Masthead logo
- Masthead toolbar items
- Breadcrumbs below the masthead
- Items in the toolbar group
- Checkboxes in the rows
- Pagination at the bottom of the table

wise-king-sullyman:
I'm seeing a `ReferenceError: number is not defined` error when navigating to the demo now.

wise-king-sullyman:
I'm still seeing a handful of differences in the toolbars, specifically at larger window widths. I.e.

image
 vs
 
image

I'm not sure if this is significant since this demo seems to focus on small window usage.

jenny-s51:
Thank you for your feedback @wise-king-sullyman ! I've added some changes that should make the toolbar and dashboard header align with the core demos more closely.

Per @tlabaj and @mcarrano, we are separating the implementation of additional toolbar logic from these table PRs. We will address that in an upcoming sprint as these must be resolved first:
- https://github.com/patternfly/patternfly-design/issues/1157 
- https://github.com/patternfly/patternfly-react/issues/7384

Please kindly re-review when you have a moment so we can prioritize getting these table changes in for the release 🙂

wise-king-sullyman (approved):
If the additional toolbar differences will be resolved in future issues this looks perfect 🔥

tlabaj (commented):
Looks pretty good.  Only thing I am seeing is the the "Responsive action hidden on small button" is not getting hidden on small for me.

gitdallas (approved):

tlabaj (changes requested):
@jenny-s51 this still is not working correctly.  I am also noticing that the structure does not match core.  Core's toolbar seems to be in a card and this one is not.  I am also seeing other discrepancies lie that toolbar content section in core has "pf-m-nowrap" applied.

tlabaj (changes requested):
It is looking better!  however I see the "sort" button disappearing when the screen is very small (phone size) 

image

Here it shows when it is bigger (tablet size).

image

mcarrano:
Good catch @tlabaj 

@jenny-s51 for reference, this core demo shows how sorting should behave when the view port shrinks: https://www.patternfly.org/v4/components/table/html-demos/sortable/  Basically a sort menu is exposed in the toolbar after the table converts to its stacked orientation.

jenny-s51:
@tlabaj @mcarrano I've updated with a fix that keeps the sort menu in the toolbar when using a phone size viewport.

mcarrano (approved):
This looks great. Thanks @jenny-s51 !

tlabaj (changes requested):
The responsive behaviour is still slightly off.  I see the sorting button before the table transitions to a stacked layout.

image

packages/react-table/src/demos/Table.md
```diff
@@ -2397,52 +2394,77 @@ ComposableTableSortable = () => {
...
+                          onSelect: (_event, isSelecting) => onSelectNode(row, rowIndex, isSelecting),
+                          isSelected: isNodeSelected(row)
+                        }}
+                      />
+                      <Td dataLabel={columns[0]}>
+                        <div>{row[0]}</div>
+                        <a href="#">siemur/test-space</a>
+                      </Td>
+                      <Td dataLabel={columns[1]}>
+                        <CodeBranchIcon key="icon" /> {row[1]}
```
mcoker, line 2444: Just a nit, but looks like this whitepsace character between the icon and text is getting removed. Is there a way to preserve that? I suppose we could use `&nbsp;`, though not sure if there are downsides to that. We could also use a flex layout with an `xs` or `sm` spacer - I think an `sm` spacer looks better there anyways.

@mcarrano wdyt? This first cell has a small spacer between the icon and number after it. The other cells just use a whitespace (single space) character between the icon and number.

Screen Shot 2022-05-11 at 5 16 27 PM

packages/react-table/src/demos/Table.md
```diff
@@ -2397,52 +2394,77 @@ ComposableTableSortable = () => {
...
+                        sortBy: {
+                          index: activeSortIndex,
+                          direction: activeSortDirection
+                        },
+                        onSort,
+                        columnIndex
+                      }
+                    };
+                    return (
+                      <Th key={columnIndex} {...sortParams}>
```
mcoker, line 2421: The core demo starts off with the first column header `<th>` marked as sorted. The react demo content is also visually sorted by the first column, but the column header isn't marked as sorted.

mcoker (approved):
This LGTM, left a few comments but none are blockers.

packages/react-table/src/demos/Table.md
mcarrano: Ah, I see what you mean @mcoker Which do you feel is more correct? Agree that these should be consistent. We talked about creating a common data set for all these demos so we shouldn't need to be implementing one-off fixes for things like this.

mcoker: @mcarrano since the core and react examples can handle whitespace differently, it's probably best practice to use a layout for these icons. It also gives us control over the spacing, alignment, etc. For example, we use a flex layout  for icons + text in the card demos

Screen Shot 2022-05-12 at 9 05 25 AM

packages/react-table/src/demos/Table.md
jenny-s51: Nice catch - updated!

packages/react-table/src/demos/Table.md
jenny-s51: Using a `Flex` layout here is a great idea @mcoker ! I've updated this code to wrap the row data and icons with `Flex` using a `sm` spacer as you suggested/preferred.

packages/react-core/src/demos/examples/DashboardHeader.js
```diff
@@ -127,6 +128,11 @@ export default class DashboardHeader extends React.Component {
                 <AttentionBellIcon />
               </Button>
             </ToolbarItem>
+            <ToolbarItem>
+              <Button aria-label="Notifications" variant={ButtonVariant.plain}>
```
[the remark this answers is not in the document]
jenny-s51: I've updated these button definitions to use the structure you specified 👍

packages/react-table/src/demos/Table.md
mcoker: It's so pretty now!

packages/react-core/src/demos/examples/DashboardHeader.js
```diff
@@ -123,20 +124,17 @@ export default class DashboardHeader extends React.Component {
             spacer={{ default: 'spacerNone', md: 'spacerMd' }}
           >
             <ToolbarItem>
-              <Button aria-label="Notifications" variant={ButtonVariant.plain}>
-                <AttentionBellIcon />
-              </Button>
+              <Button aria-label="Notifications" variant={ButtonVariant.plain} icon={<AttentionBellIcon />} />
+            </ToolbarItem>
+            <ToolbarItem>
+              <Button aria-label="App launcher" variant={ButtonVariant.plain} icon={<ThIcon />} />
```
mcoker, line 130: In core we use the app launcher component here - that seems to make sense. Though I suspect that should be a menu toggle/menu combo?

packages/react-core/src/demos/examples/DashboardHeader.js
```diff
@@ -123,20 +124,17 @@ export default class DashboardHeader extends React.Component {
...
             </ToolbarItem>
             <ToolbarGroup variant="icon-button-group" visibility={{ default: 'hidden', lg: 'visible' }}>
               <ToolbarItem>
-                <Button aria-label="Settings actions" variant={ButtonVariant.plain}>
-                  <CogIcon />
-                </Button>
+                <Button aria-label="Settings actions" variant={ButtonVariant.plain} icon={<CogIcon />} />
```
mcoker, line 134: This is a dropdown in core. I don't know that the "actions" text is helpful either, seems like "Settings" is fine and that's what it is in core.

packages/react-core/src/demos/examples/DashboardHeader.js
```diff
@@ -123,20 +124,17 @@ export default class DashboardHeader extends React.Component {
...
               </ToolbarItem>
               <ToolbarItem>
-                <Button aria-label="Help actions" variant={ButtonVariant.plain}>
-                  <QuestionCircleIcon />
-                </Button>
+                <Button aria-label="Help actions" variant={ButtonVariant.plain} icon={<QuestionCircleIcon />} />
```
mcoker, line 137: This is also a dropdown in core, and label is "Help".

packages/react-core/src/demos/examples/DashboardHeader.js
```diff
@@ -123,20 +124,17 @@ export default class DashboardHeader extends React.Component {
             spacer={{ default: 'spacerNone', md: 'spacerMd' }}
           >
             <ToolbarItem>
-              <Button aria-label="Notifications" variant={ButtonVariant.plain}>
-                <AttentionBellIcon />
-              </Button>
+              <Button aria-label="Notifications" variant={ButtonVariant.plain} icon={<AttentionBellIcon />} />
```
mcoker, line 127: Can we just use the regular bell icon here? I'm pretty sure this would need to turn red (`variant="attention"`) if the attention bell icon is used.

mcoker (approved):
Looks super duper! Double approved. Though I did notice a couple of things with the actions in the header - I think we can remove "actions" from 2 of the buttons' aria-labels, and all but the notification bell have menus - I'm assuming we should make menus for those?

Also with the notification badge, we should just use the default bell, as the attention bell is intended to convey state. And I'm assuming we want that button to toggle a notification drawer?

I'm not sure if some of these issues (maybe making menus out of the actions and implementing a notification drawer?) are part of separate efforts? Just things I noticed so wanted to mention them. Again, nothing is blocking - it's all pre-existing and dashboard wrapper stuff unrelated to the table.

tlabaj:
I agree Jenny updating the wrapper is outside the scope of this PR.  I have created this issue #7429

mmenestr (approved):

mcoker (changes requested):
Looks like the sort icon is hidden on xs (< sm breakpoint) viewports - I think this line can just be updated to `{{ md: 'hidden' }}`?

https://github.com/patternfly/patternfly-react/blame/1e7104949ff012ac1246d2eb65250131458b653e/packages/react-table/src/demos/Table.md#L2282

mcoker (approved):
LGTM! Great job on this!!

tlabaj (approved):
Looks great Jenny!

PR #7293, opened 2022-04-22
docs(Modal): add example preserving a11y when content overflows

### LINKED ISSUE #7079: Add example for Modal to show how to add tabindex when there is overflow
Add example for accessible overflow in Modal

wise-king-sullyman:
**What**: Closes #7079

**Additional issues**:

Convenience link to the added example: https://patternfly-react-pr-7293.surge.sh/components/modal/#with-overflowing-content

### DIFF
--- a/packages/react-core/src/components/Modal/ModalContent.tsx
+++ b/packages/react-core/src/components/Modal/ModalContent.tsx
@@ -47,6 +47,10 @@ export interface ModalContentProps extends OUIAProps {
   'aria-label'?: string;
   /** Id of Modal Box description */
   'aria-describedby'?: string;
+  /** Accessible label applied to the modal box body */
+  bodyAriaLabel?: string;
+  /** Accessible role applied to the modal box body */
+  bodyAriaRole?: string;
   /** Flag to show the close button in the header area of the modal */
   showClose?: boolean;
   /** Default width of the content. */
@@ -82,6 +86,8 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
   'aria-label': ariaLabel = '',
   'aria-describedby': ariaDescribedby,
   'aria-labelledby': ariaLabelledby,
+  bodyAriaLabel,
+  bodyAriaRole,
   showClose = true,
   footer = null,
   actions = [],
@@ -120,10 +126,17 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
     actions.length > 0 && <ModalBoxFooter>{actions}</ModalBoxFooter>
   );
 
+  const defaultModalBodyAriaRole = bodyAriaLabel ? 'region' : undefined;
+
   const modalBody = hasNoBodyWrapper ? (
     children
   ) : (
-    <ModalBoxBody {...props} {...(!description && !ariaDescribedby && { id: descriptorId })}>
+    <ModalBoxBody
+      aria-label={bodyAriaLabel}
+      role={bodyAriaRole || defaultModalBodyAriaRole}
+      {...props}
+      {...(!description && !ariaDescribedby && { id: descriptorId })}
+    >
       {children}
     </ModalBoxBody>
   );
--- a/packages/react-core/src/components/Modal/ModalContent.tsx
+++ b/packages/react-core/src/components/Modal/ModalContent.tsx
@@ -47,9 +47,9 @@ export interface ModalContentProps extends OUIAProps {
   'aria-label'?: string;
   /** Id of Modal Box description */
   'aria-describedby'?: string;
-  /** Accessible label applied to the modal box body */
+  /** Accessible label applied to the modal box body. This should be used to communicate important information about the modal box body div if needed, such as that it is scrollable */
   bodyAriaLabel?: string;
-  /** Accessible role applied to the modal box body */
+  /** Accessible role applied to the modal box body. This will default to region if a body aria label is applied. Set to a more appropriate role as applicable based on the modal content and context */
   bodyAriaRole?: string;
   /** Flag to show the close button in the header area of the modal */
   showClose?: boolean;
--- a/packages/react-core/src/components/Modal/examples/Modal.md
+++ b/packages/react-core/src/components/Modal/examples/Modal.md
@@ -869,3 +869,10 @@ class HelpModal extends React.Component {
 ```ts file="ModalWithForm.tsx"
 
 ```
+
+### With overflowing content
+
+If the content that you're passing to the modal is likely to overflow the modal content area, pass `tabIndex={0}` to the modal. Failing to do so will cause the modal content to not be scrollable via keyboard. 
+
+```ts file="ModalWithOverflowingContent.tsx"
+```
--- a/packages/react-core/src/components/Modal/examples/Modal.md
+++ b/packages/react-core/src/components/Modal/examples/Modal.md
@@ -872,7 +872,7 @@ class HelpModal extends React.Component {
 
 ### With overflowing content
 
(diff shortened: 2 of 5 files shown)


packages/react-core/src/components/Modal/examples/Modal.md
```diff
@@ -869,3 +869,10 @@ class HelpModal extends React.Component {
 ```ts file="ModalWithForm.tsx"
 
 ```
+
+### With overflowing content
+
+If the content that you're passing to the modal is likely to overflow the modal content area, pass `tabIndex={0}` to the modal. Failing to do so will cause the modal content to not be scrollable via keyboard.
```
tlabaj, line 875: @wolfeallison could you please review the copy here.

wolfeallison: I think we could be a little more explicit on what "failing to do so" is. Not sure if this is technically correct, but what about "Failing to pass will cause the modal content to not be scrollable via keyboard." Otherwise, looks great!

wise-king-sullyman: @wolfeallison what do you think about rewording that whole line to just be:

"If the content that you're passing to the modal is likely to overflow the modal content area, pass `tabIndex={0}` to the modal to enable keyboard accessible scrolling."

packages/react-core/src/components/Modal/examples/ModalWithOverflowingContent.tsx
```diff
@@ -0,0 +1,70 @@
...
+    setIsModalOpen(prevIsModalOpen => !prevIsModalOpen);
+  };
+
+  return (
+    <React.Fragment>
+      <Button variant="primary" onClick={handleModalToggle}>
+        Show modal
+      </Button>
+      <Modal
+        tabIndex={0}
```
nicolethoen, line 17: Recently, @seanforyou23 made a comment on some a11y docs about this sort of update and suggested adding an accessible label. This is because the non-negative tabindex will make the container register as a form-control on rotor menus or elements lists and without an explicit label, various browsers or screen readers may label it differently/incorrectly.
Comment: https://github.com/patternfly/patternfly-org/pull/2893#discussion_r841910449

That said, I dont think we need to update the example docs - I am adding this suggestion to the a11y docs. But maybe we should add an aria-label or aria-labelledby to elements we are intentionally adding a tabindex to?

wise-king-sullyman: Makes sense. I'm on board with adding an aria-label here.

jessiehuff: I agree with the suggestion of adding a label, but I wonder if we would need to add a role in that scenario as well? I think technically just having an `aria-label` on a `<div>` will throw an Axe error since this approach is not well supported. For example, we actually have [this issue ](https://www.patternfly.org/v4/components/wizard/react/basic/)right now on our Wizard. I'm not sure which role would make sense in this scenario though. We could do a basic "region" role, but there might be [a better sectioning role](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#content_sectioning). It might require a little bit of research to see what the best approach is to add a label.

wise-king-sullyman: @jessiehuff from the research I've done so far I think the ideal role might depend on what the modal is for / what the content in the modal is.

For that reason I think the best approach here is probably to add another prop that the consumer can use to apply a more suitable role if applicable, but default to having it as a "region" if an aria-label is applied to ensure that (at the least) Axe errors aren't caused. For a little additional context from my research [Carbon](https://carbondesignsystem.com/components/modal/usage#live-demo) sets the role to "region" in their modal if it's scrollable. Thoughts?

wise-king-sullyman: Also @nicolethoen the modal currently applies passed aria-labels to the overall modal box rather than letting them go to the modal box body where the content is / the tab index is applied, so another prop will need to be added for consumers to be able to apply labels to the modal box body.

nicolethoen (approved):
I think this looks great!

thatblindgeye (approved):

kmcfaul (approved):

jessiehuff (approved):
LGTM! Thanks for your work on this! 😄

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -120,10 +126,17 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
...
   );
 
+  const defaultModalBodyAriaRole = bodyAriaLabel ? 'region' : undefined;
+
   const modalBody = hasNoBodyWrapper ? (
     children
   ) : (
-    <ModalBoxBody {...props} {...(!description && !ariaDescribedby && { id: descriptorId })}>
+    <ModalBoxBody
+      aria-label={bodyAriaLabel}
```
tlabaj, line 135: Should the aria-label have a default?

packages/react-core/src/components/Modal/ModalContent.tsx
```diff
@@ -120,10 +126,17 @@ export const ModalContent: React.FunctionComponent<ModalContentProps> = ({
...
+      role={bodyAriaRole || defaultModalBodyAriaRole}
```
tlabaj, line 136: So if the consumer defines an `aria-label` but not an `aria-role`, the role defaults to 'region'?

tlabaj (changes requested):

packages/react-core/src/components/Modal/ModalContent.tsx
wise-king-sullyman: I'm not sure. If it should what would you want the default to be?

packages/react-core/src/components/Modal/ModalContent.tsx
wise-king-sullyman: Correct the role defaults to 'region' if the consumer passes an aria-role to prevent Axe violations.

packages/react-core/src/components/Modal/ModalContent.tsx
jessiehuff: Hmm this is an interesting question and actually makes me also question the default role. If we're going to default the role then I think we should default an `aria-label`, but do we always want the content of a model to have a role and an `aria-label`? When we were talking about it before, I admittedly was just thinking about the overflow example, but I'm not sure if it always makes sense to have both? For example, it seems a little heavy handed if the modal content is a sentence of information, and we have it segmented out with a role and label.

wise-king-sullyman: I'm not opposed to adding a default label at all, I just have no idea what a good default would be. I think that any good label here is going to be highly variable based on what the content of the modal is.

Combined with the fact that we currently have the modal itself labeled by the modal heading, and described by the modal body, I'm not sure if having a default would be for the best? I'm not an expert on this topic by any means though and am happy to proceed however you all think is best on it 🙂

tlabaj: I will defer to @jessiehuff here.
Jessie, are you saying that the role should only default to `region` if there is an overflow?  Should this props be tied to that having overflow?

jessiehuff: I think I like Austin's approach of defaulting the region when there's an `aria-label` to prevent a11y issues, but I'm leaning towards not having a default `aria-label`. The only case that I can see that being helpful is in the case of the overflow so perhaps it's better to just allow users to customize it when they'd like.

tlabaj: it might be beneficial to mention in the prop description when it should be used?

tlabaj (approved):

PR #7294, opened 2022-04-22
Topology dark theme

### LINKED ISSUE #7181: Topology - update to support dark theme
Topology styling should be updated to ensure dark theme compatibility.  Design specs available here:
- https://marvelapp.com/prototype/i195ic1/screen/85808125

srambach:
Adds dark styling adjustments.
Closes #7181 

Note:
As part of this PR, I added Patternfly-style variables for all color variables, since a large number of them needed to be overridden in dark theme. However, it does not exactly match Patternfly conventions because the associated classnames are slightly different that the Patternfly convention, and it seemed better to match the class names for clarity.
Also, the edges already had a variable system in place, so I did not bring this into the same convention either, but instead left it as is and initialized its values with values matching the convention above.

Exceptions to dark styling as spec'd in https://marvelapp.com/prototype/i195ic1/screen/85808125:
- Node badge colors are currently set directly on the rect, not in CSS, so this will need to be addressed in a future issue.
- Compatible/incompatible connectors are not currently supported, so this has not been implemented.
- Shadows on the critical default nodes are hard coded in the react code, so these were not addressed here.

### DIFF
--- a/packages/react-styles/src/css/components/Topology/topology-components.css
+++ b/packages/react-styles/src/css/components/Topology/topology-components.css
@@ -39,6 +39,10 @@
   --pf-topology__node__label__text--m-secondary--Fill: var(--pf-global--secondary-color--100);
 
   --pf-topology__node--m-selected--node__label__text--Fill: var(--pf-global--Color--light-100);
+  --pf-topology__node--m-selected--m-info--node__label__text--Fill: var(--pf-global--Color--light-100);
+  --pf-topology__node--m-selected--m-danger--node__label__text--Fill: var(--pf-global--Color--light-100);
+  --pf-topology__node--m-selected--m-warning--node__label__text--Fill: var(--pf-global--Color--light-100);
+  --pf-topology__node--m-selected--m-success--node__label__text--Fill: var(--pf-global--Color--light-100);
 
   --pf-topology__node__label__background--Fill: var(--pf-global--BackgroundColor--100);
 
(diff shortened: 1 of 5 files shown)


jeff-phillips-18 (approved):
LGTM. Thanks!

mceledonia (approved):
Looks good thanks Sarah!

andrew-ronaldson:
Thanks Sarah. Good work

tlabaj (approved):

packages/react-styles/src/css/components/Topology/topology-components.css
```diff
@@ -1,13 +1,250 @@
...
+  --pf-topology__group__label__node__label__background--StrokeWidth: 2px;
+  --pf-topology__group__label__node__label__background--Stroke: var(--pf-global--palette--black-300);
+  --pf-topology__group--m-hover--label__node__label__background--Stroke: var(--pf-global--palette--black-100);
+  --pf-topology__group--m-selected--label__node__label__background--Stroke: var(--pf-global--primary-color--300);
+
+  --pf-topology__group--m-selected--group__label__node__label__background--Fill: var(--pf-global--primary-color--300);
+
+  /* dark edge */
+  --pf-topology__edge--HoverStroke: var(--pf-global--palette--black-300);
+  --pf-topology__edge--m-hover--background--Stroke: var(--pf-global--palette--black-600);
```
mcoker, line 224: Could these use `border-color--100` & `200`?

PR #7297, opened 2022-04-25
feat(Tabs): allow dynamic close/add

### LINKED ISSUE #7199: Tabs- support dynamic tabs
Want to add the option to add and remove tabs from a set of horizontal tabs. Clicking the Add button adds a new tab to the end of the list. Clicking remove deletes the tab and shifts following tabs to the left.

Core issue/request: https://github.com/patternfly/patternfly/issues/4757

kmcfaul:
**What**: Closes #7199

Adds `onClose` and `onAdd` callbacks to `Tabs`. Adding these props will add their respective buttons.
Adds `addAriaLabel` and `closeAriaLabel` to `Tabs`/`Tab` to control aria label of buttons.


packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -102,10 +109,29 @@ const TabBase: React.FunctionComponent<TabProps> = ({
...
+        eventKey === localActiveKey && styles.modifiers.current,
+        handleTabClose && styles.modifiers.action,
+        childClassName
+      )}
       role="presentation"
     >
       {tooltip ? <Tooltip {...tooltip.props}>{tabButton}</Tooltip> : tabButton}
+      {handleTabClose !== undefined && (
+        <span className={css(styles.tabsItemClose)}>
+          <button
```
mcoker, line 123: Just curious - why not import and use `<Button>`?

packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -49,6 +53,7 @@ const TabBase: React.FunctionComponent<TabProps> = ({
   href,
   innerRef,
   tooltip,
+  closeAriaLabel,
```
mcoker, line 56: Should this have a default or warning if not set since we render the icon (meaning it will always be an icon - no text - I'm assuming)? Same for the add button `aria-label`. Not sure if that's a best practice or not.

packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -102,10 +109,29 @@ const TabBase: React.FunctionComponent<TabProps> = ({
 
   return (
     <li
-      className={css(styles.tabsItem, eventKey === localActiveKey && styles.modifiers.current, childClassName)}
+      className={css(
+        styles.tabsItem,
+        eventKey === localActiveKey && styles.modifiers.current,
+        handleTabClose && styles.modifiers.action,
```
mcoker, line 115: If a `.pf-m-action` tab is `disabled` or `.pf-m-aria-disabled`, this `<li>` needs the class `.pf-m-disabled`. This turns the entire item (tab and close button) grey, instead of just the tab which you will see if you disable a tab in the PR preview.

Also the core examples show the close buttons as enabled when a tab is disabled, meaning it's possible to have a disabled tab with either an enabled or disabled close button. Maybe @mcarrano or @mmenestr can provide clarification if that's valid, and if we want to update our examples to show a best practice. Assuming it's valid, I think we'll need a way to handle the disabled state of the close button in the react component.

mcoker (changes requested):
Looks great! Left a few comments for review. Also it looks like adding/removing tabs doesn't trigger the detection of overflow in the tab list and toggle the scroll buttons - should it?

packages/react-core/src/components/Tabs/Tab.tsx
mcarrano: That's a really good question, and I'm not sure what's correct here. @mmenestr do you have thoughts? Maybe it's best to keep this flexible to allow for different use cases. There might be situations where you'd want to remove a disabled tab. Depends why the tab is disabled.

mmenestr: Hmm yeah that's a good question. I think I agree with you @mcarrano, there could be a use case for being able to dismiss a disabled tab, so we should probably keep that option.

packages/react-core/src/components/Tabs/Tab.tsx
kmcfaul: Ah you're right, I think was following the HTML a bit too closely. Will update that.

packages/react-core/src/components/Tabs/Tab.tsx
kmcfaul: Added an additional prop to control `isDisabled` for the close button separately.

kmcfaul:
Updated to refine how tabs are selected when new tabs are added/removed and reducing the starting number of tabs in the demo. The scroll buttons will pop up after overflowing, but only after you scroll the tabs manually first, and they don't seem to disappear when the tabs shrink enough. I'm not sure about a way to force the scroll buttons to update, any thoughts @tlabaj @nicolethoen?

Still working on getting the newly added tab focused when it overflows.

mcarrano:
This is looking much better @kmcfaul ! In navigating this with the keyboard, I noticed something weird about how the focus rectangle gets drawn around the Add button.

Screen Shot 2022-04-28 at 2 25 28 PM

This is Firefox. @mcoker ideas?

mcoker (changes requested):
Looking good! Left a couple of comments about aria-labels. Also if you don't pass `closeAriaLabel`, the autogenerated one is `aria-label="[object Object] close button"`.

Screen Shot 2022-05-02 at 8 01 02 PM

There are a couple of edge cases where disabled tabs are made active. I'm assuming we would want to check first and not make a disabled tab active?

* If you add a disabled tab (why would you do that? I dunno...) it makes the tab current and turns the text black. I added a core issue to fix the text color - https://github.com/patternfly/patternfly/issues/4824
* If you have a disabled tab and a non-disabled/active tab, and close the active tab, the disabled tab is made active. 

I'm assuming in both cases that we don't want to make a tab current if its disabled?

kmcfaul:
I'm guessing with the disabled tab issue would have to be handled in the demo code by expanding the tabs data array to track the disabled state of each tab, versus using an index (because that will dynamically change when something is removed). It should be possible without additional changes to the component.

Will update the aria labels.

nicolethoen (commented):
Couple quirks i've noticed:
- I think that the number used in the new tab name should be something that is incrementing everytime i click the add button. Right now, if I have tabs 1 - 4 and delete tab three, then add a new tab, it creates another tab 4 I'd expect it to add a tab 5.
- @mcarrano @mcoker should the cursor change to the 'no action' cursor when hovering over a disabled tab button?

mcarrano:
> @mcarrano @mcoker should the cursor change to the 'no action' cursor when hovering over a disabled tab button?

Good question. Are we doing that in other places for disabled elements? I can't remember, but think we should be consistent.

mcoker:
@mcarrano it's come up before and we aren't able to implement it consistently because with a lot of `disabled` (things a browser can disable - basically form elements that support the `disabled` attribute - buttons, inputs, etc) and "disabled" (things we disable manually that aren't disabled by the browser, such as links) elements we use CSS to disable click/hover/etc mouse interactions with the element via `pointer-events: none`, and that disables a cursor change on hover since we're basically disabling hovering the thing using `pointer-events`.

Here is an example with the cursor set to `not-allowed` on different disabled things - https://codepen.io/mcoker/pen/rNJVYZO

"Disabled button" uses the `disabled` attribute and "Disabled link" uses `.pf-m-disabled` - both of these we disable mouse interaction via `pointer-events: none;` so the cursor doesn't change.

"ARIA disabled", however, preserves mouse interactions (doesn't use `pointer-events: none`) so that you can trigger a tooltip or popover on the element when hovered, so you see the cursor change there.

If we wanted to universally set the cursor on disabled things across the library, we would likely need to start with enabling `pointer-events` on these elements, then looking at other ways of keeping the user from interacting with them.

mcarrano:
Thanks for the detailed explanation of that @mcoker . In that case I'm OK with not having the cursor change for disabled tabs.

packages/react-core/src/components/Tabs/Tab.tsx
```diff
@@ -33,6 +35,10 @@ export interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement | HTMLB
   innerRef?: React.Ref<any>;
   /** Optional Tooltip rendered to a Tab. Should be <Tooltip> with appropriate props for proper rendering. */
   tooltip?: React.ReactElement<any>;
+  /** Aria-label for the close button added by passing the onClose property to Tabs. */
+  closeAriaLabel?: string;
```
tlabaj, line 39: Can we call this `closeButtonAriaLabel` instead.  Other components have this.

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -31,6 +32,12 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   defaultActiveKey?: number | string;
   /** Callback to handle tab selection */
   onSelect?: (event: React.MouseEvent<HTMLElement, MouseEvent>, eventKey: number | string) => void;
+  /** Callback to handle tab closing */
+  onClose?: (event: React.MouseEvent<HTMLElement, MouseEvent>, eventKey: number | string) => void;
+  /** Callback for the add button. Passing this property inserts the add button */
+  onAdd?: () => void;
+  /** Aria-label for the add button */
+  addAriaLabel?: string;
```
tlabaj, line 40: `addButtonAriaLabel`

tlabaj (changes requested):

tlabaj:
@kmcfaul there is an error when you remove that last tab. 
@mcarrano what would you expect when the last tab is remove?  I would expect to still see the add button.

mcarrano:
> @mcarrano what would you expect when the last tab is remove? I would expect to still see the add button.

Thanks a great question. Yes, at a minimum the user should be able to add new tabs, but I think this is also highly context dependent. I could see dropping the user into some type of empty state, for example. Can we just attach this to a callback function to also the consumer to handle this as they see fit? @christianvogt I'm curious what you you would expect to in your terminal view if the user deletes the last remaining tab.

tlabaj (changes requested):
@kmcfaul we spoke about the last at being removed at the leads meeting today.  WE talked about how this might be a product/context specific  implementation decision.  For the purposes our examples, it was proposed that we have the last  tab not closable.

packages/react-core/src/components/Tabs/examples/Tabs.md
```diff
@@ -1498,3 +1615,10 @@ class ToggledSeparateContent extends React.Component {
   }
 }
 ```
+
+### Dynamic
+
+To enable closeable tabs, pass the `onClose` property to `Tabs`, and to enable the add button, pass the `onAdd` property to `Tabs`. Aria labels may be controlled manually by passing `closeAriaLabel` to `Tab` and `addAriaLabel` to `Tabs`.
```
tlabaj, line 1621: the prop names need to be updated here.

mcarrano (approved):
Looks great @kmcfaul ! Thanks for making these changes to the example.

packages/react-core/src/components/Tabs/examples/TabsDynamic.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }, [tabs]);
+
+  return (
+    <Tabs
+      activeKey={activeTabKey}
+      onSelect={(event: any, tabIndex: string | number) => setActiveTabKey(tabIndex as number)}
+      onClose={onClose}
+      onAdd={onAdd}
+      aria-label="Tabs in the addable/closeable example"
+      addButtonAriaLabel="Add tab"
```
mcoker, line 48: Since this is the same as the default `aria-label`, should we consider making it unique or removing it?

mcoker (approved):
LGTM! Just one question but it isn't a blocker.

christianvogt:
If you create many tabs until the scroll buttons appear, then select the first tab and start to remove the tabs. You can get into a situation where the buttons remain but aren't necessary.

image

christianvogt:
@kmcfaul follow these steps:
 1. click the `+` button 5 times
 2. scroll all the way to the left
 3. select tab 1
 4. close tab 1
 5. continue closing the first until 1 remains

kmcfaul:
@christianvogt Added some extra logic when children change, this should improve the behavior for scroll buttons being added/removed when the build is done.

nicolethoen (approved):
LGTM!

tlabaj:
@christianvogt can you please verify the update looks good to you.

christianvogt:
The issue I observed has been fixed.

gitdallas (approved):
looks like you have some unused consts in the tests but otherwise lgtm

wise-king-sullyman (approved):
I think that some tests to ensure that the added callbacks and aria-label props function as expected (i.e. testing that the `onClose` callback fires when that tab's close button is clicked) would be beneficial here. I won't block this PR for them though.

Everything else looks great! 🚀

Issue #7307, opened 2022-04-27
Topology - Add cola group layout


jeff-phillips-18:
What is the existing component, if any?
Topology

**Describe the feature**
Add a topology layout that uses WebCola but handles lots of groups and subgroups well.

**Any other information?**
WebCola gets slow when there are lots of nodes

PR #7308, opened 2022-04-27
feat(Topology): Add ColaGroupsLayout

jeff-phillips-18:
**What**:
Closes https://github.com/patternfly/patternfly-react/issues/7307

**Additional issues**:
One known issue with the layout stacking vertically. This should be addressed in a follow up.


nicolethoen (approved):
I tested the demo-app and don't see any errors in the console, and I don't know exactly what to expect when playing with a cola layout, but I did play with it for a bit. Seems smooth.

PR #7311, opened 2022-04-28
feat(tabs): add secondary border-bottom variation, update demos

### LINKED ISSUE #7259: Tabs - add border-bottom variation for secondary tabs, update demo
Issue: https://github.com/patternfly/patternfly/issues/4753
PR: https://github.com/patternfly/patternfly/pull/4774

* Adds `.pf-m-border-bottom` to add a bottom border to secondary tabs.
* Updates secondary tabs in "Open tabs with secondary tabs" demo
  * Core: https://pf4.patternfly.org/components/tabs/html-demos/open-tabs-with-secondary-tabs/
  * React: https://www.patternfly.org/v4/components/tabs/react-demos#open-tabs-with-secondary-tabs
* Also see if we can match the structure in the core demo. We need to remove the extra space between the primary and secondary tabs, and make sure the secondary tabs extends to the edge of the container with an inset, so the border goes to the edge of the container.
  * If we can't match the core structure, you should be able to do that with the current example by:
    * Add `.pf-m-no-padding` to page section that wraps secondary tabs and tab content
    * Add `.pf-m-page-insets` to secondary tabs (and remove `.pf-m-inset-none`)
    * Add `.pf-m-padding` to the `.pf-c-tab-content__body` element that wraps the "Pod details" content
    * Side note - the first tab content element in this demo is in it's own div, and the others are in another, and those divs are wrapped in a flex layout - could the flex layout just be removed and the tab content elements all be adjacent to one another? Only one should show at a time.

wise-king-sullyman:
**What**: Closes #7259

**Additional issues**:

Convenience link to the updated demo:
https://patternfly-react-pr-7311.surge.sh/components/tabs/react-demos/open-tabs-with-secondary-tabs/

### DIFF
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -41,7 +41,7 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   isBox?: boolean;
   /** Enables vertical tab styling */
   isVertical?: boolean;
-  /** Enables no border bottom tab styling */
+  /** Enables border bottom tab styling on tabs. Defaults to true. To remove the bottom border, set this prop to false. */
   hasBorderBottom?: boolean;
   /** Enables border bottom styling for secondary tabs */
   hasSecondaryBorderBottom?: boolean;
--- a/packages/react-core/src/components/Tabs/Tabs.tsx
+++ b/packages/react-core/src/components/Tabs/Tabs.tsx
@@ -43,6 +43,8 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   isVertical?: boolean;
   /** Enables no border bottom tab styling */
   hasBorderBottom?: boolean;
+  /** Enables border bottom styling for secondary tabs */
+  hasSecondaryBorderBottom?: boolean;
   /** Aria-label for the left scroll button */
   leftScrollAriaLabel?: string;
   /** Aria-label for the right scroll button */
@@ -292,6 +294,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
       isVertical,
       isBox,
       hasBorderBottom,
+      hasSecondaryBorderBottom,
       leftScrollAriaLabel,
       rightScrollAriaLabel,
       'aria-label': ariaLabel,
@@ -361,6 +364,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
             showScrollButtons && !isVertical && styles.modifiers.scrollable,
             usePageInsets && styles.modifiers.pageInsets,
             !hasBorderBottom && styles.modifiers.noBorderBottom,
+            hasSecondaryBorderBottom && styles.modifiers.borderBottom,
             formatBreakpointMods(inset, styles),
             variantStyle[variant],
             className
--- a/packages/react-core/src/components/Tabs/__tests__/Tabs.test.tsx
+++ b/packages/react-core/src/components/Tabs/__tests__/Tabs.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { Tabs } from '../Tabs';
 import { Tab } from '../Tab';
 import { TabTitleText } from '../TabTitleText';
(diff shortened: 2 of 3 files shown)


packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -43,6 +43,8 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   isVertical?: boolean;
   /** Enables no border bottom tab styling */
   hasBorderBottom?: boolean;
+  /** Enables border bottom styling for secondary tabs */
+  hasSecondaryBorderBottom?: boolean;
```
nicolethoen, line 47: Is a new prop necessary? if the consumer specifies isSecondary and hasBorderBottom, could it do the same thing?

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -361,6 +364,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
             showScrollButtons && !isVertical && styles.modifiers.scrollable,
             usePageInsets && styles.modifiers.pageInsets,
             !hasBorderBottom && styles.modifiers.noBorderBottom,
+            hasSecondaryBorderBottom && styles.modifiers.borderBottom,
```
mcoker, line 367: would this work to keep a single prop?

```suggestion
            !isSecondary && !hasBorderBottom && styles.modifiers.noBorderBottom,
             isSedoncary && hasBorderBottom && styles.modifiers.borderBottom,
```

mcoker (commented):
LGTM! Just a couple of comments.

packages/react-core/src/components/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -350,3 +350,42 @@ test('should render tabs with no bottom border', () => {
   expect(asFragment()).toMatchSnapshot();
 });
 
+test('should not render tabs with secondary border bottom when not passed hasSecondaryBorderBottom', () => {
+  render(
+    <Tabs id="noBottomBorderTabs" aria-label="secondary bottom border">
```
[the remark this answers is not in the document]
wise-king-sullyman: It isn't consumer or user facing so it shouldn't make a big difference IMO. I'm not against making these sentence case though.

packages/react-core/src/components/Tabs/Tabs.tsx
wise-king-sullyman: I *think* that would still cause breaking changes since `hasBorderBottom` defaults to true, so it would require consumers to add `hasBorderBottom={false}` if they want to keep their secondary tabs without a bottom border.

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -43,6 +43,8 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   isVertical?: boolean;
   /** Enables no border bottom tab styling */
   hasBorderBottom?: boolean;
```
jenny-s51, line 45:
```suggestion
  hasNoBorderBottom?: boolean;
```

Based on the prop description, I'm guessing you tested a `hasNoBorderBottom` prop, which is what we want based on the core docs: https://patternfly-pr-4774.surge.sh/components/tabs/ 

The tabs have a bottom border by default, so it makes sense to have a prop that lets a consumer remove the bottom border.

jenny-s51 (changes requested):
These changes are looking good @wise-king-sullyman! I did a lot of investigating here and have suggestions for changes that will help us get this merged without breaking anything! These changes should ease the breaking change updates down the road as well. Here's what we'd have to do:

1. Add  a `hasNoBorderBottom` prop. `Tabs` (`.pf-c-tabs`) already have a bottom border by default, so instead of `hasBorderBottom` we'd want `hasNoBorderBottom` which will apply the CSS class `.pf-m-no-border-bottom` to the tabs. 
2. Set `hasSecondaryBorderBottom` to `false` under `defaultProps`. Based on the [core docs](https://patternfly-pr-4774.surge.sh/components/tabs/) `isSecondary` i.e. (`pf-m-secondary`) styles the tabs with *no* border. So when a user sets `hasSecondaryBorderBottom` it should add  the `.pf-m-border-bottom` to the secondary variant of Tabs.

The stuff we'd have left to do in a breaking change release would be to either:
1. Delete the `hasBorderBottom` prop completely (functionality is replaced by `hasNoBorderBottom`) OR
2. Rename `hasSecondaryBorderBottom` to `hasBorderBottom` and it would serve the same purpose -- based on the core docs the `.pf-m-border-bottom` class is only applied to **secondary tabs** anyway. Though `hasSecondaryBorderBottom` would likely make more sense to a consumer as it tells them directly what Tabs variant that prop should be used with.

WDYT @wise-king-sullyman ? @mcoker please comment if you have any thoughts - I think I've interpreted the core side correctly 😄

packages/react-core/src/components/Tabs/Tabs.tsx
wise-king-sullyman: Yeah `hasBorderBottom` was originally `hasNoBorderBottom` in #7183 when I first PR'd it. @kmcfaul @nicolethoen I'd appreciate your thoughts on this suggestion. I'm a fan but since you both advocated for `hasBorderBottom` I'm hesitant to change it before getting your input.

Thank you for putting this much thought into it!

packages/react-core/src/components/Tabs/Tabs.tsx
kmcfaul: Maybe we can update this with the next breaking change release? Assuming we want secondary tabs to also default to having a bottom border like default tabs.

packages/react-core/src/components/Tabs/Tabs.tsx
```diff
@@ -43,6 +43,8 @@ export interface TabsProps extends Omit<React.HTMLProps<HTMLElement | HTMLDivEle
   isVertical?: boolean;
   /** Enables no border bottom tab styling */
```
jenny-s51, line 44:
```suggestion
  /** Enables border bottom tab styling on tabs. Defaults to true. To remove the bottom border, set this prop to false. */
```

packages/react-core/src/components/Tabs/Tabs.tsx
jenny-s51: Ah okay, thank you for linking the comment thread @wise-king-sullyman and for sharing your input @kmcfaul !  

Avoiding the double negative makes sense. If that's the case then we would want to keep both `hasBorderBottom` and `hasSecondaryBorderBottom` for the styling to work as expected. I've added a suggestion to update the prop description which would clarify the use case of `hasBorderBottom` a bit better - WDYT @wise-king-sullyman?

wise-king-sullyman: That suggestion looks great to me @jenny-s51, thank you for doing it!

packages/react-core/src/components/Tabs/Tabs.tsx
wise-king-sullyman: @kmcfaul I think that would be ideal.

mcoker (approved):
LGTM!

nicolethoen (approved):

jenny-s51 (approved):
👍👍

tlabaj (approved):
LGTM

PR #7323, opened 2022-04-28
feat(wizard): add support for additional information panel (drawer)

### LINKED ISSUE #7258: Wizard - add support for additional Information panel (drawer)
Issue - https://github.com/patternfly/patternfly/issues/4611
PR - https://github.com/patternfly/patternfly/pull/4778

* Adds `.pf-m-light-200` to `.pf-c-page__main-wizard` (the page section that wraps an in-page wizard.
* Adds a "With drawer" wizard example and demo, and "With drawer, in page" wizard demo that uses `.pf-c-page__main-wizard.pf-m-light-200`

jenny-s51:
**What**: Closes #7258 

This PR adds functionality to wizard that allows a consumer to add a drawer via a new `hasDrawer` prop. React demo has been aligned with the core examples and React fullscreen demos have been added as well. 

Per accordance with @tlabaj , the functionality added to this PR adds a drawer that is always visible: further discussion and designs may be necessary regarding if/how we want to let a user toggle the drawer open/closed from within the wizard, and if we want to pass any props to `Wizard` that will allow a consumer to style the top-level `Drawer` component.


packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -18,6 +19,8 @@ export interface WizardStep {
   name: React.ReactNode;
   /** The component to render in the main body */
   component?: any;
+  /** The content to render in the drawer panel */
+  drawerPanelContent?: any;
```
nicolethoen, line 23: i've noticed that leaving this out of one of the steps appears to remove the drawer for that step, even when `hasDrawer` is true on the wizard! Not sure if its worth removing the drawer scaffolding in those cases so that drawers without drawerPanelContent defined for a step dont wrap the whole wizard in drawer components? It's probably not trivial, so as long as there are not interactive or style side effects to situations like that, then it's likely fine as is.

nicolethoen (commented):
I think that given how large the drawer appears in the new example, we likely do want to expose some props so that people can control the props/styles of the drawer - specifically drawer width.

packages/react-core/src/components/Wizard/examples/Wizard.md
```diff
@@ -859,3 +860,138 @@ class WizardInModal extends React.Component {
...
+        isExpanded: false
+      });
+    };
+  }
+
+  render() {
+    const { isOpen, isExpanded } = this.state;
+
+    const panel1Content = (
+      <DrawerPanelContent widths={{ default: 'width_25' }} colorVariant={DrawerColorVariant.light200}>
```
jenny-s51, line 922: Thank you @nicolethoen @mcarrano for your thorough reviews and helpful feedback! 

The width of the drawer panel is set through the `widths` prop of `DrawerPanelContent`, which a consumer can customize as shown in the demo here.

I've set the smallest available width as the `default` value, and the demo has been updated to reflect this!

mcarrano (approved):
Much better.  Thanks @jenny-s51 !

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -350,12 +368,16 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     const computedSteps: WizardStep[] = this.initSteps(steps);
     const firstStep = activeStep === flattenedSteps[0];
     const isValid = activeStep && activeStep.enableNext !== undefined ? activeStep.enableNext : true;
+    const DrawerWrapper: React.FunctionComponent<DrawerWrapperProps> = ({ condition, wrapper, children }) =>
```
tlabaj, line 371: Can you own this component into its own file.

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -41,7 +44,15 @@ export type WizardStepFunctionType = (
...
 
+export interface DrawerWrapperProps {
+  children: React.ReactElement;
+  condition: boolean;
+  wrapper: (children: React.ReactElement) => JSX.Element;
+}
+
 export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
+  /** Content rendered inside the wizard. */
+  children?: React.ReactNode;
```
tlabaj, line 55: I don't think you need this prop.

tlabaj (changes requested):

packages/react-core/src/components/Wizard/examples/Wizard.md
```diff
@@ -859,3 +860,138 @@ class WizardInModal extends React.Component {
   }
 }
 ```
+
+### Wizard with drawer
+
+```js isBeta
```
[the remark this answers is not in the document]
jenny-s51: Yes - good idea!

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -100,6 +111,8 @@ export interface WizardProps extends React.HTMLProps<HTMLDivElement> {
   isOpen?: boolean;
   /** Flag indicating nav items with sub steps are expandable */
   isNavExpandable?: boolean;
+  /** Flag indicating the wizard has a drawer */
```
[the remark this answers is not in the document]
jenny-s51: Good call - updated with a clearer description!

jenny-s51:
Thank you @tlabaj for your awesome suggestions - updated this PR based on your feedback!

mcoker (commented):
One comment about the need for `colorVariant` on `<DrawerContent>` - doesn't seem like we need that. 

I'm also seeing that the expandable navigation on small screens is hidden when there is a drawer. And if you shrink the screen all the way down and the drawer panel covers the drawer content (with the wizard in it), I'm not able to access anything in the wizard (nav, body, footer) because the panel covers everything.

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -462,31 +475,43 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
-                activeStep={activeStep}
-                nextButtonText={(activeStep && activeStep.nextButtonText) || nextButtonText}
-                backButtonText={backButtonText}
-                cancelButtonText={cancelButtonText}
-              />
+          <WizardDrawerWrapper
+            condition={hasDrawer && activeStep.drawerPanelContent}
+            wrapper={(children: React.ReactNode) => (
+              <Drawer isInline isExpanded>
+                <DrawerContent colorVariant={DrawerColorVariant.light200} panelContent={activeStep.drawerPanelContent}>
```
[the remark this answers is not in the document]
jenny-s51: Thank you @mcoker - removed that prop

mmenestr (approved):
This looks good! I think we should definitely consider adding the option to open/close the drawer in future demos though

tlabaj:
@m

> This looks good! I think we should definitely consider adding the option to open/close the drawer in future demos though

@mmenestr We could do that.  We would need designs on how we want to toggle the drawer open.  Would t there be a button in the masthead.  Would it be in the wizard body? Maybe we can open a design issue for this?

packages/react-core/src/components/Wizard/WizardDrawerWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+
+export interface DrawerWrapperProps {
+  children: React.ReactElement;
+  condition: boolean;
```
tlabaj, line 5: I would still add prop description her for other devs maintaining the code.
I would also make the condition name more descriptive o the condition that is being checked. to

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -66,23 +69,25 @@ export const WizardToggle: React.FunctionComponent<WizardToggleProps> = ({
...
-        onClick={() => onNavToggle(!isNavOpen)}
-        className={css(styles.wizardToggle, isNavOpen && 'pf-m-expanded')}
-        aria-label={ariaLabel}
-        aria-expanded={isNavOpen}
-      >
-        <span className={css(styles.wizardToggleList)}>
-          <span className={css(styles.wizardToggleListItem)}>
-            <span className={css(styles.wizardToggleNum)}>{activeStepIndex}</span> {activeStepName}
-            {activeStepSubName && <AngleRightIcon className={css(styles.wizardToggleSeparator)} aria-hidden="true" />}
+      {!hasDrawer && (
```
tlabaj, line 72: I am not quite understanding why we hare not rendering this if we have a drawer?

tlabaj (changes requested):

packages/react-core/src/components/Wizard/WizardToggle.tsx
jenny-s51: Good question @tlabaj ! I added this conditional to match the html structure of the core demo: 
Screen Shot 2022-05-06 at 3 43 49 PM

We don't want to render any of the stuff that comes before [.pf-c-wizard__outer-wrap](https://github.com/patternfly/patternfly-react/pull/7323/files/70ce3b170251b056b04ed7df97ec729e148c9960#diff-038302fe0ddb5ac4ec5b59ca6bf16bd2ef6fa81fcf8bb56572a40d78df73b59eR91)

packages/react-core/src/components/Wizard/WizardDrawerWrapper.tsx
jenny-s51: Great idea/suggestions - added prop descriptions here and changed `condition` to `hasDrawer`!

packages/react-core/src/components/Wizard/WizardToggle.tsx
tlabaj: I think the structure may be off.  It looks like the toggle is before the drawer in core.
image

In your pr it is missing al together
image

jenny-s51: Fixed! I moved `WizardDrawerWrapper` into `WizardToggle` which solved this structural issue - we no longer need that `hasDrawer` condition either! The HTML should match up now 👍 thank you @tlabaj

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -83,20 +88,29 @@ export const WizardToggle: React.FunctionComponent<WizardToggleProps> = ({
...
-          <WizardBody
-            mainComponent={isInPage ? 'div' : 'main'}
-            aria-label={mainAriaLabel}
-            aria-labelledby={mainAriaLabelledBy}
-            hasNoBodyPadding={hasNoBodyPadding}
-          >
-            {activeStep.component}
-          </WizardBody>
+      <WizardDrawerWrapper
+        hasDrawer={hasDrawer && activeStep.drawerPanelContent}
```
tlabaj, line 92: I think you can move this conditional out `WizardDrawerWrapper`. You should not be rendering it is these conditions are not met.

tlabaj (changes requested):

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -18,6 +18,8 @@ export interface WizardStep {
   name: React.ReactNode;
   /** The component to render in the main body */
   component?: any;
+  /** The content to render in the drawer panel */
+  drawerPanelContent?: any;
```
thatblindgeye, line 22: Nit: what do you think of tweaking the prop description to mention the `hasDrawer` prop, e.g. "The content to render in the drawer panel when hasDrawer is also passed in"? My thinking is just to connect the two props together more, so that consumers will know immediately they also need to pass `hasDrawer` in from the description.

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -29,6 +31,8 @@ export interface WizardToggleProps {
   mainAriaLabel?: string;
   /** If the wizard is in-page */
   isInPage?: boolean;
+  /** If the wizard has a drawer */
+  hasDrawer?: boolean;
```
thatblindgeye, line 35: Similar to above nit re: prop description

jenny-s51 (commented):
Great points @thatblindgeye ! Updated based on your suggestions 🙂

packages/react-core/src/components/Wizard/Wizard.tsx
jenny-s51: Good idea - definitely makes sense!

packages/react-core/src/demos/Wizard/WizardDemo.md
```diff
@@ -229,3 +394,207 @@ class FullPageWizard extends React.Component {
...
+      <DrawerPanelContent colorVariant={DrawerColorVariant.light200}>
+        <DrawerHead>
+          <span tabIndex={0} ref={this.drawerRef}>
+            Information panel content
+          </span>
+        </DrawerHead>
+      </DrawerPanelContent>
+    );
+
+    const configSubstepAContent = (
```
tlabaj, line 504: nit.  I might add panel to the name here and in other panel consts.

packages/react-core/src/demos/Wizard/WizardDemo.md
```diff
@@ -229,3 +394,207 @@ class FullPageWizard extends React.Component {
...
+        </DrawerHead>
+      </DrawerPanelContent>
+    );
+
+    const steps = [
+      { id: 0, name: 'Information', component: <p>Step 1 content</p>, drawerPanelContent: informationPanelContent },
+      {
+        id: 1,
+        name: 'Configuration',
+        steps: [
```
tlabaj, line 549: it looks like "configuration step" does not have. drawer panel content, but it seems to be rendering substep A panel. content when i look at demo

jenny-s51: This is the expected behavior - each of the substeps on the Configuration step have `drawerPanelContent` which render substep A panel content and substep B panel content!

wise-king-sullyman (approved):
I have one prop description nit, and I think that it would be beneficial to add some tests for the new component and props.

None of those are blockers for me though and everything else LGTM!

thatblindgeye (approved):

tlabaj (commented):
Looking great Jenny!  if you could had a unit test that verity's the drawer gets rendered, like Austin suggested, that would be great.

tlabaj (approved):
LGTM.  Thanks Jenny.

packages/react-core/src/demos/Wizard/WizardDemo.md
```diff
@@ -229,3 +393,206 @@ class FullPageWizard extends React.Component {
...
+          mainContainerId={pageId}
+        >
+          <PageSection variant={PageSectionVariants.light}>
+            <TextContent>
+              <Text component="h1">Main title</Text>
+              <Text component="p">A demo of a wizard in a page.</Text>
+            </TextContent>
+          </PageSection>
+          <PageSection type={PageSectionTypes.wizard} variant={PageSectionVariants.light}>
+            <Wizard hasDrawer navAriaLabel={`${title} steps`} mainAriaLabel={`${title} content`} steps={steps} />
```
mattnolting, line 591: Applying an `aria-label` here `mainAriaLabel={`${title} content`}` results in an Axe error. 

`aria-label attribute is not well supported on a div with no valid role attribute`

I think you can safely remove it.

packages/react-core/src/demos/Wizard/WizardDemo.md
```diff
@@ -86,7 +88,169 @@ class BasicWizardDemo extends React.Component {
...
+        variant={ModalVariant.large}
+        showClose={false}
+        hasNoBodyWrapper
+        aria-describedby="wiz-modal-demo-description"
+        aria-labelledby="wiz-modal-demo-title"
+      >
+        <Wizard
+          navAriaLabel={`${title} steps`}
+          hasDrawer
+          mainAriaLabel={`${title} content`}
```
mattnolting, line 239: You can remove this `mainAriaLabel={`${title} content`}`

mattnolting (commented):
Great work @jenny-s51! Just saw the Axe a11y error.

mattnolting (approved):
Awesome work! LPTM 🥇

mcoker:
@jenny-s51 I think we're going to need clarification from @mmenestr @mcarrano. I'm looking at the design in the original issue (https://github.com/patternfly/patternfly/issues/4611) and I don't see mobile designs included anywhere. There may be a couple of solutions:

* Figure out a way to toggle the drawer panel
  * Should the panel be closable in all viewports or just mobile?
  * When it's closable, what's the trigger to re-open it?
* Handle the drawer panel differently on mobile so it doesn't cover the main area
  * Don't show it (doesn't seem ideal)
  * Move it to come after the content or in some other kind of presentation that isn't the drawer panel (may be fairly complicated and require a custom implementation)

mcarrano:
@mcoker You raise a great question. It may be that we need to make the drawer closable in all cases so it can automatically close when the viewport in shrunk beyond a certain breakpoint. There is currently an open design issue here to consider how to open/close the drawer: https://github.com/patternfly/patternfly-design/issues/1160  @mmenestr perhaps you can look at this repressiveness issue as part of that?

mmenestr:
Ok, I have come up with the following proposal on how we should handle the mobile experience for a non-dismissible drawer in a wizard. Here is the mock: https://marvelapp.com/prototype/9389af6/section/1619656

The idea is that when in a mobile viewport, the non-dismissable drawer would become dismissible, and an "Open drawer" li button would surface on the top right of the wizard body. When clicked on, the drawer would open (from right to left, as you would expect from a drawer), and a user could dismiss it via the X button on the top right. One important change to note is that the drawer height would not go past the wizard body area, and would stop right above the button footer. That way, we don't have the problem of the buttons being covered when the drawer is open.

We chose to go this route to match the eventual demo I propose for a[ dismissible drawer](https://marvelapp.com/prototype/9389af6/section/1615614), which would have the same mobile behavior.

Let me know if you see any issues with this from a development standpoint!

jenny-s51 (commented):
@mmenestr @mcarrano @mcoker per your feedback, I've updated this PR with a dismissable drawer that matches the [design Margot shared](https://marvelapp.com/prototype/9389af6/screen/86742793). Thanks again @mattnolting for the Wizard updates in Core!

mcarrano:
Other than the issue @mcoker flagged above, this looks good to me. Thanks for all the work on this @jenny-s51 and @mattnolting !

@tlabaj @nicolethoen feel free to merge this once the issue referenced above is fixed.

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -29,6 +32,12 @@ export interface WizardToggleProps {
   mainAriaLabel?: string;
   /** If the wizard is in-page */
   isInPage?: boolean;
+  /** Flag indicating the wizard has a drawer for at least one of the wizard steps */
```
tlabaj, line 35: I would mark all props for new feature beta.

tlabaj (changes requested):
Looking good Jenny.  Just a few comments.

mattnolting:
@jenny-s51 to @mcoker's comment, `.pf-c-wizard__footer` needs to move to a sibling of `.pf-c-drawer__main` to match the requested design update

Screen Shot 2022-06-23 at 9 35 23 AM

rather than a sibling of `.pf-c-drawer`

Screen Shot 2022-06-23 at 9 36 22 AM

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -18,6 +18,10 @@ export interface WizardStep {
   name: React.ReactNode;
   /** The component to render in the main body */
   component?: any;
+  /** The content to render in the drawer panel (use when hasDrawer prop is set on the wizard).   */
+  drawerPanelContent?: any;
+  /** Button which toggles the drawer button open */
```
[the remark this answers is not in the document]
jenny-s51: Definitely sounds better! Updated!

packages/react-core/src/components/Wizard/WizardToggle.tsx
jenny-s51: Ah yes, good idea! Updated

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -83,20 +95,40 @@ export const WizardToggle: React.FunctionComponent<WizardToggleProps> = ({
...
+              aria-label={mainAriaLabel}
+              aria-labelledby={mainAriaLabelledBy}
+              hasNoBodyPadding={hasNoBodyPadding}
+            >
+              {activeStep.hasOpenDrawerButton && !isDrawerExpanded && (
+                <Button
+                  className="pf-u-float-right pf-u-ml-md"
+                  isInline
+                  variant="link"
+                  onClick={() => onDrawerToggle(!isDrawerExpanded)}
```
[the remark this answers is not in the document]
jenny-s51: This is a good point - ideally the drawer should toggle without the user needing to provide the `onDrawerToggle` callback. However since the `isDrawerExpanded` state is managed at the top level by the consumer, it looks like this method needs to stay at the same level in order for the toggle functionality to work as expected.

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -83,20 +95,40 @@ export const WizardToggle: React.FunctionComponent<WizardToggleProps> = ({
...
+                >
+                  Open Drawer
```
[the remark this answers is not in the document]
jenny-s51: Updated!

mattnolting (approved):
Nailed it! LPTM 🚀

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -340,6 +357,10 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
       titleId,
       descriptionId,
       isNavExpandable,
+      hasDrawer,
+      isDrawerExpanded,
+      onDrawerToggle,
+      drawerToggleButtonText,
```
tlabaj, line 363: Should we have a default for the text.  Otherwise maybe we should not render the button at all if not text is provided.  When the there is not text, the button is rendered but you can't click it since it has not width. 
 @mcoker What do you think?

tlabaj (changes requested):

packages/react-core/src/components/Wizard/Wizard.tsx
jenny-s51: I like this idea @tlabaj! Setting "Open Drawer" as the default for `drawerToggleButtonText`. I can add "defaults to Open Drawer" in the prop description, and the the consumer would know to set that prop on `Wizard` if they want some other custom text

jenny-s51: On second thought, it might be a better idea to have this text customizable at every step! Will push with a new update in a moment...

mcoker: @tlabaj hmm... looking at this again, we updated the open drawer functionality to be a demo. The user should be able to use any kind of button, anywhere they want to toggle the drawer. If I'm reading the example code correctly, I don't think we want to include that button in the component itself - this should be defined in the component example (or demo example). https://github.com/patternfly/patternfly-react/pull/7323/files#diff-038302fe0ddb5ac4ec5b59ca6bf16bd2ef6fa81fcf8bb56572a40d78df73b59eR111-R120

Also as a side note, the `.pf-u-*` classes used will not work if users are not importing the utility classes themselves, since those don't ship with patternfly by default. They would need to import https://www.patternfly.org/v4/utilities/float and https://www.patternfly.org/v4/utilities/spacing

tlabaj: So the wizard is not composable. I think the best we can do on the react side is pass a button rather than the text for the button we create internally.  That would also take care of the utility class issue, since they will not be part of the component.

tlabaj: @jenny-s51 It looks like the implementation would work for the scenario where the drawer toggle button is in the wizard content.  What is not accounted for here is the second scenario in the [design](https://marvelapp.com/prototype/9389af6/screen/86742796) where the button is inline as part of a paragraph in the content. 
Core also has this [demo](https://pf4.patternfly.org/components/wizard/html-demos#with-drawer-and-informational-step) with the inline variation

jenny-s51: @tlabaj thank you! Updated to include an inline drawer toggle demo in modal and in page!

mcoker (approved):
LGTM! One thing I wonder is if the utility class belongs in the component example or not, assuming users copy/paste example code, though it's just a question and not a blocker.

packages/react-core/src/components/Wizard/WizardToggle.tsx
```diff
@@ -91,7 +97,11 @@ export const WizardToggle: React.FunctionComponent<WizardToggleProps> = ({
             aria-label={mainAriaLabel}
             aria-labelledby={mainAriaLabelledBy}
             hasNoBodyPadding={hasNoBodyPadding}
+            activeStep={activeStep}
+            isDrawerExpanded={isDrawerExpanded}
+            hasDrawer={hasDrawer}
           >
+            {!isDrawerExpanded && activeStep.drawerToggleButton}
```
tlabaj, line 104: We should also check for `hasDrawer` here so the button will not render if there is not drawer.

tlabaj (changes requested):

packages/react-core/src/components/Wizard/examples/Wizard.md
```diff
@@ -859,3 +861,141 @@ class WizardInModal extends React.Component {
...
+import React from 'react';
+import { Button, DrawerActions, DrawerCloseButton, DrawerHead, DrawerPanelContent, Wizard } from '@patternfly/react-core';
+
```
tlabaj, line 870: Shouldn't we be importing https://www.patternfly.org/v4/utilities/float and https://www.patternfly.org/v4/utilities/spacing here? if the consumer copied and pasted this it would not work unless the imported these. Isn't hat correct @mcoker

tlabaj (changes requested):

packages/react-core/src/components/Wizard/examples/Wizard.md
nicolethoen: If they hadn't done so yet, they'd have to import the utilities style sheet, not the classes themselves. i'm not sure we'd want to do that in the example (re-import a whole stylesheet that we already import). But maybe instructions about importing the utilities style sheet can be added as a comment alongside the imports?

jenny-s51: That's a great idea @nicolethoen ! @tlabaj I combed through our repo, and we already happen to have a lot of demos that use utility classes like this without any kind of documentation. So I think for now, ultimately those changes would be out of scope of the functionality presented in this PR, since we would need to figure out the best way to present those instructions!

 Maybe for all those demos that use utility classes, we can open a separate issue to add those instructions as a comment alongside the imports for all those files? 

Do you think that would be a good game plan so we can get this merged @tlabaj @nicolethoen @mcoker ?

tlabaj: So maybe @mcokers comment is correct.  We should not use the utilities in the example.  but if we leave them out, the layout not as shown in the design.
image

Maybe the example should demo the  button inline in a paragraph variation Then we just demo this variation with the action demo?
Another question I would ask is would we want adding the toggle button and having it aligned correctly work without needing to apply utility classes, or are we ok with it not aligning by default, if that is ok, we can just remove the classes from the example like @mcoker mentioned above.

cc @mcarrano

mcoker: > are we ok with it not aligning by default, if that is ok, we can just remove the classes from the example 

@tlabaj my thinking is that the wizard component doesn't implement the layout/design, and they can do it any number of ways. So I think just presenting the default functionality of the prop in the example is good, and users can look at the demos for design/layout examples.

mmenestr:
This looks great overall!! Just have a few small things:

- Can we change to "Open drawer" with lower case D to follow sentence case guidelines
- Should the drawer only span that tiny little box when it's in a small screen? Shouldn't it at least reach the footer? 

image

- And then finally, should the footer have a shadow behind it? It looks weird when the drawer is fully open but doesn't go all the way down. At least when there's a show behind the footer, it makes sense for it to stop where it stops if that makes sense. Not sure if we only do that for sticky footers and that's why there's no shadow but thought I'd suggest it.
- 
image

packages/react-core/src/components/Wizard/examples/Wizard.md
tlabaj: @mcoker I agree.  So @jenny-s51 Can you just remove the utility class from the example.  The demo will show how to lay it out.

mcoker:
> Should the drawer only span that tiny little box when it's in a small screen? Shouldn't it at least reach the footer?

It's because this wizard doesn't have a height assigned. Either the wizard or its parent needs some sort of a height. I see we have an inline `height: 400px` style on all wizards in the examples. We could add that to this drawer example, or remove those inline styles and add a local stylesheet for the examples page that styles `.ws-preview { height: 450px; }` (the extra 50px is to account for the top/bottom padding of `.ws-preview`). With that approach, individual examples could still be overridden by targeting the id of that example, e.g. `#ws-react-c-wizard-wizard-with-drawer { height: [something else]; }`

> should the footer have a shadow behind it?

It's because we only enable the footer shadow when we know the footer is sticky, so it's currently limited to use cases where the wizard is a child of the page component's wizard section element, a modal, or a drawer. We have a task to normalize this style in an upcoming breaking change release - https://github.com/patternfly/patternfly/blob/e530391ee29e1b9f719b40881b635914a99fb80e/src/patternfly/components/Wizard/wizard.scss#L550-L557

packages/react-core/src/components/Wizard/examples/Wizard.md
jenny-s51: Good call - updated to remove the utility classes

tlabaj (approved):

mcoker (approved):
🚀

mmenestr (approved):

Issue #7021, opened 2022-03-08
DualListSelectorControl tooltips remain visible after VO blur


thatblindgeye:
**Describe the issue. What is the expected and unexpected behavior?**
The tooltips on the Dual List Selector component's controls do not currently get read by screen readers; whether this should be fixed since most likely the tooltip will match the `aria-label` of the button may require additional input. UPDATE: this portion of the issue looks to be resolved.

The tooltips also do not disappear immediately when navigating via VO (you have to navigate outside the button controls area it looks like, rather than simply navigating away from the button that has the tooltip currently displayed).

If the tooltips should be announced by screen readers, even if the contents of them matches the button's `aria-label`, then this might be similar to PR #6940 in that instead of using the Tooltip's `reference` prop, it should instead act as a wrapper around the DualListSelectorControl inner button element.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

Go to the PatternFly [Dual List Selector](https://www.patternfly.org/v4/components/dual-list-selector#basic-with-tooltips) page and use VO on the With Tooltips example. Notice that the buttons between the two option lists do not have tooltips announced by screen readers, nor do they immediately disappear after moving focus away from a button.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Could be considered a bug

thatblindgeye:
Updates made in PR https://github.com/patternfly/patternfly-react/pull/7332 look to have resolve the issue of tooltips not getting announced. However, there is still a small issue where when navigating via VO controls from an active button to a disabled button, the tooltip remains visible. See video below:

https://user-images.githubusercontent.com/70952936/167696901-7564eb7e-603c-4252-afe9-d190e8fd8fd8.mov

thatblindgeye:
This may be an issue with Voiceover/screen reader focus than anything to do with PF. Will keep this issue closed due to that and we can revisit if this is brought up again.

Issue #7096, opened 2022-03-21
CodeBlock: Tooltips in code block examples no VO accessible


nicolethoen:
The tooltip text indicating that the text has been successfully copied to clipboard does not get read by VO.
![Mar-21-2022 11-06-14](https://user-images.githubusercontent.com/26305082/159290333-9d3403c6-efd9-4007-a8a1-de35ae6fa4ec.gif)

Issue #4446, opened 2020-06-24
Tooltip examples aren't read by VO in Safari


jessiehuff:
**Describe the issue. What is the expected and unexpected behavior?**
The tooltip examples don't seem to be read in VO when using Safari, except for the positions example. I've noticed that it does currently work in Chrome, but seems to have some difficulty in Safari. It only reads the item that the tooltip is on and then tells you what type of element it is. So in this case, it reads "I have a tooltip" then "You are currently on a text element." I would expect it to read out the tooltip as well.

<img width="1025" alt="Screen Shot 2020-06-24 at 11 50 16 AM" src="https://user-images.githubusercontent.com/28281340/85588686-e8d18a80-b610-11ea-8371-306caab31c4f.png">

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
Bug

thatblindgeye:
The tooltip examples are resolved in Safari as the examples have been updated to use buttons. Per the epic this is linked to, tooltips on static elements generally aren't well supported so it's less a Safari issue and more an issue with how AT may announce accessible names/descriptions. So this can be closed.

 However I will open a followup issue to investigate the tooltips on MenuItem components as the tooltips aren't being announced in any browser for me there (tested VO and NVDA).

PR #7332, opened 2022-05-02
fix(Tooltip): make ref variant announced by assistive tech

### LINKED ISSUE #7331: Bug - Tooltip - contents for ref variant not announced by assistive tech
**Describe the problem**
Currently when using a React or selector ref for the Tooltip component, the tooltip contents does not get announced by assistive technologies. Only using the Tooltip component as a wrapper has the contents announced.

1. Go to the [Tooltip component](https://www.patternfly.org/v4/components/tooltip/)
2. Start Voice Over or another screen reader
3. Navigate to the examples that use a React or selector Ref
4. Notice that the tooltip contents may not get announced (tested using VO in Chrome)

**Expected behavior**
The tooltip contents gets announce when using the React or selector ref variants

An alternative is using the Tooltip component as laid out in the "basic" example in the above link (using the Tooltip component as a wrapper).

**Screenshots**
If applicable, add screenshots to help explain the issue.

**What is your environment?**

**Any other information?**

thatblindgeye:
**What**: Closes #7331

Convenience links:

[Tooltip component](https://patternfly-react-pr-7332.surge.sh/components/tooltip)
[Tabs component](https://patternfly-react-pr-7332.surge.sh/components/tabs)

**Additional issues**:
This PR also updates the Tab component that was previously updated in PR #6940.

Additionally, this PR could also resolve issues #7021 and #4446, but only when using the React or selector ref variants (or by using `aria="none"` and `aria-live="polite"` props). So this could potentially require altering how the Tooltip prop is setup or just a larger discussion, depending on if another solution can be found when not using a ref. This I believe was alluded to in another issue in this epic #1655: "One possible alternative is to render the tooltip as an aria-live region."

This PR getting merged would also go towards fixing #7096, as that would be updated to follow the "Icon with dynamic content" Tooltip example in this PR.

### DIFF
--- a/packages/react-core/src/components/Tabs/examples/Tabs.md
+++ b/packages/react-core/src/components/Tabs/examples/Tabs.md
@@ -106,6 +106,8 @@ class SimpleTabs extends React.Component {
 
 When using a React ref to link a Tooltip to a Tab component via the `reference` prop, you should avoid manually passing in a value of "off" to the `aria-live` prop. Doing so may lead to the tooltip becoming less accessible to assistive technologies.
 
+The tooltip should also have the `id` prop passed in. The value given to this prop should then be passed into the tab's `aria-describedby` prop. This ensures a tooltip used with a React ref will be announced by the JAWS and NVDA screen readers.
+
 ```js
 import React from 'react';
 import { Tabs, Tab, TabTitleText, Checkbox, Tooltip } from '@patternfly/react-core';
@@ -163,11 +165,13 @@ class SimpleTabs extends React.Component {
             title={<TabTitleText>ARIA Disabled (Tooltip)</TabTitleText>}
             isAriaDisabled
             ref={tooltipRef}
+            aria-describedby="tooltip-ref1"
           >
             ARIA Disabled (Tooltip)
           </Tab>
         </Tabs>
         <Tooltip
+          id="tooltip-ref1"
           content="Aria-disabled tabs are like disabled tabs, but focusable. Allows for tooltip support."
           reference={tooltipRef}
         />
--- a/packages/react-core/src/components/Tabs/examples/Tabs.md
+++ b/packages/react-core/src/components/Tabs/examples/Tabs.md
@@ -56,7 +56,12 @@ class SimpleTabs extends React.Component {
 
     return (
       <div>
-        <Tabs activeKey={activeTabKey} onSelect={this.handleTabClick} isBox={isBox} aria-label="Tabs in the default example">
+        <Tabs
+          activeKey={activeTabKey}
+          onSelect={this.handleTabClick}
+          isBox={isBox}
+          aria-label="Tabs in the default example"
+        >
           <Tab eventKey={0} title={<TabTitleText>Users</TabTitleText>}>
             Users
           </Tab>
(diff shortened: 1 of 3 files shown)


jessiehuff (commented):
I think this is definitely a really interesting approach. It likely would eliminate some of the differences we've been seeing across browsers and screen readers. I know that some thought leaders haven't been in favor of this approach (for example, Sarah Higley who works at Microsoft, wrote [this article warning against it](https://sarahmhigley.com/writing/tooltips-in-wcag-21/)). However, I've also read others [like this one](https://pressbooks.library.ryerson.ca/wafd/chapter/tooltips/) that speak to our experience. I think if we're finding that this is the most accessible and consistent across browsers and screen readers, then we should go with what works. Open to what others think. 🙂 I'm wondering if I should test this with other screen readers outside of VO.

thatblindgeye:
> I think this is definitely a really interesting approach. It likely would eliminate some of the differences we've been seeing across browsers and screen readers. I know that some thought leaders haven't been in favor of this approach (for example, Sarah Higley who works on Microsoft's a11y, wrote [this article warning against it](https://sarahmhigley.com/writing/tooltips-in-wcag-21/)). However, I've also read others [like this one](https://pressbooks.library.ryerson.ca/wafd/chapter/tooltips/) that speak to our experience. I think if we're finding that this is the most accessible and consistent across browsers and screen readers, then we should go with what works. Open to what others think. 🙂 I'm wondering if I should test this with other screen readers outside of VO.

Yeah I think it's really iffy either way. I do agree with that first article in regards to users not being able to opt out when a tooltip uses `aria-live`, but I do feel the preferred method in that article also suffers the drawback of not being reliably read by screen readers (though I think slightly worse when it comes to using refs in React, and then there is also Safari which doesn't seem to read tooltips properly even when used as a wrapper with `aria-describedby`).

It can really come down to not allowing users to opt out of the tooltip content vs not having the tooltip content read whatsoever depending on the browser (which even if tooltip content shouldn't be required for users to understand something, I'm not sure if preventing it from users because of browser is the right move).

kmcfaul (changes requested):

nicolethoen:
Testing the changes works great! I just want to sanity check my understanding that the only html markup change with this approach is the aria-live and aria attributes? It doesn't seem too risky of a breaking change unless there is something else I'm missing

thatblindgeye:
> Testing the changes works great! I just want to sanity check my understanding that the only html markup change with this approach is the aria-live and aria attributes? It doesn't seem too risky of a breaking change unless there is something else I'm missing

@nicolethoen That's correct, ideally this shouldn't cause any breaking issues since it's only adding the optional aria-live prop, and if any consumers are already using that attribute it should still work as they intended.

nicolethoen (approved):

jessiehuff (approved):
LGTM! 🙂

thatblindgeye:
@nicolethoen Per convo with @jessiehuff, the Tooltip and Tabs examples that use the `reference` prop on the Tooltip have the `aria-describedby` attribute added back in. Both JAWS and NVDA didn't announce the tooltip contents with only `aria-live`, so adding the aria-describedby back in resolves that while aria-live helps with VO.

Further tweaking of the Tooltip component may be needed down the line to find a solution that helps with these three main screen readers, but this does help improve things a bit still.

tlabaj (approved):

nicolethoen (approved):

kmcfaul (approved):

PR #7334, opened 2022-05-02
fix: tree view menu keyboard handling

### LINKED ISSUE #7083: Composable Menu demo: tree view - not keyboard accessible
The [composable menu tree view demo](https://www.patternfly.org/v4/demos/composable-menu#composable-tree-view-menu) does not provide a way to navigate into the menu via keyboard navigation.

kmcfaul:
**What**: Closes #7083

The demo looks like it was broken when I added the KeyboardHandler to Menu, which expects the typical menu structure with menu components. I've added a property to allow a user to override the handler for custom menus and updated the demo accordingly.

There's a minor issue I'm still tracking down where toggling a tree folder somehow adds a tabindex to the group header and focuses it instead. If anyone has any ideas about this LMK.

### DIFF
--- a/packages/react-core/src/demos/ComposableMenu/examples/ComposableTreeViewMenu.tsx
+++ b/packages/react-core/src/demos/ComposableMenu/examples/ComposableTreeViewMenu.tsx
@@ -168,7 +168,7 @@ export const ComposableTreeViewMenu: React.FunctionComponent = () => {
       return;
     }
     if (menuRef.current.contains(event.target as Node) || toggleRef.current.contains(event.target as Node)) {
-      if (event.key === 'Escape' || event.key === 'Tab') {
+      if (event.key === 'Escape') {
         setIsOpen(!isOpen);
         toggleRef.current.focus();
       }
@@ -194,7 +194,7 @@ export const ComposableTreeViewMenu: React.FunctionComponent = () => {
     ev.stopPropagation(); // Stop handleClickOutside from handling
     setTimeout(() => {
       if (menuRef.current) {
-        const firstElement = menuRef.current.querySelector('li > button:not(:disabled)');
+        const firstElement = menuRef.current.querySelector('li button:not(:disabled)');
         firstElement && (firstElement as HTMLElement).focus();
       }
     }, 0);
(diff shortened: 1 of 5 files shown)


jessiehuff (commented):
Really nice catch - thanks for fixing this!

The only random comment I have is that the jump links seem connected to the wrong headers? I checked org and the surge site and didn't see this. Not quite sure why this is happening.

packages/react-core/src/demos/ComposableMenu/examples/ComposableTreeViewMenu.tsx
```diff
@@ -168,10 +169,19 @@ export const ComposableTreeViewMenu: React.FunctionComponent = () => {
       return;
     }
     if (menuRef.current.contains(event.target as Node) || toggleRef.current.contains(event.target as Node)) {
-      if (event.key === 'Escape' || event.key === 'Tab') {
+      if (event.key === 'Escape') {
```
nicolethoen, line 172: I think a comment before the `handleMenuKeys` function to explain the goals or use cases covered in this function would be super helpful.

nicolethoen (commented):
LGTM! I think just a little more context for why you wrote/needed the various helper functions in the example comments would help a lot.

kmcfaul:
Updated with some comments around the handlers. Also removed the KeyboardHandler override prop from Menu as it's no longer used.

thatblindgeye (commented):
Overall looks good! There were some things I noticed when using VO that would require a separate issue, but otherwise I think the only thing I'm noticing is some potential Tab behavior issue. Below is a video, but when I first open the toggle with keyboard and the Ready toggle button receives focus, pressing Tab causes the Ready checkbox to receive focus. If I open the menu toggle and press down arrow to place focus on the Not Ready toggle, pressing Tab results in the second tree view receiving focus. I wasn't sure if this was the expected behavior or not.

https://user-images.githubusercontent.com/70952936/166933331-802adf7d-bd13-4cdf-88db-26dd656924ab.mov

kmcfaul:
Tab is meant to go between trees in this demo, and moving the focus via the arrow keys should be updating the tabIndex so only that row (which currently includes all interactable elements in the row like the button and input) is tabbable. You can see this behavior in the Treeview checkbox example as well - where after tabbing through the first row it skips to the next interactable element after the Treeview.

nicolethoen (approved):

jessiehuff (approved):

Issue #7360, opened 2022-05-04
Charts: Add style for tooltip cursor


dlabrecq:
To support the dark theme, we need to style the vertical cursor used with tooltips. In order to do that, we must override Victory's cursor component.

<img width="779" alt="Screen Shot 2022-05-04 at 7 18 39 PM" src="https://user-images.githubusercontent.com/17481322/166840534-a19d272e-e20f-4de3-a930-10ea33f0962c.png">

dlabrecq:
Merged https://github.com/patternfly/patternfly-react/pull/7361

PR #7361, opened 2022-05-04
fix(charts): Add style for tooltip cursor

dlabrecq:
To support the dark theme, we need to style the vertical cursor used with tooltips. In order to do that, we must override Victory's cursor component.

https://github.com/patternfly/patternfly-react/issues/7360

Testing: Changed stroke style to red
Screen Shot 2022-05-04 at 7 23 59 PM


mcoker (approved):
works great!

PR #7363, opened 2022-05-05
Add missing 'children' prop in several places

jonkoops:
Adds 'children' prop to the types of various components where they were missing.  In React 18 a component type (e.g. `FunctionComponent`) [no longer has](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210) an implicit `children` prop. 

This is needed to land #7142.


wise-king-sullyman (approved):

tlabaj (approved):
LGTM.

jonkoops:
Thanks for the quick response y'all!

PR #7364, opened 2022-05-05
Topology dark theme follow-on

srambach:
**What**: This is a follow-on to address a few of the comments in #7294 that came after merging. Not all comments have been addressed because they are suggested changes to colors deviating from existing light theme and the spec'd dark theme. @mceledonia I defer to you on any changes to colors.

- Fixes css variable fallbacks
- Adds css variables for stroke width
- Uses one named color where the same palette one was used


mcoker (approved):
LGTM!

jeff-phillips-18 (approved):

tlabaj (approved):

PR #7368, opened 2022-05-05
feat(Menu): Added support for checkbox menu

### LINKED ISSUE #7166: [Menu] - [Add checkbox menu]
Menu

**Describe the feature**
Add checkbox menu
https://github.com/patternfly/patternfly/issues/4663

tlabaj:
**What**: Closes #7166  #7367 #7433

link to checkbox menu example:https://patternfly-react-pr-7368.surge.sh/components/menu#with-checkbox

A new demo was added to composable menus: https://patternfly-react-pr-7368.surge.sh/demos/composable-menu#composable-simple-checkbox-select

Issue to track  keyboard interactions #7382.

### DIFF
--- a/packages/react-core/src/components/Menu/MenuItem.tsx
+++ b/packages/react-core/src/components/Menu/MenuItem.tsx
@@ -295,7 +295,7 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
       onMouseOver={onMouseOver}
       {...(flyoutMenu && { onKeyDown: handleFlyout })}
       ref={ref}
-      role="none"
+      role={!hasCheck ? 'none' : 'menuitem'}
       {...props}
     >
       <GenerateId>
@@ -305,7 +305,7 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
             className={css(styles.menuItem, getIsSelected() && !hasCheck && styles.modifiers.selected, className)}
             aria-current={getAriaCurrent()}
             {...(!hasCheck && { disabled: isDisabled })}
-            role={!hasCheck ? 'none' : 'menuitem'}
+            {...(!hasCheck && { role: 'menuitem' })}
             ref={innerRef}
             {...(!hasCheck && {
               onClick: (event: any) => {
--- a/packages/react-core/src/components/Menu/MenuItem.tsx
+++ b/packages/react-core/src/components/Menu/MenuItem.tsx
@@ -305,7 +305,7 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
             className={css(styles.menuItem, getIsSelected() && !hasCheck && styles.modifiers.selected, className)}
             aria-current={getAriaCurrent()}
             {...(!hasCheck && { disabled: isDisabled })}
-            role="menuitem"
+            role={hasCheck ? 'none' : 'menuitem'}
             ref={innerRef}
             {...(!hasCheck && {
               onClick: (event: any) => {
@@ -327,6 +327,7 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
                 <span className={css('pf-c-menu__item-check')}>
                   <Checkbox
                     id={randomId}
+                    component="span"
                     isChecked={isSelected || false}
                     onChange={event => onItemSelect(event, onSelect)}
                     isDisabled={isDisabled}
--- a/packages/react-core/src/components/Menu/MenuItem.tsx
+++ b/packages/react-core/src/components/Menu/MenuItem.tsx
@@ -25,7 +25,7 @@ export interface MenuItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'onC
   /** Target navigation link */
   to?: string;
   /** @beta Flag indicating the item has a checkbox */
-  isCheck?: boolean;
+  hasCheck?: boolean;
   /** Flag indicating whether the item is active */
   isActive?: boolean;
   /** Flag indicating if the item is favorited */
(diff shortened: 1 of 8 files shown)


mcarrano (commented):
@tlabaj this looks fine, but in the example on the Menu page, how come clicking the checkbox does not toggle the state?

tlabaj:
> @tlabaj this looks fine, but in the example on the Menu page, how come clicking the checkbox does not toggle the state?

The "Basic" menu example did not show selected state so I made this in match.  I suppose that is a little different since that can be used as an action menu.  I can make the update to toggle the checkbox.

nicolethoen (commented):
The basic checkbox select example also includes a count badge which increments/decrements as items are checked/unchecked. Is that something we should add to the composable menu example?

mcarrano:
Good catch @nicolethoen . I missed that. We should add it if we can.

mcarrano (approved):
Looks great @tlabaj !

jessiehuff (approved):
Looks good to me since we have the other issue for keyboard interaction. :)

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -22,6 +24,8 @@ export interface MenuItemProps extends Omit<React.HTMLProps<HTMLLIElement>, 'onC
   itemId?: any;
   /** Target navigation link */
   to?: string;
+  /** @beta Flag indicating the item has a checkbox */
+  isCheck?: boolean;
```
kmcfaul, line 28: Would `hasCheck` make more sense here? Not a blocker as I'm happy with either prop name.

kmcfaul (approved):

gitdallas (approved):

packages/react-core/src/components/Menu/MenuItem.tsx
tlabaj: I was waiting for someone to comment.  I was debating on the name here.

wise-king-sullyman (approved):

kmcfaul (approved):

nicolethoen (approved):

mmenestr (approved):

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -290,49 +298,66 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
...
-            <span className={css(styles.menuItemSelectIcon)}>
-              <CheckIcon aria-hidden />
+      <GenerateId>
+        {randomId => (
+          <Component
+            tabIndex={-1}
+            className={css(styles.menuItem, getIsSelected() && !hasCheck && styles.modifiers.selected, className)}
+            aria-current={getAriaCurrent()}
+            {...(!hasCheck && { disabled: isDisabled })}
+            role="menuitem"
```
mcoker, line 308: From the core docs, if the menu item is a checkbox item (`<label>`), this item is set to `role="none"` and the parent `.pf-c-menu__list-item` element gets `role="menuitem"`

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -290,49 +298,66 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
...
+            <span className={css(styles.menuItemMain)}>
+              {direction === 'up' && (
+                <span className={css(styles.menuItemToggleIcon)}>
+                  <AngleLeftIcon aria-hidden />
+                </span>
+              )}
+              {icon && <span className={css(styles.menuItemIcon)}>{icon}</span>}
+              {hasCheck && (
+                <span className={css('pf-c-menu__item-check')}>
+                  <Checkbox
```
mcoker, line 328: Ideally this should be a `<span>` because it is a child of a `<label>`, and [`<label>` only supports phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#technical_summary)

Doesn't look like the checkbox supports changing the element via `component` - could we add a follow up for that? https://github.com/patternfly/patternfly-react/blob/6e50f2379b8cd89bce6d5542a422def46162c3c4/packages/react-core/src/components/Checkbox/Checkbox.tsx#L96

mcoker (changes requested):
Looks great! Just one thing about a `role` change I'm seeing as different from what's in core.

packages/react-core/src/components/Menu/MenuItem.tsx
tlabaj: I opened issue #7433 and addressed it in this PR.

packages/react-core/src/components/Menu/MenuItem.tsx
```diff
@@ -305,7 +305,7 @@ const MenuItemBase: React.FunctionComponent<MenuItemProps> = ({
             className={css(styles.menuItem, getIsSelected() && !hasCheck && styles.modifiers.selected, className)}
             aria-current={getAriaCurrent()}
             {...(!hasCheck && { disabled: isDisabled })}
-            role={hasCheck ? 'none' : 'menuitem'}
+            role={!hasCheck ? 'none' : 'menuitem'}
```
mcoker, line 308: I think this was right for the most part before - if it's not a check, it should have `role="menuitem"`,  if it's a checkbox it _shouldn't have a `role` attribute set_, which I'm assuming will just use whatever default role a `<label>` is. Sorry my previous suggestion said to set it to `role="none"` but looking at the core examples, that isn't correct.

Here are screenshots from the core examples. 

Regular non-checkbox menu
Screen Shot 2022-05-17 at 5 35 13 PM


Checkbox menu
Screen Shot 2022-05-17 at 5 34 47 PM

mcoker (changes requested):
Left a comment about the role on the menu item. Also looks like we need to update the menu list item to have `role="menuitem"` if it's a checkbox item.

mcoker (approved):
🤩

PR #7370, opened 2022-05-05
feat(table): add fullscreen loading state demo

jenny-s51:
**What**: Closes #7342

**Convenience link to demo:** https://patternfly-react-pr-7370.surge.sh/components/table/react-demos/loading/

PR #7371, opened 2022-05-05
feat(table): add fullscreen empty state demo

### LINKED ISSUE #7343: Feat(Table): Full screen demo for empty state

jenny-s51:
**What**: Closes #7343 

**Convenience link to demo**: https://patternfly-react-pr-7371.surge.sh/components/table/react-demos/#empty


mcarrano (changes requested):
@jenny-s51 Just a couple of requests...

Can we make all three demos for empty state full screen? Seems odd that only the first one is. Also, the third demo (Error) should include an icon as in the design guidelines here: https://www.patternfly.org/v4/components/empty-state/design-guidelines#back-end-failure

@mmenestr please also take a look at this one to see if there is anything else we should update.

mmenestr:
Don't see anything else - agree with @mcarrano's comments, otherwise looks good!

jenny-s51:
Thank you Matt, yes per your comment in the loading state PR: https://github.com/patternfly/patternfly-react/pull/7370#pullrequestreview-964683682, I am adding the empty state demos separately and should have the error state up soon! Thank you @mcarrano @mmenestr

mattnolting (approved):
LPTM 👍

mcarrano (commented):
@jenny-s51 I think this is fine. I do notice subtle differences when I switch between this and the HTML demo. I think there are mostly due to navigation and masthead differences. Is this part us what we still need to sync up on for parity? It would be great if we can get to a common and consistent page shell.

tlabaj:
> @jenny-s51 I think this is fine. I do notice subtle differences when I switch between this and the HTML demo. I think there are mostly due to navigation and masthead differences. Is this part us what we still need to sync up on for parity? It would be great if we can get to a common and consistent page shell.

@mcarrano we have an issue open to update the dashboard wrapper to match core.  I think once this is done, the issues you are noticing will be resolved.

tlabaj (approved):
LGTM

mcarrano (approved):
> @mcarrano we have an issue open to update the dashboard wrapper to match core. I think once this is done, the issues you are noticing will be resolved.

In that case, I approve! Thanks @jenny-s51 .

PR #7377, opened 2022-05-06
fix(Banner): add screen reader text

### LINKED ISSUE #5890: Banner examples need screen reader text/label indicating the context
The banner component uses color to communicate information/context to users. This information/context is not given to screen reader users who cannot see the color. Since we allow users to add any different types of context (for example, Gov't systems uses these to indicate whether the information on a page is classified or not), we should at least provide an example in which the context is provided through a label or screen reader text.

Go to the [banner documentation](https://www.patternfly.org/v4/components/banner) and notice that the colors can potentially indicate state such as success, danger, etc. 

Example bug

bartoval:
**What**: Closes #5890

### DIFF
--- a/packages/react-core/src/components/Banner/Banner.tsx
+++ b/packages/react-core/src/components/Banner/Banner.tsx
@@ -11,7 +11,7 @@ export interface BannerProps extends React.HTMLProps<HTMLDivElement> {
   variant?: 'default' | 'info' | 'danger' | 'success' | 'warning';
   /** If set to true, the banner sticks to the top of its container */
   isSticky?: boolean;
-  /** Text announced by screen reader when variant is set to indicate the type of banner*/
+  /** Text announced by screen readers to indicate the type of banner. Defaults to "${variant} banner" if this prop is not passed in */
   screenReaderText?: string;
 }
 
--- a/packages/react-core/src/components/Banner/Banner.tsx
+++ b/packages/react-core/src/components/Banner/Banner.tsx
@@ -11,12 +11,15 @@ export interface BannerProps extends React.HTMLProps<HTMLDivElement> {
   variant?: 'default' | 'info' | 'danger' | 'success' | 'warning';
   /** If set to true, the banner sticks to the top of its container */
   isSticky?: boolean;
+  /** Text announced by screen reader when variant is set to indicate the type of banner*/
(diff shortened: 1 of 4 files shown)


packages/react-core/src/components/Banner/Banner.tsx
```diff
@@ -11,12 +11,15 @@ export interface BannerProps extends React.HTMLProps<HTMLDivElement> {
   variant?: 'default' | 'info' | 'danger' | 'success' | 'warning';
   /** If set to true, the banner sticks to the top of its container */
   isSticky?: boolean;
+  /** Text announced by screen reader when variant is set to indicate the type of banner*/
+  screenReaderText?: string;
```
thatblindgeye, line 15: Thinking some slight tweaking of the prop description could be done here, since the variant will always be set and to mention what the default is:

"Text announced by screen readers to indicate the type of banner. Defaults to `${variant} banner` if this prop is not passed in.".

thatblindgeye (commented):
Looks good! Just had a small suggestion/comment below, let me know what you think.

packages/react-core/src/components/Banner/Banner.tsx
bartoval: Thanks!
Now it's more clear 👍🏼

thatblindgeye (approved):

jessiehuff (approved):
LGTM! Thanks! 🙂

packages/react-core/src/components/Banner/Banner.tsx
```diff
@@ -11,12 +11,15 @@ export interface BannerProps extends React.HTMLProps<HTMLDivElement> {
   variant?: 'default' | 'info' | 'danger' | 'success' | 'warning';
   /** If set to true, the banner sticks to the top of its container */
   isSticky?: boolean;
+  /** Text announced by screen readers to indicate the type of banner. Defaults to ${variant} banner if this prop is not passed in */
```
tlabaj, line 14: Nit
```suggestion
  /** Text announced by screen readers to indicate the type of banner. Defaults to "${variant} banner" if this prop is not passed in */
```

tlabaj (commented):
Looks great!  Just one little nit.

tlabaj (approved):

PR #7385, opened 2022-05-06
fix(Popper): update Popper modifiers when child DOM changes

### LINKED ISSUE #7162: Bug - Popover - Popover positioning can result in contents outside viewport
**Describe the problem**
It appears `Popover` is rendering (and thus positioning itself) before `children` components passed to `bodyContent` have finished rendering, resulting in poor placement of the Popover.

Screen Shot 2022-04-01 at 9 06 52 AM

If I replace `children` with static JSX (e.g., `<div>Lorem ipsum.</div>`, the bug doesn’t occur.

Screen Shot 2022-04-01 at 9 05 32 AM

Is there some way to tell the parent Popover not to render until the children have fully rendered? 

See https://bugzilla.redhat.com/show_bug.cgi?id=2070457

**Expected behavior**
`Popover` doesn't show and position until contents are finished rendering.

Yes.  See https://bugzilla.redhat.com/show_bug.cgi?id=2070457.  We've had similar issues in OpenShift console as well.

**Screenshots**
See above.

**What is your environment?**
 - OS: MacOS
 - Browser:  Chrome 
 - Version: 100.0.4896.60

OpenShift 4.11

**Any other information?**

nicolethoen:
**What**: Closes #7162 

With the help of @mturley, we identified that the contents of openshift's popover were being lazy loaded and stealthily swapping out the 'loading' indicator from their popover without triggering a state change or rerender.

We added a new observer to Popper (on which Popover is built) which will watch the entire contents of the Popper for any DOM tree mutations or updates. Whenever a mutation is detected, we will call Popper's 'update' function which reruns all of Popper's modifiers to make sure it's positioned appropriately. Popper's 'update' function uses debounce and is already being used to check Popper's position everytime the user scrolls or resizes the page, so it shouldn't introduce a performance problem. 

Also, I removed the `useEffect` which watches for changes to the `bodyContent` and calls `forceUpdate` since my change also effectively does that - with the added benefit of using debouce to prevent updates from being performed too many times.

### DIFF
--- a/packages/react-core/src/helpers/Popper/Popper.tsx
+++ b/packages/react-core/src/helpers/Popper/Popper.tsx
@@ -165,6 +165,10 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
       element.removeEventListener(event, listener);
     }
   };
+
+  // Obeserver added to address https://github.com/patternfly/patternfly-react/issues/7162
+  // and trigger a Popper update when content changes.
+  // Also accounts for https://github.com/patternfly/patternfly-react/issues/5620
   const observer = new MutationObserver(() => {
     update && update();
   });
--- a/packages/react-core/src/helpers/Popper/Popper.tsx
+++ b/packages/react-core/src/helpers/Popper/Popper.tsx
@@ -165,6 +165,10 @@ export const Popper: React.FunctionComponent<PopperProps> = ({
       element.removeEventListener(event, listener);
     }
   };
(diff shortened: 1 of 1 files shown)


tlabaj (approved):

tlabaj (commented):
Looks good. just the nit about the comment.

mturley (approved):

tlabaj (approved):
Looks good. just the nit about the comment.

PR #7387, opened 2022-05-09
fix(Chip): add custom maxWidth prop

### LINKED ISSUE #7167: Chip - allow customizable max-width
Currently, the max width of a chip's text is set via a css variable `--pf-c-chip__text--MaxWidth` to be 16ch. If the text is wider than that, it is truncated and the chip is wrapped in a tooltip. 

We should create a react prop to override the value of the max-width so that it can be wider or never truncate as the consumer desires.

bartoval:
**What**: Closes #7167

### DIFF
--- a/packages/react-core/src/components/Chip/Chip.tsx
+++ b/packages/react-core/src/components/Chip/Chip.tsx
@@ -38,6 +38,9 @@ export interface ChipProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
     | 'left-end'
     | 'right-start'
     | 'right-end';
+
+  /** Css property expressed in percentage or every css unit that override the default value of the max-width  */
+  maxWidthText?: string;
 }
 
 interface ChipState {
@@ -73,6 +76,8 @@ export class Chip extends React.Component<ChipProps, ChipState> {
     });
   }
 
+  setMaxWidthText = () => this.props.maxWidthText || 'auto';
+
   renderOverflowChip = () => {
     const { children, className, onClick, ouiaId } = this.props;
     const Component = this.props.component as any;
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/Chip/Chip.tsx
```diff
@@ -83,7 +88,9 @@ export class Chip extends React.Component<ChipProps, ChipState> {
         {...(this.props.component === 'button' ? { type: 'button' } : {})}
         {...getOUIAProps('OverflowChip', ouiaId !== undefined ? ouiaId : this.state.ouiaStateId)}
       >
-        <span className={css(styles.chipText)}>{children}</span>
+        <span style={{ maxWidth: this.setMaxWidthText() }} className={css(styles.chipText)}>
```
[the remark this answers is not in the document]
bartoval: You are totally right @mcoker 🥇 
Many thanks for addressing me on the right path and for your detailed explanation and links. 👍🏼

mcoker (approved):
LGTM!

I just review the CSS, but I imagine you'll also want to preserve any styles someone may be passing as `style` currently, like

```
        {...(this.props.maxWidthText && {
          style: {
            this.setChipStyle(),
            ...props.style
          }
        })}
```

packages/react-core/src/components/Chip/Chip.tsx
```diff
@@ -38,6 +38,9 @@ export interface ChipProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
     | 'left-end'
     | 'right-start'
     | 'right-end';
+
+  /** Css property expressed in percentage or every css unit that override the default value of the max-width  */
```
tlabaj, line 42: I would updated to this 
```suggestion
  /** Css property expressed in percentage or any css unit that overrides the default value of the max-width of the chip's text  */
```

packages/react-core/src/components/Chip/Chip.tsx
```diff
@@ -38,6 +38,9 @@ export interface ChipProps extends React.HTMLProps<HTMLDivElement>, OUIAProps {
...
+  maxWidthText?: string;
```
tlabaj, line 43: nit.  I Think `textMaxWidth` would be more inline with how we name our props.

kmcfaul (approved):

PR #7388, opened 2022-05-09
chore(Dropdown): convert examples to TypeScript

### LINKED ISSUE #7383: Dropdown - convert examples to TypeScript
Dropdown

**Describe the feature**
The examples for the Dropdown component need to be converted to TypeScript and functional components.

N/A

**Any other information?**
N/A

thatblindgeye:
**What**: Closes #7383 

**Additional issues**:


thatblindgeye:
> The split button action and split button primary action examples are missing a gap between the dropdowns. It's a little confusing because it looks like a single dropdown toggle with a lot of icons/buttons on it.

Ah good catch! I added back in the `{' '}` that the examples originally had which adds that spacing back.

jenny-s51 (changes requested):
Demos LGTM, nice work Eric !

The [checkbox guidelines](https://www.patternfly.org/v4/components/checkbox/design-guidelines/#when-to-use-switches-versus-checkboxes) you linked refer to the 3rd state as an "intermediate" state so we should try to match those! 

Added some comments where these occurrences can be updated 🙂

jenny-s51 (approved):
Awesome! Yes, indeterminate definitely makes more sense here. In that case LGTM! 🚀

tlabaj (approved):

Issue #7392, opened 2022-05-10
Charts - add dark theme tooltip border styles


mcoker:
From core issue - https://github.com/patternfly/patternfly/issues/4841

PR #7393, opened 2022-05-10
fix(charts): support customizing tooltip border for dark theme

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/7392

@dlabrecq I'm using the var for a border width here - as far as I can tell, everything still seems to be positioned correctly. Is this OK or should we try and make this change some other way?


dlabrecq (commented):
Looking more closely, the border seems to make the pointer extend beyond the vertical cursor line. Perhaps if the border were set at 2 or less, that would not happen?

dlabrecq (approved):

nicolethoen (approved):

PR #7399, opened 2022-05-10
feat(Popper): add mouse events for popper content

### LINKED ISSUE #1457: Tooltip - persistent on hover
To meet AA WCAG 2.1 compliance, Tooltips should be persistent on hover (i.e. when you move the mouse from the tooltip toggle to the tooltip contents.

The WCAG item with more details about this: [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/quickref/#content-on-hover-or-focus)

thatblindgeye:
**What**: Closes #1457 

[Preview build - Tooltip](https://patternfly-react-pr-7399.surge.sh/components/tooltip)

Two new mouse events were added to the Popper component, both to resolve the above issue and for other components that use Popper to potentially utilize these events.

I also updated the default `exitDelay` for the tooltip with an explanation. Below is a video for the behavior when `exitDelay={0}` (_technically_ it works). I only chose a value of "300" to match the `entryDelay`, but if anyone has a default they'd prefer it can be tweaked.

https://user-images.githubusercontent.com/70952936/167679634-966e7641-6b79-48f6-ad1d-d94f83d3a50b.mov

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react-core/src/components/Tooltip/Tooltip.tsx
@@ -63,7 +63,10 @@ export interface TooltipProps extends Omit<React.HTMLProps<HTMLDivElement>, 'con
   enableFlip?: boolean;
   /** Delay in ms before the tooltip appears */
   entryDelay?: number;
-  /** Delay in ms before the tooltip disappears */
+  /** Delay in ms before the tooltip disappears, Avoid passing in a value of "0", as users should
+   * be given ample time to move their mouse from the trigger to the tooltip content without the content
+   * being hidden.
+   */
   exitDelay?: number;
   /**
    * The desired position to flip the tooltip to if the initial position is not possible.
@@ -149,7 +152,7 @@ export const Tooltip: React.FunctionComponent<TooltipProps> = ({
   enableFlip = true,
   className = '',
   entryDelay = 300,
-  exitDelay = 0,
+  exitDelay = 300,
   appendTo = () => document.body,
   zIndex = 9999,
   maxWidth = tooltipMaxWidth.value,
@@ -318,6 +321,8 @@ export const Tooltip: React.FunctionComponent<TooltipProps> = ({
       placement={position}
       onMouseEnter={triggerOnMouseenter && show}
       onMouseLeave={triggerOnMouseenter && hide}
+      onPopperMouseEnter={triggerOnMouseenter && show}
+      onPopperMouseLeave={triggerOnMouseenter && hide}
       onFocus={triggerOnFocus && show}
       onBlur={triggerOnFocus && hide}
       onDocumentClick={triggerOnClick && onDocumentClick}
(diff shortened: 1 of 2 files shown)


mcarrano:
I don't off the top of my head and would need to research this. But what's there now feels pretty good.

mcoker:
@jessiehuff I don't know either, it's not anything we've defined in core. Our flyout menu may also have an exit delay.

tlabaj:
@jessiehuff I know at one point Zack had updated our delays for tooltip.  I wonder if the exit delay here should match tooltip?

tlabaj:
@jessiehuff @mcarrano here is the tool tip PR https://github.com/patternfly/patternfly-react/pull/5746
I that PR we discussed whether changing the default is breaking as well since it would change behaviour in applications.  It was decided it was ok and more of a bug fix for the tooltip.

tlabaj (changes requested):
This may be considered a  breaking change.

tlabaj:
> @tlabaj if this would be considered a breaking change I can revert the `exitDelay` default back to 0 and add that update as something to make with our breaking change release. I think at least having the prop description be what it's been updated to should help and inform consumers about it in the mean time.

I don't know that it would be considered breaking.  @mcarrano with the tooltip change you said the exit delay update was warranted.  How do you feel about a change here?

thatblindgeye:
This PR mainly resolves the issue of keeping tooltips active when a mouse hovers over the actual tooltip contents rather than only the trigger. Tooltip just uses Popper under the hood, so I had to make an update to Popper to implement this feature.

I had also updated the `exitDelay` prop on the Tooltip due to the behavior seen in the video from my original comment (though that behavior is less because of the `exitDelay` being too short and more because the `exitDelay` occurs quicker than the `entryDelay` does). I do think having the exitDelay by default be longer than 0ms is worth making an update for at some point, though, just to ensure by default users have enough time to hover the tooltip should they need to.

tlabaj:
Ok, I was thinking the change was in popper for some reason.  My mistake. Now that I see it is in tooltip,  PR #5746 had changed the exit delay from 100 to zero.  So, I don't know that we want to increase it here.  @mcarrano what are your thoughts?

jessiehuff:
If the exit delay is zero, would the user be able to hover over the tooltip or would it interfere (like Eric's example above shows)? I'm wondering if the tooltip would disappear before they're able to get to it?

thatblindgeye:
You are able to hover the tooltip with an exitDelay of 0, but since the entryDelay is longer it causes that effect of the tooltip starting to hide and then re-appear when you hover it. Setting both entry and exit delays to the same number, or making the exit delay longer, resolves that visual issue.

mcarrano:
I don't know why we set the exit delay to 0 in the past, but seems to me that the entry and exit delays should be consistent. Can we just make them the same? Do you see any problem with this @tlabaj ?

tlabaj:
> I don't know why we set the exit delay to 0 in the past, but seems to me that the entry and exit delays should be consistent. Can we just make them the same? Do you see any problem with this @tlabaj ?
@mcarrano it would change the current default behaviour.   If we consider it a bug fix, that should be ok.

tlabaj (approved):
lgtm

PR #7402, opened 2022-05-11
Add additional type-casts for React nodes

jonkoops:
Since React 18 has stricter types than before some type-casts need to be added and expanded in order for the code to compile successfully.

Needed to land #7142

### DIFF
--- a/packages/react-core/src/components/DualListSelector/DualListSelector.tsx
+++ b/packages/react-core/src/components/DualListSelector/DualListSelector.tsx
@@ -134,10 +134,10 @@ export class DualListSelector extends React.Component<DualListSelectorProps, Dua
   private removeAllButtonRef = React.createRef<HTMLButtonElement>();
   static defaultProps: PickOptional<DualListSelectorProps> = {
     children: '',
-    availableOptions: [] as React.ReactNode[],
+    availableOptions: [],
     availableOptionsTitle: 'Available options',
     availableOptionsSearchAriaLabel: 'Available search input',
-    chosenOptions: [] as React.ReactNode[],
+    chosenOptions: [],
     chosenOptionsTitle: 'Chosen options',
     chosenOptionsSearchAriaLabel: 'Chosen search input',
     id: getUniqueId('dual-list-selector'),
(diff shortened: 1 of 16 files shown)


jonkoops:
@nicolethoen @tlabaj could this get reviewed and merged? I'd like to rebase the React 18 PR on top.

packages/react-core/src/components/DualListSelector/DualListSelector.tsx
```diff
@@ -271,16 +275,20 @@ export class DualListSelector extends React.Component<DualListSelectorProps, Dua
     this.setState(prevState => {
       const movedOptions =
         prevState.availableTreeFilteredOptions ||
-        flattenTreeWithFolders(prevState.availableOptions as DualListSelectorTreeItemData[]);
+        flattenTreeWithFolders((prevState.availableOptions as unknown) as DualListSelectorTreeItemData[]);
```
tlabaj, line 278: Curious.  Why `unknown` here and not `any`?

jonkoops: The `unknown` keyword still allows a stricter subset of type checking, whereas `any` disables all checks. But in this case I used `unknown` as it was the recommendation by the compiler.

nicolethoen (commented):
So im realizing that you're needing to cast these list as unknown since the chosenOptions and availableOptions props are both typed as React.ReactNode[] | DualListSelectorTreeItemData[].

It seems as though, since the tree feature was added later, we could tighten the typing of these various lists of options used throughout the component to avoid using the 'unknown' casting. Unless casting to unknown was something recommended as part of the migration to react 18 and is a fine solution. WDYT?

jonkoops:
Sorry for the late response, I have been focused on some other projects so I haven't had the time to get back to this.

> It seems as though, since the tree feature was added later, we could tighten the typing of these various lists of options used throughout the component to avoid using the 'unknown' casting.

Yes, I think it makes total sense to re-work these types. Ideally there would never be any `unknown` or `any` types present in the project. This is however not always possible and in this case I wanted to avoid a large refactor to this component.

> I do not know if that is blocker for this PR though. we could open a follow up issue to address that.

Agreed, let's open up an issue for this. Especially if this is still a 'beta' feature I think it makes sense to stabilize this before making it set in stone.

I think we can merge this as-is so I can continue creating new PRs to help with the React 18 migration. WDYT @nicolethoen @tlabaj?

tlabaj:
I opened this follow up issue for the dual list selector
https://github.com/patternfly/patternfly-react/issues/7548

nicolethoen (approved):

PR #7408, opened 2022-05-12
fix(Slider): add support for aria-labelledby ad aria-describedby

### LINKED ISSUE #7164: Slider - add support for aria-describedby / aria-labelled by
Slider

**Describe the feature**
We want to use the slider component in the context of a form. Similar to other form inputs, there are situations where we need to display messaging to the user about options they selected, and link the message to the input using `aria-describedby`. This is not currently supported in the current version of the Slider component. I expected that I could define `<Slider aria-describedby="xxx" ... /> and have that attribute passed down to the div with `role="slider"`.

While looking into this, I also noticed that `aria-labelledby` also does not appear to be supported for the slider component. This could be a nice enhancement, too, to provide a choice to reuse existing text in the UI.

none

**Any other information?**

bartoval:
**What**: Closes #7164


packages/react-core/src/components/Slider/Slider.tsx
```diff
@@ -61,6 +61,10 @@ export interface SliderProps extends Omit<React.HTMLProps<HTMLDivElement>, 'onCh
   leftActions?: React.ReactNode;
   /** Actions placed to the right of the slider */
   rightActions?: React.ReactNode;
+  /** Id to use for Slider descriptor */
+  'aria-describedby'?: string;
+  /** Id to use for Slider label */
+  'aria-labelledby'?: string;
```
thatblindgeye, line 67: The descriptions could be tweaked here to reflect that multiple IDs could be passed in, as well as what they're being attached to:

```suggestion
  /** One or more id's to use for the slider thumb description */
  'aria-describedby'?: string;
  /** One or more id's to use for the slider thumb label */
  'aria-labelledby'?: string;
```

thatblindgeye (commented):
Looks good! Only have one suggestion and one clarifying question below.

packages/react-core/src/components/Slider/_tests_/Slider.test.tsx
```diff
@@ -79,3 +79,13 @@ describe('slider', () => {
...
   });
 });
+
+test('renders slider with aria-labelledby', () => {
+  const { asFragment } = render(<Slider value={50} aria-labelledby="slider aria-labelledby" />);
+  expect(asFragment()).toMatchSnapshot();
+});
+
+test('renders slider with aria-describedby', () => {
+  const { asFragment } = render(<Slider value={50} aria-labelledby="slider aria-describedby" />);
```
thatblindgeye, line 89: This should be `aria-describedby="..."`

thatblindgeye (changes requested):
Just a small typo noted below.

packages/react-core/src/components/Slider/_tests_/Slider.test.tsx
```diff
@@ -79,3 +79,13 @@ describe('slider', () => {
...
+
+test('renders slider with aria-labelledby', () => {
+  const { asFragment } = render(<Slider value={50} aria-labelledby="slider aria-labelledby" />);
+  expect(asFragment()).toMatchSnapshot();
+});
+
+test('renders slider with aria-describedby', () => {
+  const { asFragment } = render(<Slider value={50} aria-describedby="slider aria-describedby" />);
+  expect(asFragment()).toMatchSnapshot();
+});
```
thatblindgeye, line 91: @wise-king-sullyman do you think instead of snapshots these tests could instead check for `toHaveAccessibleName` or `toHaveAttribute`?

thatblindgeye (approved):
Looks good! I pinged Austin in a comment below, but approving it for now and depending what he responds with I'll re-review if any further changes are made.

packages/react-core/src/components/Slider/_tests_/Slider.test.tsx
wise-king-sullyman: @thatblindgeye I *believe* so, another approach could probably be to use the `ByLabelText` selector. 

I'm not positive in the case of aria-describedby what the best route would be. I'm going to do some testing/investigation on that topic and report back shortly.

wise-king-sullyman: I think the first test would be more true to the way this functionality would be used by refactoring to something like:

```js
test('renders slider with aria-labelledby', () => {
  render(
    <>
      <p id="test">label text</p>
      <Slider value={50} aria-labelledby="test" />
    </>
  );

  const slider = screen.getByRole('slider', { name: 'label text' });

  expect(slider).toBeVisible();
});
```

The reasoning being that this actually tests that the label is being applied in a manner that is similar to how a user would experience it.

For the second test a very similar:

```js
test('renders slider with aria-describedby', () => {
  render(
    <>
      <p id="test">descriptive text about the slider</p>
      <Slider value={50} aria-describedby="test" />
    </>
  );

  const slider = screen.getByRole('slider', { description: 'descriptive text about the slider' });

  expect(slider).toBeVisible();
});
```

Is what I would recommend, *but* the `description` option for `getByRole` is a new feature in RTL, so the version for that package would need to be bumped to incorporate it. I'll put up a bump PR for it in just a moment as that option is definitely something I'd like to leverage going forward.

jessiehuff (approved):
LGTM other than the testing comments that Eric and Austin brought up.

jgiardino (approved):
Thanks for making this update! I was able to test in Safari and heard the aria-describedby text announced.

wise-king-sullyman (changes requested):
Looks perfect other than the test concerns that @thatblindgeye brought up. Once those are addressed I'll be happy to approve.

wise-king-sullyman (approved):
Looks perfect! 🚀

PR #7411, opened 2022-05-12
fix(Modal): tooltip visibility on modal open

### LINKED ISSUE #7112: If modal title has ellipsis, its tooltip is shown on modal open
If modal has a title that is too long to display, it will shorten it with ellipsis (...), hovering over shortened text shows tooltip with the full text that was shortened. This is correct and does work, but when you open a modal, this tooltip is displayed automatically without having to hover over the title. This can sometimes break automation as the tooltip blocks clicks (see [ESSNTL-1821](https://issues.redhat.com/browse/ESSNTL-1821)). In this example the tooltip is even rendered in the wrong place, but that would not happen if the tooltip didn't get shown on modal open. The tooltip is again closed only after you hover over the title again.

tooltip

Code sandbox: https://codesandbox.io/s/elated-hertz-0v6mb5?file=/index.js
Notice how clicking the button to open the modal shows a tooltip on the title and does not hide it until you hover over the title again.

This is a bug, which affects automation. For regular user might be visually unpleasant as it blocks view until you hover over the title again.
tooltip

Red Hat Insights Remediations

thatblindgeye:
**What**: Closes #7112 

While this PR resolves the issue linked, this does potentially produce a problem for keyboard users as they will be unable to display the tooltip since the title can't be focused. If this would be considered an issue (which if tooltips are being utilized, there's argument that the modal title *is* important), one solution to this would be to add `{...(hasTooltip && { tabIndex: 0 })}` on the ModalBoxTitle component so that keyboard users can tab to the title. This, however, a) goes against recommendations of not using tooltip on static elements (*this* could be resolved with a quick "don't use tooltips on static elements, except in cases of truncation"), and b) could also affect tests, as whether the title is focusable would have to be determined (though this might be easier to handle as it's more a case of taking into account an extra Tab press or not rather than a tooltip that blocks an unknown part of the page).

Another option would be to re-visit whether Modal titles should be truncated, or possibly adding a button that can be clicked to reveal the tooltip with the entire title contents (more of a popover/toggletip, though).

To test this fix:

1. Go to the [Modal preview build]() (link will be edited in upon build)
2. Copy the code below and replace the code in the first example code editor with it
3. Open the modal and notice that the tooltip does not automatically display

<details>
<summary>Click to display code block</summary>

```
import React from "react";
import { Modal, ModalVariant, Button, Bullseye } from "@patternfly/react-core";

class LargeModal extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      isModalOpen: false
    };
    this.handleModalToggle = () => {
      this.setState(({ isModalOpen }) => ({
        isModalOpen: !isModalOpen
      }));
    };
  }

  render() {
    const { isModalOpen } = this.state;

    return (
      <React.Fragment>
        <Bullseye>
          <Button variant="primary" onClick={this.handleModalToggle}>
            Show medium modal
          </Button>
        </Bullseye>
        <Modal
          variant={ModalVariant.medium}
          title="Very long title such that it overflows and causes the text to be shortened using an elipsis."
          isOpen={isModalOpen}
          onClose={this.handleModalToggle}
          actions={[
            <Button
              key="confirm"
              variant="primary"
              onClick={this.handleModalToggle}
            >
              Confirm
            </Button>,
            <Button
              key="cancel"
              variant="link"
              onClick={this.handleModalToggle}
            >
              Cancel
            </Button>
          ]}
        >
          Notice how clicking the button to open the modal shows a tooltip on
          the title and does not hide it until you hover over the title again.
        </Modal>
      </React.Fragment>
    );
  }
}
```

</details>

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Modal/ModalBoxTitle.tsx
+++ b/packages/react-core/src/components/Modal/ModalBoxTitle.tsx
@@ -37,7 +37,7 @@ export const ModalBoxTitle: React.FunctionComponent<ModalBoxTitleProps> = ({
   titleLabel = '',
   ...props
 }: ModalBoxTitleProps) => {
-  const [isTooltipVisible, setIsTooltipVisible] = React.useState(false);
+  const [hasTooltip, setHasTooltip] = React.useState(false);
   const h1 = React.useRef<HTMLHeadingElement>();
   const label = titleLabel || (isVariantIcon(titleIconVariant) ? `${capitalize(titleIconVariant)} alert:` : titleLabel);
   const variantIcons = {
@@ -50,7 +50,7 @@ export const ModalBoxTitle: React.FunctionComponent<ModalBoxTitleProps> = ({
   const CustomIcon = !isVariantIcon(titleIconVariant) && titleIconVariant;
 
   useIsomorphicLayoutEffect(() => {
-    setIsTooltipVisible(h1.current && h1.current.offsetWidth < h1.current.scrollWidth);
+    setHasTooltip(h1.current && h1.current.offsetWidth < h1.current.scrollWidth);
   }, []);
 
   const content = (
@@ -70,12 +70,6 @@ export const ModalBoxTitle: React.FunctionComponent<ModalBoxTitleProps> = ({
     </h1>
   );
 
-  return isTooltipVisible ? (
-    <Tooltip content={title} isVisible>
-      {content}
-    </Tooltip>
-  ) : (
-    content
-  );
+  return hasTooltip ? <Tooltip content={title}>{content}</Tooltip> : content;
 };
 ModalBoxTitle.displayName = 'ModalBoxTitle';


nicolethoen (approved):
I think design/interaction design needs to consider how this usage of tooltip currently goes against our best practices. It's not the only place we do this - I think table cells can be set to truncate with a tooltip when their contents are long as well.
So I think i'd recommend opening a follow up issue in pf-design to consider usage of tooltips with truncation in general. It could be as simple as adding the same styles that progress steps have when the step has an associate tooltip to indicate that they are interactable.
Not sure if @mmenestr, @mcarrano, or @mceledonia have additional thoughts about this.

mmenestr:
This is a tough one.  If I'm understanding this issue correctly, @nicolethoen I agree this is an issue that is present in multiple different cases where truncation happens so whatever we decide here should work for all the other cases of truncation. 

In terms of a solution, I'm inclined to go with the first one proposed which is to change the recommendation to "don't use tooltips on static elements, except in cases of truncation", and let a static element be focusable to reveal a tooltip. 

Or like @thatblindgeye suggests, maybe we could link into wrapping a modal header as an option, instead of truncating it ? But I wouldn't recommend that solution for tables, so that becomes tricky.

mcarrano (commented):
I'll defer to @mmenestr on this one.

jessiehuff (commented):
This is definitely an interesting and good question. A few articles out there on this subject:

- To support the statement that tooltips should not be used for static elements, this is a good article: https://sarahmhigley.com/writing/tooltips-in-wcag-21/#best-practices-summary
- And then Heydon Pickering also notes that tooltips should be used sparingly: https://inclusive-components.design/tooltips-toggletips/
- In this article the author makes a good point that truncation usually favors the designer, not the user: https://www.bennadel.com/blog/3624-reconsidering-text-overflow-ellipsis-as-a-design-smell-and-accessibility-concern.htm

I also spoke to Jenn on this, and I think we lean towards the opinion that truncation simplifies design and development, often at the expense of the user. In the context of a table where we want to provide a short snippet of the contents to help with finding the table item you're looking for, truncation could make sense, but seeing the full contents in a tooltip should not be the only place where full contents are available. In a table view, it's possible to provide the option to navigate to the full details view. In a modal, it's much more complicated. 

Jenn also had an idea that I figured I could just throw out there and get opinions on:

> Some of the issues that need to be addressed by displaying the full title are how to handle scrolling. If the title does occupy 25% of the viewport height vs 10% of it, that also hurts the user. Coming up with a solution that is elegant and solves all the problems is definitely tricky. I’m not sure if there’s an option to limit truncation to display only on scrolling. For example, display the full title initially, then as the user scrolls down, the title text becomes truncated to a single line. I have no idea if that’s even possible, but something like that provides the best of both worlds (title contents fully visible in context vs utilizing as much of the viewport height as possible for the contents of the modal)

Basically, it does seem a bit odd to always have the modal title truncated and only have the full context available in a tooltip. In general, that feels like it favors a good looking design over usability, but that's just my opinion. However, if we decide that makes the most sense here then I would at least think it would need to be keyboard accessible like Eric points out.

mmenestr:
I do tend to always favor the tooltip approach, although I know it's not the most accessible. The only alternative I see is wrapping like we mentioned above - but we do get into the potential problem of having an extremely long header that takes up most of the height of the modal - which we don't want. 

So I'd either 
1) Use the tooltip, but make sure its accessible via keyboard like you suggest @thatblindgeye OR
2) Go ahead and wrap, with the risk of having modals with very long headers, and rely on the design guidelines to advise against modal headers that are longer than two lines (?). Overall, I don't see it being a common occurrence.. I feel like designers usually have full modal headers in their mocks, and they don't truncate elements in their designs - they usually wrap (at least I do). We always try to make headers as concise as possible, and put any varying information (like item names) in the body of a modal, to account for a potential long header scenario.

mcarrano:
Getting back to the original issue... I remember that we made a change some time ago to support modal titles that overflowed by truncating them with a tooltip. Maybe that was the wrong approach. @mcoker @mceledonia do you remember the discussion about that? Would allowing the title to wrap be an alternative without causing other issues?

IMO, modal titles should always be short and in theory we should not have this problem. But I understand that we need to account for it. I'd accept an ugly solution if it reduces complexity.

mcoker:
@mcarrano as far as I remember, we truncate modal titles so that this doesn't happen - the body content is too short to read since we limit the modal height to the viewport height:

Screen Shot 2022-05-24 at 6 34 56 PM

However I question if truncation should be the default - it should probably be left to the user to ensure there is enough space left over from the other stuff in the modal to show the body content in this edge case, instead of truncating the modal title for everyone.

thatblindgeye:
In regards to the issue at hand, @jessiehuff and @mcarrano and/or @mmenestr , do you feel comfortable with the solution made with this PR to resolve a tooltip displaying automatically when a modal opens? Or would the issue of modal title not being fully visible to users navigating via keyboard that is introduced be more counterproductive than resolving the linked issue here is worth?

Because the truncation discussion isn't a simple thing to decide on, I can make a separate issue and link to this PR for the background, and that could have the breaking change label added to it if necessary (even if only for now). Jenn's idea does seem interesting, and we have a couple other options to consider as well.

mcarrano:
@jessiehuff I will reply with a comment on the original issue.

tlabaj (approved):
lgtm

mcarrano (approved):
I'm good to make this change. Thanks @thatblindgeye !

jessiehuff (approved):
Thanks Eric! :)

PR #7414, opened 2022-05-13
refactor(Search Input): use TextInputGroup instead native HTML elements

### LINKED ISSUE #7121: Search input - refactor to use Text Input Group under the hood
Follow up to https://github.com/patternfly/patternfly/issues/4614

bartoval:
**What**: Closes #7121

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -56,7 +56,10 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   isAdvancedSearchOpen?: boolean;
   /** Label for the button which opens the advanced search form menu */
   openMenuButtonAriaLabel?: string;
-
+  /** Label for the button to navigate to previous result  */
+  previousNavigationButtonAriaLabel?: string;
+  /** Label for the button to navigate to next result */
+  nextNavigationButtonAriaLabel?: string;
   /** Function called when user clicks to navigate to next result */
   onNextClick?: (event: React.SyntheticEvent<HTMLButtonElement>) => void;
   /** Function called when user clicks to navigate to previous result */
@@ -98,6 +101,8 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
   'aria-label': ariaLabel = 'Search input',
   resetButtonLabel = 'Reset',
   openMenuButtonAriaLabel = 'Open advanced search',
+  previousNavigationButtonAriaLabel = 'Previous',
+  nextNavigationButtonAriaLabel = 'Next',
   submitSearchButtonLabel = 'Search',
   isDisabled = false,
   ...props
@@ -197,13 +202,18 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
             <div className="pf-c-text-input-group__group">
               <Button
                 variant={ButtonVariant.plain}
-                aria-label="Previous"
+                aria-label={previousNavigationButtonAriaLabel}
                 isDisabled={isDisabled}
                 onClick={onPreviousClick}
               >
                 <AngleUpIcon />
               </Button>
-              <Button variant={ButtonVariant.plain} aria-label="Next" isDisabled={isDisabled} onClick={onNextClick}>
+              <Button
+                variant={ButtonVariant.plain}
+                aria-label={nextNavigationButtonAriaLabel}
+                isDisabled={isDisabled}
+                onClick={onNextClick}
+              >
                 <AngleDownIcon />
               </Button>
             </div>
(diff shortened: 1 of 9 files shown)


nicolethoen (changes requested):
The resulting HTML markup of these search input examples should match that of [these text input group examples in the html docs](https://www.patternfly.org/v4/components/text-input-group/html#search-input-group).
Ultimately, you should (hopefully) be able to remove `import styles from '@patternfly/react-styles/css/components/SearchInput/search-input';`

bartoval:
> The resulting HTML markup of these search input examples should match that of [these text input group examples in the html docs](https://www.patternfly.org/v4/components/text-input-group/html#search-input-group). Ultimately, you should (hopefully) be able to remove `import styles from '@patternfly/react-styles/css/components/SearchInput/search-input';`

Hey!
Thanks for these suggestions 💯

nicolethoen (commented):
This is looking good! There is still an extra wrapper `div` which is always wrapping the text-input-group in the react examples that is not in the html examples. Can we remove that?

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -178,116 +178,115 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
...
+            <div className="pf-c-text-input-group__group">
+              <Button
+                variant={ButtonVariant.plain}
+                aria-label="Previous"
+                isDisabled={isDisabled}
+                onClick={onPreviousClick}
+              >
+                <AngleUpIcon />
+              </Button>
+              <Button variant={ButtonVariant.plain} aria-label="Next" isDisabled={isDisabled} onClick={onNextClick}>
```
tlabaj, line 206: we need the ability to customize this label for localization

tlabaj (changes requested):

tlabaj (approved):
LGTM

bartoval:
> This looks great! 👍 A few questions about `disabled` states for buttons.
> 
> Looks like the previous button in [match with navigable options](https://patternfly-react-pr-7414.surge.sh/components/search-input#match-with-navigable-options) should be disabled at match `1`.
> 
> Screen Shot 2022-05-26 at 9 06 58 AM
> 
> Should the submit button in [With submit button](https://patternfly-react-pr-7414.surge.sh/components/search-input#with-submit-button) and [Advanced](https://patternfly-react-pr-7414.surge.sh/components/search-input#advanced) should be disabled by default as well?
> 
> Screen Shot 2022-05-26 at 9 11 28 AM
> 
> Screen Shot 2022-05-26 at 9 14 17 AM

Thanks, @mattnolting, you are right 👍🏼 
So I implemented them.

mattnolting:
@bartoval Great, thanks! Looks like the [Advanced example](https://patternfly-react-pr-7414.surge.sh/components/search-input/#advanced) has a last glitch.

Screen Recording 2022-05-26 at 1 02 15 PM

bartoval:
> @bartoval Great, thanks! Looks like the [Advanced example](https://patternfly-react-pr-7414.surge.sh/components/search-input/#advanced) has a last glitch.
> 
> Screen Recording 2022-05-26 at 1 02 15 PM [ ](https://user-images.githubusercontent.com/5385435/170538806-69bdc2b8-fa9e-40a4-bd80-e2ea873125ae.gif) Screen Recording 2022-05-26 at 1 02 15 PM [ ](https://user-images.githubusercontent.com/5385435/170538806-69bdc2b8-fa9e-40a4-bd80-e2ea873125ae.gif)

thanks again for your eagle eye 🥇 
Fixed

thatblindgeye (commented):
Awesome work on this!

Similar to @mattnolting comment regarding [match with navigable options](https://patternfly-react-pr-7414.surge.sh/components/search-input#match-with-navigable-options), I think it'd make sense to also add similar functionality to disable next button when the current result is at the last result.

thatblindgeye (approved):
💪🏼

mattnolting (approved):
Fantastic! 🚢  it!

bartoval:
> Looks like we were both pressing a button on this PR at the same time! Approving again just for visibility in the "reviewers" sidebar.

lol https://media.giphy.com/media/TbYgHMnICI1A4/giphy.gif

mmenestr:
Overall, this looks good to me - I'm just wondering why we decided to disable the submit button in the case, until someone actually types something into the search area? I haven't seen us do that in the past (for example, when we used the input group for search).

I don't think there's anything wrong with that approach, but just wonder why we're changing the usual approach here!

mmenestr:
Oh, missed that! Ok, from a consistency standpoint, we don't usually do this, but I guess it makes sense, so we can ship it!! :)

mmenestr (approved):

wise-king-sullyman (approved):
Looks 🔥

bartoval:
> > Overall, this looks good to me - I'm just wondering why we decided to disable the submit button in the case, until someone actually types something into the search area? I haven't seen us do that in the past (for example, when we used the input group for search).
> > I don't think there's anything wrong with that approach, but just wonder why we're changing the usual approach here!
> 
> @mattnolting had suggested this above.

I also think this feat improves the accessibility (the aria-l
Screenshot from 2022-05-31 15-48-10
abel became true) when you can't proceed

nicolethoen (commented):
The only thing I noticed is that the original advanced search example does not append the advanced form inline. It lays over the code editor and other example controls. In the updated example, it moves those controls down. Is that intended? @mmenestr @mattnolting do you have thoughts on that? I fear that is a change in default behavior that will surprise consumers.

Old example:
Screen Shot 2022-05-31 at 9 58 55 AM

updated example:
Screen Shot 2022-05-31 at 9 59 19 AM

nicolethoen (approved):

garrett:
When attempting to upgrade to the latest release of PatternFly, we spotted an issue due to failing tests. It appears to be related to this PR, as it changes the previous widgets used from HTML to the PF Panel component. Previously, it used the `pf-c-search-input__menu` class, but now it uses these classes: `pf-c-panel pf-m-raised pf-c-search-input__menu`.

However, pf-c-panel makes the popover `position: relative` now, instead of `position: absolute` and the width of the popover is now `auto` instead of being 100% wide.

As a result of the changes, the panel popover for the advanced filter still reserves the space in the toolbar, causing a layout shift when it appears, which looks like this:

relative whoops

(You can see the panel below the toolbar still... but as it's relative positioning instead of absolute, it now affects the flow of the content in the toolbar.)

PR #7417, opened 2022-05-13
fix(ComposableMenu): update appendTo in demos

### LINKED ISSUE #7291: Composable menu - ensure examples/demo are appending menu correctly
Most of our composable menu demos are appending the menu to the document body.

The MenuToggle is a button by default - so we cannot append the Menu to the button or the Menu will be an interactable element inside an interactable button. 

Some possible suggestions:
- Make the menu toggle a div rather than a button for the sake of being able to append a menu to it
- build a container to place a menutoggle and menu inside
- reevaluate the appendto parent recommendation

thatblindgeye:
**What**: Closes #7291

For the demos I followed how the application launcher demo had it setup by wrapping the returned Popper component in a `div`. It looked to work well for all demos, though the flyout one rendered flyout menus incorrectly (this behavior also appears in the React workspace without this wrapper `div` however).

[Composable menu preview build](https://patternfly-react-pr-7417.surge.sh/demos/composable-menu) 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/demos/ComposableMenu/examples/ComposableActionsMenu.tsx
+++ b/packages/react-core/src/demos/ComposableMenu/examples/ComposableActionsMenu.tsx
@@ -10,6 +10,7 @@ export const ComposableActionsMenu: React.FunctionComponent = () => {
   const [selectedItems, setSelectedItems] = React.useState<number[]>([]);
   const toggleRef = React.useRef<HTMLButtonElement>();
   const menuRef = React.useRef<HTMLDivElement>();
+  const containerRef = React.useRef<HTMLDivElement>();
 
   const handleMenuKeys = (event: KeyboardEvent) => {
     if (isOpen && menuRef.current.contains(event.target as Node)) {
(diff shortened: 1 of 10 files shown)


nicolethoen (commented):
It seems like the screen reader is not interacting well for me with the drilldown demo, nor the flyout demo.
I cant seem to trigger the flyouts to open in the flyout demo
firstone
The screen reader things the drilldown submenus are next in the DOM order even when they are hidden.
second one
These maybe simple fixes, but if not, we can open follow up issues. They were pre-existing as far as I can tell.

thatblindgeye:
In regards to the issue seen with the drilldown variant, I opened https://github.com/patternfly/patternfly/issues/4870 in Core.

For the flyout variant, I opened #7461

tlabaj (approved):
LGTM

jenny-s51 (approved):

jessiehuff (approved):
Looks fairly good. I see the same issues that you mention above, but since you opened issues for those, I'll approve. :) I think it's worth noting though that these issues also seem to appear visually for keyboard users: 
8E4638B4-9A9A-4095-B72C-39A0B567A386

mcoker (approved):
LGTM!

Issue #2879, opened 2019-09-06
Checkbox <input> never has checked atrribute in DOM


redallen:
Unless we pass `defaultChecked` to a checkbox, React doesn't change the `checked` attribute in the DOM. This is default browser behavior. [React has a related issue that's been open for years](https://github.com/facebook/react/issues/6321#issuecomment-201040512). However, changing the attribute makes it easier for Selenium testing to take place.

We should evaluate whether we want to make the attribute change in the DOM like with #2749, or revert it and instead rely on checking the `value` property of the `<input>`.

redallen:
> The `<input type="checkbox">` defines a checkbox.
- [W3Schools](https://www.w3schools.com/tags/att_input_type_checkbox.asp)

By default, the browser does not touch the `checked` attribute that may be defined on an `<input>` element like `<input checked="">` when a user checks or unchecks a checkbox. This is apparent in [a simple example.](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_checkbox)

As you can tell, after checking the "I have a car," the `checked` attribute does not appear in the DOM. After unchecking the "I have a boat," the `checked=""` attribute still remains in the DOM.
![image](https://user-images.githubusercontent.com/47335686/65518882-040d3f00-deb3-11e9-812a-a80dbf084fa1.png)

This implementation is consistent across Firefox and Chrome. Indeed, the only way to tell if the checkbox is checked or not is to look directly at the DOM element's `checked` property:
![image](https://user-images.githubusercontent.com/47335686/65519147-8138b400-deb3-11e9-91fd-80ea86a2e06b.png)

I opened #2749 to make it easier for QE to just look at the `checked` attribute rather than the `checked` property in the DOM for our [Switch component.](https://patternfly-react.surge.sh/patternfly-4/components/switch/) I believe it is better is to revert it and instead rely on [Selenium's getProperty](), for which I have provided a simple example using Python 2 that works:
```python
  1 from selenium import webdriver
  2 from selenium.webdriver.support.ui import WebDriverWait
  3 from selenium.webdriver.support import expected_conditions as EC
  4 from selenium.webdriver.common.by import By
  5 from selenium.common.exceptions import TimeoutException
  6 
  7 driver = webdriver.Chrome()
  8 driver.get("https://patternfly-react.surge.sh/patternfly-4/components/checkbox")
  9 delay = 3 # seconds
 10 try:
[...]

Issue #2791, opened 2019-08-28
Uncontrolled switch is not working


boaz0:
The uncontrolled examples of switch don't work correctly.
Clicking the switch doesn't check or uncheck it.

boaz0:
@rachael-phillips note that this may be fixed by https://github.com/patternfly/patternfly-react/pull/2749

tlabaj:
closed by PR #2749

PR #7418, opened 2022-05-13
fix/docs(Switch): fix uncontrolled examples

### LINKED ISSUE #7255: Bug - [Switch] - [uncontrolled examples are not working]
**Describe the problem**
The uncontrolled switch examples do not toggle.

Provide steps to reproduce. A [codesandbox](https://codesandbox.io/s/serverless-cherry-q9t3f ) demonstrating the problem is appreciated.
https://www.patternfly.org/v4/components/switch#uncontrolled

kmcfaul:
**What**: Closes #7255

@tlabaj Did these work before? Uncontrolled inputs shouldn't be setting `isChecked`, and half of the example switches are redundant now that it works. I'm curious if this was meant as a static demo initially.

### DIFF
--- a/packages/react-core/src/components/Switch/examples/Switch.md
+++ b/packages/react-core/src/components/Switch/examples/Switch.md
@@ -176,19 +176,9 @@ import { Switch } from '@patternfly/react-core';
     aria-label="Message when on"
     label="Message when on"
     labelOff="Message when off"
-    isChecked
+    isChecked={null}
   />
   <br />
-  <Switch
-    id="uncontrolled-switch-off"
-    aria-label="Message when on"
-    label="Message when on"
-    labelOff="Message when off"
-    isChecked={false}
-  />
-  <br />
-  <Switch id="uncontrolled-no-label-switch-on" aria-label="Message when on" isChecked />
-  <br />
-  <Switch id="uncontrolled-no-label-switch-off" aria-label="Message when on" isChecked={false} />
+  <Switch id="uncontrolled-no-label-switch-on" aria-label="Message when on" isChecked={null} />
 </React.Fragment>;
 ```
--- a/packages/react-core/src/components/Switch/examples/Switch.md
+++ b/packages/react-core/src/components/Switch/examples/Switch.md
@@ -171,14 +171,8 @@ import React from 'react';
 import { Switch } from '@patternfly/react-core';
 
 <React.Fragment>
-  <Switch
-    id="uncontrolled-switch-on"
-    aria-label="Message when on"
-    label="Message when on"
-    labelOff="Message when off"
-    isChecked={null}
-  />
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/Switch/examples/Switch.md
```diff
@@ -176,19 +176,9 @@ import { Switch } from '@patternfly/react-core';
     aria-label="Message when on"
     label="Message when on"
     labelOff="Message when off"
-    isChecked
+    isChecked={null}
```
jenny-s51, line 179:
```suggestion
    isChecked={null}
    defaultChecked
```

jenny-s51 (changes requested):
Great update @kmcfaul !! It's really cool that we can get an uncontrolled input by setting isChecked to null! 

According to the react docs, it looks like we can [specify an initial value using defaultChecked but leave subsequent updates uncontrolled](https://reactjs.org/docs/uncontrolled-components.html#default-values).  

I see some `defaultChecked` logic in Checkbox.tsx`, `Radio.tsx`, both of which also have uncontrolled examples, maybe we can expose it in the props and use it in a similar way here? WDYT?

tlabaj (commented):
@kmcfaul I am not sure if it did work before.  I would need tog look at the history

packages/react-core/src/components/Switch/examples/Switch.md
```diff
@@ -176,19 +176,9 @@ import { Switch } from '@patternfly/react-core';
     aria-label="Message when on"
```
nicolethoen, line 176: Should we also update the aria-label when appropriate to demonstrate the intention? it's possible that, when it's uncontrolled, there are a few extra considerations to make to make sure the aria-label is meaningful. Like, what should we have the aria-label say if it wont change when the switch is on/off?

tlabaj (changes requested):
Hmmm.  Katie, I think you were right, this probably never worked before.  I would expected it to work like the checkbox or radio in that you do not need to pass `isChecked` to make it uncontrolled.  I think we need to update the code so that is the case.  We can also support `isChecked` being passed as `null` in order to not break people...

tlabaj:
@kmcfaul it looks like this issue came up before, there was an attempt to fix it that was reverted.  Here is the original issue https://github.com/patternfly/patternfly-react/issues/2791
See this related issue too
https://github.com/patternfly/patternfly-react/issues/2879

packages/react-core/src/components/Switch/examples/Switch.md
thatblindgeye: I opened issue #7468 for this issue as well as an issue I noticed with the `aria-labelledby` attribute.

tlabaj (approved):
LGTM

jenny-s51 (approved):
LGTM!  👍  Nice work on this @kmcfaul @thatblindgeye

PR #7426, opened 2022-05-16
feat(HelperText): add screenreader text for dynamic variant

### LINKED ISSUE #7263: HelperText - dynamic variant demo should include a live region
HelperText

**Describe the feature**
Currently the [dynamic variant with static text demo](https://patternfly-react-pr-7250.surge.sh/components/helper-text/react-demos/#dynamic-variant-with-static-text) does not notify screen readers when helper text items update from an "error" to a "success", due to the icons having the `aria-hidden` attribute and the text itself not updating outside of styling.

Screen reader users should be notified when the status/variant of these helper text items get updated. Conversation regarding this feature can be see in PR #7250 . Depending on the implementation used, the a11y docs for this component may also need updating.

N/A

**Any other information?**
N/A

thatblindgeye:
**What**: Closes #7263 

[HelperText dynamic demo preview build](https://patternfly-react-pr-7426.surge.sh/components/helper-text/react-demos#dynamic-variant-with-static-text)

Rather than utilizing `aria-live`, I instead added an element to the HelperTextItem with the `pf-u-screen-reader` class. That plus `aria-describedby` does seem to work with VO (would have to test with JAWS and NVDA). I wasn't sure whether we would want to expose the ability to customize this message for screen readers so I opted to not include that right now.

Another option could be using the `output` element which does work nicely with its implicit live region, but the issue then becomes repetitive text being announced from the live region and `aria-describedby` when a user begins typing in the input field. 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/HelperText/HelperTextItem.tsx
+++ b/packages/react-core/src/components/HelperText/HelperTextItem.tsx
@@ -70,7 +70,11 @@ export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
           {variant === 'error' && <ExclamationCircleIcon />}
         </span>
       )}
-      <span className={css(styles.helperTextItemText)}>{children}</span>
+
+      <span className={css(styles.helperTextItemText)}>
+        {children}
+        {isDynamic && <span className="pf-u-screen-reader">: {variant} status;</span>}
+      </span>
     </Component>
   );
 };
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/HelperText/HelperTextItem.tsx
```diff
@@ -70,7 +70,11 @@ export const HelperTextItem: React.FunctionComponent<HelperTextItemProps> = ({
           {variant === 'error' && <ExclamationCircleIcon />}
         </span>
       )}
-      <span className={css(styles.helperTextItemText)}>{children}</span>
+
+      <span className={css(styles.helperTextItemText)}>
+        {children}
+        {isDynamic && <span className="pf-u-screen-reader">: {variant} status;</span>}
```
nicolethoen, line 76: I don't think you can hardcode the word status, because it needs to be able to be translated programmatically, but it could use the word status by default.

tlabaj (approved):

PR #7434, opened 2022-05-17
fix(Select): popper menu flips correctly

### LINKED ISSUE #7286: [Dropdown like components using menuAppendTo and popper] - [add ability to enable flip for popper]
Select, Context Selector, Dropdown

**Describe the feature**
Add a prop to allow the consumer to enable flip for popper.

### LINKED ISSUE #7398: Bug - OverflowMenu- Always placed on right
**Describe the problem**
OverflowMenu cannot be placed to left.(like tooltip) It opens always as "placed to the right" which makes it unusable on the righthand side of the screen

Place overflowMenu dropdown on right hand side of screen

**Expected behavior**
some property to open it as left placed(like tooltip)

**Screenshots**
Screenshot 2022-05-06 at 4 58 43 PM

Screenshot 2022-05-05 at 7 05 04 PM

**What is your environment?**
Chrome on MacOS

Openshift 4.11

**Any other information?**

thatblindgeye:
**What**: Closes #7286 

This PR does not resolve issue #7398 as the Overflow Menu component itself doesn't use Popper or `menuAppendTo`

[Context selector](https://patternfly-react-pr-7434.surge.sh/components/context-selector/)
[Dropdown](https://patternfly-react-pr-7434.surge.sh/components/dropdown)
[Select](https://patternfly-react-pr-7434.surge.sh/components/select)

To test, copy+paste the following to the returned component in an example from the above links:

```
menuAppendTo="parent"
isFlippable
```

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
+++ b/packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
@@ -29,7 +29,8 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
     direction: DropdownDirection.down,
     onSelect: (): void => undefined,
     autoFocus: true,
-    menuAppendTo: 'inline'
+    menuAppendTo: 'inline',
+    isFlippable: false
   };
 
   constructor(props: DropdownProps & OUIAProps) {
@@ -75,6 +76,7 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
       toggle,
       autoFocus,
       menuAppendTo,
+      isFlippable,
       ...props
     } = this.props;
     const id = toggle.props.id || `pf-dropdown-toggle-id-${DropdownWithContext.currentId++}`;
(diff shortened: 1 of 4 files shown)


thatblindgeye:
> Looking good! I'm noticing Select's Panel example is not flipping. It's probably built differently than the other select variants' menus.

Below is the behavior I'm seeing with the Panel example (on both appending to parent and document body + the `isFlippable` prop):

https://user-images.githubusercontent.com/70952936/169140728-f7fbbbc2-ccc9-4956-9d44-4df124ca6c98.mov

jenny-s51 (approved):

tlabaj (changes requested):
looks good jut some small suggestions..

tlabaj (approved):

packages/react-core/src/components/Dropdown/DropdownWithContext.tsx
```diff
@@ -96,6 +98,9 @@ export class DropdownWithContext extends React.Component<DropdownProps & OUIAPro
           const BaseComponent = baseComponent as any;
           const menuContainer = (
             <DropdownMenu
+              // This removes the `position: absolute` styling from the `.pf-c-dropdown__menu`
+              // allowing the menu to flip correctly
+              {...(isFlipEnabled && { style: { position: 'revert', minWidth: 'min-content' } })}
```
mcoker, line 103: Looks like this `min-width` is needed because the positioned popper element is assigned a `width`, which is the same width as the toggle (as far as I can tell). So the `__menu` is currently overflowing the positioned popper element, even though it looks OK.

On the context selector, however, here is what a non-positioned versus popper positioned menu looks like since the positioned menu has a fixed width:

non-positioned
Screen Shot 2022-05-26 at 2 58 15 PM

popper positioned
Screen Shot 2022-05-26 at 2 58 28 PM

Can we nix the width? Or maybe assign a `min-width` instead?

Issue #7436, opened 2022-05-17
Topology - Need ability to show details of nodes when zoomed out


jeff-phillips-18:
**What is the existing component, if any?
Topology

**Describe the feature**
When zoomed out nodes are displayed with less info, just the shape and colored status. We need a way for users to get back the full details when at full scale. On hover, the node should scale back up and be shown at full size with all the features of the original node.

PR #7437, opened 2022-05-17
feat(Topology): Add node scaling on hover at low details scale level

jeff-phillips-18:
**What**: 
Closes https://github.com/patternfly/patternfly-react/issues/7436

**Screen Shot**:

topology-zoom-scale

/cc @christianvogt

### DIFF
--- a/packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/components/StyleNode.tsx
+++ b/packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/components/StyleNode.tsx
@@ -2,16 +2,21 @@ import * as React from 'react';
 import {
   Decorator,
   DEFAULT_DECORATOR_RADIUS,
+  DEFAULT_LAYER,
   DefaultNode,
   getDefaultShapeDecoratorCenter,
+  Layer,
   Node,
   NodeShape,
   NodeStatus,
   observer,
   ScaleDetailsLevel,
   ShapeProps,
+  TOP_LAYER,
   TopologyQuadrant,
+  useHover,
   WithContextMenuProps,
+  WithCreateConnectorProps,
   WithDragNodeProps,
   WithSelectionProps
 } from '@patternfly/react-topology';
@@ -40,6 +45,7 @@ type StyleNodeProps = {
   regrouping?: boolean;
   dragging?: boolean;
 } & WithContextMenuProps &
+  WithCreateConnectorProps &
   WithDragNodeProps &
   WithSelectionProps;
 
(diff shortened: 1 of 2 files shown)


jeff-phillips-18:
Demo: https://pf-topology-zoom-scale.surge.sh/

packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/components/StyleNode.tsx
```diff
@@ -138,24 +145,32 @@ const StyleNode: React.FunctionComponent<StyleNodeProps> = ({
...
-      onContextMenu={data.showContextMenu ? onContextMenu : undefined}
-      contextMenuOpen={contextMenuOpen}
-      labelIcon={LabelIcon && <LabelIcon noVerticalAlign />}
-      attachments={
-        detailsLevel === ScaleDetailsLevel.high && renderDecorators(element, passedData, rest.getShapeDecoratorCenter)
-      }
-    >
-      {detailsLevel !== ScaleDetailsLevel.low && renderIcon(passedData, element)}
-    </DefaultNode>
+    <Layer id={hover ? TOP_LAYER : DEFAULT_LAYER}>
```
christianvogt, line 148: This is causing every hovered node to swap to the top layer. I don't believe that was intended behavior?

packages/react-topology/src/components/nodes/DefaultNode.tsx
```diff
@@ -192,9 +204,52 @@ const DefaultNode: React.FunctionComponent<DefaultNodeProps> = ({
...
+      const scaleGoal = 1 / scale;
+      const scaleBump = (scaleGoal - 1) / SCALE_UP_STEPS;
+      const bumpScale = (scaleStep: number = 0) => {
+        setNodeScale(1 + scaleBump * scaleStep);
+        if (scaleStep < SCALE_UP_STEPS) {
+          handle = setTimeout(() => {
+            bumpScale(scaleStep + 1);
+          }, SCALE_UP_INTERVAL / SCALE_UP_STEPS);
+        }
+      };
```
christianvogt, line 224: I find the scaling animation a bit slow.
You should use `requestAnimationFrame` instead of timeout for animations.

Furthermore the scaling value should be based on time and not step intervals.
That way the scaling timing is consistent always and let `requestAnimationFrame` dictate how many frames we can actually render smoothly in the allotted time.

packages/react-integration/demo-app-ts/src/components/demos/TopologyDemo/components/StyleNode.tsx
jeff-phillips-18: It was intended. Without this when the node is enlarged, there can be other nodes that intersect and draw over portions of the enlarged node.

christianvogt: but even when at full scale, the same hover behavior is new.

packages/react-topology/src/components/nodes/DefaultNode.tsx
jeff-phillips-18: Updated and handled unmount.

jessiehuff (approved):
LGTM! :)

PR #7444, opened 2022-05-19
fix(Select): allow enter to create when there is a single option

### LINKED ISSUE #7260: Bug - Select Typeahead isCreatable - Cannot create using a keayboard
**Describe the problem**

Previously, it was possible to add new values with pressing enter:

```jsx
    userEvent.type(screen.getByLabelText('input field'), 'option 1{enter}option 2{enter}option 3{enter}');
```

Now, it does not work. 

1. Go to https://www.patternfly.org/v4/components/select#typeahead

2. Check `isCreatable` to true.

3. Try to create a new value using only a keyboard

**Expected behavior**

Pressing enter when typing should enter a new value.

It is blocking our testing suite, it is possible to avoid it using a pointer.

**What is your environment?**
 - OS: MacOS
 - Browser Chrome
 - Version latest

WebRCA

kmcfaul:
**What**: Closes #7260

This issue was caused by #6966 which prevents enter from selecting an option in typeahead until `ArrowDown` is pressed to navigate into the menu. Creating still works with `ArrowDown` + `Enter`, and I'm not sure if this is a pattern we want to hold for all cases in typeahead. This PR will add an exception for `isCreatable` typeaheads where the create option is the only option in the list (creating a new option where there are still multiple options shown will still require arrow navigation to the create button).

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -679,13 +679,13 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
   };
 
   handleTypeaheadKeys = (position: string, shiftKey: boolean = false) => {
-    const { isOpen, onFavorite } = this.props;
+    const { isOpen, onFavorite, isCreatable } = this.props;
     const { typeaheadCurrIndex, tabbedIntoFavoritesMenu } = this.state;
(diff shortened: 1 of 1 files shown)


tlabaj (approved):

thatblindgeye (approved):
Only thing I noticed is the "create" option doesn't have any focus style applied when navigating to it, but not something to block this particular PR over I don't think.

PR #7446, opened 2022-05-19
feat(table): convert column mgmt demos to fullscreen

### LINKED ISSUE #7357: Feat(Table): Full screen demo for Column Management Modal

jenny-s51:
**What**: Closes #7357

### DIFF
--- a/packages/react-table/src/demos/Table.md
+++ b/packages/react-table/src/demos/Table.md
@@ -293,10 +293,11 @@ class BulkSelectTableDemo extends React.Component {
 
 ### Column management
 
-```js
+```js isFullscreen
 import React from 'react';
 import {
   Button,
+  Card,
   DataList,
   DataListCheck,
   DataListItem,
@@ -312,6 +313,7 @@ import {
   OverflowMenuItem,
   OptionsMenu,
   OptionsMenuToggle,
+  PageSection,
   Pagination,
   PaginationVariant,
   Text,
@@ -325,6 +327,7 @@ import CodeBranchIcon from '@patternfly/react-icons/dist/esm/icons/code-branch-i
 import CubeIcon from '@patternfly/react-icons/dist/esm/icons/cube-icon';
 import FilterIcon from '@patternfly/react-icons/dist/esm/icons/filter-icon';
 import SortAmountDownIcon from '@patternfly/react-icons/dist/esm/icons/sort-amount-down-icon';
+import DashboardWrapper from '@patternfly/react-core/src/demos/examples/DashboardWrapper';
 
 class ColumnManagementAction extends React.Component {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


mcarrano (changes requested):
@jenny-s51 This looks great, but there seems to be a bug with column reordering via drag & drop. If I reorder columns and then click Save, upon returning to the table, it's empty.

jenny-s51 (commented):
Great catch @mcarrano ! I've been thoroughly investigating here; this is not a trivial bug to fix, so I opened #7471 to address this issue as part of a necessary refactor. 

Our demo is using a deprecated API on `DataList` that implemented the drag/drop functionality. It would make more sense to refactor the code here in favor of the new `DragDrop` component, which will likely fix this bug and also keep us up to date with how we currently demo `DragDrop` in [DataList](https://github.com/patternfly/patternfly-react/blob/18817c800714e4d743cca5c8d4eca01b5da00616/packages/react-core/src/components/DataList/examples/DataListDraggable.tsx#L64)! WDYT @tlabaj ?

mmenestr:
Looks good, apart from what @mcarrano pointed out! Another thing, that maybe is outside the scope of this issue, is that I expected that while I'm dragging a row around, the other rows would move to "make room" for the dragged row, so that I would be able to have a visual of what i'm doing before I drop the row where I want it. Idk if that makes sense, but I just feel like it's hard to tell where the row is going to land while i'm dragging it..

jenny-s51:
Good point @mmenestr - the visual look of the current drag/drop implementation here is different from the drag/drop visuals that we see in the DataList draggable demo. This is also due to the newer `DragDrop` component that's being used there! 

The dragging visuals will be addressed separately in #7471 when this code is refactored to use `DragDrop`, which will align the appearance with the draggable Data list demo!

tlabaj:
> Great catch @mcarrano ! I've been thoroughly investigating here; this is not a trivial bug to fix, so I opened #7471 to address this issue as part of a necessary refactor.
> 
> Our demo is using a deprecated API on `DataList` that implemented the drag/drop functionality. It would make more sense to refactor the code here in favor of the new `DragDrop` component, which will likely fix this bug and also keep us up to date with how we currently demo `DragDrop` in [DataList](https://github.com/patternfly/patternfly-react/blob/18817c800714e4d743cca5c8d4eca01b5da00616/packages/react-core/src/components/DataList/examples/DataListDraggable.tsx#L64)! WDYT @tlabaj ?

@mcarrano @jenny-s51, I agree, we would want to suggest refactoring to use new drag and drop.

mmenestr:
kk sounds good! Just tried it out again, and the drag and drop is working as expected now!! But one thing I noticed is that when I go back into the manage columns modal after I've re-ordered something, the order that the columns show up in does not match the order they're in inside the table. Shouldn't the order of the columns in the modal, match the order of the columns in the table?

jenny-s51:
@mmenestr that will also be addressed as part of the refactor in #7471! The way the rows in the modal are currently ordered/implemented can be fixed/updated once we use `DragDrop` here instead 👍

packages/react-table/src/demos/Table.md
```diff
@@ -1661,24 +1673,30 @@ class ColumnManagementAction extends React.Component {
...
+          <PageSection padding={{ default: 'noPadding', xl: 'padding' }}>
+            <Card component="div">
+              <Table
+                gridBreakPoint="grid-xl"
+                header={
+                  <React.Fragment>
+                    <Toolbar id="page-layout-table-column-management-action-toolbar-top">{toolbarItems}</Toolbar>
+                  </React.Fragment>
+                }
+                aria-label="This is a table with checkboxes"
```
thatblindgeye, line 1686:
```suggestion
                aria-label="This is a table with draggable column management"
```

packages/react-table/src/demos/Table.md
```diff
@@ -898,24 +901,30 @@ class ColumnManagementAction extends React.Component {
...
```
thatblindgeye, line 914:
```suggestion
                aria-label="This is a table with column management"
```

thatblindgeye (changes requested):
Had a couple of nits below, but I'm also noticing an issue with both of the examples updated. The Pagination looks to be breaking outside of the main page content when I scroll to the very bottom of the page:

Pagination breaking out of page

versus how the bulk select demo looks:

Pagination in bulk select demo

Issue #6295, opened 2021-09-10
Hoverable/selectable datalist + table rows not VO accessible


nicolethoen:
How will we notify screen reader users that the item is selectable or selected. I've often questioned this in other components as well such as with the Selectable Datalist. With that one though, we had it in the name of the example so it at least had some indication that it was selectable (perhaps not an ideal solution though).

tlabaj:
@mcarrano can you please add this one to a milestone

wise-king-sullyman:
I'm addressing the table portion of this issue under this issue number and creating a separate issue for the datalist portion.

PR #7447, opened 2022-05-19
fix(Table): improve a11y for selectable table rows

wise-king-sullyman:
**What**: Addresses part of #6295 

**Additional issues**:

Convenience link: https://patternfly-react-pr-7447.surge.sh/components/table/react-composable/composable-row-click-handler-hoverable--selected-rows/

### DIFF
--- a/packages/react-table/src/components/TableComposable/TableComposable.tsx
+++ b/packages/react-table/src/components/TableComposable/TableComposable.tsx
@@ -59,6 +59,10 @@ export interface TableComposableProps extends React.HTMLProps<HTMLTableElement>,
   isExpandable?: boolean;
   /** Collection of column spans for nested headers. Deprecated: see https://github.com/patternfly/patternfly/issues/4584 */
   nestedHeaderColumnSpans?: number[];
+  /** Flag to apply a caption element that improves a11y for tables with selectable rows */
+  hasSelectableRowCaption?: boolean;
+  /** Visible text to add into the table selectable row a11y caption */
+  selectableRowCaptionText?: string;
 }
 
 interface TableComposableContextProps {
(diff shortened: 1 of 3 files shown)


wise-king-sullyman:
@thatblindgeye yes, this approach doesn't announce anything when navigating the table cell-by-cell as would intuitively be done when using a screen reader. As you said we could change that, but I feel like having each cell in a row announce if its row is selected or not would be a less than ideal user experience.

I think that the best fix for that issue would probably be for me to revise the demo to have the table announce rows are selectable as you suggested.

I'll update that so that we can see how well it ends up working. If anyone else has thoughts about this I'd love to hear them 🙂

thatblindgeye (commented):
With my previous comments kept in mind, I feel like something along the lines of the latest update might be one of the better ways to go about letting users know via screen reader that the table has rows that are selectable without getting overly verbose.

One issue with hardcoding the caption as it is is that passing in your own Caption component will visually render, but it doesn't get announced by screen reader. If instead of a caption element being applied when `hasSelectableRows` is true, a plain `div` with an ID is used to add to the `table` element's `aria-describedby` attribute, the same problem occurs where the `caption` element gets ignored by VO (since the table is being described by that `div` now).

After toying around, adding a `div class="pf-screen-reader">...</div>` *inside* the `caption` element allows the actual visible caption to be announced and the hidden instructions to be announced. This may not be any closer to a perfect solution, but in theory this could allow a consumer to pass in a custom visible caption to the table and still have screen reader instructions hidden visually (we could even potentially allow passing in custom instructions if needed).

Here's what the DOM looks like after I toyed around a bit:

hidden div inside a table caption element

wise-king-sullyman:
@thatblindgeye when attempting to add a div inside of the caption like you suggested I'm running into an issue where the div itself pushes the table down.

image

I think I've found a solution that works around this while still enabling consumers to add content in to the caption by only placing the assistive text inside a hidden div when additional caption content is passed and otherwise rendering the assistive text directly in the caption. I should have this approach working and pushed shortly.

wise-king-sullyman:
Also in the last commit I added a `hasSelectableRowCaption` prop to make this new behavior opt-in and prevent breaking changes. I've created #7473 to remove this prop in the next breaking change release to enable this behavior to be automatic in the future.

thatblindgeye (commented):
Things are sounding how I had them when editing things via dev tools, so I'm glad that was able to be done without that shifting.

With the new props added, I think an issue that could arise is if the Caption component is also passed in as the first child of the Table as it results in that double caption element issue from occurring. Would adding these props to the Caption component work do you think?

kmcfaul (approved):

mattnolting (approved):
Seems like a good enhancement to me. I'm not sure how we make the entire row selectable in VO?? But this is definitely an improvement. 👍

PR #7450, opened 2022-05-23
fix(CodeEditor): enable external updating of editor value on file upload

### LINKED ISSUE #7341: Bug - CodeEditor - Upload file doesn't trigger onChange

wise-king-sullyman:
**What**: Closes #7341

**Additional issues**:

I will create a follow-up issue shortly for figuring out how to effectively test components using Monaco and/or React Dropzone in accordance with our new testing guidelines.


wise-king-sullyman:
@batzionb I didn't change the behavior of the `onChange` prop itself to prevent breaking changes from occurring, and instead added a new `onCodeChange` prop which is called whenever `onChange` is called or when a file is uploaded.

Does this resolve the issue for you?

batzionb:
@wise-king-sullyman 
Yes, this is a good solution. Thanks!

tlabaj (approved):
LGTM

thatblindgeye (approved):

PR #7451, opened 2022-05-23
feat(table): full screen demo for sticky first column

### LINKED ISSUE #7359: Sticky first column

jenny-s51:
**What**: Closes #7359


mmenestr:
Is there any way you could add a couple more columns so that the scrolling from side to side can be seen in full screen mode, and not just when the screen is minimized? Otherwise looks great and behaves as expected!! :)

packages/react-table/src/demos/table-demos/StickyFirstColumn.jsx
```diff
@@ -0,0 +1,128 @@
...
+                  <Th modifier="truncate">{columnNames.header6}</Th>
+                  <Th modifier="truncate">{columnNames.header7}</Th>
+                  <Th modifier="truncate">{columnNames.header8}</Th>
+                  <Th modifier="truncate">{columnNames.header9}</Th>
+                </Tr>
+              </Thead>
+              <Tbody>
+                {sortedFacts.map(fact => (
+                  <Tr key={fact.name}>
+                    <Th isStickyColumn stickyMinWidth="100px" hasRightBorder modifier="truncate">
```
mattnolting, line 92: This `MinWidth` needs to match the `thead > tr > th` `MinWidth`. Currently, one is `120px` and the other is `100px`. It doesn't really matter which value you choose as long as they're the same. Core uses `100px`.

mattnolting (commented):
Looks great! Saw one thing.

wise-king-sullyman (approved):
Seems to work well and looks great from a React perspective!

thatblindgeye (approved):

packages/react-table/src/demos/table-demos/StickyFirstColumn.jsx
jenny-s51: Nice catch @mattnolting , thank you! I've updated both `MinWidth`s to `100px` to match core.

packages/react-table/src/demos/table-demos/StickyFirstColumn.jsx
```diff
@@ -0,0 +1,204 @@
...
+                  <Th modifier="truncate">{columnNames.header7}</Th>
+                  <Th modifier="truncate">{columnNames.header8}</Th>
+                  <Th modifier="truncate">{columnNames.header9}</Th>
+                  <Th modifier="truncate">{columnNames.header10}</Th>
+                  <Th modifier="truncate">{columnNames.header11}</Th>
+                  <Th modifier="truncate">{columnNames.header12}</Th>
+                  <Th modifier="truncate">{columnNames.header13}</Th>
+                  <Th modifier="truncate">{columnNames.header14}</Th>
+                  <Th modifier="truncate">{columnNames.header15}</Th>
+                  <Th modifier="truncate">{columnNames.header16}</Th>
```
mattnolting, line 141: `thead` cells are set to truncate by default, so you don't need `modifier="truncate"`. I'm not sure why the core demo has that, but it needs to be updated. I've opened [4882](https://github.com/patternfly/patternfly/issues/4882) to do that.  I'll set the first two `th` cells to `pf-m-fit-content` and remove the subsequent `pf-m-truncate` mods, so you can match that pattern.

mattnolting (commented):
Great work! Looks like I need to make some Core updates for you...

jenny-s51 (commented):
Thank you @mattnolting - the table is looking awesome! 

I set `fitContent` as the modifier for the first couple `Th` cells and now the content is no longer being truncated. Should address your comment as well @mmenestr ! 

Initially I made sure to align this demo with the current core demo -- updating the core demo 
in https://github.com/patternfly/patternfly/issues/4882 to align with your comments here sounds like a good plan 👍

mmenestr (approved):

mattnolting (approved):
Great work! LPTM 🎉

PR #7456, opened 2022-05-24
fix(Select): aria-describedby is added properly

### LINKED ISSUE #7251: Bug - Select Typeahead - aria-describedby is not present
**Describe the problem**
The Select component states that `aria-describedby` can be added to the Select component. When I add this to the Select component that is a typeahead variant, I expect to see `aria-describedby` added to the input `.pf-c-select__toggle-typeahead` so that any error messages we display are announced to users of assistive technology.

Add `aria-describedby` to the `<Select>` component.

**Expected behavior**
I expect to see `aria-describedby` added to the input `.pf-c-select__toggle-typeahead`

No

tompsota:
The aria-describedby field is now added regardless of state of validation

**What**: Closes #7251

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -103,6 +103,8 @@ export interface SelectProps
   'aria-invalid'?: boolean;
   /** Label for input field of type ahead select variants */
   typeAheadAriaLabel?: string;
+  /** Id of div for input field of type ahead select variants */
+  typeAheadAriaDescribedby?: string;
   /** Label for clear selection button of type ahead select variants */
   clearSelectionsAriaLabel?: string;
   /** Label for toggle of type ahead select variants */
@@ -209,6 +211,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
     'aria-describedby': '',
     'aria-invalid': false,
     typeAheadAriaLabel: '',
+    typeAheadAriaDescribedby: '',
     clearSelectionsAriaLabel: 'Clear all',
     toggleAriaLabel: 'Options menu',
     removeSelectionAriaLabel: 'Remove',
@@ -929,6 +932,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       validated,
       selections: selectionsProp,
       typeAheadAriaLabel,
+      typeAheadAriaDescribedby,
       clearSelectionsAriaLabel,
       toggleAriaLabel,
       removeSelectionAriaLabel,
@@ -1229,7 +1233,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
           className
         )}
         {...(width && { style: { width } })}
-        {...(validated !== ValidatedOptions.default && { 'aria-describedby': ariaDescribedby })}
+        {...(ariaDescribedby && { 'aria-describedby': ariaDescribedby })}
         {...(validated !== ValidatedOptions.default && { 'aria-invalid': ariaInvalid })}
       >
         {isOpen && menuContainer}
@@ -1250,7 +1254,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
         ref={this.parentRef}
         {...getOUIAProps(Select.displayName, ouiaId !== undefined ? ouiaId : this.state.ouiaStateId, ouiaSafe)}
         {...(width && { style: { width } })}
-        {...(validated !== ValidatedOptions.default && { 'aria-describedby': ariaDescribedby })}
+        {...(ariaDescribedby && { 'aria-describedby': ariaDescribedby })}
         {...(validated !== ValidatedOptions.default && { 'aria-invalid': ariaInvalid })}
       >
         <SelectToggle
@@ -1319,6 +1323,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   aria-activedescendant={typeaheadActiveChild && typeaheadActiveChild.id}
                   id={`${selectToggleId}-select-typeahead`}
                   aria-label={typeAheadAriaLabel}
+                  {...(typeAheadAriaDescribedby && { 'aria-describedby': typeAheadAriaDescribedby })}
                   placeholder={placeholderText as string}
                   value={
                     typeaheadInputValue !== null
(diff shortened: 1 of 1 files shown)


packages/react-core/src/components/Select/Select.tsx
```diff
@@ -1250,7 +1250,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
         ref={this.parentRef}
         {...getOUIAProps(Select.displayName, ouiaId !== undefined ? ouiaId : this.state.ouiaStateId, ouiaSafe)}
         {...(width && { style: { width } })}
-        {...(validated !== ValidatedOptions.default && { 'aria-describedby': ariaDescribedby })}
+        {...(ariaDescribedby !== '' && { 'aria-describedby': ariaDescribedby })}
```
jenny-s51, line 1253:
```suggestion
        {...(ariaDescribedby && { 'aria-describedby': ariaDescribedby })}
```

jenny-s51 (changes requested):
Nice update @tompsota - I'm seeing `aria-describedby`when I enter a value for it!

Based on @jgiardino 's description in https://github.com/patternfly/patternfly-react/issues/7251, it looks like we want `aria-describedby` added to the input .pf-c-select__toggle-typeahead, whereas currently it's showing up on `pf-c-select`.

Easy fix - you should be able to relocate `{...(ariaDescribedby && { 'aria-describedby': ariaDescribedby })}` to the [typeahead](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Select/Select.tsx#L1317-L1334) and [typeahead multi](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Select/Select.tsx#L1344-L1357) `<input>`s.

What do you think @tompsota ?

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -1229,7 +1229,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
           className
         )}
         {...(width && { style: { width } })}
-        {...(validated !== ValidatedOptions.default && { 'aria-describedby': ariaDescribedby })}
+        {...(ariaDescribedby !== '' && { 'aria-describedby': ariaDescribedby })}
```
jenny-s51, line 1232:
```suggestion
        {...(ariaDescribedby && { 'aria-describedby': ariaDescribedby })}
```

packages/react-core/src/components/Select/Select.tsx
tompsota: Thanks for the comments. 

If we relocate `ariaDescribedby` to `<input>`s, we will be unable to use it with different variants of `Select` as some of them do not have inputs. 

I see 2 possibilities how to avoid this:

1. Add `aria-describedby` to `<input>`s and keep it also where it is now
2. Add new prop `typeAheadAriaDescribedBy`, apply it to `<input>`s and keep `aria-describedby` where it is now. Similarly to [typeAheadAriaLabel](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Select/Select.tsx#L1321)

I would prefer the second option, what about you @jenny-s51?

jenny-s51: Thank you for clarifying @tompsota - I like the second option as well! It would make sense to follow this existing convention the way it's used for typeAheadAriaLabel

jenny-s51 (commented):
LGTM! Looks like we just need to rebuild/update the snapshots for `Select.test.tsx`

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -1250,7 +1254,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
         ref={this.parentRef}
         {...getOUIAProps(Select.displayName, ouiaId !== undefined ? ouiaId : this.state.ouiaStateId, ouiaSafe)}
         {...(width && { style: { width } })}
-        {...(validated !== ValidatedOptions.default && { 'aria-describedby': ariaDescribedby })}
+        {...(ariaDescribedby && { 'aria-describedby': ariaDescribedby })}
```
thatblindgeye, line 1257: I believe if this is moved to the `<SelectToggle>` on line 1260, the attribute would apply to the correct element.

thatblindgeye (commented):
Sounds good when testing in VO! I am noticing an issue with the original `aria-describedby` prop with the non-typeahead variant in that VO doesn't announce the description, however. It looks like it's because the `aria-describedby` is being applied to the div with class `.pf-c-select` rather than the SelectToggle like the `aria-labelledby` is.

EDIT: Since this is pre-existing I'm going to open a follow up issue, so no need to make this update on this PR.

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -1319,6 +1323,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   aria-activedescendant={typeaheadActiveChild && typeaheadActiveChild.id}
                   id={`${selectToggleId}-select-typeahead`}
                   aria-label={typeAheadAriaLabel}
+                  aria-describedby={typeAheadAriaDescribedby}
```
thatblindgeye, line 1326: It might be worth following the `aria-describedby` prop here so that the attribute doesn't get added if the prop isn't passed in:

```suggestion
                  {...(typeAheadAriaDescribedby && { 'aria-describedby': typeAheadAriaDescribedby })}
```

packages/react-core/src/components/Select/Select.tsx
```diff
@@ -1347,6 +1352,7 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
                   id={`${selectToggleId}-select-multi-typeahead-typeahead`}
                   aria-label={typeAheadAriaLabel}
                   aria-invalid={validated === ValidatedOptions.error}
+                  aria-describedby={typeAheadAriaDescribedby}
```
thatblindgeye, line 1355: Same here in regards to comment above.

thatblindgeye (changes requested):

thatblindgeye (approved):

jessiehuff (approved):
LGTM!

jenny-s51 (approved):
🚀

PR #7458, opened 2022-05-24
chore(DualListSelector): convert examples to TypeScript

### LINKED ISSUE #7396: Dual list selector - convert examples to TypeScript
Dual List Selector

**Describe the feature**
The examples for the Dual List Selector component need to be converted to TypeScript and functional components.

N/A

**Any other information?**
N/A

thatblindgeye:
**What**: Closes #7396 

[Dual list selector preview build](https://patternfly-react-pr-7458.surge.sh/components/dual-list-selector)

### Updated comment

Per my original comment below, I utilized `JSON.stringy'`s replacer argument to resolve the error that was occurring (also noted below). An alternative could be to utilize equality checking similar to [how to compare objects in JavaScript](https://dmitripavlutin.com/how-to-compare-objects-in-javascript/#4-deep-equality), which did work when I tried it out, but because you can pass in strings, objects, or HTML elements (based on the examples we provide), I wasn't sure whether we would ever need to further check whether an option is a particular type and do different checking based on that. Using this replacer argument seems like a simpler fix and it works with both the updated functional example and the  class example.

I also made a change to the replacer method in that it checks for any key that starts with an underscore (rather than just the specific `_owner` key) just to be safe.

### Original comment below

**A note on this PR**

When running this code locally, the "Using more complex options with actions" causes the below error to occur whenever you try interacting with the dual list selector in the example (clicking an option, typing in the search bar, etc):

dual list example error

This only occurs when the `availableOptions` state in the example is set to an array of HTML elements like `<span>Option 1</span>...` when the example is converted to a functional component. An array of strings works fine, and the array of HTML elements also works fine on the current example on the PF site that uses a class. From trying to look into it, the below snippet of code on lines 195-196 within DualListSelector is causing it:

```
if (
      JSON.stringify(this.props.availableOptions) !== JSON.stringify(this.state.availableOptions) ||
      JSON.stringify(this.props.chosenOptions) !== JSON.stringify(this.state.chosenOptions)
    )
```
Utilizing JSON.stringify's replacer argument does seem to avoid this error with a function similar to below, but I haven't yet delved into figuring out why that resolves it/why the error occurs in the first place:

```
  replacer = (key: any, value: any) => {
    if (key === '_owner') {
      return undefined;
    }
    return value;
  };

// replacing the previous code snippet with the below...
if {
  JSON.stringify(this.props.availableOptions, this.replacer) !== JSON.stringify(this.state.availableOptions, this.replacer) ||
  JSON.stringify(this.props.chosenOptions, this.replacer) !== JSON.stringify(this.state.chosenOptions, this.replacer)
}
```

Because the example seems to intend to show how to use more complex options than just strings, it wouldn't seem like simply changing the options to strings would suffice here.

**Additional issues**:


nicolethoen:
Ah, so stringify was being used here to do a deeper comparison between the complex objects. @kmcfaul I think you were the one who landed on this solution here?
I think you could replace that bit of logic that does something else to determine if there's been a change to the available or chosen options when they are possibly complex objects?

packages/react-core/src/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -23,806 +24,86 @@ import PficonSortCommonAscIcon from '@patternfly/react-icons/dist/esm/icons/pfic
...
 import React from 'react';
-import { DualListSelector, DualListSelectorPane, DualListSelectorList, DualListSelectorListItem, DualListSelectorControlsWrapper, DualListSelectorControl } from '@patternfly/react-core';
+import {
+  DualListSelector,
+  DualListSelectorPane,
+  DualListSelectorList,
+  DualListSelectorListItem,
+  DualListSelectorControlsWrapper,
+  DualListSelectorControl
+} from '@patternfly/react-core';
```
nicolethoen, line 63: Could you just remove these imports from the `noLive` region? They are clutter and not helping to demonstrate structure as was the intent of this documentation.

nicolethoen (commented):
One nit - looks great to me!

nicolethoen (approved):

PR #7464, opened 2022-05-25
docs(ExpandableSection): add example with custom component

### LINKED ISSUE #6967: it would be nice if ExpandableSection toggleText could be an arbitrary component
For example, what if i want to place an icon in the toggler? I realize that we could take control of this by having detached toggler, but... that really isn't our use case. We just want to put an icon in the toggle area.

e.g. from [pymdown Details](https://facelessuser.github.io/pymdown-extensions/extensions/details/)

Screen Shot 2022-02-02 at 3 36 46 PM

tompsota:
Added a new example demonstrating that any component can be used as toggle text.

**What**: Closes #6967


packages/react-core/src/components/ExpandableSection/examples/ExpandableSection.md
```diff
@@ -178,3 +180,48 @@ class SimpleExpandableSection extends React.Component {
   }
 }
 ```
+
+### With custom component
+
+Any `React.ReactNode` object can be used in place of toggle text.
```
tlabaj, line 186: I don't think we need this text here.

packages/react-core/src/components/ExpandableSection/examples/ExpandableSection.md
```diff
@@ -178,3 +180,48 @@ class SimpleExpandableSection extends React.Component {
   }
 }
 ```
+
+### With custom component
```
tlabaj, line 184: I would rename this to something like "With icon in toggle".  What do you think for a name here @mcarrano. 
I think we want to get across the the consumer that the toggle can be something other than a string.

tlabaj (changes requested):

packages/react-core/src/components/ExpandableSection/examples/ExpandableSection.md
mcarrano: Maybe say "With additional non-text elements?" We shouldn't imply that this is limited to an icon. I'd also recommend that we note that this is limited to non-interactive elements (you can't put a button here, for example).

mcarrano (approved):
Looks good to me. Thanks @tompsota !

thatblindgeye (approved):

tlabaj (approved):
Looks great!

Issue #7469, opened 2022-05-26
React tokens failing on dark theme vars


mcoker:
Doesn't look like react tokens is going to work with dark theme, and it's currently failing trying to resolve some dark theme specific vars. For the time being, we can just ignore the dark theme blocks.

PR #7470, opened 2022-05-26
chore(react-tokens): exclude dark theme blocks

mcoker:
fixes https://github.com/patternfly/patternfly-react/issues/7469


jschuler:
Here's a diff between the current react-tokens and after the change FYI
https://diffy.org/diff/7263e28925bd5

mcoker:
@evwilkin @jschuler the problem is that react tokens just does a regex match for CSS vars in a global CSS var file - https://github.com/patternfly/patternfly-react/blob/35dea6aaa7ed4c6f49ae0d3ee6644dda3efb2889/packages/react-tokens/scripts/generateTokens.js#L103-L107

That file has a handful of selectors in it that redefine global variables (https://unpkg.com/browse/@patternfly/patternfly@4.196.3/base/patternfly-variables.css), so the vars in those selectors will be returned multiple times. The dark theme code was in there, and I'm assuming since those vars are required to be defined last, their value was getting used as the value in the token.

I'm sure there's a better way to do this, but I'm using `css` to `parse` that stylesheet instead, excluding blocks that match `/\.pf-theme-dark/` and `stringify` to put it back in stylesheet form before running the existing regex match - https://github.com/patternfly/patternfly-react/blob/bfd2f40f8a72bc1777882d9fb2904aaab839a7b4/packages/react-tokens/scripts/generateTokens.js#L108-L116

jschuler:
Hi @mcoker , thanks for the fix! The diff looks much smaller now
https://diffy.org/diff/243e6fc7161a1
Assuming these changes are intended, so LGTM!

jschuler (approved):

tlabaj (approved):
LGTM

PR #7474, opened 2022-05-27
fix(DataList): improve a11y for selectable data list items

### LINKED ISSUE #7472: Hoverable/selectable datalist not VO accessible
This issue is addressing the datalist portion of #6295.

wise-king-sullyman:
**What**: Closes #7472

**Additional issues**:

Convenience link: https://patternfly-react-pr-7474.surge.sh/components/data-list/react/selectable-rows/

### DIFF
--- a/packages/react-core/src/components/DataList/DataList.tsx
+++ b/packages/react-core/src/components/DataList/DataList.tsx
@@ -47,6 +47,8 @@ export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, '
   wrapModifier?: DataListWrapModifier | 'nowrap' | 'truncate' | 'breakWord';
   /** @deprecated Order of items in a draggable DataList */
   itemOrder?: string[];
+  /** Flag indicating that the data list should render a hidden input to make it selectable */
+  hasSelectableInput?: boolean;
 }
 
 interface DataListState {
@@ -60,6 +62,7 @@ interface DataListContextProps {
   isSelectable: boolean;
   selectedDataListItemId: string;
   updateSelectedDataListItem: (id: string) => void;
+  hasSelectableInput?: boolean;
   isDraggable: boolean;
   dragStart: (e: React.DragEvent) => void;
   dragEnd: (e: React.DragEvent) => void;
@@ -308,6 +311,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
       onDragFinish,
       gridBreakpoint,
       itemOrder,
+      hasSelectableInput,
       /* eslint-enable @typescript-eslint/no-unused-vars */
       ...props
     } = this.props;
@@ -330,6 +334,7 @@ export class DataList extends React.Component<DataListProps, DataListState> {
           isSelectable,
           selectedDataListItemId,
           updateSelectedDataListItem,
+          hasSelectableInput,
           isDraggable: this.html5DragDrop,
           dragStart: this.dragStart,
           dragEnd: this.dragEnd,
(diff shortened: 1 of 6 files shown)


thatblindgeye (commented):
Things sounds good going through the selectable example! One thing that we may need to align on is whether clicking a row itself should select *and* unselect that row, as right now you can select and unselect a row via the hidden input, but clicking a row itself multiple times doesn't unselect it. This could be made into a separate issue, though. 

Only other thing would be do you think it'd be worth exposing a prop to pass in a custom function for when the hidden input changes? Similar to how was implemented on the [Card PR](https://github.com/patternfly/patternfly-react/pull/7144/files)

wise-king-sullyman:
@thatblindgeye I agree completely that exposing the onChange for the input as I did in Card is the way to go, great callout.

As far as the behavior that we demo, I implemented toggle behavior since the input is a checkbox. But after thinking through it I believe that unlike on Card we can actually just open up the choice between a checkbox and radio to the consumer here since the `DataList` component can set the type for all selectable items.

If it sounds good to you I'll update this to allow the consumer to pass in the onChange and input type, and change the selectable rows demo to use a radio input that won't toggle shortly.

tlabaj (changes requested):

packages/react-core/src/components/DataList/DataList.tsx
```diff
@@ -47,6 +47,11 @@ export interface DataListProps extends Omit<React.HTMLProps<HTMLUListElement>, '
   wrapModifier?: DataListWrapModifier | 'nowrap' | 'truncate' | 'breakWord';
   /** @deprecated Order of items in a draggable DataList */
   itemOrder?: string[];
+  /** @beta Object that causes the data list to render hidden inputs which improve selectable item a11y */
+  selectableInput?: {
```
tlabaj, line 51: I like this API better!

tlabaj (approved):
LGTM!

packages/react-core/src/components/DataList/DataList.tsx
wise-king-sullyman: As do I, it was a great suggestion!

mcoker (approved):
LGTM! Just one comment about the prop name and values to try and avoid breaking changes going forward. Not sure if it's a good idea or not here.

packages/react-core/src/components/DataList/DataList.tsx
tlabaj: I like Coker's suggestion.  Also, can we create an interface for the object so we can document each property in the object like we do in [slider](https://github.com/patternfly/patternfly-react/blob/c2faf7db0e1006d1660d572d7f507053685cb9b8/packages/react-core/src/components/Slider/Slider.tsx#L10).  `selectabelInput` would thenhave the type of whatever you call the interface. You can add the interface to the `propComponents` in the md file so it is documented on the website.

tlabaj (changes requested):

mattnolting (commented):
This looks really great @wise-king-sullyman! As I was testing, in noticed that the invisible checkbox is highlight in VO (not tabbing). So I did a little tinkering and found that, instead of using `pf-screen-reader`, if we wrote the css a little differently, we could have the entire row selected on VO checkbox focus. 

Currently, when navigating visual focus appears in top left or most recent, relatively positioned ancestor. 

Screen Recording 2022-06-01 at 11 23 07 AM

With a css, the row is highlighted on checkbox focus.

Screen Recording 2022-06-01 at 11 28 22 AM

Here's the css update:

Screen Shot 2022-06-01 at 10 56 37 AM

I tested in FF, Chrome, and Safari. They all seem to work as expected.

tlabaj (approved):

jessiehuff (approved):
Awesome!! :)

PR #7477, opened 2022-05-27
fix(Table): filterable table demo icon text spacing fix

### LINKED ISSUE #6262: Filterable table demo updates
Spotted in this demo - https://www.patternfly.org/v4/components/table/react-demos#filterable. Seems like we should exclude the checkbox in the body if there is an empty state result.

Screen Shot 2021-09-02 at 1 22 40 PM

Also the dropdown toggle text and left icon are a little close. We could pass the toggle icon as an `icon` prop to the toggle instead. It would look like this:

Screen Shot 2021-09-02 at 1 33 30 PM

andyyvo:
**What**: Fixed bug where icon and text for Filterable Table Demos was too close. Passed icon in as `icon` prop as suggested. Closes #6262

**Additional issues**: N/A

### DIFF
--- a/packages/react-table/src/demos/Table.md
+++ b/packages/react-table/src/demos/Table.md
@@ -1878,8 +1878,8 @@ class FilterTableDemo extends React.Component {
           onSelect={this.onCategorySelect}
           position={DropdownPosition.left}
           toggle={
-            <DropdownToggle onToggle={this.onCategoryToggle} style={{ width: '100%' }}>
-              <FilterIcon /> {currentCategory}
+            <DropdownToggle onToggle={this.onCategoryToggle} style={{ width: '100%' }} icon={<FilterIcon /> }>
+              {currentCategory}
             </DropdownToggle>
           }
           isOpen={isCategoryDropdownOpen}
--- a/packages/react-table/src/demos/Table.md
+++ b/packages/react-table/src/demos/Table.md
@@ -1872,24 +1872,25 @@ class FilterTableDemo extends React.Component {
   buildCategoryDropdown() {
     const { isCategoryDropdownOpen, currentCategory } = this.state;
 
+    const categoryMenuItems = [
+      <SelectOption key="cat1" value="Location" />,
+      <SelectOption key="cat2" value="Name" />,
+      <SelectOption key="cat3" value="Status" />,
+    ];
+
     return (
       <ToolbarItem>
-        <Dropdown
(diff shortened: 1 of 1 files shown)


jenny-s51 (changes requested):
Nice work @andyyvo ! Icon is definitely showing up where we want it to now and the spacing looks great.

However when I shrink the screen to a smaller viewport I'm seeing some interesting behavior with the alignment of the dropdown icon, text, and toggle. 
Screen Shot 2022-05-31 at 1 43 10 PM

Looks like this is because we're using a `Dropdown` above a `Select` here, which both seem to have different styling at different viewport sizes.

Based on the HTML structure used in the table core demos, it looks like we want to use a `Select` here instead of a `Dropdown`. Swapping out the `Dropdown` for a `Select` would solve both the alignment issue and the icon spacing issue. WDYT @mcoker ?

mcoker (approved):
Nailed it!

mattnolting:
> Based on the HTML structure used in the table core demos, it looks like we want to use a Select here instead of a Dropdown. Swapping out the Dropdown for a Select would solve both the alignment issue and the icon spacing issue. WDYT @mcoker ?

Good work @andyyvo!

@jenny-s51 That's correct, the dropdown should be a select. What's happening here is that there are two versions of the toolbar, mobile and desktop. The dropdown referenced gets moved to the expandable content section on mobile devices, where it should be a select. However, this does expose a `.pf-c-dropdown__toggle` bug. If a `.pf-c-dropdown` has a set width, `justify-content: space-between` spaces the icon, text and toggle unexpectedly, resulting in the bug your noted.

Screen Shot 2022-06-07 at 10 27 49 AM
 
[Opened an issue to address](https://github.com/patternfly/patternfly/issues/4894).

andyyvo:
Switched from `Dropdown` to `Select` for the Filterable Table demo and I've tested it on both Chrome devtools and Safari iphone view and got the same result. I have screenshots of the changes just to verify that this is what we're looking for @jenny-s51 @mattnolting

Screen Shot 2022-06-15 at 4 45 38 PM

Screen Shot 2022-06-15 at 4 45 35 PM

mattnolting (approved):
Well done 👏 LGTM

jenny-s51:
Looks 💯

jenny-s51 (approved):

nicolethoen (approved):

PR #7489, opened 2022-06-01
chore(FormSelect): convert examples to TypeScript

### LINKED ISSUE #7475: Form select: convert examples to TypeScript

tompsota:
chore(FormSelect): convert examples to TypeScript

**What**: Closes #7475


packages/react-core/src/components/FormSelect/examples/FormSelectDisabled.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { FormSelect, FormSelectOption } from '@patternfly/react-core';
+
+export const FormSelectBasic: React.FunctionComponent = () => {
```
jenny-s51, line 4:
```suggestion
export const FormSelectDisabled: React.FunctionComponent = () => {
```

packages/react-core/src/components/FormSelect/examples/FormSelectGrouped.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { FormSelect, FormSelectOption, FormSelectOptionGroup } from '@patternfly/react-core';
+
+export const FormSelectGrouped: React.FunctionComponent = () => {
+  const [formSelectValue, setFormSelectValue] = React.useState<string>('2');
```
jenny-s51, line 5:
```suggestion
  const [formSelectValue, setFormSelectValue] = React.useState('2');
```

packages/react-core/src/components/FormSelect/examples/FormSelectIconSpriteVariant.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { Form, FormGroup, FormSelect, FormSelectOption, ValidatedOptions } from '@patternfly/react-core';
+
+export const FormSelectIconSpriteInput: React.FunctionComponent = () => {
+  const [formSelectValue, setFormSelectValue] = React.useState<string>('');
+  const [validated, setValidated] = React.useState<ValidatedOptions>(ValidatedOptions.default);
+  const [invalidText, setInvalidText] = React.useState<string>('You must choose something');
```
jenny-s51, line 7:
```suggestion
  const [invalidText, setInvalidText] = React.useState('You must choose something');
```

packages/react-core/src/components/FormSelect/examples/FormSelectValidated.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { Form, FormGroup, FormSelect, FormSelectOption, ValidatedOptions } from '@patternfly/react-core';
+
+export const FormSelectValidated: React.FunctionComponent = () => {
+  const [formValue, setFormValue] = React.useState<string>('');
+  const [invalidText, setInvalidText] = React.useState<string>('You must choose something');
```
jenny-s51, line 6:
```suggestion
  const [invalidText, setInvalidText] = React.useState('You must choose something');
```

packages/react-core/src/components/FormSelect/examples/FormSelectValidated.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const [helperText, setHelperText] = React.useState<string>('');
```
jenny-s51, line 7:
```suggestion
  const [helperText, setHelperText] = React.useState('');
```

jenny-s51 (changes requested):
Awesome work on this @tompsota ! The examples are looking great. Left some comments below.

In a lot of cases, TS can infer the type of your hooks itself e.g. when you set the value to a string, so you don't need the extra `<string>` type declaration for those!

Just a few easy fixes and otherwise LGTM 👍

jenny-s51 (approved):
Looks 👍👍!

packages/react-core/src/components/FormSelect/examples/FormSelectDisabled.tsx
```diff
@@ -0,0 +1,28 @@
...
+    { value: 'mr', label: 'Mr', disabled: false },
+    { value: 'miss', label: 'Miss', disabled: false },
+    { value: 'mrs', label: 'Mrs', disabled: false },
+    { value: 'ms', label: 'Ms', disabled: false },
+    { value: 'dr', label: 'Dr', disabled: false },
+    { value: 'other', label: 'Other', disabled: false }
+  ];
+
+  return (
+    <FormSelect value={formSelectValue} onChange={onChange} aria-label="FormSelect Input">
```
thatblindgeye, line 22: The FormSelect is missing the `isDisabled` prop

packages/react-core/src/components/FormSelect/examples/FormSelectInvalid.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const FormSelectInvalid: React.FunctionComponent = () => {
+  const [formSelectValue, setFormSelectValue] = React.useState('');
+  const [validated, setValidated] = React.useState<ValidatedOptions>(ValidatedOptions.default);
+
+  const isEmpty = () => formSelectValue !== '';
+  const onChange = (value: string) => {
+    const validated = isEmpty() && ValidatedOptions.error;
+    setFormSelectValue(value);
+    setValidated(validated);
+  };
```
thatblindgeye, line 13: It doesn't look like this part of the example is working how it may have been intended when originally added. 

Rather than trying to set both the validated and value states in the onChange function, it might work better to have the onChange function only call `setFormSelectValue` like other examples have, and moving the `setValidated` and checking if `formSelectValue` is an empty string to a useEffect hook that only runs when the `formSelectValue` updates. The `isEmpty` function would just also have to be updated to check that `formSelectValue === ''`

thatblindgeye (changes requested):
Great job on the conversions! I also had a couple change requests below, let me know if you have any questions! 🙂

packages/react-core/src/components/FormSelect/examples/FormSelectInvalid.tsx
tompsota: Yes, this works better. Thanks for pointing it out.

I also renamed `const validated` variable so it doesn't collide with state variable.

packages/react-core/src/components/FormSelect/examples/FormSelectInvalid.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [formSelectValue, setFormSelectValue] = React.useState('');
+  const [validated, setValidated] = React.useState<ValidatedOptions>(ValidatedOptions.default);
+
+  const isEmpty = () => formSelectValue === '';
+  const onChange = (value: string) => {
+    setFormSelectValue(value);
+  };
+
+  React.useEffect(() => {
+    const validationState = !isEmpty() && ValidatedOptions.error;
```
thatblindgeye, line 14: This line just needs to be updated so that it sets the validated state correctly; right now it's having the opposite effect of a valid selection rendering the error icon in the example.

```suggestion
    const validationState = isEmpty() && ValidatedOptions.error;
```

thatblindgeye (changes requested):

packages/react-core/src/components/FormSelect/examples/FormSelectInvalid.tsx
tompsota: Okay so iiuc, only `Select a number` option should be invalid? If this is the case, won't it be better to use e.g. 3 as the only valid option and add to corresponding `FormSelectOption`s label that this is the only valid option? Similarly as in `Validated` example.

thatblindgeye: > Okay so iiuc, only Select a number option should be invalid?

That's what it seemed like the intention originally was, at least based on the variable being named `isEmpty`.

> If this is the case, won't it be better to use e.g. 3 as the only valid option and add to corresponding FormSelectOptions label that this is the only valid option? Similarly as in Validated example.

I think either way could work really. While both Validated and Invalid examples are very similar, the main difference just seems to be that the Validate example shows various states of validation instead of only an error like the Invalid example does. As long as that remains true, reusing the `onChange` from the Validate example and tweaking it to not use helper text or simulate a network call could work (this would also make the useEffect unnecessary)

tompsota: Okay, so I kept the original example, it does what we need it to do so it's probably not necessary to make it more complicated.

tlabaj (commented):
@mcarrano I think we can actually remove the "invalid" example.  It is from before we hd the different "validated" states.  I don't remember if we kept it for a reason.

tlabaj (approved):
LGTM!

PR #7503, opened 2022-06-03
fix(Calendar month): update to match Core

### LINKED ISSUE #7459: Bug - Calendar month - update the month prev/next icons to match core
**Describe the problem**
At some point the icons used to change the month on the calendar got changed from the original design and core implementation. They should be using the left and right carets vs the arrows. See discussion here: https://github.com/patternfly/patternfly/pull/4862

thatblindgeye:
**What**: Closes #7459 

[Calendar month preview build](https://patternfly-react-pr-7503.surge.sh/components/calendar-month)

@mcarrano It looks like there's another discrepancy between Core and React where the Core version shows 5 weeks and React shows 6. A comment in the React code mentions, "We will always show 6 weeks like google calendar," though it looks like Google calendar shows 5 weeks at a time (unsure if this is the default). Do you think this should also be updated in React to match Core's 5 week view?

**Additional issues**:
N/A

### DIFF
--- a/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
+++ b/packages/react-core/src/components/CalendarMonth/CalendarMonth.tsx
@@ -2,8 +2,9 @@ import React, { useEffect } from 'react';
 import { TextInput } from '../TextInput/TextInput';
 import { Button } from '../Button/Button';
 import { Select, SelectOption } from '../Select';
-import ArrowLeftIcon from '@patternfly/react-icons/dist/esm/icons/arrow-left-icon';
-import ArrowRightIcon from '@patternfly/react-icons/dist/esm/icons/arrow-right-icon';
+import { InputGroup } from '../InputGroup';
+import AngleLeftIcon from '@patternfly/react-icons/dist/esm/icons/angle-left-icon';
+import AngleRightIcon from '@patternfly/react-icons/dist/esm/icons/angle-right-icon';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/CalendarMonth/calendar-month';
 import { getUniqueId } from '../../helpers/util';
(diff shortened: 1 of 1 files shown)


mmenestr:
From[ these examples](https://material.io/components/date-pickers#anatomy), it does seem like google uses 6 weeks (idk if this is really the right place to look though) but that's because the specific month they're using in their examples requires the "6 lines". I also notice that they don't show the dates that aren't part of the month, so the examples feel visually lighter. 

Meanwhile, in our "6 line" example, we not only show the dates that aren't part of the month, but we also have a full line of numbers that pertain to the next month - which doesn't really seem useful to me? It feels like it just creates clutter... all that to say, I wonder if the 5 vs. 6 should just depend on the month - and whether we should make sure that you never have a full line of numbers that don't even pertain to the month selected!

Screen Shot 2022-06-03 at 8 20 22 PM

nicolethoen (approved):

mmenestr:
No, I personally don't mind having the extra dates in a lighter grey since it doesn't really change the space being used, and kind of helps with situating yourself within months. I also don't think it should be opt-in/opt-out from a consistency standpoint - we probably want all calendar months to look the same and to not have unnecessary variations!

What I do feel strongly about is that I don't think there should be a line of **only** dates that don't belong in that month! In the image above the 6 to 12 of December are shown even though the month selected is November, so that does feel like unnecessary space being added/used. Does that make sense?

thatblindgeye:
@mmenestr The calendar month now only shows weeks that include at least 1 date within the current month, up to a maximum of 6 weeks. You can see this in the latest [calendar month preview build](https://patternfly-react-pr-7503.surge.sh/components/calendar-month/).

mmenestr (approved):
Looks great @thatblindgeye thank you so much!!

PR #7513, opened 2022-06-04
fix(Table): update expandable examples to set column widths

### LINKED ISSUE #7460: Table - add/update expandable examples/demos to set column widths
re: https://github.com/patternfly/patternfly/pull/4852

The column widths in expandable tables may shift a bit as expandable rows are toggled open/closed, since the expandable rows are part of the table and the table columns adapt to the content in the table cells. To keep the column widths static when toggling, we've added an example that shows adding set width variations to the expandable table header cells.

For this update in react, I guess it's a question of whether we want to update all of the expandable examples (or at least the ones whose content shifts), or add a new example/demo with the column widths set? @mcarrano wdyt?

bartoval:
**What**: Closes #7460

### DIFF
--- a/packages/react-table/src/components/TableComposable/Tr.tsx
+++ b/packages/react-table/src/components/TableComposable/Tr.tsx
@@ -13,7 +13,9 @@ export interface TrProps extends React.HTMLProps<HTMLTableRowElement>, OUIAProps
   innerRef?: React.Ref<any>;
   /** Flag indicating the Tr is hidden */
   isHidden?: boolean;
-  /** Only applicable to Tr within the Tbody: Makes the row expandable and determines if it's expanded or not */
+  /** Only applicable to Tr within the Tbody: Makes the row expandable and determines if it's expanded or not.
+   * To prevent column widths from responding automatically when expandable rows are toggled, the width prop must also be passed into either the th or td component
+   */
   isExpanded?: boolean;
   /** Only applicable to Tr within the Tbody: Whether the row is editable */
   isEditable?: boolean;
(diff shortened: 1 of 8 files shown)


thatblindgeye (changes requested):
Awesome job, the examples look great! I think it might be worth also updating the ComposableTableCompoundExpandable example, as testing that one out with the "lorem ipsum..." text has the same column width changing behavior without the set widths.

Another thing might be including the verbiage that was included on the Core example (copy pasted below), but since this is updating multiple examples I'm not sure if we'd want that text repeated 5+ times on the page. If you or anyone else has any thoughts on this they'd be appreciated.

Verbiage from Core example:
>Note: Table column widths will respond automatically when toggling expanded rows. To retain column widths between expanded and collapsed states, column header and/or data cell widths must be set.

thatblindgeye:
>For the second question, I don't have a strong opinion. I believe people can see the code example easily if they have trouble.
In case we can put this note only for the first one, that represents the main example.
Otherwise, we can move these examples and group them under the title expandable (like in the HTML example, but I don't know if currently we are following a logic to order our examples)

I'd agree with adding the note on the first/main expandable example. Depending what you think, another option in addition to adding the note on the main example could be to add onto the `isExpanded` prop description on the `Tr` component to include something along the lines of, "To prevent column widths from responding automatically when expandable rows are toggled, the width prop must also be passed into either the th or td component" (though I don't feel super strong about it, more just an additional option).

bartoval:
> To prevent column widths from responding automatically when expandable rows are toggled, the width prop must also be passed into either the th or td component

👍🏼  I think is a good idea to add this info into the TR doc

kmcfaul (approved):

tlabaj (approved):

PR #7515, opened 2022-06-06
feat(LabelGroup): add editable label examples

### LINKED ISSUE #7351: Label Group - add editable label group
Label Group

**Describe the feature**
Adds an editable label group according to the designs here: https://docs.google.com/document/d/1XfP0mgrIrTSUTTClCMl3hvU_CVE14jJmVv-LEOXdfuY/edit#

There are three ways that this can work that should be illustrated in the examples:

1. Clicking Add will add an editable label where the user can enter the label text.
2. Clicking Add will expose a select menu of choices to pick from and that text will be applied to the new label, which is always added at the start of the list.
3. Clicking Add will open a modal where the user can specify different label options based on project requirements.

kmcfaul:
**What**: Closes #7351

@mcarrano LMK if there are any adjustments I should make to the new examples and whether these should live in a demo page or remain on the examples page. I followed the scenarios pretty closely but didn't cover every shown variation (re: label colors, within groups or out, etc).

Added a new property `addLabel` to `LabelGroup` which puts the label within the label group but not included in the overflow list.

### DIFF
--- a/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
+++ b/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
@@ -58,6 +58,8 @@ export interface LabelGroupProps extends React.HTMLProps<HTMLUListElement> {
   hasEditableTextArea?: boolean;
   /** @beta Additional props passed to the editable textarea. */
   editableTextAreaProps?: any;
+  /** @beta Control for adding new labels */
+  addLabel?: React.ReactNode;
 }
 
 interface LabelGroupState {
@@ -149,6 +151,7 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
       isEditable,
       hasEditableTextArea,
       editableTextAreaProps,
+      addLabel,
       /* eslint-enable @typescript-eslint/no-unused-vars */
       ...rest
     } = this.props;
@@ -189,6 +192,7 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
                 </Label>
               </li>
             )}
+            {addLabel && <li className={css(styles.labelGroupListItem)}>{addLabel}</li>}
             {isEditable && hasEditableTextArea && (
               <li className={css(styles.labelGroupListItem, styles.modifiers.textarea)}>
                 <textarea className={css(styles.labelGroupTextarea)} rows={1} tabIndex={0} {...editableTextAreaProps} />
--- a/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
+++ b/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
@@ -59,7 +59,7 @@ export interface LabelGroupProps extends React.HTMLProps<HTMLUListElement> {
   /** @beta Additional props passed to the editable textarea. */
   editableTextAreaProps?: any;
   /** @beta Control for adding new labels */
-  addLabel?: React.ReactNode;
+  addLabelControl?: React.ReactNode;
 }
 
 interface LabelGroupState {
@@ -151,7 +151,7 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
       isEditable,
       hasEditableTextArea,
       editableTextAreaProps,
-      addLabel,
+      addLabelControl,
       /* eslint-enable @typescript-eslint/no-unused-vars */
       ...rest
     } = this.props;
@@ -192,7 +192,7 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
                 </Label>
               </li>
             )}
-            {addLabel && <li className={css(styles.labelGroupListItem)}>{addLabel}</li>}
+            {addLabelControl && <li className={css(styles.labelGroupListItem)}>{addLabelControl}</li>}
             {isEditable && hasEditableTextArea && (
               <li className={css(styles.labelGroupListItem, styles.modifiers.textarea)}>
                 <textarea className={css(styles.labelGroupTextarea)} rows={1} tabIndex={0} {...editableTextAreaProps} />
@@ -232,7 +232,7 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
       );
     };
 
-    return numChildren === 0 ? null : (
+    return numChildren === 0 && addLabelControl === undefined ? null : (
       <GenerateId>{randomId => renderLabelGroup(this.props.id || randomId)}</GenerateId>
     );
   }
--- a/packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
+++ b/packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
@@ -22,7 +22,7 @@ export const LabelGroupEditableAdd: React.FunctionComponent = () => {
 
   const onEdit = (nextText: string, index: number) => {
     const copy = [...labels];
-    copy[index] = { name: nextText, props: labels[index].editableProps };
+    copy[index] = { name: nextText, props: labels[index].props };
     setLabels(copy);
   };
 
@@ -46,15 +46,15 @@ export const LabelGroupEditableAdd: React.FunctionComponent = () => {
       categoryName="Label group 1"
       numLabels={5}
       isEditable
-      addLabel={
+      addLabelControl={
         <Label color="blue" variant="outline" isOverflowLabel onClick={onAdd}>
-          Add Label
+          Add label
         </Label>
       }
     >
       {labels.map((label, index) => (
         <Label
-          key={index}
+          key={`${label.name}-${index}`}
           id={`${label.name}-${index}`}
           color="blue"
           onClose={() => onClose(label.name)}
--- a/packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
+++ b/packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
@@ -24,7 +24,7 @@ export const LabelGroupEditableAdd: React.FunctionComponent = () => {
 
   const onEdit = (nextText: string, index: number) => {
     const copy = [...labels];
-    copy[index] = { name: nextText, props: labels[index].props };
+    copy[index] = { name: nextText, props: labels[index].props, id: labels[index].id };
     setLabels(copy);
   };
 
--- a/packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
+++ b/packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { LabelGroup, Label } from '@patternfly/react-core';
+
+export const LabelGroupEditableAdd: React.FunctionComponent = () => {
+  const [labels, setLabels] = React.useState<any>([
+    { name: 'Label 1' },
+    { name: 'Label 2' },
+    {
+      name: 'Label 3',
+      props: {
+        isEditable: true,
+        editableProps: {
+          'aria-label': 'label editable text'
+        }
+      }
+    }
+  ]);
+
+  const onClose = (label: string) => {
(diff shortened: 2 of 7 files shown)


mcarrano:
This is looking good @kmcfaul . Just a couple of comments...

* When using the dropdown to pick from a list of labels, I would not expect the resulting new label to be editable. The "pick" use case is for situations where the label name is constrained to some pre-existing list. 
* For the modal example, seems like the resulting label is editable even if I set Editable=No.

I think that you raise an interesting question about whether any of these examples are really demos. The Modal example definitively feels like it should be a demo since the content of the modal could be anything. I'd like to not imply that we must expose all of the options. Maybe the other two are OK as examples? Somehow we should make it clear that you can customize the Add function to be what you need it to be.

@mmenestr perhaps also take a look at this.

mmenestr:
This is looking good!
Just a few comments:
• For editable labels with add button, I would expect that when I click into the edit the label, the text doesn't change automatically to something else - I think it should stay saying "New label" until I manually type something else in? I also think I would expect it to remain editable after I've renamed it something other than "New label". @mcarrano please tell me if that's not expected...
• Agreed with Matt that for the dropdown one it probably shouldn't be editable post dropdown selection!
• For the modal one, when I select everything and submit, everything looks great and works as expected, but once I go in and try to edit the label that was created, it reverts back to what seems to be the default label? (Ie: A blue filled label with the words "Label 1"). I would expect it to retain the characteristics I had selected in the modal!

mcarrano:
> I would expect it to remain editable after I've renamed it something other than "New label". @mcarrano please tell me if that's not expected...

Yes, I agree with that.

tlabaj:
Other than the example where the control is alabel

> This is looking good @kmcfaul . Just a couple of comments...
> 
> * When using the dropdown to pick from a list of labels, I would not expect the resulting new label to be editable. The "pick" use case is for situations where the label name is constrained to some pre-existing list.
> * For the modal example, seems like the resulting label is editable even if I set Editable=No.
> 
> I think that you raise an interesting question about whether any of these examples are really demos. The Modal example definitively feels like it should be a demo since the content of the modal could be anything. I'd like to not imply that we must expose all of the options. Maybe the other two are OK as examples? Somehow we should make it clear that you can customize the Add function to be what you need it to be.
> 
> @mmenestr perhaps also take a look at this.

Other than the example that uses a label as a control, the other two feel like demos.  I wonder if we should move the dropdown and modal examples to demos.  We could add some text to the label example saying see demos for more advance/other ways to add a label?

packages/react-core/src/components/LabelGroup/LabelGroup.tsx
```diff
@@ -58,6 +58,8 @@ export interface LabelGroupProps extends React.HTMLProps<HTMLUListElement> {
   hasEditableTextArea?: boolean;
   /** @beta Additional props passed to the editable textarea. */
   editableTextAreaProps?: any;
+  /** @beta Control for adding new labels */
+  addLabel?: React.ReactNode;
```
tlabaj, line 62: Maybe `addLabelControl` or `addLabelAction` for a name would be more descriptive of prop intent?

srambach:
If I delete all the labels in the group, the "add label" and the entire label group also disappears. Should it stick around?

srambach:
Also, two other things I notice.
- If you created an editable label, then focus on it with mouse or keyboard to edit it, it always changes to be "Label 1". I'd expect it to retain its current value.
- In the modal, if you use the keyboard to navigate, selecting something from a dropdown throws focus to the close button. I'd expect it to go back to the dropdown.

mcarrano:
> If I delete all the labels in the group, the "add label" and the entire label group also disappears. Should it stick around?

That's a great catch @srambach . This was not covered in the design spec, but the way I'd propose this to work is that when the last label is removed, some placeholder text is added to the group that defaults to something like, "Add a new label," and the Add label button remains active so the user can add more labels to the group. What do you think @kmcfaul @mmenestr ?

Also, in reviewing this again, I noticed a small core issue that slipped through the cracks. Per our content guidelines, the button label should be "Add label" and not "Add Label."

mmenestr:
Idk that having an "add a new label" text would even be needed ? I think just making sure the "Add label" button never disappears is enough!

kmcfaul:
Updates: [@mcarrano @mmenestr]
- Removed editable flag for new options added by dropdown
- Fixed add via modal not respecting editable option
- Fixed label text changing for newly added labels
- Fixed label not being editable after editing
- Fixed label changing props after editing
- Renamed `addLabel` to `addLabelControl`
- Moved dropdown + modal add to demos; added description to point to demos
- Prevented label group disappearing when number of labels is 0 & `addLabelControl` is defined
- Updated add label button content to "Add label"

@evwilkin Is it possible to put demo `md` files inside the component folder? I was only able to get the docs build to pick up the new file when it is located in the demo folder (for now I have the `md` there and the code files are still in the component folder).

@srambach I'm looking into the focus issue, I believe it has to do with the fact the toggle is a `div` wrapping the `Label` rather than a typical `MenuToggle`-like component it was built to expect.

mcarrano (changes requested):
@kmcfaul This looks great now. I just stumbled over one odd bug. If I add multiple labels with the same text (i.e. "Label text 1), and then I remove one of them, it removes all the labels with the same text.

mmenestr:
Looks really good!! Just have a few more random things:

1. One random comment - and i'm not sure what other people would think, but: If I go into edit a label, and then open the dropdown before finishing my edit, the label I was editing goes into a weird in between state...

https://user-images.githubusercontent.com/51165119/172721100-e62ddf33-967b-4f58-a855-46533b4350a0.mov

But what I would expect to have happen is that if the label is mid edit when I click on the dropdown, it would revert back to its default state as soon as I click the dropdown. 

2. In the modal example, I'm wondering if the fields should be required or not? If I don't select anything, I notice that it just defaults to "No" on everything.. I don't think that's wrong per say, but seeing as the form asks the user to select a yes or no, It feels weird not to force them to select something? @mcarrano what are your thoughts on that?

Also as a side note (and might be something that needs to be fixed outside of this issue) but I noticed that when I try to edit a label by highlighting all the text (to eventually delete it), if I move my cursor past the label (like left of the label while I highlight) it comes out of the edit mode when I release my click - which is an annoying interaction. I feel like it should stay in the edit mode until I click away from it, not when I release a click on highlight. Let me know if that doesn't make sense 

https://user-images.githubusercontent.com/51165119/172721513-299c03a1-26fb-4a72-b9fe-0470ef251f6d.mov

mcarrano:
> In the modal example, I'm wondering if the fields should be required or not? If I don't select anything, I notice that it just defaults to "No" on everything.. I don't think that's wrong per say, but seeing as the form asks the user to select a yes or no, It feels weird not to force them to select something? @mcarrano what are your thoughts on that?

Since this is only a demo, I'm not too concerned with whether the fields are indicated as required or not. In a real use case I would expect a dialog that presents more application specific options. As for the yes/no options, it's generally good practice to have a default selection when using radio buttons so if the default is "no" then the radio buttons should be set to indicate that.

kmcfaul:
- Added default values for yes/no modal options and filled/outline modal option
- Adjusted dropdown example to not prevent event propagation on toggle click (which prevented editable label handler). This now has race condition between the label turning off the editable state and dropdown opening, where the menu opens off-center. I'm not sure how best to address this.
- Adjusted examples to close based on an ID to prevent matching labels being removed all at once.

tlabaj:
> @kmcfaul these two lines in the `source` file in react-docs define the file structure where the PF theme is looking for .md files and which tabs it's mapping them to:
> 
> * https://github.com/patternfly/patternfly-react/blob/main/packages/react-docs/patternfly-docs.source.js#L39-L40
> 
> Based on this you could move the demo code into a new `demos` directory like one of these options:
> 
> * `LabelGroup/examples/demos/[file-here].md`
> * `LabelGroup/demos/[file-here].md`
> 
> For consistency I'd check with @tlabaj if you want to do this as I don't believe we organize any other existing demos this way currently.

@kmcfaul currently all our demos live in `patternfly-react/packages/react-core/src/demos`.  Probably left over from when the demos where in a separate section on the website.  We can discuss whether it makes more sense to restructure where we put the demos in the code.  I agree that it may make more sense for them to live with the component.

tlabaj (approved):

mcarrano (approved):

mmenestr (approved):
Looks great to me too !! Thanks @kmcfaul

srambach:
A couple of keyboard things -
- When I edit a label and hit return to persist, the focus is not shown. Tab puts it back onto the control but if it could be there immediately that would be better.
- If I add some labels, then add but just press enter on the dialog, all the ones I added disappear
- I'm having trouble consistently reproducing it, but a couple of times, using just keyboard, I've added some editable labels, edited one or two, and hit one of the delete buttons and all the ones I created disappeared.

wise-king-sullyman (approved):
Looks great!

packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const onClose = (labelId: string) => {
+    setLabels(labels.filter((l: any) => l.id !== labelId));
+  };
+
+  const onEdit = (nextText: string, index: number) => {
+    const copy = [...labels];
+    copy[index] = { name: nextText, props: labels[index].props };
+    setLabels(copy);
+  };
```
thatblindgeye, line 29: For the bug shown in the video above, it looks like the issue occurs here; the `label.id` ends up getting removed after editing.

thatblindgeye (commented):
This looks great! I believe I'm also seeing the same behavior that @srambach is seeing, albeit this video is using mouse (I was able to recreate the behavior with keyboard only, though):

https://user-images.githubusercontent.com/70952936/173415160-4e2db066-ffa5-44e1-9aa3-1ea87cb72be0.mov

I would also agree with focus being placed on the editable label after editing has concluded with the Enter key, though since this component is still beta I wonder if addressing a11y concerns in a separate issue would be better.

packages/react-core/src/components/LabelGroup/examples/LabelGroupEditableAdd.tsx
kmcfaul: Ah good catch, I missed preserving that when I updated everything to use an id instead of name.

kmcfaul:
I will open a follow up to address keyboard navigation for editable labels.

thatblindgeye (approved):
Looks good! 🎉

tlabaj:
> @srambach The `onClick` event uses the position of the cursor on mouse-up so if the cursor is outside when released it will be detected as an outside click (cancelling the edit). You can properly select the text as long as the cursor is still within the input, or by clicking several times to select all.
> 
> Would we rather have editing be active using `onMouseDown`?

cc @mmenestr

mmenestr:
If i'm understanding this correctly, I would vote MouseUp for editing being active!

kmcfaul:
> If i'm understanding this correctly, I would vote MouseUp for editing being active!

That's currently what is happening (because the click event position is the MouseUp position). The caveat is that when you click and drag, to highlight text, you potentially will trigger a cancelled edit if you release outside the input.

thatblindgeye:
@kmcfaul @mmenestr If using MouseDown would resolve that issue without causing any other major issues, I think it would be more beneficial to use that instead. One benefit could be for users who may not have the accuracy to keep the mouse within the input the entire time when selecting that text.

srambach:
> One benefit could be for users who may not have the accuracy to keep the mouse within the input the entire time when selecting that text.

+1

kmcfaul:
@srambach @thatblindgeye @tlabaj @mmenestr 

Editing will now activate `onMouseDown` for mouse interaction, keyboard interaction will be the same. There shouldn't be a major difference in feel for the mouse interaction, just the editable field activating a fraction earlier.

srambach (approved):
Yay! Much better feel now when selecting text to edit. Once the focus issues are addressed later, this will be really slick. Thanks for addressing the comments. :)

PR #7516, opened 2022-06-06
feat(Form): add functionality for form group roles

### LINKED ISSUE #6097: Form - support for role="group" and role="radiogroup"
https://github.com/patternfly/patternfly/pull/4240

A form group with a single label and multiple form elements (text inputs, checkboxes, etc), should have `role="group"` with `aria-labelledby` the `id` of the `<label>` in that form group.
  * Exception to this is if the elements are a grouping of radio buttons where only one can be selected at a time, use `role="radiogroup"`

cc @jessiehuff

### LINKED ISSUE #7517: [Form] - [add group role to section and field group]
Follow up to https://github.com/patternfly/patternfly/pull/4424/

thatblindgeye:
**What**: Closes #6097, closes #7517

Followup from https://github.com/patternfly/patternfly/pull/4240 and  https://github.com/patternfly/patternfly/pull/4424

[Form preview build](https://patternfly-react-pr-7516.surge.sh/components/form)

Added a prop object to the FormGroup component, which when passed in will:

- render a `span` instead of a `label` element
- link the above `span` to the `.pf-c-form__group` element via the `aria-labelledby` attribute
- apply either `role="group"` or `role="radiogroup"` to the `.pf-c-form__group` element, depending on whether the `multipleInputs.isRadioGroup` prop is set to true or not

Also update examples and added a radio form group to more closely resemble the Core example.

**Additional issues**:


packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -41,6 +47,9 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   helperTextInvalidIcon?: React.ReactNode;
   /** ID of the included field. It has to be the same for proper working. */
   fieldId: string;
+  /** Object for identifying whether the form group has multiple inputs for a single label, and
+   * whether the multiple inputs are radio inputs or not. */
+  multipleInputs?: MultipleInputsObject;
```
wise-king-sullyman, line 52: In some examples you're passing `multipleInputs` as a boolean, but it's just typed as an object here. I think that either those examples should be updated to pass an empty object or the type here should be updated to be an object or a boolean.

If you go the type updating route though I feel like this prop name could use a slight edit.

I do also wonder if (as we've talked about a bit) this API would maybe make more sense with two separate props instead of an object prop that wraps another singular prop. I'm not positive how I feel about this idea though and would love input from others 🙂

nicolethoen (approved):

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
```diff
@@ -203,4 +203,52 @@ describe('FormGroup', () => {
...
+          <input />
+          <input />
+        </FormGroup>
+      );
+
+      const labelElement = screen.getByRole('group').querySelector('.pf-c-form__label');
+
+      expect(labelElement.tagName).toBe('SPAN');
+    });
+  });
```
wise-king-sullyman, line 253: One test I would add here: a sanity check test for the expected behavior when `multipleInputs` is not passed or is otherwise falsy.

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
```diff
@@ -203,4 +203,52 @@ describe('FormGroup', () => {
...
+      render(
+        <FormGroup label="label" fieldId="correct-label-element" multipleInputs={{ isRadioGroup: false }}>
+          <input />
+          <input />
+        </FormGroup>
+      );
+
+      const labelElement = screen.getByRole('group').querySelector('.pf-c-form__label');
+
+      expect(labelElement.tagName).toBe('SPAN');
```
wise-king-sullyman, line 251: This feels very non-RTL to me, but I'll have to think about if there's a way to test this that would be more in line with a real user experience and circle back to you. There very well might not be.

wise-king-sullyman (changes requested):
Overall this looks great! I think the testing is really solid but I did make one suggestion and have one thing I'll need to look into more and get back to you about.

I also noticed one slight typing/demo issue and that's the only real "hard" blocker out of all my comments.

packages/react-core/src/components/Form/FormGroup.tsx
thatblindgeye: I definitely understand the concern with the API here. A few ways that we could go about this off the top of my head:

- Update the interface so that `isRadioGroup` is required when passing in `multipleInputs`, and updating examples so that either `multipleInputs={{ isRadioGroup: true }}` or `multipleInputs={{ isRadioGroup: false }}` is passed in (as opposed to *just* `multipleInputs`
- Renaming the prop to something along the lines of `groupOptions` or `groupType`, then adding in another prop to the interface along the lines of `inputAmount: 'single' | 'multiple'` (more similar to your DataList implementation). This would result in a prop looking like `groupOptions={{ inputAmount: 'multiple', isRadioGroup: 'true' }}`. Not personally sure if this would make the most sense, however, since the default behavior is already a single input per form group, and this might require passing in props to retain that default behavior.
- Just breaking things out to two separate props like you mentioned, maybe `hasMultipleInputs` and `isRadioGroup`, and just noting in the `isRadioGroup` description that it should only be passed in when `hasMultipleInputs` is also passed in.

Tagging @nicolethoen as well for input on the above proposals

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
thatblindgeye: Do you think testing against a snapshot would make more sense here? Once unit tests are rewritten to align better with our current intentions, this would ideally only result in 2 snapshots for the component.

Another possibility *might* be testing user events. If there's a form group with a single input, clicking the label should place focus on the input; if there's a form group with multiple inputs, however, clicking the span won't place focus anywhere.

packages/react-core/src/components/Form/FormGroup.tsx
wise-king-sullyman: Yeah those are all the direct options I can think of as well, and honestly I don't love any of them. Your current API implementation is probably my favorite of these options.

Is there any way we could automatically determine if a group has multiple inputs rather than requiring the consumer to pass that in as a prop? I feel like that would really be the best approach from the consumer side.

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
wise-king-sullyman: > Another possibility might be testing user events. If there's a form group with a single input, clicking the label should place focus on the input; if there's a form group with multiple inputs, however, clicking the span won't place focus anywhere.

That seems like it aligns super well with the actual difference in UX and is the way I would go, awesome idea!

packages/react-core/src/components/Form/FormGroup.tsx
thatblindgeye: We could possibly check whether the `children` prop of the Form Group is an array (more than 1 input will return true), but this could possibly get skewed if any non-input content is rendered within the Form Group, such as helper text or validation text. I tinkered around a little and there's some workaround to that, but it may not be 100% accurate all the time due to there being various variables to consider (if it's a native input element, whether it's a React component, whether that React component *is* an input, etc.)

This is just from some quick tinkering around, though, so I might be missing something more obvious.

wise-king-sullyman: Yeah that doesn't seem ideal. I have one vague idea taking shape which would involve in some way adding registration behavior of some sort to our input components. If that would even work though it would probably be out of scope for this issue, so for the moment at least I'm going to join in saying that your implementation is slick and I think I like it.

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
```diff
@@ -203,4 +204,70 @@ describe('FormGroup', () => {
...
+
+  test('input should receive focus when clicking label', () => {
+    render(
+      <FormGroup data-testId="form-group-test-id" label="label" fieldId="native-label-element">
+        <input id="native-label-element" />
+      </FormGroup>
+    );
+
+    const labelElement = screen.getByTestId('form-group-test-id').querySelector('.pf-c-form__label');
+    const input = screen.getByTestId('form-group-test-id').querySelector('input');
```
wise-king-sullyman, line 216: To better replicate the real user experience I would like if we could get away from relying on test ids to select these elements if possible.

I haven't had the time to test this myself yet to confirm, but I think that this should be possible via something like:

```suggestion
    const labelElement = screen.getByText('label');
    const input = screen.getByRole('textbox');
```

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
```diff
@@ -203,4 +204,70 @@ describe('FormGroup', () => {
...
+    test('inputs should not receive focus when clicking label with multipleInputs passed in', () => {
+      render(
+        <FormGroup label="label" fieldId="span-label-element" multipleInputs={{ isRadioGroup: false }}>
+          <input id="span-label-element" />
+          <input />
+        </FormGroup>
+      );
+
+      const labelElement = screen.getByRole('group').querySelector('.pf-c-form__label');
+      const inputs = screen.getByRole('group').querySelectorAll('input');
```
wise-king-sullyman, line 232: And for selecting multiple inputs you should be able to just use `getAllByRole` rather than `getByRole`.

wise-king-sullyman (commented):
I really like the testing approach here! I just have a couple of small things about the selectors being used.

packages/react-core/src/components/Form/__tests__/FormGroup.test.tsx
thatblindgeye: Ah good catch! The `getByText` is more obvious in hindsight, but didn't think of the `textbox` role 🔥

thatblindgeye:
Based on conversation, the latest update renames the new prop to simply "role", and all tests and examples have been updated accordingly.

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -47,9 +41,10 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   helperTextInvalidIcon?: React.ReactNode;
   /** ID of the included field. It has to be the same for proper working. */
   fieldId: string;
-  /** Object for identifying whether the form group has multiple inputs for a single label, and
-   * whether the multiple inputs are radio inputs or not. */
-  multipleInputs?: MultipleInputsObject;
+  /** Sets the role of the form group if multiple inputs are being passed in. If multiple radio inputs
+   * are passed in use "radiogroup", otherwise use "group" when passing in multiple of any other input.
+   */
+  role?: 'group' | 'radiogroup';
```
wise-king-sullyman, line 47: One last thought about this, would it be better to just type this as `string` and just conditionally check for `group` or `radiogroup` explicitly as needed, so that if a consumer wants a different role to apply they can do so?

cc @nicolethoen

wise-king-sullyman (commented):
I really like this, just one question.

packages/react-core/src/components/Form/FormGroup.tsx
nicolethoen: I agree - I think it can be string in case someone wants to explicitly set role=none or role=presentation or who knows what else

mcarrano (changes requested):
Seems like there is still a problem with the way checkboxes and radio buttons are used in the examples. I should never be able to select multiple radio button options like this:

Screen Shot 2022-06-09 at 12 16 49 PM

Can we change this to be consistent with the core demo that looks like this?

Screen Shot 2022-06-09 at 12 16 32 PM

packages/react-core/src/components/Form/FormGroup.tsx
thatblindgeye: In terms of the scope of the Form Group fix, if we allow any role to be passed in that might require another update in Core (right now I believe they're only applying either `group` or `radiogroup`). If the intent of a form group is to act more like a `fieldset` (which by default gets announced as a "group"), only allowing either of those two options seems like the better solution. If we'd want to allow any role to be passed, however, it might also be worth adding a disclaimer in the prop description about using other roles (basically a "be careful" sort of thing).

wise-king-sullyman: I think that we should allow any role to be passed through for now at least, since like @nicolethoen mentioned yesterday FormGroup is currently often used in a non-fieldset-like fashion. I also wonder if any consumer might currently be applying a different role to a FormGroup for some reason, in which case limiting it to only `group`/`radiogroup` could be breaking.

You bring up good points though, and this seems like a topic that deserves further discussion and maybe a change in the next breaking change release?

wise-king-sullyman (approved):
Looks great!

nicolethoen (approved):

mattnolting (approved):
LGTM 🎉

srambach (approved):
LGTM 👍

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -41,6 +41,10 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   helperTextInvalidIcon?: React.ReactNode;
   /** ID of the included field. It has to be the same for proper working. */
   fieldId: string;
+  /** Sets the role of the form group. Pass in "radiogroup" when the form group contains multiple
+   * radio inputs, or pass in "group" when the form group contains multiple of any other input type.
+   */
+  role?: string;
```
tlabaj, line 47: Can this only be `radiogroup` or `group`?  If so i would suggest defining a union.

jessiehuff (approved):
LGTM! Tested in VO, JAWS, and NVDA and all announce them as a group as expected. :)

PR #7518, opened 2022-06-06
chore(DashboardHeader): align DashboardHeader with core

### LINKED ISSUE #7429: [Dashboard wrapper] - [Update to Match core demos]
Dashboard wrapper

**Describe the feature**
Core has updated their fullscreen dashboard demos.  React should update the dashboard wrapper so that our full screen demos match core.

jenny-s51:
**What**: Closes #7429

### DIFF
--- a/packages/react-core/src/demos/examples/DashboardHeader.js
+++ b/packages/react-core/src/demos/examples/DashboardHeader.js
@@ -27,7 +27,6 @@ import CogIcon from '@patternfly/react-icons/dist/esm/icons/cog-icon';
 import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
 import QuestionCircleIcon from '@patternfly/react-icons/dist/esm/icons/question-circle-icon';
 import BellIcon from '@patternfly/react-icons/dist/esm/icons/bell-icon';
-import imgAvatar from '@patternfly/react-core/src/components/Avatar/examples/avatarImg.svg';
 
 export default class DashboardHeader extends React.Component {
   constructor(props) {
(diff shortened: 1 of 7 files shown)


nicolethoen (commented):
Looks like the left/right padding in the masthead, as well as the image size of the PatternFly logo are both a little larger on the react dashboard wrapper than the core demos. 

Jun-07-2022 13-35-14

Also, should this update include updates to the masthead in the page demos? Those are the only full page demos, i believe, that don't use the dashboard wrapper.

jenny-s51 (commented):
Thank you for your feedback @nicolethoen ! I think that gif is showing the difference between the react workspaces, i.e. on main vs on this branch. Did you mean to show the difference between the react and core workspaces?

I've increased the size of the masthead inset padding to match the size of the page inset as shown in the core demo, which might be causing the toolbar item spacing to look different by a few pixels. I've also increased the Brand/logo size to match the exact size in the core demos too.

I checked all the HTML elements and the components/styling all do seem to match core. Looks like the toolbar items are still a few pixels away from their placement in core though... would be awesome to get some opinions/feedback on this styling from the core folks! @mattnolting @mcoker 

Also updating the page demos is a great idea! I opened

jenny-s51 (commented):
Also, yes, it looks like we do need to update `Page` with the new DashboardWrapper @nicolethoen ! I've opened and assigned myself to https://github.com/patternfly/patternfly-react/issues/7526! Will plan to update those full screen demos with these changes once they are approved 👍

nicolethoen:
> Did you mean to show the difference between the react and core workspaces?

oh - you're right. I thought I had the core demo up... my bad.

nicolethoen (approved):

thatblindgeye (approved):

packages/react-core/src/demos/examples/DashboardHeader.js
```diff
@@ -147,7 +146,7 @@ export default class DashboardHeader extends React.Component {
                 <Button aria-label="Help" variant={ButtonVariant.plain} icon={<QuestionCircleIcon />} />
               </ToolbarItem>
             </ToolbarGroup>
-            <ToolbarItem visibility={{ default: 'hidden', sm: 'visible', md: 'visible', lg: 'hidden' }}>
+            <ToolbarItem visibility={{ default: 'visible', sm: 'visible', md: 'visible', lg: 'hidden', xl: 'hidden' }}>
```
mcoker, line 149: Untested, but I believe all but `lg` are needed - it should be visible by default, and once it's set at a breakpoint, it persists through all breakpoints by default.

```suggestion
            <ToolbarItem visibility={{ lg: 'hidden' }}>
```

mcoker (approved):
LGTM!!!

PR #7520, opened 2022-06-06
docs(JumpLinks): add jump links with drawer demo

### LINKED ISSUE #6582: JumpLinks or Drawer?: build demo for a page with jumplinks and a drawer
to be completed after https://github.com/patternfly/patternfly/issues/4512

requested by @jgiardino and @srambach

wise-king-sullyman:
**What**: Closes #6582 

**Additional issues**:

One item on this that I'm not sure about: I didn't see how the drawer was to be toggled in the core demo. I took a similar approach to the switch used to demonstrate vertical vs horizontal capabilities in the other JumpLinks demo, but am more than happy to change it.

Convenience link to the jump links demos page: https://patternfly-react-pr-7520.surge.sh/components/jump-links/react-demos

And the new demo itself: https://patternfly-react-pr-7520.surge.sh/components/jump-links/react-demos/with-drawer/

### DIFF
--- a/packages/react-core/src/demos/JumpLinks.md
+++ b/packages/react-core/src/demos/JumpLinks.md
@@ -136,5 +136,7 @@ ScrollspyH2 = () => {
 
 This demo shows how jump links can be used in combination with a drawer.
 
+The scrollable selector passed to the jump links component is an id that was placed on the drawer content component.
+
 ```js isFullscreen file="./examples/JumpLinks/JumpLinksWithDrawer.js"
 ```
--- a/packages/react-core/src/demos/JumpLinks.md
+++ b/packages/react-core/src/demos/JumpLinks.md
@@ -130,3 +130,11 @@ ScrollspyH2 = () => {
   );
 };
 ```
+
+
+### With drawer
+
+This demo shows how jump links can be used in combination with a drawer.
+
+```js isFullscreen file="./examples/JumpLinks/JumpLinksWithDrawer.js"
+```
--- a/packages/react-core/src/demos/JumpLinks.md
+++ b/packages/react-core/src/demos/JumpLinks.md
@@ -136,7 +136,7 @@ ScrollspyH2 = () => {
 
 This demo shows how jump links can be used in combination with a drawer.
 
-The scrollable selector passed to the jump links component is an id that was placed on the drawer content component.
+The `scrollableSelector` prop passed to the jump links component is an `id` that was placed on the `DrawerContent` component.
 
 ```js isFullscreen file="./examples/JumpLinks/JumpLinksWithDrawer.js"
 ```
--- a/packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
+++ b/packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
@@ -29,9 +29,10 @@ export const JumpLinksWithDrawer = () => {
 
   React.useEffect(() => {
     const masthead = document.getElementsByClassName('pf-c-masthead')[0];
+    const drawerToggleSection = document.getElementsByClassName('pf-c-page__main-section')[0];
 
     getResizeObserver(masthead, () => {
-      setOffsetHeight(masthead.offsetHeight);
+      setOffsetHeight(masthead.offsetHeight + drawerToggleSection.offsetHeight);
     });
   }, []);
 
--- a/packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
+++ b/packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
@@ -11,6 +11,7 @@ import {
   JumpLinks,
   JumpLinksItem,
   JumpLinksList,
+  PageSection,
   Sidebar,
   SidebarContent,
   SidebarPanel,
@@ -79,7 +80,9 @@ export const JumpLinksWithDrawer = () => {
                   </JumpLinks>
                 </SidebarPanel>
                 <SidebarContent>
-                  <Button onClick={onToggleClick}>Toggle drawer</Button>
+                  <PageSection sticky="top" variant="light">
+                    <Button onClick={onToggleClick}>Toggle drawer</Button>
+                  </PageSection>
                   <TextContent>
                     {headings.map(heading => (
                       <div key={heading} style={{ maxWidth: '800px', marginBottom: '32px' }}>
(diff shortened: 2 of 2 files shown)


packages/react-core/src/demos/JumpLinks.md
```diff
@@ -130,3 +130,11 @@ ScrollspyH2 = () => {
   );
 };
 ```
+
+
+### With drawer
+
+This demo shows how jump links can be used in combination with a drawer.
```
nicolethoen, line 137: Could you also call out what you used as the `scrollableSelector` in this case? Just for clarity since I believe that was the major question which lead to the development of these demos

nicolethoen (changes requested):
For clarity - maybe double the length of the gibberish text in each jump links section? At my default window width, there is no need to use jumplinks/no scrolling.

packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
```diff
@@ -0,0 +1,124 @@
...
+                        <JumpLinksList></JumpLinksList>
+                      </JumpLinksItem>
+                    ))}
+                  </JumpLinks>
+                </SidebarPanel>
+                <SidebarContent>
+                  <Button onClick={onToggleClick}>Toggle drawer</Button>
+                  <TextContent>
+                    {headings.map(heading => (
+                      <div key={heading}>
```
jenny-s51, line 85: We should probably match the `div` styling here to the `div` styling in the other Jump Links demo! 

Looks like there's currently some missing spacing between the text content and the headers. Setting the `maxWidth` here keeps all the text visible when the drawer is toggled too 🙂

```suggestion
                      <div key={heading} style={{ maxWidth: '800px', marginBottom: '32px' }}>
```

jenny-s51 (changes requested):
This is looking good @wise-king-sullyman! Just one comment.

jenny-s51 (approved):

mcarrano (commented):
This looks good @wise-king-sullyman . If it's not too much trouble, can the toggle be placed in a sticky page section so it always stays at the top when the content scrolls?

@jgiardino @srambach looks like the original request for this came from you. Is this what you were hoping to see?

wise-king-sullyman:
@mcarrano does this implementation of your sticky page section suggestion look good to you? The default variant looked a bit out of place to me so I went with the light variant, but I'm happy to change it.

srambach:
This looks great; definitely showing what I had hoped, how to put both a drawer and jumplinks together in a page. I did notice though that the Skip to Content link doesn't go anywhere - I'd expect it to jump me to the main content. What do you think?

wise-king-sullyman:
@srambach it seems like VO is flaky on the skip to content in this demo for me for some reason. Sometimes it doesn't go anywhere, sometimes it accurately jumps to main, and sometimes it jumps to the jump links container and I can't figure out why.

If you/anyone have any ideas about why that is I am all ears.

srambach:
> @srambach it seems like VO is flaky on the skip to content in this demo for me for some reason. Sometimes it doesn't go anywhere, sometimes it accurately jumps to main, and sometimes it jumps to the jump links container and I can't figure out why.

Hmm - I wasn't even trying VO, I was just using the tab key.

wise-king-sullyman:
Ah, using tab it seems like the page jumps back to the top but focus does seem to stay on the masthead. This seems to be the case in the other jump links demo and some other demos (the primary detail demos as an example) as well from some quick testing.

I wonder if this might be a bug with the dashboard wrapper.

thatblindgeye (approved):
Looks good! In terms of the skip link issue mentioned, the behavior I'm seeing I believe matches @wise-king-sullyman where focus looks to remain in the masthead/on the skip link button after clicking it.

Otherwise the skip link does work for me, albeit it's difficult to tell since the main content doesn't have a very large scroll (if I scroll in the content area with the "First section" etc headings, the `main` element scrolls a little and then clicking the skip link jumps the page back up the tiny bit).

nicolethoen:
> I wonder if this might be a bug with the dashboard wrapper.

 If you want to create an issue, that can be something we look at in the future.

packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
```diff
@@ -0,0 +1,128 @@
...
+        </DrawerActions>
+      </DrawerHead>
+    </DrawerPanelContent>
+  );
+
+  return (
+    <DashboardWrapper breadcrumb={null} mainContainerId="scrollable-element">
+      <Drawer isExpanded={isExpanded}>
+        <DrawerContent panelContent={panelContent} id="jump-links-drawer-drawer-scrollable-container">
+          <DrawerContentBody>
```
mattnolting, line 64: Looks like there's a duplicate `<DrawerContentBody>` here

packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
```diff
@@ -0,0 +1,128 @@
...
+            <DrawerContentBody hasPadding>
+              <Sidebar>
+                <SidebarPanel variant="sticky">
```
mattnolting, line 67: The `<SidebarPanel variant="sticky">` is jumping on scroll because `--pf-c-sidebar__panel--Top` is set to `0`. To correct this, either the `.pf-c-sidebar__panel` has to receive padding or `--pf-c-sidebar__panel--Top` has to be changed to match the `.pf-c-drawer__body.pf-m-padding` padding value.

packages/react-core/src/demos/examples/JumpLinks/JumpLinksWithDrawer.js
thatblindgeye: The HTML demo also has this issue:

HTML jump links demo DOM

Would HTML need to be updated to fix that, or since this React demo aligns with HTML would it be fine leaving it as-is?

mcarrano (approved):
Looks good to me.

srambach:
Note that if you use the Skip to Content link, the jumplinks stop working. If you think this is because of the bug with Skip to Content noted above, then I guess that issue should be updated; otherwise it would be a new one.

thatblindgeye:
@srambach Created #7561 as a followup to that behavior.

srambach (approved):

nicolethoen (approved):

PR #7521, opened 2022-06-07
chore(Form): convert examples to TypeScript/functional components

### LINKED ISSUE #5767: Form: examples: incorrect usage of `fieldId` property in example code
In https://www.patternfly.org/v4/components/form/#horizontal-stacked-helper-text-on-top for example the `fieldId` is set to `options`, the latter Id does not exist.

### LINKED ISSUE #7466: Form

andyyvo:
**What**: Closes #7466 . Closes #5767 . Converted all Form component React demo examples into TypeScript. Two things to watch: **FormInvalid.tsx** and **FormInvalidWithFormAlert.tsx** swapped `noval` to `default`; **FormValidated.tsx** originally had a small bug where the `setTimeout` callback would stack based on the number of key inputs, but was fixed using `useEffect`.

**Additional issues**: N/A


packages/react-core/src/components/Form/examples/FormGroupLabelInfo.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { Form, FormGroup, TextInput, Popover } from '@patternfly/react-core';
+import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+
+export const FormGroupLabelInfo: React.FunctionComponent = () => {
+  const [value1, setValue1] = React.useState(''); // name
```
jenny-s51, line 6: Looks like you only have one field here, so you could probably just go with `value` or `name`!

```suggestion
  const [name, setName] = React.useState('');
```

packages/react-core/src/components/Form/examples/FormLimitWidth.tsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import { Form, FormGroup, TextInput, Checkbox, Popover, ActionGroup, Button } from '@patternfly/react-core';
+import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+
+export const SimpleForm: React.FunctionComponent = () => {
+  const [value1, setValue1] = React.useState(''); // name
```
jenny-s51, line 6:
```suggestion
  const [name, setName] = React.useState('');
```

nicolethoen (changes requested):

packages/react-core/src/components/Form/examples/FormBasic.tsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import { Form, FormGroup, TextInput, Checkbox, Popover, ActionGroup, Button } from '@patternfly/react-core';
+import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+
+export const SimpleForm: React.FunctionComponent = () => {
+  const [value1, setValue1] = React.useState(''); // name
```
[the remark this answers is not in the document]
nicolethoen: across all examples

packages/react-core/src/components/Form/examples/FormBasic.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const [value2, setValue2] = React.useState(''); // email
+  const [value3, setValue3] = React.useState(''); // phone
+
+  const handleValue1Change = value1 => {
```
jenny-s51, line 10:
```suggestion
  const handleNameChange = name => {
```

jenny-s51 (changes requested):
This is looking great @andyyvo! Nice update to the Validated example with `useEffect` 👍

I agree with @nicolethoen and was just about to comment the same - for best practice + readability, it would be good to update the hook variable names and value change methods with names that correspond to the fields directly!

packages/react-core/src/components/Form/examples/FormBasic.tsx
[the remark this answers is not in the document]
jenny-s51: +1 @nicolethoen

andyyvo:
Sounds great, thanks @nicolethoen and @jenny-s51 !! I have a few questions that I'd love to learn more about for future reference.

1. When it comes to naming conventions, should the state variable (and corresponding handler function) names be written based on their respective input? For example, the Full Name input uses `const [name, setName]` and Email uses `const [email, setEmail]`. I'd love to know how to best approach components with one state variable vs. a couple state variables vs. many state variables (Field groups).
2. When making changes to an existing PR, should I go back into my code editor, make the changes there, and then push a new PR? I hope I don't sound like too much of a novice but while I've worked with Git for awhile, class projects usually don't use PRs often.

Thank you so much!!

packages/react-core/src/components/Form/examples/FormHorizontalHelper.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const [helperText] = React.useState('Select all that apply:');
+
+  return (
+    <Form isHorizontal>
+      <FormGroup
+        label="Label"
+        helperText={helperText}
+        isHelperTextBeforeField
+        hasNoPaddingTop
+        fieldId="options"
```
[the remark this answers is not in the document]
nicolethoen: I think that eric is actually resolving this exact issue in this PR: https://github.com/patternfly/patternfly-react/pull/7516 so disregard this :)

andyyvo:
Just added all the cleaned up files. **FormFieldGroups.tsx** is still something that needs work. I'm running into an issue with object destructuring. It appears that `handleChange` is running into an issue on line 65 where it can't read `name`. I think the problem is that in the HTML attributes where I'm referencing `handleChange`, the `name` attribute for `TextInput` has hyphens in the string.

andyyvo:
Just wrapped up **FormFieldGroups.tsx**. Thank you so much, @nicolethoen !!!

nicolethoen (commented):
Looks great! one nit

tlabaj:
@andyyvo can you verify that this issue is fixed with this PR please.  If not, can you take a look at it sine you are in the code?
https://github.com/patternfly/patternfly-react/issues/5767

andyyvo:
@nicolethoen is #5767 the issue that you, me, and Eric were discussing about `fieldID` or `fieldGroups`?

nicolethoen:
@andyyvo it is not - I was talking about [this core issue](https://github.com/patternfly/patternfly/issues/4893)

packages/react-core/src/components/Form/examples/FormFieldGroups.tsx
```diff
@@ -0,0 +1,388 @@
...
+  const initialValues = {
+    'form-expandable-field-groups-label1': '',
+    'form-expandable-field-groups-label2': '',
+    'form-expandable-field-groups-field-group1-label1': '',
+    'form-expandable-field-groups-field-group1-label2': '',
+    'form-expandable-field-groups-field-group2-label1': '',
+    'form-expandable-field-groups-field-group2-label2': '',
+    'form-expandable-field-groups-field-group3-label1': '',
+    'form-expandable-field-groups-field-group3-label2': '',
+    'form-expandable-field-groupsform-expandable-field-groups-field-group1-label1': '',
```
jenny-s51, line 23: I think `form-expandable-field-groupsform-expandable-field-groups-field-group1-label1` is referring to one of the nested fields? 

We could call it `field-group1-nested-group1-label1` - would clear things up a bit 🙂

packages/react-core/src/components/Form/examples/FormFieldGroups.tsx
```diff
@@ -0,0 +1,388 @@
...
+    'form-expandable-field-groupsform-expandable-field-groups-field-group1-label2': '',
+    'form-expandable-field-group2-label1': '',
+    'form-expandable-field-group2-label2': '',
+    'form-expandable-field-group3-label1': '',
+    'form-expandable-field-group3-label2': '',
+    'form-expandable-field-groups-field-group7-label1': '',
```
jenny-s51, line 29: Same idea for these... it looks like in the demo there's no "Field group 7", "Field Group 8", or "Field group 10"  so it's tricky to understand which inputs these are referring to. 

We could have something like `field-group3-nested-field-group1-label1` here - WDYT?

packages/react-core/src/components/Form/examples/FormFieldGroups.tsx
```diff
@@ -0,0 +1,388 @@
...
+  FormFieldGroupExpandable,
+  FormFieldGroupHeader,
+  TextInput,
+  Button
+} from '@patternfly/react-core';
+import TrashIcon from '@patternfly/react-icons/dist/esm/icons/trash-icon';
+
+export const FieldGroups: React.FunctionComponent = () => {
+  const initialValues = {
+    'form-expandable-field-groups-label1': '',
```
jenny-s51, line 15: Is there any way we could clean these names up? Would be a lot more DRY and readable if we removed the `form-expandable-field-groups-` text from all of these!
```suggestion
    'label1': '',
```

andyyvo: Looks good and I totally agree @jenny-s51 ! The only issue I ran into with this was that the names of the labels in `initialValues` had to match the `id` and other props in the components used in order for the state variables to update the input. If I'm able to, I can change all of the id props to match it to something simpler?

jenny-s51: That sounds like a good plan @andyyvo !

packages/react-core/src/components/Form/examples/FormBasic.tsx
```diff
@@ -0,0 +1,122 @@
...
+          isRequired
+          type="tel"
+          id="simple-form-phone-01"
+          name="simple-form-phone-01"
+          placeholder="555-555-5555"
+          value={phone}
+          onChange={handlePhoneChange}
+        />
+      </FormGroup>
+      <FormGroup role="group" isInline fieldId="basic-form-checkbox-group" label="How can we contact you?">
```
nicolethoen, line 100: I don't think these `fieldId` props are service a purpose when you set a `role`

andyyvo: Should I remove all `fieldId` where there is a `role` then?

packages/react-core/src/components/Form/examples/FormFieldGroups.tsx
```diff
@@ -0,0 +1,324 @@
...
+    '1-label1': '',
+    '1-label2': '',
+    'group1-label1': '',
+    'group1-label2': '',
+    'group2-label1': '',
+    'group2-label2': '',
+    'group3-label1': '',
+    'group3-label2': '',
+    'exapnded-group1-label1': '',
+    'exapnded-group1-label2': '',
```
jenny-s51, line 24: Looks like we just need to make this is fixed throughout `FormFieldGroups.tsx`! I'm seeing this in a few places

```suggestion
    'expanded-group1-label2': '',
```

packages/react-core/src/components/Form/examples/FormFieldGroups.tsx
```diff
@@ -0,0 +1,324 @@
...
+            }
+          />
+        }
+      >
+        <FormGroup label="Label 1" isRequired fieldId="exapnded-group2-label1">
+          <TextInput
+            isRequired
+            id="exapnded-group2-label1"
+            name="exapnded-group2-label1"
+            value={inputValues['exapnded-group2-label1']}
```
jenny-s51, line 302: We should make sure the non-expandable example doesn't have a field with `expanded` in the name!

jenny-s51 (changes requested):
This is looking awesome @andyyvo !! Recent updates are excellent 👌

Left a couple comments about some small naming-related things - last few updates! Once those are addressed this PR should be good to  🚀!!

packages/react-core/src/components/Form/examples/FormFieldGroups.tsx
andyyvo: sounds good! I'll get right on that :D

jenny-s51 (approved):
Thanks @andyyvo !! Great job on this - LGTM 🤩

nicolethoen (changes requested):
Just need to resolve merge conflicts 👍

nicolethoen (approved):

nicolethoen (changes requested):
Looks like the a11y tests failed because of this:
```
@patternfly/react-docs: 240/873   /components/form/react/basic
@patternfly/react-docs: violations: duplicate-id-aria
@patternfly/react-docs: 241/873   /components/form/react/horizontal
@patternfly/react-docs: violations: duplicate-id-aria
@patternfly/react-docs: 242/873   /components/form/react/limit-width
@patternfly/react-docs: violations: duplicate-id-aria
```

More information here: https://patternfly-react-pr-7521-a11y.surge.sh/

andyyvo:
It appears that the accessibility issue derives from `role` and `fieldId`. More specifically, because `role` prop essentially removes the purpose of `fieldId`, we no longer define a `fieldId`. However, **FormGroup.tsx** uses the `fieldId` prop to create `aria-labelledby` and `id` (as seen on line 126 and 136 respectively of **FormGroup.tsx** of my most recent commit). Since there's no `fieldId`, it is undefined. This is resulting in a duplicate of `aria-labelledby` and `id` for the two instances of checkboxes/radio buttons in **FormBasic.tsx**, **FormHorizontal.tsx**, and **FormLimitWidth.tsx**. One possible solution I have in mind is to get rid of one example in each. However, that would still conflict for the consumer if they do try to implement both. A solution I just discovered is to define a `fieldId` when the `role` prop is used because `fieldId` is still processed for props mentioned before.

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -40,7 +40,7 @@ export interface FormGroupProps extends Omit<React.HTMLProps<HTMLDivElement>, 'l
   /** Icon displayed to the left of the helper text when the field is invalid. */
   helperTextInvalidIcon?: React.ReactNode;
   /** ID of the included field. It has to be the same for proper working. */
-  fieldId: string;
+  fieldId?: string;
```
thatblindgeye, line 43: If we decide to have this prop continue working as-is, i.e. setting that `aria-labelledby` on FormGroup, then this description could use updating. Something along the lines of below, but I'd hold off until Nicole responds to my other comment regarding this prop.

"ID of an individual field or a group of multiple fields. Required when a role of "group" or "radiogroup" is passed in. If only one field is included, its ID attribute and this prop must be the same."

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -120,38 +123,48 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
...
-          className={css(
-            styles.formGroupLabel,
-            labelInfo && styles.modifiers.info,
-            hasNoPaddingTop && styles.modifiers.noPaddingTop
-          )}
-          {...(isGroupOrRadioGroup && { id: `${fieldId}-legend` })}
+          className={css(styles.formGroup, className)}
+          {...(role && { role })}
+          {...(isGroupOrRadioGroup && { 'aria-labelledby': `${fieldId}-legend` })}
+          {...(!fieldId && { 'aria-labelledby': `${randomId}-legend` })}
```
thatblindgeye, line 132: Rather than having both of these lines to set `aria-labelledby`, they could be condensed down to one:

```suggestion
          {...(isGroupOrRadioGroup && { 'aria-labelledby': `${fieldId || randomId}-legend` })}
```

packages/react-core/src/components/Form/FormGroup.tsx
```diff
@@ -120,38 +123,48 @@ export const FormGroup: React.FunctionComponent<FormGroupProps> = ({
...
-            </React.Fragment>
+          {label && (
+            <div
+              className={css(
+                styles.formGroupLabel,
+                labelInfo && styles.modifiers.info,
+                hasNoPaddingTop && styles.modifiers.noPaddingTop
+              )}
+              {...(isGroupOrRadioGroup && { id: `${fieldId}-legend` })}
+              {...(!fieldId && { id: `${randomId}-legend` })}
```
thatblindgeye, line 143: Similar to the comment above about `aria-labelledby`

packages/react-core/src/components/Form/examples/FormGroupLabelInfo.tsx
```diff
@@ -0,0 +1,71 @@
...
+              onClick={e => e.preventDefault()}
+              aria-describedby="form-group-label-info"
+              className="pf-c-form__group-label-help"
+            >
+              <HelpIcon noVerticalAlign />
+            </button>
+          </Popover>
+        }
+        isRequired
+        helperText="Include your middle name if you have one."
```
thatblindgeye, line 57: Looks like `fieldId` was remove from here from the original example. Once that gets added back in it should resolve that a11y issue

thatblindgeye (changes requested):

thatblindgeye (approved):
🎉 Looks good!

nicolethoen (approved):

PR #7529, opened 2022-06-08
fix(Table): account for added columns with expandable

### LINKED ISSUE #3613: PF4 Table: fullWidth = true does not render correctly when rows are selectable or have actions in expanded content

I'm using the `expanded content`, `selectable` and `actions`, I configured my expanded content with `fullWidth=true`, but the output does not render in full width.

See here: https://codesandbox.io/s/recursing-keldysh-8qlci
The two tables are equal, but one is selectable while the other is not.
edit: Updated to include samples of code having actions on the parent row, but not on the contents.

Screenshot_2020-01-29_21-35-41

Bug, I don't know any workaround.

kmcfaul:
**What**: Closes #3613

Updates collapsible transform logic to account for additional added columns (actions, selectable, etc). Currently using the internal `firstUserColumnIndex` flag but more fine grained control can be added by checking each feature flag individually instead.

### DIFF
--- a/packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
+++ b/packages/react-table/src/components/Table/utils/decorators/collapsible.tsx
@@ -67,7 +67,7 @@ export const collapsible: IFormatter = (
 export const expandable: IFormatter = (value: IFormatterValueType, { rowData }: IExtra) =>
   rowData && rowData.hasOwnProperty('parent') ? <ExpandableRowContent>{value}</ExpandableRowContent> : value;
 
-export const expandedRow = (colSpan?: number, additionalColSpan?: number) => {
+export const expandedRow = (colSpan?: number, additionalColSpan: number = 0) => {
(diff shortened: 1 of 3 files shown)


nicolethoen (changes requested):
The Legacy table looks great!!

I'm not sure if the updates conflict with the composable table example, or if the expandable composable table examples just need a little work - It looks like colspan is calculated by the consumer in the example.

[This code sandbox](https://codesandbox.io/s/qjh5k8) was built using your PR, and highlights the fact that colspan is not being calculated the same way in the composable examples.

nicolethoen:
Should we fix the composable examples as part of this issue, @kmcfaul ?

kmcfaul:
@nicolethoen I'm having trouble tracing how the example is trying to apply the colspans. Will update if I can find a fix before freeze.

nicolethoen (approved):

tlabaj (approved):

PR #7530, opened 2022-06-08
fix(Select): check for updates in grouped select options

### LINKED ISSUE #7134: Bug - Select - Typeahead does not update immediatly when using SelectGroup
**Describe the problem**
Select component when using variant typeahead and typeaheadmulti does not get updated immediatly. You have to click something else or open/close the select multiple types to see the new content

- Use a select with groups
- Dynamically disable a field when the select is open
https://codesandbox.io/s/eager-turing-pdbn8v?file=/index.js

**Expected behavior**
New state of the select should be updated immediatly

No - I was going to switch to the checkbox mode, which updates immediatly, but came across this.

**Screenshots**
Peek 2022-03-29 07-41

**What is your environment?**
 - OS: Fedora
 - Browser Chrome
 - Version Version 99.0.4844.51 (Official Build) (64-bit)

Sidenote: I also saw something else when reproducing, on a previous version the selection was working OK - but now it only shows 1 selection even when using `typeaheadMulti`

kmcfaul:
**What**: Closes #7134

Adds logic to Select update function to account for select option groups.

### DIFF
--- a/packages/react-core/src/components/Select/Select.tsx
+++ b/packages/react-core/src/components/Select/Select.tsx
@@ -320,9 +320,11 @@ export class Select extends React.Component<SelectProps & OUIAProps, SelectState
       (this.props.isGrouped &&
         prevProps.children.some(
           (prevChild, index) =>
-            (prevChild.props.children && prevChild.props.children.length) !==
-              (this.props.children[index].props.children && this.props.children[index].props.children.length) ||
-            checkUpdatedChildren(prevChild.props.children, this.props.children[index].props.children)
+            prevChild.type === 'SelectGroup' &&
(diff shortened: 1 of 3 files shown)


tlabaj (changes requested):
Given the propensity for select to break, can you please add an integration test or a unit test to verify this behaviour?

nicolethoen (approved):
LGTM!

thatblindgeye (approved):
Looks good! Tested the preview build using the code from the codesandbox in the original issue, and the bug that was previously seen no longer occurs.

tlabaj (approved):
LGTM

PR #7532, opened 2022-06-09
fix(ProgressStepper): added onTriggerEnter prop to Popover

### LINKED ISSUE #7200: ProgressStepper - Popovers should be interactive via keyboard
Existing

**Describe the feature**
For the [Progress Stepper with Popover](https://www.patternfly.org/v4/components/progress-stepper#with-help-popover) example, the popovers can only be opened via mouse click or with Voice Over's VO + Space keyboard interaction (which acts as a click event I believe). Using just the keyboard, the popover cannot be opened with the Enter or Space keys like a button normally should.

The issue might be due to the fact that when a Popover is rendered for a ProgressStep it renders a `<span>` with a role attribute and value of "button" rather than a native `<button>` element. Using a `<button>` element might require a similar fix to be made in Core, however.

N/A

**Any other information?**
N/A

andyyvo:
**What**: Closes #7200 . The `Popover` for the `ProgressStepper` component can now be accessed via the keyboard using **Space** and **Enter** keys. The issue was due to the `Popover` component not utilizing the `onTriggerEnter` prop in **Popover.tsx**. It now works like a native button but in a discussion with @thatblindgeye , this issue could be resolved in core by changing the `span` element to a `button` element.

**Additional issues**: No issues but read above. There are alternative solutions that may be easier resolutions. Another point to consider is simplifying the code for the `onTriggerEnter` cases.

### DIFF
--- a/packages/react-core/src/components/Popover/Popover.tsx
+++ b/packages/react-core/src/components/Popover/Popover.tsx
@@ -368,6 +368,24 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
       }
     }
   };
+  const onTriggerEnter = (event: KeyboardEvent) => {
+    if (event.keyCode === KEY_CODES.ENTER || event.keyCode === KEY_CODES.SPACE) {
+      event.preventDefault();
+      if (triggerManually) {
+        if (visible) {
+          shouldClose(null, hide, event);
+        } else {
+          shouldOpen(show, event);
+        }
+      } else {
+        if (visible) {
+          hide();
+        } else {
+          show(true);
+        }
+      }
+    }
+  };
   const onContentMouseDown = () => {
     if (focusTrapActive) {
       setFocusTrapActive(false);
@@ -461,6 +479,7 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
         distance={distance}
         placement={position}
         onTriggerClick={onTriggerClick}
+        onTriggerEnter={onTriggerEnter}
         onDocumentClick={onDocumentClick}
         onDocumentKeyDown={onDocumentKeyDown}
         enableFlip={enableFlip}
--- a/packages/react-core/src/components/Popover/Popover.tsx
+++ b/packages/react-core/src/components/Popover/Popover.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import * as React from 'react';
-import { KeyTypes, KEY_CODES } from '../../helpers/constants';
+import { KeyTypes } from '../../helpers/constants';
 import styles from '@patternfly/react-styles/css/components/Popover/popover';
 import { css } from '@patternfly/react-styles';
 import { PopoverContext } from './PopoverContext';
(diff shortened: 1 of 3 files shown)


nicolethoen (commented):
I think this solution is good - I don't think we can change the span element to a button before the breaking change release since Progress stepper is not a beta component. Maybe we should open an issue in the core repo to document the thought for the breaking change release.

Can you also add a unit test to make sure that this onTriggerEnter prop continues working and we don't reintroduce this issue in the future?

nicolethoen:
Edit: Progress stepper is still beta - idk what I was looking at. 
I did open the issue in core, so maybe we can ask @mcoker and @tlabaj if they think it's better to do it on the core side first rather than pull this in.

tlabaj (commented):
Progress Stepper is still beta so we should be able to change the structure.  I think we should discuss with @mcoker if there are any concerns on the core side and wait for them to fix it there first so we stay in sync. 
cc @mcarrano

tlabaj (changes requested):
We should wait for https://github.com/patternfly/patternfly/issues/4898
 to be resolved before we take this one in.

mcoker:
@andyyvo yep! I just verified the changes needed to support this from core are in the main branch of the patternfly-react repo. The core change is just to add `border: 0` to the step-title element so that you can use a `<button>` HTML element there instead of a `<span>`. You can also remove the `tabindex` and `role` attribute when you switch the element to a `<button>` as those are redundant for the button element, though be sure and leave `type="button"` since a button's `type` is `submit` by default, and these buttons that trigger a popover are not submit buttons.

Screen Shot 2022-06-21 at 3 48 55 PM

packages/react-core/src/components/Popover/Popover.tsx
```diff
@@ -368,6 +368,24 @@ export const Popover: React.FunctionComponent<PopoverProps> = ({
...
+        }
+      } else {
+        if (visible) {
+          hide();
+        } else {
+          show(true);
+        }
+      }
+    }
+  };
```
thatblindgeye, line 388: I'm wondering if we still need this addition to Popover now that Progress Stepper is using a button. @tlabaj what do you think? I could see it being useful if someone *is* using a Popover on a non-button element at the very least.

thatblindgeye (changes requested):
Starting to look good! Left some comments/changes below.

packages/react-core/src/components/ProgressStepper/ProgressStep.tsx
```diff
@@ -92,6 +92,7 @@ export const ProgressStep: React.FunctionComponent<ProgressStepProps> = ({
           ref={stepRef}
           {...(popoverRender && { tabIndex: 0, role: 'button', type: 'button' })}
```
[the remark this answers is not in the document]
andyyvo: just made these changes

packages/react-core/src/components/Popover/Popover.tsx
andyyvo: I've pushed everything except for this suggested change (keep or remove). Let me know what the final verdict is and we should be all good to go on this!

tlabaj (approved):

thatblindgeye (approved):
Awesome work on this, especially for implementing that original logic to trigger the Popover on keyboard interaction! 💪🏼

nicolethoen (approved):

mcoker (approved):
Nice work!!

PR #7544, opened 2022-06-13
fix(NumberInput): leading zeros removed from inputs

### LINKED ISSUE #7117: Bug - NumberInput - Leading zeros
Leading zeros should be stripped, for example if the user types `0`, then adds another positive integer like `3`, the input should strip the leading zero and display `3` only.

Expected:
1. Type `405`
2. Delete first digit `4`
3. Input should update to show the value `5`

Actual:
1. Type `405`
2. Delete first digit `4`
3. Input shows `05`

Looks like if the final value assigned to the input in the jsx is a string, it works fine, but right now we're passing a number to it.

Current:
```
<input value={value} />
```

Possible solution:
```
<input value={Number(value).toString()} />
```

Related issue:
https://github.com/facebook/react/issues/9402

andyyvo:
**What**: Closes #7117 . Leading zero is removed from `NumberInput` component by implementing an `onBlur` prop for all inputs. If the user types 0, then adds another positive integer like 3, the input now strips the leading zero and display 3 only. This works for negative values and for decimal places.

**Additional issues**: The only issue was the **With Unit** component examining monetary amounts currently takes in a string and was thus, not implemented with the leading zero. Furthermore, the `toFixed(2)` function should probably be called on `onChange` as well as to prevent user inputs of anything passed two decimals.


packages/react-core/src/components/NumberInput/examples/NumberInput.md
```diff
@@ -37,6 +37,10 @@ class BasicNumberInput extends React.Component {
         value: this.state.value + 1
       });
     };
+
+    this.onBlur = (e) => {
+      e.target.value = Number(e.target.value).toString();
```
jenny-s51, line 42: It looks like we want to directly update the `value` state variable - this aligns with the implementation of all the other event handlers. 

Can we use state wherever `this.onBlur` is declared in `NumberInput.md`? 

```suggestion
      const newValue = Number(e.target.value).toString();
       this.setState({
         value: newValue
       });
```

jenny-s51 (changes requested):
Looking good @andyyvo - nice work; inputs are behaving as intended now. Left a comment about `onBlur`, will be just a small update! 

Otherwise LGTM 🙂

packages/react-core/src/components/NumberInput/examples/NumberInput.md
andyyvo: Good catch! I'll add that to all of the `onBlur` instances :)

andyyvo: We might not need to update the state in `onBlur` since we do that in `onChange` and console logging `value` in both `onChange` and `onBlur` appear to have the leading zero cut. I'm going to keep looking into this!

jenny-s51: Ah you're right! Did some investigating here too... since we need to keep the `toString()` for the leading zeros fix, I'm OK approving as-is -- matches how @jschuler wrote it in #7117 👍

andyyvo: Sounds great! I just made sure that `onBlur` was also added to the second example with the monetary amount

nicolethoen (commented):
Okay - if this should be the default behavior of the NumberInput, should we bring the `Number(e.target.value).toString();` logic into the component so the consumer doesn't have to implement this fix themself?

Something sort of like how the `onChange` prop is passed to the `TextInput` component and then called from an internal function (which could also include your extra logic) which is ultimately what's passed to the input element.

mcarrano (approved):
Looks good from a UX perspective. Thanks @andyyvo !

andyyvo:
> Okay - if this should be the default behavior of the NumberInput, should we bring the `Number(e.target.value).toString();` logic into the component so the consumer doesn't have to implement this fix themself?
> 
> Something sort of like how the `onChange` prop is passed to the `TextInput` component and then called from an internal function (which could also include your extra logic) which is ultimately what's passed to the input element.

+1 on this -- how would I go about changing the component itself to implement this logic?

jenny-s51:
@andyyvo you can try to put the `.toString()` conversion directly on `value` here: https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/NumberInput/NumberInput.tsx#L117

andyyvo:
Shout to @nicolethoen for this one! Just added a fix to **NumberInput.tsx** where a `handleBlur` function now renders the logic after any `onChange` or `onBlur` event so that the consumer no longer would have to write out this logic themselves. @jenny-s51 if you could take a look -- that would be great!

nicolethoen (approved):

tlabaj (changes requested):
Can you please add unit or integration test to verify the behaviour for the different scenarios (negative number, decimals, etc).

packages/react-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -87,6 +87,16 @@ export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
   const keyDownHandler =
     inputProps && inputProps.onKeyDown ? inputProps.onKeyDown : defaultKeyDownHandler({ inputName, onMinus, onPlus });
 
+  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    event.target.value = Number(event.target.value).toString();
```
tlabaj, line 91: I would create a local event variable and set it to the event then modify that.
```suggestion
    let ev = event.target.value;
    ev.target.value = Number(event.target.value).toString();
```

packages/react-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -87,6 +87,16 @@ export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
...
+    if (onChange) {
+      onChange(event);
```
tlabaj, line 93:
```suggestion
      onChange(ev);
```

packages/react-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -87,6 +87,16 @@ export const NumberInput: React.FunctionComponent<NumberInputProps> = ({
...
+    }
+    if (onBlur) {
+      onBlur(event);
```
tlabaj, line 96:
```suggestion
      onBlur(ev);
```

tlabaj (changes requested):

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
```diff
@@ -60,4 +61,90 @@ describe('numberInput', () => {
...
+
+  test('renders unit with negative decimal number', () => {
+    const { asFragment } = render(<NumberInput value={-12.4} unit="%" unitPosition="after" />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test('renders unit & position with negative decimal number', () => {
+    const { asFragment } = render(<NumberInput value={-12.4} unit="$" unitPosition="before" />);
+    expect(asFragment()).toMatchSnapshot();
+  });
```
wise-king-sullyman, line 108: @tlabaj are these tests for all these different possible combinations of input values and units something you were looking for?

I feel like the tests that I've highlighted here could be trimmed down to just a test for `unit` with `"before"` and a test for `unit` with `"after"` since the unit and its position aren't dependent on what the input value is.

tlabaj: @wise-king-sullyman I think that is fine to trim that down.  I was not specific enough with my comment.
I was actually looking for test to verify the trimming of the leading zeros works with different values ( decimal, negative numbers) as stated in description.

andyyvo: @wise-king-sullyman @tlabaj I'll remove the tests on units and position then and it should be all set for review

wise-king-sullyman: @andyyvo I think that @tlabaj is looking for tests which verify that the new behavior with leading zeros being removed works as expected, which I'm not currently seeing in the test suite. Please correct me if I'm wrong about my read of your last comment @tlabaj.

To do so I'm guessing you'll need to use `userEvent` to type a leading zero into the input then click outside of the input, then make an assertion about the value of the input, rather than just being able to pass it into `NumberInput` as a value.

tlabaj: @wise-king-sullyman that is correct.  This could also be done with a cypress integration test.

andyyvo:
Struggling to write a test for leading zeros. Here are all the tests I've tried that have failed:

```
test('removes leading zero', () => {
    render(<NumberInput value={0320} />); // this is an octal literal which is its own error
    expect(screen.getByDisplayValue('320')).toBeInTheDocument();

    const input = screen.getByRole('spinbutton'); // tried textbox and it doesn't work
    userEvent.type(input, '0'); // if initial value is 320, it remains 320

    expect(screen.getByRole('spinbutton')).toHaveValue('0320');
});
```

```
test('removes leading zero', () => {
    const val = (123).toString().padStart(5, '0'); // the issue here is that value takes in a number and not string
    render(<NumberInput value={parseInt(val)} />); // parseInt defeats the purpose of the test
    expect(screen.getByDisplayValue('00123')).toBeInTheDocument();
});
```

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
```diff
@@ -60,4 +61,82 @@ describe('numberInput', () => {
...
+  test('removes leading zero', () => {
+    render(<NumberInput value={10} onChange={() => {}} />);
+  
+    const input = screen.getByRole('spinbutton');
+    userEvent.type(input, '{arrowleft}{arrowleft}0');
+    expect(input).toHaveDisplayValue('010');
+  
+    userEvent.click(document.body)
+  
+    expect(input).toHaveDisplayValue('10');
```
wise-king-sullyman, line 140: Can you add tests to verify that the leading zeros are removed on decimal and negative numbers as well?

wise-king-sullyman (changes requested):
Looking good! I think there are a few other behaviors to test still though.

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
```diff
@@ -60,4 +61,128 @@ describe('numberInput', () => {
...
+
+  test('removes trailing zero', () => {
+    render(<NumberInput value={120.0} />);
+    expect(screen.getByDisplayValue('120')).toBeInTheDocument();
+  });
+
+  test('removes many trailing zeroes', () => {
+    render(<NumberInput value={120.0100} />);
+    expect(screen.getByDisplayValue('120.01')).toBeInTheDocument();
+  });
```
wise-king-sullyman, line 129: I don't think that these trailing zero tests are needed because I believe number inputs have this behavior by default.

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
```diff
@@ -60,4 +61,128 @@ describe('numberInput', () => {
...
+
+  test('removes a trailing zero from a negative number', () => {
+    render(<NumberInput value={-40.0} />);
+    expect(screen.getByDisplayValue('-40')).toBeInTheDocument();
+  });
+
+  test('removes trailing zeroes from a negative number', () => {
+    render(<NumberInput value={-40.0100} />);
+    expect(screen.getByDisplayValue('-40.01')).toBeInTheDocument();
+  });
```
wise-king-sullyman, line 151: See my previous comment regarding trailing zero testing.

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
```diff
@@ -60,4 +61,128 @@ describe('numberInput', () => {
...
+    render(<NumberInput value={-55.0} onChange={() => {}} />);
+  
+    const input = screen.getByRole('spinbutton');
+    userEvent.type(input, '{arrowleft}{arrowleft}0');
+    expect(input).toHaveDisplayValue('-055');
+  
+    userEvent.click(document.body)
+  
+    expect(input).toHaveDisplayValue('-55');
+  });
```
wise-king-sullyman, line 187: I don't believe these tests are needed either since the leading zero removal behavior is already covered in the preceding tests.

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
```diff
@@ -60,4 +61,128 @@ describe('numberInput', () => {
...
 });
```
wise-king-sullyman, line 188: Can you add a test for leading zero removal for decimal numbers?

wise-king-sullyman (changes requested):

packages/react-core/src/components/NumberInput/__tests__/NumberInput.test.tsx
andyyvo: Do we not need to do the same tests with negative numbers?

wise-king-sullyman: Yes, keep the leading zero removal test for negative numbers that you currently have, or do you mean a test for negative decimal numbers?

wise-king-sullyman (approved):
Looks great! Thanks for all your work on this and especially these tests 🚀

tlabaj (approved):
LGTM! Nice work!

jenny-s51 (approved):
🚀

nicolethoen (approved):

PR #7545, opened 2022-06-13
chore(OverflowMenu): use position and isFlipEnabled props correctly in docs

nicolethoen:
Addresses #7398 

Update various examples to pass `menuAppendTo='parent'` and `isFlipEnabled` to Dropdowns housed in OverflowMenus when it makes sense.

Examples updated:

- OverflowMenu
- Toolbar
- Card view demo
- Primary detail demo
- Sortable table demo

*note: the Dropdown menu has been configured so that isFlipEnabled flips the menus up/down depending on the available vertical space. The consumer can use the `position` prop to right align the menu when appropriate. That is the recommendation made to the issue reporter in this case.

### DIFF
--- a/packages/react-core/src/components/OverflowMenu/examples/OverflowMenu.md
+++ b/packages/react-core/src/components/OverflowMenu/examples/OverflowMenu.md
@@ -61,6 +61,8 @@ class SimpleOverflowMenu extends React.Component {
             isOpen={isOpen}
             isPlain
             dropdownItems={dropdownItems}
+            isFlipEnabled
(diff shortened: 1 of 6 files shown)


thatblindgeye (approved):

jenny-s51 (approved):
LGTM! Thank you @thatblindgeye for the `padding-top` suggestion! I was able to test the flipping behavior on the `Card` and `PrimaryDetail` demos and both look great now with your recent updates @nicolethoen

kmcfaul (approved):

PR #7550, opened 2022-06-15
chore(List): converted all examples to typescript

### LINKED ISSUE #7549: List: convert examples to TypeScript

andyyvo:
**What**: Closes #7549 . List examples have all been converted from JS to TS.

**Additional issues**: This is not a critical issue. However, when writing out the TS examples for **ListIcons.tsx** and **ListLargeIcons.tsx**, there was an error on line 6110 of `node_modules/typescript/lib/typescript.js` regarding `replace` being undefined in the `normalizeSlashes` function. This was solved by changing the name of the imports of the icons from `BookOpen` to `BookOpenIcon` in both the markdown and the TS file, which works since the icons are export default. I'm not entirely sure as to what was going on but this issue was resolved with that solution.

### DIFF
--- a/packages/react-core/src/components/List/examples/List.md
+++ b/packages/react-core/src/components/List/examples/List.md
@@ -5,97 +5,35 @@ cssPrefix: pf-c-list
 propComponents: ['List', 'ListItem']
 ---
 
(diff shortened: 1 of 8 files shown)


kmcfaul (approved):
lgtm!

PR #7557, opened 2022-06-15
fix(Switch): aria labelledby to update with change

### LINKED ISSUE #7468: Bug - Switch - aria label attributes don't update with checked value
**Describe the problem**
The `aria-labelledby` attribute doesn't seem to update based on the Switch's checked status despite there being rendered a `span` element with an id of `[id passed in]-off`, which makes it seem like the intent is that the `aria-labelledby` attribute _should_ update based on the checked status.

Additionally, the uncontrolled variant should explain (and possibly provide an example of how to do so) that the `aria-label` prop should dynamically update if the message would be different for the "on" or "off" state of the switch. See comment https://github.com/patternfly/patternfly-react/pull/7418#discussion_r874782499

1. Go to the [switch component](https://www.patternfly.org/v4/components/switch)
2. Open dev tools and enter the below code snippet:
    ```
    console.log(document.getElementById('simple-switch').getAttribute('aria-labelledby'), document.getElementById('simple-switch').checked)
    ```
3. Click the switch in the Basic example, then enter the above code snippet in the console again. Notice that while the checked state updates, the aria-labelledby attribute has the same value

**Expected behavior**
Because an element is rendered with the `labelOff` text similarly to how the `label` text gets rendered, the assumption is that the aria-labelledby attribute should update to announce the correct label.

gitdallas:
**What**: Closes #7468

### DIFF
--- a/packages/react-core/src/components/Switch/Switch.tsx
+++ b/packages/react-core/src/components/Switch/Switch.tsx
@@ -94,7 +94,9 @@ export class Switch extends React.Component<SwitchProps & OUIAProps, { ouiaState
           {...([true, false].includes(defaultChecked) && { defaultChecked })}
           {...(![true, false].includes(defaultChecked) && { checked: isChecked })}
           disabled={isDisabled}
-          aria-labelledby={isAriaLabelledBy ? `${this.id}-on` : null}
+          aria-labelledby={
+            !isAriaLabelledBy ? null : `${this.id}-${isChecked === true && defaultChecked !== false ? 'on' : 'off'}`
+          }
           {...props}
         />
         {label !== undefined ? (
--- a/packages/react-core/src/components/Switch/Switch.tsx
+++ b/packages/react-core/src/components/Switch/Switch.tsx
@@ -93,9 +93,7 @@ export class Switch extends React.Component<SwitchProps & OUIAProps, { ouiaState
           onChange={event => onChange(event.target.checked, event)}
           {...(defaultChecked !== undefined ? { defaultChecked } : { checked: isChecked })}
           disabled={isDisabled}
-          aria-labelledby={
-            !isAriaLabelledBy ? null : `${this.id}-${isChecked === true && defaultChecked !== false ? 'on' : 'off'}`
-          }
+          aria-labelledby={!isAriaLabelledBy ? null : `${this.id}-${isChecked !== true ? 'off' : 'on'}`}
           {...props}
         />
         {label !== undefined ? (
--- a/packages/react-core/src/components/Switch/Switch.tsx
+++ b/packages/react-core/src/components/Switch/Switch.tsx
@@ -91,8 +91,7 @@ export class Switch extends React.Component<SwitchProps & OUIAProps, { ouiaState
           className={css(styles.switchInput)}
           type="checkbox"
           onChange={event => onChange(event.target.checked, event)}
-          {...([true, false].includes(defaultChecked) && { defaultChecked })}
-          {...(![true, false].includes(defaultChecked) && { checked: isChecked })}
+          {...(defaultChecked !== undefined ? { defaultChecked } : { checked: isChecked })}
           disabled={isDisabled}
           aria-labelledby={
             !isAriaLabelledBy ? null : `${this.id}-${isChecked === true && defaultChecked !== false ? 'on' : 'off'}`
--- a/packages/react-core/src/components/Switch/examples/Switch.md
+++ b/packages/react-core/src/components/Switch/examples/Switch.md
@@ -170,9 +170,29 @@ import { Switch } from '@patternfly/react-core';
 import React from 'react';
 import { Switch } from '@patternfly/react-core';
 
-<React.Fragment>
-  <Switch id="uncontrolled-switch-on" label="Message when on" labelOff="Message when off" defaultChecked={false} />
-  <br />
-  <Switch id="uncontrolled-no-label-switch-on" aria-label="Message when on" defaultChecked={true} />
-</React.Fragment>;
+class UncontrolledSwitch extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      checked: true
+    };
+    this.handleChange = checked => {
+      this.setState({
+        checked
+      });
+    };
+  }
+  render() {
+    const { defaultChecked } = this.state;
+    return (
+      <Switch 
+        id="uncontrolled-switch"
+        label="Message when on"
+        labelOff="Message when off"
+        defaultChecked={this.state.checked}
+        onChange={this.handleChange}  
+      />
+    )
+  }
+}
 ```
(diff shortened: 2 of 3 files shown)


kmcfaul (approved):

packages/react-core/src/components/Switch/Switch.tsx
```diff
@@ -94,7 +94,9 @@ export class Switch extends React.Component<SwitchProps & OUIAProps, { ouiaState
           {...([true, false].includes(defaultChecked) && { defaultChecked })}
           {...(![true, false].includes(defaultChecked) && { checked: isChecked })}
           disabled={isDisabled}
-          aria-labelledby={isAriaLabelledBy ? `${this.id}-on` : null}
+          aria-labelledby={
+            !isAriaLabelledBy ? null : `${this.id}-${isChecked === true && defaultChecked !== false ? 'on' : 'off'}`
```
gitdallas, line 98: This may seem a little confusing... but docs say never to use both `isChecked` and `defaultChecked` together. 

`isChecked` is default `true`, so when using `defaultChecked` it's just checking that it isn't `false` to see whether or not we should use "on" or "off"

`defaultChecked` does not have a default checked, so `!== false` would always be true and now we are just checking that `isChecked === true`

packages/react-core/src/components/Switch/examples/Switch.md
```diff
@@ -170,9 +170,29 @@ import { Switch } from '@patternfly/react-core';
 import React from 'react';
 import { Switch } from '@patternfly/react-core';
 
-<React.Fragment>
-  <Switch id="uncontrolled-switch-on" label="Message when on" labelOff="Message when off" defaultChecked={false} />
-  <br />
-  <Switch id="uncontrolled-no-label-switch-on" aria-label="Message when on" defaultChecked={true} />
-</React.Fragment>;
+class UncontrolledSwitch extends React.Component {
```
kmcfaul, line 173: I missed that this was supposed to be an uncontrolled example. There shouldn't be any state updates that affect the Switch. What is the purpose of tracking the checked state here?

gitdallas: I guess I'm still not sure what the purpose of uncontrolled would be, but I am added tracking the state to show a consumer a way they would be able to track it, and passed the value back into the component as a way for the component to update the aria-labeledby value.

nicolethoen: I think it might make a little more sense or feel a little more intuitive to have a console statement in the onChange function in the example rather than updating an unused state variable.

And I also would have expected that the `defaultChecked` would be a hard coded true (for the case of the example). If the switch should be checked on page load, then they could use the `defaultChecked` prop. 

I don't think a new `ref` is necessary - as long as the `label` and `labeloff` props are updating the `aria-label` appropriately. Are we imagining the consumer needs another way to further update the aria-label?

gitdallas: @nicolethoen - label and labeloff will switch fine, but the aria-labelledby will not update along with it. this is why i was updating the defaultChecked value.. but perhaps i could find another way, like onchange or ref within the switch component?

thatblindgeye: @nicolethoen with Dallas' current solution, the `aria-labelledby` gets updated as I expected based on the `isChecked` prop value. With an uncontrolled variant, though, it gets trickier to update the `label`/`labelOff` or `aria-label` without that `isChecked` prop (we wouldn't necessarily need to update the `aria-labelledby` if we only pass the `label` prop and not `labelOff` since they'd both end up with the same text content that way).

My thoughts for adding a `ref` prop were basically to allow consumers to make updates/validations elsewhere based on the input value from the DOM since they wouldn't have that `isChecked` state to do so. This isn't something we do in other uncontrolled variants, though, so perhaps would be a separate discussion.

An alternative could be to just advise consumers to only use a static label/aria-label for an uncontrolled variant (rather than an on and off label like the other examples have) for now.

nicolethoen (commented):
I think you don't want to include a hardcoded aria-label in the `Checked with label1 example, because now the aria-labelledby is not being set at all in this example.

packages/react-core/src/components/Switch/examples/Switch.md
```diff
@@ -171,8 +171,12 @@ import React from 'react';
...
-  <Switch id="uncontrolled-switch-on" label="Message when on" labelOff="Message when off" defaultChecked={false} />
-  <br />
-  <Switch id="uncontrolled-no-label-switch-on" aria-label="Message when on" defaultChecked={true} />
+  <Switch
+    id="uncontrolled-switch-on"
+    aria-label="Uncontrolled switch"
+    label="Message when on"
+    labelOff="Message when off"
+    defaultChecked={false}
+  />
```
thatblindgeye, line 180: Similar to @nicolethoen comment from above. For now I could see either just removing the `aria-label` or showing two separate switches (one using the `aria-label` without a visible text label, and one using the `label`/`labelOff` props).

Updating the label/aria-label for uncontrolled might be something to revisit in the future.

thatblindgeye (changes requested):
Only had the one comment below with the current implementation, otherwise I think the update to the logic for setting `aria-labelledby` is awesome

packages/react-core/src/components/Switch/examples/Switch.md
gitdallas: I guess I'm a little unclear about this and @nicolethoen comments. I added the label/labelOff because this works properly for rendering on the screen.

The issue is that with uncontrolled, the aria-labelledby doesn't switch with it so might not make sense (a switch to "on" from "off" would still point to label, not labelOff). Adding a generic aria-label that indicates the checkbox and doesn't point to a specific label seemed like a solution for this... screen readers will still note the checkbox change I believe?

It seems like we'd either have to do what I did, remove the label/labelOff and use aria-label, or remove the labelOff. What would be preferred?

Fine but no visible label:

```
  <Switch
    id="uncontrolled-switch-no-label"
    aria-label="Uncontrolled switch"
    defaultChecked={false}
  />
```

Odd because though the label/labelOff will display correctly with the change, the labelledby will always point to label (which has something such as "on")

```
  <Switch
    id="uncontrolled-switch-labels"
    label="Message when on"
    labelOff="Message when off"
    defaultChecked={false}
  />
```

Will switch between the labels, but may not matter as the labels will have the same text content

```
  <Switch
    id="uncontrolled-switch-labels"
    label="An uncontrolled switch"
    defaultChecked={false}
  />
```

thatblindgeye: Ah good catch. Based on your latest comment, I would personally say one of the following:

1. Include both your first and third example in your comment above (one using only `aria-label` and one only using `label`), just to show the uncontrolled variant with a visible and non-visible label
2. Only include your third example that uses only the `label` prop

thatblindgeye (approved):

nicolethoen (approved):
Sorry if I added confusion to this - I think I was pretty confused myself - but it LGTM!

PR #7562, opened 2022-06-15
fix(CardViewDemo): remove highlight select cards (use checkboxes only), include some checked indicator

### LINKED ISSUE #7420: Bug - [Card view demo] - clean up card selection
**Describe the problem**
There are two issues with the existing card view demo that should be corrected:

1. Currently implementation uses selectable cards together with checkboxes. These should not be used together. The design intent was for these to be standard cards with checkboxes to allow selection for bulk actions but not selectable cards.

2. When some cards (but not all) are selected, the bulk select checkbox in the toolbar should display a mixed state (-). It currently does not. See https://staging.patternfly.org/v4/components/table/react-demos/bulk-select/ for an example of how this is intended to work.

gitdallas:
**What**: Closes #7420

### DIFF
--- a/packages/react-core/src/demos/Card/Card.md
+++ b/packages/react-core/src/demos/Card/Card.md
@@ -106,6 +106,13 @@ class CardViewBasic extends React.Component {
       totalItemCount: 10
     };
 
+    this.checkAllSelected = (selected, total) => {
+      if (selected && selected < total) {
+        return null;
+      }
+      return selected === total;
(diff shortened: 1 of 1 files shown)


mcarrano (approved):
Looks great. Thanks @gitdallas !

nicolethoen (approved):
LGTM!

thatblindgeye (approved):

PR #7569, opened 2022-06-16
feat(Wizard): allow for an optional step in the wizard

### LINKED ISSUE #7302: Allow for an optional step in the Wizard
After a lot of quick-and-dirty hacks, I was finally able to get an optional step in the `Wizard`. Only when the option for OpenShift is chosem we need to ask for the necessary pull secret. When you would choose Podman, it would not need this.

image

I came up with the following:
  * define a state that control this choice `this.state.preset`
  * create the steps `2 => 'Choose your preset'`, `3 => 'Provide pull secret'` and `4 => 'Review selection'`
    * when `preset === "podman" step 3 is disabled
    * otherwise; `preset === "openshift"` step 3 is necessary
  *  step 3 has `canJumpTo: stepIdReached >= 2 && this.state.preset === "openshift"`
  * `WizardContextConsumer`
    * when `activeStep.name === "Choose your preset")` then 
    *  `Button` has specific `onClick` logic to perform the navaigation
 
```jsx
<Button style={{width: 120}} variant="primary" type="submit" onClick={() =>
    {
        if(this.state.preset === "podman") {
            goToStepById(4);    // to skip to the review
        } else {
            onNext();   // perform normal behaviour to navigate to 'Provide pull secret'
        }
    }}>
    Next
</Button>
```

I had to resort to `activeStep.name` checks. The `id` was unreliable in use as I originally wanted to use substeps to peform this option step. As for the substep you can easily use a conditional (ternary) operator `? : ` to include a step an additional like:

```javascript
const steps = [
   { 
         id: 2, 
         name: 'Preset selection', 
         steps: [
            {
               name: 'Choose your preset',
               component: < />
            }, 
            {this.state.preset === "openshift") ? 
            {
               name: 'Provide pull secret',
               component: < />
            } : {}
         ]
   },
   {
      id: 3,
      name: 'Review selection',
...
}
```

this can not be skipped or navigated around easily. It will be removed (and therefore be very visible), but the issue is that the `id` values for a substep aren't returned?

This should be easier.

bartoval:
**What**: Closes #7302

Note: modify  the prop **canJumpTo** creates conflict with the progressive wizard, so I added a new prop called **isDisabled**

### DIFF
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -38,7 +38,7 @@ export interface WizardStep {
   hideCancelButton?: boolean;
   /** (Unused if footer is controlled) True to hide the Back button */
   hideBackButton?: boolean;
-  /** Force to disable the step in the navigation.*/
+  /** Flag to disable the step in the navigation */
   isDisabled?: boolean;
 }
 
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -38,6 +38,8 @@ export interface WizardStep {
   hideCancelButton?: boolean;
   /** (Unused if footer is controlled) True to hide the Back button */
   hideBackButton?: boolean;
+  /** Force to disable the step in the navigation.*/
+  isDisabled?: boolean;
 }
 
 export type WizardStepFunctionType = (
@@ -187,7 +189,15 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
       }
       return onClose();
     } else {
-      const newStep = currentStep + 1;
+      let newStep = currentStep;
+
+      for (let nextStep = currentStep; nextStep <= maxSteps; nextStep++) {
+        if (!flattenedSteps[nextStep].isDisabled) {
+          newStep = nextStep + 1;
+          break;
+        }
+      }
+
       this.setState({
         currentStep: newStep
       });
@@ -208,7 +218,15 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
         currentStep: adjustedStep
       });
     } else {
-      const newStep = currentStep - 1 <= 0 ? 0 : currentStep - 1;
+      let newStep = currentStep;
+
+      for (let prevStep = currentStep; prevStep >= 0; prevStep--) {
+        if (!flattenedSteps[prevStep - 2].isDisabled) {
+          newStep = prevStep - 1 <= 1 ? 1 : prevStep - 1;
+          break;
+        }
+      }
+
       this.setState({
         currentStep: newStep
       });
@@ -219,9 +237,14 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
   };
 
   private goToStep = (step: number): void => {
+    const flattenedSteps = this.getFlattenedSteps();
+
+    if (flattenedSteps[step].isDisabled) {
+      return;
+    }
+
     const { onGoToStep } = this.props;
     const { currentStep } = this.state;
-    const flattenedSteps = this.getFlattenedSteps();
     const maxSteps = flattenedSteps.length;
     if (step < 1) {
       step = 1;
@@ -413,7 +436,7 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
                         return;
                       }
                       navItemStep = this.getFlattenedStepsIndex(flattenedSteps, childStep.name);
-                      enabled = childStep.canJumpTo;
+                      enabled = childStep.canJumpTo && !childStep.isDisabled;
                       return (
                         <WizardNavItem
                           key={`child_${indexChild}`}
@@ -431,7 +454,8 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
               );
             }
             navItemStep = this.getFlattenedStepsIndex(flattenedSteps, step.name);
-            enabled = step.canJumpTo;
+            enabled = step.canJumpTo && !step.isDisabled;
+
             return (
               <WizardNavItem
                 {...step.stepNavItemProps}
--- a/packages/react-core/src/components/Wizard/Wizard.tsx
+++ b/packages/react-core/src/components/Wizard/Wizard.tsx
@@ -192,6 +192,10 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
       let newStep = currentStep;
 
       for (let nextStep = currentStep; nextStep <= maxSteps; nextStep++) {
+        if (!flattenedSteps[nextStep]) {
+          return;
+        }
+
         if (!flattenedSteps[nextStep].isDisabled) {
           newStep = nextStep + 1;
           break;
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/Wizard/examples/Wizard.md
```diff
@@ -161,7 +186,7 @@ class SimpleWizard extends React.Component {
         name: 'First step',
         steps: [
           { name: 'Substep A', component: <p>Substep A content</p> },
-          { name: 'Substep B', component: <p>Substep B content</p> }
+          { name: 'Substep B', component: <p>Substep B content</p>,  isDisabled: true }
```
wise-king-sullyman, line 189: I'm curious what the reason was for marking this substep as disabled? I'm not saying it's a bad thing, I'm just unsure what the reasoning is.

wise-king-sullyman (commented):
This looks really solid, I just have a couple of small questions. Neither are really blockers though.

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -174,7 +176,15 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
...
     } else {
-      const newStep = currentStep + 1;
+      let newStep = currentStep;
+
+      for (let nextStep = currentStep; nextStep <= maxSteps; nextStep++) {
+        if (!flattenedSteps[nextStep].isDisabled) {
+          newStep = nextStep + 1;
+          break;
+        }
+      }
```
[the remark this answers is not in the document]
bartoval: actually, I think the code will be less readable and there are many differences between the 2 functions:
indexes, conditions, and loop orientation are different...if we create a new function probably we will need to pass more parameters/flags

packages/react-core/src/components/Wizard/examples/Wizard.md
bartoval: I wanted to cover a basic example and a case where we use a substep.

Maybe I can change the new example I created with this one to show both cases

packages/react-core/src/components/Wizard/Wizard.tsx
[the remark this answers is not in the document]
wise-king-sullyman: You're right, after some more playing around with it I don't think there is a good abstraction here.

packages/react-core/src/components/Wizard/examples/Wizard.md
thatblindgeye: I think I would lean that way as well, but we also have other examples that showcase some functionality that might be considered "out of scope" of what the example is primarily showing. I'd be fine leaving it as-is for now, and at another point a description could be added in to state what the example is showcasing.

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -34,6 +34,8 @@ export interface WizardStep {
   hideCancelButton?: boolean;
   /** (Unused if footer is controlled) True to hide the Back button */
   hideBackButton?: boolean;
+  /** Force to disable the step in the navigation.*/
+  isDisabled?: boolean;
```
thatblindgeye, line 38: I think the main thing for me is whether we can just use the `canJumpTo` prop for this new logic instead of introducing a prop with the same use-case. What conflict did you notice when modifying `canJumpTo`?

bartoval: I added this comment on the PR: 
` modify the prop canJumpTo creates conflict with the progressive wizard, so I added a new prop called isDisabled.`

This is because canJumpTo is calculated following a condition.

packages/react-core/src/components/Wizard/__tests__/Wizard.test.tsx
```diff
@@ -205,3 +205,50 @@ describe('Wizard', () => {
     expect(main).toHaveAttribute('aria-labelledby', 'main-aria-labelledby');
   });
 });
+
+  test.only('wiz with disabled steps', () => {
```
wise-king-sullyman, line 209: Can you remove the `.only` usage here and on the other test you added?

packages/react-core/src/components/Wizard/__tests__/Wizard.test.tsx
```diff
@@ -205,3 +205,50 @@ describe('Wizard', () => {
...
+      <Wizard
+        navAriaLabel="nav aria-label"
+        description="This wizard step is disabled"
+        steps={steps}
+      />
+    );
+    const button1Disabled = screen.getByRole('button',{ name: "B" });
+    const button2Disabled = screen.getByRole('button',{ name: "E" });
+
+    expect(button1Disabled).toHaveAttribute('disabled');
```
wise-king-sullyman, line 229: Rather than using `toHaveAttribute('disabled')` I would like to see the `toBeDisabled()` matcher from jest-dom used for these tests.

packages/react-core/src/components/Wizard/__tests__/Wizard.test.tsx
```diff
@@ -205,3 +205,50 @@ describe('Wizard', () => {
...
+      <Wizard
+        navAriaLabel="nav aria-label"
+        description="This wizard step is disabled"
+        steps={steps}
+      />
+    );
+    const buttonDisabled = screen.getByRole('button',{ name: "Substep B" });
+
+    expect(buttonDisabled).toHaveAttribute('disabled');
+  });
```
wise-king-sullyman, line 254: I think it would be beneficial to also test `next` and `back` behavior to ensure that disabled steps are skipped over as we expect them to be. Maybe also a test which verifies that clicking on the name of a disabled step doesn't bring its content up.

wise-king-sullyman (changes requested):
This is looking good! I just have a couple of small changes that I'd like to see and a request for a few additional test cases.

packages/react-core/src/components/Wizard/__tests__/Wizard.test.tsx
```diff
@@ -222,3 +223,91 @@ describe('Wizard', () => {
...
+      <Wizard
+        navAriaLabel="nav aria-label"
+        description="This wizard step is disabled"
+        steps={steps}
+      />
+    );
+
+    const navItemButton = screen.getByRole('button',{ name: "C" })
+    
+    userEvent.click(navItemButton);
```
wise-king-sullyman, line 289: Shouldn't it be the disabled step that gets clicked here?

bartoval: oh, shrap! thanks 👍🏼

wise-king-sullyman (approved):
Thanks for making all those test updates, looks great!

I think your new example was a good add but it being removed isn't a blocker for me.

thatblindgeye (approved):

packages/react-core/src/components/Wizard/Wizard.tsx
```diff
@@ -38,6 +38,8 @@ export interface WizardStep {
   hideCancelButton?: boolean;
   /** (Unused if footer is controlled) True to hide the Back button */
   hideBackButton?: boolean;
+  /** Force to disable the step in the navigation.*/
```
tlabaj, line 41: nit "Flag to disable the step in the navigation."

tlabaj (commented):
Looks good.  I am just wandering if we want an example of the behaviour on the website @mcarrano @mmenestr.

kmcfaul (approved):

mmenestr:
That example looks good! I guess I'm just wondering what scenario this would occur on. Would steps be enabled or disabled based on the "response" in the previous step?

bartoval:
> That example looks good! I guess I'm just wondering what scenario this would occur on. Would steps be enabled or disabled based on the "response" in the previous step?

Yep, 
this guy has this kind of issue https://github.com/patternfly/patternfly-react/issues/7302

tlabaj (approved):
LGTM

mmenestr (approved):

PR #7579, opened 2022-06-16
chore(Modal): convert examples to TypeScript

### LINKED ISSUE #7574: Modal - convert examples to TypeScript
Modal

**Describe the feature**
The examples for the modal component need to be converted to TypeScript and functional components

thatblindgeye:
**What**: Closes #7574 

For the examples "custom header and footer" and "no header", I updated some verbiage about the `span` that acts as the description of the modal. For the "custom header/footer" example I removed the `aria-label` prop from the example; while it did announce both the aria-label and aria-labelledby in Safari, the use of aria-label in the example didn't seem needed.

I also updated the "warning alert" example name to just "Title icon", to flow into the "custom title icon" example.

**Additional issues**:


jenny-s51 (approved):
LGTM! All demos work as expected. Nice change to the `span` verbiage in the "custom header and footer" and "no header" examples! Update Title Icon demo looks good too 👍

nicolethoen (approved):

PR #7580, opened 2022-06-17
chore(Table): clean up a11y violations in table docs

### LINKED ISSUE #7578: fix a11y violations in table docs
Address legacy table docs, composable table docs, table demos
Many duplicate ids

nicolethoen:
**What**: Closes #7578 

Cleans up `id` props duplicated across demos/examples
Cleans up non-unique aria-labels on similar elements across demos/examples

### DIFF
--- a/packages/react-table/src/components/Table/base/types.tsx
+++ b/packages/react-table/src/components/Table/base/types.tsx
@@ -81,6 +81,8 @@ export interface TdExpandType {
   columnIndex?: number;
   /** On toggling the expansion */
   onToggle?: OnCollapse;
+  /** Id prefix for expandable rows **/
+  expandId?: string;
 }
 
 export interface TdCompoundExpandType {
@@ -88,6 +90,8 @@ export interface TdCompoundExpandType {
   isExpanded: boolean;
   /** Callback on toggling of the expansion */
   onToggle?: OnExpand;
+  /** Id prefix for expandable cells **/
+  expandId?: string;
 }
 
 export interface TdFavoritesType {
--- a/packages/react-table/src/components/Table/base/types.tsx
+++ b/packages/react-table/src/components/Table/base/types.tsx
@@ -92,6 +92,10 @@ export interface TdCompoundExpandType {
   onToggle?: OnExpand;
   /** Id prefix for expandable cells **/
   expandId?: string;
+  /** The row index */
+  rowIndex?: number;
+  /** The column index */
+  columnIndex?: number;
 }
 
 export interface TdFavoritesType {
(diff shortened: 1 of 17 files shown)


packages/react-table/src/components/TableComposable/Td.tsx
```diff
@@ -134,7 +134,8 @@ const TdBase: React.FunctionComponent<TdProps> = ({
           },
           column: {
             extraParams: {
-              onCollapse: expand?.onToggle
+              onCollapse: expand?.onToggle,
+              expandId: expand?.expandId
```
thatblindgeye, line 138: Should the `const compoundParams...` code that follows the lines above include a similar line addition as this? Right now the "composable table compound expandable" example doesn't have any ID set on the toggles.

packages/react-table/src/components/Table/base/types.tsx
```diff
@@ -81,13 +81,17 @@ export interface TdExpandType {
   columnIndex?: number;
   /** On toggling the expansion */
   onToggle?: OnCollapse;
+  /** Id prefix for expandable rows **/
+  expandId?: string;
```
thatblindgeye, line 85: Not a blocker: since it looks like this prop is mainly being used to set an ID on toggles, an alternative name could be `toggleId`.

thatblindgeye (commented):
Looks good! Didn't see any ID violations when running aXe on pages. Only had a couple of comments below.

packages/react-table/src/components/Table/base/types.tsx
nicolethoen: I borrowed this name from the `expandId` prop on `Table.tsx` which also sets this value on the legacy tables. There was not a way to set the equivalent value on the ComposableTable.

packages/react-table/src/components/TableComposable/Td.tsx
nicolethoen: Possibly - I will double check

wise-king-sullyman:
@nicolethoen I'm seeing axe reporting issues in some of the examples themselves still, such as empty table headers. Here's one example from the Composable: Selectable radio input example:

image

It does just say that it's a "needs review" level issue though, so if that isn't something we're concerned with this looks great.

nicolethoen:
hm... that's true - I'll open a design issue to consider how to handle those situations. Nice catch.

nicolethoen:
> Again though I'm not sure if these are _real_ issues or not 😅

I'll investigate - it may be hard to make a call on the 'needs review' flagged issues but upon further review, there may be a relatively clear way to improve it.

packages/react-table/src/components/TableComposable/Td.tsx
nicolethoen: Actually - upon double checking, I did add an `expandId` to the `TdCompoundExpand` type :) so both expand and compoundexpand should be good here :)

tlabaj (approved):

dlabrecq (approved):

wise-king-sullyman (approved):
Looks great!

thatblindgeye (approved):

Issue #7585, opened 2022-06-17
Charts: Add ResizeObserver example page


dlabrecq:
Want to create a charts page showing how to use the `ResizeObserver`.

Although we have many examples showing how to resize charts, it seems some folks are not aware they are available. Creating a dedicated page should help make that more clear, similar to the pages for legends, themes, and tooltips.

PR #7587, opened 2022-06-17
Charts: Add ResizeObserver examples

dlabrecq:
Created a charts page showing how to resize charts using a `ResizeObserver`.

Although we have many examples showing how to resize charts, it seems some folks are not aware they're available. Creating dedicated examples should help make that more clear, similar to the pages for legends, themes, and tooltips.

Example: https://patternfly-react-pr-7587.surge.sh/charts/resize-observer
Closes https://github.com/patternfly/patternfly-react/issues/7585


nicolethoen (commented):
Maybe above each of the examples in a couple sentences you could identify which specific props you are manipulating and functions you're creating in order to enable the resizable features you describe?

dlabrecq:
Updated the example descriptions

tlabaj (approved):

PR #7595, opened 2022-06-21
feat(Table): add nested sticky header support

### LINKED ISSUE #7478: Table - fix sticky, nested table header bottom border
Placeholder for core work in https://github.com/patternfly/patternfly/pull/4857 to fix the header border when you have a sticky, nested header.

kmcfaul:
**What**: Closes #7478

Adds `isBorderRow` to `Tr`.
Adds new example for nested and sticky header.

@mattnolting There's a small difference between core/react DOM in that I couldn't render an empty `Tr`, so I added a child `Th` that contains the colspan. At least this is the difference going off the hbs structure since the surge is down atm.

### DIFF
--- a/packages/react-table/src/components/TableComposable/examples/ComposableTableNestedStickyHeader.tsx
+++ b/packages/react-table/src/components/TableComposable/examples/ComposableTableNestedStickyHeader.tsx
@@ -153,7 +153,7 @@ export const ComposableTableNestedHeaders: React.FunctionComponent = () => {
               {columnNames.destination}
             </Th>
           </Tr>
-          <Tr isBorderRow>
+          <Tr isBorderRow aria-hidden="true">
             <Td colSpan={9}></Td>
           </Tr>
         </Thead>


mattnolting (commented):
> @mattnolting There's a small difference between core/react DOM in that I couldn't render an empty Tr, so I added a child Th that contains the colspan. At least this is the difference going off the hbs structure since the surge is down atm.

I think this looks good. Core has a `tr` with an empty `td` (rather than `th`) inside the row that spans all columns. I have the empty `td` spanning a set number of columns, but we could also set the `colspan` to something like `10000`, if that has a performance advantage.

mmenestr (approved):
visually LGTM!

packages/react-table/src/components/TableComposable/examples/ComposableTableNestedStickyHeader.tsx
```diff
@@ -0,0 +1,201 @@
...
+            </Th>
+            <Th isSubheader modifier="fitContent" sort={getSortParams(3)}>
+              {columnNames.source}
+            </Th>
+            <Th isSubheader modifier="fitContent" hasRightBorder sort={getSortParams(4)}>
+              {columnNames.destination}
+            </Th>
+          </Tr>
+          <Tr isBorderRow>
+            <Th colSpan={9}></Th>
```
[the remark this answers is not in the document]
kmcfaul: I can, but don't trs in a `thead` typically contain `th` instead of `td`?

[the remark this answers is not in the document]
mattnolting: @kmcfaul, yes, if they contain textual table content and related `tbody td` content. If you look at empty `thead` cells, they're all `td`s as we want assistive technologies to ignore them.

packages/react-table/src/components/TableComposable/examples/ComposableTableNestedStickyHeader.tsx
```diff
@@ -0,0 +1,201 @@
...
+            </Th>
+            <Th isSubheader modifier="fitContent" sort={getSortParams(3)}>
+              {columnNames.source}
+            </Th>
+            <Th isSubheader modifier="fitContent" hasRightBorder sort={getSortParams(4)}>
+              {columnNames.destination}
+            </Th>
+          </Tr>
+          <Tr isBorderRow>
+            <Td colSpan={9}></Td>
```
mattnolting, line 157: 👍

packages/react-table/src/components/TableComposable/examples/ComposableTableNestedStickyHeader.tsx
```diff
@@ -0,0 +1,201 @@
...
+              {columnNames.datetime}
+            </Th>
+            <Th isSubheader modifier="fitContent" sort={getSortParams(3)}>
+              {columnNames.source}
+            </Th>
+            <Th isSubheader modifier="fitContent" hasRightBorder sort={getSortParams(4)}>
+              {columnNames.destination}
+            </Th>
+          </Tr>
+          <Tr isBorderRow>
```
mattnolting, line 156: Can you also add `aria-hidden="true"` here? Looks like doing so removes the row from total table row count.

mattnolting: I'll create a followup up issue for Core.

tlabaj (approved):

PR #7598, opened 2022-06-21
Callback to know when we changed to a particular step

### LINKED ISSUE #7702: Wizard - Make it easier to track the current step
Wizard component

**Describe the feature**
Tracking the current step is complicated, this is an internal state to the component and its value is passed to multiple functions.

If one wants to track the state of the current step we need to implement multiple callbacks (`onNext`, `onBack`, `onGoToStep`). Moreover, if we use a custom footer, we would have to implement that logic on the footer, as these are not used.

Any other modification that might change the current step while using other callback might introduce bugs if using above method.
An example of this could be the sample in the [documentation](https://patternfly-react.surge.sh/components/wizard/#remember-last-step)  that does not implement the `onGoToStep` and the current step is out of sync with the Wizard.

There should be a callback for advertising the change of step.

No visual changes

**Any other information?**

josejulio:
TLDR; I'm consolidating the way to know the current step to a single callback that fires everytime we change to a new step.

The problem I'm trying to solve here is to prevent from having a separate source of truth in the user of a Wizard. 

As an user, we have to keep track of the current step for external computing. i.e. know if we want to transverse from step `a` to step `b` or what to validate. 

To do this, the current recomended way (as show in this [example](https://patternfly-react.surge.sh/components/wizard)) is to create a local state variable and pass some callbacks to the Wizard, then "spy" for the values that are sent in there to sync our current step value.

This has some drawbacks:
 - We have to keep a separate source of truth for the current step - even the example is not complete - if we click on the steps on the left menu it will get out of sync - we still have to implement other callback (`onGoToStep`).
 - If we implement the Wizard footer, the `onNext`, `onBack` and `onSave` are not used anymore, and we need to mimic this behavior with the WizardFooter implementation.
 - Anything else that is added that modifies the current step will break the implementation (e.g. ability to jump programatically to a specific step?)

I didn't create an issue for this, the change was relatively straighforward and I guess easier to show in the test.

Tried some things (can be seen in the first commit), but in the end I'm using a callback `onCurrentStepChanged` so we don't have to go fishing which callback to use. The function receives the step we transitioned to.

Fixes #7702


josejulio:
@tlabaj Could you take a look? 
 Should I create an issue?

tlabaj:
@josejulio yes, please create an issue.

josejulio:
Created the [issue](https://github.com/patternfly/patternfly-react/issues/7702) and fixed conflicts

jenny-s51 (approved):
Thank you for your contribution @josejulio ! 

Based on the prior state of the demo, definitely makes sense to have a separate prop/callback here to keep track of the current step. Renamed demo looks good and works as expected when navigating using the footer buttons, or when clicking the sidebar. I'm seeing the correct id for each step as well; thanks for adding a test.

LGTM

nicolethoen (approved):

PR #7611, opened 2022-06-24
feat(CalendarMonth): add onMonthChange callback

### LINKED ISSUE #6654: CalendarMonth RFEs

nicolethoen:
**What**: Closes #6654 

Adds onMonthChange callback and updates the default calendar to make dates selectable and log month change events.

### DIFF
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
@@ -1,4 +1,0 @@
-import React from 'react';
-import { CalendarMonth } from '@patternfly/react-core';
-
-export const CalendarMonthDefault: React.FunctionComponent = () => <CalendarMonth date={new Date()} />;
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
@@ -1,4 +1,12 @@
 import React from 'react';
 import { CalendarMonth } from '@patternfly/react-core';
 
-export const CalendarMonthDefault: React.FunctionComponent = () => <CalendarMonth date={new Date()} />;
+export const CalendarMonthDefault: React.FunctionComponent = () => {
+  const [date, setDate] = React.useState(null);
+  const onMonthChange = (newDate: Date) => {
+    // eslint-disable-next-line no-console
+    console.log(newDate);
+  };
+
+  return <CalendarMonth date={date} onChange={setDate} onMonthChange={onMonthChange} />;
+};
--- a/packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
+++ b/packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
@@ -1,12 +1,4 @@
 import React from 'react';
 import { CalendarMonth } from '@patternfly/react-core';
 
-export const CalendarMonthDefault: React.FunctionComponent = () => {
-  const [date, setDate] = React.useState(null);
-  const onMonthChange = (newDate: Date) => {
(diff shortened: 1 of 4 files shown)


mcarrano:
@nicolethoen is there any specific I should be looking at here from a UX perspective?

andyyvo:
Looks good! I noticed that the console log updates the date with the `onMonthChange` callback. However, the calendar UI itself doesn't update based on that new date when we change months
Screen Shot 2022-06-24 at 1 37 40 PM

nicolethoen:
@andyyvo the `onCalendarMonth` callback returns the new month and/or new date when the user interacts with the controls at the top of the calendar month. It is not (usually) going to be used to set the selected date. When the component loads, if a starting date is not provided, it opens the calendar month to the current month and today's date is a light grey. I clarified the console statement in the example onCalendarMonth to try and clarify.

@mcarrano I guess I tagged you so you'd know I addressed the request. There are no new UX features in this example.

andyyvo (approved):
LGTM!!

packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
```diff
@@ -1,4 +1,12 @@
 import React from 'react';
 import { CalendarMonth } from '@patternfly/react-core';
 
-export const CalendarMonthDefault: React.FunctionComponent = () => <CalendarMonth date={new Date()} />;
+export const CalendarMonthDefault: React.FunctionComponent = () => {
```
thatblindgeye, line 4: Re: this comment from Titani back when I made an update to the component: https://github.com/patternfly/patternfly-react/pull/7082#issuecomment-1079408467

Do we still want the "Default" example to be static, and then just update the "Selectable date" example to showcase this new callback?

thatblindgeye (commented):
Very nice! Just had a comment below regarding the example update

tlabaj (changes requested):
I agree with Eric.  I don't think we want to show selection here.  we have a separate example for that.

packages/react-core/src/components/CalendarMonth/examples/CalendarMonthDefault.tsx
nicolethoen: I think that the default example starting with no date selected, but still allowing a date to be selected is more in line with how most of our component examples work - even just the basic/default examples? I dont think we have many non-interactive react examples for components which are built for interaction.

nicolethoen: @tlabaj ^

tlabaj (changes requested):
This example is still not quite working as it was before.  When you click on the new date it appears selected
image

Here is how it worked before:
image

nicolethoen:
The image of what it showed before was just showing the focus state not the selected state. 
The original default calendar month example lets you change the month/year but not select a date. I didn't feel like that's usually what we would demonstrate as an example of intended behavior.
I'll change it back and update the selectable example to be the one which demonstrates the month/year change handler.

mcarrano:
I don't know. I've always found this a bit confusing, but am good with whatever you think makes sense.

tlabaj:
So maybe the answer here is to remove one of the examples.  It does not make sense to have 2 that technically do the same thing.
cc @mcarrano

thatblindgeye (approved):

mcarrano (approved):
Looks good to me. Thanks @nicolethoen !

kmcfaul (approved):

tlabaj (approved):

PR #7619, opened 2022-06-27
chore: fix various a11y violations in examples (3)

### LINKED ISSUE #7610: chore: address a11y violations from various component docs (3)
-  Notification drawer examples
-  Input group examples
-  Dual list selector examples
-  Data list examples
-  File upload examples
-  Card examples
-  Context selector examples
-  Code block examples
-  Accordion examples
-  Action list examples

nicolethoen:
**What**: Closes #7610 

- Notification drawer examples
- Input group examples
- Dual list selector examples
- Data list examples
- File upload examples
- Card examples
- Context selector examples
- Code block examples
- Accordion examples
- Action list examples

few notes:

- Context selector needed a more reliable way to generate unique ids. So I updated how they were randomly generated and also passed an `id` to the unit test cases so they would be the same across snapshots. Also, the items inside context selector menus did not have the `role='menuitem'` so I added it.
- In the Dual list selector, the `aria-multiselectable`, `aria-labelledby`, `aria-activedescendant`, and `role=tree|listbox` should only be applied when they have the requisite valid children with valid roles. In the case of empty lists, they should not have those roles (according to the axe tests).
- The DualListSelectorTree example had the same name as the DualListSelectorTree component, which was causing a conflict when generating the prop descriptions tables, so i renamed the example to DualListSelectorTreeExample.

### DIFF
--- a/packages/react-core/src/components/DualListSelector/DualListSelectorControlsWrapper.tsx
+++ b/packages/react-core/src/components/DualListSelector/DualListSelectorControlsWrapper.tsx
@@ -76,7 +76,7 @@ DualListSelectorControlsWrapperBase.displayName = 'DualListSelectorControlsWrapp
 
 export const DualListSelectorControlsWrapper = React.forwardRef(
   (props: DualListSelectorControlsWrapperProps, ref: React.Ref<HTMLDivElement>) => (
-    <DualListSelectorControlsWrapperBase innerRef={ref as React.MutableRefObject<any>} {...props} role="region" />
+    <DualListSelectorControlsWrapperBase innerRef={ref as React.MutableRefObject<any>} {...props} role="group" />
   )
 );
 
--- a/packages/react-core/src/components/DualListSelector/DualListSelectorControlsWrapper.tsx
+++ b/packages/react-core/src/components/DualListSelector/DualListSelectorControlsWrapper.tsx
@@ -76,7 +76,7 @@ DualListSelectorControlsWrapperBase.displayName = 'DualListSelectorControlsWrapp
 
 export const DualListSelectorControlsWrapper = React.forwardRef(
   (props: DualListSelectorControlsWrapperProps, ref: React.Ref<HTMLDivElement>) => (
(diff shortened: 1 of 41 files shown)


packages/react-core/src/components/DualListSelector/DualListSelectorControlsWrapper.tsx
```diff
@@ -76,7 +76,7 @@ DualListSelectorControlsWrapperBase.displayName = 'DualListSelectorControlsWrapp
 
 export const DualListSelectorControlsWrapper = React.forwardRef(
   (props: DualListSelectorControlsWrapperProps, ref: React.Ref<HTMLDivElement>) => (
-    <DualListSelectorControlsWrapperBase innerRef={ref as React.MutableRefObject<any>} {...props} />
+    <DualListSelectorControlsWrapperBase innerRef={ref as React.MutableRefObject<any>} {...props} role="region" />
```
thatblindgeye, line 79: For the role, would "region" or "group" be better in this context? Currently the selector controls appear in VO's rotor under "landmarks", and wasn't sure if that may have been part of the intent or not.

thatblindgeye (commented):
Looks good! 🎉  Only thing I noticed is on the Dual List Selector examples is aXe is catching the error "Ensures landmarks are unique" due to the dual list selector controls with `role="region"` and the same `aria-label`. Otherwise I only had a question below.

packages/react-core/src/components/DualListSelector/DualListSelectorControlsWrapper.tsx
nicolethoen: ooo good thought

thatblindgeye (commented):
Looks like dual list selector snapshots just need to be updated, but otherwise looks good to me!

thatblindgeye (approved):

dlabrecq (approved):

jessiehuff (approved):
LGTM! :)
Testing this does remind me that Dual List selector with tree items isn't accessible by VO though. Do we already have an issue open for that? I thought we did, but I can't find it at the moment.

tlabaj (approved):

PR #7621, opened 2022-06-28
fix(Search Input): use popper to control the popup in advanced search

### LINKED ISSUE #7592: Bug - SearchInput - breaking change in advanced filter's panel-based popover; positioning is no longer absolute
When attempting to upgrade to the latest release of PatternFly, we spotted an issue due to failing tests. It appears to be related to this PR, as it changes the previous widgets used from HTML to the PF Panel component. Previously, it used the `pf-c-search-input__menu` class, but now it uses these classes: `pf-c-panel pf-m-raised pf-c-search-input__menu`.

However, pf-c-panel makes the popover `position: relative` now, instead of `position: absolute` and the width of the popover is now `auto` instead of being 100% wide.

As a result of the changes, the panel popover for the advanced filter still reserves the space in the toolbar, causing a layout shift when it appears, which looks like this:

relative whoops

(You can see the panel below the toolbar still... but as it's relative positioning instead of absolute, it now affects the flow of the content in the toolbar.)

_Originally posted by @garrett in https://github.com/patternfly/patternfly-react/issues/7414#issuecomment-1160546704_

bartoval:
**What**: Closes #7592

### DIFF
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import { css } from '@patternfly/react-styles';
-import styles from '@patternfly/react-styles/css/components/SearchInput/search-input';
 import { Button, ButtonVariant } from '../Button';
 import { Badge } from '../Badge';
 import AngleDownIcon from '@patternfly/react-icons/dist/esm/icons/angle-down-icon';
@@ -12,6 +11,7 @@ import ArrowRightIcon from '@patternfly/react-icons/dist/esm/icons/arrow-right-i
 import { AdvancedSearchMenu } from './AdvancedSearchMenu';
 import { TextInputGroup, TextInputGroupMain, TextInputGroupUtilities } from '../TextInputGroup';
 import { InputGroup } from '../InputGroup';
+import { Popper } from '../../helpers';
 
 export interface SearchAttribute {
   /** The search attribute's value to be provided in the search input's query string.
@@ -277,11 +277,9 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
 
   if (!!onSearch || attributes.length > 0 || !!onToggleAdvancedSearch) {
     if (attributes.length > 0) {
-      return (
-        <div className={css(className, styles.searchInput)} ref={searchInputRef} {...props}>
-          {buildSearchTextInputGroupWithExtraButtons()}
+      const advancedSearch = (
+        <span>
           <AdvancedSearchMenu
-            className={styles.searchInputMenu}
             value={value}
             parentRef={searchInputRef}
             parentInputRef={searchInputInputRef}
@@ -297,6 +295,18 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
             advancedSearchDelimiter={advancedSearchDelimiter}
             getAttrValueMap={getAttrValueMap}
             isSearchMenuOpen={isSearchMenuOpen}
+          />
+        </span>
+      );
+
+      return (
+        <div className={css(className)} ref={searchInputRef} {...props}>
+          <Popper
+            trigger={buildSearchTextInputGroupWithExtraButtons({ id: 'custom-advanced-search' })}
+            popper={advancedSearch}
+            isVisible={isSearchMenuOpen}
+            enableFlip={false}
+            appendTo={() => document.querySelector('#custom-advanced-search')}
           />
         </div>
       );
--- a/packages/react-core/src/components/SearchInput/SearchInput.tsx
+++ b/packages/react-core/src/components/SearchInput/SearchInput.tsx
@@ -83,6 +83,13 @@ export interface SearchInputProps extends Omit<React.HTMLProps<HTMLDivElement>, 
   /** Delimiter in the query string for pairing attributes with search values.
    * Required whenever attributes are passed as props */
   advancedSearchDelimiter?: string;
+  /** The container to append the menu to.
+   * If your menu is being cut off you can append it to an element higher up the DOM tree.
+   * Some examples:
+   * appendTo={() => document.body}
+   * appendTo={document.getElementById('target')}
+   */
+  appendTo?: HTMLElement | (() => HTMLElement) | 'inline';
 }
 
 const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
@@ -112,6 +119,7 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
   nextNavigationButtonAriaLabel = 'Next',
   submitSearchButtonLabel = 'Search',
   isDisabled = false,
+  appendTo,
   ...props
 }: SearchInputProps) => {
   const [isSearchMenuOpen, setIsSearchMenuOpen] = React.useState(false);
@@ -277,7 +285,7 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
 
   if (!!onSearch || attributes.length > 0 || !!onToggleAdvancedSearch) {
     if (attributes.length > 0) {
-      const advancedSearch = (
+      const AdvancedSearch = (
         <span>
           <AdvancedSearchMenu
             value={value}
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -277,11 +277,9 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
 
   if (!!onSearch || attributes.length > 0 || !!onToggleAdvancedSearch) {
     if (attributes.length > 0) {
-      return (
-        <div className={css(className, styles.searchInput)} ref={searchInputRef} {...props}>
-          {buildSearchTextInputGroupWithExtraButtons()}
+      const advancedSearch = (
+        <span>
```
bartoval, line 281: we need a native HTML markup here, otherwise, the popper is not going to calculate the position and set the item absolute

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -298,6 +296,18 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
...
+      );
+
+      return (
+        <div className={css(className)} ref={searchInputRef} {...props}>
+          <Popper
+            trigger={buildSearchTextInputGroupWithExtraButtons({ id: 'custom-advanced-search' })}
+            popper={advancedSearch}
+            isVisible={isSearchMenuOpen}
+            enableFlip={false}
+            appendTo={() => document.querySelector('#custom-advanced-search')}
```
nicolethoen, line 309: We cannot hardcode this value. It should be something that is configurable the prop could be the same or inspired by the appendTo prop from TimePicker.tsx
```
/** The container to append the advanced search form to. Defaults to 'parent'.
   * If your form is being cut off you can append it to an element higher up the DOM tree.
   * Some examples:
   * appendTo="parent"
   * appendTo={() => document.body}
   * appendTo={document.getElementById('target')}
   */
  appendTo?: HTMLElement | (() => HTMLElement) | 'inline' | 'parent';
```

nicolethoen (changes requested):

packages/react-core/src/components/SearchInput/SearchInput.tsx
bartoval: I am not sure about that. 
We want to append this menu to 'SearchTextInputGroup' (we set the value 'custom-advanced-search' as id) every time and not to an external DOM element. 

Maybe I am missing something.
Could you please provide an example of where we need to configure this element outside the component?

nicolethoen: It's possible that someone will put an advanced search into a Wizard modal for example and that would get clipped or formatted incorrectly relative to the wizard or modal. Those tend to be some the edge cases we are allowing for. 

Additionally, if someone wants an inline advanced search, where opening the modal does (for whatever reason) push down the contents of the page so the advanced search is not overlaying the results, that should be possible using the 'inline' value of 'appendTo'

bartoval: Got it, thanks! 👍🏼 
So I think we can:
- keep the current behavior as default (without passing the prop appendTo). (popper)
- provide the inline option to push down the content ( no popper)
- add the possibility to pass an external DOM element (popper)

nicolethoen (changes requested):
sorry, I have one request to update the prop name

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -277,11 +285,9 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
 
   if (!!onSearch || attributes.length > 0 || !!onToggleAdvancedSearch) {
     if (attributes.length > 0) {
-      return (
-        <div className={css(className, styles.searchInput)} ref={searchInputRef} {...props}>
-          {buildSearchTextInputGroupWithExtraButtons()}
+      const AdvancedSearch = (
+        <span>
           <AdvancedSearchMenu
-            className={styles.searchInputMenu}
```
wise-king-sullyman, line 284: Should this class be getting removed? Based on the screenshots provided in the issue I believe that this class was present before the bug was introduced.

bartoval: yes,
we changed the implementation using the popper and we don't need to these class to set the right style/position

wise-king-sullyman: Got it, thanks!

packages/react-core/src/components/SearchInput/__tests__/SearchInput.test.tsx
```diff
@@ -102,7 +103,42 @@ describe('SearchInput', () => {
 
     userEvent.click(screen.getByRole('button', { name: 'Search' }));
 
+    userEvent.click(screen.getByRole('button', { name: 'Open advanced search' }));
+    expect(screen.getByTestId('test-id').querySelector('.pf-c-panel')).toBeInTheDocument()
```
wise-king-sullyman, line 107: Nit: rather than using `querySelector` + `toBeInTheDocument` to determine where the search menu is being placed in the DOM can you use an RTL query to select the form (or something in it) and maybe the `toContainElement` matcher?

Reason being that I believe this would better replicate the actual user experience with the component.

wise-king-sullyman (changes requested):
Just a couple nits!

packages/react-core/src/components/SearchInput/__tests__/SearchInput.test.tsx
bartoval: Good idea. Thanks 👍🏼

kmcfaul (approved):

wise-king-sullyman (approved):
Looks great! 🚀

nicolethoen (approved):

packages/react-core/src/components/SearchInput/SearchInput.tsx
```diff
@@ -277,11 +285,9 @@ const SearchInputBase: React.FunctionComponent<SearchInputProps> = ({
 
   if (!!onSearch || attributes.length > 0 || !!onToggleAdvancedSearch) {
     if (attributes.length > 0) {
-      return (
-        <div className={css(className, styles.searchInput)} ref={searchInputRef} {...props}>
-          {buildSearchTextInputGroupWithExtraButtons()}
+      const AdvancedSearch = (
+        <span>
```
mcoker, line 289: @bartoval Apologies for the super late comment, I was following up on something I made a note about a while ago. FWIW with the original issue, we should always try and avoid applying multiple component styles to an element (ie, `.pf-c-panel.pf-c-search-intput__menu`). 

This element should be a `<div>` though, a `<span>` is an inline element and its permitted content is [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content). So technically it's invalid HTML to put non-phrasing content in a `<span>`, though it works just fine. Created an issue for that here - https://github.com/patternfly/patternfly-react/issues/7796

Also it's worth noting that `appendTo="inline"` just drops the panel in the layout and pushes everything below it down (it doesn't display on top of the stuff below it). A few questions around that:
* Should `appendTo` be `menuAppendTo` like the other components that have a menu in them that popper can position?
* Is that the intended behavior of `appendTo="inline"` or should that display on top of the stuff below it like it used to - it just won't use popper. Not sure how we'd do that if we want to continue to use the panel component there, but that would be inline with the other menus that support `menuAppendTo`. 
* If we are going to leave it so that it pushes the stuff below the panel down, should we remove the "raised" modifier from the panel since the shadow implies depth/overlay?

bartoval: @mcoker thanks for your comments! They are very interesting.

- I will change the markup span in div

Regarding your questions:
- there is comment above about that and using menuAppendTo is a little misleading since the Menu is a component we have and we are not using a Menu in this case.

- Yes this behaviour is predicted like the TimePicker (@nicolethoen )

-  In my opinion the shadow is not associated to an overlay. Ie: A card have a shadow but it doesn't implies overlay.  However this is just a personal tought.

mcoker: @bartoval thanks!! 

> there is comment above about that and using menuAppendTo is a little misleading since the Menu is a component we have and we are not using a Menu in this case.

👍 got it. I agree, that could be confusing. I would probably argue that menu is more of a concept there to refer to the thing that pops open in the component, and that could be used with whatever component pops up. It seems like it might be more predictable/consistent that way. As in if we stopped using the "menu" component and used some other component for the thing that pops up, `menuAppendTo` would still be fine and wouldn't necessarily need to change IMO since we shouldn't be tying the prop name to the component used in the first place and we're using menu as a concept. Also I'd probably say that `[menu/thing]AppendTo` would be used on the component that has a thing that pops up, and `appendTo` would be more appropriate on the thing that pops up itself. Just my $0.02 😁

PR #7626, opened 2022-06-28
chore(Hint): react testing library test revamps

### LINKED ISSUE #7616: Hint test revamp
Updating `Hint` component unit tests to meet new RTL standards

andyyvo:
**What**: Closes #7616 . Added new tests for the Hint component (and subcomponents).

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/Hint/__tests__/Hint.test.tsx
+++ b/packages/react-core/src/components/Hint/__tests__/Hint.test.tsx
@@ -34,21 +34,21 @@ describe('Hint', () => {
     expect(hint).toHaveClass('custom-classname');
   });
 
-  // it('does not render actions options when not passed', () => {
-  //   render(<Hint>Test</Hint>);
+  it('does not render actions options when not passed', () => {
+    render(<Hint>Test</Hint>);
 
-  //   const hint = screen.getByText("Test");
+    const actions = screen.queryByText('actions');
 
-  //   expect(hint.firstChild).not.toHaveClass('pf-c-hint__actions');
-  // });
+    expect(actions).not.toBeInTheDocument();
+  });
 
-  // it('renders actions options', () => {
-  //   render(<Hint actions="actions">Test</Hint>);
+  it('renders actions options', () => {
+    render(<Hint actions="actions">Test</Hint>);
 
-  //   const hint = screen.getByText("Test");
+    const actions = screen.getByText("actions");
 
-  //   expect(hint.firstChild).toHaveClass('pf-c-hint__actions');
-  // });
+    expect(actions).toBeVisible();
+  });
 
   it('renders hint with body, title, and footer', () => {
     const { asFragment } = render(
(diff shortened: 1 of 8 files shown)


packages/react-core/src/components/Hint/__tests__/Hint.test.tsx
```diff
@@ -1,17 +1,63 @@
...
+  it('renders hint with body, title, and footer', () => {
+    const { asFragment } = render(
+      <Hint>
+        <HintTitle>Title</HintTitle>
+        <HintBody>Body</HintBody>
+        <HintFooter>Footer</HintFooter>
+      </Hint>
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
```
wise-king-sullyman, line 62: Can you add a test to check for the `pf-c-hint__actions` class being applied to the actions?

packages/react-core/src/components/Hint/__tests__/HintBody.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    expect(body).toHaveClass('custom-classname');
+  });
+
+  it('matches snapshot', () => {
+    const { asFragment } = render(<HintBody>{<button>Test</button>}</HintBody>);
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
```
wise-king-sullyman, line 41: Can you add a newline at the end of the new files which are missing it?

packages/react-core/src/components/Hint/__tests__/Hint.test.tsx
wise-king-sullyman: Also I would like to see a test for each of these components which verifies that other props which aren't explicitly declared are spread properly.

wise-king-sullyman (changes requested):
This is looking great! I just have a handful of small requests.

packages/react-core/src/components/Hint/__tests__/Hint.test.tsx
wise-king-sullyman: It doesn't seem like it's causing problems, but I did notice that in [the core demo without an actions element ](https://www.patternfly.org/v4/components/hint/html#default-with-no-title)the whole actions div actually isn't rendered.

Regardless for now since (currently at least) the actions div always renders there is no need to check if it applies when an `actions` is passed IMO. We just need to test that what *is* passed to `actions` is placed in that div as we expect it to be.

packages/react-core/src/components/Hint/__tests__/Hint.test.tsx
```diff
@@ -1,17 +1,70 @@
...
+
+test('renders with inherited element props spread to the component', () => {
+  render(
+    <Hint aria-label="labelling-id">Test</Hint>
+  );
+
+  expect(screen.getByText('Test')).toHaveAccessibleName('labelling-id');
+});
+
+test('renders hint with body, title, and footer', () => {
```
wise-king-sullyman, line 65: Small nit: I think this test name should be updated to reflect that the body, title, and footer elements are no longer being included in the snapshot?

wise-king-sullyman (commented):
I have one tiny nit about a test name, but other than that this is looking really solid!

wise-king-sullyman (approved):
🚀 🚀 🚀 awesome work on this!

nicolethoen (approved):

PR #7636, opened 2022-06-29
chore(NotificationBadge):convert examples to typescript

### LINKED ISSUE #7630: Notification Badge

janwright73:
**What**: Closes #7630

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/NotificationBadge/examples/NotificationBadgeWithCount.tsx
+++ b/packages/react-core/src/components/NotificationBadge/examples/NotificationBadgeWithCount.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { NotificationBadge, NotificationBadgeVariant } from '@patternfly/react-core';
+
+export const NotificationBadgeWithCount: React.FunctionComponent = () => {
+  const [firstVariant, setFirstVariant] = React.useState(NotificationBadgeVariant.unread);
+  const [firstVariantCount, setFirstVariantCount] = React.useState(30);
+  const [secondVariant, setSecondVariant] = React.useState(NotificationBadgeVariant.attention);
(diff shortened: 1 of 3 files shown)


packages/react-core/src/components/NotificationBadge/examples/NotificationBadgeWithCount.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const NotificationBadgeWithCount: React.FunctionComponent = () => {
+  const [firstVariant, setFirstVariant] = React.useState(NotificationBadgeVariant.unread);
+  const [firstVariantCount, setFirstVariantCount] = React.useState(30);
+  const [secondVariant, setSecondVariant] = React.useState(NotificationBadgeVariant.attention);
+  const [secondVariantCount, setSecondVariantCount] = React.useState(30);
+
+  const onFirstClick = (value: NotificationBadgeVariant) => {
+    setFirstVariant(value);
+    setFirstVariantCount(90);
```
gitdallas, line 12: Why 90? Just seems weird to see it go from 30 to 90 when marking read. I'd think if we are going to change it, perhaps 0 would make more sense.

janwright73: The existing example set the exact same value (30), so I was just doing a different number.  More notifications could have come in??

janwright73:
"Count (optional): When used, the count field will indicate the number of notifications in the notification drawer."  I am totally fine with the count decreasing instead of increasing.  How about going from 30 to 10?

nicolethoen (approved):

gitdallas (approved):

PR #7645, opened 2022-06-30
chore(Accordion): update tests to new standards

### LINKED ISSUE #7605: Accordion test revamp

wise-king-sullyman:
**What**: Closes #7605

**Additional issues**:


nicolethoen (approved):
LGTM!
And you may not love the nodeName property, but I imagine we'll be using it often whenever we have a configurable element - like heading levels and things like that.

PR #7651, opened 2022-07-05
Bug(ContextSelector) - Updated Param Type of toggleText in ContentSelector

### LINKED ISSUE #7457: Bug - ContextSelector - toggleText prop should allow for JSX
Right now the type is set to string, there is a use case to pass in text with an icon, so the type should be set to something like `ReactNode` instead.
https://github.com/patternfly/patternfly-react/blob/5bbf5776f7146b03c36382aa0e8de32781ded672/packages/react-core/src/components/ContextSelector/ContextSelector.tsx#L37

This issue came up in https://github.com/theforeman/foreman/pull/9222

adithyaakrishna:
Signed-off-by: Adithya Krishna <adikrish@redhat.com>

**What**: Closes #7457

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -40,7 +40,7 @@ export interface ContextSelectorProps extends Omit<ToggleMenuBaseProps, 'menuApp
   /** Component or string that will be used in the Context Selector.
    * This prop is placed in a button, and cannot contain any interactive elements
    */
-  toggleText?: React.ReactNode | string;
+  toggleText?: React.ReactNode;
   /** Aria-label for the Context Selector Search Button */
   searchButtonAriaLabel?: string;
   /** Value in the Search field */
@@ -77,7 +77,7 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
     onToggle: () => undefined as any,
     onSelect: () => undefined as any,
     screenReaderLabel: '',
-    toggleText: null as React.ReactNode,
+    toggleText: '',
     searchButtonAriaLabel: 'Search menu items',
     searchInputValue: '',
     onSearchInputChange: () => undefined as any,
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -37,8 +37,8 @@ export interface ContextSelectorProps extends Omit<ToggleMenuBaseProps, 'menuApp
   isFullHeight?: boolean;
   /** Labels the Context Selector for Screen Readers */
   screenReaderLabel?: string;
-  /** Text that appears in the Context Selector Toggle */
-  toggleText?: string;
+  /** Component that will be used in the Context Selector Toggle */
+  toggleText?: React.ReactNode;
   /** Aria-label for the Context Selector Search Button */
   searchButtonAriaLabel?: string;
   /** Value in the Search field */
@@ -75,7 +75,7 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
     onToggle: () => undefined as any,
     onSelect: () => undefined as any,
     screenReaderLabel: '',
-    toggleText: '',
+    toggleText: null as React.ReactNode,
     searchButtonAriaLabel: 'Search menu items',
     searchInputValue: '',
     onSearchInputChange: () => undefined as any,
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -38,7 +38,7 @@ export interface ContextSelectorProps extends Omit<ToggleMenuBaseProps, 'menuApp
   /** Labels the Context Selector for Screen Readers */
   screenReaderLabel?: string;
   /** Component or string that will be used in the Context Selector.
-   * This prop is placed in a button, and cannot contain any interactive elements 
+   * This prop is placed in a button, and cannot contain any interactive elements
    */
   toggleText?: React.ReactNode | string;
   /** Aria-label for the Context Selector Search Button */
--- a/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
+++ b/packages/react-core/src/components/ContextSelector/ContextSelector.tsx
@@ -37,8 +37,10 @@ export interface ContextSelectorProps extends Omit<ToggleMenuBaseProps, 'menuApp
   isFullHeight?: boolean;
   /** Labels the Context Selector for Screen Readers */
   screenReaderLabel?: string;
-  /** Component that will be used in the Context Selector Toggle */
-  toggleText?: React.ReactNode;
+  /** Component or string that will be used in the Context Selector.
+   * This prop is placed in a button, and cannot contain any interactive elements 
+   */
+  toggleText?: React.ReactNode | string;
   /** Aria-label for the Context Selector Search Button */
   searchButtonAriaLabel?: string;
   /** Value in the Search field */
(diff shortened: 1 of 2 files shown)


nicolethoen:
@adithyaakrishna 
These proposed changes look good!
I have a few small notes and answers to your questions :)

- in the issue, you asked about changing the prop name, and that is something we are actually not supposed to do because that constitutes an API breaking change. So leaving it the way you have and simply updating the prop description in the comment above the prop is perfect!
- Could you also add an additional note in the prop description so that consumers are aware that the toggletext is placed in a button. Therefore, they should not include interactable elements in their toggleText to keep the html valid. Only non-interactive ReactNodes are valid.

adithyaakrishna:
Hi @nicolethoen 
Apologies for the delay regarding this PR :(

I have added an additional note/desc for the prop as well, could you please let me know if it looks good?

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -10,8 +10,9 @@ export interface ContextSelectorToggleProps {
   id: string;
   /** Classes applied to root element of toggle */
   className?: string;
-  /** Text that appears in the Context Selector Toggle */
-  toggleText?: string;
+  /** Component or string that will be used in the Context Selector Toggle */
+  /** This prop is placed in a button */
```
nicolethoen, line 14: This is great! 
Let's also be just a bit more explicit - maybe something like `This prop is placed in a button, and cannot contain any interactive elements`

And maybe both prop descriptions for `toggleText` should say the same thing. Could you update the description in `ContextSelector.tsx` as well?

nicolethoen (changes requested):

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
adithyaakrishna: Oops! My bad, I had missed updating it in the `ContextSelector.tsx`, thanks for pointing it out @nicolethoen :)

adithyaakrishna: Updated the description with the suggestion as well 🚀

nicolethoen (approved):

nicolethoen (changes requested):
Actually, looks like the linter isn't happy.

adithyaakrishna:
@nicolethoen This is strange, husky missed this commit while I was updating the PR 🤔 

Fixed it in 774b5c7

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -32,7 +34,7 @@ export class ContextSelectorToggle extends React.Component<ContextSelectorToggle
   static displayName = 'ContextSelectorToggle';
   static defaultProps: PickOptional<ContextSelectorToggleProps> = {
     className: '',
-    toggleText: '',
+    toggleText: null as React.ReactNode,
```
adithyaakrishna, line 37: Also @nicolethoen I don't think we will need this change, however would love to know your opinion on this 👀

nicolethoen: I agree that it's likely unnecessary - an empty string is a valid value for a React.ReactNode

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -37,8 +37,10 @@ export interface ContextSelectorProps extends Omit<ToggleMenuBaseProps, 'menuApp
   isFullHeight?: boolean;
   /** Labels the Context Selector for Screen Readers */
   screenReaderLabel?: string;
-  /** Text that appears in the Context Selector Toggle */
-  toggleText?: string;
+  /** Component or string that will be used in the Context Selector.
+   * This prop is placed in a button, and cannot contain any interactive elements
+   */
+  toggleText?: React.ReactNode | string;
```
tlabaj, line 43: You do not need the "string" here.  ReactNode includes string type.

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -75,7 +77,7 @@ export class ContextSelector extends React.Component<ContextSelectorProps, { oui
     onToggle: () => undefined as any,
     onSelect: () => undefined as any,
     screenReaderLabel: '',
-    toggleText: '',
+    toggleText: null as React.ReactNode,
```
tlabaj, line 80: I  would keep the empty string default here.

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -10,8 +10,10 @@ export interface ContextSelectorToggleProps {
   id: string;
   /** Classes applied to root element of toggle */
   className?: string;
-  /** Text that appears in the Context Selector Toggle */
-  toggleText?: string;
+  /** Component or string that will be used in the Context Selector.
+   * This prop is placed in a button, and cannot contain any interactive elements
+   */
+  toggleText?: React.ReactNode | string;
```
tlabaj, line 16: Sam comment as above.  ReactNode includes string type.

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
tlabaj, line 37: You can keep the empty string default here.

tlabaj (changes requested):

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
adithyaakrishna: Oki Doki, updated the PR with the changes 🚀

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
adithyaakrishna: Gotcha, I have updated the PR with the changes 👀

thatblindgeye (approved):

nicolethoen (approved):

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
```diff
@@ -10,8 +10,10 @@ export interface ContextSelectorToggleProps {
   id: string;
   /** Classes applied to root element of toggle */
   className?: string;
-  /** Text that appears in the Context Selector Toggle */
-  toggleText?: string;
+  /** Component or string that will be used in the Context Selector.
```
tlabaj, line 13: Can this be sentence case please.  "Component or string that will be used in the context selector."

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
```diff
@@ -37,8 +37,10 @@ export interface ContextSelectorProps extends Omit<ToggleMenuBaseProps, 'menuApp
   isFullHeight?: boolean;
   /** Labels the Context Selector for Screen Readers */
   screenReaderLabel?: string;
-  /** Text that appears in the Context Selector Toggle */
-  toggleText?: string;
+  /** Component or string that will be used in the Context Selector.
```
tlabaj, line 40: Can this be sentence case please. "Component or string that will be used in the context selector."

tlabaj (changes requested):

packages/react-core/src/components/ContextSelector/ContextSelectorToggle.tsx
adithyaakrishna: Yes, I have updated it in the latest commit below :)

packages/react-core/src/components/ContextSelector/ContextSelector.tsx
adithyaakrishna: Done ✅

tlabaj (approved):

adithyaakrishna:
Yayy, thanks @nicolethoen @tlabaj and @thatblindgeye 🚀

Issue #7657, opened 2022-07-06
Primary Detail - align demo with core


jenny-s51:
**Is this a new component or an extension of an existing one?**
What is the existing component, if any?
While working on #7312 , I noticed there were many visual differences between the core demo and the current implementation of the react demo. 

**Describe the feature**
Missing search input in toolbar, and several buttons need to be aligned to match the core demo. 

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
Include screenshots or links to Marvel or other mockups.
https://pf4.patternfly.org/demos/primary-detail/html-demos/primary-detail-card-view-expanded/
https://www.patternfly.org/v4/demos/primary-detail/react-demos/primary-detail-card-view/

**Any other information?**

PR #7658, opened 2022-07-06
Primary Detail: use new raised card variant in demo, add bulk select

### LINKED ISSUE #7312: Primary-detail demo - update to use new selectable card
Want to update this demo to utilize the new selectable card styling: https://www.patternfly.org/v4/demos/primary-detail/react-demos/primary-detail-card-view-expanded/

Core PR: https://github.com/patternfly/patternfly/pull/4810

jenny-s51:
**What**: Closes #7312 

Added some work here that applies towards #7657 - this demo now supports bulk select, and the ability to delete cards. 

However in order to fully align this demo with core, a lot of additional logic/functionality is required, which is out of scope of the original issue to use the new raised variant for the card. For that reason I opened #7657

**Additional issues**: #7657

### DIFF
--- a/packages/react-core/src/demos/PrimaryDetail.md
+++ b/packages/react-core/src/demos/PrimaryDetail.md
@@ -1057,14 +1057,6 @@ class PrimaryDetailCardView extends React.Component {
       });
     };
 
-    this.deleteItem = item => _event => {
-      const filter = getter => val => getter(val) !== item.id;
-      this.setState({
-        res: this.state.res.filter(filter(({ id }) => id)),
-        selectedItems: this.state.selectedItems.filter(filter(id => id))
-      });
-    };
-
     this.checkAllSelected = (selected, total) => {
       if (selected && selected < total) {
         return null;
@@ -1266,9 +1258,11 @@ class PrimaryDetailCardView extends React.Component {
 
     this.deleteItem = item => _event => {
       const filter = getter => val => getter(val) !== item.id;
+      const filteredCards = this.state.res.filter(filter(({ id }) => id));
       this.setState({
-        res: this.state.res.filter(filter(({ id }) => id)),
-        selectedItems: this.state.selectedItems.filter(filter(id => id))
+        res: filteredCards,
+        selectedItems: this.state.selectedItems.filter(filter(id => id)),
+        totalItemCount: this.state.totalItemCount - 1
       });
     };
   }
@@ -1340,7 +1334,7 @@ class PrimaryDetailCardView extends React.Component {
         Select none (0 items)
       </DropdownItem>,
       <DropdownItem key="item-2" onClick={this.selectPage}>
-        Select page ({this.state.perPage} items)
+        Select page ({this.state.res.length} items)
       </DropdownItem>,
       <DropdownItem key="item-3" onClick={this.selectAll}>
         Select all ({this.state.totalItemCount} items)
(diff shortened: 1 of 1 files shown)


packages/react-core/src/demos/PrimaryDetail.md
```diff
@@ -1140,6 +1170,107 @@ class PrimaryDetailCardView extends React.Component {
...
+      });
+    };
+
+    this.onCardKebabDropdownSelect = (key, _event) => {
+      this.setState({
+        [key]: !this.state[key]
+      });
+    };
+
+    this.deleteItem = item => _event => {
```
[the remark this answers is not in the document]
jenny-s51: Nice catch @nicolethoen - thank you! Updated.

nicolethoen (approved):
LGTM!

nicolethoen:
@mmenestr this is the direct URL to the updated demo: https://patternfly-react-pr-7658.surge.sh/demos/primary-detail/react-demos/primary-detail-card-view/

thatblindgeye (commented):
Overall looks good! A couple of things I'm noticing (which these may be better as followup issues) are:

- When clicking anywhere in the card, the drawer gets toggled open. This includes clicking the checkbox or kebab toggle. The expectation I'd have is clicking either of those would only trigger those specific elements, and clicking anywhere else in the card would select it.
- This I can create a spinoff issue for since it might require tweaking the logic of selectable cards a bit, but should we have the card drawers closable in ways other than the "close" button? When navigating via keyboard, once I click a card to open the drawer it remains open and blocks other cards completely, and you'd have to tab through every single card to get to the close button.

@jenny-s51 what do you think?

jenny-s51 (commented):
Thank you for your awesome feedback @thatblindgeye !! 

> * When clicking anywhere in the card, the drawer gets toggled open. This includes clicking the checkbox or kebab toggle. The expectation I'd have is clicking either of those would only trigger those specific elements, and clicking anywhere else in the card would select it.

Definitely makes sense to have the checkbox, drawer toggle, and kebab toggles all behave independently! I've updated the code to separate all the different toggling behaviors you listed here. 

> * This I can create a spinoff issue for since it might require tweaking the logic of selectable cards a bit, but should we have the card drawers closable in ways other than the "close" button? When navigating via keyboard, once I click a card to open the drawer it remains open and blocks other cards completely, and you'd have to tab through every single card to get to the close button.

Nice catch!! I fixed this by making sure the drawer closes if a user selects the card a second time either via mouse or keyboard. Seems like a good, working solution for this! WDYT @thatblindgeye ?

packages/react-core/src/demos/PrimaryDetail.md
```diff
@@ -1095,22 +1101,26 @@ class PrimaryDetailCardView extends React.Component {
...
+      );
     };
 
     this.onCardClick = event => {
       if (event.currentTarget.id === this.state.activeCard) {
+        this.setState({
+          isDrawerExpanded: !this.state.isDrawerExpanded
+        });
         return;
       }
```
thatblindgeye, line 1125: Updating the `activeCard` state to `null` when the drawer gets closed when clicking a selected card again, to remove the card styling/uncheck the hidden radio button.

packages/react-core/src/demos/PrimaryDetail.md
```diff
@@ -1140,6 +1150,132 @@ class PrimaryDetailCardView extends React.Component {
...
+        return;
+      }
+
+      if ([13, 32].includes(event.keyCode)) {
+        if (event.currentTarget.id === this.state.activeCard) {
+          this.setState({
+            isDrawerExpanded: !this.state.isDrawerExpanded
+          });
+          return;
+        }
```
thatblindgeye, line 1270: Same thing here

thatblindgeye (changes requested):
The latest updates work great @jenny-s51! Just had one minor addition noted below based on these updates.

jenny-s51 (commented):
Thank you @thatblindgeye ! Updated 👍

thatblindgeye (approved):
🎉 Awesome job on this demo!

PR #7669, opened 2022-07-07
chore(ActionList): update tests to new standards

### LINKED ISSUE #7665: Action list test revamp

wise-king-sullyman:
**What**: Closes #7665 

**Additional issues**:

### DIFF
--- a/packages/react-core/src/components/ActionList/__tests__/ActionList.test.tsx
+++ b/packages/react-core/src/components/ActionList/__tests__/ActionList.test.tsx
@@ -1,14 +1,49 @@
-import * as React from 'react';
-import { render } from '@testing-library/react';
+import React from 'react';
+import { render, screen } from '@testing-library/react';
 import { ActionList } from '../ActionList';
 
-describe('action list', () => {
-  test('renders successfully', () => {
-    const { asFragment } = render(<ActionList>test</ActionList>);
-    expect(asFragment()).toMatchSnapshot();
-  });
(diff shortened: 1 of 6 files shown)


packages/react-core/src/components/ActionList/__tests__/ActionList.test.tsx
```diff
@@ -1,14 +1,49 @@
...
+test('Renders children', () => {
+  render(<ActionList>Test</ActionList>);
+
+  expect(screen.getByText('Test')).toBeVisible();
+});
+
+test('Renders with class pf-c-action-list', () => {
+  render(<ActionList>Test</ActionList>);
+
+  expect(screen.getByText('Test')).toHaveClass('pf-c-action-list');
```
nicolethoen, line 19: Would it be reasonable to combine this test with the last since they are rendering the same thing?

nicolethoen: and there are a couple other below that also render the same thing?

wise-king-sullyman: I definitely wouldn't call it *unreasonable* to combine them, but I personally would advocate for a testing approach that only makes assertions about one aspect of component behavior in each test.

The primary reason being that I believe single behavior tests provide maximum utility from the test suite, as with this approach a failing test (or combination of them) can often point you straight to where a regression was introduced.

nicolethoen (approved):

PR #7679, opened 2022-07-11
fix(CalendarMonth) fix first day of month not showing when weekstart …

### LINKED ISSUE #7654: Bug - Datepicker - impossible select some days
If you try to set Monday as the first week day instead of Sunday it becomes impossible to select some days.

### How do you reproduce the problem?

- Set `weekStart={Weekday.Monday}`
- Select 2023 January 

### Expected behavior
We can select 1th January 2023 here

### Actual behavior
As you can see on the screenshot, it is impossible to select 1th January 2023 because it is not present in the datepicker.
You need to select December 2022 to have possibility to select it. 

For most users, this is highly non-obvious behavior. Will be nice if the last week of the previous month will be visible if at least 1 day of selected month is situated there. An example of a possible implementation can be seen in the Bootstrap CSS, PrimeReact or Mac OS / Windows calendars

Screenshot_550

Example of the possible improvement:

Screenshot_551

Dominik-Petrik:
…set to monday

**What**: Closes #7654


thatblindgeye (commented):
Nice work on this! I did notice one issue that I noted below, just let me know what you think! 🙂

nicolethoen:
@thatblindgeye and @mmenestr might have some insight into when we may or may not want 6 weeks displayed. I think they recently made some decisions regarding this topic together.

mmenestr:
Yes, we basically said that it should preferably vary based on whether the 6th week has any numbers from the current month. In other words, if the month needs 6 weeks to show up fully, then the calendar should show 6 weeks, but if the month only needs 5 weeks to show up fully (which would mean that the 6th week would have no dates from the month you're currently on, and instead only show calendar days for the following month) then only 5 weeks should show.

I just thought it was weird that in a lot of cases, if we were to always show 6 weeks, you'd have a full line for dates that were not part of the month you were on.

Hopefully that makes sense!

Dominik-Petrik:
I fixed the last row not showing in specific cases, now it works as requested, so number of rows displayed varies based on whether the 6th week contains days from same month as selected. @wise-king-sullyman also shared some useful resources on testing with me, so I should be able to create tests to prevent this from regressing.

mmenestr (approved):

nicolethoen (approved):
LGTM!

thatblindgeye (approved):

wise-king-sullyman (changes requested):

packages/react-core/src/components/CalendarMonth/__tests__/CalendarMonth.test.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import { render, screen } from '@testing-library/react';
+
+import { CalendarMonth } from '../CalendarMonth';
+
+test('Renders the first date in a month when a custom weekStart is passed', () => {
+  const formatAria = (date: Date) =>
```
wise-king-sullyman, line 8: Can you add a comment here just stating that this custom aria label is needed because of a bug with automatic aria label generation as it is now, and that it can be removed once that bug is fixed?

wise-king-sullyman (changes requested):
Two super small nits and then this will be good to merge!

packages/react-core/src/components/CalendarMonth/__tests__/CalendarMonth.test.tsx
wise-king-sullyman: That's a fair point, though the same could be said for this approach and its reliance on the `cellAriaLabel` prop functioning properly. This line of thinking could also extend to many other behaviors of the component, such as the `weekStart` or `date` props working properly, or the structure of the component.

I totally get not wanting this test to be able to fail for something that isn't the actual behavior under test, but I think that it's generally reasonable for a test to rely on other behavior of the component under test since totally avoiding that would be next to impossible.

Plus if the generated aria labels are properly tested developers should be able to see that test failing as well if the generated aria label is changed, which should make it apparent what the true cause of the failure here is.

All of that being said I don't think what you've got here is bad by any means and I don't think it neccesarily needs to go, but I would still like a comment saying *why* it's here and that it shouldn't be *needed* once the bug is fixed 🙂

wise-king-sullyman (approved):
🚀 This is really awesome work!

PR #7681, opened 2022-07-11
fix(LogViewer): render correctly when there is '\n' in the string array

### LINKED ISSUE #7673: Bug - LogViewer - Log item height is miss calculated when isTextWrapped is true
**Describe the problem**
When `isTextWrapped` is `true` and the string provided to `data` prop has `\n` in it the generated lines have miss calculated height. The calculated height is less than enough to display the whole content causing the items to overlap each other.

https://codesandbox.io/s/testing-patternfly-item-height-problem-0uhc01?file=/index.js

**Expected behavior**
Expect each line to have its height calculated properly

Yes

**Screenshots**
image

**What is your environment?**
 - OS: Ubuntu 20.04.4 LTS
 - Browser chrome
 - Version 103.0.5060.53 (Official Build) (64-bit)

DaoDaoNoCode:
**What**: Closes #7673 
The log viewer only processed the string data before and split them if there are `\n` in it. Now it also checks the string array data and splits them if needed.

### DIFF
--- a/packages/react-log-viewer/src/LogViewer/LogViewer.tsx
+++ b/packages/react-log-viewer/src/LogViewer/LogViewer.tsx
@@ -143,7 +143,7 @@ const LogViewerBase: React.FunctionComponent<LogViewerProps> = memo(
 
     /* Parse data every time it changes */
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


nicolethoen (approved):
LGTM! Thanks for the quick fix!

PR #7684, opened 2022-07-12
docs: multiple typeahead select menu demo

kmcfaul:
Typeahead input toggle is built on TextInputGroup.
Keyboard interaction is a bit different than Select, and MenuItem needs a focus css class (to visually display focus while maintaining real focus on the input) if we want to better match it. Currently ArrowDown from the input will navigate into the menu items, and then the Menu keyboard interaction takes over. To get back to the input, Escape needs to be pressed (or shift-tab back to the input). 

Closes: #7700

### DIFF
--- a/packages/react-core/src/demos/ComposableMenu/examples/ComposableMultipleTypeaheadSelect.tsx
+++ b/packages/react-core/src/demos/ComposableMenu/examples/ComposableMultipleTypeaheadSelect.tsx
@@ -144,19 +144,6 @@ export const ComposableMultipleTypeaheadSelect: React.FunctionComponent = () => 
       isFullWidth
     >
       <TextInputGroup isPlain>
-        <ChipGroup>
-          {selected.map((selection, index) => (
-            <Chip
-              key={index}
-              onClick={ev => {
-                ev.stopPropagation();
-                onMenuSelect(selection);
-              }}
-            >
-              {selection}
-            </Chip>
-          ))}
-        </ChipGroup>
         <TextInputGroupMain
           value={inputValue}
           onClick={toggleMenuOpen}
@@ -165,7 +152,21 @@ export const ComposableMultipleTypeaheadSelect: React.FunctionComponent = () => 
           id="typeahead-select-input"
           autoComplete="off"
           innerRef={textInputRef}
-        />
+        >
+          <ChipGroup>
+            {selected.map((selection, index) => (
+              <Chip
+                key={index}
+                onClick={ev => {
+                  ev.stopPropagation();
+                  onMenuSelect(selection);
+                }}
+              >
+                {selection}
+              </Chip>
+            ))}
+          </ChipGroup>
+        </TextInputGroupMain>
         <TextInputGroupUtilities>
           {selected.length > 0 && (
             <Button


gitdallas (approved):

nicolethoen (changes requested):
I think we need a 'no results' message in the empty menu when all options have been filtered out.

Also, I think that Jeff's open composable typeahead select demo is handling keyboard focus the way you describe to better match the existing select implementations. Maybe his solution would be helpful here, too?

kmcfaul:
Yes it does, I'll update this PR to match

mcarrano (changes requested):
I did a quick test and their seem to be some issues. Clear does not work as expected. Should clear all chips. Also seems like if I type into the field and then select a value from the list with my mouse, the input does not clear (still retains what I typed)

packages/react-core/src/demos/ComposableMenu/examples/ComposableMultipleTypeaheadSelect.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const [isOpen, setIsOpen] = React.useState<boolean>(false);
+  const [selected, setSelected] = React.useState<string[]>([]);
+  const [inputValue, setInputValue] = React.useState<string>('');
+  const toggleRef = React.useRef<HTMLDivElement>();
+  const inputRef = React.useRef<HTMLInputElement>();
+  const menuRef = React.useRef<HTMLDivElement>();
+  const containerRef = React.useRef<HTMLDivElement>();
+
+  const options = [
+    <MenuItem key={0} itemId="Option 1">
```
tlabaj, line 28: Can we update the data a little her so we can get varying results when we start typing?  Right now you only get matches on all items or one.

tlabaj (changes requested):
The side nav links do not seem to take me to the rights examples.

packages/react-core/src/demos/ComposableMenu/examples/ComposableMultipleTypeaheadSelect.tsx
```diff
@@ -0,0 +1,219 @@
...
+    if (itemId) {
+      setSelected(
+        selected.includes(itemId) ? selected.filter(selection => selection !== itemId) : [...selected, itemId]
+      );
+    }
+  };
+
+  const toggle = (
+    <MenuToggle
+      variant="typeahead"
```
mcoker, line 140: This doesn't appear to be working - maybe you just need to rebase? Seems to work fine on main.

The typeahead looks like this with extra padding around the text input
Screen Shot 2022-09-01 at 8 25 02 PM

It should look like this - 36px tall (matches all the other form stuff and buttons)

Screen Shot 2022-09-01 at 8 25 15 PM
Screen Shot 2022-09-01 at 8 28 00 PM

mcoker (changes requested):
Structure looks good to me, though I'm seeing some style issue. I'm wondering if a rebase will fix them.

packages/react-core/src/demos/ComposableMenu/examples/ComposableMultipleTypeaheadSelect.tsx
tlabaj: @mcoker yes a rebase should fix that.  Jeff fixed the styling issue in this PR #7905

nicolethoen:
i think this also needs a rebase to take care of they syntax error you fixed in the other composable menu examples?

nicolethoen (approved):

jessiehuff (commented):
It looks pretty good! 🙂 My only concern is it doesn't seem to announce the options in VO. Looking at the Select version, we might have used `aria-activedescendant`, but I'm not sure if there were additional things we added to make this work for Select. I'm ok with us doing this in a separate issue if we want to get this through though.

packages/react-core/src/demos/ComposableMenu/examples/ComposableMultipleTypeaheadSelect.tsx
```diff
@@ -0,0 +1,219 @@
...
+        </ChipGroup>
+        <TextInputGroupMain
+          value={inputValue}
+          onClick={toggleMenuOpen}
+          onChange={onTextInputChange}
+          onKeyDown={onInputKeyDown}
+          id="typeahead-select-input"
+          autoComplete="off"
+          innerRef={textInputRef}
+        />
```
mcoker, line 168: Getting some left-edge-no-space action with the chip group:

Screen Shot 2022-09-07 at 6 13 40 PM

From the [text input group filters example](https://www.patternfly.org/v4/components/text-input-group#filters), looks like `<ChipGroup />` needs to be a child of `<TextInputGroupMain>`

```suggestion
        <TextInputGroupMain
          value={inputValue}
          onClick={toggleMenuOpen}
          onChange={onTextInputChange}
          onKeyDown={onInputKeyDown}
          id="typeahead-select-input"
          autoComplete="off"
          innerRef={textInputRef}
        >
          <ChipGroup>
            {selected.map((selection, index) => (
              <Chip
                key={index}
                onClick={ev => {
                  ev.stopPropagation();
                  onMenuSelect(selection);
                }}
              >
                {selection}
              </Chip>
            ))}
          </ChipGroup>
        </TextInputGroupMain>
```

mcoker (changes requested):
LGTM! Just one thing about the structure that should fix a small spacing issue.

mcoker (approved):
🥸👍

tlabaj (commented):
I know this is out of the scope of the PR, but the other typeahead example is not marked beta.  Can you mark that one beta too.

tlabaj (approved):
LGTM

mcarrano (approved):

kmcfaul:
@jessiehuff Opened #8125 to follow up for the VO issues

@tlabaj Are we good to try to get this merged this sprint?

jessiehuff (approved):
So I'm not quite sure what has changed, but it seems like the options are getting announced for me now. Or perhaps I was using a different screen reader than VO or I saw something that I'm not seeing right now. I did notice that I wasn't able to escape/close the menu when going through the options though. So perhaps we can keep that issue open and investigate that interaction.

PR #7686, opened 2022-07-12
chore(Truncate): component unit test revamp

### LINKED ISSUE #7643: Truncate

andyyvo:
**What**: Closes #7643 . Unit tests revamp -- note that there was an issue with the `position` prop where if `position=start`, then only `pf-c-truncate__end` would render. However, `getByText` could not detect the inputted text as a result of `&lrm;` so a snapshot test was performed instead.

**Additional issues**: N/A.

### DIFF
--- a/packages/react-core/src/components/Truncate/__tests__/Truncate.test.tsx
+++ b/packages/react-core/src/components/Truncate/__tests__/Truncate.test.tsx
@@ -5,6 +5,7 @@ import { Truncate } from '../Truncate';
 jest.mock('../../Tooltip', () => ({
   Tooltip: ({ content, position, children, ...props }) => (
     <div data-testid="Tooltip-mock" {...props}>
+      <div data-testid="Tooltip-mock-content-container">Test {content}</div>
       <p>{`position: ${position}`}</p>
       {children}
     </div>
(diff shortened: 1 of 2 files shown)


packages/react-core/src/components/Truncate/__tests__/Truncate.test.tsx
```diff
@@ -28,3 +130,26 @@ test('renders middle truncation', () => {
   );
   expect(asFragment()).toMatchSnapshot();
```
wise-king-sullyman, line 131: Can you refactor this test to not be a snapshot and explicitly assert against the default behavior when position=middle and traillingNumChars isn't passed?

packages/react-core/src/components/Truncate/__tests__/Truncate.test.tsx
```diff
@@ -1,15 +1,117 @@
...
+    <Truncate
+      content={'Testing truncate content'}
+      position='end'
+    />
+  );
+
+  const start = screen.getByText('Testing truncate content');
+
+  expect(start).toHaveClass('pf-c-truncate__start');
+});
```
wise-king-sullyman, line 105: Since position=end is the default value I think this test should actually omit the position prop so that we have a test about the default behavior here.

wise-king-sullyman (changes requested):
This is looking really good! I only have a few changes I'd like to see 🙂

wise-king-sullyman (approved):
This looks great! Thanks for taking the time to make all these changes and put this together!

